PR #7378, opened 2020-03-04
MRG, FIX: Fix fetching bugs

larsoner:
In maybe 10% of PRs we get sporadic failures downloading the test data or in `test_constants.py` [of the form](https://travis-ci.org/mne-tools/mne-python/jobs/657984986):
```
RuntimeError: URL could not be parsed properly (total size 1030996061 != file size 1)
```
I could replicate it locally by running `pytest mne/io/tests/test_constants.py` a lot (like 10-20 times) -- I'd see the failure every once in a while.

This has to do with our not handling `chunked` data protocol properly. The chunked protocol causes the `headers['Content-Length']` to be None/missing, which we currently treat as a file size of `1` but then later don't handle properly. We can and should still read data even if we don't know the total size rather than fail out. This PR simplifies a lot of the logic in our downloader in order hopefully to help. It also reduces the number of times that we open URLs to get information from them (hopefully it's only once now).

I simplified our code for following redirects, under the assumption that modern Python will not require us to go through and manually follow multiple redirects. In testing it seems okay and `nilearn` (where we originally adapted our code) does not even try to do this, so I think it'll be okay. I'm running with `[circle full]` just to test all of our downloads, though.

### DIFF
--- a/mne/utils/fetching.py
+++ b/mne/utils/fetching.py
@@ -9,53 +9,49 @@ import shutil
 import sys
 import time
 from urllib import parse, request
+from urllib.error import HTTPError, URLError
 
 from .progressbar import ProgressBar
 from .numerics import hashfunc
 from .misc import sizeof_fmt
-from ._logging import warn, logger, verbose
+from ._logging import logger, verbose
 
 
 # Adapted from nilearn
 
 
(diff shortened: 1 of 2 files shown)


agramfort:
provided Circle CI failure is unrelated LGTM. MErging. thx @larsoner

Issue #7162, opened 2019-12-28
ENH: New interface for Brain


GuillaumeFavelier:
⚠️ **Requires `pyvista>=0.24`** ⚠️

This PR summarizes the ideas to implement in `Brain`, previously called `_TimeViewer` (suggested in https://github.com/mne-tools/mne-python/pull/7153#issuecomment-569378110):

<details>
<summary><b>Milestone 0.20</b></summary>

- [x] Get it working for VectorSourceEstimates (added in #7060)
- [x] Improve the [color cycle of point picking to allow point color re-use](https://github.com/mne-tools/mne-python/pull/7247#issuecomment-580763125) (#7516)
- [x] Check that the latest version 0.24 of pyvista is used (added in https://github.com/mne-tools/mne-python/pull/7506)
- [x] Put `time_viewer='auto'` and `show_traces='auto'` as the default for `stc.plot` / `plot_source_estimates` (added in https://github.com/mne-tools/mne-python/pull/7380)
- [x] Use the latest version 0.24 of pyvista (added in https://github.com/mne-tools/mne-python/pull/7458)
- [x] Use interpolation `nearest` as the default (suggested in https://github.com/mne-tools/mne-python/pull/7247#issuecomment-589447242, added in https://github.com/mne-tools/mne-python/pull/7257, reverted in 95ad0eec8a2e1e9c7c74e25852157e650aea7df0, added in https://github.com/mne-tools/mne-python/pull/7469)
- [x] Improve performance of interpolation (added in https://github.com/mne-tools/mne-python/pull/7257, reverted in 95ad0eec8a2e1e9c7c74e25852157e650aea7df0, added in https://github.com/mne-tools/mne-python/pull/7469)
- [x] Display current time with a line in `MplCanvas` (suggested in https://github.com/mne-tools/mne-python/pull/7247#issuecomment-582899553, added in https://github.com/mne-tools/mne-python/pull/7423)
- [x] Link the time plot of `MplCanvas` with the time slider (suggested in https://github.com/mne-tools/mne-python/pull/7247#issuecomment-582899553, added in https://github.com/mne-tools/mne-python/pull/7423)
- [x] Add a way to clear all the added points in `show_traces=True` mode (added in #7386)
</details>

<details>
<summary><b>Milestone 0.21</b></summary>

- [x] Set PyVista as the default 3d backend (done in https://github.com/mne-tools/mne-python/pull/8220)
[...]

PR #7257, opened 2020-01-27
MRG: Add toggle to save_movie in _TimeViewer

GuillaumeFavelier:
This PR adds a toggle to save the scene displayed in `_TimeViewer` as a [GIF using PyVista movie feature](https://docs.pyvista.org/examples/02-plot/gif.html?highlight=gif). The shortcut for the toggle is `o`.

**TL:DR**: for implementation specific reason, I would push more for offline `save_movie()` instead, similar to `PySurfer`. I would like to have your opinion on this @larsoner 

**Note/Known issue**

From my experience of this feature, the interface stutters using the default `GIF-PIL` (Pillow) format for imageio `append_data()`. This is probably because [disk access is performed at each call](https://github.com/imageio/imageio/blob/master/imageio/plugins/pillowmulti.py#L108-L116). I had better success with `GIL-FI` (FreeImage) format: the interface was still responsive during recording and suffers for lag only while saving to disk (end of recording).

For reference:
* [Pillow `append_data()`](https://github.com/imageio/imageio/blob/master/imageio/plugins/pillowmulti.py#L108-L116)
* [FreeImage `append_data()`](https://github.com/imageio/imageio/blob/master/imageio/plugins/freeimage.py#L111-L132)

It's an item of https://github.com/mne-tools/mne-python/issues/7162

PR #7380, opened 2020-03-04
WIP: Set time_viewer and show_traces to auto

GuillaumeFavelier:
This PR sets `time_viewer='auto'` and `show_traces='auto'`. It's still a work in progress.

It's an item of #7162


GuillaumeFavelier:
To achieve `time_viewer=auto`, the following 3d viz tests need to be ported to the `renderer_interactive` fixture:
* test_plot_sparse_source_estimates
* test_process_clim_plot

GuillaumeFavelier:
By default the 3d tests are done with `off_screen=True` to avoid opening a window but in order to use the `PyVista` widgets, `off_screen` should be disabled (to obtain an interactive window).
The goal of the `renderer_interactive` fixture is just that.

GuillaumeFavelier:
The latest commit is marked as test because the `renderer_interactive` fixture does not include `mayavi` for now. It needs to be fixed.

GuillaumeFavelier:
The `interactive` parameter of `_testing_context()` is unused in the `_pysurfer_mayavi` 3d backend:

https://github.com/mne-tools/mne-python/blob/499f917dd301826606f706e775c611c3e215d1b0/mne/viz/backends/_pysurfer_mayavi.py#L464-L472

compared to `_pyvista`:

https://github.com/mne-tools/mne-python/blob/499f917dd301826606f706e775c611c3e215d1b0/mne/viz/backends/_pyvista.py#L647-L661

So I think it's totally safe to add it to the list:

https://github.com/mne-tools/mne-python/blob/499f917dd301826606f706e775c611c3e215d1b0/mne/conftest.py#L238-L243

GuillaumeFavelier:
The tests using `renderer_interactive` are:

```
test_brain_timeviewer
test_brain_timeviewer_traces
test_brain_linkviewer
test_plot_sparse_source_estimates
test_process_clim_plot
test_link_brains
```

Among those, only:

```
test_plot_sparse_source_estimates
test_process_clim_plot
```

are supported by `mayavi`, the rest should raise `NotImplementedError`.

GuillaumeFavelier:
I forgot to update Circle config yesterday... The issue with pytest-sugar is back

GuillaumeFavelier:
Oh no. It's actually another release of pytest:

image

GuillaumeFavelier:
Can we just do `pytest<5.4.0` for now?

GuillaumeFavelier:
I notice a few warnings of the type:

```
UserWarning: Attempting to set identical left == right == 0 results in singular transformations; automatically expanding.
```

Also, there is an issue with `time_label`:

```
AttributeError: 'TimeSlider' object has no attribute 'time_label'
```

I'll push a fix.

GuillaumeFavelier:
Waiting for https://github.com/mne-tools/mne-python/pull/7257 before merging conflicts.

GuillaumeFavelier:
The current issue is a timeout of Circle during doc building a bit after the visualization folder.

larsoner:
Merge conflict

GuillaumeFavelier:
Since #7458 is merged, the workaround scraper needs to be removed. I'll fix this

mne/viz/tests/test_3d.py
```diff
@@ -137,11 +137,8 @@ def test_plot_sparse_source_estimates(renderer):
     stc_data[1, 4] = 2.
     vertices = [vertices[inds], np.empty(0, dtype=np.int)]
     stc = SourceEstimate(stc_data, vertices, 1, 1)
-    surf = plot_sparse_source_estimates(sample_src, stc, bgcolor=(1, 1, 1),
-                                        opacity=0.5, high_resolution=False)
-    if renderer.get_3d_backend() == 'mayavi':
-        import mayavi  # noqa: F401 analysis:ignore
-        assert isinstance(surf, mayavi.modules.surface.Surface)
```
larsoner: Shouldn't we keep some variant of this? Can you instead change it to `renderer_interactive.get_3d_backend()`?

GuillaumeFavelier: Indeed that was a bit too agressive, I'll add a branch for mayavi

GuillaumeFavelier:
I'll try `[circle full]` once https://github.com/mne-tools/mne-python/pull/7480 is merged

larsoner:
Merged @GuillaumeFavelier

GuillaumeFavelier:
#7480 was not enough apparently since we got a timeout too, I'm trying with `show_traces=False` and at the same time see if I can improve the memory footprint of `_Brain`

GuillaumeFavelier:
So far my changes seem to help freeing memory but nothing big for `_Brain` (`master` in black, `local changes` in blue)

image

GuillaumeFavelier:
@agramfort the timeout seems related to `show_traces=True` because now CircleCI passes:

https://18788-1301584-gh.circle-artifacts.com/0/dev/index.html

Is it good enough to have only `time_viewer='auto'`?

agramfort:
yes ok. It can be improved later after the release.

>

GuillaumeFavelier:
This is ready for reviews then

larsoner:
> > Is it good enough to have only time_viewer='auto'?

> yes ok. It can be improved later after the release.

I'm not so sure because this would be a change in behavior with no deprecation cycle.

A lesser evil would probably be to add a private `_MNE_TRACES_AUTO=false` env var that we set in CircleCI and detect in `_brain` to make it so `'auto'` means False. This would make the output different from what a user would see when they ran the script with the PyVista backend and more like what it would be like with the Mayavi backend -- but I doubt anyone would really notice.

GuillaumeFavelier:
> A lesser evil would probably be to add a private _MNE_TRACES_AUTO=false env var

I guess it's possible indeed, what do you think @agramfort ?

Why private though @larsoner ?

larsoner:
> Why private though @larsoner ?

Hopefully eventually we can fix the memory/timeout problem and kill it entirely. We don't want it actually to be a permanent solution (or give users yet another way to set things, they should use `kwargs`)

GuillaumeFavelier:
> Hopefully eventually we can fix the memory/timeout problem and kill it entirely

I'll continue to investigate locally

larsoner:
Yes please do, but in the meantime the private env var workaround would make it so that we don't have to figure it out before release

larsoner:
Pushed a commit with the CircleCI workaround and setting both `show_traces` and `time_viewer` to `'auto'` as default, let's see if it gets us around the CI problem

GuillaumeFavelier:
Thank you @larsoner

agramfort:
if @drammock and @hoechenberger are happy I am happy. Happy @drammock and @hoechenberger ?

hoechenberger:
LGTM!

PR #7382, opened 2020-03-04
Fix window name in _TimeViewer

GuillaumeFavelier:
This PR fixes the window title of `_TimeViewer`. The diff is very short, the `title` parameter was not passed to `_get_renderer()`.

It's an item of #7162

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -202,7 +202,8 @@ class _Brain(object):
             _check_3d_figure(figure)
         self._renderer = _get_renderer(size=fig_size, bgcolor=background,
                                        shape=(n_row, n_col), fig=figure)
-        _set_3d_title(figure=self._renderer.figure, title=self._title)
+        if self._title is not None:
+            _set_3d_title(figure=self._renderer.figure, title=self._title)
 
         for h in self._hemis:
             # Initialize a Surface object as the geometry
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -199,7 +199,8 @@ class _Brain(object):
 
         if figure is not None and not isinstance(figure, int):
             _check_3d_figure(figure)
-        self._renderer = _get_renderer(size=fig_size, bgcolor=background,
+        self._renderer = _get_renderer(name=title, size=fig_size,
+                                       bgcolor=background,
                                        shape=(n_row, n_col), fig=figure)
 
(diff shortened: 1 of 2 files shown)


larsoner:
Can you check to make sure that however PyVista handles the `name` here, it does *not* make it that passing the same `name`/`title` twice leads to the same figure handle? See for example https://github.com/nipy/PySurfer/commit/5aaf0430e43c700dc358ad946e193f78d1137ad4 where if we did `mlab.figure(title, ...)` twice it would point to the same figure handle because of how `mlab.figure` works, similar to how doing `fig1 = plt.figure(1); fig2 = plt.figure(1)` will have `fig1` and `fig2` point to the same handle. This can be desirable in some situations but it's not good for our purposes in how we use the 3D plotters.

mne/viz/_brain/_brain.py
```diff
@@ -199,7 +199,8 @@ def __init__(self, subject_id, hemi, surf, title=None,
 
         if figure is not None and not isinstance(figure, int):
             _check_3d_figure(figure)
-        self._renderer = _get_renderer(size=fig_size, bgcolor=background,
+        self._renderer = _get_renderer(name=title, size=fig_size,
```
agramfort: `name=title` raises some skepticism but I guess it's internal code so...

GuillaumeFavelier:
`name` is a renderer mechanic, I implemented it in `mne/viz/backends_pyvista.py` to reproduce the way `mayavi`. I thought it was a required feature

GuillaumeFavelier:
See:

https://github.com/mne-tools/mne-python/blob/c4e9cbf7af9b9b3d92ce41b148eaae3f69f82fc1/mne/viz/backends/_pyvista.py#L122-L133

larsoner:
Okay looks good

GuillaumeFavelier:
I discussed offline with @agramfort and I think using`name=title` can be misleading.

I think using `set_3d_title` is a better idea now.

GuillaumeFavelier:
Sorry for the confusion.

agramfort:
wait we clarified some stuff with Guillaume IRL

>

GuillaumeFavelier:
This is a now different approach: `_set_3d_title()` will add a text label in the scene.

larsoner:
Failure is real:

https://travis-ci.org/mne-tools/mne-python/jobs/658249771

agramfort:
thx @GuillaumeFavelier !

PR #7383, opened 2020-03-04
Fix orientation slider when hemi=both

GuillaumeFavelier:
This PR fixes the overlap of the sliders when `hemi=both`. 

It's an item of #7162


agramfort:
can you share a test script to check/replicate?

GuillaumeFavelier:
Running the following:

```py
import os
import mne
    brain = stc.plot(subjects_dir=subjects_dir,
                     time_viewer=True, hemi='both')
```

This is what I get:

`master` | PR
------------|-------
2020-03-04_1920x1080 | 2020-03-04_1920x1080

PR #7386, opened 2020-03-04
Add clear_points in _TimeViewer

GuillaumeFavelier:
This PR adds a shortcut `c` to clear all traces, the help window is also updated.

Here is an animation with `show_traces='separate'`:

output

It's an item of #7162

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -488,6 +488,7 @@ class _TimeViewer(object):
             'i': self.toggle_interface,
             's': self.apply_auto_scaling,
             'r': self.restore_user_scaling,
+            'c': self.clear_points,
             ' ': self.toggle_playback,
         }
         menu = self.plotter.main_menu.addMenu('Help')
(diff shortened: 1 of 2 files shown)


mne/viz/_brain/_timeviewer.py
```diff
@@ -744,6 +753,7 @@ def help(self):
             ('i', 'Toggle interface'),
             ('s', 'Apply auto-scaling'),
             ('r', 'Restore original clim'),
+            ('c', 'Clear all traces'),
```
agramfort: pressing c is not working for me. Did you forget to configure the callback?

agramfort: also pressing i or s does not work anymore for me :(

GuillaumeFavelier: Did you try with pyvista 0.23.1?

GuillaumeFavelier: This is now the recommended version:
https://github.com/mne-tools/mne-python/issues/7162#issuecomment-586287143

agramfort:
great it works with 0.23.1 !

+1 for MRG !

>

PR #7388, opened 2020-03-04
MRG, MAINT: Decrease memory usage of plot_forward.py

larsoner:
@drammock okay for you? It turns out that the `np.eye(n_sources)` was causing ~500 MB of mem usage (`8192 ** 2 * 8 = 536870912`) so just making a comment about it should fix our recent CircleCI failures, as in local testing it goes from ~2100 MB in `master` down to ~1500 MB in this PR.


drammock:
merge away when CIs are happy.

Issue #7107, opened 2019-12-02
BUG - mne coreg issues


bloyl:
I'm observing odd behavior in `mne coreg` on a clean install on a macbook.

<details>
  <summary>sys.info</summary>

```
(mne-trunk) a483e75df2f6:~ bloyl$  python -c "import mne; mne.sys_info()"
Platform:      Darwin-18.7.0-x86_64-i386-64bit
Python:        3.7.5 (default, Oct 25 2019, 10:52:18)  [Clang 4.0.1 (tags/RELEASE_401/final)]
Executable:    //anaconda3/envs/mne-trunk/bin/python
CPU:           i386: 8 cores
Memory:        16.0 GB

mne:           0.19.2
numpy:         1.17.4 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.3.1
matplotlib:    3.1.1 {backend=MacOSX}

sklearn:       0.21.3
numba:         0.46.0
nibabel:       2.5.1
cupy:          Not found
pandas:        0.25.3
dipy:          1.0.0
mayavi:        4.7.1 {qt_api=pyqt5, PyQt5=5.13.2}
pyvista:       0.22.4
vtk:           8.1.2
```
</details>

The issue is that head surfaces aren't rendering and points (fids, dig etc) are only sometimes visible depending on how the view is rotated etc.

Interestingly this only happens when called this way. `mne coreg`
If the subjects dir is supplied, `mne coreg -s sample -d ~/mne_data/MNE-sample-data/subjects`,  it seems to work correctly.

Here is a gif showing the issue with `mne coreg`
![mne_coreg_short](https://user-images.githubusercontent.com/5585821/69994022-f9a69b80-151a-11ea-85a6-aa3431e5e71b.gif)

and a screenshot of is working with `mne coreg -s sample -d ~/mne_data/MNE-sample-data/subjects`
<img width="1197" alt="image" src="https://user-images.githubusercontent.com/5585821/69991917-97e43280-1516-11ea-823b-0d4515283557.png">

hoechenberger:
I've witnessed the same behavior on my Mac too using 0.19 and, I believe, `master`.

agramfort:
I could not replicate.

can it be a pb with the fsaverage folder you have? in the first case
it's fsaverage and with full command line you use sample

$ python -c "import mne; mne.sys_info()"

Platform:      Darwin-18.7.0-x86_64-i386-64bit

[...]

PR #7392, opened 2020-03-05
MRG, MAINT: Use VTK 8.2

larsoner:
Anaconda now has a newer VTK (8.2) than pip (8.1.2), so let's use it. It's possible that this might fix some depth buffer problems for people -- @bloyl you had an issue open about this #7107, it's possible 8.2 already fixes the problem if you feel like trying it.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,10 +29,10 @@ dependencies:
 - traits>=4.6.0
 - pyface>=6
 - traitsui>=6
+- vtk
 - pip:
   - mne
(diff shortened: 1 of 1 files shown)


larsoner:
... and actually I think we might *need* this change because otherwise anaconda might allow Python 3.8 to be installed, and VTK is not available for that version (they are going to make it part of the VTK 9 release apparently)

agramfort:
whatever works !

larsoner:
Failures are unrelated

PR #7393, opened 2020-03-05
BUG, ENH: Added ch_type option to _apply_reference so custom_ref_applied bool b…

### LINKED ISSUE #7350: Setting custom_ref_applied for ECoG data
#### Setting average reference for ECoG data
When using the `raw.set_eeg_reference(ref_channels='average')` to perform a common average reference on ECoG data, the `custom_ref_applied` bool flag in raw.info remains set to `False`. I looked at `io/reference.py` and in the function definition for `set_eeg_reference` it appears that when `_apply_reference` is called, it looks only for channels of type 'eeg' is 'custom_ref_applied' set to `True`. 

Also, in the comments for `_apply_reference`, it appears to say that `info['custom_ref_applied']=True` refers to a "non-CAR" reference, but the test is whether the reference includes EEG electrodes (which it would for an average reference, right?)

My general question then, is, for consistency, should `custom_ref_applied=True` after applying an average reference to ECoG data, or am I misunderstanding the use of this flag?  It seems like a nice marker for whether this operation has happened or not. Happy to help fix if this makes sense.  I see this is similar to this issue: https://github.com/mne-tools/mne-python/issues/6454, but not exactly.

libertyh:
#### Reference issue
Example: Fixes #7350 

This changes `_apply_reference` in `mne/io/reference.py` so that `custom_ref_applied` is set to True for ECoG and SEEG electrodes (not only EEG electrodes as before). 

#### Additional information
In previous discussions of this change, it was suggested to add `ch_type` to `set_eeg_reference`, but this keyword argument already exists and so did not need to be added. It turns out that `ch_type` was not being used in `_apply_reference` when the boolean flag for `custom_ref_applied` was changed, which could lead to confusion when setting a CAR or other reference with ECoG and sEEG data because it wouldn't be clear that the reference had been set and applied. The previous version only looked for channels of type `'eeg'` before setting `custom_ref_applied` to `True` in the `info` structure.


larsoner:
Looks good so far! Last things would be to

- update `doc/changes/latest.inc` in the `BUG` section mentioning this fix, with your name (added to `doc/changes/names.inc` if not there yet)
- add a test

For the test, you want something that fails in `master` but passes on this PR. You can for example make a `raw = RawArray(np.ones((2, 1000)), create_info(2, 1000., 'ecog')` and do `raw.set_eeg_reference(projection=True)`, and then `assert` some stuff. For example you could `assert len(raw.info['projs']) == 1` and that `assert_array_equal(raw.apply_proj().get_data(), 0.)`.

larsoner:
(If the OSX run times out on Travis you can ignore it)

doc/changes/latest.inc
```diff
@@ -124,6 +124,8 @@ Bug
 
 - Fix :class:`mne.Report` silently suppressing exceptions when used as a context manager by `Marijn van Vliet`_
 
+- Fix bug in :func:`mne.io.reference._apply_reference` to set `info.custom_ref_applied` to `True` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
```
drammock: FYI, "True" needs *two* backticks on either side of it to get properly formatted, not one (it's reStructuredText, not Markdown). This won't error out, it will just make it italicized instead of formatted as code.

libertyh:
> Looks good so far! Last things would be to
> 
> * update `doc/changes/latest.inc` in the `BUG` section mentioning this fix, with your name (added to `doc/changes/names.inc` if not there yet)
> * add a test
> 
> For the test, you want something that fails in `master` but passes on this PR. You can for example make a `raw = RawArray(np.ones((2, 1000)), create_info(2, 1000., 'ecog')` and do `raw.set_eeg_reference(projection=True)`, and then `assert` some stuff. For example you could `assert len(raw.info['projs']) == 1` and that `assert_array_equal(raw.apply_proj().get_data(), 0.)`.

For this test, is the preference that this goes in `test_proj.py`, or that I make a new file (e.g. `test_reference.py` or similar? Thanks for your patience!

larsoner:
> For this test, is the preference that this goes in test_proj.py, or that I make a new file (e.g. test_reference.py or similar? 

I would put it in some existing file. Based on `git grep set_eeg_reference` it looks like the right place would be `mne/io/tests/test_reference.py`.

larsoner (approved):
@libertyh I pushed a merge commit to deal with the conflicts, plus a tiny commit 7719bc6 to:

1. update `latest.inc` to mention public rather than private functions
2. deduplicate / DRY some code
3. add a test that fails on `master` but passes on this PR (TDD; test-driven development)

Feel free to have a look when you get a chance. In the meantime, this LGTM +1 for merge

agramfort (approved):

larsoner:
Thanks @libertyh !

libertyh:
Thank you!! I will be more careful about the tests next time, but glad this went through and thank you for finishing it up!

PR #7399, opened 2020-03-06
Update documentations in montage.py

dominikwelke:
before submitting.

#### Reference issue
addresses my request via mne mailing list (2020.03.03)

specify that head_size parameter for `def read_custom_montage` and `def make_standard_montage` expects a radius

#### Additional information
-

### DIFF
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -967,8 +967,8 @@ def read_custom_montage(fname, head_size=HEAD_SIZE_DEFAULT, coord_frame=None):
         '.elc', '.txt', '.csd', '.elp' (BESA spherical),
         '.bvef' (BrainVision files).
     head_size : float | None
-        The size of the head in meters (radius). If `None`, returns the values read from
-        the montage file with no modification. Defaults to 0.095m.
+        The size of the head (radius, in [m]). If ``None``, returns the values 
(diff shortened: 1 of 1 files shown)


mne/channels/montage.py
```diff
@@ -967,7 +967,7 @@ def read_custom_montage(fname, head_size=HEAD_SIZE_DEFAULT, coord_frame=None):
         '.elc', '.txt', '.csd', '.elp' (BESA spherical),
         '.bvef' (BrainVision files).
     head_size : float | None
-        The size of the head in meters. If `None`, returns the values read from
+        The size of the head in meters (radius). If `None`, returns the values read from
```
drammock: @dominikwelke this line is failing our tests because we adhere to the [Python style guidelines](https://www.python.org/dev/peps/pep-0008/); specifically [maximum line length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) of 79 characters is the problem here.  While you're fixing that, could you also make a small improvement:  "None" should be surrounded by two backticks before & after.

dominikwelke: line length is 79 characters now, but still failing..

dominikwelke: ok, i'd introduced a trailing whitespace

larsoner:
Thanks @dominikwelke !

PR #7168, opened 2019-12-29
MRG, ENH: Extend reduce_rank=True support and better deficiency check

larsoner:
Closes #5880.

@agramfort `reduce_rank=True` was still only implemented for `pick_ori == 'max_power' and weight_norm is not None`. This adds support for the other modes, plus a sanity check on the gain matrix directly rather than relying on `linalg` functions to fail.

@britta-wstnr @wmvanvliet it would be good if you could take a look.

PR #7401, opened 2020-03-06
FIX: Fix rank reduction for beamformers

britta-wstnr:
As discussed in https://github.com/mne-tools/mne-python/pull/7168, the option `reduce_rank=True` does not behave as one might expect:
The code is actually not implementing lead field rank reduction but reducing the rank of the denominator of the beamformer formula.

This PR fixes this, and completely swaps the old rank reduction for the lead field one.

Tests are left to fail for first round, so we can see what will change with this in place.

### DIFF
--- a/mne/beamformer/_compute_beamformer.py
+++ b/mne/beamformer/_compute_beamformer.py
@@ -18,7 +18,8 @@ from ..io.proj import make_projector, Projection
 from ..minimum_norm.inverse import _get_vertno, _prepare_forward
 from ..source_space import label_src_vertno_sel
 from ..utils import (verbose, check_fname, _reg_pinv, _check_option, logger,
-                     _pl, _check_src_normal, check_version, _validate_type)
+                     _pl, _check_src_normal, check_version, _validate_type,
+                     warn)
 from ..time_frequency.csd import CrossSpectralDensity
 
 from ..externals.h5io import read_hdf5, write_hdf5
@@ -248,9 +249,18 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
         pinv_kwargs['hermitian'] = True
 
     # leadfield rank and optional rank reduction
-    _validate_type(reduce_rank, bool, "reduce_rank", "a boolean")
+    if reduce_rank is True:
+        reduce_rank = 'denominator'
+        warn('reduce_rank=True will reduce the rank of the denominator of the '
+             'beamformer formula. If you meant to reduce the rank of the '
+             'leadfield instead, set reduce_rank to "leadfield".')
+    _check_option('reduce_rank', reduce_rank, ('leadfield', 'denominator',
+                                               False))
+
+    # inversion of the denominator
     _check_option('inversion', inversion, ('matrix', 'single'))
     if reduce_rank and inversion == 'single':
+        # TODO: what about the leadfield rank reduction with backprojection?
         raise ValueError('reduce_rank cannot be used with inversion="single"; '
                          'consider using inversion="matrix" if you have a '
                          'rank-deficient forward model (i.e., from a sphere '
@@ -263,7 +273,21 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
             raise ValueError(
                 'Singular matrix detected when estimating spatial filters. '
                 'Consider reducing the rank of the forward operator by using '
-                'reduce_rank=True.')
+                'reduce_rank="leadfield" or of the beamformer denominator by '
+                'using reduce_rank="denominator".')
+
+    # rank reduction of the lead field
+    if reduce_rank == 'leadfield':
+        # decompose lead field
+        u, s, v = np.linalg.svd(Gk)
+        s[:, np.argmin(s, axis=1)] = 0.  # set the smalles singular value to 0.
+        # expand s to match dimension of u
+        s_full = np.zeros(Gk.shape)
+        for s_full_vox, s_vox in zip(s_full, s):
+            np.fill_diagonal(s_full_vox, s_vox)
+
+        # backproject
+        Gk = np.matmul(u, np.matmul(s_full, v))
 
     with _noop_indentation_context():
         if (inversion == 'matrix' and pick_ori == 'max-power' and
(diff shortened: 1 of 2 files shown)


mne/beamformer/_compute_beamformer.py
```diff
@@ -116,7 +116,25 @@ def _sym_inv(x, reduce_rank):
...
+    """Reduce the rank of the leadfield."""
+    # decompose lead field
+    u, s, v = np.linalg.svd(G)
+    s[:, np.argmin(s, axis=1)] = 0.  # set the smallest singular value to 0.
+
+    # expand s to match dimension of u
+    s_full = np.zeros(G.shape)
+    # TODO: vectorize?
+    for s_full_vox, s_vox in zip(s_full, s):
+        np.fill_diagonal(s_full_vox, s_vox)
```
britta-wstnr: @larsoner, can you do some magic on this? I am certain this can be done faster and more elegant ...

mne/beamformer/_compute_beamformer.py
```diff
@@ -116,7 +116,25 @@ def _sym_inv(x, reduce_rank):
     return np.matmul(u * s[:, np.newaxis], u.transpose(0, 2, 1))
 
 
-def _normalized_weights(Wk, Gk, Cm_inv_sq, reduce_rank, nn, sk):
+def _reduce_leadfield_rank(G):
+    """Reduce the rank of the leadfield."""
+    # decompose lead field
+    u, s, v = np.linalg.svd(G)
+    s[:, np.argmin(s, axis=1)] = 0.  # set the smallest singular value to 0.
```
larsoner: SVD is sorted descending so you can just set `-1` to zero

mne/beamformer/_compute_beamformer.py
```diff
@@ -116,7 +116,25 @@ def _sym_inv(x, reduce_rank):
...
+
+    # expand s to match dimension of u
+    s_full = np.zeros(G.shape)
+    # TODO: vectorize?
```
larsoner: Don't actually construct these diagonal matrices. Instead, promote `s`'s dimensionality and operate elementwise using broadcasting (this is equiv to matmul with a diag mat) like:
```
s = s[..., np.newaxis, :]
G = np.matmul(u * s, v)
```

larsoner:
> Tests are left to fail for first round, so we can see what will change with this in place.

Oh okay :)

britta-wstnr:
@larsoner on my machine one DICS test fails and actually has a higher localization error after this fix. To me that makes sense as we scratch one orientation and then acutally pick orientation in that test ... But I'd like to see if you agree.

I will get to your comments soon, thank you!

larsoner:
> on my machine one DICS test fails and actually has a higher localization error after this fix. To me that makes sense as we scratch one orientation and then acutally pick orientation in that test ... But I'd like to see if you agree.

I see the same failure locally. Let me see if I can make the test a bit more robust/clear by refactoring a bit

britta-wstnr:
@larsoner, I worked on your comments and rebased. Locally, 3 tests fail for me in `test_dics`, let's see whether Travis agrees and what to do about it!

mne/beamformer/_compute_beamformer.py
```diff
@@ -291,7 +308,7 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
                     assert Ck.shape[1:] == (3, 3)
                     # Invert for all dipoles simultaneously using matrix
                     # inversion.
-                    norm = _sym_inv(Ck, reduce_rank)
+                    norm = _sym_inv(Ck, reduce_rank=False)
```
larsoner: Leaving this as `reduce_rank=reduce_rank` fixes two out of the three errors. I think it makes sense to reduce in both places. During inversion you don't want to invert three components, just two in the case of `reduce_rank`.

larsoner: (If we were to make the leadfield actually have dimensionality 2 instead of 3, we could have `reduce_rank=False` here and maybe simplify things. But then we'd have to keep track of the rotation / singular vectors, and rotate back after finishing the computations. Either way should be equivalent, and maybe the 2-dim way a bit faster, not sure which way you want to go...)

larsoner: Sorry I mean `n_orient=2` instead of `n_orient=3` in `reduce_rank=True` mode (not "dimensionality")

britta-wstnr: Hummm, I see what you mean and I can totally see that this would work. It might be unexpected beamformer behaviour for the user though. Ultimately, you reduced the rank of the leadfield already at this point ... 
But on the other hand, it basically is just a "pinv" alternative, right?

britta-wstnr: Well, having `reduce_rank` also reducing the number of orientations could be a way to handle things as well (I think FieldTrip for example supports both, with or without backprojection). It would indeed add a bit of tracking though ... And I am not sure how attractive this option would be in the real world. I find it intuitive (to reduce the output to 2 orientations, in case of a vector beamformer) but probably a lot of our orientation selection code might break?

larsoner: > But on the other hand, it basically is just a "pinv" alternative, right?

Yeah basically you need to pass `reduce_rank` to the inversion functions to make sure they don't blow something up that they shouldn't. If I use a much smaller `tol` in the effective `pinv` calculation it also works, but it seems safer just to explicitly reduce it

larsoner: Having tried now to quickly implement the lower-dimensional solution, I don't think it's worth it. We also would need to transform `Ck` to the lower-dimensional space of size `2`, do everything there, and then project back to the space of size `3`. I doubt the speed gains are meaningful and the code would require more variables to be passed around.

britta-wstnr:
Should I update the bug description if we now _also_ reduce rank on the denominator?

larsoner:
I guess so -- up to you

mne/beamformer/_compute_beamformer.py
britta-wstnr: Yeah that was my fear as well. So let's stick with `sym_inv(data, reduce_rank=True)`

britta-wstnr:
Let's do that to have it clean. I just need to commit changes from another branch then I'll push it.

britta-wstnr:
Good to go from my end, @larsoner.

larsoner:
Thanks @britta-wstnr !

PR #7402, opened 2020-03-06
MRG, MAINT: Parametrize test idx

larsoner:
Useful for #7401 and other beamformer testing because it generalizes the test to multiple source space vertices / labels. It should help test that things get generally better even if some get better and some get worse.

Ready for review/merge from my end @britta-wstnr @wmvanvliet @agramfort . This is a pretty straightforward (if you like pytest :) ) extension of what we have currently, plus cleanups of some `TODO`s in the code, which is nice.

I say we merge this then rebase #7401 so we can more clearly see what happens there based on how many tolerances go up versus down (or if we need to introduce new tolerances).

### DIFF
--- a/mne/beamformer/tests/test_dics.py
+++ b/mne/beamformer/tests/test_dics.py
@@ -18,6 +18,7 @@ from mne.beamformer import (make_dics, apply_dics, apply_dics_epochs,
 from mne.time_frequency import csd_morlet
 from mne.utils import run_tests_if_main, object_diff, requires_h5py
 from mne.proj import compute_proj_evoked, make_projector
+from mne.surface import _compute_nearest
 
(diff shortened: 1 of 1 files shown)


larsoner:
Failures are spurious, restarted (but can be ignored)

larsoner:
https://18383-1301584-gh.circle-artifacts.com/0/dev/generated/mne.minimum_norm.make_inverse_resolution_matrix.html#mne.minimum_norm.make_inverse_resolution_matrix

Shrink the window width to see the magic happen

larsoner:
Whoops, wrong PR!

larsoner:
I'll merge this so that @britta-wstnr can rebase #7401 and see what needs to be tweaked here

PR #7403, opened 2020-03-06
MRG, STY: Allow wrapping long API lines in HTML

larsoner:
Allows long API titles like `mne.minimum_norm.source_induced_power` to wrap.

A [better solution](https://stackoverflow.com/questions/23107990/word-wrap-on-non-alphabetical-character) would be to change `autosummary` to put each `X.`/`Y.` of `X.Y.Z` in a separate `span`, then you can make it break on the periods. But this is a lot more work for a marginal benefit.

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -480,3 +480,7 @@ a[class^="sphx-glr-backref-module-mne"] {  /* make all MNE things bold */
 a[class^="sphx-glr-backref-module-mne"]:hover {
     color: var(--a-color-light);
 }
+/* Long API titles need to wrap for mobile */
+div[id^="mne-"] h1 {
+    word-break: break-word;
+}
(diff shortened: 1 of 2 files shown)


larsoner:
https://18383-1301584-gh.circle-artifacts.com/0/dev/generated/mne.minimum_norm.make_inverse_resolution_matrix.html#mne.minimum_norm.make_inverse_resolution_matrix

Shrink the window width to see the magic happen

drammock:
Can we do it for the "examples using ..." text too?  See bottom of screenshot:

Screenshot_2020-03-06_10-10-56

larsoner:
@drammock should be fixed now

larsoner:
(also slipped in a fix to use `d-block` instead of `row` -- `row` should always be in a `container` div in boostrap otherwise you get horizontal scrollbars due to the `margin` settings, but if we do this it will only fill the middle portion of a page -- `d-block` does not have these problems)

larsoner:
https://18384-1301584-gh.circle-artifacts.com/0/dev/generated/mne.minimum_norm.source_induced_power.html#mne.minimum_norm.source_induced_power

drammock:
changes shouldn't affect azure CIs, no need to wait for them.

drammock:
tx @larsoner

PR #7408, opened 2020-03-09
MRG: Add nirs support to plot_evoked_topo

rob-luke:
#### Reference issue
Adding general NIRS support #7057 

Allows plotting of evoked topo map for nirs data

### DIFF
--- a/mne/viz/tests/test_topo.py
+++ b/mne/viz/tests/test_topo.py
@@ -70,9 +70,9 @@ def _get_epochs_delayed_ssp():
             proj='delayed', reject=reject)
     return epochs_delayed_ssp
 
+
 def _gen_nirs_data():
     """Generate test NIRS data."""
-    # Test plotting of fnirs types
     montage = make_standard_montage('biosemi16')
     ch_names = montage.ch_names
     ch_types = ['eeg'] * 16
@@ -81,6 +81,7 @@ def _gen_nirs_data():
     evokeds = EvokedArray(evoked_data, info=info, tmin=-0.2, nave=4)
     evokeds.set_montage(montage)
     return evokeds
+
 
 def test_plot_joint():
     """Test joint plot."""
--- a/mne/viz/tests/test_topo.py
+++ b/mne/viz/tests/test_topo.py
@@ -2,6 +2,7 @@
 #          Denis Engemann <denis.engemann@gmail.com>
 #          Martin Luessi <mluessi@nmr.mgh.harvard.edu>
 #          Eric Larson <larson.eric.d@gmail.com>
+#          Robert Luke <mail@robertluke.net>
 #
 # License: Simplified BSD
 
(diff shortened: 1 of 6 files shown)


rob-luke:
@larsoner could you please review. The function now behaves as I would expect. The CI errors seem unrelated to the code changes.

rob-luke:
Sorry to modify this after you approved @agramfort  
But I realised I had repeated some code that will probably be common as I add more nirs tests, so I have created a helper function to generate the data. Let me know if you want me to revert this change.

mne/viz/tests/test_topomap.py
```diff
@@ -138,13 +138,7 @@ def test_plot_topomap_animation():
...
 
     # Test plotting of fnirs types
-    montage = make_standard_montage('biosemi16')
-    ch_names = montage.ch_names
-    ch_types = ['eeg'] * 16
-    info = create_info(ch_names=ch_names, sfreq=20, ch_types=ch_types)
-    evoked_data = np.random.randn(16, 30)
-    evokeds = EvokedArray(evoked_data, info=info, tmin=-0.2, nave=4)
-    evokeds.set_montage(montage)
+    evokeds = _gen_nirs_data()
```
agramfort: it's weird that _gen_nirs_data gets you some data with no NIRS channel. Can you put the set_channel_types in the private function _gen_nirs_data?

rob-luke: Yes that was an odd choice. Should be fixed now.

mne/viz/tests/test_topomap.py
```diff
@@ -34,6 +33,7 @@
 from mne.viz.topomap import (_get_pos_outlines, _onselect, plot_topomap,
                              plot_arrowmap, plot_psds_topomap)
 from mne.viz.utils import _find_peaks, _fake_click
+from mne.viz.tests.test_topo import _gen_nirs_data
```
larsoner: The modern way to do this would be to create a `mne/viz/tests/conftest.py` and add a:
```
@pytest.fixture()
def fnirs_evoked():
    """Create a fnirs evoked."""
    ...
    return evoked
```

mne/viz/tests/test_topomap.py
larsoner: ... then in this func you just do:
```
def test_plot_topomap_animation(fnirs_evoked):
    ...
```
And `fnirs_evoked` will be the one returned by the fixture

mne/viz/tests/test_topo.py
```diff
@@ -161,6 +176,13 @@ def test_plot_topo():
         break
     plt.close('all')
 
+    # Test plotting of fnirs types
+    evokeds = _gen_nirs_data()
```
larsoner: same here, you'd do:
```
def test_plot_topo(fnirs_evoked):
    ...
```
And it's available to you

rob-luke:
Ok don’t merge. And I will try this fancy pytest stuff.

rob-luke:
Thanks @larsoner I think I have implemented your suggestion. Can you please review again.

larsoner:
Thanks @rob-luke !

Issue #7354, opened 2020-02-25
BUG: bonferroni correction output p-values > 1.


kingjr:
```python
from mne.stats import bonferroni_correction
_, p_corrected = bonferroni_correction([.2, .9])
```
p_corrected:
```
[0.4, 1.8]
```

Formally, I'm not sure what should be done, but I think clipping below 1 could be reasonable?

larsoner:
Agreed we can just clip at 1., PR welcome

PR #7409, opened 2020-03-09
Ensure corrected p-values are always <= 1.0

hoechenberger:
#### Reference issue
Fixes GH-7354.

Clips p-values produced by Bonferroni correction at 1.0


hoechenberger:
Whoops. There's a bug and the change breaks another test locally. Will look into it.

hoechenberger:
Ok so the new test function `test_bonferroni_pval_clip()`, which checks if the p-vals don't exceed 1, is _implicitly_ also included in the modified `test_multi_pval_correction()` function (line 34). So there is a little "test redundancy" here now; however I propose to keep `test_bonferroni_pval_clip()` as it's a very straightforward function and not as convoluted as `test_multi_pval_correction()`. WDYT?

larsoner:
Thanks @hoechenberger

PR #7414, opened 2020-03-10
MRG: Take mean of nirs channels that are too close for topoplot

rob-luke:
#### Reference issue
Alternative approach to #7410 as suggested by @larsoner 

If nirs channels are too close for the topoplot to represent then those channels are combined by taking the mean.

#### Additional information
Initial attempt. No tests yet, just looking for feedback to see if I am on the right path. But this works on my data.

### DIFF
--- a/mne/channels/layout.py
+++ b/mne/channels/layout.py
@@ -907,6 +907,42 @@ def _merge_grad_data(data, method='rms'):
     return data.reshape(data.shape[:1] + orig_shape[1:])
 
 
+def _merge_nirs_data(data, merged_names):
+    """Merge data from multiple nirs channel using the mean.
+
+    Channel names that have an x in them will be merged. The first channel in
+    the name is replaced with the mean of all listed channels. The other
+    channels are removed.
+
+    Parameters
+    ----------
+    data : array, shape = (n_channels, ..., n_times)
+        Data for channels.
+    merged_names : list
+        List of strings containing the channel names. Channels that are to be
+        merged contain an x between them.
+
+    Returns
+    -------
+    data : array
+        Data for channels with requested channels merged. Channels used in the
+        merge are removed from the array.
+    """
+    to_remove = np.empty(0, dtype=np.int32)
+    for idx, ch in enumerate(merged_names):
+        if 'x' in ch:
+            indices = np.empty(0, dtype=np.int32)
+            channels = ch.split("x")
+            for sub_ch in channels[1:]:
+                indices = np.append(indices, merged_names.index(sub_ch))
+            data[idx] = np.mean(data[np.append(idx, indices)], axis=0)
+            to_remove = np.append(to_remove, indices)
+    for rem in sorted(to_remove, reverse=True):
+        del merged_names[rem]
+        data = np.delete(data, rem, 0)
+    return data, merged_names
+
+
 def generate_2d_layout(xy, w=.07, h=.05, pad=.02, ch_names=None,
                        ch_indices=None, name='ecog', bg_image=None,
                        normalize=True):
--- a/mne/channels/layout.py
+++ b/mne/channels/layout.py
@@ -5,6 +5,7 @@
 #          Marijn van Vliet <w.m.vanvliet@gmail.com>
 #          Jona Sassenhagen <jona.sassenhagen@gmail.com>
 #          Teon Brooks <teon.brooks@gmail.com>
+#          Robert Luke <mail@robertluke.net>
 #
 # License: Simplified BSD
 
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -15,13 +15,14 @@ from numbers import Integral
 import warnings
 
 import numpy as np
+from scipy.spatial.distance import pdist, squareform
 
 from ..baseline import rescale
 from ..defaults import HEAD_SIZE_DEFAULT
 from ..channels.channels import _get_ch_type
 from ..channels.layout import (
     _find_topomap_coords, _merge_grad_data, find_layout, _pair_grad_sensors,
-    Layout)
+    Layout, _merge_nirs_data)
 from ..io.pick import (pick_types, _picks_by_type, channel_type, pick_info,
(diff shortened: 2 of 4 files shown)


mne/viz/topomap.py
```diff
@@ -15,13 +16,14 @@
 import warnings
 
 import numpy as np
+from scipy.spatial.distance import pdist, squareform
```
larsoner: This will probably need to be a nested import

https://travis-ci.org/mne-tools/mne-python/jobs/660587770#L3259

mne/viz/topomap.py
```diff
@@ -92,6 +94,46 @@ def _prepare_topomap_plot(inst, ch_type, layout=None, sphere=None):
         picks, _ = _pair_grad_sensors(info, layout)
         pos = _find_topomap_coords(info, picks[::2], sphere=sphere)
         merge_grads = True
+    elif ch_type in ['hbo', 'hbr', 'fnirs_raw', 'fnirs_od']:
+        # Special case for merging nirs channels
+        picks = pick_types(info, meg=False, ref_meg=False, fnirs=ch_type,
+                           exclude='bads')
+        chs = [info['chs'][i] for i in picks]
```
larsoner: This is enough extra lines that it should probably all live in a `picks, pos, merge_grads = _average_fnirs_overlaps(...)` call or so. It makes the code more readable, and clearer what values we will use in the rest of this function (in my example, `picks`, `pos`, and `merge_grads`), what's required to compute the overlaps, etc.

mne/viz/topomap.py
```diff
@@ -100,9 +142,6 @@ def _prepare_topomap_plot(inst, ch_type, layout=None, sphere=None):
         elif ch_type == 'csd':
             picks = pick_types(info, meg=False, csd=True, ref_meg=False,
                                exclude='bads')
-        elif ch_type in ['hbo', 'hbr', 'fnirs_raw', 'fnirs_od']:
```
larsoner: I would make a `_fnirs_types = ('hbo', 'hbr', 'fnirs_raw', 'fnirs_od')` somewhere above outside this function and reuse it here, above, below, and in `plot_evoked_topomap`.

mne/viz/topomap.py
```diff
@@ -113,10 +152,23 @@ def _prepare_topomap_plot(inst, ch_type, layout=None, sphere=None):
...
-    if merge_grads:
+    if ch_type in ['hbo', 'hbr', 'fnirs_raw', 'fnirs_od']:
+        # Remove the chroma label type for cleaner labeling.
+        ch_names = [k[:-4] for k in ch_names]
+
+    if merge_grads and ch_type == 'grad':
         # change names so that vectorview combined grads appear as MEG014x
         # instead of MEG0142 or MEG0143 which are the 2 planar grads.
         ch_names = [ch_names[k][:-1] + 'x' for k in range(0, len(ch_names), 2)]
+    elif merge_grads and ch_type in ['hbo', 'hbr', 'fnirs_raw', 'fnirs_od']:
```
larsoner: I would make the conditional:
```
if merge_grads:
    if ch_type == 'grad':
        ...
    else:
        assert fnirs_ch_type  # should be guaranteed above
        ...
```

mne/viz/topomap.py
```diff
@@ -1630,8 +1683,11 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None, layout=None,
                          'Check your input.')
 
     data *= scaling
-    if merge_grads:
+    if merge_grads and ch_type == 'grad':
         data = _merge_grad_data(data)
+    elif merge_grads and ch_type in ['hbo', 'hbr', 'fnirs_raw', 'fnirs_od']:
```
larsoner: I would rename `merge_grads` to `merge_channels` everywhere

larsoner:
Also I would recommend `git grep _merge_grad_data`'ing to see where else you might need to change things.

rob-luke:
> Also I would recommend `git grep _merge_grad_data`'ing to see where else you might need to change things.

So it appears `_merge_grad_data` is also used in `get_peak`, `plot_projs_topomap`, `_plot_ica_topomap`, `plot_ica_components`, `plot_tfr_topomap`, `plot_epochs_psd_topomap` and like a dozen helper functions.

So I think I will tackle each topic in turn to keep things bite sized. I will look at `psd_topomap`, `ica`, `tfr`, then `projs` in subsequent PRs. But first I will take and upload a short test measurement with overlapping channels so that I can make useful tests. @larsoner is this plan ok?

mne/viz/topomap.py
rob-luke: I have changed it within the range of code that I have modified. Or do you think I should change this globally?

larsoner: I would just change it locally and eventually we can change it to be more global, maybe by putting it in `mne/io/pick.py` and importing it here and elsewhere

larsoner:
> So I think I will tackle each topic in turn to keep things bite sized.

Sounds reasonable to me. Another approach *might* be to make `_merge_grad_data` a more general `_merge_channel_data` that handles both channel types, but based on its arguments I think this approach will not be straightforward (even if it ends up the way we go after you change a few of these code paths over). +1 for working in bite sized increments and we can see if patterns are repeated and can refactored.

rob-luke:
#7057

rob-luke:
@agramfort and @larsoner could you please review. This is working nicely for me.

mne/channels/layout.py
```diff
@@ -907,6 +908,42 @@ def _merge_grad_data(data, method='rms'):
...
+        List of strings containing the channel names. Channels that are to be
+        merged contain an x between them.
+
+    Returns
+    -------
+    data : array
+        Data for channels with requested channels merged. Channels used in the
+        merge are removed from the array.
+    """
+    to_remove = np.empty(0, dtype=np.int32)
```
larsoner: For linked lists the builtin `list` actually works quite well and probably faster than `np.append`, but it won't make any appreciable difference here

PR #7419, opened 2020-03-10
DOC: improve docstring for extract_label_time_course

drammock:
gives (hopefully) better explanation for the various `mode` options; converts func & method to use docdict along the way.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -728,6 +728,50 @@ transparent : bool | None
     If True, use a linear transparency between fmin and fmid.
     None will choose automatically based on colormap type.
 """
+
+# STC label time course
+docdict['eltc_labels'] = """
+labels : Label | BiHemiLabel | list of Label or BiHemiLabel
+    The labels for which to extract the time course.
+"""
+docdict['eltc_src'] = """
+src : list
+    Source spaces for left and right hemisphere.
+"""
+docdict['eltc_mode'] = """
+mode : str
+    Extraction mode, see Notes.
+"""
+docdict['eltc_allow_empty'] = """
+allow_empty : bool
+    Instead of emitting an error, return all-zero time courses for labels
+    that do not have any vertices in the source estimate. Default is ``False``.
+"""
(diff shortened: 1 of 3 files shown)


mne/utils/docs.py
```diff
@@ -728,6 +728,50 @@
...
+docdict['eltc_mode_notes'] = """
+Valid values for ``mode`` are:
+
+``'max'``
+    Maximum value across vertices at each time point within each label.
+``'mean'``
+    Average across vertices at each time point within each label. Ignores
+    orientation of sources.
+``'mean_flip'``
+    Finds the dominant direction of source orientations within each label,
```
larsoner: Somewhere in here it should say some variant of "based on the source space normal directions" to make it clear that the "dominant direction" is entirely based on anatomy (not the time series data in `stc`)

drammock:
rendered [function docs](https://18465-1301584-gh.circle-artifacts.com/0/dev/generated/mne.extract_label_time_course.html) and [method docs](https://18465-1301584-gh.circle-artifacts.com/0/dev/generated/mne.SourceEstimate.html#mne.SourceEstimate.extract_label_time_course)

mne/utils/docs.py
```diff
@@ -728,6 +728,49 @@
...
+"""
+docdict['eltc_allow_empty'] = """
+allow_empty : bool
+    Instead of emitting an error, return all-zero time courses for labels
+    that do not have any vertices in the source estimate. Default is ``False``.
+"""
+docdict['eltc_mode_notes'] = """
+Valid values for ``mode`` are:
+
+``'max'``
```
larsoner: I usually find that:
```
- X
     desc
- Y
    desc
```
looks better when rendered (otherwise they look like headings rather than a list of options), but I'm fine with this. Feel free to merge if you're happy @drammock

drammock: addressed by 4e3d1e2

agramfort:
Thx @drammock

PR #7423, opened 2020-03-11
MRG: Link _TimeViewer time slider with vertical time line

GuillaumeFavelier:
This PR adds a vertical line to indicate the current time on the `matplotlib` canvas.

I also did quite the refactor there as `_TimeViewer.__init__` was getting too long.

output

It's an item of #7162

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -115,7 +115,7 @@ class TimeSlider(object):
         """Update the time slider."""
         value = float(value)
         if not time_as_index:
-            value = self.brain.to_time_index(value)
+            value = self.brain._to_time_index(value)
         self.brain.set_time_point(value)
         if self.callback is not None:
             self.callback()
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -114,10 +114,8 @@ class TimeSlider(object):
     def __call__(self, value, update_widget=False, time_as_index=True):
         """Update the time slider."""
         value = float(value)
-        # XXX this should probably use some brain method
         if not time_as_index:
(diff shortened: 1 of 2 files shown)


mne/viz/_brain/_timeviewer.py
```diff
@@ -753,6 +775,26 @@ def plot_time_course(self, hemi, vertex_id, color):
...
+        if not hasattr(self, "mpl_canvas"):
+            return
+        if isinstance(self.show_traces, bool) and self.show_traces:
+            # add time information
+            current_time = self.brain._current_time
+            xdata = [current_time, current_time]
+            if not hasattr(self, "time_line_height"):
+                self.time_line_height = np.max(self.brain._data["array"])
+            if not hasattr(self, "time_line"):
+                self.time_line = self.mpl_canvas.plot(
```
larsoner: Why not just `axvline` here?

GuillaumeFavelier:
> feel free to take a look and fix if need be.

Could not have done better.

I just added `to_time_index()` to `_Brain`

larsoner:
@GuillaumeFavelier can we make that private and also get rid of `index_for_time`? Things like `index_for_time` seem like cruft left over from `PySurfer` that we don't want in the public interface of our new `Brain` class.

GuillaumeFavelier:
@larsoner there are multiple places where I consider that `time_idx` is `int` (which was guaranteed by `index_for_time` with the default parameters). This might take more time than expected to refactor.

larsoner:
Okay, might not be worth it then...

GuillaumeFavelier:
I took care of it in the latest commit 363e81f, what do you think?

larsoner:
Thanks @GuillaumeFavelier

PR #7427, opened 2020-03-12
MRG: Add nirs support to ICA topoplot

rob-luke:
#### Reference issue
Ongoing fNIRS improvements for #7057 

Adds support for plotting Ica components.

### DIFF
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -312,7 +312,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
     def plot_topo(self, layout=None, layout_scale=0.945, color=None,
                   border='none', ylim=None, scalings=None, title=None,
                   proj=False, vline=[0.0], fig_background=None,
-                  merge_channels=False, legend=True, axes=None,
+                  merge_grads=False, legend=True, axes=None,
                   background_color='w', noise_cov=None, show=True):
         """
         Notes
@@ -323,7 +323,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
             self, layout=layout, layout_scale=layout_scale, color=color,
             border=border, ylim=ylim, scalings=scalings, title=title,
             proj=proj, vline=vline, fig_background=fig_background,
-            merge_channels=merge_channels, legend=legend, axes=axes,
+            merge_grads=merge_grads, legend=legend, axes=axes,
             background_color=background_color, noise_cov=noise_cov, show=show)
 
     @copy_function_doc_to_method_doc(plot_evoked_topomap)
@@ -503,7 +503,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         return out
 
     def get_peak(self, ch_type=None, tmin=None, tmax=None,
-                 mode='abs', time_as_index=False, merge_channels=False,
+                 mode='abs', time_as_index=False, merge_grads=False,
                  return_amplitude=False):
         """Get location and latency of peak amplitude.
 
@@ -526,7 +526,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
             Defaults to 'abs'.
         time_as_index : bool
             Whether to return the time index instead of the latency in seconds.
-        merge_channels : bool
+        merge_grads : bool
             If True, compute peak from merged gradiometer data.
         return_amplitude : bool
             If True, return also the amplitude at the maximum response.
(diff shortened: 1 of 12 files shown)


agramfort:
tell us when to look @rob-luke

rob-luke:
Tests look to be passing but codecov has decreased, so trying the test dataset that has overlapping channels to hit those lines.

rob-luke:
@larsoner and @agramfort could you please review this PR

rob-luke:
Sorry I’ve broken the tests. Please don’t review. I will fix tomorrow

rob-luke:
Ok finally all green 😡 

@larsoner and @agramfort could you please review.

agramfort:
rather than branching with an if in each function that calls _merge_grad_data I would suggest to rename _merge_grad_data to _merge_ch_data and do the if branch in a single place. Doing a `git grep _merge_grad_data ` shows you that you made it otherwise do it in other places later

rob-luke:
Thanks @agramfort. Have I understood your suggestion correctly (based on my last commit)? Not ready for another review yet (CI is still yellow), just want to know if I understood your comment correctly.

agramfort:
almost :)

I would rename _merge_grad_data to _merge_ch_data so _merge_ch_data is now
used anywhere you were using so far _merge_grad_data

rob-luke:
Thanks @agramfort. `_merge_ch_data` takes different arguments to `_merge_grad_data` and returns two variables not one. So will take more extensive changes than I can do now. But glad I understand what you mean, I'll get on it.

rob-luke:
@larsoner I can't understand what's causing these CIs to fail. Could you please give me hint.

larsoner:
Travis and Azure are having segfaults in the last few days, unfortunately for now you'll have to look at the logs to make sure your tests are passing :(

rob-luke:
Ok my tests are passing in most CI, except random error. And on my PC.

@agramfort and @larsoner could you please review.

mne/evoked.py
```diff
@@ -312,7 +312,7 @@ def plot_image(self, picks=None, exclude='bads', unit=True, show=True,
     def plot_topo(self, layout=None, layout_scale=0.945, color=None,
                   border='none', ylim=None, scalings=None, title=None,
                   proj=False, vline=[0.0], fig_background=None,
-                  merge_grads=False, legend=True, axes=None,
+                  merge_channels=False, legend=True, axes=None,
```
agramfort: you cannot change public APIs like that. We need deprecation cycles.

mne/evoked.py
```diff
@@ -526,7 +526,7 @@ def get_peak(self, ch_type=None, tmin=None, tmax=None,
             Defaults to 'abs'.
         time_as_index : bool
             Whether to return the time index instead of the latency in seconds.
-        merge_grads : bool
+        merge_channels : bool
             If True, compute peak from merged gradiometer data.
```
agramfort: same here. And docstrings should then be updated. To be honest I would revert this one unless you need it already for NIRS

mne/evoked.py
agramfort: To be honest I would revert this one unless you need it already for NIRS

rob-luke: No don't need this at all. I just misunderstood a previous comment and thought I was meant to change all occurrences of merge grad to merge channels. I'll change this back

agramfort:
sorry I meant internal calls not public API

>

rob-luke:
all good. I will leave all internal calls at merge_channels, and revert all public api to merge_grads.

mne/viz/topo.py
```diff
@@ -568,7 +568,7 @@ def _plot_evoked_topo(evoked, layout=None, layout_scale=0.945, color=None,
                       border='none', ylim=None, scalings=None, title=None,
                       proj=False, vline=(0.,), hline=(0.,), fig_facecolor='k',
                       fig_background=None, axis_facecolor='k', font_color='w',
-                      merge_grads=False, legend=True, axes=None, show=True,
+                      merge_channels=False, legend=True, axes=None, show=True,
```
rob-luke: Is this considered public API? Should I refer this change too?

agramfort:
what is public API is any function or class that does **not** start with _

so here you're good.

mne/viz/topomap.py
```diff
@@ -334,14 +334,15 @@ def plot_projs_topomap(projs, info, cmap=None, sensors=True,
         data = proj['data']['data'].ravel()
         info_names = _clean_names(info['ch_names'], remove_whitespace=True)
         use_info = pick_info(info, pick_channels(info_names, ch_names))
-        data_picks, pos, merge_grads, names, _, this_sphere, clip_origin = \
+        data_picks, pos, merge_channels, names, _, this_sphere, clip_origin = \
```
agramfort: revert this is public

agramfort: my bad

larsoner:
Thanks @rob-luke

PR #7428, opened 2020-03-12
MRG: Add interpolation for NIRS signals

rob-luke:
#### Reference issue
Ongoing fNIRS improvements #7057 

Adds the most basic type of "fixing" of bad channels, simply replaces them by the closest good channel.

#### Additional information
Maybe interpolation isn't the right word for what I wrote, but it seemed inline with previous functions. And I would like to expand to something smarter like actual interpolation

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -962,7 +962,7 @@ class InterpolationMixin(object):
 
     @verbose
     def interpolate_bads(self, reset_bads=True, mode='accurate',
-                         fnirs_method='nearest', origin='auto', verbose=None):
+                         origin='auto', verbose=None):
         """Interpolate bad MEG and EEG channels.
 
         Operates in place.
@@ -975,9 +975,6 @@ class InterpolationMixin(object):
             Either ``'accurate'`` or ``'fast'``, determines the quality of the
             Legendre polynomial expansion used for interpolation of MEG
             channels.
-        fnirs_method : str
-            Method to be used for fNIRS interpolation. Currently only 'nearest'
-            is supported.
         origin : array-like, shape (3,) | str
             Origin of the sphere in the head coordinate frame and in meters.
             Can be ``'auto'`` (default), which means a head-digitization-based
@@ -1007,7 +1004,7 @@ class InterpolationMixin(object):
         origin = _check_origin(origin, self.info)
         _interpolate_bads_eeg(self, origin=origin)
         _interpolate_bads_meg(self, mode=mode, origin=origin)
-        _interpolate_bads_nirs(self, method=fnirs_method)
+        _interpolate_bads_nirs(self)
 
         if reset_bads is True:
             self.info['bads'] = []
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -962,7 +962,7 @@ class InterpolationMixin(object):
 
     @verbose
     def interpolate_bads(self, reset_bads=True, mode='accurate',
-                         origin='auto', verbose=None):
+                         fnirs_method='nearest', origin='auto', verbose=None):
         """Interpolate bad MEG and EEG channels.
 
         Operates in place.
@@ -975,6 +975,9 @@ class InterpolationMixin(object):
             Either ``'accurate'`` or ``'fast'``, determines the quality of the
             Legendre polynomial expansion used for interpolation of MEG
             channels.
+        fnirs_method : str
+            Method to be used for fNIRS interpolation. Currently only 'nearest'
+            is supported.
         origin : array-like, shape (3,) | str
             Origin of the sphere in the head coordinate frame and in meters.
             Can be ``'auto'`` (default), which means a head-digitization-based
@@ -993,7 +996,8 @@ class InterpolationMixin(object):
         .. versionadded:: 0.9.0
         """
         from ..bem import _check_origin
-        from .interpolation import _interpolate_bads_eeg, _interpolate_bads_meg
+        from .interpolation import _interpolate_bads_eeg,\
+            _interpolate_bads_meg, _interpolate_bads_nirs
 
         _check_preload(self, "interpolation")
 
@@ -1003,6 +1007,7 @@ class InterpolationMixin(object):
         origin = _check_origin(origin, self.info)
         _interpolate_bads_eeg(self, origin=origin)
         _interpolate_bads_meg(self, mode=mode, origin=origin)
+        _interpolate_bads_nirs(self, method=fnirs_method)
 
         if reset_bads is True:
             self.info['bads'] = []
--- a/mne/channels/interpolation.py
+++ b/mne/channels/interpolation.py
@@ -248,12 +248,13 @@ def _interpolate_bads_nirs(inst, method='nearest', verbose=None):
 
     chs = [inst.info['chs'][i] for i in picks_nirs]
     locs3d = np.array([ch['loc'][:3] for ch in chs])
-    dist = pdist(locs3d)
-    dist = squareform(dist)
 
     _check_option('fnirs_method', method, ['nearest'])
 
     if method == 'nearest':
+
+        dist = pdist(locs3d)
+        dist = squareform(dist)
 
         for bad in picks_bad:
             dists_to_bad = dist[bad]
--- a/mne/channels/interpolation.py
+++ b/mne/channels/interpolation.py
@@ -11,6 +11,7 @@ from ..utils import logger, warn, verbose
 from ..io.pick import pick_types, pick_channels, pick_info
 from ..surface import _normalize_vectors
 from ..forward import _map_meg_channels
+from ..utils import _check_option
 
 
 def _calc_h(cosang, stiffness=4, n_legendre_terms=50):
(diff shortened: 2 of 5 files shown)


larsoner:
Is there an established correct way to do this?

If not, I wonder if the spherical spline interpolation used in EEG would be good enough. Maybe the thing to do is to get a clean dataset that's good for testing, e.g., with five sensors arranged in an X pattern (maybe you already have this?). Then you could try interpolating the middle sensor just by taking the mean of the other four, or by using the edges of a Delaunay triangulation, or by spherical spline interpolation (what we use for EEG) and see which one explains the most variance of the original sensor data. If they are all close then we can just use the EEG code path.

rob-luke:
> Is there an established correct way to do this?

Absolutely not. You are massively overestimating the state of the field. In most of what I read people discard channels with bad scalp coupling index then do not discuss the impact of this. From talking to people, most studies create a grand average with different number of averages per channel (depending on how many channels were discarded).

What prompted me to write this code is that in the group analysis MNE will take only the channels that have no individual recording marked as bad. It is more common to have poor channels in NIRS than EEG, as the optodes don't have that nice pin like ending to get through peoples hair that EEG has. So when analysing a group of 10 recordings, you end up with no channels left.

> Is there an established correct way to do this?

Wonderful idea. I was actually hoping you would come up with a better solution than mine, and that's why I included the `method` parameter in the function definition. I can take a recording like this in the next few weeks.

This code to find the closest channel will also be useful when we write the short channel correction code, which in the basic approaches uses the closest short channel to each long channel to remove systemic responses.

@larsoner so do we keep this `closest` approach and then add a spline interpolation method later? Or hold off on this until I have time to collect the data and look at the spline method?

larsoner:
> so do we keep this closest approach and then add a spline interpolation method later? Or hold off on this until I have time to collect the data and look at the spline method?

Let's wait until 0.20 is merged to merge any PR that implements a method, then we have ~6 mo to decide what the default should be without needing a deprecation cycle. I'm then okay with merging `nearest` to start and keep PRs manageable.

rob-luke:
Great. Sounds like a plan. I will keep working on this and ping you when it needs another look. Thanks

rob-luke:
@larsoner can you take another look at merging this now that 0.2 is released. Thanks

mne/channels/interpolation.py
```diff
@@ -213,3 +213,62 @@ def _interpolate_bads_meg(inst, mode='accurate', origin=(0., 0., 0.04),
...
+            closest_idx = np.where(dists_to_bad == np.amin(dists_to_bad))[0]
+            # Return the same frequency
+            closest_idx = closest_idx[bad % 2]
+            inst._data[bad] = inst._data[closest_idx]
+
+        inst.info['bads'] = []
+
+    else:
+        warn('No interpolation applied. Unknown NIRS interpolation '
+             'method: ' + method)
```
larsoner, line 272: This should be an error instead of a warning. You can just replace it with (before the conditional):
```
_check_option('method', method, ['nearest'])
```

mne/channels/interpolation.py
```diff
@@ -213,3 +213,62 @@ def _interpolate_bads_meg(inst, mode='accurate', origin=(0., 0., 0.04),
     info_to = pick_info(inst.info, picks_bad)
     mapping = _map_meg_channels(info_from, info_to, mode=mode, origin=origin)
     _do_interp_dots(inst, mapping, picks_good, picks_bad)
+
+
+@verbose
+def _interpolate_bads_nirs(inst, method='nearest', verbose=None):
```
larsoner, line 219: This should be added to `InterpolateMixin.interpolate_bads` in `mne/channels/channels.py`. You can add a `fnirs_method` kwarg to the function.

mne/channels/tests/test_interpolation.py
```diff
@@ -219,4 +219,23 @@ def test_interpolation_ctf_comp():
...
+    from itertools import compress
+    from mne.channels.interpolation import _interpolate_bads_nirs
+
+    fname = op.join(data_path(download=False),
+                    'NIRx', 'nirx_15_2_recording_w_overlap')
+    raw_intensity = read_raw_nirx(fname, preload=False)
+    raw_od = optical_density(raw_intensity)
+    sci = scalp_coupling_index(raw_od)
+    raw_od.info['bads'] = list(compress(raw_od.ch_names, sci < 0.5))
+    _interpolate_bads_nirs(raw_od)
```
larsoner, line 237: Then this can just be `raw_od.interpolate_bads()`

rob-luke:
@larsoner could you please review again. I believe I have made the changes you requested. And the CI failure seems unrelated

mne/channels/interpolation.py
```diff
@@ -213,3 +214,59 @@ def _interpolate_bads_meg(inst, mode='accurate', origin=(0., 0., 0.04),
...
+        return
+    picks_bad = pick_channels(inst.info['ch_names'], bads_nirs, exclude=[])
+    bads_mask = [p in picks_bad for p in picks_nirs]
+
+    chs = [inst.info['chs'][i] for i in picks_nirs]
+    locs3d = np.array([ch['loc'][:3] for ch in chs])
+    dist = pdist(locs3d)
+    dist = squareform(dist)
+
+    _check_option('method', method, ['nearest'])
```
larsoner, line 254: one last tweak, the user-facing parameter is `fnirs_method` so it should be `'fnirs_method'` here

rob-luke:
@larsoner my attempt at git magic has removed your commit!!! 😢 

The code change you made is still there, but you credit via commit history has disappeared.

mne/channels/interpolation.py
```diff
@@ -213,3 +214,59 @@ def _interpolate_bads_meg(inst, mode='accurate', origin=(0., 0., 0.04),
...
+    if method == 'nearest':
+
+        for bad in picks_bad:
+            dists_to_bad = dist[bad]
+            # Ignore distances to self
+            dists_to_bad[dists_to_bad == 0] = np.inf
+            # Ignore distances to other bad channels
+            dists_to_bad[bads_mask] = np.inf
+            # Find closest remaining channels
+            closest_idx = np.where(dists_to_bad == np.amin(dists_to_bad))[0]
```
larsoner, line 265: `closest_idx = np.argmin(dists_to_bad)`

mne/channels/tests/test_interpolation.py
```diff
@@ -219,4 +219,22 @@ def test_interpolation_ctf_comp():
     assert raw.info['bads'] == []
 
 
+@testing.requires_testing_data
+def test_interpolation_nirs():
+    """Test interpolating bad nirs channels."""
+    from mne.preprocessing.nirs import optical_density, scalp_coupling_index
+    from mne.datasets.testing import data_path
+    from mne.io import read_raw_nirx
+    from itertools import compress
```
larsoner, line 228: You don't need to nest imports in `test_` files

rob-luke:
Thanks I will fix these here

rob-luke:
@larsoner and @agramfort could you please review. I believe the failing test is unrelated.

mne/channels/channels.py
```diff
@@ -975,6 +975,9 @@ def interpolate_bads(self, reset_bads=True, mode='accurate',
             Either ``'accurate'`` or ``'fast'``, determines the quality of the
             Legendre polynomial expansion used for interpolation of MEG
             channels.
+        fnirs_method : str
+            Method to be used for fNIRS interpolation. Currently only 'nearest'
+            is supported.
```
agramfort, line 980: I would not expose this parameter at this stage. Interpolation is done differently for each channel type now.

mne/channels/interpolation.py
```diff
@@ -213,3 +214,57 @@ def _interpolate_bads_meg(inst, mode='accurate', origin=(0., 0., 0.04),
...
+    bads_nirs = [ch for ch in inst.info['bads'] if ch in nirs_ch_names]
+    if len(bads_nirs) == 0:
+        return
+    picks_bad = pick_channels(inst.info['ch_names'], bads_nirs, exclude=[])
+    bads_mask = [p in picks_bad for p in picks_nirs]
+
+    chs = [inst.info['chs'][i] for i in picks_nirs]
+    locs3d = np.array([ch['loc'][:3] for ch in chs])
+    dist = pdist(locs3d)
+    dist = squareform(dist)
```
agramfort, line 252: I would compute dist in the if block as it's really useful for nearest.

mne/channels/tests/test_interpolation.py
```diff
@@ -219,4 +223,17 @@ def test_interpolation_ctf_comp():
...
+def test_interpolation_nirs():
+    """Test interpolating bad nirs channels."""
+    fname = op.join(data_path(download=False),
+                    'NIRx', 'nirx_15_2_recording_w_overlap')
+    raw_intensity = read_raw_nirx(fname, preload=False)
+    raw_od = optical_density(raw_intensity)
+    sci = scalp_coupling_index(raw_od)
+    raw_od.info['bads'] = list(compress(raw_od.ch_names, sci < 0.5))
+    raw_od.interpolate_bads()
+    assert raw_od.info['bads'] == []
```
agramfort, line 236: can you test the values in some way?

rob-luke:
Thanks for the feedback @agramfort

larsoner:
Thanks @rob-luke

PR #7431, opened 2020-03-12
Make current time window more obvious

cbrnr:
It is hard to see the location of the currently visible time segment in the overview bar (x scroll bar), so I increased the line width.

Before:
before

After:
after


larsoner:
Thanks @cbrnr

PR #7432, opened 2020-03-12
WIP, FIX: Offload filtering step for find_bad_channels_maxwell

larsoner:
Moves the filtering step out of `find_bad_channels_maxwell`. This makes our outputs differ from MaxFilter a bit more, but:

1. It is better in principle (and practice) to filter all the data, then process in chunks, rather than filter each chunk independently (more edge artifacts)
2. By moving the filtering out of the function, it allows the user to choose the filtering tradeoffs
3. Based on inspecting the results with @hoechenberger , it seems like we do at least as well on this PR as we do in `master`

This PR also make a couple of tweaks found while working on this:

1. Improves `verbose=True` and `verbose='debug'` output messaging
2. Cleans up how `raw.plot` filtering triages FIR vs IIR to match what `mne/filter.py` does (leads to better `padlen` in `raw.plot`)
3. Cleans up unused var in `_get_lims_cola`

WIP for now while @hoechenberger and I test a bit more.

### DIFF
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -205,7 +205,8 @@ def maxwell_filter(raw, origin='auto', int_order=8, ext_order=3,
            Physics in Medicine and Biology, vol. 51, pp. 1759-1768, 2006.
            https://doi.org/10.1088/0031-9155/51/7/008
     """  # noqa: E501
-    return _maxwell_filter(
+    logger.info('Maxwell filtering raw data')
+    params, update_kwargs = _prep_maxwell_filter(
         raw=raw, origin=origin, int_order=int_order, ext_order=ext_order,
         calibration=calibration, cross_talk=cross_talk,
         st_duration=st_duration, st_correlation=st_correlation,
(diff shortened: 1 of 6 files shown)


mne/preprocessing/maxwell.py
```diff
@@ -224,7 +224,7 @@ def _maxwell_filter(raw, origin='auto', int_order=8, ext_order=3,
                     head_pos=None, st_fixed=True, st_only=False,
                     mag_scale=100.,
                     skip_by_annotation=('edge', 'bad_acq_skip'),
-                    reconstruct='in', verbose=None):
+                    reconstruct='in', proc_msg=True, verbose=None):
```
agramfort: what's the difference between proc_msg and verbose?

larsoner: Sometimes you want some but not all messages. `proc_msg` makes it so that some only get displayed when `verbose` is sufficient *and* `proc_msg` is True.

agramfort: you'll need to add this to a docstring somewhere?

larsoner: no, it's in the private `_maxwell_filter` / just for internal use

larsoner:
WIP because after #7434 is merged it will be good to optimize this function, it's probably around 2x slower than it needs to be due to recalculated quantities :(

larsoner:
Whoops, merged the wrong PR, will roll back...

agramfort:
create new PR with these commits?

PR #7435, opened 2020-03-12
handle _Brain size gracefully with pyvista

### LINKED ISSUE #7426: pyvista _Brain init too stringent about size parameter
I'm hitting the error `'"size" parameter must be int or tuple.'` I'm generating lots of similar figures, so I store several project-level parameters (including brain plotting kwargs like figure size) in an external YAML file, and pyYAML's "safe loader" doesn't support tuples, only lists.  Instead of checking `isinstance(size, tuple)` maybe we could check if it is length-2, and then convert it to tuple?  Or do it however the mayavi code did it; it works to pass `size=[1200, 1200]` with the mayavi backend.

@GuillaumeFavelier since you're the PyVista master I wanted to run this by you before just making the change... but I'm happy to do it myself if you're OK with it and you're too busy.

drammock:
closes #7426 

Also removes `u"` at beginning of docstrings in the edited file.

It turns out that `pysurfer.Brain` doesn't really check size at all:

https://github.com/nipy/PySurfer/blob/master/surfer/viz.py#L557-L561

So I just did what seemed sensible to me.

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -169,13 +169,10 @@ class _Brain(object):
         n_col = col_dict[hemi]
 
         size = tuple(np.atleast_1d(size).round(0).astype(int).flat)
-        if len(size) == 1:
-            fig_size = size * 2  # 1-tuple to 2-tuple
-        elif len(size) == 2:
-            fig_size = size
-        else:
+        if len(size) not in (1, 2):
             raise ValueError('"size" parameter must be an int or length-2 '
                              'sequence of ints.')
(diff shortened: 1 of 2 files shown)


GuillaumeFavelier:
Any idea why the CIs are not happy?

GuillaumeFavelier:
Nevermind. This has been reported in https://github.com/Teemu/pytest-sugar/issues/187

drammock:
@GuillaumeFavelier this has been rebased with the pytest-sugar fix, ready to go when CIs are green.

GuillaumeFavelier:
Thanks @drammock

PR #7439, opened 2020-03-13
MRG, DOC: revise evoked visualization tutorial

drammock:


tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,247 @@
...
+   :local:
+   :depth: 2
+
+As usual we'll start by importing the modules we need:
+"""
+
+import os
+import numpy as np
+import mne
+from mayavi import mlab
```
agramfort: please don't use mayavi in the examples. We've tried to get rid of it everywhere with @GuillaumeFavelier

tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,247 @@
...
+grads = evks['aud/right'].copy().pick_types(meg='grad')
+eegs = evks['aud/right'].copy().pick_types(meg=False, eeg=True)
+sensor_dict = dict(magnetometers=mags, gradiometers=grads, EEG=eegs)
+
+for title, evk in sensor_dict.items():
+    _map = mne.make_field_map(evk, trans=sample_data_trans_file,
+                              subject='sample', subjects_dir=subjects_dir,
+                              meg_surf='head')
+    fig = evk.plot_field(_map, time=0.1)
+    mlab.title(title)
```
agramfort: you can use ` mne.viz.set_3d_title`

drammock: we'll see if it works on Circle...  locally `mne.viz.set_3d_title` silently fails to change anything (the window title still says "mayavi scene 1" and no text is added to the scene)

agramfort:
ping @GuillaumeFavelier if needs be

>

drammock:
@GuillaumeFavelier  it works if I explicitly call `mne.viz.set_3d_backend('pyvista')` first, but if I don't do that, it defaults to using mayavi, and `set_3d_title` does nothing.

drammock:
rendering: https://18583-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_20_visualize_evoked.html

ready for review/merge.

tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,249 @@
...
+
+###############################################################################
+# To make our life easier, let's convert that list of :class:`~mne.Evoked`
+# objects into a :class:`dictionary <dict>` (make sure the order is correct!).
+# We'll use ``/``-separated dictionary keys to encode the conditions, like is
+# often done when epoching, because some of the plotting methods can take
+# advantage of that style of coding.
+
+conds = ('aud/left', 'aud/right', 'vis/left', 'vis/right')
+evks = dict(zip(conds, evokeds_list))
```
agramfort: this is a neat syntax but I fear it's black magic for new comers and this tuto is a pretty entry level one.
2 options  write something more naive but more explicit or keep it like this but explain with very stupid code what it does.
Eg saying that it does

`evks` = {'aud/left': evokeds_list[0], 'aud/right': evokeds_list[1], ...}`

tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,249 @@
...
+
+###############################################################################
+# Plotting signal traces
+# ^^^^^^^^^^^^^^^^^^^^^^
+#
+# The most basic plot of :class:`~mne.Evoked` objects is a butterfly plot of
```
agramfort: maybe you can add a note why it's called a butterfly plot.

```
.. note:: Such plot is called a butterfly plot because a source in the brain generates
simultaneously positive and negative values on sensors so the lines deviate positively
and negatively producing what can appear as wings of a butterfly.
```

tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,249 @@
...
+###############################################################################
+# Additional examples of plotting scalp topographies can be found in
+# :ref:`ex-evoked-topomap`.
+#
+#
+# Current flow maps
+# ^^^^^^^^^^^^^^^^^
+#
+# Scalp topographies at a given time point can be augmented with arrows to show
+# the estimated direction of current flow, using the function
```
agramfort: it's not current flow but orientation of the magnetic field which is a vector field.

tutorials/evoked/plot_20_visualize_evoked.py
```diff
@@ -0,0 +1,249 @@
...
+# You can also use MEG sensors to estimate the *scalp* field by passing
+# ``meg_surf='head'``. By selecting each sensor type in turn, you can compare
+# the scalp field estimates from each.
+
+mags = evks['aud/right'].copy().pick_types(meg='mag')
+grads = evks['aud/right'].copy().pick_types(meg='grad')
+eegs = evks['aud/right'].copy().pick_types(meg=False, eeg=True)
+sensor_dict = dict(magnetometers=mags, gradiometers=grads, EEG=eegs)
+
+for title, evk in sensor_dict.items():
```
agramfort: I woudl avoid the definition of mags, grads etc. variables and do:

```
for ch_type in ['mag', 'grad', 'eeg']:
    evk = evks['aud/right'].copy().pick(ch_type)
   ... 
```

then you can use ch_type as title.

agramfort:
circle is green. Other failures are unrelated.

PR #7440, opened 2020-03-13
DOC: Fix typo

zuxfoucault:
A few "beamformer" typos

### DIFF
--- a/examples/inverse/plot_dics_source_power.py
+++ b/examples/inverse/plot_dics_source_power.py
@@ -1,9 +1,9 @@
 """
 .. _ex-inverse-source-power:
 
-=========================================
+==========================================
 Compute source power using DICS beamformer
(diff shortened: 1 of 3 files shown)


examples/inverse/plot_dics_source_power.py
```diff
@@ -2,7 +2,7 @@
 .. _ex-inverse-source-power:
 
 =========================================
-Compute source power using DICS beamfomer
+Compute source power using DICS beamformer
 =========================================
```
agramfort: @zuxfoucault you need to change too the length of the lines with ===

zuxfoucault: Done! Sorry I didn't notice that.

larsoner:
Thanks @zuxfoucault

PR #7441, opened 2020-03-13
Shorten repr of Annotations if necessary

### LINKED ISSUE #7429: Annotations repr too long
I don't like that the `repr` for an `Annotations` object exceeds 79 characters. For example, the following repr has 94 characters:

```
<Annotations  |  30 segments : T0 (15), T2 (7), T1 (8), orig_time : 2009-08-12 16:15:00+00:00>
```

It would be nice if the maximum character length of this (and any other) repr was 79, otherwise these get hard to read because of line breaks.

In the specific case of `Annotations`, I would remove the `orig_time` information because I don't think this is important enough to show in the repr.

cbrnr:
Fixes #7429.

Before (94 characters):
```
<Annotations  |  14 segments : a (1), b (1), c (1)..., orig_time : 2013-11-04 14:35:26.150000>
```

After (76 characters):
```
<Annotations | 14 segments: a (1), b (1), c (1), d (1), e (1), f (1), g ...>
```

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -206,7 +206,7 @@ class Annotations(object):
     def __repr__(self):
         """Show the representation."""
         counter = collections.Counter(self.description)
-        kinds = ', '.join(['%s (%s)' % k for k in counter.items()])
+        kinds = ', '.join(['%s (%s)' % k for k in sorted(counter.items())])
         kinds = (': ' if len(kinds) > 0 else '') + kinds
         s = ('Annotations  |  %s segment%s%s' %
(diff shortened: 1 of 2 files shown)


cbrnr:
Python 3.5 build fails because I'm relying on the behavior that dicts retain insertion order. Should I sort the dict keys?

Also, when do we require Python 3.6 (can't wait to finally use f-strings here)?

agramfort:
Please makes sure it works with 3.5. Thanks

agramfort:
thx @cbrnr

Issue #7442, opened 2020-03-13
ENH: Multi-line evoked topomap plot


fraimondo:
#### Describe the problem
When I analyze some ERPs, I want to plot evoked topomaps every few ms. The current maximum is 20, and they are plot in one single line, spanning all across the screen.

#### Describe your solution
It might be a good idea to plot them in several rows. Like this:

![Screenshot 2020-03-12 at 19 56 13](https://user-images.githubusercontent.com/4493699/76605880-8812bc00-6511-11ea-8fc1-8592a57eb7d0.png)

#### Additional context
I've been doing this for years by keeping a copy of the `plot_evoked_topomap` function with an extra parameter `max_cols`. I'm now tired of updating this function every time that MNE updates, so if you think its a good idea, I can do the PR.

Here's the proposed modification: https://github.com/fraimondo/mne-python/commit/d57e486151976620ffb82609f00affde4b39a4af

agramfort:
I would then mimic the API names of plt.subplots ie nrows, ncols
making nrows=1 the default and letting ncols default to 'auto' being
np.ceil(n_times // nrows) or something like that

fraimondo:
So I modified it an added `nrows` and `ncols` with the posibility of using `'auto'` in at least one of them: https://github.com/fraimondo/mne-python/commit/38f4b1cc20963b1593210d6f188e2086caaae49c

The only doubt I still have, is that I had to remove the limit of 20 times in `mne.viz.utils._process_times`.

Is it a desired feature? Shall I proceed with the PR? Maybe modify the test `test_plot_topomap_basic`?

agramfort:
Please do it with a PR so can comment on the diff

fraimondo:
done!

PR #7444, opened 2020-03-13
Multiline evoked topomap plot

fraimondo:
#### Reference issue
Implements #7442 

Adds `nrows` and `ncols` parameter to `plot_evoked_topomap` to allow multiline evoked topomaps plots.

This is the result with `ncols=10` and `nrows='auto'`:
Screenshot 2020-03-13 at 14 15 27

### DIFF
--- a/mne/viz/tests/test_topomap.py
+++ b/mne/viz/tests/test_topomap.py
@@ -329,6 +329,9 @@ def test_plot_topomap_basic():
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
+    with pytest.warns(RuntimeWarning, match='More than 25 topomaps plots'):
+        plot_evoked_topomap(evoked, [0.1] * 26, res=4)
+
     pytest.raises(ValueError, plot_evoked_topomap, evoked, [-3e12, 15e6],
                   time_unit='s')
 
--- a/mne/viz/tests/test_topomap.py
+++ b/mne/viz/tests/test_topomap.py
@@ -32,7 +32,7 @@ from mne.time_frequency.tfr import AverageTFR
 from mne.utils import run_tests_if_main
 from mne.datasets.testing import data_path
 
-from mne.viz import plot_evoked_topomap, plot_projs_topomap
+from mne.viz import plot_evoked_topomap, plot_projs_topomap, topomap
 from mne.viz.topomap import (_get_pos_outlines, _onselect, plot_topomap,
                              plot_arrowmap, plot_psds_topomap)
 from mne.viz.utils import _find_peaks, _fake_click
@@ -150,7 +150,7 @@ def test_plot_topomap_animation_nirs(fnirs_evoked):
 
 
 @pytest.mark.slowtest
-def test_plot_topomap_basic():
+def test_plot_topomap_basic(monkeypatch):
     """Test basics of topomap plotting."""
     evoked = read_evokeds(evoked_fname, 'Left Auditory',
                           baseline=(None, 0))
@@ -329,8 +329,11 @@ def test_plot_topomap_basic():
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
-    with pytest.warns(RuntimeWarning, match='More than 25 topomaps plots'):
-        plot_evoked_topomap(evoked, [0.1] * 26, res=4)
+    with monkeypatch.context() as m:  # speed it up by not actually plotting
+        m.setattr(topomap, '_plot_topomap',
+                  lambda *args, **kwargs: (None, None, None))
+        with pytest.warns(RuntimeWarning, match='More than 25 topomaps plots'):
+            plot_evoked_topomap(evoked, [0.1] * 26, colorbar=False)
 
     pytest.raises(ValueError, plot_evoked_topomap, evoked, [-3e12, 15e6],
                   time_unit='s')
--- a/mne/viz/tests/test_topomap.py
+++ b/mne/viz/tests/test_topomap.py
@@ -329,8 +329,6 @@ def test_plot_topomap_basic():
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
-    # pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
-    #               np.repeat(.1, 50), time_unit='s')
     pytest.raises(ValueError, plot_evoked_topomap, evoked, [-3e12, 15e6],
                   time_unit='s')
 
--- a/mne/viz/tests/test_topomap.py
+++ b/mne/viz/tests/test_topomap.py
@@ -329,8 +329,8 @@ def test_plot_topomap_basic():
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
-    pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
-                  np.repeat(.1, 50), time_unit='s')
+    # pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
+    #               np.repeat(.1, 50), time_unit='s')
     pytest.raises(ValueError, plot_evoked_topomap, evoked, [-3e12, 15e6],
                   time_unit='s')
 
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -1451,7 +1451,7 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None, layout=None,
                         mask_params=None, outlines='head', contours=6,
                         image_interp='bilinear', average=None, head_pos=None,
                         axes=None, extrapolate='box', sphere=None, border=0,
-                        max_cols=None):
+                        nrows=1, ncols='auto'):
     """Plot topographic maps of specific time points of evoked data.
 
     Parameters
@@ -1575,7 +1575,14 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None, layout=None,
         .. versionadded:: 0.18
     %(topomap_sphere_auto)s
     %(topomap_border)s
-
+    nrows : int | 'auto'
+        The number of rows of topographies to plot. Defaults to 1. If 'auto',
+        obtains the number of rows depending on the amount of times to plot
+        and the number of cols. Not valid when times == 'interactive'.
+    ncols : int | 'auto'
+        The number of columns of topographies to plot. If 'auto' (default),
+        obtains the number of columns depending on the amount of times to plot
+        and the number of rows. Not valid when times == 'interactive'.
     Returns
     -------
     fig : instance of matplotlib.figure.Figure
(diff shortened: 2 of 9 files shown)


agramfort:
can you update https://mne.tools/dev/auto_tutorials/evoked/plot_visualize_evoked.html#sphx-glr-auto-tutorials-evoked-plot-visualize-evoked-py ?

drammock:
Multiline topomap plotting is already implemented for ICA and SSP. Please check how we do it there, maybe there's already a helper function we could use. If not, this would be a good time to unify the layout code for all 3 uses.

Also, despite what @agramfort suggests, don't spend too much time modifying the "visualizing evoked" tutorial because there's an open PR to overhaul it (#7439). Just tweak one plot to test out the new code

agramfort:
yes don't listen to me  :)

>

fraimondo:
> Multiline topomap plotting is already implemented for ICA and SSP. Please check how we do it there, maybe there's already a helper function we could use. If not, this would be a good time to unify the layout code for all 3 uses.

Both SSP and ICA do not expose the parameter to the user:
SSP: https://github.com/mne-tools/mne-python/blob/41978e88e37b254f2adc3c92acf81f57bfc9e215/mne/viz/topomap.py#L353-L366

ICA (uses _prepare_trellis with `max_cols=5`): https://github.com/mne-tools/mne-python/blob/41978e88e37b254f2adc3c92acf81f57bfc9e215/mne/viz/utils.py#L450-L467

The code I'm proposing exposes both `ncols` and `nrows` to the user, with optional 'auto' parameters.

It's a bit tricky because of the colorbar. None of the other plots require/allows for a colorbar. And here there's the possibility of adding a colorbar.

I guess the best would be to use `prepare_trellis` and modify to allow two extra (optional) parameters: `nrows='auto'` and `colorbar=False`.

drammock:
> I guess the best would be to use `prepare_trellis` and modify to allow two extra (optional) parameters: `nrows='auto'` and `colorbar=False`.

Sounds good to me.

fraimondo:
This should do the trick

agramfort:
@fraimondo feel free to share code snippets to replicate some plots so we can more quickly evaluate the usability.

fraimondo:
I used this all the time to plot an MMN/P3a ERP like this:
Unknown

This is a woking snippet based on the [MNE example for plot_evoked_topomap](https://mne.tools/dev/auto_examples/visualization/plot_evoked_topomap.html#sphx-glr-auto-examples-visualization-plot-evoked-topomap-py):

```Python
# Authors: Christian Brodbeck <christianbrodbeck@nyu.edu>
#          Tal Linzen <linzen@nyu.edu>
#          Denis A. Engeman <denis.engemann@gmail.com>
#          Mikołaj Magnuski <mmagnuski@swps.edu.pl>
#
# License: BSD (3-clause)

import numpy as np

from mne import read_evokeds
from mne.viz import plot_evoked_topomap

print(__doc__)

path = sample.data_path()
fname = path + '/MEG/sample/sample_audvis-ave.fif'

# load evoked corresponding to a specific condition
# from the fif file and subtract baseline
condition = 'Left Auditory'
evoked = read_evokeds(fname, condition=condition, baseline=(None, 0))
times = np.arange(-0.2, 0.5, 0.02)
plot_evoked_topomap(evoked, times, ch_type='mag', time_unit='s', 
                    ncols=10, nrows='auto')
```

Screenshot 2020-03-15 at 17 16 57

drammock:
this is looking good. At least some of the travis failures seem related / legitimate though:

```
――――――――――――――――――――――――――― test_plot_topomap_basic ――――――――――――――――――――――――――――

mne/viz/tests/test_topomap.py:333: in test_plot_topomap_basic

    np.repeat(.1, 50), time_unit='s')

E   Failed: DID NOT RAISE <class 'RuntimeError'>
```

Remaining tasks seem to be:

- update tests so that any new or changed error messages are tested
- make at least one minor change (even just changing where a linebreak is) to any examples / tutorials where these topomaps get generated; that will cause the documentation CI to render those tutorials.
- change at least one of those examples / tutorials to use the new `nrows` / `ncols` parameters so we can see it working on the doc CI server.

fraimondo:
I'll tackle that tomorrow.

The issue with the error is that previously more than 20 topos would raise
an error. I removed that check. However, now users can plot 50 topos in one
row.

Shall I add some check with warning/error? Or just leave it like that and
remove the test?

agramfort:
preventing people from freezing their computer by plotting too many topos
remains for me a good idea.
20 topos in one row will not be readable anyway.

one solution is to prevent plotting more than 20 topos per row and then
make sure that nrows / ncols
is manually changed if more than 20 topos are requested

>

agramfort:
@fraimondo Circle failed https://circleci.com/gh/mne-tools/mne-python/18623?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

can you ping me when CIs are green?

mne/viz/topomap.py
```diff
@@ -1502,6 +1499,14 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None, layout=None,
...
     %(topomap_sphere_auto)s
     %(topomap_border)s
+    nrows : int | 'auto'
+        The number of rows of topographies to plot. Defaults to 1. If 'auto',
+        obtains the number of rows depending on the amount of times to plot
+        and the number of cols. Not valid when times == 'interactive'.
+    ncols : int | 'auto'
+        The number of columns of topographies to plot. If 'auto' (default),
+        obtains the number of columns depending on the amount of times to plot
+        and the number of rows. Not valid when times == 'interactive'.
```
larsoner: What happens if both are auto (needs to be documented)? If you haven't already decided, I would say in this case we could have `ncols` set to `min(20, len(times))` and `nrow` set to whatever it needs to be to plot everything.

Also this is missing `.. versionadded:: 0.20` directives

fraimondo: > What happens if both are auto (needs to be documented)? If you haven't already decided, I would say in this case we could have `ncols` set to `min(20, len(times))` and `nrow` set to whatever it needs to be to plot everything.

@larsoner Good point. Currently the behaviour is to raise an error. `plot_projs_topomap` has another behaviour (but does not expose the `nrows` and `ncols` parameters):
```Python
nrows = math.floor(math.sqrt(n_projs))
ncols = math.ceil(n_projs / nrows)
```

Shall I add `ncols = min(20, len(n_cells))` or use the same behaviour as `plot_projs_topomap`?

> Also this is missing `.. versionadded:: 0.20` directives
Just added it.

larsoner: I guess unifying with `plot_projs_topomap` (making it square) would be the best option. The difference would be that `plot_projs_topomap` would have `n_rows='auto'` and `evoked.plot_topomap` would have `n_rows=1` by default. The justification would be that it's a lot easier to accidentally plot hundreds of evoked topomaps and kill your machine than it is to do it with projectors (which are typically very limited in number).

fraimondo: I agree that it's easy to accidentally ask for too many plots. If the plot is not multiline and more than 20 timepoints are asked, then `_process_times` raises an error.

So in the case of `nrows=1` then the plot is not multiline and `_process_times` will raise the error. If you specify at least nrows=2, then you can still have 2 rows of 100 of plots.

The other way would be to restrict the number of columns to 20, which will also allow you to plot hundreds of topomaps. We can also restrict the number of rows to 20, giving a maximum of 400 topomaps in the worst case scenario.

Or we can do as matplotlib does, where more than 20 open figures raises a warning.

I can implement any of these 3 options. Not sure which one is the best.

larsoner: > The other way would be to restrict the number of columns to 20, which will also allow you to plot hundreds of topomaps. We can also restrict the number of rows to 20, giving a maximum of 400 topomaps in the worst case scenario.

Sounds reasonable to me

drammock: I don't like restricting the number of columns to 20.  What if someone wants to generate a landscape figure that is 25 columns x 10 rows, for example?  During exploratory analysis, or for a supplementary figure, this doesn't seem wildly unreasonable to me.

I think I prefer this: 

- `nrows=1, ncols='auto'` is the default
- `nrows='auto', ncols='auto'`: do the `sqrt` thing
- number of times > 25: warn, don't error (25 is the limit for the ICA topomap function to trigger a second, separate figure, that's why I picked 25)
- don't put any upper limit on the specific numbers passed to `nrows` or `ncols`

agramfort:
ok for me. I'll evaluate looking and playing with the evoked visualization
tutorial

>

fraimondo:
@agramfort I think the CIs are green or not related to this PR.

agramfort:
can you share a link to the rendered doc updated on circleci?

>

fraimondo:
Here it's the doc:

https://18687-1301584-gh.circle-artifacts.com/0/dev/generated/mne.viz.plot_evoked_topomap.html#mne.viz.plot_evoked_topomap

And here the example: 
https://18687-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_evoked_topomap.html#sphx-glr-auto-examples-visualization-plot-evoked-topomap-py

mne/viz/tests/test_topomap.py
```diff
@@ -326,8 +326,8 @@ def get_texts(p):
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
-    pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
-                  np.repeat(.1, 50), time_unit='s')
+    # pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
+    #               np.repeat(.1, 50), time_unit='s')
```
agramfort: cleanup

agramfort:
you should also update latests.inc file

drammock:
circle error

```
changes/latest.inc:142: WARNING: Unknown target name: "fede raimondo".
```
You need to add yourself to `doc/changes/names.inc`

fraimondo:
> You need to add yourself to `doc/changes/names.inc`

Thanks! I could not find the problem.

mne/viz/tests/test_topomap.py
```diff
@@ -329,8 +329,6 @@ def get_texts(p):
     # make sure projector gets toggled
     assert (np.max(fig1.axes[0].images[0]._A) != data_max)
 
-    pytest.raises(RuntimeError, plot_evoked_topomap, evoked,
-                  np.repeat(.1, 50), time_unit='s')
```
larsoner: Rather than eliminate this entirely, it's a warning now, right? So you could do:
```
with pytest.warns(RuntimeWarning, match='some regex substring to match'):
    plot_evoked_topomap(evoked, [0.1] * 26, res=4)
```
But if it adds a bunch of time to the test (please do check) then it's not worth it just to catch the one warning line

fraimondo: Just added that.

larsoner: It added 0.5 sec to the test which is pretty rough for something so simple to test (our tests are already quite slow). Every little bit helps, so I pushed a commit using `monkeypatch` to speed things up.

larsoner:
Thanks @fraimondo !

fraimondo:
Always a pleasure. Keep up the good work!

PR #7438, opened 2020-03-12
MNT: Exclude pytest 5.4.0

GuillaumeFavelier:
This PR skips `pytest 5.4.0` because of the issues with `pytest-sugar`.

Closes https://github.com/mne-tools/mne-python/issues/7437

PR #7445, opened 2020-03-13
MNT: Freeze pytest before 5.4.0

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/7438, `pytest 5.4.1` released a few hours ago, I don't think we can just use `!=5.4.0` anymore.

Reference:
https://github.com/mne-tools/mne-python/pull/7380#issuecomment-598781443

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ before_install:
         fi;
       fi
     # Always install these via pip so we get the latest possible versions (testing bugfixes)
-    - pip install --upgrade pytest!=5.4.0 pytest-sugar pytest-cov pytest-mock pytest-timeout codecov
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -66,7 +66,7 @@ before_install:
         fi;
       fi
     # Always install these via pip so we get the latest possible versions (testing bugfixes)
-    - pip install --upgrade pytest!=5.4.0 pytest-sugar pytest-cov pytest-mock pytest-timeout codecov
+    - pip install --upgrade pytest<5.4.0 pytest-sugar pytest-cov pytest-mock pytest-timeout codecov
```
cbrnr: I think you can't use `<` in a shell command, so you need to put this package in quotes. I've already done that in #7446.

PR #7448, opened 2020-03-13
MRG, ENH: Raw data reading refactor

### LINKED ISSUE #7436: Reading single epochs from a fif file
I noticed that reading single epochs is significantly slower when the raw data is saved as a fif file vs. an edf file. Also, the reading time from fif files linearly increases with how far in the recording the epochs are, which is not the case with edf files.

I'm interested in getting this reading time as low as possible since I have to repeatedly load single epochs. I was wondering whether this difference between fif and edf is expected (maybe part of it is due to the lower precision of edf files).

#### Steps to reproduce
I created a gist that creates .fif, .edf and .hdf5 files out of the same fake raw data, and then times the loading of each individual epoch: https://gist.github.com/hubertjb/1b434688dba296d25eba8fab2866a50f

I included direct reading from an hdf5 file as a baseline. 

#### Expected results
I was expecting fif and edf to yield similar results. Specifically, I was expecting the loading time to be more or less constant across the recording for fif. 

#### Actual results
As shown below, reading from fif is slower than from edf, and reading times increase as the epochs get further from the beginning of the recording.
image

Mean loading times are
{'edf': 0.0028056514029409373,
 'fif': 0.008060473086787204,
 'hdf5': 0.0009250746053807876}

#### Additional information
Platform:      Linux-4.15.0-70-generic-x86_64-with-debian-buster-sid
Python:        3.7.6 (default, Jan  8 2020, 19:59:22)  [GCC 7.3.0]
Executable:    /storage/store/work/hjacobba/miniconda3/envs/braindecode/bin/python
CPU:           x86_64: 72 cores
Memory:        376.5 GB

mne:           0.20.dev0
numpy:         1.18.1 {blas=openblas, lapack=openblas}
scipy:         1.4.1
matplotlib:    3.1.3 {backend=module://ipykernel.pylab.backend_inline}

sklearn:       0.22.1
numba:         0.48.0
nibabel:       Not found
cupy:          Not found
pandas:        1.0.1
dipy:          Not found
mayavi:        Not found
pyvista:       Not found
vtk:           Not found

@agramfort @larsoner

larsoner:
In `mne/io/fiff/raw.py` we iterate over tags to figure out which ones to read. This will slow down as you access data later and later in the file, so a better strategy is to construct arrays of starts/stops and then figure out which tags to read based on that (a simple `searchsorted` or even `np.where` will yield the answers). 

First step is then to refactor the reading function to "reformat raw_extras to be a dict of list/ndarray rather than list of dict (faster access)", which is done here. Hopefully all is green and we can move on to the speedup part.

At least `pytest mne/io` works locally...

- [x] Refactor list of dict to dict of ndarray
- [x] Refactor reading function to use start and stop points
- [x] @hubertjb tests speed

This should in the end be pretty straightforward so we can get it into 0.20.

Closes #7436

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -91,6 +91,8 @@ Changelog
 
 - Speed up :func:`mne.make_forward_solution` using Numba, by `Eric Larson`_
 
+- Speed up :func:`mne.io.read_raw_fif` data reading when the recording is long and there are many data tags repeatedly accessed, by `Eric Larson`_
+
 - For KIT systems without built-in layout, :func:`mne.channels.find_layout` now falls back on an automatically generated layout, by `Christian Brodbeck`_
 
 - :meth:`mne.Epochs.plot` now takes a ``epochs_colors`` parameter to color specific epoch segments by `Mainak Jas`_
--- a/mne/io/fiff/raw.py
+++ b/mne/io/fiff/raw.py
@@ -102,10 +102,10 @@ class Raw(BaseRaw):
 
         # Add annotations for in-data skips
         for extra in self._raw_extras:
-            start = np.array([e['first'] for e in extra if e['ent'] is None])
-            stop = np.array([e['last'] for e in extra if e['ent'] is None])
+            mask = [ent is None for ent in extra['ent']]
+            start = extra['first'][mask]
+            stop = extra['last'][mask]
             duration = (stop - start + 1.) / self.info['sfreq']
-
             annot = Annotations(onset=(start / self.info['sfreq']),
                                 duration=duration,
                                 description='BAD_ACQ_SKIP',
@@ -263,6 +263,14 @@ class Raw(BaseRaw):
 
             next_fname = _get_next_fname(fid, fname_rep, tree)
 
+        # reformat raw_extras to be a dict of list/ndarray rather than
+        # list of dict (faster access)
+        raw_extras = {key: [r[key] for r in raw_extras]
+                      for key in raw_extras[0]}
+        for key in raw_extras:
+            if key != 'ent':  # dict or None
+                raw_extras[key] = np.array(raw_extras[key], int)
+
         raw.last_samp = first_samp - 1
         raw.orig_format = orig_format
 
@@ -279,8 +287,7 @@ class Raw(BaseRaw):
                     float(raw.last_samp) / info['sfreq']))
 
         # store the original buffer size
-        buffer_size_sec = np.median(
-            [r['nsamp'] for r in raw_extras]) / info['sfreq']
+        buffer_size_sec = np.median(raw_extras['nsamp']) / info['sfreq']
 
         raw.info = info
         raw.verbose = verbose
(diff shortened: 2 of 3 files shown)


larsoner:
@hubertjb can you try your script again on this branch?

larsoner:
On `master` with @hubertjb's script I see:

Figure_1


And on this PR:

Figure_2

This is the resulting profile, so I don't think there is much more we can do:

<details>

```
File: /home/larsoner/python/mne-python/mne/io/fiff/raw.py
Function: _read_segment_file at line 335

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   335                                               def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
   336                                                   """Read a segment of data from a file."""
   337      2040      45723.0     22.4      1.3          with _fiff_get_fid(self._filenames[fi]) as fid:
   338      2040       2345.0      1.1      0.1              bounds = self._raw_extras[fi]['bounds']
   339      2040       1102.0      0.5      0.0              ents = self._raw_extras[fi]['ent']
   340      2040      49962.0     24.5      1.4              use = (stop > bounds[:-1]) & (start < bounds[1:])
   341      2040       1132.0      0.6      0.0              offset = 0
   342     63240      66049.0      1.0      1.8              for ei in np.where(use)[0]:
   343     61200      48496.0      0.8      1.3                  first = bounds[ei]
   344     61200      56473.0      0.9      1.6                  last = bounds[ei + 1]
   345     61200      30378.0      0.5      0.8                  nsamp = last - first
   346     61200      32890.0      0.5      0.9                  ent = ents[ei]
   347     61200      73540.0      1.2      2.0                  first_pick = max(start - first, 0)
   348     61200      54260.0      0.9      1.5                  last_pick = min(nsamp, stop - first)
   349     61200      38536.0      0.6      1.1                  picksamp = last_pick - first_pick
   350                                                           # only read data if it exists
   351     61200      29504.0      0.5      0.8                  if ent is not None:
   352    122400    1953383.0     16.0     53.9                      one = read_tag(fid, ent.pos,
   353     61200      37410.0      0.6      1.0                                     shape=(nsamp, self.info['nchan']),
   354     61200      26218.0      0.4      0.7                                     rlims=(first_pick, last_pick)).data
   355     61200      88811.0      1.5      2.5                      one.shape = (picksamp, self.info['nchan'])
   356    122400     886736.0      7.2     24.5                      _mult_cal_one(data[:, offset:(offset + picksamp)],
   357     61200      40100.0      0.7      1.1                                    one.T, idx, cals, mult)
   358     61200      50098.0      0.8      1.4                  offset += picksamp
   359      2040       9013.0      4.4      0.2              assert offset == stop - start
```

</details>

The only thing I can think of is *maybe* we can `numba.jit` some things to make them faster to avoid creating a bunch of Python objects, but this seems like a lot of work...

Ready for review/merge from my end.

hubertjb:
Thanks a lot @larsoner for looking into this so quickly! I've rerun the test script as well and got similar results to yours. The mean fif loading time went from 0.0081 to 0.0057 s.

PR #7456, opened 2020-03-16
ENH: Small mne.grand_average() changes

hoechenberger:
#### Reference issue
Required for mne-tools/mne-study-template#57

Raises when an empty sequence was passed, and emits a warning when a sequence of length 1 was passed.

Previous behavior was to always assert a length >1.

This is required to implement mne-tools/mne-study-template#57

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -135,6 +135,8 @@ Changelog
 
 - Support for saving movies of source time courses (STCs) with ``brain.save_movie`` method and from graphical user interface by `Guillaume Favelier`_
 
+- :func:`mne.grand_average` now produces a warning when only a single dataset was passed, instead of raising by `Richard Höchenberger`_
+
 Bug
 ~~~
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -135,7 +135,7 @@ Changelog
 
 - Support for saving movies of source time courses (STCs) with ``brain.save_movie`` method and from graphical user interface by `Guillaume Favelier`_
 
-- :func:`mne.grand_average` now produces a warning when only a single dataset was passed, instead of raising by `Richard Höchenberger`_
+- :func:`mne.grand_average` now produces a warning when only a single dataset was passed, instead of raising an error by `Richard Höchenberger`_
 
 Bug
 ~~~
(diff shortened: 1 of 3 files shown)


doc/changes/latest.inc
```diff
@@ -133,6 +133,8 @@ Changelog
 
 - Expose the number of ICA iterations during the fitting procedure via the ``n_iter_`` attribute of :class:`mne.preprocessing.ICA` by `Richard Höchenberger`_
 
+- :func:`mne.grand_average` now produces a warning when only a single dataset was passed, instead of raising by `Richard Höchenberger`_
```
agramfort:
```suggestion
- :func:`mne.grand_average` now produces a warning when only a single dataset was passed, instead of raising an error by `Richard Höchenberger`_
```

mne/utils/tests/test_numerics.py
```diff
@@ -465,3 +465,19 @@ def test_julian_conversions():
...
+
+def test_grand_average_len_1():
+    base_dir = op.join(op.dirname(__file__), '..', '..', 'io', 'tests',
+                       'data')
+    fname = op.join(base_dir, 'test-ave.fif')
+    evokeds = read_evokeds(fname, condition=[0], proj=True)  # list of length 0
+
+    with pytest.warns(RuntimeWarning, match='Only a single evoked'):
+        gave = grand_average(evokeds)
+        assert_allclose(gave.data, evokeds[0].data)
```
larsoner: Usually you put only what is necessary inside context managers, it makes clear what needs them. So this line should be unindented

mne/utils/tests/test_numerics.py
```diff
@@ -465,3 +465,19 @@ def test_julian_conversions():
...
+
+def test_grand_average_empty_sequence():
+    with pytest.raises(ValueError, match='Please pass a list of Evoked'):
+        grand_average([])
+
+
+def test_grand_average_len_1():
+    base_dir = op.join(op.dirname(__file__), '..', '..', 'io', 'tests',
+                       'data')
+    fname = op.join(base_dir, 'test-ave.fif')
```
larsoner: by convention these go at the top of the file. It helps with consistency and also helps us know what files are required by each test file (to some extent)

hoechenberger: Incidentally, the name of the file I need to access was already defined at the top level, so will use that variable now

hoechenberger:
Should be all good now!

hoechenberger:
Rebased and force-pushed to resolve the merge conflict.

PR #7416, opened 2020-03-10
MRG, MAINT: Work around PyVista rendering bug

larsoner:
Workaround for PyVista scraping. @GuillaumeFavelier feel free to look and decide if it needs a backport (not sure what the PyVista scraper duplication status is). All I really did was add the `if plotter in seen_plotters:` check and accumulation and it appears to fix things.

Alternative to #7398
Closes #7398
Closes #7228

PR #7458, opened 2020-03-16
WIP: Prepare _TimeViewer for next version of pyvista

GuillaumeFavelier:
This PR is a draft to test the next version of pyvista. This is still a work in progress.

⚠️ **Requires `pyvista 0.24`** ⚠️


GuillaumeFavelier:
Also the workaround for the scraper (https://github.com/mne-tools/mne-python/pull/7416) is not necessary anymore in this PR and should be removed.

larsoner:
Please do remove it here then

GuillaumeFavelier:
All green here and no timeout, I'll proceed with doc building.

GuillaumeFavelier:
Circle was successful too: https://18688-1301584-gh.circle-artifacts.com/0/dev/index.html

GuillaumeFavelier:
This is ready when we need it @agramfort, @larsoner

larsoner:
Tentatively marking for 0.20. It would be good if we could wait for their release, but let's see how long it takes them.

In the meantime @GuillaumeFavelier you have conflicts

GuillaumeFavelier:
@larsoner 0.24 is out since yesterday

GuillaumeFavelier:
I'll fix the conflicts

larsoner:
@GuillaumeFavelier should we merge then?

GuillaumeFavelier:
After giving it more thought, I think it's better to wait for 0.21 except if we really need it for a specific reason before the release.

The version 0.23.1 of pyvista has been tested extensively compared the latest one.

agramfort:
does it work fine for you? shall we try to test? we can have 3-4 people
testing quickly

>

GuillaumeFavelier:
Testing is always welcome for ` pyvista 0.24` and (maybe) `mne-python 0.21`. Let's see how it goes @agramfort :+1:

larsoner:
Agreed we should probably wait until 0.21 to advertise PyVista stuff. Having an extra 6 months of testing during which core users/contributors/maintainers, some of whom have asked about 3D viz semi-recently (@bloyl, @drammock) start actually using it in scripts would be beneficial in finding bugs before we force less experienced users to do it.

@GuillaumeFavelier are there `latest.inc` entries that need to be removed for now until 0.21 is released? Anything else we'd need to remove? Technically people can still do `MNE_3D_BACKEND=...` but if we don't advertise it I think it's okay to leave it as a "hidden" feature for 0.20.

GuillaumeFavelier:
AFAIK, the only entry about the new `_TimeViewer` is `link_brains`:

https://github.com/mne-tools/mne-python/blob/master/doc/changes/latest.inc#L130

agramfort:
I am a bit reluctant to wait to advertise pyvista. With 0.20 I would like
to have user feedback
and keep improving as bugs are reported. Users can still use mayavi if need
be. I don't think
we broke any mayavi code. So let's take the release opportunity to
advertise pyvista with nice
screenshots so we keep the ball rolling at a good speed.

my 2c

>

larsoner:
Pushed an empty commit to get Circle to build the frontpage examples so we can at least see if one plot works

larsoner:
Thanks @GuillaumeFavelier

PR #7459, opened 2020-03-16
MRG, ENH: Allow just line freq filtering

larsoner:
In principle we should be able to run `filter_chpi` to remove line frequencies even if cHPI is off. This is useful for example when processing files in batches where some have cHPI on/enabled and others don't.

### DIFF
--- a/mne/chpi.py
+++ b/mne/chpi.py
@@ -1031,7 +1031,7 @@ def filter_chpi(raw, include_line=True, t_step=0.01, t_window=None,
     ----------
     raw : instance of Raw
         Raw data with cHPI information. Must be preloaded. Operates in-place.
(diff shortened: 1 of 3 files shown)


mne/chpi.py
```diff
@@ -1025,12 +1031,17 @@ def filter_chpi(raw, include_line=True, t_step=0.01, t_window=None,
     ----------
     raw : instance of Raw
         Raw data with cHPI information. Must be preloaded. Operates in-place.
-    include_line : bool
+    include_line : bool | int
         If True, also filter line noise.
```
agramfort: and if int?

agramfort: you never use an int in any of the diff below

PR #7460, opened 2020-03-16
MRG, FIX: Offload filtering step for find_bad_channels_maxwell

### LINKED ISSUE #7466: mne.preprocessing.find_bad_channels_maxwell()'s crosstalk_file kwarg cannot be pathlib.Path
#### Describe the bug
`mne.preprocessing.find_bad_channels_maxwell()` features a `crosstalk_file` kwarg. Unlike in many (all?) other places in MNE, this argument cannot be a `pathlib.Path` object, and is required to be a string.

#### Steps to reproduce
```python
import pathlib
import mne

sample_data_folder = pathlib.Path(mne.datasets.sample.data_path())
sample_data_raw_file = (sample_data_folder / 'MEG' / 'sample' / 
                        'sample_audvis_raw.fif')
fine_cal_file = sample_data_folder / 'SSS' / 'sss_cal_mgh.dat'
crosstalk_file = sample_data_folder / 'SSS' / 'ct_sparse_mgh.fif'

raw.crop(tmax=60).load_data()

mne.preprocessing.find_bad_channels_maxwell(raw, cross_talk=crosstalk_file,
                                            calibration=fine_cal_file) 
```

#### Expected results
```
Scanning for bad channels in 12 intervals (5.0 sec) ...
    Bad MEG channels being reconstructed: ['MEG 2443']
    Processing 204 gradiometers and 102 magnetometers
    Low-pass filtering data at 40.0 Hz
    Static bad channels: []
[done]
[]
```

#### Actual results
```python
~/Development/mne-python/mne/preprocessing/maxwell.py in _maxwell_filter(***failed resolving arguments***)
    316     #
    317     if cross_talk is not None:
--> 318         sss_ctc = _read_ctc(cross_talk)
    319         ctc_chs = sss_ctc['proj_items_chs']
    320         meg_ch_names = [info['ch_names'][p] for p in meg_picks]

~/Development/mne-python/mne/io/proc_history.py in _read_ctc(fname)
    167     """Read cross-talk correction matrix."""
    168     if not isinstance(fname, str) or not op.isfile(fname):
--> 169         raise ValueError('fname must be a file that exists, not %s' % fname)
    170     f, tree, _ = fiff_open(fname)
    171     with f as fid:

ValueError: fname must be a file that exists, not /Users/hoechenberger/mne_data/MNE-sample-data/SSS/ct_sparse_mgh.fif
```

#### Additional information
It works as expected when converting `crosstalk_file` to a string before passing it to  `find_bad_channels_maxwell()`.

larsoner:
Moves the filtering step out of `find_bad_channels_maxwell`. This makes our outputs differ from MaxFilter a bit more, but:

1. It is better in principle (and practice) to filter all the data, then process in chunks, rather than filter each chunk independently (more edge artifacts)
2. By moving the filtering out of the function, it allows the user to choose the filtering tradeoffs
3. Based on inspecting the results with @hoechenberger , it seems like we do at least as well on this PR as we do in `master`

This PR also make a couple of tweaks found while working on this:

1. Improves `verbose=True` and `verbose='debug'` output messaging
2. Cleans up how `raw.plot` filtering triages FIR vs IIR to match what `mne/filter.py` does (leads to better `padlen` in `raw.plot`)
3. Cleans up unused var in `_get_lims_cola`
4. Separates Maxwell filtering steps a bit so that not as many things need to be recalculated

WIP for now while @hoechenberger and I test a bit more. Redo of #7432 which I errantly merged then force-pushed out of `master`.

Closes #7466

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3069,7 +3069,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
                 % (len(epochs.events)))
     if not np.array_equal(epochs.events[:, 0], np.unique(epochs.events[:, 0])):
         raise RuntimeError('Epochs must have monotonically increasing events')
-    meg_picks, mag_picks, grad_picks, good_picks, _ = \
+    meg_picks, mag_picks, grad_picks, good_mask, _ = \
         _get_mf_picks(epochs.info, int_order, ext_order, ignore_ref)
     coil_scale, mag_scale = _get_coil_scale(
         meg_picks, mag_picks, grad_picks, mag_scale, epochs.info)
@@ -3078,7 +3078,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
     data = np.zeros((n_channels, n_times))
     count = 0
     # keep only MEG w/bad channels marked in "info_from"
-    info_from = pick_info(epochs.info, good_picks, copy=True)
+    info_from = pick_info(epochs.info, meg_picks[good_mask], copy=True)
     all_coils_recon = _prep_mf_coils(epochs.info, ignore_ref=ignore_ref)
     all_coils = _prep_mf_coils(info_from, ignore_ref=ignore_ref)
     # remove MEG bads in "to" info
@@ -3089,7 +3089,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
     n_in, n_out = _get_n_moments([int_order, ext_order])
     S_decomp = 0.  # this will end up being a weighted average
     last_trans = None
-    decomp_coil_scale = coil_scale[good_picks]
+    decomp_coil_scale = coil_scale[good_mask]
     exp = dict(int_order=int_order, ext_order=ext_order, head_frame=True,
                origin=origin)
     for ei, epoch in enumerate(epochs):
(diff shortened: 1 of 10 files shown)


hoechenberger:
@larsoner I'm running some more tests with "real-world" datasets now. Will report back.

hoechenberger:
Ok it seems to work ok for me, but I _do_ get some (only a few, really) false-positives (i.e. channels I personally wouldn't remove, but which get marked as bad when using the default settings). How much do we worry about those?

agramfort:
anything more quantitative to share?

larsoner:
MaxFilter also gets some false positives, so I think it's okay

hoechenberger:
> anything more quantitative to share?

Well I could share data and screenshots … but like I said, the number of false-positives is low, and if @larsoner says that's okay (probably unavoidable), I'll give it my go.

From a researcher's perspective, I'd prefer the algorithm to generate some false alarms to it overlooking some bad channels. (Although I know that's not always the case with this algorithm either)

@agramfort Maybe this is something we should discuss in person, looking at a single screen (or potentially via video chat with screen sharing). I'm still learning about the properties of MEG data, so an expert's opinion would help me calibrate my very personal built-in classifier too :)

At any rate: LGTM!

larsoner:
This PR definitely decreases our agreement with MaxFilter, but in the files that I spot checked it looks like the MNE version does a better job (fewer false alarms generated) by filtering the entire file in one pass. If I apply the same filtering (e.g., 80 Hz w/1 Hz transition band as done in `master` on some of my files) but do it on a block-by-block basis rather than the entire file at once, the agreement with MaxFilter goes back up. However, this mode suffers from edge artifacts and by eye seems to generate false alarms.

hoechenberger:
> This PR definitely decreases our agreement with MaxFilter, but in the files that I spot checked it looks like the MNE version does a better job

+5 on merging this :) It's definitely also very much an improvement to the previous algorithm.

agramfort:
Ok for you @wmvanvliet ?

wmvanvliet:
I'm currently analyzing a dataset. I'll give this function a try to see how it behaves. Unfortunately, with all schools closed, I'm being kept hostage by 3 kids. No idea when I would get to this. If it's important to merge this quick (for 0.20 release) then, by all means, do so. We can come back to it later.

larsoner:
Okay sounds good. Looking forward to the feedback when you can get to testing it out. Let's go ahead and merge so it does make 0.20 though

agramfort:
+1

PR #7461, opened 2020-03-16
MRG, FIX: Fix intersphinx links

larsoner:
Should fix CircleCI errors in `master`.

### DIFF
--- a/mne/utils/numerics.py
+++ b/mne/utils/numerics.py
@@ -563,11 +563,13 @@ def grand_average(all_inst, interpolate_bads=True, drop_bads=True):
 
     Raises
     ------
-    If ``all_inst`` is empty, or if not all elements are evokeds.
+    ValueError
+        If ``all_inst`` is empty, or if not all elements are evokeds.
 
     Warns
     -----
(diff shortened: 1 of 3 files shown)


mne/utils/numerics.py
```diff
@@ -563,11 +563,13 @@ def grand_average(all_inst, interpolate_bads=True, drop_bads=True):
 
     Raises
     ------
-    If ``all_inst`` is empty, or if not all elements are evokeds.
+    ValueError
+        If ``all_inst`` is empty, or if not all elements are evokeds.
```
drammock: should we be doing `Raises` and `Warns` docstring sections on all functions/methods?  I've been assuming we were following the convention of "only document warnings/errors in the docstring if it raises a MNE-Python-specific warning/error class" (which I don't think we define anyway?)

larsoner: Agreed, I wasn't too keen on adding this in #7456 either, as we do raise lots of warnings and errors in lots of places. I'll just remove it for now and we can add it back later if we agree on a plan.

larsoner:
Hotfix in `master`

PR #7462, opened 2020-03-16
Use _time_interp_funcs in _interpolate_data

GuillaumeFavelier:
This PR fixes the recent issues with Circle. For short, `_interpolate_data()` now uses `_time_interp_funcs`

Reference: https://github.com/mne-tools/mne-python/pull/7257#discussion_r393315334


larsoner:
Can you touch at least one of the broken examples so that we can verify the fix?

larsoner:
CircleCI is fixed so I'll merge. It looks like something else has broken most of our other CIs

PR #7469, opened 2020-03-17
MRG, FIX: Fix for singleton and efficiency

larsoner:
An attempt to recover some commits from the rolled-back #7257. In theory this should fix CircleCI in `master`. Tests pass locally, let's see if the CIs agree. This refactors `_brain` to make several improvements:

1. `smoothing_matrix` is only used in one place
2. time interpolation is only constructed once (in one place) and reused
3. singleton dimensions are hopefully handled properly

@GuillaumeFavelier if things are green I think we should merge this, then maybe bring back the rest of #7257 ?


agramfort:
thx @larsoner

PR #7470, opened 2020-03-18
MRG, DOC: Clustering docstrings and connectivity=None

### LINKED ISSUE #7463: RFC: result of connectivity=None to clustering functions may be misleading
Summarizing a discussion I just had offline with @larsoner:

In `mne.stats.spatio_temporal_cluster{_1samp}_test`, the input array `X` takes shape (n_observations, n_times, n_vertices). Normally, the vertices come from an STC object, and a (sparse) matrix encoding the connectivity is easily obtained from the associated source space.

However, it is allowed to pass `connectivity=None`, which will default to using `scipy.ndimage.generate_binary_structure(2, 1)` to define a 2D regular lattice connectivity. The problem with that is that the 2D lattice connectivity may or may not be appropriate, because (again) the input shape is (n_observations, n_times, n_vertices), so it's only useful if your n_vertices dimension is in fact spatially one-dimensional.  But when would a neuroscience application be dealing with a series of spatial points that actually have one-dimensional connectivity?

It seems like supporting `connectivity=None` is a case of YAGNI, and might lead to unexpected results for users who, by analogy with the usage with STCs, think "OK, I have this 2D array, I need to ravel it, but I don't need to compute connectivity because it says the default is to use regular lattice connectivity, which is what I want".  

I think `permutation_cluster{_1samp}_test` has the same problem: it [flattens out the last dimension](https://github.com/mne-tools/mne-python/blob/master/mne/stats/cluster_level.py#L837-L838) prior to eventually [running the default lattice connectivity code](https://github.com/mne-tools/mne-python/blob/master/mne/stats/cluster_level.py#L492-L493).

My recommendation is to simply remove `connectivity=None` as a valid option to those functions. @agramfort WDYT?

crossref #4859

drammock:
In preparation for addressing #7463, this cleans up the docstrings for the 4 public-facing clustering functions:

- `mne.stats.permutation_cluster_test`
- `mne.stats.permutation_cluster_1samp_test`
- `mne.stats.spatio_temporal_cluster_test`
- `mne.stats.spatio_temporal_cluster_1samp_test`

remaining tasks:

- [x] figure out whether to change behavior of `connectivity=None` (see #7463 for discussion) *after extensive testing: I was wrong, everything works as expected with the current implementation.* closes #7463
- [x] revise entry for `connectivity` parameter in docstrings accordingly.

### DIFF
--- a/doc/python_reference.rst
+++ b/doc/python_reference.rst
@@ -1033,6 +1033,7 @@ Logging and Configuration
    set_log_level
    set_log_file
    set_config
+   set_cache_dir
    sys_info
    verbose
 
--- a/mne/stats/cluster_level.py
+++ b/mne/stats/cluster_level.py
@@ -1050,10 +1050,10 @@ def permutation_cluster_test(
         The data to be clustered. Each array in ``X`` should contain the
         observations for one group. The first dimension of each array is the
         number of observations from that group; remaining dimensions comprise
-        the size of a single observation. For example if X = [X1, X2]
-        with X1.shape = (20, 50, 4) and X2.shape = (17, 50, 4), then X has
-        2 groups with respectively 20 and 17 observations in each, and
-        each data point is of shape (50, 4).
+        the size of a single observation. For example if ``X = [X1, X2]``
+        with ``X1.shape = (20, 50, 4)`` and ``X2.shape = (17, 50, 4)``, then
+        ``X`` has 2 groups with respectively 20 and 17 observations in each,
+        and each data point is of shape ``(50, 4)``.
     %(clust_thresh_f)s
     %(clust_nperm_int)s
     %(clust_tail)s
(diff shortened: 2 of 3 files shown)


drammock:
here are the new docstrings:

[permutation_cluster_test](https://18747-1301584-gh.circle-artifacts.com/0/dev/generated/mne.stats.permutation_cluster_test.html)
[permutation_cluster_1samp_test](https://18747-1301584-gh.circle-artifacts.com/0/dev/generated/mne.stats.permutation_cluster_1samp_test.html)
[spatio_temporal_cluster_test](https://18747-1301584-gh.circle-artifacts.com/0/dev/generated/mne.stats.spatio_temporal_cluster_test.html)
[spatio_temporal_cluster_1samp_test](https://18747-1301584-gh.circle-artifacts.com/0/dev/generated/mne.stats.spatio_temporal_cluster_1samp_test.html)

drammock:
@larsoner @agramfort still waiting for slowpoke travis, but all other CIs pass. ready for review/merge.

larsoner:
Thanks @drammock

PR #7474, opened 2020-03-18
MRG: [DOC] MNE does no longer auto apply average ref

### LINKED ISSUE #7471: Update docs and code: set_eeg_reference([]) and the automatic average
Based on this the thread on referencing in the [mailing list](https://mail.nmr.mgh.harvard.edu/pipermail/mne_analysis/2020-March/006577.html), I got the idea that the past recommendation to use `set_eeg_reference([])` to prevent MNE from adding an automatic average reference is no longer needed.

Yet throughout the codebase, there are still references to this old behavior (a list follows).

https://github.com/mne-tools/mne-python/blob/25f6e5e097d8492bfcc3abb6e77804d372f68a9d/examples/preprocessing/plot_rereference_eeg.py#L47-L50

https://github.com/mne-tools/mne-python/blob/25f6e5e097d8492bfcc3abb6e77804d372f68a9d/tutorials/evoked/plot_eeg_erp.py#L72-L75

https://github.com/mne-tools/mne-python/blob/25f6e5e097d8492bfcc3abb6e77804d372f68a9d/tutorials/evoked/plot_eeg_erp.py#L93

... and there are many more :-)

Is it worth it to go through the code and remove these wrong docs? 

cc @dominikwelke @larsoner 

PS: Through my browser (Firefox), I get an encryption warning for the MNE mailing list. Just fyi.

sappelhoff:
closes #7471 

These are the files I could catch that did still assume MNE-Python to auto apply an average ref projection. Also, the docstr of `set_eeg_reference` was still assuming that.

I wasn't sure with `tutorials/preprocessing/plot_55_setting_eeg_reference.py` --> see my inline "todo" comment there marked with `XXX`

PS: What was the PR that removed the auto average ref behavior of MNE?

### DIFF
--- a/tutorials/evoked/plot_eeg_erp.py
+++ b/tutorials/evoked/plot_eeg_erp.py
@@ -20,6 +20,10 @@ raw_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
 event_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw-eve.fif'
 raw = mne.io.read_raw_fif(raw_fname, preload=True)
 
+# This particular dataset already has an average reference projection added
+# that we now want to remove for the sake of this example.
+raw.set_eeg_reference([])
+
 ###############################################################################
 # Let's restrict the data to the EEG channels
 raw.pick_types(meg=False, eeg=True, eog=True)
@@ -65,8 +69,8 @@ raw.plot_sensors('3d')  # in 3D
 # Setting EEG reference
 # ---------------------
 #
-# Let's first inspect our Raw object with its original reference
-# (i.e., after loading).
+# Let's first inspect our Raw object with its original reference that was
+# applied during the recording of the data.
 # We define Epochs and compute an ERP for the left auditory condition.
 reject = dict(eeg=180e-6, eog=150e-6)
 event_id, tmin, tmax = {'left/auditory': 1}, -0.2, 0.5
@@ -81,7 +85,9 @@ evoked_no_ref.plot(titles=dict(eeg=title), time_unit='s')
 evoked_no_ref.plot_topomap(times=[0.1], size=3., title=title, time_unit='s')
 
 ###############################################################################
-# **Common average reference** (car)
+# **Common average reference (car)**: We add back the average reference
+# projection that we removed at the beginning of this example (right after
+# loading the data.)
 raw_car, _ = mne.set_eeg_reference(raw, 'average', projection=True)
 evoked_car = mne.Epochs(raw_car, **epochs_params).average()
 del raw_car  # save memory
--- a/tutorials/evoked/plot_eeg_erp.py
+++ b/tutorials/evoked/plot_eeg_erp.py
@@ -87,7 +87,7 @@ evoked_no_ref.plot_topomap(times=[0.1], size=3., title=title, time_unit='s')
 ###############################################################################
 # **Common average reference (car)**: We add back the average reference
 # projection that we removed at the beginning of this example (right after
-# loading the data.)
+# loading the data).
 raw_car, _ = mne.set_eeg_reference(raw, 'average', projection=True)
 evoked_car = mne.Epochs(raw_car, **epochs_params).average()
 del raw_car  # save memory
--- a/tutorials/evoked/plot_eeg_erp.py
+++ b/tutorials/evoked/plot_eeg_erp.py
@@ -18,7 +18,6 @@ from mne.datasets import sample
 data_path = sample.data_path()
 raw_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
 event_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw-eve.fif'
-# these data already have an EEG average reference
 raw = mne.io.read_raw_fif(raw_fname, preload=True)
 
 ###############################################################################
@@ -50,7 +49,6 @@ raw.set_channel_types(mapping={'EOG': 'eog'})
 # The EEG channels in the sample dataset already have locations.
 # These locations are available in the 'loc' of each channel description.
 # For the first channel we get
-
 print(raw.info['chs'][0]['loc'])
 
 ###############################################################################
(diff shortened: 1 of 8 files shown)


tutorials/preprocessing/plot_55_setting_eeg_reference.py
```diff
@@ -192,6 +192,7 @@
 # the source modeling is performed. In contrast, applying an average reference
 # by the traditional subtraction method offers no such guarantee.
 #
+# XXX: IS THIS TRUE?
 # For these reasons, when performing inverse imaging, *MNE-Python will
 # automatically average-reference the EEG channels if they are present and no
 # reference strategy has been specified*. If you want to perform inverse
```
sappelhoff: note: I am uncertain about how to change the lines above. Will MNE-Python or will it NOT apply auto average refs?

drammock: As far as I can tell, it is *not* automatically applied:

- `_check_reference` is called in `_apply_inverse` [here](https://github.com/mne-tools/mne-python/blob/master/mne/minimum_norm/inverse.py#L885)
- `_check_reference` ([here](https://github.com/mne-tools/mne-python/blob/master/mne/minimum_norm/inverse.py#L761-L773)) calls `_needs_eeg_average_ref_proj` ([here](https://github.com/mne-tools/mne-python/blob/master/mne/io/proj.py#L756-L766)), which returns True if  (1) there are EEG channels present, and (2) a custom ref has not been applied, and (3) EEG average reference proj is not already there.

Instead, the user will get a `ValueError`.

In fact, it doesn't even seem to be possible any more to avoid this constraint by setting average reference to an empty list:

```python
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False).crop(tmax=30).load_data()               
raw = raw.set_eeg_reference([])                                                                        
mne.minimum_norm.inverse._check_reference(raw)
# ValueError: EEG average reference is mandatory for inverse modeling, use set_eeg_reference method.
```

sappelhoff: Thanks @drammock

sappelhoff:
https://github.com/mne-tools/mne-python/blob/25f6e5e097d8492bfcc3abb6e77804d372f68a9d/mne/channels/channels.py#L281 has a `set_eeg_reference` method which is just pointing to `mne.set_eeg_reference`.

Instead of duplicating the docstring in that class method, can we somehow "import it" from the original `mne.set_eeg_reference`?

drammock:
> Instead of duplicating the docstring in that class method, can we somehow "import it" from the original `mne.set_eeg_reference`?

You can take a look at `mne/utils/docs.py` to see `docdict`; frequently re-used docstring elements are put there, and then inserted via `%(docdict_key_name)s` where they're needed. There are also decorators `@copy_function_doc_to_method_doc` and `@copy_doc`.

sappelhoff:
Cool, this decorator would be perfect for it: 

https://github.com/mne-tools/mne-python/blob/25f6e5e097d8492bfcc3abb6e77804d372f68a9d/mne/utils/docs.py#L900

except for two things:

1. the method I would like to receive a copy of the function docstring does not have a parameter that the function has  (`copy`, the 3rd param of the function)
1. the function has two return values, the method only one (that one is congruent though).

perhaps the decorator could be extended with an optional parameter to delete a list of params by indices (currently, it's only deleting the first param, index 0) ... and extend the param deletion to a deletion of return values as well?

larsoner:
> perhaps the decorator could be extended with an optional parameter to delete a list of params by indices (currently, it's only deleting the first param, index 0) ... and extend the param deletion to a deletion of return values as well?

Maybe but it's not too difficult just to split the docstring into needed parts. Also we have to be wary of adding complexity to our docstring modifiers since they are executed on import.

See how I dealt with this sort of thing recently in MaxFilter:

https://github.com/mne-tools/mne-python/blob/master/mne/preprocessing/maxwell.py#L63-L93

There is a function that uses all the same params except for a couple (e.g., `st_duration`, `st_correlation` which are not allowed) so I grouped params and made entries in `docs.py`:

https://github.com/mne-tools/mne-python/blob/master/mne/utils/docs.py#L298-L384

sappelhoff:
okay, thanks. Ready from my side.

drammock:
@sappelhoff you have a "double line break" error in the docstring test. to iterate more quickly to find a solution, you can do `make docstring` locally from the root of the repo, to see if your fixes work.

tutorials/evoked/plot_eeg_erp.py
```diff
@@ -18,9 +18,12 @@
 data_path = sample.data_path()
 raw_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
 event_fname = data_path + '/MEG/sample/sample_audvis_filt-0-40_raw-eve.fif'
-# these data already have an EEG average reference
 raw = mne.io.read_raw_fif(raw_fname, preload=True)
 
+# This particular dataset already has an average reference projection added
+# that we now want to remove for the sake of this example.
```
agramfort:
```suggestion
# that we now want to remove it for the sake of this example.
```

tutorials/preprocessing/plot_55_setting_eeg_reference.py
```diff
@@ -192,16 +192,6 @@
...
 #
-# For these reasons, when performing inverse imaging, *MNE-Python will
-# automatically average-reference the EEG channels if they are present and no
-# reference strategy has been specified*. If you want to perform inverse
-# imaging and do not want to use an average reference (and hence you accept the
-# risks presented in the previous paragraphs), you can force MNE-Python to
-# relax its average reference requirement by passing an empty list to
-# :meth:`~mne.io.Raw.set_eeg_reference` (i.e., by calling
-# ``raw.set_eeg_reference(ref_channels=[])``) prior to performing inverse
-# imaging.
```
agramfort: you remove this as it's not true anymore? or MNE will already give you a meaningful message?

sappelhoff: yes, it's not true anymore - and MNE puts a ValueError when users try to do inverse without an average ref, see https://github.com/mne-tools/mne-python/pull/7474/commits/b00a3f806a8ed2ee3659238e6f36f72b153105dc#r394477027

sappelhoff:
> @sappelhoff you have a "double line break" error in the docstring test. to iterate more quickly to find a solution, you can do make docstring locally from the root of the repo, to see if your fixes work.

I tried several things now but couldn't find the issue :-/

agramfort:
ok !

>

tutorials/preprocessing/plot_55_setting_eeg_reference.py
drammock: I'd be inclined to still say *something* here. Maybe this?

```suggestion
# For these reasons, when performing inverse imaging, *MNE-Python will
# raise a ``ValueError`` if there are EEG channels present and
# something other than an average
# reference strategy has been specified*.
```

Although it seems odd to me that it's no longer possible to override using `[]`

drammock:
@sappelhoff I fixed the double line break error (locally, at least).  The problem was that there was a line break between the end of the `Returns` section and the `%(docdict_key)s` code, and the thing that gets filled in by the `%(docdict_key)s` code begins with another linebreak.  I also pushed a fix for slightly nicer formatting in the notes section.

tutorials/preprocessing/plot_55_setting_eeg_reference.py
larsoner: Change looks reasonable to me

I don't know if it was ever actually possible to relax the average ref requirement, that sounds like a documentation error to me

larsoner: (by "change" looking reasonable I mean these lines with your suggestion incorporated @drammock -- the documentation error looks like it will be removed even with your suggested changes  applied)

sappelhoff:
Thanks a lot @drammock

larsoner:
Thanks @sappelhoff

PR #7477, opened 2020-03-18
MRG, ENH: Speed up info copy

larsoner:
Sometimes I find that `info.copy()` is a significant bottleneck in some operations. It turns out this can be avoided if we leverage knowledge of our internal structures and object mutability by overridding a few `__deepcopy__` defs:

1. three entries in `info` (`chs`, `ch_names`, `hpi_meas`), plus two others that affect it:
2. `DigPoint` (hence `info['dig']`)
3. `Projection` (hence `info['projs']`)

This code:
```Jupyter
import mne
fname = mne.datasets.testing.data_path() + '/SSS/test_move_anon_raw.fif'
info = mne.io.read_info(fname)
%timeit info.copy()
```
on master (top) and this PR (bottom; ~10x faster) gives:
```
6.08 ms ± 333 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
580 µs ± 2.57 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
Switching to `sample_audvis_raw.fif` we see similar timings (~8x speedup):
```
3.88 ms ± 23.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
487 µs ± 4.66 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Along the way I added some extra checks to `info._check_consistency()` that showed inconsistencies in dimensionality in `cnt` and `fieldtrip` code that I fixed.

This is mostly internal with some speedups throughout the code base so doesn't seem like it needs a `latest.inc` update.

### DIFF
--- a/mne/io/_digitization.py
+++ b/mne/io/_digitization.py
@@ -101,6 +101,12 @@ class DigPoint(dict):
         pos = ('(%0.1f, %0.1f, %0.1f) mm' % tuple(1000 * self['r'])).ljust(25)
         return ('<DigPoint | %s : %s : %s frame>' % (id_, pos, cf))
 
+    # speed up info copy by only deep copying the mutable item
+    def __deepcopy__(self, memodict):
+        return DigPoint(
+            kind=self['kind'], r=self['r'].copy(),
+            ident=self['ident'], coord_frame=self['coord_frame'])
+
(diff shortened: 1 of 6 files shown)


agramfort:
thx @larsoner

PR #7480, opened 2020-03-19
MRG: Improve memory footprint of _TimeViewer

GuillaumeFavelier:
This PR improves the memory footprint of `_TimeViewer` by decreasing the reference count and solving reference cycles.

This is the comparison between `master` (black curve) and this PR (blue curve) using the following script with `mprof`:

```py
import os
import mne
import pyvista
initial_time = 0.1
for j in range(5):
    for i in range(5):
                         hemi='lh',
                         time_viewer=True, show_traces=True)
    pyvista.close_all()
```

image

There is probably room for improvement but this is a step forward.

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -14,14 +14,14 @@ from ...source_space import vertex_to_mni
 class MplCanvas(object):
     """Ultimately, this is a QWidget (as well as a FigureCanvasAgg, etc.)."""
 
-    def __init__(self, timeviewer, width, height, dpi):
+    def __init__(self, time_viewer, width, height, dpi):
         from PyQt5 import QtWidgets
         from matplotlib.figure import Figure
         from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
I don't think I can go further with this method, there is the latest result:

image

mne/viz/_brain/_timeviewer.py
```diff
@@ -837,6 +838,50 @@ def help(self):
...
+        if hasattr(self, "mpl_canvas"):
+            self.mpl_canvas.close()
+            self.mpl_canvas.time_viewer = None
+            self.mpl_canvas.canvas = None
+            self.mpl_canvas = None
+        self.time_actor = None
+        self.picked_renderer = None
+        self.act_data["lh"] = None
+        self.act_data["rh"] = None
+        self.act_data = None
```
larsoner: Oh wow :)

GuillaumeFavelier: I took no risk xD

larsoner:
Failure is just a timeout, thanks @GuillaumeFavelier

PR #7483, opened 2020-03-20
MRG, ENH: Return flats with find_bad_channels_maxwell

larsoner:
@hoechenberger feel free to try.

Should be in 0.20 to make it within-cycle API change.

### DIFF
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1889,12 +1889,12 @@ def find_bad_channels_maxwell(
 
     Returns
     -------
-    bads : list
-        List of bad MEG channels that were automatically detected among
-        the good MEG channels.
-    flats : list
-        List of MEG channels that were detected as being flat at least
-        ``min_count`` times.
+    noisy_chs : list
+        List of bad MEG channels that were automatically detected as being
+        noisy among the good MEG channels.
+    flat_chs : list
+        List of MEG channels that were detected as being flat in at least
+        ``min_count`` segments.
 
     See Also
     --------
@@ -1962,8 +1962,8 @@ def find_bad_channels_maxwell(
     del origin, int_order, ext_order, calibration, cross_talk, coord_frame
     del regularize, ignore_ref, bad_condition, head_pos, mag_scale
     good_meg_picks = params['meg_picks'][params['good_mask']]
-    bads = Counter()
-    flats = Counter()
+    noisy_chs = Counter()
+    flat_chs = Counter()
     flat_limits = dict(grad=0.01e-13, mag=0.01e-15)
     these_limits = np.array([
         flat_limits['grad']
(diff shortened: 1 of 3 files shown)


agramfort:
to me flats are part of bads so it's weird to return 2 items one being called bads and the other one flats.

larsoner:
I think it's useful to have them split and it's super easy for users to do `bads + flats` (we do this in the example) so I don't see harm in splitting them. However, combining them makes it so that you can't for example keep track as easily how often a given channel was bad vs flat in a large dataset without parsing logs (which is annoying to do)

hoechenberger:
> I think it's useful to have them split and it's super easy for users to do `bads + flats` (we do this in the example) so I don't see harm in splitting them.

I generally agree, because at a certain level, they are conceptually different things:
`flats` is for _lack of variability_, while `bads` is for _too much variability_.

But I'm not happy with the nomenclature. To me, a "bad" channel is one that's problematic and should be removed before conducting further analysis. This is also how this term is used in MNE: we only have `.info['bads']`, but no `.info['flats']`. (Which I think is good) Not, typically, we need to remove both channels that are producing noise only, and those that produce pretty much no signal at all (i.e. are flat).

Maybe the return values of `find_bad_channels_maxwell` shouldn't be called `bads, flats`, but sth like `fluctuating_chs, flat_chs` instead?

larsoner:
How about `outliers`, `flats`?

larsoner:
Or `noisy`, `flat`?

hoechenberger:
I thougt about `noise, flat` first, but we'd need to use plural forms to be consistent with the "MNE naming scheme", and `noisies, flats` sounds a bit weird I believe… `outliers, flats` could work... 
How about `noisy_chs, flat_chs`..?

hoechenberger:
`irregulars, flats`? `deviants, flats`?

hoechenberger:
`eccentrics, flats` :D

drammock:
+1 for `noisy_chs`, `flat_chs`

larsoner:
Pushed commit with `noisy_chs, flat_chs`

hoechenberger:
@larsoner LGTM!

hoechenberger:
Hehe, great timing @agramfort!

hoechenberger:
Seems Travis is having big trouble these days, currently cannot even access the test results. I believe the macOS Test simply timed out, though...

agramfort:
they are under maintenance

>

hoechenberger:
Let's wait until this afternoon and see if they get things sorted out… Otherwise I'd be inclined to just merge: I remember seeing that only the macOS test hadn't started yesterday, and the others had successfully finished.

drammock:
Yesterday I restarted the Travis macos test twice for this PR, after 2 timeouts. +1 for merging without it.

agramfort:
fair enough. thanks @larsoner !

PR #7484, opened 2020-03-20
MRG, FIX: Remove cruft

larsoner:
Just removing some version cruft we no longer need because our minimum is 0.19. Also adding one more ignore needed for latest dev `pandas` with `statsmodels` importing something deprecated


drammock:
tx @larsoner

PR #7486, opened 2020-03-20
MRG, MAINT: standardize the way we get channel types

drammock:
This PR tries to clean up / standardize how channel types are found in the code base.

- the instance methods `<raw/epochs/evoked>.get_channel_types()` now internally use the existing `_get_channel_types` function instead of a pared-down reimplementation of it.  Accordingly, they also now have method parameters `picks=None, unique=False, only_data_chs=False`.
- Various places in the codebase have been switched from list comprehensions like `[channel_type(info, x) for x in range(info['nchans'])]` to simply call the `_get_channel_types(info)` or `_get_channel_types(info, picks)`. A few more complicated list comprehensions have been left alone, as they are simpler as-is than they would have been if changed to use the `_get_channel_types` function.
- the instance method `raw.get_channel_types()` has been added to the "info data structure" tutorial, as a contrast/alternative to `channel_type` function.

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -214,7 +214,7 @@ class ContainsMixin(object):
 
         Parameters
         ----------
-        %(picks_nostr)s
+        %(picks)s
         unique : bool
             Whether to return only unique channel types. Default is ``False``.
         only_data_chs : bool
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -22,7 +22,8 @@ from ..io.constants import FIFF
 from ..io.meas_info import anonymize_info, Info, MontageMixin
 from ..io.pick import (channel_type, pick_info, pick_types, _picks_by_type,
                        _check_excludes_includes, _contains_ch_type,
-                       channel_indices_by_type, pick_channels, _picks_to_idx)
+                       channel_indices_by_type, pick_channels, _picks_to_idx,
+                       _get_channel_types)
 
 
 def _get_meg_system(info):
@@ -207,16 +208,25 @@ class ContainsMixin(object):
         """The current gradient compensation grade."""
         return get_current_comp(self.info)
 
-    def get_channel_types(self):
+    @fill_doc
+    def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
         """Get a list of channel type for each channel.
+
+        Parameters
+        ----------
+        %(picks_nostr)s
+        unique : bool
+            Whether to return only unique channel types. Default is ``False``.
+        only_data_chs : bool
+            Whether to ignore non-data channels. Default is ``False``.
 
         Returns
         -------
         channel_types : list
             The channel types.
         """
-        return [channel_type(self.info, n)
-                for n in range(len(self.info['ch_names']))]
+        return _get_channel_types(self.info, picks=picks, unique=unique,
+                                  only_data_chs=only_data_chs)
 
 
 # XXX Eventually de-duplicate with _kind_dict of mne/io/meas_info.py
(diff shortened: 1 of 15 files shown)


mne/channels/channels.py
```diff
@@ -207,16 +208,25 @@ def compensation_grade(self):
...
         return get_current_comp(self.info)
 
-    def get_channel_types(self):
+    @fill_doc
+    def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
         """Get a list of channel type for each channel.
 
+        Parameters
+        ----------
+        %(picks_nostr)s
```
larsoner: I don't think this should be `nostr`, that's usually reserved for when `info` is unavailable...?

agramfort:
thanks so much @drammock for cleaning this up. Let me know when to merge (CIs green). Diff looks good.

cbrnr:
What about `mne.io.pick.get_channel_types`? Do you still want to rename it? Can this function use `mne.io.pick._get_channel_types` to avoid duplication?

agramfort:
the idea is that we should avoid having 2 functions with same name doing
different things

cbrnr:
So what does that mean then?

drammock:
> What about mne.io.pick.get_channel_types? Do you still want to rename it? 

@cbrnr yes, see #7487

agramfort:
rename get_channel_types to get_all_channel_types and
have get_channel_types as method of data containers with an info. I think...

>

mne/channels/channels.py
drammock: it was accurate given how the method was defined when you reviewed. Originally it only handled integer picks, and I figured YAGNI for string-based picking, since if you know the channel name or type you want to pick, you probably know its type already. But now I realize it was easy enough to add string-based picking, and also decided there's a real use case, e.g., disambiguating mag vs grad among `MEG_xxxx` channel names.  So now that works, and I've changed `picks_nostr` to `picks`.

mne/io/pick.py
```diff
@@ -15,7 +15,7 @@
                      _check_option)
 
 
-def get_channel_types():
+def get_channel_type_constants():
```
agramfort: It’s public function we need to deprecate it properly. It’s used in mnelab

drammock: oops. yep, I already forgot my own suggestion :)  Will fix.

drammock:
ok @agramfort @larsoner other than the perpetually failing Travis OSX tests, this is green and ready for review/merge

larsoner:
If you want to roll some community service into your PR, you can take the top half dozen or so tests [here](https://travis-ci.org/github/mne-tools/mne-python/jobs/665213265) and add `@pytest.mark.slowtest  # slow-ish on Travis OSX` decorators to them

mne/io/pick.py
```diff
@@ -12,10 +12,16 @@
...
+                     _check_option, deprecated)
 
 
+@deprecated('The function mne.io.pick.get_channel_types() has changed names. '
+            'Please use mne.io.pick.get_channel_type_constants() instead.')
 def get_channel_types():
+    return get_channel_type_constants()
+
+
+def get_channel_type_constants():
```
larsoner: Should we add this to `python_reference.rst` or no?

drammock: Personally I don't see its utility for the average user... If you disagree feel free to push that addition

drammock:
> If you want to roll some community service into your PR, you can take the top half dozen or so tests [here](https://travis-ci.org/github/mne-tools/mne-python/jobs/665213265) and add `@pytest.mark.slowtest  # slow-ish on Travis OSX` decorators to them

This whole PR is community service 😀. But my laptop is closed for the rest of the weekend, so if you want it,  it won't happen until Monday

agramfort:
just a mac os failure. Let's merge this one. @larsoner feel free to open a new one on top of it.

PR #7488, opened 2020-03-21
DOC: Annotations class accepts float as duration param

hoechenberger:
`mne.Annotations` can handle `duration` being a float or array with only a single element. Add this info to the docstring.

This use case is already covered by the existing tests.


drammock:
Merging because Travis error is just another OSX timeout.

PR #7490, opened 2020-03-22
MRG, MAINT: Skip more on OSX

larsoner:
Speed up OSX on Travis by skipping a few more, using "slow-ish" in the wording so it's easier to `git grep` later if it speeds back up.


agramfort:
thx @larsoner

Issue #7492, opened 2020-03-22
Cannot read FIF file with invalid date


hoechenberger:
#### Describe the bug
When trying to read a `.fif` file that contains an invalid date field (in my case, an invalid birthday), an exception is raised. Previously (0.19.2, if I'm not mistaken), the file would be read, and I could manually check the date to see if it was reasonable:

```python
        subject_info = raw.info['subject_info']
        birthday = subject_info['birthday']
        if (not 1900 < birthday[0] < 2100 or
                not 1 <= birthday[1] <= 12 or
                not 1 <= birthday[2] <= 31):
            del subject_info['birthday']
```
I cannot use this approach anymore as reading the file fails in the first place.

#### Steps to reproduce
Try to read a file with an invalid date in its `info` dict (measurement date or, for that matter, birthdaty) via `mne.io.read_raw_fif()`.

(I can provide such a file if need be)

#### Expected results
Previously, the file would be loaded, with the `birthday` entry in the `info` dict being obviously nonsensical.

#### Actual results
An exception is raised.

```python
Traceback (most recent call last):
  File "/home/parietal/rhochenb/Development/Cam-CAN/Code/convert_camcan.py", line 87, in <module>
    raw = mne.io.read_raw_fif(raw_fname)
  File "/home/parietal/rhochenb/Development/mne-python/mne/io/fiff/raw.py", line 452, in read_raw_fif
    return Raw(fname=fname, allow_maxshield=allow_maxshield,
  File "<decorator-gen-169>", line 21, in __init__
  File "/home/parietal/rhochenb/Development/mne-python/mne/io/fiff/raw.py", line 79, in __init__
    self._read_raw_file(next_fname, allow_maxshield,
  File "<decorator-gen-170>", line 21, in _read_raw_file
  File "/home/parietal/rhochenb/Development/mne-python/mne/io/fiff/raw.py", line 151, in _read_raw_file
    info, meas = read_meas_info(fid, tree, clean_bads=True)
  File "<decorator-gen-24>", line 21, in read_meas_info
  File "/home/parietal/rhochenb/Development/mne-python/mne/io/meas_info.py", line 1244, in read_meas_info
    tag = read_tag(fid, pos)
  File "/home/parietal/rhochenb/Development/mne-python/mne/io/tag.py", line 451, in read_tag
    tag.data = fun(fid, tag, shape, rlims)
[...]

PR #7494, opened 2020-03-22
BF: Allow reading of FIF files with incorrect date

hoechenberger:
#### Reference issue
Fixes GH-7492.

Currently, FIF files that have an invalid birthday or measurement date set cannot be read, as an exception is raised (see #7492). This PR catches the exception, skips the entry (effectively adding it as `None` to the `info` dict), and emits a warning. This way, users can at least load the data, even though the metadata may then be partially incomplete.

#### Additional information
Tested with the problematic Cam-CAN data mentioned in #7492.

I need advise on how to best come up with a test for this.

### DIFF
--- a/mne/datasets/utils.py
+++ b/mne/datasets/utils.py
@@ -239,7 +239,7 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     path = _get_path(path, key, name)
     # To update the testing or misc dataset, push commits, then make a new
     # release on GitHub. Then update the "releases" variable:
-    releases = dict(testing='0.81', misc='0.5')
+    releases = dict(testing='0.82', misc='0.5')
     # And also update the "md5_hashes['testing']" variable below.
 
     # To update any other dataset, update the data archive itself (upload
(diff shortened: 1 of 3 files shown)


hoechenberger:
I just discovered https://github.com/mne-tools/mne-python/blob/586a8711fc94048876ea23fdf5a638783770d50b/mne/io/meas_info.py#L1315-L1317

This should probably be adjusted too? Not sure what exactly it refers to – measurement date of the helium level? (whatever _that_ means – supply level? pressure? temperature?)

agramfort:
certain meas_date which are not info['meas_date'] are meant to stay as
tuple of int. So not clear.

>
problem comes when you make a datetime object from these tuple of ints that
is bad.

bloyl:
I think a potential alternative would be would be to catch the exception here:
https://github.com/mne-tools/mne-python/blob/943dee405fcc68f985107a8f7c6912ee2adfa87a/mne/io/tag.py#L369-L371
and return `None`. I think that would cover all julian dates.

hoechenberger:
@bloyl ok, thanks, will look into this tomorrow :)

drammock:
To test this, you could read a file with a legitimate date, change it to a bad date, save it (to a temp dir), then read it back in while catching the warning, and finally testing that the date was None. Seems pretty contorted, but if you tacked it onto an existing test that was already doing some of those steps, it might be worth it?

hoechenberger:
> To test this, you could read a file with a legitimate date, change it to a bad date, save it (to a temp dir)

I was thinking about this but since dates are converted to datetime objects now before writing, I cannot write out "invalid" dates, I'm afraid

hoechenberger:
I have now followed a suggestion by @agramfort and created a very small raw file with an invalid birthday via MNE-Python 0.19; this is now included in the testing data at `mne/io/tests/data`.

When looking at how the measurement date was stored, I figured that apparently already in 0.19 and potentially even earlier (?) datetime objects had been used, and it's only the birthday for which a (YYYY, MM, DD) tuple was used, which lead to the issue I'm trying to solve here (i.e., invalid dates could end of in the metadata). Therefore, I'm limiting my adjustments and testing to the `info['subject_info']['birthday']` field now.

This also made me believe that catching an exception when reading the data quite late, namely in `meas_info.py`, should be sufficient – I'm just doing this for the birthday now, anyway. You okay with that, @bloyl?

agramfort:
Thx @hoechenberger

PR #7495, opened 2020-03-23
MRG, FIX: Fix bug with rank checking in lcmv

larsoner:
@britta-wstnr feel free to try on your data

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -148,6 +148,8 @@ Bug
 
 - Fix the ``reduce_rank`` parameter in :func:`mne.beamformer.make_lcmv` and :func:`mne.beamformer.make_dics` to reduce the rank of the leadfield first before then reducing the rank of the denominator of the beamformer formula for the inversion by `Britta Westner`_ and `Eric Larson`_.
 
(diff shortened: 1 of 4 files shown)


britta-wstnr:
@larsoner works on my example with `rank=None` !

doc/changes/latest.inc
```diff
@@ -148,6 +148,8 @@ Bug
 
 - Fix the ``reduce_rank`` parameter in :func:`mne.beamformer.make_lcmv` and :func:`mne.beamformer.make_dics` to reduce the rank of the leadfield first before then reducing the rank of the denominator of the beamformer formula for the inversion by `Britta Westner`_ and `Eric Larson`_.
 
+- Fix bug with :func:`mne.beamformer.make_lcmv` with MaxFiltered data where rank was incorrectly checked by `Eric Larson`_.
+
```
britta-wstnr: I think it is not only for maxfiltered data. Could also happen with ICA'ed data etc - if no noise covariance matrix is supplied.

agramfort:
ok to merge @britta-wstnr ? if so go for it.

britta-wstnr:
thanks @larsoner :)

PR #7497, opened 2020-03-23
MRG, STY: use unicode micro symbol wherever possible

drammock:
This PR:

- replaces μ (`03BC` greek letter mu) with µ (`00B5` micro sign) where appropriate (i.e., as an SI prefix). Has been left as `03BC` when used as a variable name in formulae / explanations.
- replaces `'uV'` with `'µV`' everywhere (except in tests where it looked like I might break something, e.g., `io` tests, I tried to be conservative there)
- along the way, removes some unicode string prefixes left over from py2.x compatibility.

There's a decent chance I broke something, so let the CIs run to completion on this one.


mne/io/edf/edf.py
```diff
@@ -576,7 +576,7 @@ def _read_edf_header(fname, exclude):
         for i, unit in enumerate(units):
             if i in exclude:
                 continue
-            if unit == 'uV':
+            if unit in ('uV', 'µV'):
```
drammock: IDK if this is unnecessarily conservative, or on the contrary, totally pointless. I *think* it might be pointless because if it wasn't needed before, then EDF headers will never say "µV" anyway.

mne/io/edf/edf.py
```diff
@@ -723,7 +723,7 @@ def _read_gdf_header(fname, exclude):
             exclude = _find_exclude_idx(ch_names, exclude)
             sel = list()
             for i, unit in enumerate(units):
-                if unit[:2] == 'uV':
+                if unit[:2] in ('uV', 'µV'):
```
drammock: also pointless?

mne/io/fieldtrip/utils.py
```diff
@@ -20,6 +20,7 @@
               'V': 1,
               'mV': 1e-3,
               'uV': 1e-6,
+              'µV': 1e-6,
```
drammock: also pointless?

mne/io/utils.py
```diff
@@ -50,8 +50,8 @@ def _check_orig_units(orig_units):
 
         # Common "invalid units" can be remapped to their valid equivalent
         remap_dict = dict()
-        remap_dict['uv'] = u'µV'
-        remap_dict[u'μv'] = u'µV'  # greek letter mu vs micro sign. use micro
+        remap_dict['uv'] = 'µV'
+        remap_dict['μv'] = 'µV'  # greek letter mu vs micro sign. use micro
```
drammock: here I'm only removing the unicode string designator, not changing the string content at all.

mne/io/tests/test_utils.py
```diff
@@ -9,11 +9,11 @@
...
+                      microSign='µV')
     orig_units = _check_orig_units(orig_units)
     assert orig_units['FC1'] == 'nV'
     assert orig_units['Hfp3erz'] == 'n/a'
-    assert orig_units['Pz'] == u'µV'
-    assert orig_units['greekMu'] == u'µV'
-    assert orig_units['microSign'] == u'µV'
+    assert orig_units['Pz'] == 'µV'
+    assert orig_units['greekMu'] == 'µV'
+    assert orig_units['microSign'] == 'µV'
```
drammock: here also I'm only removing unicode string designator, not changing string content.

mne/io/edf/edf.py
larsoner: Agreed we shouldn't need to modify internal values until we find files that use these new chars

mne/io/edf/edf.py
larsoner: I would say so

mne/io/fieldtrip/utils.py
larsoner: I think so

larsoner:
Approved. Though I am a bit disappointed you did not name your branch `µ` to see if `git` and GitHub could handle it :)

agramfort:
thx @drammock 

and @larsoner for the nerdy joke to wake me up :)

sappelhoff:
some more nerd information, proudly presented by the unicode consortium:

> Micro  sign is  included  in  several  parts  of ISO/IEC  8859,  and  therefore  supported  in  many  legacy  environments  where  U+03BC μ GREEKLETTERSMALLMU is not available. Implementations therefore need to be able to recognize the micro sign, **even though U+03BC μ is the preferred character in a Unicode context.**

(emphasis added by me)

it's not really relevant to this PR, but I was on a crusade to purge greek-small-letter mu from everywhere and it was recently brought to my attention that it is actually preferred over the micro sign :man_shrugging: 

see in http://www.unicode.org/reports/tr25/

drammock:
🤦

PR #7064, opened 2019-11-14
MRG: Change way NIRS frequency is encoded

rob-luke:
#### Reference issue
Addresses issue #7057

#### What does this implement/fix?
Change the way NIRS light frequency is encoded as [suggested by @larsoner](https://github.com/mne-tools/mne-python/pull/6674#issuecomment-546908907)

PR #7499, opened 2020-03-23
MRG: Add test that NIRS naming matches stored frequency

rob-luke:
#### Reference issue
General fNIRS tweaks #7057.

In #7064 the NIRS light frequency was moved from being stored in the channel name to being stored in `info['chs'][ii]['loc'][9]`. However, the frequency is still stored in the channel name for ease of reading, plotting, and because MNE throws an error if two channels have the same name. 

This PR adds a check that the light frequency stored in the `ch_name` and in `loc` match. And adds tests.

#### Additional information
People shouldn't mess with the channel names and this is specified in the docs already. https://github.com/mne-tools/mne-python/blob/master/tutorials/io/plot_30_reading_fnirs_data.py#L26

### DIFF
--- a/mne/preprocessing/nirs/nirs.py
+++ b/mne/preprocessing/nirs/nirs.py
@@ -81,6 +81,18 @@ def _check_channels_ordered(raw, freqs):
         ch2_name_info = re.match(r'S(\d+)_D(\d+) (\d+)',
                                  raw.info['chs'][ii + 1]['ch_name'])
 
+        if raw.info['chs'][ii]['loc'][9] != \
+                float(ch1_name_info.groups()[2]) or \
+                raw.info['chs'][ii + 1]['loc'][9] != \
+                float(ch2_name_info.groups()[2]):
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


rob-luke:
Just fixed a typo. But I am done now.

rob-luke:
@larsoner I always prefer more checks, but I guess these things slow the code down. Whats the general preference around here?

larsoner:
> I guess these things slow the code down. Whats the general preference around here?

In general more checks are better. Typically we have slowdowns in the hundreds of ms to seconds range (even plotting is here), so something like this that adds what I assume is < 1 ms is fine.

larsoner:
I should say typically we have these slowdowns *when we want to do something interesting*, not in checks like this...

rob-luke:
> In general more checks are better.

Thanks.

circle is failing, but looks unrelated

larsoner:
The 3.7 pip build on Travis is probably also going to fail, both hopefully will be fixed by #7501

agramfort:
thx @rob-luke

PR #7507, opened 2020-03-24
MRG, ENH: Allow picking without preload

### LINKED ISSUE #1766: Preload subset of channels or channel types from raw.fif
I have simultaneous MEG/EEG data sampled at 5 kHz (because I'm also recording brainstem activity) and it would nice to be able to preload say only EEG or only subsets of channels etc. and do analysis/eye-balling..  There are probably other good use cases for this too.. 

I can see this being possible quickly in a few different ways:
(1) Add kwargs like eeg=True, meg=False etc. to pick only certain channel types..
(2) Take channel names (as opposed to indices) list for a pickNames argument.. 
(3) Add a regular picks argument and then the user does two steps... (i) load with preload=False and determine picks from the info and then (ii) re-run the file read with preload=True and picks=picks

Thoughts?

larsoner:
I realized that a clean path to picking without preload is to follow this to-do list:

- [x] Ensure that `self._read_segment_file` for all readers is implemented such that:
    1. `self._raw_extras` is always a list of dict, and `self._raw_extras[fi]['orig_nchan']` gets automatically added by `BaseRaw` to store the original number of channels on disk (as it's used by many readers, might as well add it in one place). 
    2. It only uses the attributes `self._filenames` and `self._raw_extras`. This mostly requires removing uses of `self.info`, but these uses are fairly rare.

    Most of our readers were already close to being stuctured this way, so this was a fairly simple refactoring. This is in the current commit. It uses a little wrapper `_ReadSegmentFileProtector` to make sure only the correct two attributes are used. It also changes `idx` to always be a ndarray of int rather than possibly being a slice because this simplifies some code and shouldn't incur much of any performance penalty (will check against master before merge).

    The one annoying thing I had to do to make this work was update `egimff` annotation setting, which was previously being done using a hack marked with `XXX` but now uses `self.set_annotations` like elsewhere in `mne/io/*/*`.

- [x] Make a new variable `self._read_picks` that is used to sub-index when reading. Currently this is just `np.arange(self._raw_extras[fi]['orig_nchan'])`, so it's a no-op, but should allow us to ...

- [x] Add support for `raw.pick_types` without preload by subselecting `self._read_picks`. Currently this is not done because I want to make sure the changes thus far lead to green CIs (I think they will).

- [x] Add tests for pick-without-preload.

- [x] Update examples where picking is used to do it before `load_data` to save memory.

- [x] Double-check that this PR does not slow down I/O with profiling.

- [x] Update `latest.inc`

TL;DR: `BaseRaw` keeps track of which picks from the original data on disk we actually need and doing the appropriate subselection (the new thing), and each individual reader just needs to care about reading a set of `idx` indices of data from the original ones on disk (which has always been the case, just clearer now).

Marking for 0.21 because this is a big enough change that we should probably do it at the beginning of a release cycle rather than the end.

Closes #1766.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -15,6 +15,8 @@ Current (0.21.dev0)
 Changelog
 ~~~~~~~~~
 
+- Allow picking channels in raw instances (e.g., :meth:`mne.io.Raw.pick_types`) without preloading data, by `Eric Larson`_
+
 - Add support for signals in mV for :func:`mne.io.read_raw_brainvision` by `Clemens Brunner`_
 
 Bug
--- a/examples/connectivity/plot_mixed_source_space_connectivity.py
+++ b/examples/connectivity/plot_mixed_source_space_connectivity.py
@@ -69,9 +69,9 @@ vol_src = setup_volume_source_space(
 src += vol_src
 
 # Load data
-raw = read_raw_fif(fname_raw, preload=True)
+raw = read_raw_fif(fname_raw)
+raw.pick_types(meg=True, eeg=False, eog=True, stim=True).load_data()
 events = mne.find_events(raw)
-raw.pick_types(meg=True, eeg=False, eog=True)
 noise_cov = mne.read_cov(fname_cov)
 
 # compute the fwd matrix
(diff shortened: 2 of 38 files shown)


larsoner:
Okay so using Hubert's benchmarking script `benchmark_fif_edf` (uploaded a variant [here](https://gist.github.com/larsoner/a6110c5afb890d78517a070be3cccabe)) I did see a 1-2 ms slowdown due to `idx` not being a slice, so I put code back in to slice-ify it where applicable. Now timings are the same here as on `master`.

larsoner:
This is now ready for review/merge

larsoner:
It was red due to download problems and I restarted it, no idea why GitHub didn't update to reflect the pass

PR #7508, opened 2020-03-24
MRG, ENH: Dont require preload for raw resample

larsoner:
Relaxes the requirement that data be preloaded before resampling. Should help on low-memory machines. Probably worth doing early in 0.21 rather than before cutting 0.20, hence the WIP.

Todo:

- [x] Update `latest.inc`

### DIFF
--- a/mne/io/fiff/tests/test_raw_fiff.py
+++ b/mne/io/fiff/tests/test_raw_fiff.py
@@ -965,6 +965,16 @@ def test_crop():
 
 
 @testing.requires_testing_data
+def test_resample_equiv():
+    """Test resample (with I/O and multiple files)."""
+    raw = read_raw_fif(fif_fname).crop(0, 1)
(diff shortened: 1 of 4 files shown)


larsoner:
This is now ready for review/merge

mne/io/fiff/tests/test_raw_fiff.py
```diff
@@ -965,10 +965,12 @@ def test_crop():
 
 
 @testing.requires_testing_data
-def test_resample(tmpdir):
+@pytest.mark.parametrize('preload', (True, False))
+def test_resample(tmpdir, preload):
```
agramfort, line 969: I would have tested that the values are the same if you preload or not.

larsoner:
@drammock next time you run into memory problems with some example, this and #7507 might make it easier to lower the memory usage

PR #7247, opened 2020-01-23
MRG: Enable point picking in _TimeViewer

GuillaumeFavelier:
This PR enables point picking and display the vertex id as text in the top left. It's the first step to achieve #7235. Shortcut is `p` to pick under the mouse. This is still a work in progress.

**Animation**

output

**ToDo**

https://github.com/mne-tools/mne-python/issues/7235#issuecomment-577959594

Closes #7235 
It's an item of https://github.com/mne-tools/mne-python/issues/7162

PR #7509, opened 2020-03-25
MRG: Change _TimeViewer slider style

GuillaumeFavelier:
It occurs to me that I never asked for feedbacks for the slider style, I just went with what I found practical but apparently the current design has some limitations. For example, since the slider tube is very thin it's easy to miss it and rotate the brain instead (reported in https://github.com/mne-tools/mne-python/pull/7247#issuecomment-583536539 and offline by @hoechenberger). This PR tries to address that by changing the style of the slider (and by doing so, modify it's hitbox).

I suggest 2 styles of sliders. In both styles I removed the end cap. In style 1, I increased the tube width but the overall style stays similar. The style 2 is a bit extreme but fixes the original issue quite easily.

<details>

```py
import os
import mne
initial_time = 0.1
                 hemi='lh',
                 )
```

</details>

`master` | Style 1 | Style 2
------------|-----------|-----------
image | image | image

Except the size, it's also possible to change the colors (*c.f.* https://github.com/pyvista/pyvista/pull/511):

image

Sadly, changing the shape or replacing the handle by a texture would require way more work on the VTK part.

It's an item of #7162

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -417,10 +417,9 @@ class _TimeViewer(object):
         if slider is not None:
             slider_rep = slider.GetRepresentation()
             slider_rep.SetSliderLength(0.02)
-            slider_rep.SetSliderWidth(0.04)
-            slider_rep.SetTubeWidth(0.005)
-            slider_rep.SetEndCapLength(0.01)
-            slider_rep.SetEndCapWidth(0.02)
+            slider_rep.SetSliderWidth(0.06)
+            slider_rep.SetTubeWidth(0.02)
+            slider_rep.GetCapProperty().SetOpacity(0)
             slider_rep.GetSliderProperty().SetColor((0.5, 0.5, 0.5))
             if not show_label:
                 slider_rep.ShowSliderLabelOff()
@@ -541,7 +540,7 @@ class _TimeViewer(object):
         # Colormap slider
         pointa = np.array((0.82, 0.26))
         pointb = np.array((0.98, 0.26))
-        shift = np.array([0, 0.08])
+        shift = np.array([0, 0.12])
         fmin = self.brain._data["fmin"]
         self.fmin_call = BumpColorbarPoints(
             plotter=self.plotter,
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -417,8 +417,8 @@ class _TimeViewer(object):
         if slider is not None:
             slider_rep = slider.GetRepresentation()
             slider_rep.SetSliderLength(0.02)
-            slider_rep.SetSliderWidth(0.06)
-            slider_rep.SetTubeWidth(0.06)
+            slider_rep.SetSliderWidth(0.04)
+            slider_rep.SetTubeWidth(0.04)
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
What do you think @agramfort, @hoechenberger, @larsoner, @drammock ?

hoechenberger:
+1 for style 2 with slightly less thick tube & slider. This then also mimics the sensor-level plots (e.g. `Raw.plot()`) and it's also the kind of slider used in all (?) modern operating systems / GUIs: tube has the same width as the slider.

GuillaumeFavelier:
Style 3 is just a rework of 2) but with the same size used in `master`.

GuillaumeFavelier:
Also feel free to share if you have any suggestions for the colors.

hoechenberger:
> Style 3 is just a rework of 2) but with the same size used in `master`.

That could work! What do the others think? Sure, I think it's more aesthetically pleasant to look at those sliders where the tube is narrower than the slider element, but I think this here is an important tradeoff to make between eye candy and usability.

Regarding the colors, can you adjust the tube color to be slightly less salient? Very-light-gray-ish or something?

GuillaumeFavelier:
> Regarding the colors, can you adjust the tube color to be slightly less salient? Very-light-gray-ish or something?

Good idea!

hoechenberger:
It's not by any chance possibly to specify a color for the _outline_ (border) of the tube, is it? Bc then you could make the tube black (as the background) and still indicate that this is a slider widget by having a fine outline drawn, inside which the slider can move.

GuillaumeFavelier:
AFAIK not just with sliders but maybe with vtkBorderWidgets...

Reference: https://vtk.org/Wiki/VTK/Examples/Cxx/Widgets/BorderWidget

hoechenberger:
> AFAIK not just with sliders but maybe with vtkBorderWidgets...

Seems too complicated… let's see if a less prominent tube color does the trick for now :)

GuillaumeFavelier:
Trying with my firefox slider on KDE:

image

I obtain:

image

hoechenberger:
@GuillaumeFavelier looks good to me!!

hoechenberger:
@GuillaumeFavelier I just pulled and tried this out and I noticed another (minor) oddity:

When I click a slider somewhere left or right of its horizontal center point and then start dragging, as soon as dragging begins, the slider jumps such that it's centered horizontally below the cursor.

GuillaumeFavelier:
If you click the tube it's normal that the slider jumps but when I drag, the slider moves only at the end of the interaction for me. I don't think I can reproduce this

hoechenberger:
@GuillaumeFavelier I've recorded a video:
https://drive.google.com/open?id=1EkPfPqRT47euD87_E8Idrr4P98PQX2hp

agramfort:
I think we can live with that :)

>

GuillaumeFavelier:
>  I've recorded a video

Thank you, now I understand what you mean.

larsoner:
Looks great to me, I do like the slider filling up the entire height of the tube

GuillaumeFavelier:
This is ready for merge then

larsoner:
Thanks @GuillaumeFavelier

PR #7060, opened 2019-11-14
WIP: Use _Brain in plot_vector_source_estimates

GuillaumeFavelier:
This PR is a follow-up of #6731 suggested also in [this comment](https://github.com/mne-tools/mne-python/pull/6731#issuecomment-536042199). The goal is to make `stc_vec.plot()` use the `_Brain` class which relies on the abstract 3d backend layer `_Renderer`.

**ToDo**

- [x] Fix scaling of the glyphs by using scalar values (merged [upstream](https://github.com/pyvista/pyvista/pull/451))
- [x] Enable Depth Peeling (merged [upstream](https://github.com/pyvista/pyvista/pull/450))
- [x] Increase arrow thickness
- [x] Fix colorbar range
- [x] Update the 3D tests with the `renderer` fixture (c.f. https://github.com/mne-tools/mne-python/pull/7028#issue-337436335)
- [x] Experiment 3D arrows glyphs (feedback from #7030)
- [x] Colormap has a 'grey tone'
- [x] Refactor smoothing matrix generation
- [x] The left and right hemi data have different scales
- [x] Rendering dark filter bug on CircleCI
- [x] remove glyph options (should be future PR)
- [x] `time_viewer=True` only playback only updates mesh, not vectors
- [x] hitting spacebar at the end of playback aborts Python (should restart from beginning), fix was in the reverted PR
- [x] fix glyph flashing during playback (if still an issue)
- [x] get CIs green
- [x] add support for `show_traces=True` if it's not too difficult
- [x] fix smoothing steps bug

It's an item of #7162

PR #7511, opened 2020-03-25
MRG, ENH: Use uniform scale factor

larsoner:
Part of #7060 to see if we can replicate the `test_report.py` error.

Has stubs with `if True` to facilitate future code, but can still be merged I think.


larsoner:
I killed Travis. Looks like Azure is headed toward success, will merge once it's green

PR #7352, opened 2020-02-24
MAINT: Disable problematic test on Azure

larsoner:
Every PR is red because of Azure. I'm going to try:

1. Skipping the `plot_vector_source_estimates` test since it seems to show up in [the tracebacks](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=5634&view=logs&j=b195a6e3-7484-5943-e627-0b35c27605d6&t=6dcb3581-08d4-5e76-74f2-6edaf20ff912&l=257).
2. Removing `mayavi` and `PySurfer` from the Azure builds.

I hope that (1) is enough because (2) is drastic, but we should get Azure green for most PRs and work on fixing this annoying bug separately.

PR #7512, opened 2020-03-25
WIP, MAINT: Restore Mayavi tests to Azure

larsoner:
Reverts #7352. Let's see if things work any better now, though I doubt it will.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,7 +125,6 @@ jobs:
       pip install --upgrade numpy scipy vtk
       pip install --upgrade -r requirements.txt
       pip install codecov
-      pip uninstall -yq pysurfer mayavi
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
I was not very optimistic either. I'll focus on this in particular after 0.20

larsoner:
I think it's probably related to the same thing as the vector PR, likely having to do with setting rendering parameters before the OpenGL context has been activated. In `mne coreg` we listen for the `scene:activated` event:

https://github.com/mne-tools/mne-python/blob/master/mne/gui/_coreg_gui.py#L1866

If you try to plot before this things go badly. I suspect it might be what's happening in some of these cases. This might also be part of the reason why in PySurfer we force GUI event processing -- for example it could force the window to actually be created and OpenGL context to be initialized so that plotting and other OpenGL-related things actually work.

PR #7513, opened 2020-03-25
MRG, MAINT: Update PyVista backend

larsoner:
Let's see if this part of #7060 breaks Azure.


larsoner:
@GuillaumeFavelier this breaks it! I'll try removing the AA stuff and see if that fixes things

GuillaumeFavelier:
+1 for testing AA

larsoner:
Okay @GuillaumeFavelier pushed a commit that hopefully will "fix" things for us

larsoner:
I'll kill these CIs for now and if the fix doesn't work, keep iterating here

PR #7516, opened 2020-03-25
MRG, ENH: Allow color reuse in clicking

larsoner:
Allow color reuse during picking with `_TimeViewer`. It's an item of #7162.

Probably worth adding to 0.20


agramfort:
thx @larsoner

PR #7518, opened 2020-03-26
Disable _TimeViewer when no time info is provided

GuillaumeFavelier:
This PR locally tests the given brain and disables the `_TimeViewer` wrapping when no time information is provided. For instance this happens in `examples/forward/plot_source_space_morphing.py`.

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1753,7 +1753,8 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
                              spacing=spacing, time_viewer=time_viewer,
                              colorbar=colorbar, transparent=transparent)
     time_viewer, show_traces = _check_time_viewer_compatibility(time_viewer,
-                                                                show_traces)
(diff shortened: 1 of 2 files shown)


GuillaumeFavelier:
Please bear with me, I found a cleaner way based on `stc` and `hemi`

mne/viz/_3d.py
```diff
@@ -1849,6 +1865,8 @@ def _check_time_viewer_compatibility(time_viewer, show_traces):
         if not check_version('surfer', '0.9'):
             raise RuntimeError('This function requires pysurfer version '
                                '>= 0.9')
+    if time_viewer:
+        time_viewer = _check_stc_time(stc, hemi)
```
larsoner: We don't want to always disable it, it controls more than time so can still be useful even if there is a single time point

GuillaumeFavelier:
Closing since #7520 suggests a better compromise instead of disabling `_TimeViewer` completely.

PR #7520, opened 2020-03-26
MRG, FIX: Fix single time point

larsoner:
Alternative to #7518.

I think we should show the (incorrectly-named) `_TimeViewer` even if there is a single time point because it shows useful controls like clim, view, etc. (and others if we add them later). This hides the playback speed and time slider as well.

### DIFF
--- a/examples/forward/plot_source_space_morphing.py
+++ b/examples/forward/plot_source_space_morphing.py
@@ -4,8 +4,8 @@ Use source space morphing
 =========================
 
 This example shows how to use source space morphing (as opposed to
-SourceEstimate morphing) to create data that can be compared between
-subjects.
+:class:`~mne.SourceEstimate` morphing) to create data that can be compared
+between subjects.
 
 .. warning:: Source space morphing will likely lead to source spaces that are
              less evenly sampled than source spaces created for individual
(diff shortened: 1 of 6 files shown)


larsoner:
Pushed a commit to fix the PyQt-abort-on-error problem

GuillaumeFavelier:
How do we select the functions wrapped by `safe_event`? Because I would like to suggest `self.play()` as well.

larsoner:
In principle, anything that could cause an error should be guarded. But in the special case of `play` we probably want to do more, because we should stop the playback, otherwise the user will get spammed with messages

larsoner:
@GuillaumeFavelier @agramfort this should be ready for review/merge now

larsoner:
Maybe it would require a PyVista PR? Or is there some general VTK renderer update blocking that we could use?

GuillaumeFavelier:
I think a trick similar to https://github.com/mne-tools/mne-python/blob/master/mne/viz/_brain/_brain.py#L947-L963 would be necessary where we configure the representation of the widget to be hidden by default. This would require a PR on pyvista.

>  is there some general VTK renderer update blocking

hm... I don't know any. I tried hiding the window (or interactor) and display at the end but although the window reappears ready, there is the fact that it disappears out of the blue.

GuillaumeFavelier:
Maybe the rendering thread could be paused...

Reference: https://github.com/pyvista/pyvista/blob/master/pyvista/plotting/qt_plotting.py#L449

PR #7522, opened 2020-03-26
MRG, DOC: Add data channels to glossary

larsoner:
Adds `data channels` to our glossary with a couple of refs (can add more later as we find them). Also clarifies a bit what happens by default with `create_info`, as these two conspired to trip up @arokem on Gitter.

@drammock can you look at the output and see if you'd like to tweak, then merge?


larsoner:
This is what the output looks like locally:

Screenshot from 2020-03-26 17-08-38

larsoner:
Setting to WIP because I realized that we should also provide the mapping to human-readable names in there...

drammock:
Based on your screenshot, the definition and the list don't jive. It looks like you just listed all channel types.

larsoner:
Indeed misc is in there. These private variables need some renaming...

larsoner:
[Should be fixed](https://18964-1301584-gh.circle-artifacts.com/0/dev/glossary.html#term-data-channels)

Also I deprecated `ch_types=None` because there is no reason to have it instead of just `ch_types='misc'`.

mne/defaults.py
```diff
@@ -37,8 +37,8 @@
     titles=dict(mag='Magnetometers', grad='Gradiometers', eeg='EEG', eog='EOG',
                 ecg='ECG', emg='EMG', misc='misc', seeg='sEEG', bio='BIO',
                 dipole='Dipole', ecog='ECoG', hbo='Oxyhemoglobin',
-                ref_meg='Reference Magnetometers', fnirs_raw='fNIRS',
-                fnirs_od='fNIRS', hbr='Deoxyhemoglobin',
+                ref_meg='Reference Magnetometers', fnirs_raw='fNIRS (raw)',
+                fnirs_od='fNIRS (OD)', hbr='Deoxyhemoglobin',
```
larsoner: @rob-luke okay for you?

rob-luke: All good by me thanks :)

PR #7523, opened 2020-03-26
DOC: Fixes a typo in ICA docstring

arokem:

"principle component" => "principal components"


mmagnuski:
I'm not sure about this one: wikipedia gives singular "component"
https://en.wikipedia.org/wiki/Principal_component_analysis
is there an error there too?
but I'm no expert on english language. :)

mmagnuski:
BTW, the wikipedia also uses the plural but less frequently.

arokem:
I think that you are right. I think that both are allowable, but "component" is in more common use (e.g., google search for "principal components analysis" gives about 2.7 M pages and "principal component analysis" gives >10 M pages.

Fixed back in recent commit.

larsoner:
Thanks @arokem

Issue #7526, opened 2020-03-27
Website examples after #7520


GuillaumeFavelier:
I inspected the website frontpage after #7520 and I noticed 3 issues:

1) Taking screenshot with time viewer is not consistent with what we had previously
https://18971-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_publication_figure.html#sphx-glr-auto-examples-visualization-plot-publication-figure-py

![image](https://user-images.githubusercontent.com/18143289/77763381-428ade80-703b-11ea-9643-0eee85cb49ee.png)

2) The titles overlap
https://18971-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_psf_ctf_vertices_lcmv.html#sphx-glr-auto-examples-inverse-plot-psf-ctf-vertices-lcmv-py

![image](https://user-images.githubusercontent.com/18143289/77763395-4b7bb000-703b-11ea-9407-60d7cd66bcaa.png)

3) The figure is still registered after `close()`
https://18971-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_publication_figure.html#sphx-glr-auto-examples-visualization-plot-publication-figure-py

![image](https://user-images.githubusercontent.com/18143289/77763591-9c8ba400-703b-11ea-8e58-d03bd43414fe.png)

larsoner:
Can you make a PR to fix these?

GuillaumeFavelier:
I have something locally, I'll take care of 1)

PR #7527, opened 2020-03-27
Taking a screenshot with _TimeViewer

GuillaumeFavelier:
This PR fixes the issue with taking screenshot with `time_viewer=True` by wrapping `_Brain.screenshot()`

Fixes one item of #7526 

*Note: it is the second time I do such function wrapping, maybe using inheritance instead of composition should be considered*


GuillaumeFavelier:
The following does not seem related:

```
doc/changes/0.20.inc:399: Sherif  ==> Sheriff
Makefile:111: recipe for target 'codespell-error' failed
make: *** [codespell-error] Error 1
```

larsoner:
I'm not sure whether or not we want the interface disabled during the screenshot, as it won't match what the user sees (a key goal of sphinx-gallery capturing). In this case I'd rather just do `time_viewer=False` in the calls

examples/visualization/plot_publication_figure.py
```diff
@@ -44,7 +44,7 @@
 evoked.plot()
 
 stc.plot(views='lat', hemi='split', size=(800, 400), subject='sample',
-         subjects_dir=subjects_dir, initial_time=max_t)
+         subjects_dir=subjects_dir, initial_time=max_t, time_viewer=False)
```
larsoner: show_traces=False, too?

GuillaumeFavelier: normally, only `time_viewer=False` should be necessary but it's maybe clearer with `show_traces=False` too

larsoner:
Appears to still require merge @GuillaumeFavelier

GuillaumeFavelier:
Oh sorry I forgot to set `show_traces=False`

larsoner:
Thanks for the quick fix @GuillaumeFavelier

Issue #7521, opened 2020-03-26
BUG: Overflow error with raw.plot()


cbrnr:
I have a specific BrainVision file that I cannot plot:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py", line 152, in draw_path
    self._renderer.draw_path(gc, path, transform, rgbFace)
OverflowError: In draw_path: Exceeded cell block limit

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/matplotlib/backends/backend_macosx.py", line 74, in _draw
    self.figure.draw(renderer)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/figure.py", line 1735, in draw
    mimage._draw_list_compositing_images(
  File "/usr/local/lib/python3.8/site-packages/matplotlib/image.py", line 137, in _draw_list_compositing_images
    a.draw(renderer)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 2630, in draw
    mimage._draw_list_compositing_images(renderer, self, artists)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/image.py", line 137, in _draw_list_compositing_images
    a.draw(renderer)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/artist.py", line 38, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/matplotlib/lines.py", line 802, in draw
    renderer.draw_path(gc, tpath, affine.frozen())
  File "/usr/local/lib/python3.8/site-packages/matplotlib/backends/backend_agg.py", line 154, in draw_path
    raise OverflowError("Exceeded cell block limit (set "
OverflowError: Exceeded cell block limit (set 'agg.path.chunksize' rcparam)

If you suspect this is an IPython 7.13.0 bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@python.org

[...]

PR #7529, opened 2020-03-27
MRG, ENH: Allow float clipping

larsoner:
This improves our clipping to use `matplotlib` to cut off values rather than using `np.nan`. This is better because `np.nan` can cause the entire trace to be removed, this will cause it to still exist but just be invisible outside the appropriate "box".

Also proposes changing the default `clipping=None` to `clipping=1.5`, but this is certainly up for debate. WIP because we need to run some examples and decide, and it should go in 0.21.

Todo:

- [x] Decide on default
- [x] `latest.inc`

Some related discussion in #7521

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -37,6 +37,8 @@ Changelog
 
 - Add functionality to interpolate bad NIRS channels by `Robert Luke`_
 
+- Add support to in :meth:`mne.io.Raw.plot` for passing ``clipping`` as a float to clip to a proportion of the dedicated channel range by `Eric Larson`_
+
 - Add function :func:`mne.preprocessing.annotate_muscle_zscore` to annotate periods with muscle artifacts. by `Adonay Nunes`_
 
 - Support for saving movies of source time courses (STCs) with ``brain.save_movie`` method and from graphical user interface by `Guillaume Favelier`_
@@ -45,6 +47,8 @@ Bug
 ~~~
 
 - Fix bug with :func:`mne.minimum_norm.compute_source_psd_epochs` and :func:`mne.minimum_norm.source_band_induced_power` raised errors when ``method='eLORETA'`` by `Eric Larson`_
+
+- The default plotting mode for :func:`mne.io.Raw.plot` and :ref:`gen_mne_browse_raw` has been changed to ``clipping=3.`` to facilitate data analysis with large deflections, by `Eric Larson`_
 
 - Fix handling of NaN when using TFCE in clustering functions such as :func:`mne.stats.spatio_temporal_cluster_1samp_test` by `Eric Larson`_
 
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -90,6 +90,7 @@ def run():
     lowpass = options.lowpass
     filtorder = options.filtorder
     clipping = options.clipping
+    clipping = None if clipping.lower() == 'none' else clipping
     try:
         clipping = float(clipping)  # allow float and convert it
     except Exception:
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -94,7 +94,7 @@ def run():
         clipping = None if clipping.lower() == 'none' else clipping
     try:
         clipping = float(clipping)  # allow float and convert it
-    except ValueError:
+    except (TypeError, ValueError):
         pass
     filterchpi = options.filterchpi
     verbose = options.verbose
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -20,6 +20,7 @@ def run():
     """Run command."""
     import matplotlib.pyplot as plt
     from mne.commands.utils import get_optparser, _add_verbose_flag
+    from mne.viz import _RAW_CLIP_DEF
 
     parser = get_optparser(__file__, usage='usage: %prog raw [options]')
 
@@ -65,7 +66,7 @@ def run():
                       default=4)
     parser.add_option("--clipping", dest="clipping",
                       help="Enable trace clipping mode, either 'clamp' or "
-                      "'transparent'", default=1.5)
+                      "'transparent'", default=_RAW_CLIP_DEF)
     parser.add_option("--filterchpi", dest="filterchpi",
                       help="Enable filtering cHPI signals.", default=None,
                       action="store_true")
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -90,7 +90,8 @@ def run():
     lowpass = options.lowpass
     filtorder = options.filtorder
     clipping = options.clipping
-    clipping = None if clipping.lower() == 'none' else clipping
+    if isinstance(clipping, str):
+        clipping = None if clipping.lower() == 'none' else clipping
     try:
         clipping = float(clipping)  # allow float and convert it
     except Exception:
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -65,7 +65,7 @@ def run():
                       default=4)
     parser.add_option("--clipping", dest="clipping",
                       help="Enable trace clipping mode, either 'clamp' or "
-                      "'transparent'", default=None)
+                      "'transparent'", default=1.5)
     parser.add_option("--filterchpi", dest="filterchpi",
                       help="Enable filtering cHPI signals.", default=None,
                       action="store_true")
@@ -89,6 +89,10 @@ def run():
     lowpass = options.lowpass
     filtorder = options.filtorder
     clipping = options.clipping
+    try:
+        clipping = float(clipping)  # allow float and convert it
+    except Exception:
+        pass
     filterchpi = options.filterchpi
     verbose = options.verbose
 
--- a/mne/commands/mne_browse_raw.py
+++ b/mne/commands/mne_browse_raw.py
@@ -94,7 +94,7 @@ def run():
         clipping = None if clipping.lower() == 'none' else clipping
     try:
         clipping = float(clipping)  # allow float and convert it
-    except Exception:
+    except ValueError:
         pass
     filterchpi = options.filterchpi
     verbose = options.verbose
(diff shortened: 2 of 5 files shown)


larsoner:
Here is an example of `1.5` on `sample` data:

Screenshot from 2020-03-27 11-51-21

larsoner:
@drammock feel free to try this one if you want

drammock:
I think `clipping=1.5` is too aggressive to be a useful default. It makes artifacts look weird. Here's a screenshot from the SSP tutorial, built locally using this branch:

Screenshot_2020-04-01_10-27-19

EEG data with large movement-induced drifts, bad electrodes, etc, is going to be even worse, and I worry that people *won't notice* extremely bad channels because of this setting.  I think I'd be fine with either leaving the default as `None` and adding some tutorial text about the `clipping` paramter, or setting the clipping value fairly high as default (3?  5?  10?  The best answer to "how many channels should a trace cut across before it disappears?" seems like it might depend on how many channels are currently being displayed...)

larsoner:
Want to try setting the default to 3 or 5 and playing with some data to see if it in general works better than `None`?

cbrnr:
We don't even need to change the default as long as we make the error message in the currently failing example clearer (i.e. suggest to use a specific `clipping` value).

larsoner:
> We don't even need to change the default as long as we make the error message in the currently failing example clearer (i.e. suggest to use a specific clipping value).

Agreed this is not meant to be a fix for the problem in your failing example, this is more about potentially changing the default so that it generally works better for data of all sorts

drammock:
I tried some fairly noisy EEG data at a few different values. TLDR: `clipping=3.` was a good compromise for that data.  I realized also how `clipping` interacts with the `-` and `+` keys for changing scaling on the fly; IMO a demonstration of how these two interact is a good idea to add to one of the tutorials.  Screenshots follow:

Default (`clipping=1.5`)

new-default

`clipping=3`

clipping-3

`clipping=5`

clipping-5

larsoner:
Looks reasonable to me. Also makes me realize that I need to `np.clip` the yvals for the clipping mask to make sure nothing exceeds the plotting box...

larsoner:
After some testing with noisy data, I agree with @drammock that `3.` is a good default. `latest.inc` updated, ready for review/merge from my end

doc/changes/latest.inc
```diff
@@ -46,6 +48,8 @@ Bug
 
 - Fix bug with :func:`mne.minimum_norm.compute_source_psd_epochs` and :func:`mne.minimum_norm.source_band_induced_power` raised errors when ``method='eLORETA'`` by `Eric Larson`_
 
+- The default plotting mode for :func:`mne.io.Raw.plot` and :ref:`gen_mne_browse_raw` has been changed to ``clipping=3.`` to facilitate data analysis with large deflections, by `Eric Larson`_
```
hoechenberger, line 51: Shouldn't that go to the API section?

doc/changes/latest.inc
```diff
@@ -83,8 +87,4 @@ Bug
 API
 ~~~
 
-- Python 3.5 is no longer supported, Python 3.6+ is required, by `Eric Larson`_
-
-- ``vmin`` and ``vmax`` parameters are deprecated in :meth:`mne.Epochs.plot_psd_topomap` and :func:`mne.viz.plot_epochs_psd_topomap`; use new ``vlim`` parameter instead, by `Daniel McCloy`_.
-
-- The method ``stc_mixed.plot_surface`` for a :class:`mne.MixedSourceEstimate` has been deprecated in favor of :meth:`stc.surface().plot(...) <mne.MixedSourceEstimate.surface>` by `Eric Larson`_
+(??)- Nothing yet
```
hoechenberger, line 90: Don't forget to remove this before merging :)

mne/commands/mne_browse_raw.py
```diff
@@ -89,6 +90,12 @@ def run():
     lowpass = options.lowpass
     filtorder = options.filtorder
     clipping = options.clipping
+    if isinstance(clipping, str):
+        clipping = None if clipping.lower() == 'none' else clipping
+    try:
+        clipping = float(clipping)  # allow float and convert it
+    except Exception:
```
hoechenberger, line 97: Can we be more specific here? We only want to proceed in case of `TypeError` and `ValueError`, right?

```suggestion
    except TypeError, ValueError:
```

mne/viz/raw.py
```diff
@@ -97,12 +97,15 @@ def _pick_bad_channels(event, params):
     _plot_update_raw_proj(params, None)
 
 
+_RAW_CLIP_DEF = 3.
```
hoechenberger, line 100: Should this maybe be a `config` setting? Or wouldn't that make any sense?

mne/viz/raw.py
```diff
@@ -185,11 +188,17 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
...
            Support for ``filtorder=0`` to use FIR filtering.
-    clipping : str | None
+    clipping : str | float | None
         If None, channels are allowed to exceed their designated bounds in
         the plot. If "clamp", then values are clamped to the appropriate
         range for display, creating step-like artifacts. If "transparent",
         then excessive values are not shown, creating gaps in the traces.
+        If float, values are outside the float factor ``clipping`` beyond their
+        dedicated range, so ``clipping=1.`` is an alias for
+        ``clipping='transparent'``.
```
hoechenberger, line 198: I find this hard to understand. How about sth like:

```suggestion
        If float, clipping occurs for values exceeding their dedicated range by
        the factor ``clipping``. Consequently, ``clipping=1.`` is an alias for
        ``clipping='transparent'``.
```

However I can see that "exceeding X by factor N" would actually imply that, if `N=1`, we would _excced_ X by `1 X`, i.e., the value would be at `2 X`.

Yet this seems to be common language usage in the US, I suppose? E.g. tech companies always promote their products as being "4x faster" when actually the new speed is 4x the old speed, i.e. they're actually only 3 times faster, but _4 times as fast_.

Opinions?

mne/viz/raw.py
```diff
@@ -785,6 +795,20 @@ def _plot_raw_traces(params, color, bad_color, event_lines=None,
...
 
+            # clip to range (if relevant)
+            if params['clipping'] == 'clamp':
+                np.clip(this_data, -1, 1, out=this_data)
+            elif params['clipping'] is not None:
+                l, w = this_t[0], this_t[-1] - this_t[0]
+                ylim = params['ax'].get_ylim()
+                b = offset - params['clipping']  # max(, ylim[0])
+                h = 2 * params['clipping']  # min(, ylim[1] - b)
+                assert ylim[1] <= ylim[0]  # inverted
```
hoechenberger, line 806: Why is this inverted..?

doc/changes/latest.inc
larsoner: More of a bugfix (and no deprecation cycle)

mne/commands/mne_browse_raw.py
larsoner: I wasn't sure what all the conditions were for `float(str)` raising an error (didn't want to bother looking), but if you're sure it's just those two I can change it

larsoner: In some testing it seems just to be `ValueError` so I'll go with that for now

mne/viz/raw.py
larsoner: No this is just for internal convenience, not user setting

mne/viz/raw.py
larsoner: Pushing a commit to reword now

mne/viz/raw.py
larsoner: It's just internally how we plot these, I don't remember the reasons...

mne/commands/mne_browse_raw.py
hoechenberger: `float(None)` throws a `TypeError` :) So I guess my proposal shoud be safe

Marking as "Unresolved" so the `Commit suggestion` button becomes active again.

hoechenberger: Oh, you had already pushed a change to this line. Never mind.

mne/viz/raw.py
hoechenberger: Hahaha ok fair enough 😅

mne/commands/mne_browse_raw.py
larsoner: `None` at this point shouldn't be able to get through given our default, I guess I could change it in case we change our default back to `None` at some point?

mne/commands/mne_browse_raw.py
```diff
@@ -89,6 +90,12 @@ def run():
     lowpass = options.lowpass
     filtorder = options.filtorder
     clipping = options.clipping
+    if isinstance(clipping, str):
+        clipping = None if clipping.lower() == 'none' else clipping
+    try:
+        clipping = float(clipping)  # allow float and convert it
+    except ValueError:
+        pass
```
hoechenberger, line 98: `float(x)` will raise `TypeError` if `x is None`. We should explicitly handle this case as we're setting `clipping = None` in certain situations.


```suggestion


    if isinstance(clipping, str):
        clipping = None if clipping.lower() == 'none' else clipping

    if clipping is not None:
        try:
            clipping = float(clipping)  # allow float and convert it
        except ValueError:
            pass
```

mne/commands/mne_browse_raw.py
hoechenberger: > `None` at this point shouldn't be able to get through given our default,

But we _do_ set `clipping = None` in certain situations in line 94 :)
see https://github.com/mne-tools/mne-python/pull/7529#pullrequestreview-398231697

mne/commands/mne_browse_raw.py
larsoner: done

(Didn't use the suggestion because it had PEP8 errors)

larsoner:
Moved

larsoner:
(though if we decide at some point to change `_RAW_CLIP_DEFAULT=None`, the TypeError needs to go back -- so I'm going to put it back just to be safe, shouldn't be harm in keeping it)

larsoner:
Pushed a commit to be as limiting as possible with the conditionals and `try/except`

hoechenberger:
Perfect, thanks @larsoner. I'm happy now :)

hoechenberger:
@cbrnr @drammock You happy too?

cbrnr:
Thanks @larsoner!

PR #7531, opened 2020-03-27
MRG, MAINT: Remove deprecated code

larsoner:
The most fun step of releasing, code purge. This one looks pretty good at +113/-1118, let's see if CIs agree with the changes.


drammock:
the `to_data_frame` deprecations should be left in, I think?  The changes to `to_data_frame` haven't been through a stable release yet.

larsoner:
The marking for this is admittedly confusing. 0.20 has been branched and pushed to PyPi, so this changeset, which goes in 0.21, is marked as 0.20 only because it's part of the 0.20 release process.

I'll remove the milestone, though, since it's more confusion than it's worth...

larsoner:
Okay in it goes!

PR #7533, opened 2020-03-28
DOC: Make compute_raw_covariance easier to find

hoechenberger:

A quick support request from @virvw made it obvious to me that `mne.compute_raw_covariance` is currently not so easy to discover as it should be; users were trying to use `mne.compute_covariance` for empty-room data (and, consequently, running into trouble),
without realizing that `compute_raw_covariance` was readily available. This commit adds a clearly-visible note to the docstring to help guide users better.

### DIFF
--- a/mne/cov.py
+++ b/mne/cov.py
@@ -361,6 +361,9 @@ def compute_raw_covariance(raw, tmin=0, tmax=None, tstep=0.2, reject=None,
     It is typically useful to estimate a noise covariance from empty room
     data or time intervals before starting the stimulation.
 
+    .. note:: To estimate the noise covariance from epoched data, use
+              :func:`mne.compute_covariance` instead.
+
     Parameters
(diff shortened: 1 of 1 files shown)


agramfort:
You could also add the see also sections

mne/cov.py
```diff
@@ -693,7 +697,8 @@ def compute_covariance(epochs, keep_sample_mean=True, tmin=None, tmax=None,
 
     See Also
     --------
-    compute_raw_covariance : Estimate noise covariance from raw data.
+    compute_raw_covariance : Estimate noise covariance from raw data, such as
+        empty-room recordings.
```
agramfort: Can you link in the reverse direction too?

hoechenberger: Done

agramfort:
thx @hoechenberger

PR #7535, opened 2020-03-30
MRG: Fix bug in scalp coupling index.

rob-luke:
The filter parameters for scalp coupling index were being passed incorrectly to the downstream function.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -20,7 +20,7 @@ Changelog
 Bug
 ~~~
 
-- Nothing yet
+- Fix bug with :func:`mne.preprocessing.nirs.scalp_coupling_index` where filter transition was incorrectly assigned by `Robert Luke`_.
 
 API
 ~~~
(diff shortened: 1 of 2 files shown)


rob-luke:
@larsoner @agramfort can you please review

agramfort:
can you document the bug fix in the latest.inc and then we'll need to backport this to 0.20

larsoner:
Backported without `0.20.inc` update. I realized that those should be done and pushed only once we've actually pushed the `0.20.1` release to PyPi, otherwise they'll show up on the `stable` website's `whats_new` page before they're actually out...

larsoner:
So I propose with each thing to backport:

1. Backport code changes with `cherry-pick`, push to `maint/0.20`
2. Label each backported PR as `backported`
3. Wait until it's time to put out a new minor release
4. Go through the list of `backported`-labeled PRs and update `maint/0.20`'s `0.20.inc` file, push to `maint/0.20`
3. Push 0.20.1 to PyPi

This should keep PyPi's version and `stable` in sync

drammock:
Would it be easier to put the "backport" label right in the changelog entry, rather than on the PR?

agramfort:
thanks

>

larsoner:
> Would it be easier to put the "backport" label right in the changelog entry, rather than on the PR?

I like the GitHub labels because as they are added one by one you can remove labels one by one, and subselect in the `PR` list by that label, etc. If it's just in the changelog, I need to look at that list, then figure out which PR added it (pretty easy in VScode with `git` auto-highlighting but still not as easy as using GH labels)

PR #7539, opened 2020-03-31
MRG: Add mV support to BrainVision reader

### LINKED ISSUE #7537: read_raw_brainvision should support mV
While working on https://github.com/bids-standard/pybv/pull/39 (an option to explicitly specify the unit for the exported file), I noticed that `read_raw_brainvision` only supports signals in V, µV or nV, but not mV (https://github.com/mne-tools/mne-python/blob/master/mne/io/brainvision/brainvision.py#L504). Would it be OK if I added the missing mV unit?

cbrnr:
Fixes #7537.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -15,12 +15,12 @@ Current (0.21.dev0)
 Changelog
 ~~~~~~~~~
 
-- Nothing yet
+- Add support for signals in mV for :func:`mne.io.read_raw_brainvision` by `Clemens Brunner`_
 
 Bug
 ~~~
 
-- Fix bug with :func:`mne.preprocessing.nirs.scalp_coupling_index` where filter transition was incorrectly assigned by `Robert Luke`_.
+- Fix bug with :func:`mne.preprocessing.nirs.scalp_coupling_index` where filter transition was incorrectly assigned by `Robert Luke`_
 
 API
 ~~~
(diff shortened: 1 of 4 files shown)


agramfort:
any test that can be trivially updated to check that it works?

cbrnr:
For a test we would need to add at least one additional .vhdr file (which would specify channel units in mV). Are you OK with that?

I'm already testing the behavior in https://github.com/bids-standard/pybv/pull/39 but I guess we should have our own test here.

cbrnr:
Or I could modify the existing `test_nV.vhdr` and include other units (i.e. `V` and `mV`).

larsoner:
You can probably just fake one. Copy some existing .eeg file to a tmpdir and copy it's vhdr line by line, modifying the one you need. Then read the data and make sure it's allclose with the original data times 1e-3

You could for example reuse this fixture which does some of the annoying steps for you

https://github.com/mne-tools/mne-python/blob/master/mne/io/brainvision/tests/test_brainvision.py#L109

cbrnr:
I edited the existing alternative header file, and for some reason mV does not work yet (V works fine though).

cbrnr:
It seems like the `'range'` entry in the `ch` dict is incorrect for the mV case:

- For nV 'cal` is `500` and `range` is `1e-09`.
- For mV 'cal' is `0.0005` and `range` is `1.0` (should be `1e-03` IMO).
- For V 'cal' is `5e-07` and `range` is `1.0`.

Any ideas where this could be happening?

cbrnr:
Found it, pushing the working version.

agramfort:
don't forget to update latest.inc

thanks

cbrnr:
Done.

larsoner:
Thanks @cbrnr

larsoner:
Backported

PR #7542, opened 2020-03-31
MRG, DOC: cosmetic changes to what's new for 0.20

drammock:
tiny fix to the `whats_new` page for 0.20:

- @bloyl's name was failing to link in one case because it followed directly after a comma (lacked a leading space)
- tiny grammar tweak


larsoner:
backported

PR #7546, opened 2020-04-01
MRG, ENH: Better progressbar smoothing

larsoner:
In testing ([this gist](https://gist.github.com/larsoner/c56b2521925273094208f33aa1e5cff1) and also long, slow clustering operations) this seems to be a better smoothing param. Turns out that `smoothing=0.9` was doing almost nothing. Using `smoothing=0.05` gives much smoother time remaining and ops/sec estimates.

@drammock I think you mentioned that the PB updates were a bit erratic, can you see if this fixes it for you?

### DIFF
--- a/mne/chpi.py
+++ b/mne/chpi.py
@@ -858,20 +858,18 @@ def compute_chpi_amplitudes(raw, t_step_min=0.01, t_window='auto',
         (len(sin_fits['times']),
          len(hpi['freqs']),
          len(sin_fits['proj']['data']['col_names'])))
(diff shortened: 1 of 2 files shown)


larsoner:
Pushed one more commit to fix a "flashing progressbar" problem, because `ProgressBar` should only be used as a context manager when `joblib`/`n_jobs` is being used

larsoner:
Failure is fixed by #7551

larsoner:
Backported

PR #7552, opened 2020-04-02
MAINT: Catch un-nested reads

larsoner:
Seeing the un-nested `read_raw_fif` in #7551 made me wonder if there was a way to catch this sort of thing. Turns out there are ways of tracking refs that shouldn't cause ref cycle problems or anything. Adds < 1 ms to test runs. Gives this output for un-nested reads (where here I've intentionally added a bad one):
```
larsoner@bunk:~/python/mne-python$ pytest mne/tests/test_surface.py 
Test session starts (platform: linux, Python 3.8.0, pytest 5.4.1.dev62+g2d9dac95e, pytest-sugar 0.9.2)
rootdir: /home/larsoner/python/mne-python, inifile: setup.cfg
plugins: mock-1.11.2, hypothesis-5.7.0, cov-2.8.1, timeout-1.3.3, sugar-0.9.2
collecting ... 
------------------------------------------------------ generated xml file: /home/larsoner/python/mne-python/junit-results.xml ------------------------------------------------------
! _pytest.outcomes.Exit: 
Loaded data found during test collection:

    /home/larsoner/python/mne-python/mne/tests/test_surface.py line 35: <Raw | sample_audvis_trunc_raw.fif, 376 x 6007 (20.0 s), ~3.6 MB, data not loaded>

Please nest data loading in tests. !

Results (0.22s):
```
Seems potentially worth it to give people a nice message (and automatically catch) when they don't nest raw/epochs/evoked reads in tests. But maybe it's too magical. I'll let @agramfort decide :)

### DIFF
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -130,6 +130,20 @@ def matplotlib_config():
     cbook.CallbackRegistry = CallbackRegistryReraise
 
 
+def pytest_collectreport(report):
+    """Check that no MNE classes were instantiated during collection."""
+    from mne.io.base import TrackingMixin
+    instances = list()
+    for key in TrackingMixin.__refs__:
+        instances.extend(key._get_instances())
+    if instances:
+        pytest.exit('\nLoaded data found during test collection:\n\n'
+                    '    %s\n\nPlease nest data loading in tests.\n'
+                    % ('\n    '.join('%s line %s: %s'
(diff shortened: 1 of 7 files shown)


larsoner:
I think this probably is too magical / not worth any possible additional maintenance burden, but it passed on the full Travis run so we should be in the clear currently at least.

PR #7554, opened 2020-04-02
DOC: (Hopefully) small clarification to plot_evoked_topomap() docstring

hoechenberger:

I slightly changed the docstrings of the `vmin` and `vmax` parameters of `plot_evoked_topomap()`, and hope they are easier to understand now.

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -1425,12 +1425,14 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None,
         If None, then channels are chosen in the order given above.
     vmin : float | callable | None
         The value specifying the lower bound of the color range.
-        If None, and vmax is None, -vmax is used. Else np.min(data).
-        If callable, the output equals vmin(data). Defaults to None.
+        If None, and vmax is None too, will use the negative of the maximum
+        absolute value of the data (i.e., 0 will be the midpoint). Else
+        np.min(data). If callable, the output equals callable(data). Defaults
+        to None.
     vmax : float | callable | None
         The value specifying the upper bound of the color range.
         If None, the maximum absolute value is used. If callable, the output
-        equals vmax(data). Defaults to None.
+        equals callable(data). Defaults to None.
     cmap : matplotlib colormap | (colormap, bool) | 'interactive' | None
         Colormap to use. If tuple, the first value indicates the colormap to
         use and the second value is a boolean defining interactivity. In
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -1423,16 +1423,13 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None,
         The channel type to plot. For 'grad', the gradiometers are collected in
         pairs and the RMS for each pair is plotted.
         If None, then channels are chosen in the order given above.
-    vmin : float | callable | None
-        The value specifying the lower bound of the color range.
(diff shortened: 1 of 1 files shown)


drammock:
This seems like a case where it would be easier to document both parameters in one entry. Are we allowed to do `vmin, vmax : None | float | callable` as a single entry?

larsoner:
> Are we allowed to do vmin, vmax : None | float | callable as a single entry?

I would say try it and see if `mne/tests/test_docstring_parameters.py` complains, and check the API page rendering.

Another option is just to document everything in `vmin : ` as we do now, and change `vmax : ` just to say "See vmin.".

drammock:
In that case, I would suggest something like this:

```
vmin, vmax : float | callable | None
    Lower and upper bounds of the colormap, in the same units as the data.
    If ``vmin`` and ``vmax`` are both ``None``, they are set at ± the maximum
    absolute value of the data (yielding a colormap with midpoint at 0).
    If only one of ``vmin``, ``vmax`` is ``None``, will use ``min(data)``
    or ``max(data)``, respectively. If callable, should accept a
    :ref:`numpy array <numpy.ndarray>` of data and return a float.
```

hoechenberger:
@drammock sounds excellent, will try this. Great phrasing too!

hoechenberger:
@drammock Do I need to add `numpy.ndarray` to `numpydoc_xref_aliases` in `conf.py`, or how do I get rid of the warning Circle throws?

```
/home/circleci/project/mne/evoked.py:docstring of mne.Evoked.plot_topomap:22: WARNING: undefined label: numpy.ndarray (if the link has no caption the label must precede a section header)
/home/circleci/project/mne/evoked.py:docstring of mne.EvokedArray.plot_topomap:22: WARNING: undefined label: numpy.ndarray (if the link has no caption the label must precede a section header)
/home/circleci/project/mne/viz/topomap.py:docstring of mne.viz.plot_evoked_topomap:25: WARNING: undefined label: numpy.ndarray (if the link has no caption the label must precede a section header)
```

agramfort:
I would avoid the cross-link with numpy and keep it simple and stupid

hoechenberger:
Circle is happy now, Travis is struggling but unrelated.

btw @larsoner I don't have that "Restart Travis job" button, could you grant me the required permission to restart individual jobs?

hoechenberger:
@agramfort If you're happy I believe we can merge, since @drammock kinda implicitly approved this, as the change I added was proposed by him, sans for my small NumPy reference adjustmen.

agramfort:
LGTM

https://19114-1301584-gh.circle-artifacts.com/0/dev/generated/mne.viz.plot_evoked_topomap.html?highlight=plot_evoked_topomap#mne.viz.plot_evoked_topomap

drammock:
Thanks @hoechenberger. FYI, I think the problem was that the link to ndarray should have been a `func` instead of a `ref`

hoechenberger:
Ok, thanks! Should I change that in another PR or is the class-ref we have now good enough?

drammock:
Oh, never mind. I didn't realize you'd resolved the linking; `class` is correct. I was just going by Alex's comment that you should remove it, and assuming that was what had happened. And then I mis-remembered the directive. That'll teach me to comment before I've had breakfast.

hoechenberger:
Thx for the clarification, @drammock  :) And keep those comments coming, even before breakfast! ;)

PR #7558, opened 2020-04-03
ENH: Custom figure title when plotting Dipole locations

hoechenberger:
Allow to specify a custom figure title when plotting `Dipole` locations via `Dipole.plot_locations()` or `viz.plot_dipole_locations()`.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -994,6 +994,14 @@ df : instance of pandas.DataFrame
     packages.
 """
 
+# Dipole
+docdict['dipole_locs_fig_title'] = """
+title : str | None
+    The title of the figure if ``mode='ortoview'`` (ignored for all other
+    modes). If ``None``, dipole number and its properties (amplitude,
+    orientation etc.) will be shown. Defaults to ``None``.
+"""
+
 # Finalize
 docdict = unindent_dict(docdict)
 fill_doc = filldoc(docdict, unindent_params=False)
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2685,7 +2685,7 @@ def plot_dipole_locations(dipoles, trans=None, subject=None, subjects_dir=None,
                           mode='orthoview', coord_frame='mri', idx='gof',
                           show_all=True, ax=None, block=False, show=True,
                           scale=5e-3, color=None, highlight_color='r',
-                          fig=None, verbose=None):
+                          fig=None, verbose=None, title=None):
     """Plot dipole locations.
 
     If mode is set to 'arrow' or 'sphere', only the location of the first
@@ -2766,6 +2766,9 @@ def plot_dipole_locations(dipoles, trans=None, subject=None, subjects_dir=None,
 
         .. versionadded:: 0.19.0
     %(verbose)s
+    %(dipole_locs_fig_title)s
+
+        .. versionadded:: 0.21.0
 
     Returns
     -------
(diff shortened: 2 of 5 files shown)


mne/utils/docs.py
```diff
@@ -994,6 +994,14 @@
     packages.
 """
 
+# Dipole
+docdict['dipole_locs_fig_title'] = """
+title : str | None
+    The title of the figure if ``mode='ortoview'`` (ignored for all other
```
agramfort, line 1000:
```suggestion
    The title of the figure if ``mode='orthoview'`` (ignored for all other
```

hoechenberger:
> can't you do what you want with
> 
> plt.gcf().suptitle('')
> 
> after the plot?
> 
> matplotlib figures are easy to modify even after they have been displayed

Yes that would certainly be possible, however since many other viz functions already have `title` or `window_title` parameters to allow setting the titles, I thought it would be good to have similar support while `Dipole` plotting as well, for the sake of consistency (and, obviously, usability!)

agramfort:
ok fair enough.

however adding 1s of test for this seems too much. Our test suite is
already way to long.
Can't you just update a existing test?

hoechenberger:
> Can't you just update a existing test?

Done in c9e0cd9
Thx for pointing this out

mne/viz/tests/test_3d.py
```diff
@@ -494,11 +494,14 @@ def test_plot_dipole_mri_orthoview():
     """Test mpl dipole plotting."""
     dipoles = read_dipole(dip_fname)
     trans = read_trans(trans_fname)
-    for coord_frame, idx, show_all in zip(['head', 'mri'],
-                                          ['gof', 'amplitude'], [True, False]):
+    for coord_frame, idx, show_all, title in zip(['head', 'mri'],
+                                                 ['gof', 'amplitude'],
+                                                 [True, False],
+                                                 ['Test', None]):
```
drammock, line 500: now would be a good time to change this to a pytest parametrize approach, if you're up for it.

mne/viz/_3d.py
```diff
@@ -3117,10 +3121,15 @@ def _plot_dipole(ax, data, vox, idx, dipole, gridx, gridy, ori, coord_frame,
...
-        dipole.amplitude[idx] * 1e9, coord_frame_name) +
-        '(%0.1f, %0.1f, %0.1f) mm' % tuple(xyz[idx]))
+
+    if title is None:
+        plt.suptitle('Dipole #%s / %s @ %.3fs, GOF: %.1f%%, %.1fnAm\n%s: ' % (
+            idx + 1, len(dipole.times), dipole.times[idx], dipole.gof[idx],
+            dipole.amplitude[idx] * 1e9, coord_frame_name) +
+            '(%0.1f, %0.1f, %0.1f) mm' % tuple(xyz[idx]))
+    else:
+        plt.suptitle(title)
```
larsoner, line 3131: I would just do:
```
if title is None:
    title = ...
fig.suptitle(title)
```
It modernizes to use the class, plus is more DRY. But I can live with this variant

hoechenberger: Totally agree, just didn't dare to touch the existing `plt` use because I wasn't sure whether it's there for a reason. Will change tomorrow. :)

mne/viz/tests/test_3d.py
hoechenberger: Definitely, thanks for your suggestion.

mne/viz/_3d.py
hoechenberger: Addressed in ba3e62e

mne/viz/tests/test_3d.py
hoechenberger: Addressed in 382586b

larsoner:
Thanks @hoechenberger

PR #7560, opened 2020-04-03
MRG, FIX: Fix read_epochs w/complex data, preload=False

larsoner:
Fixes a bug pointed out on the mailing list.


ualsbombe:
Did indeed fix the bug on my end, thank you!

PR #7567, opened 2020-04-06
MRG, FIX: Fix toggle all

larsoner:
Fixes a bug where "Toggle all" did not actually toggle all projectors. Test fails on `master` and passes on this PR.


agramfort:
backported

PR #7569, opened 2020-04-07
Rename artifact detection to movement detection

AdoNunes:
A small PR.
In a previous PR, the filename for the movement artifact detection was named 'plot_artifact_detection.py' but I realize that it should be "plot_movement_detection.py". This is more in accordance with the new "plot_muscle_detection.py".


agramfort:
thx @AdoNunes 

backported

PR #7572, opened 2020-04-07
MRG: Remove toolbars in PyVista plotter

GuillaumeFavelier:
This small PR disables the default camera toolbars which are not compatible with `_TimeViewer`. Its features are either unused or duplicated.

This simplifies the window design overall.

`master` | PR
------------|-----
image | image

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -159,6 +159,8 @@ class _Renderer(_BaseRenderer):
             with _disabled_depth_peeling():
                 self.plotter = self.figure.build()
             self.plotter.hide_axes()
+            self.plotter.default_camera_tool_bar.close()
+            self.plotter.saved_cameras_tool_bar.close()
             _enable_aa(self.figure, self.plotter)
 
     def subplot(self, x, y):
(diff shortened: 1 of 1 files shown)


agramfort:
@hoechenberger ok for you?

hoechenberger:
Works for me. It's a little jumpy because the windows first opens _with_ the toolbars, only to close them during the initialization process. But if that's the easiest way for you to to it, let's just leave it so.

larsoner:
It would indeed be nice to make this a PyVista on-window-instantiation option (i.e., have it never show up at all). @GuillaumeFavelier want to make an upstream PR that avoids this jerkiness in the long term?

GuillaumeFavelier:
> want to make an upstream PR that avoids this jerkiness in the long term?

I'll take care of it!

GuillaumeFavelier:
I opened https://github.com/pyvista/pyvista/pull/667

But meanwhile I think closing the toolbar is the nicest way to achieve this.

larsoner:
Agreed, and once 0.25 is out we can add some `_get_args` introspection to figure out if we can remove during instantiation (0.25+) or if we need to remove after the fact (0.24)

larsoner:
Thanks @GuillaumeFavelier

PR #7579, opened 2020-04-08
DOC: Fixes in tutorial

hoechenberger:

Fixes a typo in the "EEG source localization given electrode locations on an MRI" tutorial and corrects the transformation matrix going from milli-meters to meters.

### DIFF
--- a/tutorials/source-modeling/plot_eeg_mri_coords.py
+++ b/tutorials/source-modeling/plot_eeg_mri_coords.py
@@ -65,7 +65,7 @@ fname_mon = op.join(misc_path, 'sample_eeg_mri', 'sample_mri_montage.elc')
 # which does a maximum intensity projection (easy to see the fake electrodes).
 # This plotting function requires data to be in MNI space.
 # Because ``img.affine`` gives the voxel-to-world (RAS) mapping, if we apply a
-# RAS-to-MRI transform to it, it becomes the voxel-to-MNI transformation we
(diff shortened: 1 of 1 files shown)


tutorials/source-modeling/plot_eeg_mri_coords.py
```diff
@@ -95,7 +95,7 @@
 #         >>> pos_vox = ...  # loaded from a file somehow
 #         >>> img = nibabel.load(fname_T1)
 #         >>> vox2mri = img.header.get_vox2ras_tkr()  # voxel -> mri
-#         >>> vox2mri[:3] /= 1000.  # mm -> m
+#         >>> vox2mri[:3, -1] /= 1000.  # translation column: mm -> m
```
hoechenberger, line 98: I'm actually not so sure this is really correct, because probably the scaling should not only be applied to the _translation_, but also to the … well, to the "scaling part" of the transformation matrix. That is, the diagonal of `vox2mri[:3, :3]` should be multiplied with `1/1000`… But I haven't really worked with affine transformations "hand-on" in may years so I'm a little rusty :)

larsoner: It's going to depend on whether or not the affine changes the data units I think. In this case it does (voxel to physical), so this now looks incorrect

larsoner:
Thanks @hoechenberger

PR #7580, opened 2020-04-08
Rework _Brain show

GuillaumeFavelier:
**TL:DR**: `_TimeViewer` is displayed only once, when the UI is ready.

The default plotter of the pyvista backend is interactive as soon as it opens that's why one can notice every piece of the interface being added when `time_viewer=True` resulting in the jerkiness mentioned by @hoechenberger in https://github.com/mne-tools/mne-python/pull/7572#issuecomment-610382365 and @jasmainak in https://github.com/mne-tools/mne-python/pull/7247#issuecomment-587833079.

This PR fixes this behaviour by showing the `_TimeViewer` only when when it's ready (once every GUI element is set up). To achieve that I have to make `_Brain` aware that `time_viewer` is set. So I introduced the `show` parameter.

This way, the behaviour of `_Brain` is more consistent with all the basic `_3d.py` primitives and the `show` parameter is now connected to the rest of the viz pipeline.

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1788,12 +1788,17 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
 
     if title is None:
         title = subject if len(hemis) > 1 else '%s - %s' % (subject, hemis[0])
+    kwargs = {
+        "subject_id": subject, "hemi": hemi, "surf": surface,
+        "title": title, "cortex": cortex, "size": size,
+        "background": background, "foreground": foreground,
+        "figure": figure, "subjects_dir": subjects_dir,
+        "views": views
+    }
+    if get_3d_backend() == "pyvista":
+        kwargs["show"] = not time_viewer
     with warnings.catch_warnings(record=True):  # traits warnings
-        brain = Brain(subject, hemi=hemi, surf=surface,
-                      title=title, cortex=cortex, size=size,
-                      background=background, foreground=foreground,
-                      figure=figure, subjects_dir=subjects_dir,
-                      views=views)
+        brain = Brain(**kwargs)
     center = 0. if diverging else None
     for hemi in hemis:
         hemi_idx = 0 if hemi == 'lh' else 1
@@ -1816,7 +1821,7 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
                 kwargs["min"] = scale_pts[0]
                 kwargs["mid"] = scale_pts[1]
                 kwargs["max"] = scale_pts[2]
-            else:
+            else:  # pyvista
                 kwargs["fmin"] = scale_pts[0]
                 kwargs["fmid"] = scale_pts[1]
                 kwargs["fmax"] = scale_pts[2]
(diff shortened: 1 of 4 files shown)


GuillaumeFavelier:
What do you think of this change @agramfort, @larsoner ? Does it fix the 'jerkiness' for you @hoechenberger ?

larsoner:
@GuillaumeFavelier it fixes the jerkiness of the controls. Is it possible to also fix the jerkiness of the bottom matplotlib plot when `show_traces=True`? In theory you could create that window and embed the matplotlib and (complete) pyvista views, then show it, right?

GuillaumeFavelier:
> could create that window and embed the matplotlib and (complete) pyvista views, then show it

should be possible indeed, I'll take a look.

hoechenberger:
Just tried it out with @GuillaumeFavelier together via Zoom on macOS. Works great, if one ignores the "traces" window issue @larsoner mentioned. If this could be fixed, I'd be happy. :)

GuillaumeFavelier:
I switched my desktop window manager from `i3` to `KDE` but I can't reproduce the jumpy behaviour with `show_traces=True`. For me, the window appears complete. I tried PyQt5 5.11.3 and 5.14.1.

So I'll try by showing the matplotlib canvas hopefully to update the layout calculation before the final `show()` request.

larsoner:
The latest commit appears to have fixed all jerkiness for me, thanks @GuillaumeFavelier !

GuillaumeFavelier:
> The latest commit appears to have fixed all jerkiness for me

Nice! And what about you @hoechenberger ?

hoechenberger:
There is only a _minor_ jerkiness left here, which occurs when `show_traces=True`: at a very late stage of window composition, the status bar pops in (or maybe it's the window being resized so the status bar becomes visible, cannot tell). Aside from that, it's great!

GuillaumeFavelier:
Oh the status bar! I should enable it once and for all xD

agramfort:
don't wait for me to merge.

>

hoechenberger:
Sorry it's the exact opposite: status bar pops in when `show_traces=False`.

In both cases, `show_traces=False` and `show_traces=True`, the window is created with a status bar eventually. It just seems that when `show_traces=False`, the bar is either created later or, as I said, becomes visible due to a resize.

GuillaumeFavelier:
What I mean is that the status bar is not supposed to pop in any way. This is not intended.

GuillaumeFavelier:
I don't see the point to enable it in this PR without anything to populate it right now. It was done in https://github.com/mne-tools/mne-python/pull/7257 to get info about the calculation but got reverted.

Is it okay for you if I work on the status bar in another PR and work towards fixing the jerkiness at that time?

hoechenberger:
> What I mean is that the status bar is not supposed to pop in any way. This is not intended.

Maybe it's a macOS & Qt thing then. BUUUT I have some ideas how to use the status bar ;))

hoechenberger:
@GuillaumeFavelier Only now I see your previous comment. Yes, that's totally alright, do what you need to do :)

larsoner:
Thanks @GuillaumeFavelier

PR #7581, opened 2020-04-08
MRG, FIX: Datetime call in gdf 2.x age calculation

SveaMeyer13:
#### Reference issue
WIth regards to  #7124 .

#### What does this implement/fix?
While importing a GDF 2.x file with

```python
from mne import io
io.read_raw_gdf(fname, preload=True)
```

we encountered the following error:
 
```python
Extracting EDF parameters from /home/svea/Desktop/mne_examples/sample_data/B0103T.gdf...
GDF file detected
Traceback (most recent call last):
  File "/home/svea/Desktop/mne_examples/online_linear_demo.py", line 17, in <module>
    raw = io.read_raw_gdf(raw_fname, preload=True)
  File "/home/svea/Desktop/mne_examples/venv/lib/python3.7/site-packages/mne/io/edf/edf.py", line 1327, in read_raw_gdf
    verbose=verbose)
  File "<decorator-gen-164>", line 21, in __init__
  File "/home/svea/Desktop/mne_examples/venv/lib/python3.7/site-packages/mne/io/edf/edf.py", line 189, in __init__
    exclude, preload)
  File "/home/svea/Desktop/mne_examples/venv/lib/python3.7/site-packages/mne/io/edf/edf.py", line 377, in _get_info
    edf_info, orig_units = _read_header(fname, exclude)
  File "/home/svea/Desktop/mne_examples/venv/lib/python3.7/site-packages/mne/io/edf/edf.py", line 366, in _read_header
    return _read_gdf_header(fname, exclude), None
  File "/home/svea/Desktop/mne_examples/venv/lib/python3.7/site-packages/mne/io/edf/edf.py", line 865, in _read_gdf_header
    today = datetime.today(tzinfo=timezone.utc)
TypeError: today() takes no keyword arguments
```

After closer investigation, we suspect that it was caused by the changes introduced by #7124 and released in v0.20.0.

As stated in the [Python docs for datetime](https://docs.python.org/3.7/library/datetime.html#datetime.datetime.now),
datetime.now(tz=timezone.utc) should be the syntactically correct functional equivalent.

To then make today and patient['birthday'] comparable, we add tzinfo to birthday (in line 860).

PR #7582, opened 2020-04-08
MRG, FIX: Fix today

larsoner:
Fixes an issue from the listserv.

@nbara have you checked to see if the birthday / age stuff actually gives correct values?

Needs #7581, then I'll rebase

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -42,6 +42,8 @@ Bug
 
 - Fix bug with :func:`mne.make_forward_dipole` where :func:`mne.write_forward_solution` could not be used by `Eric Larson`_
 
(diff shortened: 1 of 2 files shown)


larsoner:
Also I noticed that even though there is a birthday, `info['subject_info'] is None`, which probably shouldn't be the case.

@nbara any interest in trying to actually populate that dict? I don't have files to test with so I'm probably not the best one to implement it. You can see the entries that can go in here:

https://mne.tools/dev/generated/mne.Info.html

nbara:
Ooof, I haven't used GDF files in ages! I don't have other files to test I'm afraid...

larsoner:
Merged by rebase

Issue #7576, opened 2020-04-07
Cannot plot data after filtering


Andrea-V:
#### Describe the bug
I cannot plot the data again after applying filtering.

#### Steps to reproduce

		data_highpass = data.copy().filter(l_freq=0.5, h_freq=None)
		
		data.plot(n_channels=61, scalings="auto") # this works as expected
		plt.show()

		data_highpass.plot(n_channels=61, scalings="auto") # here I get the error
		plt.show()

#### Expected results
It should be able to plot data after filtering

#### Actual results
ValueError: cannot convert float NaN to integer

#### Additional information
Platform:      Linux-5.3.0-45-generic-x86_64-with-glibc2.10
Python:        3.8.1 (default, Jan  8 2020, 22:29:32)  [GCC 7.3.0]
Executable:    /home/andrea/anaconda3/envs/bci_competition/bin/python
CPU:           x86_64: 8 cores
Memory:        7.6 GB

mne:           0.19.2
numpy:         1.18.1 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.4.1
matplotlib:    3.2.0 {backend=TkAgg}

sklearn:       0.22.2.post1
numba:         Not found
nibabel:       Not found
cupy:          Not found
pandas:        1.0.1
dipy:          Not found
mayavi:        Not found
pyvista:       Not found
vtk:           Not found

agramfort:
it seems filtering creates NaN

larsoner:
@Andrea-V we can't tell from the last line of the traceback where the error is actually happening. Can you paste the full traceback?

Andrea-V:
here is the full error:

`
Traceback (most recent call last):
  File "preprocessing_upper_limb.py", line 54, in <module>
    data_highpass.plot(n_channels=61, scalings="auto")		
  File "/home/andrea/anaconda3/envs/bci_competition/lib/python3.8/site-packages/mne/io/base.py", line 1562, in plot
    return plot_raw(self, events, duration, start, n_channels, bgcolor,
  File "</home/andrea/anaconda3/envs/bci_competition/lib/python3.8/site-packages/mne/externals/decorator.py:decorator-gen-132>", line 2, in plot_raw
  File "/home/andrea/anaconda3/envs/bci_competition/lib/python3.8/site-packages/mne/utils/_logging.py", line 90, in wrapper
    return function(*args, **kwargs)
[...]

PR #7583, opened 2020-04-08
MRG, FIX: Fix labeling for broken raw

larsoner:
Why the data are `nan` is another issue, but at least `raw.plot` should not die with a cryptic message. We could perhaps also add a warning, this PR will make it so the user sees something like this, which pretty clearly indicates a problem (so maybe it's enough already):

Screenshot from 2020-04-08 10-29-21

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -40,6 +40,8 @@ Bug
 
 - Fix bug with :func:`mne.make_forward_dipole` where :func:`mne.write_forward_solution` could not be used by `Eric Larson`_
 
+- Fix bug with :meth:`mne.io.Raw.plot` when ``scalings='auto'`` where bad data would prevent channel plotting by `Eric Larson`_
+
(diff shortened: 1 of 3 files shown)


larsoner:
Pushed a commit to deal with `nan` in autoscale, now we get:

Screenshot from 2020-04-08 11-49-09

Eventually we might want to add a check like this:
```
    if check_finite and not np.isfinite(x).all():
        warn('Non-finite (inf or NaN) found in data, these will spread '
             'when %s. Consider picking different channels, or Annotating '
             'segments as bad if working with raw data' % (kind,))
```
But it's non-trivial to do this in a way that is actually informative for the user. For example in #7576 we hit this during EDF channel resampling, and the user gets spammed with messages. So it should probably be a separate PR someday.

larsoner:
@agramfort this would be good to get into 0.20.1, too

agramfort:
backported

PR #7584, opened 2020-04-08
MRG, DOC: Simplify Darwin installation

larsoner:
Apparently conda supports platform selectors (?) so let's use those instead.


hoechenberger:
This is only working behause this selector appears in the pip section and is passen straight to pip. Selectors won’t work for conda packages in environment files. Just FYI :)

PR #7592, opened 2020-04-09
WIP: Add basic tool bar with unicode symbols

GuillaumeFavelier:
This PR, following a discussion with @hoechenberger and strongly inspired by cbrnr/mnelab#61, adds a tool bar with useful bindings to the features of _TimeViewer:

output

It's also an item of #7162 since it adds buttons to the interface.

---

This is not compatible with https://github.com/mne-tools/mne-python/pull/7589

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -329,6 +329,8 @@ class _TimeViewer(object):
         self.color_cycle = None
         self.picked_points = {'lh': list(), 'rh': list()}
         self._mouse_no_mvt = -1
+        self.icons = dict()
+        self.actions = dict()
         self.orientation = [
             'lateral',
             'medial',
@@ -359,6 +361,7 @@ class _TimeViewer(object):
         self.plotter = brain._renderer.plotter
         self.main_menu = self.plotter.main_menu
         self.window = self.plotter.app_window
+        self.tool_bar = self.window.addToolBar("toolbar")
         self.status_bar = self.window.statusBar()
         self.interactor = self.plotter.interactor
         self.interactor.keyPressEvent = self.keyPressEvent
@@ -374,12 +377,14 @@ class _TimeViewer(object):
             self.show_traces = show_traces
             self.separate_canvas = False
 
+        self.load_icons()
         self.configure_time_label()
         self.configure_sliders()
         self.configure_scalar_bar()
         self.configure_playback()
         self.configure_point_picking()
         self.configure_menu()
+        self.configure_tool_bar()
 
         # show everything at the end
         self.toggle_interface()
@@ -393,6 +398,12 @@ class _TimeViewer(object):
 
     def toggle_interface(self):
         self.visibility = not self.visibility
+
+        # update tool bar icon
+        if self.visibility:
+            self.actions["visibility"].setText(self.icons["visibility_on"])
+        else:
+            self.actions["visibility"].setText(self.icons["visibility_off"])
 
         # manage sliders
         for slider in self.plotter.slider_widgets:
@@ -432,6 +443,13 @@ class _TimeViewer(object):
 
     def toggle_playback(self):
         self.playback = not self.playback
+
+        # update tool bar icon
+        if self.playback:
+            self.actions["play"].setText(self.icons["pause"])
+        else:
+            self.actions["play"].setText(self.icons["play"])
+
         if self.playback:
             time_data = self.brain._data['time']
             max_time = np.max(time_data)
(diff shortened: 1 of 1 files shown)


agramfort:
I prefer unicode so no need to add any file in the repo.

>

agramfort:
works like a charm for me!

hoechenberger:
(I honestly think this looks so much worse than the SVG icons you had in the other PR… Looks like a proof of concept in comparison, but not like a refined UI. just my 2 ct)

hoechenberger:
@agramfort @larsoner Why is it such a big deal to ship with a few tiny SVGs? Honest question :)

hoechenberger:
So, regarding which toolbar looks better, this one or the SVG one – maybe it is just a matter of taste.

But if the reasoning behind using an (imho) inferior UI design is technical limitations (e.g. keeping the sdist slightly smaller or sth), I really believe this decision should be reconsidered. TimeViewer is quickly gaining in features and is already one of the BIG selling points of our 0.20 release. It will only continue to get even more useful and at one point will be the "face" of MNE when it comes to source analysis. I think it's important that this face is looking as great as possible: a beautiful app creates a better UX and higher user affinity! UI candy is really important to many users (and me personally – hey, I'm using a Mac for a reason ;)) Just my… USD 2, this time ;)

agramfort:
looking at the size of this diff:

https://github.com/mne-tools/mne-python/pull/7592/files

it seems easier to maintain (1 file) and yes it keeps the sdist small.

>

hoechenberger:
It really looks like a KDE or Gnome app from 20 years ago on my computer. I don't like it.

Screenshot 2020-04-09 at 18 57 17

mne/viz/_brain/_timeviewer.py
```diff
@@ -755,6 +773,47 @@ def configure_point_picking(self):
...
+    def load_icons(self):
+        self.icons["help"] = "❓"
+        self.icons["play"] = "⏯️"
+        self.icons["pause"] = "⏸️"
+        self.icons["scale"] = "📏"
+        self.icons["clear"] = "🗑️"
+        self.icons["restore"] = "↩️"
+        self.icons["screenshot"] = "📷"
+        self.icons["visibility_on"] = "👀"
+        self.icons["visibility_off"] = "👀"
```
larsoner, line 785: Maybe try builtin Qt icons?

https://doc.qt.io/qt-5/qstyle.html#StandardPixmap-enum

drammock:
Maybe I missed it, but why is FontAwesome not an option?  If distribution size is the only concern, it is possible to create a derivative font with only the icons we need (e.g., online tool [fontello](http://fontello.com/) does this).

hoechenberger:
> but why is FontAwesome not an option?

I proposed FA but I think it hasn't been considered / discussed yet :) Thanks for bumping this, this could really be a middle ground? :)

larsoner:
How big was the size diff, looks like it has been force-pushed over?

+1 for the trim-font idea from @drammock. It will keep `pip install` small. As we add icons it will hurt our repo size diff, but seeing as the entire `fontawesome` in our `doc/_static/fonts` is 165 kB and we will use a tiny fraction, I think this will be negligible.

hoechenberger:
Just for a "newcomer", why is the repo & sdist size so crucial? Users download the sdist… how many times a year? And I clone the full repo only when I move to a new computer. Why is size so important in 2020, where we have TB hard drives and fast internet connections?

cbrnr:
I also think that file sizes in the kB range shouldn't matter in 2020. Moreover, SVG icons are much smaller than a font; MNELAB's icons are between 186 bytes and 2 kB. In addition, SVG icons are text files and thus better suited for version control than binary (font) files (although I doubt that we will ever change SVG icons).

drammock:
Screenshot_2020-04-09_10-46-47

Are people happy with these?  I included 2 options for "help", 2 options for "trash", and 3 options for "scale" (personally I like the sliders best for that one)

GuillaumeFavelier:
Your idea looks very promising @drammock and I really want to see how it looks but I do not know how to integrate it. Do you create a new font that should be loaded by PyQt?

drammock:
@GuillaumeFavelier you can take [this JSON config file](https://gist.github.com/drammock/407f9c43b5b240b597cc37476b20dbb5) and load it into the Fontello website, to pick up where I left off.  It will provide fonts in TTF, WOFF, WOFF2, EOT, and SVG format. I'm assuming PyQt can handle at least one of those.

agramfort:
I have the impression that rather than saving a day of work to different
people by trying to KISS
i now create more work. I will just say "be pragmatic" and keep in mind
that timing is not negligible

>

hoechenberger:
> i now create more work. I will just say "be pragmatic" and keep in mind that timing is not negligible

I see what you mean, and sorry if I've been utilizing everybody's time here a little too much too :\ But I think that UI design is _extemely_ crucial to make users happy. No-one enjoys using ugly or badly designed applications. I mean, there's a reason why big companies like Apple conduct extensive studies on their new products to finetune the UI and ensure good UX. Therefore I, personally, think that any time to discuss UI eye candy is time very well spent, if the goal is to increase the MNE adoption rate :)

cbrnr:
I 100% agree with @hoechenberger. In my limited Qt experience, (SVG) icon files are the standard way to create icons. I've never seen someone use characters from special fonts. I'd just use the SVGs from the previous issue, they look good and weigh only a few kB.

cbrnr:
Also, I'd try to make sure that the app looks as native as possible. This mean that the menu bar should not be attached to the window, but should go in the global menu bar at the top of the screen.

agramfort:
do what you think is best. I remove my vote from the polling booth

>

agramfort:
I would propose to merge this for now and improve later. Any objection?

hoechenberger:
Current status:
- Alex has "retracted" his vote
- Clemens is in favor of SVG icons, based on looks and his experience with Qt (and the use by other apps)
- Daniel suggested to look into FontAwesome symbols, however this would require further work from Guillaume's side
- I'm in strong favor of the SVG approach in #7589 too
- So far I'm not aware of any strong technical implications "forcing" us to save on a few KB required for bundling SVGs, but correct me if I'm mistaken here

I would, therefore, like to **suggest** to **close this one in favor of a re-opened (and merged :)) #7589** :) The work has already been done there, and we could move forward with a solution with which, to my understanding, we all could live.

cbrnr:
I agree with @hoechenberger. I also want to add that if we have to bundle a new font this will take much more space than all SVG icons we're ever going to need. Still in the kB range though so not even important.

agramfort:
I won't vote again but please don't spend days on this...

>

hoechenberger:
+100 on moving ahead! @GuillaumeFavelier the stage is yours. 👌

GuillaumeFavelier:
Let's continue the effort on https://github.com/mne-tools/mne-python/pull/7589 then

PR #7574, opened 2020-04-07
MRG: Brain save movie

GuillaumeFavelier:
This PR brings back the part of https://github.com/mne-tools/mne-python/pull/7257 about `save_movie` in `_Brain`.

It's an item of https://github.com/mne-tools/mne-python/issues/7162

PR #7596, opened 2020-04-10
TST: Check Circle

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/7574#issuecomment-612106164 and tries to reproduce the warnings.

### DIFF
--- a/tutorials/intro/plot_20_events_from_raw.py
+++ b/tutorials/intro/plot_20_events_from_raw.py
@@ -72,7 +72,7 @@ raw.crop(tmax=60).load_data()
 # What is a STIM channel?
 # ^^^^^^^^^^^^^^^^^^^^^^^
 #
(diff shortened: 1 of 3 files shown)


GuillaumeFavelier:
Okay so I'm able to reproduce on `master` but I swim in very dark waters here. From what I can decipher, `STIM channel` and `hpi` are not registered in a certain glossary.

GuillaumeFavelier:
I suppose those two terms should be added in `doc/glossary.rst`

GuillaumeFavelier:
Actually `stim channel` already exists so it could be an issue with case-sensitive stuff.

larsoner:
Yes we build with `--pre` releases of Sphinx nowadays to catch these bugs early

larsoner:
Thanks @GuillaumeFavelier

Issue #7577, opened 2020-04-07
RFC: Remove "spectrum_fit" from notch filter


larsoner:
I have never really gotten it to work in practice, and users on Gitter are unhappy that it takes ages to run. I wonder if we should deprecate and remove it, as I'm not sure if fixing it is even really feasible.

mmagnuski:
Maybe we should move it to sandbox? But I'm fine either way.

sappelhoff:
this is relevant to https://github.com/sappelhoff/pyprep/issues/18

@yjmantilla

yjmantilla:
What I can contribute to this is that in the tests mentioned in sappelhoff/pyprep#21 it seems that the agreement between the prep cleanline method and the mne spectrum notch is high but I haven't really tested it focusing on that part. Maybe doing some simulations would help us determining if it is worthwhile leaving it there.

larsoner:
@mmagnuski suggested that this deficiency might be due to not operating on windows of raw data, which is entirely possible. I've opened a PR to make a general overlap-add processing class in #7598 that would hopefully make it easy to do so.

christian-oreilly:
I just want to confirm, in case that can be useful, that, as it has been mentioned in #7598, at least part of the very long processing for this function comes from the fact that the algorithm is applied on the whole raw file (or at least, much too long part of it). I pinpointed for my case where it was hanging and it was at the line `dpss = _get_dpss()(N, half_nbw, Kmax)` in the function `dpss_windows(...)`. It resulted in a call equivalent to `scipy.signal.windows.dpss(100000, 3383.0, 6766)` which takes forever. It now works quite well and within an acceptable time with #7609, using filter_length='10s'.

PR #7598, opened 2020-04-10
ENH: Add overlap add classes

larsoner:
This PR just refactors existing code to:

1. Make use of a generic overlap-add processing class (OTP), and
2. Make `Interp2` more flexible

This will eventually allow combining the two methods and using them more easily. For examlpe, @mmagnuski suggested that the reason `spectrum_fit` does not work and takes forever (#7577) is perhaps because it does not work in overlapping windows, but rather the entire raw file. With the `_OLA` class it should be much easier to make it operate on overlapping windows.


larsoner:
Basically as long as the raw `simulation` tests pass (Interp2) and the OTP tests pass (_OLA), along with the builtin tests for Interp2 and _OLA, we should be able to trust the results here (I hope!).

agramfort:
it's just a refactoring using a class / generator?

larsoner:
Yes to make it easier to use it elsewhere

agramfort:
+1 for MRG if green then

>

Issue #7601, opened 2020-04-13
BUG: Extreme topomap valuse still too strong


larsoner:
@agramfort pointed out that the `examples/time_frequency/plot_source_power_spectrum_opm.py` example had edges that are too strong:

![master](https://mne.tools/dev/_images/sphx_glr_plot_source_power_spectrum_opm_001.png)

Doing `border='mean'` plus `extrapolate='local'` gives:

![Screenshot from 2020-04-13 09-44-01](https://user-images.githubusercontent.com/2365790/79125259-e1b61280-7d6b-11ea-87b8-c20e826581ab.png)

I wonder if:

1. The default extrapolation mode of `box` should actually add the local points, then also add the four points in the corners. It might be a good compromise.
2. `border='mean'` should be the default

If we did these two things, this is what the new default would look like (hacked together the box change for now and set `border='mean'`):

![Screenshot from 2020-04-13 10-03-41](https://user-images.githubusercontent.com/2365790/79126553-2b076180-7d6e-11ea-8cdb-64d21e9aeb84.png)

I know we've had some conversations about (2) at least, but I'd also like (1) because it might combine the local focality of `'local'` with the smoothness (bounded by matplotlib mask instead of interpolation grid extent) of `'box'`. Thoughts @mmagnuski ?

larsoner:
Also, using `np.median` rather than `np.mean` seems to work a tiny bit better to deal with the blob-like appearance at the edges of `local (or box+local) interp:

![Screenshot from 2020-04-13 10-13-30](https://user-images.githubusercontent.com/2365790/79127506-044a2a80-7d70-11ea-9594-b7579361765a.png)

I tried weighting the points by inverse distance and it didn't seem to help much. But we can always improve this aspect of appearance later, the real questions are the two posed above.

agramfort:
is it possible to extrapolate a bit less far from sensors with "
border='mean' plus extrapolate='local'"
maybe divide by 2 the extra area beyond the convex hull?
I would consider changing then our defaults here.

>

larsoner:
Halving the distance fixes the blob problem, and also makes sense in terms of the "represented area" each channel gets, here shown for the new `'box'` mode (which does local then adds box; #7602)

![Screenshot from 2020-04-13 10-25-11](https://user-images.githubusercontent.com/2365790/79128208-25f7e180-7d71-11ea-8f02-4e2d0d660161.png)

I prefer this over pure `'local'` mode, which has jagged edges:

[...]

PR #7193, opened 2020-01-11
MRG+1, ENH: add border argument for topomap extrapolation

mmagnuski:
#### Reference issue
Continuation of [topomap extrapolation pull request](GH-5754).

#### What does this implement/fix?
This adds a `border` argument to `mne.viz.plot_topomap` that allows to change the values set to extrapolation points. The default is zero, just as before, but `'mean'` allows to extrapolate to mean of channels that are neighbours to given extrapolation point. See the example below:
image

(I know, I will have to fix the channel positions after the recent changes to plot_topomap channel position handling)

`border='mean'` is especially useful when plotting log-power (and setting `vmin`, `vmax`):
image

This is a draft at this point, but feel free to comment.

#### TODOs:
- [x] add tests
- [x] `%()s` docs
- [x] expose `border` in other topomap functions?

PR #7602, opened 2020-04-13
WIP: Use mean for borders and local for box extrapolation

larsoner:
Follow-up for #7193. Implements ideas from #7601.


larsoner:
I'm just going to kill CircleCI on this one, I don't think we'll need to look at the outputs. At some point we might consider using the local+box interpolation mode, but I think for now we don't need to bother

Issue #7472, opened 2020-03-18
Create canonical template channel locations


cbrnr:
Currently there are some issues with plotting montages, especially with those consisting of template locations (as opposed to realistic locations), see e.g. #7141 and #7455.

At the moment, MNE ships 24 montage files, and it is not obvious which of these contain template locations and which contain realistic locations. Moreover, among these montage files there are several files containing template locations, which is really confusing.

Ideally, there should be only one template containing all possible locations. This is something we have already discussed in #5269. Unfortunately, the discussion went into a different direction and the issue was closed without solving the original problem.

Therefore, I'd like to bring up this issue again. I suggest that we get rid of the various template location files and unify them in one central place. We could use a simple text file where we could also support multiple labels per location to accommodate different naming schemes. We could compute all locations with the tools provided by @sappelhoff, and we could either include his calculation tools or just the resulting text file in MNE.

agramfort:
disclairmer neither me or @larsoner or @massich with who I did most of the
refactoring of DigMontage
have strong experience with EEG data requiring a template. The history is
that we took files from fieldtrip
and aimed at offering the same options with MNE. Ideally we should as well
or better than fieldtrip or eeglab
with little effort. Now I don't have any strong opinion of how to achieve
this. I just have the impression that
depending of eeg system it's easier to have a file per EEG cap. my 2c

>

cbrnr:
In my experience, topomaps are often used to visualize (channel-based) results. In such cases, it is more important that the maps look pretty (symmetrical), because exact locations are usually not available. This is what template locations are mainly used for.

Most caps use some variation of the extended 10-20, 10-10 or 10-05 systems, which are all based on the same underlying subdivision of certain distances on a sphere. However, sometimes they use different names for the same locations. Therefore, I think having only one template with a large number of channels (e.g. 345 as @sappelhoff proposed) would be sufficient to create all kinds of different montages (if we allow for duplicate labels).

Right now the whole montage system is rather messy. It is nice that there are some realistic montages, but there should be only one template montage as opposed to many (partially or completely) overlapping files.

sappelhoff:
> Right now the whole montage system is rather messy. It is nice that there are some realistic montages, but there should be only one template montage as opposed to many (partially or completely) overlapping files.

I agree.

Some manufacturers (e.g., Biosemi) however do use their own system (not a standard), and thus the template positions of the 10-05 system do not apply.

cbrnr:
[...]

PR #7603, opened 2020-04-13
MRG, BUG: Use auto extrap and mean borders

### LINKED ISSUE #7601: BUG: Extreme topomap valuse still too strong

larsoner:
Probably a better solution than #7602.

WIP because:

- [x] viz examples/tutorials need to be updated
- [ ] [examples](https://19301-1301584-gh.circle-artifacts.com/0/dev/index.html) need to be checked

Closes #7601
Closes #7602

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -611,10 +611,18 @@ class _GridData(object):
             v_extra = np.zeros(self.n_extra)
             indices, indptr = self.tri.vertex_neighbor_vertices
             rng = range(n_points, n_points + self.n_extra)
+            used = np.zeros(len(rng), bool)
             for idx, extra_idx in enumerate(rng):
                 ngb = indptr[indices[extra_idx]:indices[extra_idx + 1]]
                 ngb = ngb[ngb < n_points]
-                v_extra[idx] = v[ngb].mean()
+                if len(ngb) > 0:
+                    used[idx] = True
+                    v_extra[idx] = v[ngb].mean()
+            if not used.all() and used.any():
+                # Eventually we might want to use the value of the nearest
+                # point or something, but this case should hopefully be
+                # rare so for now just use the average value of all extras
+                v_extra[~used] = np.mean(v_extra[used])
         else:
             v_extra = np.full(self.n_extra, self.border, dtype=float)
 
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -265,7 +265,7 @@ def plot_projs_topomap(projs, info, cmap=None, sensors=True,
                        colorbar=False, res=64, size=1, show=True,
                        outlines='head', contours=6, image_interp='bilinear',
                        axes=None, vlim=(None, None),
-                       sphere=None, extrapolate='box', border=0):
+                       sphere=None, extrapolate='local', border=0):
     """Plot topographic maps of SSP projections.
 
     Parameters
@@ -460,6 +460,7 @@ def _get_extra_points(pos, extrapolate, sphere):
     x, y, _, head_radius = sphere
 
     # the old method of placement - large box
+    mask_pos = None
     if extrapolate == 'box':
         extremes = np.array([pos.min(axis=0), pos.max(axis=0)])
         diffs = extremes[1] - extremes[0]
@@ -469,7 +470,7 @@ def _get_extra_points(pos, extrapolate, sphere):
             *([[0] * (pos.shape[1] - 1) + [1]] * pos.shape[1]))))
         pidx = np.tile(np.arange(pos.shape[1])[np.newaxis], (len(eidx), 1))
         outer_pts = extremes[eidx, pidx]
-        return outer_pts, Delaunay(np.concatenate((pos, outer_pts)))
+        return outer_pts, mask_pos, Delaunay(np.concatenate((pos, outer_pts)))
 
     # check if positions are colinear:
     diffs = np.diff(pos, axis=0)
@@ -516,25 +517,35 @@ def _get_extra_points(pos, extrapolate, sphere):
                 new_pos = new_pos[good_extra]
 
             tri = Delaunay(np.concatenate([pos, new_pos], axis=0))
-            return new_pos, tri
+            return new_pos, new_pos, tri
 
         # get the convex hull of data points from triangulation
         hull_pos = pos[tri.convex_hull]
 
         # extend the convex hull limits outwards a bit
-        channels_center = pos.mean(axis=0, keepdims=True)
-        radial_dir = hull_pos - channels_center[np.newaxis, :]
+        channels_center = pos.mean(axis=0)
+        radial_dir = hull_pos - channels_center
         unit_radial_dir = radial_dir / np.linalg.norm(radial_dir, axis=-1,
                                                       keepdims=True)
         hull_extended = hull_pos + unit_radial_dir * distance
+        mask_pos = hull_pos + unit_radial_dir * distance * 0.5
         hull_diff = np.diff(hull_pos, axis=1)[:, 0]
         hull_distances = np.linalg.norm(hull_diff, axis=-1)
+        del channels_center
+
+        # Construct a mask
+        mask_pos = np.unique(mask_pos.reshape(-1, 2), axis=0)
+        mask_center = np.mean(mask_pos, axis=0)
+        mask_pos -= mask_center
+        mask_pos = mask_pos[
+            np.argsort(np.arctan2(mask_pos[:, 1], mask_pos[:, 0]))]
+        mask_pos += mask_center
 
         # add points along hull edges so that the distance between points
         # is around that of average distance between channels
         add_points = list()
         eps = np.finfo('float').eps
-        n_times_dist = np.round(hull_distances / distance).astype('int')
+        n_times_dist = np.round(0.25 * hull_distances / distance).astype('int')
         for n in range(2, n_times_dist.max() + 1):
             mask = n_times_dist == n
             mult = np.arange(1 / n, 1 - eps, 1 / n)[:, np.newaxis, np.newaxis]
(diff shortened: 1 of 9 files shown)


larsoner:
[Examples have rendered](https://19301-1301584-gh.circle-artifacts.com/0/dev/index.html) in case anyone wants to look

larsoner:
Okay the MEG looks reasonable, but @mmagnuski you are right that the local convex hull is not always good -- for EEG in `plot_evoked_topomap` it does not look good.

This makes me wonder if we want to add `extrapolate='auto'`, which means `'local'` for MEG and `'head'` for EEG. Or maybe we want to add some softening / expanding of the edges, but that's probably going to be tricky to get right.

drammock:
The renderings in [this example](https://19301-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_evoked_topomap.html) look good to me. The renderings in [this example](https://19301-1301584-gh.circle-artifacts.com/0/dev/auto_examples/io/plot_read_proj.html) also look good as far as the "local"/convex hull question goes, but it looks like the head outline is slightly off-center left-to-right?  I realize that's not what you're addressing in this PR, but it's rather puzzling since both examples use the sample data.

agramfort:
MEG topos look much nicer:

https://19301-1301584-gh.circle-artifacts.com/0/dev/auto_examples/io/plot_read_proj.html#sphx-glr-auto-examples-io-plot-read-proj-py

but our EEG friends may have some concerns.

extrapolate='auto' is maybe a good option.

@mmagnuski @cbrnr thoughts @sappelhoff thoughts?

agramfort:
it's still a bit unclear to me why the EEG sensors on sample are so
centered inside the head circle. There
are electrodes closer to the ears in 3d.

>

larsoner:
They are that way in `master`, it's just harder to see when the default interp goes to the circle

- [master](https://mne.tools/dev/auto_examples/visualization/plot_evoked_topomap.html#additional-plot-topomap-options)
- [PR](https://19309-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_evoked_topomap.html#additional-plot-topomap-options) (unchanged pos)

It's possible this will be fixed by #7455, but I'd rather deal with the problem separately since it's orthogonal and not added by this PR.

mmagnuski:
@larsoner
> This makes me wonder if we want to add extrapolate='auto', which means 'local' for MEG and 'head' for EEG. Or maybe we want to add some softening / expanding of the edges, but that's probably going to be tricky to get right.

I  don't think an auto of `extrapolate='head'` would be good - as this generates artifacts when some channels are outside of the head circle. As such cases are pretty common now - this might lead to more surprise than the proposed always 'local' scenario. (`extrapolate='circle'` might be a good alternative to what 'head' does currently, but that's another issue)
When it comes to softening etc. - that's why I was referring to using two mask at the same time: the circle and convex hull. The circle smoothes and the convex hull removes whatever is not masked by the circle. But it would likely require some tweaking, and I know you don't like this idea very much, so I'm not pressing. :)

There seems to be some pixelization not masked that is visible here (third topo from the left):
image
but it seems related to the low `res` in that code cell (`res=32`).

larsoner:
> When it comes to softening etc. - that's why I was referring to using two mask at the same time: the circle and convex hull. The circle smoothes and the convex hull removes whatever is not masked by the circle. But it would likely require some tweaking, and I know you don't like this idea very much, so I'm not pressing. :)

This will not work for the case linked above -- you have a (correct) hexagonal convex hull, so adding an additional circular mask on top of this will either be a no-op or it will remove some sensors. If we want to keep `local` the default but soften it, a better approach might be to "push" some points out a bit from the radius. Or figure out how to get Bezier points or something... but I'd rather avoid these complications if possible.

> There seems to be some pixelization not masked that is visible here (third topo from the left):

Ahh yes, we'll need to tweak how far out we put the extra points, how many of them we have, etc. I changed some values here that might have made the "jagginess" problem worse, but we'll have to evaluate that in the context of not introducing "blobs", too.

Ultimately if someone uses a low res, though, I don't mind them ending up with jagged lines...

> I don't think an auto of extrapolate='head' would be good - as this generates artifacts when some channels are outside of the head circle. 

Where do you see this? My understanding is that "head" really means "minimum of the head circle and the circle that encompasses all sensors", so it should always give at least a somewhat sensible one.

It seems like the simplest solution is to make the default `head` for EEG and `local` for MEG, and then fix the problems with `head` plotting for EEG (since we need to do those anyway.

mmagnuski:
> This will not work for the case linked above -- you have a (correct) hexagonal convex hull, so adding an additional circular mask on top of this will either be a no-op or it will remove some sensors. If we want to keep local the default but soften it, a better approach might be to "push" some points out a bit from the radius. Or figure out how to get Bezier points or something... but I'd rather avoid these complications if possible.

I didn't mean this in the context of the example I pasted. If you increase the distance between channels and extrapolation points (and thus the distance of convex hull mask) the circle mask trick would work. But I understand your point, and it is a small thing - its good as it is.

> Where do you see this? My understanding is that "head" really means "minimum of the head circle and the circle that encompasses all sensors", so it should always give at least a somewhat sensible one.

It is more visible with `border=0`, but you can see that the extrapolation points are at the head circle:
image
with `border='mean'` it's less visible but the extrapolation points are still at the head circle. I think they should be where the circle mask is (or slightly outside to be precise) - but some time ago we discussed this and you preferred it stayed at the head circle.

> It seems like the simplest solution is to make the default head for EEG and local for MEG, and then fix the problems with head plotting for EEG (since we need to do those anyway.

Sure, if we fix `extrapolate='head'` then it would be ok. :)

> Ultimately if someone uses a low res, though, I don't mind them ending up with jagged lines...

:+1:

cbrnr:
Plotting IC topomaps with my limited set of channels (I still think they should be closer to the head circle but this is the other issue we've not solved) doesn't look great with the new defaults:

Screen Shot 2020-04-14 at 12 03 50

I'd prefer if we'd extrapolate at least to the head circle.

hoechenberger:
@cbrnr 
> I'd prefer if we'd extrapolate at least to the head circle.

+1 on that one

agramfort:
what do we do here to move on?

larsoner:
My vote is to do `extrapolate='auto'` which means 'local' for MEG and 'head' for EEG. Also maybe try to reduce the "edge blobs" here if reasonable to do so. Then I think the remaining issues can be dealt with in separate PRs (probably mostly for EEG / head / skirt modes?).

agramfort:
that works for me !

>

larsoner:
> I think they should be where the circle mask is (or slightly outside to be precise) - but some time ago we discussed this and you preferred it stayed at the head circle.

> Sure, if we fix extrapolate='head' then it would be ok. :)

Okay done now along with `'auto'` mode. Running `[circle full]` we'll see how it looks!

larsoner:
I have not looked at all examples, but this looks okay at least. If others agree I can take a look at all examples:

https://19486-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_evoked_topomap.html#additional-plot-topomap-options

mmagnuski:
Looks ok. So, in a subsequent PR we can change the behavior of extrapolate='head' so that extrapolation points are not placed on the head outline if some channels extend beyond this outline, right?

larsoner:
It already should extend outside the head, which is what you wanted, right?

mmagnuski:
Ah, you changed it? Sorry, didn't notice.

mne/viz/topomap.py
```diff
@@ -546,18 +559,19 @@ def _get_extra_points(pos, extrapolate, sphere):
...
     else:
+        assert extrapolate == 'head'
         # return points on the head circle
-        angle = np.arcsin(distance / 2 / head_radius)
+        angle = np.arcsin(distance / 2 / np.mean(radii))
         points_l = np.arange(0, 2 * np.pi, angle)
-        use_radius = head_radius * 1.1
-        points_x = np.cos(points_l) * use_radius + x
-        points_y = np.sin(points_l) * use_radius + y
+        use_radii = radii * 1.1
```
mmagnuski, line 566: Ok, so `radii` are x and y clip radius now, not the head_radius, 👍

larsoner: Yes indeed!

cbrnr:
My ICA components example looks good!

sappelhoff:
Had a quick look at some examples, looks good to me as well :+1: 

Could you post your example once more with the new settings @cbrnr ?

cbrnr:
Here are my IC topomaps - except for the issue that these now look much more compressed than before (#7472 and #7141) I'm happy:

Screen Shot 2020-04-21 at 09 38 11

mmagnuski:
@larsoner
Could we just change the docs for `extrapolate='head'` so that it is clear that the extrapolation is to the head outline or outside?

hoechenberger:
Looks great, awesome work!

larsoner:
@mmagnuski changed in https://github.com/mne-tools/mne-python/pull/7603/commits/15e5baef6c447e2d1cb3864cae09178ffb12a3ca#diff-c0cc0663d36fcb3b69d6e33a22573db1R87-R90 to:

        Extrapolate out to the edges of the clipping circle. This will be on
        the head circle when the sensors are contained within the head circle,
        but it can extend beyond the head when sensors are plotted outside
        the head circle.

Okay?

mmagnuski:
Perfect!

larsoner:
Okay, hopefully it's easier to fix the remaining issues now!

sappelhoff:
Thanks @larsoner!

Issue #7599, opened 2020-04-11
BUG: Bugs with VTK 9


larsoner:
@GuillaumeFavelier you can now [pretty easily build and install VTK 9](https://gitlab.kitware.com/vtk/vtk/blob/master/Documentation/dev/build.md#python-wheels) (which is good because they are on RC3, and it should fix some bugs and bring 3.8 support):
```
git clone --recursive https://gitlab.kitware.com/vtk/vtk.git
mkdir vtk-build
cd vtk-build
cmake -GNinja -DVTK_BUILD_TESTING=OFF -DVTK_WHEEL_BUILD=ON -DVTK_PYTHON_VERSION=3 -DVTK_WRAP_PYTHON=ON ../vtk
ninja -j 4
python setup.py bdist_wheel
pip install dist/vtk-*.whl
```
However, there is at least one rendering bug (depth peeling again maybe?) where the surface overlay gets cut off at some distance based on rotation (note the rightward "v" cut of the overlay):

![Screenshot from 2020-04-11 09-36-36](https://user-images.githubusercontent.com/2365790/79045278-4b9db300-7bd8-11ea-9ab6-a33474a3e803.png)

We might need to figure out how to properly resolve the coincident topology here -- relying on VTK to do it automagically might not work anymore. Or maybe it's something else...?

larsoner:
FWIW, PyVista worked "out of the box" this way, while [Mayavi fails to compile](https://github.com/enthought/mayavi/issues/908) (which I'll look into) :)

larsoner:
Can someone else on macOS see if you can replicate https://gitlab.kitware.com/vtk/vtk/-/issues/17846 ?

GuillaumeFavelier:
Thanks for starting this @larsoner. I'll follow the procedure your explained to be better prepared for VTK 9.

Also relevant for https://github.com/pyvista/pyvista/issues/562

banesullivan:
Can you share that data or how to launch that scene? I'll try to look into this and see what's happening to the overlain mesh.

larsoner:
WARNING: Will download 1.5 GB of data to `~/mne_data` (unless you give it a different path to use)
```
import faulthandler
import os.path as op
import mne

faulthandler.enable()
data_path = mne.datasets.sample.data_path()
sample_dir = op.join(data_path, 'MEG', 'sample')
subjects_dir = op.join(data_path, 'subjects')
inv = mne.minimum_norm.read_inverse_operator(op.join(
[...]

PR #7607, opened 2020-04-15
WIP: Resolve coincident topology in _Brain

GuillaumeFavelier:
This PR fixes the part of https://github.com/mne-tools/mne-python/issues/7599 about overlapping surfaces.

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -474,6 +474,10 @@ class _Brain(object):
                 actor, mesh = mesh_data
                 # add metadata to the mesh for picking
                 mesh._hemi = hemi
+                mapper = actor.GetMapper()
+                mapper.SetResolveCoincidentTopologyToPolygonOffset()
+                mapper.SetRelativeCoincidentTopologyPolygonOffsetParameters(
+                    -1., -1.)
             else:
                 actor, mesh = mesh_data, None
             self._data[hemi]['actor'].append(actor)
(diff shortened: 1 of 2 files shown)


mne/viz/_brain/_brain.py
```diff
@@ -474,6 +474,10 @@ def add_data(self, array, fmin=None, fmid=None, fmax=None,
                 actor, mesh = mesh_data
                 # add metadata to the mesh for picking
                 mesh._hemi = hemi
+                mapper = actor.GetMapper()
+                mapper.SetResolveCoincidentTopologyToPolygonOffset()
+                mapper.SetRelativeCoincidentTopologyPolygonOffsetParameters(
+                    -1., -1.)
```
larsoner, line 480: What I would do here is make it implicitly that the order of addition sets the zorder (things plotted later end up on top). So `brain` should have something like a `self._polygon_offset_counter = 0` in init and anything you add afterward does `self._polygon_offset_counter += 1` and then `mapper.SetRelativeCoincidentTopologyPolygonOffsetParameters(-self._polygon_offset_counter, -self._polygon_offset_counter)` or so. WDYT?

GuillaumeFavelier: From my experience with polygon offset, especially if many surfaces overlap, some artifacts can appear. Also, I checked the doc of SetRelativeCoincidentTopologyPolygonOffsetParameter and it accepts double so maybe we have some room for tweaks here.

Ref: https://vtk.org/doc/nightly/html/classvtkMapper.html#ad49f0d5f3086b3a79f542cb9e801d8ac

GuillaumeFavelier: I'll go with `self._polygon_offset_counter` and modify accordingly if we run into artifacts :+1:

larsoner: It seems like as soon as we do a `brain.add_annotation` plus a `brain.add_data` we will have a problem, but I guess we can deal with it then

PR #7609, opened 2020-04-15
MRG, ENH: Use overlap-add in spectrum_fit mode

### LINKED ISSUE #7577: RFC: Remove "spectrum_fit" from notch filter

larsoner:
Closes #7577.

Based on some local testing I'm not convinced it works. But @yjmantilla @sappelhoff @mmagnuski if you want to try it just pass `filter_length='10s'` or whatever to `notch_filter`. Maybe this will make it easier for people to test against `cleanline` or other implementations, and/or push fixes.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -308,8 +308,12 @@ filter_length : str | int
 """
 docdict['filter_length_notch'] = docdict['filter_length'] + """
     When ``method=='spectrum_fit'``, this sets the effective window duration
-    over which fits are computed. The default in 0.21 is None, but this will
-    change to 10. in 0.22.
+    over which fits are computed. See :func:`mne.filter.create_filter`
+    for options. Longer window lengths will give more stable frequency
+    estimates, but require (potentially much) more processing and are not able
+    to adapt as well to non-stationarities.
+
+    The default in 0.21 is None, but this will change to ``'10s'`` in 0.22.
 """
 docdict['l_trans_bandwidth'] = """
 l_trans_bandwidth : float | str
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -306,6 +306,11 @@ filter_length : str | int
     * **int**: Specified length in samples. For fir_design="firwin",
       this should not be used.
 """
+docdict['filter_length_notch'] = docdict['filter_length'] + """
+    When ``method=='spectrum_fit'``, this sets the effective window duration
+    over which fits are computed. The default in 0.21 is None, but this will
+    change to 10. in 0.22.
+"""
 docdict['l_trans_bandwidth'] = """
 l_trans_bandwidth : float | str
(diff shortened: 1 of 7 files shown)


yjmantilla:
Thanks , Ive been really busy but I will test this as soon as I can.

christian-oreilly:
I did not do a full test and rigorous comparison with Matlab `cleanline`, but using this PR with `pyprep`, it worked quite well. Before these changes, `pyprep` was hanging forever because of the `spectrum_fit` notch filtering (#7577). With this modif (and the corresponding modif in `pyprep`) it ran in an acceptable time and the results look reasonable.

PSD for the original data:

image
       
I am generally not a big fan of notch filters because they clearly over-filter. Here is the result with a standard `firwin` MNE notch-filter. 

image

The spectral distortion is arguably even worst than on the original spectrum. Now, using the `spectrum_fit` through `pyprep` : 

image

There is still a bit of overfitting, but much better. In any case, the result seems to agree with what it aims to do.

larsoner:
Changing this to MRG then. I think any other fixes would be to the basic routines and can be in separate PRs. This already seems to help fix something that is broken/missing in `master`.

larsoner:
Rebased again. Should be ready for review/merge @agramfort

mne/utils/docs.py
```diff
@@ -306,6 +306,11 @@
     * **int**: Specified length in samples. For fir_design="firwin",
       this should not be used.
 """
+docdict['filter_length_notch'] = docdict['filter_length'] + """
+    When ``method=='spectrum_fit'``, this sets the effective window duration
+    over which fits are computed. The default in 0.21 is None, but this will
+    change to 10. in 0.22.
```
agramfort, line 312: Say it’s in seconds

agramfort: And explain what is the benefit or trade off here

larsoner:
@agramfort done:

https://21675-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_30_filtering_resampling.html#power-line-noise

Also fixed a bug with inplace division that was making it work less well for weird sample rates -- @christian-oreilly you might want to re-test it on your data. It should give better results now.

christian-oreilly:
> @agramfort done:
> 
> https://21675-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_30_filtering_resampling.html#power-line-noise
> 
> Also fixed a bug with inplace division that was making it work less well for weird sample rates -- @christian-oreilly you might want to re-test it on your data. It should give better results now.

Thanks @larsoner I'll rerun PyPrep on my dataset with the master once this branch is merged.

drammock:
thanks @larsoner

PR #7610, opened 2020-04-16
MRG: Refactor _TimeViewer shortcuts

GuillaumeFavelier:
Thanks to https://github.com/mne-tools/mne-python/pull/7589 we can now drastically simplify the way we manage the shortcuts in `_TimeViewer`. This PR also improves stability because we don't rely on the `keyPressEvent` callback anymore and flexibility because we use `QAction` instead of a `dict` for the bindings. It's also easier to maintain because the shortcuts are set in one place.

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -347,14 +347,6 @@ class _TimeViewer(object):
             'frontal',
             'parietal'
         ]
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
I have to update `_LinkViewer` as well.

hoechenberger:
Thanks @GuillaumeFavelier

PR #7606, opened 2020-04-14
MRG: _TimeViewer resource guide

GuillaumeFavelier:
This PR adds a small README that describes the instructions to update the `resource.py` file used for `_TimeViewer`.

PR #7612, opened 2020-04-16
MRG: TimeViewer save movie

GuillaumeFavelier:
This PR brings back the part of https://github.com/mne-tools/mne-python/pull/7257 about `save_movie` in `_TimeViewer`. The default shortcut is `CTRL+SHIFT+s` and a new icon in the tool bar has been added:

image


GuillaumeFavelier:
What do you think @agramfort, @hoechenberger , @larsoner?

hoechenberger:
Thanks @GuillaumeFavelier, will test this ASAP.

Wondering if an even better icon could be: 
video_call-24px

(from https://material.io/resources/icons/?search=video&icon=video_call&style=outline)

WDYT? Any preferences there?

GuillaumeFavelier:
No strong opinion here. Both does the job. I can change it.

larsoner:
Slight preference for something that implies save (usually meaning: have a floppy disk), like



https://findicons.com/icon/163988/film_save

GuillaumeFavelier:
> Slight preference for something that implies save 

Hm... This is a good point. It's about `save_movie` after all

larsoner:
Bonus points if you can find (or make) an image for "screenshot" with the same floppy disk image superimposed in the same location of the icon, so that we have some icon-consistency for what will open a save dialog

GuillaumeFavelier:
I tried to find something in the same style but I could not find anything so I made this one with Inkscape:

image

> the same floppy disk image

With this one:

image

I will see how it goes when integrated in the tool bar

GuillaumeFavelier:
> I tried to find something in the same style but I could not find anything so I made this one with Inkscape:

I should probably increase the resolution of the icons. On my display, it's super small. You cannot notice the disk:

image

GuillaumeFavelier:
@larsoner is it okay for you if I rework the icons (icon-consistency, resolution) in another PR just after https://github.com/mne-tools/mne-python/pull/7606? I can open an issue to keep track of this.

hoechenberger:
> floppy disk image

Do people these days even know WTH a floppy disk is? I'm being serious, if you turn 20 this year, chances are you never ever had to deal with floppies in your entire life. And I haven't seen such a "Save" icon in … ages, either? Only in old or open-source software ;) On the Mac, these days you always have some autosave or use the `File / Save (as)` menu entry.

This just to say that I'm not against a floppy disk symbol, I just doubt it carries any useful meaning these days unless you're, like, in your 40s or older.

hoechenberger:
Also wonder if it shouldn't be called "Capture Movie" or "Record Movie" or "Record Clip" instead of "Save Movie". Apple simply uses "Record" (without any further noun) in their Screencast app.

hoechenberger:
@larsoner Thanks for this extensive response! You convinced me, and I trust you there regarding language use. So – "Save Movie"? + a small floppy icon? :)

GuillaumeFavelier:
Useful material for the recording feature can be found in https://github.com/mne-tools/mne-python/pull/7257#issuecomment-578783579

larsoner:
> is it okay for you if I rework the icons (icon-consistency, resolution) in another PR just after #7606? I can open an issue to keep track of this.

Yeah you can just add it to the big tracking issue

> So – "Save Movie"? + a small floppy icon? :)

I trust @GuillaumeFavelier to come up with something amazing at some point :)

larsoner:
Thanks @GuillaumeFavelier !

hoechenberger:
@GuillaumeFavelier @larsoner Just a thought, if it's difficult to find a good solution with "icon indicating _movie_ **plus** small icon suggesting _save_": it just occurred to me that the "film save" icon @larsoner posted above:

film_save

actually would probably already carry the desired meaning _even without the floppy disk_. Because this … how do you even call this? "film reel snippet" suggests that something has or will be stored permanently.

Now I don't really like [this particular icon](https://material.io/resources/icons/?search=film&style=outline) they provide on `material.io`:
Screenshot 2020-04-17 at 10 22 01


BUT then I came across another idea, as [this icon](https://material.io/resources/icons/?search=film&icon=movie&style=outline) popped up:
Screenshot 2020-04-17 at 10 22 13


**A clapperboard**! Couldn't that be a solution?  Not sure how easy it is to _identify_ this icon as a clapperboard – but if we modified it such that the… movable thing that claps down is lifted / tilted upwards, it could work, no?

Unknown

GuillaumeFavelier:
I like the last clapperboard icon!

I'll open a PR soon to experiment with icons with a better resolution and better consistency. I will link your suggestion for movie in the PR :+1:

Issue #7611, opened 2020-04-16
TimeViewer Play/Pause button should toggle state when tmax is reached


hoechenberger:
#### Describe the bug
Currently the Play/Pause button doesn't switch from Pause to Play when the the last time point has been reached.

#### Steps to reproduce
Open TimeViewer with an STC with multiple time points, press the play button, wait until the last time point is reached.

#### Expected results
Play/Pause button state should toggle.

#### Actual results
Play/Pause button does not toggle:
https://youtu.be/xnFgfJ1-s-M

#### Additional information
Instead of toggling automatically once tmax is reached, we could also simply loop until the user explicitly requests to pause playback?

hoechenberger:
Also needs a callback from time slider drag events.

GuillaumeFavelier:
> Also needs a callback from time slider drag events.

I am not sure to understand what you mean by that. Drag event for the time slider and matplotlib time line are 'active user event'. As opposed to the 'play/pause' button which indicates that the time points are updated automatically.

hoechenberger:
What meant is that we need to decide what happens in the following scenario:

- user has pressed the Play/Pause button
- movie is still playing
- user grabs the PyVista time slider and moves it along the tube
- should playback be paused?
- or should playback immediately continue once the slider drag event has been finalized?

larsoner:
I would say it should restart as soon as you let go of the slider, starting at the new slider location. This allows viewing the same activity over and over again just by clicking a point on the slider during playback, which is nice.

GuillaumeFavelier:
>  it should restart as soon as you let go of the slider, starting at the new slider location. 

I believe this is the current behaviour right?

![output](https://user-images.githubusercontent.com/18143289/79464596-8cdbfb80-7ffa-11ea-89ef-b299303ed99a.gif)

larsoner:
It was last time I tested and might indeed still be true. I was just responding to @hoechenberger's question above:

> - should playback be paused?
> - or should playback immediately continue once the slider drag event has been finalized?

hoechenberger:
> It was last time I tested and might indeed still be true. I was just responding to @hoechenberger's question above:

Yes, it's still working that way! :)

PR #7613, opened 2020-04-16
WIP: Fix play/pause button in _TimeViewer

GuillaumeFavelier:
This PR fixes the play/pause button behaviour in `_TimeViewer`. It is now updated correctly when it reaches the last timestep.

Closes https://github.com/mne-tools/mne-python/issues/7611

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -448,8 +448,11 @@ class _TimeViewer(object):
         self.fmax_slider_rep.SetValue(self.brain._data['fmax'])
         self.plotter.update()
 
-    def toggle_playback(self):
-        self.playback = not self.playback
+    def toggle_playback(self, value=None):
+        if value is None:
+            self.playback = not self.playback
+        else:
+            self.playback = value
 
         # update tool bar icon
         if self.playback:
@@ -473,7 +476,7 @@ class _TimeViewer(object):
             try:
                 self._advance()
             except Exception:
-                self.playback = False
+                self.toggle_playback(value=False)
                 raise
 
     def _advance(self):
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
This should do the trick @hoechenberger. If there is a use case for the `loop`, it's totally possible to do it too. Although I would prefer to have a toggle for it instead of making it the default.

mne/viz/_brain/_timeviewer.py
```diff
@@ -473,7 +476,7 @@ def play(self):
             try:
                 self._advance()
             except Exception:
-                self.playback = False
```
hoechenberger, line 476: I wonder if we can be a little more specific regarding which exception to catch?

hoechenberger:
> . If there is a use case for the `loop`, it's totally possible to do it too. Although I would prefer to have a toggle for it instead of making it the default.

I agree, can be solved in another PR if anyone desired this behavior.

mne/viz/_brain/_timeviewer.py
GuillaumeFavelier: Actually the goal here is to stay as general as possible to behave like a fail-safe, catch any unexpected Qt failures without crashing the Python interpreter. You can take a look at `safe_event()`:

https://github.com/mne-tools/mne-python/blob/2fc2e704976e7319ea200c1a979ad9392fd297c5/mne/viz/_brain/_timeviewer.py#L25-L31

larsoner:
LGTM, @GuillaumeFavelier feel free to merge if this one is also good to go

mne/viz/_brain/_timeviewer.py
hoechenberger: Haha alright then :)

PR #7614, opened 2020-04-16
MRG, ENH: Allow pick_ori="vector" with fixed ori inv

larsoner:
One advantage of `plot_vector_source_estimates` is that the arrows + glass brain represent the true geometry of the source space better than an inflated view. It would be nice to be able to use this function, even if you are working with fixed orientations. This PR enables `pick_ori='vector'` with fixed-orientation inverses.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -21,6 +21,8 @@ Changelog
 
 - Allow resampling raw data with :func:`mne.io.Raw.resample` without preloading data, by `Eric Larson`_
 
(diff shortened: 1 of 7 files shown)


larsoner:
Failures are unrelated (Azure timeout and OSX download failure)

agramfort:
can you add this to documentation? update an example and share a link
in artifact? thanks

larsoner:
https://19360-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_vector_mne_solution.html#sphx-glr-auto-examples-inverse-plot-vector-mne-solution-py

PR #7619, opened 2020-04-16
[WIP] Adding updated montage to allow np.nan based on keyword 'on_missing'

adam2392:
#### Reference issue
Fixes: https://github.com/mne-tools/mne-bids/issues/381

(Related issue in mne-bids)

Allows raw to set a montage even if there are missing coordinates. This is accomplished w/ an explicit `on_missing=('ignore', 'raise', 'warn')` term.


agramfort:
cc @cbrnr 

I think we should not make this the default and for your usecase my fix does the job

cbrnr:
I don't know, I still don't understand why a warning isn't sufficient. If there are functions that don't work with missing locations then the error should be raised there.

agramfort:
my experience is that people don't read warnings and it's source of silent
bugs

>

cbrnr:
This might be true, but the error should still be raised in a downstream function where it actually causes an error.

hoechenberger:
Why not add a flag to this specific function to determine whether it should throw a warning or an exception?

cbrnr:
Because I think it makes more sense to have a parameter that is more specific than the decision between error or warning. As a user, I don't care about raising an error or a warning, but if the parameter influences the subset directly I would immediately know what I'm actually changing (and this gives me the option to resolve the issue of unknown electrode locations, at least for specific use cases).

agramfort:
then we should add an missing='error' | 'warn' to set_montage to be more flexible for users we want to live on the edge :)

cbrnr:
I'd prefer `missing=True` (no error) and `missing=False` (error) - or something along these lines.

larsoner:
More consistent with what we do elsewhere would be `on_missing='raise'` by default, allowing `'ignore'`

drammock:
+1 for `on_missing='raise'`...  to me it's more explicit / obvious what it means, compared to `missing=True`

cbrnr:
Fine with me.

hoechenberger:
> +1 for `on_missing='raise'`... to me it's more explicit / obvious what it means, compared to `missing=True`

+1 on this one
~This is what I meant in the first place, sorry for not being clear there, @cbrnr

adam2392:
Sweet! I like to live on the edge :), and can go ahead and add that option to `set_montage`.

mne/channels/montage.py
```diff
@@ -679,20 +680,25 @@ def _backcompat_value(pos, ref_pos):
...
+                # set ch coordinates and names from digmontage or nan coords
+                _ch_pos_use = dict()
+                for name, use in zip(info_names, info_names_use):
+                    if use in ch_pos_use:
+                        _ch_pos_use[use] = ch_pos_use[use]
+                    else:
+                        _ch_pos_use[use] = [np.nan, np.nan, np.nan]
+                ch_pos_use = _ch_pos_use
+
+                warn(missing_coord_msg)
```
cbrnr, line 701: This warning should be removed because if setting `on_missing='ignore'` users are already aware and explicitly opt in to this behavior.

doc/changes/latest.inc
```diff
@@ -31,6 +31,8 @@ Changelog
 
 - Support for saving movies of source time courses (STCs) with ``brain.save_movie`` method and from graphical user interface by `Guillaume Favelier`_
 
+- Add "on_missing='raise'" to :func:`set_montage` to allow ignoring of missing electrode coordinates by `Adam Li`_
```
larsoner, line 34:
```suggestion
- Add "on_missing='raise'" to :meth:`mne.io.Raw.set_montage` and related functions to allow ignoring of missing electrode coordinates by `Adam Li`_
```

mne/channels/montage.py
```diff
@@ -606,7 +606,7 @@ def _get_montage_in_head(montage):
 
 
 @fill_doc
-def _set_montage(info, montage, match_case=True):
+def _set_montage(info, montage, match_case=True, on_missing=True):
```
larsoner, line 609: This is not a valid value. Somewhere below you should do:
```
_check_option('on_missing', on_missing, ('raise', 'ignore'))
```

mne/channels/montage.py
larsoner: Agreed, probably most consistent to make it just a `logger.info`

mne/io/meas_info.py
```diff
@@ -150,13 +150,18 @@ class MontageMixin(object):
...
         """Set EEG sensor configuration and head digitization.
 
         Parameters
         ----------
         %(montage)s
         %(match_case)s
+        on_missing : str
+            Either 'raise' to raise an error when
+            channels have missing coordinates,
+            or 'ignore' to set channels to np.nan and set montage.
```
larsoner, line 164: This can be moved to `mne/utils/docs.py` and here you can just do `(on_missing_montage)s`, but it would be good to add:

```suggestion
            or 'ignore' to set channels to np.nan and set montage.

            .. versionadded:: 0.20.1
```

because we will want to backport this

mne/channels/montage.py
```diff
@@ -621,6 +621,7 @@ def _set_montage(info, montage, match_case=True):
         The measurement info to update.
     %(montage)s
     %(match_case)s
+    %(on_missing)s
```
larsoner, line 624: There is already an `on_missing` in `mne/utils/docs.py` and it is not the correct one for this use case. You should create a new `on_missing_montage` or so in `mne/utils/docs.py` and use it here and in `meas_info.py`

mne/channels/montage.py
agramfort: this is not fixed

mne/channels/montage.py
```diff
@@ -634,6 +635,8 @@ def _set_montage(info, montage, match_case=True):
         montage = make_standard_montage(montage)
 
     if isinstance(montage, DigMontage):
+        _check_option('on_missing', on_missing, ('raise', 'ignore'))
```
agramfort, line 638: on_missing='warning' should be supported to be consistent with other occurence of this parameter in the rest of the  code

agramfort:
please add tests

adam2392:
> please add tests

Is the best place to add tests for this functionality to be in: mne/channels/tests/test_montage.py?

agramfort:
yes. theory is that if the function is in ./foo.py then tests are in ./tests/test_foo.py

mne/channels/montage.py
```diff
@@ -690,18 +692,20 @@ def _backcompat_value(pos, ref_pos):
                                    not_in_montage)
             if on_missing == 'raise':
                 raise ValueError(missing_coord_msg)
+            elif on_missing == 'warn':
+                logger.warning(missing_coord_msg)
```
agramfort, line 696: you should use 

from mne.utils import warn

function

mne/channels/montage.py
```diff
@@ -690,18 +692,20 @@ def _backcompat_value(pos, ref_pos):
...
                 logger.info(missing_coord_msg)
 
+            # set ch coordinates and names from digmontage or nan coords
+            _ch_pos_use = dict()
+            for name, use in zip(info_names, info_names_use):
+                if use in ch_pos_use:
+                    _ch_pos_use[use] = ch_pos_use[use]
+                else:
+                    _ch_pos_use[use] = [np.nan, np.nan, np.nan]
+            ch_pos_use = _ch_pos_use
```
agramfort, line 707: for me the logic here is broken.

```
            ch_pos_use = dict(ch_pos_use)  # make a copy
            for name in info_names:
                if name not in ch_pos_use:
                    _ch_pos_use[name] = [np.nan, np.nan, np.nan]
```

or something like this

mne/channels/montage.py
agramfort, line 609: @adam2392 why do you keep resolving this conversation? we both agree with @larsoner that on_missing takes string values and has 3 options. It should not be a book anywhere

adam2392: ah sorry, I added the option:

```
        _check_option('on_missing', on_missing, ['raise',
                                                 'ignore',
                                                 'warn'])
```

and forgot to change the function header.

agramfort:
@cbrnr ok for you.

Let's make CIs green first though

mne/utils/docs.py
```diff
@@ -794,6 +794,14 @@
 
     .. versionadded:: 0.20
 """
+docdict['on_missing_montage'] = """
+on_missing : str
+    Either 'raise', or 'warn' to raise an error/warning when
+    channels have missing coordinates,
+    or 'ignore' to set channels to np.nan and set montage.
+
```
larsoner, line 802:
```
Running flake8
mne/utils/docs.py:802:1: W293 blank line contains whitespace
```

agramfort:
I would not backport this as in 0.20 you had the option to allow subsets. I had a bunch of conflicts when trying. Unless someone objects...

hoechenberger:
+1 on not backporting if that causes these kinds of issues. What is your stance on this, @cbrnr?

agramfort:
Then I think we need a dedicated PR on maint/0.20 branch to avoid breaking
things

>

cbrnr:
@agramfort you mean if we backport we'd have to replace `raise_on_subset=False` with `on_missing='ignore'`? This should be doable, but I'm fine with not backporting because 0.20 has an option to disable the error.

agramfort:
ok let's **not** backport then

>

PR #7620, opened 2020-04-17
MRG: Plot NIRS data between source and detector

rob-luke:
#### Reference issue
Discussed in #7057 and https://github.com/mne-tools/mne-nirs/issues/59

Add option to plot line between source detector pairs in `plot_alignment` .

#### Additional details

This PR originally contained code to plot data over the head, but this has been moved to MNE-NIRS

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -685,9 +685,8 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     if not all(x in ('original', 'projected') for x in eeg):
         raise ValueError('eeg must only contain "original" and '
                          '"projected", got %s' % (eeg,))
-    if not all(x in ('channels', 'pairs') for x in fnirs):
-        raise ValueError('fnirs must only contain "channels" or "pairs", '
-                         'got %s' % (fnirs,))
+    for xi, x in enumerate(fnirs):
+        _check_option('fnirs[%d]' % xi, x, ('channels', 'pairs'))
 
     info = create_info(1, 1000., 'misc') if info is None else info
     _validate_type(info, "info")
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -594,8 +594,10 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         the subjects bem and bem/flash folders are searched. Defaults to None.
     seeg : bool
         If True (default), show sEEG electrodes.
-    fnirs : bool
-        If True (default), show fNIRS electrodes.
+    fnirs : str | list | bool | None
+        Can be "channels" or "pairs" to show the fNIRS channel locations or
+        line between source-detector pairs, or a combination like
+        ``('pairs', 'channels')``. True translates to ``('channels')``.
 
         .. versionadded:: 0.20
     show_axes : bool
@@ -664,8 +666,15 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     if isinstance(eeg, str):
         eeg = [eeg]
 
+    if fnirs is True:
+        fnirs = ['channels']
+    elif fnirs is False:
+        fnirs = list()
+    elif isinstance(fnirs, str):
+        fnirs = [fnirs]
+
     _check_option('interaction', interaction, ['trackball', 'terrain'])
-    for kind, var in zip(('eeg', 'meg'), (eeg, meg)):
+    for kind, var in zip(('eeg', 'meg', 'fnirs'), (eeg, meg, fnirs)):
         if not isinstance(var, (list, tuple)) or \
                 not all(isinstance(x, str) for x in var):
             raise TypeError('%s must be list or tuple of str, got %s'
@@ -676,6 +685,9 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     if not all(x in ('original', 'projected') for x in eeg):
         raise ValueError('eeg must only contain "original" and '
                          '"projected", got %s' % (eeg,))
+    if not all(x in ('channels', 'pairs') for x in fnirs):
+        raise ValueError('fnirs must only contain "channels" or "pairs", '
+                         'got %s' % (fnirs,))
 
     info = create_info(1, 1000., 'misc') if info is None else info
     _validate_type(info, "info")
@@ -719,8 +731,10 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     ref_meg = 'ref' in meg
     meg_picks = pick_types(info, meg=True, ref_meg=ref_meg)
     eeg_picks = pick_types(info, meg=False, eeg=True, ref_meg=False)
-    other_bools = dict(ecog=ecog, seeg=seeg, fnirs=fnirs)
-    del ecog, seeg, fnirs
+    fnirs_picks = pick_types(info, meg=False, eeg=False,
+                             ref_meg=False, fnirs=True)
+    other_bools = dict(ecog=ecog, seeg=seeg, fnirs=('channels' in fnirs))
+    del ecog, seeg
     other_keys = sorted(other_bools.keys())
     other_picks = {key: pick_types(info, meg=False, ref_meg=False,
                                    **{key: True}) for key in other_keys}
@@ -1134,6 +1148,12 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
                               fwd_nn[:, ori, 1],
                               fwd_nn[:, ori, 2],
                               color=color, mode='arrow', scale=1.5e-3)
+    if 'pairs' in fnirs:
+        fnirs_loc = np.array([info['chs'][k]['loc'][3:9] for k in fnirs_picks])
+        logger.info('Plotting %d fnirs pairs' % (fnirs_loc.shape[0]))
+        renderer.tube(origin=fnirs_loc[:, :3],
+                      destination=fnirs_loc[:, 3:])
+
     renderer.set_camera(azimuth=90, elevation=90,
                         distance=0.6, focalpoint=(0., 0., 0.))
     renderer.show()
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -597,7 +597,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     fnirs : str | list | bool | None
         Can be "channels" or "pairs" to show the fNIRS channel locations or
         line between source-detector pairs, or a combination like
-        ``('pairs', 'channels')``. True translates to ``('channels',)``.
+        ``('pairs', 'channels')``. True translates to ``('pairs',)``.
 
         .. versionadded:: 0.20
     show_axes : bool
@@ -667,7 +667,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         eeg = [eeg]
 
     if fnirs is True:
-        fnirs = ['channels']
+        fnirs = ['pairs']
     elif fnirs is False:
         fnirs = list()
     elif isinstance(fnirs, str):
@@ -1147,7 +1147,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
                               fwd_nn[:, ori, 1],
                               fwd_nn[:, ori, 2],
                               color=color, mode='arrow', scale=1.5e-3)
-    if 'pairs' in fnirs:
+    if 'pairs' in fnirs and len(fnirs_picks) > 0:
         fnirs_loc = np.array([info['chs'][k]['loc'][3:9] for k in fnirs_picks])
         logger.info('Plotting %d fnirs pairs' % (fnirs_loc.shape[0]))
         renderer.tube(origin=fnirs_loc[:, :3],
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -597,7 +597,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     fnirs : str | list | bool | None
         Can be "channels" or "pairs" to show the fNIRS channel locations or
         line between source-detector pairs, or a combination like
-        ``('pairs', 'channels')``. True translates to ``('channels')``.
+        ``('pairs', 'channels')``. True translates to ``('channels',)``.
 
         .. versionadded:: 0.20
     show_axes : bool
--- a/mne/viz/backends/_pysurfer_mayavi.py
+++ b/mne/viz/backends/_pysurfer_mayavi.py
@@ -179,22 +179,31 @@ class _Renderer(_BaseRenderer):
         origin = np.atleast_2d(origin)
         destination = np.atleast_2d(destination)
         if scalars is None:
-            surface = self.mlab.plot3d([origin[:, 0], destination[:, 0]],
-                                       [origin[:, 1], destination[:, 1]],
-                                       [origin[:, 2], destination[:, 2]],
-                                       tube_radius=radius,
-                                       color=color,
-                                       figure=self.fig)
+            for idx in range(origin.shape[0]):
+                surface = self.mlab.plot3d([origin[idx, 0],
+                                            destination[idx, 0]],
+                                           [origin[idx, 1],
+                                            destination[idx, 1]],
+                                           [origin[idx, 2],
+                                            destination[idx, 2]],
+                                           tube_radius=radius,
+                                           color=color,
+                                           figure=self.fig)
         else:
-            surface = self.mlab.plot3d([origin[:, 0], destination[:, 0]],
-                                       [origin[:, 1], destination[:, 1]],
-                                       [origin[:, 2], destination[:, 2]],
-                                       [scalars[:, 0], scalars[:, 1]],
-                                       tube_radius=radius,
-                                       vmin=vmin,
-                                       vmax=vmax,
-                                       colormap=colormap,
-                                       figure=self.fig)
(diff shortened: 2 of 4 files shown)


rob-luke:
I don't know why this is throwing an error. It runs on my computer.

rob-luke:
Here is an example of what the code currently looks like on my computer (with different sensor position and random data). I am happy with the overall look. Now I need to make the code nice.

image

agramfort:
@GuillaumeFavelier can you review this one ? thx

rob-luke:
Thanks. Any review is appreciated. Its early days but any push in the right direction is appreciated.

agramfort:
At first glance, we need to decide if this code will only work with
pyvista or if we use the _Renderer
abstraction to be able to work with mayavi and future notebook backend. I
would prefer you use the _Renderer

>

rob-luke:
great. I will look in to _Renderer then.

rob-luke:
Ok using _Renderer it now looks like (again a different dataset)...

image

larsoner:
Nice!

Might be worth setting `sphinx_gallery_thumbnail_number = 16` in the example so that the brain is the thumbnail?

rob-luke:
People do love a brain picture :)

GuillaumeFavelier:
Very cool plot and usage of `_Renderer` here ^^

larsoner:
I had another API thought. I'm not sure we even need a separate function. For MEG we have `meg=('sensors', 'helmet')`. We should probably just have (by default) `fnirs=('sensors', 'pairs')`, where `'pairs'` means "draw a line between them". Then we don't even need a new keyword argument if we can live with users not getting to choose the colormap.

larsoner:
(I mean this should all be done in `plot_alignment`)

mne/viz/nirs.py
```diff
@@ -0,0 +1,172 @@
...
+        Can also be a list of these options, or an empty list (``[]``,
+        equivalent of ``False``).
+    fwd : instance of Forward
+        The forward solution. If present, the orientations of the dipoles
+        present in the forward solution are displayed.
+    dig : bool | 'fiducials'
+        If True, plot the digitization points; 'fiducials' to plot fiducial
+        points only.
+    ecog : bool
+        If True (default), show ECoG sensors.
```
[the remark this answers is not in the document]
larsoner: Ignore this if you agree with my API proposal below about just using `plot_alignment` directly

rob-luke:
Thanks for the feedback.

> I mean this should all be done in plot_alignment

How does one set the values/colours of sensors in plot_alignment? In this PR I’m passing in a data array, I can’t see an equivalent in plot alignment?

I wanted the ability to set the line colour to represent the output of analysis. Either hrf or peak values etc

> if we can live with users not getting to choose the colormap.

I’m a bit of a visualisation nerd. And would prefer being able to set colormaps. If we move to plot alignment can I add a keyword?

larsoner:
> How does one set the values/colours of sensors in plot_alignment? In this PR I’m passing in a data array, I can’t see an equivalent in plot alignment? ... I wanted the ability to set the line colour to represent the output of analysis. Either hrf or peak values etc

You can't currently. Indeed if you want to be able to "plot values" or something in 3D then a separate function would be better.

To me there are two separate use cases:

1. Showing the sensor arrangement relative to the head (i.e., geometry / configuration)
2. Showing some data values (i.e., some activation)

To me the sensor pairings with tubes between them falls at least partly under (1), and thus would be well suited to being included in `plot_alignment` at least to start.

Showing data doesn't really fit in the scope of `plot_alignment`, so a separate function would be better. Currently for (2) you have topo, topomap, and `plot_evoked_field` (though it's not used that much). It sounds like you're proposing a third function that is akin to `plot_alignment` in that it plots a brain/head/whatever plus sensor locations, but the sensor colors reflect some data values. It sounds like the first image in this ECoG example (though it's constructed by hacking the 3D transforms):

https://mne.tools/dev/auto_examples/visualization/plot_3d_to_2d.html#project-3d-electrodes-to-a-2d-snapshot

I could imagine the utility of making this sort of brain + electrode visualization work natively, rather than via some projection scheme. Clearly there is/was already a use case and need for it in ECoG, and now you're proposing something that seems quite similar in principle.

So my vote would be to first do this pairing scheme in `plot_alignment`. Then you don't need a colormap, since it's not showing data, right?

Then once we have that, we could build a generic sensor+data+brain plotting function. I don't think it matters much whether it's fNIRS or EEG/ECoG/MEG data (except that MEG data must have the coordinate frame changed). WDYT?

rob-luke:
Sounds like a plan. My need is for use case 2. But lets knock over 1 and 2 in the same go.

So I will expand `plot_alignment` to allow for `fnirs=('sensors', 'pairs')`. Which will show the location of source-detector pairs, but not overlay any data. All pairs will have the same color.

I will create a function (generalise the one already in this PR) for generic sensor+data+brain plotting.

rob-luke:
Ok `plot_alignment` with `pairs` seems to be working now and can be seen in the [ci circle build doc](https://19451-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#view-location-of-sensors-over-brain-surface). And looks as expected (see below). Now I will move on to the generic data display function ([use case 2](https://github.com/mne-tools/mne-python/pull/7620#issuecomment-615533313)).

mne/viz/nirs.py
```diff
@@ -0,0 +1,162 @@
+"""Functions to visualise NIRS data."""
+
+import numpy as np
+
+from . import plot_alignment
+
+
+def plot_nirs_source_detector(data, info=None, radius=0.001,
```
agramfort, line 8: would you agree to kill this function and just add nirs support to plot_alignment?

rob-luke: Definitely agree to killing this function.

But I would like to add a function to overlay channel analysis over a brain.

Would you prefer I make this PR for the extension to plot_alignment and move discussion of a data overlay to another issue/PR? Or move to MNE-NIRS?

agramfort:
what do you mean by "channel analysis"?

>

rob-luke:
Sorry, I should be clearer with my messages.

In NIRS a channel is between a light source and detector. These have been marked as lines in the figures above. For each channel we analyse the measurement and extract a metric (peak value, hrf analysis, etc) and usually call this channel level analysis. I guess its equivalent to scalp level analysis in EEG.

Plotting the output of the analysis for a source-detector pair is traditionally done by drawing a line between the source and detector and setting the line equal to the extracted metric. This is what I would like to create / that's what the `plot_nirs_source_detector` does.

agramfort:
so the color of the tube or its size would match a scalar quantity?

>

rob-luke:
> so the color of the tube or its size would match a scalar quantity?

Various tools do it differently. My preferred approach is colour represents a scalar quantity. Size remains constant. And you have dashed or solid lines to represent a Boolean (usually to represent if a significant response was observed).

I think it will be difficult to make different line types with Renderer. So maybe tube size could represent significance. Small tube size for no significant response. Large tube size for significant response.

One of the reasons I like the approach of showing the source detectors as a line/tube is that you can see how far apart the optodes are. Optode separation distance is proportional to how deep in the brain the photons will reach, and what brain regions the pair is sensitive too (larger separation=deeper). So you gain additional insight to what’s available in the topomap.

Also happy to move this to MNE-NIRS if that’s more appropriate. Or happy to discuss alternative visualisation approaches. It’s always good to hear another opinion from someone outside the nirs field.

agramfort:
I would start by putting this in mne-nirs for now. It will allow you to
iterate faster and release at a much faster rate

>

rob-luke:
The code is passing on PyVista but failing on Mayavi ([CI fail](https://travis-ci.org/github/mne-tools/mne-python/jobs/677529469#L3384)). Could this be a bug?

If I change [this line](https://github.com/mne-tools/mne-python/blob/6dda707b0714618c8cbf4484b3aa780b048b4122/mne/viz/backends/_pysurfer_mayavi.py#L175) to below the tests pass, but its uglier.

```python
def tube(self, origin, destination, radius=0.001, color='white',
         scalars=None, vmin=None, vmax=None, colormap='RdBu',
         normalized_colormap=False, reverse_lut=False):
    color = _check_color(color)
    origin = np.atleast_2d(origin)
    destination = np.atleast_2d(destination)
    if scalars is None:
        for idx in range(origin.shape[0]):
            surface = self.mlab.plot3d([origin[idx, 0], destination[idx, 0]],
                                       [origin[idx, 1], destination[idx, 1]],
                                       [origin[idx, 2], destination[idx, 2]],
                                       tube_radius=radius,
                                       color=color,
                                       figure=self.fig)
```

rob-luke:
The other code has been moved to https://github.com/mne-tools/mne-nirs/pull/70 as suggested by agramfort

agramfort:
@GuillaumeFavelier can you have a look?

>

mne/viz/_3d.py
```diff
@@ -1134,6 +1148,12 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
                               fwd_nn[:, ori, 1],
                               fwd_nn[:, ori, 2],
                               color=color, mode='arrow', scale=1.5e-3)
+    if 'pairs' in fnirs:
+        fnirs_loc = np.array([info['chs'][k]['loc'][3:9] for k in fnirs_picks])
+        logger.info('Plotting %d fnirs pairs' % (fnirs_loc.shape[0]))
+        renderer.tube(origin=fnirs_loc[:, :3],
+                      destination=fnirs_loc[:, 3:])
```
GuillaumeFavelier, line 1155: It's about being able to plot multiple tubes with only one instruction. I could not make it work with mayavi. This is what I ended up using:

```suggestion
        for loc in range(len(fnirs_loc)):
            x1, y1, z1 = fnirs_loc[loc, :3]
            x2, y2, z2 = fnirs_loc[loc, 3:]
            renderer.tube(origin=np.c_[x1, y1, z1],
                          destination=np.c_[x2, y2, z2])
```

rob-luke: @GuillaumeFavelier i have attempted to fix this by modifying the mayavi code to be consistent with PyVista in the commit below.

rob-luke: Or would you prefer to address this in plot|allignment as you suggest above?

rob-luke:
I can’t tell if this single CI failure due to segmentation fault is related to this PR. I recall seeing similar faults on other commits. But it does occur near the test I altered.

rob-luke:
@larsoner @agramfort @GuillaumeFavelier  could you please review. I have changed the status to mrg as I am happy with the code and scope of work now.

mne/viz/_3d.py
```diff
@@ -594,8 +594,10 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         the subjects bem and bem/flash folders are searched. Defaults to None.
     seeg : bool
         If True (default), show sEEG electrodes.
-    fnirs : bool
-        If True (default), show fNIRS electrodes.
+    fnirs : str | list | bool | None
+        Can be "channels" or "pairs" to show the fNIRS channel locations or
+        line between source-detector pairs, or a combination like
+        ``('pairs', 'channels')``. True translates to ``('channels')``.
```
larsoner, line 600:
```suggestion
        ``('pairs', 'channels')``. True translates to ``('channels',)``.
```
Also should we change the default to `('channels', 'pairs')`? If it's the sort of thing people will want most of the time then we should do it.

mne/viz/_3d.py
```diff
@@ -676,6 +685,9 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     if not all(x in ('original', 'projected') for x in eeg):
         raise ValueError('eeg must only contain "original" and '
                          '"projected", got %s' % (eeg,))
+    if not all(x in ('channels', 'pairs') for x in fnirs):
+        raise ValueError('fnirs must only contain "channels" or "pairs", '
+                         'got %s' % (fnirs,))
```
larsoner, line 690: FYI you could also do:
```
for xi, x in enumerate(fnirs):
    _check_option('fnirs[%d]' % xi, x, ('channels', 'pairs'))
```

mne/viz/_3d.py
larsoner: Agreed it's better to make `renderer` support it natively

mne/viz/backends/_pysurfer_mayavi.py
```diff
@@ -179,22 +179,31 @@ def tube(self, origin, destination, radius=0.001, color='white',
...
+            for idx in range(origin.shape[0]):
+                surface = self.mlab.plot3d([origin[idx, 0],
+                                            destination[idx, 0]],
+                                           [origin[idx, 1],
+                                            destination[idx, 1]],
+                                           [origin[idx, 2],
+                                            destination[idx, 2]],
+                                           tube_radius=radius,
+                                           color=color,
+                                           figure=self.fig)
```
larsoner, line 191: This could potentially be much slower if there is a long tube (adding thousands or tens of thousands of single tubes will probably have much worse drawing overhead versus a single contiguous tube).

- Maybe there should be a `mode='continuous'` vs `mode='separate'` or something?
- Alternatively, we could support different modes implicitly depending on whether you get an array of shape `(n_tubes, n_segments, 3)` (`n_tubes` contiguous tubes, each separate from one another) or just `(n_segments, 3)` (single contiguous tube).
- The best solution would probably be to figure out why PyVista and VTK are different. @GuillaumeFavelier at some point it would be good to look deeper into how PyVista and Mayavi do these differently at the VTK level. It seems worth it in the long run to be able to change these properties without having to create a ton of objects. 

For now @rob-luke can you just add a comment listing this as a TODO for when we run into a performance problem with this in the future?

rob-luke: Added a TODO and link to issue that @GuillaumeFavelier raised

rob-luke:
> Is it meant for there to be some tubes going not between sensors? 

Yes, these are the short channels. I have added an additional line to the tutorial to mention what they are.

mne/viz/_3d.py
rob-luke: I made it channels by default as that seemed most similar to other modalities. But happy to change to pairs if you prefer. It's explained in the tutorial, so should be accessible to users.

larsoner: What do you think fNIRS people will want to see by default? I wouldn't worry about what the other channels do but rather what's appropriate for your modality...

rob-luke: Pairs it is then :) Change committed below.

tutorials/preprocessing/plot_70_fnirs_processing.py
```diff
@@ -29,12 +29,35 @@
...
+# -------------------------------------------
+#
+# Here we validate that the location of sources-detector pairs and channels
+# are in the expected locations. Source-detector pairs are shown as lines
+# between the optodes, channels (the mid point of source-detector pairs) are
+# shown as dots.
+
+mne.datasets.fetch_fsaverage()
+subjects_dir = mne.datasets.sample.data_path() + '/subjects'
+mne.viz.set_3d_backend('pyvista')
```
larsoner, line 43: it shouldn't be necessary to have this in the example

tutorials/preprocessing/plot_70_fnirs_processing.py
```diff
@@ -29,12 +29,35 @@
...
+###############################################################################
+# View location of sensors over brain surface
+# -------------------------------------------
+#
+# Here we validate that the location of sources-detector pairs and channels
+# are in the expected locations. Source-detector pairs are shown as lines
+# between the optodes, channels (the mid point of source-detector pairs) are
+# shown as dots.
+
+mne.datasets.fetch_fsaverage()
```
larsoner, line 41: usually you do `fetch_fsaverage` or `sample.data_path()`, you should not need both. For this example I'd go with `sample` since it's more generally useful for learning MNE

If for some reason you do need `fetch_fsaverage`, it should go after the `data_path()` call and then the `subjects_dir` should be passed to `fetch_fsaverage`. But I don't think you should need it...

tutorials/preprocessing/plot_70_fnirs_processing.py
```diff
@@ -29,12 +29,35 @@
...
+subjects_dir = mne.datasets.sample.data_path() + '/subjects'
+mne.viz.set_3d_backend('pyvista')
+
+fig = mne.viz.create_3d_figure(size=(800, 600), bgcolor='white')
+fig = mne.viz.plot_alignment(raw_intensity.info, show_axes=True,
+                             subject='fsaverage',
+                             trans='fsaverage', surfaces=['brain'],
+                             fnirs=['channels', 'pairs'],
```
larsoner, line 49: You changed the default `True` to mean `('pairs',)` but then use `['channels', 'pairs']` in the example. Would `('channels', 'pairs')` might actually be the best default for `plot_alignment`? Or do you indeed want to do something non-default in this example? Sorry for the back and forth on this one...

Also would this look better with `surfaces=['brain', 'head']`? Might look cooler?

rob-luke: I leave both channels and pairs in here as they are described above and I see this tutorial as a tool to show people the nirs processing options, then they can choose the option they prefer. I think pairs is still the best default.  Also I prefer brain alone without head here.

tutorials/preprocessing/plot_70_fnirs_processing.py
rob-luke: Correct. It built fine when removed below

larsoner:
Thanks @rob-luke !

PR #7624, opened 2020-04-17
ENH: Allow path-like values passed to set_config

### LINKED ISSUE #7622: mne.set_config doesn't accept Path objects
#### Describe the bug
Passing a `pathlib.Path` value to `mne.set_config()` raises an error.

#### Steps to reproduce
```python
import pathlib
import mne

path = pathlib.Path('/tmp')
mne.set_config(key='SUBJECTS_DIR', value=path)
```
#### Expected results
Configuration would be updated.

#### Actual results
```python
Untitled-1 in 
      4 
      5 path = pathlib.Path('/tmp')
----> 6 mne.set_config(key='SUBJECTS_DIR', value=path)

~/Development/mne-python/mne/utils/config.py in set_config(key, value, home_dir, set_env)
    264     # While JSON allow non-string types, we allow users to override config
    265     # settings using env, which are strings, so we enforce that here
--> 266     _validate_type(value, (str, type(None)), "value",
    267                    "None or string")
    268 

~/Development/mne-python/mne/utils/check.py in _validate_type(item, types, item_name, type_name)
    377                 type_name[-1] = 'or ' + type_name[-1]
    378                 type_name = ', '.join(type_name)
--> 379         raise TypeError('%s must be an instance of %s, got %s instead'
    380                         % (item_name, type_name, type(item),))
    381 

TypeError: value must be an instance of None or string, got  instead
```

#### Additional information
As we support `Path` objects in other places in MNE, `set_config()` should not be an exception.

hoechenberger:
#### Reference issue
Fixes #7622

Allows to pass<strike> arbitrary</strike> `path-like` values to `set_config` <strike>as long as they're `None` or can be cast into a string. No further type checking is performed.</strike>

#### Additional information
<strike>Also this PR includes a small fix to the `config` test suite, as one of the tests was expecting a `TypeError`, while in fact a `RuntimeWarning` is raised.</strike>

### DIFF
--- a/mne/utils/config.py
+++ b/mne/utils/config.py
@@ -263,8 +263,9 @@ def set_config(key, value, home_dir=None, set_env=True):
     _validate_type(key, 'str', "key")
     # While JSON allow non-string types, we allow users to override config
     # settings using env, which are strings, so we enforce that here
-    _validate_type(value, (str, type(None)), "value",
-                   "None or string")
+    _validate_type(value, (str, 'path-like', type(None)), 'value')
+    if value is not None:
+        value = str(value)
 
     if key not in known_config_types and not \
             any(k in key for k in known_config_wildcards):
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -54,6 +54,7 @@ Bug
 
 - Fix bug with :func:`mne.io.read_raw_edf` where recording ID was not read properly for non-ASCII characters by `Lx37`_
 
+- Make :func:`mne.set_config` accept arbitrary input values as long as they can be represented as a string by `Richard Höchenberger`_
 
 API
 ~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -54,7 +54,7 @@ Bug
 
 - Fix bug with :func:`mne.io.read_raw_edf` where recording ID was not read properly for non-ASCII characters by `Lx37`_
 
-- Make :func:`mne.set_config` accept arbitrary input values as long as they can be represented as a string by `Richard Höchenberger`_
+- Make :func:`mne.set_config` accept path-like input values by `Richard Höchenberger`_
 
 API
 ~~~
--- a/mne/utils/tests/test_config.py
+++ b/mne/utils/tests/test_config.py
@@ -1,6 +1,7 @@
 from io import StringIO
 import os
 import pytest
+from pathlib import Path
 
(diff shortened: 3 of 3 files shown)


larsoner:
Actually to be safe maybe it's a good idea to keep the type checking so that people don't mistakenly think you actually can pass `float` for example and have it saved as float. One way to do this and solve your problem would be to `_validate_type(value, (None, 'path-like'), 'value')`, and cast to `str` afterward.

hoechenberger:
> One way to do this and solve your problem would be to `_validate_type(value, (None, 'path-like'), 'value')`, and cast to `str` afterward.

Can do, but isn't this exactly what @cbrnr wanted us to avoid in #7622? :)

larsoner:
His suggestion was not to restrict types because JSON dumping would handle it:, but JSON dumps allow for things like float and int to pass through:
```
>>> type(json.loads(json.dumps({'a': 1}))['a'])
<class 'int'>
```
This is a different behavior from casting to str, which we do internally for compatibility with env var setting, which is necessarily string-based. So it's actually good for us if people get errors with `int` and `float` because it helps them understand that we only store strings.

hoechenberger:
@larsoner changed accordingly

hoechenberger:
Not sure what's up with Travis…

larsoner:
Grrr `numba`... I'll look into it

mne/utils/config.py
```diff
@@ -263,8 +263,9 @@ def set_config(key, value, home_dir=None, set_env=True):
     _validate_type(key, 'str', "key")
     # While JSON allow non-string types, we allow users to override config
     # settings using env, which are strings, so we enforce that here
-    _validate_type(value, (str, type(None)), "value",
-                   "None or string")
+    _validate_type(value, (str, 'path-like', type(None)), 'value')
```
larsoner, line 266: you actually get a better error message if you pass `None` directly here:
```
>>> mne.utils._validate_type(1, None, 'test')
  File "/home/larsoner/python/mne-python/mne/utils/check.py", line 380, in _validate_type
    raise TypeError('%s must be an instance of %s, got %s instead'
TypeError: test must be an instance of None, got <class 'int'> instead
>>> mne.utils._validate_type(1, type(None), 'test')
  File "/home/larsoner/python/mne-python/mne/utils/check.py", line 380, in _validate_type
    raise TypeError('%s must be an instance of %s, got %s instead'
TypeError: test must be an instance of NoneType, got <class 'int'> instead
```
but we can always tweak this later

PR #7625, opened 2020-04-17
MRG, FIX: Fix typing for numba

larsoner:
I can replicate the problem seen on Travis locally and will file an upstream bug soon. In the meantime let's keep CIs green

https://travis-ci.org/github/mne-tools/mne-python/jobs/676241896

### DIFF
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1426,6 +1426,8 @@ def _find_nearest_tri_pts(rrs, pt_triss, pt_lens,
         pqs = np.empty((len(drs), 2))
         dists = np.empty(len(drs))
         dist = np.inf
+        # make life easier for numba var typing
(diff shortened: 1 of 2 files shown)


mne/surface.py
```diff
@@ -1426,6 +1426,8 @@ def _find_nearest_tri_pts(rrs, pt_triss, pt_lens,
         pqs = np.empty((len(drs), 2))
         dists = np.empty(len(drs))
         dist = np.inf
+        # make life easier for numba var typing
+        p, q, pt = np.float64(0.), np.float64(1.), np.int64(0)
```
larsoner, line 1430: Figured out how to actually fix it. Also opened upstream bug:

https://github.com/numba/numba/issues/5584

PR #7634, opened 2020-04-20
FIX: Circle timeout after plot_visualize_stc

GuillaumeFavelier:
I realized that since #7612 has been merged, Circle reaches timeout after `plot_visualize_stc`. I can't reproduce locally so I decided to investigate in here.

- [x] Find that the issue is related to the wrapping of `_Brain.save_movie`
- [ ] Resolve `save_movie` reference cycle (waiting for CircleCI)
- [x] Fix `UnboundLocalError`: `playback_speed_slider` used without being declared

### DIFF
--- a/mne/viz/_brain/_timeviewer.py
+++ b/mne/viz/_brain/_timeviewer.py
@@ -13,11 +13,10 @@ import sys
 
 import numpy as np
 
-from . import _Brain
 from ..utils import _check_option, _show_help, _get_color_list, tight_layout
 from ...externals.decorator import decorator
 from ...source_space import vertex_to_mni
-from ...utils import _ReuseCycle, warn, copy_doc
+from ...utils import _ReuseCycle, warn
 
 
 @decorator
@@ -362,8 +361,6 @@ class _TimeViewer(object):
         # Direct access parameters:
         self.brain = brain
         self.brain.time_viewer = self
-        self.brain._save_movie = self.brain.save_movie
-        self.brain.save_movie = self.save_movie
         self.plotter = brain._renderer.plotter
         self.main_menu = self.plotter.main_menu
         self.window = self.plotter.app_window
@@ -467,7 +464,7 @@ class _TimeViewer(object):
         self.interactor.setCursor(QCursor(Qt.WaitCursor))
 
         try:
-            self.brain._save_movie(
+            self.brain.save_movie(
                 filename=filename,
                 time_dilation=(1. / self.playback_speed),
                 callback=frame_callback,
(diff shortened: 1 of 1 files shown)


larsoner:
@GuillaumeFavelier I think it's more likely a general memory problem. Sometimes it dies on `plot_visualize_stc` but also sometimes on `plot_mne_solutions`. Here is the `mprof run plot_mne_solutions.py`:

Screenshot from 2020-04-20 08-54-39

And `plot_visualize_stc`:

Screenshot from 2020-04-20 08-58-16

It's not the same as what happens on CircleCI, because Sphinx-gallery will close and GC at the end of each plotting block. And the SG report confirms this:

https://mne.tools/dev/auto_tutorials/source-modeling/sg_execution_times.html

Same on this PR, so maybe it does indeed fix it (?):

https://19456-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/sg_execution_times.html

In any case, this all indicates that the memory usage of `_Brain` can be high so we need to make sure it gets closed and GC'ed properly.

GuillaumeFavelier:
So if I understand correctly you think that the failures on CircleCI are not related to memory but we should pay attention to it? I can open a PR about that and make it my priority. Using those 2 scripts as benchmarks. What do you think?

larsoner:
> So if I understand correctly you think that the failures on CircleCI are not related to memory but we should pay attention to it?

Either that or somehow they are related to memory and this PR fixes it (?). I have no idea why they would, though. `copy_doc` should work with `str` so it should make no reference cycles and having `_Brain` in the space doesn't seem like it should matter. Any idea why this PR seems to fix things?

Also should we merge it?

GuillaumeFavelier:
I am not sure myself. I wanted to investigate more actually:

- restoring `copy_doc`
- restoring the original function of `_Brain` during `clean`

GuillaumeFavelier:
Any way of getting more information really.

mne/viz/_brain/_timeviewer.py
```diff
@@ -362,8 +361,6 @@ def __init__(self, brain, show_traces=False):
         # Direct access parameters:
         self.brain = brain
         self.brain.time_viewer = self
-        self.brain._save_movie = self.brain.save_movie
-        self.brain.save_movie = self.save_movie
```
larsoner, line 366: Ahh, actually this could create a reference cycle that it has trouble resolving

GuillaumeFavelier:
I will push a hotfix for the missing slider too.

GuillaumeFavelier:
I will push a hotfix for the missing slider too.

GuillaumeFavelier:
The patch did not work. I suggest to revert back to something similar to edb8ccb3b9cad1e48b98eef72be60cbc1efc26e4 and merge with master. 

I will find a way to put this feature back in another PR.

GuillaumeFavelier:
This is now ready to merge @larsoner

larsoner:
Thanks @GuillaumeFavelier !

PR #7638, opened 2020-04-20
MRG, MAINT: Support only 3.5+

### LINKED ISSUE #7635: MAINT: Require 3.6+
Python 3.5 reaches EOL middle of Sept this year. Upstream libraries such as SciPy have dropped support for it. I say we do the same for the next release (0.20 being the last to support 3.5).

Bringing this up because I just did `from collections import OrderedDict` and I'd rather not have to ...

larsoner:
Closes #7635


README.rst
```diff
@@ -50,7 +50,7 @@ To install the latest stable version of MNE-Python, you can use pip_ in a termin
 
 **Note** that MNE-Python 0.17 was the last release to support Python 2.
 MNE-Python 0.18 only works under Python 3, and MNE-Python 0.19 requires
-Python 3.5 or higher.
+Python 3.6 or higher.
```
drammock, line 53:
```suggestion
MNE-Python 0.18 and later only work with Python 3. MNE-Python 0.21 requires
Python 3.6 or higher.
```

mne/channels/montage.py
```diff
@@ -119,8 +118,6 @@ def make_dig_montage(ch_pos=None, nasion=None, lpa=None, rpa=None,
         ch_names = None
     else:
         ch_names = list(ch_pos.keys())
```
drammock, line 120: shouldn't need the `.keys()` here, right?

mne/io/_digitization.py
```diff
@@ -426,9 +426,7 @@ def _make_dig_points(nasion=None, lpa=None, rpa=None, hpi=None,
                         'kind': FIFF.FIFFV_POINT_EXTRA,
                         'coord_frame': coord_frame})
     if dig_ch_pos is not None:
-        keys = dig_ch_pos.keys()
-        if not isinstance(dig_ch_pos, OrderedDict):
-            keys = sorted(keys)
+        keys = list(dig_ch_pos.keys())
```
drammock, line 429: shouldn't need `.keys()` here right?

cbrnr:
:+1: for type-checking against an ABC.

larsoner:
> Do you think it's worth it to start using abstract base class Collection in type checking?

Maybe but it will probably need more testing, so I'd rather let someone else tackle it in another PR

agramfort:
would someone with python 2.5 still be able to use mne 0.21 and have silent bugs?

larsoner:
> would someone with python 2.5 still be able to use mne 0.21 and have silent bugs?

No because:

1. imports like `from collections.abc` will fail, and
2. we won't even allow them to install with our new `python_requires` line in `setup.py`

agramfort:
all good then

>

larsoner:
> Any parts of the codebase that would benefit from namedtuple?

In principle anywhere we return a tuple could probably benefit from it, I guess

larsoner:
Rebased, okay to merge once green @agramfort ?

agramfort:
+1 for MRG when CIs are happy

README.rst
```diff
@@ -48,9 +48,9 @@ To install the latest stable version of MNE-Python, you can use pip_ in a termin
 
     pip install -U mne
 
-**Note** that MNE-Python 0.17 was the last release to support Python 2.
-MNE-Python 0.18 only works under Python 3, and MNE-Python 0.19 requires
-Python 3.5 or higher.
+- MNE-Python 0.17 was the last release to support Python 2.7
+- MNE-Python 0.18 requires Python 3.5 or higher
```
cbrnr, line 52: Why not mention 0.18-0.20? It's not important though.

larsoner: We're only mentioning the points at which requirements change

mne/channels/tests/test_montage.py
```diff
@@ -847,6 +847,16 @@ def test_read_dig_captrack(tmpdir):
...
+    EXPECTED_CH_NAMES = [
+        'T7', 'FC5', 'F7', 'C5', 'FT7', 'FT9', 'TP7', 'TP9', 'P7', 'CP5',
+        'PO7', 'C3', 'CP3', 'P5', 'P3', 'PO3', 'PO9', 'O1', 'Oz', 'POz', 'O2',
+        'PO4', 'P1', 'Pz', 'P2', 'CP2', 'CP1', 'CPz', 'Cz', 'C1', 'FC1', 'FC3',
+        'REF', 'F3', 'F1', 'Fz', 'F5', 'AF7', 'AF3', 'Fp1', 'GND', 'F2', 'AF4',
+        'Fp2', 'F4', 'F8', 'F6', 'AF8', 'FC2', 'FC6', 'FC4', 'C2', 'C4', 'P4',
+        'CP4', 'PO8', 'P8', 'P6', 'CP6', 'PO10', 'TP10', 'TP8', 'FT10', 'T8',
+        'C6', 'FT8'
+    ]
+    assert set(EXPECTED_CH_NAMES) == set(EXPECTED_CH_NAMES_OLD)
```
sappelhoff, line 859: @larsoner why this change?

larsoner: Instead of being alphabetical it now should be the order on disk. I figured it shouldn't affect being able to actually use it with our montage functions...

PR #7597, opened 2020-04-10
MRG, ENH: Improve mixed source estimate support

larsoner:
- [x] Redefine "mixed" source estimate as two cortical surface source spaces, plus others
- [x] Add MixedVectorSourceEstimate
- [x] Add tests for generating all source space types in inversion (mixed not tested, mixed vector not tested)
- [x] Add MixedVectorSourceEstimate.surface() and .volume() to allow splitting (someday later maybe we'll add putting them back together but not for now)
- [x] Add round-trip I/O tests for Mixed(Vector)SourceEstimate
- [x] Add support for MixedSourceEstimate to `extract_label_time_course`
- [x] Deprecate `MixedSourceEstimate.plot_surface` in favor of `stc.surface().plot()`

Next PRs:

- Add support for `*VectorSourceEstimate` to `extract_label_time_course`
- Extend `SourceMorph` to handle mixed source spaces + estimates
- Add to tutorials and definitions of various SourceEstimate flavors

In the long run:

- Add support for `Vol/MixedSourceEstimate.plot` to `_Brain` somehow (volume rendering?)

Closes #7564

Issue #7564, opened 2020-04-06
issue with volumetric source estimate morphing


fleurgaudfernau:
#### Describe the bug
I have mixed source estimates that I wish to morph to the fsaverage brain. To do so, I separated the mixed source estimates into a surface source estimate and a volume source estimate. When trying to morph the lastest (using morph_apply), I get the following error:
ValueError: vertices do not match between morph (77) and stc (33) for the volume:
(also, I believe there is a small mistake in morph_apply and what it means is actually that vertices do not match between morph (33) and stc (77) for the volume.)

I believe the error is coming from the fact that when computing the source morph, mne only considers the first of the two volume source spaces in cereb_used, hence generating a morph with 33 vertices instead of 77 vertices, but I do not know how to fix it.

#### Steps to reproduce
```python
import mne
import os
import os.path as op
import numpy as np
from mne.datasets import sample, fetch_fsaverage
from mne.minimum_norm import apply_inverse, read_inverse_operator, make_inverse_operator

data_path = mne.datasets.sample.data_path()
subject = 'sample'
data_dir = op.join(data_path, 'MEG', subject)
subjects_dir = op.join(data_path, 'subjects')
bem_dir = op.join(subjects_dir, subject, 'bem')

fname_mixed_src = op.join(bem_dir, '%s-oct-6-mixed-src.fif' % subject)
fname_aseg = op.join(subjects_dir, subject, 'mri', 'aseg.mgz')
fname_model = op.join(bem_dir, '%s-5120-bem.fif' % subject)
fname_bem = op.join(bem_dir, '%s-5120-bem-sol.fif' % subject)
fname_evoked = data_dir + '/sample_audvis-ave.fif'
fname_trans = data_dir + '/sample_audvis_raw-trans.fif'
fname_fwd = data_dir + '/sample_audvis-meg-oct-6-mixed-fwd.fif'
fname_cov = data_dir + '/sample_audvis-shrunk-cov.fif'

#create a mixed source space
labels_vol = ['Left-Cerebellum-Cortex',
              'Right-Cerebellum-Cortex']
src = mne.setup_source_space(subject, spacing='oct5',
                             add_dist=False, subjects_dir=subjects_dir)

vol_src = mne.setup_volume_source_space(
    subject, mri=fname_aseg, pos=10.0, bem=fname_model,
    volume_label=labels_vol, subjects_dir=subjects_dir,
    add_interpolator=False,  # just for speed, usually this should be True
    verbose=True)

[...]

Issue #7007, opened 2019-10-31
DOC: morph generated from src does not work because fwd exlcude some vertices


kingjr:
```python
"""Prep example from scratch to make sure we have a unique pipeline"""
import mne
import os
from mne.datasets import sample
from mne.minimum_norm import apply_inverse, make_inverse_operator

# Paths
data_path = sample.data_path()
subject = 'sample'
subjects_dir = os.path.join(data_path, 'subjects')
sample_dir = os.path.join(data_path, 'MEG', subject)

# Functional pipeline
raw_fname = sample_dir + '/sample_audvis_filt-0-40_raw.fif'
events_fname = sample_dir + '/sample_audvis_filt-0-40_raw-eve.fif'
raw = mne.io.read_raw_fif(raw_fname)
events = mne.read_events(events_fname)
epochs = mne.Epochs(raw, events, tmin=-.100, tmax=.200)
cov = mne.compute_covariance(epochs, tmax=0)
evo = epochs.average()

# Structural pipeline
trans_fname = sample_dir + '/sample_audvis_raw-trans.fif'
src = mne.setup_source_space(subject, spacing='oct6',
                             subjects_dir=subjects_dir, 
                             add_dist=False)
morph = mne.compute_source_morph(src,
                                 subject_to='fsaverage',
                                 subjects_dir=subjects_dir)
model = mne.make_bem_model(subject=subject, 
                           ico=4,
                           conductivity=(.3,),
                           subjects_dir=subjects_dir)
bem = mne.make_bem_solution(model)
fwd = mne.make_forward_solution(epochs.info, 
                                trans=trans_fname, 
                                src=src, bem=bem,
                                meg=True, eeg=False, mindist=5.0, n_jobs=-1)

# Inverse modeling
inv = make_inverse_operator(epochs.info, fwd, cov, loose=0., depth=0.8)
stc = apply_inverse(evo, inv)
# -----

```

critically
```python
# morphing from stc to stc works
morph = mne.compute_source_morph(stc,
                                 subject_from=subject,
                                 subject_to='fsaverage',
                                 subjects_dir=subjects_dir)

stc_success = morph.apply(stc)

# morphing from src to stc does not
morph = mne.compute_source_morph(src,
                                 subject_from=subject,
                                 subject_to='fsaverage',
                                 subjects_dir=subjects_dir)
stc_fails = morph.apply(stc)
```

returns 
```
[...]

PR #7639, opened 2020-04-20
MRG, ENH: Add volumetric atlas support

### LINKED ISSUE #6155: ENH: Volume labeling
At some point we should have a way of "label-izing" volumes like we do for surfaces. Freesurfer probably has utilities/files for this but I've never looked or worked with them (probably aseg.mgz), and in general we should have a way of making volume labels. Basically we need (API proposal):

1. `read_labels_from_aseg` (or whatever the correct Freesurfer file is), which returns a list of...
2. `VolumeLabel` or some equivalent (needs suitabl API and I/O functions)
3. `extract_label_time_course` extended to work for `VolSourecEstimate` + `VolLabel`

Eventually we should make everything work for mixed source spaces, too, but there are probably other things to sort out before then (e.g., mixed orientation priors).

larsoner:
Adds volume label support to `extract_label_time_course`. Requires #7597 (and follow-up PRs).

- [x] Add `mne.read_freesurfer_lut` as suggested by @vferat [here](https://github.com/mne-tools/mne-python/issues/6155#issuecomment-612420561) to make `setup_volume_source_space` more general and pave the way for more generic atlas support
- [x] Speed up `setup_volume_source_space` with many labels by refactoring `_make_volume_source_space`
- [x] Add test that `setup_volume_source_space` with all volume labels of `aseg.mgz` creates a complete source space
- [x] Add `extract_label_time_course` VolSourceEstimate support via atlas
- [x] Add `vol_stc.in_label` volume atlas support

Next PRs:

- Ensure all source estimate types work with SourceMorph (#7564)
- Update docs about `src` in SourceMorph (#7007)

Closes #6155.

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -1638,12 +1638,16 @@ class SourceEstimate(_BaseSurfaceSourceEstimate):
         if hemi == 'lh':
             use = self.__class__(
                 self.lh_data, [self.vertices[0], []], self.tmin, self.tstep)
+            meth = use.get_peak
         elif hemi == 'rh':
             use = self.__class__(
                 self.rh_data, [[], self.vertices[1]], self.tmin, self.tstep)
+            meth = use.get_peak
         else:
-            use = self
-        return use.get_peak(tmin, tmax, mode, vert_as_index, time_as_index)
+            meth = super().get_peak
+        return meth(tmin=tmin, tmax=tmax, mode=mode,
+                    vert_as_index=vert_as_index,
+                    time_as_index=time_as_index)
 
     @fill_doc
     def center_of_mass(self, subject=None, hemi=None, restrict_vertices=False,
@@ -2806,11 +2810,11 @@ def _gen_extract_label_time_course(stcs, labels, src, mode='mean',
                 label_tc[i] = func(flip, stc.data[vertidx])
 
         # extract label time series for the vol src space (only mean supported)
-        offset = nvert[:-n_aseg].sum()  # effectively :2 or :0
+        offset = nvert[:-n_mean].sum()  # effectively :2 or :0
         for i, nv in enumerate(nvert[2:]):
             if nv != 0:
                 v2 = offset + nv
-                label_tc[n_aparc + i] = np.mean(stc.data[offset:v2], axis=0)
+                label_tc[n_mode + i] = np.mean(stc.data[offset:v2], axis=0)
                 offset = v2
 
         # this is a generator!
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -23,14 +23,15 @@ from .surface import read_surface, _get_ico_surface, mesh_edges
 from .source_space import (_ensure_src, _get_morph_src_reordering,
                            _ensure_src_subject, SourceSpaces, _get_src_nn,
                            _import_nibabel, _get_mri_info_data,
-                           _get_atlas_values)
+                           _get_atlas_values, _check_volume_labels,
+                           read_freesurfer_lut)
 from .transforms import _ensure_trans, apply_trans
 from .utils import (get_subjects_dir, _check_subject, logger, verbose, _pl,
                     _time_mask, warn, copy_function_doc_to_method_doc,
                     fill_doc, _check_option, _validate_type, _check_src_normal,
                     _check_stc_units, _check_pandas_installed, deprecated,
                     _check_pandas_index_arguments, _convert_times,
-                    _build_data_frame, _check_time_format)
+                    _build_data_frame, _check_time_format, _check_path_like)
 from .viz import (plot_source_estimates, plot_vector_source_estimates,
                   plot_volume_source_estimates)
 from .io.base import TimeMixin
@@ -2803,18 +2804,35 @@ def _volume_labels(src, labels, trans, mri_resolution):
     # given volumetric source space when used with extract_label_time_course
     from .label import Label
     assert src.kind == 'volume'
-    nib = _import_nibabel('use volume atlas labels')
-    _validate_type(labels, (str, nib.MGHImage),
-                   'labels when using a volume source space')
-    logger.info('Reading atlas %s' % (labels,))
-    vol_info = _get_mri_info_data(labels, data=True)
+    extra = ' when using a volume source space'
+    _import_nibabel('use volume atlas labels')
+    _validate_type(labels, ('path-like', list, tuple), 'labels' + extra)
+    if _check_path_like(labels):
+        mri = labels
+        infer_labels = True
+    else:
+        if len(labels) != 2:
+            raise ValueError('labels, if list or tuple, must have length 2, '
+                             'got %s' % (len(labels),))
+        mri, labels = labels
+        infer_labels = False
+        _validate_type(mri, 'path-like', 'labels[0]' + extra)
+    logger.info('Reading atlas %s' % (mri,))
+    vol_info = _get_mri_info_data(str(mri), data=True)
     atlas_values = np.unique(vol_info['data'])
     atlas_values = atlas_values[np.isfinite(atlas_values)]
-    atlas_values = atlas_values[atlas_values != 0]
     if not (atlas_values == np.round(atlas_values)).all():
         raise RuntimeError('Non-integer values present in atlas, cannot '
                            'labelize')
     atlas_values = np.round(atlas_values).astype(np.int64)
+    if infer_labels:
+        labels = {
+            k: v for k, v in read_freesurfer_lut()[0].items()
+            if v in atlas_values}
+    labels = _check_volume_labels(labels, mri, name='labels[1]')
+    assert isinstance(labels, dict)
+    del atlas_values
+
     vox_mri_t = vol_info['vox_mri_t']
     want = src[0].get('vox_mri_t', None)
     if want is None:
@@ -2833,17 +2851,17 @@ def _volume_labels(src, labels, trans, mri_resolution):
                            'shape %s' % (atlas_shape, src_shape))
     if mri_resolution:
         # Upsample then just index
-        labels = list()
+        out_labels = list()
         nnz = 0
         interp = src[0]['interpolator']
         # should be guaranteed by size checks above and our src interp code
         assert interp.shape[0] == np.prod(src_shape)
         assert interp.shape == (vol_info['data'].size, len(src[0]['rr']))
         interp = interp[:, src[0]['vertno']]
-        for v in atlas_values:
+        for v in labels.values():
             mask = vol_info['data'].ravel(order='F') == v
             csr = interp[mask]
-            labels.append(dict(csr=csr))
+            out_labels.append(dict(csr=csr))
             nnz += csr.shape[0] > 0
     else:
         # Use nearest values
@@ -2854,13 +2872,13 @@ def _volume_labels(src, labels, trans, mri_resolution):
             rr = apply_trans(_ensure_trans(trans, 'head', 'mri'), rr)
         del src
         src_values = _get_atlas_values(vol_info, rr[vertno])
-        vertices = [vertno[src_values == val] for val in atlas_values]
-        labels = [Label(v, hemi='lh', name=val)
-                  for v, val in zip(vertices, atlas_values)]
+        vertices = [vertno[src_values == val] for val in labels.values()]
+        out_labels = [Label(v, hemi='lh', name=val)
+                      for v, val in zip(vertices, labels.keys())]
         nnz = sum(len(v) != 0 for v in vertices)
     logger.info('%d/%d atlas regions had at least one vertex '
-                'in the source space' % (nnz, len(atlas_values)))
-    return labels
+                'in the source space' % (nnz, len(out_labels)))
+    return out_labels
 
 
 def _gen_extract_label_time_course(stcs, labels, src, mode='mean',
@@ -2926,7 +2944,8 @@ def _gen_extract_label_time_course(stcs, labels, src, mode='mean',
                     assert vertidx.shape[1] == stc.data.shape[0]
                     this_data = np.reshape(stc.data, (stc.data.shape[0], -1))
                     this_data = vertidx * this_data
-                    this_data.shape = (this_data.shape[0],) + stc.data.shape[1:]
+                    this_data.shape = \
+                        (this_data.shape[0],) + stc.data.shape[1:]
                     label_tc[i] = func(flip, this_data)
                 else:
                     this_data = stc.data[vertidx]
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -20,7 +20,11 @@ from .filter import resample
 from .fixes import einsum
 from .surface import read_surface, _get_ico_surface, mesh_edges
 from .source_space import (_ensure_src, _get_morph_src_reordering,
-                           _ensure_src_subject, SourceSpaces, _get_src_nn)
+                           _ensure_src_subject, SourceSpaces, _get_src_nn,
+                           _import_nibabel, _get_mri_info_data,
+                           _get_atlas_values)
+from .transforms import (_angle_between_quats, rot_to_quat,
+                         _print_coord_trans)
 from .utils import (get_subjects_dir, _check_subject, logger, verbose,
                     _time_mask, warn, copy_function_doc_to_method_doc,
                     fill_doc, _check_option, _validate_type, _check_src_normal,
@@ -2745,11 +2749,30 @@ def _prepare_label_extraction(stc, labels, src, mode, allow_empty):
 def _volume_labels(src, labels):
     # This will create Label objects that should do the right thing for our
     # given volumetric source space when used with extract_label_time_course
-    import nibabel
+    from .label import Label
     assert src.kind == 'volume'
-    _validate_type(labels, (str, nibabel.spatialimages.SpatialImage),
+    nib = _import_nibabel('use volume atlas labels')
+    _validate_type(labels, (str, nib.MGHImage),
                    'labels when using a volume source space')
-    raise NotImplementedError
+    logger.info('Reading atlas %s' % (labels,))
+    vol_info = _get_mri_info_data(labels, data=True)
+    vox_mri_t = vol_info['vox_mri_t']
+    want = src[0].get('vox_mri_t', None)
+    if want is None:
+        raise RuntimeError(
+            'Cannot use volumetric atlas if no mri was supplied during '
+            'source space creation')
+    _print_coord_trans(vox_mri_t, 'MRI volume : ')
+    vox_mri_t, want = vox_mri_t['trans'], want['trans']
+    if not np.allclose(vox_mri_t, want, atol=1e-6):
+        raise RuntimeError(
+            'atlas vox_mri_t does not match that used to create the source '
+            'space')
+    vertno = src[0]['vertno']
+    src_values = _get_atlas_values(vol_info, src[0]['rr'][vertno])
+    labels = [Label(vertno[src_values == val])
+              for val in np.unique(vol_info['data'])]
+    return labels
 
 
 def _gen_extract_label_time_course(stcs, labels, src, mode='mean',
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -25,12 +25,12 @@ from .source_space import (_ensure_src, _get_morph_src_reordering,
                            _import_nibabel, _get_mri_info_data,
                            _get_atlas_values, _check_volume_labels,
                            read_freesurfer_lut)
-from .transforms import _ensure_trans, apply_trans
+from .transforms import _get_trans, apply_trans
 from .utils import (get_subjects_dir, _check_subject, logger, verbose, _pl,
                     _time_mask, warn, copy_function_doc_to_method_doc,
                     fill_doc, _check_option, _validate_type, _check_src_normal,
                     _check_stc_units, _check_pandas_installed, deprecated,
-                    _check_pandas_index_arguments, _convert_times,
+                    _check_pandas_index_arguments, _convert_times, _ensure_int,
                     _build_data_frame, _check_time_format, _check_path_like)
 from .viz import (plot_source_estimates, plot_vector_source_estimates,
                   plot_volume_source_estimates)
@@ -567,8 +567,7 @@ class _BaseSourceEstimate(TimeMixin):
 
         Returns
         -------
-        label_tc : array, shape (n_labels, n_times)
-            Extracted time course for each label.
+        %(eltc_returns)s
 
         See Also
         --------
@@ -1640,13 +1639,8 @@ class SourceEstimate(_BaseSurfaceSourceEstimate):
             or index.
         """
         _check_option('hemi', hemi, ('lh', 'rh', None))
-        if hemi == 'lh':
-            use = self.__class__(
-                self.lh_data, [self.vertices[0], []], self.tmin, self.tstep)
-            meth = use.get_peak
-        elif hemi == 'rh':
-            use = self.__class__(
-                self.rh_data, [[], self.vertices[1]], self.tmin, self.tstep)
+        if hemi is not None:
+            use = self._hemilabel_stc(hemi)
             meth = use.get_peak
         else:
             meth = super().get_peak
@@ -1836,8 +1830,7 @@ class _BaseVolSourceEstimate(_BaseSourceEstimate):
 
         Returns
         -------
-        label_tc : array, shape (n_labels, n_times)
-            Extracted time course for each label.
+        %(eltc_returns)s
 
         See Also
         --------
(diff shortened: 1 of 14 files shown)


larsoner:
Okay @agramfort this is ready for review/merge as well. Might be good for you to take a first pass, then once we're happy with the API ask others who would be interested to look

larsoner:
The new tutorial [image](https://19612-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_visualize_stc.html#sphx-glr-auto-tutorials-source-modeling-plot-visualize-stc-py) showing `extract_label_time_course` on a volume STC:

mne/source_estimate.py
```diff
@@ -1791,6 +1808,89 @@ def plot(self, src, subject=None, subjects_dir=None, mode='stat_map',
...
+        --------
+        extract_label_time_course : Extract time courses for multiple STCs.
+
+        Notes
+        -----
+        %(eltc_mode_notes)s
+        """
+        return extract_label_time_course(
+            self, labels, src, mode=mode, return_generator=False,
+            allow_empty=allow_empty, verbose=verbose)
```
agramfort, line 1845: you don't pass `trans` or `mri_resolution` parameters here. I am confused

larsoner: Just a bug, need to write tests to use this method in addition to the function !

larsoner: Fixed

poojaprabhu9:
I was trying to read the volume stc which were computed previously using mne.read_source_estimate in mne.v0.21.dev0.
I met with an error
TypeError: vertices must be an instance of list, got <class 'numpy.ndarray'> instead.
without reading stcs i cannot check the extract_label_time_course ,correct?

larsoner:
@poojaprabhu9 I pushed a fix that should fix that bug hopefully

poojaprabhu9:
I tried the same by installing mne.v0.21.dev0. 
I error still exist.

agramfort:
you need to use this branch. Try:

pip install -U https://github.com/larsoner/mne-python/archive/atlas.zip

poojaprabhu9:
that helped, thanks.
I had computed stc after morphing.
Thus, now when i am extracting the time course, i have to setup the volume source space for  fsaverage  to get ''src" (same src was used while doing morphing)
```
src_fs = mne.setup_volume_source_space( 'fsaverage', pos=5.,mri=aseg_fname, 
             subjects_dir=subjects_dir,verbose=True)
```
then i tried to extract the time course from the stc

`tc=stc_morph.extract_label_time_course(labels='/path/fsaverage/mri/aparc+aseg.mgz',src=src_fs,mode='mean',allow_empty=True)`

i am encountering the unexpected error:
ValueError: 9900/20377 left hemisphere stc vertices missing from the source space, likely mismatch.

How is that possible when i am using the same src to extract the time course as well as to compute the morphed stc?

agramfort:
can you share a code snippet to replicate ?

>

poojaprabhu9:
I tried to replicate the code  [here](https://gist.github.com/poojaprabhu9/22e36e4d44d1d53da83fbec0c54ddd07)
two observations are done,
1. when i execute the code in mne.v.0.20,  the stc obtained after morphing has 20377 vertices whereas when executed in development version(pip install -U https://github.com/larsoner/mne-python/archive/atlas.zip ) it has 24303 vertices. I can extract the time course from the morphed stc computed from development version. In my case i have the morphed stc computed using mne.v.0.20, when i read this in development version and try to extract the time course i get the ValueError: 9900/20377 left hemisphere stc vertices missing from the source space, likely mismatch.
2.  when i try to execute (line 49 of gist)
`src = mne.setup_volume_source_space('sample', mri=aseg_fname,pos=5., sphere=sphere, mindist=3.,subjects_dir=subjects_dir)`
 in development version i get 
MemoryError: Unable to allocate 402. TiB for an array with shape (48003, 48003, 48003) and data type int32
whereas it works fine in mne.v.0.40.
so you can't run the whole code in development version. So i computed the forward solution in mne.v.0.20 and then i read the forward solution in development version.

larsoner:
You have to be careful about sphere_units, they changed between versions. You probably need to set this to mm, the default is now m.

In general if you stick to using dev versions you need to follow changes or read the release notes very closely because deprecation cycles can be effectively days or weeks in duration instead of 6 months.

poojaprabhu9:
> You have to be careful about sphere_units, they changed between versions. You probably need to set this to mm, the default is now m.
> 
> In general if you stick to using dev versions you need to follow changes or read the release notes very closely because deprecation cycles can be effectively days or weeks in duration instead of 6 months.

that solved the second point.
first problem is solved by adding sphere=(0,0,0,90) to 
```
src_fs = mne.setup_volume_source_space( 'fsaverage', pos=5.,mri=aseg_fname, 
             subjects_dir=subjects_dir,verbose=True)
```
Then compute the extract_time_course. 
I could extract the time course. Thanks!

larsoner:
Okay for you @vferat ?

mne/source_space.py
```diff
@@ -2535,36 +2615,37 @@ def get_volume_labels_from_aseg(mgz_fname, return_colors=False):
...
-                     lut[lut['id'] == ii]['A'][0]]
-                    for ii in np.unique(mgz_data)]
-
-    order = np.argsort(label_names)
-    label_names = [label_names[k] for k in order]
-    label_colors = [label_colors[k] for k in order]
-
+    atlas = nib.load(mgz_fname)
+    want = np.unique(_get_img_fdata(atlas))
+    ids, colors = read_freesurfer_lut()
```
vferat, line 2632: Would it be possible to add a way to use specific freesurfer LUT here ? Adding an input either an object returned by or a file path to be read by `read_freesurfer_lut` :

`def get_volume_labels_from_aseg(mgz_fname, lut=None, return_colors=False):`

larsoner: So you want both `read_freesurfer_lut(fname=None)`, and `get_volume_labels_from_aseg(mgz_fname, return_colors=False, lut=None)` where non-None inputs mean "don't use the builtin one"? Seems doable

larsoner: Actually `read_freesurfer_lut` already takes `fname`, so we just need `lut` support for `get_volume_labels_from_aseg`

larsoner: added `atlas_ids` argument in latest commit

larsoner:
@vferat should be fixed now

larsoner:
Oh wait I just fixed the error you were getting (pre-edit). Let me look...

larsoner:
@vferat I get:
```
>>> len(mne.get_volume_labels_from_aseg('sample_Schaefer2018_100Parcels_7Networks.mgz'))
114
```
Maybe some bug with your code?

larsoner:
Okay I see the problem. When you do only:
```
label_tc = stc.extract_label_time_course(fname_aseg, src=src, trans=inv['mri_head_t'])
```
It will only use the builtin FreeSurfer LUT with the `aseg` you have provided. You need to pass the `atlas_ids` you want to use with this non-standard MRI, e.g.:
```
atlas_ids, lut_colors = mne.read_freesurfer_lut('Schaefer2018_100Parcels_7Networks_LUT.txt')
label_names = mne.get_volume_labels_from_aseg(fname_aseg, atlas_ids=atlas_ids)
# not sure if you need this, but just to be safe:
atlas_ids = {k: atlas_ids[k] for k in label_names}
label_tc = stc.extract_label_time_course((fname_aseg, atlas_ids), src=src, trans=inv['mri_head_t'])
```

vferat:
> 
> 
> Okay I see the problem. When you do only:
> 
> ```
> label_tc = stc.extract_label_time_course(fname_aseg, src=src, trans=inv['mri_head_t'])
> ```
> 
> It will only use the builtin FreeSurfer LUT with the `aseg` you have provided. You need to pass the `atlas_ids` you want to use with this non-standard MRI, e.g.:
> 
> ```
> atlas_ids, lut_colors = mne.read_freesurfer_lut('Schaefer2018_100Parcels_7Networks_LUT.txt')
> label_names = mne.get_volume_labels_from_aseg(fname_aseg, atlas_ids=atlas_ids)
> # not sure if you need this, but just to be safe:
> atlas_ids = {k: atlas_ids[k] for k in label_names}
> label_tc = stc.extract_label_time_course((fname_aseg, atlas_ids), src=src, trans=inv['mri_head_t'])
> ```

Ok well I see, thanks ! I tested it on another dataset and it's working fine. 👍

larsoner:
Since @agramfort approves and it seems to be working for people, let's get this in so that people can start using it. Thanks for testing and iterating @poojaprabhu9 @vferat !

poojaprabhu9:
stc.extract_label_time_course( labels, src, mode='auto', allow_empty=False,trans=None, mri_resolution=True, verbose=None)
We will be using the HCPMMP1 parcellation. This parcellation has atlas in .nii format. 
While passing the atlas path to the 'labels', is it mandatory to have the atlas in .mgz format?

larsoner:
Yes right now, because everything is done in Freesurfer surface RAS coordinates (`vox_mri_t` in MNE, `get_vox2ras_tkr` in nibabel) and this doesn't exist AFAIK for NIfTI. Right now you have to make sure that whatever atlas image you use has been transformed properly to the subject's MRI space so that things like `mri_resolution=True` work properly, and that a given voxel in your atlas corresponds to the same voxel in your `T1.mgz`. The builtin Freesurfer `aseg.mgz` meets these requirements.

Have you already transformed the HCPMMP1 parcellation to your subject's MRI space such that a given voxel in your HCPMMP1 volume corresponds to the same voxel in their `aseg.mgz` / `T1.mgz`? If so, how did you do this transformation (we should document it)? If on the other hand you just have HCPMMP1 in NIfTI for fsaverage, we'll need to figure out how to transform it to the subject's space in the first place. It should be doable, but we need to figure it out. Once we figure that out, we can document it and then figure out what needs to be done at the MNE end to facilitate using other atlases.

poojaprabhu9:
Check this link: https://cjneurolab.org/2016/11/22/hcp-mmp1-0-volumetric-nifti-masks-in-native-structural-space/
As per the link, the code to create volume parcellation using HCPMMP1 is [here](https://gist.github.com/poojaprabhu9/81bd61e081fc52c17392ffaaf0673e7a). the final volume will be in .nii format.
We just need it for fsaverage brain. 
We need to apply this parcellation on the already morphed volume estimates. For this purpose we have to extract the time courses for each labels of HCPMMP1 parcellation.  Currently, to extract the time courses we have the atlas in nifti. So the previous question.

larsoner:
So `fsaverage` is pretty nice:
```
>>> import nibabel as nib
>>> img = nib.load('/mnt/bakraid/data/structurals/fsaverage/mri/aseg.mgz')
>>> import numpy as np
>>> np.array_equal(img.affine, img.header.get_vox2ras_tkr())
True
```
You don't have to worry about these differences. So in theory if you just save your `nii` to `mgz`, the MNE code should work, because (I think) `nibabel` will write the `vox2ras_tkr` as being equivalent to `vox2ras` / `affine`.

larsoner:
... and personally I would probably update that script you used to output .mgz files with the vox2ras_tkr set properly. Then it would even work for non-fsaverage subjects.

poojaprabhu9:
You mean to say I can use atlas in .nii format to extract time courses from stc?
(Without updating the MNE)

larsoner:
> Or you mean I need to convert .nii to .mgz? Then try to extract the time course

You almost certainly need to convert it to `mgz`. Even just doing `mri_convert /path/whatever.nii /path/whatever.mgz` might be enough for things to work.

virvw:
Hi there,

I am understanding Pooja wishes to extract volumic parcels using the
HCPMMP1 for our source analyses.
We are slowly advancing. Bear in mind you are not talking to coders but
simple scientists trying to make whatever receipe you guys mention work :)

I ran 45 min of a code to realize that "wm.mgz" was missing in the
fsaverage. It should be mentionned beforehand.
So before I continue solving error after bug sequentially: where can I get
a sense of the ingredients that are first needed to get the final receipe
to work?

------------------------------------------------
Right now, I came up with a rather unsatiscatory recipe because I think a
lot of initial variables are missing:

[This procedure takes 45 min +]

To create HCPMMP volumetric parcels on NeuroSpin server:
1) copy "fsaverage" in the MRI folder of your study
2) copy "lh.HCPMMP1.annot" and "rh.HCPMMP1.annot" in /fsaverage/label/
https://cjneurolab.org/2016/11/22/hcp-mmp1-0-volumetric-nifti-masks-in-native-structural-space/

3) create a "subject_list.txt" simply listing "fsaverage"
4) type to initialize freesurfer:
freesurfer_init
5) type to temporarily set your default $SUBJECTS_DIR to your study:
export $SUBJECTS_DIR=/neurospin/meg/meg_tmp/mystudy/MRI
6) type to inialize fsl  :
fsl_init
7) type to create the parcel into the folder HCPMMP1_parcellation:
bash create_subj_volume_parcellation.sh -L subject_list.txt -a HCPMMP1 -d
HCPMMP1_parcellation -s YES -t YES
---------------------------------------

Thanks a lot, much appreciated


Virginie van Wassenhove
*https://brainthemind.com/ <https://brainthemind.com/>*

larsoner:
We can probably ship some HCPMMP1 + aseg volumes for `fsaverage` at least. I think that doing the conversion is out of scope for us. But it sounds like having the volumes for `fsaverage` is already enough, assuming you're morphing all subject data there anyway (?).

larsoner:
... looking into this a bit, the solution they posted is a start but it's not exactly straightforward to get it working. I trimmed the script a bit just to use with fsaverage and output an mgz file. However, it does not output the actual region LUT that is to be used with the volume, so this needs to be created based on the math it actually does. And doing these things in Bash is not exactly straightforward.

I think it would actually be easier to do most of this in Python with nibabel. All that we really need FreeSurfer for (I think) is to generate a aseg volume from the annotation:
```
mri_aparc2aseg --s fsaverage --o HCPMMP1.mgz  --annot HCPMMP1
```
From there in Python we can:

1. Load `HCPMMP1.mgz`
2. Remove hippocampal voxels (they do this in the script)
3. Shift the values by some amount (they do this in the script)
4. Load the standard `aseg.mgz` for the subject, and add it to the HCPMMP one (including only whatever regions should be included)
5.  Keep track of which vertex values correspond to which regions, and then write out a custom LUT.

Eventually this could be done for other subjects, too, for example by using `read_labels_from_annot` followed by `morph_labels` then `write_labels_to_annot` for each subject (they do this sort of operation in their bash script).

For now, perhaps try using `aparc.a2009s+aseg.mgz`. Each FreeSurfer-reconstructed subject should already have this. It's a much less challenging path, and you should already have the files you need. In principle later HCPMMP1 can be substituted as the atlas+LUT combination, but in the meantime you can set up your scripts to work without this being a bottleneck.

larsoner:
@poojaprabhu9 @virvw can you try this FreeSurfer command (after running `mne.datasets.fetch_hcpmmp_parcellation`):
```
$ mri_aparc2aseg --s fsaverage --volmask --annot HCPMMP1
$ mri_aparc2aseg --s fsaverage --volmask --annot HCPMMP1_combined
```
Then run [this gist](https://gist.github.com/larsoner/8e664205cd8285ca7c46211403ad12ce), which will create `SUBJECTS_DIR/fsaverage/mri/HCPMMP1ColorLUT.txt` and `HCPMMP1_combinedColorLUT.txt`. You can then view these with FreeSurfer (from `SUBJECTS_DIR/fsaverage/mri` directory:
```
$ freeview -v HCPMMP1+aseg.mgz:colormap=lut:lut=HCPMMP1ColorLUT.txt
```
Screenshot from 2020-05-08 12-30-43
```
freeview -v HCPMMP1_combined+aseg.mgz:colormap=lut:lut=HCPMMP1_combinedColorLUT.txt
```
Screenshot from 2020-05-08 12-42-29

Then in MNE you can do something like the following to extract label time courses:
```
mri_dir = op.join(mne.get_config('SUBJECTS_DIR'), 'fsaverage', 'mri')
labels = mne.read_labels_from_annot('fsaverage', 'HCPMMP1', 'both')
lut = mne.read_freesurfer_lut(op.join(mri_dir, 'HCPMMP1ColorLUT.txt'))
# limit to just the HCPMMP labels with this, or omit it to keep subcortical ones:
lut = {label.name: lut[label.name] for label in labels}
atlas = op.join(mri_dir, 'HCPMMP1+aseg.mgz')
label_tc = mne.extract_label_time_course(stc, (atlas, lut), fwd['src'])
```

poojaprabhu9:
I was trying to use the above code but unfortunately i could not read the stc file (i updated the mne using, pip install https://api.github.com/repos/mne-tools/mne-python/zipball/master) because of this error.
TypeError: vertices must be an instance of list, got <class 'numpy.ndarray'> instead
I remember this problem was solved (when i was using, pip install -U https://github.com/larsoner/mne-python/archive/atlas.zip)
Did i miss something?

larsoner:
> pip install https://api.github.com/repos/mne-tools/mne-python/zipball/master

This might not have done anything, try doing it again with `--upgrade` in there. If it still fails, upload the STC somewhere and I can see if I can load it

poojaprabhu9:
I tried upgrading by 
`pip --upgrade install https://api.github.com/repos/mne-tools/mne-python/zipball/master`
Still the error exist.

You can find the stc file [here](https://mab.to/wp1AuZW4D)
Please try` stc.in_label` and `stc.extract_time_courses`.

larsoner:
Can you upload the source space that you used? I can't try those functions without it

poojaprabhu9:
just run this line
```
src = mne.setup_volume_source_space(
                        'fsaverage', pos=5., subjects_dir=subjects_dir, sphere=(0,0,0,90),
                        verbose=True,sphere_units='mm')
```

larsoner:
Okay -- it's better to use the bem/brain to define the set of sources to keep rather than `sphere`

poojaprabhu9:
> Okay -- it's better to use the bem/brain to define the set of sources to keep rather than `sphere`

will try that for the further computations.

virvw:
@larsoner @poojaprabhu9 
This worked:
$ mri_aparc2aseg --s fsaverage --volmask --annot HCPMMP1
$ mri_aparc2aseg --s fsaverage --volmask --annot HCPMMP1_combined

I am stuck with the gist for several reasons independent of the code itself
@poojaprabhu9 if you manage to run that bit on the fsaverage and forward the outcome, I'll finish up...

virvw:
@larsoner @poojaprabhu9 

The gist was ran removing the initial option "sort" because not available in our version.
 
Note that the created file "FreeSurferColorLUT.txt" was found by the script "create_subj_volume_parcellation.sh" when placed in the same folder as the script and not in the "fsaverage/label/"  

Now I am encountering issue when running "create_subj_volume_parcellation.sh" since we are missing "wm.mgz" for fsaverage


-----------------
FreeSurfer/vv221713@is152869:/neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI$ bash create_subj_volume_parcellation.sh -L subject_list.txt -a HCPMMP1 -d HCPMMP1_parcellation -s YES -t YES

         >>>>         Current FreeSurfer subjects folder is /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI

reading colortable from annotation file...
colortable with 181 entries read (originally ./left.fsaverage164.label.gii)
reading colortable from annotation file...
colortable with 181 entries read (originally ./left.fsaverage164.label.gii)
reading colortable from annotation file...
colortable with 181 entries read (originally ./right.fsaverage164.label.gii)
reading colortable from annotation file...
colortable with 181 entries read (originally ./right.fsaverage164.label.gii)

         >>>>         PREPROCESSING fsaverage         <<<<
         >>>>         START TIME: Sun May 10 16:49:56 CEST 2020         <<<<
>>>>    Annotation files lh.fsaverage_HCPMMP1.annot and rh.fsaverage_HCPMMP1.annot already exist in fsaverage/label. Won't perform transformations
reading colortable from annotation file...
colortable with 182 entries read (originally HCPMMP1_parcellation/temp_1_0_1246/colortab_HCPMMP1_L)
reading colortable from annotation file...
colortable with 182 entries read (originally HCPMMP1_parcellation/temp_1_0_1246/colortab_HCPMMP1_R)
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
>> Creating subcortical aseg masks for subject fsaverage
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
fslmaths: error while loading shared libraries: libopenblas.so.0: cannot open shared object file: No such file or directory
computing statistics for each annotation in /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/label/lh.fsaverage_HCPMMP1.annot.
reading volume /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz...
mghRead(/neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz, -1): could not open file
mris_anatomical_stats: could not read input volume /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz
computing statistics for each annotation in /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/label/rh.fsaverage_HCPMMP1.annot.
reading volume /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz...
mghRead(/neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz, -1): could not open file
mris_anatomical_stats: could not read input volume /neurospin/meg/meg_tmp/2015_MiniManip_rsMEG/POLTI_Ignacio_RSretroTIME/MRI/fsaverage/mri/wm.mgz
(standard_in) 2: syntax error
sed: -e expression n°1, caractère 3: `,' inattendue
sed: -e expression n°1, caractère 3: `,' inattendue
sed: -e expression n°1, caractère 3: `,' inattendue
sed: -e expression n°1, caractère 3: `,' inattendue

         >>>>         fsaverage STARTED AT Sun May 10 16:49:56 CEST 2020, ENDED AT: Sun May 10 16:51:04 CEST 2020

larsoner:
> The gist was ran removing the initial option "sort" because not available in our version.

It will not work properly without that option. The regions will be mislabeled. It was added to `master` recently, so if you update it should be available.

> Note that the created file "FreeSurferColorLUT.txt" was found by the script

You should not run this at all anymore, just [the two steps mentioned in the comment above](https://github.com/mne-tools/mne-python/pull/7639#issuecomment-625907891):

1. `$ mri_aparc2aseg --s fsaverage --volmask --annot HCPMMP1`
2. `python hcpmmp_lut.py` from [here](https://gist.github.com/larsoner/8e664205cd8285ca7c46211403ad12ce)

This should create all the files for `fsaverage` that you need to be able to extract label time courses (i.e., run the Python code from [this comment](https://github.com/mne-tools/mne-python/pull/7756#issuecomment-625986409)).

poojaprabhu9:
> python hcpmmp_lut.py from here

There is something missing in the line 17 of hcpmmp_lut.py. is it so?

virvw:
@larsoner considering thet the sort option is set to FALSE in your script,
does that matter at all?

If it does then I have to wait unless someone can run your script because I
can't update.

@pooja due to this issue, please neglect the files I just sent you which
will be mislabeled...

Virginie van Wassenhove
*https://brainthemind.com/ <https://brainthemind.com/>*

larsoner:
> There is something missing in the line 17 of hcpmmp_lut.py. is it so?

That's the start of a multi-line string, looks okay to me. Did you try and and get some error?

> considering thet the sort option is set to FALSE in your script,
does that matter at all?

In `master` the default is `True` (and anything before `master` implicitly had `sort=True`), so yes unfortunately you need it.

Also, if you plan to use the `HCPMMP1_combined` version, you probably need to (for now at least) regenerate it using latest `master`, which you can do by deleting the `lh/rh.HCPMMP1_combined.annot` files from `fsaverage/label` the doing `mne.datasets.fetch_hcpmmp_parcellation()`. And if you already did the FreeSurfer `mri_aparc2aseg` command, redo it as well, and then re-run my `gist`. This also has to do with the `sort` option defauting to True, which causes problems when generating the low-res/combined parc...

poojaprabhu9:
> That's the start of a multi-line string, looks okay to me. Did you try and and get some error?
Everything is fine. It was a mistake from my side.

virvw:
Ok I think things are sorted out.... and it is now back in Pooja's hands...!


Virginie van Wassenhove
*https://brainthemind.com/ <https://brainthemind.com/>*

poojaprabhu9:
@larsoner @virvw 
It is working!!
Thanks a lot @larsoner

PR #7641, opened 2020-04-20
MRG, FIX: Fix bug when EDF has null bytes

larsoner:
Takes care of a bug where you could get during EDF reads:
```
E       ValueError: invalid literal for int() with base 10: '36096\x00\x00\x00'
```
Stripping the null bytes fixes the issue.


drammock:
@tayyabafaisal you can either wait 5-6 months for version 0.21 to be released, or you can switch to using the development version ("current master"), see instructions [here](https://mne.tools/dev/install/advanced.html#using-the-development-version-of-mne-python-latest-master)

drammock:
@larsoner @agramfort should this one get backported?

larsoner:
Yeah, I'll do it now

PR #7642, opened 2020-04-20
[FIX][WIP] Fix handling of complex matrices in DICS beamformer

britta-wstnr:
This PR does the following things:

1. Rename, move, and test the inversion function for positive semidefinite matrices, formerly `_sym_inv`, now `_pos_semidef_inv`. _Background:_ make the name and documentation clear as this function only handles pos. semidefinite matrices. Move to `utils.linalg.py` and add testing.

2. Fix the transpose of non-Hermitian matrices that can be complex in the case of DICS.

3. Adjust a testing value in `test_dics.py`.

Please double check the use of `conjugate()` to see whether it is really necessary everywhere. I checked everything with an example of the `sample` data set as well and had a quick check against FieldTrip for the non-weightnorm version.

@wmvanvliet once @larsoner is happy, it would be great to have the simulations run on this. Hopefully things will improve!

### DIFF
--- a/mne/beamformer/_compute_beamformer.py
+++ b/mne/beamformer/_compute_beamformer.py
@@ -339,10 +339,10 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
         if pick_ori in [None, 'vector'] and n_orient > 1:
             # Rescale each set of 3 filters
             W = W.reshape(-1, 3, W.shape[1])
-            noise_norm = np.sqrt(np.sum(W ** 2, axis=(1, 2), keepdims=True))
+            noise_norm = np.linalg.norm(W, axis=(1, 2), keepdims=True)
         else:
             # Rescale each filter separately
-            noise_norm = np.sqrt(np.sum(W ** 2, axis=1, keepdims=True))
+            noise_norm = np.linalg.norm(W, axis=1, keepdims=True)
 
         if weight_norm == 'nai':
             # Estimate noise level based on covariance matrix, taking the
@@ -396,12 +396,12 @@ def _compute_power(Cm, W, n_orient):
     source_power = np.zeros(n_sources)
     for k in range(n_sources):
         Wk = W[n_orient * k: n_orient * k + n_orient]
-        power = Wk.dot(Cm).dot(Wk.T)  # XXX why no conj on Wk?
+        power = Wk.dot(Cm).dot(Wk.conj().T)
 
         if n_orient > 1:  # Pool the orientations
-            source_power[k] = np.abs(power.trace())
+            source_power[k] = power.trace().real
         else:
-            source_power[k] = np.abs(power)
+            source_power[k] = power.real
 
     return source_power
 
--- a/mne/beamformer/_compute_beamformer.py
+++ b/mne/beamformer/_compute_beamformer.py
@@ -7,7 +7,6 @@
 # License: BSD (3-clause)
 
 from copy import deepcopy
-from contextlib import contextmanager
 
 import numpy as np
 from scipy import linalg
@@ -191,12 +190,6 @@ def _normalized_weights(Wk, Gk, Cm_inv_sq, reduce_rank, nn, sk):
     Wk[:] = Wk_max[:, np.newaxis]
 
 
-@contextmanager
-def _noop_indentation_context():
-    """Context manager that does nothing."""
-    yield
-
-
 def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
                         reduce_rank, rank, inversion, nn, orient_std):
     """Compute a spatial beamformer filter (LCMV or DICS).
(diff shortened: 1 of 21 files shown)


mne/utils/linalg.py
```diff
@@ -156,3 +156,19 @@ def sqrtm_sym(A, rcond=1e-7, inv=False):
         use_s = s
     a = np.matmul(u * use_s[..., np.newaxis, :], u.swapaxes(-2, -1))
     return a, s
+
```
britta-wstnr, line 159: @larsoner you asked me to also check `sqrtm_sym` (above). Is it for Hermitian matrices? If so, it should be fine?

britta-wstnr:
It seems like the Travis failure relates to `numpy 1.13` where it fails in `numpy.linalg.pinv`:
`The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

Here's the link to the line: https://github.com/numpy/numpy/blob/maintenance/1.13.x/numpy/linalg/linalg.py#L1652

@larsoner can we handle this from within our `test_linalg.py`? Or am I understanding the error wrong?

larsoner:
We could make it work with older versions, but I'll push an easier fix that should still be fine :)

britta-wstnr:
@larsoner can you point to the failing test? For me all Travis output except the latest one appears as "canceled" so I can't see where it failed.

britta-wstnr:
I did not see any local failures before I pushed my commits

larsoner:
> can you point to the failing test? For me all Travis output except the latest one appears as "canceled" so I can't see where it failed.

They all pass currently because I removed a `.conj` you had put in (see below)

> I did not see any local failures before I pushed my commits

I think your tests were incomplete. Once I added more complete tests, the `.conj()` was definitely needed both for equiv with `np.linalg.pinv(..., hermitian=True)` and also for `assert_allclose(np.dot(mat_inv, mat), np.eye(3))`. So that one I'm pretty sure needs to be there.

With that in place (and maybe before?), when I did the `mne/beamformer/tests/test_dics.py` there was a failure on the (now cleaner as) `assert_allclose(noise_power.data, 1.)` line [here](https://github.com/mne-tools/mne-python/pull/7642/files#diff-d5ec8a3903870d83cb763f9ae005b798R328). To get that to pass, I had to remove the `.conj()` in [this line](https://github.com/mne-tools/mne-python/pull/7642/files#diff-530ba070f2fd62733fd85b24b9bf8c39R399), which does otherwise seem like it should be there conceptually.

britta-wstnr:
I am not sure I understand the test. I assume it wants to test the unit-gain for the noise? It is called `noise_csd` but passed to the beamformer as the signal, no?

wmvanvliet:
Hrm, very strange this `Wk.dot(Cm).dot(Wk.T)` line without the `conj()`. The result is close but note completely a real value. Need to trace this line by line to see what's going on.

mne/utils/linalg.py
larsoner: it's only for real matrices so it should be fine

wmvanvliet:
In the `'unit-noise-gain'` case, `Wk.dot(Wk.conj().T)` should be 1. But it isn't. Could it be that we compute the weight norm wrong? That would explain some simulation results!

larsoner:
> The result is close but not completely a real value. Need to trace this line by line to see what's going on.

At some point (I think with the `.conj()` in there?) the lines that are currently `np.abs(np.trace(...))` [here](https://github.com/mne-tools/mne-python/blob/master/mne/beamformer/_compute_beamformer.py#L415-L418) could just be changed to `np.trace(...).real`, because the imaginary part was effectively/numerically zero (i.e., 1e-14 below the real part).

larsoner:
> Could it be that we compute the weight norm wrong? That would explain some simulation results!

It would explain the test failure when the presumably correct `.conj()` is put back in -- definitely worth investigating (but outside my expertise, so I call "not it")!

britta-wstnr:
> In the `'unit-noise-gain'` case, `Wk.dot(Wk.conj().T)` should be 1. But it isn't. Could it be that we compute the weight norm wrong? That would explain some simulation results!

But the output of the beamformer is `Wk.dot(Cm).dot(Wk.T)`. The CSD matrix is in there, as well.

Well the weight norm is fine for the LCMV. I think DICS follows the same path?

larsoner:
> Well the weight norm is fine for the LCMV. I think DICS follows the same path?

If there is some `.conj` missing, it wouldn't matter for real-valued LCMV but would for DICS...

britta-wstnr:
@larsoner that is true.

wmvanvliet:
> But the output of the beamformer is Wk.dot(Cm).dot(Wk.T). The CSD matrix is in there, as well.

Yes, but in the case of `'unit-noise-gain'` the "noise" CSD is a unit matrix, hence you can leave it out for quick testing. It follows the same code path for LCMV, but that is likely wrong. Because we compute the weight normalization like this:

`W_norm = W * (1 / np.sqrt(np.sum(W ** 2, axis=1, keepdims=True)))`

This is correct for real values, since this will mean `W_norm.dot(W_norm.T) == 1`, but not for complex values!

larsoner:
>  np.sqrt(np.sum(W ** 2, axis=1, keepdims=True)))

This could just be `np.linalg.norm(W, axis=1, keepdims=True)`, then we should be good? Someone try it!

britta-wstnr:
I can have a look

wmvanvliet:
> np.linalg.norm(W, axis=1, keepdims=True)

Exactly the solution I just arrived at, that works!

britta-wstnr:
I'll do it then I can also add changes to the log @larsoner

larsoner:
Hopefully that allows restoring the `.conj` where it should be, and also allows us to improve some tolerances. But even if it doesn't do the latter, we should be good to go once you've tested it a bit @britta-wstnr

britta-wstnr:
Will be on it and report back!

wmvanvliet:
> Hopefully that allows restoring the .conj where it should be

It should be restored, otherwise it will be in error

larsoner:
I'd also like to put in `.real`s [here](https://github.com/mne-tools/mne-python/blob/master/mne/beamformer/_compute_beamformer.py#L415-L418) -- okay with you @wmvanvliet ? In principle the imaginary part should just be round-off / numerical error I think...

wmvanvliet:
Yes, we should return a plain float array, not a complex one

larsoner:
CIs refuse to trigger, will run tests locally and merge if they are green

larsoner:
@wmvanvliet should we wait for your simulation results to check these, or go ahead and merge? Mathematically it seems to make sense so I'm tempted just to merge here

wmvanvliet:
Simulation results are in. Oddly, the fixes in this PR had little effect on the results. Something is still wrong with the weight normalization:

dics_simulation

In the figure above, each dot represents a combination of parameters. The location of the dot indicates two performance metrics: distance to the true simulated source (lower is better) and a measure of how focal the source estimate is around the source (higher is better). I've limited the parameter combinations in this plot to only show gradiometers and no whitening.

We see that for all cases in which `weight_norm='unit-noise-gain'` (brown points) the performance is way worse than for all cases in which `normalize_fwd=True`.

britta-wstnr:
@wmvanvliet Does the selection you plotted include both complex and real filters? Compared to our full simulations from last time, the full simulation from this time (not shown here) seems to get rid of a lot of the fuzziness of our complex filter results - the results for that were all over the place.
I agree however, that the weight norm still performs badly. Let's check if this is for both real and complex filters - if it is, the conjugate transpose is not the reason for what we see here and there must be some additional problem.

mne/beamformer/_compute_beamformer.py
```diff
@@ -396,12 +396,12 @@ def _compute_power(Cm, W, n_orient):
     source_power = np.zeros(n_sources)
     for k in range(n_sources):
         Wk = W[n_orient * k: n_orient * k + n_orient]
-        power = Wk.dot(Cm).dot(Wk.T)  # XXX why no conj on Wk?
```
britta-wstnr: @wmvanvliet I am confused, I corrected all of these before?

wmvanvliet: sorry, git mistake on my end.

wmvanvliet:
> Does the selection you plotted include both complex and real filters?

It does. I agree with your assessment that in the full simulation, this PR gets rid of a lot of the problems with complex filters. However, it does not address the weight norm issue.

britta-wstnr:
I will do some checks on the weightnorm and report back.

larsoner:
Does this PR make weight normalization performance worse than master? If not then my vote is to merge this one and bake another PR for weight normalization fixes.

britta-wstnr:
I don't think there is a change except the fixes for complex values. So +1 for merging.

wmvanvliet:
Yeah, this PR definitely makes things better

larsoner:
Thanks @britta-wstnr @wmvanvliet ! backported.

PR #7643, opened 2020-04-20
DOC: better permutation cluster docstring

drammock:
docstring improvement spurred by [this question](https://mail.nmr.mgh.harvard.edu/pipermail/mne_analysis/2020-April/006756.html) on the MNE-Analyze mailing list.


larsoner:
Thanks @drammock

PR #7647, opened 2020-04-21
MRG, ENH: Add label extraction for vector source estimates

larsoner:
Helpful when working in free or even loose orientation because you can extract the vector source activations, and then take the norm (or do PCA, a Hotelling test, etc.) rather than having to take the norm or just the normal component.

Eventually I see this as being especially useful with volumetric source models with free orientation, so it lays the groundwork for #6155 to be useful.


larsoner:
https://19528-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_label_source_activations.html#using-vector-solutions

PR #7653, opened 2020-04-21
MRG, FIX: Use nearest in grid in setup_volume_source_space

larsoner:
The old code for subselecting volumes based on `volume_label` used `_compute_nearest`, which is fine if you use *the entire set of volume grid voxels* and see if what you're closest to matches the desired ID.

However, it used `_compute_nearest` with the *voxels matching the given ID*, and then checked to make sure that the source space `rr` were within some distance of these nearest neighbors based on the voxel spacing. This is problematic for two reasons:

1. Non-uniform `zooms` could break this, but in practice they are probably always the same. More importantly...
2. Checking the distance from the source space point its the nearest voxel center effectively forms a sphere in 3D space around each voxel center. Geometrically, at the boundaries of the sub-volume this forms a "bumpy surface" of inclusion composed of spheres when what you should have is a surface composed entirely of regular-grid cubes. Thus there are some vertices/locations that will belong to more than one ROI when using `volume_labels`, because they sit within the `dist`-based neighborhood/sphere of two different voxels (one having the volume ID of interest, and the other not).

Our old code used a maximal distance of `linalg.norm(vox_mri_t['trans'][:3, :3].sum(0) / 2.)`. Assuming unit length (WLOG) this forms a radius of 0.866 around each voxel center. The maximum error would thus be that minus 0.5 -- for a point directly in the middle of the line between two neighboring voxels -- so we could have been off by a factor of `0.366 * pos`, I think. Hence the line in `latest.inc` about over-extending in a worse case to 37% times pos.

This PR adds a test that fails on `master` and passes on this PR that the overlap between two neighboring regions is zero. To run it on `master` you'd need to run it with the commented-out `nuse`s, and you should see an overlap of 0.47. It's pretty bad, but I intentionally selected two complex surfaces with a high surface-to-volume ratio to exacerbate the problem, so hopefully it's a worst case...

It uses a procedure that should both be both faster and more robust, namely:

1. take our source `rr` and transforming to voxel i/j/k, which we know are on a regular grid
2. `np.round`
3. make sure they are actually in the volume
4. use them to index into our volume
5. check whether or not you get value of interest

It also will allow us to easily iterate over volume IDs for volumetric labeling, which is how I originally stumbled onto this issue, doing it this way was not equivalent...

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -44,6 +44,8 @@ Changelog
 Bug
 ~~~
 
+- Fix bug with :func:`mne.setup_volume_source_space` when ``volume_label`` was supplied where voxels slightly (in a worst case, about 37% times ``pos`` in distance) outside the voxel-grid-based bounds of regions were errantly included, by `Eric Larson`_
+
 - Fix bug with :func:`mne.minimum_norm.compute_source_psd_epochs` and :func:`mne.minimum_norm.source_band_induced_power` raised errors when ``method='eLORETA'`` by `Eric Larson`_
 
 - Fix handling of NaN when using TFCE in clustering functions such as :func:`mne.stats.spatio_temporal_cluster_1samp_test` by `Eric Larson`_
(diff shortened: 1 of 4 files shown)


larsoner:
backported

agramfort:
thanks !

>

PR #7021, opened 2019-11-05
[WIP] Add CSD whitening to DICS beamformer

wmvanvliet:
Working on adding whitening support to the DICS beamformer, so we can mix different channel types.

Todo:
 - [x] Implement whitening of CSD and Leadfield
 - [x] Modify DICS example to use both grads and mags
 - [ ] Do proper check to see if data and noise CSDs are compatible (working on part of this in #7029)
 - [ ] Add new unit tests for DICS whitening
 - [x] Have only one whitening matrix for all frequencies
 - [ ] Update What's New

PR #7656, opened 2020-04-22
MRG, BUG: Refactor beamformer code

larsoner:
To be fixed with @britta-wstnr

- [x] Remove specialized `_normalized_weights` path
- [x] Fix bug with `_reg_pinv` not maintaining Hermitian symmetry
- [x] Make ori picking step independent from filter calc
- [x] Fix max power ori tests (and FT equiv)
- [x] Reorganize according to the order @britta-wstnr prefers / the formulas suggest
- [x] Update testing data to have signed data from FieldTrip
- [x] Update FT equiv test to use sign flipping
- [x] Figure out the right time to pick the lead field for `'pick_ori='max-power'` and `pick_ori='normal'`
- [x] Deprecate `tf_lcmv`
- [x] Figure out the right thing to do with `'unit-noise-gain'` norm / rotation invariance
- [x] Check that everything works with DICS and LCMV in simulations
- [x] Add whitening for DICS
- [x] Add tests for unit-noise-gain properties for LCMV and DICS variants

Closes #7021


britta-wstnr:
@larsoner The test looks fine to me, thanks for putting it together! The vector case will be a whole other story I fear but let's discuss this later - let's try and get the scalar version green for this one first!

britta-wstnr:
So the test fails (at least) for the two filters ('unit-noise-gain' and 'unit-gain' filters) not relying on the same lead field if the orientation is picked.
We are not storing the updated lead field in the `filters` object - maybe we should? Furthermore, it would also make sense to store the selected orienation in that case (`max_power_ori` in our code).

larsoner:
Can the orientation picking be done last, after all other computations? If so, then maybe we shouldn't even bother doing it during filter creation but rather during `apply_lcmv`?

britta-wstnr:
No, it should ideally be done before the filter computation and then we pick the lead field and recompute the filter.

larsoner:
Why even bother storing the leadfield in the `filters` object?

britta-wstnr:
Well in this case it would come in handy for the test. I do not know how many use-cases there are, one argument for it is that we _change_ the lead field if we do orientation selection. The plotting of the orientation itself is sometimes done, to get an overview of which orientations the filter uses, so that would in any case be useful (and would give the actually used lead field as well, if needed).

larsoner:
+1 for storing the selected orientations (surprised we didn't already do this?). Whether you store the entire leadfield or the selected one is up to you. The selected one is just the `np.dot` in the direction of the orientation selection for each vertex, right? I guess you have more flexibility if you store the entire leadfield and let people do the dot when they want it ... ?

BTW would it be possible at some point to provide the "variance explained" like we do for minimum norm and mixed norm solvers? Probably not for this PR but a follow-up one, it would be nice to have.

mne/beamformer/_compute_beamformer.py
```diff
@@ -253,92 +199,108 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
...
+    if pick_ori == 'max-power':
+        # Compute the power
+        if inversion == 'matrix':
+            # Compute power by applying the spatial filters to the cov matrix
+            mult = Wk
+        else:
+            assert inversion == 'single'
+            # First make the filters unit gain, then apply them to the
+            # cov matrix to compute power.
+            mult = Wk / np.linalg.norm(Wk, axis=2, keepdims=True)
```
britta-wstnr, line 228: I don't think this works here, the code snippet comes from further down, where `Wk` already refers to a later step in filter calculation. I will see if I can fix it!

mne/beamformer/_compute_beamformer.py
```diff
@@ -253,92 +199,108 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
...
-            # Reapply source covariance after inversion
-            norm *= sk[:, :, np.newaxis]
-            norm *= sk[:, np.newaxis, :]
-        else:  # n_orient == 1
-            assert Ck.shape[1:] == (1, 1)
-            # Fixed source orientation
+    #
+    # 2. Compute max power orientation
+    #
+    if pick_ori == 'max-power':
```
britta-wstnr, line 219: The unit-noise-gain filter also should have a different formula for the orientation computation - I will put the full formulas here and then we can see if we can reduce them / make them faster again later.

britta-wstnr:
**What's new / rationale:**
- We strictly follow the formulas from Sekihara & Nagarajan 2008 for the orientation selection of the unit-noise-gain beamformer and pick the leadfield.
- The numerator and denominator of the beamformer are computed in one place and named (no step-by-step changes of `Wk` into different products anymore).
- The weight normalization is done by normalizing the unit-gain beamformer, rationale is stated in comments.
- General documentation of where code relates directly to formulas.

**What works now:**
- The new unit-noise-gain test works now for scalar beamformer with orientation selection.
- external tests are passing again, even at higher precision
- the orientation selection for unit-noise-gain beamformers is fixed.

**What does not work yet:**
- this pass added 5 more failing tests to the already 7 failing tests (fix in `_sym_inv`)
- the single inversion is not implpemented in the orienation selection yet
- the correction regarding the sign flipping (orienation arbitrary by 180 deg.) will need to be brought back - where it is now it makes external tests fail horrendeously (correlation of 0.5 instead of 1.0)

So there is still some stuff to do, but I find the code way clearer now and hope it will be easier to maintain in the future!

mne/beamformer/_lcmv.py
```diff
@@ -157,14 +157,19 @@ def make_lcmv(info, forward, data_cov, reg=0.05, noise_cov=None, label=None,
 
     # Whiten the data covariance
     Cm = np.dot(whitener, np.dot(Cm, whitener.T))
+    # Restore to positive semi-definite, as
+    # (negative eigenvalues are errant / due to massive scaling differences)
+    s, u = np.linalg.eigh(Cm)
+    Cm = np.dot(u * np.abs(s), u.T.conj())
```
wmvanvliet, line 163: Do we need this for DICS too?

larsoner: Not sure. I did make some changes to `_reg_inv` that might take care of it

larsoner:
@wmvanvliet I just pushed a fix that refactors a bit more, now the `inversion='single'` is in a single code path/function, at least as far as `_compute_beamformer` is concerned.

@britta-wstnr I:

- adjusted tolerances so that all tests should pass
- fixed the `normal` unit-noise-gain formula test by doing the normal picking at the same time as the max-power picking -- seems more principled (if you're going to pick a single ori, it should be done in the same place both in normal and in max-power mode it seems). This will make it so that computing a vector beamformer and taking the `stc.normal()` will not be the same as computing a beamformer with `pick_ori='vector'`, but I guess this is desired behavior (right?). If it's not, then we need to move the normal orientation picking to the very end of the beamformer calculation, then it's not clear why we wouldn't also do this for max power ori...

To me it seems like the one remaining issue (for LCMV at least) is the `pick_ori='vector'` case for the unit-noise-gain. I've updated the code comment to make it clear why I think they differ currently, and which of three choices we have to make them equivalent. I'll call it out in-line with a comment...

mne/beamformer/_compute_beamformer.py
```diff
@@ -253,96 +228,94 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
...
+
+        # Compute the lead field for the optimal orientation,
+        # and adjust numer/denom
+        Gk = np.matmul(Gk, max_power_ori[..., np.newaxis])
+        n_orient = 1
+    else:
+        max_power_ori = None
+        if pick_ori == 'normal':
+            Gk = Gk[..., 2:3]
+            n_orient = 1
```
larsoner, line 285: Moved `pick_ori='normal'` up here, now it uses gain matrix subselection just like `pick_ori='max-power'`, and from then on the formulas proceed like scalar beamformers.

Again, don't know if it's correct, but it's more consistent...

mne/beamformer/_compute_beamformer.py
```diff
@@ -253,96 +228,94 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
...
+    W = np.matmul(bf_denom_inv, bf_numer)
+    assert W.shape == (n_sources, n_orient, n_channels)
+    del bf_numer, bf_denom, bf_denom_inv, sk
+
+    #
+    # 4. Re-scale filter weights according to the selected weight_norm
+    #
+
+    # Weight normalization is done by computing:
+    # W_ung = inv(sqrt(W_ug @ W_ug.T)) @ W_ug
```
larsoner, line 311: What will matter here is what sizes these things actually are. In other words, what is the size of W_ug, is it for a single ori `(1, n_channels)` or all three oris `(3, n_channels)`, and hence whether:

1. the `sqrt` is a plain square root
2. a matrix square root (only possible in the 3x3 case), or even
3. square root of the diagonal entries of the matrix only (unlikely but possible)

Currently we do a fourth thing, which is just to take the norm over the `(3, n_channels)` array, square root this singleton, and invert it, which doesn't act like any of the three above so is probably wrong.

Note that this only matters when `pick_ori='vector'` (or None) because if `pick_ori='normal'` or `pick_ori='max_power'`, we are in the 1-direction-per-source / `n_orient=1` case by this point in the code now.

mne/beamformer/tests/test_lcmv.py
```diff
@@ -697,6 +697,95 @@ def test_lcmv_ctf_comp():
...
+    #       of one -- depends on what the units are above, if the sqrt is a
+    #       matrix or single square root, etc.:
+    #
+    # - Here we do a proper matrix square root + inversion
+    # - in _compute_beamformer we just take the norm over the
+    #   (n_orient, n_channels) array for each source, then invert the
+    #   single value for that source
+    # - It's possible the paper computes the orientation for each output
+    #   direction instead, which would be a third way (but this is probably not
+    #   rotation invariant and sensitive to nulls, so seems unlikely?)
```
larsoner, line 772: Code comment restating what I said above about the options (though I think the list here is less complete than my inline GitHub comment above)

mne/beamformer/_compute_beamformer.py
larsoner: I guess this issue still remains to be fixed, too...

larsoner: Actually it looks like you've done this below in the `else` block, right? If so please "resolve" this @britta-wstnr

mne/beamformer/_compute_beamformer.py
larsoner: For now I've just implemented the direct formula with a Frobenius norm for each source point as this is both rotation invariant and had the lowest bias in the tests, but again I don't know if it's correct. I also implemented the non-rotation-invariant one below based on the formula you gave me @britta-wstnr, you can try it by uncommenting it (and commenting out the "direct way" code) if you want

mne/beamformer/_compute_beamformer.py
britta-wstnr: I just kept `pick_ori='normal'` along where it used to be - I am not very familiar with it as I always compute beamformers in volume source space so this never really comes up. It is not an option that is seen very often in the wild with beamformer either I think - not saying it's not useful, but not seen often. If we are not sure it is correct, is it worth keeping? CC @wmvanvliet

mne/beamformer/_compute_beamformer.py
britta-wstnr: So far, the case where the matrix can be 3 x 3 has been wrong, I believe (vector case). I don't think it is a simple matrix square root in that case. 
I will check the different versions and re-check the formulas.

mne/beamformer/_compute_beamformer.py
larsoner: Actually come to think of it I think this is a reasonable thing to do. You should get the same result from this as you would from passing a fixed-orientation forward. If you want the normal component from a vector free-orientation solution, you should use `stc_free.normal()` and this will differ from the fixed-orientation-forward result in the minimum-norm case, too. So I'll mark this one as resolved

mne/beamformer/_compute_beamformer.py
```diff
@@ -113,67 +113,33 @@ def _reduce_leadfield_rank(G):
...
-    # Compute the filter in the orientation of max power
-    Wk_max = np.matmul(max_power_ori[:, np.newaxis], Wk)[:, 0]
-    Gk_max = np.matmul(Gk, max_power_ori[:, :, np.newaxis])
-    denom = np.matmul(Gk_max.transpose(0, 2, 1),
-                      np.matmul(Cm_inv_sq[np.newaxis], Gk_max))[:, 0]
-    np.sqrt(denom, out=denom)
-    Wk_max /= denom
-    # All three entries get the same value from this operation
-    Wk[:] = Wk_max[:, np.newaxis]
+def _sym_inv_sm(x, reduce_rank, inversion, sk):
```
britta-wstnr, line 116: Aaah, very nice to have this take care of the inversion scheme now!

mne/beamformer/_lcmv.py
```diff
@@ -589,6 +576,8 @@ def tf_lcmv(epochs, forward, noise_covs, tmin, tmax, tstep, win_lengths,
     .. [2] Sekihara & Nagarajan. Adaptive spatial filters for electromagnetic
            brain imaging (2008) Springer Science & Business Media
     """
+    # TODO this has an entirely different code path, need to decide if we
+    # can unify it, or deprecate this, etc.
```
britta-wstnr, line 580: I am pro deprecation. There are better methods now (that we have better computational power etc) than chopping data and computing covariance matrices on those snippets ... e.g., using Hilbert-data per frequency band and getting good time resolution, or using DICS with many frequencies and getting good frequency resolution. 
If someone still wants to do it, I feel it would be straight-forward to write a loop over time windows and frequency bands and do covariance computation and beamformer in that loop.

mne/beamformer/tests/test_lcmv.py
```diff
@@ -697,6 +697,74 @@ def test_lcmv_ctf_comp():
         make_lcmv(info_comp, fwd, data_cov)
 
 
+@testing.requires_testing_data
+@pytest.mark.parametrize('pick_ori', ['max-power', 'normal', 'vector'])
+@pytest.mark.parametrize('reg', (0.05, 0.))
```
britta-wstnr, line 702: The formula I implemented here does definitely not hold for vector beamformers ....

mne/beamformer/_lcmv.py
larsoner: Okay I'll deprecate then

mne/beamformer/tests/test_lcmv.py
larsoner: I'll just kill `vector` for now then

mne/beamformer/tests/test_lcmv.py
```diff
@@ -697,6 +698,72 @@ def test_lcmv_ctf_comp():
...
+        assert pick_ori == 'vector'
+        n_orient = 3
+    assert fwd_sol.shape == (498, 102, n_orient)
+
+    # compute the inverse of the covariance matrix and square it
+    Cm_inv, _, _ = _reg_pinv(data_cov.data, reg, rank)
+
+    # compute the unit-noise-gain beamformer in pedestrian mode:
+    # formula: G.T @ Cm_inv / sqrt(G.T @ Cm_inv @ Cm_inv @ G)
+    # Sekihara & Nagarajan 2008, eq. 4.15
```
britta-wstnr, line 753: Do we want to keep this test now, @larsoner ? It now just redoes what the code does as well, no?

britta-wstnr: OTOH, it might make sense if we decide to solve `unit-noise-gain`vector version differently ...

mne/utils/linalg.py
```diff
@@ -148,7 +152,7 @@ def sqrtm_sym(A, rcond=1e-7, inv=False):
     """
```
britta-wstnr, line 152: `inv` is missing in documentation

mne/beamformer/tests/test_lcmv.py
```diff
@@ -698,72 +698,6 @@ def test_lcmv_ctf_comp():
...
-    forward = mne.pick_types_forward(forward, meg='mag')  # restrict to MAG
-    fwd_sol = forward['sol']['data'].T
-    n_sources, n_channels = forward['nsource'], forward['nchan']
-    fwd_sol = np.reshape(fwd_sol, (n_sources, 3, n_channels))
-    fwd_sol = fwd_sol.transpose(0, 2, 1)
-    if pick_ori == 'normal':
-        fwd_sol = fwd_sol[:, :, 2:3]
-        n_orient = 1
-    elif pick_ori == 'max-power':
-        fwd_sol = np.matmul(fwd_sol, filters['max_power_ori'][..., np.newaxis])
```
britta-wstnr: @larsoner I am referring to this for a test of the `max_power_ori` application in `_compute_beamformer`

mne/beamformer/_lcmv.py
```diff
@@ -198,14 +210,7 @@ def _apply_lcmv(data, filters, info, tmin, max_ori_out):
...
 
-        if filters['is_ssp']:
-            # check whether data and filter projs match
-            _check_proj_match(info, filters)
-            if filters['whitener'] is None:
-                M = np.dot(filters['proj'], M)
-
-        if filters['whitener'] is not None:
-            M = np.dot(filters['whitener'], M)
+        M = _proj_data(M, info['projs'], filters)
```
wmvanvliet, line 211: What do we do with `filters['whitener']`?

mne/beamformer/_lcmv.py
```diff
@@ -182,11 +170,23 @@ def make_lcmv(info, forward, data_cov, reg=0.05, noise_cov=None, label=None,
...
+        rank=rank_int, max_power_ori=max_power_ori)
 
     return filters
 
 
+def _proj_data(M, proj, filters):
+    if filters['is_ssp']:
+        # check whether data and filter projs match
+        _check_proj_match(proj, filters)
+        if filters['whitener'] is None:
```
larsoner, line 182: @wmvanvliet we use it here

`_proj_data` is perhaps not the best name, could rename to `_proj_whiten_data`

mne/beamformer/tests/test_lcmv.py
```diff
@@ -874,4 +911,72 @@ def test_lcmv_maxfiltered():
...
+            # only the diagonal is correct
+            got = got.reshape(n_sources, -1)[:, ::n_orient + 1]
+            desired = np.ones_like(got)
+        assert_allclose(got, desired, atol=1e-7, err_msg='w @ w.conj().T')
+        # Check that the result here is a constant times a diagonal matrix
+        # (for Sekihara) or a rotated version of that (for sqrtm)
+        # got = w @ use_G.swapaxes(-2, -1)
+        # s = np.linalg.svd(got, compute_uv=False)
+        # theta = np.repeat(s[..., :1], s.shape[-1], axis=-1)
+        # assert_allclose(s, theta, rtol=1e-1, err_msg='G.T @ w = θI')
```
britta-wstnr, line 982: @larsoner  This is not right I think - it is easier than you tested for.
`w @ G.T` should not have a constant on the diagonal, it just should be a diagonal matrix. This holds for me in the test above:
`np.allclose(np.diag(np.diag(got[0])), got[0]) `

I just did not know how to do this on our 3-dimensional matrix ....

larsoner: Last I checked it, it wasn't diagonal I don't think. But I can look early next week. If you want to try something, feel free to just make it a simple loop and I can vectorize the check later

larsoner:
@agramfort do you want to see if the changes make sense to you based on the doc updates, etc.?

larsoner:
Thanks @britta-wstnr @wmvanvliet !

agramfort:
Congrats 🎉 🍾🍻

PR #7658, opened 2020-04-22
FIX: Fix RAS coordinates in watershed bem surfaces

### LINKED ISSUE #7570: mne.bem.make_watershed_bem() Coordinate System Error
#### Describe the bug
Incorrect extraction of volume info by the _extract_volume_info method in make_watershed_bem() in bem.py. If you look at the extracted BEM surfaces using the ancient mne freeview_bem_surfaces method from terminal that calls Freeview, you will see that the BEM surfaces have flipped yz coordinates. This is due to incorrect use of the header['Mdc'].T matrix. 

#### Steps to reproduce
Sorry I can't provide a snippet since this bug requires examining the BEM surfaces outside MNE-Python. But if you just run mne.bem.make_watershed_bem() on any Freesurfer T1 recon, and then use freeview to check the surfaces. For instance, you can use the MNE-C method mne freeview_bem_surfaces -s subject to check (or you can manually do it by freeview -f surfacefile) in the watershed folder, you will see the yz dimensions are flipped. 

[MWE]: https://en.wikipedia.org/wiki/Minimal_Working_Example
[built-in datasets]: https://mne.tools/dev/overview/datasets_index.html
[sample]: https://mne.tools/dev/overview/datasets_index.html#sample

#### Expected results
Well.. the BEM surface should have the same volume-geometry as the T1.mgz from Freesurfer recon. It is not in version 0.20.0. 

#### Actual results
It's flipped in yz dimensions. 

#### Additional information
Ok here's some historical information: this is an inherited bug since long time ago when initially there was the bug on the line "write_surface(s, rr, tris, volume_info=volume_info)" that should have been "write_surface(surf_ws_out, rr, tris, volume_info=volume_info)". I don't know at which version this is fixed but it is correct in 0.20.0. However the _extract_volume_info method is wrong. header['Mdc'].T is not [xras, yras, bras]. Instead, it is [rxyz, axyz, sxyz]. So you need to transpose the matrix before assigning the ras direction vectors in vol_info. 

Here is an easy fix: change line 1229 from 
vol_info['xras'], vol_info['yras'], vol_info['zras'] = header['Mdc'].T
to 
vol_info['xras'], vol_info['yras'], vol_info['zras'] = list(zip(*header['Mdc'].T))
and you will fix it. 

In theory if everyone's using freesurfer correctly then the outputted surfaces will already be in the right Freesurfer coordinates. So just like you don't update volume_geometry in mne.bem.make_flash_bem(), you don't need to update watershed surfaces in mne.bem.make_watershed_bem() either. I could see the initial thought was probably to accommodate T1.mgz beyond the default output volume from recon-all pipeline, so I guess it makes more sense to fix this transpose error than to remove the volume-geometry update codes all together. Just as a further suggestion, there is a useful freesurfer command mris_convert --vol-geom volume function that handles all these different structural scan type issues. So you might just want to call mris_convert in a subprocess.run() call that will save you a lot of trouble and avoid having to define your own _extract_volume_info() function, that is dependent on how nibable reads header files. Cheers.

yh-luo:
#### Reference issue
Fixes #7570.

In v20, the RAS coordinates of the surfaces created by `mne.bem.make_water_shed_bem` were wrong (refer to #7570 for details). This PR fixes the problem and adds tests to `mne/commands/test_commands/test_watershed_bem` to check the RAS coordinates of the watershed BEM surfaces.

#### Additional information
The bug was found thanks to @mh105, replicated by @larsoner.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -82,6 +82,8 @@ Bug
 
 - Fix bug with :func:`mne.beamformer.make_dics` where complex conjugates were not applied properly by `Britta Westner`_ and `Eric Larson`_
 
+- Fix bug with :func:`mne.bem.make_watershed_bem` where the RAS coordinates of watershed bem surfaces were not updated correctly from the volume file by `Yu-Han Luo`_
+
 API
 ~~~
 
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -32,7 +32,7 @@ from .surface import (read_surface, write_surface, complete_surface_info,
 from .transforms import _ensure_trans, apply_trans, Transform
 from .utils import (verbose, logger, run_subprocess, get_subjects_dir, warn,
                     _pl, _validate_type, _TempDir, _check_freesurfer_home,
-                    _check_fname)
+                    _check_fname, has_nibabel)
 from .fixes import einsum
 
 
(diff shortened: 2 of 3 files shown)


larsoner:
backported

PR #7659, opened 2020-04-22
MRG: _Brain add_annotation

GuillaumeFavelier:
This PR adds `add_annotation` in PyVista-compatible `_Brain`. This is still a work in progress.

- [x] Fix colormap table
- [x] Fix color interpolation
- [x] Fix region borders
- [x] Add `get_brain_class()`
- [x] Refactor `mesh_edges`
- [x] Add support for input `cortex` parameter
- [x] Check compatibility with VTK9
- [x] Use `bone_r` colormap instead of `bone`
- [x] Improve coverage

image


GuillaumeFavelier:
It's slightly better. From `examples/visualization/plot_parcellation.py`:

image

agramfort:
can you show what is the target/objective rendered output with mayavi?

GuillaumeFavelier:
PyVista | Mayavi
--|--
image | image

GuillaumeFavelier:
PyVista | Mayavi
--|--
image | image

GuillaumeFavelier:
The range does not seem to be *exactly* the same leading to some differences in the resulting values for the regions

larsoner:
Could it just be a lighting difference?

GuillaumeFavelier:
I investigated further and it's causd by the casting of the `ids` to `np.float`. To obtain the same result, only `np.uint8` is allowed for the `LookupTable`. I'll push a commit to patch it.

larsoner:
Actually no. There is a region that is red on the left and green on the right (sixth from the left in the line of green regions on the temporal love). Probably still some bug

larsoner:
Ahh okay

GuillaumeFavelier:
I'm not sure about the naked use of `_Brain` here but I don't see any other convenient way to achieve that:

https://github.com/mne-tools/mne-python/blob/49d200d6e81afd670e93f407ffcabb499b778aee/examples/visualization/plot_parcellation.py#L33-L36

larsoner:
One option would be a tiny wrapper `mne.viz.get_brain_class()` that either returns PySurfer's or our internal `_Brain`, then you just do:
```
Brain = mne.viz.get_brain_class()
brain = Brain(...)
```
`get_brain_class` basically just needs your conditional inside it.

Also we should probably at some point:

- make the class name `Brain` instead of `_Brain`
- make it public (`mne.viz.Brain`) with proper documentation, etc., and a clear preference for it just like we used to have for PySurfer

If it's this cycle (~5 mo timeframe) that we properly document everything, make it public, and prefer PyVista for everything, then we can probably even get rid of the `get_brain_class` shim above before release. But putting the shim in now is a lightweight solution for delaying the need to make those choices...

mne/viz/_brain/surface.py
```diff
@@ -159,3 +160,30 @@ def load_curvature(self):
...
+    a, b, c = faces.T
+    edges = sparse.coo_matrix((np.ones(nfaces), (a, b)),
+                              shape=(npoints, npoints))
+    edges = edges + sparse.coo_matrix((np.ones(nfaces), (b, c)),
+                                      shape=(npoints, npoints))
+    edges = edges + sparse.coo_matrix((np.ones(nfaces), (c, a)),
+                                      shape=(npoints, npoints))
+    edges = edges + edges.T
+    edges = edges.tocoo()
+    return edges
```
larsoner, line 189:
```
$ git grep "def mesh_edges"
mne/surface.py:def mesh_edges(tris):
```
Did you check to see if that one will work already? If so, then you shouldn't need this version, but rather import the other one with your `from ...surface import ...` line. There have been other duplicates like this in `_brain` before and I've been trying to remove them as I've found them. Let's try not to add more...

larsoner:
Comparing [the PR](https://19671-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py) to [master](https://mne.tools/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py), it looks like `low_contrast` as the cortex type is not working / ignored by PyVista.

GuillaumeFavelier:
[master example](https://mne.tools/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py) against [PR example](https://19707-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py) : 

master | PR
----|----
image | image

hoechenberger:
> [master example](https://mne.tools/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py) against [PR example](https://19707-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_parcellation.html#sphx-glr-auto-examples-visualization-plot-parcellation-py) :

Looks great

mne/viz/_brain/_brain.py
```diff
@@ -234,22 +241,44 @@ def __init__(self, subject_id, hemi, surf, title=None,
...
+                                          opacity=alpha), False),
+                            high_contrast=(dict(colormap="Greys",
+                                                vmin=-.1, vmax=1.3,
+                                                opacity=alpha), False),
+                            low_contrast=(dict(colormap="Greys",
+                                               vmin=-5, vmax=5,
+                                               opacity=alpha), False),
+                            bone=(dict(colormap="bone",
+                                       vmin=-.2, vmax=2,
+                                       opacity=alpha), True))
```
larsoner, line 279: Instead of dealing with any of this `reverse_lut` business, I would just do `bone_r` as the colormap here. Then you don't need any to pass any `reverse_lut` code.

larsoner:
Other than a minor comment and coverage, LGTM

larsoner:
Thanks @GuillaumeFavelier !

Issue #7661, opened 2020-04-23
ENH: Remove redundant tests on make_flash_bem


yh-luo:
#### Describe the problem

`mne.bem.make_flash_bem` was tested in both `mne/commands/test_commands.py` and `mne/tests/test_bem.py`. Since the function is time-consuming, merge them into `test_flash_bem` in `test_commands.py` would reduce testing time (~200 sec).

The flash bem surfaces in the testing dataset are different from surfaces computed by `mne/commands/test_commands/test_flash_bem` in that `test_flash_bem` uses both mef05.mgz (FLASH5) and mef30.mgz (FLASH30) to create the flash bem surfaces while the surfaces in testing data were created from only flash5.mgz (FLASH5). Is it necessary to assure the generated flash bem surfaces are the exactly same as those in the testing data?

The coordinates (`x` from test_flash_bem, `y` from testing data) are different, reported by pytest:
```
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatch: 100%
Max absolute difference: 1.82610321
Max relative difference: 43.68559136
x: array([[  -1.028   ,   -0.8233  ,  115.167198],
  [  72.226601,   -0.7632  ,   67.861504],
  [  21.6248  ,   83.526398,   66.532303],...
y: array([[  -1.3635  ,   -0.8711  ,  115.845596],
  [  72.295403,   -0.5315  ,   68.657402],
  [  20.939199,   83.773201,   67.367401],...
```

#### Describe your solution

Maybe testing the shape of surfaces like in `test_watershed_bem` is enough?

#### Describe possible alternatives

+ Keep both tests
+ Increase the tolerance (the tolerance would be large)

agramfort:
nothing is necessary as long as we have tests to ensure that we don't break
things

>

larsoner:
+1 for updating `mne-testing-data` with the correct surfaces and using `assert_allclose` with something like `atol=1e-5`

yh-luo:
> +1 for updating `mne-testing-data` with the correct surfaces and using `assert_allclose` with something like `atol=1e-5`

So `assert_close` with `kwargs = dict(rtol=1e-7, atol=1e-5)` passed. The generated surfaces are compared to the assumed correct surfaces (I copied the output surfaces from pytest temp folder).
[...]

PR #7663, opened 2020-04-23
MRG, ENH: Remove redundant tests on make_flash_bem

yh-luo:
#### Reference issue
Enhance tests as stated in #7661.

Merge `mne/tests/test_bem.py::test_make_flash_bem` to `mne/commands/test_commands.py::test_flash_bem` to reduce test time.
Testing data would be more compact (does not need `COR*` files) and efficient (use resampled `T1.mgz` to save time) using the current test.

#### Additional information
Testing dataset should be updated if this PR is accepted, or the test may fail.

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1786,8 +1786,6 @@ def make_flash_bem(subject, overwrite=False, show=True, subjects_dir=None,
     """
     from .viz.misc import plot_bem
 
-    is_test = os.environ.get('MNE_SKIP_FS_FLASH_CALL', False)
-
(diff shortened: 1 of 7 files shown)


larsoner:
@yh-luo you need to follow step 5 here

https://github.com/mne-tools/mne-testing-data#add-or-change-files-in-the-repo-for-use-with-mne

larsoner:
I'll push a quick commit to do it

larsoner:
This works on my system and it won't be tested by CIs, I'll go ahead and merge then rebase #7704, which should then test it. Thanks @yh-luo !

Issue #7654, opened 2020-04-22
Add option to only plot requested properties for plot_filter


rob-luke:
#### Describe the problem
I would like to just plot one aspect of a filter (e.g. just magnitude). Further I would like to specify the axes the data is plotted on.

#### Describe your solution
Add additional arguments to the function (perhaps... `magnitude`, `phase`, `time`) that change enable/disable plotting of each subplot (all default to True to maintain current functionality).

Add an axes parameter that can be passed in to plot over existing axes ([possibly discussed before](https://github.com/mne-tools/mne-python/blob/master/mne/viz/misc.py#L893)).

#### Describe possible alternatives
Leave as is.

larsoner:
Maybe instead of separate kwargs it can just be `plot=('magnitude', 'phase', 'time')` and the order and members of the iterable determine what is plotted, and `axes=None` if passed must match that length.

PR #7664, opened 2020-04-23
MRG: Add plot and axes option to plot_filter

rob-luke:
#### Reference issue
#7654

Add `plot` keyword to allow choice of filter properties to plot.  
Add `axes` keyword.

#### Additional information
Order of plots is not as specified by the user, not in fixed order as before

### DIFF
--- a/mne/viz/misc.py
+++ b/mne/viz/misc.py
@@ -769,7 +769,8 @@ _DEFAULT_ALIM = (-80, 10)
 
 def plot_filter(h, sfreq, freq=None, gain=None, title=None, color='#1f77b4',
                 flim=None, fscale='log', alim=_DEFAULT_ALIM, show=True,
-                compensate=False):
+                compensate=False, plot=('time', 'magnitude', 'delay'),
+                axes=None):
     """Plot properties of a filter.
 
     Parameters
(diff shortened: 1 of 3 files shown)


rob-luke:
@larsoner @agramfort could you please review. The CI fail seems unrelated

mne/viz/misc.py
```diff
@@ -827,13 +837,20 @@ def plot_filter(h, sfreq, freq=None, gain=None, title=None, color='#1f77b4',
     from scipy.signal import (
         freqz, group_delay, lfilter, filtfilt, sosfilt, sosfiltfilt)
     import matplotlib.pyplot as plt
+
     sfreq = float(sfreq)
     _check_option('fscale', fscale, ['log', 'linear'])
+    if isinstance(plot, str):
+        plot = [plot]
+    for xi, x in enumerate(plot):
+        _check_option('plot[%d]' % xi, x, ('magnitude', 'delay', 'time'))
```
larsoner, line 846: It's starting to seem like this should be a new:
```
_check_iterable_option('plot', plot, ('magintude', 'delay', 'time'))
```
Eventually we can make `_check_iterable_option` smart, like raising an error if it's not iterable for example. There is a risk to over-engineering these sorts of things, but it seems like you would have been able to use it half a dozen times or so, and I imagine there are a lot of other similar cases in our codebase, so seems reasonable to me. WDYT?

If you agree it's good, let's do it in another PR since this is already good to go

mne/viz/misc.py
```diff
@@ -809,6 +810,15 @@ def plot_filter(h, sfreq, freq=None, gain=None, title=None, color='#1f77b4',
           by squaring it.
 
         .. versionadded:: 0.18
+    plot : list | tuple | str
+        A list of the requested plots from `time`, `magnitude` and `delay`.
+        Default is to plot all three filter properties
+        ('time', 'magnitude', 'delay').
```
larsoner, line 816: ... actually since I need to find something to gripe about
```suggestion
        ('time', 'magnitude', 'delay').
        
        .. versionadded:: 0.21.0
```
Same for other param below

mne/viz/misc.py
rob-luke: Yeah sure. I will take a crack it implementing this over the next few days in a different PR.

agramfort:
thx @rob-luke

Issue #7652, opened 2020-04-21
get_montage and get_coordinates for Raw and Montage objects


adam2392:
#### Describe the problem
Once a user has set the montage, then getting the montage itself and data from it is not trivial.

Ideally, the users can get the actual xyz coordinates whether its an array, or dictionary via just a `get` function.

You have to do things like:
1. `ch_locs = raw.info['chs']['loc'][0:3]`
or
2. `ch_locs = [raw.info['dig'][idx]['r'][0:3] for idx in len(raw.info['dig'])]`

In 1., it took me a bit to end up figuring out that the 12 point locations in are not relevant for the most part if I need the xyz coordinates. And the xyz are stored in the first 3 indices. This is at least for iEEG. 

**Note:** In addition, idk if this is on purpose, but the raw.info['dig'] can be not in the same order as raw.info['ch_names'] especially if there was some reordering done when making `ch_pos` for the DigMontage. I don't think it affects internal code, but it further makes it non-trivial if you need access to the coordinates you want to work w/.

#### Describe your solution
An implementation at the Raw level for `get_montage()`, it simply returns the montage that is set for a Raw object, or returns an error message if no montage.

An implementation at the Montage level for `get_coordinates()`, which returns the xyz coordinates in it's 'm' form as a 3D array. 

Ideally it looks like this:

```
montage = raw.montage
ch_locs = montage.get_coordinates()
```

#### Describe possible alternatives
`get_coordinates` could also be at the Raw level?

#### Additional context
See related discussion: https://github.com/mne-tools/mne-bids/pull/393

Happy to submit a PR after discussion of if this is valid.

larsoner:
I would not do `get_montage`, a `montage` is just a way to set channel positions.

I would add a `inst.get_channel_positions()` that for example returned a `dict` of the channel name: pos mapping. It would really just loop over `info['chs']` to look at `loc` but indeed a user should not have to do this.

agramfort:
Get_montage that returns a dig montage that could be fed to set_montage is to me the most consistent. I would like to get all dig points etc

larsoner:
Okay, fair enough

drammock:
I'd also note that `get_ch_positions` (rather than `get_channel_positions`) is more consistent with other methods like `inst.get_ch_types()`.  So if you end up doing both `get_montage` and `get_xxx_positions` then I think `xxx` should be `ch`, not `channel`.

cbrnr:
[...]

PR #7667, opened 2020-04-23
[MRG] Adding get_montage for montage to BaseRaw objects

adam2392:
#### Reference issue
Fixes: #7652 

Adds `get_montage()` and `get_ch_positions` for `ContainsMixin` mixin class, which can be used by all instances to get the montage or channel positions.


mne/channels/channels.py
```diff
@@ -228,6 +229,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
         return _get_channel_types(self.info, picks=picks, unique=unique,
                                   only_data_chs=only_data_chs)
 
+    def get_ch_positions(self, picks=None):
+        """Get a dictionary of channel positions for each channel.
+
+        Parameters
+        ----------
+        picks : str | list | slice | None
+            None gets good data indices.
```
agramfort, line 238: insert empty line

agramfort: also docstring should be expanded

mne/channels/channels.py
```diff
@@ -228,6 +229,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        Returns
+        -------
+        ch_positions : dict
+            A dictionary with keys as the ch_names and values as
+             corresponding list of xyz coordinates.
```
agramfort, line 243: in which coordinate frame?

mne/channels/channels.py
```diff
@@ -228,6 +229,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        coord_frame = _frame_to_str[coord_frame_int]
+
+        # extract landmark coords
+        landmark_coords = _extract_landmarks(self.info['dig'])
+
+        # create montage and return it
+        montage = make_dig_montage(
+            ch_pos=ch_pos,
+            coord_frame=coord_frame,
+            **landmark_coords
```
agramfort, line 277: you're omitting all the dig points that are not EEG channels or fids.

have a look at function `_foo_get_data_from_dig` that should probably be renamed if you start using it.  Don't copy the code blindly as it would make a code duplication

mne/channels/channels.py
```diff
@@ -228,6 +229,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+    landmarks = {d['ident']: d for d in dig
+                 if d['kind'] == FIFF.FIFFV_POINT_CARDINAL}
+    if landmarks:
+        if FIFF.FIFFV_POINT_NASION in landmarks:
+            coords['nasion'] = landmarks[FIFF.FIFFV_POINT_NASION]['r'].tolist()
+        if FIFF.FIFFV_POINT_LPA in landmarks:
+            coords['lpa'] = landmarks[FIFF.FIFFV_POINT_LPA]['r'].tolist()
+        if FIFF.FIFFV_POINT_RPA in landmarks:
+            coords['rpa'] = landmarks[FIFF.FIFFV_POINT_RPA]['r'].tolist()
+    return coords
```
agramfort, line 294: see _get_fid_coords in mne/io/_digitization.py

mne/channels/tests/test_montage.py
```diff
@@ -1205,6 +1205,34 @@ def test_set_montage_with_missing_coordinates():
...
+    raw.set_channel_types({ch: 'ecog' for ch in raw.ch_names})
+    ch_names = raw.ch_names
+    n_channels = len(ch_names)
+    ch_coords = np.arange(n_channels * 3).reshape(n_channels, 3)
+    montage = make_dig_montage(
+        ch_pos=dict(zip(ch_names, ch_coords,)),
+        nasion=[0, 1, 0], lpa=[1, 0, 0], rpa=[-1, 0, 0],
+        coord_frame='head'
+    )
+    raw.set_montage(montage)
```
agramfort, line 1222: make it work with testing data as it contains all kinds of dig points you don't want to loose with get_montage

the test should do:
m = inst.get_montage()
inst.set_montage()
and inst should be unchanged

adam2392: What is the best way to compare if two instances are the same? Is what I am doing currently not enough? 

If not, is there a function in mne-python testing for this?

adam2392:
@agramfort lmk if you think there's any issues still?

> make it work with testing data as it contains all kinds of dig points you don't want to loose with get_montage
> 
> the test should do:
> m = inst.get_montage()
> inst.set_montage()
> and inst should be unchanged

I used both the: `fif_fname = op.join(io_dir, 'tests', 'data', 'test_raw.fif')` and the `bv_fname = op.join(io_dir, 'brainvision', 'tests', 'data', 'test.vhdr')` along w/ the `fif_dig_montage_fname = op.join(data_path, 'montage', 'eeganes07.fif')` test datasets. I used a standard montage template, as well as one loaded from data. Note that the `on_missing` is required to allow full roundtrip of the montage setting.

**FYI some issues w/  montage ordering?**
I tried testing the actual montage received, but this doesn't work and I still am not sure why, but I think  it's because the `raw.info['dig']` does not maintain any ordering, so when you try to compare using `_get_dig_montage_pos`, the assertion doesn't work nicely. 

```
    # XXX: dig montage doesn't maintain ordering...
    assert_array_equal(_get_dig_montage_pos(montage),
                        _get_dig_montage_pos(test_montage))
```

mne/channels/channels.py
```diff
@@ -228,6 +230,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+    def get_ch_positions(self, picks=None):
+        """Get a dictionary of channel positions for each channel.
+
+        Channel positions are in meters in the specified coordinate frame
+        of the montage.
+
+        Parameters
+        ----------
+        picks : str | list | slice | None
+            None gets good data indices.
```
agramfort, line 242: get ch position for MEG channel is not enough. You need orientation.

```suggestion
            None corresponds to all data channels that are not MEG.
```

mne/channels/channels.py
```diff
@@ -228,6 +230,70 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+            The coordinate frame of the montage's xyz coordinates.
+        """
+        # get the channel positions
+        picks = _picks_to_idx(self.info, picks)
+        chs = self.info['chs']
+        ch_names = self.info['ch_names']
+        pos = np.array([chs[k]['loc'][:3] for k in picks])
+
+        # extract landmark coords and coordinate frame
+        _, coord_frame_int = _get_fid_coords(self.info['dig'])
```
agramfort, line 260: you should use the coordframe provided in chs structure. Don't use dig here.

mne/channels/tests/test_montage.py
```diff
@@ -1205,6 +1205,83 @@ def test_set_montage_with_missing_coordinates():
     )
 
 
+def test_get_channel_coordinates():
+    """Test get montage and its coordinates."""
+    N_CHANNELS = 3
+
+    # create toy dataset
+    raw = _make_toy_raw(N_CHANNELS)
```
agramfort, line 1213: make your test with testing data as you'll have more channel types. Having only EEG or iEEG is too easy.

doing

raw.get_ch_positions()

seems to miss certain EEG channels on testing data.

mne/channels/channels.py
adam2392: I see, I then can create a dictionary of `{ch_name: ch_coord_frame}` then? That is what I return now instead of a single coord frame.

mne/channels/tests/test_montage.py
adam2392: I am now using 'test_raw.fif', 'test.vhdr', and 'sample_audvis_raw.fif' to do `get_montage()` -> `set_montage()`.

I added the unit tests for `get_ch_positions()`.

agramfort:
@adam2392 can you see why CIs complain? I'll review when it's green. thanks

adam2392:
> @adam2392 can you see why CIs complain? I'll review when it's green. thanks

Sorry I totally dropped the ball on this, but will ping you when the CIs pass. Doing a check rn.

adam2392:
Okay I just fixed the codecov issue and now I think that once CI is green this is good to go assuming the changes I made addressed your comments sufficiently. (i.e. using the test dataset)

mne/channels/channels.py
```diff
@@ -235,6 +237,87 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        picks = _picks_to_idx(self.info, picks)
+        chs = [self.info['chs'][ipick] for ipick in picks]
+
+        # extract name, position, orientation (for MEG)
+        ch_names, ch_pos = [], []
+        for ch in chs:
+            if ch['kind'] == FIFF.FIFFV_MEG_CH:
+                pos = ch['loc']
+            else:
+                pos = ch['loc'][:3]
```
larsoner, line 276: I wouldn't triage like this, I would just always take `[:3]`. The remaining terms are not a position but an orientation, and they aren't even in a particularly human-readable form. We should find some other way to get people that info about MEG channels if they want it.

Also, one sneaky problem that we should be careful about is copies vs views. Here `info` is never copied, `info['chs'][ii]` i.e. `ch` is a dict, `ch['loc']` is an ndarray, and `ch['loc'][:3]` is a view because it's a slice. So someone doing on the output `ch_pos_dict['EEG 001'][:] = np.nan` will set `inst.info['chs'][ii]['loc'][:3] = np.nan` effectively. So better here to do:
```suggestion
            pos = ch['loc'][:3].copy()
```

mne/channels/tests/test_montage.py
```diff
@@ -43,10 +43,14 @@
                     read_fiducials, __file__ as _MNE_IO_FILE)
 
 from mne.io import RawArray
-from mne.datasets import testing
+from mne.datasets import testing, sample
```
larsoner, line 46: This still uses `sample` this should not be necessary...

mne/io/_digitization.py
```diff
@@ -201,7 +212,7 @@ def _foo_get_data_from_dig(dig):
             # XXX: point_names.append('HPI%03d' % d['ident'])
         elif d['kind'] == FIFF.FIFFV_POINT_EXTRA:
             hsp.append(d['r'])
-        elif d['kind'] == FIFF.FIFFV_POINT_EEG:
+        else:
```
larsoner, line 215: Any reason to change this? IIRC ECoG/SEEG etc. still get named `_EEG` in this structure

mne/channels/channels.py
adam2392: Good catch! thanks.

mne/channels/tests/test_montage.py
adam2392: Hmm do you think I should keep it for the sample MEG data? I'm using `fif_fname = op.join(io_dir, 'tests', 'data', 'test_raw.fif')` right now mainly otherwise. I'm not familiar with what's in the testing data, so if that is sufficient, then sounds good!

mne/io/_digitization.py
adam2392: Hmm not sure why I ended up changing this. Changed it back and it didn't break, so assuming it was a byproduct of an earlier iteration.

mne/channels/tests/test_montage.py
larsoner: You can use sample_audvis_trunc_raw.fif in the testing dataset, it has the same channels it has just been truncated

mne/channels/channels.py
```diff
@@ -235,6 +237,84 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        ----------
+        picks : str | list | slice | None
+            None corresponds to all data channels that are not MEG.
+
+        Returns
+        -------
+        ch_positions : dict
+            A dictionary with keys as the ch_names and values as
+             corresponding list of xyz coordinates in the montage
+             coordinate frame.
```
agramfort, line 261: above you say it's not only xyz but also orientation for MEG

mne/channels/channels.py
```diff
@@ -235,6 +237,84 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        ch_frame_dict : dict
```
agramfort, line 262:
```suggestion
        ch_coord_frame : dict
```
you don't have _dict for ch_positions

mne/channels/channels.py
```diff
@@ -235,6 +237,84 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+            A dictionary with keys as the ch_names and values as
+             corresponding montage coordinate frame.
+        """
+        # only get channels that were picked
+        picks = _picks_to_idx(self.info, picks)
+        chs = [self.info['chs'][ipick] for ipick in picks]
+
+        # extract name, position, orientation (for MEG)
```
agramfort, line 270: where do you get orientation?

mne/channels/channels.py
```diff
@@ -235,6 +237,84 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+            ch_names.append(ch['ch_name'])
+
+        # extract landmark coords and coordinate frame
+        ch_coord_frame = [_frame_to_str[ch['coord_frame']] for ch in chs]
+
+        # construct channel position dictionary and coord frame
+        ch_pos_dict = dict(zip(ch_names, ch_pos))
+        ch_frame_dict = dict(zip(ch_names, ch_coord_frame))
+
+        return ch_pos_dict, ch_frame_dict
```
agramfort, line 284:
```suggestion
        return ch_positions, ch_coord_frame
```

mne/channels/channels.py
```diff
@@ -235,6 +237,84 @@ def get_channel_types(self, picks=None, unique=False, only_data_chs=False):
...
+        ch_pos = dict(zip(ch_names, ch_locs))
+        montage = make_dig_montage(
+            ch_pos=ch_pos,
+            coord_frame=coord_frame,
+            nasion=montage_bunch.nasion,
+            lpa=montage_bunch.lpa,
+            rpa=montage_bunch.rpa,
+            hsp=montage_bunch.hsp,
+            hpi=montage_bunch.hpi,
+        )
```
agramfort, line 315: I am fine with this code that I trust but why having method get_ch_positions if you can just do:

raw.get_montage().ch_pos

to get channel positions?

agramfort: it's also not using the same code so it's error prone

mne/channels/channels.py
adam2392: Oh yeah, leaving out orientation now per @larsoner  comment on https://github.com/mne-tools/mne-python/pull/7667#discussion_r481502708

mne/channels/channels.py
adam2392: So rn, I'm not sure why but the latest commit fails when doing this. That is, i) I get a montage and its coordinates, then ii) I `raw.set_montage()`, then iii) I try to get the coordinates using either `chs['loc'][:3]`, or `.get_montage()._get_ch_pos()`.

In iii), the `chs['loc'][:3]` works fine, but for some reason `.get_montage()._get_ch_pos()` does not, and so I want to suspect that there is a transformation that occurs when doing `set_montage()` because the roundtrip of `get_montage` and `set_montage` works fine.

adam2392: I just pushed up the change and you'll see that the tests fail here:

```
mne/channels/tests/test_montage.py:1329: in test_get_ch_positions
    assert_array_equal(montage_ch_pos[ch], pos)
E   AssertionError: 
E   Arrays are not equal
E   
E   Mismatched elements: 3 / 3 (100%)
E   Max absolute difference: 0.06785223
E   Max relative difference: 0.78861487
E    x: array([-0.001064,  0.018188,  0.127724], dtype=float32)
E    y: array([-0.001133,  0.08604 ,  0.081454], dtype=float32)
```

larsoner: It's a ref channel problem, pushed a commit

agramfort:
I woudl suggest we limit this PR to get_montage that returns channels that it make sense to use in set_montage (EEG, seeg, ecog, fnirs). I would leave the MEG out of this.

adam2392:
> I woudl suggest we limit this PR to get_montage that returns channels that it make sense to use in set_montage (EEG, seeg, ecog, fnirs). I would leave the MEG out of this.

Sorry for the stupid questions (mne-python codebase is still a big one for me to understand), but... does what I have rn meet that? Since @larsoner suggested I just drop the orientations, I don't think I am doing it anymore(?) Or do I have to add an explicit picks somewhere in the `get_ch_positions()` function?

larsoner:
I think the idea is to get rid of `get_ch_positions` entirely and just have `get_montage` which returns positions only for the same set of channels that `set_montage` can set (EEG, sEEG, ECoG, not sure about fNIRS...)

agramfort:
> I think the idea is to get rid of get_ch_positions entirely and just have get_montage which returns positions only for the same set of channels that set_montage can set (EEG, sEEG, ECoG, not sure about fNIRS...)
+1

adam2392:
Okay got rid of `get_ch_positions` and just left `get_montage` along w/ the related tests.

mne/channels/tests/test_montage.py
```diff
@@ -1298,6 +1298,112 @@ def test_set_montage_with_missing_coordinates():
...
+    # if montage gets set to None
+    raw.set_montage(None)
+    assert raw.get_montage() is None
+
+    # read in BV test dataset and make sure montage
+    # fulfills roundtrip on non-standard montage
+    dig_montage = read_dig_fif(fif_dig_montage_fname)
+
+    # Make a BrainVision file like the one the user would have had
+    # with testing datasete 'test.vhdr'
```
agramfort, line 1380:
```suggestion
    # with testing dataset 'test.vhdr'
```

agramfort (commented):
I am really not sure why tests are so complicated.

I would read in the testing fif file. do a get_montage and set_montage and assert with objecf_diff that chs and dig are the same.

I would then do a set_montage to an EEG file without montage and then do a get_montage and assert it's also consistent with object_diff

adam2392:
Not sure why the round trip for this test dataset is failing: `fif_fname = op.join(io_dir, 'tests', 'data', 'test_raw.fif')`

I get a mismatch say for example:
```
# original location of info['chs'][316]['loc']
[-0.01491282  0.11079782  0.07650087  0.          0.          0.
  0.          1.          0.          0.          0.          1.        ]

# new location of info['chs'][316]['loc']
[-0.01491282  0.11079782  0.07650087  0.00235201  0.11096951 -0.03500458
  0.          1.          0.          0.          0.          1.        ]
```

So I'm not sure if that's coming from the fact that there is a orientation change due to `_get_data_as_dict_from_dig`? However, the coordinates are the same...

larsoner:
For EEG, first three entries in `loc` are the electrode pos, next three are the ref electrode pos (if present)

adam2392:
> For EEG, first three entries in `loc` are the electrode pos, next three are the ref electrode pos (if present)

It seems that the ref position gets set to 0 if it was already set in the `fif` file. Namely here: https://github.com/mne-tools/mne-python/blob/f6cb5110deddeee724f153ecd93bc5a5aa13adca/mne/channels/montage.py#L692-L695

Then if one has a raw file without those 2 REF names, then it sets the ref location to [0, 0, 0]. So then the `get_montage` -> `set_montage` will change the ref locations. 

Is there a suggested way to either modify the test (e.g. only test the first 3 coordinates in `'loc'`), or modify the `_set_montage` function to account for this?

larsoner:
> It seems that the ref position gets set to 0 even if it was already set in the fif file; it will do this when the set of channel names does not have one called REF, or EEG000. Namely here:

That code looks to me like it looks for EEG000 or REF and uses that as the value for `loc[3:6]`.

I would expect the fix to be, in your code, look at `ch['loc'][3:6]` for all EEG channels. They should all match (either all zero or all something else), and if it's not all zero, then in `get_montage` set either `EEG000` or `REF` (depending on the channel naming scheme) to that value.

agramfort:
yes get_montage should return a REF channel if ch['loc'][3:6] is not 0 for
EEG channels

>

mne/channels/tests/test_montage.py
```diff
@@ -1304,6 +1304,75 @@ def test_set_montage_with_missing_coordinates():
...
+    # 1. read in testing data and assert montage roundtrip
+    # for testing dataset: 'test_raw.fif'
+    raw = read_raw_fif(fif_fname)
+    raw = raw.rename_channels(lambda name: name.replace('EEG ', 'EEG'))
+    raw2 = raw.copy()
+    # get montage and then set montage and
+    # it should be the same
+    montage = raw.get_montage()
+    raw.set_montage(montage, on_missing='ignore')
+    assert_object_equal(raw.info['chs'], raw2.info['chs'])
```
agramfort, line 1321: can you test dig too?

mne/channels/tests/test_montage.py
```diff
@@ -1304,6 +1304,75 @@ def test_set_montage_with_missing_coordinates():
...
+    # set the montage; note renaming to make standard montage map
+    raw.set_montage(montage)
+
+    # get montage back and set it
+    # the channel locations should be the same
+    orig_chs = raw.info['chs']
+    test_montage = raw.get_montage()
+    raw.set_montage(test_montage, on_missing='ignore')
+    test_chs = raw.info['chs']
+    assert_object_equal(orig_chs, test_chs)
```
agramfort, line 1334: can you test that test_montage is the same as montage?

mne/channels/tests/test_montage.py
```diff
@@ -1304,6 +1304,75 @@ def test_set_montage_with_missing_coordinates():
...
+        ch['kind'] = FIFF.FIFFV_EEG_CH
+
+    # Set the montage and roundtrip
+    raw_bv.set_montage(dig_montage)
+    raw_bv2 = raw_bv.copy()
+
+    # reset the montage
+    test_montage = raw_bv.get_montage()
+    raw_bv.set_montage(test_montage, on_missing='ignore')
+    assert_object_equal(raw_bv2.info['chs'], raw_bv.info['chs'])
```
agramfort, line 1369: can you test dig too?

mne/channels/tests/test_montage.py
adam2392: So I'm not sure why this is the case for this test dataset, but `info['dig']` contains an extra `EEG #0` digPoint, whereas when we do the set_montage that goes away since there is no `EEG000` or `REF` channel inside the actual `info['ch_names']`.

This results in 145 digpoints after set_montage, but 146 digpoints before. This occurs at index 85. Everything else matches up. I confirmed that this is the Reference DigPoint because it has the exact same positions as the reference location in all the `info['chs'][ii]['loc'][3:6]`.

**Possible solutions?**
1. Fix when I call `set_montage`: I need to keep track of the fact that there might be a REF location set in the raw file, but no actual REF channel in the dataset by name (EEG000, REF), and if so, add that DigPoint to the `info['dig']. The issue now is when I assert objects are equal, the order is not the same as before I do `set_montage()`

2. Fix when I call `get_montage`: Here, it's not clear to me how I would do this, but I would need a way of making a montage that makes sure to set the dig later with an extra DigPoint even if there is no channel name present.

3. If I don't compare the dig, then everything else matches. That is `info['chs']` because the REF location is encoded into the [3:6] inside the `'loc'` dictionary.

adam2392: Sorry if this is not the right interpretation (I'm not 100% sure I fully understand Digs vs channel locations yet, so I might be missing something). I did a bit of debugging and pushed up the updated unit test so you can see where it breaks.

agramfort:
In dig for me it's still a pb of EEG #0 

```
(Pdb) raw.info['dig'][85:90]
[<DigPoint |     EEG #1 : (-37.4, 105.7, 73.3) mm   : head frame>, <DigPoint |     EEG #2 : (-14.9, 110.8, 76.5) mm   : head frame>, <DigPoint |     EEG #3 : (12.1, 109.6, 77.4) mm    : head frame>, <DigPoint |     EEG #4 : (-65.7, 78.2, 65.5) mm    : head frame>, <DigPoint |     EEG #5 : (-50.6, 84.5, 88.3) mm    : head frame>]
(Pdb) raw2.info['dig'][85:90]
[<DigPoint |     EEG #0 : (2.4, 111.0, -35.0) mm    : head frame>, <DigPoint |     EEG #1 : (-37.4, 105.7, 73.3) mm   : head frame>, <DigPoint |     EEG #2 : (-14.9, 110.8, 76.5) mm   : head frame>, <DigPoint |     EEG #3 : (12.1, 109.6, 77.4) mm    : head frame>, <DigPoint |     EEG #4 : (-65.7, 78.2, 65.5) mm    : head frame>]
```

adam2392:
Okay so it's just a problem with ordering the Dig inside the info, but on the test side, this can be fixed by just doing two loops over both digPoints and checking existence in each other. Since the `info['chs']` are the same, the digPoints ordering does not seem to matter that much? 

I added the updated code to `_set_montage`, seen here:

```Python
# get a copy of the old dig
        if info['dig'] is not None:
            old_dig = info['dig'].copy()
        else:
            old_dig = []

        # determine if needed to add an extra EEG REF DigPoint
        if custom_eeg_ref_dig:
            # ref_name = 'EEG000' if match_case else 'eeg000'
            ref_dig_dict = {'kind': FIFF.FIFFV_POINT_EEG,
                            'r': eeg_ref_pos,
                            'ident': 0,
                            'coord_frame': info['dig'].pop()['coord_frame']}
            ref_dig_point = _format_dig_points([ref_dig_dict])[0]
            # only append the reference dig point if it was already
            # in the old dig
            if ref_dig_point in old_dig:
                digpoints.append(ref_dig_point)
        info['dig'] = _format_dig_points(digpoints)
```

Which will account for the extra EEGREF Digpoints when it was present in the original Dig.

adam2392:
On the test side, I updated the tests to do a "set" check and a coordinate check on the DigPoints and the montage roundtrip.

mne/channels/montage.py
```diff
@@ -746,13 +764,34 @@ def _backcompat_value(pos, ref_pos):
...
+            # ref_name = 'EEG000' if match_case else 'eeg000'
+            ref_dig_dict = {'kind': FIFF.FIFFV_POINT_EEG,
+                            'r': eeg_ref_pos,
+                            'ident': 0,
+                            'coord_frame': info['dig'].pop()['coord_frame']}
+            ref_dig_point = _format_dig_points([ref_dig_dict])[0]
+            # only append the reference dig point if it was already
+            # in the old dig
+            if ref_dig_point in old_dig:
+                digpoints.append(ref_dig_point)
```
larsoner, line 793: It would be better not to mess with the order -- it's pretty standard to have the ref be the first of all EEG electrodes. I would use `.insert` it to the `digpoeints` list or add it to the list *first* rather than *last*. Then even the round-trip order should match, which is nice.

agramfort:
+1 to try to keep the order if one can.

can we do something to help?

adam2392:
Okay that sounds fair to me. I think what will be needed is to first parse the digpoints about to be set and then keep track of type and ident such that they are ordered in the following way:

- Fiducials
- None
- EEG
- ecog/seeg(?)

Is there an acceptable order to hardcode? In addition that brings up a question as I type this out: how are digpoints for IEEG represented exactly... (are they differentiated due to coord frame)?

larsoner:
AFAIK we don't differentiate between different electrode types (EEG, ECoG, sEEG) in dig. In theory just the channel names matching in the `montage` instance should be good enough for those use cases.

Yes the standard order is fiducials/HPI (if present)/EEG/extra (headshape) points. So I think all that should need to be tweaked here really is ensuring that the ref gets `ident`==0 and is first in the list of EEG points...

adam2392:
> AFAIK we don't differentiate between different electrode types (EEG, ECoG, sEEG) in dig. In theory just the channel names matching in the `montage` instance should be good enough for those use cases.
> 
> Yes the standard order is fiducials/HPI (if present)/EEG/extra (headshape) points. So I think all that should need to be tweaked here really is ensuring that the ref gets `ident`==0 and is first in the list of EEG points...

Got it working by adding a "sorting" algorithm applied to the `_format_dig_points` function.

agramfort (approved):

larsoner:
Thanks @adam2392 !

PR #7673, opened 2020-04-24
TST: Check Travis

GuillaumeFavelier:
I investigate https://github.com/mne-tools/mne-python/pull/7620#issuecomment-617100952 in this PR


larsoner:
@GuillaumeFavelier also #7674 should help, it should I think make the test show up as a failure rather than killing the whole test suite with a segfault. At least that's what I've seen happen when `pytest-xdist` has been used in SciPy

GuillaumeFavelier:
I'll add it in the next run, thanks for the tip :+1:

GuillaumeFavelier:
For reference:

https://github.com/mne-tools/mne-python/pull/7674#issuecomment-619207074

larsoner:
Things seem more stable now so I'll go ahead and close this

PR #7676, opened 2020-04-24
MRG, ENH: Speed up examples

larsoner:
Tries to speed up our two slowest examples:
```
    - ../tutorials/source-modeling/plot_forward.py: 115.75 sec    448.3 MB
    - ../tutorials/source-modeling/plot_mne_solutions.py: 96.71 sec    383.4 MB
```
In the first one, if we use fewer sources it's faster. In the second one, just showing the loose inverse seems like enough (no need to compute and show fixed and free).

### DIFF
--- a/tutorials/source-modeling/plot_forward.py
+++ b/tutorials/source-modeling/plot_forward.py
@@ -59,8 +59,9 @@ subject = 'sample'
 # assumes that you have the ``bem`` folder of your subject's FreeSurfer
 # reconstruction, containing the necessary files.
 
+slices = [60, 128, 196]  # just a few for speed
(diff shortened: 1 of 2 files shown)


agramfort:
I am not too enclined to sacrifice pedagogy for CI speed. Users will copy paste from tutorials without realizing they don't use good default or will complain that tutorials use precomputed stuff that they don't know how to produce with their own data

larsoner:
Okay we can live with the fwd being slow then, and maybe I'll redo the `mne_solutions` to still show free and fixed, but not all four variants for each (the loose eloreta is the slow one anyway)

PR #7682, opened 2020-04-27
MRG, DOC: Properly document str option to concatenate_raws

### LINKED ISSUE #6380: ENH: Add preload=str to concatenate_raw and concatenate_epochs
Hello all, I am using mne.concatenate_raws to collect all my raw data for a single subject. This often caps out my memory on my computer. Is there a way to better manage memory in this situation. Another potential solution is to concatenate epochs which is much better for memory. However, I am concerned that filtering short epochs would cause other issues in my data.

@larsoner has suggested the following:
can you open an issue about concatenate_raws supporting preload as a string?
basically it would allow using a memmap'ed file
in the meantime you can filter each raw instance and save it to disk
and load each of those filtered ones with preload=False (don't load into memory)
then do concatenate_raws on those
then do mne.Epochs on the concatenated instance
the epochs will pull data from the concatenated instance, which will pull them from disk instead of memory

larsoner:
Closes #6380

We already have `str` support in `append `, just need to document it in `concatenate_raws` that calls `append` internally. That issue also mentions making `concatenate_epochs` allow a `str` but I think this would require extensive refactoring of our `Epochs` class, and hopefully is YAGNI knowing that this is available.

Also updates a link from our roadmap to GitHub.

### DIFF
--- a/doc/overview/roadmap.rst
+++ b/doc/overview/roadmap.rst
@@ -114,6 +114,8 @@ of by MNE. Subgoals consist of:
   code to use cloud computing (optionally, based on config) rather than local
   resources.
 
+See also :gh:`6086`.
+
 
 Tutorial / example overhaul
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


agramfort:
thx @larsoner

PR #7685, opened 2020-04-27
MRG, ENH: Add DigMontage.rename_channels

### LINKED ISSUE #7528: ENH: Add montage.rename_channels
It should just be a simple dict remapping, I think. It seems like a nicer thing to do than ask people to rename their raw data channels (and possibly rename them back).

larsoner:
Closes #7528


agramfort:
thx @larsoner

PR #7687, opened 2020-04-28
DOC: Add "ambidextrous" to docstring

hoechenberger:
Adds "ambidextrous" to `Info` docstring.

#### Additional information
As is defined in `io.constants` already.


agramfort:
thx @hoechenberger

Issue #7683, opened 2020-04-27
viz.get_3d_backend() should return None if no 3d backend was found, instead of raising an exception


hoechenberger:
#### Describe the problem
Currently, `mne.viz.get_3d_backend()` will raise a `RuntimeError` if no usable backend has been found:

https://github.com/mne-tools/mne-python/blob/d0b8bcf5201d7b5c278e7410e4035dc5aead30b8/mne/viz/backends/renderer.py#L119-L147

#### Describe your solution
I believe it should return `None` instead: if there is currently no 3d backend, then, well, there is none ;)

#### Describe possible alternatives
I currently have to be ready to catch a `RuntimeError` when I simply want to query whether I can produce a 3d visualization or not. Background: I'm plotting STCs in environments that sometimes don't have working `pyvista` or `mayavi` installations. I want to find out when that's the case, and switch to using matplotlib instead.

#### Additional context
I checked how `get_3d_backend()` is currently used, and without exception, any use is a comparison like`if get_3d_backend() != 'pyvista':`; therefore, returning `None` if no working backend was found should not break any existing code.

cc @GuillaumeFavelier

agramfort:
wdyt @GuillaumeFavelier ?

GuillaumeFavelier:
`get_3d_backend()` was not designed for such a use. All the functions of `mne/viz/backends/renderer` expect it to return a valid 3d backend. So it's possible to make it return `None` instead of raising an error but I can't evaluate the impact of this change to the rest of those functions, which are used extensively in `_3d.py`. What would be the default behaviour of those functions if no 3d backend is found?

If the goal is to check if you can use a 3d backend, I could refactor a function that tests just that and return `None` otherwise. What do you think?

hoechenberger:
> `get_3d_backend()` was not designed for such a use. All the functions of `mne/viz/backends/renderer` expect it to return a valid 3d backend. So it's possible to make it return `None` instead of raising an error but I can't evaluate the impact of this change to the rest of those functions, which are used extensively in `_3d.py`. What would be the default behaviour of those functions if no 3d backend is found?

Yes I was concerned about this too, bc then these functions all would have to check if `None` was returned or not, which seems cumbersome. However like I said, I've searched the sources for calls of this function, and there's always a comparison like `if get_3d_backend() == 'pyvista'`, so it should be safe to make such a change, but …

> 
[...]

PR #7292, opened 2020-02-05
Refactor reload_backend()

GuillaumeFavelier:
This PR changes the API of renderer so that the private `_reload_backend()` only reloads the given backend and `get_3d_backend()` always returns a valid backend name (to avoid getting  `None` for example). It also limits the amount of backend reloads to the bare minimum:

scenario |`master` | PR
----------|---------------|-------
example.py | Using mayavi 3d backend.<br>Using mayavi 3d backend. | Using mayavi 3d backend.
with `use_3d_backend('pyvista')` | Using pyvista 3d backend.<br>Using pyvista 3d backend.<br>Using mayavi 3d backend.<br>Using mayavi 3d backend. | Using mayavi 3d backend.<br>Using pyvista 3d backend.<br>Using mayavi 3d backend.
with `set_3d_backend('pyvista')` | Using pyvista 3d backend.<br>Using pyvista 3d backend. | Using pyvista 3d backend.
with `MNE_3D_BACKEND=pyvista` | Using pyvista 3d backend. | Using pyvista 3d backend.

PR #7688, opened 2020-04-28
WIP: Refactor get_3d_backend primitive

GuillaumeFavelier:
This PR follows the discussion on https://github.com/mne-tools/mne-python/issues/7683 and introduces a new `have_3d_backend` public function that returns `True` if any 3d backend is available and `False` otherwise.

Closes https://github.com/mne-tools/mne-python/issues/7683


mne/viz/backends/renderer.py
```diff
@@ -33,6 +33,26 @@ def _get_renderer(*args, **kwargs):
...
+    status: bool
+        Return True is a valid 3d backend is found and False otherwise.
+    """
+    for name in VALID_3D_BACKENDS:
+        try:
+            importlib.import_module(name=name)
+        except ImportError:
+            pass
+        else:
+            return True
```
larsoner, line 52: Why not just:
```
try:
    get_3d_backend()
except RuntimeError:
    return False
else:
    return True
```
? It's more DRY

drammock:
Sorry to chime in late (time zone problem), but I'd like to suggest a slightly different solution. Change existing `get_3d_backend` to private function and add a thin public wrapper with same name that returns none when appropriate. I feel like as soon as this functionality is out in the wild, some user will say "can this return actual backend instead of True" so let's just anticipate that now.

agramfort:
I like this option as it avoids a new public function

>

hoechenberger:
I like @drammock's suggestion :)

mne/viz/backends/renderer.py
GuillaumeFavelier: The backend is reloaded in `get_3d_backend` since the lazy backend loading I did in https://github.com/mne-tools/mne-python/pull/7292#issue-371545900

GuillaumeFavelier:
Thanks all for suggestions! I will change the name of the PR to refactor of `get_3d_backend` and see how to achieve:

> Change existing get_3d_backend to private function and add a thin public wrapper with same name that returns none when appropriate

drammock:
@GuillaumeFavelier I think a variation of @larsoner's suggestion should do it; change existing function to start with an underscore, and then the public function is:

```python
def get_3d_backend():
    try:
        backend = _get_3d_backend()
    except RuntimeError:
        return None
    return backend
```

GuillaumeFavelier:
I'll try it right now!

hoechenberger:
@GuillaumeFavelier maybe I'm just missing something here after a long day, but don't you also want to adjust all your "internal" code to make use of the private function now instead of the public interface?

mne/viz/backends/renderer.py
```diff
@@ -124,6 +124,15 @@ def get_3d_backend():
     backend_used : str
         The 3d backend currently in use.
```
drammock, line 125:
```suggestion
    backend_used : str | None
        The 3d backend currently in use. If no backend is found, returns ``None``.
```

drammock:
> @GuillaumeFavelier maybe I'm just missing something here after a long day, but don't you also want to adjust all your "internal" code to make use of the private function now instead of the public interface?

yes, I was also intending this, but failed to mention it explicitly.

GuillaumeFavelier:
I ran into an issue with `test_renderer` in `mne/viz/tests/test_3d.py`:

```
            if 'output' in _get_args(err_fun):
>               raise subprocess.CalledProcessError(p.returncode, command, output)
E               subprocess.CalledProcessError: Command '['/home/guillaume/source/anaconda3/envs/mne/bin/python', '-uc', "import mne; mne.viz.create_3d_figure((800, 600)); backend = mne.viz._get_3d_backend(); assert backend == 'mayavi', backend"]' returned non-zero exit status 1.
```

GuillaumeFavelier:
Probably caused by:

```
"AttributeError: module 'mne.viz' has no attribute '_get_3d_backend'\n"
```

hoechenberger:
> ```
> "AttributeError: module 'mne.viz' has no attribute '_get_3d_backend'\n"
> ```

Which is correct, since you don't import `_get_3d_backend` globally in that module, as far as I can see. How did you do that previously? Wanna quickly look at the code together? It's coffee time anyway ;)

GuillaumeFavelier:
> Wanna quickly look at the code together? It's coffee time anyway ;)

Lead the way

larsoner:
Thanks @GuillaumeFavelier

PR #7689, opened 2020-04-28
FIX: patch doctest

larsoner:
Iterating with @Aniket-Pradhan until things work for Fedora. Then I'll merge this and cut a 0.20.2, and port changes up to `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,9 @@ matrix:
         - os: osx
           env: CONDA_ENV="environment.yml"
 
-        # PIP + non-default stim channel
+        # PIP + non-default stim channel + log level info
         - os: linux
-          env: MNE_STIM_CHANNEL=STI101
+          env: MNE_STIM_CHANNEL=STI101 MNE_LOGGING_LEVEL=info
           language: python
           python: "3.7"
 
--- a/doc/changes/0.20.inc
+++ b/doc/changes/0.20.inc
@@ -20,7 +20,7 @@ Bug
 - Fix bug with :func:`mne.preprocessing.nirs.scalp_coupling_index` where filter transition was incorrectly assigned by `Robert Luke`_
 
 - Fix support for signals in mV for :func:`mne.io.read_raw_brainvision` by `Clemens Brunner`_
-￼
+
 - Fix bug with :func:`mne.minimum_norm.compute_source_psd_epochs` and :func:`mne.minimum_norm.source_band_induced_power` raised errors when ``method='eLORETA'`` by `Eric Larson`_
 
 ￼- Fix bug with :func:`mne.make_forward_dipole` where :func:`mne.write_forward_solution` could not be used by `Eric Larson`_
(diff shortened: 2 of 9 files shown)


Aniket-Pradhan:
Another issue that I had been facing is here: `mne/utils/tests/test_logging.py:test_get_call_line`

The error log I get is:
```
>           assert my_line == 'my_line = foo(verbose=v)  # testing'
E           AssertionError: assert 'unknown' == 'my_line = foo(verbose=v)  # testing'
E             - unknown
E             + my_line = foo(verbose=v)  # testing
```

I was running the test case inside a [`mock`](https://github.com/rpm-software-management/mock/wiki) (basically in a VM). Now, I am not sure whether the test fails because of the environment or not, but when I run the test using a python virtual environment, it works fine.

I tried debugging this issue, for which I checked the second-last frame (instead of the [third-last](https://github.com/mne-tools/mne-python/blob/master/mne/utils/_logging.py#L305) and I got `return _get_call_line()` which is correct. The third last frame gives a `None` value, for which the test fails.

If you think that this is happening because of running the test in mock, I can skip this test, since it's working perfectly fine for me in a virtual environment.

larsoner:
Yes mock is almost certainly the problem because that test relies on `inspect` giving correct information

Aniket-Pradhan:
Ach... no worries then.

One last doctest failure for you: [`mne/filter.py:621`](https://github.com/mne-tools/mne-python/blob/master/mne/filter.py#L621)

I am sort of confused, as to why these doctest failures don't come up on Travis builds?

Thanks a lot for working on this, and helping me out here!

larsoner:
> I am sort of confused, as to why these doctest failures don't come up on Travis builds?

Tests might not be run at all / default logging levels, or it's possible that some test modifies the logging level without setting it back. Basically we have some cruft somewhere...

larsoner:
Pushed a commit to fix that bug and modify a CI run to use MNE_LOGGING_LEVEL=info (the default for new users) rather than warning

Aniket-Pradhan:
Okay, just one last failure.

I am not getting a `UserWarning` as expected in [mne/viz/tests/test_ica.py:180](https://github.com/mne-tools/mne-python/blob/master/mne/viz/tests/test_ica.py#L179).

```
        with pytest.warns(UserWarning, match='Infinite value .* for epoch 0'):
>           ica.plot_properties(epochs)
E           Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) was emitted. The list of emitted warnings is: [].
```

I just want to ask if I should really expect any warnings in this section? I was unable to find any mention of UserWarning with ICA

larsoner:
It should warn but it's not critical that we verify it. Pushed a commit to be less stringent

Aniket-Pradhan:
Well, that fixes all the tests.

Thanks a lot, @larsoner for everything :heart: 

I will be waiting for the next release to update python-mne for Fedora

larsoner:
Should be able to push it to PyPi tonight or tomorrow morning. Want to wait for CIs to actually come back happy, and also see it work on `master`

larsoner:
@Aniket-Pradhan you should be good to go

https://pypi.org/project/mne/

larsoner:
(merged by push with version and `0.20.inc` updates)

Aniket-Pradhan:
@larsoner looks like you accidentally didn't include #7690 in v0.20.2. Or did I get you wrong and the commit will be available in the next release?

larsoner:
The changes should be there, are they not?

larsoner:
Argh, they are not. Will cut another one shortly

larsoner:
Okay should be fixed now

https://pypi.org/project/mne/#files

Aniket-Pradhan:
Thanks for your help! :smile:

PR #7695, opened 2020-04-29
MRG: Refactor _get_ps_kwarg

GuillaumeFavelier:
This PR refactors `_3d.py` and removes a dependency to `surfer`. It should help with the following item of #7162 : 

image


agramfort:
thx @GuillaumeFavelier

PR #7211, opened 2020-01-15
WIP: Add text above orientation slider

GuillaumeFavelier:
This PR updates the orientation slider of `_TimeViewer` and adds a `vtkTextActor` above the slider.

**Known issues**

For now, the text does not scale with the window size

Fullscreen | Resized
--------------|----------
2020-01-15_1920x1080 | 2020-01-15_1920x1080

It's an item of #7162

PR #7697, opened 2020-04-29
WIP: Port 3d testing to BackgroundPlotter completely

GuillaumeFavelier:
This PR ports testing of the PyVista 3d backend to the `BackgroundPlotter` completely.

As a consequence, the variable `MNE_3D_BACKEND_TESTING` needs refactoring.

**Known issue**
- The dimensions of a screenshot taken with the `BackgroundPlotter` are not equal to `win_height x win_width x 3` but slightly less. I suppose that the real dimensions taken into account are effectively those of the render view (which is smaller than the window).


GuillaumeFavelier:
I reported this behaviour in https://github.com/pyvista/pyvista/issues/703

GuillaumeFavelier:
One way to mitigate this issue is to use `setMinimumSize()` on the render view widget so that the window don't automatically shrink it.

GuillaumeFavelier:
> the window does not automatically shrink it.

The drawback is that the user is no longer able to resize the window to a geometry smaller then `window_size`.

GuillaumeFavelier:
This can be relevant to https://github.com/mne-tools/mne-python/pull/7211 (about minimum/fixed widget geometry)

larsoner:
> The drawback is that the user is no longer able to resize the window to a geometry smaller then window_size.

Can you set this until the `super` init finishes, then un-set it?

GuillaumeFavelier:
I use it after `BackgroundPlotter` init to resize the render view. I'll see if there is a way to un-set it without updating the geometry of the window.

larsoner:
> I use it after BackgroundPlotter init to resize the render view. 

Can you see if it's possible to do *before* initializing the view? It should help prevent more jerky resizes, if those happen with the current approach

GuillaumeFavelier:
The window is resized before being shown so I don't think this will lead to jerkiness. But this is a workaround without a way to setting it at init. So far, locally the main issue I notice is with `show_traces=True`. When the window opens:

image

I have to manually resize to get all the content:

image

Also, the CIs do not seem to like those changes.

larsoner:
Probably whatever `setMinSize` you do for the 3D area, you need to do for the 2D area as well. Maybe at 1/4 of whatever the requested 3D vertical size is?

larsoner:
For now you could `assert_allclose(shape, expected_shape, atol=40)  # XXX undo once size fixed` or so to get things to pass. Seems worth it to get the actual BackgroundPlotter used, then we can deal with fixing the size separately

GuillaumeFavelier:
Awesome idea, thank you. I was stuck with `SetMinimumSize()` tbh

larsoner:
The Azure failure is fixed by #7740, but I couldn't see the failed OSX log so I restarted that build.

larsoner:
Failures on OSX apparently seem real

https://travis-ci.org/github/mne-tools/mne-python/builds/683971748

GuillaumeFavelier:
Actually if https://github.com/mne-tools/mne-python/pull/7758 is chosen, I would prefer to keep testing on both plotters.

PR #7699, opened 2020-04-29
safer indexing/slicing in internal multitaper code

drammock:
This is a tiny fix to internal multitaper code. It shouldn't affect anything that uses the public API; these are changes I needed to make to use some of the private multitaper functions in my analysis pipeline, to allow using the private functions with both epochs and evoked objects. Should be fully backward-compatible, and should make future expansion of the multitaper API slightly easier.

(for those interested, in an SSVEP experiment I needed to keep the complex values --- which meant keeping the separate spectral estimates from each taper --- so that later I could hopefully get some phase-cancelling noise reduction in the non-carrier / non-oddball frequency bins).


agramfort:
thx @drammock

PR #7704, opened 2020-04-30
MAINT: Add ultraslow run

larsoner:
As discussed in the flash bem issues/PRs, it would be good to have a run that actually executes our ultraslow tests. This PR adds one to Azure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,18 +10,15 @@ jobs:
         - run:
             name: Set BASH_ENV
             command: |
+              set -e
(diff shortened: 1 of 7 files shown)


larsoner:
This appears to be working:

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=6899&view=logs&j=bf0f2fcf-46b6-5ffd-e30e-4688aa34951e&t=f37d81b4-3e73-5034-37ae-dd1a020c26b3

I'll get the CircleCI refactor working, then we should be good

PR #7705, opened 2020-04-30
MRG, ENH: Add example of projection to source space

larsoner:
@SherazKhan asked if we had an example of projecting decoder coefficients to source space yet and I didn't think we did. This adds it to the decoding tutorial, and cleans up some refs to use the new footbib magic.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,6 +258,12 @@ jobs:
               if grep -q html_dev-pattern build.txt || grep -q html_dev-noplot build.txt; then
                 tar czf doc/_build/html/_downloads.tgz doc/_build/html/_downloads
                 rm -Rf doc/_build/html/_downloads
+                rm -f doc/auto_*/*/*.pickle
+                rm -f doc/auto_*/*/*.codeobj
+                rm -f doc/auto_*/*/*.md5
+                rm -f doc/auto_*/*/*.py
+                rm -f doc/auto_*/*/*.ipynb
+                rm -f doc/generated/*.examples
               fi
 
         # Save the JUnit file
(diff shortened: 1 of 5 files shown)


dengemann:
The way it’s coded here you add back the mean (via inverting scaler) which gives you essentially the grand average ERF.
This could be fixed by resubtracting the scaler.mean_

larsoner:
It seems weird to add in this manual step. Isn't it informative to start to know that it is basically making use of the grand average ERF to do the classification, even if it does some processing like mean subtraction and standardization, PCA, whatever else, along the way?

dengemann:
But the model uses the de-meaned data. You essentially add back the intercept which will suffocate the model specific-patterns.

larsoner:
Here I just make use of an existing `evoked_time_gen` that's already part of the narrative. So if we're going to make that change, it's actually orthogonal to this one (should be separate PR). Can you make a PR to improve this part of the tutorial and make `evoked_time_gen` more informative in this regard?

dengemann:
Yeah it’s an old issue just seeing that brings it back to memory.

I think we should make these considerations part of the tutorial and explore the different outputs.

larsoner:
Updated doc:

https://19760-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/machine-learning/plot_sensors_decoding.html#projecting-sensor-space-patterns-to-source-space

It seems to look wacky because we are getting "403 (Forbidden)" errors when downloading bootstrap JS, it's doing that on other PRs, too.

SherazKhan:
Thanks, @larsoner, and @dengemann. I think we need to discuss this in detail.

PR #7707, opened 2020-05-01
MRG, FIX: Fix path

larsoner:
In #7704 I removed a `SUBJECTS_DIR` env var in `.circleci/config.yml` to ensure it's always set properly in examples. This led to a CircleCI failure:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/1377/workflows/28bc8b26-4c73-4c33-ae3b-1dd76b7f1d20/jobs/19789/steps

This should fix it by passing a path when doing `fetch_hcpmmp_parcellation`.


agramfort:
thx @larsoner

PR #7708, opened 2020-05-01
BF: Improvements to viz.get_3d_backend()

hoechenberger:
This builds upon #7688.

On `master`, I observed the following behavior on a machine without a working 3D backend:

MWE:
```python
import mne

print(mne.viz.get_3d_backend())
print(mne.viz.get_3d_backend())
```

`master`:
```
None
pyvista
```

Reason is that when probing for usable backends, `MNE_3D_BACKEND` is being set _before the probing is carried out_, so when it fails, the variable has already been changed, and is not reset to its previous state. df2e395 fixes the behavior and only changes `MNE_3D_BACKEND` is probing the 3D backend was successful:
```
None
None
```

<strike>I also realized that `viz.get_3d_backend()` can be simplified much more: since `MNE_3D_BACKEND` should always contain the currently used backend, `get_3d_backend()` can simply return this value. There's no need to invoke `_get_3d_backend()`. These changes are implemented in 2729896</strike>

### DIFF
--- a/mne/viz/backends/renderer.py
+++ b/mne/viz/backends/renderer.py
@@ -135,13 +135,12 @@ def _get_3d_backend():
         MNE_3D_BACKEND = get_config(key='MNE_3D_BACKEND', default=None)
         if MNE_3D_BACKEND is None:  # try them in order
             for name in VALID_3D_BACKENDS:
-                MNE_3D_BACKEND = name
                 try:
                     _reload_backend(name)
                 except ImportError:
-                    pass
+                    continue
                 else:
-                    break
+                    MNE_3D_BACKEND = name
             else:
                 raise RuntimeError('Could not load any valid 3D backend: %s'
                                    % (VALID_3D_BACKENDS))
--- a/mne/viz/backends/renderer.py
+++ b/mne/viz/backends/renderer.py
@@ -121,11 +121,7 @@ def get_3d_backend():
         The 3d backend currently in use. If no backend is found,
         returns ``None``.
     """
-    try:
-        backend = _get_3d_backend()
-    except RuntimeError:
-        return None
-    return backend
+    return MNE_3D_BACKEND
 
 
 def _get_3d_backend():
--- a/mne/viz/backends/renderer.py
+++ b/mne/viz/backends/renderer.py
@@ -137,6 +137,7 @@ def _get_3d_backend():
                     continue
                 else:
                     MNE_3D_BACKEND = name
+                    break
             else:
                 raise RuntimeError('Could not load any valid 3D backend: %s'
                                    % (VALID_3D_BACKENDS))
--- a/mne/viz/backends/renderer.py
+++ b/mne/viz/backends/renderer.py
@@ -121,7 +121,11 @@ def get_3d_backend():
         The 3d backend currently in use. If no backend is found,
         returns ``None``.
     """
-    return MNE_3D_BACKEND
+    try:
+        backend = _get_3d_backend()
+    except RuntimeError:
+        return None
+    return backend
 
 
 def _get_3d_backend():
--- a/mne/viz/backends/tests/test_renderer.py
+++ b/mne/viz/backends/tests/test_renderer.py
@@ -10,7 +10,6 @@ import os
 import pytest
 import numpy as np
 
-import mne
 from mne.viz.backends.tests._utils import (skips_if_not_mayavi,
                                            skips_if_not_pyvista)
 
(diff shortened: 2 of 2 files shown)


mne/viz/backends/tests/test_renderer.py
```diff
@@ -152,3 +153,25 @@ def test_3d_backend(renderer):
...
+    mne.viz.backends.renderer.MNE_3D_BACKEND = None
+    assert mne.viz._get_3d_backend() is None
+    assert mne.viz._get_3d_backend() is None
+
+    # Now test the public function.
+    mne.viz.backends.renderer.MNE_3D_BACKEND = None
+    assert mne.viz.get_3d_backend() is None
+
+    mne.viz.set_3d_backend('pyvista')
+    assert mne.viz.get_3d_backend() == 'pyvista'
```
larsoner, line 174: Your code as-is is a bit unsafe. For example if this `assert` fails, now the 3D backend has been set for any other tests that also run. Any time you modify a global state or var you should enclose it effectively in a `try/except` or equivalent pattern (context manager, Pytest fixture). Here I'd probably use some pytest magic:
```
@pytest.fixture()
def restore_backend():
    import mne
    orig_backend = mne.viz.backends.renderer
    yield
    mne.viz.backends.renderer = orig_backend


def test_get_3d_backend(restore_backend):
    ...
```
Then pytest will ensure that no matter what happens, it will get set back.

hoechenberger: Very good point, will try to adopt this.

hoechenberger: Adressed in 2adeb3d. Turns out there was already a fixture I could reuse.

mne/viz/backends/renderer.py
```diff
@@ -121,11 +121,7 @@ def get_3d_backend():
         The 3d backend currently in use. If no backend is found,
         returns ``None``.
     """
-    try:
-        backend = _get_3d_backend()
-    except RuntimeError:
-        return None
-    return backend
+    return MNE_3D_BACKEND
```
larsoner, line 124: It looks like this will no longer cause any checking to actually be done. And indeed it appears to be the case -- on `master` this seems correct:
```
$ python -c "import mne; print(mne.viz.get_3d_backend())"
mayavi
$ MNE_3D_BACKEND=pyvista python -c "import mne; print(mne.viz.get_3d_backend())"
pyvista
```
On this PR it does not:
```
$ python -c "import mne; print(mne.viz.get_3d_backend())"
None
$ MNE_3D_BACKEND=pyvista python -c "import mne; print(mne.viz.get_3d_backend())"
None
```

hoechenberger: Oh. I always just set the backend via `viz.set_3d_backend()`. Need to rethink this entire PR then.

hoechenberger:
@larsoner I removed most of the test in 787c7d9 in favor of an already-existing test:

https://github.com/mne-tools/mne-python/blob/30646229f23cee554884fe144726baf7947ff0a1/mne/viz/backends/tests/test_renderer.py#L26-L41

mne/viz/backends/renderer.py
hoechenberger: I reverted that particular change, could you tell me if it works for you now as expected?

larsoner: Yes it works again

mne/viz/backends/tests/test_renderer.py
```diff
@@ -152,3 +153,11 @@ def test_3d_backend(renderer):
...
     rend.reset_camera()
     rend.show()
+
+
+def test_get_3d_backend(backend_mocker):
+    """Test get_3d_backend private function call for side-effects."""
+    # Test twice to ensure the first call had no side-effect
+    mne.viz.backends.renderer.MNE_3D_BACKEND = None
+    assert mne.viz._get_3d_backend() is None
+    assert mne.viz._get_3d_backend() is None
```
GuillaumeFavelier, line 163: I suggest to migrate to local import of `renderer` in the test to have better control over the backends. The function `get_3d_backend` will try all the valid backends in the list `VALID_3D_BACKENDS`, the default being `mayavi` so testing with `None` will not work. This is my suggestion:

```py
def test_get_3d_backend():
    """Test get_3d_backend private function call for side-effects."""
    # Test twice to ensure the first call had no side-effect
    from mne.viz.backends import renderer
    backend_name = renderer.MNE_3D_BACKEND
    assert renderer.get_3d_backend() == backend_name
    assert renderer.get_3d_backend() == backend_name
```

GuillaumeFavelier: Or using the `renderer` fixture:

```py
def test_get_3d_backend(renderer):
    """Test get_3d_backend private function call for side-effects."""
    # Test twice to ensure the first call had no side-effect
    backend_name = renderer.MNE_3D_BACKEND
    assert renderer.get_3d_backend() == backend_name
    assert renderer.get_3d_backend() == backend_name
```

GuillaumeFavelier:
Also thank you @hoechenberger for taking care of this :)

mne/viz/backends/tests/test_renderer.py
hoechenberger: You are right. I only thought about this from the perspective of my test environment, where I had uninstalled mayavi and pyvista. Of course if they're available, my proposed test setup doesn't make too much sense. Will look into this this afternoon

hoechenberger:
The failing test seems to be unrelated…

larsoner:
Indeed, I'll open a PR to fix that

larsoner:
@GuillaumeFavelier feel free to merge if you're happy

GuillaumeFavelier:
Thanks @hoechenberger !

PR #7709, opened 2020-05-01
[MRG] Adding temporary fix for montage transformation.

adam2392:
#### Reference issue
Partially addresses problem in #7691 

Adds a boolean parameter to `set_montage` to allow head transformation of the montage to be turned off. This allows 'unknown' montages to be set on the Raw data structure.

#### Additional information
Defaults to true for backwards compatibility.


agramfort:
@adam2392 one cannot merge this one too fast. Please use it for your tests using sEEG and see how   it flies.

PR #7016, opened 2019-11-04
FIX: anonymize annotations

bloyl:
Correctly set `annotations.orig_time` from `meas_date`.

Thanks @alexrockhill for noticing it. should help with https://github.com/mne-tools/mne-bids/pull/280

@alexrockhill can you see if it solves your issue?

PR #6892, opened 2019-10-03
ENH: Basic anonymization

bloyl:
This modifies the way we were anonymizing meas_info objects to mainly permit time shifting dates.
It also adds an executable for anonymizing raw fif files from the command line.

The goal was to match the current plans in mne_cpp (https://github.com/mne-tools/mne-cpp/pull/393) which is still under development. hence the [WIP] tag.

closes  #6844 and #4982

@LorenzE

PR #7711, opened 2020-05-02
fix set_annotation with anonymize data

agramfort:
aims to close https://github.com/mne-tools/mne-bids/issues/403

cc @larsoner @hoechenberger 

this is still WIP as it's no so clear why this is needed. There is something fishy/inconsistent
somewhere.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -19,6 +19,8 @@ Changelog
 
 - Add support for signals in mV for :func:`mne.io.read_raw_brainvision` by `Clemens Brunner`_
 
+- Fixed bug with events_from_annotations when ``info['meas_date'] is None`` and ``first_samp > 0`` by `Alex Gramfort`_
+
 - :meth:`mne.Epochs.plot_psd_topomap` and :func:`mne.viz.plot_epochs_psd_topomap` now allow joint colorbar limits across subplots, by `Daniel McCloy`_.
 
 - Add :class:`mne.MixedVectorSourceEstimate` for vector source estimates for mixed source spaces, by `Eric Larson`_
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -1008,7 +1008,9 @@ def events_from_annotations(raw, event_id="auto",
 
     if chunk_duration is None:
         inds = raw.time_as_index(annotations.onset, use_rounding=use_rounding,
-                                 origin=annotations.orig_time) + raw.first_samp
+                                 origin=annotations.orig_time)
+        if raw.info['meas_date'] is not None:
+            inds += raw.first_samp
 
         values = [event_id_[kk] for kk in annotations.description[event_sel]]
         inds = inds[event_sel]
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -1026,7 +1026,8 @@ def events_from_annotations(raw, event_id="auto",
                 _inds = raw.time_as_index(_onsets,
                                           use_rounding=use_rounding,
                                           origin=annotations.orig_time)
-                _inds += raw.first_samp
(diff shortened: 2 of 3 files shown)


hoechenberger:
This does inded seem to fix the issue. I've tested with the localizer datast mentioned in the original issue, and with MNE's sample data `MEG/sample/sample_audvis_raw.fif`, and both seem to work!

agramfort:
basically it's a nasty bug that occurs when meas_date is None and first_samp > 0 ...

@hoechenberger one should not have this issue with BIDS data as meas_date should be shifted in the  past when converting to BIDS and not removed. Pb is due to the fact that data was anonymized with mne before the days_back logic of raw.anonymize was implemented.

ok for you @larsoner ? we should ideally backport this...

hoechenberger:
> @hoechenberger one should not have this issue with BIDS data as meas_date should be shifted in the past when converting to BIDS and not removed. Pb is due to the fact that data was anonymized with mne before the days_back logic of raw.anonymize was implemented.

Thanks for the explanation, I need to read up on the `days_back` thing, never really understood how it works (but hopefully soon!)

agramfort:
have a look at https://github.com/mne-tools/mne-python/pull/6892/files
https://github.com/mne-tools/mne-python/pull/7016/files
for some background

>

hoechenberger:
> have a look at https://github.com/mne-tools/mne-python/pull/6892/files https://github.com/mne-tools/mne-python/pull/7016/files for some background

Thanks @agramfort

hoechenberger:
@agramfort I just discovered that this PR doesn't solve the entirety of the issue.

So far I had only looked at event sample numbers, which were off in `master`, and your PR fixes those.

Now I looked into the `Annotations` onset times, and there's still a problem. Have a look at this MWE:

```python
from tempfile import TemporaryDirectory
import mne
from mne_bids import make_bids_basename, read_raw_bids, write_raw_bids

mne.set_log_level(verbose=False)


subject = 'SB01'
raw_fname = 'SB01_Localizer_raw.fif'
raw = mne.io.read_raw_fif(raw_fname, allow_maxshield=True)
events = mne.find_events(raw, min_duration=2/raw.info['sfreq'])
bids_basename = make_bids_basename(subject=subject)
bids_fname = bids_basename + '_meg.fif'

with TemporaryDirectory() as bids_root:
    write_raw_bids(raw=raw,
                   bids_basename=bids_basename,
                   bids_root=bids_root,
                   events_data=events,
                   overwrite=True,
                   verbose=False)

    raw_read = read_raw_bids(bids_fname=bids_fname,
                             bids_root=bids_root,
                             extra_params=dict(allow_maxshield=True),
                             verbose=False)
    events_read, _ = mne.events_from_annotations(raw_read)

print(f'First event sample before writing:   {events[0, 0]}')
print(f'First event sample after reading:    {events_read[0, 0]}')
print(f'… after first_samp subtraction:      {events_read[0, 0] - raw_read.first_samp}')

print(f'Last time point after reading:       {raw_read.times[-1]}')
print(f'Last annotation onset after reading: {raw_read.annotations.onset[-1]}')
print(f'… after first_time subtraction:      {raw_read.annotations.onset[-1] - raw_read.first_time}')
```

Produces:
```
First event sample before writing:   24863
First event sample after reading:    24863
… after first_samp subtraction:      2363
Last time point after reading:       270.7495
Last annotation onset after reading: 281.966
… after first_time subtraction:      270.716
```

hoechenberger:
@agramfort How should we proceed here?

agramfort:
no time to look today :(

>

hoechenberger:
> no time to look today :(

No worries, it's not urgent. Let me know if I can be of any help

PR #7725, opened 2020-05-04
MRG, FIX: Fix MRI orientations

### LINKED ISSUE #7221: Orientation in plot_bem
#### Describe the bug
Two things. The first one is more of a feature request: It might be nice to have an optional mri arguments to mne.viz.plot_bem to explicitly pass the location of the MRI file to use in mne.viz.plot_bem.

Second, in the current code of MNE, the orientation argument of mne.viz.plot_bem only produce correct visualization when the MRI volumes are in the "conform" space (e.g, running  mri_convert with --conform). E.g., here is the result with orientation == coronal for a volume that is not in the "conform" space:

image

To the extent that the volumes contain the metadata for the orientation, the plot_bem function should check the orientation metadata and either 1) throw an exception if using non-conform volumes would cause to many issues or 2) give an appropriate view of the volume if managing internally non-conform volumes is reasonable for MNE. I don't think the volume should be made "conform" by MNE, unless MNE also manage rotating all the other artifacts (e.g., BEM surfaces) so that they stay in registration with the volume.

Question: Are there other functions that would be impacted by non "conform" volume?

larsoner:
Closes #7221.

@christian-oreilly can you try it with your MRIs that are in non-standard orientation? I was able to check that the FreeSurfer-standard ones look okay, but don't have good MRIs to test for non-standard orientations. If you don't have them I could maybe resample these and regenerate BEMs but that sounds like too much work :)

This actually makes our code simpler and easier to read I think, which is nice.

### DIFF
--- a/mne/viz/misc.py
+++ b/mne/viz/misc.py
@@ -28,9 +28,10 @@ from ..surface import read_surface
 from ..io.proj import make_projector
 from ..io.pick import (_DATA_CH_TYPES_SPLIT, pick_types, pick_info,
                        pick_channels)
-from ..source_space import read_source_spaces, SourceSpaces, _read_mri_info
+from ..source_space import (read_source_spaces, SourceSpaces, _read_mri_info,
+                            _check_mri)
 from ..transforms import invert_transform, apply_trans
-from ..utils import (logger, verbose, get_subjects_dir, warn, _check_option,
+from ..utils import (logger, verbose, warn, _check_option,
                      _mask_to_onsets_offsets, _pl)
 from ..io.pick import _picks_by_type
 from ..filter import estimate_ringing_samples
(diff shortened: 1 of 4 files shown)


hoechenberger:
friendly ping to @christian-oreilly

mne/viz/misc.py
```diff
@@ -300,36 +301,28 @@ def _plot_mri_contours(mri_fname, surfaces, src, orientation='coronal',
...
+    axcodes = axcodes.replace('L', 'R').replace('P', 'A').replace('I', 'S')
     if orientation == 'coronal':
-        x, y, z = 0, 1, 2
+        x, y, z = axcodes.index('R'), axcodes.index('S'), axcodes.index('A')
     elif orientation == 'axial':
-        x, y, z = 2, 0, 1
+        x, y, z = axcodes.index('A'), axcodes.index('R'), axcodes.index('S')
     else:  # orientation == 'sagittal'
-        x, y, z = 2, 1, 0
+        x, y, z = axcodes.index('A'), axcodes.index('S'), axcodes.index('R')
```
jona-sassenhagen, line 320: I think I would do

```
orientation_to_xyz = {'coronal': ('R', 'S', 'A'), 'axial': ('A', 'R', 'S'), 'sagittal': ('A', 'S', 'R')}
x, y, z = [axcodes.index(coordinate) for coordinate in orientation_to_xyz[orientation]]
```

larsoner: Indeed that's nicer, pushing now

agramfort:
thx @larsoner

christian-oreilly:
@larsoner  Sorry for the late feedback. I do not have the MRI that prompted me to submit the issue #7221 anymore because I went around the issue by preprocessing all my MRI to a "conform" space in the meantime. But I always end up having to move data between pipelines and almost always run into issues with incompatible coordinate systems. So I found some MRI from a different project (originally processed in CIVET rather than FreeSurfer) that were not preprocessed to a conform to space and that is what I got:

Before to PR:
Screenshot from 2020-05-11 12-52-06

After the PR:
Screenshot from 2020-05-11 12-50-50

So... the fix works to the extent that it puts the MRI in coronal view when it is supposed to put it in this view. I think the rest is more about what level of user-friendliness you want to support (i.e., whether or not you would like to have it 100% aware of coordinate system so that it would invert axes if needed to display it in a standard orientation or leave it to the user to provide their MRI in a given coordinate system). At least, right now, it does the volume along the requested axis.

FYI, there appears to be some bug in the management of paths:

```
In [6]: import mne                                                              

In [7]: import os                                                               

In [8]: mne.viz.plot_bem("ShNe_V1_anlm0.5r", os.environ["SUBJECTS_DIR"])        
Using surface: /usr/local/freesurfer/subjects/ShNe_V1_anlm0.5r/bem/inner_skull.surf
Using surface: /usr/local/freesurfer/subjects/ShNe_V1_anlm0.5r/bem/outer_skull.surf
Using surface: /usr/local/freesurfer/subjects/ShNe_V1_anlm0.5r/bem/outer_skin.surf
Out[8]: <Figure size 780x706 with 12 Axes>

In [9]: mne.viz.plot_bem("ShNe_V1_anlm0.5r")                                    
<ipython-input-9-1980c44bfbfe> in <module>
----> 1 mne.viz.plot_bem("ShNe_V1_anlm0.5r")

~/code/mne-python/mne/viz/misc.py in plot_bem(subject, subjects_dir, orientation, slices, brain_surfaces, src, show, show_indices, mri)
    460 
    461     # Get the BEM surface filenames
--> 462     bem_path = op.join(subjects_dir, subject, 'bem')
    463 
    464     if not op.isdir(bem_path):

~/anaconda3/lib/python3.7/posixpath.py in join(a, *p)
     78     will be discarded.  An empty last part will result in a path that
     79     ends with a separator."""
---> 80     a = os.fspath(a)
     81     sep = _get_sep(a)
     82     path = a

TypeError: expected str, bytes or os.PathLike object, not NoneType
```

The call at "In [9]" should works as, according to the docstring, the subjects_dir argument can be left to None, in which case "the path is obtained by using the environment variable SUBJECTS_DIR." This error seems to have been introduced with this fix because it works fine when I revert to the MNE version I had before applying this fix.

larsoner:
Indeed I can see how this error was created. I can fix the flipping problem more easily if you can share their T1.mgz and `bem` directory with me privately...? In the meantime I can work on a fix either way

christian-oreilly:
@larsoner Sure, I'll sent these to you.

larsoner:
@christian-oreilly can you first try #7763 ? I think I only need the files if that PR does not work.

christian-oreilly:
@larsoner Sure, I'll do.

christian-oreilly:
@larsoner Yep, works fine! Both the orientation and the path thing. Nice work! :)

PR #7701, opened 2020-04-30
TST: Speedup testing of pyvista backend

GuillaumeFavelier:
This PR disables anti-aliasing, line smoothing, smooth shading and decreases the polygon counts of tubes during testing of the pyvista 3d backend. Hopefully this will decrease testing time especially on macOS.

PR #7726, opened 2020-05-04
MRG: Speedup testing of _Brain

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/7701#issuecomment-623471954 idea and tries to decrease the number of calls to `_Brain.__init__()` which seems costly. Hopefully this will reduce testing time especially on macOS.

### DIFF
--- a/mne/viz/_brain/tests/test_brain.py
+++ b/mne/viz/_brain/tests/test_brain.py
@@ -76,19 +76,13 @@ def test_brain(renderer):
         _Brain(subject_id=subject_id, hemi='foo', surf=surf)
 
     # cortex
-    for cortex in ['classic', 'low_contrast', 'high_contrast', 'bone']:
-        brain = _Brain(subject_id, hemi=hemi, surf=surf, size=size,
-                       subjects_dir=subjects_dir, title='test',
-                       cortex=cortex)
-        brain.close()
-
+    brain = _Brain(subject_id, hemi=hemi, surf=surf, size=size,
+                   subjects_dir=subjects_dir, title='test',
+                   cortex='low_contrast')
     # add_data
     stc = read_source_estimate(fname_stc)
     fmin = stc.data.min()
     fmax = stc.data.max()
-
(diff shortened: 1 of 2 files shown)


mne/viz/_brain/tests/test_brain.py
```diff
@@ -71,114 +75,81 @@ def test_brain_init(renderer):
     with pytest.raises(KeyError):
         _Brain(subject_id=subject_id, hemi='foo', surf=surf)
 
+    # cortex
     for cortex in ['classic', 'low_contrast', 'high_contrast', 'bone']:
```
larsoner, line 79: This does 4 brain inits, `'classic'` is the default (right?) so maybe just test one of these other ones?

mne/viz/_brain/tests/test_brain.py
```diff
@@ -71,114 +75,81 @@ def test_brain_init(renderer):
...
-    hemi = 'lh'
-    hemi_data = stc.data[:len(stc.vertices[0]), 10]
-    hemi_vertices = stc.vertices[0]
     fmin = stc.data.min()
     fmax = stc.data.max()
 
-    brain_data = _Brain(subject_id, hemi, surf, size=300,
-                        subjects_dir=subjects_dir)
+    brain = _Brain(subject_id, hemi=hemi, surf=surf, size=size,
+                   subjects_dir=subjects_dir)
```
larsoner, line 91: ... then don't bother creating a new `brain` here just reuse the one above

larsoner:
Next targets (another PR someday) might be:
```
24.02s teardown mne/viz/tests/test_3d.py::test_plot_sparse_source_estimates[pyvista]
17.28s teardown mne/viz/tests/test_3d.py::test_link_brains[pyvista]
```
That is a very long time for a teardown. I guess this means that closing the windows is taking forever (?). Maybe for some reason calling close causes a series of draws to happen. Not sure.

In any case, this is already good progress, thanks @GuillaumeFavelier !

PR #7728, opened 2020-05-04
MRG, FIX: Fix volume morph with multiple labels

larsoner:
See `latest.inc`, basically only the first volume label would be interpolated when morphing a source space with multiple subvolumes and using `mri_resolution=True`.

This was found while working toward fixing #7564, so this has a bunch of stubs where support for `'mixed'` will eventually need to be added (it actually doesn't look so bad!).


agramfort:
thanks @larsoner

PR #7729, opened 2020-05-04
MRG, FIX: Fix for complex

larsoner:
Fixes a problem found by @drammock working with complex Evoked data where the variance explained computation / message would give a warning, which made it clear it was doing something wrong.


agramfort:
thx @larsoner

PR #7731, opened 2020-05-05
MRG: Add ability to load NIRX data by passing path to header file

rob-luke:
#### Reference issue
https://github.com/cbrnr/mnelab/issues/144#issuecomment-623066162

Allows user to pass the path to a NIRX header file to load data, rather than passing the path to the directory.

This was suggested as a way to easily add data loading to MNELAB (see comment above).

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -5,6 +5,7 @@
 from configparser import ConfigParser, RawConfigParser
 import glob as glob
 import re as re
+import os.path as op
 
 import numpy as np
 
@@ -23,7 +24,7 @@ def read_raw_nirx(fname, preload=False, verbose=None):
     Parameters
     ----------
     fname : str
-        Path to the NIRX data folder.
+        Path to the NIRX data folder or header file.
     %(preload)s
     %(verbose)s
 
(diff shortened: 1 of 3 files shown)


mne/io/nirx/nirx.py
```diff
@@ -36,6 +37,10 @@ def read_raw_nirx(fname, preload=False, verbose=None):
     --------
     mne.io.Raw : Documentation of attribute and methods.
     """
+    if fname.endswith('.hdr'):
+        return read_raw_nirx(op.dirname(op.abspath(fname)),
+                             preload=preload, verbose=verbose)
+
     return RawNIRX(fname, preload, verbose)
```
agramfort, line 44: I woudl do this RawNIRX init and do:

```
if fname.endswith('.hdr'):
    fname  = op.dirname(op.abspath(fname))
```

rob-luke: Thanks

cbrnr:
Can you also add a what's new entry please?

rob-luke:
@larsoner could you please review. I think the CI failure is unrelated.

agramfort:
thx @rob-luke

Issue #7732, opened 2020-05-05
Bug: event_repeated metadata incompatibility


kingjr:
`mne.Epochs` `event_repeated` parameter does not play well `metadata`:

```python
import mne
import numpy as np
import pandas as pd
raw_data = np.random.randn(2, 10000)
info = mne.create_info(2, 1000.)
raw = mne.io.RawArray(raw_data, info)

opts = dict(raw=raw, tmin=0, tmax=.001, baseline=None)
```

No problem when there are no repeated events and metadata
```python
events = [[0, 0, 1], [1, 0, 1]]
metadata = pd.DataFrame(events, columns=['onset', 'duration', 'value'])

epochs = mne.Epochs(events=events, metadata=metadata, **opts)
```

No problem when repeated events and no metadata
```python
events = [[1, 0, 1], [1, 0, 1]]
epochs = mne.Epochs(events=events, event_repeated='drop', **opts)
```

problem when repeated events and metadata
```python
events = [[1, 0, 1], [1, 0, 1]]
epochs = mne.Epochs(events=events, event_repeated='drop', metadata=metadata, **opts)
```

kingjr:
BTW, there should be an event_repeated='ignore' option. 

It is common to have distinct events types which end up with simultaneous onsets e.g. :
- during streamed stimulation,  button presses may be simultaneous to stim onset; 
- the beginning of words if always the beginning of a phoneme.

In such cases I don't think there is a good reason to force users to build different `Epochs` objects for each event type.

agramfort:
event_repeated='ignore' would lead to a duplicated epoch right?

if so no objection if it's well documented

>

kingjr:
duplicate in terms of m/eeg, but not in terms of metadata

agramfort:
so you expect the sample idx to be like an index in pandas sense in the
metadata?

kingjr:
not sure what you have in mind exactly; let me clarify my view

IMO the events and metadata should in principle be a unique object, i.e. a
sparse time series. Historically, they have been built separately but they
should remain connected during e.g. drop_bad, epoch indexing etc.

The ideal case would be for metadata  to be explicitly indexed by the event
onset (in raw sample, i.e. events[:,0]), but I *think* that would create
some backward incompatibilities.

The alternative is to link the events rows and the metadata iloc. One risk
I foresee with pd dataframes is the loc versus iloc indexing and the
reset_index mess, which could confuse users: dealing with pandas
[...]

PR #7733, opened 2020-05-05
MRG, FIX: Fix bug with metadata and event_repeated

larsoner:
Addresses the first part of #7732

@kingjr can you try it?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -120,6 +120,8 @@ Bug
 
 - Fix bug with :meth:`mne.Epochs.save` where the file splitting calculations did not account for the sizes of non-data writes by `Eric Larson`_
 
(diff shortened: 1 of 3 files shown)


kingjr:
The `merge` behavior is kind of weird for the metadata, but I think that's a separate and non-trivial issue.

I would still recommend to add the repeated_events='ignore' option, but this is also separate.

Else LGTM

larsoner:
> The merge behavior is kind of weird for the metadata, but I think that's a separate and non-trivial issue.

Indeed, maybe we should not allow merging when metadata is present.

> I would still recommend to add the repeated_events='ignore' option, but this is also separate.

Let's leave #7732 open to track and discuss this (it's why I didn't add the "closes" tag to the PR description)

PR #7737, opened 2020-05-05
MRG: Motion Annotation

bloyl:
`mne.preprocessing.annotate_movement` computes HPI displacements from an average position. Bad data segments (in `raw.annotations`) where meant to be excluded from this averaging. However there is a bug in computing the weights for averaging

Basically the weights (`seg_good`) are defined from `pos` which has different time samples from `raw`. So the onsets/durations of the bad annotations do line up with the indices of the `seg_good` array.

This PR uses `compute_average_dev_head_t` to compute an average transform (ignoring bad segments) and uses that to compute the average HPI locations.

Additionally, I'd like to propose adding a flag `use_raw_head_trans` (or similarly named) that if set would use `raw.info['dev_head_t']` as opposed to `compute_average_dev_head_t`. The default would be `False`

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -127,7 +127,6 @@ API
 
 - The method ``stc_mixed.plot_surface`` for a :class:`mne.MixedSourceEstimate` has been deprecated in favor of :meth:`stc.surface().plot(...) <mne.MixedSourceEstimate.surface>` by `Eric Larson`_
 
-- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head
-transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
+- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
 
 - The function ``mne.channels.read_dig_captrack`` will be deprecated in version 0.22 in favor of :func:`mne.channels.read_dig_captrak` to correct the spelling error: "captraCK" -> "captraK", by `Stefan Appelhoff`_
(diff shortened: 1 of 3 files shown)


larsoner:
> Additionally, I'd like to propose adding a flag use_raw_head_trans (or similarly named) that if set would use raw.info['dev_head_t'] as opposed to compute_average_dev_head_t. The default would be False

Maybe `use_dev_head_t='average' (default) | 'info'`? That way we can add other options later if need be, and it's fairly readable.

AdoNunes:
Nice catch @bloyl.

I am just worried that the same could happen in the ``compute_average_dev_head_t``.

In function example, the pos had the same number of samples as the raw. I guess if the fit is not good it will remove pos samples? If so ``compute_average_dev_head_t`` would be wrong or crash.

```
    # Get time indices
    ts = np.concatenate((hp[:, 0], [(raw.last_samp + 1) / sfreq]))
    assert (np.diff(ts) > 0).all()
    ts -= raw.first_samp / sfreq
    idx = raw.time_as_index(ts, use_rounding=True)

    assert idx[-1] == len(seg_good)
```
Here the ts (pos time stamps) are converted to idx of the raw samples

```
   # Mark times bad that are bad according to annotations
    onsets, ends = _annotations_starts_stops(raw, 'bad')
    for onset, end in zip(onsets, ends):
        seg_good[onset:end] = 0
    dt = np.diff(np.cumsum(np.concatenate([[0], seg_good]))[idx])
```
then, ``seg_good`` initially ones are zeroed if bad_annotation.  The good time periods indexed by `idx` are used for computing the average dev_head_t.

OK, after writing this I think the function is fine. 
LGTM

bloyl:
What about this as a docstring for the new parameter?
```
    use_dev_head_trans : 'average' (default) | 'info'
        Identify the device to head transform used to define the
        fixed HPI locations for computing moving distances.
        If ``average`` the average device to head transform is
        computed using ``compute_average_dev_head_t``.
        If ``info``, ``raw.info['dev_head_t']`` is used.
```

I'm not sure I'm using the `` correctly?

larsoner:
Backticks look correct and description LGTM

bloyl:
Codecov failed because of a testing timeout. I restarted the test but I'm not sure how to redo codecov.

larsoner:
Thanks @bloyl !

PR #7740, opened 2020-05-06
MRG, FIX: Fix epochs split size

larsoner:
From the mailing list.


ualsbombe:
Fixed my issue

_Dear list

Every now and then I cannot save my epochs objects due to this error. This happens only for one of my subjects (out of thirty) and only on one particular (bandpass filtered and hilbert transformed) raw file. I have tried both the development version and newest stable version.

I guess it should take care of it automatically as I didn't change the default split size of 2GB of epochs_hilbert.save()

All the best

Lau_

PR #7717, opened 2020-05-03
MRG, ENH: import code for optical imaging data from BOXY, fNIRS and fast signal

kylemath:
#### Reference Issue
Continuing from discussion in https://github.com/mne-tools/mne-python/issues/7057 regarding importing of other file formats of optical imaging data.
Issue https://github.com/mne-tools/mne-python/issues/7718 introduced to discuss the goals and issues as well.

#### What does this implement/fix?
This creates io functionality for BOXY files created for the Imagent and Imagent 2 from ISS, which collect both overall and modulated light levels, but also phase delay measures used to measure fast scattering changes in neural tissue.

#### Additional information
Some important discussions to have regarding 

- [x] data types (3 data types, overall light (DC), modulated light (AC), and phase delay (Ph), from each channel)

- [x] Wavelength

- [x] Modulation frequency - new imagent 2 machine has 4 modulation frequencies of light, which are used to measure AC and Ph in the measured signal, and may be relevant for light modeling and analysis

- [x] Channel names - are based on a letter/number grid on our headgear, but need to be flexible

- [x] digitization and coregistration - common head, individual heads, fiducial alignment, other head points, sensor locations/midpoints

- [x] markers from data files

- [x] importing multiple files per participant (old systems use on file for each detector), all files use one file for each montage (arrangement of sources and detectors), usually collected in separate sessions, for both these cases only a single digitization file is needed, but different files of source and detector locations used.

- [x] renaming files - BOXY outputs time stamped files, and usually matlab script is used to rename files by subject number, experiment, etc., build into this? 

- [x] preprocessing functions - heart beat removal - regression based heart beat removal for fast signal needed as pulse wave spectra overlaps with fast signals (Corballis and Gratton) (ppod software - https://code.google.com/archive/p/p-pod/)

- [x] preprocessing functions - phase wrapping correction - phase signal is modified to a continuous signal (ppod software - https://code.google.com/archive/p/p-pod/)

References:
boxy optical imaging workflow:
http://sites.psych.ualberta.ca/kylemathewson/optical-imaging/

PR #7746, opened 2020-05-07
x

kylemath:
## Reference issue
Example: Addressing discussion in other PR on loading BOXY data here: https://github.com/mne-tools/mne-python/pull/7717.

In addition to plotting the midpoint of source and detector, the source and detector locations should also be plot-able, to evaluate coregistration. Plotting is done in plot_alignment based on other_picks.

Example of desired output currently working in first pass, black are detectors, red sources, midpoints not shown: (-Colours - what is reserved, what is good?, )
Screen Shot 2020-05-06 at 11 56 00 PM

First try by creating two new keys in other_loc based on other points, resetting other_keys, and adding colours and scales to the DEFAULTS

```py 
    for key, picks in other_picks.items():
        if other_bools[key] and len(picks):
            if key == 'fnirs':
                # other_loc[key] = np.array([info['chs'][pick]['loc'][:3]
                #                for pick in picks])
                other_loc['source'] = np.array([info['chs'][pick]['loc'][3:6]
                                               for pick in picks])
                other_loc['detector'] = np.array([info['chs'][pick]['loc'][6:9]
                                               for pick in picks])
                other_keys = sorted(other_loc.keys())
                logger.info('Plotting %d %s location%s'
                            % (len(other_loc[key]), key, _pl(other_loc[key])))
```

#### Additional information

Potential API needed to add:
@rob-luke says:
Currently plot alignment takes the arguments fnirs=('pairs', 'channels'). You could open another PR to extend this to fnirs=('pairs', 'channels', 'sources', 'detectors')?

@rob-luke says: 
Currently channel plotting uses other_picks, you would need to extend the code to plot ['loc'][3:6], and ['loc'][6:9] based on if sources or detectors was passed somewhere around here https://github.com/mne-tools/mne-python/blob/master/mne/viz/_3d.py#L1039

#### Other ideas outside scope
- perhaps think about plotting banana shaped path between source and detector instead of shortest path (easiest way is the 3d shape made by a string of spheres: https://github.com/kylemath/MOpt3d/blob/master/setuplookup.m)

- perhaps think about plotting path along scalp curve instead of shortest linear path


kylemath:
ok this one is cleaner in terms of files changed, but still has some incorrect commit history from the other PR attached to it I can't seem to remove, @rob-luke

PR #7747, opened 2020-05-07
MRG, ENH - Plot optodes in plot_alignment for fNIRS channels

kylemath:
## Reference issue
Example: Addressing discussion in other PR on loading BOXY data here: https://github.com/mne-tools/mne-python/pull/7717.

In addition to plotting the midpoint of source and detector, the source and detector locations should also be plot-able, to evaluate coregistration. Plotting is done in plot_alignment based on other_picks.

Example of desired output currently working in first pass, black are detectors, red sources, midpoints not shown: (-Colours - what is reserved, what is good?, )
Screen Shot 2020-05-06 at 11 56 00 PM

First try by creating two new keys in other_loc based on other points, resetting other_keys, and adding colours and scales to the DEFAULTS

```py 
    for key, picks in other_picks.items():
        if other_bools[key] and len(picks):
            if key == 'fnirs':
                # other_loc[key] = np.array([info['chs'][pick]['loc'][:3]
                #                for pick in picks])
                other_loc['source'] = np.array([info['chs'][pick]['loc'][3:6]
                                               for pick in picks])
                other_loc['detector'] = np.array([info['chs'][pick]['loc'][6:9]
                                               for pick in picks])
                other_keys = sorted(other_loc.keys())
                logger.info('Plotting %d %s location%s'
                            % (len(other_loc[key]), key, _pl(other_loc[key])))
```

#### Additional information

Potential API needed to add:
@rob-luke says:
Currently plot alignment takes the arguments fnirs=('pairs', 'channels'). You could open another PR to extend this to fnirs=('pairs', 'channels', 'sources', 'detectors')?

@rob-luke says: 
Currently channel plotting uses other_picks, you would need to extend the code to plot ['loc'][3:6], and ['loc'][6:9] based on if sources or detectors was passed somewhere around here https://github.com/mne-tools/mne-python/blob/master/mne/viz/_3d.py#L1039

#### Other ideas outside scope
- perhaps think about plotting banana shaped path between source and detector instead of shortest path (easiest way is the 3d shape made by a string of spheres: https://github.com/kylemath/MOpt3d/blob/master/setuplookup.m)

- perhaps think about plotting path along scalp curve instead of shortest linear path


kylemath:
@rob-luke third times a charm, here is a nice clean PR finally as a start for plotting the optodes

rob-luke:
The attached figure looks great.

> black are detectors, red sources, midpoints not shown: (-Colours - what is reserved, what is good?, )

Like I mentioned in the other PR, I think red is an intuitive choice for an infrared emitter. Others would need to comment if black is already reserved.

> perhaps think about plotting banana shaped path

I think this would be more appropriate for MNE-NIRS. Do these "banana" shapes usually represent the most probable photon migration path? If so, would we need to implement https://github.com/mne-tools/mne-nirs/issues/62 first?

> perhaps think about plotting path along scalp curve instead of shortest linear path

Yes, improving the `pairs` option to plot along the scalp would be a nice improvement.

larsoner:
> Others would need to comment if black is already reserved.

Can you look at `plot_alignment` and see if black is used anywhere? Maybe just use a dark gray (0.85 or so?), I doubt we've used that for anything but it should still hopefully look okay against the gray surfaces.

kylemath:
@rob-luke - just thinking of api now, you suggest:  fnirs=('pairs', 'channels', 'sources', 'detectors')

Is the idea with plotting the "channels" midpoint that this is where the data is assignment in the latter head plots in your analysis pipeline?

kylemath:
maybe something like this:
Screen Shot 2020-05-08 at 2 21 43 PM

```py
    if fnirs is True:
        fnirs = ['pairs', 'sources', 'detectors']
```
```py
    for xi, x in enumerate(fnirs):
        _check_option('fnirs[%d]' % xi, x, ('channels', 'pairs', 'sources', 'detectors'))
```

```py
    for key, picks in other_picks.items():
        if other_bools[key] and len(picks):
            if key == 'fnirs':
                if 'channels' in fnirs:
                    other_loc[key] = np.array([info['chs'][pick]['loc'][:3]
                                   for pick in picks])
                if 'sources' in fnirs:
                    other_loc['source'] = np.array([info['chs'][pick]['loc'][3:6]
                                                   for pick in picks])
                if 'detectors' in fnirs:
                    other_loc['detector'] = np.array([info['chs'][pick]['loc'][6:9]
                                               for pick in picks])
                other_keys = sorted(other_loc.keys())
                logger.info('Plotting %d %s location%s'
                            % (len(other_loc[key]), key, _pl(other_loc[key])))
```

kylemath:
@rob-luke - this is ready for an initial review but I think everything wanted is in the current PR

rob-luke:
I would prefer to leave the default plotting option as it is. I think having the additional sources and detectors creates too much visual clutter (particularly when you extend this to multimodal datasets). Further, I think that having pairs and sources and detectors is redundant. So please leave the default as is, and if the users want the additional points they can pass in `sources` and `detectors` to the `fnirs` argument.

You will need to add some tests too. Previously this was done by checking that the correct items were listed in the logs. And there are still seems to be some debugging print statements in the PR.

kylemath:
thanks @rob-luke good feedback, agree on the default changes,
can you point me to where the tests for this function would live

rob-luke:
Tests tend to live in a `test` directory in the same location as the file you are editing. You can also click on the details of the failing tests below it will show you the tests that are failing and the path to them ([here is a link for current failing CI](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=7075&view=logs&j=2e1ef983-081d-5c63-b72a-d9f329364ee0&t=748a1ae9-793d-59c6-b8d5-d00add274aaa&l=196)). Its also handy to look at the commit history of the file you are editing, the previous changes are usually accompanied by changes to tests and documentation that you can mimic.

kylemath:
@rob-luke - fixed the default behaviour and reverted the colour of the channels to the red it was before, fixed bug, and added tests of logs

kylemath:
saw the tests, got my flake8 plugin working, and fixing these style problems now

rob-luke:
> saw the tests, got my flake8 plugin working, and fixing these style problems now

Great. It took me a for attempts to understand all the different tests, but after you've been through it once its easier. Having flake8 and pydocstyle installed locally saved me quite some time too. 

Following the process [here](https://mne.tools/stable/install/contributing.html#github-workflow)... "When you believe a PR is ready to be reviewed and merged, convert it from a draft PR to a normal PR, change its title to begin with MRG, and add a comment to the PR asking for reviews (changing the title does not automatically notify maintainers)." Then people will review your code.

kylemath:
@rob-luke @larsoner I believe this is ready to go, could I please have a review?

mne/defaults.py
```diff
@@ -83,9 +87,7 @@ def _handle_default(k, v=None):
     """Avoid dicts as default keyword arguments.
 
     Use this function instead to resolve default dict values. Example usage::
-
```
rob-luke, line 86: This PR doesn't relate to `_handle_default`, so no need to change the doctoring here.

mne/viz/_3d.py
```diff
@@ -595,9 +595,10 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     seeg : bool
         If True (default), show sEEG electrodes.
     fnirs : str | list | bool | None
-        Can be "channels" or "pairs" to show the fNIRS channel locations or
-        line between source-detector pairs, or a combination like
-        ``('pairs', 'channels')``. True translates to ``('pairs',)``.
+        Can be "channels", "pairs", "detectors", and/or "sources" to show the
+        fNIRS channel locations, optode locations, or line between
+        source-detector pairs, or a combination like ``('pairs', 'channels')``.
+        True translates to ``('pairs', 'sources', 'detectors')``.
```
rob-luke, line 601: True translates to ``('pairs',)``.

rob-luke:
Not required, but you could also modify the tutorial to show off your new visualisation here https://github.com/mne-tools/mne-python/blob/master/tutorials/preprocessing/plot_70_fnirs_processing.py#L44

If so, please also add explanation in the accompanying text for what each colour represents.

kylemath:
@rob-luke - maybe the "channels" should be white to match the pairs ? I think its too confusing with these three red shades:
Screen Shot 2020-05-10 at 6 52 06 PM

mne/defaults.py
kylemath: I am failing tests because of this function

rob-luke:
> @rob-luke - maybe the "channels" should be white to match the pairs ? I think its too confusing with these three red shades:

I agree its a bit too similar now. But I think it will be too hard to distinguish white dots on white lines. If you are looking for other options, what about sources=red, detectors=aqua, channels=black, pairs=white? Or keep similar to as is, but add more contrast between the orange and red.

kylemath:
This is white: 
Screen Shot 2020-05-10 at 8 14 20 PM

kylemath:
This is your suggestion with cyan,
Screen Shot 2020-05-10 at 8 17 02 PM

kylemath:
here is it with our full head montage (24 Sources X 16 channels X 2 sessions):
Screen Shot 2020-05-10 at 8 41 08 PM

kylemath:
I guess I like the white sphere in the middle of the pair as it looks more like the true banana path of light from source to detector, here is how my mopt3d software plots those and the intersection with the brain:

Screen Shot 2020-05-10 at 9 10 06 PM

rob-luke:
Thanks for sharing the figures. After looking at them I definitely don't like the white channels and white pairs, its too hard to see the channels. Currently my preference is for your original colouring (sorry to send you in circles), but somehow accentuating the difference between the red and orange.

Perhaps get the tests passing, then when others review they can weigh in on the colours.

kylemath:
ok thanks,  looks like this for now while I work on tests
Screen Shot 2020-05-10 at 11 50 56 PM

kylemath:
And with that last change the tests will all pass now @larsoner @rob-luke

mne/defaults.py
```diff
@@ -83,9 +87,7 @@ def _handle_default(k, v=None):
...
         scalings = _handle_default('scalings', scalings)
-
```
rob-luke, line 88: As above.

tutorials/preprocessing/plot_70_fnirs_processing.py
```diff
@@ -36,15 +36,17 @@
 # Here we validate that the location of sources-detector pairs and channels
 # are in the expected locations. Source-detector pairs are shown as lines
 # between the optodes, channels (the mid point of source-detector pairs) are
-# shown as dots.
+# shown as orange dots. Source are optionally shown as red dots and detectors
```
rob-luke, line 39: Channels are also not shown by default / are optional.

doc/changes/latest.inc
```diff
@@ -59,6 +59,8 @@ Changelog
 
 - Add ``plot`` option to :meth:`mne.viz.plot_filter` allowing selection of which filter properties are plotted and added option for user to supply ``axes`` by `Robert Luke`_
 
+- Add ``sources`` and ``detectors`` options for :meth:`mne.viz.plot_alignment` allowing plotting of optode locations in addition to channel midpoint ``channels`` and ``path`` between optodes by `Kyle Mathewson`_
```
rob-luke, line 62: Specify this is for fNIRS.

mne/defaults.py
kylemath: If I don't remove this my tests fail?

tutorials/preprocessing/plot_70_fnirs_processing.py
kylemath: added

doc/changes/latest.inc
kylemath: done

mne/defaults.py
larsoner: It shouldn't cause tests to fail to leave these lines in (otherwise all PRs would be failing due to this). Maybe you added some whitespace to them, and that's why flake8 complained?

mne/viz/_3d.py
```diff
@@ -595,9 +595,10 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     seeg : bool
         If True (default), show sEEG electrodes.
     fnirs : str | list | bool | None
-        Can be "channels" or "pairs" to show the fNIRS channel locations or
-        line between source-detector pairs, or a combination like
-        ``('pairs', 'channels')``. True translates to ``('pairs',)``.
+        Can be "channels", "pairs", "detectors", and/or "sources" to show the
+        fNIRS channel locations, optode locations, or line between
+        source-detector pairs, or a combination like ``('pairs', 'channels')``.
+        True translates to ``('pairs')``.
```
larsoner, line 601:
```suggestion
        True translates to ``('pairs',)``.
```
There is a difference, what you wrote is actually a `str` and the suggested change is a `tuple` with one `str` element

mne/viz/_3d.py
```diff
@@ -1038,10 +1044,27 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
...
+                if 'detectors' in fnirs:
+                    other_loc['detector'] = np.array(
+                        [info['chs'][pick]['loc'][6:9]
+                         for pick in picks])
+                    logger.info('Plotting %d %s detector%s'
+                                % (len(other_loc['detector']),
+                                   key, _pl(other_loc['detector'])))
+                other_keys = sorted(other_loc.keys())
+                logger.info('Plotting %d %s location%s'
+                            % (len(other_loc[key]), key, _pl(other_loc[key])))
```
larsoner, line 1067: You have changed the logic here. Previously this would `logger.info` for any `key`, now it only does so for `key == 'fnirs'`. You probably want to unindent this one level

mne/viz/_3d.py
kylemath: fixed

mne/defaults.py
kylemath: ok reverted

mne/viz/_3d.py
kylemath: agreed

kylemath:
@larsoner @rob-luke addressed all those comments in last commit

mne/defaults.py
```diff
@@ -81,11 +85,8 @@
 
 def _handle_default(k, v=None):
     """Avoid dicts as default keyword arguments.
-
```
larsoner, line 84: This does not appear reverted

kylemath:
Sorry about that @larsoner, Should be unchanged from master now, 
thanks for guiding me through slowly,

mne/viz/_3d.py
larsoner, line 1067: This still has not been unindented

larsoner: (GitHub tip: after pushing some commits, you can view the diff yourself to quickly make sure that everything is as it should be before requesting review. I usually do this when making a PR just to make sure there isn't extraneous stuff in there or things that I missed. But even with that sometimes I miss stuff, so no big deal...)

kylemath: now it has sorry, forgot to save file

kylemath:
not sure what happened with the travis test: 
```
FAILED mne/inverse_sparse/tests/test_gamma_map.py::test_gamma_map - assert 3....
Results (1805.46s):
    1794 passed
       2 xpassed
       1 failed
         - mne/inverse_sparse/tests/test_gamma_map.py:46 test_gamma_map
       3 xfailed
      29 skipped
```

larsoner:
Fixed by #7764, I restarted the runs

rob-luke:
The [updated tutorial figure](https://20000-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#view-location-of-sensors-over-brain-surface) looks great. Thanks @kylemath.

larsoner:
Thanks @kylemath !

kylemath:
awesome thanks for the help learning,

PR #7750, opened 2020-05-07
Remove area edge lines in average PSD plots

### LINKED ISSUE #7738: raw.plot_psd: edge line of area plot
when plotting raw psd with `n_fft=len(raw.times), average=True` to get high def in frequency, the edge of the grey shaded area overplots, which creates an ambiguous darker grey area around the normal grey area
image
Here's a zoomed in version:
image

would be nice to be able to disable this edge line in the call to `raw.plot_psd`

hoechenberger:
#### Reference issue
Closes #7738

Do not plot lines on the edges of PSD areas

### DIFF
--- a/tutorials/raw/plot_40_visualize_raw.py
+++ b/tutorials/raw/plot_40_visualize_raw.py
@@ -192,3 +192,4 @@ raw.plot_projs_topomap(colorbar=True)
 # .. LINKS
 #
 # .. _spectral density: https://en.wikipedia.org/wiki/Spectral_density
+
--- a/tutorials/raw/plot_40_visualize_raw.py
+++ b/tutorials/raw/plot_40_visualize_raw.py
@@ -188,7 +188,6 @@ raw.plot_sensors(ch_type='eeg')
 
 raw.plot_projs_topomap(colorbar=True)
 
-# XXX
 ###############################################################################
 # .. LINKS
 #
--- a/tutorials/raw/plot_40_visualize_raw.py
+++ b/tutorials/raw/plot_40_visualize_raw.py
@@ -188,8 +188,8 @@ raw.plot_sensors(ch_type='eeg')
 
 raw.plot_projs_topomap(colorbar=True)
 
+# XXX
 ###############################################################################
 # .. LINKS
 #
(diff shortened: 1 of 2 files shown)


tutorials/raw/plot_40_visualize_raw.py
```diff
@@ -192,3 +192,4 @@
 # .. LINKS
 #
 # .. _spectral density: https://en.wikipedia.org/wiki/Spectral_density
+
```
drammock, line 195: you'll need to revert this before we can merge; it will fail flake8

hoechenberger: Yep, just wanted to ensure the tutorial would be rendered… or would that have happened anyway?

hoechenberger: Oh I see, it doesn't render anything now haha

drammock: yeah, the triggering is determined by comparing the file against master.

What I should have said is "when you touch a tutorial to trigger it to render, a safe way to touch it is to just add/change a line break in a paragraph of text.  Adding blank lines in the middle of the document (e.g., in between paragraphs or between a paragraph and a code block) should also be safe.  Blank lines at the end make flake angry, and `XXX` comments will show up in the rendered tutorial (and so are also not a good choice).

hoechenberger:
@agramfort @larsoner @dnacombo 
WDYT?

https://19932-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/raw/plot_40_visualize_raw.html#plotting-spectral-density-of-continuous-data

larsoner:
Agreed

tutorials/raw/plot_40_visualize_raw.py
```diff
@@ -188,6 +188,7 @@
 
 raw.plot_projs_topomap(colorbar=True)
 
+# XXX
```
agramfort, line 191:
```suggestion
```

larsoner:
Thanks @hoechenberger

tutorials/raw/plot_40_visualize_raw.py
hoechenberger: Thanks for the explanation, @drammock!

PR #7752, opened 2020-05-08
Montage subset hotfix

### LINKED ISSUE #7615: `.set_montage()` when some channel locations are not available
In past mne version (something like 0.19 or 0.18) I could do `raw.set_montage(montage)` even when some channel positions were not available. I would get a warning in such case, but it was possible. Later the default behavior was changed to throwing an error if some channel positions were missing, but allowed to override this using `raise_if_subset=False` argument. This argument is no longer present and I am not sure what is the preferred way to read custom montages that have a few channels missing.

Creating a digmontage using `mne.channels.make_dig_montage` and passing it to `raw.set_montage()` leads to an error. I can set `np.full(3, np.nan)` positions for those missing channels when creating the montage, but I am not sure if nan channel positions are handled well everywhere (for example such nan-position channel is still plotted in `raw.plot_sensors()` as if it had a meaningful position).

I think the past behavior of raising a warning when some channel positions are missing in the montage made sense. Throwing an error was still fine as long as it was possible to turn the error off with `raise_if_subset=False`.

cbrnr:
Fixes #7615. This re-enables setting `raise_if_subset` to `False` in 0.20.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ docstring:
 	@$(PYTESTS) --doctest-modules mne/tests/test_docstring_parameters.py
 
 check-manifest:
-	check-manifest --ignore .circleci*,doc,logo,mne/io/*/tests/data*,mne/io/tests/data,mne/preprocessing/tests/data,.DS_Store
+	check-manifest --ignore .circleci/config.yml,doc,logo,mne/io/*/tests/data*,mne/io/tests/data,mne/preprocessing/tests/data,.DS_Store
 
 check-readme: clean wheel_quiet
 	twine check dist/*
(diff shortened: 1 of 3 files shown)


cbrnr:
It would be great if we could make a new release soon. If possible, backporting #7731 would be really helpful too!

agramfort:
@cbrnr can you look at the CIs errors? Sure for the release we can do it quickly.

cbrnr:
Ready for review!

agramfort:
Thx @larsoner

agramfort:
I mean thanks @cbrnr

cbrnr:
@larsoner I don't know how to backport #7731 without manually copying the changes. There must be a git command for that I assume. Can you help?

drammock:
@cbrnr the merge message says that @agramfort merged into `maint/0.20` so it's already backported

drammock:
ah, oops, just realized you're talking about a different commit.  I use `git cherry-pick` but I'm sure there are several ways to do it

PR #7753, opened 2020-05-08
WIP: Enable usage of _nirs.fif

rob-luke:
Similar to #6416 I wish to enable support for fif filenames ending in _nirs.fif for compliance with BIDS. This is also discussed in #4950

#### Additional information
The BIDS specification is not complete yet. But the NIRS spec is heavily based on previous specs such as MEG.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -57,6 +57,8 @@ Changelog
 
 - Add ``plot`` option to :meth:`mne.viz.plot_filter` allowing selection of which filter properties are plotted and added option for user to supply ``axes`` by `Robert Luke`_
 
+- Accept filenames of raw .fif files to end on `_nirs.fif` to enable complicance with the Brain Imaging Data Structure by `Robert Luke`_
+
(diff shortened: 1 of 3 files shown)


rob-luke:
@sappelhoff having been through this before, do you have any comments on this?

sappelhoff:
Is `.fif` under consideration to be a file format for NIRS-BIDS? If not, why is this needed?

rob-luke:
Very good question. The plan is to have SNIRF as the format for BIDS. But SNIRF isn't completed yet, and no converters exist from other file formats to SNIRF. So I can't test with it. In the meantime I was planning to use fif as it seems well supported in MNE.

rob-luke:
If this seems like a bad idea I will modify the bids PRs instead.

agramfort:
I also feel it's premature and I doubt fif will be accepted a NIRS standard
format by BIDS community

>

rob-luke:
Thanks for the feedback. Closing

PR #7754, opened 2020-05-08
Ica annotations

bloyl:
When calling `ica.plot_sources(raw)` annotations were not displayed.

This PR mimics what is done in `raw.plot` so they are displayed. I'm not sure what the best way to provide a test is. Maybe I should modifiy the tutorial to have an annotation?

Thoughts?


larsoner:
For a test, I would ensure that there is some annotation in the first few seconds, call the plotter, then ensure that there is are more `fig.axes.patches` than there are on `master`. There are probably zero on `master` and one on your PR I'm guessing?

PR #7761, opened 2020-05-11
MRG, FIX: Fix find_bad_channels_maxwell indexing

### LINKED ISSUE #7760: find_bad_channels_maxwell errors on CTF data
find_bad_channels_maxwell raises an exception for ctf data if the data is longer then one duration.

to replicate:

```
import os.path as op
from mne.io import read_raw_ctf
from mne.preprocessing import find_bad_channels_maxwell

ctf_fname_continuous = op.join(testing.data_path(download=True),
                               'CTF', 'testdata_ctf.ds')

raw = read_raw_ctf(ctf_fname_continuous)
raw.apply_gradient_compensation(0)
bads, flats = find_bad_channels_maxwell(raw_cp)
```

which raises
```
/opt/anaconda3/envs/chop_mne-dev/lib/python3.7/site-packages/mne/preprocessing/maxwell.py in find_bad_channels_maxwell(***failed resolving arguments***)
   2021             # p2p
   2022             range_ = np.ptp(delta, axis=-1)
-> 2023             range_ *= params['coil_scale'][these_picks, 0]
   2024             mean, std = np.mean(range_), np.std(range_)
   2025             # z score

IndexError: index 298 is out of bounds for axis 0 with size 298
```

I suspect that ` _prep_maxwell_filter` and/or `_get_coil_scale` aren't dealing with ctfs axial gradiometers (that get labeled as magnetometers) correctly. So it is probably also an issue with KIT and Artemis sensors.

larsoner:
Closes #7760.

@bloyl can you check to see if it solves your problem? Currently it finds three ref channels bad, no idea if that's sensible. If not, might be worth doing `ignore_ref=True`. Also, it would be good to know in general whether or not this function works for your datasets now that it won't just error out on you :) But we should merge this PR either way as it fixes a clear bug, if it does not work well for your data we might be able to fix it in another PR.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -126,7 +126,7 @@ Bug
 
 - Fix bug with :func:`mne.preprocessing.find_bad_channels_maxwell` where good data of exactly ``step`` duration would lead to an error by `Eric Larson`_
 
-- Fix bug with :func:`mne.preprocessing.find_bad_channels_maxwell` where indices were not handled properly when MEG channels were not first in the raw instance by `Eric Larson`_
+- Fix bug with :func:`mne.preprocessing.find_bad_channels_maxwell` where indices were not handled properly when MEG channels were not first in the raw instance, and logging messages incorrectly reported the interval used by `Eric Larson`_
 
 - Make :func:`mne.set_config` accept path-like input values by `Richard Höchenberger`_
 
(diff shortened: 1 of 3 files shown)


bloyl:
This gets me past the error although something isn't quite right with the time steps.

It keeps saying:
```
Using 88/95 harmonic components for    0.000  (73/80 in, 15/15 out)
```
For each time segment. Hopefully its just a string formatting error?

Overall, I have no idea how this will work. I'd like to compare it to the old `Faster` method that had been in mne-sandbox, but I'm having trouble building that.

larsoner:
> For each time segment. Hopefully its just a string formatting error?

Indeed it's just a logging error, should be fixed in the latest commit

bloyl:
Logging looks great and it runs through. Thanks for taking a look so quickly!

larsoner:
Hope it ends up actually being useful!

PR #7763, opened 2020-05-11
MRG, FIX: Fixes for BEM contours

larsoner:
Should fix the bug mentioned in #7725. @christian-oreilly can you try it with your out-of-standard-ori MRI? I don't have any to test with but in principle I think this should work...

This PR also:

1. Should ensure that we always (by default) proceed through slices in L-to-R, P-to-A, and I-to-S order (previously it depended on the order in of the MRI).
2. Adds orientation markers to the edges of the plots by default (can be disabled with `show_orientations=False`)
3. Points out that there is a redundant and probably also wrong (for non-conform'ed MRIs) function `mne.viz._3d._plot_mri_contours` that we should really refactor with the `mne.viz.misc._plot_mri_contours` -- the function names match and there is a lot of similar code so it's unfortunate that we ended up with two copies, presumably due to @agramfort 6 years ago accidentally making two copies in https://github.com/mne-tools/mne-python/commit/5c1d64b1045f96c7c3a7e45bc4068e19fce695e7 :)
4. Changes the default `slices=None` from:
    ```
    np.linspace(0, n_slices, 12, endpoint=False)
    ```
    to:
    ```
    np.linspace(0, n_slices - 1, 14, endpoint=True)
    ```
    There is almost never any reason to show the first or last slice. On `master` we correctly avoid the last but not the first and unevenly sample in between, we now evenly sample from near-to-first to near-to-last instead by making 14 points evenly spaced between first and last slice (inclusive) and then excluding the first and last of those 14 to obtain 12.

On master for `plot_forward.py` we get:
Screenshot from 2020-05-11 14-10-37

And on this PR:

Screenshot from 2020-05-11 14-20-32

Note that there is a left-to-right flip here. On `master` the subject's left was shown on the right (radiological convention, probably not intentional) and this PR sets it to the more standard (for us at least) neurological convention of having the subject's left on the left. You can tell this by the fact that in the image from this PR, slice 177 (bottom left) has the pill on the right, and then the last slice is flipped left-right relative to the `master` plots.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -100,7 +100,7 @@ Bug
 
 - Fix bug when using :meth:`mne.Epochs.crop` to exclude the baseline period would break :func:`mne.Epochs.save` / :func:`mne.read_epochs` round-trip by `Eric Larson`_
 
-- Fix bug with :func:`mne.viz.plot_bem` when MRIs are not in standard FreeSurfer orientation by `Eric Larson`_
+- Fix bug with :func:`mne.viz.plot_bem` and :class:`mne.Report` when plotting BEM contours when MRIs are not in standard FreeSurfer orientation by `Eric Larson`_
 
 - Fix :ref:`gen_mne_setup_forward_model` to have it actually compute the BEM solution in addition to creating the BEM model by `Eric Larson`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -55,7 +55,7 @@ Changelog
 
 - Support for saving movies of source time courses (STCs) with ``brain.save_movie`` method and from graphical user interface by `Guillaume Favelier`_
 
-- Add ``mri`` argument to :func:`mne.viz.plot_bem` by `Eric Larson`_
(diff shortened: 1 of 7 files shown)


larsoner:
Reported as fixing orientations for @christian-oreilly [here](https://github.com/mne-tools/mne-python/pull/7725#issuecomment-626880174), so this one should be good to go

agramfort:
thanks for cleaning my mess @larsoner :)

larsoner:
Decided to take care of the deduplication here, it was't so bad actually. On `master` the BEM looks like:

Screenshot from 2020-05-11 17-07-24

And on this PR it looks like `plot_bem`:

Screenshot from 2020-05-11 17-05-25

larsoner:
Going to push one more fix to adjust the `axial` view, the left should really be on the left and the right on the right, not left on the bottom right on the top:

Screenshot from 2020-05-11 17-10-44

larsoner:
Style failure is just the one from #7767

agramfort:
thx @larsoner

PR #7764, opened 2020-05-11
MAINT: Bump tol for gamma map test

larsoner:
All PRs are currently failing due to a `gamma_map` test being too strict about distances:

https://travis-ci.org/github/mne-tools/mne-python/jobs/685787842

This one is only off by 3.6 mm so here we just bump the tol. I cannot reproduce locally so we'll have to see what the CIs say.

It's not clear to me what this is due to, possibly some version bump in mkl ... ?


agramfort:
yes it can be an update in MKL. As solver is non-convex you easily end up is a slightly different local minimum

PR #7770, opened 2020-05-12
Minor plot_raw aes improvement

cbrnr:
This just removes the title "Ch." of the vertical (channel) scroll bar. I think it looks cleaner without the title, which can also interfere with annotations/events. It should also be pretty clear even without the title what that scrollbar is.

I also wanted to change the `hsel_patch` to a uniformly transparent filled rect. Currently, this rectangle (which indicates the current position in the scroll bar) has a darker border which isn't really necessary. However, unless someone has an idea how to achieve this (the patch should still extend over the upper and lower scrollbar boundaries) I think it is probably not possible (https://stackoverflow.com/questions/53803531/matplotlib-circle-patch-with-alpha-produces-overlap-of-edge-and-facecolor).


larsoner:
Thanks @cbrnr

PR #7771, opened 2020-05-12
WIP, ENH: BrainVision impedance parsing

honzaseda:
I was using the MNE tool for parsing BrainVision data files in a paper where we needed to parse all the information from the segmentation and impedance sections of a header file. As i did not find these informations in the current raw parsed data, an enhancement of the current parser was implemented that adds these sections to the raw info.

Considering a possible wider use case for these changes, we decided that this enhancement could be included in the tool directly.

---
**Enhancements:**

The function `parse_segmentation` takes any of the key-value pairs in the segmentation subsections of the BrainVision file starting with `S e g m e n t a t i o n  /  A v e r a g i n g`, as well as some of the added key-values in the Common Infos. The segmentation is stored in a new info item `segmentation`.

The function `parse_impedance` creates a `dict` of all the channel impedances from the header file. All of the channels in the raw info are iterated over, and if an impedance record exists, they are appended to the channel dict.

The header file that these were tested on is in `mne/io/brainvision/tests/data/test_segmentation_impedance.vhdr` (also used for the unit tests)

---

**Problems:**

However, even though these changes work well in our local environment and the parsing of the values in itself works good, implementing these changes directly to MNE causes some of the already existing unit tests to fail, and I'm not really sure how to solve these issues. When i try to add any new attributes to the info structure, even an empty list.

These are the failing tests:
```
FAILED mne/io/brainvision/tests/test_brainvision.py::test_brainvision_data_highpass_filters - RuntimeWarning: Could not get size for self.info
FAILED mne/io/brainvision/tests/test_brainvision.py::test_brainvision_data_lowpass_filters - RuntimeWarning: Could not get size for self.info
FAILED mne/io/brainvision/tests/test_brainvision.py::test_brainvision_data_partially_disabled_hw_filters - AssertionError: assert {'acq_pars', ... 'comps', ...} == {'acq_pars', ... 'comps', ...}
FAILED mne/io/brainvision/tests/test_brainvision.py::test_brainvision_data_software_filters_latin1_global_units - AssertionError: assert {'acq_pars', ... 'comps', ...} == {'acq_pars', ... 'comps', ...}
FAILED mne/io/brainvision/tests/test_brainvision.py::test_brainvision_data - RuntimeWarning: Could not get size for self.info
```

Any ideas where the problem could be for the existing tests and how this could be improved?

### DIFF
--- a/mne/io/brainvision/brainvision.py
+++ b/mne/io/brainvision/brainvision.py
@@ -96,7 +96,7 @@ class RawBrainVision(BaseRaw):
         settings, cfg, cinfo, _ = _aux_vhdr_info(vhdr_fname)
         split_settings = settings.splitlines()
         self.impedances = parse_impedance(split_settings)
-        self.segmentation = parse_segmentation(split_settings, cfg, cinfo)
+        self.segmentation = _parse_segmentation(split_settings, cfg, cinfo)
 
         # Get annotations from vmrk file
         annots = read_annotations(mrk_fname, info['sfreq'])
@@ -853,12 +853,16 @@ def _check_bv_annot(descriptions):
 def parse_impedance(settings):
     """Parse impedances from the header file.
 
-    :param settings: header settings lines
-    :type settings: list
-    :param info: BrainVision info parsed from the original parser
-    :type info: dict
+    Parameters
+    ----------
+    settings : list
+        the header settings lines
+
+    Returns
+    -------
+    dict : A dictionary of all the channels and their impedances
     """
-    ranges = parse_impedance_ranges(settings)
+    ranges = _parse_impedance_ranges(settings)
     impedance_setting_lines = [i for i in settings if
                                i.startswith('Impedance')]
     impedances = dict()
@@ -893,13 +897,17 @@ def parse_impedance(settings):
     return impedances
 
 
-def parse_impedance_ranges(settings):
+def _parse_impedance_ranges(settings):
     """Parse the selected electrode impedance ranges from the header.
 
-    :param settings: header settings lines
-    :type settings: list
-    :returns parsed electrode impedances
-    :rtype dict
+    Parameters
+    ----------
+    settings : list
+        the header settings lines
+
+    Returns
+    -------
+    dict : A dictionary of impedance ranges for each type of electrode
     """
     impedance_ranges = [item for item in settings if
                         "Selected Impedance Measurement Range" in item]
@@ -924,23 +932,28 @@ def parse_impedance_ranges(settings):
     return electrode_imp_ranges
 
 
-def parse_segmentation(settings, cfg, common_info):
+def _parse_segmentation(settings, cfg, common_info):
     """Parse the segmentation/averaging section of the header.
 
-    :param settings: header settings lines
-    :type settings: list
-    :param cfg: cfg of the header file returned by _aux_vhdr_info
-    :type cfg: ConfigParser
-    :param common_info: cinfo from the BrainVision header parser
-    :type common_info: str
-    :returns the parsed segmentation as a dict
-    :rtype dict
+    Parameters
+    ----------
+    settings : list
+        the header settings lines
+    cfg : ConfigParser
+        cfg of the header file returned by _aux_vhdr_info
+    common_info : str
+        cinfostr from the BrainVision header parser
+
+    Returns
+    -------
+    dict : The parsed segmentation sections with their
+        key-value pairs as a dict
     """
     segmentation = dict()
     if 'S e g m e n t a t i o n  /  A v e r a g i n g' in settings:
         idx = settings.index('S e g m e n t a t i o n  /  A v e r a g i n g')
         segmentation_settings = settings[idx:]
-        segmentation = parse_basic_segmentation(cfg, common_info)
+        segmentation = _parse_basic_segmentation(cfg, common_info)
 
         if "Markers" in segmentation_settings:
             idx = segmentation_settings.index("Markers")
@@ -967,20 +980,20 @@ def parse_segmentation(settings, cfg, common_info):
             segmentation["intervals"] = intervals
 
         if "Averaging" in segmentation_settings:
-            segmentation["averaging"] = get_segmentation_key_values(
+            segmentation["averaging"] = _get_segmentation_key_values(
                 segmentation_settings,
                 segmentation_settings.index("Averaging"),
                 " is "
             )
 
         if "Artifact Rejection" in segmentation_settings:
-            segmentation["artifact_rejection"] = get_segmentation_key_values(
+            segmentation["artifact_rejection"] = _get_segmentation_key_values(
                 segmentation_settings,
                 segmentation_settings.index("Artifact Rejection")
             )
 
         if "Miscellaneous" in segmentation_settings:
-            segmentation["miscellaneous"] = get_segmentation_key_values(
+            segmentation["miscellaneous"] = _get_segmentation_key_values(
                 segmentation_settings,
                 segmentation_settings.index("Miscellaneous")
             )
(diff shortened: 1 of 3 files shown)


larsoner:
> As i did not find these informations in the current raw parsed data, an enhancement of the current parser was implemented that adds these sections to the raw info.

The set of values / things that can be added to `Info` is restricted by the FIF format spec, so we can't add things there arbitrarily. Instead it might be worth adding these to the `RawBrainvision` class as attributes or methods, such as `raw_bv.get_impedances()` or so. They'll be lost on save-read round-trip (to FIF) but that shouldn't be too much of a problem for getting things like impedances.

honzaseda:
Thanks for the suggestion, I will look into this.

honzaseda:
Moved the parsed data into attributes of the RawBrainvision class

mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,204 @@ def _check_bv_annot(descriptions):
     bv_markers = (set(_BV_EVENT_IO_OFFSETS.keys())
                   .union(set(_OTHER_ACCEPTED_MARKERS.keys())))
     return len(markers_basename - bv_markers) == 0
+
+
+def parse_impedance(settings):
```
agramfort, line 853: this should be a private function
```suggestion
def _parse_impedance(settings):
```

mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,204 @@ def _check_bv_annot(descriptions):
...
+    """Parse impedances from the header file.
+
+    :param settings: header settings lines
+    :type settings: list
+    :param info: BrainVision info parsed from the original parser
+    :type info: dict
```
agramfort, line 859: this is not how we document parameters. See numpydoc standard

mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,204 @@ def _check_bv_annot(descriptions):
...
+                    channel_impedance.update(ranges['Ground'])
+                elif 'Data' in ranges:
+                    channel_impedance.update(ranges['Data'])
+                impedances[channel_name] = channel_impedance
+            else:
+                break
+    return impedances
+
+
+def parse_impedance_ranges(settings):
```
agramfort, line 896:
```suggestion
def _parse_impedance_ranges(settings):
```

mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,204 @@ def _check_bv_annot(descriptions):
...
+                electrode_range = electrode_range.split()
+                electrode_imp_ranges[electrode_range[0]] = {
+                    "imp_lower_bound": float(electrode_range[6]),
+                    "imp_upper_bound": float(electrode_range[8]),
+                    "imp_range_unit": electrode_range[9]
+                }
+    return electrode_imp_ranges
+
+
+def parse_segmentation(settings, cfg, common_info):
```
agramfort, line 927:
```suggestion
def _parse_segmentation(settings, cfg, common_info):
```

mne/io/brainvision/tests/data/test_segmentation_impedance.vhdr
```diff
@@ -0,0 +1,143 @@
...
+DataFormat=BINARY
+; Data orientation: MULTIPLEXED=ch1,pt1, ch2,pt1 ...
+DataOrientation=MULTIPLEXED
+NumberOfChannels=19
+; Sampling interval in microseconds
+SamplingInterval=1000
+SegmentationType=MARKERBASED
+SegmentDataPoints=1100
+Averaged=YES
+AveragedSegments=80
```
agramfort, line 17: does it mean it should be read as an Evoked dataset?

honzaseda: Yes, if Averaged is set to YES, it means that the data file contains data that have been averaged across segments.

agramfort: this tells me that we should add a read_evoked_brainvision no?
I would rather not have raw_raw_brainvision read such a header without complaining

honzaseda: Probably? I'm not really sure as i do not know the inner workings of MNE that well.

agramfort:
@sappelhoff or @cbrnr can you pitch in?

mne/io/brainvision/brainvision.py
sappelhoff: @honzaseda you forgot to make this method private as well

cbrnr:
There is `raw._raw_extras` to store this kind of stuff. For example, almost all of the EDF header goes into that attribute. Not sure if putting it in that dict is a better option than creating a `raw.impedances` attribute or a getter method.

larsoner:
> BrainVision raw files will be the only files to have this attribute?

I think this is okay. It will also be lost on FIF round-trip but that's also okay. If there are other readers that support impedances we can think about a more global way of storing them or something. As long as we pick a sensible attribute here we can always extend it to other readers later via a mixin or something.

> The alternative would be to optionally return the impedances as a data table at read time

This would also be okay, but it's not all that extensible of there are other attributes people will want from other readers. You end up with a bunch of `return_...` kwargs or otherwise changing the API signature. Seems cleaner than just 

> There is raw._raw_extras to store this kind of stuff. 

Users should not expect private attributes to be stable, they can be changed in a backward compatible way at any time during development with no warning. So we should not advertise that people store information they want in there, or make use of private attributes in general.

BTW ideally / by design the only things that should actually live in `_raw_extras` are the things you need to actually `_read_segment_file`.

cbrnr:
Yes, if this should be used by users it's better not to rely on that private attribute. I'd vote for a simple `impedances` attribute then (instead of a getter method).

agramfort:
one strategy could be to use the chs['loc'][6] entry for impedance.
We've come up with such a strategy to store wavelength for fNIRS data.

sappelhoff:
> one strategy could be to use the chs['loc'][6] entry for impedance.

can we have unlimited space in the `chs['loc'][X]` entries? Or is this somehow bounded by FIFF rules?

If it's bounded, using the limited space for impedance values would seem like a waste to me.

If it's not bounded +1

Otherwise, with Eric's description I know also vote for a simple attribute.

And I also vote for @honzaseda to split the two enhancements into 2 PRs =)

agramfort:
chs pos is 12 values for all channels. for MEG is 3x3 rotation + 3
translation
for EEG it's 6 : location of channel + location of reference. So
technically we have 6 spare entries unused

>

honzaseda:
Removed the segmentation parsing, I will create a separate PR for this and leave only the impedance parsing here.

Left the impedances as an attribute of the `RawBrainvision` class for now.

mne/io/brainvision/tests/data/test_segmentation_impedance.vhdr
```diff
@@ -0,0 +1,143 @@
...
+
+A m p l i f i e r  S e t u p
+============================
+Number of channels: 19
+Sampling Rate [Hz]: 1000
+Sampling Interval [µS]: 1000
+
+Channels
+--------
+#     Name      Phys. Chn.    Resolution / Unit   Low Cutoff [s]   High Cutoff [Hz]   Notch [Hz]    Series Res. [kOhm] Gradient         Offset
```
larsoner, line 57: See comment in other PR, this should eventually be added to `mne-testing-data`

sappelhoff: for the sake of this PR you could add the impedances to one of the existing VHDR files in https://github.com/mne-tools/mne-python/tree/master/mne/io/brainvision/tests/data and then use that file for testing.

sappelhoff:
just a note, this PR could also come in handy for MNE-BIDS to automatically populate the `impedance` field in [`electrodes.tsv` BIDS files](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv).

We're having a short "unconference" at my institute soon, and maybe I'll find time there to finish this. Unless you have other plans @honzaseda ?

sappelhoff:
> We're having a short "unconference" at my institute soon, and maybe I'll find time there to finish this. Unless you have other plans @honzaseda ?

I just went ahead and finished it, and included you in the credits. Thanks for making a start with this PR @honzaseda!

See #7974

PR #7772, opened 2020-05-12
MRG, DOC: Notes about info

larsoner:
Make it clearer that people (including devs) shouldn't generally add new stuff to `info`.

### DIFF
--- a/mne/io/meas_info.py
+++ b/mne/io/meas_info.py
@@ -184,14 +184,17 @@ class Info(dict, MontageMixin):
     """Measurement information.
 
     This data structure behaves like a dictionary. It contains all metadata
-    that is available for a recording.
-
-    This class should not be instantiated directly. To create a measurement
-    information strucure, use :func:`mne.create_info`.
+    that is available for a recording. However, its keys are restricted to
+    those provided by the
+    `FIF format specification <https://github.com/mne-tools/fiff-constants>`__,
+    so new entries should not be manually added.
 
     The only entries that should be manually changed by the user are
     ``info['bads']`` and ``info['description']``. All other entries should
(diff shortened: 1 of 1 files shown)


cbrnr:
What about `info["ch_names"]` - this can also be changed manually because there's no function.

larsoner:
No, it shouldn't be -- it creates an inconsistency with other entries in info. You should use [inst.rename_channels](https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.rename_channels)

mne/io/meas_info.py
```diff
@@ -184,14 +184,17 @@ class Info(dict, MontageMixin):
...
-    information strucure, use :func:`mne.create_info`.
+    that is available for a recording. However, its keys are restricted to
+    those provided by the
+    `FIF format specification <https://github.com/mne-tools/fiff-constants>`__,
+    so new entries should not be manually added.
 
     The only entries that should be manually changed by the user are
     ``info['bads']`` and ``info['description']``. All other entries should
-    be considered read-only, or should be modified by functions or methods.
+    be considered read-only, or should be modified by MNE functions or methods.
```
drammock, line 194:
```suggestion
    be considered read-only, though they can be modified by various MNE-Python
    functions or methods (which have safeguards to ensure all fields remain in
    sync).
```

larsoner:
> You should use inst.rename_channels

It just occurred to me that we should probably document at least some common methods like these in the `info` class itself, so if someone wants to change them, they get a link to the method that does it. @drammock worth adding to some DOC todo?

drammock:
> we should probably document at least some common methods like these in the `info` class itself

#7774

agramfort:
thx @larsoner

PR #7775, opened 2020-05-12
MRG, ENH: Add resizing when time_viewer is present

larsoner:
On this PR this code snippet correctly sets the render area to the requested `size=(800, 600)`, and yet is resizeable afterward:

<details>
<summary>MWE</summary>

```
import os.path as op
import mne

initial_time = 0.1

size = (800, 600)
                     size=size)
sc = brain.screenshot()
assert sc.shape[:2] == size[::-1]
```

</details>

Screenshot from 2020-05-12 14-17-05

@GuillaumeFavelier can you take over to actually incorporate this properly? For example this also needs to work when `time_viewer=False` (it does not currently) -- you can probably use similar techniques. But I'll let you decide how to refactor to make it work for both cases.


mne/viz/_brain/_timeviewer.py
```diff
@@ -355,7 +357,21 @@ def __init__(self, brain, show_traces=False):
...
+        sz = self.brain._size
+        adjust_mpl = self.show_traces and not self.separate_canvas
+        if adjust_mpl:
+            self.mpl_canvas.canvas.setMinimumSize(
+                sz[0], int(round(sz[1] / self.interactor_stretch)))
+        try:
+            yield
+        finally:
+            if adjust_mpl:
+                self.mpl_canvas.canvas.setMinimumSize(0, 0)
```
larsoner, line 374: This flow control seems cleaner to me, WDYT?
```suggestion
        if not adjust_mpl:
            yield
        else:
            self.mpl_canvas.canvas.setMinimumSize(
                sz[0], int(round(sz[1] / self.interactor_stretch)))
            try:
                yield
            finally:
                self.mpl_canvas.canvas.setMinimumSize(0, 0)
```

GuillaumeFavelier: One if to rule them all

larsoner:
@GuillaumeFavelier can you add tests for size? I started to add some but the size was wrong for one run, things don't work with latest pyvista + pyvistaqt, etc. so it was not trivial

GuillaumeFavelier:
I'll take care of it

PR #7781, opened 2020-05-13
MRG, API: change out_type default in permutation_cluster_(1samp_)test

drammock:


drammock:
@larsoner FYI there was a Circle error:
```
Looking in links: https://vtk.org/download/
ERROR: Could not find a version that satisfies the requirement vtk>=9.0.0rc3 (from versions: none)
ERROR: No matching distribution found for vtk>=9.0.0rc3
```
restarted the pipeline and it got past that point, so maybe just transient internet weirdness?  Mentioning it in case it's useful since I know you've been working on VTK stuff lately.

larsoner:
Just needs an update to `latest.inc` in the `API` section then +1 for merge

mne/utils/docs.py
```diff
@@ -1128,7 +1128,17 @@
...
+"""
+docdict['clust_out_none'] = """
+out_type : None | 'mask' | 'indices'
+    Output format of clusters. If ``'mask'``, returns boolean arrays the same
+    shape as the input data, with ``True`` values indicating locations that are
+    part of a cluster. If ``'indices'``, returns a list of lists, where each
+    sublist contains the indices of locations that together form a cluster.
+    Note that for large datasets, ``'indices'`` may use far less memory than
+    ``'mask'``. ``None`` (the default) is equivalent to ``'mask'`` but will
+    change to ``'indices'`` in version 0.22.
```
larsoner, line 1141: Typically with deprecations we don't officially support `None` as a value. So you just document as:
```
out_type : None | 'mask' | 'indices'
    ...
    The default in 0.21 is ``'mask'`` but will change to ``'indices'`` in 0.22.
```

larsoner: Edited, forgot to actually remove the `None` in the original (whoops!)

drammock: but then there's no way to suppress the deprecation warning?  The way I've done it, the warning won't happen if `out_type='mask'` or `out_type='indices'`.

larsoner: You still suppress by passing a str. I'm not asking to change any code behavior, just the documentation string here. With deprecations we don't add `None` to the list of accepted values even though technically it will still work (and is the default). It's just a temporary stand-in alias, no user should actually pass `out_type=None` and expect it to be supported in future versions, so we don't document it as being accepted.

drammock: ah, understood.

PR #7784, opened 2020-05-14
MRG, ENH: Minor speedup of permutation tests

yh-luo:
Minor refactoring to speed up permutation tests. This PR brings about 1~2% speedup (best case 5%).

#### Setup

```python
import numpy as np
import time
from mne.stats import permutation_cluster_1samp_test

def large_samples_perm(n_perm=100, n_space=100, n_time=100):
    noise_level = 20
    n_time_1 = n_time
    n_time_2 = n_time
    normfactor = np.hanning(20).sum()
    rng = np.random.RandomState(42)
    condition1_1d = rng.randn(n_time_1, n_space) * noise_level
    for c in condition1_1d:
        c[:] = np.convolve(c, np.hanning(20), mode="same") / normfactor
    pseudoekp = 10 * np.hanning(round(n_space / 2))[None, :]
    condition1_1d[:, round(n_space / 2):] += pseudoekp

    thresh_tfce = dict(start=0, step=0.2)
    permutation_cluster_1samp_test(condition1_1d, threshold=thresh_tfce,
                                   n_permutations=n_perm, tail=0, seed=1,
                                   buffer_size=None)
```

#### n_perm=100, n_space=100, n_time=100

```python
%timeit large_samples_perm()
# master: 436 ms ± 13.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# opt: 413 ms ± 2.04 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
5.2% speedup

#### n_perm=1000, n_space=100, n_time=100

```python
%timeit large_samples_perm(n_perm=1000, n_space=100, n_time=100)
# master: 3.85 s ± 13.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# opt: 3.75 s ± 6.31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
2.6% speedup

#### n_perm=1000, n_space=300, n_time=300

```python
%timeit large_samples_perm(n_perm=1000, n_space=300, n_time=300)
# master: 5.83 s ± 22.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# opt: 5.73 s ± 31 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
1.7% speedup

#### n_perm=1000, n_space=1000, n_time=1000

```python
t0 = time.time()
large_samples_perm(n_perm=1000, n_space=1000, n_time=1000)
t1 = time.time()
elapsed = t1 - t0
# master: 17.224292039871216
# opt: 16.965471029281616
```
1.5% speedup

### DIFF
--- a/mne/stats/cluster_level.py
+++ b/mne/stats/cluster_level.py
@@ -501,6 +501,7 @@ def _find_clusters_1dir(x, x_in, connectivity, max_step, t_power, ndimage):
         if x.ndim == 1:
             # slices
             clusters = ndimage.find_objects(labels, n_labels)
+            # equivalent to if len(clusters) == 0 but faster
             if not clusters:
                 sums = list()
             else:
--- a/mne/stats/cluster_level.py
+++ b/mne/stats/cluster_level.py
@@ -454,8 +454,8 @@ def _find_clusters(x, threshold, tail=0, connectivity=None, max_step=1,
                 else:
                     len_c = len(c)
                 scores[c] += h * (len_c ** e_power)
-    # turn sums into ndarray after running
-    sums = np.concatenate(sums) if sums else np.empty(0)
+    # turn sums into array
+    sums = np.concatenate(sums) if sums else np.array([])
     if tfce:
         # each point gets treated independently
         clusters = np.arange(x.size)
@@ -514,15 +514,15 @@ def _find_clusters_1dir(x, x_in, connectivity, max_step, t_power, ndimage):
         else:
             # boolean masks (raveled)
             clusters = list()
-            sums = list()
-            for label in range(1, n_labels + 1):
-                c = labels == label
+            sums = np.empty(n_labels)
+            for label in range(n_labels):
+                c = labels == label + 1
                 clusters.append(c.ravel())
                 if t_power == 1:
-                    sums.append(np.sum(x[c]))
+                    sums[label] = np.sum(x[c])
                 else:
-                    sums.append(np.sum(np.sign(x[c]) *
-                                       np.abs(x[c]) ** t_power))
+                    sums[label] = np.sum(np.sign(x[c]) *
+                                         np.abs(x[c]) ** t_power)
     else:
         if x.ndim > 1:
             raise Exception("Data should be 1D when using a connectivity "
--- a/mne/stats/cluster_level.py
+++ b/mne/stats/cluster_level.py
@@ -252,16 +252,16 @@ def _get_clusters_st(x_in, neighbors, max_step=1):
     if len(cl_goods) > 0:
         keepers = [np.array([], dtype=int)] * n_times
         row, col = np.unravel_index(cl_goods, (n_times, n_src))
+        lims = [0]
         if isinstance(row, int):
             row = [row]
             col = [col]
-            lims = [0]
         else:
             order = np.argsort(row)
             row = row[order]
             col = col[order]
-            lims = [0] + (np.where(np.diff(row) > 0)[0] +
-                          1).tolist() + [len(row)]
+            lims += (np.where(np.diff(row) > 0)[0] + 1).tolist()
+            lims.append(len(row))
 
         for start, end in zip(lims[:-1], lims[1:]):
             keepers[row[start]] = np.sort(col[start:end])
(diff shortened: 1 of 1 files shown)


mne/stats/cluster_level.py
```diff
@@ -455,6 +454,8 @@ def _find_clusters(x, threshold, tail=0, connectivity=None, max_step=1,
                 else:
                     len_c = len(c)
                 scores[c] += h * (len_c ** e_power)
+    # turn sums into ndarray after running
+    sums = np.concatenate(sums) if sums else np.empty(0)
```
larsoner, line 458: `sums` is 1D so it's simpler just to do `sums = np.array(sums)` as `np.array([])` is the same as `np.empty(0)`

mne/stats/cluster_level.py
```diff
@@ -500,7 +501,7 @@ def _find_clusters_1dir(x, x_in, connectivity, max_step, t_power, ndimage):
         if x.ndim == 1:
             # slices
             clusters = ndimage.find_objects(labels, n_labels)
-            if len(clusters) == 0:
+            if not clusters:
```
larsoner, line 504: I find this less clear actually because before it was clear we were considering if there were any elements in an iterable (len-0), with the current code it's not clear if it's checking for boolean / None / len-0, etc.

mne/stats/cluster_level.py
```diff
@@ -513,15 +514,15 @@ def _find_clusters_1dir(x, x_in, connectivity, max_step, t_power, ndimage):
...
+                c = labels == label
                 clusters.append(c.ravel())
                 if t_power == 1:
-                    sums[label] = np.sum(x[c])
+                    sums.append(np.sum(x[c]))
                 else:
-                    sums[label] = np.sum(np.sign(x[c]) *
-                                         np.abs(x[c]) ** t_power)
+                    sums.append(np.sum(np.sign(x[c]) *
+                                       np.abs(x[c]) ** t_power))
```
larsoner, line 525: This ends up being less efficient than before, because now you'll end up with two copies of `sums` -- the first in a list and then another made later with `atleast_1d`.

In general we tend to preallocate and assign when weknow the size we will have (as is done here in `master`) and reserve list-append-convert-ndarray for cases where we don't know the size ahead of time (as is the case above, and why we use lists there), so conceptually this goes against that a bit, too...

mne/stats/cluster_level.py
yh-luo: `sums` is a list of `np.array`, if we do `sums = np.array(sums)` it breaks other functions.
I would change it to `sums = np.concatenate(sums) if sums else np.array([])`.

mne/stats/cluster_level.py
yh-luo: I originally thought that because `_find_clusters_1dir` is called in a loop (of another loop...) and `clusters` is always a list of None or slices, `not clusters` would be better to save some time. I would run some more tests to check.

mne/stats/cluster_level.py
yh-luo: Thanks for explaining! I will change it back

mne/stats/cluster_level.py
yh-luo: `if not clusters` cost 14.7904 s while `if len(clusters) == 0` cost 14.9306 s when n_perm=1000, n_space=1000, n_time=1000. `not clusters` could bring a 0.93% speedup (I found this [answer](https://stackoverflow.com/a/45778282) helpful). Running permutation test on my dataset cost about 7 hours, that would save about 4 minutes. Not much time saved indeed...Not sure what to do, what do you think?

larsoner: That's more than I expected actually. Maybe just add a comment that it's equiv to checking len(x) == 0 but faster

yh-luo:
### Current results

#### n_perm=100, n_space=100, n_time=100
370 ms ± 4.16 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
15.13% speedup

#### n_perm=1000, n_space=100, n_time=100
3.39 s ± 39.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
11.95% speedup

#### n_perm=1000, n_space=300, n_time=300
5.11 s ± 39.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
12.35% speedup

#### n_perm=1000, n_space=1000, n_time=1000
14.764962434768677
14.28% speedup

The performance is better than I think :partying_face:

larsoner:
Thanks @yh-luo !

larsoner:
(FYI, we are not notified of title changes, so thanks for the ping to review -- it probably would have taken me a while to notice otherwise)

PR #7791, opened 2020-05-15
MRG: Prepare migration to PyVista 0.25

GuillaumeFavelier:
This PR tests the migration of `BackgroundPlotter` to https://github.com/pyvista/pyvista_qt

**ToDo**
- [x] Build successfully on `[circle front]`
- [x] Bump sys_info (suggested in https://github.com/mne-tools/mne-python/pull/7791#issuecomment-648142199)
- [x] Catch deprecation warnings (suggested in https://github.com/mne-tools/mne-python/pull/7791#issuecomment-648148968)
- [x] Avoid regression with the `notebook` backend (suggested in https://github.com/mne-tools/mne-python/pull/7791#issuecomment-648148242)

Relevant to https://github.com/pyvista/pyvista/pull/719

Closes https://github.com/mne-tools/mne-python/pull/7697

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -472,6 +472,7 @@ class _Renderer(_BaseRenderer):
         self.plotter.reset_camera()
 
     def screenshot(self, mode='rgb', filename=None):
+        self.plotter.app.processEvents()
         return _take_3d_screenshot(figure=self.figure, mode=mode,
                                    filename=filename)
 
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -667,6 +667,7 @@ def _process_events(plotter, show=False):
         if show:
             plotter.app_window.show()
 
+
 def _set_colormap_range(actor, ctable, scalar_bar, rng=None):
     from vtk.util.numpy_support import numpy_to_vtk
     mapper = actor.GetMapper()
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -46,7 +46,7 @@ class _Figure(object):
                  size=(600, 600),
                  shape=(1, 1),
                  background_color='black',
-                 smooth_shading=False,
+                 smooth_shading=True,
                  off_screen=False,
                  notebook=False):
         self.plotter = plotter
@@ -224,7 +224,8 @@ class _Renderer(_BaseRenderer):
                 from matplotlib.colors import ListedColormap
                 colormap = ListedColormap(colormap)
 
-            actor = self.plotter.add_mesh(
+            actor = _add_mesh(
+                plotter=self.plotter,
                 mesh=mesh, color=color, scalars=scalars,
                 rgba=rgba, opacity=opacity, cmap=colormap,
                 backface_culling=backface_culling,
@@ -253,13 +254,16 @@ class _Renderer(_BaseRenderer):
             if kind == 'tube':
                 contour = contour.tube(radius=width, n_sides=self.tube_n_sides)
                 line_width = 1.0
-            self.plotter.add_mesh(mesh=contour,
-                                  show_scalar_bar=False,
-                                  line_width=line_width,
-                                  color=color,
-                                  cmap=colormap,
-                                  opacity=opacity,
-                                  smooth_shading=self.figure.smooth_shading)
+            _add_mesh(
+                plotter=self.plotter,
+                mesh=contour,
+                show_scalar_bar=False,
+                line_width=line_width,
+                color=color,
+                cmap=colormap,
+                opacity=opacity,
+                smooth_shading=self.figure.smooth_shading
+            )
 
     def surface(self, surface, color=None, opacity=1.0,
                 vmin=None, vmax=None, colormap=None,
@@ -275,13 +279,16 @@ class _Renderer(_BaseRenderer):
             mesh = PolyData(vertices, triangles)
             if scalars is not None:
                 mesh.point_arrays['scalars'] = scalars
-            self.plotter.add_mesh(mesh=mesh, color=color,
-                                  rng=[vmin, vmax],
-                                  show_scalar_bar=False,
-                                  opacity=opacity,
-                                  cmap=cmap,
-                                  backface_culling=backface_culling,
-                                  smooth_shading=self.figure.smooth_shading)
+            _add_mesh(
+                plotter=self.plotter,
+                mesh=mesh, color=color,
+                rng=[vmin, vmax],
+                show_scalar_bar=False,
+                opacity=opacity,
+                cmap=cmap,
+                backface_culling=backface_culling,
+                smooth_shading=self.figure.smooth_shading
+            )
 
     def sphere(self, center, color, scale, opacity=1.0,
                resolution=8, backface_culling=False,
@@ -299,7 +306,8 @@ class _Renderer(_BaseRenderer):
             mesh = PolyData(np.array(center))
             glyph = mesh.glyph(orient=False, scale=False,
                                factor=factor, geom=geom)
-            actor = self.plotter.add_mesh(
+            actor = _add_mesh(
+                self.plotter,
                 glyph, color=color, opacity=opacity,
                 backface_culling=backface_culling,
                 smooth_shading=self.figure.smooth_shading
@@ -321,15 +329,18 @@ class _Renderer(_BaseRenderer):
                 else:
                     scalars = None
                 tube = line.tube(radius, n_sides=self.tube_n_sides)
-                self.plotter.add_mesh(mesh=tube,
-                                      scalars=scalars,
-                                      flip_scalars=reverse_lut,
-                                      rng=[vmin, vmax],
-                                      color=color,
-                                      show_scalar_bar=False,
-                                      cmap=cmap,
-                                      smooth_shading=self.
-                                      figure.smooth_shading)
+                _add_mesh(
+                    plotter=self.plotter,
+                    mesh=tube,
+                    scalars=scalars,
+                    flip_scalars=reverse_lut,
+                    rng=[vmin, vmax],
+                    color=color,
+                    show_scalar_bar=False,
+                    cmap=cmap,
+                    smooth_shading=self.
+                    figure.smooth_shading
+                )
         return tube
 
     def quiver3d(self, x, y, z, u, v, w, color, scale, mode, resolution=8,
@@ -355,12 +366,15 @@ class _Renderer(_BaseRenderer):
             if mode == '2darrow':
                 return _arrow_glyph(grid, factor)
             elif mode == 'arrow' or mode == '3darrow':
-                self.plotter.add_mesh(grid.glyph(orient='vec',
-                                                 scale=scale,
-                                                 factor=factor),
-                                      color=color,
-                                      opacity=opacity,
-                                      backface_culling=backface_culling)
+                _add_mesh(
+                    self.plotter,
+                    grid.glyph(orient='vec',
+                               scale=scale,
+                               factor=factor),
+                    color=color,
+                    opacity=opacity,
+                    backface_culling=backface_culling
+                )
             elif mode == 'cone':
                 cone = vtk.vtkConeSource()
                 if glyph_height is not None:
@@ -372,14 +386,16 @@ class _Renderer(_BaseRenderer):
                 cone.Update()
 
                 geom = cone.GetOutput()
-                self.plotter.add_mesh(grid.glyph(orient='vec',
-                                                 scale=scale,
-                                                 factor=factor,
-                                                 geom=geom),
-                                      color=color,
-                                      opacity=opacity,
-                                      backface_culling=backface_culling)
-
+                _add_mesh(
+                    self.plotter,
+                    grid.glyph(orient='vec',
+                               scale=scale,
+                               factor=factor,
+                               geom=geom),
+                    color=color,
+                    opacity=opacity,
+                    backface_culling=backface_culling
+                )
             elif mode == 'cylinder':
                 cylinder = vtk.vtkCylinderSource()
                 cylinder.SetHeight(glyph_height)
@@ -397,13 +413,16 @@ class _Renderer(_BaseRenderer):
                 trp.Update()
 
                 geom = trp.GetOutput()
-                self.plotter.add_mesh(grid.glyph(orient='vec',
-                                                 scale=scale,
-                                                 factor=factor,
-                                                 geom=geom),
-                                      color=color,
-                                      opacity=opacity,
-                                      backface_culling=backface_culling)
+                _add_mesh(
+                    self.plotter,
+                    grid.glyph(orient='vec',
+                               scale=scale,
+                               factor=factor,
+                               geom=geom),
+                    color=color,
+                    opacity=opacity,
+                    backface_culling=backface_culling
+                )
 
     def text2d(self, x_window, y_window, text, size=14, color='white',
                justification=None):
@@ -488,6 +507,12 @@ class _Renderer(_BaseRenderer):
         if not self.figure.store['off_screen']:
             for renderer in self.plotter.renderers:
                 renderer.enable_depth_peeling()
+
+
+def _add_mesh(plotter, *args, **kwargs):
+    if hasattr(plotter, "app"):
+        plotter.app.processEvents()
+    return plotter.add_mesh(*args, **kwargs)
 
 
 def _deg2rad(deg):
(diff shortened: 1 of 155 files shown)


GuillaumeFavelier:
Building the website frontpage is necessary too.

GuillaumeFavelier:
[pyvistqt is on pypi](https://github.com/pyvista/pyvistaqt/issues/4)! I can move forward with the migration

GuillaumeFavelier:
We have a few errors:

```
AttributeError: This plotter is closed and unable to save a screenshot.
```

```
UserWarning: VTK 9 no longer accepts an offset array
```

A crash:

```
―――――――――――――――――――――― mne/viz/_brain/tests/test_brain.py ――――――――――――――――――――――

[gw2] darwin -- Python 3.7.4 /Users/travis/miniconda/bin/python

worker 'gw2' crashed while running 'mne/viz/_brain/tests/test_brain.py::test_brain_timeviewer_traces[pyvista-both]'
```

A backend selection failure:

```
――――――――――――――――――――――――――――― test_get_3d_backend ――――――――――――――――――――――――――――――

[gw3] darwin -- Python 3.7.4 /Users/travis/miniconda/bin/python

mne/viz/backends/tests/test_renderer.py:162: in test_get_3d_backend

    assert renderer.get_3d_backend() == orig_backend

E   AssertionError: assert 'mayavi' == None

E    +  where 'mayavi' = <function get_3d_backend at 0x112db94d0>()

E    +    where <function get_3d_backend at 0x112db94d0> = <module 'mne.viz.backends.renderer' from '/Users/travis/build/mne-tools/mne-python/mne/viz/backends/renderer.py'>.get_3d_backend
```

larsoner:
Indeed the screenshot error is the one I got when I briefly tried to get it working. Not sure what the problem is there, but clearly just the `from pyvistaqt import ...` is not equivalent to what used to be done in `pyvista` :(

GuillaumeFavelier:
Let's try with `BUILDING_GALLERY = True`

Reference: https://github.com/pyvista/pyvista/pull/783

GuillaumeFavelier:
I decided to change my approach. PyVista's architecture has changed a lot recently and this flow of issues is too much for me to handle at the same time. I will proceed from the latest working version, iterate one step at a time and adapt MNE to digest the changes.

GuillaumeFavelier:
Good start. `master`(78bc2a4c6d7f2a0ad545d242fa715b711510eb02) with `0.24.2` makes all the CIs happy and looking particularly at Azure here, I think it could be nice to pin to `0.24.2` instead of `0.24.0` as a short term solution @larsoner

GuillaumeFavelier:
Azure seems a tad slower but all good with `0.24.3` too, good news.

agramfort:
🎉

>

larsoner:
😭

larsoner:
(I speak for the CIs)

GuillaumeFavelier:
Not everything is green, it's not failed tests but timeouts.

GuillaumeFavelier:
The render window is cleared with `_clear_ren_win()` before the screenshot, that's why we get:

```
AttributeError: This plotter is closed and unable to save a screenshot.
```

GuillaumeFavelier:
Migrating completely to `BackgroundPlotter` should fix this one (work in progress in https://github.com/mne-tools/mne-python/pull/7697).
Those PRs are oddly connected.

larsoner:
Feel free to combine into a single PR if you think it would help

GuillaumeFavelier:
And this one:

```
UserWarning: VTK 9 no longer accepts an offset array
```

has been fixed upstream normally.

GuillaumeFavelier:
I take 35faa8e6dece95b2dcba8ade472f87b750ab1082 from https://github.com/mne-tools/mne-python/pull/7758 to fix the backend selection failure in `test_get_3d_backend()`.

GuillaumeFavelier:
This is still here in `pyvista 0.25.3`:

```
UserWarning: VTK 9 no longer accepts an offset array
```

larsoner:
We can add that as an `ignores` in `mne/conftest.py`, I'm more worried about the segfaults here

https://travis-ci.org/github/mne-tools/mne-python/jobs/699043753

GuillaumeFavelier:
Indeed a serie of crashes on Travis and Azure...

GuillaumeFavelier:
The Travis who failed early did not manage to install mayavi for some reason @larsoner:

https://travis-ci.org/github/mne-tools/mne-python/jobs/699235108#L859

The other one on MacOS gets `mayavi` but fails in the pyvista tests, I'll investigate:

https://travis-ci.org/github/mne-tools/mne-python/jobs/699235109#L3277

larsoner:
Ahh the:
```
    qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
```
we've gotten on CircleCI and IIRC it has to do with necessary libraries not being found

GuillaumeFavelier:
Do we get this only in this PR? Could it be related to a new version of PyQt5?

larsoner:
No it's all PRs. Yes Qt 5.15 probably. I'll look into it, you can ignore for now...

larsoner:
(or if you need things to pass, pin to <5.15 temporarily)

larsoner:
I killed all Travis runs except for OSX (the other important / full ones will fail)

GuillaumeFavelier:
@agramfort @larsoner do you reproduce the segfault on Windows or MacOS in you checkout this branch on 7b6c3446ab0d3df4df82249e5b654a9e6ff7d99c : 

with (latest version) : 

`PyQt5==5.15.0`
`pyvista==0.25.3`
`pyvistaqt==0.1.0`
`vtk==9.0.0`

when : 

```
pytest -v --cov=mne mne/viz/_brain/tests/test_brain.py
```

or in general :

```
pytest -v mne/viz
```

GuillaumeFavelier:
Meanwhile Azure is green since a long time when I disable traces, I'll restart to confirm if it's spurious.

GuillaumeFavelier:
The issue on Windows seems related to `show_traces`. I'll trigger again but it's already 3 times.

For MacOS, it's a different fight though. For this one, maybe the cause is `vtk 9.0.0`

GuillaumeFavelier:
Also as an unexpected insight, `show_traces=True` has a bigger overhead than I thought since removing it speedup testing on azure significantly (except some caching magic I would not be aware of):

image

I'll investigate further on my Windows tomorrow.
Still need help to reproduce on MacOS.

larsoner:
Based on what the PyVista folks wrote I think we more or less know VTK9 is broken on OSX, we should probably do something like this (hopefully it will work):
```
- vtk<9; platform_system == "Darwin"
- vtk; platform_system != "Darwin"
```
in both `requirements.txt` and `environment.yml`.

larsoner:
macOS ended up with no VTK because it tried to use the 8.2 source distribution (!) from vtk.org; changed the req on Darwin to `<8.2` so that it picks up the PyPi version instead of the vtk.org one.

larsoner:
VTK9 is not the problem on macOS, still segfaults with 8.1.2:

https://travis-ci.org/github/mne-tools/mne-python/jobs/700078046#L3312

mne/viz/backends/_pyvista.py
```diff
@@ -168,7 +168,6 @@ def __init__(self, fig=None, size=(600, 600), bgcolor='black',
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", category=FutureWarning)
             if MNE_3D_BACKEND_TESTING:
-                self.figure.plotter_class = Plotter
```
larsoner, line 171: Perhaps what you need to do when switching to the other plotting class is ensure that, during closure, events get processed before actually closing the window. For example back in the day for vispy I had to call `app.process_events()` [a couple of times](https://github.com/vispy/vispy/blob/master/vispy/app/backends/_qt.py#L770-L778), and this made things more stable / fixed segfaults.

But to test these sorts of things you really need to iterate locally on a machine with the OS rather than wait for CI 30-minute round trips.

BTW you can also speed up this iterative process on the CIs when you think you have something working by running just `pytest ... mne/viz` instead of `pytest ... mne`. Might be worth switching this to that now to ensure that the segfaults still happen with the reduced tests.

GuillaumeFavelier:
Thanks for the confirmation. I am currently investigating the issue with Windows locally but i'll have to use the CIs for MacOS.
Next in my list is PyQt5. Otherwise I have to pinpoint where this happens.
I'll try `app.process_events()` too but why only on MacOS? 🤔

GuillaumeFavelier:
I can reproduce locally on Windows. The segfault happens during picking.

agramfort:
only one failure: https://travis-ci.org/github/mne-tools/mne-python/jobs/700120874

do you want me to test something on my mac?

GuillaumeFavelier:
Yes, can you try https://github.com/mne-tools/mne-python/pull/7791#issuecomment-646195854 or on the latest commit (6147a497ab15ac2a25c7903de0f2c4a99975d0ba)?

GuillaumeFavelier:
I notice a serie of `DeprecationWarning` on Azure. I guess it's related to the latest numpy?

```
DeprecationWarning: `np.bool` is a deprecated alias for the builtin `bool`. Use `bool` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.bool_` here.
```

GuillaumeFavelier:
Okay now MacOS then Circle

GuillaumeFavelier:
So `_sphere` did the trick for Windows, `processEvents` for `Qt` and `vtk<8.2` for MacOS.

GuillaumeFavelier:
Now circle

larsoner:
If you expect to have to iterate, `[circle front]` will build 2 examples (I think?) that require PyVista and only take ~20 minutes instead of ~2 hours

larsoner:
... actually once something does fail on CircleCI, you can "Rebuild with SSH" on circle, wait for it to fail again, then SSH into the machine (it gives an address at the end of the run) and then do:
```
source $BASH_ENV
set +e
```
then you can tweak and try things iteratively on the machine where the failure occurs.

GuillaumeFavelier:
I'll use SSH with this run and `[circle front]` if I really have to commit again.

larsoner:
Once you have things working, can you cherry-pick bc2e8da04 ? Don't want to push it to your branch now because it'll kill CircleCI

GuillaumeFavelier:
Sure, I'll take care of it :+1:

agramfort:
@GuillaumeFavelier I tried this PR on my mac

```
$ conda list | grep vista
pyvista                   0.25.3                   pypi_0    pypi
pyvistaqt                 0.1.0                    pypi_0    pypi
$ mne sys_info
Platform:      Darwin-19.0.0-x86_64-i386-64bit
Python:        3.7.5 (default, Oct 25 2019, 10:52:18)  [Clang 4.0.1 (tags/RELEASE_401/final)]
Executable:    /Users/alex/miniconda3/bin/python
CPU:           i386: 16 cores
Memory:        32.0 GB

mne:           0.21.dev0
numpy:         1.17.4 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.4.1
matplotlib:    3.2.1 {backend=TkAgg}

sklearn:       0.23.1
numba:         0.46.0
nibabel:       3.0.1
cupy:          Not found
pandas:        1.0.5
dipy:          1.0.0
mayavi:        4.7.1
pyvista:       0.25.3
vtk:           8.2.0
PyQt5:         5.13.2
```

some good news.

These tests pass:

$ pytest mne/viz/tests/test_3d.py mne/viz/backends/tests/

and 3d viz from ipython works.

now the 3d notebook support now breaks with a segfault.

let me know if you want to talk

larsoner:
Locally I also need 99eccfc27 to make `make clean; MNE_3D_BACKEND=pyvista make html_dev-front` work, but it does otherwise work! If you get errors about offset array and/or NumPy deprecations, that commit should fix them.

GuillaumeFavelier:
Thanks for trying it out. I'll test the notebook and fix local building too.

larsoner:
@GuillaumeFavelier the build is already in the process of failing due to the offset array bug:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/2248/workflows/427fd7d3-7283-421d-a5ac-14af16f98a58/jobs/20745/steps

Might be worth cherry-picking my commits then `git commit --allow-empty -m 'Build [circle front]'`'ing to see if it works...

mne/viz/backends/_pyvista.py
```diff
@@ -773,6 +773,18 @@ def _glyph(dataset, scale_mode='scalar', orient=True, scalars=True, factor=1.0,
     return alg.GetOutput()
 
 
+def _sphere(plotter, center, color, radius):
```
larsoner, line 776: Can you add some comment about why this is necessary / `renderer.sphere` did not work?

larsoner:
Looks like it worked

mne/viz/backends/_pyvista.py
GuillaumeFavelier: I was able to reproduce a segfault when I was clicking/picking a lot in a very short period of time. I assume here that this is what happens during testing. I compared the 2 functions and the only noticeable difference is the use of a glyph filter before adding a sphere to the scene. So I guess that the pipeline is heavier with glyph and maybe some form of synchronisation could mitigate the issue with `renderer.sphere()`. On the other side, using glyph to only add one sphere looked a bit overkill to me.

larsoner: Okay can you put this in the code comments? At some point we might consider just adding some delay or a `processEvents` after the clicking.

Also I would just set `rtol=1e-6` in to fix https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=7774&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=f8799db5-b972-5f58-516d-bc88b8357f99&l=50

GuillaumeFavelier:
@agramfort I tried locally with `vtk 9.0.0` and `vtk 8.1.2` on my linux and both the notebook I used for testing and `test_notebook.py` are working. But we can discuss that offline too. Meanwile, this is still WIP.

agramfort:
green !!!! is this still WIP?

GuillaumeFavelier:
It is WIP at least until the situation with the `notebook` backend is clarified I think.

agramfort:
actually It's not so clear what's going on for me.

I have some segfaults with mne.viz.plot_alignment in plain ipython so my
crash is not related to notebook.
Also stc.plot does work in the notebook for me.

it's black magic for me...

>

GuillaumeFavelier:
Nice catch @agramfort, I was able to reproduce on my linux with:

```
examples/visualization/plot_montage.py
```

GuillaumeFavelier:
I'm concerned by the fact that the CIs did not catch this though.

GuillaumeFavelier:
Oh no... Actually it's because `smooth_shading` is disabled during testing...

GuillaumeFavelier:
@agramfort, can you try again on the latest commit (4d92ed67fdc9a3ccafb51aeaa9034546cb585cee) ?

agramfort:
does not help. We can do a screen sharing session in the afternoon if you
want

>

larsoner:
On my macOS machine I changed `environment.yml` to have `python<3.8`, then ran:
```
conda env create -n pyvistaqt environment.yml
conda activate pyvistaqt
cd mne/viz/_brain/tests
jupyter notebook
```
Then ran `stc.py`, it worked. Jupyter notebook had a "unexpected keyword argument 'io_loop'" that I then fixed with `pip install tornado==5.4.3`. Then `pip install pyimpl` (probably should have used `server_environment.yml`. Then segfault in `_sphere->_add_mesh->pyvista/plotting/plotting.py->core/filters.py line 3001 in compute_normals`. So I think this is probably related to things not being properly initialized at the VTK end.

GuillaumeFavelier:
Many thanks! Your detailed report gives me some more material to work with to debug the situation.

Should we disable `smooth_shading` then as a short term solution?

GuillaumeFavelier:
Disabling for now to move forward. We can always enable it later once the situation is clarified.

larsoner:
Yes, that at least fixes `test_notebook.py` at my end!

larsoner:
Sadly, this fixes the segfault but the plots do not show brains (colorbars, etc. show up but not the brains) for me in the notebook

GuillaumeFavelier:
Any error log by any chance? What script did you use?

larsoner:
Never mind, actually I was testing in the wrong environment! Things work locally for me on macOS now so +1 for merge

GuillaumeFavelier:
Is it working for you too @agramfort?

larsoner:
@GuillaumeFavelier pushed 8be2205 to DRY our `processEvents` calls.

GuillaumeFavelier:
Wow, now it's packed with `processEvents()`. Fingers crossed for https://github.com/mne-tools/mne-python/pull/7927

larsoner:
@GuillaumeFavelier my plan is to merge this pending @agramfort's approval, then #7931, then #7775, then #7927 (I'm about to enable traces there), probably with a merge or rebase at each step, sound good?

agramfort:
Let me try later Tonight

GuillaumeFavelier:
Sounds like a plan

agramfort:
works like a charm !

awesome work @GuillaumeFavelier and @larsoner !

PR #7792, opened 2020-05-15
WIP, ENH: BrainVision segmentation parsing

honzaseda:
This was originally in the #7771 Pull Request.

Added parsing of the segmentation metadata from the BrainVision header file. 

**Enhancements:**

The function `_parse_segmentation` takes any of the key-value pairs in the segmentation subsections of the BrainVision file starting with `S e g m e n t a t i o n  /  A v e r a g i n g`, as well as some of the added key-values in the Common Infos. The segmentation is stored in a new attribute `segmentation` of the `RawBrainvision` class.

In the original PR there was a suggestion that the dataset should be read as an Evoked dataset for `Averaged=YES`?

### DIFF
--- a/mne/io/brainvision/brainvision.py
+++ b/mne/io/brainvision/brainvision.py
@@ -93,6 +93,11 @@ class RawBrainVision(BaseRaw):
 
         self.set_montage(montage)
 
+        settings, cfg, cinfo, _ = _aux_vhdr_info(vhdr_fname)
+        split_settings = settings.splitlines()
+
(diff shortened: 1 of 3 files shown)


agramfort:
@honzaseda first thing is to share a tiny file containing averaged data. Then we can add it to mne-testing-data repo and then add a read_evoked_brainvision function.

larsoner:
I'm going to close this since it seems like it's no longer being worked on (?), but feel free to reopen if I'm mistaken @honzaseda !

PR #7795, opened 2020-05-18
[MRG] update to pymatreader 0.0.23

thht:
pymatreader received an update. it now works correctly with complex numbers and a few crashes were fixed.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -71,6 +71,8 @@ Changelog
 
 - BIDS conformity: When saving FIF files to disk and the files are split into parts, the ``split_naming='bids'`` parameter now uses a "_split-%d" naming instead of the previous "_part-%d", by `Stefan Appelhoff`_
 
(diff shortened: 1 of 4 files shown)


thht:
mhhh.... something wrong with eeglab stuff it seems.....

larsoner:
For the:
```
E   UserWarning: Complex objects (like classes) are not supported. They are imported on a best effort base but your mileage will vary.
```
I would just enclose it with a `with pytest.warns(UserWarning, match='Complex objects'):`.

For the event count one, that will require more investigation and could be real. Let me know if you need help looking into it

thht:
ok, this looks good... should i put a note in changes?

thht:
ok, seems to be all good...

thht:
the failed job is caused by some error in `mne\minimum_norm\tests\test_inverse.py:182: in _compare_inverses_approx` so it does not seem to have anything to do with my changes...

larsoner:
Thanks @thht

PR #7796, opened 2020-05-18
MRG, MAINT, DOC: Remove spyder

### LINKED ISSUE #7788: MAINT: remove spyder from environment file?
Seems like every 6-12 months, one of my colleagues says "Spyder won't load, and in trying to fix it I borked my environment, please help."  I'm wondering why we implicitly recommend spyder by including it in our `environment.yml` file.  Spyder also imposes version constraints (e.g., currently it requires `PyQt5<5.13`), which has the potential to cause conflicts / inconsistent environments if (for example) pyvista and spyder impose incompatible dependency constraints.

Shouldn't we leave it up to the end user to pick their IDE?  If not, can we be a bit more intentional about our recommendations?

drammock:
closes #7788 

- removes spyder from our environment file
- adds docs section on installing IDEs
- small clarification to pre-install docs

### DIFF
--- a/doc/install/mne_python.rst
+++ b/doc/install/mne_python.rst
@@ -241,6 +241,45 @@ check out the :doc:`advanced` page to see if your problem is discussed there.
 If not, the `MNE mailing list`_ and `MNE gitter channel`_ are
 good resources for troubleshooting installation problems.
 
+
+Installing a Python IDE
+^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


drammock:
here's [the new installation docs section](https://20138-1301584-gh.circle-artifacts.com/0/dev/install/mne_python.html#installing-a-python-ide)

agramfort:
thx @drammock

PR #7797, opened 2020-05-18
FIX: Correct a bug in find_bads_eog

### LINKED ISSUE #6014: Mne.find_bads_eog () only takes one EOG channel into account I have 3
Hello Dear Mne Developers, 
I am new to MNE python and this EEG field. I am using mne.ICA for removing EOG artifacts. I have 3 EOG channels and I am following the example published on the web. The problem is after fitting the ICA I use the **find_bads_eog** for selecting bad components but this one only takes into account the first EOG channel and removes only one component. But on display, I can clearly see a lot of artifacts are coming from other 2 channels. Is there a way to force ICA to take into account all the available EOG channels? 

Thanks in advance :)

christian-oreilly:
Closes #6014

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -72,6 +72,8 @@ Changelog
 Bug
 ~~~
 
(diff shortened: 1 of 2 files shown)


agramfort:
thx @christian-oreilly

Issue #6266, opened 2019-05-03
DOC, TODO: advanced git page?


drammock:
I'd been hoping we could avoid it and let the rest of the internet handle advanced git advice, but based on conversations in #6254, it seems like a troubleshooting / advanced git page might be wanted (at least by some)?

- [ ] when, how, and why to rebase
- [ ] tips like `git push -u`
- [ ] `git commit --amend` and `git push --force`
- [ ] `git stash`

@jeythekey @massich @larsoner @mmagnuski

PR #7800, opened 2020-05-19
Alias temporal_derivative_distribution_repair to tddr

### LINKED ISSUE #7798: Rename temporal_derivative_distribution_repair to tddr
Can we rename `temporal_derivative_distribution_repair` to just `tddr`? I really think that the function name is too long (39 characters) - it makes it impossible to stay within the 79 character limit without using ugly `\` line breaks, e.g.:

```python
from mne.preprocessing.nirs import temporal_derivative_distribution_repair as tddr  # 82 characters
```

I think renaming to `tddr` is not problematic because the function is in the `nirs` subpackage, and TDDR is a commonly known abbreviation in the NIRS community (googling for "nirs tddr" immediately brings up several results with the correct method).

This issue came up in https://github.com/cbrnr/mnelab/pull/145 that @rob-luke is working on. He'd probably be fine with renaming because that's what he initially proposed.

And while we're at it, I'd like to rename `beer_lambert_law` to `beer_lambert` as well (we'd save 4 chars without losing any meaning).

cbrnr:
Fixes #7798. I've also made the docstring of the internal `_TDDR` function more Pythonic.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repar` with shortened alias ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repair` with shortened alias ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
 
 - Add :func:`mne.read_freesurfer_lut` to make it easier to work with volume atlases by `Eric Larson`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Added temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.tddr` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.tddr` by `Robert Luke`_
 
 - Add :func:`mne.read_freesurfer_lut` to make it easier to work with volume atlases by `Eric Larson`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Add temporal derivative distribution repair (TDDR) ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repar` with shortened alias ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
 
 - Add :func:`mne.read_freesurfer_lut` to make it easier to work with volume atlases by `Eric Larson`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.tddr` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
 
 - Add :func:`mne.read_freesurfer_lut` to make it easier to work with volume atlases by `Eric Larson`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Added temporal derivative distribution repair :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repair` by `Robert Luke`_
+- Added temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.tddr` by `Robert Luke`_
 
 - Add :func:`mne.read_freesurfer_lut` to make it easier to work with volume atlases by `Eric Larson`_
 
--- a/examples/preprocessing/plot_fnirs_artifact_removal.py
+++ b/examples/preprocessing/plot_fnirs_artifact_removal.py
@@ -16,7 +16,8 @@ how artifact correction techniques attempt to correct the data.
 import os
 import mne
 
-from mne.preprocessing.nirs import optical_density, tddr
+from mne.preprocessing.nirs import (optical_density,
+                                    temporal_derivative_distribution_repair)
 
 ###############################################################################
 # Import data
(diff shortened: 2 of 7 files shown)


examples/preprocessing/plot_fnirs_artifact_removal.py
```diff
@@ -17,7 +17,7 @@
 import mne
 
 from mne.preprocessing.nirs import (
-    optical_density, temporal_derivative_distribution_repair)
+    optical_density, tddr)
```
hoechenberger, line 20: Should be:
```python
from mne.preprocessing.nirs import optical_density, tddr
```

cbrnr: Thanks, PyCharm didn't catch this automatically!

hoechenberger: Well it's not a bug, and it probably complies with PEP8, so there's not much to catch :)

cbrnr: Automatic refactoring should be able to take care of such cases IMO, but maybe I'm wrong.

mne/preprocessing/nirs/_tddr.py
```diff
@@ -30,6 +30,11 @@ def tddr(raw):
     raw : instance of Raw
          Data with TDDR applied.
 
+    Notes
+    -----
+    There is a shorter alias :func:`mne.preprocessing.nirs.tddr` that can be
+    used instead of this function (e.g. if PEP8 character limit is an issue).
```
hoechenberger, line 36:
```suggestion
    used instead of this function (e.g. if line length is an issue).
```

cbrnr:
Not really related to this PR, but while we're cleaning up shouldn't `test_nirs.py` be in `mne/preprocessing/nirs/tests` (it is currently in `mne/preprocessing/tests`?

examples/preprocessing/plot_fnirs_artifact_removal.py
```diff
@@ -75,7 +74,7 @@
 # This approach corrects baseline shift and spike artifacts without the need
 # for any user-supplied parameters :footcite:`FishburnEtAl2019`.
 
-corrected_tddr = temporal_derivative_distribution_repair(corrupted_od)
+corrected_tddr = tddr(corrupted_od)
```
rob-luke, line 77: Can we leave this as it was? As a tutorial I think it is appropriate to have the extended name for ease of reading.

cbrnr: Sure!

hoechenberger: +1 on using the long name here, and potentially adding a comment that `from ... import tddr` is also available if preferred

rob-luke:
Thanks for fixing the docs strings too. I had just done a direct copy paste from the original author.

cbrnr:
@larsoner where do I need to exclude the `tddr` function for the docstring checks to pass?

mne/preprocessing/nirs/_tddr.py
```diff
@@ -30,6 +30,11 @@ def temporal_derivative_distribution_repair(raw):
     raw : instance of Raw
          Data with TDDR applied.
 
+    Notes
+    -----
+    There is a shorter alias :func:`mne.preprocessing.nirs.tddr` that can be
```
larsoner, line 35: Use double-backticks here. There is no entry in `python_reference.rst` for this to link to, hence on CircleCI:
```
/home/circleci/project/mne/preprocessing/nirs/_tddr.py:docstring of mne.preprocessing.nirs.temporal_derivative_distribution_repair:30: WARNING: py:func reference target not found: mne.preprocessing.nirs.tddr
```

doc/changes/latest.inc
```diff
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Added temporal derivative distribution repair :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repair` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.tddr` by `Robert Luke`_
```
larsoner, line 42: Same here, double-backticks
```
changes/latest.inc:42: WARNING: py:func reference target not found: mne.preprocessing.nirs.tddr
```

mne/preprocessing/nirs/_tddr.py
```diff
@@ -47,29 +52,34 @@ def temporal_derivative_distribution_repair(raw):
...
+    signal : ndarray
+        Uncorrected optical density data (sample x channel).
+    sample_rate : float
+        Sampling rate in Hz.
+
+    Returns
+    -------
+    signal_corrected : ndarray
+        Corrected optical density data (sample x channel).
+    """
```
larsoner, line 82: Please revert this, the old code was closer to copy-paste from the upstream repo which we'd like to preserve (easier for updating, etc.), see comment above about only modifications...

cbrnr:
> Probably you'll need to add it here:
>
> https://github.com/mne-tools/mne-python/blob/master/mne/tests/test_docstring_parameters.py#L199
>
> I guess it wasn't obvious from looking at the file yourself? If so, please suggest better names to use there.

It wasn't obvious because this list is not used in `test_docstring_parameters` (which I thought was responsible for performing the tests). I saw the list and the name is pretty obvious, it just wasn't clear where exactly it is used.

larsoner:
Ctrl-F for the name shows here:

https://github.com/mne-tools/mne-python/blob/master/mne/tests/test_docstring_parameters.py#L258-L300

mne/preprocessing/nirs/_tddr.py
cbrnr: I didn't know that there was an upstream .py file - I thought @rob-luke ported MATLAB code to Python. OK, I'll revert.

mne/preprocessing/nirs/_tddr.py
cbrnr: Done

doc/changes/latest.inc
cbrnr: Done

cbrnr:
If I had taken a bit more time to look at the code I would have figured it out, but my focus was too narrow because you explicitly mentioned the other function. So I guess no need to change anything there.

cbrnr:
@larsoner anything else to do for this PR?

doc/changes/latest.inc
```diff
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Added temporal derivative distribution repair :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repair` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
```
larsoner, line 42: Would be good to revert this one, too, since we want to link to the rendered doc. Or maybe
```suggestion
- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repar` with shortened alias ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
```

doc/changes/latest.inc
```diff
@@ -39,7 +39,7 @@ Changelog
 
 - :func:`mne.viz.plot_dipole_locations` and :meth:`mne.Dipole.plot_locations` gained a ``title`` argument to specify a custom figure title in ``orthoview`` mode by `Richard Höchenberger`_
 
-- Add temporal derivative distribution repair (TDDR) ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
+- Add temporal derivative distribution repair (TDDR) :func:`mne.preprocessing.nirs.temporal_derivative_distribution_repar` with shortened alias ``mne.preprocessing.nirs.tddr`` by `Robert Luke`_
```
drammock, line 42: Typo in crosstef (repar -> repair)

larsoner:
CircleCI failure is a false alarm, shouldn't cause problems on `master`

cbrnr:
Pretty hard to spot these warnings. I assume the log explicitly mentions `plot_fnirs_artifact_removal.py` because it changed (and everything else is cached somewhere)?

larsoner:
> I assume the log explicitly mentions plot_fnirs_artifact_removal.py because it changed (and everything else is cached somewhere)?

CircleCI only builds modified examples (makes it take ~10 minutes instead of ~2h). If you look at the output you will see that only that example was built

hoechenberger:
> Pretty hard to spot these warnings.

Yeah that's something I'm also struggling with quite frequently. Would there be any way to highlight those errors / warnings better?

larsoner:
I personally don't find it too difficult, I open the log and ctrl-F "warning" and it's always one of the first few results. In theory we could make the script more complex by teeing the output to a log and, on failure, `grep`ping this to make it show up at the bottom.

cbrnr:
It would be nice to have a summary of all warnings and errors at the bottom, but if you know what to search ("warning" and "error") it probably isn't too inconvenient.

larsoner:
Thanks @cbrnr

drammock:
all of this back and forth about reading the circle logs tells me that we need this info in an "advanced dev docs" page.  I'll add that to #6266

PR #7801, opened 2020-05-20
WIP, MAINT: Simpler VTK 9 on PyPI

larsoner:
VTK has [landed on PyPI](https://pypi.org/project/vtk/9.0.0/#files) so we can simplify our installs accordingly.


larsoner:
Actually apparently the Linux ones aren't there yet, but should be by the end of the week

Issue #7790, opened 2020-05-15
Error in reading data from Fieldtrip


gbrookshire:
#### Describe the bug

I'm unable to read data saved by Fieldtrip using the functions in `mne.io.fieldtrip`. Any suggestions would be greatly appreciated!

#### Steps to reproduce

Download the Fieldtrip data from this link:
<ftp://ftp.fieldtriptoolbox.org/pub/fieldtrip/tutorial/coherence/data.mat>

```python
import mne
data_from_ft = mne.io.fieldtrip.read_epochs_fieldtrip('data.mat', None)
```

#### Expected results
Load the data saved in the Fieldtrip object.

#### Actual results
```
In [5]: data_from_ft = mne.io.fieldtrip.read_epochs_fieldtrip('data.mat', None)                                
<ipython-input-5-6596278ed022>:1: RuntimeWarning: Importing FieldTrip data without an info dict from the original file. Channel locations, orientations and types will be incorrect. The imported data cannot be used for source analysis, channel interpolation etc.
  data_from_ft = mne.io.fieldtrip.read_epochs_fieldtrip('data.mat', None)
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-5-6596278ed022> in <module>
----> 1 data_from_ft = mne.io.fieldtrip.read_epochs_fieldtrip('data.mat', None)

~/anaconda3/envs/mne/lib/python3.7/site-packages/mne/io/fieldtrip/fieldtrip.py in read_epochs_fieldtrip(fname, info, data_name, trialinfo_column)
    117     ft_struct = ft_struct[data_name]
    118 
--> 119     info = _create_info(ft_struct, info)  # create info structure
    120     data = np.array(ft_struct['trial'])  # create the epochs data array
    121     events = _create_events(ft_struct, trialinfo_column)

~/anaconda3/envs/mne/lib/python3.7/site-packages/mne/io/fieldtrip/utils.py in _create_info(ft_struct, raw_info)
     60         pick_info(info, ch_idx, copy=False)
     61     else:
---> 62         montage = _create_montage(ft_struct)
     63 
     64         info = create_info(ch_names, sfreq, montage=montage)

~/anaconda3/envs/mne/lib/python3.7/site-packages/mne/io/fieldtrip/utils.py in _create_montage(ft_struct)
    143                 cur_labels[available_channels])
    144             montage_pos.extend(
--> 145                 cur_ch_struct['chanpos'][available_channels])
[...]

PR #7802, opened 2020-05-20
[MRG] [FIX] FieldTrip Import: Missing channels in epoched data and better error messages

thht:
This PR fixes problems made apparent by #7790 namely:

1. An error was thrown when epoched data was imported with a `Info` object and channels were missing.
2. FieldTrip supports non-uniform times for epochs which MNE does not. This used to lead to a non-meaningful exception.
3. FieldTrip updated the format of their sensor definitions a while back. Importing files that still use the old format now leads to a meaningful error telling the user how to fix the issue.

This PR needs an updated of the testing data. I will issue the PR right away.


thht:
ok, no idea why circleci is still pending. the documentation seems to have built ok.
azure crashes when it wants to install mayavi.

larsoner:
> ok, no idea why circleci is still pending

Just some weird artifact redirection bug

> azure crashes when it wants to install mayavi.

Fixed by #7801 and restarted the run

LGTM +1 for merge

larsoner:
Thanks @thht

PR #7804, opened 2020-05-20
MRG: warning banner for old versions of docs

### LINKED ISSUE #7782: DOC: old versions of docs website should link to current stable
Lots of packages (numpy, scipy, etc) have prominent messages on their outdated docs websites, saying something like

> You are viewing docs for version XX, click here for current version (YY) docs

We need to do this too; a colleague of mine just got routed by google to the docs from version 0.15 (!) and didn't immediately notice the outdated version.

Additionally, I've mentioned before somewhere that the old docs version dropdown doesn't show versions newer than itself:

Screenshot_2020-05-13_11-33-34

But that's a more annoying problem to fix than just adding a big colorful banner at the top saying "these docs are old", and I think if we have the big banner then we can ignore the dropdown problem.

drammock:
closes #7782 

- adds a call to `versionwarning.js` to the layout template
- styles the links in the resulting devbar

WIP until we iron out the style-related questions in [mne-tools.github.io#20](https://github.com/mne-tools/mne-tools.github.io/pull/20)

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -122,6 +122,11 @@ blockquote {
   border-radius: 0 0 4px 4px !important;
 }
 
+.devbar a {
+    color: #fff;
+    font-weight: bold;
+}
+
 .pad-top-30 {
   padding-top: 30px;
 }
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -2,6 +2,7 @@
 
 {% block extrahead %}
 
+    <link rel="canonical" href="https://mne.tools/stable/index.html" />
     <script type="text/javascript" src="{{ pathto('_static/copybutton.js', 1) }}"></script>
 
 {% if use_google_analytics|tobool %}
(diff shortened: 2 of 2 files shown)


larsoner:
Indeed it currently has the double-bar

https://20208-1301584-gh.circle-artifacts.com/0/dev/index.html

drammock:
> Indeed it currently has the double-bar
> 
> https://20208-1301584-gh.circle-artifacts.com/0/dev/index.html

fixed the Javascript, no more double-bar on Circle builds.  This is now ready for review/merge; I'm OK with how it is but open to suggestions about how to change the wording / formatting of the devbar message.

drammock:
BTW, this is how it currently looks on the old version of the docs (so far only on 0.11): https://mne.tools/0.11/index.html

agramfort:
I don's see it here https://mne.tools/0.15/ normal?

>

drammock:
@agramfort so far it's only been hacked into the homepage of 0.11 as a proof of concept (see link in my prev comment).  I'm right this minute working on a PR into mne-tools.github.io to put on every page of all old versions.

larsoner:
Thanks @drammock

larsoner:
Should be working now

PR #7806, opened 2020-05-21
MRG, MAINT: Smaller PNGs during doc build

larsoner:
Also modernizes our inclusion of backrefs, and considates CircleCI `apt-get` to a single line.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,12 @@ jobs:
               - pip-cache
 
         - run:
-            name: Install 3D rendering libraries
+            name: Install 3D rendering libraries \ PyQt5 dependencies \ graphviz \ optipng (for optimized images)
             command: |
-              sudo apt-get install libosmesa6 libglx-mesa0 libopengl0 libglx0 libdbus-1-3
+              sudo apt-get install libosmesa6 libglx-mesa0 libopengl0 libglx0 libdbus-1-3 \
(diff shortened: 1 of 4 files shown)


larsoner:
Manually killing Travis and Azure non-style because they are not relevant

larsoner:
Seems to work, in it goes

agramfort:
did you quantify the gain?

larsoner:
In the Sphinx-gallery PR I did some testing suggesting you could get 20-25% gains with little added build time, but didn't actually test here. Looking at the `mne-tools.github.io` right now, we have `du -hs` of `dev/_images` of 98M and the last couple builds took 1:21:55 and 1:28:04. The `master` build from merging this PR timed out on `plot_mne_dspm_source_localization`, so we'll have to wait for another hour and a half to see the sizes and times now...

larsoner:
I was wrong about the times above, should have listed them as 1:45:40 and 1:53:23. CircleCI finished and uploaded, looks like before/after:

- `dev/_images` size decreases from 98M to 79M (thumbs only: 2.9M to 2.4M)
- Build time increases from 1:53:23 to 2:00:30

There is a lot of variability in the build times, so this could actually be a bit better or worse. But it doesn't look like it's going to end up being too bad.

agramfort:
ok so a 20% decrease roughly

>

larsoner:
yeah

PR #7809, opened 2020-05-22
Add wrapper read_raw for specific read_raw_xxx readers

### LINKED ISSUE #7769: Add generic reader function?
At some point we've already discussed if adding a generic reader function (`mne.io.read_raw`) would be useful. I can't find the relevant discussion, but I think that we decided that users should stick to the separate readers.

I've written such a wrapper function for MNELAB (https://github.com/cbrnr/mnelab/pull/148), and I think this might still be a good idea to include in MNE. It makes things easier for users - the file extension already determines the file type, so why do they have to explicitly pick the right function if this could be done automatically?

Of course this is not meant to replace the individual readers - it's just a convenience wrapper that should work for the most common file types.

```python
mne.io.read_raw("test.vhdr")
mne.io.read_raw("test.edf")
mne.io.read_raw("test.bdf")
mne.io.read_raw("test.set")
# ... you get the point
```

I'm fine if people decide (again) not to include this ("There should be one-- and preferably only one --obvious way to do it"). But in that case, I'd like to ask if anyone has a good idea how to include the true MNE reader function calls in MNELAB's history without creating too big of a mess. Currently, I stick to just using `mnelab.io.read_raw` for history entries because it keeps the code so much simpler. Suggestions are very much appreciated (either here or at https://github.com/cbrnr/mnelab/pull/148)!

cbrnr:
Fixes #7769.

This PR adds `mne.io.read_raw`, which is "just" a wrapper for specific readers. Based on the file extension, the wrapper dispatches known specific readers (and raises an error for unknown file types). I've included a suggested file type in the error message if the user tried to open a wrong file from formats consisting of multiple files (e.g. the BrainVision format has `.eeg`, `.vmrk`, and `.vhdr` files, of which only the last can be opened).

This is a first suggestion, please comment. We should also add a test - any suggestions what the test should do? I thought that we'd use `read_raw` to load each supported file type, and probably also an unsupported one.

### DIFF
--- a/mne/io/_read_raw.py
+++ b/mne/io/_read_raw.py
@@ -1,0 +1,80 @@
+"""Generic wrapper function read_raw for specific read_raw_xxx readers."""
+
+# Authors: Clemens Brunner <clemens.brunner@gmail.com>
+#
+# License: BSD (3-clause)
+
+
+from pathlib import Path
+from functools import partial
+
+from . import (read_raw_edf, read_raw_bdf, read_raw_gdf, read_raw_brainvision,
+               read_raw_fif, read_raw_eeglab, read_raw_cnt, read_raw_egi,
+               read_raw_eximia, read_raw_nirx, read_raw_fieldtrip,
+               read_raw_artemis123, read_raw_nicolet, read_raw_kit,
+               read_raw_ctf)
+
+
+def _read_unsupported(fname, **kwargs):
+    ext = "".join(Path(fname).suffixes)
+    msg = f"Unsupported file type ({ext})."
+    suggest = kwargs.get("suggest")
+    if suggest is not None:
+        msg += f" Try reading a {suggest} file instead."
+    msg += " Consider using a dedicated reader function for more options."
+    raise ValueError(msg)
+
+
+# supported read file formats
+supported = {".edf": read_raw_edf,
+             ".bdf": read_raw_bdf,
+             ".gdf": read_raw_gdf,
+             ".vhdr": read_raw_brainvision,
+             ".fif": read_raw_fif,
+             ".fif.gz": read_raw_fif,
+             ".set": read_raw_eeglab,
+             ".cnt": read_raw_cnt,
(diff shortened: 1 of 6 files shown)


mne/io/_read_raw.py
```diff
@@ -0,0 +1,80 @@
...
+readers = {**supported, **suggested}
+
+
+def read_raw(fname, *, preload=False, verbose=None, **kwargs):
+    """Read raw file.
+
+    Parameters
+    ----------
+    fname : str
+        File name to load.
```
agramfort, line 60: add doc for preload and verbose

cbrnr: How exactly does this Sphinx `%(preload)s` business work? It doesn't seem to work in this function so I'm guessing I have to add it to some configuration.

larsoner: You need to add the `@verbose` or `@fill_doc` decorator to the function, these do `%` formatting using our `docdict`. You can just use the latter because each reader will have its own `@verbose` and there is no reason to apply it twice

cbrnr: Thanks @larsoner!

cbrnr:
I assume the error on Azure (Windows 3.7 conda) is unrelated? It said this job has failed several times in the past, so is there a way to fix it?

`FAILED mne/minimum_norm/tests/test_inverse.py::test_make_inverse_operator_diag[testing_data-testing_data]`

larsoner:
Yes we'll need to fix it. My money is on it being a hard-to-track-down (and even harder to get fixed) MKL bug. So for now feel free to add a `if os.getenv('AZURE_CI_WINDOWS'`-style skip with a big TODO comment saying that it should be fixed, and/or open an issue to fix it

cbrnr:
Failing Azure job issue: #7814

I've added some tests, this should be good to go.

cbrnr:
All green!

hoechenberger:
Awesome work, @cbrnr!

PR #7811, opened 2020-05-22
MRG, FIX: Fix tick labeling

larsoner:
Latest `matplotlib` complains if `set_x/yticklabels(...)` count does not match the number of ticks, which is reasonable. This PR makes it so that we set them consistently, and tests now pass on my machine other than some issue with `annotations`, which I plan to also fix in this PR (hence the WIP).


larsoner:
Annotation problem is upstream and unreleased (AFAIK) https://github.com/matplotlib/matplotlib/pull/17489, so I'll merge once CIs come back happy here

Issue #7815, opened 2020-05-23
ENH: Automatically compute threshold for CTPS cardiac artifacts detection


yh-luo:
#### Describe the problem
Current default threshold is 0.25 for `mne.preprocessing.ICA.find_bads_ecg` with `method='ctps'`.
However, the value seems arbitrary. Would it make sense to implement automatic computation for CTPS threshold?

0.25 is actually the threshold for data with 400 Hz sampling rate. I found that 0.25 is sometimes not sensitive enough for clean data with sampling rate higher than 400 Hz.

For example, the sampling rate of `sample` dataset is 600 Hz. After filtering, `mne.preprocessing.ICA.find_bads_ecg` does not detect any ECG related IC using default (0.25) threshold. Using 0.21 (computed from the formula), it is able to detect one ECG related IC.

The ECG Evoked
![ecg_evoked](https://user-images.githubusercontent.com/29604854/82733885-a71b9e80-9d49-11ea-86a8-1de0815b70f9.png)

The ECG scores of ICs
![ecg_scores](https://user-images.githubusercontent.com/29604854/82733896-b7cc1480-9d49-11ea-8469-59e9cb324f77.png)

The detected IC (ICA001) with `threshold=0.21`
![ecg_ic_properties](https://user-images.githubusercontent.com/29604854/82733876-9a974600-9d49-11ea-8e9b-87fce4139efd.png)
![ecg_ic_overlay](https://user-images.githubusercontent.com/29604854/82733881-a2ef8100-9d49-11ea-812f-7344db050382.png)
![ica_components_on_raw](https://user-images.githubusercontent.com/29604854/82733964-31fc9900-9d4a-11ea-9d9f-20a95f4b8839.png)

#### Describe your solution

In the [reference](http://ieeexplore.ieee.org/document/4536072/), the threshold was set to Pk >= 10^-20. Pk can be computed from formula (13) with λ and λ will change according to the number of data points. Thus, the threshold will also change according to the sampling rate.

The threshold of the Kuiper index can be determined easily:

```python
import numpy as np


def _get_ctps_threshold(N=1000, Pk_threshold=1e-20):
    vs = [x / 100 for x in range(1, 100)]
    Pks = list()
    C = np.sqrt(N) + 0.155 + 0.24 / np.sqrt(N)
    # because when k gets large, only k=1 matters for the summation
    # k*v*C thus becomes v*C
    for v in vs:
[...]

PR #7819, opened 2020-05-24
MRG, ENH: Automatically compute threshold for CTPS ECG detection

yh-luo:
#### Reference issue
Implementation in #7815 

#### Additional information
After a deprecation cycle, TODO part will replace the current default value 0.25.
If that's ok, I will update `latest.inc`

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -75,6 +75,8 @@ Changelog
 
 - Add generic reader function :func:`mne.io.read_raw` that loads files based on their extensions (it wraps the underlying specific ``read_raw_xxx`` functions) by `Clemens Brunner`_
 
+- Add ``'auto'`` option to :meth:`mne.preprocessing.ICA.find_bads_ecg` to automatically determine the threshold for CTPS method by `Yu-Han Luo`_
+
 Bug
 ~~~
 
@@ -174,3 +176,5 @@ API
 - Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
 
 - The function ``mne.channels.read_dig_captrack`` will be deprecated in version 0.22 in favor of :func:`mne.channels.read_dig_captrak` to correct the spelling error: "captraCK" -> "captraK", by `Stefan Appelhoff`_
+
+- The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -1165,11 +1165,12 @@ class ICA(ContainsMixin):
             The name of the channel to use for ECG peak detection.
             The argument is mandatory if the dataset contains no ECG
             channels.
-        threshold : float | str | None
-            The value above which a feature is classified as outlier. If
-            None and method is 'ctps', defaults to 0.25. Can be 'auto' to
-            automatically determine the threshold for 'ctps' method. If None
-            and method is 'correlation', defaults to 3.0.
+        threshold : float | str
+            The value above which a feature is classified as outlier. If 'auto'
+            and method is 'ctps', automatically compute the threshold. If
+            'auto' and method is 'correlation', defaults to 3.0. The default
+            translates to 0.25 for 'ctps' and 3.0 for 'correlation' in version
+            0.21 but will change to 'auto' in version 0.22.
 
             .. versionchanged:: 0.21
         start : int | float | None
@@ -1235,10 +1236,9 @@ class ICA(ContainsMixin):
 
         if method == 'ctps':
             if threshold is None:
-                warn('The default for "threshold" for CTPS method will change '
-                     'from 0.25 to "auto" in version 0.22. To avoid this '
-                     'warning, explicitly set threshold to "auto" for '
-                     'automatic computation.',
+                warn('The default for "threshold" will change from None to'
+                     '"auto" in version 0.22. To avoid this warning, '
+                     'explicitly set threshold to "auto".',
                      DeprecationWarning)
                 threshold = 0.25
             elif threshold == 'auto':
@@ -1272,10 +1272,12 @@ class ICA(ContainsMixin):
             self.labels_['ecg/%s' % ch_name] = list(ecg_idx)
         elif method == 'correlation':
             if threshold is None:
+                warn('The default for "threshold" will change from None to'
+                     '"auto" in version 0.22. To avoid this warning, '
+                     'explicitly set threshold to "auto".',
+                     DeprecationWarning)
                 threshold = 3.0
             elif threshold == 'auto':
-                warn('Only CTPS method supports automatic computation for '
-                     'threshold. Threshold is set to 3.0 (default)')
                 threshold = 3.0
             self.labels_['ecg'], scores = self._find_bads_ch(
                 inst, [ecg], threshold=threshold, start=start, stop=stop,
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -1115,6 +1115,28 @@ class ICA(ContainsMixin):
 
         return labels, scores
 
+    def _get_ctps_threshold(self, Pk_threshold=1e-20):
+        """Automatically decide the Kuiper index threshold for CTPS method.
+
+        References
+        ----------
+        [1] Dammers, J., Schiek, M., Boers, F., Silex, C., Zvyagintsev,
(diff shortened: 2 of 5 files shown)


mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
 
         return labels, scores
 
+    def _get_ctps_threshold(self, Pk_threshold=1e-20):
```
agramfort, line 1118:
```suggestion
    def _get_ctps_threshold(self, pk_threshold=1e-20):
```
you should not have capitals in snake_case variable names if you follow pep8 rule.

mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
...
+        References
+        ----------
+        [1] Dammers, J., Schiek, M., Boers, F., Silex, C., Zvyagintsev,
+            M., Pietrzyk, U., Mathiak, K., 2008. Integration of amplitude
+            and phase statistics for complete artifact removal in independent
+            components of neuromagnetic recordings. Biomedical
+            Engineering, IEEE Transactions on 55 (10), 2353-2362.
+        """
+        N = self.info['sfreq']
+        vs = [x / 100 for x in range(1, 100)]
```
agramfort, line 1130:
```suggestion
        vs = np.arange(1, 100) / 100
```

mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
...
+        Pks = list()
+        C = np.sqrt(N) + 0.155 + 0.24 / np.sqrt(N)
```
agramfort, line 1132: use sqrt from math

```
import math as m
m.sqrt
```

makes it explicit that its the sqrt of a float and not an array

mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
...
+        # when k gets large, only k=1 matters for the summation
+        # k*v*C thus becomes v*C
+        for v in vs:
+            Pk = 2 * (4 * (v * C)**2 - 1) * (np.exp(-2 * (v * C)**2))
+            Pks.append(abs(Pk - Pk_threshold))
```
agramfort, line 1137: please do this with numpy operations and without for loop

mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
...
+        return vs[Pks.index(min(Pks))]
```
agramfort, line 1138: np.argmin is your friend

mne/preprocessing/ica.py
```diff
@@ -1198,7 +1220,14 @@ def find_bads_ecg(self, inst, ch_name=None, threshold=None, start=None,
 
         if method == 'ctps':
             if threshold is None:
+                warn('The default threshold will change from 0.25 to'
+                     'automatical computation in version 0.22.',
```
agramfort, line 1224:
```suggestion
                     'automatic computation in version 0.22.',
```
you should tell users set threshold to 'auto' to avoid the warning

mne/preprocessing/ica.py
```diff
@@ -1115,6 +1115,28 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
 
         return labels, scores
 
+    def _get_ctps_threshold(self, Pk_threshold=1e-20):
+        """Automatically decide the Kuiper index threshold for CTPS method.
+
```
agramfort, line 1120: can you point to equations in paper to that explain how this works?

mne/preprocessing/ica.py
yh-luo: In the paper, Pk was log-transformed to pk: pk = -log10(Pk). Because the threshold value used here was Pk, I wondered if `Pk_threshold` would be more appropriate than `pk_threshold`? or maybe `significance_level` would be better?

agramfort:
so Pk is a p-value?

>

yh-luo:
Pk is a value to evaluate whether an IC is ECG-related.

Pk

agramfort:
I would just call it threshold and explain in the docstring to what it
corresponds in the paper.

mne/preprocessing/ica.py
```diff
@@ -1133,9 +1165,13 @@ def find_bads_ecg(self, inst, ch_name=None, threshold=None, start=None,
             The name of the channel to use for ECG peak detection.
             The argument is mandatory if the dataset contains no ECG
             channels.
-        threshold : float
+        threshold : float | str | None
```
agramfort, line 1168: I think we should deprecate the None option and move to 'auto' in next versions. In 0.21 we get a warning to set it to auto to avoid warning and in 0.22 we make 'auto' the default and deprecate None.

yh-luo: ok

mne/preprocessing/tests/test_ica.py
```diff
@@ -651,7 +656,7 @@ def f(x, y):
     idx, scores = ica.find_bads_eog(evoked, ch_name='MEG 1441')
     assert_equal(len(scores), ica.n_components_)
 
-    idx, scores = ica.find_bads_ecg(evoked, method='correlation')
+    idx, scores = ica.find_bads_ecg(evoked, method='correlation', threshold=3)
```
larsoner, line 659: Can this be changed to `threshold='auto'` and still work?

mne/preprocessing/ica.py
```diff
@@ -1115,12 +1116,43 @@ def _find_bads_ch(self, inst, chs, threshold=3.0, start=None,
...
+            and phase statistics for complete artifact removal in independent
+            components of neuromagnetic recordings. Biomedical
+            Engineering, IEEE Transactions on 55 (10), pp.2356.
+        """
+        N = self.info['sfreq']
+        Vs = np.arange(1, 100) / 100
+        C = math.sqrt(N) + 0.155 + 0.24 / math.sqrt(N)
+        # in formula (13), when k gets large, only k=1 matters for the
+        # summation. k*V*C thus becomes V*C
+        Pks = 2 * (4 * (Vs * C)**2 - 1) * (np.exp(-2 * (Vs * C)**2))
```
larsoner, line 1142: we have PEP8 requirements for spaces around operators (I think?) so `C)**2` is likely going to be a problem...

yh-luo: When I put spaces around `**`, both `autopep8` and `yapf` removed the surrounding spaces. Also Flake passed in the current version, I'm afraid that put spaces around `**` would violate PEP8...

larsoner: Indeed you're right, `flake8` passes locally!

agramfort:
good to go?

>

mne/preprocessing/tests/test_ica.py
larsoner: This has not been checked yet

yh-luo: Thanks for reminding, I've checked and updated the influenced tests and examples.

yh-luo:
`threshold='auto'` passed for both ctps and correlation methods locally, so I updated `test_ica.py` to use the new `threshod='auto'`

BTW, the computed threshold in `plot_run_ica.py` is 0.41 (the original default is 0.25) because the data were resampled to 150 Hz. Though both thresholds converged to the same components, I think it would be better to determine the threshold based on data sampling rate. Thus the threshold in `plot_run_ica.py` for CTPS method was updated to `'auto'` and not the current default 0.25.
To make sure the detected components are ECG related, the estimated source of the components are visualized.

agramfort:
@yh-luo you will need to update latest.inc file to get credit and document this improvement. thanks

doc/changes/latest.inc
```diff
@@ -174,3 +176,5 @@ API
 - Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
 
 - The function ``mne.channels.read_dig_captrack`` will be deprecated in version 0.22 in favor of :func:`mne.channels.read_dig_captrak` to correct the spelling error: "captraCK" -> "captraK", by `Stefan Appelhoff`_
+
+- The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
```
agramfort, line 180: you need you add your name to names.inc too

yh-luo: thank you! I had added my name to `names.inc` in previous PR, so I did not in this one. Locally built documentation works as expected.

agramfort:
thx @yh-luo !

PR #7822, opened 2020-05-25
Remove double spaces in reprs

cbrnr:
This PR changes `  |  ` (double spaces) to ` | ` (single spaces) in various reprs to make them consistent with the recently changed reprs for `Annotation`, `Info`, and `Raw`.


larsoner:
Thanks @cbrnr

PR #7823, opened 2020-05-25
Deprecate meg=True in pick_types

### LINKED ISSUE #7710: Set all types to False in pick_types
Would it make sense to set all arguments in `pick_types` to `False` by default? In other words, currently everything is `False` except for `meg`. I know where this is coming from, but for people working with EEG, NIRS, or other non-MEG signals, we always have to include `meg=False` in addition to e.g. `eeg=True`.

cbrnr:
Fixes #7710 (change the default `meg=True` to `meg=False` in a deprecation cycle.

- [x] Fix/prepare examples/tutorials/etc.
- [x] Change also in `pick_types_forward`?
- [x] ~~Any additional MEG bias we should remove in this PR?~~
- [x] ~~Make arguments kw-only?~~

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -183,4 +183,4 @@ API
 
 - The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
 
-- The default argument ``meg=True`` in func:`mne.io.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
+- The default argument ``meg=True`` in :func:`mne.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -177,8 +177,10 @@ API
 
 - The method ``stc_mixed.plot_surface`` for a :class:`mne.MixedSourceEstimate` has been deprecated in favor of :meth:`stc.surface().plot(...) <mne.MixedSourceEstimate.surface>` by `Eric Larson`_
 
-- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
+- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates by `Luke Bloy`_
 
 - The function ``mne.channels.read_dig_captrack`` will be deprecated in version 0.22 in favor of :func:`mne.channels.read_dig_captrak` to correct the spelling error: "captraCK" -> "captraK", by `Stefan Appelhoff`_
 
 - The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
+
+- The default argument ``meg=True`` in func:`mne.io.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -633,7 +633,7 @@ class UpdateChannelsMixin(object):
     """Mixin class for Raw, Evoked, Epochs, AverageTFR."""
 
     @verbose
-    def pick_types(self, meg=True, eeg=False, stim=False, eog=False,
+    def pick_types(self, meg=None, eeg=False, stim=False, eog=False,
                    ecg=False, emg=False, ref_meg='auto', misc=False,
                    resp=False, chpi=False, exci=False, ias=False, syst=False,
                    seeg=False, dipole=False, gof=False, bio=False, ecog=False,
@@ -644,10 +644,9 @@ class UpdateChannelsMixin(object):
         Parameters
         ----------
         meg : bool | str
-            If True include all MEG channels. If False include None.
-            If string it can be 'mag', 'grad', 'planar1' or 'planar2' to select
-            only magnetometers, all gradiometers, or a specific type of
-            gradiometer.
+            If True include MEG channels. If string it can be 'mag', 'grad',
+            'planar1' or 'planar2' to select only magnetometers, all
+            gradiometers, or a specific type of gradiometer.
         eeg : bool
             If True include EEG channels.
         stim : bool
@@ -659,8 +658,10 @@ class UpdateChannelsMixin(object):
         emg : bool
             If True include EMG channels.
         ref_meg : bool | str
-            If True include CTF / 4D reference channels. If 'auto', the
-            reference channels are only included if compensations are present.
+            If True include CTF / 4D reference channels. If 'auto', reference
+            channels are included if compensations are present and ``meg`` is
+            not False. Can also be the string options for the ``meg``
+            parameter.
         misc : bool
             If True include miscellaneous analog channels.
         resp : bool
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -421,10 +421,12 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
                                'fnirs_raw', 'fnirs_od')
             if ch_type in ('grad', 'mag'):
                 pick[k] = _triage_meg_pick(info['chs'][k], meg)
-                deprecation_warn = True
+                if meg_default_arg:
+                    deprecation_warn = True
             elif ch_type == 'ref_meg':
                 pick[k] = _triage_meg_pick(info['chs'][k], ref_meg)
-                deprecation_warn = True
+                if meg_default_arg:
+                    deprecation_warn = True
             else:  # ch_type in ('hbo', 'hbr')
                 pick[k] = _triage_fnirs_pick(info['chs'][k], fnirs)
 
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -373,9 +373,12 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
     # NOTE: Changes to this function's signature should also be changed in
     # PickChannelsMixin
     if meg is None:
-        meg = True
-        warn("The default of meg=True will change to meg=False in version 0.22"
-             ", set meg explicitly to avoid this warning.", DeprecationWarning)
+        meg = True  # previous default arg
+        meg_default_arg = True  # default argument for meg was used
+    # only issue deprecation warning if there are MEG channels in the data and
+    # if the function was called with the default arg for meg
+    deprecation_warn = False
+
     exclude = _check_info_exclude(info, exclude)
     nchan = info['nchan']
     pick = np.zeros(nchan, dtype=np.bool)
@@ -407,6 +410,8 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
             param_dict[key] = fnirs
     for k in range(nchan):
         ch_type = channel_type(info, k)
+        if ch_type in ('grad', 'mag') and meg_default_arg:
+            deprecation_warn = True
         try:
             pick[k] = param_dict[ch_type]
         except KeyError:  # not so simple
@@ -414,8 +419,10 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
                                'fnirs_raw', 'fnirs_od')
             if ch_type in ('grad', 'mag'):
                 pick[k] = _triage_meg_pick(info['chs'][k], meg)
+                deprecation_warn = True
             elif ch_type == 'ref_meg':
                 pick[k] = _triage_meg_pick(info['chs'][k], ref_meg)
+                deprecation_warn = True
             else:  # ch_type in ('hbo', 'hbr')
                 pick[k] = _triage_fnirs_pick(info['chs'][k], fnirs)
 
@@ -437,6 +444,9 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
     else:
         sel = pick_channels(info['ch_names'], myinclude, exclude)
 
+    if deprecation_warn:
+        warn("The default of meg=True will change to meg=False in version 0.22"
+             ", set meg explicitly to avoid this warning.", DeprecationWarning)
     return sel
 
 
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -374,8 +374,8 @@ def pick_types(info, meg=None, eeg=False, stim=False, eog=False, ecg=False,
     # PickChannelsMixin
     if meg is None:
         meg = True
-        warn("The default of meg=True will change to meg=False in version "
-             "0.22.", DeprecationWarning)
+        warn("The default of meg=True will change to meg=False in version 0.22"
+             ", set meg explicitly to avoid this warning.", DeprecationWarning)
     exclude = _check_info_exclude(info, exclude)
     nchan = info['nchan']
     pick = np.zeros(nchan, dtype=np.bool)
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -656,7 +656,7 @@ def pick_channels_forward(orig, include=[], exclude=[], ordered=False,
     return fwd
 
 
-def pick_types_forward(orig, meg=True, eeg=False, ref_meg=True, seeg=False,
+def pick_types_forward(orig, meg=None, eeg=False, ref_meg=True, seeg=False,
                        ecog=False, include=[], exclude=[]):
     """Pick by channel type and names from a forward operator.
 
@@ -664,10 +664,10 @@ def pick_types_forward(orig, meg=True, eeg=False, ref_meg=True, seeg=False,
     ----------
     orig : dict
         A forward solution.
-    meg : bool or str
-        If True include all MEG channels. If False include None
-        If string it can be 'mag' or 'grad' to select only gradiometers
-        or magnetometers.
+    meg : bool | str
+        If True include MEG channels. If string it can be 'mag', 'grad',
+        'planar1' or 'planar2' to select only magnetometers, all gradiometers,
+        or a specific type of gradiometer.
     eeg : bool
         If True include EEG channels.
     ref_meg : bool
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -12,7 +12,7 @@ import numpy as np
 
 from .constants import FIFF
 from ..utils import (logger, verbose, _validate_type, fill_doc, _ensure_int,
-                     _check_option)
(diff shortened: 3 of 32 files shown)


mne/io/pick.py
```diff
@@ -374,6 +372,10 @@ def pick_types(info, meg=True, eeg=False, stim=False, eog=False, ecg=False,
     """
     # NOTE: Changes to this function's signature should also be changed in
     # PickChannelsMixin
+    if meg is None:
+        meg = True
+        warn("The default of meg=True will change to meg=False in version "
+             "0.22.", DeprecationWarning)
```
larsoner, line 378: ", set `meg` explicitly to avoid this warning"

... and you should only emit this if there are actually MEG channels present. Otherwise the value of `meg` does not matter and people should not care/be warned (e.g., when working with just EEG or fNIRS data the `meg=True` default won't matter)

cbrnr: You don't want to warn if people use the default? If we only warn if there are MEG channels, then people won't know that something changed if they just happen to load a file without MEG channels.

I can do it, but this also makes the code a bit more complicated too.

larsoner: > You don't want to warn if people use the default?

Only want to warn them when it matters/changes behavior. This is what we usually do, and it cuts down on noise for people unaffected by the change.

> people won't know that something changed if they just happen to load a file without MEG channels.

This will only happen if someone uses a MEG+EEG file in 0.20 and got used to / expected that behavior, never use one in 0.21 (when they would get a warning), and then go back to using them again in 0.22, and they do not pay attention to the release notes. I say MEG+EEG because for MEG-only users, doing `pick_types(info, meg=False)` (default in 0.22) when you only have MEG channels things will either very clearly be wrong or you will get an error (I think).

This seems like an acceptable risk to take on given the benefits of not false-alarm notifying all our EEG, fNIRS, etc. users about things that are irrelevant to them.

cbrnr: OK, then I'll try to keep the changes to issue the warning only in those cases as minimal as possible. Also, I think picking `meg=False` on a MEG-only data set will return an empty selection and not an error - at least this is the current behavior and I wouldn't change it.

cbrnr:
@larsoner can you check if the logic when to warn is OK like this?

cbrnr:
Now do I catch all deprecation warnings individually or do I just disable this specific `DeprecationWarning` globally (do we have something like a `pytest.ini` config file?)?

larsoner:
You will probably need to update all tests that currently fail to now explicitly use meg=True  to ensure they work the same way now and once we switch to meg=False as the default

mne/io/pick.py
```diff
@@ -374,6 +372,15 @@ def pick_types(info, meg=True, eeg=False, stim=False, eog=False, ecg=False,
     """
     # NOTE: Changes to this function's signature should also be changed in
     # PickChannelsMixin
+    if meg is None:
+        meg = True  # previous default arg
+        meg_default_arg = True  # default argument for meg was used
+    else:
+        meg_default_arg = False
+    # only issue deprecation warning if there are MEG channels in the data and
```
larsoner, line 380: Indeed this looks like it should work, but I would write a little 10 line test with read_info to make sure / assert it is the case (can use pytest.deprecated_call for ensuring it warns at the right time)

larsoner:
Also you should push a commit with `[circle full]` in the message, it will show how many examples also break (we do not allow examples to give dep warnings) and thus also need to be updated

mne/io/pick.py
cbrnr: Done and works, except for "planar1" and "planar2" which I need to handle correctly. However, I don't understand what these values do. In the test, I have `info1 = create_info(6, 256, ["eeg", "mag", "grad", "misc", "grad", "hbo"])`. Why do I get `[2]` for `mne.pick_types(info1, meg="planar1")` and `[]` for `mne.pick_types(info1, meg="planar2")`?

larsoner: > Why do I get [2] for mne.pick_types(info1, meg="planar1") and [] for mne.pick_types(info1, meg="planar2")?

There are two types of grads in Neuromag systems (one computes the gradient in the X direction, the other Y). When you pass `meg='grad'` in `create_info`, it creates a `planar1` variant, not `planar2`. I wouldn't bother adding a `planar2`, it should be good enough to assert that you you do get `[2]` for `planar1` and `[]` for `planar2` in your example.

larsoner:
> Sphinx-gallery successfully executed 97 out of 190 files subselected by:

Looks like that's 93 examples to change :(

cbrnr:
> Looks like that's 93 examples to change :(

Yep, I'm on it.

mne/io/pick.py
cbrnr: But wait, what about index 4? That's also a 'grad' channel - why is it neither `'planar1'` nor `'planar2'`?

larsoner: Ah that's weird. Indeed it should be one or the other. Must be a bug somewhere in that function I guess. Do you get `[2, 4]` when you do `meg='grad`, and/or `[1, 2, 4]` when you do `meg=True`?

cbrnr: Yes, that's exactly what I get for these two queries.

larsoner: Want me to look into it, or do you want to fix it here?

cbrnr: Please go for it!

larsoner: Okay, will look tomorrow if you don't beat me to it

larsoner:
> Change also in pick_types_forward?

Yes I think we should. I imagine this is used much more often internally than it is by users, so this won't inflict nearly as much pain for people than the primary `pick_types` deprecation will.

> Any additional MEG bias we should remove in this PR?

`pick_types` and related functions are the only MEG-biased ones I can think of offhand

> Make arguments kw-only?

`pick_types` has been around for 10 years now and I don't think we've ever tried to reorder the arguments, so I don't see too much point in doing so. There are probably people who have done `pick_types(info, False, True)` or `evoked.pick_types(False, True) knowing that meg/eeg are the first two non-info arguments, and I don't see a reason to break their code.

mne/io/pick.py
```diff
@@ -928,7 +944,8 @@ def _pick_aux_channels(info, exclude='bads'):
 def _pick_data_or_ica(info, exclude=()):
     """Pick only data or ICA channels."""
     if any(ch_name.startswith('ICA') for ch_name in info['ch_names']):
-        picks = pick_types(info, exclude=exclude, misc=True)
+        # FIXME: is meg=True really correct here?
+        picks = pick_types(info, exclude=exclude, misc=True, meg=True)
```
cbrnr, line 948: @larsoner I don't really understand why `meg=True` here - shouldn't this pick only ICA channels (i.e. `misc` channels)?

cbrnr:
Except for my question about `_pick_data_or_ica` this is ready!

mne/io/pick.py
larsoner: It's probably not correct. If you want to be paranoid you could do it both ways (make a `picks2`), `np.testing.assert_array_equal` the two, and run all `preprocessing` tests to double check, then remove the `meg=True` one knowing it never mattered. But it shouldn't, and if it does, it's probably a bug...

cbrnr: But we could do that in a separate issue so that this PR replicates the previous behavior? Or should I correct it (provided that `picks` equals `picks2` in the tests)?

larsoner: Either way is fine by me

cbrnr: OK, then I'd prefer tackling this in a new issue/PR.

cbrnr:
Last question solved, ready from my side!

larsoner:
needs rebase/merge, conflict

doc/changes/latest.inc
```diff
@@ -175,6 +175,8 @@ API
 
 - The method ``stc_mixed.plot_surface`` for a :class:`mne.MixedSourceEstimate` has been deprecated in favor of :meth:`stc.surface().plot(...) <mne.MixedSourceEstimate.surface>` by `Eric Larson`_
 
-- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates By `Luke Bloy`_
+- Add ``use_dev_head_trans`` parameter to :func:`mne.preprocessing.annotate_movement` to allow choosing the device to head transform is used to define the fixed cHPI coordinates by `Luke Bloy`_
 
 - The function ``mne.channels.read_dig_captrack`` will be deprecated in version 0.22 in favor of :func:`mne.channels.read_dig_captrak` to correct the spelling error: "captraCK" -> "captraK", by `Stefan Appelhoff`_
+
+- The default argument ``meg=True`` in func:`mne.io.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
```
larsoner, line 182: This is documented at the root level so this will be an error

https://mne.tools/stable/generated/mne.pick_types.html
```suggestion
- The default argument ``meg=True`` in func:`mne.pick_types` and related methods like :meth:`mne.io.Raw.pick_types` and :meth:`mne.Evoked.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
```

cbrnr: Just so I understand, where will this be an error (if the doc contains func:`mne.io.pick_types`)?

larsoner: > where will this be an error

CircleCI will say something about not being able to link to the ref properly, and the build will fail

larsoner: (alternatively, locally you can do `make html_dev-noplot` in the `doc/` directory then `echo $?` afterward it should not be zero, and if you look up you should see a warning about not being able to link properly)

larsoner: ... though I seem to remember your previous build being green, so maybe it figures out to go up to `mne.pick_types` when the `mne.io.pick_types` link is requested somehow?

cbrnr: The previous CircleCI build was all green with exit code 0, yes.

larsoner: Looks like it [was indeed green](https://app.circleci.com/pipelines/github/mne-tools/mne-python/1839/workflows/5bbacad8-9713-404d-b997-25aa61ef8818/jobs/20306/steps) but shouldn't have been because the [link did not get created correctly](https://20306-1301584-gh.circle-artifacts.com/0/dev/whats_new.html#api), maybe some latest Sphinx change made this fail to link without raising an error, not sure...

mne/channels/channels.py
```diff
@@ -644,10 +644,9 @@ def pick_types(self, meg=True, eeg=False, stim=False, eog=False,
         Parameters
         ----------
         meg : bool | str
-            If True include all MEG channels. If False include None.
-            If string it can be 'mag', 'grad', 'planar1' or 'planar2' to select
-            only magnetometers, all gradiometers, or a specific type of
-            gradiometer.
+            If True include MEG channels. If string it can be 'mag', 'grad',
+            'planar1' or 'planar2' to select only magnetometers, all
+            gradiometers, or a specific type of gradiometer.
```
hoechenberger, line 649:
```suggestion
            If ``True``, include MEG channels. If string, it can be ``'mag'``, 
            ``'grad``', ``'planar1'`` or ``'planar2'`` to select only magnetometers
             all gradiometers, or a specific type of gradiometer.
```

mne/io/pick.py
```diff
@@ -325,10 +324,9 @@ def pick_types(info, meg=True, eeg=False, stim=False, eog=False, ecg=False,
     emg : bool
         If True include EMG channels.
     ref_meg : bool | str
-        If True include CTF / 4D reference channels. If 'auto', the
-        reference channels included if compensations are present
-        and ``meg`` is not False. Can also be the string options allowed
-        for the ``meg`` parameter.
+        If True include CTF / 4D reference channels. If 'auto', reference
+        channels are included if compensations are present and ``meg`` is not
+        False. Can also be the string options for the ``meg`` parameter.
```
hoechenberger, line 329:
```suggestion
        If ``True``, include CTF / 4D reference channels. If ``'auto'``, reference
        channels are included if compensations are present and ``meg`` is not
        False. Can also be the string options for the ``meg`` parameter.
```

mne/io/pick.py
```diff
@@ -640,18 +655,18 @@ def pick_channels_forward(orig, include=[], exclude=[], ordered=False,
...
     orig : dict
         A forward solution.
-    meg : bool or str
-        If True include all MEG channels. If False include None
-        If string it can be 'mag' or 'grad' to select only gradiometers
-        or magnetometers.
+    meg : bool | str
+        If True include MEG channels. If string it can be 'mag', 'grad',
+        'planar1' or 'planar2' to select only magnetometers, all gradiometers,
+        or a specific type of gradiometer.
```
hoechenberger, line 669:
```suggestion
        If ``True``m include MEG channels. If string, it can be ``'mag'``,
        ``'grad'``, ``'planar1'`` or ``'planar2'`` to select only
        magnetometers, all gradiometers, or a specific type of gradiometer.
```

doc/changes/latest.inc
drammock: the linking failure is due to lack of a colon at the beginning of `func`

larsoner: Grrr... thanks @drammock . I'll see if changing our `default_role` to something other than `"autolink"` linke `"py:obj"` would have made this error show up, in which case we should change it

cbrnr: I've inserted the missing colon and now use `mne.pick_types`.

cbrnr:
@hoechenberger do we really always have to double-backtick literals in docstrings? I know that this makes it show up as `<code>`, but IIRC we've had this discussion in the past and decided that it's not worth the effort (and additional characters).

drammock:
> IIRC we've had this discussion in the past and decided that it's not worth the effort (and additional characters)

I don't understand the reasoning. We're not charged per-character, and the effort is about as minimal as it gets (press the key your finger is already on, twice instead of once).  I'll concede that the changelog is probably the least important place to get this right (much more important in tutorials and docstrings), but remembering / enforcing to do it correctly here is a good way to reinforce the behavior so that it also happens everywhere else.

hoechenberger:
+100 @drammock

cbrnr:
It's harder to read in plain text when I see everything surrounded by double backticks. But whatever, if we now do that I will change it.

cbrnr:
```python
meg : bool | str
            If ``True` include MEG channels. If string it can be ``'mag'``,
            ``'grad'``, ``'planar1'`` or ``'planar2'`` to select only
            magnetometers, all gradiometers, or a specific type of gradiometer.
```

You don't think this is harder to read (in plain text) (especially the literal strings)? Also, why change this only in a few selected parameter descriptions and not all? I don't want to do it here, if you think this is important please do it in a new PR.

larsoner:
> You don't think this is harder to read (in plain text) (especially the literal strings)?

I find it a little bit harder, but not too much. And to me the RST rendering gain is worth the plain-text rendering loss. I'm fine with someone else working on this in another PR, though

hoechenberger:
> You don't think this is harder to read (in plain text) (especially the literal strings)? 

I think it's not easy to read without syntax highlighting that helps with this.

That being said, our users view the rendered output, and many IDEs provide rendered docstrings as well. So unless you're actually working on the docstrings, this shouldn't really affect you, no?

> Also, why change this only in a few selected parameter descriptions and not all?

Because we're already inconsistent anyway and changing some bits "as we go" in PRs that change some lines of docstrings brings us a little closer to overall more consistently rendered docs.

> I don't want to do it here

There's no need to, it was merely suggestions, no blockers from my end. Sorry if it came across that way!

hoechenberger:
FWIW, I'm using VS Code with the [Highlight extension](https://marketplace.visualstudio.com/items?itemName=fabiospampinato.vscode-highlight) and with the following setting:

```javascript
    "highlight.regexes": {
        "(``)(.+?)(``)": [
            {"color": "#d88e73"},
            {"color": "#3c9ddb"},
            {"color": "#d88e73"}
        ]
    }
```

This produces:
Screenshot 2020-05-28 at 08 24 11

Maybe some of you may find this useful.

cbrnr:
That's pretty neat! Is there anything similar for PyCharm?

hoechenberger:
> That's pretty neat! Is there anything similar for PyCharm?

I'm afraid not, at least I haven't found anything so far :( There is a related support request from ~2 years ago:
https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000082599-Custom-syntax-highlighting

cbrnr:
OK, that's a pity. Though it wouldn't really solve the issue, because users (as opposed to devs) will read docstrings mostly in their REPL (e.g. IPython), which means they will see just plain text without any fancy rST highlighting. I think enclosing every literal in double backticks is just overkill.

Also, it violates the numpydoc style recommendation that variables should be enclosed in single backticks: https://numpydoc.readthedocs.io/en/latest/format.html.

hoechenberger:
Thanks @cbrnr 
@drammock, would there be a way to change the way we mark variable names and constants to make things more readable in plain text? Maybe we should open a new issue about this…

drammock:
> users (as opposed to devs) will read docstrings mostly in their REPL (e.g. IPython)

@cbrnr Do we actually have evidence for this one way or the other? Anyone who uses Spyder will typically see HTML-rendered docs, for example.

> would there be a way to change the way we mark variable names and constants to make things more readable in plain text?

@hoechenberger our sphinx config was recently changed so that anything surrounded by single ticks (which previously rendered as italics, redundant with single asterisks for yielding italics) will now always be treated as an attempted link or cross-reference by the build system. This allows Sphinx to catch errors like the recently seen

```
func:`some_function_name`
```
where `func:` was missing a leading colon. Previously it rendered as `func:italicized_function_name` and raised no error / warning. This change also means that the role prefix (`:func:`, `:class:` etc) can be omitted in most cases, which will allow us to simplify the plain-text docstrings considerably.  It's a trade-off.

cbrnr:
> @cbrnr Do we actually have evidence for this one way or the other? Anyone who uses Spyder will typically see HTML-rendered docs, for example.

I only have anecdotal evidence from coworkers and students in my Python course, where the majority seems to be using plain-text docstrings (mostly via the IPython shortcut `?command` or `command?`). Plain-text docs are always available no matter what editor or IDE you're using (even if HTML rendering is supported). Many editors don't support HTML rendering of docstrings out of the box (e.g. VS Code, PyCharm) - or at least I haven't figured out how to view HTML-rendered docs. It seems like Spyder is the only editor where this works out of the box.

👍 for the simplification! Regarding the use of double backticks, do we have docstring guidelines somewhere? If not I think it would be a good addition to have such a document in the contributing section.

cbrnr:
@drammock @hoechenberger do you have any additional comments (let's do docstring formatting changes in another PR though)?

hoechenberger:
@cbrnr LGTM!

hoechenberger:
@drammock I think you can merge this if you're happy

drammock:
> do we have docstring guidelines somewhere? If not I think it would be a good addition to have such a document in the contributing section.

added to #6266

drammock:
thanks @cbrnr !

PR #7824, opened 2020-05-25
MRG: Add support for reading and writing sufaces to .obj

wmvanvliet:
Add support for exporting FreeSurfer surfaces to .obj files so you can import them in blender.
blender

See also: https://github.com/ezemikulan/blender_freesurfer

Todo:

 - [x] Implement writing .obj files
 - [x] Implement reading .obj files
 - [x] Basic unit tests
 - [x] Finish unit tests for full coverage
 - [x] Add tutorial on how to use this to fix `Surface inner skull is not completely inside surface outer skull` errors
 - [x] Update what's new


mne/surface.py
```diff
@@ -959,33 +1009,52 @@ def write_surface(fname, coords, faces, create_stamp='', volume_info=None,
...
+                fid.write(_serialize_volume_info(volume_info))
+
+    elif file_format == 'obj':
+        with open(fname, 'w') as fid:
+            for line in create_stamp.splitlines():
+                fid.write('# {}\n'.format(line))
+            for v in coords:
+                fid.write('v {} {} {}\n'.format(*v))
+            for f in faces:
+                fid.write('f {} {} {}\n'.format(*(f + 1)))
```
larsoner, line 1057: @wmvanvliet we are allowed to use f-strings now because we require 3.6+

mne/tests/test_surface.py
```diff
@@ -244,4 +252,4 @@ def test_normal_orth():
         assert_allclose(ori[2], nn, atol=1e-12)
 
 
-run_tests_if_main()
+#run_tests_if_main()
```
larsoner, line 255: remove

larsoner: BTW now that we have `pytest` and it's (much) easier to use, I find these not so helpful. I'd be okay with getting rid of them in the codebase...

wmvanvliet: Alternatively, is there a good way to run a single test inside a Jupyter console? (I usually need the postmortem debugger)

larsoner:
> Alternatively, is there a good way to run a single test inside a Jupyter console? (I usually need the postmortem debugger)

Personally I just do:
```
pytest mne/tests/test_whatever.py -k test_something --tb=short --pdb
```
this drops to a `(pdb)` prompt where you can even do `interactive` and then have a standard python terminal (with matplotlib interactive, etc.). Not sure if there is a way to get this to be an ipython rather than python terminal, maybe with something like https://cprohm.de/article/ipytest-running-pytest-in-ipython-notebooks.html/ ?

wmvanvliet:
`Sphinx-gallery gallery_conf["plot_gallery"] was False, so no examples were executed.`

Does that mean that CircleCI is complaining that no examples needed to be build or something?

larsoner:
No, it's a warning about optipng (warnings are treated as errors) that you can safely ignore. It's because your branch is out of date and CircleCI doesn't / can't merge the `yaml` before executing

wmvanvliet:
I'm still working on the "fix my BEM with Blender" tutorial, which I feel should be part of this PR.

wmvanvliet:
Can anyone help me getting the static images (blender screenshots) properly embedded the the tutorial text?

larsoner:
I can push a commit to fix the warnings and add a thumbnail (hopefully)

larsoner:
(FYI testing locally with `PATTERN=in_blender make html_dev-pattern` allows for fairly quick iteration to fix bugs)

Grr... something is wrong with the SG static thumb embedding. I'll probably need to fix that at the SG end, but the syntax here is correct so shouldn't be a blocker for merge

larsoner:
... turns out I just can't spell

mne/surface.py
```diff
@@ -698,7 +708,39 @@ def read_surface(fname, read_metadata=False, return_dict=False, verbose=None):
...
+                split = line.split()
+                if split[0] == "v":  # vertex
+                    coords.append([float(item) for item in split[1:]])
+                elif split[0] == "f":  # face
+                    dat = [int(item.split("/")[0]) for item in split[1:]]
+                    if len(dat) != 3:
+                        raise RuntimeError('Only triangle faces allowed.')
+                    # In .obj files, indexing starts at 1
+                    faces.append([d - 1 for d in dat])
+        ret = (np.array(coords), np.array(faces))
```
agramfort, line 740: can you put this in a private function to have the same design as with _get_read_geometry?

wmvanvliet:
When CI comes back green, this is ready to merge from my end.

larsoner:
Green other than timeouts that are not your fault that we can ignore

larsoner:
The files are actually kind of big (500 / 700 kB), maybe some suitable JPG compression could be used?

doc/changes/latest.inc
```diff
@@ -224,3 +228,5 @@ API
 - The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
 
 - The default argument ``meg=True`` in :func:`mne.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
+
+- Add ``file_format`` parameter to :func:`mne.read_surface` and :func:`mne.write_surface` by `Marijn van Vliet`_
```
larsoner, line 232: This shouldn't actually go here as it's a new feature, not something that someone will need to change to keep their code working as it did before (unless they are using very strange filenames), so probably better just to combine this with your Wavefront entry above

wmvanvliet:
Image filesizes are around 100kb now

agramfort:
looks good https://20582-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_fix_bem_in_blender.html#sphx-glr-auto-tutorials-source-modeling-plot-fix-bem-in-blender-py

wmvanvliet:
CI's were timing out

ezemikulan:
Hi, sorry for the late reply, looks great to me!

wmvanvliet:
Hi @ezemikulan, thanks! If you can state you full name I can add you properly instead of just your username.

ezemikulan:
It is Ezequiel Mikulan

Thanks again

PR #7810, opened 2020-05-22
MRG, ENH: Validated new algorithm for ICA MEG reference channel method 

larsoner:
Supersedes #7212. Just adds one more plot

Closes #7212

PR #7825, opened 2020-05-25
MRG, ENH: Support n_col keyword in ica.plot_score

bloyl:
In playing with the recent ica ref channels artifact rejection (#7810), I'd sometimes have 4-5 score plots in one column, which I couldn't see.

This PR adds a parameter ('n_col') to 'ica_plot_scores' to use a grid with multiple columns.

There is a question about the following from the test.
```ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1], labels='ecg')```
it fails on this PR and is currently commented out.

On master it passed but did not make a plot, because:
```
if labels == 'ecg':
        labels = [label for label in ica.labels_ if label.startswith('ecg/')]
```
returned `labels = []` and nothing was plotted.

I don't know enough about the ica labeling so someone should advice what the correct behavior is and how to test it.

### DIFF
--- a/mne/viz/ica.py
+++ b/mne/viz/ica.py
@@ -638,10 +638,10 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
     if n_cols is None:
         # prefer more rows.
         n_rows = int(np.ceil(np.sqrt(n_scores)))
-        n_cols = int(np.ceil(n_scores / float(n_rows)))
+        n_cols = (n_scores - 1) // n_rows + 1
     else:
         n_cols = min(n_scores, n_cols)
-        n_rows = int(np.ceil(n_scores / float(n_cols)))
+        n_rows = (n_scores - 1) // n_cols + 1
 
     if figsize is None:
         figsize = (6.4 * n_cols, 2.7 * n_rows)
--- a/mne/viz/ica.py
+++ b/mne/viz/ica.py
@@ -585,7 +585,8 @@ def _plot_ica_sources_evoked(evoked, picks, exclude, title, show, ica,
 
 
 def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
-                    title='ICA component scores', figsize=None, show=True):
+                    title='ICA component scores', figsize=None,
+                    n_cols=None, show=True):
     """Plot scores related to detected components.
 
     Use this function to asses how well your score describes outlier
@@ -612,6 +613,11 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
         The figure title.
     figsize : tuple of int | None
         The figure size. If None it gets set automatically.
+    n_cols : int | None
+        Scores are plotted in a grid. This parameter controls how
+        many to plot side by side before starting a new row. By
+        default, a number will be chosen to make the grid as square as
+        possible.
     show : bool
         Show figure if True.
 
@@ -627,15 +633,26 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
     exclude = np.unique(exclude)
     if not isinstance(scores[0], (list, np.ndarray)):
         scores = [scores]
-    n_rows = len(scores)
+    n_scores = len(scores)
+
+    if n_cols is None:
+        n_cols = int(np.ceil(np.sqrt(n_scores)))
+    else:
+        n_cols = min(n_scores, n_cols)
+
+    n_rows = int(np.ceil(n_scores / float(n_cols)))
+
     if figsize is None:
-        figsize = (6.4, 2.7 * n_rows)
-    fig, axes = plt.subplots(n_rows, figsize=figsize, sharex=True, sharey=True)
+        figsize = (6.4 * n_cols, 2.7 * n_rows)
+    fig, axes = plt.subplots(n_rows, n_cols, figsize=figsize,
+                             sharex=True, sharey=True)
+
     if isinstance(axes, np.ndarray):
         axes = axes.flatten()
     else:
         axes = [axes]
-    axes[0].set_title(title)
+
+    fig.suptitle(title)
 
     if labels == 'ecg':
         labels = [label for label in ica.labels_ if label.startswith('ecg/')]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by bloyl** [comment]
@larsoner or @dengemann could either of you take a look at the commented out test? I'm not sure what it was trying to test.

File: mne/viz/tests/test_ica.py [hunk 1]
```diff
@@ -296,9 +296,23 @@ def test_plot_ica_scores():
     ica.labels_['eog'] = 0
     ica.labels_['ecg'] = 1
     ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1])
+    ica.plot_scores([[0.3, 0.2], [0.3, 0.2]], axhline=[0.1, -0.1])
     ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1], labels='foo')
     ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1], labels='eog')
-    ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1], labels='ecg')
+    # ica.plot_scores([0.3, 0.2], axhline=[0.1, -0.1], labels='ecg')
```
agramfort, line 302: cleanup

bloyl: That's an old test. I'm not sure what the point of it is nor what the desired/intended behavior is in that situation. see my comments on the PR.

I'm fine with deleting it just want to make sure I'm not changing some behavior that is used/usefull

larsoner: Can you change the ECG threshold so that it *does* find a component, and thus can plot? Alternatively, keep this line in but nested in a proper `pytest.raises`? Though I think the former would be preferable..

bloyl: I don't think there is any actual ECG thresholding going on.

The test manually sets the labels as : 
```
    ica.labels_ = dict()
    ica.labels_['eog/0/foo'] = 0
    ica.labels_['eog'] = 0
    ica.labels_['ecg'] = 1
```
but the plotter only looks for nested label names `label.startswith('ecg/')`

one solution would be to first look for nested labels ('ecg/') and if none are found look for non-nested ('ecg')

I don't like keeping the line and nesting in a `pytest.raises` that would convert a test that used to check if something didn't raise an error to a test that the same code now does raise and error.

larsoner: > one solution would be to first look for nested labels ('ecg/') and if none are found look for non-nested ('ecg')

I have no idea if the labeling is supposed to work this way, it would require digging into the code...

Maybe just add:
```
ica.labels_['ecg/0/bar'] = 1
```
? This should allow it to plot

bloyl:
I just pushed a change to the function that

1) Lets the test pass
2) Does what I think was expected before

I'm good to MRG on my end if everyone is happy.

mne/viz/ica.py
```diff
@@ -627,30 +633,48 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
...
-    n_rows = len(scores)
+    n_scores = len(scores)
+
+    if n_cols is None:
+        # prefer more rows.
+        n_rows = int(np.ceil(np.sqrt(n_scores)))
+        n_cols = int(np.ceil(n_scores / float(n_rows)))
+    else:
+        n_cols = min(n_scores, n_cols)
+        n_rows = int(np.ceil(n_scores / float(n_cols)))
```
larsoner, line 644: It's better to stick with int arithmetic where possible, so:
```
n_rows = (n_scores - 1) // n_cols + 1
```
same in the `n_cols` calculation above

mne/viz/ica.py
```diff
@@ -627,30 +633,48 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
     exclude = np.unique(exclude)
     if not isinstance(scores[0], (list, np.ndarray)):
         scores = [scores]
-    n_rows = len(scores)
+    n_scores = len(scores)
+
+    if n_cols is None:
+        # prefer more rows.
```
larsoner, line 639: I would have assumed the opposite given that screens are wider than they are tall. Bonus of computing `n_cols` first is that the the `n_rows` can be unified/deindented. WDYT?

mne/viz/ica.py
```diff
@@ -627,30 +633,48 @@ def plot_ica_scores(ica, scores, exclude=None, labels=None, axhline=None,
...
         axes = axes.flatten()
     else:
         axes = [axes]
-    axes[0].set_title(title)
+
+    fig.suptitle(title)
 
     if labels == 'ecg':
         labels = [label for label in ica.labels_ if label.startswith('ecg/')]
+        labels.sort(key=lambda l: l.split('/')[1])  # sort by index
```
larsoner, line 660: Alternatively (avoids the conditional):
```suggestion
        labels = [label for label in ica.labels_ if label.startswith('ecg')]
        # sort by index, allowing just `ecg`
        labels.sort(key=lambda l: l.split('/')[:2][-1])
```

mne/viz/ica.py
bloyl: That is how I originally wrote it as well.

I changed it because:
1. Most common use case is plotting 1 or 2 scores from ECG and EOG artifact rejection, backward compatibility would put these as 1 column.
2. The subplots are wider then they are tall. so it makes sense to add a new row before adding a new column

bloyl: I'm fine with either.

mne/viz/ica.py
bloyl: Assuming I need the same logic in both the 'ecg' and 'eog' cases.
This change fails in the case of : 
```
    ica.labels_ = dict()
    ica.labels_['ecg/0/foo'] = 0
    ica.labels_['ecg'] = 0
```
because it would return `['ecg', 'ecg/0/foo']` when we just want `['ecg']`.

The fact that the current code on master didn't work for the last several years and no complained suggests this isn't a heavily used case.

larsoner: Ahh okay, I assumed that we might actually want both `ecg` and `ecg/0/foo`, but sounds like that's not the case

mne/viz/ica.py
larsoner: Okay I can live with this then!

mne/viz/ica.py
larsoner: ... but it would be nice to move to int arithmetic here still

larsoner:
Thanks for working to appease codecov :)

agramfort:
Thx @bloyl

Issue #7671, opened 2020-04-24
ENH: Add classes for Spectra / PSD


thht:
#### Describe the problem
mne features dedicated classes for time-series data as well as time-frequency data. however, for a current project i need to work a lot with spectra. this would preferably include good exploratory plotting solutions like we already have for time-series data, including interactive features.

#### Describe your solution
as i am going to need these features anyhow, i am going to start working on such a class right now in my own collection.

i would like to know whether there is any motivation to have such a class in mne python.

and if yes, i would look forward to any suggestions / input from you guys.

larsoner:
@drammock is also fighting with this issue. One solution is to consider these TFRs with a single time point, and make our computation functions like compute_source_psd return them.

thht:
ok, but does this let me plot spectra and then mark a time-window which results in a topoplot etc?

larsoner:
Not sure, need to check the docs for those classes. But if not, these seem like core features that should be part of the class anyway so we should add them

thht:
ok... so i am going to start by creating a subclass of AverageTFR in my own repo, add the methods and as soon as it works i start an mne branch and refactor it into the original class... and then we can see how to make in mne-ready....

larsoner:
Sounds reasonable to me but let's wait to hear from at least @drammock and @agramfort that they agree before you put in too much effort

larsoner:
... actually I'm not sure why it would need to be a subclass rather than an instance of `AverageTFR` directly. If it's going to have to be a subclass we'd probably just want to make a new class altogether.

I know @drammock's use case involves storing multitaper estimates, so you still end up with 3D data of shape `(n_channels, n_freqs, n_tapers)`, so we could put this in `AverageTFR` easily, but it isn't really a *time* frequency representation, it is just a single frequency representation computed with multiple window functions.

thht:
sorry, this seems to be a misunderstanding.

in the final mne python version this would be all implemented in `AverageTFR`. but as i need this now and still want to use the stable mne version in my analysis project, i will create my custom subclass of `AverageTFR` just adding the features i need because i do not want to change the mne code in this particular scenario.

when i am ready to integrate it into mne, i will refactor the code so that my custom subclass is not needed anymore but instead all new logic/functionality would be implemented directly in `AverageTFR`.

i hope this clears up the confusion?

larsoner:
Indeed, makes sense

drammock:
[...]

PR #7828, opened 2020-05-26
WIP: Add CrossSpectralDensity.compute_coherence method

wmvanvliet:
Computing sensor level coherence is quite simple, and useful if you for
example want to compute coherence between each MEG sensor and an
external source.

Todo:

 - [x] Implement `csd.compute_coherence` method
 - [x] Refactor `csd.plot(mode='coh')` to make use of the new method
 - [x] Unit test
 - [ ] Add example for computing sensor-level coherence with external source
 - [ ] Add entry to what's new

### DIFF
--- a/mne/time_frequency/csd.py
+++ b/mne/time_frequency/csd.py
@@ -378,6 +378,37 @@ class CrossSpectralDensity(object):
 
         return _vector_to_sym_mat(self._data[:, index])
 
+    def compute_coherence(self, frequency=None, index=None):
+        """Compute a sensor-to-sensor coherence matrix.
+
+        If there is only one matrix defined in the CSD object, calling this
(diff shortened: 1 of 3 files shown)


larsoner:
Maybe a test that on some small set of `epochs` computing `spectral_connectivity` gives the same result as computing `CrossSpectralDensity` + `csd.compute_coherence`?

wmvanvliet:
I actually wonder if this is needed at all, given that `spectral_connectivity` already exists. Probably not.

wmvanvliet:
Still, we should probably have a tutorial for computing coherence with an external source.

larsoner:
Indeed, some SSVEP/ASSR dataset would be nice for this, and would also work well for sorting out #7671

PR #7831, opened 2020-05-26
MRG, ENH: Add axes to plot_evoked_white

larsoner:
1. Add charset spec to `Report` so that unicode characters actually render (on master they are messed up)
2. Add `axes` arg to `plot_evoked_white`
3. Modernize some `evoked.plot_white` tests
4. Actually test the multi-cov case of `plot_evoked_white` (previously only used in examples)


larsoner:
One more commit to push -- remove `90%` width reduction for thumbnails and increase slider from `col-md-6` size. Things like `evoked.plot_joint` did not fit with these constraints, so it makes sense to make them larger.

Issue #7774, opened 2020-05-12
DOC: add crossrefs in the info docstring


drammock:
We should add crossrefs in the info docstring to the various MNE functions/methods that can safely change info fields (e.g., `inst.rename_channels` etc)

cbrnr:
Is there a list of these functions somewhere? If not, can we start collecting them here?

larsoner:
No list AFAIK, I would just start a PR with a few of them and people can push commits with others when they think of / find them

drammock:
we don't typically assign issues, feel free to open a PR

PR #7832, opened 2020-05-26
Clearer info docstring

### LINKED ISSUE #7818: Use properties to set values in Info
As I've mentioned in #7816 I think it would be more intuitive if users could just set values in the `Info` dict using simple assignments, e.g. `info['meas_date'] = None`, instead of setter methods (e.g. `info.set_meas_date(None)`). This is consistent with values that we currently allow to be set directly, e.g. `info["bads"] = []`.

This would also allow us to make read-only properties for fields which users should not set manually.

cbrnr:
Fixes #7818. @hoechenberger @larsoner @agramfort OK with you?

### DIFF
--- a/mne/io/meas_info.py
+++ b/mne/io/meas_info.py
@@ -189,14 +189,14 @@ class Info(dict, MontageMixin):
     `FIF format specification <https://github.com/mne-tools/fiff-constants>`__,
     so new entries should not be manually added.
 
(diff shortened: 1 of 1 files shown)


drammock:
@cbrnr care to address #7774 at the same time?

larsoner:
This already helps , thanks @cbrnr ! Feel free to tackle the cross references next if you're up for it

PR #6522, opened 2019-07-03
DOC: add list of side-packages

drammock:
per discussion on MNE_Analysis email list, this adds a list of "side-packages" to the installation docs.

PR #7835, opened 2020-05-27
DOC: added reference to sesameeg package

sarasommariva:
Added reference to package [sesameeg](https://pybees.github.io/sesameeg/) into section [install](https://mne.tools/stable/install/pre_install.html) of the documentation.

### DIFF
--- a/doc/install/pre_install.rst
+++ b/doc/install/pre_install.rst
@@ -56,6 +56,8 @@ MNE-Python, including packages for:
 - post-hoc modification of linear models (`posthoc`_)
 - a python implementation of the Preprocessing Pipeline (PREP) for EEG data
   (`pyprep`_)
+- automatic multi-dipole localization and uncertainty quantification with
+  the Bayesian algorithm SESAME (`sesameeg`_)
 
 
 What should I install?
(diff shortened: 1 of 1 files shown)


larsoner:
I think we need a new page for this, see #5484. Would you be up for creating a new page? Then we can push commits with other packages and merge a first version of it, and others can add things later...

agramfort:
+1000

drammock:
I added that "related projects" section to the "pre-install" page in #6522. And @sarasommariva is adding sesameeg into that list, which is the correct place in my mind.  @larsoner @agramfort what is the reasoning behind giving it a dedicated page?  To me the pre-install page doesn't seem overly crowded or long...

agramfort:
+1 for MRG

we can deliberate later :)

>

larsoner:
Ahh, forgot about that. Agreed it's not too long and we can live with it there for now at least.

drammock:
thanks @sarasommariva

sarasommariva:
thanks to you all guys! I'll be happy to help with the plug-in page when/if needed.. for now I'm lazy enough to agree with @drammock :)

PR #7839, opened 2020-05-27
Add CZI to list of funders

drammock:


larsoner:
It looks less abbreviated compared to the other entries:

https://20314-1301584-gh.circle-artifacts.com/0/dev/index.html

But maybe this just indicates that we should abbreviate the others less at some point. +1 for merge from me

drammock:
> It looks less abbreviated compared to the other entries

fair point.  We could abbreviate inst. name to CZI, and/or use their initialism for this funidng call (EOSS).  Or, as you suggest, we could spell out "Department of Energy", "National Institutes of Health", etc.  WDYT?

larsoner:
Up to you -- I would lean toward abbreviating both since that's what we do on the others now. But also feel free to try expanding the others (maybe just the funders is the way to go?) to see if it looks good

larsoner:
[Longer one](https://20316-1301584-gh.circle-artifacts.com/0/dev/index.html) LGTM, +1 for merge

Issue #7842, opened 2020-05-28
Pyvista foreground bug


ualsbombe:
#### Describe the bug
〈Replace this text with a clear and concise description of the bug.〉

When changing the background to 'white' and foreground to 'black', text is not formatted accordingly (works as intended in mayavi) when using mne.viz.plot_source_estimates with pyvista

#### Steps to reproduce
```
import mne
from os.path import join

data_path = mne.datasets.sample.data_path()
subjects_dir = join(data_path, 'subjects')
filename = 'sample_audvis-meg'
stc = mne.read_source_estimate(join(data_path, 'MEG', 'sample', filename))

## works as intended

mne.viz.set_3d_backend('mayavi')
stc.plot('sample', subjects_dir=subjects_dir, background='white',
         foreground='black', figure=1)

## does not

mne.viz.set_3d_backend('pyvista')
stc.plot('sample', subjects_dir=subjects_dir, background='white',
         foreground='black', figure=2)
```

#### Expected results
That the numbers for the scale would be written in black, now they are (presumably) written in white

#### Actual results
Mayavi
![image](https://user-images.githubusercontent.com/17312936/83109776-45995e00-a0c2-11ea-9890-426b48e0a7a2.png)
Pyvista
![image](https://user-images.githubusercontent.com/17312936/83109802-4e8a2f80-a0c2-11ea-94e6-c9e5bebf2526.png)

#### Additional information
mne.sys_info()
Platform:      Linux-5.3.0-53-generic-x86_64-with-debian-buster-sid
Python:        3.7.6 (default, Jan  8 2020, 19:59:22)  [GCC 7.3.0]
Executable:    /home/lau/miniconda3/envs/mnedev/bin/python
CPU:           x86_64: 8 cores
Memory:        31.2 GB

mne:           0.21.dev0
numpy:         1.18.1 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.4.1
matplotlib:    3.1.3 {backend=Qt5Agg}

sklearn:       0.22.1
numba:         0.48.0
nibabel:       3.0.2
cupy:          Not found
pandas:        1.0.2
dipy:          1.1.1
mayavi:        4.7.1
pyvista:       0.24.0
vtk:           8.2.0
PyQt5:         5.9.2

PR #7843, opened 2020-05-28
MRG: Add support for foreground in _Brain

GuillaumeFavelier:
This PR adds support for the `foreground` parameter in `_Brain`. This parameter influences the color of the labels displayed (scalar bar label, time label and slider title and label).

Following the example given in https://github.com/mne-tools/mne-python/issues/7842#issue-626271633, this is what I obtain now:

time_viewer=False | time_viewer=True
---|---
image | image

Closes https://github.com/mne-tools/mne-python/issues/7842
It's an item of https://github.com/mne-tools/mne-python/issues/7162

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -177,10 +177,12 @@ class _Brain(object):
 
         if isinstance(background, str):
             background = colorConverter.to_rgb(background)
+        self._bg_color = background
         if foreground is None:
-            foreground = "white"
+            foreground = 'w' if sum(self._bg_color) < 2 else 'k'
         if isinstance(foreground, str):
             foreground = colorConverter.to_rgb(foreground)
+        self._fg_color = foreground
         if isinstance(views, str):
             views = [views]
         n_row = len(views)
@@ -193,7 +195,6 @@ class _Brain(object):
                              'sequence of ints.')
         fig_size = size if len(size) == 2 else size * 2  # 1-tuple to 2-tuple
 
-        self._foreground = foreground
         self._hemi = hemi
         self._units = units
         self._subject_id = subject_id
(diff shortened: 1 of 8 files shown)


mne/viz/backends/base_renderer.py
```diff
@@ -332,13 +332,16 @@ def text3d(self, x, y, z, text, width, color='white'):
...
 
         Parameters
         ----------
         source:
             The object of the scene used for the colormap.
+        color:
+            The color of the label text.
         title: str | None
             The title of the scalar bar.
         n_labels: int | None
```
agramfort, line 347: bgcolor not documented?

mne/viz/_brain/_brain.py
```diff
@@ -177,6 +177,8 @@ def __init__(self, subject_id, hemi, surf, title=None,
 
         if isinstance(background, str):
             background = colorConverter.to_rgb(background)
+        if foreground is None:
+            foreground = "white"
```
larsoner, line 181: We should set it to white or black automatically depending on the luminance of the `background`, see [how it's done in PySurfer](https://github.com/nipy/PySurfer/blob/master/surfer/viz.py#L563-L564)

larsoner:
Can you add to the TODO list to make the `matplotlib` figure use the same foreground/background colors as the plotter above it? It seems like they should match...

larsoner:
(not necessarily for this PR)

larsoner:
Pushed a commit to fix `pyvistaqt` and to fix our default to `foreground=None`

I did notice a bug with the VectorSourceEstimate plotting -- it looks like the colormap of the arrows are not interactively updated, see how the brain surface and arrow colormaps do not match here:

Screenshot from 2020-05-28 10-41-39

Sample code to reproduce, just click around the sliders and you should see the brain surface change color but the color/alpha of the arrows stays the same.

GuillaumeFavelier:
I updated #7162 with https://github.com/mne-tools/mne-python/pull/7843#issuecomment-635384571 and https://github.com/mne-tools/mne-python/pull/7843#issuecomment-635395395

agramfort:
merge when happy !

>

larsoner:
Thanks @GuillaumeFavelier

Issue #7720, opened 2020-05-04
Add viz.plot_annotations() & Annotations.plot()


hoechenberger:
#### Describe the problem
Currently there is no easy way to directly visualize `Annotations`. I always have to convert them to events first and then use `mne.viz.plot_events()`, which seems pretty awkward.

#### Describe your solution
Similarly to `mne.viz.plot_events()`, there should be `mne.viz.plot_annotations()` and `Annotations.plot()`. They would require the user to pass an `info` structure, and otherwise work similarly to `plot_events()`.

PR #7845, opened 2020-05-28
Allow find_bad_channels_maxwell() to return scores

hoechenberger:
Inspired by https://github.com/mne-tools/mne-study-template/issues/118#issuecomment-630847578, I have added a new parameter `return_scores` to `preprocessing.find_bad_channels_maxwell()`.

If True, we will now return a dictionary with alle the scores, limits, and corresponding channel names and types. 

<strike>
- the edges of the time windows analyzed
- the corresponding evaluation of a channel being flat in this window
- the corresponding evaluation of a channel being noisy in this window

I currently refer to the latter two values as "scores", but actually they are just boolean values. So the returned ndarrays `scores_flat` and `scores_noisy` only contain True and False for each channels in all evaluated time segments. Wonder if there's a better name we could use?
</strike>

#### Additional information
I first wanted to ask for feedback before starting to add tests.

### DIFF
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1888,7 +1888,7 @@ def find_bad_channels_maxwell(
 
         .. warning:: This feature is experimental and may change in a future
                      version of MNE-Python without prior notice. Please
-                     report any problems and improvement proposals to the
+                     report any problems and enhancement proposals to the
                      developers.
     %(maxwell_origin_int_ext_calibration_cross)s
     %(maxwell_coord)s
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -2108,6 +2108,7 @@ def find_bad_channels_maxwell(
     flat_chs = sorted((f for f, c in flat_chs.items() if c >= min_count),
                       key=lambda x: raw.ch_names.index(x))
 
+    # Only include MEG channels.
     scores_flat = scores_flat[params['meg_picks']]
     thresh_flat = thresh_flat[params['meg_picks']]
     scores_noisy = scores_noisy[params['meg_picks']]
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1885,6 +1885,11 @@ def find_bad_channels_maxwell(
     return_scores : bool
         If ``True``, return a dictionary with scoring information for each
         evaluated segment of the data. Default is ``False``.
+
+        .. warning:: This feature is experimental and may change in a future
+                     version of MNE-Python without prior notice. Please
+                     report any problems and improvement proposals to the
+                     developers.
     %(maxwell_origin_int_ext_calibration_cross)s
     %(maxwell_coord)s
     %(maxwell_reg_ref_cond_pos)s
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1911,8 +1911,9 @@ def find_bad_channels_maxwell(
         - ``ch_types`` : ndarray, shape (n_meg,)
             The types of the MEG channels in ``ch_names`` (``'mag'``,
             ``'grad'``).
-        - ``bin_edges`` : ndarray, shape (n_windows + 1,)
-            The window boundaries (in seconds) used to calculate the scores.
+        - ``bins`` : ndarray, shape (n_windows, 2)
+            The inclusive window boundaries (start and stop; in seconds) used
+            to calculate the scores.
         - ``scores_flat`` : ndarray, shape (n_meg, n_windows)
             The scores for testing whether MEG channels are flat.
         - ``limits_flat`` : ndarray, shape (n_meg, 1)
@@ -2007,7 +2008,7 @@ def find_bad_channels_maxwell(
     all_flats = set()
 
     # Prepare variables to return if `return_scores=True`.
-    bin_edges = np.r_[starts, stops[-1]]  # Ensure we include the endpoint!
+    bins = np.empty((len(starts), 2))  # To store start, stop of each segment
     # We create ndarrays with one row per channel, regardless of channel type
     # and whether the channel has been marked as "bad" in info or not. This
     # makes indexing in the loop easier. We only filter this down to the subset
@@ -2037,6 +2038,7 @@ def find_bad_channels_maxwell(
         delta = np.std(data, axis=-1).min(-1)  # min std across segments
 
         # We may want to return this later if `return_scores=True`.
+        bins[si, :] = chunk_raw.times[0], chunk_raw.times[-1]
         scores_flat[good_meg_picks, si] = delta
         thresh_flat[good_meg_picks] = these_limits.reshape(-1, 1)
 
@@ -2115,7 +2117,7 @@ def find_bad_channels_maxwell(
     if return_scores:
         scores = dict(ch_names=ch_names,
                       ch_types=ch_types,
-                      bin_edges=bin_edges,
+                      bins=bins,
                       scores_flat=scores_flat,
                       limits_flat=thresh_flat,
                       scores_noisy=scores_noisy,
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1858,7 +1858,7 @@ def _trans_sss_basis(exp, all_coils, trans=None, coil_scale=100.):
 # st_only
 @verbose
 def find_bad_channels_maxwell(
-        raw, limit=7., duration=5., min_count=5,
+        raw, limit=7., duration=5., min_count=5, return_scores=False,
         origin='auto', int_order=8, ext_order=3, calibration=None,
         cross_talk=None, coord_frame='head', regularize='in', ignore_ref=False,
         bad_condition='error', head_pos=None, mag_scale=100.,
@@ -1881,6 +1881,9 @@ def find_bad_channels_maxwell(
     min_count : int
         Minimum number of times a channel must show up as bad in a chunk.
         Default is 5.
+    return_scores : bool
+        If True (default False), return the scores and time of the window edges
+        used.
     %(maxwell_origin_int_ext_calibration_cross)s
     %(maxwell_coord)s
     %(maxwell_reg_ref_cond_pos)s
@@ -1896,6 +1899,15 @@ def find_bad_channels_maxwell(
     flat_chs : list
         List of MEG channels that were detected as being flat in at least
         ``min_count`` segments.
+    bin_edges : ndarray, shape (n_windows + 1,)
+        The window boundaries (in seconds) used to calculate the scores.
+        Only returned when ``return_scores`` is ``True``.
+    scores_flat : ndarray, shape (n_meg, n_windows)
+        The scores for testing whether MEG channels are flat. Only returned
+        when ``return_scores`` is ``True``.
+    scores_noisy : ndarray, shape (n_meg, n_windows)
+        The scores for testing whether MEG channels are noisy. Only returned
+        when ``return_scores`` is ``True``.
 
     See Also
     --------
@@ -1904,10 +1916,11 @@ def find_bad_channels_maxwell(
 
     Notes
     -----
-    All arguments after ``raw``, ``limit``, ``duration``, and ``min_count``
-    are the same as :func:`~maxwell_filter`, except that the following are
-    not allowed in this function because they are unused: ``st_duration``,
-    ``st_correlation``, ``destination``, ``st_fixed``, and ``st_only``.
+    All arguments after ``raw``, ``limit``, ``duration``, ``min_count``, and
+    ``return_scores`` are the same as :func:`~maxwell_filter`, except that the
+    following are not allowed in this function because they are unused:
+    ``st_duration``, ``st_correlation``, ``destination``, ``st_fixed``, and
+    ``st_only``.
 
     This algorithm, for a given chunk of data:
 
@@ -1973,8 +1986,14 @@ def find_bad_channels_maxwell(
         if pick in params['grad_picks'] else
         flat_limits['mag']
         for pick in good_meg_picks])
+
     flat_step = max(20, int(30 * raw.info['sfreq'] / 1000.))
     all_flats = set()
+    bin_edges = np.r_[starts, stops[-1]]  # Ensure we include the endpoint!
+    scores_flat = np.empty((len(good_meg_picks), len(starts)), dtype=bool)
+    scores_flat.fill(False)
+    scores_noisy = scores_flat.copy()
+
     for si, (start, stop) in enumerate(zip(starts, stops)):
         n_iter = 0
         orig_data = raw.get_data(None, start, stop, verbose=False)
@@ -1988,6 +2007,7 @@ def find_bad_channels_maxwell(
         data.shape = (data.shape[0], -1, flat_step)
         delta = np.std(data, axis=-1).min(-1)  # min std across segments
         chunk_flats = delta < these_limits
+        scores_flat[:, si] = chunk_flats  # We may want to return this later
         chunk_flats = np.where(chunk_flats)[0]
         chunk_flats = [raw.ch_names[good_meg_picks[chunk_flat]]
                        for chunk_flat in chunk_flats]
@@ -2035,6 +2055,9 @@ def find_bad_channels_maxwell(
                          % (name, max_))
             these_picks.pop(idx)
             chunk_noisy.append(name)
+            # We may want to return this later
+            ch_idx = raw.copy().pick_types(meg=True).ch_names.index(name)
+            scores_noisy[ch_idx, si] = True
         noisy_chs.update(chunk_noisy)
     noisy_chs = sorted((b for b, c in noisy_chs.items() if c >= min_count),
                        key=lambda x: raw.ch_names.index(x))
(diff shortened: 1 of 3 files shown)


agramfort:
what do you want to do with that? how do you plan visualise this?

hoechenberger:
> what do you want to do with that? how do you plan visualise this?

My idea was that we could plot it with time (or segment number) on the abscissa and channels on the ordinate. Bad segments for individual channels would be colored in based on the type of bad (flat, noisy, or manually set).

My idea is to have sth that looks a little like the top-right subplot here:

agramfort:
I think it's important to show what it can bring visually before updating
public API

hoechenberger:
@agramfort 
> I think it's important to show what it can bring visually before updating
> public API

Very first very drafty draft:
Screenshot 2020-05-28 at 20 05 20

I also want to add a topoplot with sensors shown in the respective colors.

drammock:
@hoechenberger it seems this kind of plot could be profitably used for #7720 too. Maybe we should implement that, and use it to plot *annotations* from `find_bads_maxwell`?

hoechenberger:
> @hoechenberger it seems this kind of plot could be profitably used for #7720 too. Maybe we should implement that, and use it to plot _annotations_ from `find_bads_maxwell`?

I had been thinking about this too. The issue with Annotations is that they extend spatially across all channels, while here we're only interested temporal extent of a signal across _individual_ channels… Any ideas to reconcile this dichotomy?

hoechenberger:
@drammock We can also talk off GH if you think it could make sense, just drop me a mail at richard.hoechenberger@gmail.com

hoechenberger:
> I currently refer to the latter two values as "scores", but actually they are just boolean values. So the returned ndarrays `scores_flat` and `scores_noisy` only contain True and False for each channels in all evaluated time segments. Wonder if there's a better name we could use?

Just to elaborate on this a little more. What I currently return are just the auto-detected bad segments of channels which the algorithm found to be bad based on these segments. I simply tag these segments as "bad because flat", "bad because noisy".

We could also return the _actual_ scores, which would make for a nice visualization: for each channel and time window, we would get a specific _number_ (or score) that shows how "bad" that segment was. The plot would be very colorful and not as sparse as the one I posted above.

**Now there would be a few issues with this approach:**

1. Thresholding of the scores – which score qualifies as "bad" – is done based on the present data. This means that absolute score values are meaningless and wouldn't help users to decide just _how_ bad (or good!) certain segments of their data are.

2. We would have to create two subplots to represent the "flat" and "noisy" scores.

3. Also I believe that currently, the "noisy" channel detection skips all channels that had previously been identified as being "flat", so we wouldn't get any "noisy" scores for those.

4. In a graphical representation of all scores, it would be more difficult to find a way to make those channels that were found to be bad clearly stick out. This is the beauty of the sparse approach I presented in the figure above.

**Potential solutions**

1. Instead of absolute scores, we return the ration `score / threshold`; so values `<=1` would mean "does not qualify as bad", and values `>1` would mean "bad segment".

2. Today's displays are wide, so… maybe it's not a problem to place those two plots on a horizontal grid.

3. Should be easy to fix.

4. Keep this sparse "binary" figure in addition to the score figure.

… but that's also a lot of figures :)

Thoughts?

agramfort:
thoughts: play with it on real data see what is the most informative and ask
feedback from users like @SophieHerbst

>

drammock:
> The issue with Annotations is that they extend across _all_ channels, while here we're only interested in the temporal extent of a signal across _individual_ channels… Any ideas how to reconcile this dichotomy?

fair point; either you would need one `Annotations` object per channel detected as bad, and a plotting function that took in multiple annotations objects (seems unweildy), or a single `Annotations` object that labels the annotations something like `BAD_MAXWELL_MEG_1033` (or similar) and a plotting function that plots each annotation type on a different row / y-value of your figure.  That latter option seems maybe workable for this case, and maybe even sensible behavior for what you would want for any old annotation file (i.e., `BAD_EOG` on one line, `BAD_MOVEMENT` on a different line, etc).  WDYT?

agramfort:
adding support for subsets of channels in annotations was requested in the
past by some EEG users I know.
I said for now YAGNI but we may reconsider

>

hoechenberger:
@drammock
> or a single `Annotations` object that labels the annotations something like `BAD_MAXWELL_MEG_1033` (or similar) and a plotting function that plots each annotation type on a different row / y-value of your figure.

This could be a workaround, yes. But feels very hacky…

> That latter option seems maybe workable for this case, and maybe even sensible behavior for what you would want for any old annotation file (i.e., `BAD_EOG` on one line, `BAD_MOVEMENT` on a different line, etc). WDYT?

I think I do understand the general concept you're proposing, but what do you mean by "old annotation file"?

@agramfort 
> adding support for subsets of channels in annotations was requested in the past by some EEG users I know. I said for now YAGNI but we may reconsider

I have to say that this obviously sounds like the cleanest solution to me. I'm mostly worried about how to expose this in the Raw browser plot, i.e. how to make it so that you can select a time range & add an Annotation to only a subset of channels.

agramfort:
yes supporting this in UI will be tricky. I propose to sleep over it a bit
more

>

drammock:
> what do you mean by "old annotation file"?

sorry, "any old annotation file" meant "any annotation file not created by `find_bad_channels_maxwell` (i.e., one whose annotations were not channel-specific)

larsoner:
Combining this with #1869 could potentially be quite useful

hoechenberger:
Great idea, @larsoner!

I will pick up working again on this PR shortly.

hoechenberger:
@drammock After discussing with @agramfort and critically reconsidering my specific use case, I have decided that for now I will just work on the implementation of "return more elaborate scoring results" within the scope of this PR. I believe your suggestion does make a lot of sense, but I would address this in a followup PR.

So what I've done now is I've changed the return value(s): When `return_scores=True`, we return a single dictionary `scores`, which currently has the keys `bin_edges`, `scores_flat`, `scores_noisy`, `limits_flat`, `limits_noisy`.

With very little code, I can now produce helpful visualizations (I know, the colormap is not great):

```python
    auto_noisy_chs, auto_flat_chs, scores = find_bad_channels_maxwell(
        raw=raw_lp_filtered_for_maxwell,
        calibration=config.mf_cal_fname,
        cross_talk=config.mf_ctc_fname,
        return_scores=True)

    bad_mask = np.ones_like(scores['scores_noisy'])
    bad_mask[scores['scores_noisy'] > scores['limits_noisy']] = 0

    fig, ax = plt.subplots(1, 2, figsize=(12, 5))
    sns.heatmap(scores['scores_noisy'], cmap='bwr', ax=ax[0])
    sns.heatmap(scores['scores_noisy'], mask=bad_mask, cmap='bwr', ax=ax[1])

    ax[0].set_xlabel('Data Segment', fontweight='bold')
    ax[0].set_ylabel('Channel Index', fontweight='bold')
    fig.suptitle('Automated Noisy Channel Detection', fontweight='bold')
    plt.show()
```
Figure_1

(absolute scores on the left, thresholded on the right)

WDYT?

drammock:
> for now I will just work on the implementation of "return more elaborate scoring results"

seems reasonable.

mne/preprocessing/maxwell.py
```diff
@@ -1896,6 +1900,29 @@ def find_bad_channels_maxwell(
...
+
+        - ``scores_flat`` : ndarray, shape (n_meg, n_windows)
+
+          The scores for testing whether MEG channels are flat.
+
+        - ``limits_flat`` : ndarray, shape (n_meg, n_windows)
+
+          The thresholds above which a score classified a segment as "flat".
+
+        - ``scores_noisy`` : ndarray, shape (n_meg, n_windows)
```
agramfort, line 1920: insert empty line to be consistent. Or remove it everywhere to avoid a too long docstring

agramfort:
you will need to add tests and expose this in an example as an illlustration.

mne/preprocessing/maxwell.py
hoechenberger: I've added an empty line for now.

I'm not sure what's the best approach, also considering that I want a nice formatting in the end :) @drammock, do you have any suggestions here?

agramfort:
look at rendered doc on circleci artifact to decide

>

mne/preprocessing/tests/test_maxwell.py
```diff
@@ -1118,5 +1137,34 @@ def test_find_bad_channels_maxwell(fname, bads, annot, add_ch, ignore_ref,
...
+                got_scores['limits_noisy'].shape ==
+                (len(meg_chs), len(got_scores['bin_edges']) - 1))
+
+        for want_bad in want_bads:
+            ch_idx = list(got_scores['ch_names']).index(want_bad)
+            scores = got_scores['scores_noisy'][ch_idx]
+            limits = got_scores['limits_noisy'][ch_idx]
+            n_segments_exceed_limit = (scores > limits).sum()
+            assert (n_segments_exceed_limit >=
+                    min(min_count, len(got_scores['bin_edges']) - 1))
```
larsoner, line 1159: Rather than iterate over the ones it should find, why not just redo the computation with something like:
```
n_segments_exceed_limit = (got_scores['scores_noisy'] >= got_scores['limits_noisy']).sum(-1)
new_bad = set(got_scores['ch_names'][idx] for idx in np.where(n_segments_exceed_limit >= min(...))
assert new_bad == set(want_bads)
```
It checks both hits and correct rejections rather than just hits.

hoechenberger: I have now implemented your suggestion in 694a8b2

hoechenberger:
Ok this is good for a review!

hoechenberger:
Tutorial:
https://20498-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_60_maxwell_filtering_sss.html#using-sss-and-maxwell-filtering-in-mne-python

API doc:
https://20498-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.find_bad_channels_maxwell.html?highlight=find_bad_channels_maxwell#mne.preprocessing.find_bad_channels_maxwell

mne/preprocessing/maxwell.py
```diff
@@ -1973,8 +2001,23 @@ def find_bad_channels_maxwell(
         if pick in params['grad_picks'] else
         flat_limits['mag']
         for pick in good_meg_picks])
+
     flat_step = max(20, int(30 * raw.info['sfreq'] / 1000.))
     all_flats = set()
+
+    # Prepare variables to return if `return_scores=True`.
+    bin_edges = np.r_[starts, stops[-1]]  # Ensure we include the endpoint!
```
larsoner, line 2009: This is only correct if the data has no BAD annotations in it. So it would be better to keep the `starts` and `stops` rather than go with `bin_edges`.

mne/preprocessing/maxwell.py
```diff
@@ -1973,8 +2001,23 @@ def find_bad_channels_maxwell(
...
+    # We create ndarrays with one row per channel, regardless of channel type
+    # and whether the channel has been marked as "bad" in info or not. This
+    # makes indexing in the loop easier. We only filter this down to the subset
+    # of MEG channels after all processing is done.
+    ch_names = np.array(raw.ch_names)
+    ch_types = np.array(raw.get_channel_types())
+    scores_flat = np.empty((len(ch_names), len(starts)))
+    thresh_flat = scores_flat.copy()
+    scores_noisy = scores_flat.copy()
```
larsoner, line 2018: np.empty_like(scores_flat) is better and clearer here

tutorials/preprocessing/plot_60_maxwell_filtering_sss.py
```diff
@@ -106,18 +110,96 @@
...
+# the automated detection of *noisy* channels.
+
+for ch_type in ('mag', 'grad'):
+    # Only select the data for mag or grad channels.
+    ch_subset = auto_scores['ch_types'] == ch_type
+    ch_names = auto_scores['ch_names'][ch_subset]
+    scores = auto_scores['scores_noisy'][ch_subset]
+    limits = auto_scores['limits_noisy'][ch_subset]
+    # Number of time windows
+    segments = range(1, len(auto_scores['bin_edges']))
```
larsoner, line 144: Can you instead use something like (after refactoring according to the above):
```
t_centers = (start + stop) / 2.
```
Then the plot would go according to time rather than segments? Not sure if this would actually work, but if there are gaps in the data (e.g., if there are BAD annotations) it will actually match the time course better.

Personally I would probably work through how to use `plt.pcolormesh` with bin edges to represent the chunks correctly, but if there are gaps in the data will require filling in gaps with `np.nan` or so, so might not be worth the trouble.

hoechenberger:
> That pattern of vertically similar colors appears to be due to the nature of that test data…

Scratch that, there must still be a bug somewhere. I get a similar pattern for noise detection with actual experimental data. Ugh…

hoechenberger:
> > That pattern of vertically similar colors appears to be due to the nature of that test data…
> 
> Scratch that, there must still be a bug somewhere. I get a similar pattern for noise detection with actual experimental data. Ugh…

Ok found it. Will be fixed shortly.

mne/preprocessing/maxwell.py
```diff
@@ -1896,6 +1900,36 @@ def find_bad_channels_maxwell(
...
+            The inclusive window boundaries (start and stop; in seconds) used
+            to calculate the scores.
+        - ``scores_flat`` : ndarray, shape (n_meg, n_windows)
+            The scores for testing whether MEG channels are flat.
+        - ``limits_flat`` : ndarray, shape (n_meg, 1)
+            The score thresholds above which a segment was claffified as
+            "flat".
+        - ``scores_noisy`` : ndarray, shape (n_meg, n_windows)
+            The scores for testing whether MEG channels are noisy.
+        - ``limits_noisy`` : ndarray, shape (n_meg, 1)
```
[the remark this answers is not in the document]
hoechenberger: > why `ndarray, shape (n_meg, 1)`? It should be `float`, no? I see below you only set it for the good channels, but personally I would just leave it as a float (it's okay to set it for all channels because the `bad` ones get values of `-np.inf` so comparisons pass).

Also while this would work for the noisy channels, it wouldn't be applicable to the flat detection, because there we use different thresholds depending on channel type (mag, grad)

larsoner:
> I expect this to make it easier to create an interactive visualization based on the arrays returned here. I'm thinking about something like the viz I put into the tutorial, but also displaying channels from info['bads'], and where clicking on a tile of the heatmap would take you straight to the corresponding raw segment. 

If the algorithm doesn't operate on / modify / analyze some channels, it doesn't seem like they should show up in the output at all. For example if you pass `ignore_ref` to `maxwell_filter`, those channels shouldn't show up even though they are MEG channels. Downstream viz should be built to take into account the channels that were actually processed, regardless of actual data ordering, channel selection, etc., so I don't see simplifying the viz step as sufficient motivation for including `info['bads']` MEG channels in the output.

hoechenberger:
Thanks for the input,  @larsoner! I will discuss this with @SophieHerbst to get a better idea of what users might need and expect. 

--

hoechenberger:
@larsoner 
After thinking and also discussing this with @agramfort and @SophieHerbst, I have decided to:

- return a score array for `n_meg` channels, not just for `n_meg_good` channels, because it makes downstream handling easier. For those channels that were marked as `bad` before running `find_bad_channels_maxwell()`, `np.nan` scores and limits will be returned. It's just as simple to omit those NaNs e.g. while plotting (in fact, most plotting functions will ignore them by default), and it seems more correct to have these  value be `np.nan` than `+ / - np.inf`. So if you have no strong objections to this – this is how I'd ike to do it.

I've also added a warning to the `return_scores` param, stating this feature is experimental and may break without deprecation cycle. This should give is the flexibility to adjust the feature when needed – in fact, we're still busy exploring real-world use cases.

Let me know what you think!

Otherwise, this is all good to go from my side.

larsoner:
I can live with it, thanks @hoechenberger !

mne/preprocessing/maxwell.py
```diff
@@ -1896,6 +1905,33 @@ def find_bad_channels_maxwell(
...
+            The score thresholds above which a segment was claffified as
+            "noisy".
+
+        .. note:: The scores and limits for channels marked as ``bad`` in the
+                  input data will will be set to ``np.nan``.
```
agramfort, line 1934: will will -> will

mne/preprocessing/maxwell.py
```diff
@@ -1877,10 +1877,19 @@ def find_bad_channels_maxwell(
...
         Minimum number of times a channel must show up as bad in a chunk.
         Default is 5.
+    return_scores : bool
+        If ``True``, return a dictionary with scoring information for each
+        evaluated segment of the data. Default is ``False``.
+
+        .. warning:: This feature is experimental and may change in a future
+                     version of MNE-Python without prior notice. Please
+                     report any problems and enhancement proposals to the
+                     developers.
```
agramfort, line 1892: we need ..versionadded

tutorials/preprocessing/plot_60_maxwell_filtering_sss.py
```diff
@@ -106,18 +111,100 @@
...
-raw.info['bads'].extend(auto_noisy_chs + auto_flat_chs)
 
 ###############################################################################
-# But this algorithm is not perfect. For example, it misses ``MEG 2313``,
-# which has some flux jumps, because there are not enough flux jumps in the
-# recording. So it can still be useful to manually inspect and mark bad
-# channels:
+# Now we can update the list of bad channels in the dataset.
+
+bads = [*raw.info['bads'], *auto_noisy_chs, *auto_flat_chs]
```
agramfort, line 123: bads = raw.info['bads'] + auto_noisy_chs + auto_flat_chs

seems simpler especially for non python gurus

PR #7847, opened 2020-05-28
Fix eeg rereferencing error when ch_type is defined

dtxe:
When calling raw.set_eeg_reference with a pre-defined channel type, the error "local variable 'type_' referenced before assignment" occurs.
eg. `raw.set_eeg_reference(ref_channels='average', projection=False, ch_type='seeg')`

type_ is set while detecting the channel type that should be referenced. This commit uses the ch_type variable instead of type_, which is guaranteed to be set.


mne/io/reference.py
```diff
@@ -392,7 +392,7 @@ def set_eeg_reference(inst, ref_channels='average', copy=True,
         logger.info('EEG data marked as already having the desired reference.')
     else:
         logger.info('Applying a custom %s '
-                    'reference.' % DEFAULTS['titles'][type_])
+                    'reference.' % DEFAULTS['titles'][ch_type])
```
larsoner, line 395: This suggests we have some branches not covered properly. Can you add a loop [here](https://github.com/mne-tools/mne-python/blob/002a5cfc9baecdd0ecdaff1f38555b44a141de35/mne/io/tests/test_reference.py#L236-L238) to do something like this:
```
    for ch_type in ('auto', 'ecog'):
        raw = RawArray(data, create_info(3, 1000., ['ecog'] * 2 + ['misc']))
        reref, ref_data = set_eeg_reference(raw.copy(), ch_type=ch_type)
        _test_reference(raw, reref, ref_data, ['0', '1'])
```
This would fail on `master` but pass on this PR, right?

larsoner:
@dtxe any chance to add some variant of the test suggested above? If not I can try pushing some commits to your pull request. We will release 0.21 in a couple of weeks and it would be good to have this fix

larsoner:
Thanks @dtxe !

PR #7852, opened 2020-05-29
MRG, DOC: Add evoked movecomp to example

larsoner:
I realized that we didn't have an example of evoked movement compensation in our docs. This adds one to the end of the simulation. It also fixes a tiny bug with the weight calculation. In testing it doesn't actually affect the values, but worth doing anyway.


agramfort:
thx @larsoner !

Issue #7862, opened 2020-06-03
mne.preprocessing_current_source_density gives wrong CSD estimates


mservantufc:
#### Describe the bug
Attempt to compute CSD on evoked object called "monopolar_av_ref.fif" (see attached). The evoked object contains grand-averaged EEG waveforms (64 EEG channels (averaged reference), 'biosemi64' montage + 1 EOG channel).  

#### Steps to reproduce
GA_comp_corr_stim = mne.read_evokeds('monopolar_av_ref.fif')
GA_comp_corr_stim_CSD = mne.preprocessing.compute_current_source_density(GA_comp_corr_stim)#should be similar to the attached CSD.fif file

plots obtained:
times = np.arange(.3, 1.,.05)
fig1 = GA_comp_corr_stim.plot_topomap(times = times ,ch_type = 'eeg', cmap = 'viridis', nrows = 2, average = .03, time_unit='s')
fig3 = GA_comp_corr_stim_CSD.plot_topomap(times = times ,ch_type = 'csd', cmap = 'viridis', nrows = 2, average = .03, time_unit='s')

#### Actual results
plots "topo_CSD.png" and "topo_monopolar.png" in the attached. It is clear that CSD estimates are wrong, since the spatial selectivity appears lower than the spatial selectivity of the monopolar data, and differ from those obtained with a python implemenbtation of Mike Cohen's matlab code ('topo_CSD_Cohen.png')
Maybe it's a problem with the sphere= auto argument that returns:
Fitted sphere radius:         95.0 mm
Origin head coordinates:      0.0 -0.0 40.1 mm
Origin device coordinates:    0.0 -0.0 40.1 mm

#### Additional information
Platform:      Windows-10-10.0.18362-SP0
Python:        3.7.6 (default, Jan  8 2020, 20:23:39) [MSC v.1916 64 bit (AMD64)]
Executable:    C:\Users\mservant\Anaconda3\pythonw.exe
CPU:           Intel64 Family 6 Model 158 Stepping 10, GenuineIntel: 12 cores
Memory:        15.8 GB

mne:           0.20.5
numpy:         1.18.1 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.4.1
matplotlib:    3.1.3 {backend=Qt5Agg}

sklearn:       0.22.1
numba:         0.48.0
nibabel:       3.0.1

cupy:          Not found
pandas:        1.0.1
dipy:          1.1.1
[...]

PR #7863, opened 2020-06-04
MRG, BUG: Updated Current Source Density

alexrockhill:
Addresses https://github.com/mne-tools/mne-python/issues/7862.

Passes all tests and matches example from https://github.com/alberto-ara/Surface-Laplacian but testing data needs to be added to mne-testing-data and some code review would be appreciated. I think the for loops could or mostly all be changed to matrix multiplication or called through numpy/scipy but I'll need to look into that. Thanks @mservantufc for bringing this to attention. It appears as if the previous version was not working.

Maybe @larsoner and/or @jasmainak might have a minute to review since they were nice enough to help me with the original PR :) ?

- [x] Figure out units (don't necessarily use unit sphere)
- [x] Figure out cosine similarity versus 1 minus half the squared Euclidean distance


agramfort:
@alexrockhill can you touch the example file to we see how it looks now on the circleci artifact?

mne/preprocessing/tests/data/test_eeg_csd.csv
```diff
@@ -0,0 +1,64 @@
...32323229267e-07,5.390363636363642629e-08,-5.524502020201985514e-08,-3.029441848484847305e-07,-2.607717171717170900e-07,-4.825616161616158453e-08,-2.538674747474747907e-07,4.101132323232323078e-07,2.658551616161619900e-07,8.892952020202020153e-07,-2.229958585858585517e-07,3.693985858585853795e-07,2.057048090909090783e-07,6.391091616161618726e-07,4.053931515151513576e-07,-1.009909393939393781e-06,-4.271014141414098877e-08,4.522194444444443320e-07,1.714669696969694697e-07,-2.702303131313129525e-07,-6.859908686868686411e-07,-3.714364242424240003e-07,1.414630888888890034e-07,-8.757466060606058561e-07,-8.691700000000002053e-07,-9.448299999999993886e-07,-6.156907777777778286e-07,-7.489006161616161631e-07,-4.194075050505051570e-07,-1.208135555555555185e-06,-1.170046677777777669e-06,-1.325560434343433855e-06,-1.229559535353535155e-06,-1.506960131313131181e-06,-1.228605131313130989e-06,-2.758265959595958823e-07,-7.359676868686865129e-07,-7.129717171717171820e-07,-5.529046464646465404e-07,-1.052502424242424204e-06,-1.641118696969696343e-06,-1.781973242424242774e-06,-1.361436161616160979e-06,-2.032656161616161136e-06,-1.681479292929292518e-06,-1.770453131313130879e-06,-1.318636767676767499e-06
```
larsoner, line 1: Since this PR is going to require adding data to `mne-testing-data` anyway, let's not add these files here but in `mne-testing-data` instead

larsoner:
Happy to help make it more efficient, etc. but we need to get the test files over to `mne-testing-data` first. If you want, you can just email them to me and I can do it

larsoner:
Oh wait, I see ... actually these are small enough that they can probably just stay here. I'll fix the code

agramfort:
can you compress the files a bit? maybe ship them as npz?

larsoner:
> can you compress the files a bit? maybe ship them as npz?

Oof actually I didn't realize that `test_eeg.csv` is 30 MB. This should indeed be moved to `mne-testing-data`. I can do this quickly

larsoner:
(I will probably also just save it as `.mat` for ease of cross-platform loading)

agramfort:
+1

>

mne/preprocessing/_csd.py
```diff
@@ -23,7 +23,59 @@
...
+            g = 0
+            h = 0
+            for ni in range(n_legendre_terms):
+                g = g + (two_n1[ni] * leg_poly[ni, i, j]) / g_denom[ni]
+                h = h - (two_n1[ni] * leg_poly[ni, i, j]) / h_denom[ni]
+            G[i, j] = g / (4 * np.pi)
+            H[i, j] = -h / (4 * np.pi)
+    # add transposes
+    G += G.T - np.identity(n_chs) * G[1, 1] / 2
+    H += H.T - np.identity(n_chs) * H[1, 1] / 2
```
larsoner, line 77: This step is weird, I think it's incorrect. Shouldn't it be:
```
    G += G.T - np.diag(np.diag(G))
    H += H.T - np.diag(np.diag(H))
```
i.e., you add it to its transpose, then remove the duplicate summation along the diagonal?

larsoner: (it's also not great that tests pass regardless of whether or not I make this change :( )

alexrockhill: Hmmm agreed

larsoner:
@alexrockhill I pushed a commit to:

1. Vectorize everything
2. Move test files to `.mat` in `mne-testing-data`
3. Touch the CSD example

The CSD example looks bad now. Also, I noticed that the norm of the matrix `G` is about six orders of magnitude different between the EEG-file test and the FIF-file test, which suggests that there is some scale factor problem going on (maybe a units problem, everything `radius`/`pos` should be done in meters, maybe one at some point is in mm?

On `master` the norm of `G` appears to be around `0.001`, whereas on this PR it's much larger. This effectively changes the usefulness of `lambda2` quite a bit. We're going to need to fix this -- perhaps we should always use `lambda2 * linalg.norm(G)` or so, which will make it more invariant to these sorts of scale changes. We'll also need to adjust the default from 1e-5 to something much higher (0.05? maybe set by eye) to match the new norm levels.

alexrockhill:
> @alexrockhill I pushed a commit to:
> 
> 1. Vectorize everything
> 2. Move test files to `.mat` in `mne-testing-data`
> 3. Touch the CSD example
> 
> The CSD example looks bad now. Also, I noticed that the norm of the matrix `G` is about six orders of magnitude different between the EEG-file test and the FIF-file test, which suggests that there is some scale factor problem going on (maybe a units problem, everything `radius`/`pos` should be done in meters, maybe one at some point is in mm?
> 
> On `master` the norm of `G` appears to be around `0.001`, whereas on this PR it's much larger. This effectively changes the usefulness of `lambda2` quite a bit. We're going to need to fix this -- perhaps we should always use `lambda2 * linalg.norm(G)` or so, which will make it more invariant to these sorts of scale changes. We'll also need to adjust the default from 1e-5 to something much higher (0.05? maybe set by eye) to match the new norm levels.

Thanks so much, I am definitely not as skilled at vectorizing.

I think maybe `pos * 1e-3` can be added with `sphere` updated accordingly and it should still pass the tests. I can check.

larsoner:
Pushed one more efficiency commit, no more redundant `lpn` calculations, or unnecessary diagonal ones.

This might make it more difficult to debug, though, so feel free to go back to your old code for `_calc_G_H` for testing and iteration if you need to. I can easily restore the more efficient version later (or feel free to copy-paste it back in once you have everything working).

larsoner:
But `radius` also shows up in a later calculation. I guess this means that it's not scale invariant...?

larsoner:
BTW you can see in the output of this example that something is broken (and that `lambda2` does not currently matter):

https://20396-1301584-gh.circle-artifacts.com/0/dev/auto_examples/preprocessing/plot_eeg_csd.html#sphx-glr-auto-examples-preprocessing-plot-eeg-csd-py

alexrockhill:
Hmmm yeah changing to mm and then changing 85 radius to 0.085 didn't work, I agree on it not being scale invariant. Not sure what's going on here though.

alexrockhill:
Although the tests are passing and the evoked data is the same, when I went back and plotted this example https://github.com/alberto-ara/Surface-Laplacian/blob/master/surface%20laplacian%20example.ipynb the topomap it was uniform... I think the tests aren't covering the issue for some reason
csd_example

larsoner:
Maybe the `atol` is wrong/meaningless. Does the topomap look okay with your code?

Also, is it possible that the electrode locations are supposed to be projected onto a sphere? They currently are not, unless by happenstance you are using a spherical montage with zero origin.

larsoner:
... based on code comments for the old `_calc_g`, indeed it looks like `cos_dist` should be between 0 and 1 because it's meant to be "cosine of angles between pairs of points on a spherical surface". So really it shouldn't be calculated using the squared euclidean distance, but rather directly via the dot product of unit vectors. I'll see if it fixes the FIF data...

larsoner:
Okay this might have fixed things, let me push...

alexrockhill:
Ahh I think you're right the the error tolerance is meaningless. The scale of the example data is 1e-6 to 1e-7 so etol would have to be much lower.

larsoner:
If I go back to c2586be (when you opened the PR) and fix the path problems and use a sensible `atol`, you can see things are already broken there:
```
>       assert_allclose(epochs_csd.average().data, csd, atol=1e-7)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-07
E       
E       Mismatched elements: 38831 / 40960 (94.8%)
E       Max absolute difference: 1.23793879e-05
E       Max relative difference: 9.19420201
E        x: array([[-3.616917e-10, -7.107747e-10, -6.617863e-10, ..., -7.082218e-10,
E               -6.186906e-10, -7.052339e-10],
E              [-3.351042e-10, -7.403796e-10, -7.674658e-10, ...,  1.577782e-10,...
E        y: array([[-1.023747e-06, -5.399796e-07, -1.163600e-06, ..., -1.681479e-06,
E               -1.770453e-06, -1.318637e-06],
E              [-8.542492e-07, -5.915384e-07, -1.396437e-06, ...,  9.873015e-07,...
```
So maybe go back to that commit and fix things first? Then you can force-push and I can cherry-pick my commits on top...

alexrockhill:
Totally my fault I just realized the test data is likely the before data and not the after laplacian data because of a very stupid switching return error, my bad!!!

larsoner:
We can update the testing data to fix it, if you want to work from c2586be and push-force over my changes I can fix the testing data. But we should wait until we have something that works with a reasonable `atol` first

larsoner:
If the CSV you pushed is the one to compare against and it passes on c2586be, I can take care of the rest of the annoying steps.

larsoner:
... with that file on top of c2586be I get:
```
mne/preprocessing/tests/test_csd.py:84: in test_csd_matlab
    assert_allclose(epochs_csd.average().data, csd, atol=1e-10)
E   AssertionError: 
E   Not equal to tolerance rtol=1e-07, atol=1e-10
E   
E   Mismatched elements: 40960 / 40960 (100%)
E   Max absolute difference: 4.65739658e-05
E   Max relative difference: 4.92637861
E    x: array([[-3.616917e-10, -7.107747e-10, -6.617863e-10, ..., -7.082218e-10,
E           -6.186906e-10, -7.052339e-10],
E          [-3.351042e-10, -7.403796e-10, -7.674658e-10, ...,  1.577782e-10,...
E    y: array([[ 3.122692e-06,  6.683897e-06,  9.506682e-06, ..., -2.452742e-05,
E           -2.426614e-05, -2.801882e-05],
E          [ 7.973345e-06,  9.962245e-06,  1.211932e-05, ..., -9.860589e-06,...
```
So hopefully there is more to come and I just should have been more patient :)

alexrockhill:
I can go back and fix tests, sorry I may need a bit of time to clean it all up from where it worked

larsoner:
Sounds good, as I said feel free to force push over my changes, it's possible that my calculations are not equivalent since I was assuming the tests would tell me if something broke. Once you have something that matches, knowing the efficiency bits and the project-to-sphere-for-cosine-distance things I think I can reincorporate my stuff fairly easily!

larsoner:
@alexrockhill nice commit message :) just let me know when you want me to take a look

alexrockhill:
Sorry give me a bit more time, I'll send you a message, just writing more commits so it's easier to go back than before

alexrockhill:
Ok @larsoner, if you have time to optimize at all, that would be fantastic. I have to leave this for now and pick it up tonight but it has a working test and I triple checked that it replicates the example from https://github.com/alberto-ara/Surface-Laplacian/blob/master/surface%20laplacian%20example.ipynb. There is still a lot of optimizing to do and the data needs to be updated in mne-testing-data. I can get to those later and send another message then but if you did have some time to give it a pass over that might save me optimizing it and then realizing I don't know how `np.clip` etc. works and that that would work much better. Either way, I'll work on it more tonight. Again, really sorry for the faulty test script!!

larsoner:
> Again, really sorry for the faulty test script!!

Don't sweat it! Figuring out there was a problem cost less time than actually doing the optimizations, and now that I've done them they should be easy to put back. I'll see if I can fix things without too much effort, and hand it back to you when I get stuck :)

larsoner:
(Plus the bonus of social coding is that it's not just *your* fault that a test you coded did not work, but also *mine* as a reviewer because I should have looked more carefully / caught it and told you to fix it, or fixed it myself, before even going down the optimization route... yay shared responsibility!)

mne/preprocessing/_csd.py
```diff
@@ -152,14 +189,17 @@ def compute_current_source_density(inst, sphere='auto', lambda2=1e-5,
...
     pos -= (x, y, z)
-
-    G = _calc_g(np.dot(pos, pos.T), stiffness=stiffness,
-                n_legendre_terms=n_legendre_terms)
-    H = _calc_h(np.dot(pos, pos.T), stiffness=stiffness,
-                n_legendre_terms=n_legendre_terms)
+    # Original code does not project onto the sphere, probably not good:
+    # pos /= radius
+    # from scipy.spatial.distance import squareform, pdist
+    # cos_dist = 1 - squareform(pdist(pos, 'sqeuclidean')) / 2.
```
larsoner, line 195: @alexrockhill I think this formulation might be incorrect (ignoring for now whether you need `pos /= radius` or `pos /= np.linalg.norm(pos, axis=1)`, let's assume you have `radius` right and it's actually spherical so they are the same). The [cosine similarity](https://en.wikipedia.org/wiki/Cosine_similarity) is given by the dot product divided by the magnidtudes (which are both one here). I haven't read the actual papers (probably should), but assuming that it's supposed to measure the cosine distance, using some variant of the squared Euclidean distance will not be correct.

mne/preprocessing/tests/test_csd.py
```diff
@@ -3,168 +3,142 @@
...
-
-    # test raw
-    csd_raw = compute_current_source_density(raw, sphere=sphere)
+    evoked, csd, sphere = evoked_csd_sphere
+    evoked_csd = compute_current_source_density(evoked, sphere=sphere)
+    assert_allclose(linalg.norm(csd), 0.00177, atol=1e-5)
+    # If we don't project onto the sphere, we get 1e-12 accuracy here,
+    # but it's a bad assumption for real data!
+    # Also, we divide by (radius ** 2) to get to units of V/m², unclear
+    # why this isn't done in the upstream implementation
```
larsoner, line 73: This is the other thing I changed @alexrockhill -- the `_compute_csd` function has a `/ radius ** 2` to get the units to be a source density. It's unclear why the upstream implementation does not have this, as I think it will end up with units of `V` instead of `V/m²`. The latter seems like the right units, but again maybe I need to read the papers...

Otherwise, what you had and what `master` did were nearly equivalent.

mne/preprocessing/_csd.py
```diff
@@ -152,14 +189,17 @@ def compute_current_source_density(inst, sphere='auto', lambda2=1e-5,
...
+    # Project onto the sphere and do the distance (effectively)
+    pos /= np.linalg.norm(pos, axis=1, keepdims=True)
+    cos_dist = np.clip(np.dot(pos, pos.T), -1, 1)
+    del pos
+    G, H = _calc_G_H(n_legendre_terms, stiffness, cos_dist)
```
larsoner, line 200: The one thing I need to check is why `_calc_G_H` does not give the same result as `_calc_g` and `_calc_h`. That also seems to differ between this PR and `master`, but I haven't investigated exactly why.

examples/preprocessing/plot_eeg_csd.py
```diff
@@ -3,7 +3,8 @@
 Transform EEG data using current source density (CSD)
 =====================================================
 
-This script shows an example of how to use CSD [1]_ [2]_ [3]_.
+This script shows an example of how to use CSD :footcite`PerrinEtAl1987`
+:footcite:`PerrinEtAl1989` :footcite:`KayserTenke2015`.
```
drammock, line 7:
```suggestion
This script shows an example of how to use CSD
:footcite:`PerrinEtAl1987,PerrinEtAl1989,KayserTenke2015`.
```

mne/preprocessing/_csd.py
larsoner: ... oh wait, they actually match. I'll remove them in favor of `_calc_g` and `_calc_h`, since they're (much) simpler

mne/preprocessing/_csd.py
```diff
@@ -91,12 +92,7 @@ def compute_current_source_density(inst, sphere='auto', lambda2=1e-5,
...
-    .. [1] Perrin F, Bertrand O, Pernier J. "Scalp current density mapping:
-           Value and estimation from potential data." IEEE Trans Biomed Eng.
-           1987;34(4):283–288.
-    .. [2] Perrin F, Pernier J, Bertrand O, Echallier JF. "Spherical splines
-           for scalp potential and current density mapping."
-           Electroenceph Clin Neurophysiol. 1989;72(2):184–187.
+    .. footbibliography::
     .. [3] Kayser J, Tenke CE. "On the benefits of using surface Laplacian
            (Current Source Density) methodology in electrophysiology.
            Int J Psychophysiol. 2015 Sep; 97(3): 171–173.
```
drammock, line 98: cruft

mne/defaults.py
```diff
@@ -15,19 +15,19 @@
     units=dict(mag='fT', grad='fT/cm', eeg='µV', eog='µV', ecg='µV', emg='µV',
                misc='AU', seeg='mV', dipole='nAm', gof='GOF', bio='µV',
                ecog='µV', hbo='µM', hbr='µM', ref_meg='fT', fnirs_raw='V',
-               fnirs_od='V', csd='V/m²'),
+               fnirs_od='V', csd='mV/m²'),
```
larsoner, line 18: @alexrockhill I think these should be correct now. Natively they are `V/m²`, but if we display them that way then `scalings['csd']` must be `1.`. Instead, if what we show in `evoked.plot` is `mV/m²`, then we need `scalings['csd'] = 1.e3`. So at least these are consistent now. They need to be related to one another. We should probably add a test for this, but it involves checking units and converting and would be a pain...

mne/defaults.py
```diff
@@ -15,19 +15,19 @@
...
-                  csd=1e5),
+                  csd=1e3),
     # rough guess for a good plot
     scalings_plot_raw=dict(mag=1e-12, grad=4e-11, eeg=20e-6, eog=150e-6,
                            ecg=5e-4, emg=1e-3, ref_meg=1e-12, misc='auto',
                            stim=1, resp=1, chpi=1e-4, exci=1, ias=1, syst=1,
                            seeg=1e-4, bio=1e-6, ecog=1e-4, hbo=10e-6,
                            hbr=10e-6, whitened=10., fnirs_raw=2e-2,
-                           fnirs_od=2e-2, csd=20e-4),
+                           fnirs_od=2e-2, csd=200e-4),
```
larsoner, line 30: This scale factor, on the other hand, just says how to scale the SI-unit data to make it generally look nice in `raw.plot`, so it does not need to maintain the scale-factor-to-SI relationship.

mne/preprocessing/_csd.py
```diff
@@ -153,9 +151,18 @@ def compute_current_source_density(inst, sphere='auto', lambda2=1e-5,
...
+    # cos_dist = 1 - squareform(pdist(pos, 'sqeuclidean')) / 2.
+    # Project onto the sphere and do the distance (effectively)
+    pos /= np.linalg.norm(pos, axis=1, keepdims=True)
+    cos_dist = np.clip(np.dot(pos, pos.T), -1, 1)
+    del pos
+
+    G = _calc_g(cos_dist, stiffness=stiffness,
                 n_legendre_terms=n_legendre_terms)
-    H = _calc_h(np.dot(pos, pos.T), stiffness=stiffness,
+    H = _calc_h(cos_dist, stiffness=stiffness,
```
larsoner, line 165: ... *and* @alexrockhill our old code did use the actual cosine distance, so that's both the name and someone else's implementation that make it seem like it should be the cosine similarity, not one minus the squared euclidean distance divided by two. But this is just anecdotal, someone might need to go to the papers to check.

larsoner:
[PR](https://20415-1301584-gh.circle-artifacts.com/0/dev/auto_examples/preprocessing/plot_eeg_csd.html#sphx-glr-auto-examples-preprocessing-plot-eeg-csd-py) vs [master](https://mne.tools/dev/auto_examples/preprocessing/plot_eeg_csd.html#sphx-glr-auto-examples-preprocessing-plot-eeg-csd-py)

alexrockhill:
Here is the relevant code from the original MATLAB source as far as I can tell

```
% CSD - Current Source Density (CSD) transformation based on spherical spline
%       surface Laplacian as suggested by Perrin et al. (1989, 1990)
%
% (published in appendix of Kayser J, Tenke CE, Clin Neurophysiol 2006;117(2):348-368)
%
% Usage: [X, Y] = CSD(Data, G, H, lambda, head);
%
% Implementation of algorithms described by Perrin, Pernier, Bertrand, and
% Echallier in Electroenceph Clin Neurophysiol 1989;72(2):184-187, and 
% Corrigenda EEG 02274 in Electroenceph Clin Neurophysiol 1990;76:565.
% 
% Input parameters:
%   Data = surface potential electrodes-by-samples matrix
%      G = g-function electrodes-by-electrodes matrix
%      H = h-function electrodes-by-electrodes matrix
% lambda = smoothing constant lambda (default = 1.0e-5)
%   head = head radius (default = no value for unit sphere [µV/m²])
%          specify a value [cm] to rescale CSD data to smaller units [µV/cm²]
%          (e.g., use 10.0 to scale to more realistic head size)
%
% Output parameter:
%      X = current source density (CSD) transform electrodes-by-samples matrix
%      Y = spherical spline surface potential (SP) interpolation electrodes-
%          by-samples matrix (only if requested)
%
% Copyright (C) 2003 by Jürgen Kayser (Email: kayserj@pi.cpmc.columbia.edu)
% GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
% Updated: $Date: 2005/02/11 14:00:00 $ $Author: jk $
%        - code compression and comments 
% Updated: $Date: 2007/02/07 11:30:00 $ $Author: jk $
%        - recommented rescaling (unit sphere [µV/m²] to realistic head size [µV/cm²])
%   Fixed: $Date: 2009/05/16 11:55:00 $ $Author: jk $
%        - memory claim for output matrices used inappropriate G and H dimensions
%   Added: $Date: 2009/05/21 10:52:00 $ $Author: jk $
%        - error checking of input matrix dimensions
%
function [X, Y] = CSD(Data, G, H, lambda, head)
[nElec,nPnts] = size(Data);            % get data matrix dimensions
if ~(size(G,1) == size(G,2)) | ...     % matrix dimension error checking: 
   ~(size(H,1) == size(H,2)) | ...     % G and H matrix must be nElec-by-nElec
   ~(size(G,1) == nElec) | ...         
   ~(size(H,1) == nElec)
   X = NaN; Y = NaN;                   % default to invalid output 
   help CSD
   disp(sprintf(strcat('*** Error: G (%d-by-%d) and H (%d-by-%d) matrix', ...
                       ' dimensions must match rows (%d) of data matrix'), ...
                       size(G),size(H),nElec));
   return 
end   
mu = mean(Data);                       % get grand mean
Z = (Data - repmat(mu,nElec,1));       % compute average reference
Y = Data;  X = Y;                      % claim memory for output matrices
if nargin < 5; head = 1.0; end;        % initialize scaling variable [µV/m²]
head = head * head;                    % or rescale data to head sphere [µV/cm²]
if nargin < 4; lambda = 1.0e-5; end;   % initialize smoothing constant
for e = 1:size(G,1);                   % add smoothing constant to diagonale
  G(e,e) = G(e,e) + lambda; 
end; 
Gi = inv(G);                           % compute G inverse
for i = 1:size(Gi,1);                  % compute sums for each row
  TC(i) = sum(Gi(i,:));
end;
sgi = sum(TC);                         % compute sum total
for p = 1:nPnts
  Cp = Gi * Z(:,p);                    % compute preliminary C vector
  c0 = sum(Cp) / sgi;                  % common constant across electrodes
  C = Cp - (c0 * TC');                 % compute final C vector
  for e = 1:nElec;                     % compute all CSDs ...
    X(e,p) = sum(C .* H(e,:)') / head; % ... and scale to head size
  end;
  if nargout > 1; for e = 1:nElec;     % if requested ...
    Y(e,p) = c0 + sum(C .* G(e,:)');   % ... compute all SPs
  end; end;
end;
```

```
% function [G,H] = GetGH (M, m)
% 
% Modification of GetCSD.m to obtain the G- and H-matrices, which are 
% invariant for a given EEG montage, which are then used as input with
% the EEG/ERP data to the CSD.m routine
%
% Usage: [G,H] = GetGH (M, m)
% 
% Input parameter:
%         M = a MatLab structure for an EEG montage consisting of <n> sites
%             as returned by the function ExtractEEGMontage.m;
%             M consists of:  <n> cell string array 'lab'
%                             <n> double 'theta'
%                             <n> double 'phi'
%                             <n * 2> double matrix 'xy'
%                             ('xy' values are not used in this function)
%         m = positive integer constant that affects spherical spline flexibility
%             values of m can be:    2 - most flexible interpolation
%                                    3 - medium flexibility
%                                    4 - more rigid splines (default)
%                                5..10 - increasingly rigid splines  
%
% Output parameters:
%         G = G electrodes-by-electrodes matrix (SP spline)
%         H = H electrodes-by-electrodes matrix (CSD spline)
%
% Notes: 1) GetGH.m differs from GetCSD.m by removing the first three lines
%           in the GetCSD.m code because spherical angles are defined through
%           <M>, and by removing the last line handling the EEG/ERP data
%        2) In contrast to GetCSD.m, the <m> constant may be optionally
%           specified as an input parameter (default: m = 4)
%        3) Using the G and H matrices with data not conforming to the original
%           EEG montage will not produce valid CSDs
%        4) The orientation of Theta and Phi is defined in the appendix of
%           Kayser J, Tenke CE, Clin Neurophysiol 2006;117(2):348-368), and
%           is not identical to other spherical notifications using the
%           same spherical angle names (e.g., as in EEGlab)
%
% -----------------------------------------------------------------------
% GetCSD - Compute Current Source Density (CSD) estimates using the spherical spline
%          surface Laplacian algorithm suggested by Perrin et al. (1989, 1990)
%
% (published in appendix of Kayser J, Tenke CE, Clin Neurophysiol 2006;117(2):348-368)
%
% Usage: [CE] = GetCSD(EEGmont, ERPdata);
%
% Implementation of algorithms described by Perrin, Pernier, Bertrand, and
% Echallier in Electroenceph Clin Neurophysiol 1989;72(2):184-187, and 
% Corrigenda EEG 02274 in Electroenceph Clin Neurophysiol 1990;76:565.
% 
% Input parameters:
%   EEGmont = name of EEG montage spherical coordinates file in ASCII format,
%             with rows consisting of six columns: electrode label, two 
%             spherical angles (theta, phi), and Cartesian coordinates x,y,z
%   ERPdata = name of ERP data file in ASCII format stored as 
%             electrodes-by-samples (rows-by-columns) matrix
%
% Output parameter:
%        CE = CSD estimates as electrodes-by-samples matrix
%
% Copyright (C) 2003 by Jürgen Kayser (Email: kayserj@pi.cpmc.columbia.edu)
% GNU General Public License (http://www.gnu.org/licenses/gpl.txt)
% Updated: $Date: 2005/02/11 14:00:00 $ $Author: jk $
% -----------------------------------------------------------------------
% Modified: $Date: 2009/05/12 19:12:00 $ $Author: jk $
%        - eliminate <ERPdata> input argument and direct use of CSD.m
%        - return G and H transformation matrices
%    Added: $Date: 2009/05/14 17:20:00 $ $Author: jk $
%        - input argument <m> to specify constant for spline flexibility
%    Added: $Date: 2009/05/22 11:28:00 $ $Author: jk $
%        - progress bar to indicate status of iteration 
%  Changed: $Date: 2010/04/11 19:46:00 $ $Author: jk $
%        - allow more rigid values for spline constant m [2..10] 
%
function [G,H] = GetGH (M, m)
ThetaRad = (2 * pi * M.theta) / 360;     % convert Theta and Phi to radians ...
PhiRad = (2 * pi * M.phi) / 360;         % ... and Cartesian coordinates ...
[X,Y,Z] = sph2cart(ThetaRad,PhiRad,1.0); % ... for optimal resolution
nElec = length(M.lab);                   % determine size of EEG montage
EF(nElec,nElec) = 0;                     % initialize interelectrode matrix ...
for i = 1:nElec; for j = 1:nElec;        % ... and compute all cosine distances
  EF(i,j) = 1 - ( ( (X(i) - X(j))^2 + ...
    (Y(i) - Y(j))^2 + (Z(i) - Z(j))^2 ) / 2 );
end; end;
if nargin < 2
   m = 4;                                % set m constant default
end
if ~ismember(m,[2:10])                    % verify m constant
%   disp(sprintf('Error: Invalid m = %d [use 2, 3, or 4]',[m]));
   disp(sprintf('Error: Invalid m = %d [use an integer between 2 and 10]',[m]));
   G = NaN;
   H = NaN;
   return
end
disp(sprintf('Spline flexibility:  m = %d',[m])); 
N = 50;                                  % set N iterations
G(nElec,nElec) = 0; H(nElec,nElec) = 0;  % claim memory for G- and H-matrices
fprintf('%d iterations for %d sites [',N,nElec); % intialize progress bar
for i = 1:nElec; for j = 1:nElec;
  P = zeros(N);                          % compute Legendre polynomial
  for n = 1:N;
    p = legendre(n,EF(i,j));
    P(n) = p(1);
  end;
  g = 0.0; h = 0.0;                      % compute h- and g-functions
  if j == 1; fprintf('*'); end;          % show progress
  for n = 1:N;
    g = g + ( (( 2.0*n+1.0) * P(n)) / ((n*n+n)^m    ) );
    h = h + ( ((-2.0*n-1.0) * P(n)) / ((n*n+n)^(m-1)) );
  end;
  G(i,j) =  g / 4.0 / pi;                % finalize cell of G-matrix
  H(i,j) = -h / 4.0 / pi;                % finalize cell of H-matrix
end; end;
disp(']');                               % finalize progress bar
```

Just so that this is here as reference.

larsoner:
(FYI, fortunately we got permission from Jurgen a couple of years ago to relicense his implementation under BSD, otherwise we could not use / look at this snippet because of the GPL license)

mne/preprocessing/tests/test_csd.py
larsoner: Based on the code snippets you pasted, it looks like perhaps you only factor in `radius` here (`head` in his code) if you want to scale away from the default SI units (except his is in uV, but I assume the input data are assumed to be in uV already). So indeed perhaps we should get rid of the `/ radius ** 2` in our code

alexrockhill:
> (FYI, fortunately we got permission from Jurgen a couple of years ago to relicense his implementation under BSD, otherwise we could not use / look at this snippet because of the GPL license)

Great I think Dennis had mentioned that earlier. Thanks so much for finishing everything @larsoner! 

It might be worth also citing:

- Cohen, M.X. (2014). Surface Laplacian In Analyzing neural time series data: theory and practice 
   (pp. 275-290). London, England: The MIT Press.

The formatting for that looks complex, I'll have to look into the documentation on how citations are done now.

larsoner:
Just add entries like I did already in this PR. It's standard bibtex entries, should be easier rather than more difficult

mne/preprocessing/_csd.py
larsoner: > [X,Y,Z] = sph2cart(ThetaRad,PhiRad,1.0)

So this takes the locations (wherever they were) and implicitly projects them onto the unit sphere. So we shouldn't do `pos /= radius` but rather `pos /= np.linalg.norm(pos, axis=1, keepdims=True)`. The remaining question in my mind is whether the 1 minus half the squared Euclidean distance is the correct thing to use or the cosine similarity between the positions. I'll read the paper to figure it out, but to me (given the spherical projection and variable naming) I'm guessing it will be the cosine similarity.

alexrockhill:
Ok doing now, where are you on projecting onto the sphere or not? Is it just preliminary commenting it out?

mne/preprocessing/_csd.py
```diff
@@ -153,9 +151,18 @@ def compute_current_source_density(inst, sphere='auto', lambda2=1e-5,
         raise ValueError('Zero or infinite position found in chs')
     pos -= (x, y, z)
 
-    G = _calc_g(np.dot(pos, pos.T), stiffness=stiffness,
+    # Original code does not project onto the sphere, probably not good:
+    # pos /= radius
+    # from scipy.spatial.distance import squareform, pdist
+    # cos_dist = 1 - squareform(pdist(pos, 'sqeuclidean')) / 2.
+    # Project onto the sphere and do the distance (effectively)
+    pos /= np.linalg.norm(pos, axis=1, keepdims=True)
```
larsoner, line 159: @alexrockhill this implicitly projects to a unit sphere. It preserves the phi/theta but sets r=1, which is the same thing as doing cart2sph, set r=1, then sph2cart...

alexrockhill: Ok so ideally the sphere would have the radius of the head? Is that what's done in the original?

mne/preprocessing/_csd.py
larsoner: one minus half the squared euclidean appears to be an equivalent way of computing the cosine similarity, so this is solved

mne/preprocessing/tests/test_csd.py
larsoner: Leaving `radius ** 2` in the code makes sense conceptually because it's a current source density -- once you do all your work on the unit sphere, dividing by the actual radius gets you to the correct units (I think). Intuitively it makes sense because the smaller the head, the higher the current *source density* needs to be to achieve the same observed EEG voltages. This is also consistent with `master` and Jurgen's code, so I think we should keep it.

mne/preprocessing/_csd.py
larsoner: In Mike and Jurgen's code they both effectively do this (at least for a spherical montage it's equivalent), and based on computing a cosine similarity I think it's the right thing to do.

Mike never uses an actual radius, but Jurgen does (divides by `head * head` effectively), which conceptually makes sense to me.

larsoner:
Okay @alexrockhill I think I'm happy with our code now, having read through Mike's and Jurgen's code.

mne/defaults.py
larsoner: SI unit test has been added that would have caught the unit problem in master

larsoner:
Sorry @alexrockhill I had to rebase, so while I was in there I squashed our commits, hope that's okay. This one can be rebase+merged to preserve history if whoever hits the green button can remember to do so...

alexrockhill:
> Sorry @alexrockhill I had to rebase, so while I was in there I squashed our commits, hope that's okay. This one can be rebase+merged to preserve history if whoever hits the green button can remember to do so...

Of course, that's great. Thanks for doing the hard parts :)

agramfort:
you mean the fun part for @larsoner :)

>

larsoner:
Thanks @alexrockhill !

agramfort:
@larsoner you merged this without squashing so we now have in the history the big CSV files that were pushed by @alexrockhill in the first commits.... grrrrr

I saw this as it required 15s for me to fetch master :(

rewriting history to remove these files may put a mess on the current PRs...

@larsoner wdyt?

agramfort:
https://github.com/mne-tools/mne-python/commit/0f8788ffbe6ceff6fb64ddfab27a8d7f1d841478 is the commit now in master

larsoner:
Argh sorry forgot about those. The rebase and merge was intentional to preserve code history/blame, but I forgot about the file commits. Thanks for push forcing to fix it

agramfort:
@cbrnr @larsoner @drammock @alexrockhill @hoechenberger @GuillaumeFavelier FYI I force pushed to master to remove the 60MB of csv files that were commited by mistake from this PR. I hope it does not create too much of a mess on your PRs....

PR #7875, opened 2020-06-07
FIX : EDF+ Annotation Timestamps missing sub-second accuracy

### LINKED ISSUE #7872: EDF+ Annotation Timestamps missing submillisecond accuracy
It was noticed that the onset time for annotations in an EDF+ file might be missing the submillisecond accuracy accociated with the start time of the recording. This results in annotation timestamps being delayed from the true event time. 

#### Expected results

The EDF+ file used here contains patient data and cannot be shared. However, if this bug is confirmed all EDF+ files will be affected. 

If the EDF+ file is brought into [EDFbrowser](https://www.teuniz.net/edfbrowser/) it will be noticed that the millisecond aspect of the start time is displayed in the `Start` field:

thumbnail_Outlook-ibjnd242

For instance, this results in the annotation timestamps being:

thumbnail_Outlook-of2qriv1

#### Actual results

When loading the data into MNE the annotation timestamps return:

thumbnail_Outlook-dnk5ugfi

If you take the difference between the MNE timestamp and the EDFbrowser timestamp it is equal to the submillisecond portion of the recording start time.

#### Additional information

I reached out to Teunis van Beelen, the developer of [EDFbrowser](https://www.teuniz.net/edfbrowser/), to get an explanation. The explanation provided was as follows:

> EDF does not have support for subsecond precision of the starttime but EDF+ does. Because EDF+ needed to be backwards compatible with EDF and EDF doesn't have the possibility to write subsecond precision for the starttime into the header, the subsecond part of the starttime is written outside of the EDF header. The way the subsecond part of the starttime is stored in EDF+ is described here: https://www.edfplus.info/specs/edfplus.html#timekeeping

Teunis van Beelen provided a quote from the EDF specs:

> "The startdate/time of a file is specified in the EDF+ header fields 'startdate of recording' and 'starttime of recording'. These fields must indicate the absolute second in which the start of the first data record falls. So, the first TAL in the first data record always starts with +0.X2020, indicating that the first data record starts a fraction, X, of a second after the startdate/time that is specified in the EDF+ header. If X=0, then the .X may be omitted."

In other words, the subsecond part of the starttime is written in the EDF Annotations "signal".

agramfort:
closes #7872

### DIFF
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -1383,14 +1383,14 @@ def _read_annotations_edf(annotations):
             if description:
                 events.append([onset, duration, description])
             elif k == 0:
-                # "The startdate/time of a file is specified in the EDF+ header
+                # The startdate/time of a file is specified in the EDF+ header
                 # fields 'startdate of recording' and 'starttime of recording'.
                 # These fields must indicate the absolute second in which the
                 # start of the first data record falls. So, the first TAL in
-                # the first data record always starts with +0.X2020, indicating
+                # the first data record always starts with +0.X, indicating
                 # that the first data record starts a fraction, X, of a second
                 # after the startdate/time that is specified in the EDF+
-                # header. If X=0, then the .X may be omitted."
+                # header. If X=0, then the .X may be omitted.
                 offset = -onset
 
     return zip(*events) if events else (list(), list(), list())
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -1375,12 +1375,23 @@ def _read_annotations_edf(annotations):
         triggers = re.findall(pat, tals.decode('latin-1'))
 
     events = []
-    for ev in triggers:
-        onset = float(ev[0])
+    offset = 0.
+    for k, ev in enumerate(triggers):
+        onset = float(ev[0]) + offset
         duration = float(ev[2]) if ev[2] else 0
         for description in ev[3].split('\x14')[1:]:
             if description:
                 events.append([onset, duration, description])
+            elif k == 0:
+                # "The startdate/time of a file is specified in the EDF+ header
+                # fields 'startdate of recording' and 'starttime of recording'.
+                # These fields must indicate the absolute second in which the
+                # start of the first data record falls. So, the first TAL in
+                # the first data record always starts with +0.X2020, indicating
+                # that the first data record starts a fraction, X, of a second
+                # after the startdate/time that is specified in the EDF+
+                # header. If X=0, then the .X may be omitted."
+                offset = -onset
 
     return zip(*events) if events else (list(), list(), list())
 
--- a/mne/io/edf/tests/test_edf.py
+++ b/mne/io/edf/tests/test_edf.py
@@ -56,7 +56,7 @@ bdf_multiple_annotations_path = op.join(data_path, 'BDF',
                                         'multiple_annotation_chans.bdf')
 test_generator_bdf = op.join(data_path, 'BDF', 'test_generator_2.bdf')
 test_generator_edf = op.join(data_path, 'EDF', 'test_generator_2.edf')
-edf_annot_sub_ms_path = op.join(data_path, 'EDF', 'subsecond_starttime.edf')
+edf_annot_sub_s_path = op.join(data_path, 'EDF', 'subsecond_starttime.edf')
 
 eog = ['REOG', 'LEOG', 'IEOG']
 misc = ['EXG1', 'EXG5', 'EXG8', 'M1', 'M2']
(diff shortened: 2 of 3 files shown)


agramfort:
@greydongilmore the onsets are now as you expect them but I am wondering if I should not also update the info['meas_date'] as it can contain miliseconds

wdyt?

greydongilmore:
This information would be useful to have access to upon heading the EDF+ file header. However, the header block in the EDF+ file will not contain the millisecond value information, and adding it to the `info` dict would require moving outside the header block to decode the TAL channel. Would this cause an issue for any other EDF methods?

If implemented, a conditional statement could be added based on the EDF type (whether EDF or EDF+). A zero millisecond value could be added to the `info['meas_date']` if EDF file is detected, otherwise read the first block of the TAL channel to get the value.

From the [EDF spec documentation](https://www.edfplus.info/specs/edfplus.html#tal):

> Onset must start with a '+' or a '-' character and specifies the amount of seconds by which the onset of the annotated event follows ('+') or precedes ('-') the startdate/time of the file, that is specified in the header.

So the subsecond part of the starttime must be added to the onset time of the events
at the moment of writing them into the file. The subsecond part of the starttime must be subtracted from the onset time of the events at the moment of reading them from the file.

agramfort:
ok then I guess what I did does what you expect.

with my branch my onsets are now:

[  1.951172   3.492188 290.501953 583.572266]

>

greydongilmore:
@agramfort I have tested on a few other files locally and it performs as expected now.

mne/io/edf/edf.py
```diff
@@ -1375,12 +1375,23 @@ def _read_annotations_edf(annotations):
...
         duration = float(ev[2]) if ev[2] else 0
         for description in ev[3].split('\x14')[1:]:
             if description:
                 events.append([onset, duration, description])
+            elif k == 0:
+                # "The startdate/time of a file is specified in the EDF+ header
+                # fields 'startdate of recording' and 'starttime of recording'.
+                # These fields must indicate the absolute second in which the
+                # start of the first data record falls. So, the first TAL in
+                # the first data record always starts with +0.X2020, indicating
```
cbrnr, line 1390: `+0.X2020` should be `+0.X` (or what does `2020` mean?)

mne/io/edf/edf.py
```diff
@@ -1375,12 +1375,23 @@ def _read_annotations_edf(annotations):
...
-        onset = float(ev[0])
+    offset = 0.
+    for k, ev in enumerate(triggers):
+        onset = float(ev[0]) + offset
         duration = float(ev[2]) if ev[2] else 0
         for description in ev[3].split('\x14')[1:]:
             if description:
                 events.append([onset, duration, description])
+            elif k == 0:
+                # "The startdate/time of a file is specified in the EDF+ header
```
cbrnr, line 1386: You can remove the initial double quote (and the closing one at the end).

cbrnr:
A test would be nice. I've cropped the file linked by @greydongilmore to 5s (and removed most channels as well as downsampled to 256Hz). If @greydongilmore gives us permission to use this cropped file (17kB) I'd upload it to mne-testing-data so that @agramfort can add a test. There are only 2 annotations in it (`XLSpike` and `Clip Note`).

agramfort:
@cbrnr ok for you?

cbrnr:
Yes, perfect! Now let's wait for permission to use the small test file.

agramfort:
the file shared with me is 30MB

do you have an edf file that makes you pass in the if block with an offset
different from 0?

none of our test files have this.

>

agramfort:
cool for the tiny file. Can you share it with me?

>

cbrnr:
I cropped the file that @greydongilmore shared with you. It's 17kB and has the sub-second start time.

agramfort:
where is the file?

cbrnr:
[subsecond_starttime.edf.zip](https://github.com/mne-tools/mne-python/files/4745184/subsecond_starttime.edf.zip)

agramfort:
I am putting this in the testing data. thanks

>

agramfort:
@cbrnr merge if you're happy

mne/io/edf/tests/test_edf.py
```diff
@@ -385,3 +386,10 @@ def test_edf_lowpass_zero():
     with pytest.warns(RuntimeWarning, match='too long.*truncated'):
         raw = read_raw_edf(edf_stim_resamp_path)
     assert_allclose(raw.info["lowpass"], raw.info["sfreq"] / 2)
+
+
+@testing.requires_testing_data
+def test_edf_annot_ms_onset():
+    """Test reading of sub-milisecond annotation onsets."""
```
cbrnr, line 393: To be exact, this should be "sub-second annotation onsets". Accordingly, the name for the test data should be `edf_annot_sub_s_path`.

cbrnr:
Thanks @agramfort!

greydongilmore:
@cbrnr @agramfort thanks for quick action on this! 👍

agramfort:
great team effort !

>

PR #7877, opened 2020-06-08
MRG, ENH: Allow disabling FXAA

larsoner:
@bloyl @kingjr can you see if this fixes your OSMESA / remote rendering experience in `mne coreg` and `stc.plot`? For the latter you'll also need to pass `stc.plot(..., antialias=False)` and if you use PySurfer you'll need to be on https://github.com/nipy/PySurfer/pull/295

@GuillaumeFavelier can you take a look and make sure I disabled it in the right place(s)?

### DIFF
--- a/mne/gui/_coreg_gui.py
+++ b/mne/gui/_coreg_gui.py
@@ -2027,8 +2027,10 @@ class CoregFrame(HasTraits):
             renderer.use_depth_peeling = 0
             renderer.vtk_window.multi_samples = 8
             renderer.vtk_window.alpha_bit_planes = 0
-            if hasattr(renderer, 'use_fxaa'):
-                self.scene.renderer.use_fxaa = True
+            # This can break things on MESA and it's not really necessary
+            # for coreg, so let's not set it.
+            # if hasattr(renderer, 'use_fxaa'):
+            #     self.scene.renderer.use_fxaa = True
         self.scene.render()
 
     @on_trait_change('lock_fiducials')
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -1586,7 +1586,7 @@ class SourceEstimate(_BaseSurfaceSourceEstimate):
              cortex="classic", size=800, background="black",
              foreground=None, initial_time=None, time_unit='s',
              backend='auto', spacing='oct6', title=None,
-             show_traces='auto', verbose=None):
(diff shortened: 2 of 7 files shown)


GuillaumeFavelier:
Everything looks okay to me @larsoner

bloyl:
I'll look more closely later today but this looks like it only fixes `stc.plot`?

Would it be possible to use a global config variable that is respected by the 3d renderers? That would have the benefit of fixing all 3d plotting (`plot_alignment`, etc.) and also let the same code work in a number of settings.

larsoner:
Ahh right, we'd need to update every 3D plotting function to take this argument. That's probably not so nice. We could have something like a `mne.viz.set_3d_options(antialias=None)` -- for now it would just be `antialias` as the only arg but eventually we might want to add other things (depth peeling?). @GuillaumeFavelier WDYT? Other API ideas? And also maybe we want a new `MNE_3D_OPTION_ANTIALIAS` so that users that always want this off can just set a bash env var and never worry about it.

GuillaumeFavelier:
>  update every 3D plotting function to take this argument

If it's global, why update every function? It's not like you would like to have specific plots with AA and some without it. I think we could just modify the renderers directly and `mne.coreg`. It's less painful to maintain too.

I like `set_3d_options`. Do we have a place to save such settings or we can just modify `os.environ`?

larsoner:
> I like set_3d_options. Do we have a place to save such settings or we can just modify os.environ?

We shouldn't use `os.environ` to pass things around unless we need to. I think I know a simple way to implement it, let me give it a shot...

larsoner:
> If it's global, why update every function?

I meant if we followed what I was originally doing we'd need to. With the global approach (implemented now), indeed we don't.

@GuillaumeFavelier can you look at the changes and let me know what you think? @bloyl in principle this should now work with all 3D functions in MNE. You can do either `mne.viz.set_3d_options(antialias=False)` in a given session, or set `MNE_3D_OPTIONS_ANTIALIAS=false` as an env var.

GuillaumeFavelier:
`mne/viz/_3d.py` is a nice place indeed, thanks for adding this. Also, LGTM

GuillaumeFavelier:
I pushed a commit to fix the import in `_pyvista.py`

larsoner:
Okay @bloyl @kingjr let us know if this helps and we'll go ahead and merge

PR #7878, opened 2020-06-08
MRG, FIX: Fix get_channel_types

### LINKED ISSUE #7837: Magnetometer positions look odd in PSD plot
#### Steps to reproduce
Let's first visualize the magnetometer channel positions via `Raw.plot_sensors()`:
```python
import mne


fname = data_path + '/MEG/sample/sample_audvis_raw.fif'
raw = mne.io.read_raw_fif(fname)
raw.plot_sensors(ch_type='mag')
```
This produces:
sensors

Looking good. Now let's plot PSDs of the data.

```python
raw.plot_psd()
```
Result cropped so it only shows the sensor positions (top to bottom: mag, grad, eeg):
Figure_1 copy

The magnetometer sensor layout looks incorrect, and doesn't match the one produced by `plot_sensors()`. Am I missing something here?

cc @agramfort

larsoner:
Fixes a bug with `plot_psd` that was introduced by #7486. Basically the order in `picks` was not preserved, which was required by `plot_raw_psd` when there are multiple channel types.

Closes #7837

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -180,6 +180,10 @@ Bug
 
 - Fix bug with :func:`mne.bem.make_watershed_bem` where the RAS coordinates of watershed bem surfaces were not updated correctly from the volume file by `Yu-Han Luo`_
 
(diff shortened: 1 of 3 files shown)


agramfort:
thanks heaps @larsoner !

>

PR #7880, opened 2020-06-08
MAINT: Tweak versions for doc builds [circle front]

larsoner:


drammock:
@larsoner since you're already PRing into `0.20`, I never did backport #7804...  would it make sense to cherry-pick 683613a into this PR?

larsoner:
Good idea, done

PR #7883, opened 2020-06-09
MRG, ENH: Add method to project onto max power ori

larsoner:
Useful for:

1. Converting `pick_ori='vector'` solutions to max-power-ori solutions
2. Eventually for `unit-noise-gain` vector beamformers we'll want to do this (probably)

In the beamformer PR, we can also check to see the extent to which `pick_ori='max-power'` agrees with `pick_ori='vector'` followed by `stc.project_onto()`.

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -1807,12 +1807,80 @@ class _BaseVectorSourceEstimate(_BaseSourceEstimate):
             cortex.
         """
         _check_src_normal('normal', src)
+        normals = self._get_src_normals(src, use_cps)
+        return self.project_onto(normals)
+
+    def _get_src_normals(self, src, use_cps):
         normals = np.vstack([_get_src_nn(s, use_cps, v) for s, v in
-                             zip(src, self.vertices)])
-        data_norm = einsum('ijk,ij->ik', self.data, normals)
-        return self._scalar_class(
+                            zip(src, self.vertices)])
+        return normals
+
+    def project_onto(self, directions=None, src=None, use_cps=True):
+        """Project the data for each vertex in a given direction.
+
+        Parameters
+        ----------
+        directions : ndarray, shape (n_src, 3) | None
+            If None, SVD will be used to project onto the direction of
+            maximal power.
+        src : instance of SourceSpaces | None
+            Only used when ``directions is None``. See Notes.
+        %(use_cps)s
+            Should be the same value that was used when the forward model
+            was computed (typically True).
+
+        Returns
+        -------
+        stc : instance of SourceEstimate
+            The projected source estimate.
+        normals : ndarray, shape (n_src, 3)
+            Only returned if ``normals is None``, the normals computed
+            using SVD.
+
+        Notes
+        -----
+        When using SVD, there is a sign ambiguity for the direction of maximal
+        power. When ``src is None``, the direction is chosen that makes the
+        resulting time waveform sum positive (i.e., have positive amplitudes).
+        When ``src`` is provided, the directions are flipped in the direction
+        of the source normals, i.e., outward from cortex for surface source
+        spaces and in the +Z / superior direction for volume source spaces.
+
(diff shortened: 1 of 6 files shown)


mne/tests/test_source_estimate.py
```diff
@@ -1188,6 +1213,37 @@ def test_vec_stc_basic(tmpdir, klass, kind):
...
+            x = x / np.linalg.norm(x)
+            return -np.linalg.norm(np.dot(x, data[ii]))
+        want_nn[ii] = fmin_cobyla(objective, x0, (), rhobeg=0.1, rhoend=1e-6)
+    want_nn /= np.linalg.norm(want_nn, axis=1, keepdims=True)
+
+    stc = VolVectorSourceEstimate(data, [np.arange(n_src)], 0, 1)
+    stc_max, directions = stc.project_onto()
+    flips = np.sign(np.sum(directions * want_nn, axis=1, keepdims=True))
+    directions *= flips
+    assert_allclose(directions, want_nn, atol=1e-6)
```
larsoner, line 1244: @agramfort here is your proof-by-COBYLA :)

britta-wstnr:
I will have a look later whether this performs like what I wrote yesterday (which is how  @sarangnemo implemented it in NutMEG back then) - I guess @larsoner got so excited by the plots I shared that he had to try by himself! :smile: 

For what it's worth, the code I wrote up yesterday performs close to a scalar beamformer with optimal orientation picking - I have some preliminary code to also allow orientations to vary across time, which might make things more interesting, which I'd be happy to contribute.

britta-wstnr:
image

So our methods agree - left image is my version (minus polarity flipping), right image is your version !

larsoner:
Excellent -- in that case this is good to go @agramfort . Once this is merged I'll rebase the beamformer refactoring PR and we can properly test the two `vector` "unit-noise-gain" methods with and without SVD afterward

mne/source_estimate.py
```diff
@@ -1807,12 +1806,81 @@ def normal(self, src, use_cps=True):
...
+        %(use_cps)s
+            Should be the same value that was used when the forward model
+            was computed (typically True).
+
+        Returns
+        -------
+        stc : instance of SourceEstimate
+            The projected source estimate.
+        normals : ndarray, shape (n_vertices, 3)
+            Only returned if ``normals is None``, the normals computed
```
agramfort, line 1837:
```suggestion
            Only returned if ``directions is None``.
```

larsoner:
> should directions be a public parameter?

It was easy enough to add and validation was simple. People could want to for example project onto the max power orientations they got from some other filter. So I don't see a problem/maintenance cost from keeping it.

> we have stc.normal so should we rather of stc.principal_component?

Starting from scratch I'd actually prefer the API to be `stc.project_onto('normal' | 'pca')` or so. I can rename it to `principal_component` but if we think of other ways to do the projections our namespace gets more annoying -- I'd rather keep and potentially in the future add additional options to a generic `project_onto` function...

agramfort:
ok I propose to deprecate VectorSourceEstimate.normal and introduce

VectorSourceEstimate.project(directions, src=None, use_cps=True, return_directions=False)

where directions can be
'normal'  -> requires src
'pca' -> does not require src? and use return_directions if you want to get the estimated directions
array of shape (n_dipoles, 3)

sounds like a plan?

larsoner:
I'd rather just always return the directions, return_ kwargs I'd rather leave as a mechanism for backward compatibility. It's very easy to do [0] to get the first output of a function.

src is mandatory in both cases to make the sign flip directions not arbitrary

larsoner:
(deal with sign flip ambiguity in pca)

agramfort:
ok then to remove return_ param and always return directions.

we have a plan?

larsoner:
Yep, I'll push shortly

agramfort:
Thx @larsoner

PR #7888, opened 2020-06-10
MRG, ENH: Add mri_resolution="sparse"

### LINKED ISSUE #7886: Using surface source spaces with SourceSpaces.export_volume writes out all possible surface vertices
## Describe the bug
Not necessarily a bug, but unexpected behaviour (at least for me):
Writing out surface source spaces from a mixed source space using export_volume always writes out the full list of vertices in the surface used to create the source space, not the vertices that are "inuse", i.e. sources. Is this intentional? If yes, would it be sensible to include an option to write only  source vertices in use?

#### Steps to reproduce
'''
subjects_dir = f'{data_path}/subjects'
spacing = 'ico2'
n_jobs=1
sphere = (0,0,0,0.12)
surf_src = mne.setup_source_space('fsaverage', spacing=spacing,
                                   subjects_dir=subjects_dir,
                                   n_jobs=n_jobs, add_dist=False)
lh_cereb= mne.setup_volume_source_space("fsaverage", mri=file_path+"/fsaverage/mri/aseg.mgz", sphere=sphere,
                                        volume_label="Left-Cerebellum-Cortex", sphere_units="m")
rh_cereb= mne.setup_volume_source_space("fsaverage", mri=file_path+"/fsaverage/mri/aseg.mgz", sphere=sphere,
                                        volume_label="Right-Cerebellum-Cortex", sphere_units="m")
src=surf_src + lh_cereb + rh_cereb
nii_name=file_path+"/MNESource_cm2.nii"
src.export_volume(nii_name, mri_resolution=True, use_lut=False, overwrite=True)

'''

#### Expected results
When exporting surface source spaces along with volumetric ones I would expect it to export only the surface vertices that are in use, and label voxels within the volume that do not coincide with the location of a source vertex that is in use as background/ 0 

#### Actual results
When exporting surface sources along with volume sources the full vertex list gets written, regardless of how many vertices are in use

larsoner:
Closes #7886.

@susamerz can you see if this takes care of your use case? It allows only filling in the nearest neighbor voxel to the actual source space points for both surfaces and volumes.

I didn't add it to an example because it doesn't look good with the native `nilearn` plotters because it's so sparse...

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -73,6 +73,8 @@ Changelog
 
 - Add estimation method legend to :func:`mne.viz.plot_snr_estimate` by `Eric Larson`_
 
+- Add support to `mne.SourceSpaces.export_volume` for ``mri_resolution='sparse'`` to color only the nearest-neighbor voxels instead of entire regions by `Eric Larson`_
+
 - Add ``axes`` argument to :func:`mne.viz.plot_evoked_white`, :meth:`mne.Evoked.plot_white`, and :func:`mne.viz.plot_snr_estimate` by `Eric Larson`_
 
 - Add ``sources`` and ``detectors`` options for fNIRS use of :meth:`mne.viz.plot_alignment` allowing plotting of optode locations in addition to channel midpoint ``channels`` and ``path`` between fNIRS optodes by `Kyle Mathewson`_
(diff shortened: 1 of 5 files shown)


susamerz:
Yes, this is exactly what I was looking for, thank you

Issue #7313, opened 2020-02-13
Can't load raw nirx data, due to encoding issue


benoitvalery:
When I try to load NirX data with the `mne.io.read_raw_nirx` function, I'm facing encoding issue with the `.hdr` file. For now, I'm solving this manually by converting the `.hdr` to `utf8` with a simple text file editor like `geany`.

#### MWE

I recorded a test dataset (in a `test_github` folder), which is composed of the following files :
```bash
./NIRS-2020-02-13_001.evt  --  inode/x-empty; charset=binary
./NIRS-2020-02-13_001.set  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001.wl1  --  text/plain; charset=us-ascii
./Standard_probeInfo.mat  --  application/octet-stream; charset=binary
./NIRS-2020-02-13_001.dat  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001.tpl  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001_config.txt  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001.wl2  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001.hdr  --  application/x-wine-extension-ini; charset=iso-8859-1
./NIRS-2020-02-13_001.inf  --  text/plain; charset=us-ascii
./NIRS-2020-02-13_001.avg  --  application/octet-stream; charset=binary
```
These informations were obtained with the following command (linux) : 
```bash
for f in `find | egrep -v Eliminate`; do echo "$f" ' -- ' `file -bi "$f"` ; done
```
On python side, here is the code that I intend to use to load the data. I'm using the last version of MNE (0.20.dev0).
```python
#!/usr/bin/env python3
import os
import mne

path = os.sep.join([os.getcwd(), 'test_github'])
raw_intensity = mne.io.read_raw_nirx(path, verbose=True).load_data()
```
As mentioned [here](https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html?highlight=nirs), I should obtain this kind of output :
```python
Loading /home/circleci/mne_data/MNE-fNIRS-motor-data/Participant-1
Reading 0 ... 23238  =      0.000 ...  2974.464 secs...
```
But actually, the `read_raw_nirx` command raises the following Traceback :
```python


Loading /home/bvaler01/Documents/programmes/NBack/test_github
Traceback (most recent call last):
[...]

PR #7891, opened 2020-06-12
MRG: Extract measurement date and age for NIRX files

rob-luke:
Extract measurement date and birthday for NIRX files.

#### Additional information
The vendor software only allows saving of age as an integer. So I have taken the measurement date, subtracted the age integer from the years, and stored that as the birthday. Open to alternative suggestions.

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -6,6 +6,8 @@ from configparser import ConfigParser, RawConfigParser
 import glob as glob
 import re as re
 import os.path as op
+import datetime as dt
+import locale
 
 import numpy as np
 
@@ -15,8 +17,7 @@ from ..constants import FIFF
 from ..meas_info import create_info, _format_dig_points
 from ...annotations import Annotations
 from ...transforms import apply_trans, _get_trans
-from ...utils import logger, verbose, fill_doc
-from ...utils import warn
+from ...utils import logger, verbose, fill_doc, warn
 
 
 @fill_doc
@@ -118,22 +119,51 @@ class RawNIRX(BaseRaw):
 
         # Parse required header fields
 
+        # Extract measurement date and time
+        datetime_str = hdr['GeneralInfo']['Date'] + hdr['GeneralInfo']['Time']
+        meas_date = None
+        # Several formats have been observed so we try each in turn
+        try:
+            dt_code = '"%a, %b %d, %Y""%H:%M:%S.%f"'
+            meas_date = dt.datetime.strptime(datetime_str, dt_code)
+            meas_date = meas_date.replace(tzinfo=dt.timezone.utc)
+        except ValueError:
+            pass
+        try:
+            dt_code = '"%a, %d %b %Y""%H:%M:%S.%f"'
+            meas_date = dt.datetime.strptime(datetime_str, dt_code)
+            meas_date = meas_date.replace(tzinfo=dt.timezone.utc)
+        except ValueError:
+            pass
+        if meas_date is None:
+            warn("Extraction of measurement date from NIRX file failed. "
+                 "This can be caused by files saved in certain locales. "
+                 "Please report this as a github issue. "
+                 "The date is being set to January 1st, 2000, "
+                 "instead of {}".format(datetime_str))
+            meas_date = dt.datetime(2000, 1, 1, 0, 0, 0,
+                                    tzinfo=dt.timezone.utc)
+
         # Extract frequencies of light used by machine
         fnirs_wavelengths = [int(s) for s in
                              re.findall(r'(\d+)',
-                             hdr['ImagingParameters']['Wavelengths'])]
+                                        hdr['ImagingParameters'][
+                                            'Wavelengths'])]
 
         # Extract source-detectors
         sources = np.asarray([int(s) for s in re.findall(r'(\d+)-\d+:\d+',
-                              hdr['DataStructure']['S-D-Key'])], int)
+                                                         hdr['DataStructure'][
+                                                             'S-D-Key'])], int)
         detectors = np.asarray([int(s) for s in re.findall(r'\d+-(\d+):\d+',
-                                hdr['DataStructure']['S-D-Key'])], int)
+                                                           hdr['DataStructure']
+                                                           ['S-D-Key'])],
+                               int)
 
         # Determine if short channels are present and on which detectors
         if 'shortbundles' in hdr['ImagingParameters']:
             short_det = [int(s) for s in
                          re.findall(r'(\d+)',
-                         hdr['ImagingParameters']['ShortDetIndex'])]
+                                    hdr['ImagingParameters']['ShortDetIndex'])]
             short_det = np.array(short_det, int)
         else:
             short_det = []
@@ -150,6 +180,7 @@ class RawNIRX(BaseRaw):
         # Note: NIRX also records "Study Type", "Experiment History",
         #       "Additional Notes", "Contact Information" and this information
         #       is currently discarded
+        # NIRStar does not record an id, or handedness by default
         subject_info = {}
         names = inf['name'].split()
         if len(names) > 0:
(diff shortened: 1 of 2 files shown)


rob-luke:
@larsoner this is failing due to processing of French date strings. Is there any inbuilt multilingual date string parsers in MNE already or do I need to write one?

rob-luke:
HHmmmm my changes also seem to break other things, Ill fix that. But still curious about the best way to process French dates.

agramfort:
what do you mean French dates? like month with letter with accents?

larsoner:
Like Travis says:
```
ValueError: time data '"jeu. 13 févr. 2020""09:08:47.511"' does not match format '"%a, %b %d, %Y""%H:%M:%S.%f"'
```

agramfort:
is it fixable by setting the LOCALE properly?

>

larsoner:
The docs suggest that `locale` is used, so indeed if the `locale` can be extracted from the file header, then we can make a `@contextlib.contextmanager` to `orig = locale.set_locale(locale.LC_something, whatever)` and then `set_locale(locale.LC_something, orig)` afterward. Even if it isn't embedded in the header (which would be sad), we could start making a list of ones to iterate over and try, and always add others later if need be. (Or maybe it won't be so bad to iterate over all of them, if we can get a list?)

rob-luke:
Thanks for the helpful suggestions. The locale is not embedded in the header. I had started writing a list of locales to iterate over, that's when I stopped and decided to ask here before progressing too far. So it sounds like there is no solution in MNE for this already, so I will continue to code along this path.

larsoner:
Indeed AFAIK we have never had to deal with locale when parsing datetime. We can try adding some other `LC_` env var to some Travis or Azure run to see if we can get some existing `datetime` stuff to break. Let's do that after you fix the reading of this particular file, though 
-- it will probably tell us which env var(s) are relevant.

agramfort:
here is an example that works:

import locale
from datetime import datetime
locale.setlocale(locale.LC_TIME, "fr_FR")
s = "jeu. 13 fév. 2020""09:08:47.511"
f = "%a. %d %b. %Y""%H:%M:%S.%f"
datetime.strptime(s, f)

however I had to use fév. and févr. which seems to be the standard.

rob-luke:
> however I had to use fév. and févr. which seems to be the standard.

Thanks for the code snippet. I came up with something very similar and ran in to the same problem. It's the `févr` that is the problem, `datetime.strptime` seems to expect `fev`. I considered some sort of replace approach for different languages, but this seems unwieldily (how many languages to I then support?). I also considered other packages that are made for dealing with dates, but don't want to add more requirements.

rob-luke:
Next Im going to check where this `févr` actually comes from. This may be a problem of our own creating, I have never seen a file from France and don't know if the device will actually save as `fev` or `févr`, this seems to have been[ generated somewhere in `_test_raw_reader`.](https://github.com/mne-tools/mne-python/blob/e69e581d58bdeada61c8b863885387aa28ede1da/mne/io/tests/test_raw.py#L43)

agramfort:
I see with company that produces this file how the come up with févr while
it appears the standard across languages is fév

I would use  try except here to avoid failing to load such files but I
agree the mne code cannot fix all IO corner cases.

can you see with them how it happens? and maybe how they deal with this
internally?

>

larsoner:
@rob-luke will you have time to revisit this in the next couple of weeks? I'll optimistically mark for 0.21

rob-luke:
Yes I would like to cross this off. The milestone has sept 15 marked, that should be achievable

rob-luke:
@larsoner I am not sure what to do here with the non standard French encoding of dates.

It seems to me that `fév` is what datetime expects, yet we have 'févr' in the tests because that was provided in #7313 (I assume, I don't actually have the file).

So if NIRX is providing non standard french date formats, and if I know all the non standard naming, I could create a translation function. But I don't know all the non standard naming and don't have access to a machine at the moment to determine this. Plus this would become quite unwieldy, what other non standard language formats do we then support.

Would it be acceptable to try and parse the dates in the current `locale` and throw a warning if it doesn't work. For dates that we can't extract it will set them to 0 or something else? This will get English working and downstream usage of the date, and when someone with access to a French system gets some data they can just tweak the French parsing. Thoughts?

agramfort:
yes I fear we need something custom. But I would reach out to
engineers at the company
that produces these files to report the issue.

rob-luke:
> I would reach out to engineers at the company that produces these files to report the issue.

Good idea. But I would like to verify the issue myself first if I am to contact them. Ideally someone could send me some files recorded in France. Failing that, I could set the local of our lab pc to France and record my own file, but Im currently locked out of the lab (at least for the next 3 weeks). So this won't be done by 0.21 release date. I suggest I create an English date PR for 0.21 and raise an issue for the non English dates to be solved when someone sends me some files or I can get in the lab. Acceptable? Or do you prefer I wait and do everything at once?

agramfort:
acceptable

rob-luke:
@agramfort @larsoner can you please review.

mne/io/nirx/nirx.py
```diff
@@ -222,7 +253,8 @@ def __init__(self, fname, preload=False, verbose=None):
         def prepend(list, str):
             str += '{0}'
             list = [str.format(i) for i in list]
-            return(list)
+            return (list)
```
agramfort, line 256:
```suggestion
            return list
```

besides you should not call a variable list

mne/io/nirx/tests/test_nirx.py
```diff
@@ -79,6 +79,12 @@ def test_nirx_15_2_short():
     # Test data import
     assert raw._data.shape == (26, 145)
     assert raw.info['sfreq'] == 12.5
+    assert raw.info['meas_date'].month == 8
+    assert raw.info['meas_date'].day == 23
+    assert raw.info['meas_date'].year == 2019
+    assert raw.info['meas_date'].hour == 7
+    assert raw.info['meas_date'].minute == 37
+    assert raw.info['meas_date'].second == 4
```
agramfort, line 87: I would do it with == after creating a datetime instance with proper timezone. It's less much lines.

mne/io/nirx/tests/test_nirx.py
```diff
@@ -268,16 +278,24 @@ def test_nirx_15_2():
     # Test data import
     assert raw._data.shape == (64, 67)
     assert raw.info['sfreq'] == 3.90625
+    assert raw.info['meas_date'].month == 10
+    assert raw.info['meas_date'].day == 2
+    assert raw.info['meas_date'].year == 2019
+    assert raw.info['meas_date'].hour == 9
+    assert raw.info['meas_date'].minute == 8
+    assert raw.info['meas_date'].second == 47
```
agramfort, line 286: same here one assert should do it

mne/io/nirx/tests/test_nirx.py
```diff
@@ -312,6 +330,12 @@ def test_nirx_15_0():
     # Test data import
     assert raw._data.shape == (20, 92)
     assert raw.info['sfreq'] == 6.25
+    assert raw.info['meas_date'].month == 10
+    assert raw.info['meas_date'].day == 27
+    assert raw.info['meas_date'].year == 2019
+    assert raw.info['meas_date'].hour == 13
+    assert raw.info['meas_date'].minute == 53
+    assert raw.info['meas_date'].second == 34
```
agramfort, line 338: idem

mne/io/nirx/nirx.py
```diff
@@ -118,22 +118,51 @@ def __init__(self, fname, preload=False, verbose=None):
...
+            meas_date = dt.datetime.strptime(datetime_str, dt_code)
+            meas_date = meas_date.replace(tzinfo=dt.timezone.utc)
+        except ValueError:
+            pass
+        try:
+            dt_code = '"%a, %d %b %Y""%H:%M:%S.%f"'
+            meas_date = dt.datetime.strptime(datetime_str, dt_code)
+            meas_date = meas_date.replace(tzinfo=dt.timezone.utc)
+        except ValueError:
+            pass
```
agramfort, line 136:
```suggestion
        try:
            for dt_code in ['"%a, %b %d, %Y""%H:%M:%S.%f"', '"%a, %d %b %Y""%H:%M:%S.%f"']:
                meas_date = dt.datetime.strptime(datetime_str, dt_code)
                meas_date = meas_date.replace(tzinfo=dt.timezone.utc)
                break
        except ValueError:
            pass
```

rob-luke:
Thanks for the review. @agramfort and @larsoner could you please review again.
CI fail seems to be a time out error

agramfort (approved):

agramfort:
thx @rob-luke

PR #7896, opened 2020-06-12
MRG, BUG: Fix bug with volume morph and subject_to!="fsaverage"

larsoner:
We had a bug where `src_to` did not work when the MRI and RAS coordinate frames were not identical. Fortunately they are for fsaverage (hopefully by far the most common destination subject), unfortunately I have never seen them be identical for any other FreeSurfer subject (hopefully a rare destination use case).

WIP because ~~using non-fsaverage subjects is still not totally satisfactory (though they at least now are more reasonable instead of all zero), want to see if there is anything to fix or if it's just a parameter problem at my end.~~ :

- [x] Need to make tests better

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -33,6 +33,8 @@ Changelog
 
 - Add support for mixed source spaces to :func:`mne.compute_source_morph` by `Eric Larson`_
 
+- Add support for omitting the SDR step in volumetric morphing by passing ``n_iter_sdr=()`` to `mne.compute_source_morph` by `Eric Larson`_
+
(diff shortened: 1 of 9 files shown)


larsoner:
Okay this one should be good to go. I noticed that the volumetric morphing has an issue where the edge voxel values do not round-trip to 1. because of the reslicing / interpolation steps using linear interpolation, but I'm not sure that there is a good way around this. That's a problem on `master` and this PR, so we can tackle that later if need be.

agramfort:
@larsoner any idea why the CIs are failing?

larsoner:
Just timeouts that should be fixed by #7904, this should be good to review

agramfort:
thx @larsoner

PR #7898, opened 2020-06-14
MRG: Warn if NIRX directory structure has been modified from original format

### LINKED ISSUE #7894: [Question][fNIRS] The use of .dat file
Dear MNE devs and users,

I'm not quite used to fNIRS nor the MNE package (started my internship last week) and I can't figure out the use the *.dat file.

In your [tutorial](https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#tut-fnirs-processing) concerning the Preprocessing of fNIRS data, the function `mne.io.read_raw_nirx(fnirs_raw_dir, verbose=True)` (line 11 of the first box) requires a *.dat file to create a **RawNIRX** instance, although my tutor explained to me that *.dat are essentially files used to display data in real-time (i.e. during the experiment) and as such, are based on standard parameters : they're not usable after the experiment because they do not contain useful (to be processed) data, since the probed data is stored in *.wl1 and *.wl1, events in *.evt and the metadata in *.hdr and *_probeInfo.mat.

I searched through the code and even though you do [load](https://github.com/mne-tools/mne-python/blob/f6c6ff2842ff020816bfc203f00e03410c9e0ea3/mne/io/nirx/nirx.py#L74) the file, I can't find where you use it (or if you use it all)? Am I missing something ?

Thanks in advance,

David

rob-luke:
#### Reference issue
Fixes #7894. @Swy7ch noted that the NIRX file reader checks for a dat file, but it is not used.

Instead of failing if the `.dat` is missing this now throws a warning to the user that the file structure has been modified from what the measurement machine saved.

#### Additional information
We have had some discussion about what the correct behaviour is. See the above linked issue for discussion.

### DIFF
--- a/mne/io/nirx/tests/test_nirx.py
+++ b/mne/io/nirx/tests/test_nirx.py
@@ -5,6 +5,7 @@
 
 import os.path as op
 import shutil
+import os
 
 import pytest
 from numpy.testing import assert_allclose, assert_array_equal
@@ -34,6 +35,17 @@ def test_nirx_hdr_load():
     # Test data import
     assert raw._data.shape == (26, 145)
     assert raw.info['sfreq'] == 12.5
+
+
+@requires_testing_data
+def test_nirx_dat_warn(tmpdir):
+    """Test reading NIRX files when missing data."""
+    shutil.copytree(fname_nirx_15_2_short, str(tmpdir) + "/data/")
+    os.rename(str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.dat",
+              str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.tmp")
+    fname = str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.hdr"
+    with pytest.raises(RuntimeWarning, match='A single dat'):
+        read_raw_nirx(fname, preload=True)
 
 
 @requires_testing_data
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -15,6 +15,7 @@ from ..meas_info import create_info, _format_dig_points
 from ...annotations import Annotations
 from ...transforms import apply_trans, _get_trans
 from ...utils import logger, verbose, fill_doc
+from ...utils import warn
 
 
 @fill_doc
@@ -71,7 +72,7 @@ class RawNIRX(BaseRaw):
 
         # Check if required files exist and store names for later use
         files = dict()
-        keys = ('dat', 'evt', 'hdr', 'inf', 'set', 'tpl', 'wl1', 'wl2',
+        keys = ('evt', 'hdr', 'inf', 'set', 'tpl', 'wl1', 'wl2',
                 'config.txt', 'probeInfo.mat')
         for key in keys:
             files[key] = glob.glob('%s/*%s' % (fname, key))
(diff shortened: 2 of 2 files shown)


mne/io/nirx/tests/test_nirx.py
```diff
@@ -36,6 +37,18 @@ def test_nirx_hdr_load():
...
+@requires_testing_data
+def test_nirx_dat_warn():
+    """Test reading NIRX files when missing data."""
+    os.rename(fname_nirx_15_2_short + "/NIRS-2019-08-23_001.dat",
+              fname_nirx_15_2_short + "/NIRS-2019-08-23_001.tmp")
+    fname = fname_nirx_15_2_short + "/NIRS-2019-08-23_001.hdr"
+    with pytest.raises(RuntimeWarning, match='A single dat'):
+        read_raw_nirx(fname, preload=True)
+    os.rename(fname_nirx_15_2_short + "/NIRS-2019-08-23_001.tmp",
+              fname_nirx_15_2_short + "/NIRS-2019-08-23_001.dat")
```
agramfort, line 49: this is for me dangerous as if the lines above fail you will never revert the renaming.

rob-luke: Agreed. Could I not have a test for this? It will drop the code coverage :(

Or is there another way you would suggest?

agramfort:
maybe a try finally construct?

>

mne/io/nirx/tests/test_nirx.py
larsoner: @rob-luke the way we usually solve this is with the pattern:
```
def test_whatever(tmpdir):  # pytest builtin tmpdir fixture
    shutuil.copytree(orig_dir, str(tmpdir))
    # mess with files in tmpdir
    ...
    read_raw_nirx(tmpdir.join('whatever.ext'))
```
pytest takes care of creating and destroying the `tmpdir` directory

rob-luke: beautiful. Thanks

rob-luke:
@larsoner @agramfort could you please review, this failure seems due to the Mac CI exceeding its time limit.

rob-luke:
a 16 second turn around. Surely that's a record, thanks @agramfort

agramfort:
:)

>

mne/io/nirx/tests/test_nirx.py
```diff
@@ -36,6 +37,17 @@ def test_nirx_hdr_load():
...
 
 
+@requires_testing_data
+def test_nirx_dat_warn(tmpdir):
+    """Test reading NIRX files when missing data."""
+    shutil.copytree(fname_nirx_15_2_short, str(tmpdir) + "/data/")
+    os.rename(str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.dat",
+              str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.tmp")
+    fname = str(tmpdir) + "/data" + "/NIRS-2019-08-23_001.hdr"
+    with pytest.raises(RuntimeWarning, match='A single dat'):
```
larsoner, line 47: usually we use `pytest.warns`, but I suppose this works, too

larsoner:
Thanks @rob-luke

Issue #7901, opened 2020-06-15
Rename `master` branch


hoechenberger:
Hello,

amid the global [BLM protests](https://en.wikipedia.org/wiki/Black_Lives_Matter), it seems a number of free/open-source software projects are now [reconsidering](https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/) the naming of their main source-code management branch, historically called `master` in `git`. Microsoft seems to be [actively working](https://twitter.com/natfriedman/status/1271253144442253312) on bringing out-of-the-box support for other naming schemes to GitHub.

Despite its widespread use in the contexts of software development and hardware interfaces, the term "master" **may** be considered offensive by some. While one could argue that this perception entirely disregards context, I would argue that nothing is lost by dropping the name. I would consider moving to a different name as a token of support for the struggles of ethnic minorities and marginalized people; and it would be a public display of some (granted, cheap, but still visible) efforts to actively embrace an inclusive community. Lastly I would argue that our Code of Conduct actually doesn't leave us much of a choice anyway, as it clearly [states](https://github.com/mne-tools/.github/blob/master/CODE_OF_CONDUCT.md#our-standards) we vow to use "welcoming and inclusive language". Now if the `master` term has been identified to be non-inclusive – regardless of whether **we** feel that this is true or not – it should be dropped.

Looking forward to reading your thoughts on this!

Best wishes,

–Richard

cbrnr:
By all means change it if it makes people happy. GitHub will soon default to `main` anyway. However, I think that in this case the term master has nothing to do with master/slave (yes I know the BitKeeper background, but in a plain Git repo there is a master branch but no slave branch(es)). Same for blacklisting/whitelisting, which actually originates from the term black book (which were actual black books).

I'd only caution to wait until we know all implications of that name change. I've read that it can be non-trivial in some cases (e.g. it seems to be hard-coded for the docs), so let's do it only after GitHub releases detailed instructions how to do it.

hoechenberger:
Hi @cbrnr 
> However, I think that in this case the term master has nothing to do with master/slave (yes I know the BitKeeper background, but in a plain Git repo there is a master branch but no slave branch(es)). Same for blacklisting/whitelisting, which actually originates from the term black book (which were actual black books).

Yes I do agree, my perception is similar. But others seem to disagree, and I don't mind making a small change to them them feel more welcome ☺️

[...]

PR #7902, opened 2020-06-15
MRG, MAINT: Check usage

larsoner:
Codespell recently added a `usage` dictionary to move toward #7901. This just has us use it, with an exception for `master` for now (no other changes needed for our codebase).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,10 +105,10 @@ flake:
 	@echo "flake8 passed"
 
 codespell:  # running manually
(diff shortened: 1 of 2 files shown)


cbrnr:
Not sure why the Python 3.8 job keeps failing because it times out after 1 hour - is this OK?

larsoner:
I'm working on adding more tests to slow to try to work around it

PR #7904, opened 2020-06-16
MRG, FIX: Speed up I/O tests, mark some slow

larsoner:
Speeds up the H5 info round-trip test by changing some list of dict to dict of array and back (saves 9 sec locally in the BTi test because there are >3k dig points). Eventually we can make use of this in things that use `write_hdf5` if we find that they are slow in practice. I looked and currently TFR seems to be the only place this might be useful but the implementation was not going to be trivial because we support writing lots at once, but we can revisit it later if need be.

Tightens up our use of types to be more consistent across readers (it makes `object_diff` happy but should be a good idea in general).

Also marks a couple more tests as slow that keep showing up in macOS like:

https://travis-ci.org/github/mne-tools/mne-python/jobs/697952162#L3269

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,7 +141,7 @@ script:
         pip install -e .;
         python mne/tests/test_evoked.py;
       fi;
-    - echo pytest -m "${CONDITION}" --cov=mne -n 1 ${USE_DIRS}
+    - echo 'pytest -m "${CONDITION}" --cov=mne -n 1 ${USE_DIRS}'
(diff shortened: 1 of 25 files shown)


larsoner:
OSX is much safer at 42 minutes now (50 min limit), pushing a commit to change `not ultraslowtest` to `not slowtest` on Azure since those are all very close to their 60 minute limit currently. After that I'll merge, but @agramfort if you have comments I can make a follow-up PR

agramfort:
thx @larsoner !

PR #7905, opened 2020-06-17
WIP: Warn if untested NIRX device

### LINKED ISSUE #7913: [fNIRS] Misleading error message when the raw_data path does not exist
#### Describe the bug

When the path provided by the user does not exist (i.e. when the directory does not exist), the error message raised by the **RAWNirx** `__init__` function is 
```
RuntimeError: Expect one evt file, got 0
```

I think it should be something like 
```
RuntimeError: The path you specified does not exist
```
since the error is not about the existence of a file, but rather about the directory's

#### Steps to reproduce
Create a **RAWNirx** instance by reading the data of a non-existent directory :
```
fnirs_raw_dir = 'a/path/that/doesnt/exist'

raw_intensity = mne.io.read_raw_nirx(fnirs_raw_dir,
                                         verbose=True).load_data()
```

#### Expected results
The error message tells the user the path to the specified directory does not exist.

#### Actual results
The error message tells the user the specified directory lacks an `*.evt` file (since it is the first file extension that is checked) but not that the directory does not exist: the existence of the directory is not checked, which is misleading the user.

#### Additional information
Platform:        Windows-10-10.0.18362-SP0
Python:          3.8.3 (default, May 19 2020, 06:50:17) [MSC v.1916 64 bit (AMD64)]
Executable:     C:\Users\juliend\Anaconda3\envs\mne_dev\python.exe
CPU:              Intel64 Family 6 Model 58 Stepping 9, GenuineIntel: 8 cores
Memory:        15.6 GB

mne:              0.21.dev0
numpy:          1.18.1 {blas=mkl_rt, lapack=mkl_rt}
scipy:             1.4.1
matplotlib:     3.1.3 {backend=Qt5Agg}

sklearn:         0.22.1
numba:         0.49.1
nibabel:         3.1.0
cupy:             Not found
pandas:         1.0.3
dipy:             1.1.1
mayavi:        4.7.2.dev0
pyvista:        0.25.3
vtk:              9.0.0
PyQt5:         5.9.2

\-\-

Please note that adding
```
if not op.exists(fname):
    raise RuntimeError('The path you specified does not exist.')
```
or
```
if not op.isdir(fname):
    raise RuntimeError('The path you specified does not exist.')
```
at the start of `__init` function of **RAWNirx** does ~~not~~ [EDIT: see [this](https://github.com/mne-tools/mne-python/issues/7913#issuecomment-647976330) comment] lead to raising the error, if *fname* is a non-existent path; ~~there must be a "good" reason for that~~. I'm in to correct the behaviour if you think it should be fixed.

rob-luke:
@cbrnr confirmed in #7894 that the NIRX format for saving data is not consistent across devices. I have only added test data for the NIRScout device I have access too. The function now throws a warning if used with an untested device.

#### Additional information
Next step is to improve the reader function to work with these alternative formats. But this warning seems prudent until someone with access to these devices adds reader support.

Closes #7913

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -22,6 +22,8 @@ from ...utils import warn
 def read_raw_nirx(fname, preload=False, verbose=None):
     """Reader for a NIRX fNIRS recording.
 
+    This function has only been tested with NIRScout devices.
+
     Parameters
     ----------
     fname : str
@@ -70,6 +72,9 @@ class RawNIRX(BaseRaw):
         if fname.endswith('.hdr'):
             fname = op.dirname(op.abspath(fname))
 
+        if not op.isdir(fname):
+            raise RuntimeError('The path you specified does not exist.')
+
         # Check if required files exist and store names for later use
         files = dict()
         keys = ('evt', 'hdr', 'inf', 'set', 'tpl', 'wl1', 'wl2',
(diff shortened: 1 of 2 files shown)


cbrnr:
This warning will not be raised for NIRSport2 devices, which do not generate a .hdr file in the first place. But it will work for NIRSport1 data.

Maybe it is sufficient to include a note in the docstring that we have currently tested the function only for NIRScout devices, and that NIRSport and NIRSport2 are WIP.

cbrnr:
Also, maybe it would be a good idea to accept the .wl1 and/or .wl2 file as the file argument for the reader. These two files exists for all NIRX devices, so at least the refactored reader should change this from the current directory or the .hdr file.

cbrnr:
Here are the files recorded with NIRSport2:
[NIRSport2.zip](https://github.com/mne-tools/mne-python/files/4792163/NIRSport2.zip)
I'll upload NIRSport1 files soon (we need to record a short test session, which requires setting up the equipment).

rob-luke:
I suggest larger rewrites to support additional devices be kept for another PR, and this is just a quick fix to inform users that we currently only support NIRScout devices. I’m happy with either a doc change or a warning or both. @larsoner what do you prefer

rob-luke:
Also CI failures seem unrelated and the decrease in code coverage is because we dont have test data in the repo yet.

cbrnr:
And here are the files recorded with an old NIRSport device:

[2014-04-22_001.zip](https://github.com/mne-tools/mne-python/files/4792671/2014-04-22_001.zip)

larsoner:
@rob-luke want to take a quick look and see if it's easy to support these files? If it only takes a week or two to get a PR in that allows to read them then let's just do that.

rob-luke:
>@rob-luke want to take a quick look and see if it's easy to support these files? 

Sorry I don't have the capacity to implement new file readers now. If someone who wants to use NIRSport devices implements a reader I will be happy to review a PR.

> And here are the files recorded with an old NIRSport device:

Can we also get details of how the file was generated and the settings saved in the data? I think this is important to have to ensure we extract the correct values from the data, otherwise we are just assuming we interpret the file format correctly. Here is an example of the minimum information I think is required... https://github.com/mne-tools/mne-testing-data/pull/51

rob-luke:
Also integrates comments from #7913

larsoner:
CircleCI is unrelated. Let's get this in and we can relax the warning once someone gets a chance to work on the other reader. @cbrnr do you want to give it a shot at some point, incorporating @rob-luke's suggestion [above](https://github.com/mne-tools/mne-python/pull/7905#issuecomment-645626028) about what files are helpful to have?

cbrnr:
Yes, I'll see what I can do. It doesn't seem to be too complicated, but I need to finish some other tasks before I can start working on that. I'll create an issue to remind myself.

PR #7906, opened 2020-06-17
WIP: Fix Travis

larsoner:
Hopefully expose the failure sooner.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
             name: Install 3D rendering libraries \ PyQt5 dependencies \ graphviz \ optipng (for optimized images)
             command: |
               sudo apt-get install libosmesa6 libglx-mesa0 libopengl0 libglx0 libdbus-1-3 \
-                  libxkbcommon-x11-0 \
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ jobs:
             name: Install 3D rendering libraries \ PyQt5 dependencies \ graphviz \ optipng (for optimized images)
             command: |
               sudo apt-get install libosmesa6 libglx-mesa0 libopengl0 libglx0 libdbus-1-3 \
-                  libxkbcommon-x11-0 \
+                  libxkbcommon-x11-0 libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-shape0 libxcb-xfixes0 libxcb-xinerama0 \
```
larsoner, line 59: From https://doc.qt.io/qt-5/linux-requirements.html it looks like 5.15 was not compiled with `-qt-xcb` so now we need the system libs. I fixed this on CircleCI by actually installing each package individually, on Travis I just took the approach of `sudo apt-get install libxcb*` because it seemed cleaner than adding a bunch of newlines to enumerate the packages (shouldn't matter much either way we do it)

larsoner:
Okay let's see if this stabilizes things, if not we can pin PyQt5 again (but hopefully won't need to)

PR #7908, opened 2020-06-17
try to fix nullcontext

### LINKED ISSUE #7907: Python version dependency issue
#7758 introduced a dependency on `contextlib.nullcontext`, which was added to the standard library in Python 3.7.  We allegedly support Python 3.6 as our minimum version.  This came up in helping a user troubleshoot MNE on Google CoLab hosted notebooks, which run py 3.6

cc @GuillaumeFavelier

drammock:
closes #7907

### DIFF
--- a/mne/fixes.py
+++ b/mne/fixes.py
@@ -1078,3 +1078,16 @@ else:
     @jit()
     def mean(array, axis):
         return _np_apply_along_axis(np.mean, axis, array)
+
+
+###############################################################################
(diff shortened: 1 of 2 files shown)


drammock:
@larsoner I don't think these changes are actually getting tested; AFAICT our only py3.6 test run is on Travis, and it skips all the pyvista-related tests.  Feel free to look into it and push here; otherwise I'll try to dig deeper tomorrow.

larsoner:
Thanks @drammock

Issue #5745, opened 2018-11-23
TFCE on 3D spatial or 4D spatio-temporal data


ilogue:
Would `spatio_temporal_cluster_1samp_test` and the other clustering functions work on 3D spatial or 4D spatio-temporal data? I'm thinking to use it for data in MNI space. Dimensions would be approximately 256 * 65 * 77 * 53 (t * x * y * z). Would I have to provide it with a connectivity matrix or would it take into account a "3D lattice"?

What do you think @larsoner @ktavabi ? many thanks!

mmagnuski:
I don't know about volumetric data (surface- or channel- based representations have only one spatial dimension) but in general currently cluster based tests are limited to 2d in mne.

mmagnuski:
So if you have 3 spatial dimensions and one additional you would have to "unroll" the spatial dimensions and provide connectivity for them.

agramfort:
you can work with any connectivity provided it's regular grid or you
provide a sparse matrix to define the connectivity

ilogue:
alright I'll give that a go

larsoner:
Closing in favor of #4859

Issue #4859, opened 2018-01-09
Cluster-based stats improvements - potential GSoC (and future)


mmagnuski:
Currently I'm copying everything from the e-mail thread, but I'll find some time this week to organize this better and propose something more concrete.

Eric @larsoner suggested to mirror fieldtrip's cluster-based API in mne python:

> Mikolaj, perhaps instead of making a new API proposal, you could try Python-izing the FieldTrip API. No need to reinvent the wheel here if theirs is complete / extensible enough already, and it would have the added benefit of making user code porting simpler.

Sure, that makes sense, although I would prefer to change a few things, notably the design, ivar and uvar which are quite confusing for newcomers.
I can start with showing how an almost exact copy of fieldtrips API would look
(I omit method='montecarlo', because we are considering now mirroring only the cluster-based API):
```python
# within-subjects t-test
from mne.stats import ttest_rel_no_p, cluster_based_statistic

conditions = ['condA', 'condB']

erp_list = list()
for epochs in epochs_list:
    erp_list += [epochs[cond].average() for cond in conditions]

# in fieldtrip, because all the files/trials are passed
# in a cell array (or just varargin) you need additional info
# sepcifying the design:
conditions = np.array([0, 1] * n_subjects)
subjects = np.array([[i, i] for i in range(n_subjects)]).ravel()
design = np.stack([subjects, conditions], axis=0)

# later in the function uvar denotes the row containing 'units of observation',
# here a single subject identifiers, and ivar denotes levels of independent
# variable

# additonally we have adjacency structure in adjacency var

cluster_based_statistic(erp_list, statistic=ttest_rel_no_p, alpha=0.025,
                        cluster_alpha=0.05, cluster_statistic='maxsum',
                        min_nb_chan=2, neighbours=adjacency, tail=0,
                        cluster_tail=0, n_permutations=1000, design=design,
                        uvar=0, ivar=1)
# the alpha is set to 0.025 above to correct for two-tail, fieldtrip also
# allows to set another parameter to have two-tail cluster p values returned
```
I like the Fieldtrip’s cluster-based API in general but not the `design`,
`ivar`, `uvar` approach. This is something that would require some further
thought. Passing a real design matrix might be ok to add some time later for more advanced users, but
the basic API should be much simpler. A sketchy idea I have:
```python
erp_list = {cond: [epochs[cond].average() for epochs in epochs_list]
            for cond in cond_list}

[...]

PR #7916, opened 2020-06-22
MRG, ENH: Add arbitrary connectivity for stats

### LINKED ISSUE #7915: Cluster-based Permutation test for Time-Frequency Data - 3D clusters
Hello MNE-team,

Eric Larson has asked me to post this as an issue here.

I've been trying to run cluster-based permutation tests on time-frequency data considering all channels and different frequency ranges across time. 
Colleagues of mine have informed me that Fieldtrip in MATLAB provides an algorithm that allows to search for adjacent frequencies, timepoints and channels that show similar modulations, which then result in 3D clusters. In the examples on statistical analysis of time-frequency data provided on the MNE website one channel only is selected. Is there an option to run a similar analysis that runs over all channels simultaneously and produces 3D clusters? If so is there an example that can be found somewhere and are there suggestions for plotting these outputs? Thank you very much for your help!

Best,

Sebastian

larsoner:
Arguably part of #4859, originally brought up in #5745, resurfaced on the mailing list today.

Closes #7915

@mmagnuski you might have asked about this sort of thing before, interested in reviewing and/or trying?

In theory at some point we could probably exploit a lattice structure to avoid a lot of unnecessary checks (just like we do in the `*_temporal` variants of our stats functions), but this is at least a good start that should allow people to do these sorts of analyses even if they are slower than they could be.

### DIFF
--- a/mne/stats/tests/test_adjacency.py
+++ b/mne/stats/tests/test_adjacency.py
@@ -41,4 +41,4 @@ def test_adjacency_equiv(shape):
     # for now
     assert np.in1d(conn, [0, 1, 2, 3]).all()
     assert conn.shape == conn_sk.shape
-    assert_array_equal(conn, conn)
+    assert_array_equal(conn, conn_sk)
--- a/mne/stats/tests/test_adjacency.py
+++ b/mne/stats/tests/test_adjacency.py
@@ -8,7 +8,7 @@ import pytest
 import numpy as np
 from numpy.testing import assert_array_equal
 
-from mne.stats import combine_connectivity
+from mne.stats import combine_adjacency
 from mne.utils import requires_sklearn
 
 
@@ -24,13 +24,13 @@ from mne.utils import requires_sklearn
     (1, 1, 2),
     (3, 4, 5),
 ])
-def test_connectivity_equiv(shape):
-    """Test connectivity equivalence for lattice connectivity."""
+def test_adjacency_equiv(shape):
+    """Test adjacency equivalence for lattice adjacency."""
     from sklearn.feature_extraction import grid_to_graph
     # sklearn requires at least two dimensions
     sk_shape = shape if len(shape) > 1 else (shape + (1,))
     conn_sk = grid_to_graph(*sk_shape).toarray()
-    conn = combine_connectivity(*shape)
+    conn = combine_adjacency(*shape)
     want_shape = (np.prod(shape),) * 2
     assert conn.shape == conn_sk.shape == want_shape
     assert (conn.data == 1.).all()
--- a/mne/stats/tests/test_connectivity.py
+++ b/mne/stats/tests/test_connectivity.py
@@ -13,7 +13,7 @@ from mne.utils import requires_sklearn
 
 
 @requires_sklearn
-@pytest.mark.parametrize('ns', [
+@pytest.mark.parametrize('shape', [
     (1,),
     (2,),
     (1, 1),
@@ -24,13 +24,14 @@ from mne.utils import requires_sklearn
     (1, 1, 2),
     (3, 4, 5),
 ])
-def test_connectivity_equiv(ns):
+def test_connectivity_equiv(shape):
     """Test connectivity equivalence for lattice connectivity."""
     from sklearn.feature_extraction import grid_to_graph
-    sk_ns = ns if len(ns) > 1 else (ns + (1,))
-    conn_sk = grid_to_graph(*sk_ns).toarray()
-    conn = combine_connectivity(*ns)
-    want_shape = (np.prod(ns),) * 2
+    # sklearn requires at least two dimensions
+    sk_shape = shape if len(shape) > 1 else (shape + (1,))
+    conn_sk = grid_to_graph(*sk_shape).toarray()
+    conn = combine_connectivity(*shape)
+    want_shape = (np.prod(shape),) * 2
     assert conn.shape == conn_sk.shape == want_shape
     assert (conn.data == 1.).all()
     conn = conn.toarray()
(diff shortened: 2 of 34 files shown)


tutorials/stats-sensor-space/plot_stats_cluster_1samp_test_time_frequency.py
```diff
@@ -106,19 +130,24 @@
     if p_val <= 0.05:
         T_obs_plot[c] = T_obs[c]
 
+# Just plot one channel's data
+ch_idx, f_idx, t_idx = np.unravel_index(
+    np.nanargmax(np.abs(T_obs_plot)), epochs_power.shape[1:])
+# ch_idx = tfr_epochs.ch_names.index('MEG 1332')  # to show a specific one
```
drammock, line 136: cruft? or intentional?

mne/stats/tests/test_connectivity.py
```diff
@@ -0,0 +1,43 @@
...
+    (3, 4),
+    (1, 1, 1),
+    (1, 1, 2),
+    (3, 4, 5),
+])
+def test_connectivity_equiv(ns):
+    """Test connectivity equivalence for lattice connectivity."""
+    from sklearn.feature_extraction import grid_to_graph
+    sk_ns = ns if len(ns) > 1 else (ns + (1,))
+    conn_sk = grid_to_graph(*sk_ns).toarray()
```
drammock, line 31: for future maintainability, could we have slightly more transparent variable names for `ns`, `sk_ns`, and `conn_sk`?

tutorials/stats-sensor-space/plot_stats_cluster_1samp_test_time_frequency.py
larsoner: Intentional in case people want to play with it in the example (it used to only show that sensor)

larsoner:
Here is the rendered example:

https://20729-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/stats-sensor-space/plot_stats_cluster_1samp_test_time_frequency.html#sphx-glr-auto-tutorials-stats-sensor-space-plot-stats-cluster-1samp-test-time-frequency-py

mmagnuski:
@larsoner Great, I'll take a look within 24h!

tutorials/stats-sensor-space/plot_stats_cluster_1samp_test_time_frequency.py
```diff
@@ -106,19 +130,24 @@
     if p_val <= 0.05:
         T_obs_plot[c] = T_obs[c]
 
+# Just plot one channel's data
```
mmagnuski, line 133: I'd write that we find indices of the maximum absolute t value

tutorials/stats-sensor-space/plot_stats_cluster_1samp_test_time_frequency.py
```diff
@@ -75,20 +72,47 @@
...
-evoked.crop(0., 0.4)
-tfr_epochs.crop(0., 0.4)
+evoked.crop(-0.1, 0.4)
+tfr_epochs.crop(-0.1, 0.4)
 
-epochs_power = tfr_epochs.data[:, 0, :, :]  # take the 1 channel
+epochs_power = tfr_epochs.data
+
+###############################################################################
+# Define connectivity for statistics
```
mmagnuski, line 81: I must say that the naming for adjacency is a bit unfortunate in mne. I was a few times in a situation when I had to search online for mne function that finds adjacency. Because it has connectivity in its name finding it was a bit difficult: I was getting lots of results related to actual connectivity (coherence, phase slope index etc.) but not the function I needed.

larsoner:
We could rename to adjacency in another PR. I agree it's bad naming to have this collide with functional and effective connectivity, I almost named this combine_adjacency but the deprecation bothered me.

@agramfort okay to rename via deprecation spatio_src_connectivity and related functions to spatial_src_adjacency, and clustering arguments to adjacency instead of connectivity?

agramfort:
how is it called in other packages?

larsoner:
Based on a minimal-effort searching (others feel free to chime in if you know of others or better for these packages)

- FieldTrip appears to use [neighbours](http://www.fieldtriptoolbox.org/tutorial/cluster_permutation_timelock/) but in the narrative doc also mentions "connected" subsets and "adjacency"
- Brainstorm uses fieldtrip (?) but only uses the term "adjacency" in the couple of places I looked
- [Maris and Oostenveld](https://www.sciencedirect.com/science/article/pii/S0165027007001707):

    > ... Cluster the selected samples in connected sets on the basis of temporal adjacency... instead of clustering selected (sensor, time)-samples in connected sets on the basis of spatial and temporal adjacency, we now cluster the selected (frequency, time)-samples on the basis of spectral and temporal adjacency

To me "adjacency" is the winner based on this

agramfort:
ok then

>

larsoner:
Okay, deprecations complete for `connectivity`->`adjacency`. Instead of +182/−15 you now get +714/−468 @agramfort :)

mne/stats/tests/test_adjacency.py
```diff
@@ -0,0 +1,44 @@
...
+    assert conn.shape == conn_sk.shape == want_shape
+    assert (conn.data == 1.).all()
+    conn = conn.toarray()
+    # we end up with some duplicates that can turn into 2's and 3's,
+    # eventually we might want to keep these as 1's but it's easy enough
+    # with a .astype(bool) (also matches sklearn output) so let's leave it
+    # for now
+    assert np.in1d(conn, [0, 1, 2, 3]).all()
+    assert conn.shape == conn_sk.shape
+    assert_array_equal(conn, conn)
```
mmagnuski, line 44: both arrays are the same object here

mmagnuski:
Looks good! And I'm glad it's `adjacency` now! :) (just one small comment regarding one of the tests)

larsoner:
Good catch @mmagnuski, that errant naming was missing an issue with the diagonals being repeated such that they were no longer unity values (they were equal to `len(structure)` instead of always being 1.).

PR #7923, opened 2020-06-23
MRG, API: Use new fNIRS naming

larsoner:
@rob-luke can you take over? So far I've just added the new constants and provided an alias in the `FIFF` constant bunch.

I guess the next step is to rename the `'fnirs_raw'` everywhere to `'fnirs_cw_amplitude'`, and maybe add some aliases for backward compatibility (?).


rob-luke:
Thanks for kicking this off @larsoner.

I used this branch today while analysing different datasets. It is working for me. I dont think we will need any aliases for backwards compatibility because I never operate on the raw data, I always convert to optical density straight away, and I believe this is quite common.

So I think this is ready for review.

larsoner:
@kylemath can you look first?

cbrnr:
@rob-luke I think you use this data type in MNELAB, so if you are not going to add an alias we will need to introduce a change which will require a future version of MNE.

rob-luke:
> @rob-luke I think you use this data type in MNELAB, so if you are not going to add an alias we will need to introduce a change which will require a future version of MNE.

Sure, I can make the change in the 0.21 branch

larsoner:
@cbrnr okay to merge this, or do you think we need to alias `fnirs_raw` in some places like `pick_types(info, fnirs='fnirs_raw')`? I don't think it would be too onerous to support these aliases if it's for a few public functions.

cbrnr:
@larsoner yes, aliasing `fnirs_raw` and `fnirs_od` in `pick_types` would be great, because that's the only places it is currently used in MNELAB. @rob-luke can you add that?

larsoner:
Okay, added aliases that make `'fnirs_raw' in inst` and `pick_types(inst.info, fnirs='fnirs_raw')` work, with deprecation warnings that they will go away in 0.22.

larsoner:
Travis is happy must just be a status update issue, ready to go @agramfort

rob-luke:
> @rob-luke can you add that?

Sorry was too slow on this. Thanks @larsoner for jumping on it and keeping things moving.

mne/io/pick.py
```diff
@@ -786,6 +791,17 @@ def _mag_grad_dependent(info):
                for ph in info.get('proc_history', []))
 
 
+def _fnirs_raw_dep(ch_type, warned):
+    if ch_type == 'fnirs_raw':  # alias
+        if not warned[0]:
+            warn('"fnirs_raw" has been deprecated in favor of the more '
+                 'explicit "fnirs_cw_amplitude" and will be removed in 0.22',
+                 DeprecationWarning)
```
rob-luke, line 799: Just a question, how do you go about ensuring these things are removed as specified by the depreciation warnings? Do you just do a search for `DeprecationWarning` after releasing 0.21?

larsoner: Yes, it's my favorite part of the [release process (step 14)](https://github.com/mne-tools/mne-python/wiki/How-to-make-a-release)...

rob-luke: Thanks for the link. I'm just doing some prethinking for when we eventually push MNE-NIRS out (if we do).

cbrnr: @rob-luke what do you mean by MNE-NIRS?

rob-luke: https://mne.tools/mne-nirs/

rob-luke: Its where I do the experimental stuff, get everything working, tweak the API, then move to MNE-Python

cbrnr: Got it - I'm glad all of the NIRS functionality will eventually land in MNE-Python.

cbrnr:
Thanks @larsoner!

kylemath:
Thanks everyone, we tested this yesterday and it was working for us

mne/io/pick.py
rob-luke: Thats the plan. Definitely the MNE epoching style stuff goes straight in here. But the different style analysis (more MRI than EEG) will start of in MNE-NIRS.

PR #7928, opened 2020-06-25
MRG: Add smoothing controller to TimeViewer for the notebook backend

GuillaumeFavelier:
This PR adds a controller to the `notebook` backend to set the number of smoothing steps when `time_viewer=True`.

**Issue**

The rendering cell is broken because of the text given by `logger` I think (the name of the backend and the number of iterations for smoothing).

image

image

It's part of #7162.

### DIFF
--- a/mne/tests/test_cov.py
+++ b/mne/tests/test_cov.py
@@ -147,7 +147,7 @@ def test_cov_order():
     whitener_2, w_ch_names_2, n_nzero_2 = compute_whitener(
         cov_reorder, info, return_rank=True)
     assert_array_equal(w_ch_names_2, w_ch_names)
-    assert_allclose(whitener_2, whitener)
+    assert_allclose(whitener_2, whitener, rtol=1e-6)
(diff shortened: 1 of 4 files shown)


larsoner:
I would just wrap [this call](https://github.com/mne-tools/mne-python/blob/master/mne/viz/_brain/_brain.py#L1009) in `with mne.use_log_level(False):` to always suppress these messages. They aren't really useful now that our morphing is fairly fast

GuillaumeFavelier:
Your advice helped, now the messages about the number of iterations have stopped popping up.
If the first element of the rendered cell is not text, it works as expected, otherwise the cell is not flat and sliding is needed. This behaviour is totally normal.

GuillaumeFavelier:
I can cherry-pick 1355296c69e8b54ccbeacd12deb3770296d85241 here if needed?

larsoner:
Go ahead and I'll merge so it gets into master

larsoner:
Thanks @GuillaumeFavelier

PR #7929, opened 2020-06-25
MRG, ENH: Automatically fix magnetometers when maxwell filtering

larsoner:
As suggested by Jukka

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1357,13 +1357,16 @@ def _compute_ch_connectivity(info, ch_type):
     return ch_connectivity, ch_names
 
 
-def fix_mag_coil_types(info):
+def fix_mag_coil_types(info, use_cal=False):
     """Fix magnetometer coil types.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


mne/channels/channels.py
```diff
@@ -1385,24 +1388,31 @@ def fix_mag_coil_types(info):
...
     for ii in picks:
         ch = info['chs'][ii]
         if ch['coil_type'] in (FIFF.FIFFV_COIL_VV_MAG_T1,
                                FIFF.FIFFV_COIL_VV_MAG_T2):
-            old_mag_inds.append(ii)
+            if use_cal:
+                if ch['cal'] > 3e-11:
+                    old_mag_inds.append(ii)
+            else:
+                old_mag_inds.append(ii)
```
larsoner, line 1415: @mshamalainen clear enough?

mshamalainen: This should do it. I suppose the processed files will inherit the correct mag coil types.

PR #7934, opened 2020-06-25
MRG, BUG: Fix vector scaling

larsoner:
Fixes the bug with glyph colormap not being updated, tested with:

<details>

```
import os.path as op
import mne

initial_time = 0.1

size = (800, 400)
                     size=size, background='k')
```

</details>

A few things along the way:

1. Consolidate `update_lut` so that there is no longer a need for the very repeat-y `update_fmin`, `update_fmid`, and `update_fmax`
2. Refactor some time viewer slider controls to make one call to `update_lut` instead of potentially multiple calls like `update_fmin` followed by `update_fmid`
3. Make `update_fscale` private (I don't think it was meant to be public)
4. Never apply `alpha` to the LUT, but instead apply it to the `mesh(..., opacity=alpha)` instead.

Approximately 3x minus-to-plus ratio on this PR suggesting that it's DRYing out our code. No test added because I'm not sure that there is a great way to test this. In theory if there were existing vector + TimeViewer tests it could be added there, but in a quick check I did not see any. @GuillaumeFavelier feel free to correct me if I'm wrong...

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -142,6 +142,8 @@ Bug
 
 - Fix plotting with :func:`mne.viz.plot_evoked_topomap` to pre-existing axes by `Daniel McCloy`_
 
+- Fix bug with :func:`mne.viz.plot_vector_source_estimates` using the PyVista backend with ``time_viewer=True`` when updating the arrow colormaps by `Eric Larson`_
+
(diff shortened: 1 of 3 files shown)


GuillaumeFavelier:
> here were existing vector + TimeViewer tests it could be added there, but in a quick check I did not see any

Its test is `test_plot_vector_source_estimates` in `test_3d.py`. The `renderer_interactive` makes it available for the PyVista backend with `time_viewer=True` and `show_traces='auto'` (was not the case before).

agramfort:
thx @larsoner

PR #7935, opened 2020-06-26
MRG, VIZ, BUG: handle CSD channel type when topo plotting

### LINKED ISSUE #7933: VIZ, BUG: plot_compare_evokeds(..., axes='topo') fails for CSD channel types
probably related to mne.channels.layout.generate_2d_layout.  The plot succeeds, but all the axes are squished into the lower left corner and the figure is mostly empty space.  To reproduce:

```python
import mne

raw = mne.io.read_raw_fif(data_path + '/MEG/sample/sample_audvis_raw.fif',
                          preload=True)
raw.pick_types(meg=False, eeg=True, eog=True, ecg=True, stim=True, exclude='bads')
events = mne.find_events(raw)
raw.set_eeg_reference(projection=True).apply_proj()

            'visual/right': 4, 'smiley': 5, 'button': 32}
epochs = mne.Epochs(raw, events, event_id=event_id, tmin=-0.2, tmax=0.5,
                    preload=True)
evoked = epochs['auditory'].average()

evoked_csd = mne.preprocessing.compute_current_source_density(evoked)

mne.viz.plot_compare_evokeds(evoked_csd, picks='csd', axes='topo')
```

drammock:
closes #7933 

The approach in 1b9ff53 works, in that it yields a correct topo layout for the CSD-transformed EEG data. However, it breaks 3 tests in `mne/viz/tests/test_topo.py` so it's probably not an approach worth keeping.  Other possible approaches:

- add a new return value (`has_eeg_csd_coils`?) to `mne.channels.channels._get_ch_info` rather than hacking it to return `True` for `has_eeg_coils` when only CSD channels are present.
- in `mne.channels.layout.make_eeg_layout`, use a boolean `allow_csd` parameter instead of a `ch_types=['eeg', 'csd']` parameter; and make `eeg=True` always the case in the call to `pick_types` therein.

I'm not very familiar with this part of the codebase, so any suggestions are welcome.

### DIFF
--- a/mne/viz/tests/test_evoked.py
+++ b/mne/viz/tests/test_evoked.py
@@ -326,8 +326,8 @@ def test_plot_compare_evokeds():
     # test with (fake) CSD data
     csd = _get_epochs(picks=np.arange(315, 320)).average()  # 5 EEG chs
     for entry in csd.info['chs']:
-        entry['coil_type'] = 6
-        entry['unit'] = 117
+        entry['coil_type'] = FIFF.FIFFV_COIL_EEG_CSD
+        entry['unit'] = FIFF.FIFF_UNIT_V_M2
     plot_compare_evokeds(csd, picks='csd', axes='topo')
     # old tests
     red.info['chs'][0]['loc'][:2] = 0  # test plotting channel at zero
--- a/mne/viz/tests/test_evoked.py
+++ b/mne/viz/tests/test_evoked.py
@@ -42,7 +42,7 @@ picks = [0, 1, 2, 3, 4, 6, 7, 61, 122, 183, 244, 305]
 sel = [0, 7]
 
 
-def _get_epochs():
+def _get_epochs(picks=picks):
     """Get epochs."""
     raw = read_raw_fif(raw_fname)
     raw.add_proj([], remove_existing=True)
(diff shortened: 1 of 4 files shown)


larsoner:
> add a new return value (has_eeg_csd_coils?) to mne.channels.channels._get_ch_info rather than hacking it to return True for has_eeg_coils when only CSD channels are present.

Seems like it's safest just to add another return value `has_csd_coils` to `_get_ch_info`.

drammock:
All green; this one's ready for review/merge.

mne/viz/tests/test_evoked.py
```diff
@@ -323,6 +323,12 @@ def test_plot_compare_evokeds():
     figs = plot_compare_evokeds(evoked_dict, axes='topo', legend=True)
     for fig in figs:
         assert len(fig.axes[0].lines) == len(evoked_dict)
+    # test with (fake) CSD data
+    csd = _get_epochs(picks=np.arange(315, 320)).average()  # 5 EEG chs
+    for entry in csd.info['chs']:
+        entry['coil_type'] = 6
+        entry['unit'] = 117
```
agramfort, line 330: can you use more explicit FIFF constants here?

drammock: > can you use more explicit FIFF constants here?

sure, will do.

larsoner:
Thanks @drammock

Issue #7926, opened 2020-06-25
[fNIRS] use .nosatflags_wlX in case they exist


Swy7ch:
#### Describe the bug
The NIRStar software sometimes puts NaN values in the `.wlX` files when it estimates the probes saturated. It is not clear how (or even why) it does so, but it changes the data without the approval of the researcher. The software does provide the "true" values in a separate file name `.nosatflags_wlX`, which is a copy of the corresponding `.wlX` file with the true value.

Currently, the library checks for any file in the working directory that ends with 'wl1'. When the `nosatflags_wlX` files exist, the `glob.glob()` function returns a list of two paths, one for `.wlX` and another for `.nosatflags_wlX`, and then raise an error saying : 

    'Expect one wl1 file, got 2'

#### Steps to reproduce
* Get a "saturated" sample of data
* Read the data from said sample

#### Expected results
Warn the user that we don't use the `.wlX` file but the corresponding `.nosatflags_wlX` one instead, and use it.

#### Actual results
An error is raised saying there is too much `.wlX` files.

#### Additional information

Platform:      Windows-10-10.0.18362-SP0
Python:        3.8.3 (default, May 19 2020, 06:50:17) [MSC v.1916 64 bit (AMD64)]
Executable:    C:\Users\juliend\Anaconda3\envs\mne_dev\python.exe
CPU:           Intel64 Family 6 Model 58 Stepping 9, GenuineIntel: 8 cores
Memory:        15.6 GB

mne:           0.21.dev0
numpy:         1.18.1 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.4.1
matplotlib:    3.1.3 {backend=Qt5Agg}

sklearn:       0.22.1
numba:         0.49.1
nibabel:       3.1.0
cupy:          Not found
pandas:        1.0.3
dipy:          1.1.1
mayavi:        4.7.2.dev0
pyvista:       0.25.3
vtk:           9.0.0
PyQt5:         5.9.2

\-\-

I have a patch that warns the user of the switch in used files, and also use the right one, that you can find [here](https://github.com/rderollepot/mne-python/commit/fbbe655d54388da84e8844075e66f93529c0a875) (I will PR it, just wanted to open an issue before as I didn't know the right worflow)

larsoner:
Thoughts @rob-luke ?

rob-luke:
[...]

PR #7936, opened 2020-06-26
WIP: Add NIRSport support

Swy7ch:
#### Reference issue
 #7926, in [this](https://github.com/mne-tools/mne-python/issues/7926#issuecomment-650082347) comment.

Add NIRSport (v1) systems data processing in MNE:
* saturation handling
* NIRSport-specific reader ?
* more ?

#### Additional information
The two versions of NIRSport (1 & 2) do not provide data the same way, so v2 support will be handled later. 

As said [here](https://github.com/mne-tools/mne-python/issues/7926#issuecomment-650062486) by @cbrnr, NIRSport v1 and NIRScout data seem similar enough to use the same reader. We can of course debate this, and I'd make another one if needed.


Swy7ch:
Wow, that's a lot of errors for such a small patch.

Regarding `circleci` it looks like I have to link my repo with them. If it's the case I'll have to contact my tutor, as he's in charge of the repo.

Don't know about the code coverage though.

drammock:
It's actually not that bad:

Style errors:
```
mne/io/nirx/tests/test_nirx.py:58:1: E302 expected 2 blank lines, found 1
mne/io/nirx/tests/test_nirx.py:64:33: W503 line break before binary operator
mne/io/nirx/tests/test_nirx.py:71:1: E302 expected 2 blank lines, found 1
mne/io/nirx/tests/test_nirx.py:60 in public function `test_nirx_nosatflags_v1_warn`:
        D400: First line should end with a period (not 'a')
```

Typos:
```
mne\io\nirx\tests\test_nirx.py:66: in test_nirx_nosatflags_v1_warn
    with pytest.raised(RuntimeWarning, match='You provided saturated'):
E   AttributeError: module 'pytest' has no attribute 'raised'
```
(should be `pytest.raises`, lines 65 and 67)

rob-luke:
The third option may be to return the saturated data and an index/mask to which data is invalid. Not sure how this would actually be implemented. But just throwing all the options out there.

drammock:
> The third option may be to return the saturated data and an index/mask to which data is invalid. Not sure how this would actually be implemented.

`numpy.ma.masked_array` is a natural choice for the data container in that case, but it could take some real work to make sure that actually works as expected with the rest of the instance's API.

rob-luke:
> `numpy.ma.masked_array` is a natural choice for the data container in that case, but it could take some real work to make sure that actually works as expected with the rest of the instance's API.

Oh very interesting! I didn't know about this. But yeah sounds like lots of work. And I guess we can't expand `info` to have a mask. 

But I think its worth having these API discussions now before @Swy7ch puts too much work in. We need to decide which approach to take (return NaNs, return invalid data, or return data+mask), then what the defaults should be. @larsoner any suggestions here?

> Don't know about the code coverage though.

Once we get your test sets in this should fix itself.

mne/io/nirx/nirx.py
```diff
@@ -19,7 +19,7 @@
 
 
 @fill_doc
-def read_raw_nirx(fname, preload=False, verbose=None):
+def read_raw_nirx(fname, nosatflags=False, preload=False, verbose=None):
     """Reader for a NIRX fNIRS recording.
 
     This function has only been tested with NIRScout devices.
```
rob-luke, line 25:
```suggestion
    This function has only been tested with NIRScout and NIRSport1 devices.
```

mne/io/nirx/nirx.py
```diff
@@ -28,6 +28,8 @@ def read_raw_nirx(fname, preload=False, verbose=None):
     ----------
     fname : str
         Path to the NIRX data folder or header file.
+    nosatflags : bool
+        Decide to use *.nosatflags_wlX instead of standard *.wlX files
```
rob-luke, line 32: Can you expand this to describe what causes these files to be created (saturated values) and what the effect of the flag has (NaNs vs Saturated data). And mention that it is only relevant for NIRSport devices.

No need to be overly brief here, better to be longer and clearer.

mne/io/nirx/nirx.py
```diff
@@ -55,6 +57,8 @@ class RawNIRX(BaseRaw):
...
```
rob-luke, line 61: As above

mne/io/nirx/tests/test_nirx.py
```diff
@@ -54,6 +54,18 @@ def test_nirx_dat_warn(tmpdir):
     with pytest.raises(RuntimeWarning, match='A single dat'):
         read_raw_nirx(fname, preload=True)
 
+@requires_testing_data
+def test_nirx_nosatflags_v1_warn(tmpdir):
```
rob-luke, line 58: Great. Lets update with your real data once collected.

rob-luke:
https://github.com/mne-tools/mne-python/blob/965f2e3dc26c4e05bfb4260a54b25af0151dd5f2/mne/io/nirx/nirx.py#L133-L136

You can also add NIRSport1 to this somehow (depends how encoded in header file, hopefully the headers for NIRSport1 are similar format).

Swy7ch:
>You can also add NIRSport1 to this somehow (depends how encoded in header file, hopefully the headers for NIRSport1 are similar format).

Looks like the general header structure is the same, only some variables differ. For instance, in [GeneralInfo], 'APD' (for a NIRScout dataset) are replaced with 'Dual-Tip' (for a NIRSport one). Don't know if they are the same thing with a differnt name, but they're booth booleans.

>Style errors:

@drammock Fixed it, except for 

    mne/io/nirx/tests/test_nirx.py:64:33: W503 line break before binary operator

Not quite used to write Python (C dev here), and PEP8 [recommends](https://pep8.org/#break-before-or-after-binary-operator) to break *before* a binary operator, for readability purpose. Since the choice to break before or after binary operator is a project convention, which path did you chose to follow ? 

Force-pushing in the meantime to review other changes

larsoner:
We still abide by the old standard (we had a lot of code before they changed it in 2016), see our custom config here

https://github.com/mne-tools/mne-python/blob/master/setup.cfg#L19

I'd be okay with removing this requirement altogether. We can't just switch easily -- `flake8 --count` says that removing `W504` from our ignore list means we'd need to change 575 lines all at once, which is rough...

Swy7ch:
No problem, just wanted to stick with PEP8 recommendations before knowing what your were :). I've fixed it.

mne/io/nirx/nirx.py
```diff
@@ -19,15 +19,25 @@
...
 
-    This function has only been tested with NIRScout devices.
+    This function has only been tested with NIRScout and NIRSport1 devices.
 
     Parameters
     ----------
     fname : str
         Path to the NIRX data folder or header file.
+    nosatflags : bool
+        (Only relevant for NIRSport devices). Decide to use *.nosatflags_wlX
```
rob-luke, line 32: Is this also relevant for NIRSport2 devices? If not, lets be specific and say NIRSport1.

mne/io/nirx/nirx.py
```diff
@@ -19,15 +19,25 @@
...
+        instead of standard *.wlX files.
+        The NIRSport probes can saturate during the experiment. Starting with
+        NIRStar 14.2, those saturated values are replace with 'NaN' in the
+        standard *.wlX files. The probed values are stored in another file
```
rob-luke, line 36:
```suggestion
        standard *.wlX files. The measured values are stored in another file
```

rob-luke: The term probe is commonly used in NIRS literature to mean the location of optodes (montage).

mne/io/nirx/nirx.py
```diff
@@ -19,15 +19,25 @@
...
     """Reader for a NIRX fNIRS recording.
 
-    This function has only been tested with NIRScout devices.
+    This function has only been tested with NIRScout and NIRSport1 devices.
 
     Parameters
     ----------
     fname : str
         Path to the NIRX data folder or header file.
+    nosatflags : bool
```
cbrnr, line 31: Can we try to find a better parameter name? Something without any abbreviations, such as `saturated=False` or `flag_saturated=False` would be much better IMO.

mne/io/nirx/nirx.py
Swy7ch: The documentation only talks about the v1, so I guess not. I'll be more specific

mne/io/nirx/nirx.py
Swy7ch: I figured I could use the name of the file, but your request does make sense. Maybe `use_saturated` ? To be more specific than just `saturated`. @rob-luke, any idea ?

rob-luke: I think we need to establish the preferred behaviour of the function first, then answer these detail questions. But my vote is for `return_saturated` or `use_saturated`.

cbrnr: IMO `flag_saturated` is clearer, because it implies that saturated values are flagged somehow (with NaNs in this case). Even if we don't flag/mask saturated values, we will always return saturated values (either masked/flagged or not), so `use_saturated` or `return_saturated` does not tell me what they do. So I'd vote for either `flag_saturated` or `mask_saturated`, both of which return NaNs for saturated values if `True` and the raw values if `False`.

rob-luke: Lets go with `mask_saturated` then.

Swy7ch: Just to be clear : `mask_saturated` is set to **True** when we want `NaN` values, right ?

larsoner: What does `False` do? Maybe it's clearest just to name the behaviors like `saturated='nan' | 'whatevertheotherbehavioris'`.

Swy7ch: Currently with `nosatflags`, `True` means we use the *.nosatflags_wlX files, whereas `False` means we use the standard *.wlX files (which contains 'NaN' values)

cbrnr: `mask_saturated=True` masks saturated values with NaNs, `mask_saturated=False` doesn't mask saturated values.

cbrnr: If you want `saturated='nan'`, the other alternative should be "do nothing", so maybe `saturated=None`? Not sure if this is more intuitive though.

larsoner: `saturated='ignore'` would work, too, but I'm also okay with `mask_saturated` as a bool.

cbrnr: Yes, I like that! Makes for a shorter name too.

drammock: If there will only ever be 2 options, please keep as bool. `saturated='nan'` or `'ignore'` leads to internal code that is uglier and more annoying to maintain IMO, and the API is just as clear I think.

larsoner: I'm thinking exactly of leaving the possibility open for future behaviors like `saturated='annotate'`, which could for example annotate as `BAD` segments of saturation like `mne.preprocessing.mark_flat` does...

drammock: OK,that makes sense. I was envisioning the simple "do x" vs "don't do x" case, where I think Boolean has obvious advantages.

rob-luke: > If there will only ever be 2 options ...

I think this is the more important thing to be discussed first. What should be returned by default, and what should the other options be. Its unclear to me what the correct behaviour should be. And I still like the idea of returning a mask as well as the data somehow, but maybe this is too complicated as discussed earlier.

cbrnr: I thought that this was already settled. `saturated='annotate'` returns signals with NaNs, whereas `saturated='ignore'` returns the actual values.

cbrnr: And by default I would use `saturated='ignore'` (so no NaNs).

larsoner: My suggestion was saturated='nan' populates with Nan, saturated='annotate' would actually add raw annotations like mark_flat (but we can add this later, just Nan and ignore to start)

mne/io/nirx/nirx.py
```diff
@@ -19,15 +19,28 @@
 
 
 @fill_doc
-def read_raw_nirx(fname, preload=False, verbose=None):
+def read_raw_nirx(fname, saturated='ignore', preload=False, verbose=None):
     """Reader for a NIRX fNIRS recording.
 
-    This function has only been tested with NIRScout devices.
+    This function has only been tested with NIRScout and NIRSport1 devices.
```
cbrnr, line 25: I'd move this in a separate Notes section (see e.g. https://github.com/mne-tools/mne-python/blob/master/mne/io/edf/edf.py#L70).

mne/io/nirx/nirx.py
```diff
@@ -19,15 +19,28 @@
...
 
     Parameters
     ----------
     fname : str
         Path to the NIRX data folder or header file.
+    saturated : str
+        (Only relevant for NIRSport1 devices). If ``ignore`` (default),
```
cbrnr, line 32: I think we usually include the quotes, i.e. ``'ignore'`` (and ``'mask'``).

mne/io/nirx/nirx.py
rob-luke: I like this last suggestion by larsoner. Personally I would imagine myself using the annotate option, this way I could use all filtering functions without caring about NaNs. But then easily reject epochs or data segments where I know bad data lives. Are there any negatives to this approach? Also saturated=nan is intuitive to what is going to happen to the saturated data. We could also keep the saturated=ignore option which returns the data but doesn't add annotations if that's what people want.

cbrnr: Filtering should work in the presence of NaNs (at least `scipy.signal.lfilter` has no problems with NaNs).

mne/io/nirx/nirx.py
Swy7ch: Right. Shall I also move the explanation regarding the saturation in the Notes section?

mne/io/nirx/nirx.py
Swy7ch: Changing 'mask' with 'nan' then ?

rob-luke: > Filtering should work in the presence of NaNs

Nice, I didn't know this. But I guess more generally I don't write my functions expecting NaNs, perhaps this is something I should think of more in the future.

My vote is for larsoners proposal above.

mne/io/nirx/nirx.py
cbrnr: Yes please.

mne/io/nirx/nirx.py
cbrnr: My bad, I actually always agreed with @larsoner but I wrote `'annotate'` instead of `'nan'`. So yes, I think we now all agree on the solution:

- `saturated='ignore'` (the default) returns the original values (contained in .nosatflags_wl* files)
- `saturated='nan'` replaces saturated values with NaNs (contained in .wl* files)
- `saturated='annotate'` will return the original values (just like `'ignore'`) and in addition will create annotations at the locations of saturated values - however, this option is for another PR

rob-luke: Ok we are all on the same page now. Thanks for clarifying cbrnr.

> however, this option is for another PR

Can we do this all in this PR? There is no rush to get this merged as we don't even have a test set committed to the repo yet. I don't want to have incomplete support for NIRSport merged, then we are just going to have triage more issues. Let's add complete support in one hit. If annotate is too much work for one person, then multiple people can commit to this branch.

rob-luke: Also thanks for bearing with us Swy7ch. Let us know if there are any other decisions to be made that are blocking your progress.

Swy7ch: Don't worry, I actually think the process is quite active, which is a good think as far as I'm concerned :)

>Can we do this all in this PR?

Definitely ! I guess you already suggested it #7926, so I didn't ask again. I'll see what I can do concerning the annotation. A few questions in this regard:

* the data should be used as is (i.e. with NaN values), but the channels must be marked(like the 'bad' channels markin the [tutorial](https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#evaluating-the-quality-of-the-data)), right ? I don't fully understand @drammock's [solution](https://github.com/mne-tools/mne-python/pull/7936#issuecomment-650445349)
* any idea of the name of the marking ? 'saturated', I assume ?
* do you know if marking an object twice could lead to troubles ? I'll try to find a dataset with channels that are both of bad quality and containing NaN anyway, it's just to have a heads-up.

drammock: > the data should be used as is (i.e. with NaN values), but the channels must be marked (like the 'bad' channels markin the tutorial), right ? I don't fully understand @drammock's solution

TL;DR: you don't need to understand my solution, because it looks like we're not going to do that anyway (which is fine).

Full explanation: My suggestion of using `numpy.ma.masked_array` was an alternative to doing separate masked / non-masked options in the reader function. In a masked array approach, the data includes all the saturated values, but the nan-masked values are also read from disk and used to create a numpy mask over the data array. This is nice because it gives users the option to change their mind about using/not using the saturated values later on (without having to reload from disk) because masked arrays make it easy to ignore the mask and see the underlying data if you want to. But the downside is that most of the downstream code (analysis and plotting functions that operate on `Raw` objects) is not tested to work with masked arrays and I expect several things will break. The good news is, by having string options like `ignore` and `nan` and `annotate`, it leaves open the possibility to add a `masked` option later if we find that we need/want Raw objects to support masked arrays for other reasons.

Swy7ch:
Regarding the way to handle this 'annotate' flag, shall I mark the channels as 'saturated' like done [here](https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#evaluating-the-quality-of-the-data) ?

    sci = mne.preprocessing.nirs.scalp_coupling_index(raw_od)
    [...]
    raw_od.info['bads'] = list(compress(raw_od.ch_names, sci < 0.5))

That is to say, determine which channels contain saturated data, and add a 'saturated' entry in the `raw_*.info` dictionnary ?

Assuming the `_read_csv_row_cols()` [method](https://github.com/mne-tools/mne-python/blob/2965cc3644538c97b297f131bbb46ce9f954ea6f/mne/io/nirx/nirx.py#L333) can return more informations to the `_read_segment_files()` [method](https://github.com/mne-tools/mne-python/blob/2965cc3644538c97b297f131bbb46ce9f954ea6f/mne/io/nirx/nirx.py#L306), which in turn can pass the informations to the `read_raw_nirx()` method, we can update the `raw_*.info` dictionnary, just like shown in the example above.

If we're going this way, I strongly suggest (if it's not a requirement, which I think is the case since we'd be changing what is returned by several methods) to create another reader specifically for NIRSport1.

Thoughts @larsoner @rob-luke @cbrnr @drammock ?

cbrnr:
No, that's not what we want. We do not want to mark entire channels as bad, but only mark the specific segments where saturated values occur. The way we handle this is to create annotations in `raw.annotations`. Each saturated segment corresponds to an entry in this attribute.

drammock:
@Swy7ch here's a tutorial on how annotations work, if you need it: https://mne.tools/dev/auto_tutorials/raw/plot_30_annotate_raw.html

Swy7ch:
I wanted to check how the reader deals with NaN by default, so here's what I did:

- I loaded a dataset containing saturated data (i.e., a NaN value in the *.wl1 and *.wl2 files)
- At the end of `read_segment_file()`, I added a line to print the length of a row in the data object it creates
        - compare this value with the number or rows in the raw *.wl1 : they are the same value
- Then, I added a snippet to print the loaded data into files (one for each row in `data` inside `_read_segment_file()` then checked for any 'nan' value ; there are none
        - actually, when looking at the exact position where it should be, the saved value is a number that must be calculated at some point (hope so at least)

I took some screenshots to discuss on something tangible. The highlighted values are the values measured *just before* the NaN
test_wl2_23

Looking at this first picture, the value seem to match with the values before and after (that looks like some kind of regression). But then, looking at the second picture where two NaN were measured in a row, the values are quite surprising (especially the first one) : 
2018-07-31_116_21_01_1_wl1_28

I'll check deeper in the code where the load actually happens to see how those values are calculated, if they ever are.

Swy7ch:
Retgarding the way to annotate, I'm almost there. I do have one last problem: time retrieving. For now, I can't easily access raw data (in the matrix shape) to determine precisely the time of saturation. I could mod the _read_segment_file() method, but it looks dirty to me (as it is a reading method), so i ended up with this (a **RawNIRX** method):

    def annotate_saturated(self):
        times_to_annotate = []
        durations_to_annotate = []
        desc_to_annotate = []
        data, times = self.get_data(return_times=True)

        for channel_index, channel_values in enumerate(data):
            with open('script_output/output' + str(channel_index) + '.txt', 'w') as f:
                for time, value in zip(times, channel_values):
                    f.write("%s;%d\n" % (value, time))
                    if str(value) == 'nan':
                        print("%s @ %d" % ('nan', time))
                        times_to_annotate.append(time)
                        durations_to_annotate.append(1)
                        desc_to_annotate.append('nan')


The one problem with this snippet is that the `get_data()` method returns an array of times (corresponding to the data values), which is an array of seconds expressed with **integers**. Therefore, since there are multiple measurements done each second, it looks like this (value on the left, time on the right) : 

    0.8651118;2594
    0.8685657;2594
    0.867278;2594
    nan;2595
    0.8615325;2595
    0.8650122;2595
    0.8633895;2595
    0.8680118;2595

I *could* use a meter to manually retrieve the position of the saturated value in the relevant second (in the 'for' loop), but I would then need to retrieve the sampling rate to convert it back in time. Is there a better way to handle this problem ? I doubt annotating the whole second is a good idea, since "clean" data would be falsely annotate, right ?

Also have to deal with `orig_time` for the annotation, but this should be easy

larsoner:
I'd prefer to keep things as simple as possible for now -- it's generally easier to not try to implement everything in one go. How about we don't support `annotate` out of the box? For example we could instead consider adding a generic `mne.preprocessing.annotate_nan` that reads data, finds segments where any channels are `np.nan` and annotates them. Or later implement `saturated='annotate'` or whatever in the NIRSport reader.

Would this allow you to more easily add NIRSport support?

Swy7ch:
>it's generally easier to not try to implement everything in one go

Fair point !

>Would this allow you to more easily add NIRSport support?

NIRSport is already supported, only 'NaN' values are not (and natively, MNE does not read them). I can't say if writing a generic method would be easier or not, but it would have the major advantage of being reusable for other devices/datatypes. I can do it, of course !

Regarding the annotation, maybe I didn't understand correctly but one cannot annotate one single channel, right ? From what I've understood, we can annotate data during a specific timespan (like in [this](https://mne.tools/dev/auto_tutorials/raw/plot_30_annotate_raw.html#operations-on-annotations-objects) example), but it will concern all channels. I may have missunderstood though.

Swy7ch:
@larsoner Regarding the project structure, shall I create `mne.preprocesing.annotate_nan` or only add a method in `mne.preprocessing.artifact_detection`? Real question is : are 'NaN' values considered artifacts ?

mne/io/nirx/nirx.py
```diff
@@ -125,6 +125,10 @@ def __init__(self, fname, saturated, preload=False, verbose=None):
                         warn('You provided saturated data and specified '
                              'to use the standard *.wlX files.')
                         files[key] = files[key][1]
+                    elif saturated == 'annotate':
+                        warn('You provided saturated data and specified '
+                             'to annotate your data with a \'nan\' flag.')
+                        files[key] = files[key][1]
```
Swy7ch, line 131: We can't annotate data yet (since there is not), so I'm wondering when we should do it. Or leave the feature for later ?

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,41 @@
...
+                onsets.append(time)
+                durations.append(sampling_duration)
+                desc.append('nan')
+
+    if annot is None:
+        annot = Annotations(onsets, durations, desc)
+    else:
+        annot.append(onsets, durations, desc)
+
+    return annot
```
Swy7ch, line 41: Should we annotate `raw` inplace, instead of returning an **Annotations** instance containing the existing annotations and the new ones ?

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,41 @@
...
+    desc = []
+    sampling_duration = 1 / float(raw.info['sfreq'])
+    data, times = raw.get_data(return_times=True)
+    for row, channel_values in enumerate(data):
+        for time, value in zip(times, channel_values):
+            if str(value) == 'nan':
+                print("nan @ %d" % time)
+                onsets.append(time)
+                durations.append(sampling_duration)
+                desc.append('nan')
```
Swy7ch, line 34: It's a typo in regard of the docstring, but is it the good terminology ?

mne/io/nirx/nirx.py
larsoner: Are the data actually stored on disk as `NaN`, or are they just pegged to the A/D max or min value? If they're actually stored on disk as `NaN`, for the first version of this reader I would just read these directly as `np.nan`, and in a separate PR we can add a separate function `annotate_nan` or add a `kwarg` to this reader (not sure which, let's decide later)

rob-luke:
> Regarding the annotation, maybe I didn't understand correctly but one cannot annotate one single channel, right ? From what I've understood, we can annotate data during a specific timespan (like in this example), but it will concern all channels. I may have missunderstood though.

Why not just mark all channels as bad for the period that any channel has NaNs in it? If NaNs are typically caused by a movement artefact then its likely that all channels are bad for that period of time, and its only that certain channels were so bad that they saturated. Then if required in the future we can make channel specific annotations.

rob-luke:
Further, if I am processing data with NaNs in it I am probably not just going to ignore the NaN data, but all data +- a few seconds as the data would be rapidly growing before the NaNs, and the data after the NaNs is probably in the non linear part of the amplifier curve. So I think its probably overkill to mark specific channels as NaNs for specific time periods, and sufficient just to mark all channels as bad when any channel is NaN. 

But there may be use cases I haven't considered. Feel free to discuss specific uses

mne/io/nirx/nirx.py
Swy7ch: Yes, the data is actually stored as `NaN`, and is read this way. The problems rise when one tries to plot such data.

Swy7ch:
>Why not just mark all channels as bad for the period that any channel has NaNs in it? [...]  its likely that all channels are bad for that period of time.

Definitely! I was just questionning the fact that we annotate as `NaN` something that isn't *actually* `NaN`, but I didn't see the problem this way ; it does make sense.

>Further, if I am processing data with NaNs in it I am probably not just going to ignore the NaN data, but all data +- a few seconds as the data would be rapidly growing before the NaNs, and the data after the NaNs is probably in the non linear part of the amplifier curve. So I think its probably overkill to mark specific channels as NaNs for specific time periods, and sufficient just to mark all channels as bad when any channel is NaN.

Again, good point. We just need to decide what the timespan would be

cbrnr:
:+1: to annotate all channels (we can't do it any other way anyway), but let's use the time intervals defined by the NaNs. Users can expand these themselves if they want.

rob-luke:
>  but let's use the time intervals defined by the NaNs. Users can expand these themselves if they want.

Yes this is what I meant. I as an end user will post process the nans

mne/preprocessing/annotate_nan.py
Swy7ch: > 👍 to annotate all channels (we can't do it any other way anyway), but let's use the time intervals defined by the NaNs. Users can expand these themselves if they want.

Okay so this is what I did: annotating just for the duration of one measurement.

By the way, currently I made the method to annotate one NaN at a time, even if there are several that follow each other. I don't know if it's a good or a bad thing ? It is at least a simple one. I'll just take care of not making duplication if two channels have recorded a NaN value at the same time

larsoner: > By the way, currently I made the method to annotate one NaN at a time, even if there are several that follow each other. I don't know if it's a good or a bad thing ?

Bad thing performance wise, you end up with many more annotations this way. Better to compute where there are contiguous segments and annotate those directly

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,63 @@
...
+        for new_key in annotations_dict.keys():
+            old_value = annotations_dict[old_key]
+            if old_key < new_key <= (float(old_key) + old_value + tolerance):
+                # the data is processed from start to end,
+                # so keys are already sorted
+                if old_key + old_value <= new_key + annotations_dict[new_key]:
+                    diff = float(new_key) + annotations_dict[new_key] - \
+                           (old_key + old_value)
+                    annotations_dict[old_key] += diff
+                    to_delete.append(new_key)
```
Swy7ch, line 45: It seems rather dirty to me, but I couldn't come with a more simple way to merge segments. Any ideas ?

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,63 @@
...
+        for new_key in annotations_dict.keys():
+            old_value = annotations_dict[old_key]
+            if old_key < new_key <= (old_key + old_value + tolerance):
+                # the data is processed from start to end,
+                # so keys are already sorted
+                new_value = annotations_dict[new_key]
+                if old_key + old_value <= new_key + new_value:
+                    diff = new_key + new_value - (old_key + old_value)
+                    annotations_dict[old_key] += diff
+                    to_delete.append(new_key)
```
Swy7ch, line 45: This process seems rather dirty to me, but I couldn't come with a more simple way to merge segments. Any ideas ?

larsoner: First create a single book ndarray of shape `(times,)` with True where there are NaN, then use _mask_to_onsets_offsets

mne/preprocessing/annotate_nan.py
larsoner: I think in some other issue recently we decided to return a new annotations object rather than modify in place

mne/io/nirx/nirx.py
larsoner: In that case we shouldn't even have a `saturated` argument. The reader should just set these to `NaN`, and then people should make use of `mne.preprocessing.annotate_nan` or something else to fix them

Swy7ch: Fair enough. I'll remove the flags in `read_raw_nirx()` then. Shall I keep the warning concerning `NaN`, though ?

I also realized the data files that are read are the *.nosatflags_wlX (without `NaN`). Following your answer, I guess we shall let the user deal with these files and provide only one (be it *.wlX or *.nosatflags_wlX, it's up to them) ?

larsoner: > I also realized the data files that are read by default are the *.nosatflags_wlX (without NaN), not the *.wlX. Following your answer, I guess we shall let the user deal with these files and provide only one (be it *.wlX or *.nosatflags_wlX, it's up to them) ?

I don't fully understand what these files actually contain. Can you clarify? There are two files: `nosatflags_wlX` and `.wlX`. These files both contain data (?). One contains `NaN` and the other contains ... gaps where there should be NaN? The saturated values? Or do they contain something else (annotations of various things, one includes annotations of NaN channels/times and the other does not)?

Apologies if this is already specified somewhere, but the conversation has been quite extended over this PR so it's a bit difficult to find information :(

Swy7ch: >Apologies if this is already specified somewhere, but the conversation has been quite extended over this PR so it's a bit difficult to find information :(

No problem !

>These files both contain data (?).

Yes, they're actually *almost* identical. In case of saturation, the *.wlX files contain `NaN` instead of the value they're supposed to hold, like this one :
image

This a raw, saturated dataset, you're seing *.wl1, with our much loved `NaN`values instead of the value (between 0 and 1) they're supposed to be.

Now this is the *.nosatflags from the same dataset, which does not contain `NaN` values, but the value that was *measured* by the probe.
image

This is still a number between 0 and 1 (not necessarily 1, by the way), which leaves the question : "How does NIRStar conclude that this particular probe saturated?". There is no other informations provided, no annotation whatsoever

larsoner: > "How does NIRStar conclude that this particular probe saturated?"

Indeed, I expected it to be stuck at the max A/D value or something. I think I get it now. How about three options:

1. `'ignore'`: the non-NaN data are read as if there were no problem with the data
2. `'nan'`: the saturated values are replaced with NaN
3. `'annotate'`: the non-NaN data are read, but segments are annotated as `BAD_SATURATED` or something

For (3), it actually shouldn't be done by a separate `annotate_nan` function, since the actual (saturated) data values will be returned by `_read_raw_segment` (though we could still have a separate `annotate_nan`, and the `read_raw_nirx` reader could make use of the same underlying masking functionality). Or if you think that 1 and 2 are already enough, we can omit (3) for now. Or if even just (2) is enough to start, then we can add `saturated` later, and just always do (2) when the appropriate files are available.

Does this make sense to you? Sorry for the run-around on this, I should have looked into how things were actually stored on disk before asking you for changes.

Swy7ch: It's no big deal, really.

I like the 3 options way of handling things. 1 and 2 are already implemented, 3 still needs some work then.

>since the actual (saturated) data values will be returned by _read_raw_segment

Regarding th eway it's handled as of today, only one of those two files is actually loaded. I can definitely change this behavious (by creating a 'nan' entry in **keys** [here](https://github.com/mne-tools/mne-python/blob/4b282c833b9d6b86ab390b374863e2323fee131d/mne/io/nirx/nirx.py#L80), `keys('nan']` for example), to load *.nosatflags_wlX and use *.wlX to get saturated segments.

larsoner: Up to you if you want to implement 3. Maybe for now let's go with just 1 and 2, and see if we end up needing 3 eventually?

Swy7ch: 1 is already implemented (actually, the default behaviour, see [this](https://github.com/mne-tools/mne-python/pull/7936#issuecomment-656592972) comment

2 is also implemented with the 1st [commit](https://github.com/mne-tools/mne-python/pull/7936/commits/266bbab5268eebeef1f6d10c9e257656a24b69c6#diff-7c077bb43fb6aa0bdda5f6c0cdd99863R124) of this PR, but leads to problems with plotting (as `matplotlib` cannot plot `np.nan`)

>Maybe for now let's go with just 1 and 2, and see if we end up needing 3 eventually?

If you want to implement 1 and 2 first, only the first 2 commits are relevant. I'd just need to provide some data if it is really needed.

\-\-

3 is dealt with in this PR's 3rd [commit](https://github.com/mne-tools/mne-python/pull/7936/commits/4f2e5b37a3073317d2364b00ebe47f800c39cc14), so I'll finish the work (in another PR if needed, no problem !)

Regarding `annotate_nan()`, I'd say it's a convenient way to deal with `NaN` at any step of the process. We can force the annotation in the reader (even though the process would be quite the same, I think) since people will need to do that anyway if they want to plot something

larsoner: > We can force the annotation in the reader (even though the process would be quite the same, I think) since people will need to do that anyway if they want to plot something

The difference would be that the actual data values in the annotated ranges would be different between solution (3) and (2)-followed-by-annotate_nan. In the former they wouldn't be `nan`, and in the latter they would be. Not sure if this would be helpful or meaningful for people. But yes feel free to add it if you want.

It does lead to a nice way of testing things, namely that you should be able to ensure that the `raw.get_data()` from (1) matches that of (3); that the annotations of (2)-plus-`annotate_nan` match that the annotations of (3); and that the `nan` locations of (2) match the annotations in (2) (and implicitly (3) by the previous check).

Swy7ch: >It does lead to a nice way of testing things, namely that you should be able to ensure that the raw.get_data() from (1) matches that of (3); that the annotations of (2)-plus-annotate_nan match that the annotations of (3); and that the nan locations of (2) match the annotations in (2) (and implicitly (3) by the previous check).

Lovely :D 

We could wait for @rob-luke 's opinion on the subject, but I'm in to work on it : are you OK with 1, 2 and 3 as stated above by @larsoner ?

larsoner: @Swy7ch I would say go forward with the plan. If we can get it done in the next couple of weeks it'll make it into the upcoming 0.21 release

rob-luke:
@larsoner I think the thing blocking this PR from hitting 0.21 is a lack of test data. I think we need a documented recording committed to the test repo, not just some random file that's been passed around. If possible I would like to know the settings that were used to generate the data. So it might depend if labs around the world are open.

cbrnr:
IMO as long as we don't break things (which we don't), I think it's OK to include this even without official test data.

rob-luke:
Sure, but then what's the rush to get this merged if its untested? If its untested then it isn't known to be stable, and then shouldn't be in a release. If someone needs this feature then they can use this branch knowing that its a work in progress.

It also makes it hard for me to review. For example, this code introduces NaNs in the data (if `saturated='nan'`), how do we know that this works with downstream processing such as the Beer-Lambert law? My gut feeling is its going to throw uninformative errors all over the place (but maybe not, but that's the point I can't tell yet).

I know its hard to get in to the lab with the current global crisis and that is probably what's caused a hold up on getting testing data. I would much prefer real data, but if that's not possible and everyone wants this in the release untested, then could we mock the different conditions in tests to ensure that the data plays nice with downstream code?

cbrnr:
> Sure, but then what's the rush to get this merged if its untested? If its untested then it isn't known to be stable, and then shouldn't be in a release. If someone needs this feature then they can use this branch knowing that its a work in progress.

A lot of our code is based on ad-hoc implementation without weeks/months of testing (at least that's my impression). If we find bug, we can always fix them on the way. If we get it into the release, people will be able to use it - in my experience, checking out a branch is a big hurdle and will prevent most of our user base from using that feature.

> It also makes it hard for me to review. For example, this code introduces NaNs in the data (if saturated='nan'), how do we know that this works with downstream processing such as the Beer-Lambert law? My gut feeling is its going to throw uninformative errors all over the place (but maybe not, but that's the point I can't tell yet).

That's a good point, and I agree that this needs to be tested thoroughly, but it doesn't necessarily depend upon someone recording a new test data set. I thought that we already had such a data set with saturated values?

In any case, I absolutely do not want to rush you - if you think this needs more time, then don't feel obliged to get this done for 0.21.

mne/io/nirx/nirx.py
```diff
@@ -16,6 +16,7 @@
 from ...transforms import apply_trans, _get_trans
 from ...utils import logger, verbose, fill_doc
 from ...utils import warn
+from ...preprocessing import annotate_nan
```
Swy7ch, line 19: This line is causing problems during imports. There is an import loop while importing `preprocessing` modules :

    File "c:\users\juliend\documents\mne\mne\io\nirx\__init__.py", line 7, in <module>
        from .nirx import read_raw_nirx
    File "c:\users\juliend\documents\mne\mne\io\nirx\nirx.py", line 19, in <module>
        from ...preprocessing import annotate_nan
    File "c:\users\juliend\documents\mne\mne\preprocessing\__init__.py", line 11, in <module>
        from .maxfilter import apply_maxfilter
    File "c:\users\juliend\documents\mne\mne\preprocessing\maxfilter.py", line 10, in <module>
        from ..io import read_raw_fif
    ImportError: cannot import name 'read_raw_fif' from partially initialized module 'mne.io' (most likely due to a circular    import) (c:\users\juliend\documents\mne\mne\io\__init__.py)

Don't know how to resolve it since it doesnt seem to be linked to me...?

mne/io/nirx/nirx.py
```diff
@@ -363,6 +368,10 @@ def prepend(list, str):
             annot = Annotations(onset, duration, description)
             self.set_annotations(annot)
 
+        if saturated == "annotate":
+            annot = annotate_nan(self)
+            self.set_annotations(annot)
+
```
Swy7ch, line 374: Here is the block I need `preprocessing.annotate_nan`

Swy7ch (commented):
Problems rose when trying to import `preprocessing.annotate_nan` from io.nirx.nirx.py`, and I don't know what is causing the problem (nor how to resolve it)

mne/io/nirx/nirx.py
rob-luke: Could you put it with the other trouble maker here? 

https://github.com/mne-tools/mne-python/blob/806cd4785f927dbf262b7c878a2a8f143ee64a63/mne/io/nirx/nirx.py#L106

Swy7ch: It fixed the problem, thanks 🙂

rob-luke: Glad it worked 👍

mne/io/nirx/nirx.py
Swy7ch: I think I've come with something usable. I did test it on several "official" datasets from one of our studies, and everything seemed OK. Unfortunately I can't provide sample data, since my tutor is not here today and I don't know how to use NIRSport. Still need to provide tests for `preprocessing.annotate_nan` too, I'll work on it today.

I'll try to get some sample data before my intership ends on Friday, and update the PR if needed. My tutor knows about the PR though, so he may be able to help you out if we're short !

larsoner:
@Swy7ch do you need help with some aspect of this, or are we just waiting for a good tiny test dataset?

I'm going to push the milestone to 0.22 -- it'll give us 6 months to actually use the code and refine it, which is better than pushing it out in a rush and having to backport fixes.

Swy7ch:
As I as explained a few weeks ago (but this thread is soooo long that I can't blame you haha), my internship ended on August 28th, so I can't quite work on it for now (University and stuff). What I did was quite working tho, so any feedback is appreciated.

That said, my tutor @rderollepot (why can't I tag him ?) should be able to take the PR - we talked about it before I left the lab -, and provide the relevant updates. Anyway here's his account : https://github.com/rderollepot 

I will be off for the next few weeks, so I can't guarantee you any update on my side. Feel free to pm my tutor though ! :smiley:

rderollepot:
Concerning the following of the PR, I will try my best to follow @Swy7ch good work, but I must say I didn't dig as deeper in MNE as he did yet, and I'm not as used to the collaborative Github workflow as he was, so it might take me more time.

I should be able to provide test dataset(s) however. What kind of dataset(s) would we want ?
Should I just provide a direct download link, or include it somewhere in the project ?

rob-luke:
We also really appreciate the great work @Swy7ch did. Romain, I am happy to give more of a helping hand to get you up and running here (for example I can add the test dataset to the required place if you can get it to me anyway that's easy).

For a test dataset we want is the smallest recording that demonstrates the effect we want to test. So here we want to test that we can 1) load data from a nirsport device, and 2) test that saturations are dealt with. So I would suggest we need two recordings, each approximately 10s long. The first recording can just be any normal recording that you do, but ensure it does not saturate, this will allow us to test simply that we can load the data. The second recording should be the same except it saturates at some known time (is this possible? maybe this is hard to control?). Its handy to include a trigger or two in the measurement too, so that we can test that. If you can document the parameters used in the recordings similar to https://github.com/mne-tools/mne-testing-data/pull/51 that would be handy (let me know if I can help in any way).

Please ask any questions about the git process or what you plan to record in advance, easiest to discuss first and save you time later.

larsoner:
Ping @rderollepot if you can provide the testing data, we should be able to move forward with this. I think it's plausible we could merge this code for a potential 0.22 release in a few weeks

rderollepot:
Yeah sorry for the delay, I should be able to provide some good test data by the end of the week.

I had access to the lab last week and I did some recordings, in which I managed to trigger saturations simply by pulling on a detector to decouple the channel from the head. I made three recordings of approximately 15s:
- One with good calibration for every channel, doing nothing all run: no saturation
- One with good calibration except for a channel with a detector that I maintained pulled off all run (not in contact with the head): channel saturated during the all run
- A last one with good calibration for every channel, in which I pulled the same detector as above at 5s and 10s: controlled saturation on this channel at 5 and 10s

The problem I had with this acquisition is that I had no description of my probe setup at that time, so I quickly defined a *.tpl file manually to start an acquisition. This mean I don't have any probeInfo.mat file in the data. From an analysis point of view it doesn't really matter as we can always add a probe info definition later, but for our purpose I supposed this point could induce undesired side effects. So I properly defined my probe setup with NIRSite, and I only need to do the acquisition again in the same conditions. I was planning to do that this afternoon, so you should not have to wait much longer 😊 

I'll keep this "incomplete" test data though, as I think it may be useful to base some tests on it because as a user, I regularly had to update my probeInfo later in the analysis. We don't have a way to precisely measure probe locations on subject for now in the lab, so our probe definition is quite theroretical, and when there is a software improvement (like NIRSite), it's interesting to be able to define the setup more precisely a posteriori for the analysis.

rderollepot:
Just finished the PR with the three NIRSport (v1) datasets: mne-tools/mne-testing-data#78. Thanks @rob-luke for the example you linked above, it helped me a lot in the process !

I made some slight changes for this second acquisition:
- I used a proper probeInfo.mat file made with NIRSite this time
- All three datasets were recorded in a row using the same calibration, which was good (green) for all channels
- I added two triggers in each recording

Let me know if anything is missing or can be improved, or if any other testing data is needed.

rob-luke:
Looks great @rderollepot. I cut a new release at https://github.com/mne-tools/mne-testing-data/releases/tag/0.111

So you will need to update this line to point to 0.111, then it will give you a hash error, so you will need to update the hash a few lines below. Let us know if you get stuck.

https://github.com/mne-tools/mne-python/blob/fe8eb45abb779e5f8ec5739486a0dbadf8d408de/mne/datasets/utils.py#L248

rderollepot:
Sorry @rob-luke but there is something I am not sure to understand : at which point will I have the hash error ?

Is it when I will push the new commit with the updated release number ?

larsoner:
> Is it when I will push the new commit with the updated release number ?

Yes but also I just downloaded the .tar.gz and the hash I got was this if you want to try to avoid the error:
```
e7ece4615882b99026edb76fb708a3ce
```
Put this hash in when you update the release number and you should be good

rderollepot:
Ok thank you @larsoner !

larsoner:
Where are we on this one? We're going to try to release 0.23 soon and it would be nice to get this in!

rob-luke:
@rderollepot if you don't currently have the resources to complete this I can assist by pushing a few commits. But only if you wish, no pressure.

rob-luke:
@larsoner I just reviewed this code. Its great, but a decent way off finished. The docs don't reflect the code (I think later code modifications just weren't reflected in the docs) and the tests don't really test anything. The authors kindly contributed data to the testing repo, but it is not used here. I am going to put some effort in to tidying up the great work started by @rderollepot and @Swy7ch .

larsoner:
Sounds good!

rderollepot:
Oh, sorry for being late here, I'll catch up here and on the new PR.

> @rderollepot if you don't currently have the resources to complete this I can assist by pushing a few commits. But only if you wish, no pressure.

I'll gladly accept your help on this @rob-luke. Since the end of @Swy7ch's internship I've indeed been lacking some resources and I wasn't able to consistently spend some time on MNE myself. He also was much more used to the collaborative quality development process than I am, and I must confess I didn't even realize the job on this PR was not finished, sorry for that.

So thanks a lot for helping here ! I'll jump to #9348

rob-luke:
Geez, I'd actually thought this was merged @larsoner. I will take a look at the code this week and evaluate the effort required and if its possible with my schedule in the next two weeks. After changing the ppf, this is my next top priority.

PR #7940, opened 2020-06-29
MRG: Add time player for the notebook backend

GuillaumeFavelier:
This PR adds a "player" controller for the time slider with play, pause, stop and rewind. It also reorders the controllers for a cleaner UI.

image

It's part of #7162

PR #7943, opened 2020-06-29
MRG, MAINT: Tweak CIs

### LINKED ISSUE #7939: Following the installation instructions for macOS yields an error message
#### Describe the bug
Following the installation instructions for macOS  in the development version of the docs produces an error message, not a working MNE-Python installation

#### Steps to reproduce
Download the conda environment file from https://raw.githubusercontent.com/mne-tools/mne-python/master/environment.yml

Run
```bash
conda env create -f environment.yml -n mne-test
```

#### Expected results
A working environment should be created.

#### Actual results
Installation fails due to an unsatisfiable dependency:
```bash
Pip subprocess error:
ERROR: Could not find a version that satisfies the requirement vtk<8.2 (from -r /private/tmp/condaenv.fkd0fvyw.requirements.txt (line 4)) (from versions: 9.0.0, 9.0.1)
ERROR: No matching distribution found for vtk<8.2 (from -r /private/tmp/condaenv.fkd0fvyw.requirements.txt (line 4))


CondaEnvException: Pip failed
```

#### Additional information
This is due to the fact that `python` is pinned to `>= 3.5` and `vtk` is pinned to `< 8.2` on macOS. This will first install Python 3.8 (via `conda`) and bomb during `pip` install of `vtk`, because the only wheels available for Python 3.8 are for VTK 9.

VTK was pinned to version 8 on macOS because of some [issues](https://github.com/pyvista/pyvistaqt/pull/16) with version 9.

Since conda environment files do not support selectors, there are currently only 3 solutions to this problem:
- fix the bugs that keep us from using VTK 9 on macOS
- provide a separate environment file for macOS, which pins Python to < 3.8
- suggest that users install the `conda-forge` package of MNE-Python, as the "clean" way to declare system-dependent requirements in conda land is by producing separate packages

cc @GuillaumeFavelier

larsoner:
1. Permit operators to start newlines in addition to at the end of lines (we have ~575 of these in `master`) so that we can transition to or at least allow the "new" (as of 2016) way of writing code, [as discussed here](https://github.com/mne-tools/mne-python/pull/7936#issuecomment-651020811)
2. Fix or ignore a few style errors in `conf.py`
3. Run `make flake` on the entire codebase not just select folders
4. Switch to `conda` for VTK (#7939)
5. Fix CircleCI failures related to `numpydoc` seen [here](https://github.com/mne-tools/mne-python/pull/7940#issuecomment-651107331)

Closes #7939

NumpyDoc will still cause a failure, I think it needs to be fixed at the NumpyDoc end...


larsoner:
Fixed VTK9 rendering on macOS by disabling depth peeling and anti-aliasing. The remaining bug on my macOS system is that the widget controls do not work because there is a 2x mismatch between their render position and the click position needed to control them. We'll have to see if this is fixable at the PyVista end, or if this is an underlying VTK bug (my money is on the latter).

hoechenberger:
> The remaining bug on my macOS system is that the widget controls do not work because there is a 2x mismatch between their render position and the click position needed to control them. We'll have to see if this is fixable at the PyVista end, or if this is an underlying VTK bug (my money is on the latter).

This suggests that there's something off with the conversion between retina pixels and displayed points

mne/cov.py
```diff
@@ -63,8 +63,8 @@ def _get_tslice(epochs, tmin, tmax):
     return tslice
 
 
-class Covariance(dict):
-    """Noise covariance matrix.
+class Covariance(dict_):
+    """Noise covariance matrixx.
```
hoechenberger, line 67: Do I need to understand this? :)

larsoner: argh no this is debugging cruft

mne/fixes.py
```diff
@@ -1091,3 +1091,61 @@ def mean(array, axis):
     @contextmanager
     def nullcontext(enter_result=None):
         yield enter_result
+
+
+###############################################################################
+# Work around autosummary making bad :class: links
```
agramfort, line 1097: this sucks. It was working with an older version? if so can we rather pin the version rather than crippling our code?

larsoner: How does this cripple our code? It just replaces a few docstrings. Subclassing like `isinstance(x, dict)` are still fine. Overhead should be minimal.

The problem is that numpydoc improved the signature extraction, and we hit a bug now with autosummary trying to add effectively:
```
:class:`v, remove specified key and return the corresponding value.`
```
to correspond to the `D.pop() -> v, remove specified key and return the corresponding value.` docstring.

This seemed like the cleanest solution, but I'm open to other ideas.

agramfort:
it's because numpydoc does not parse docstring of std lib?

>

larsoner:
See above -- I think it's because numpydoc passes the stdlib definition unchanged, then autodoc sees the `->` in the docstring and assumes that anything after that is only a `:class:` (hence an autodoc/autosummary bug)

agramfort:
would it make sense to wait for autodoc/autosummary to fix the bug?

>

larsoner:
Maybe, but in the meantime all CircleCI builds will fail. I guess we could roll back numpydoc in the meantime, but no idea how long the autodoc fix will take.

larsoner:
Personally I don't see why the `dict_` class is so bad, it just adds one more entry in the MRO...

agramfort:
ok fair enough

let's proceed and see in a few weeks/months if we can undo this after a fix
upstream

>

hoechenberger:
I agree with @agramfort that it is kind of ugly, but I think it's okay for now :)

hoechenberger:
I believe this also begs the question if / whether we should switch to `conda-forge` packages anyway:

```bash
$ conda search vtk |grep pkgs/main
vtk                            8.1.0 py27h04d9db9_201  pkgs/main
vtk                            8.1.0 py27h87a2a2b_201  pkgs/main
vtk                            8.1.0 py35h04d9db9_201  pkgs/main
vtk                            8.1.0 py35h87a2a2b_201  pkgs/main
vtk                            8.1.0 py36h04d9db9_201  pkgs/main
vtk                            8.1.0 py36h87a2a2b_201  pkgs/main
vtk                            8.1.1 py27ha9eb873_204  pkgs/main
vtk                            8.1.1 py27hce9f6a2_204  pkgs/main
vtk                            8.1.1 py36ha9eb873_204  pkgs/main
vtk                            8.1.1 py36hce9f6a2_204  pkgs/main
vtk                            8.2.0 py27h9bafd54_200  pkgs/main
vtk                            8.2.0 py36h9bafd54_200  pkgs/main
vtk                            8.2.0 py37h9bafd54_200  pkgs/main
```

hoechenberger:
@larsoner feel free to revert 237e3e1, just want to see if this works…

GuillaumeFavelier:
@hoechenberger I think it follows the news on https://github.com/pyvista/pyvistaqt/pull/16#issuecomment-651149656 and https://github.com/pyvista/pyvistaqt/issues/15#issuecomment-651227725

Fetching `vtk 9.0.1` on PyPi for every OS could solve the problem.

hoechenberger:
> Fetching `vtk 9.0.1` on PyPi for every OS could solve the problem.

Ok so we only need to pin VTK to >=9.0.1?

GuillaumeFavelier:
For MacOS at least, yes

GuillaumeFavelier:
The failures of the notebook job are note related, see https://github.com/mne-tools/mne-python/pull/7944

GuillaumeFavelier:
The `Windows 3.7 conda` job on Azure complains that `pytest` is not a command. It's not installed?

agramfort:
try:

python -m pytest

see https://docs.pytest.org/en/stable/usage.html

larsoner:
The command failing is often an indicator that the install still failed (but didn't error out for some reason)

GuillaumeFavelier:
I pushed again, it seems there was indeed an error in the installation.

GuillaumeFavelier:
Still not happy:

```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

      File "d:\a\1\s\conda\lib\site-packages\conda\exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "d:\a\1\s\conda\lib\site-packages\conda_env\cli\main.py", line 80, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "d:\a\1\s\conda\lib\site-packages\conda_env\cli\main_update.py", line 68, in execute
        directory=os.getcwd())
      File "d:\a\1\s\conda\lib\site-packages\conda_env\specs\__init__.py", line 40, in detect
        if spec.can_handle():
      File "d:\a\1\s\conda\lib\site-packages\conda_env\specs\yaml_file.py", line 18, in can_handle
        self._environment = env.from_file(self.filename)
      File "d:\a\1\s\conda\lib\site-packages\conda_env\env.py", line 151, in from_file
        return from_yaml(yamlstr, filename=filename)
      File "d:\a\1\s\conda\lib\site-packages\conda_env\env.py", line 136, in from_yaml
        data = yaml_load_standard(yamlstr)
      File "d:\a\1\s\conda\lib\site-packages\conda\common\serialize.py", line 76, in yaml_load_standard
        return yaml.load(string, Loader=yaml.Loader, version="1.2")
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\main.py", line 640, in load
        return loader._constructor.get_single_data()  # type: ignore
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\constructor.py", line 102, in get_single_data
        node = self.composer.get_single_node()
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 75, in get_single_node
        document = self.compose_document()
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 96, in compose_document
        node = self.compose_node(None, None)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 132, in compose_node
        node = self.compose_mapping_node(anchor)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 194, in compose_mapping_node
        item_value = self.compose_node(node, item_key)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 130, in compose_node
        node = self.compose_sequence_node(anchor)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 163, in compose_sequence_node
        node.value.append(self.compose_node(node, index))
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 132, in compose_node
        node = self.compose_mapping_node(anchor)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 194, in compose_mapping_node
        item_value = self.compose_node(node, item_key)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 130, in compose_node
        node = self.compose_sequence_node(anchor)
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\composer.py", line 162, in compose_sequence_node
        while not self.parser.check_event(SequenceEndEvent):
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\parser.py", line 144, in check_event
        self.current_event = self.state()
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\parser.py", line 519, in parse_indentless_sequence_entry
        token.move_comment(self.scanner.peek_token())
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\scanner.py", line 173, in peek_token
        self.fetch_more_tokens()
      File "d:\a\1\s\conda\lib\site-packages\ruamel_yaml\scanner.py", line 310, in fetch_more_tokens
        % utf8(ch), self.reader.get_mark())
    ruamel_yaml.scanner.ScannerError: while scanning for the next token
    found character '\t' that cannot start any token
      in "<unicode string>", line 35, column 44:
         ... 0.1; platform_system == "Darwin"		
                                             ^ (line: 35)
```

There is a tab in the file XD

environment.yml
```diff
@@ -32,7 +32,7 @@ dependencies:
   - mne
   - https://github.com/numpy/numpydoc/archive/master.zip
   - imageio-ffmpeg>=0.4.1
-  - vtk>=9.0.1; platform_system == "Darwin"		
+  - vtk>=9.0.1; platform_system == "Darwin"
   - vtk; platform_system != "Darwin"
```
larsoner, line 36: Let's make things simpler and just require vtk>=9.0.1 on all platforms

agramfort:
how do we proceed? we have 2 PRs that are red but collectively could be green?

GuillaumeFavelier:
I branched out in https://github.com/mne-tools/mne-python/pull/7944. If it makes things easier, I can integrate it in here?

agramfort:
ok for me

>

agramfort:
thx a lot @larsoner and @GuillaumeFavelier !

PR #7944, opened 2020-06-30
MRG, MNT: Migrate to nbclient for notebook testing

GuillaumeFavelier:
Checks that the CIs are green on Azure for the notebook 3d backend.

### DIFF
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -297,14 +297,8 @@ def _check_skip_backend(name):
 def renderer_notebook():
     """Verify that pytest_notebook is installed."""
     from mne.viz.backends.renderer import _use_test_3d_backend
-    try:
-        with warnings.catch_warnings():
-            warnings.filterwarnings("ignore", category=DeprecationWarning)
-            from pytest_notebook import execution
-    except ImportError:
-        pytest.skip("Test skipped, requires pytest-notebook")
     with _use_test_3d_backend('notebook'):
-        yield execution
+        yield
 
 
 @pytest.fixture(scope='function', params=[testing._pytest_param()])
--- a/mne/viz/_brain/tests/test_notebook.py
+++ b/mne/viz/_brain/tests/test_notebook.py
@@ -10,13 +10,14 @@ PATH = os.path.dirname(os.path.realpath(__file__))
 @pytest.mark.slowtest
 @testing.requires_testing_data
 @requires_version('nbformat')
+@requires_version('nbclient')
 def test_notebook_3d_backend(renderer_notebook):
     """Test executing a notebook that should not fail."""
     import nbformat
+    from nbclient import NotebookClient
 
-    notebook = nbformat.read(
-        os.path.join(PATH, "test.ipynb"), as_version=4,
-    )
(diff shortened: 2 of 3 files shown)


GuillaumeFavelier:
Now there is a warning on Azure for notebook testing:

```
FutureWarning: Method cleanup(connection_file=True) is deprecated, use cleanup_resources(restart=False).
```

GuillaumeFavelier:
This is probably because of the new version of `jupyter_client`:

image

GuillaumeFavelier:
Bumping `nbconvert` solved the warning but now `pytest-notebook` is [not ready for the change](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=7866&view=logs&j=1d294aa9-8c9e-52da-e2cc-28ca059e1c9b&t=520de7d9-efc8-52cf-0de1-994cbfe115fe&l=63):

```
AttributeError: 'ExecuteCoveragePreprocessor' object has no attribute 'setup_preprocessor'
```

GuillaumeFavelier:
Think it's fine since `nbconvert 6.0.0` is not officially released yet. We could probably ignore for now and report the change on [pytest-notebook](https://github.com/chrisjsewell/pytest-notebook) ?

GuillaumeFavelier:
This is trickier than it seems. There is an API change in `nbconvert 6.0.0` leading to some part of the code to be split in `nbclient`.

GuillaumeFavelier:
As we do not gather coverage informations from this notebook, I could consider migrating to `nbclient` instead.

GuillaumeFavelier:
I reported the `FutureWarning` in https://github.com/jupyter/nbclient/issues/89.

For now, we can just pin `jupyter_client<6.1.5`.

GuillaumeFavelier:
What do you think @agramfort, @larsoner ?

agramfort:
I am confused. Why removing pytest-notebook?

GuillaumeFavelier:
We don't use the coverage information and it's not ready yet for `nbconvert 6.0.0`.

larsoner:
What do you mean we don't use it? It should be reported like all other runs (other than style)

GuillaumeFavelier:
>  It should be reported like all other runs

I was not aware of that

agramfort:
coverage report or execution? I can live without coverage but the notebook
definitely needs to be
executed by pytest

>

GuillaumeFavelier:
`nbclient` executes the notebook `test.ipynb` in the `test_notebook_3d_backend()` function of the `test_notebook.py` file.

I'm now completely lost for coverage.

agramfort:
if test.ipynb runs on CIs than I am happy

>

larsoner:
> nbclient executes the notebook test.ipynb in the test_notebook_3d_backend() function of the test_notebook.py file.
>
> I'm now completely lost for coverage.

As long as it does this in the same Python process, I think `coverage` should work. And based on the coverage numbers for the notebook backend, it should be working.

https://codecov.io/gh/mne-tools/mne-python/src/master/mne/viz/backends/_notebook.py

larsoner:
Closing in favor of #7943

PR #7948, opened 2020-06-30
MAINT: Skip a few more on macOS

larsoner:
[We're very close to the 50 minute limit on macOS](https://travis-ci.org/github/mne-tools/mne-python/jobs/703573579#L3294) so mark a few more as `slowtest`.


agramfort:
thx @larsoner !

PR #7950, opened 2020-07-01
[MRG] fix Issue with stc.project after restricting to a label

### LINKED ISSUE #7932: stc.project('pca', src) requires fwd['src']
Sometimes a few source locations are dropped in forward computation so the `stc` gets out of sync with the original `src`. this means you need to call `stc.project('pca', fwd['src'])`. It would be nice if you could use the original `src`.

This problem used to exist with `stc.extract_label_time_course` but that has been fixed maybe by #6789?

I'm not sure how to replicate this issue on sample data but I could share some data that exhibits the issue if that is needed.

bloyl:
`source_space._get_src_nn` assumed a 1-1 mapping between `stc.vertices` and `src.vertices` which broke `stc.project` if `stc` was already restricted to a label or during the forward computation. 

Closes #7932

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -116,6 +116,8 @@ Bug
 
 - Fix bug with :func:`mne.preprocessing.ICA.find_bads_eog` when more than one EOG components are present by `Christian O'Reilly`_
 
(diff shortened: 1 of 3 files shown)


larsoner:
Thanks @bloyl

PR #7951, opened 2020-07-01
MRG: Only process NIRX event files if they are present

rob-luke:
In the event that no triggers are sent during a recording (for example a resting state measurement) the event file is not created. This PR only tries to create annotations if the event file is present.

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -77,7 +77,7 @@ class RawNIRX(BaseRaw):
 
         # Check if required files exist and store names for later use
         files = dict()
-        keys = ('evt', 'hdr', 'inf', 'set', 'tpl', 'wl1', 'wl2',
+        keys = ('hdr', 'inf', 'set', 'tpl', 'wl1', 'wl2',
                 'config.txt', 'probeInfo.mat')
         for key in keys:
             files[key] = glob.glob('%s/*%s' % (fname, key))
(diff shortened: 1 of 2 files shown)


mne/io/nirx/nirx.py
```diff
@@ -288,19 +288,20 @@ def prepend(list, str):
...
-        for t_idx in range(len(t)):
-            binary_value = ''.join(t[t_idx][1:])[::-1]
-            trigger_frame = float(t[t_idx][0])
-            onset[t_idx] = (trigger_frame) * (1.0 / samplingrate)
-            duration[t_idx] = 1.0  # No duration info stored in files
-            description[t_idx] = int(binary_value, 2) * 1.
-        annot = Annotations(onset, duration, description)
-        self.set_annotations(annot)
+        if op.isfile(files['hdr'][:-4]+'.evt'):
+            with _open(files['hdr'][:-4]+'.evt') as fid:
```
cbrnr, line 292: Surround `+` with spaces. Also, you can re-use the dot from `files['hdr'][:-3]`. Bonus comment: using `pathlib` would expose the file stem and extension without having to do any string slicing (`pathlib` would actually be a better replacement for `os.path` - but this could be done in a separate PR).

rob-luke: Thanks

rob-luke:
@larsoner @agramfort could you please review

agramfort:
@larsoner merge if happy

cbrnr:
Thanks @rob-luke!

Issue #7814, opened 2020-05-23
Azure Windows 3.7 conda job keeps failing


cbrnr:
For quite some time the Azure Windows 3.7 conda job has been throwing the following error:

```
FAILED mne/minimum_norm/tests/test_inverse.py::test_make_inverse_operator_diag[testing_data-testing_data]
```

More detailed log message:

```
================================== FAILURES ===================================
_________ test_make_inverse_operator_diag[testing_data-testing_data] __________
[gw0] win32 -- Python 3.7.4 D:\a\1\s\conda\python.exe
mne\minimum_norm\tests\test_inverse.py:727: in test_make_inverse_operator_diag
    rtol=1e-1, atol=1e-1, ctol=0.99, check_K=False)
mne\minimum_norm\tests\test_inverse.py:182: in _compare_inverses_approx
    assert corr > ctol
E   assert 0.7766994175372354 > 0.99
```

@larsoner suspects that this is related to MKL and probably hard to track down.

agramfort:
@larsoner any thought?

>

larsoner:
It seems to be somewhat sporadic, I've only seen it a handful of times. Not sure how that's possible, unless somewhere memory is getting corrupted. I can try replicating but want to see if it shows up again, since the last couple of PRs I looked at did not show a problem...

cbrnr:
https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=7369&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=f8799db5-b972-5f58-516d-bc88b8357f99

larsoner:
So it is non-deterministic. Great -- to my windows boot I go, wish me luck...

GuillaumeFavelier:
What is the current situation on this @cbrnr ? Probably many things have changed since the issue has been opened.

cbrnr:
I haven't seen this lately (but I didn't have too many PRs either). @larsoner?

larsoner:
I haven't seen it in a while, let's optimistically close and reopen if need be

larsoner:
Not dead, happened in #7952

larsoner:
Okay maybe it's fixed now...

PR #7952, opened 2020-07-01
Fix some small glitches introduced via #7845

hoechenberger:


agramfort:
@hoechenberger merge when CIs are green

hoechenberger:
@larsoner Any idea why CI is failing on Python 3.7 on Windows?

larsoner:
Just a spurious linalg problem noted in #7814 that I was hoping was dead...

larsoner:
Thanks @hoechenberger

PR #7953, opened 2020-07-01
MRG, MAINT: Clean up VTK9 offset array

larsoner:
Okay for you @GuillaumeFavelier ? This seems to be the right way to fix it (?) and it seems to work in a simple testing example:
```
import os.path as op
import mne

initial_time = 0.1

size = (800, 400)
                     size=size, background='k')
```


larsoner:
Output looks correct

https://20879-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_mne_dspm_source_localization.html#dipole-orientations

PR #7954, opened 2020-07-01
MRG+1: BUG: info['bads'] order shouldn't matter in write_evokeds()

hoechenberger:

I ran into an issue where I created two Evokeds and their difference (via `combine_evoked()`); say, `e1`, `e2`, and `diff`.

I then wanted to write those Evokeds to a single file via `mne.write_evokeds([e1, e2, diff])`, which [failed](https://app.circleci.com/pipelines/github/mne-tools/mne-study-template/459/workflows/6ae202d0-9190-4192-bdfd-08f3e5f12fa3/jobs/477/steps) with a `ValueError`: "info['bads'] must match'". I looked into the `bads` lists and found that the order of bad channels in `diff` was different from the order in `e1` and `e2`; yet, the set of channels was identical.

While it probably would make sense to figure out where this difference comes from, it also got me thinking that an altered `bads` order shouldn't be a stopper when using `write_evokeds()`, as long as all passed Evoked objects mark the same channels as bad.

And this is exactly what this PR implements.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -105,11 +105,13 @@ Changelog
 
 - Add ``'auto'`` option to :meth:`mne.preprocessing.ICA.find_bads_ecg` to automatically determine the threshold for CTPS method by `Yu-Han Luo`_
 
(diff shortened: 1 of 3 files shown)


hoechenberger:
@larsoner I've added a changelog entry.

agramfort:
Thx @hoechenberger

Issue #7751, opened 2020-05-07
RFC, MAINT: changes to 2D plotting


drammock:
This issue is for planning changes to our 2D plotting capabilities ([roadmap link](https://mne.tools/dev/overview/roadmap.html?highlight=roadmap#id2)).  As a starting point, here are (I think?) all the functions/methods that make 2D plots, grouped roughly by the kind of plot they generate. Also included is a section for 3D plots that always use matplotlib rather than mayavi/pyvista. Each section is followed by some comments / proposals about what might be done. Some of the proposed changes are big, others are quite minor. When folks have weighed in, I'll convert it to a consolidated checklist for keeping track of which ones are done.

## Browse-type windows (scrollable data)

- raw.plot
- epochs.plot
- ica.plot_sources(raw)

### possible changes
1. Anything we can do to speed up the responsiveness of scrolling / scaling of browse-type windows would be desirable.
2. One possible approach to unifying the underlying code would be a new class `MNEBrowseFigure` that inherits from `matplotlib.figure.Figure` and automatically does a bunch of stuff in its `__init__` like generating the scrollbars and buttons, setting margins, etc. This would also allow the (somewhat unwieldy) `params` dictionaries that are passed back and forth to be instead stored as attributes of the Figure object (nested under an `mne` attribute so they can be accessed as `fig.mne.property_name`, thereby segregating the MNE-specific attributes from the matplotlib ones).

## Interactive lineplots (click-drag to pop up scalp topographies, click to show channel names)

- raw.plot_psd
- epochs.plot_psd
- evoked.plot
- evoked.plot_white
- evoked.plot_joint

### possible changes
1. evoked.plot_joint currently generates separate figures for each channel type; all the others generate a single figure with subplots for each channel type.  It's worth considering whether there's a way for plot_joint to be single-figure like the others.
2. all of these have options for spatial_colors; the head circle in the inset axes gets stretched to an ellipse when the window is resized. Should fix the aspect ratio of the inset axes if possible.

## scalp field maps  (possibly w/ trellis)

- {raw/epochs/evoked/mne.viz}.plot_projs_topomap
- epochs.plot_psd_topomap
- evoked.plot_topomap
- ica.plot_components
- mne.preprocessing.corrmap
- mne.viz.plot_arrowmap
- mne.viz.plot_tfr_topomap

### possible changes
1. make sure all of these have `vlim='joint'` option
[...]

PR #7955, opened 2020-07-01
MRG, VIZ, MAINT: figure class proof-of-concept

drammock:
This is an element of #7751. It defines new classes:

- `MNEFigParams` (a simple attribute container)
- `MNEFigure` (wrapper for `matplotlib.figure.Figure` with our params added)
- `MNEBrowseFigure` (Figure class for raw.plot, ica.plot, epochs.plot)

So far just laying down the infrastructure, none of the instance methods are actually using the new classes yet.  If anyone wants to play around with it:

```python
import os
import mne
plt.ion()
                                    'sample_audvis_raw.fif')
fig = mne.viz.raw.plot_raw_alt(raw)
# the "params" live under fig.mne:
print(fig.mne.ax_help)  # etc
```

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -1258,7 +1258,7 @@ class MNEBrowseFigure(MNEFigure):
             if active:
                 labels[ix] += ' (already applied)'
         # make figure
-        width = max([4, max([len(label) for label in labels]) / 6 + 0.5])
+        width = max([4, max([len(label) for label in labels]) / 8 + 0.5])
         height = (len(projs) + 1) / 6 + 1.5
         fig = self._new_child_figure(figsize=(width, height),
                                      fig_name='fig_proj',
@@ -1267,7 +1267,7 @@ class MNEBrowseFigure(MNEFigure):
         # make axes
         offset = (1 / 6 / height)
         position = (0, offset, 1, 0.8 - offset)
-        ax = fig.add_axes(position, frame_on=False)
+        ax = fig.add_axes(position, frame_on=False, aspect='equal')
         # make title
         first_line = ('Projectors already applied to the data are dimmed.\n'
                       if any(self.mne.projs_active) else '')
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -575,7 +575,11 @@ class MNEBrowseFigure(MNEFigure):
             self.mne.remove_dc = not self.mne.remove_dc
             self._redraw()
         elif key == 'p':  # toggle draggable annotations
-            self.mne.draggable_annotations = not self.mne.draggable_annotations
+            self._toggle_draggable_annotations(event)
+            if self.mne.fig_annotation is not None:
+                checkbox = self.mne.fig_annotation.mne.drag_checkbox
+                with _events_off(checkbox):
+                    checkbox.set_active(0)
         elif key == 's':  # scalebars
             self._toggle_scalebars(event)
         elif key == 'w':  # toggle noise cov whitening
@@ -813,7 +817,7 @@ class MNEBrowseFigure(MNEFigure):
 
     def _create_annotation_fig(self):
         """Create the annotation dialog window."""
-        from matplotlib.widgets import Button, SpanSelector
+        from matplotlib.widgets import Button, SpanSelector, CheckButtons
         from mpl_toolkits.axes_grid1.axes_size import Fixed
         from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
         # make figure
@@ -861,6 +865,32 @@ class MNEBrowseFigure(MNEFigure):
         fig.button = Button(button_ax, 'Add new label')
         fig.button.on_clicked(self._add_annotation_label)
         plt_show(fig=fig)
+        # add "draggable" checkbox
+        drag_ax_height = 3 * pad
+        drag_ax = div.append_axes('bottom', size=Fixed(drag_ax_height),
+                                  pad=Fixed(pad), aspect='equal')
+        checkbox = CheckButtons(drag_ax, labels=('Draggable edges?',),
+                                actives=(self.mne.draggable_annotations,))
+        checkbox.on_clicked(self._toggle_draggable_annotations)
+        fig.mne.drag_checkbox = checkbox
+        # reposition & resize axes
+        width_in, height_in = fig.get_size_inches()
+        width_ax = fig._inch_to_rel(width_in - 2 * pad)
+        aspect = width_ax / fig._inch_to_rel(drag_ax_height)
+        drag_ax.set_xlim(0, aspect)
+        drag_ax.set_axis_off()
+        # reposition & resize checkbox & label
+        rect = checkbox.rectangles[0]
+        _pad, _size = (0.2, 0.6)
+        rect.set_bounds(_pad, _pad, _size, _size)
+        lines = checkbox.lines[0]
+        for line, direction in zip(lines, (1, -1)):
+            line.set_xdata((_pad, _pad + _size)[::direction])
+            line.set_ydata((_pad, _pad + _size))
+        text = checkbox.labels[0]
+        text.set(position=(3 * _pad + _size, 0.45), va='center')
+        for artist in lines + (rect, text):
+            artist.set_transform(drag_ax.transData)
         # setup interactivity in plot window
         col = ('#ff0000' if len(fig.mne.radio_ax.buttons.circles) < 1 else
                fig.mne.radio_ax.buttons.circles[0].get_edgecolor())
@@ -871,6 +901,10 @@ class MNEBrowseFigure(MNEFigure):
         self.mne.ax_main.selector = selector
         self.mne._callback_ids['motion_notify_event'] = \
             self.canvas.mpl_connect('motion_notify_event', self._hover)
+
+    def _toggle_draggable_annotations(self, event):
+        """Enable/disable draggable annotation edges."""
+        self.mne.draggable_annotations = not self.mne.draggable_annotations
 
     def _update_annotation_fig(self):
         """Draw or redraw the radio buttons and annotation labels."""
@@ -888,7 +922,7 @@ class MNEBrowseFigure(MNEFigure):
         # populate center axes with labels & radio buttons
         ax.clear()
         title = 'Existing labels:' if len(labels) else 'No existing labels'
-        ax.set_title(title, size=None)
+        ax.set_title(title, size=None, loc='left')
         ax.buttons = RadioButtons(ax, labels)
         # adjust xlim to keep equal aspect & full width (keep circles round)
         aspect = (width - 2 * pad) / var_height
@@ -933,14 +967,16 @@ class MNEBrowseFigure(MNEFigure):
         0.3  text entry
         0.1  padding above button
         0.3  button
+        0.1  padding above checkbox
+        0.3  checkbox
         0.1  bottom margin
         ------------------------------------------
-        2.5  total fixed height
+        2.9  total fixed height
         """
         pad = 0.1
         width = 4.5
         var_height = max(pad, 0.7 * n_labels)
-        fixed_height = 2.5
+        fixed_height = 2.9
         return (width, var_height, fixed_height, pad)
 
     def _add_annotation_label(self, event):
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -270,10 +270,9 @@ class MNEBrowseFigure(MNEFigure):
 
         # additional params for browse figures
         self.mne.ch_start = 0
+        self.mne.butterfly_image = None
         self.mne.projector = None
         self.mne.projs_active = np.array([p['active'] for p in self.mne.projs])
-        self.mne.event_lines = None
-        self.mne.event_texts = list()
         self.mne.whitened_ch_names = list()
         self.mne.use_noise_cov = self.mne.noise_cov is not None
         # annotations
@@ -285,6 +284,10 @@ class MNEBrowseFigure(MNEFigure):
         self.mne.annotation_segment_colors = dict()
         self.mne.annotation_hover_line = None
         self.mne.draggable_annotations = False
+        # lines
+        self.mne.event_lines = None
+        self.mne.event_texts = list()
+        self.mne.vline_visible = False
         # scalings
         self.mne.scale_factor = 0.5 if self.mne.butterfly else 1.
         self.mne.scalebars = dict()
@@ -732,7 +735,8 @@ class MNEBrowseFigure(MNEFigure):
         self._update_picks()
         self._update_trace_offsets()
         self._redraw(annotations=True)
-        self._show_vline(self.mne.vline.get_xdata())
+        if self.mne.vline_visible:
+            self._show_vline(self.mne.vline.get_xdata()[0])
         if self.mne.fig_selection is not None:
             # Show all radio buttons as selected when in butterfly mode
             fig = self.mne.fig_selection
@@ -1754,6 +1758,7 @@ class MNEBrowseFigure(MNEFigure):
             self.draw_artist(artist)
         self.canvas.blit()
         self.canvas.flush_events()
+        self.mne.vline_visible = True
 
     def _hide_vline(self):
         """Hide the vertical line."""
@@ -1764,6 +1769,7 @@ class MNEBrowseFigure(MNEFigure):
             self.draw_artist(artist)
         self.canvas.blit()
         self.canvas.flush_events()
+        self.mne.vline_visible = False
 
 
 def _figure(toolbar=True, FigureClass=MNEFigure, **kwargs):
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -281,9 +281,9 @@ class MNEBrowseFigure(MNEFigure):
         self.mne.hscroll_annotations = list()
         self.mne.annotation_segments = list()
         self.mne.annotation_texts = list()
-        self.mne.added_labels = list()
-        self.mne.segment_colors = dict()
-        self.mne.segment_line = None
+        self.mne.new_annotation_labels = list()
+        self.mne.annotation_segment_colors = dict()
+        self.mne.annotation_hover_line = None
         self.mne.draggable_annotations = False
         # scalings
         self.mne.scale_factor = 0.5 if self.mne.butterfly else 1.
@@ -458,7 +458,7 @@ class MNEBrowseFigure(MNEFigure):
                 event.inaxes != self.mne.ax_main):
             return
         if not self.mne.draggable_annotations:
-            self._remove_annotation_line()
+            self._remove_annotation_hover_line()
             return
         from matplotlib.patheffects import Stroke, Normal
         for coll in self.mne.annotations:
@@ -478,24 +478,26 @@ class MNEBrowseFigure(MNEFigure):
                     path.vertices[mask, 0] = x0
 
                 # create or update the DraggableLine
-                if self.mne.segment_line is None:
+                hover_line = self.mne.annotation_hover_line
+                if hover_line is None:
                     line = self.mne.ax_main.plot([x, x], ylim, color=color,
                                                  linewidth=2, pickradius=5.)[0]
-                    self.mne.segment_line = DraggableLine(
+                    hover_line = DraggableLine(
                         line, self._modify_annotation, drag_callback)
                 else:
-                    self.mne.segment_line.set_x(x)
-                    self.mne.segment_line.drag_callback = drag_callback
+                    hover_line.set_x(x)
+                    hover_line.drag_callback = drag_callback
                 # style the line
-                line = self.mne.segment_line.line
+                line = hover_line.line
                 patheff = [Stroke(linewidth=4, foreground=color, alpha=0.5),
                            Normal()]
                 line.set_path_effects(patheff if line.contains(event)[0] else
                                       patheff[1:])
                 self.mne.ax_main.selector.active = False
+                self.mne.annotation_hover_line = hover_line
                 self.canvas.draw_idle()
                 return
-        self._remove_annotation_line()
+        self._remove_annotation_hover_line()
 
     def _keypress(self, event):
         """Handle keypress events."""
@@ -666,7 +668,7 @@ class MNEBrowseFigure(MNEFigure):
                     end = start + raw.annotations.duration
                     ann_idx = np.where((xdata > start) & (xdata < end))[0]
                     raw.annotations.delete(ann_idx)  # only first one deleted
-                self._remove_annotation_line()
+                self._remove_annotation_hover_line()
                 self._draw_annotations()
                 self.canvas.draw_idle()
             elif event.inaxes == ax:  # hide green line
@@ -946,7 +948,7 @@ class MNEBrowseFigure(MNEFigure):
         ax = fig.mne.radio_ax
         # get all the labels
         labels = list(set(self.mne.inst.annotations.description))
-        labels = np.union1d(labels, self.mne.added_labels)
+        labels = np.union1d(labels, self.mne.new_annotation_labels)
         # compute new figsize
         width, var_height, fixed_height, pad = \
             self._compute_annotation_figsize(len(labels))
@@ -967,7 +969,8 @@ class MNEBrowseFigure(MNEFigure):
             center = ax.transData.inverted().transform(
                 ax.transAxes.transform((0.1, 0)))
             circle.set_center((center[0], circle.center[1]))
-            circle.set_edgecolor(self.mne.segment_colors[label.get_text()])
+            circle.set_edgecolor(
+                self.mne.annotation_segment_colors[label.get_text()])
             circle.set_linewidth(4)
             circle.set_radius(radius / len(labels))
         # style the selected button
@@ -1015,7 +1018,7 @@ class MNEBrowseFigure(MNEFigure):
     def _add_annotation_label(self, event):
         """Add new annotation description."""
         text = self.mne.fig_annotation.label.get_text()
-        self.mne.added_labels.append(text)
+        self.mne.new_annotation_labels.append(text)
         self._setup_annotation_colors()
         self._update_annotation_fig()
         # automatically activate new label's radio button
@@ -1033,11 +1036,11 @@ class MNEBrowseFigure(MNEFigure):
 
         # TODO disable for epochs/ica instance types
         raw = self.mne.inst
-        segment_colors = getattr(self.mne, 'segment_colors', dict())
+        segment_colors = getattr(self.mne, 'annotation_segment_colors', dict())
         # sort the segments by start time
         ann_order = raw.annotations.onset.argsort(axis=0)
         descriptions = raw.annotations.description[ann_order]
-        color_keys = np.union1d(descriptions, self.mne.added_labels)
+        color_keys = np.union1d(descriptions, self.mne.new_annotation_labels)
         colors, red = _get_color_list(annotations=True)
         color_cycle = cycle(colors)
         for key, color in segment_colors.items():
@@ -1051,7 +1054,7 @@ class MNEBrowseFigure(MNEFigure):
                 segment_colors[key] = red
             else:
                 segment_colors[key] = next(color_cycle)
-        self.mne.segment_colors = segment_colors
+        self.mne.annotation_segment_colors = segment_colors
 
     def _select_annotation_span(self, vmin, vmax):
         """Handle annotation span selector."""
@@ -1065,11 +1068,11 @@ class MNEBrowseFigure(MNEFigure):
         self._draw_annotations()
         self.canvas.draw_idle()
 
-    def _remove_annotation_line(self):
+    def _remove_annotation_hover_line(self):
         """Remove annotation line from the plot and reactivate selector."""
-        if self.mne.segment_line is not None:
-            self.mne.segment_line.remove()
-            self.mne.segment_line = None
+        if self.mne.annotation_hover_line is not None:
+            self.mne.annotation_hover_line.remove()
+            self.mne.annotation_hover_line = None
             self.mne.ax_main.selector.active = True
             self.canvas.draw()
 
@@ -1097,7 +1100,7 @@ class MNEBrowseFigure(MNEFigure):
                            annotations.description[ann_idx],
                            annotations, ann_idx)
         self._draw_annotations()
-        self._remove_annotation_line()
+        self._remove_annotation_hover_line()
         self.canvas.draw_idle()
 
     def _clear_annotations(self):
@@ -1122,7 +1125,7 @@ class MNEBrowseFigure(MNEFigure):
         ylim = ax.get_ylim()
         for idx, (start, end) in enumerate(segments):
             descr = self.mne.inst.annotations.description[idx]
-            segment_color = self.mne.segment_colors[descr]
+            segment_color = self.mne.annotation_segment_colors[descr]
             kwargs = dict(color=segment_color, alpha=0.3)
             # draw all segments on ax_hscroll
             annot = self.mne.ax_hscroll.fill_betweenx((0, 1), start, end,
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -603,7 +603,10 @@ class MNEBrowseFigure(MNEFigure):
         if isinstance(event.artist, Text):
             ch_name = event.artist.get_text()
             ind = self.mne.ch_names[self.mne.picks].tolist().index(ch_name)
-            self._toggle_bad_channel(ind)
+            if event.mouseevent.button == 1:  # left click
+                self._toggle_bad_channel(ind)
+            elif event.mouseevent.button == 3:  # right click
+                self._create_ch_location_fig(ind)
 
     def _buttonpress(self, event):
         """Handle mouse clicks."""
@@ -688,6 +691,30 @@ class MNEBrowseFigure(MNEFigure):
         # redraw
         self._update_projector()
         self._redraw()
+
+    def _create_ch_location_fig(self, idx):
+        """Show channel location figure; idx is index of *visible* channels."""
+        from .utils import _channel_type_prettyprint
+        pick = self.mne.picks[idx]
+        ch_name = self.mne.ch_names[pick]
+        ch_type = self.mne.ch_types[pick]
+        if ch_type not in _DATA_CH_TYPES_SPLIT:
+            return
+        # create figure and axes
+        fig = self._new_child_figure(figsize=(4, 4),
+                                     fig_name='fig_help',
+                                     window_title=f'Location of {ch_name}')
+        ax = fig.add_subplot()
+        title = (f'{_channel_type_prettyprint[ch_type]} positions '
+                 f'({ch_name} highlighted)')
+        plot_sensors(self.mne.info, ch_type=ch_type, axes=ax, title=title,
+                     kind='select')
+        inds = np.in1d(fig.lasso.ch_names, [ch_name])
+        fig.lasso.disconnect()
+        fig.lasso.alpha_other = 0.3
+        fig.lasso.linewidth_selected = 2
+        fig.lasso.style_sensors(inds)
+        fig.mne.parent_fig.mne.foo = ax.collections[0]
 
     def _toggle_butterfly(self):
         """Enter or leave butterfly mode."""
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -1429,7 +1429,6 @@ class MNEBrowseFigure(MNEFigure):
 
     def _update_yaxis_labels(self):
         """Change the y-axis labels."""
-        fontstyle = 'italic' if self.mne.use_noise_cov else 'normal'
         if self.mne.butterfly and self.mne.fig_selection is not None:
             exclude = ('Vertex', 'Custom')
             ticklabels = list(self.mne.ch_selections)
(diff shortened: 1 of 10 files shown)


mne/viz/_figure.py
```diff
@@ -0,0 +1,430 @@
...
+    """Interactive figure with scrollbars, for data browsing."""
+    def __init__(self, inst, *args, xlabel='Time (s)', show_scrollbars=True,
+                 **kwargs):
+        from matplotlib.widgets import Button
+        from mpl_toolkits.axes_grid1.axes_size import Fixed
+        from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
+        from mne.viz.utils import _get_figsize_from_config
+
+        # figsize is the first arg of Figure()
+        figsize = args.pop(0) if len(args) else kwargs.pop('figsize', None)
```
jona-sassenhagen, line 61: just `if args`?

mne/viz/_figure.py
```diff
@@ -0,0 +1,430 @@
...
+        if figsize is None:
+            figsize = _get_figsize_from_config()
+        kwargs['figsize'] = figsize
```
jona-sassenhagen, line 64:
```suggestion
        kwargs['figsize'] = figsize or _get_figsize_from_config()
```

is also always an option

cbrnr:
I haven't seen any discussion on how to implement dark mode - what's the plan? Add a `theme` argument? Or just a boolean dark/light mode switch? I'm really looking forward to this feature, that way MNELAB can go all auto-dark (at least on macOS for now).

drammock:
The plan is to not get in the way of existing matplotlib theme functionality. In other words, use rcparams entries for things like foreground and background, and references to the color cycle like 'C2' for accent color needs.

drammock:
this is ready for preliminary testing / review from anyone interested (@cbrnr?  @hoechenberger?).  For ease of side-by-side comparison, the new plotting function is implemented as `mne.viz.plot_raw_alt(raw, ...)` so you can pop open one of those plus a regular `raw.plot(...)` window at the same time. Here is a snippet to get you started, that illustrates a lot of the functionality all at once:

```python
import os
import numpy as np
import mne
plt.ion()
                                    'sample_audvis_raw.fif')
raw.apply_proj()
ecg_proj_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                             'sample_audvis_ecg-proj.fif')
ecg_projs = mne.read_proj(ecg_proj_file)
raw.add_proj(ecg_projs)
sample_data_noise_cov_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                                          'ernoise-cov.fif')
noise_cov = mne.read_cov(sample_data_noise_cov_file)
event_dict = {'auditory/left': 1, 'auditory/right': 2}
event_color = {1: 'r', 2: 'g', -1: 'y'}
annotations = mne.Annotations(onset=(2, 3, 5), duration=(0.5, 2.5, 1),
                              description=('foo', 'bar', 'bad_baz'))
raw.set_annotations(annotations)
fig = mne.viz.raw.plot_raw_alt(
    raw, proj=True, events=events, event_id=event_dict,
    event_color=event_color, noise_cov=noise_cov,
    group_by='selection', butterfly=False, decim=2,
    show_options=True)
```

Known deficiencies:

1. data traces are now plotted with a matplotlib `LineCollection` which means the z-order of bad channels is not all the way at the back.  Probably the right thing to do is exclude bads from the LineCollection and plot them separately; just haven't done that yet. Other suggestions welcome.
2. clipping is implemented as a masked array rather than as `np.clip()` (for the "clamp" case) or as rectangular clipboxes (for other cases).  In principle it should be possible to still do it the old way even with a `LineCollection`, if people think the masked array looks bad.
3. not yet implemented: new feature of right-clicking to show channel location
4. not yet implemented: new feature to click-drag in non-annotation mode to get a pop-up spectrum plot

cbrnr:
@drammock I'd love to test this next week (I'm currently on vacation).

hoechenberger:
@drammock Will taste later today!

hoechenberger:
@drammock So you switched clamping back to using np.clip instead of masked arrays? Why's that?

drammock:
Because I realized that "clamp" (as in voltage clamp) had a particular meaning and users would expect to see flat lines (rather than empty space) where the signal exceeded its limits. I still use masked arrays (instead of a clipping mask) for the other case.

agramfort:
tested locally. Works great !

cbrnr:
Looks really good! Here are a few differences I've noticed, not sure if they are intentional or not, but they're maybe worth discussing:

1. The path in the window title bar now gets shortened rather abruptly (in my case `...ts/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif`). I don't know if I like that more than showing the full path. A third option would be to shorten in a more sensible manner, e.g. either make sure that we don't shorten folder names (`.../mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif` or maybe `/Users/clemens/.../sample/sample_audvis_raw.fif`). Finally, we could also just show the file name with not path at all.
2. Clicking between traces should bring up a vertical line with the associated time shown in the lower left corner. This line changed from green to yellow. Also, it is very difficult to click between traces, especially if there are many channels - this leads to channels being marked/unmarked as bad. Note that this is also true for the old plotting function, but I still think this is not ideal. We should either make sure that clicking outside traces sets the green line, or limit marking as bad to clicking on channel labels only. Or maybe setting the green line should require a keypress before clicking.
3. Event labels are now shown in the corresponding event colors instead of black. This looks nice in the example, but it might lead to labels that are difficult to read if the event color is e.g. yellow. Personally, I like the new (colored) version better.
4. The font size of event and annotation labels is now larger. I feel that with the old smaller font size annotations were sometimes easier to read because they didn't overlap that much.
5. The description in the annotation window should mention that left-clicking and dragging existing annotations works only if snap mode (p) is on. Maybe the current snap mode status should also be shown somewhere in the annotation window.
6. I'm sure there is a reason why the scale bar (s) doesn't work in your example, but this is not obvious and maybe should be explained somewhere.
7. The new plotting function prints `STI 014` in italic - does this bear any meaning?
8. I noticed that `STI 014` is always shown at the bottom (also in the old function) - is this intentional? In another example file where the STIM channel is called "Status" this behavior doesn't work (i.e. the STIM channel is not always shown).
9. The checkboxes in the projector window should always be squares and not rectangles depending on the window size.

drammock:
>     1. The path in the window title bar now gets shortened rather abruptly (in my case `...ts/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif`).

This truncation to 60 characters was previously only done when `len(raw._filenames) > 1`.   I couldn't figure out why you would want to truncate a path that said `/really/long/path/to/some/file.fif (+ 3 more)` but not truncate one that said `/really/really/long/path/to/a/single/file.fif` so I altered the logic a bit.  Based on your feedback I've now implemented a function to do "clean" truncation that omits full path elements at a time, starting from the middle and working outwards, and never omitting the rightmost element (the filename).  It doesn't guarantee you'll end up with something under 60 characters anymore, but if the filename alone is that long maybe it's OK to handle it ungracefully.

>     2. Clicking between traces should bring up a vertical line with the associated time shown in the lower left corner. This line changed from green to yellow.

The yellowish was an early decision to go with the matplotlib cycler colors for everything (in hopes of making dark mode work via `plt.style.use`). But using cycler colors seems unlikely to happen (see #8217) so I've changed the color back.

> it is very difficult to click between traces, especially if there are many channels - this leads to channels being marked/unmarked as bad. Note that this is also true for the old plotting function, but I still think this is not ideal. We should either make sure that clicking outside traces sets the green line, or limit marking as bad to clicking on channel labels only. Or maybe setting the green line should require a keypress before clicking.

Personally I'm in favor of marking bads *only* when clicking channel names (not data traces).  But this is a UI change that might upset many users.  Note that getting the "hit" logic correct for the traces is surprisingly tricky; MPL provides a `pickradius` argument in points, but in my testing it's pretty hard to get the "right" value (what is "right" seems dependent on window size and how noisy the data is).  For now I'll probably not fix this, since it's not really a regression.

> Event labels are now shown in the corresponding event colors instead of black. This looks nice in the example, but it might lead to labels that are difficult to read if the event color is e.g. yellow. Personally, I like the new (colored) version better.

I like the colored version too, but I've reverted to black for now (it was not an intentional change).

> The font size of event and annotation labels is now larger.

This was an unintentional change and has been fixed.

> The description in the annotation window should mention that left-clicking and dragging existing annotations works only if snap mode (p) is on. Maybe the current snap mode status should also be shown somewhere in the annotation window.

Good idea, will do.

> I'm sure there is a reason why the scale bar (s) doesn't work in your example, but this is not obvious and maybe should be explained somewhere.

That was a result of testing cruft (I had briefly commented out the lines that handle it).  Fixed now.

>  The new plotting function prints `STI 014` in italic - does this bear any meaning?

Italicized channel names is done when channels are whitened (toggle with `w` or disable by not passing in `noise_cov`).  But the STI channels should *not* be italicized because they never get whitened. Fixed now.

> I noticed that `STI 014` is always shown at the bottom (also in the old function)

This is a "feature" of `group_by='selection'`.  If stim channels called "status" don't work here, the fix would be to `utils/config.py:_get_stim_channel()`

> The checkboxes in the projector window should always be squares and not rectangles depending on the window size.

Surprisingly tricky. Working on it.

cbrnr:
Regarding the green line, what about continuously reporting the time under the mouse cursor somewhere in the plot or in a separate (toggle-able statusbar)? This would make the green line unnecessary (not that anyone would notice because it's so hard to get it to appear in the first place).

drammock:
@larsoner this may be close to ready for your eyes.  Summary of where things stand:

- `group_by='selection'` looks nicer and works a little smoother IMO
- new functionality: right-click on channel name -> show channel location popup
- traces are drawn with LineCollections, which seems to be a *little* faster but not much.
    - this required doing clipping as a masked array, which looks worse than proper clipping. I tried to clip the LineCollection with a PolyCollection as a clipping path (didn't work)... maybe could convert a PolyCollection to a compound path? Compound paths are possible, but [the only example](https://matplotlib.org/3.3.1/gallery/shapes_and_collections/compound_path.html) shows them being made with drawing primitives.
    - I tried to speed up butterfly mode by doing it with `imshow`.  It was fairly fast (I think faster than the LineCollection) as long as data points weren't connected, but this looked bad (a sort of "cloud of data points" rather than collection of traces, and bad channels basically disappeared into the background); as soon as I connected the dots (with a scikit-image drawing function) it looked good but took forever. It was also hard to get clipping to work right with that approach.  Code for this is local at the moment (not in this PR).
- the green vertical line is done with blitting, to prevent having to redraw all data traces each time (was especially laggy in butterfly mode). Annotation is still very laggy with butterfly mode; haven't tried to do it with blitting yet.

Do you have any suggestions / ideas / advice about which way might be most promising to go on these issues?

drammock:
I was able to create a compound path to mask the `LineCollection`, but it looks like I'll need a separate `LineCollection` for each channel type (in butterfly mode) and a separate one for each channel in regular mode (in which case, we're back to using `Line2D` :facepalm:)

Screenshot_2020-09-23_11-10-39

That is implemented on `drammock/mne-python:clipboxes`

FWIW here's what the `imshow` version of butterfly mode looks like (implemented on `drammock/mne-python:butterfly-image`)

butterfly-image

I'm beginning to think that the *only* way to significantly speed this up is to use blitting for *everything*.  In which case I suggest moving forward with reviewing/merging this one after I fix the clipping and clean up the cruft.

larsoner:
> I'm beginning to think that the only way to significantly speed this up is to use blitting for everything.

Could be. So maybe go back to Line2D and we'll try to speed up by blitting only the things we need to blit?

drammock:
> > I'm beginning to think that the only way to significantly speed this up is to use blitting for everything.
> 
> Could be. So maybe go back to Line2D and we'll try to speed up by blitting only the things we need to blit?

I've reverted the `LineCollection` stuff now, so I think this is ready for a review.  Here is what remains undone (but I think each of these deserves its own PR):

- use blitting to speed up plotting (current PR only uses blitting on the green vertical line)
- new feature: click-drag (when not annotating) to pop open a spectral plot for that span of data
- UX change: there's been talk of removing the "click data to mark bad" because it can be hard to *not* click data if you want a vertical marker line, and because we already have a way to do (un)mark bads by clicking channel names.  I expect this to be controversial and I think so far only @cbrnr and I have weighed in (we're both in favor of the change)

larsoner:
>  so I think this is ready for a review.

Want to get CIs happy first, then I'll look?

drammock:
@larsoner last CI run everything passed except coverage and one Travis run.  I've just pushed some commits that should help coverage.  The travis failures were all `AttributeError: 'NoneType' object has no attribute 'mpl_connect'` and seem to suggest that figures are not getting created properly.  Those tests are passing locally for me, so I'm not sure what I'm doing wrong. Any insight?

larsoner:
These errors look legitimate (tick numbers and BAD_ vs BAD):

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=9316&view=logs&jobId=1d294aa9-8c9e-52da-e2cc-28ca059e1c9b&j=1d294aa9-8c9e-52da-e2cc-28ca059e1c9b&t=520de7d9-efc8-52cf-0de1-994cbfe115fe

For the mpl_connect thing, maybe this:

https://stackoverflow.com/questions/31933393/event-handling-matplotlibs-figure-and-pyplots-figure

But if it's happening on our minimal/old deps build, I would try replicating that environment locally. Then you should see the same error. It could be it's a matplotlib bug that has been fixed and maybe we need to either `skip` on old mpl or bump our matplotlib requirement.

drammock:
> These errors look legitimate (tick numbers and BAD_ vs BAD):
> 
> https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=9316&view=logs&jobId=1d294aa9-8c9e-52da-e2cc-28ca059e1c9b&j=1d294aa9-8c9e-52da-e2cc-28ca059e1c9b&t=520de7d9-efc8-52cf-0de1-994cbfe115fe

Argh, that's a new one I just introduced.

> But if it's happening on our minimal/old deps build, I would try replicating that environment locally. Then you should see the same error.

Yep, now I get that locally.

> It could be it's a matplotlib bug that has been fixed and maybe we need to either `skip` on old mpl or bump our matplotlib requirement.

Will see if I can find out when this behavior changed in MPL and/or how to work around it with old MPL.

hoechenberger:
Thanks @drammock for doing this amazing work! And sorry for being so late providing feedback.

I just gave the Raw browser a shot. Here's what I've found

- looks so much better now than before 😍😍😍
- toggling draggable Annotations mode via `p` doesn't seem to work for me
- when in Annotations mode (`a`) and enabling draggable edges, hovering the mouse curser over an edge and then pressing `a` to exit Annotations mode leaves a highlighted edge behind:

    Screenshot 2020-09-26 at 00 07 09

- I was irritated that closing the channel selection window would close the entire Raw browser, this seems unexpected for a user
- I find it difficult to easily see at a quick glance which channels have been selected in the channel browser. Fill colors, edges, or size of the circles should be changed to improve accessibility:

    Screenshot 2020-09-26 at 00 04 20

- Same goes for right-clicking a channel to view its location: it takes me a fraction of a second too long to find the channel on the plot:

    Screenshot 2020-09-26 at 00 08 31

- I tried for an entire minute but I couldn't get "left-click-and-drag to view the PSD plot" to work. All that happened was either nothing or I would toggle channels between bad and not bad. What am I doing wrong?

- With all these new windows, we're becoming a little bit like that image manipulation software with the weird name in its earlier days. I wonder: would it be feasible to assign a fixes area on the left side to always keep the relevant tools visible? Or to at least open the windows such that they are arranged (and sized) like here (but that might not be possible via Matplotlib?):

    Screenshot 2020-09-26 at 00 21 31

- The `show_options` kwarg should probably be renamed to better match what it actually does?

- There should be a keyboard shortcut (that's also listed in the Help window) to toggle the SSP Projection Vectors window

- The `Help` and `Prj` buttons do not change color when hovering over them; the `Toggle all` button in the SSP Projection Vectors window does (which is better IMHO)

hoechenberger:
I would also suggest to change the Prj button label to "SSP", which should be less ambiguous. WDYT?

hoechenberger:
I was also thinking that one could dedicated a small strip at the top of the window for something resembling a toolbar. This should be fairly easy to do and I believe it would up UX a big deal. The Help and Prj buttons could be moved there as well. I can do a quick mock-up tomorrow if it's unclear what I mean.

hoechenberger:
> We could even use icons, each of them in their own axes and with an on_clicked event callback.

Generally I think using some nice icons can make for an overall "friendlier" UI and a better UX. Here's an example of a dialog box I recently built for something `mne-bids`-related. By adding the question mark icon, I think it's suddenly much better than a plain dialog box would be:

https://youtu.be/6L7jrcTUPGM?t=27

hoechenberger:
Is there any particular reason why channel names are italicized in the Raw browser? I always found sans-serif fonts and italic stylization a slightly odd combination… Nothing major of course, just wondered why you picked this?

drammock:
> I just gave the Raw browser a shot. Here's what I've found on macOS with `Qt5Agg` backend.
> 
> toggling draggable Annotations mode via `p` doesn't seem to work for me

That should work. Will investigate.

> when in Annotations mode (`a`) and enabling draggable edges, hovering the mouse curser over an edge and then pressing `a` to exit Annotations mode leaves a highlighted edge behind

That shouldn't happen. Will investigate.

> I was irritated that closing the channel selection window would close the entire Raw browser, this seems unexpected for a user

This replicates past behavior. I don't know why it was designed that way. I inferred that maybe the idea is (1) selection mode is determined when you first start up the browser, it can't easily be turned on/off once the plot is open, and (2) there isn't a way to re-spawn the selection interface once you close it, so it and the data browser should behave as a unit.

> I find it difficult to easily see at a quick glance which channels have been selected in the channel selection window. Fill colors, edges, or size of the circles should be changed to improve accessibility:

I already increased it from how it was on master. Will try to push it a little further.

> Same goes for right-clicking a channel to view its location: it takes me a fraction of a second too long to find the channel on the plot:

ibid.

> I tried for an entire minute but I couldn't get "left-click-and-drag to view the PSD plot" to work.

Not yet implemented. This will be done in a separate PR. I mentioned that above somewhere, but it's a long thread, I don't blame you for missing it.

> With all these new windows, we're becoming a little bit like that image manipulation software with the weird name in its earlier days. I wonder: would it be feasible to assign a fixed area on the left side to always keep the relevant tools visible?

`-1` to this suggestion.  Some users prefer to use the entire screen real-estate for the data traces, in order to see as many channels and as much time as possible at once and still have it be intelligible (when browsing through on a preliminary quality check of the recording).  That's also why we have zen mode and fullscreen support.  I don't disagree that the many windows is a bit unwieldy, but the alternative (eating data space to provide options GUI space) seems worse to me.  If I'm overruled, I will lobby hard to make it a separate PR, because it means re-thinking the layout of all those GUI elements (right now they're all different sizes/shapes, to make each as compact as is feasible).

> Or to at least open the windows such that they are arranged (and sized) like here (but that might not be possible via Matplotlib?):

It is possible to programmatically specify MPL window positions via `fig.canvas.manager.window.move(x,y)` (though I'm not positive that this works on all backends; trivially it won't work on notebook backends at least; some dated discussion [here](https://stackoverflow.com/q/7449585/1664024)).  So this may be technically possible, but given the variety of use cases I'm a bit wary of trying to find The One True Layout and impose it on everyone.  An alternative would be keeping track of window position (for data browser and the main child figures, but not channel position figs or future click-drag-spectrum figs) in the MNE config file (several variables of the form `MNE_BROWSE_RAW_PROJ_WINDOW_POSITION` or so... WDYT?)

> The `show_options` kwarg should probably be renamed to better match what it actually does?

What is the suggestion?  `show_proj_window`?  I don't object, though it would need a deprecation cycle.

> There should be a keyboard shortcut (that also gets listed in the Help window) to toggle the SSP Projection Vectors window

Makes sense.  What key?

> The `Help` and `Prj` buttons do not change color when hovering over them; the `Toggle all` button in the SSP Projection Vectors window does (which is better IMHO)

I don't know why this is happening, and I'm not knowingly doing anything different when creating those buttons. Maybe MPL bug?

> I would prefer a clear(er) indication about which interaction mode I'm currently in and which magic has been applied to the figure I'm viewing. Here, too, a panel docked on the left side could help.

Assuming that a panel won't happen, what would this look like and where would it go?

> `Qt5Agg` has a status bar on the bottom by default (but it seems we're disabling it). Couldn't we make use of it?

I hesitate to rely on backend-specific functionality, in part because we can't control what users use, and in part to keep the notebook and pop-up figures as close to feature-equivalent as possible.

> I would also suggest to change the Prj button label to "SSP", which should be less ambiguous. WDYT?

No strong feeling here.

> I was also thinking that one could dedicate a small strip at the top of the window for something resembling a toolbar. This should be fairly easy to do and I believe it would up UX a big deal. The Help and Prj buttons could be moved there as well. I can do a quick mock-up tomorrow if it's unclear what I mean. We could even use icons, each of them in their own axes and with an on_clicked event callback.

See above re: use of screen real-estate for data vs UI.  I don't object to a toolbar in principle, but keep in mind that this PR is first and foremost a `MAINT` PR.  I have thrown in some UI improvements because they were easy to do along the way, but it is not intended to be a full overhaul of the UI design. I think the right approach here is to collect your UX proposals into a separate `RFC` issue, and/or add them to #7751, to give people a chance to weigh in on them separately, and provide time to iterate through some mock-up proposals.

> Is there any particular reason why channel names are italicized in the Raw browser? I always found sans-serif fonts and italic stylization a slightly odd combination… Nothing major of course, just wondered why you picked this?

Channels that have been whitened are italicized. This preserves past behavior, so I didn't pick it :slightly_smiling_face:. Pressing `w` (when `noise_cov` was provided) should enable/disable whitening and also add/remove italics.  LMK if this isn't working properly for you.

drammock:
@larsoner everything is green except codecov.  I'll work on test coverage on Monday, up to you if you want to wait for that or look now.

cbrnr:
@drammock @hoechenberger I also thought about using the status bar for things like showing the time under the mouse pointer. That way we could even get rid of that green line. But whichever way we decide, this should definitely be tackled in a separate PR!

drammock:
> > I just gave the Raw browser a shot. Here's what I've found on macOS with `Qt5Agg` backend.
> > toggling draggable Annotations mode via `p` doesn't seem to work for me
> 
> That should work. Will investigate.

For me, on Linux and in an OSX VM (both using `qt5agg` backend), pressing `p` enables draggable annotation mode (and also checks/unchecks the "draggable annotations" checkbox in the annotations dialog window). What does not work is that in the OSX VM, the hover effect that highlights the edge of the annotation span doesn't work --- @hoechenberger can you try click-dragging an edge even when it's not highlighted?  If that works, any chance you'd be willing to do a little debugging on this?  Doing any serious work in an OSX VM is very slow for me, since I have all the wrong muscle memories.

> > when in Annotations mode (`a`) and enabling draggable edges, hovering the mouse curser over an edge and then pressing `a` to exit Annotations mode leaves a highlighted edge behind
> 
> That shouldn't happen. Will investigate.

Fixed.

> > I find it difficult to easily see at a quick glance which channels have been selected in the channel selection window. Fill colors, edges, or size of the circles should be changed to improve accessibility:
> 
> I already increased it from how it was on master. Will try to push it a little further.

Fixed.

> > Same goes for right-clicking a channel to view its location: it takes me a fraction of a second too long to find the channel on the plot:
> 
> ibid.

Fixed.

> > There should be a keyboard shortcut (that also gets listed in the Help window) to toggle the SSP Projection Vectors window
> 
> Makes sense. What key?

mapped to `j` now.

> > The `Help` and `Prj` buttons do not change color when hovering over them; the `Toggle all` button in the SSP Projection Vectors window does (which is better IMHO)
> 
> I don't know why this is happening, and I'm not knowingly doing anything different when creating those buttons. Maybe MPL bug?

Fixed (though annoyingly a bit laggy on my system, presumably because the hover effect is triggering a full canvas redraw :slightly_frowning_face: ). Apparently if you don't keep a reference to the buttons, then hover highlighting doesn't work?

mne/viz/tests/test_raw.py
```diff
@@ -331,19 +332,18 @@ def test_plot_raw_ssp_interaction():
...
     assert len(plt.get_fignums()) == 1
-    _child_fig_helper(fig, '?', 'fig_help')
+    _child_fig_helper(fig, '?', 'fig_help', new_mpl)
     print(fig.mne.child_figs)
-    _child_fig_helper(fig, 'j', 'fig_proj')
+    _child_fig_helper(fig, 'j', 'fig_proj', new_mpl)
     print(fig.mne.child_figs)
-    _child_fig_helper(fig, 'a', 'fig_annotation')
+    _child_fig_helper(fig, 'a', 'fig_annotation', new_mpl)
     print(fig.mne.child_figs)
```
drammock: oops, these `print` lines are cruft

drammock:
This one is ready for a careful code review, if anyone is up for it :slightly_smiling_face:  The one CI failure is CodeCov, and it's only 0.81% short of the desired 95% coverage target.

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+            if is_named:
+                setattr(self.mne.parent_fig.mne, self.mne.fig_name, None)
+
+    def _keypress(self, event):
+        """Handle keypress events."""
+        if event.key == self.mne.close_key:
+            from matplotlib.pyplot import close
+            close(self)
+        elif event.key == 'f11':  # full screen
+            self.canvas.manager.full_screen_toggle()
```
larsoner, line 58: Why not test this one? Does it throw an error on `Agg`? If not but it's a no-op at the mpl end, it's probably still worth "testing" it to hit the branch in our code.

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        w_or_h = fig_w if horiz else fig_h
+        return dim_inches / w_or_h
+
+
+class MNEAnnotationFigure(MNEFigure):
+    """Interactive dialog figure for annotations."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
```
larsoner, line 116:
```suggestion
```
I don't think these are needed. AFAICT all it does is ensure all are `kwargs` but `MNEFigure` already ensures this.

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        color = list(buttons.circles[idx].get_edgecolor())
+        color[-1] = 0.5
+        buttons.circles[idx].set_facecolor(color)
+        self.canvas.draw()
+
+
+class MNESelectionFigure(MNEFigure):
+    """Interactive dialog figure for channel selections."""
+
+    def __init__(self, **kwargs):
```
larsoner, line 195: same

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        self.mne.fig_annotation = None
+
+        # MAIN AXES: default sizes (inches)
+        l_margin = 1.
+        r_margin = 0.1
+        b_margin = 0.45
+        t_margin = 0.25
+        scroll_width = 0.25
+        hscroll_dist = 0.25
+        vscroll_dist = 0.1
```
larsoner, line 312: matplotlib 2.2 has `constrained_layout=True`. Maybe add a comment that we might be able to simplify this stuff at some point by bumping our min and using it?

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+    def _close(self, event):
+        """Handle close events (via keypress or window [x])."""
+        from matplotlib.pyplot import close
+        # write temporary info object back to instance (for bads, but don't
+        # persist changes to proj checkboxes)
+        self.mne.info['projs'] = self.mne.inst.info['projs']
+        self.mne.inst.info = self.mne.info
+        # Clean up child figures too
+        while len(self.mne.child_figs):
+            fig = self.mne.child_figs[-1]
```
larsoner, line 431: This looks like a bug and that it should be `self.mne.child_figs.pop(-1)` until looking deeper into the code. Can you add a comment here that the child figures all remove themselves from the parent figure lists in `_close` so need to `pop`?

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+            close(fig)
+
+    def _resize(self, event):
+        """Handle resize event for mne_browse-style plots (Raw/Epochs/ICA)."""
+        size = ','.join(self.get_size_inches().astype(str))
+        set_config('MNE_BROWSE_RAW_SIZE', size, set_env=False)
```
larsoner, line 437: Rather than spamming / writing config during resize, we should probably just do it once during the close event. WDYT?

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        fig._add_default_callbacks()
+        self.mne.child_figs.append(fig)
+        if isinstance(fig_name, str):
+            setattr(self.mne, fig_name, fig)
+        return fig
+
+    def _create_ch_location_fig(self, idx):
+        """Show channel location figure; idx is index of *visible* channels."""
+        pass
+        # from .utils import _channel_type_prettyprint
```
larsoner, line 689: Some comment here about a todo/what's missing/why not implemented?

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        width = 1 - 2 * left
+        height = 1 - 2 * bottom
+        fig.mne.radio_ax = fig.add_axes((left, bottom, width, height),
+                                        frame_on=False, aspect='equal')
+        div = make_axes_locatable(fig.mne.radio_ax)
+        self._update_annotation_fig()  # populate w/ radio buttons & labels
+        # append instructions at top
+        instructions_ax = div.append_axes(position='top', size=Fixed(1),
+                                          pad=Fixed(5 * pad))
+        # XXX when we support a newer matplotlib (3.0?) the instructions can
```
larsoner, line 845: 3.0.0 came out Sept 18 2018, and we're going to release next in March 2021. So it will have been out for 2.5 years then, which seems within our limit (I think we go with 2-ish years?). So if it makes things better we should consider upping our minimum, but maybe in another PR.

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+               self.mne.fig_annotation.mne.radio_ax.buttons.labels].index(text)
+        self.mne.fig_annotation._set_active_button(idx)
+        # simulate a click on the radiobutton → update the span selector color
+        self.mne.fig_annotation._radiopress(event=None)
+        # reset the text entry box's text
+        self.mne.fig_annotation.label.set_text('BAD_')
+
+    def _setup_annotation_colors(self):
+        """Set up colors for annotations."""
+        from itertools import cycle
```
larsoner, line 1010: no need to nest this, it's a builtin

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+        fig.mne.proj_all.on_clicked(partial(self._toggle_proj_checkbox,
+                                            toggle_all=True))
+        # save params
+        fig.mne.proj_checkboxes = checkboxes
+        # show figure (this should work for non-test cases)
+        try:
+            self.mne.fig_proj.canvas.draw()
+            plt_show(fig=self.mne.fig_proj, warn=False)
+        except Exception:
+            pass
```
larsoner, line 1315: Do we still need this try/except? I think I recently pushed some fixes to our `plt_show` that might make it so you don't...?

mne/viz/_figure.py
```diff
@@ -0,0 +1,1854 @@
...
+                           self.mne.ch_start + self.mne.n_channels)
+            self.mne.picks = self.mne.ch_order[_slice]
+            self.mne.n_channels = len(self.mne.picks)
+
+    def _load_data(self, start=None, stop=None):
+        """Retrieve the bit of data we need for plotting."""
+        if self.mne.instance_type == 'raw':
+            return self.mne.inst[:, start:stop]
+        else:
+            raise NotImplementedError  # TODO: support epochs, ICA
```
larsoner, line 1577: The next follow-up PRs I hope? :)

mne/viz/raw.py
```diff
@@ -317,261 +269,142 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
...
+                  ch_color_dict=color,
+                  # display
+                  butterfly=butterfly,
+                  clipping=clipping,
+                  scrollbars_visible=show_scrollbars,
+                  scalebars_visible=show_scalebars,
+                  window_title=title)
+
+    fig = _browse_figure(**params)
+    fig._post_init()  # XXX fix for older matplotlib
```
larsoner, line 381: more evidence a version bump would be nice

mne/viz/utils.py
```diff
@@ -1028,8 +1028,6 @@ def _plot_raw_onkey(event, params):
             _setup_annotation_fig(params)
         else:
             params['fig_annotation'].canvas.close_event()
-    elif event.key == 'b':
-        _setup_butterfly(params)
```
larsoner, line 1032: So butterfly no longer works for Epochs / ICA, right? I guess we can live with it temporarily since those would be the next PRs

mne/viz/utils.py
```diff
@@ -3285,6 +3222,26 @@ def _set_window_title(fig, title):
         fig.canvas.manager.set_window_title(title)
 
 
+def shorten_path_from_middle(path, max_len=60, replacement='...'):
```
larsoner, line 3225:
```suggestion
def _shorten_path_from_middle(path, max_len=60, replacement='...'):
```

larsoner (commented):
Interaction seems faster, which is awesome! However it looks like the channel order is different grad/mag/EEG on master compared to mag/grad/EEG on this PR, e.g. with:
```
$ mne browse_raw ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif 
```
Is this intentional?

mne/viz/_figure.py
drammock: It is tested in `test_plot_raw_keypresses`:

https://github.com/drammock/mne-python/blob/18de83aa6b291cf97db26572506a80078c7ebb0e/mne/viz/tests/test_raw.py#L365-L366

I don't know why Codecov thinks it's not tested.

mne/viz/_figure.py
drammock: not sure, let's find out :)

mne/viz/_figure.py
drammock: yep

larsoner:
I think it might be better to keep the raw order rather than the epochs order. I assume it's more widely used.

drammock:
> I think it might be better to keep the raw order rather than the epochs order. I assume it's more widely used.

Are you sure?  It seems really strange to me that `raw.plot()` would do something different than `_DATA_CH_TYPES_ORDER_DEFAULT`.  I was assuming that you and Stefan had left that note because the old order was considered a minor inconsistency bug.

mne/viz/utils.py
drammock: it never worked for ICA (`_setup_butterfly` is a no-op if it's an ICA object), and Epochs has its own keypress function.  So this isn't even a temporary loss of function

drammock: (I did this to allow getting rid of `_setup_butterfly_params` entirely, which is otherwise no longer needed)

larsoner:
> Are you sure? It seems really strange to me that raw.plot() would do something different than _DATA_CH_TYPES_ORDER_DEFAULT. I was assuming that you and Stefan had left that note because the old order was considered a minor inconsistency bug.

Agreed the internal consistency is not great, but I'd prioritize the raw plotting consistency over our internal consistency here. And then I'd have the `epochs.plot` conform / change to what we do with raw, since I assume the raw plotting is used (a lot) more often.

FWIW we unfortunately aren't all that internally consistent. `evoked.plot` for the sample data plots in the order EEG/grad/mag for example.  I don't see a good reason for `evoked.plot` to use a different order, but it might break people's code to change that. `evoked.plot_topomap` does mags, but this is a better choice for that type of viz because we have to take the norm of the grads to plot them on a topomap (which is less "nice" than just plotting values directly).

drammock:
ok @larsoner I've restored channel type plotting order, and CIs are still happy (except codecov, which should improve once ICA uses the figure class, because old code that has unhit lines will get removed)

larsoner (approved):
LGTM +1 for merge

@agramfort do you want to look before merge?

agramfort:
looking now !

>

mne/viz/_figure.py
larsoner: See #8374 -- my vote would be to merge this PR then that one, to make it so you don't have to do another rebase here

drammock: ok with me.  just waiting on @agramfort to sign off here.

agramfort (approved):
amazing work @drammock !

hoechenberger:
🥳

drammock:
@agramfort did you forget to sqash-merge?

larsoner:
If so we should undo and force push a squashed version because of the number of commits

drammock:
Just force-pushed to master. :fearful:  please check to make sure I didn't do something wonky.

hoechenberger:
> Just force-pushed to master. :fearful:  please check to make sure I didn't do something wonky.

Can we make master a protected branch to prevent such accidents in the future?

larsoner:
The accident was actually pushing the green button the wrong way, and the fix was a force push that would not have been allowed under the protection rules. FWIW so far I've seen more harm / wasted time from protected branches than good on other repos.

What we could maybe do is only allow the squash and merge option for the green button on this repo. That would have prevented the issue and shouldn't add much if any maintenance overhead. Rebase or standard merge is used very rarely by us nowadays.

larsoner:
Okay just changed it so that squash merge is the only choice. About once a year (at most) I've wanted to do a different type (the highly collaborative beamforming PR is the only one I remember) and I'm okay with doing that manually from the command line or temporarily enabling another mode when needed.

PR #7957, opened 2020-07-03
mrg: Add scalar mult and div operators for AverageTFR

bloyl:
Add multiplication and division operators to `AverageTFR`

This enables computing group grand averages from lists using `np.mean`

### DIFF
--- a/mne/time_frequency/tfr.py
+++ b/mne/time_frequency/tfr.py
@@ -1931,18 +1931,12 @@ class AverageTFR(_BaseTFR):
         return self
 
     def __truediv__(self, a):  # noqa: D105
-        return self.__div__(a)
-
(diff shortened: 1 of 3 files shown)


mne/time_frequency/tfr.py
```diff
@@ -1930,6 +1930,32 @@ def __isub__(self, tfr):  # noqa: D105
         self.data -= tfr.data
         return self
 
+    def __truediv__(self, a):  # noqa: D105
+        return self.__div__(a)
+
+    def __div__(self, a):  # noqa: D105
```
larsoner, line 1936: I don't think we need `__div__` anymore now that we use Python3. I would remove this and put it all under `__truediv__`

larsoner:
Thanks @bloyl !

PR #7961, opened 2020-07-06
do not forget to pass adjacency

### LINKED ISSUE #7960: tutorial bug plot_stats_cluster_erp.py
#### Describe the bug
Tutorial plot_stats_cluster_erp.py does not use the electrode connectivity matrix on the spatiotemporal cluster test. Connectivity 'con' is computed in line 79, but it is not used in line 88.

agramfort:
closes #7960


hoechenberger:
Thanks @agramfort

Issue #7959, opened 2020-07-03
Feat Requests: pyvista - Brain/timeviewer


bloyl:
The pyvista, stc plotting is great and already really useful for reviewing data.

here are a few thoughts (in no particular order) I've had using it as part of an analysis pipeline for the last week.

1) there is no `brain.remove_labels()` which is mentioned in the `add_label` docs.
2) It would be useful to return vertices that are interactively added using `timeviewer`. This would let you save these points and later use them for analysis.
3) is it possible to link two `timeviewer`s so they share the same point picking and optionally the same camera/view. I wouldn't want them time synced (at least not by default) so you can look at different types of stcs.
4) I wonder what the opportunities are for adding labels and having time_viewer display their wave forms. Exposing the parameters for `extract_label_time_course` via the gui might be a pain though.

GuillaumeFavelier:
> The pyvista, stc plotting is great and already really useful for reviewing data.

Thanks for the nice feedback!

1) This feature was not on the top of the list but this issue can definitely increase the priority 
2) Good idea! We could create a function `get_picked_points()` or `get_picked_vertices()`, what do you say?
3) Only the time controllers can be linked at the moment (colorbar in the todo list). I can see the camera controls being shared easily and picking could be tricky (but possible!).
4) @agramfort  what do you think about this one?

I'll update #7162 with your comments and link the PRs to this issue too if you want :+1:

agramfort:
4. is what mne_analyze does historically so I think there is indeed a user
base for this. I can do you a demo

>

GuillaumeFavelier:
I wanted to work on 4) but I realized that `extract_label_time_course()` needs the `src` parameter (source spaces for left and right hemisphere) which is not available with `_Brain`. How can I fix this?

I had [this example](https://mne.tools/dev/auto_examples/visualization/plot_parcellation.html?highlight=plot_parcellation) in mind.

agramfort:
you should be able to extend extract_label_time_course allowing to
pass white surfaces instead of src.
for extract_label_time_course you need the locations and normals of
the vertices present in the stc.

larsoner:
> you should be able to extend extract_label_time_course allowing to
pass white surfaces instead of src.
> for extract_label_time_course you need the locations and normals of
the vertices present in the stc.

[...]

PR #7517, opened 2020-03-26
Update _TimeViewer's plotter only when necessary

GuillaumeFavelier:
This PR updates the render window only when it is expected or necessary (meaning after a user interaction) instead of being called constantly at a very small interval. I expect this to lower the burden on the CPU.

**Note**: My CPU usage for a standard script went from ~25-30% to around 1% on idle.

It's an item of #7162

PR #7962, opened 2020-07-06
MRG: Link cameras in _LinkViewer

GuillaumeFavelier:
This PR adds parameters to `link_brains()` to decide what to link: `time` controller and/or `camera` controls.

**ToDo**
- [x] Fix camera update delay
- [x] Update testing in `test_3d.py`
- [x] Rename the `time_viewer` used as reference into `leader`
- [x] Check stc time infos
- [x] Link time course canvas

It's an item of #7162 
Related to https://github.com/mne-tools/mne-python/issues/7959


agramfort:
@bloyl can you give it a try?

mne/viz/_3d.py
```diff
@@ -1517,13 +1517,17 @@ def _plot_mpl_stc(stc, subject=None, surface='inflated', hemi='lh',
...
     """Plot multiple SourceEstimate objects with PyVista.
 
     Parameters
     ----------
     brains : list, tuple or np.ndarray
         The collection of brains to plot.
+    time : bool
+        If True, link the time controllers. Defaults to True.
+    camera : bool
+        If True, link the camera controls. Defaults to True.
```
larsoner, line 1530:
```suggestion
        If True, link the camera controls. Defaults to False.
```
Or change it above, in which case `_LinkViewer.__init__` should also change

mne/viz/_brain/_timeviewer.py
```diff
@@ -1236,6 +1241,14 @@ def link_sliders(self, name, callback, event_type):
                     event_type=event_type
                 )
 
+    def link_cameras(self):
+        leader = self.time_viewers[0]  # select a time_viewer as leader
+        camera = leader.plotter.camera
+        print(camera)
```
larsoner, line 1247: don't forget to remove cruft

bloyl:
Sorry for the delay in testing this.

This is pretty cool and does what I'd like.

A couple of things I noticed.
1. The camera doesn't force an update on the other renderer so things get laggy. I assume this is what you mean by `Fix camera update delay` in the description. but thought I'd mention it.
2. The time linking only works on the slider. if you click in the time course plot only the one `brain` updates.
3. You are currently allowed to time link brains that have different time axis. but since the linking is on time `index` this causes things to get out of sync.

My suggestion for point 3 would be use the actual time as opposed to the index to do the linking. Additionally I would add a `time='auto'` option as the default. this would use `time=True` if the linked `brain` objects all have the time array and `false` if they don't, supporting both linked multiple evoked STCs and unlinked stcs (say evoked and a PSD) with the default. Additonally if a user had evoked stcs with different time axis that were actually comparable (maybe historic data or something) they could force `time=True` to still have the linking.

GuillaumeFavelier:
> The camera doesn't force an update on the other renderer so things get laggy.

The cameras are linked but the event "the camera has been modified" is not captured simply because I don't know how to do that. And since #7517, the plotter is not updated automatically. I see 2 ways to fix this. Either capture the vtk event (interactor, camera...) or allow automatic update again.

> The time linking only works on the slider

Nice catch! I'll take care of this :)

> You are currently allowed to time link brains that have different time axis.

Hm, this is not intended... I have to check the time axes somehow. @larsoner, @agramfort any idea on how to deal with this?

agramfort:
> Hm, this is not intended... I have to check the time axes somehow. @larsoner, @agramfort any idea on how to deal with this?

stc.times should match between linked brains

GuillaumeFavelier:
2) and 3) should be fixed by now. I'll investigate 1) and see if I can capture the right VTK event to update the plotters. Otherwise, I'll have to re-enable automatic update to mitigate.

GuillaumeFavelier:
It turns out, the camera `ModifiedEvent` does the trick! I invite you to try it out @bloyl :)

output

Reference: http://vtk.1045678.n5.nabble.com/Vtk-Camera-Postion-Event-td5727757.html

mne/viz/_brain/_timeviewer.py
```diff
@@ -1199,28 +1200,49 @@ def clean(self):
...
-            callback=self.set_time_point,
-            event_type="always"
-        )
+        # check time infos
+        brain_times = [brain._times for brain in brains]
+        if not all(len(x) == len(brain_times[0]) for x in brain_times):
+            time = False
+        if not all(all(np.equal(x, brain_times[0])) for x in brain_times):
+            time = False
+
```
bloyl, line 1213: This second check throws `ValueError: operands could not be broadcast` if the `brain_times` are of different lengths and more than one.

why not only do the check if you have to (`time == True`)

```
        if time and not all(len(x) == len(brain_times[0]) for x in brain_times):
            time = False
        if time and not all(all(np.equal(x, brain_times[0])) for x in brain_times):
            time = False
```

to test just link a plot of an evoked `stc` with a cropped copy of the same `stc`

bloyl:
Other than my comment above this is working for me.

Additional question, how did you make the 4 panel figure on the right?

GuillaumeFavelier:
> how did you make the 4 panel figure on the right?

I use the following parameters: `hemi='split', views=['lat', 'med']`

GuillaumeFavelier:
Feel free to have a look or even test it @agramfort, @larsoner

mne/viz/_brain/_timeviewer.py
```diff
@@ -1199,28 +1200,49 @@ def clean(self):
...
-            name="_time_slider",
-            callback=self.set_time_point,
-            event_type="always"
-        )
+        # check time infos
+        times = [brain._times for brain in brains]
+        if time and not all(len(x) == len(times[0]) for x in times):
+            time = False
+        if time and not all(all(np.equal(x, times[0])) for x in times):
+            time = False
```
larsoner, line 1212:
```suggestion
        if time and not all(np.allclose(x, times[0]) for x in times[1:]):
            warn('stc.times do not match, not linking time')
            time = False
```

GuillaumeFavelier: Why not testing the length?

larsoner: allclose will check shapes implicitly (assuming you don't have something pathological like 1. And [1., 1., 1.], Which would pass allclose but shouldn't occur for data at this level)

PR #7968, opened 2020-07-06
MRG, ENH: Speed up epochs.copy

### LINKED ISSUE #7947: Extremely slow access to epochs after channel remapping.
I have two epoch lists. In one of the lists all channels have to be mirrored to the other side of the hemisphere (flipped). After that I want to concatenate both lists to one epoch file. So I'm using mne.rename_channels() to remap the channels of the one list. After that I'm using mne.equalize_channels() to get a uniform channel mapping of the two lists. After that I'm using mne.concatenate_epochs() to concat the lists. It works perfectly fine but if I try to access a condition after the concatenation in the resulting epoch instance, things become really slow. Before the access is really quick like instant. After it takes up to 20s. How could I resolve that I tried a lot already. Thank you very much!

larsoner:
Closes #7947

On `master` the snippet [here](https://github.com/mne-tools/mne-python/issues/7947#issuecomment-653166786) I get:
```
Time untouched epochs:    0:00:00.014442
Time concatenated epochs: 0:00:00.457064
```
On this PR:
```
Time untouched epochs:    0:00:00.014059
Time concatenated epochs: 0:00:00.022924
```
So it's still ~2x slower, which I guess is to be expected because `selection` does grow by 2x. This should at least be hopefully acceptable compared to the ~30x slower it is on `master`.

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1919,8 +1919,10 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
     perc : float
         Total percentage of epochs dropped.
     """
-    if not isinstance(drop_log, tuple) or not isinstance(drop_log[0], tuple):
-        raise ValueError('drop_log must be a tuple of tuples')
+    if not isinstance(drop_log, tuple) or \
+            not all(isinstance(d, tuple) for d in drop_log) or \
+            not all(isinstance(s, str) for d in drop_log for s in d):
+        raise TypeError('drop_log must be a tuple of tuple of str')
     perc = 100 * np.mean([len(d) > 0 for d in drop_log
                           if not any(r in ignore for r in d)])
     return perc
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -17,10 +17,8 @@ import json
 import operator
 import os.path as op
 import warnings
-from distutils.version import LooseVersion
 
 import numpy as np
-import scipy
 
 from .io.write import (start_file, start_block, end_file, end_block,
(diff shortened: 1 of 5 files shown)


drammock:
relevant failure:

```
mne\viz\tests\test_epochs.py:272: in test_plot_drop_log
    plot_drop_log([['One'], [], []])
mne\viz\epochs.py:603: in plot_drop_log
    percent = _drop_log_stats(drop_log, ignore)
mne\epochs.py:1923: in _drop_log_stats
    raise ValueError('drop_log must be a tuple of tuples')
E   ValueError: drop_log must be a tuple of tuples
```

larsoner:
Fixed by fc85ca7, appears to be coming back green

mne/epochs.py
```diff
@@ -355,7 +355,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         Iterable of indices of selected epochs. If ``None``, will be
         automatically generated, corresponding to all non-zero events.
     drop_log : list | None
-        List of lists of strings indicating which epochs have been marked to be
+        List of tuple of strings indicating which epochs have been marked to be
```
drammock, line 358: Is there a reason that `BaseEpochs` here wants a *list* of tuples, but the `Epochs` constructor wants a *tuple* of tuples?  Since users are discouraged from using `BaseEpochs` for anything but type checking, it's probably not likely to cause problems... just curious if there's a rationale.

larsoner: Just missed updating this docstring

PR #7965, opened 2020-07-06
MRG, BUG: Allow ref mags to have a comp grade

larsoner:
Closes #6488 

@Yaocoo @Nichalas can you try it? @Nichalas at least with your test data I can now read it.

PR #7972, opened 2020-07-07
MRG: Add SNIRF reader

rob-luke:
#### Reference issue
#7057 

Adds ability to read NIRS data in the SNIRF format. 

#### Additional information
This is becoming the default file format and will become the BIDS approved format.

### DIFF
--- a/mne/io/snirf/_snirf.py
+++ b/mne/io/snirf/_snirf.py
@@ -1,0 +1,201 @@
+# Authors: Robert Luke <mail@robertluke.net>
+#
+# License: BSD (3-clause)
+
+import re as re
+import numpy as np
+
+from ..base import BaseRaw
+from ..meas_info import create_info
+from ...annotations import Annotations
+from ...utils import logger, verbose, fill_doc, warn, requires_h5py
+
+
+@requires_h5py
+@fill_doc
+def read_raw_snirf(fname, preload=False, verbose=None):
+    """Reader for a SNIRF fNIRS recording.
+
+    .. note:: This reader supports the .snirf file type only,
+              not the .jnirs version.
+              Currently only support for continuous wave SNIRF data is
+              provided.
+
+    Parameters
+    ----------
+    fname : str
+        Path to the SNIRF data file.
+    %(preload)s
+    %(verbose)s
+
+    Returns
+    -------
+    raw : instance of RawSNIRF
+        A Raw object containing SNIRF data.
+
+    See Also
+    --------
+    mne.io.Raw : Documentation of attribute and methods.
+    """
+    return RawSNIRF(fname, preload, verbose)
+
+
+def _open(fname):
+    return open(fname, 'r', encoding='latin-1')
+
+
+@requires_h5py
+@fill_doc
+class RawSNIRF(BaseRaw):
+    """Raw object from a SNIRF fNIRS file.
+
+    Parameters
+    ----------
+    fname : str
+        Path to the SNIRF data file.
+    %(preload)s
+    %(verbose)s
+
+    See Also
+    --------
+    mne.io.Raw : Documentation of attribute and methods.
+    """
+
+    @verbose
+    def __init__(self, fname, preload=False, verbose=None):
+        from ...externals.pymatreader.utils import _import_h5py
+        h5py = _import_h5py()
+        logger.info('Loading %s' % fname)
+
+        with h5py.File(fname, 'r') as dat:
+
+            if 'data2' in dat['nirs'].keys():
+                print("File contains multiple recordings. "
+                      "This is not supported.")
+
+            data = np.array(dat.get('/nirs/data1/dataTimeSeries')).T
+            last_samps = data.shape[1] - 1
+
+            samplingrate_raw = np.array(dat.get('nirs/data1/time'))
+            sampling_rate = 0
+            if samplingrate_raw.shape == (2, 1):
+                # specified as onset/samplerate
+                warn("Onset sample rate SNIRF not yet supported")
+            else:
+                # specified as time points
+                fs_diff = np.around(np.diff(samplingrate_raw), decimals=4)
+                if len(np.unique(fs_diff)) == 1:
+                    # Uniformly sampled data
+                    sampling_rate = 1. / np.unique(fs_diff)
+                else:
+                    # print(np.unique(fs_diff))
+                    warn("Non uniform sampled data not supported")
+            if sampling_rate == 0:
+                warn("Unable to extract sample rate")
+
+            sources = np.array(dat.get('nirs/probe/sourceLabels'))
+            detectors = np.array(dat.get('nirs/probe/detectorLabels'))
+            sources = [s.decode('UTF-8') for s in sources]
+            detectors = [d.decode('UTF-8') for d in detectors]
+
+            # Extract source and detector locations
+            detPos3D = np.array(dat.get('nirs/probe/detectorPos3D'))
+            srcPos3D = np.array(dat.get('nirs/probe/sourcePos3D'))
+
+            assert len(sources) == srcPos3D.shape[0]
+            assert len(detectors) == detPos3D.shape[0]
+
+            if detPos3D.shape[1] == 2:
+                detPos3D = np.hstack((detPos3D, np.zeros((len(detectors), 1))))
+                srcPos3D = np.hstack((srcPos3D, np.zeros((len(sources), 1))))
+
+            # Extract wavelengths
+            fnirs_wavelengths = np.array(dat.get('nirs/probe/wavelengths'))
+            fnirs_wavelengths = [int(w) for w in fnirs_wavelengths]
+
+            # Extract channels
+            def atoi(text):
+                return int(text) if text.isdigit() else text
+
+            def natural_keys(text):
+                return [atoi(c) for c in re.split(r'(\d+)', text)]
+
+            channels = np.array([name for name in dat['nirs']['data1'].keys()])
+            channels_idx = np.array(['measurementList' in n for n in channels])
+            channels = channels[channels_idx]
+            channels = sorted(channels, key=natural_keys)
+
+            chnames = []
+            for chan in channels:
+                src_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/sourceIndex'))[0])
+                det_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/detectorIndex'))[0])
+                wve_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/wavelengthIndex'))[0])
+                ch_name = sources[src_idx - 1] + '_' +\
+                    detectors[det_idx - 1] + ' ' +\
+                    str(fnirs_wavelengths[wve_idx - 1])
+                chnames.append(ch_name)
+
+            # Create mne structure
+            info = create_info(chnames,
+                               sampling_rate,
+                               ch_types='fnirs_cw_amplitude')
+
+            subject_info = {}
+            names = np.array(dat.get('nirs/metaDataTags/SubjectID'))
+            subject_info['first_name'] = names[0].decode('UTF-8')
+            info.update(subject_info=subject_info)
+
+            LengthUnit = np.array(dat.get('/nirs/metaDataTags/LengthUnit'))
+            LengthUnit = LengthUnit[0].decode('UTF-8')
+            scal = 1
+            if "cm" in LengthUnit:
+                scal = 100
+            elif "mm" in LengthUnit:
+                scal = 1000
+
+            for idx, chan in enumerate(channels):
+                src_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/sourceIndex'))[0])
+                det_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/detectorIndex'))[0])
+                wve_idx = int(np.array(dat.get('nirs/data1/' +
+                                               chan + '/wavelengthIndex'))[0])
+                info['chs'][idx]['loc'][0:3] = srcPos3D[src_idx - 1, :] / scal
+                info['chs'][idx]['loc'][3:6] = srcPos3D[src_idx - 1, :] / scal
+                info['chs'][idx]['loc'][6:9] = detPos3D[det_idx - 1, :] / scal
+                info['chs'][idx]['loc'][9] = fnirs_wavelengths[wve_idx - 1]
+
+            super(RawSNIRF, self).__init__(info, preload, filenames=[fname],
+                                           last_samps=[last_samps],
+                                           verbose=verbose)
+
+            # Extract annotations
+            annot = Annotations([], [], [])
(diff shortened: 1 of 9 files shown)


rob-luke:
This work is on hold until I better understand the expected behaviour of SNIRF files as discussed here https://github.com/fNIRS/snirf_homer3/issues/5

rob-luke:
Test data is at https://github.com/mne-tools/mne-testing-data/pull/68

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,193 @@
+# Authors: Robert Luke <mail@robertluke.net>
+#
+# License: BSD (3-clause)
+
+import re as re
+import numpy as np
+import h5py
```
rob-luke, line 7: @larsoner is it ok to use h5py rather than `from ...externals.pymatreader import read_mat`?

If I use `read_mat` then I get the following error which appears to be caused as the upstream software does not save a mat file type version.

```python

/Users/rluke/Library/Python/3.7/lib/python/site-packages/scipy/io/matlab/mio.py:72: in mat_reader_factory
    mjv, mnv = get_matfile_version(byte_stream)
...


    def get_matfile_version(fileobj):
        """
        Return major, minor tuple depending on apparent mat file type
    
        Where:
    ...


        if maj_val in (1, 2):
            return ret
>       raise ValueError('Unknown mat file type, version %s, %s' % ret)
E       ValueError: Unknown mat file type, version 0, 0

```

larsoner: It's worth opening a pymatreader issue first. Maybe for example you can override/specify the version manually. It'll probably be easier than using h5py manually.

Also I would check to see how MATLAB handles this. pymatreader should probably do what MATLAB does, if reasonable.

I'd also bother the vendor to fix their files. It's probably 1 line of code for them to change...

But yes you can use h5py directly if you need to.

rob-luke: > I'd also bother the vendor to fix their files. It's probably 1 line of code for them to change...

Yep tried already. They aren't going to change it.

> But yes you can use h5py directly if you need to.

Ok I will try your suggestions above, modify pymatreader and see what Matlab does. But keep this as a final option.

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,200 @@
...
+        for idx in range(int(num_chans / 2)):
+            chans.append(idx)
+            chans.append(idx + int(num_chans / 2))
+        self.pick(picks=chans)
+
+    def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
+        """Read a segment of data from a file."""
+        from ...externals.pymatreader.utils import _import_h5py
+        h5py = _import_h5py()
+        dat = h5py.File(self._filenames[0], 'r')
```
larsoner, line 195: To avoid unclosed resources any `h5py.File` should be used as:
```
with h5py.File(...) as dat:
    ...
```

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,200 @@
...
+        dat = np.array(dat.get('/nirs/data1/dataTimeSeries')).T
+
+        data[:] = dat[idx, start:stop]
```
larsoner, line 198: This probably works but will be inefficient (loads entire data array). IIRC h5py supports slicing their own objects, so maybe some variant of this would work:
```suggestion
        data[:] = np.asarray(dat.get('/nirs/data1/dataTimeSeries')[start:stop, idx]).T
```

larsoner: (If this does not because of the fancy indexing of `idx`, you probably at least can slice into the time interval with `data[:] = np.asarray(blah[start:stop])).T[idx])

rob-luke: As you predicted fancy indexing was an issue, so Ive used your second solution.

mne/io/snirf/_snirf.py
rob-luke: Thanks

rob-luke:
Ping @larsoner @agramfort 

SNIRF is a very large standard that include many features that we will not support in MNE (at least for now) including time domain NIRS, a variety of data types we won't support(diffuse correlation spectroscopy), multiple simultaneous measurements on different people (hyperscanning), non uniform sampling, etc.

I have included a few checks for data types we don't support. But this is causing a decrease in coverage as I don't have test data for all these different types (and there is no way I can obtain most of these exceptions). Do you prefer a) I don't test all these obscure features and silently fail if a user passes in a gated time domain fluorescence measurement, or b) put these tests in and have a decrease in test coverage as I cant hit them?

Here is one example of a check that I am referring to 

https://github.com/mne-tools/mne-python/pull/7972/files#diff-6ddea29a0563d9d09a63ec6af1be0bf8R92

larsoner:
> I have included a few checks for data types we don't support. But this is causing a decrease in coverage as I don't have test data for all these different types (and there is no way I can obtain most of these exceptions). Do you prefer a) I don't test all these obscure features and silently fail if a user passes in a gated time domain fluorescence measurement, or b) put these tests in and have a decrease in test coverage as I cant hit them?

One way to hit these lines with `monkeypatch`. If you want to try what I'm thinking of, here is an example where I hack an MNE function to give the output that we would get if a file were exhibiting the problematic behavior:

https://github.com/mne-tools/mne-python/pull/7965/files#diff-89350124c3b525af7b8aabdf157ae7ecR396-R411

It's less robust than actually having the problematic files because you have to make sure that what you mock actually properly captures the problematic behavior, but if you can accurately capture the behavior it's nice. But we also often just live with these lines not being covered, so it might not be worth the effort. Depends on how intrinsically motivated you are to work on this sort of problem...

Feel free to give it a shot if you want to play with `pytest`, otherwise I can try a little bit tomorrow (going to bed here no) and push a commit if successful.

rob-luke:
Thanks for the info.

> Depends on how intrinsically motivated you are to work on this sort of problem...

In this situation I am not very motivated. Lots of the checks will be for situations I never expect we will see in practice. Take for example the non uniform sampling, no commercial devices do this, I doubt even research devices do.

I think we should focus on making sure we support the same features in snirf as are available with the other readers in MNE. So currently we should support continuous wave data with uniform sample rates. Then as MNE is expanded to other nirs types we add this to our snirf support (eg once we have a frequency domain file type merged we can add snirf support for this).

This way we can provide support for what I imagine is the most common workflow. Measure data and save in native format -> read data with MNE -> save in bids format as snirf -> read in snirf bids data for processing.

So how about I put in a few checks, but I don't go overboard and we allow the coverage to slightly decrease?

larsoner:
Sounds good

rob-luke:
@larsoner @agramfort could you please review this file reader. I have been using it for a week with no issues, so I think its behaving correctly.

The SNIRF file format is quite extensive, so as NIRS support expands in MNE this reader will need to be revisited (e.g. kylemath is adding frequency domain nirs support). Next I will write a SNIRF writer function.

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+import re as re
+import numpy as np
+
+from ..base import BaseRaw
+from ..meas_info import create_info
+from ...annotations import Annotations
+from ...utils import logger, verbose, fill_doc, warn, requires_h5py
+
+
+@requires_h5py
```
larsoner, line 14: This is really a testing decorator not meant to be used by user-facing functions. Usually we do this:
```
if not check_version('h5py'):
    raise ImportError('The h5py package is required to read raw SNIRF data')
```
If you want, you could add a helper to `mne/utils/check.py` to make this process more general, like:
```
def _require_version(lib, what, version='0.0'):
    if not check_version(lib, version):
        extra = f' (version >= {version})' if version != '0.0' else ''
        raise ImportError(f'The {lib} package{extra} is required to {what}')
```
WDYT?

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+@fill_doc
+def read_raw_snirf(fname, preload=False, verbose=None):
+    """Reader for a SNIRF fNIRS recording.
```
larsoner, line 17: Maybe for now:
```suggestion
    """Reader for a continuous wave SNIRF data.
```

Then the `.. note::` below can live in a standalone `Notes` section?

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+
+    See Also
+    --------
+    mne.io.Raw : Documentation of attribute and methods.
+    """
+
+    @verbose
+    def __init__(self, fname, preload=False, verbose=None):
+        from ...externals.pymatreader.utils import _import_h5py
+        h5py = _import_h5py()
```
larsoner, line 67: Ahh, if you're already using this (which probably does what I propose adding internally to MNE above) then you shouldn't have `requires_h5py` above

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+        logger.info('Loading %s' % fname)
+
+        with h5py.File(fname, 'r') as dat:
+
+            if 'data2' in dat['nirs'].keys():
```
larsoner, line 72:
```suggestion
            if 'data2' in dat['nirs']:
```

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                warn("File contains multiple recordings. "
+                     "MNE does not support this feature.")
+
+            if np.array(dat.get('nirs/data1/measurementList1/dataType')) != 1:
+                warn("File does not contain continuous wave data. "
```
larsoner, line 77: This should probably be an error

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                     "MNE only supports reading "
+                     "continuous wave amplitude SNIRF files.")
+
+            data = np.array(dat.get('/nirs/data1/dataTimeSeries')).T
```
larsoner, line 81: This will read the entire thing, right (bad memory use and efficiency)? Can you just get the information like:
```
last_samps = dat.get('/nirs/data1/dataTimeSeries').shape[0]
```
Assuming the `.get` does not actually pull all the data until you `np.array` cast it. If it does, I have to imagine there is some way to avoid this in h5py and just get the element shape.

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                info['chs'][idx]['loc'][6:9] = detPos3D[det_idx - 1, :] / scal
+                info['chs'][idx]['loc'][9] = fnirs_wavelengths[wve_idx - 1]
+
+            super(RawSNIRF, self).__init__(info, preload, filenames=[fname],
+                                           last_samps=[last_samps],
+                                           verbose=verbose)
+
+            # Extract annotations
+            annot = Annotations([], [], [])
+            for key in dat['nirs'].keys():
```
larsoner, line 178:
```suggestion
            for key in dat['nirs']:
```

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                if 'stim' in key:
+                    data = np.array(dat.get('/nirs/' + key + '/data'))
+                    if not data.shape == ():
```
larsoner, line 181:
```suggestion
                    if data.shape != ():
```
Though `shape == ()` can occur for single values:
```
>>> np.float64(2).shape
()
```
So maybe what you really want here is `if data.size > 0`?

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                        annot.append(data[:, 0], 1.0, key[4:])
+            self.set_annotations(annot)
+
+            # Reorder channels to match expected ordering in MNE
+            num_chans = len(self.ch_names)
+            chans = []
+            for idx in range(int(num_chans / 2)):
```
larsoner, line 188:
```suggestion
            for idx in range(num_chans // 2):
```
direct int arithmetic preferable to float cast + int cast where possible

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+                chans.append(idx)
+                chans.append(idx + int(num_chans / 2))
```
larsoner, line 190:
```suggestion
                chans.append(idx + num_chans // 2)
```

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+            self.pick(picks=chans)
+
+    def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
+        """Read a segment of data from a file."""
+        from ...externals.pymatreader.utils import _import_h5py
+        h5py = _import_h5py()
```
larsoner, line 196: Don't need to do this here since you've already checked in the `__init__`
```suggestion
        import h5py
```

mne/io/snirf/_snirf.py
```diff
@@ -0,0 +1,202 @@
...
+
+        with h5py.File(self._filenames[0], 'r') as dat:
+            data[:] = np.asarray(
```
larsoner, line 199: Do you even need the `np.asarray` here? Or will the assignment itself will cast for you?

mne/io/snirf/_snirf.py
rob-luke: Sounds good. Ive added this as you suggest. I have not gone through the code base to edit other functions. But will do that once this is merged as a seperate PR.

mne/io/snirf/_snirf.py
rob-luke: Ok I think I have handled this as you suggest now. But its probably worth double checking I interpreted your suggestions correctly.

mne/io/snirf/_snirf.py
rob-luke: I figure I can do a search for all uses of `check_version` and see which match this pattern. Any other suggestions?

larsoner: That's all I would do. Don't need to catch all of them, we can always migrate as we're working on bits of code anyway.

rob-luke:
Thanks for the review @larsoner. As always I have learnt a few things from your comments. Could you please check I have included your suggestions accurately. The Mac CI failed due to 50 min limit, I assume this is unrelated to these code changes.

Next I will code up a SNIRF writer (then we should have close to full support for NIRS BIDS). Where should this writer be placed? I could write a `write_raw_snirf` and put it in `mne/io/snirf/_snirf.py`, or should I extend `save` in `mne/io/base.py` to add an `fname=*.snirf` option?

larsoner:
> Next I will code up a SNIRF writer (then we should have close to full support for NIRS BIDS). Where should this writer be placed? I could write a write_raw_snirf and put it in mne/io/snirf/_snirf.py, or should I extend save in mne/io/base.py to add an fname=*.snirf option?

Let's open a separate issue about this as it will require some discussion

larsoner:
https://github.com/mne-tools/mne-python/issues/8060

larsoner:
Thanks @rob-luke !

PR #7974, opened 2020-07-07
MRG, ENH: BrainVision impedance parsing

sappelhoff:
takes over #7771 
closes #7771 

notes for posterity
- I looked at this vhdr file to adjust some of our own test files: https://gin.g-node.org/JiriVanek/DataConversionToNIX/src/master/BrainVision%20format/RAW/P3Numbers_20150618_f_10_001.vhdr


mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,98 @@ def _check_bv_annot(descriptions):
...
+    Parameters
+    ----------
+    settings : list
+        The header settings lines.
+    recording_date : datetime.datetime | None
+        The date of the recording as extracted from the VMRK file.
+
+    Returns
+    -------
+    dict : A dictionary of all the channels and their impedances.
```
agramfort, line 865:
```
impedances : dict
    A dictionary of all the channels and their impedances.
```

mne/io/brainvision/brainvision.py
```diff
@@ -843,3 +848,98 @@ def _check_bv_annot(descriptions):
...
+    """Parse the selected electrode impedance ranges from the header.
+
+    Parameters
+    ----------
+    settings : list
+        the header settings lines
+
+    Returns
+    -------
+    dict : A dictionary of impedance ranges for each type of electrode
```
agramfort, line 923: idem

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -650,4 +651,33 @@ def test_event_id_stability_when_save_and_fif_reload(tmpdir):
...
+        'imp': expected_imps[i],
+        'imp_unit': expected_imp_unit,
+        'imp_meas_time': expected_imp_meas_time,
+        'imp_lower_bound': expected_imp_lower_bound,
+        'imp_upper_bound': expected_imp_upper_bound[i],
+        'imp_range_unit': expected_imp_unit,
+    } for i, elec in enumerate(expected_electrodes)}
+
+    raw = read_raw_brainvision(vhdr_path, eog=eog)
+    assert_array_equal(expected_impedances, raw.impedances)
```
agramfort, line 680: you are using array equal on something that is not an array. Maybe mne.utils.object_diff ?

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -650,4 +651,57 @@ def test_event_id_stability_when_save_and_fif_reload(tmpdir):
...
+    expected_impedances = {elec: {
+        'imp': expected_imps[i],
+        'imp_unit': expected_imp_unit,
+        'imp_meas_time': expected_imp_meas_time,
+    } for i, elec in enumerate(expected_electrodes)}
+
+    with pytest.warns(RuntimeWarning, match='different .*pass filters'):
+        raw = read_raw_brainvision(vhdr_mixed_lowpass_path,
+                                   eog=['HEOG', 'VEOG'], misc=['ECG'])
+    object_diff(expected_impedances, raw.impedances)
```
larsoner, line 704:
```suggestion
    assert object_diff(expected_impedances, raw.impedances) == ''
```

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -650,4 +651,57 @@ def test_event_id_stability_when_save_and_fif_reload(tmpdir):
...
+        'imp': expected_imps[i],
+        'imp_unit': expected_imp_unit,
+        'imp_meas_time': expected_imp_meas_time,
+        'imp_lower_bound': expected_imp_lower_bound,
+        'imp_upper_bound': expected_imp_upper_bound[i],
+        'imp_range_unit': expected_imp_unit,
+    } for i, elec in enumerate(expected_electrodes)}
+
+    raw = read_raw_brainvision(vhdr_path, eog=eog)
+    object_diff(expected_impedances, raw.impedances)
```
sappelhoff, line 680:
```suggestion
    assert object_diff(expected_impedances, raw.impedances) == ''
```

sappelhoff:
The tests are failing because `object_diff` is comparing `np.nan` as False:

```
In [4]: np.nan == np.nan                                                        
Out[4]: False
```

so actually, everything is passing. How should I fix this?

sappelhoff:
these points go back to the discussion we had some weeks ago.

This is what @larsoner said about this in his [comment](https://github.com/mne-tools/mne-python/pull/7771#issuecomment-628737877)

> BrainVision raw files will be the only files to have this attribute?

I think this is okay. It will also be lost on FIF round-trip but that's also okay. If there are other readers that support impedances we can think about a more global way of storing them or something. As long as we pick a sensible attribute here we can always extend it to other readers later via a mixin or something.

---

If `raw.impedances` becomes a publicly documented attribute of Raw, then we also HAVE to implement proper saving of this attribute, right?

So the options I see:

1. make it an undocumented attribute of BrainVision files only, until we get more file format readers to parse impedances
1. make it a fully documented attribute of Raw, with possible FIF roundtrip

With option 2., I am a bit uncomfortable, because I don't know how many file formats even support saving the impedance values in their file formats?

I have only ever seen this for BrainVision files. Do you know of more @hoechenberger @cbrnr ?

agramfort:
does eeglab set files have a documented way of storing this info?

>

sappelhoff:
> does eeglab set files have a documented way of storing this info?

I did a quick google search, and not  alot shows up. But I am no expert on `.set`

EDF files also do NOT seem to have impedances: https://www.edfplus.info/specs/edfplus.html

hoechenberger:
> I have only ever seen this for BrainVision files. Do you know of more @hoechenberger @cbrnr ?

I've only worked with BrainVision and BioSemi data, and I only remember seeing this for BV!

cbrnr:
I haven't seen that either. If you want to make it a documented attribute, then the default could be `None` (meaning for all formats except BrainVision). Even BrainVision files normally don't have this kind of information, so it should really be an optional piece of information.

hoechenberger:
> Even BrainVision files normally don't have this kind of information, so it should really be an optional piece of information.

Oh really? Interesting, and good to know! The ones produced in our lab always contained this info. Do you know under which circumstances this info will / will not be produced?

cbrnr:
Maybe I'm wrong, but this should only work for amps that measure impedances (i.e. active electrodes) - so you won't find this info for passive amps. Also, none of our example files in the testing data repo has impedance info.

hoechenberger:
> Maybe I'm wrong, but this should only work for amps that measure impedances (i.e. active electrodes)

Ah that could be right, I've only used active electrodes in the past years!

sappelhoff:
whether or not impedances are saved with the BrainVision files really depends on a lot of things.

- the amplifier/electrodes being used
- the recording software and settings being used (BrainVision Recorder works, but also not in all settings/impedance-checks)
- whether or not you have a USB connection to the "actiCap control box"(interface between amps and electrodes) vs only the fibre optics cables

For me *active* vs *passive* was not a deciding factor --> I *can* record impedances from both activa AND passive electrodes (see e.g., the HEOG+ and HEOG- impedances I added in this PR --> that's real data from passive electrodes)

---

anyhow, it seems like reading impedance values from EEG file formats may prove to be difficult.

| possible | not possible |
| ---| ---|
| BrainVision (sometimes) | EDF |
| | BDF|
| | .set |

What about FIF?

larsoner:
Not that I know of

sappelhoff:
@agramfort in https://github.com/mne-tools/mne-python/pull/7974/commits/916f9e9779ff1b29f3c9c0ecba083f9418d1a056 the issue with the broken link from the mailing list should be fixed.

larsoner:
Needs merge/rebase now (sorry)

mne/io/brainvision/brainvision.py
```diff
@@ -51,6 +51,11 @@ class RawBrainVision(BaseRaw):
     %(preload)s
     %(verbose)s
 
+    Attributes
+    ----------
+    impedances : dict
+        A dictionary of all electrodes and their impedances.
```
larsoner, line 57: FYI we don't document the individual classes so this will only show up if you look at the doc of this class `__init__` in a terminal or IDE (which is probably slim to never). But I guess we can keep it

mne/utils/numerics.py
```diff
@@ -759,6 +759,12 @@ def object_diff(a, b, pre=''):
             for ii, (xx1, xx2) in enumerate(zip(a, b)):
                 out += object_diff(xx1, xx2, pre + '[%s]' % ii)
     elif isinstance(a, (str, int, float, bytes, np.generic)):
+        # catch non-array np.nan floats
+        if isinstance(a, float):
+            if np.isnan(a):
+                if not _array_equal_nan(a, b):
+                    out += pre + ' value mismatch (%s, %s)\n' % (a, b)
+                return out
```
larsoner, line 767: I don't like this short-circuiting logic -- now we have to know that there is another `return` path. Instead how about:
```suggestion
    elif isinstance(a, float):
        if not _array_equal_nan(a, b):
            out += pre + ' value mismatch (%s, %s)\n' % (a, b)
    elif isinstance(a, (str, int, bytes, np.generic)):
```
Basically if it's `float` we directly triage to a function that properly checks for equality. If it's any of the other types we can use the builtin `==`.

sappelhoff: yes, that looks a lot better, thanks!

larsoner:
Thanks @sappelhoff

sappelhoff:
seems like the GDF format supports impedance values: https://arxiv.org/pdf/cs/0608052v6,

but I have never used that format and have not seen it "in the wild" too much.

PR #7971, opened 2020-07-07
MRG, DOC: update install instructions

drammock:
closes #7945

PR #7979, opened 2020-07-08
backport new environment file and install instructions

drammock:
@larsoner not sure if I did the cherry-pick correctly; the diff ended up including some changes that weren't from #7971 ... but I *think* that's actually what we want in this case?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -118,6 +118,7 @@ nitpick_ignore = [
     ("py:class", "None.  Update D from dict/iterable E and F."),
     ("py:class", "an object providing a view on D's values"),
     ("py:class", "a shallow copy of D"),
+    ("py:class", "(k, v), remove and return some (key, value) pair as a"),
 ]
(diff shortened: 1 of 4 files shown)


larsoner:
Technically we don't want to bump Travis on `maint/0.20` to 3.6 because 0.20 still supports 3.5. But it probably does not matter much since we're releasing 0.21 in about two months, so if it's enough to make the CIs happy then I guess it's fine.

drammock:
hmm, circle failures are all related to _dict stuff.  I guess I need to cherry-pick the ignore-nitpicky lines in `doc/conf.py`?

larsoner:
Yep, that should do it

drammock:
@larsoner Circle is happy, and I've reverted all the changes except for the changes to the docs, and the `nitpicky` ignores in `conf.py`.  Note that I had to add a *new* nitpick ignore to make Circle happy.  Is this OK to merge then, to get the stable docs in sync with the current state of `environment.yml`?

larsoner:
Thanks @drammock

PR #7983, opened 2020-07-09
Add low-pass filter to find_bad_channels_maxwell()

hoechenberger:

The current approach to using `find_bad_channels_maxwell()` requires users to manually low-pass filter their data before passing it to the function. This is kind of cumbersome.

This PR, therefore, adds a `h_freq` kwarg that sets the cutoff frequency of an automatically-applied low-pass filter. Passing `h_freq=None` disables filtering. If the input data has already been low-pass filtered, a helpful warning is emitted.

We already have a similar API in `preprocessing.compute_proj_ecg()`, so this change makes things more consistent within the `preprocessing` module.

### DIFF
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1978,16 +1978,16 @@ def find_bad_channels_maxwell(
     .. versionadded:: 0.20
     """
     if h_freq is not None:
-        if 'lowpass' in raw.info:
+        if raw.info.get('lowpass') and raw.info['lowpass'] < h_freq:
             msg = (f'The input data has already been low-pass filtered with a '
-                   f'{raw.info["lowpass"]} Hz cutoff frequency. If you wish '
-                   f'to avoid filtering again in find_bad_channels_maxwell(), '
-                   f'please pass `h_freq=None`.')
-            logger.warning(msg)
-
-        logger.info(f'Applying low-pass filter with {h_freq} Hz cutoff '
-                    f'frequency ...')
-        raw = raw.copy().filter(l_freq=None, h_freq=h_freq)
+                   f'{raw.info["lowpass"]} Hz cutoff frequency, which is '
+                   f'below the requested cutoff of {h_freq} Hz. Not applying '
+                   f'low-pass filter.')
+            logger.info(msg)
+        else:
+            logger.info(f'Applying low-pass filter with {h_freq} Hz cutoff '
+                        f'frequency ...')
+            raw = raw.copy().filter(l_freq=None, h_freq=h_freq)
 
     limit = float(limit)
     onsets, ends = _annotations_starts_stops(
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -1897,7 +1897,7 @@ def find_bad_channels_maxwell(
     %(maxwell_skip)s
     h_freq : float | None
         The cutoff frequency (in Hz) of the low-pass filter that will be
-        applied before processing the data. This defaults to ``40.0``, which
+        applied before processing the data. This defaults to ``40.``, which
         should provide similar results to MaxFilter. If you do not wish to
         apply a filter, set this to ``None``.
     %(verbose)s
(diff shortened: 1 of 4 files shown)


mne/preprocessing/maxwell.py
```diff
@@ -1975,6 +1977,18 @@ def find_bad_channels_maxwell(
 
     .. versionadded:: 0.20
     """
+    if h_freq is not None:
+        if 'lowpass' in raw.info:
```
agramfort, line 1981: lowpass is always present but it can be None. I think you will warn all the time here.

What I would do is something like this:

`if raw.info.get('lowpass') and raw.info['lowpass'] < 'lowpass' `

hoechenberger: I've changed things in 94e5807, can you have a look?

larsoner:
-1 on this. There are lots of options for filtering. I don't see why we want to add this here

larsoner:
(This was discussed in the original PR quite a bit, too)

hoechenberger:
> -1 on this. There are lots of options for filtering. I don't see why we want to add this here

I do see your point, but I was feeling that we should provide a sane default here. Less typing and therefore fewer bugs in user code :) Similarly, [`compute_proj_ecg()`](https://mne.tools/stable/generated/mne.preprocessing.compute_proj_ecg.html) offers "built-in" filtering, so I thought it would make sense to add this here too, for the sake of consistency. For advanced users (and use cases :)) where one needs more sophisticated filters, one can still manually filter and set `h_freq=None` to work with the preconditioned data.

> (This was discussed in the original PR quite a bit, too)

I will read up on this, thanks for the pointer.

larsoner:
> Similarly, compute_proj_ecg() already offers "built-in" filtering, so I thought it would make sense to add this here too, for the sake of consistency

I actually consider that a kind of a deficiency in the design of `compute_proj_ecg`. It supports some but not all filtering options and the API is both bloated and incomplete because of this. But also, to at least somewhat justify it, we do filter there by default, I don't think we should for `find_bad_channels_maxwell` (even though we recommend doing it) because it is a bit sensitive to the parameters you choose.

One thing that might be okay is actually to use `**filt_kwargs` here. If it's empty, don't filter. If it's not, pass all of them to `raw = raw.copy().filter(...)`. Then you at least implicitly get all options. But really I'm still -0.5 since all we're really saving is a few characters:
```
raw = raw.copy().filter(2.)
find_bad_channels_maxwell(raw)

# vs
find_bad_channels_maxwell(raw, h_freq=2)
```

agramfort:
what convinces me is that it promotes best practices. If we think that by default users should filter below 40Hz it's a way to help users.

ideally we should filter automatically with the best options to detect artifacts ...

larsoner:
Makes sense. And also I was thinking about high-pass filtering, which can really mess with the performance. The low-pass characteristics I think it's less sensitive to. So I can live with this

agramfort:
would you also suggest to high pass filter? what do you use yourself?

>

larsoner:
No I don't high-pass, I just `filter_chpi`. MaxFilter does some high-passing but it didn't seem to help much in my testing, and was a big source of variability (between MaxFilter and MNE matching). I struggled with trying to make them equivalent for quite some time and couldn't, hence why I was "stuck" on the idea that this was about high-passing...

agramfort:
ok let's move on with just low pass for now

ok for you?

>

mne/preprocessing/maxwell.py
```diff
@@ -1975,6 +1977,18 @@ def find_bad_channels_maxwell(
 
     .. versionadded:: 0.20
     """
+    if h_freq is not None:
+        if raw.info.get('lowpass') and raw.info['lowpass'] < h_freq:
+            msg = (f'The input data has already been low-pass filtered with a '
+                   f'{raw.info["lowpass"]} Hz cutoff frequency, which is '
+                   f'below the requested cutoff of {h_freq} Hz. Not applying '
+                   f'low-pass filter.')
```
larsoner, line 1985: For this to be true you need to then set `h_freq = None`

larsoner: ahh wait nevermind, you don't need to because you actually just filter below!

larsoner:
Thanks @hoechenberger

hoechenberger:
Great, thanks @larsoner

PR #7986, opened 2020-07-09
[MRG] Strip base directory name from Report captions when using parse_folder

hoechenberger:

When using `Report.parse_folder()`, including the "base directory" (i.e., folder name) in all captions's unnecessarily redundant and impairs legibility. This commit changes the behavior of `Report` such that when `parse_folder()` is used, the base directory name will be stripped from the captions.

### `master`
Screenshot 2020-07-09 at 17 43 10

### PR branch
Screenshot 2020-07-09 at 17 42 39

### DIFF
--- a/tutorials/misc/plot_report.py
+++ b/tutorials/misc/plot_report.py
@@ -195,4 +195,3 @@ with mne.open_report('report.h5') as report:
 ###############################################################################
 # With the context manager, the updated report is also automatically saved
 # back to :file:`report.h5` upon leaving the block.
-# This line needs to be removed.
--- a/tutorials/misc/plot_report.py
+++ b/tutorials/misc/plot_report.py
@@ -195,3 +195,4 @@ with mne.open_report('report.h5') as report:
 ###############################################################################
 # With the context manager, the updated report is also automatically saved
 # back to :file:`report.h5` upon leaving the block.
+# This line needs to be removed.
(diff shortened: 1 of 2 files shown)


larsoner:
Can you touch `auto_tutorials/misc/plot_report.html` so that we can see how it looks on CircleCI?

hoechenberger:
> Can you touch `auto_tutorials/misc/plot_report.html` so that we can see how it looks on CircleCI?

Good idea

tutorials/misc/plot_report.py
```diff
@@ -195,3 +195,4 @@
 ###############################################################################
 # With the context manager, the updated report is also automatically saved
 # back to :file:`report.h5` upon leaving the block.
+# This line needs to be removed.
```
larsoner, line 198: FYI instead of doing this, which will require another commit (and another round of CIs), it's better to either improve the wording/tutorial in some way, or (if you can't easily find an improvement) do some harmless (PEP8-compatible) newline removal or insertion.

larsoner: Also in a commit where all you care about is Circle running, it helps to do `[skip travis]` in the commit message since Travis is usually our CI bottleneck (and Azure doesn't have an equivalent)

larsoner:
Indeed it looks better

https://21039-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py

tutorials/misc/plot_report.py
hoechenberger: Thanks @larsoner, I will note this down :)

agramfort:
thx @hoechenberger

PR #7987, opened 2020-07-09
DOC: Update estimates [skip travis]

larsoner:
Just noticed that https://mne.tools/dev/cited.html is a year plus out of date. Updating now, and added another step to the release instructions to update this at release time.


larsoner:
https://21042-1301584-gh.circle-artifacts.com/0/dev/cited.html

PR #7990, opened 2020-07-09
Report parsing

agramfort:
improves parse of folder for report

PR #7991, opened 2020-07-10
MRG: Add SSP projectors to Report

hoechenberger:
This is still WIP and depends on #7990. So will need to be rebased after #7990 has been merged.

### DIFF
--- a/mne/report.py
+++ b/mne/report.py
@@ -39,11 +39,12 @@ from .externals.h5io import read_hdf5, write_hdf5
 
 VALID_EXTENSIONS = ['raw.fif', 'raw.fif.gz', 'sss.fif', 'sss.fif.gz',
                     'eve.fif', 'eve.fif.gz', 'cov.fif', 'cov.fif.gz',
-                    'trans.fif', 'trans.fif.gz', 'fwd.fif', 'fwd.fif.gz',
-                    'epo.fif', 'epo.fif.gz', 'inv.fif', 'inv.fif.gz',
-                    'ave.fif', 'ave.fif.gz', 'T1.mgz', 'meg.fif']
-SECTION_ORDER = ['raw', 'events', 'epochs', 'evoked', 'covariance', 'trans',
-                 'mri', 'forward', 'inverse']
+                    'proj.fif', 'prof.fif.gz', 'trans.fif', 'trans.fif.gz',
+                    'fwd.fif', 'fwd.fif.gz', 'epo.fif', 'epo.fif.gz',
+                    'inv.fif', 'inv.fif.gz', 'ave.fif', 'ave.fif.gz', 'T1.mgz',
+                    'meg.fif']
+SECTION_ORDER = ['raw', 'events', 'epochs', 'ssp', 'evoked', 'covariance',
+                 'trans', 'mri', 'forward', 'inverse']
 
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


agramfort:
can you share a rendered report to see how it looks?

hoechenberger:
> can you share a rendered report to see how it looks?

Yes, once it renders :D

hoechenberger:
Built doc: https://21081-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py

See the 2nd Report in there for an example with SSP projectors.

hoechenberger:
built doc:
https://21097-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py

hoechenberger:
@larsoner @GuillaumeFavelier Any idea why the notebook test is failing? Seems to be unrelated to this PR?

agramfort:
@hoechenberger projs are often stored in separate files like a cov.fif you have proj.fif. Cf sample data and study template that have projs for ecg and eog. Here you plot the projs in the data. Please add the other option with separate files.

hoechenberger:
Doc artifact:
https://21141-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py

hoechenberger:
re Travis error see https://github.com/mne-tools/mne-python/pull/8000#issuecomment-657131358

mne/tests/test_report.py
```diff
@@ -83,7 +86,8 @@ def test_render_report(renderer, tmpdir):
     # Also, make sure crop range is wide enough to avoid rendering bug
     epochs.average().crop(0.1, 0.2).save(evoked_fname)
 
-    report = Report(info_fname=raw_fname_new, subjects_dir=subjects_dir)
+    report = Report(info_fname=raw_fname_new, subjects_dir=subjects_dir,
+                    projs=True)
```
agramfort, line 90: you should check that the proj appears in the rendered html. See how it's done elsewhere. I would just check that a certain string is present.

hoechenberger: Done in d9cc5cd

hoechenberger:
Travis errors seem unrelated. We need to fix this flake8 thing for Python 3.6…

hoechenberger:
As far as I am concerned, this is good to merge.

mne/report.py
```diff
@@ -1615,10 +1647,15 @@ def __exit__(self, type, value, traceback):
...
-            caption = f'{prefix}: {fname} {suffix}'
-        else:
-            caption = f'{prefix}: {fname[len(data_path)+1:]} {suffix}'
+        if data_path is not None:
+            # Ensure we drop leading path separators from the
+            # filename before using it as a caption. We don't know whether
+            # `data_path` has a trailing separator or not; so create a Path
+            # just to be safe (trailing separators will be removed during the
+            # roundtrip).
+            fname = fname[len(str(Path(data_path))) + 1:]
```
larsoner, line 1656: `os.path.relpath` instead?

hoechenberger: No it would have a different behavior than what I'm trying to achieve here… I know it's not pretty, but it does to the job.

larsoner: When would it be different?

hoechenberger: Never mind, `op.relpath` makes total sense here. Thanks!

hoechenberger:
I have no idea why the phsionet dataset test is failing on Travis, all dataset tests pass without any issues locally for me.

larsoner:
Thanks @hoechenberger

PR #7998, opened 2020-07-10
WIP: TFR Doc/test changes

### LINKED ISSUE #7978: DOC: order of operation in TFR functions needs to be clarified
issue came up here: https://gitter.im/mne-tools/mne-python?at=5f05de658342f4627400eee5

the tfr_morlet for time frequency decomposition has a parameter average which says:

>    If True average across Epochs.

The docs need to be updated to make it clear what the order of operations is. First averaging then decomposition? Or the other way around? What are the steps?

bloyl:
Closes #7978

@sappelhoff @larsoner @agramfort 

I have currently commented out the tests for the complex return type as they don't make sense and thus shouldn't pass.
I could remove them and just leave them as smoke tests or if anyone has a suggestion on what to regression test to I could do that. Just let me know what you think is best.

@drammock I'll admit that I don't know how all the doc linking stuff works so please let me know what needs to be changed.

### DIFF
--- a/mne/time_frequency/tests/test_tfr.py
+++ b/mne/time_frequency/tests/test_tfr.py
@@ -104,6 +104,16 @@ def test_time_frequency():
     assert_allclose(power.data, power_picks_avg.data)
     assert_allclose(itc.data, itc_picks.data)
 
+    # test on evoked
+    power_evoked = tfr_morlet(evoked, freqs, n_cycles, use_fft=True,
+                              return_itc=False)
+    # one is squared magnitude of the average (evoked) and
+    # the other is average of the squared magnitudes (epochs PSD)
+    # so values shouldn't match, but shapes should
+    assert_array_equal(power.data.shape, power_evoked.data.shape)
+    pytest.raises(AssertionError, assert_allclose,
+                  power.data, power_evoked.data)
+
     # complex output
     pytest.raises(ValueError, tfr_morlet, epochs, freqs, n_cycles,
                   return_itc=False, average=True, output="complex")
@@ -112,9 +122,13 @@ def test_time_frequency():
     epochs_power_complex = tfr_morlet(epochs, freqs, n_cycles,
                                       output="complex", average=False,
                                       return_itc=False)
-    epochs_power_2 = abs(epochs_power_complex)
-    epochs_power_3 = epochs_power_2.copy()
-    epochs_power_3.data[:] = np.inf  # test that it's actually copied
+    epochs_amplitude_2 = abs(epochs_power_complex)
+    epochs_amplitude_3 = epochs_amplitude_2.copy()
+    epochs_amplitude_3.data[:] = np.inf  # test that it's actually copied
+
+    # test that the power computed via `complex` is equivalent to power
+    # computed within the method.
+    assert_allclose(epochs_amplitude_2.data**2, epochs_power_picks.data)
 
     print(itc)  # test repr
     print(itc.ch_names)  # test property
--- a/mne/time_frequency/tests/test_tfr.py
+++ b/mne/time_frequency/tests/test_tfr.py
@@ -3,8 +3,7 @@ import datetime
 import os.path as op
 
 import numpy as np
-from numpy.testing import (assert_array_almost_equal, assert_array_equal,
-                           assert_equal)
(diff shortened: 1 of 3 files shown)


mne/time_frequency/tfr.py
```diff
@@ -689,7 +689,13 @@ def tfr_morlet(inst, freqs, n_cycles, use_fft=False, return_itc=True, decim=1,
 
         .. versionadded:: 0.13.0
     average : bool, default True
-        If True average across Epochs.
+        If False return an EpochsTFR containing separate tfrs for each epoch.
+        If True return an AverageTFR containing the average of all tfrs across
+            epochs.
```
drammock, line 694:
```suggestion
        If ``False`` return an `EpochsTFR` containing separate TFRs for each
        epoch. If ``True`` return an `AverageTFR` containing the average of all
        TFRs across epochs.
```

mne/time_frequency/tfr.py
```diff
@@ -689,7 +689,13 @@ def tfr_morlet(inst, freqs, n_cycles, use_fft=False, return_itc=True, decim=1,
...
+
+        .. note:: Using `average=True` is functionally equivlent to using
+        `average=False` followed by `EpochsTFR.average()`, but will be more
+        memory efficient.
```
drammock, line 698:
```suggestion
        .. note::
           Using ``average=True`` is functionally equivalent to using
           ``average=False`` followed by ``EpochsTFR.average()``, but is
           more memory efficient.
```

mne/time_frequency/tfr.py
```diff
@@ -832,7 +838,13 @@ def tfr_multitaper(inst, freqs, n_cycles, time_bandwidth=4.0,
...
     %(picks_good_data)s
     average : bool, default True
-        If True average across Epochs.
+        If False return an EpochsTFR containing separate tfrs for each epoch.
+        If True return an AverageTFR containing the average of all tfrs across
+            epochs.
+
+        .. note:: Using `average=True` is functionally equivlent to using
+        `average=False` followed by `EpochsTFR.average()`, but will be more
+        memory efficient.
```
drammock, line 847:
```suggestion
        .. note::
           Using ``average=True`` is functionally equivalent to using
           ``average=False`` followed by ``EpochsTFR.average()``, but is
           more memory efficient.
```

mne/time_frequency/tfr.py
```diff
@@ -832,7 +838,13 @@ def tfr_multitaper(inst, freqs, n_cycles, time_bandwidth=4.0,
     %(n_jobs)s
     %(picks_good_data)s
     average : bool, default True
-        If True average across Epochs.
+        If False return an EpochsTFR containing separate tfrs for each epoch.
+        If True return an AverageTFR containing the average of all tfrs across
+            epochs.
```
drammock, line 843:
```suggestion
        If ``False`` return an `EpochsTFR` containing separate TFRs for each
        epoch. If ``True`` return an `AverageTFR` containing the average of all
        TFRs across epochs.
```

sappelhoff:
Thanks @bloyl the docstring is now much more understandable and clear.

re: test --> I'd either remove it or make it meaningful, but not comment it out and let it sit there.

> @drammock I'll admit that I don't know how all the doc linking stuff works so please let me know what needs to be changed.

that works by using this file: https://github.com/mne-tools/mne-python/blob/master/mne/utils/docs.py

see also the explanation that I received some months ago: https://github.com/mne-tools/mne-python/pull/7474#issuecomment-600783472

mne/time_frequency/tests/test_tfr.py
```diff
@@ -118,9 +115,6 @@ def test_time_frequency():
     epochs_power_2 = abs(epochs_power_complex)
     epochs_power_3 = epochs_power_2.copy()
     epochs_power_3.data[:] = np.inf  # test that it's actually copied
-    assert_array_almost_equal(epochs_power_2.data, epochs_power_picks.data)
```
larsoner, line 121: Instead of killing these, if they should check the opposite, you could do something like:
```
assert not np.allclose(..., atol=1-20)
```
Or whatever makes the point that they actually shouldn't be equal

larsoner:
Thanks @bloyl

PR #8000, opened 2020-07-11
Ignore flake8 E741 (ambiguous variable name)

hoechenberger:
Running flake8 on master gives me:
```
./tutorials/source-modeling/plot_background_freesurfer_mne.py:133:22: E741 ambiguous variable name 'I'
./tutorials/source-modeling/plot_background_freesurfer_mne.py:136:22: E741 ambiguous variable name 'I'
```

https://github.com/mne-tools/mne-python/blob/d839a6ce32da840d7ed4f4cec5340e479f8bde55/tutorials/source-modeling/plot_background_freesurfer_mne.py#L132-L137

I don't think this rule is very useful (as we'll spot badly-named variables during the code review anyway), so this PR disables it.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ doc-files = doc
 
 [flake8]
 exclude = __init__.py,*externals*,constants.py,fixes.py,resources.py
-ignore = W503,W504,I100,I101,I201,N806,E201,E202,E221,E222,E241
+ignore = W503,W504,I100,I101,I201,N806,E201,E202,E221,E222,E241,E741
 # We add A for the array-spacing plugin, and ignore the E ones it covers above
 select = A,E,F,W,C
 


hoechenberger:
@larsoner Travis fails with
```
[gw0] linux -- Python 3.6.7 /home/travis/miniconda/envs/test/bin/python3.6
4043mne/tests/test_docstring_parameters.py:161: in test_docstring_parameters
4044    raise AssertionError(msg)
4045E   AssertionError: 
4046E   mne.source_space.SourceSpaces.clear : SS03 : Summary does not end with a period
4047E   mne.source_space.SourceSpaces.reverse : SS03 : Summary does not end with a period
4048E   2 errors
4049
4050
```
I checked and the docstrings look alright on Py 3.7 and 3.8, so this seems to be a Py 3.6 thing…

larsoner:
I would rather rename the variable and keep the check

larsoner:
Or noqa it if it's a valuable exception

hoechenberger:
@larsoner 
But I think that the naming that causes the messages here makes total sense in the given context (dict key), and we don't want #noqa strings in a tutorial, do we?

That said, maybe there's an easier solution:
instead of creating the dict via
```python
 ori_slice = dict(P='Coronal', A='Coronal', 
                  I='Axial', S='Axial', 
                  L='Sagittal', R='Saggital') 
```
we could use the `{k: v}` syntax, so the names flake8 complains about would become strings. And now thinking about this some more it actually seems like a flake8 bug to me in the first place!!

hoechenberger:
Seeing that the docstyle tests pass for #7991 **after** it has been rebased on `master`, I have to conclude that the flake complaint is probably something that only happens on my system -- at least it doesn't seem to happen on CI (anymore). Therefore, I'm closing this PR.

Still-- what do we do about the Travis poblem?

hoechenberger:
Just for reference, I've updated to:
```
flake8-3.8.3
pyflakes-2.2.0
```
and I don't see this error anymore.

PR #8002, opened 2020-07-12
MRG: When picking, remove inapplicable projectors

hoechenberger:
Pick'ing may entirely eliminate a certain channel type from the resulting object. However, the list of projectors in `info['projs']` will remain unchanged. This can lead to problems in later processing steps, where situations may arise where one is trying to apply a projector for a channel type that is not present in the data anymore.

**MWE:**
```python
import mne

raw.set_eeg_reference(projection=True)
raw.pick_types(meg=False, eeg=True)  # Drop MEG data
raw.plot_projs_topomap()
```

**With `master`:**
A problem arises because although we dropped all MEG data, the MEG projectors are still present.

```python
  File "/private/tmp/mwe.py", line 10, in <module>
    raw.plot_projs_topomap()
  File "/Users/hoechenberger/Development/mne-python/mne/io/proj.py", line 280, in plot_projs_topomap
    fig = plot_projs_topomap(self.info['projs'], self.info, cmap=cmap,
  File "/Users/hoechenberger/Development/mne-python/mne/viz/topomap.py", line 324, in plot_projs_topomap
    use_info = pick_info(info, pick_channels(info_names, ch_names))
  File "<decorator-gen-8>", line 21, in pick_info
  File "/Users/hoechenberger/Development/mne-python/mne/io/pick.py", line 489, in pick_info
    raise ValueError('No channels match the selection.')
ValueError: No channels match the selection.
```

**With this PR:**
The MEG projectors get removed during the pick:
```
Removing projector <Projection | PCA-v1, active : False, n_channels : 102>
Removing projector <Projection | PCA-v2, active : False, n_channels : 102>
Removing projector <Projection | PCA-v3, active : False, n_channels : 102>
```
Screenshot 2020-07-12 at 22 46 10

WIP because
- I'm looking for feedback, obviously
- tests are missing
- not sure if this would demand a deprecation cycle

Thanks!

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -904,7 +904,7 @@ class UpdateChannelsMixin(object):
                 drop_idx.append(idx)
 
         for idx in drop_idx:
-            logger.warning(f"Removing projector {self.info['projs'][idx]}")
+            warn(f"Removing projector {self.info['projs'][idx]}")
         if drop_idx:
             self.del_proj(drop_idx)
 
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -893,7 +893,7 @@ class UpdateChannelsMixin(object):
         return self
 
     def _pick_projs(self):
-        """Only keep projectors whose channels are still in the data."""
+        """Keep only projectors which apply to at least 1 data channel."""
         drop_idx = []
         for idx, proj in enumerate(self.info['projs']):
             if not set(self.info['ch_names']) & set(proj['data']['col_names']):
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -722,7 +722,9 @@ class UpdateChannelsMixin(object):
             ias=ias, syst=syst, seeg=seeg, dipole=dipole, gof=gof, bio=bio,
             ecog=ecog, fnirs=fnirs, include=include, exclude=exclude,
             selection=selection)
-        return self._pick_drop_channels(idx)
+        self._pick_drop_channels(idx)
+        self._pick_projs()
+        return self
 
     def pick_channels(self, ch_names, ordered=False):
         """Pick some channels.
@@ -756,8 +758,10 @@ class UpdateChannelsMixin(object):
 
         .. versionadded:: 0.9.0
         """
-        return self._pick_drop_channels(
-            pick_channels(self.info['ch_names'], ch_names, ordered=ordered))
+        self._pick_drop_channels(pick_channels(self.info['ch_names'], ch_names,
+                                               ordered=ordered))
+        self._pick_projs()
+        return self
 
     @fill_doc
     def pick(self, picks, exclude=()):
@@ -777,7 +781,9 @@ class UpdateChannelsMixin(object):
         """
         picks = _picks_to_idx(self.info, picks, 'all', exclude,
                               allow_empty=False)
-        return self._pick_drop_channels(picks)
+        self._pick_drop_channels(picks)
+        self._pick_projs()
+        return self
 
     def reorder_channels(self, ch_names):
         """Reorder channels.
@@ -888,6 +894,20 @@ class UpdateChannelsMixin(object):
             self._data = self._data.take(idx, axis=axis)
         else:
             assert isinstance(self, BaseRaw) and not self.preload
+        return self
+
+    def _pick_projs(self):
+        """Only keep projectors whose channels are still in the data."""
+        drop_idx = []
+        for idx, proj in enumerate(self.info['projs']):
+            if not set(self.info['ch_names']) & set(proj['data']['col_names']):
+                drop_idx.append(idx)
+
+        for idx in drop_idx:
+            logger.warning(f"Removing projector {self.info['projs'][idx]}")
+        if drop_idx:
+            self.del_proj(drop_idx)
+
         return self
 
     def add_channels(self, add_list, force_update_info=False):
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -901,7 +901,8 @@ class UpdateChannelsMixin(object):
 
         for idx in drop_idx:
             logger.info(f"Removing projector {self.info['projs'][idx]}")
-        if drop_idx:
+
+        if drop_idx and hasattr(self, 'del_proj'):
             self.del_proj(drop_idx)
 
         return self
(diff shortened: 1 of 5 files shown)


larsoner:
> , where situations may arise where one is trying to apply a projector for a channel type that is not present in the data anymore.

This actually shouldn't be a problem, it should just be `np.eye` in that case. If it breaks something it's a bug downstream that we should fix first (regardless of whether or not we do adopt the behavior in this PR)

hoechenberger:
> If it breaks something it's a bug downstream that we should fix first

Ok, since there's a MWE in my first comment, we can use this as a bug report :)

hoechenberger:
… although the problem here is not exactly about _applying_ a projector, but about plotting it!

larsoner:
Actually looking at the code and the problem, this probably is the right way to go. Let's get #8007 in, rebase this, and then see if we can fix any errors that crop op

hoechenberger:
Thanks @larsoner, I will pick up working on this shortly.
(current code also contains a bug, which I realized wile drafting #8003)

hoechenberger:
Will rebase on master after #8005 and #8007 have been merged.

hoechenberger:
Rebased on `master`.

In #8003, we changed the Note in `del_proj()`:

```
        .. note:: The projection vector can only be removed if it is inactive
                  (has not been applied to the data), unless the channels it
                  was applied to no longer exist in the data.
```

Once this PR gets merged, however, the case "the channels [the projector] was applied to no longer exist in the data" shouldn't be possible anymore through use of the public API, right? Because the only way to remove channels is by picking, and this PR will ensure that picking also drops unusable projs. So I'm wondering if I should drop this part from the `del_proj()` docstring as part of this PR?

hoechenberger:
I tried to capture the warnings in a test but it doesn't seem to work -- pytest claims no warning of any kind has been emitted, even though the warnings do appear on stdout!

```python
        with pytest.warns(RuntimeWarning, match='Removing projector.*'):
>           raw.pick_types(meg=False, eeg=True)
E           Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) was emitted. The list of emitted warnings is: [].
```

Any idea what might be the issue?

mne/channels/channels.py
```diff
@@ -890,6 +896,20 @@ def _pick_drop_channels(self, idx):
...
 
+    def _pick_projs(self):
+        """Only keep projectors whose channels are still in the data."""
+        drop_idx = []
+        for idx, proj in enumerate(self.info['projs']):
+            if not set(self.info['ch_names']) & set(proj['data']['col_names']):
+                drop_idx.append(idx)
+
+        for idx in drop_idx:
+            logger.warning(f"Removing projector {self.info['projs'][idx]}")
```
larsoner, line 907: don't use `logger.warning` but rather `warn` from `..utils`

hoechenberger: Weird, ok… why is that?

larsoner: We always use `warn`. It does nice things like sets `stacklevel` automatically, and does or does not emit based on the `logger.level`, etc.

hoechenberger: Thanks, that's good to know!

hoechenberger:
Ok let's see what CI says… I'm worried I will have to handle the new warning in 123456 different tests individually…

hoechenberger:
… or I could just turn it into an `logger.info`. WDYT? Where do we draw the line between info and warning?

larsoner:
This to me should just be `logger.info`. warnings should be if it's likely someone is doing something bad that could break something. Nobody's code should break (in principle at least) because we remove projs that don't apply to the data

mne/channels/channels.py
```diff
@@ -722,7 +722,9 @@ def pick_types(self, meg=None, eeg=False, stim=False, eog=False,
             ias=ias, syst=syst, seeg=seeg, dipole=dipole, gof=gof, bio=bio,
             ecog=ecog, fnirs=fnirs, include=include, exclude=exclude,
             selection=selection)
-        return self._pick_drop_channels(idx)
+        self._pick_drop_channels(idx)
```
larsoner, line 725: More DRY to just call `self._pick_projs()` at the end of `self._pick_drop_channels`, no? The one you don't technically need it for is `reorder_channels`, but that's okay because it will be a no-op there. It would however have avoided the bug that seems to be here, namely, that it seems to be missing from `def drop_channels` (right?)

hoechenberger:
CI is failing because `AverageTFR` doesn't inherit from `ProjMixin` and therefore doesn't have a `drop_proj()` method. What to do about that?

mne/channels/channels.py
hoechenberger: > It would however have avoided the bug that seems to be here, namely, that it seems to be missing from `def drop_channels` (right?)

I had forgotten about `drop_channels`, yes, that's a bug.

Good idea you're proposing here, will do that!

hoechenberger: @larsoner I now included projector handling in `drop_channels` too.

After thinking about your more DRY proposal for a bit, I felt that I would like to keep `_pick_drop_channels` (which acts on channels) and `_pick_projs` (which acts on projectors) as two clearly separated steps for now, meaning I'd prefer keep the current non-DRY approach. It's just my personal feeling, I prefer not to convolute stuff too much. 

If you have strong objections, I can change the according like you suggested; otherwise, I'd like to keep it the way it currently is.

larsoner:
I'm fine with keeping the drop projections as a separate function, but I don't see a good reason for _pick_drop_channels not to call it. I think it's less convoluted actually. pick_drop_channels should do everything required to pick and drop channels from an instance once you know the indices, one of which is (now) to drop irrelevant projectors.

Going in the direction you're advocating, we shouldn't even have a _pick_drop_channels but instead split it into one that picks just from info, one that just drops data if preloaded, etc., and replicate these calls four or five places now and more later if we add another mixin method that needs to _pick_drop...

hoechenberger:
@larsoner Ok, implemented in bed0dcd

hoechenberger:
> CI is failing because `AverageTFR` doesn't inherit from `ProjMixin` and therefore doesn't have a `del_proj()` method. What to do about that?

@larsoner Any thoughts on this one? I've never worked with  `AverageTFR`  so I'm not sure what's the best approach here.

larsoner:
You could check to see if it has that attribute and only call it if it does

hoechenberger:
@larsoner Done

hoechenberger:
@larsoner This is good to go from my end. Let me know if you would like me to rebase.

larsoner:
Thanks @hoechenberger

PR #8003, opened 2020-07-13
MRG: Allow removal of active projectors if channels they applied to have meanwhile been dropped

hoechenberger:
In cases where the channels that an applied projector applies to have been removed from the data (pick after apply_proj() call), it should be safe to allow users to remove the projector. Currently we disallow deletion of any "active" (applied) projectors, so this commit adds more flexibility there.

### DIFF
--- a/mne/io/fiff/tests/test_raw_fiff.py
+++ b/mne/io/fiff/tests/test_raw_fiff.py
@@ -739,6 +739,16 @@ def test_proj(tmpdir):
     raw.apply_proj()
     assert_allclose(raw[:, :][0][:1], raw[0, :][0])
 
+    # Read file again, apply proj, pick all channels one proj did NOT apply to;
+    # then try to delete this proj, which now exclusively refers to channels
+    # which are not present in the data anymore.
+    raw = read_raw_fif(fif_fname).apply_proj()
+    del_proj_idx = 0
+    picks = list(set(raw.ch_names) -
+                 set(raw.info['projs'][del_proj_idx]['data']['col_names']))
+    raw.pick(picks)
+    raw.del_proj(del_proj_idx)
+
 
 @testing.requires_testing_data
 @pytest.mark.parametrize('preload', [False, True, 'memmap.dat'])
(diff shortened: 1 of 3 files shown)


mne/io/proj.py
```diff
@@ -222,7 +222,8 @@ def del_proj(self, idx='all'):
         """Remove SSP projection vector.
 
         .. note:: The projection vector can only be removed if it is inactive
-                  (has not been applied to the data).
+                  (has not been applied to the data), unless the channels it
+                  was applied to no longer exist in the data.
```
agramfort, line 226:
```suggestion
                  was applied to no longer exists in the data.
```

mne/io/fiff/tests/test_raw_fiff.py
```diff
@@ -739,6 +739,16 @@ def test_proj(tmpdir):
...
 
+    # Read file again, apply proj, pick all channels one proj did NOT apply to;
+    # then try to delete this proj, which now exclusively refers to channels
+    # which are not present in the data anymore.
+    raw = read_raw_fif(fif_fname).apply_proj()
+    del_proj_idx = 0
+    picks = list(set(raw.ch_names) -
+                 set(raw.info['projs'][del_proj_idx]['data']['col_names']))
+    raw.pick(picks)
+    raw.del_proj(del_proj_idx)
```
agramfort, line 750: can you test that del_proj_idx is actually no longer there?

mne/io/proj.py
hoechenberger: It has to be "exist" here:
"unless the **channels** it was applied to no longer **exist** in the data."

mne/io/fiff/tests/test_raw_fiff.py
hoechenberger: Is 09c9a9b sufficient?

mne/io/proj.py
larsoner: Agreed what is here is already grammatically correct

larsoner:
Thanks @hoechenberger

PR #8011, opened 2020-07-15
MINOR, DOC: Add See Also

sappelhoff:
The `save` method should have a pointer to the function used to read back the data.

### DIFF
--- a/mne/time_frequency/tfr.py
+++ b/mne/time_frequency/tfr.py
@@ -986,7 +986,6 @@ class _BaseTFR(ContainsMixin, UpdateChannelsMixin, SizeMixin):
         See Also
         --------
         read_tfrs, write_tfrs
-
         """
         write_tfrs(fname, self, overwrite=overwrite)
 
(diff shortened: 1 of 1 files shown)


mne/time_frequency/tfr.py
```diff
@@ -982,6 +982,11 @@ def save(self, fname, overwrite=False):
             The file name, which should end with ``-tfr.h5``.
         overwrite : bool
             If True, overwrite file (if it exists). Defaults to False.
+
+        See Also
+        --------
+        read_tfrs, write_tfrs
+
```
larsoner, line 989:
```
E           mne.time_frequency.tfr.AverageTFR.save : GL02 : Closing quotes should be placed in the line after the last text in the docstring (do not close the quotes in the same line as the text, or leave a blank line between the last text and the quotes)
E           mne.time_frequency.tfr.AverageTFR.save : GL03 : Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
E           mne.time_frequency.tfr.EpochsTFR.save : GL02 : Closing quotes should be placed in the line after the last text in the docstring (do not close the quotes in the same line as the text, or leave a blank line between the last text and the quotes)
E           mne.time_frequency.tfr.EpochsTFR.save : GL03 : Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
E           4 errors
```

sappelhoff: probably just this change, right?
```suggestion
```

larsoner: yep

larsoner:
Thanks @sappelhoff

PR #8012, opened 2020-07-15
MRG, MAINT: Work around matplotlib deprecation

larsoner:
Uses the now-preferred method of setting a window title.


agramfort:
thx @larsoner

PR #8014, opened 2020-07-15
MRG, ENH: Combine channels

### LINKED ISSUE #7984: Feature request: Average over channels
Hello!

(This issue has been raised in the mne_analysis mailing list)

#### Describe the problem
There is currently no way to average signals over channels on a Raw, Epochs or Evoked object easily. One can average signals over channels with the "combine" parameter in some visualisation options but it is not possible to modify a Raw, Epochs or Evoked instance by performing an average over channels.

#### Describe your solution
One solution could be to add a new method to Raw, Epochs and Evoked enabling to combine channels (with mean or median options for example). It could for example take as one of the parameters a dictionary where each ROI (new pseudo-channels) would be associated with a list of channels to perform the combination on. This method could also eventually return the standard deviations in order to compute confidence intervals...

- One alternative is to use for example the [plot_image](https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs.plot_image) method with the "combine" parameter, but it only enables to plot signals and does not return a modified instance of Epochs.
- Another alternative is to extract data with [get_data](https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.get_data), perform the operation with [numpy.mean](https://numpy.org/doc/stable/reference/generated/numpy.mean.html), recreate channel information with pseudo-channels thanks to [create_info](https://mne.tools/stable/generated/mne.create_info.html) and recreate a Raw object from a numpy array with [RawArray](https://mne.tools/stable/generated/mne.io.RawArray.html).

#### Additional context
No additional context.

Please let me know if that makes sense?

Many thanks!
Johann

HanBnrd:
#### Reference issue
Closes #7984

Implementation of a way to combine channels from Raw, Epochs, or Evoked. Channels can be combined with a callable as argument or with "mean", "median", or "std".

#### Additional information
- It currently returns a new instance and does not modify the original one
- It is currently a method in `mne/channels/channels.py`, open to suggestions for a better location
- Was wondering if this sounds fine before implementing tests


HanBnrd:
Hi @drammock,
I implemented this with a single dict as argument, would be happy to have some guidance to be also able to pass multiple key-values pairs. Should I do it with `**kwargs`? Would you know an example of this kind of function/method in MNE?

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+            # Check if channel indices are out of bounds
+            if not all(idx in ch_idx for idx in these_picks):
+                raise ValueError('Some channel indices are out of bounds.')
+            # Check if heterogeneous sensor type combinations
+            this_ch_type = np.array(ch_types)[these_picks]
+            if len(set(this_ch_type)) > 1:
+                types = ', '.join(set(this_ch_type))
+                raise ValueError('Cannot combine sensors of different types; '
+                                 '"{}" contains types {}.'
+                                 ''.format(this_group, types))
```
drammock, line 1058: since our minimum supported Python version recently changed to 3.6, we can use [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) instead of `str.format()`.  Our plan is to use f-strings in new code, and gradually update old code to use f-strings as it gets edited. These lines would become:

```suggestion
                raise ValueError('Cannot combine sensors of different types; '
                                 f'"{this_group}" contains types {types}.')
```

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+            #  Check if combining less than 2 channel
+            if len(these_picks) < 2:
+                raise ValueError('Less than 2 channels in group "{}"; cannot '
+                                 'combine by method "{}".'
+                                 ''.format(this_group, method))
```
drammock, line 1063: this will prevent users from passing single-channel groups, and you haven't (yet) implemented a boolean `drop` parameter either, so currently there is no way to combine certain channels and leave other channels untouched. Ping @larsoner, @agramfort: can you weigh in on whether allowing a mix of combined and uncombined channels is something we should support?

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+
+        Returns
+        -------
+        inst : instance of Raw, Epochs, or Evoked
+            A new instance.
+        """
+        from ..io import BaseRaw, RawArray
+        from ..epochs import BaseEpochs
+        from ..evoked import Evoked
+        from .. import EpochsArray, EvokedArray
```
drammock, line 1026: a quick `git grep` shows me that `_BaseTFR` inherits from `UpdateChannelsMixin`, so this will need to support TFR objects too, I think.

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+
+        try:
+            # Extract stim channels if any
+            inst = self.copy().pick_types(meg=False, stim=True)
+            # Create clean info to avoid inconsistencies
+            inst.info = create_info(sfreq=self.info['sfreq'],
+                                    ch_names=inst.info['ch_names'],
+                                    ch_types=_get_channel_types(inst.info))
+        except ValueError:
+            inst = None
```
drammock, line 1076: can you explain the rationale here?  I can imagine cases where keeping stim channels would be desirable, and other cases where the presence of stim channels would be an unexpected surprise...  seems like something the user should be able to control.

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+        # Convert string values of ``method`` into callables
+        if isinstance(method, str):
+            method_dict = {key: partial(getattr(np, key), axis=ch_axis)
+                           for key in ('mean', 'median', 'std')}
+            try:
+                method = method_dict[method]
+            except KeyError:
+                raise ValueError('"method" must be a callable, or one of '
+                                 '"mean", "median", or "std"; got {}'
+                                 ''.format(method))
```
drammock, line 1045: is there a way to modify the `_make_combine_callable` function to be usable here (rather than copy-pasting parts of it)?

https://github.com/mne-tools/mne-python/blob/056c273f1bb96bb3229e60874122f830f946746c/mne/viz/utils.py#L2975-L2998

HanBnrd:
> 
> 
> I had a quick look and the code looks reasonable (and is easy to read; thanks for using clear variable names!). See my comments below. For now let's not worry about the `dict.update()` approach, it would be nice to have but is not a blocker.

Thanks for your detailed feedback :)

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
             assert all(len(r) == self.info['nchan'] for r in self._read_picks)
         return self
 
+    def combine_channels(self, group_by, method='mean'):
+        """Combine channels based on regions of interest (ROIs).
+
+        Parameters
+        ----------
+        group_by : dict
```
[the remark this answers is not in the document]
HanBnrd: Will change it to `groups` then, as it is simpler and "channels" is already in the method name.

mne/channels/channels.py
HanBnrd: Good point on giving the user control on this, I can implement a `keep_stim` boolean argument. Should it defaults to something? Maybe to `False` to avoid unexpected behaviour?

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+                                        tmin=self.times[0])
+            elif isinstance(self, Evoked):
+                this_inst = EvokedArray(this_group_data, this_info,
+                                        tmin=self.times[0])
+            if inst is None:
+                inst = this_inst
+            else:
+                inst.add_channels([this_inst])
+
+        return inst
```
HanBnrd, line 1101: The [Contributing to MNE webpage](https://mne.tools/stable/install/contributing.html#return-types) says that methods should return `self` and functions return copies, should I then return a modified `self` instead, @drammock?

mne/channels/channels.py
```diff
@@ -987,6 +989,118 @@ def add_channels(self, add_list, force_update_info=False):
...
+        group_by : dict
+            Specifies which channels are aggregated into a single channel, with
+            aggregation method determined by the ``method`` parameter. One new
+            pseudo-channel is made per dict entry; the dict values must be
+            lists of picks (integer indices of ``ch_names``). For example::
+
+                group_by=dict(Left_ROI=[1, 2, 3, 4], Right_ROI=[5, 6, 7, 8])
+
+            Note that within a dict entry all channels must have the same type.
+        %(method)s
```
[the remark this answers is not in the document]
HanBnrd: I think I can probably write the doc manually for the `method` argument. My initial plan was to use the docdict of the `combine` argument found in multiple viz functions, but I think "combine" makes less sense than "method" in combine_channels. Also the `method` argument is a bit different as it does not accept None or 'gfp' (but should it?).
*&rarr; Linked to https://github.com/mne-tools/mne-python/pull/8014#discussion_r455322276*

mne/channels/channels.py
HanBnrd: This function was pretty handy but not exactly what I wanted, as `combine_channels` does not accept None or 'gfp' currently. In this case I feel like it would probably make `_make_combine_callable` messy to modify it (in terms of documentation with the docdict, and error messages). Please let me know what you think.
*&rarr; Linked to https://github.com/mne-tools/mne-python/pull/8014#discussion_r455740903*

agramfort:
2 things:
- method should operate inplace
- I am -1 on having both real EEG channels and virtual/combined EEG
channels in the same object.

2 options:
- use a function not a method that returns a new instance of an XxxArray
class
- remove all data / channels from the instance when adding these new
signals.

I am leaning towards a simple function.

>

mne/channels/channels.py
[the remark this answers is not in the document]
drammock: fair enough.  The functions are quite similar but not identical; we can worry about a possible deduplication refactor later.

mne/channels/channels.py
drammock: +1 for `keep_stim` with default `False`

mne/channels/channels.py
drammock: I think I agree with @agramfort that supporting a mix of real and virtual channels is probably a bad idea (at the very least it might cause havoc for topography plotting functions). And if I had to guess, I would say that in some cases users will want to keep the original object around (i.e., *not* modify in place), and the resulting object will at least be smaller than the original, so let's turn this into a function rather than a method.  Another consequence of that choice is that you don't have to support TFR objects right away, we can add that support later if you're not up for it right now.

HanBnrd:
Just did the modifications based on your comments @drammock and @agramfort, thanks. Will implement tests now.

HanBnrd:
I was actually thinking about turning verbose off when creating the new instances with XxxArray, is that something I should do with eg. `RawArray(..., verbose=False)`?

larsoner:
I would set it to whatever the parent instance's `.verbose` value is

HanBnrd:
> 
> 
> I would set it to whatever the parent instance's `.verbose` value is

Thanks @larsoner, I could do that if that's fine. The thing is that it is creating a new instance with one channel for every ROI, the console may become flooded with many ROIs if verbose is on for the parent's instance.
The function appends an new instance with one channel at every loop iteration to gradually create the instance with all the ROIs. I have never worked with instances bigger than memory so far but I though (maybe wrongly) that it could eventually save memory to append instances instead of numpy arrays? If not, I can take a different approach where I append numpy arrays and finally create a new instance for better verbose?

larsoner:
> he thing is that it is creating a new instance with one channel for every ROI, the console may become flooded with many ROIs if verbose is on for the parent's instance.

I would create and return a single new instance with all the new (averaged) channels

agramfort:
> I would create and return a single new instance with all the new (averaged) channels

+1

HanBnrd:
I changed the code to create a single new instance. I also implemented tests, could you please give me some guidance on that?

rob-luke:
Looking great. I have one question about throwing an error if you try and combine less than two channels (one channel 😄).

Here is my situation. I have a dataset which I want to split in to two regions of interest. Each region of interest has 4 channels. But for one participant they have lots of long hair and a bad connection between sensors and the scalp. And for this participant 3 of the 4 channels in the ROI is marked as bad. However, the remaining channel is clean and usable. In this situation, to be consistent with the rest of the dataset, I would want to create an ROI even though it has one channel. I could write a custom script for this one participant, or have an if statement, but I think it would be cleaner to just use the same code and have a combined channel from one channel. 

Could we perhaps print a warning to the user but not error in this situation?

rob-luke:
> I also implemented tests, could you please give me some guidance on that?

They look nice. I checked the codecov patch report and you have hit most of your lines of code. And the tests make sense and are documented well. 

But because you asked... you have only tested with combining two channels (unless I have read wrong), so you could check it works for 3 + channels. You also check the mean result is the same a numpy. You could do the same for the other methods.

HanBnrd:
> 
> 
> Looking great. I have one question about throwing an error if you try and combine less than two channels (one channel 😄).
> 
> Here is my situation. I have a dataset which I want to split in to two regions of interest. Each region of interest has 4 channels. But for one participant they have lots of long hair and a bad connection between sensors and the scalp. And for this participant 3 of the 4 channels in the ROI is marked as bad. However, the remaining channel is clean and usable. In this situation, to be consistent with the rest of the dataset, I would want to create an ROI even though it has one channel. I could write a custom script for this one participant, or have an if statement, but I think it would be cleaner to just use the same code and have a combined channel from one channel.
> 
> Could we perhaps print a warning to the user but not error in this situation?

Yes you are right, thanks for pointing that out. I think that would be very handy, and I can see myself being in the same situation.

Actually, I'm thinking that also adding a parameter (eg. `drop_bad` as a boolean, default to False) to keep or drop bad channels when combining the ROIs would make this even easier. It would enable to specify only one ROI dict and iterate over many instances of Raw (or Epochs or Evoked) with different bad channels. It could eventually print a warning for dropped channels as well. What do you think?

HanBnrd:
Should I care about the order of the ROI dict as the insertion-order preservation nature of dict seems to be present only since [Python 3.7](https://docs.python.org/3/whatsnew/3.7.html) and the minimum supported Python version is 3.6?

larsoner:
We have been assuming dict order is preserved since it was in standard cpython even in 3.6. But looking into it more yes to be safe use OrderedDict. And I need to look into reverting our previous removals of OrderedDict elsewhere in the code base in case people use some other python implementation that does not guarantee order :(

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
                       for selection, picks in selections.items()}
 
     return selections
+
+
+def combine_channels(instance, groups, method='mean', keep_stim=False,
+                     drop_bad=False):
+    """Combine channels based on regions of interest (ROIs).
```
rob-luke, line 1546: You use the terms `groups` further down and only use the acronym ROI in the example below. Combinations are also not specified based on regions, but lists of channels/indices. So the current wording seems more about how you would use the function rather than what it does. Perhaps..

```suggestion
    """Combine channels based on specified channel grouping.
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    keep_stim : bool
+        If True (default False), keep stimulus channels.
+    drop_bad : bool
+        If True (default False), drop channels marked as bad.
+
+    Returns
+    -------
+    combined_inst : instance of Raw, Epochs, or Evoked
+        An MNE-Python object with channels combined for each ROI defined in the
+        ``groups`` parameter.
```
HanBnrd, line 1583:
```suggestion
        An MNE-Python object with channels combined for each group defined in
        the ``groups`` parameter.
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    instance : instance of Raw, Epochs, or Evoked
+        An MNE-Python object to combine the channels for. The object can be of
+        type Raw, Epochs, or Evoked.
+    groups : dict
+        Specifies which channels are aggregated into a single channel, with
+        aggregation method determined by the ``method`` parameter. One new
+        pseudo-channel is made per dict entry; the dict values must be lists of
+        picks (integer indices of ``ch_names``). For example::
+
+            groups=dict(Left_ROI=[1, 2, 3, 4], Right_ROI=[5, 6, 7, 8])
```
HanBnrd, line 1559:
```suggestion
            groups=dict(Left=[1, 2, 3, 4], Right=[5, 6, 7, 8])
```

HanBnrd:
Thanks @rob-luke 😃

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    method : str | callable
+        Which method to use to combine channels. If a :class:`str`, must be one
+        of 'mean', 'median', or 'std' (standard deviation). If callable, the
+        callable must accept one positional input (data of shape ``(n_channels,
+        n_times)``, ``(n_epochs, n_channels, n_times)``, or ``(n_evokeds,
+        n_channels, n_time)``) and return an :class:`array <numpy.ndarray>` of
+        shape ``(n_times)``, ``(n_epochs, n_times)``, or ``(n_evokeds,
+        n_times)``. For example with an instance of Raw::
+
+            method = lambda data: np.median(data, axis=0)
```
agramfort, line 1571: give an example for instance of Epochs too.

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+            raise ValueError('"method" must be a callable, or one of '
+                             f'"mean", "median", or "std"; got "{method}".')
+
+    # Create new instance
+    if not isinstance(keep_stim, bool):
+        raise TypeError('"keep_stim" must be of type bool, not '
+                        f'{type(keep_stim)}.')
+    if keep_stim is True:
+        try:
+            combined_inst = instance.copy().pick_types(meg=False, stim=True)
```
agramfort, line 1615: this is bad practice. I you have raw of size 2GB you allocated another 2GB of RAM to get only one channel out. Use mne.pick_types and raw[] syntax. It will be way more efficient.

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    if drop_bad is False:
+        ch_idx_bad = []
+    elif not instance.info['bads']:
+        ch_idx_bad = []
+    else:
+        ch_idx_bad = pick_channels(instance.info['ch_names'],
+                                   instance.info['bads'])
+
+    # Check correctness of combinations
+    for this_group, these_picks in groups.items():
```
agramfort, line 1640:
```suggestion
    for this_group, this_picks in groups.items():
```
is more common idiom.

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+                             f'"{this_group}" contains types {types}.')
+        # Remove bad channels
+        these_bads = [idx for idx in these_picks if idx in ch_idx_bad]
+        these_picks = [idx for idx in these_picks if idx not in ch_idx_bad]
+        if these_bads:
+            logger.info('Dropped the following channels in group '
+                        f'{this_group}: {these_bads}')
+        #  Check if combining less than 2 channel
+        if len(set(these_picks)) < 2:
+            warn(f'Less than 2 channels in group "{this_group}"; cannot '
```
agramfort, line 1658: you warn here so you actually can

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+        # If all good create more detailed dict without bad channels
+        groups[this_group] = dict(picks=these_picks, ch_type=this_ch_type[0])
+
+    # Combine channels and add them to the new instance
+    group_ch_names, group_ch_types, group_data = [], [], []
+    for this_group, this_group_dict in groups.items():
+        group_ch_names.append(this_group)
+        group_ch_types.append(this_group_dict['ch_type'])
+        these_picks = this_group_dict['picks']
+        this_data = np.take(instance._data, these_picks, axis=ch_axis)
```
agramfort, line 1669: idem. Don't use private attributes. If its starts with _ it's a private attribute you should not use outside of the class definition.

mne/channels/tests/test_channels.py
```diff
@@ -360,4 +361,58 @@ def test_equalize_channels():
...
+    combine_channels(raw, good, drop_bad=True)
+    combine_channels(raw_ch_bad, good, drop_bad=True)
+
+    # Test result with one ROI
+    good_single = dict(foo=[0, 1, 3, 4])  # good grad
+    combined_mean = combine_channels(raw, good_single, method='mean')._data
+    combined_median = combine_channels(raw, good_single, method='median')._data
+    combined_std = combine_channels(raw, good_single, method='std')._data
+    foo_mean = np.mean(raw._data[good_single['foo']], axis=0)
+    foo_median = np.median(raw._data[good_single['foo']], axis=0)
```
agramfort, line 387: don't use _data

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+
+            groups=dict(Left=[1, 2, 3, 4], Right=[5, 6, 7, 8])
+
+        Note that within a dict entry all channels must have the same type.
+    method : str | callable
+        Which method to use to combine channels. If a :class:`str`, must be one
+        of 'mean', 'median', or 'std' (standard deviation). If callable, the
+        callable must accept one positional input (data of shape ``(n_channels,
+        n_times)``, ``(n_epochs, n_channels, n_times)``, or ``(n_evokeds,
+        n_channels, n_time)``) and return an :class:`array <numpy.ndarray>` of
```
drammock, line 1567:
```suggestion
        n_channels, n_times)``) and return an :class:`array <numpy.ndarray>` of
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+        shape ``(n_times)``, ``(n_epochs, n_times)``, or ``(n_evokeds,
```
drammock, line 1568:
```suggestion
        shape ``(n_times,)``, ``(n_epochs, n_times)``, or ``(n_evokeds,
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+        n_times)``. For example with an instance of Raw::
+
+            method = lambda data: np.median(data, axis=0)
+
+        Defaults to ``mean``.
+    keep_stim : bool
+        If True (default False), keep stimulus channels.
```
drammock, line 1575:
```suggestion
        If ``True``, include stimulus channels in the resulting object.
        Defaults to ``False``.
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+        of 'mean', 'median', or 'std' (standard deviation). If callable, the
+        callable must accept one positional input (data of shape ``(n_channels,
+        n_times)``, ``(n_epochs, n_channels, n_times)``, or ``(n_evokeds,
+        n_channels, n_time)``) and return an :class:`array <numpy.ndarray>` of
+        shape ``(n_times)``, ``(n_epochs, n_times)``, or ``(n_evokeds,
+        n_times)``. For example with an instance of Raw::
+
+            method = lambda data: np.median(data, axis=0)
+
+        Defaults to ``mean``.
```
drammock, line 1573:
```suggestion
        Defaults to ``'mean'``.
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    keep_stim : bool
+        If True (default False), keep stimulus channels.
+    drop_bad : bool
+        If True (default False), drop channels marked as bad.
```
drammock, line 1577:
```suggestion
        If ``True``, drop channels marked as bad before combining. Defaults to
        ``False``.
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+
+    Returns
+    -------
+    combined_inst : instance of Raw, Epochs, or Evoked
+        An MNE-Python object with channels combined for each group defined in
+        the ``groups`` parameter.
```
drammock, line 1583:
```suggestion
        An MNE-Python object of the same type as the input ``inst``, containing
        one virtual channel for each group in ``groups`` (and, if ``keep_stim``
        is ``True``, also containing stimulus channels).
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,150 @@ def make_1020_channel_selections(info, midline="z"):
...
+    """
+    from ..io import BaseRaw, RawArray
+    from ..epochs import BaseEpochs
+    from ..evoked import Evoked
```
drammock, line 1587: you could import these from `..` as well; they're exposed at the root level of the package. (unimportant; FYI)

mne/channels/tests/test_channels.py
```diff
@@ -360,4 +361,58 @@ def test_equalize_channels():
...
+    assert np.array_equal(combined_std, np.expand_dims(foo_std, axis=0))
+
+    # Test bad cases
+    raw_no_preload = read_raw_fif(raw_fname, preload=False)
+    raw_no_stim = read_raw_fif(raw_fname, preload=True)
+    raw_no_stim.pick_types(meg=True, stim=False)
+    bad1 = dict(foo=[0, 376], bar=[5, 2])  # out of bounds
+    bad2 = dict(foo=[0, 2], bar=[5, 2])  # type mix in same group
+    pytest.raises(RuntimeError, combine_channels, raw_no_preload, good)
+    pytest.raises(ValueError, combine_channels, raw, good, method='bad_method')
```
drammock, line 400: In new or updated tests, we are migrating to using `pytest.raises` as a context manager. Example:

```suggestion
    with pytest.raises(ValueError, match='"method" must be a callable, or '):
        combine_channels(raw, good, method='bad_method') 
```

Where `match` is a regex that will match the actual error message that is raised.

mne/channels/tests/test_channels.py
```diff
@@ -360,4 +361,58 @@ def test_equalize_channels():
...
+    pytest.raises(ValueError, combine_channels, raw_no_stim, good,
+                  keep_stim=True)
+    pytest.raises(ValueError, combine_channels, raw, bad1)
+    pytest.raises(ValueError, combine_channels, raw, bad2)
+
+    # Test warnings
+    warn1 = dict(foo=[375, 375], bar=[5, 2])  # same channel in same group
+    warn2 = dict(foo=[375], bar=[5, 2])  # one channel (last channel)
+    warn3 = dict(foo=[0, 4], bar=[5, 2])  # one good channel left
+    pytest.warns(RuntimeWarning, combine_channels, raw, warn1)
```
drammock, line 412: Same deal here:

```suggestion
    with pytest.warns(RuntimeWarning, match='Less than 2 channels in group'):
        combine_channels(raw, warn1)
```

mne/channels/channels.py
```diff
@@ -1536,3 +1539,141 @@ def make_1020_channel_selections(info, midline="z"):
...
+                        f'{type(keep_stim)}.')
+    if keep_stim:
+        stim_ch_idx = list(pick_types(inst.info, meg=False, stim=True))
+        if stim_ch_idx:
+            new_ch_names = [ch_names[idx] for idx in stim_ch_idx]
+            new_ch_types = [ch_types[idx] for idx in stim_ch_idx]
+            new_data = [np.take(inst_data, idx, axis=ch_axis)
+                        for idx in stim_ch_idx]
+        else:
+            warn('Could not find stimulus channels.')
```
HanBnrd, line 1625: Changed this to be a warning instead of an error, as I though it might be better if one wants to use this function when iterating over many instances and for some reason one has no stim channels... It's more convenient probably.

HanBnrd:
> 
> 
> last request. Can this be added to a tutorial or an example for documentation? maybe the EEG tutorial? basically to be put where you think users will expect to find it.

Sure, I think this function is probably suited for fNIRS, maybe in [Preprocessing functional near-infrared spectroscopy (fNIRS) data](https://mne.tools/stable/auto_tutorials/preprocessing/plot_70_fnirs_processing.html)? What do you think @rob-luke?

HanBnrd:
> 
> 
> Just looking through and its unclear to me if we are coping over all of the useful information from the original `inst` to the new one.
> I unfortunately don't have a full list of the useful information but the things that jump out to me are
> 
>     1. `raw.annotations` - you could just check if `inst` has an `annotations` attribute and copy it over if it does.
> 
>     2. stuff in `raw.info` like `lowpass`, `highpass`, `events` - I'm not sure of the best thing to do here. maybe somethign like the following might be helpful.
> 
> 
> ```
>     from mne.io.meas_info import _merge_info
> 
>     info = # create the info for just your new data. i.e what you already have
> 
>     # combine with inst.info
>     info_out = _merge_info([inst.info.copy(), info], force_update_to_first=True)
>     # only keep the new channels.
>     info_out.pick_channels(info.ch_names)
> ```
> 
> feel free to disregard if these points have already been addressed.

Thanks for your interest in this PR @bloyl. At the moment it is only keeping the bare minimum to create an MNE-Python object from an array. So ch_names, ch_types and sfreq.

Should I add support for this here @drammock @agramfort @larsoner @rob-luke or maybe create another feature request and PR after this one?

drammock:
> > last request. Can this be added to a tutorial or an example for documentation? maybe the EEG tutorial? basically to be put where you think users will expect to find it.
> 
> Sure, I think this function is probably suited for fNIRS, maybe in [Preprocessing functional near-infrared spectroscopy (fNIRS) data](https://mne.tools/stable/auto_tutorials/preprocessing/plot_70_fnirs_processing.html)?

Averaging channels isn't what I normally think of as "preprocessing".  I think this tutorial is probably a good spot for it: https://mne.tools/dev/auto_tutorials/evoked/plot_eeg_erp.html  It's one that I haven't had a chance to overhaul yet (I've been working through them all gradually); since I'll likely change the tutorial a lot when I get round to that one, any additions you make now could be pretty skeletal --- i.e., just a sentence saying "if you want to average across sensors to make an ROI, do this" and then a couple lines of code to average and then a plot.  Since the tutorial already defines `left` and `right` evoked objects, maybe plot a left vs right ROI during a left stimulus?

HanBnrd:
I added a few lines to the [EEG processing and Event Related Potentials (ERPs)](https://mne.tools/dev/auto_tutorials/evoked/plot_eeg_erp.html) tutorial, please let me know if that's okay for you @drammock :)

Also some tests are failing, but I'm not sure if that's related to my changes...

larsoner:
Windows 3.8 pip pre you can ignore for now, but CircleCI has a new warning (we don't allow warnings in tests):

https://app.circleci.com/pipelines/github/mne-tools/mne-python/2794/workflows/b19611f5-966c-4e98-8b27-04421717f505/jobs/21348

```
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/gen_gallery.py", line 164, in call_memory
    mem, out = memory_usage(func, max_usage=True, retval=True,
  File "/home/circleci/.local/lib/python3.8/site-packages/memory_profiler.py", line 343, in memory_usage
    returned = f(*args, **kw)
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/gen_rst.py", line 464, in __call__
    exec(self.code, self.fake_main.__dict__)
  File "/home/circleci/project/tutorials/evoked/plot_eeg_erp.py", line 117, in <module>
    raw_car, _ = mne.set_eeg_reference(raw, 'average', projection=True)
  File "<decorator-gen-211>", line 21, in set_eeg_reference
  File "/home/circleci/project/mne/io/reference.py", line 343, in set_eeg_reference
    warn('An average reference projection was already added. The data '
  File "/home/circleci/project/mne/utils/_logging.py", line 293, in warn
    warnings.warn_explicit(
RuntimeWarning: An average reference projection was already added. The data has been left untouched.
```
Which would be this line:

https://github.com/mne-tools/mne-python/pull/8014/files#diff-cdeb04d3326fc19e15cb55419752e417R117

You can either do `verbose='error'` or just write a comment saying that the data already have an average ref proj

larsoner:
I killed Travis, it will fail until #8050 lands, then I'll restart CIs

larsoner:
@HanBnrd if you merge with `master` or rebase, CIs should fix themselves

larsoner:
@HanBnrd let me know if you want help with the merge, but in theory you should just be able to do:
```
# git remote add upstream https://github.com/mne-tools/mne-python.git  <-- if you haven't already done this
git fetch upstream
git merge upstream master
git push
```

HanBnrd:
> 
> 
> @HanBnrd let me know if you want help with the merge, but in theory you should just be able to do:
> 
> ```
> # git remote add upstream https://github.com/mne-tools/mne-python.git  <-- if you haven't already done this
> git fetch upstream
> git merge upstream master
> git push
> ```

Sorry @larsoner I'm not sure I can merge `HanBnrd:combine-channels` into `mne-tools:master` myself as I don't have write access to `mne-tools/mne-python`. I cannot do it from GitHub and it does not work from command line with your instructions. If I should normally be able to do it, I would be happy to have some help please.

larsoner:
> If I should normally be able to do it, I would be happy to have some help please.

To be clear, I'm not saying that you should push/merge your changes into `github.com/mne-tools/mne-python.git` / click the green button like maintainers do. What I'm saying is that *on your local machine* you can perform a merge, and push that merge commit here. You can merge any branch into any other branch on your local machine, there are no restrictions (on at least trying; sometimes there can be conflicts).

For example, I can do it locally on my machine:
```
larsoner@bunk:~/python/mne-python$ git remote add HanBnrd git@github.com:/HanBnrd/mne-python.git

larsoner@bunk:~/python/mne-python$ git fetch HanBnrd 
remote: Enumerating objects: 107, done.
remote: Counting objects: 100% (107/107), done.
remote: Total 133 (delta 107), reused 107 (delta 107), pack-reused 26
...

larsoner@bunk:~/python/mne-python$ git checkout -b combine-channels HanBnrd/combine-channels 
Branch 'combine-channels' set up to track remote branch 'combine-channels' from 'HanBnrd'.
Switched to a new branch 'combine-channels'
# Now I am where you are on your local machine

larsoner@bunk:~/python/mne-python$ git merge upstream/master
# It pops up a vim (for me) window, and I just `:wq` to save-and-quit -- the merge commit message is fine
Auto-merging mne/channels/channels.py
Auto-merging doc/python_reference.rst
Auto-merging doc/changes/latest.inc
Merge made by the 'recursive' strategy.
...
```
And at this point, I could now do:
```
git push
```
And it would push the merge commit to your branch (assuming I have permission to do so, which is the case if you checked the "allow contributions from maintainers").

But you should be able to do it, too. If you do:
```
git remote add upstream git://github.com/mne-tools/mne-python.git
git fetch upstream
```
Now your system knows about all up-to-date branches out on GitHub. Then if you're locally on your combine-channels branch and you do:
```
git merge upstream/master
```
Then save-and-exit the prompt that comes up about what merge commit message to use, then you should be able to do:
```
git push origin combine-channels
```
Then we should see a merge commit pop up in your branch in this PR.

HanBnrd:
> 
> 
> > If I should normally be able to do it, I would be happy to have some help please.
> 
> To be clear, I'm not saying that you should push/merge your changes into `github.com/mne-tools/mne-python.git` / click the green button like maintainers do. What I'm saying is that _on your local machine_ you can perform a merge, and push that merge commit here. You can merge any branch into any other branch on your local machine, there are no restrictions (on at least trying; sometimes there can be conflicts).
> 
> For example, I can do it locally on my machine:
> 
> ```
> larsoner@bunk:~/python/mne-python$ git remote add HanBnrd git@github.com:/HanBnrd/mne-python.git
> 
> larsoner@bunk:~/python/mne-python$ git fetch HanBnrd 
> remote: Enumerating objects: 107, done.
> remote: Counting objects: 100% (107/107), done.
> remote: Total 133 (delta 107), reused 107 (delta 107), pack-reused 26
> ...
> 
> larsoner@bunk:~/python/mne-python$ git checkout -b combine-channels HanBnrd/combine-channels 
> Branch 'combine-channels' set up to track remote branch 'combine-channels' from 'HanBnrd'.
> Switched to a new branch 'combine-channels'
> # Now I am where you are on your local machine
> 
> larsoner@bunk:~/python/mne-python$ git merge upstream/master
> # It pops up a vim (for me) window, and I just `:wq` to save-and-quit -- the merge commit message is fine
> Auto-merging mne/channels/channels.py
> Auto-merging doc/python_reference.rst
> Auto-merging doc/changes/latest.inc
> Merge made by the 'recursive' strategy.
> ...
> ```
> 
> And at this point, I could now do:
> 
> ```
> git push
> ```
> 
> And it would push the merge commit to your branch (assuming I have permission to do so, which is the case if you checked the "allow contributions from maintainers").
> 
> But you should be able to do it, too. If you do:
> 
> ```
> git remote add upstream git://github.com/mne-tools/mne-python.git
> git fetch upstream
> ```
> 
> Now your system knows about all up-to-date branches out on GitHub. Then if you're locally on your combine-channels branch and you do:
> 
> ```
> git merge upstream/master
> ```
> 
> Then save-and-exit the prompt that comes up about what merge commit message to use, then you should be able to do:
> 
> ```
> git push origin combine-channels
> ```
> 
> Then we should see a merge commit pop up in your branch in this PR.

Okay, thanks for the explanation :)

drammock:
Here is [the rendered tutorial](https://21408-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_eeg_erp.html#evoked-response-averaged-across-channels-by-roi).  The plot looks odd; the two ROIs appear to be mirror images of one another?

drammock:
Perhaps I'm being too nitpicky, but maybe a more realistic example is to hand-construct a dict of just a few channels from each side (usually an ROI is not "all electrodes on the left half of the head") instead of using the for-loop to sort by odd/even channel loc.  WDYT?

HanBnrd:
Yeah okay, I can select channels around the motor cortex left vs. right.

HanBnrd:
This is what I have when I average those sensors (left vs. right) for left auditory:
sensors
combine_channels
What do you think?

drammock:
That looks better. Thanks, and sorry for the last-minute change request...  I didn't foresee (but should have) how strange the left half / right half ROI signals would look.

drammock:
thanks @HanBnrd!

HanBnrd:
Thank you all for your support :)

PR #8017, opened 2020-07-15
MRG, BUG: Fix legend bgcolor and Travis macOS

larsoner:
Non-scraping-related parts of #7927, which is on hold until we figure out how to avoid memory problems.


agramfort:
thx @larsoner

PR #8019, opened 2020-07-16
fix plot method of TFR

### LINKED ISSUE #8013: BUG: TFR.plot misc issues
I am experiencing several issues with the `.plot` method of TFR objects. I suspect that there are several things off, but for now I found this obvious bug:

These are valid inputs for the `title` parameter:

https://github.com/mne-tools/mne-python/blob/638847ea5539be32919b6ac4dd85ef8c4c3b79aa/mne/time_frequency/tfr.py#L1136-L1139

Yet in the code, we have this logic:

https://github.com/mne-tools/mne-python/blob/638847ea5539be32919b6ac4dd85ef8c4c3b79aa/mne/time_frequency/tfr.py#L1269-L1278

Hence, when I pass `title='auto'`, my figure will have the title `auto`

sappelhoff:
closes #8013

### DIFF
--- a/mne/time_frequency/tfr.py
+++ b/mne/time_frequency/tfr.py
@@ -1132,15 +1132,17 @@ class AverageTFR(_BaseTFR):
             If true, colorbar will be added to the plot. For user defined axes,
             the colorbar cannot be drawn. Defaults to True.
         show : bool
-            Call pyplot.show() at the end.
+            Whether to call pyplot.show() for each plotted figure or not.
+            Defaults to ``True``.
         title : str | 'auto' | None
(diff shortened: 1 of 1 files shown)


sappelhoff:
@larsoner I am not really using this anymore, and my fixes here are not perfect either (e.g., it's a bit weird to return multiple figs IMO).

Should I just close this PR or do you (or somebody else) want to salvage it?

If so, please feel free to take over or close :-)

larsoner:
I'll close for now but leave the issue open -- will resurrect this if/when necessary

PR #8020, opened 2020-07-16
MRG: bump spyder version

### LINKED ISSUE #8009: MAINT, DOC: bump spyder-kernels
Once Spyder 4.1.4 hits the anaconda `defaults` channel, we'll want to update the install instructions to state that the `base` env should have `spyder>=4.1.4`, and update our `environment.yml` to require `spyder-kernels>=1.9.2`. This is necessary because on OSX Catalina (and presumably also on Big Sur) there is an issue with spyder running under py3.7 in base being unable to set the console to use a py3.8 executable (which currently happens by default when installing with our env file).

Check here periodically to see if the relevant versions are out yet:

- https://anaconda.org/anaconda/spyder
- https://anaconda.org/anaconda/spyder-kernels  (can use pip to get 1.9.2 if the conda release lags spyder 4.1.4)

Crossref spyder-ide/spyder#13252

drammock:
closes #8009

currently WIP because the fix doesn't actually work in my local testing; waiting to hear back from Spyder devs.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -23,7 +23,7 @@ dependencies:
 - traitsui
 - imageio
 - tqdm
-- spyder-kernels
+- spyder-kernels>=1.9.2
 - imageio-ffmpeg>=0.4.1
 - vtk>=9.0.1
 - pyvista>=0.24
--- a/environment.yml
+++ b/environment.yml
@@ -23,7 +23,7 @@ dependencies:
 - traitsui
 - imageio
 - tqdm
-- spyder-kernels>=1.9.2
+- spyder-kernels>=1.10.0
 - imageio-ffmpeg>=0.4.1
 - vtk>=9.0.1
 - pyvista>=0.24
--- a/doc/install/mne_python.rst
+++ b/doc/install/mne_python.rst
@@ -165,7 +165,7 @@ your operating system.
    with ``curl``) and use it to create a new environment (named ``mne`` by
    default)::
 
-       $ conda install --name base nb_conda_kernels
+       $ conda install --name base nb_conda_kernels "spyder>=4.2.1"
        $ curl --remote-name https://raw.githubusercontent.com/mne-tools/mne-python/master/environment.yml
        $ conda env update --file environment.yml
 
@@ -173,7 +173,7 @@ your operating system.
 
    - Download the `environment file`_
    - Open an Anaconda command prompt
-   - Run :samp:`conda install --name base nb_conda_kernels`
+   - Run :samp:`conda install --name base nb_conda_kernels "spyder>=4.2.1"`
    - :samp:`cd` to the directory where you downloaded the file
    - Run :samp:`conda env update --file environment.yml`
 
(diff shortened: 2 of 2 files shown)


drammock:
@ccordoba12 says he can successfully set a py3.8 interpreter within an instance of spyder running in a py3.7 environment, on a bare-metal mac (although I've not been able to replicate this when running OSX catalina in a VM); see spyder-ide/spyder#13252 for discussion. Do any of our devs have a mac running catalina where they could double-check / confirm that this works?  The procedure is:

```bash
conda create -n test-spyder python=3.7 spyder
conda create -n test-py38 python=3.8 spyder-kernels
conda activate test-spyder
spyder &
```
...and then go into spyder preferences and try to set `Tools > Preferences > Python Interpreter > Use the following interpreter` to the value of `conda activate test-py38 && python -c "import sys; print(sys.executable)"`.  Then close the running ipython console (should be py3.7) and see if the automatically spawned replacement console is py3.8.

Once we get confirmation, this PR is ready for review/merge.  All CIs are happy except codecov (why, codecov? this PR only touches docs and the environment file)

larsoner:
@drammock I'm on Catalina and ran those commands (after `conda update -n base conda` just to be safe) and when I do Python->Preferences, then Python Interpreter, then paste in the value of `conda activate test-py38 && python -c "import sys; print(sys.executable)"` I get the error:
```
Invalid file path:
conda activate test-py38 && python -c "import sys; print(sys.executable)"
```

larsoner:
FWIW it looks like it's Spyder 4.1.5 according to the "About Spyder"

drammock:
@larsoner my instructions were not clear. **in a terminal**, run `conda activate test-py38 && python -c "import sys; print(sys.executable)"` and paste the path that it returns into the "use the following interpreter" box.

larsoner:
For me that's `/Users/larsoner/opt/miniconda3/envs/test-py38/bin/python` and it appears not to work:

Screen Shot 2020-10-14 at 4 50 39 PM

drammock:
@larsoner can you confirm that you do indeed have `spyder-kernels` installed in that env and it's a high enough version?
`conda activate test-py38 && conda list spyder-kernels`

larsoner:
```
$ conda activate test-py38 && conda list spyder-kernels
# packages in environment at /Users/larsoner/opt/miniconda3/envs/test-py38:
#
# Name                    Version                   Build  Channel
spyder-kernels            1.9.4                    py38_0  
```

larsoner:
@drammock thoughts on this for 0.22, or just bump to 0.23?

drammock:
spyder devs have failed to replicate.  Personally I don't have the patience to debug this any further; working inside the OSX VM is really slow for me (I have all the wrong muscle memories, and some keyboard shortcuts get stolen by the host OS).  Unless an actual mac user is willing to take over here and in spyder-ide/spyder#13252, I think we have 2 choices:

1. close this and pretend nothing is wrong.
2. assume that yours and my experiences are some kind of fluke or user error, and go ahead and rebase+merge this PR.

larsoner:
I dug into it and replied to the upstream issue -- they require spyder_kernels < 1.10 but 1.10 is the only version available for 3.8. Let's see what they say

drammock:
this one should (finally) be good to go

environment.yml
```diff
@@ -23,7 +23,7 @@ dependencies:
 - traitsui
 - imageio
 - tqdm
-- spyder-kernels
+- spyder-kernels>=1.9.2
```
ccordoba12, line 26: This should be `>=1.10.0`, if you plan to use Spyder 4.2.1

drammock: thanks for the heads-up @ccordoba12 !

drammock:
```suggestion
- spyder-kernels>=1.10.0
```

larsoner:
Thanks @drammock @ccordoba12 !

PR #8021, opened 2020-07-17
MNT: Notebook testing

GuillaumeFavelier:
This PR fixes the failures with the notebook job on Azure.

### DIFF
--- a/server_environment.yml
+++ b/server_environment.yml
@@ -1,4 +1,4 @@
-name: base
+name: notebook
 channels:
 - conda-forge/label/vtk_dev
 - conda-forge
@@ -20,5 +20,5 @@ dependencies:
   - pyvista==0.24
   - ipywidgets
   - nbformat
-  - nbclient
+  - nbclient<=0.4.0
   - jupyter_client!=6.1.5
--- a/server_environment.yml
+++ b/server_environment.yml
@@ -9,16 +9,16 @@ dependencies:
 - mesalib
 - ffmpeg
 - vtk
-- numpy
 - scipy
 - traits
 - pip:
   - mne
+  - numpy<1.19.0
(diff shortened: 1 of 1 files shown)


server_environment.yml
```diff
@@ -9,15 +9,15 @@ dependencies:
 - mesalib
 - ffmpeg
 - vtk
-- numpy
 - scipy
 - traits
 - pip:
   - mne
+  - numpy<1.19.0
```
larsoner, line 16: This is weird, should probably just as the version constraint above, or do everything in pip rather than first installing NumPy above via conda then killing it by installing one over it in pip

larsoner: (scipy depends on NumPy, so removing just NumPy from the conda list is essentially the same as leaving it in)

GuillaumeFavelier: The kernel dies so I will move scipy and try again

larsoner: FWIW I'm sure this solution works, is just not clean. It also might do some weird things with MKL at some point (for example if we switch to the base channel rather than conda forge).

Basically I'm suggesting either pin above (easier if conda actually supports or), or move everything that requires NumPy to the pip section (harder to ensure is implemented and maintained properly)

GuillaumeFavelier: Almost everything uses pip. We need conda mainly because of mesalib and vtk (not sure it's needed anymore).

larsoner: Actually, `vtk` also requires `numpy` so it should also be moved. I'll push a commit

larsoner:
Okay even if the solution is not ideal, it at least seems to work, so I'll merge

larsoner:
Thanks @GuillaumeFavelier

PR #8023, opened 2020-07-17
MRG, ENH: Add mne.viz.centers_to_edges for pcolormesh

larsoner:
Should fix https://app.circleci.com/pipelines/github/mne-tools/mne-python/2708/workflows/e86b53b9-e443-45f8-bc27-108e24b27ad1/jobs/21253. I've wanted this function a few times before when using `pcolormesh`, so might as well make it public.


agramfort:
thx @larsoner !

PR #8033, opened 2020-07-20
MRG, ENH: Add proj="reconstruct" option

### LINKED ISSUE #3184: ENH: Reconstruction?
Would people be interested in looking at the extent to which e.g. Evoked data can be reconstructed following projections (un-doing the induced signal bias)? Here is the idea:

```
import os.path as op

import mne
from mnefun import plot_reconstruction

raw = mne.io.read_raw_fif(op.join(data_path, 'MEG', 'sample',
                          'sample_audvis_raw.fif'), preload=True)
raw.pick_types(meg=True, eeg=False, eog=True, stim=True, exclude=())
events = mne.find_events(raw, 'STI 014')
raw.add_proj(mne.preprocessing.compute_proj_eog(raw)[0])
raw.add_proj(mne.preprocessing.compute_proj_ecg(raw)[0])
epochs = mne.Epochs(raw, events, {'Left auditory': 1}, -0.2, 0.5,
                    baseline=(None, 0), proj='delayed',
                    reject=dict(grad=4000e-13, mag=4e-12, eog=150e-6))
evoked = epochs.average()
evoked.plot_reconstruction()
```

Would give you something like:

figure_1

The "reconstructed" data is obtained by projecting the (SSP-projected) data to a sphere using MNE and coming back out. The critical step is that the projection operator is applied to the lead fields as well as the data, thereby undoing the bias to the data.

In Xfit, this is known as reconstruction "on" vs "off", and helps visualize whether or not your projectors have eliminated a true signal of interest. Techniques like dipole fitting and minimum-norm estimates both do this type of compensation during estimation, so it gives you a better idea in sensor space of how much of your signal of interest is still there.

larsoner:
Closes #3184

### DIFF
--- a/tutorials/preprocessing/plot_50_artifact_correction_ssp.py
+++ b/tutorials/preprocessing/plot_50_artifact_correction_ssp.py
@@ -182,8 +182,9 @@ fig = epochs.average().plot_topomap(times, proj='interactive')
 
 ###############################################################################
 # Plotting the ERP/F using ``evoked.plot()`` or ``evoked.plot_joint()`` with
-# and without projectors applied can also be informative.
-#
+# and without projectors applied can also be informative, as can plotting with
+# ``proj='reconstruct'``, which can reduce the signal bias introduced by
+# projections (see :ref:`tut-artifact-ssp-reconstruction` below).
 #
 # Example: EOG and ECG artifact repair
 # ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -282,9 +283,9 @@ mne.viz.plot_projs_topomap(ecg_projs, info=raw.info)
 #
 # By default, the filtered epochs will be averaged together
 # before the projection is computed; this can be controlled with the boolean
-# ``average`` parameter.
-#
-# .. TODO what is the (dis)advantage of **not** averaging before projection?
+# ``average`` parameter. In general this improves the signal-to-noise (where
+# "signal" here is our artifact!) ratio because the artifact temporal waveform
+# is fairly similar across epochs and well time locked to the detected events.
 #
 # To get a sense of how the heartbeat affects the signal at each sensor, you
 # can plot the data with and without the ECG projectors:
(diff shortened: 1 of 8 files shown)


larsoner:
https://21286-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#reducing-ssp-bias-in-sensor-space-via-signal-reconstruction

tutorials/preprocessing/plot_50_artifact_correction_ssp.py
```diff
@@ -402,6 +403,40 @@
...
 #
+# .. _tut-artifact-ssp-reconstruction:
+#
+# Reducing SSP bias in sensor space via signal reconstruction
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# In addition to looking at unprojected versus projected data, it's also useful
+# to look at sensor space data reconstructed to reduce the bias introduced by
+# projection. This reconstruction is performed internally by mapping to
+# source space and back via a minimum-norm solution. The reconstructed data
+# can be show by using ``proj='reconstruct'`` in evoked plotting functions,
```
drammock, line 414:
```suggestion
# can be shown by using ``proj='reconstruct'`` in evoked plotting functions,
```

tutorials/preprocessing/plot_50_artifact_correction_ssp.py
```diff
@@ -402,6 +403,40 @@
 # projectors are capturing something else other than a heartbeat artifact (and
 # thus may be removing brain signal and should be discarded).
 #
+# .. _tut-artifact-ssp-reconstruction:
+#
+# Reducing SSP bias in sensor space via signal reconstruction
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# In addition to looking at unprojected versus projected data, it's also useful
+# to look at sensor space data reconstructed to reduce the bias introduced by
```
drammock, line 411: I think this is the first mention in the tutorial of "bias introduced by projection" and thus I think it needs explaining. Can you add a new sentence before this one, explaining that projection introduces bias, what kind of bias it is, how/why it happens, when it might be a problem for your analysis, etc?

larsoner: [Better](https://21289-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#visualizing-ssp-sensor-space-bias-via-signal-reconstruction) ?

drammock: yep. thanks!

agramfort:
@larsoner merge if happy

larsoner:
> I think this can dramatically change the way sensor level stats are done

Only if we make `evoked.reconstruct_proj()` public, right now you can only do it in plotting ... :)

agramfort:
can you open an issue so we don't forget about ICA option?

PR #8035, opened 2020-07-20
WIP, MAINT: Fix Azure python pre failures

larsoner:
Let's see if we can get more information about where things are crashing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,15 +142,15 @@ script:
         pip install -e .;
         python mne/tests/test_evoked.py;
       fi;
-    - echo 'pytest -m "${CONDITION}" --cov=mne -n 1 ${USE_DIRS}'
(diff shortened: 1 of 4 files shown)


larsoner:
I have no idea why disabling `pytest-xdist` by removing `-n 1` seems to have fixed things on Azure. I'll restart the problematic Azure run to double check. Maybe there was a new version of NumPy or SciPy pre that kicked in. (It also broke our tab test due to some weird `tqdm` thing that I'm pushing a commit to fix.)

At some point we could also try `pytest-forked` and use `-n 1 --forked`, though it has a warning that it's not really maintained :(

larsoner:
This doesn't necessarily fix the problem, but it does clean up our CI approach, so I'll go ahead and merge

PR #8037, opened 2020-07-21
MRG, DOC: note on using picard for same solution as other algos

sappelhoff:
Picard can be used to obtain the same solution as with FastICA, Infomax, or extended Infomax --> and apparently (according to Picard :wink: ) converges faster.

This sounds like a win for me, so I added this to the docs. WDYT?

@larsoner @cbrnr @agramfort @mmagnuski @pierreablin 

Info taken from README: https://github.com/pierreablin/picard

Previously discussed in [gitter](https://gitter.im/mne-tools/mne-python?at=5f16ce4815e5e72fb3679875), Thanks @cbrnr for making me aware of it.

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -297,6 +297,17 @@ class ICA(ContainsMixin):
     see :class:`~sklearn.decomposition.FastICA`, :func:`~picard.picard`,
     :func:`~mne.preprocessing.infomax`.
 
+    .. note:: Picard can be used to solve the same problems as FastICA,
+              Infomax, and extended Infomax, but typically converges faster
+              than either of those methods. To make use of Picard's speed while
+              still obtaining the same solution as with other algorithms, you
+              need to specify ``method='picard'`` and ``fit_params`` as a
+              dictionary with the following combination of keys:
+
+              - ``dict(ortho=False, extended=False)`` for Infomax
+              - ``dict(ortho=False, extended=True)`` for extended Infomax
+              - ``dict(ortho=True, extended=True)`` for FastICA
+
     Reducing the tolerance (set in ``fit_params``) speeds up estimation at the
     cost of consistency of the obtained results. It is difficult to directly
     compare tolerance levels between Infomax and Picard, but for Picard and


cbrnr:
Looks good - I think this is something that people are not very much aware of. We should advertise this on the EEGLAB mailing list 😄.

sappelhoff:
however, digging into this, I also saw that picard is currently not tested for Python 3 --> https://github.com/pierreablin/picard/pull/22

cbrnr:
CIs show some package conflicts (traits) - do we fix these before merging?

GuillaumeFavelier:
I would suggest to update with `master` (0b3ad0344d041590635bb380c3880849a0d21881) because I don't see any `-n` used in `.travis` anymore.

cbrnr:
@sappelhoff can you rebase?

cbrnr:
Alright, now seeing the errors described in #8040. Any ideas?

GuillaumeFavelier:
I compared with a recently merged PR and `pyface` has been bumped from `6.1.2` to `7.0.1`. That could explain the new requirement?

cbrnr:
Probably. But which package is causing this runtime error then? There is no `conda` error when installing packages.

cbrnr:
Testing in #8042.

GuillaumeFavelier:
I think it's because `pyface 7.0.1` is available on conda only recently?

image

cbrnr:
You mean they might have overlooked the dependency on `traits`? `conda info pyface` shows `traits` as a dependency without any version restriction (but maybe it should be `>=6).

GuillaumeFavelier:
It's just a guess since it's the first big change I notice. But this could explain why there is no error during installation.

cbrnr:
Yes. And according to https://docs.enthought.com/pyface/changelog.html#release-7-0-1, `pyface` >= 7 requires `traits` >= 6 (which is not available in `conda` yet).

cbrnr:
`traits` 6.1.0 is already in the channel (https://anaconda.org/anaconda/traits/files), so maybe this issue will resolve itself.

cbrnr:
Some package apparently pins `traits` to <6 because `conda update traits` says that the latest version is already installed (5.2.0).

cbrnr:
Maybe someone who actually knows what they are doing could help, because I don't use `conda` and I can't figure out what the conflict is (@hoechenberger maybe?).

cbrnr:
Also, let's fix this issue in #8040 or #8042 instead of cluttering this PR.

hoechenberger:
Great move, @sappelhoff! @agramfort was **waiting** for somebody to step in and finally file this kind of PR ;) Good to see this happen!! We've been using Picard in my old lab and it worked really well. It's also our default now in the `mne-study-template` :)

mne/preprocessing/ica.py
```diff
@@ -297,6 +297,17 @@ class ICA(ContainsMixin):
     see :class:`~sklearn.decomposition.FastICA`, :func:`~picard.picard`,
     :func:`~mne.preprocessing.infomax`.
 
+    .. note:: Picard can be used to solve the same problems as FastICA,
+              Infomax, and extended Infomax, but typically converges faster
+              than either of those methods. To make use of Picard's speed while
+              still obtaining the same solution as with other algorithms, you
+              need to specify ``method='picard'`` and ``fit_params`` as a
+              dict with the following combination of keys:
```
hoechenberger, line 305:
```suggestion
              dictionary with the following combination of keys:
```

agramfort:
thx @sappelhoff

PR #8041, opened 2020-07-22
MRG: Multiple fixes in PyVista 3d backend

GuillaumeFavelier:
This PR fixes multiple things with the PyVista 3d backend.

ToDo:

- [x] Fixes the bug with `representation` in the PyVista 3d backend.
- [x] Send pre-computed normals to the `renderer.mesh()` function when possible
- [x] Retrieve `nn` infos from the surface parameter and use them as normals data in `renderer.surface()`
- [x] ~Use `smooth_shading=True` when the normals are not given~ **OR** use same parameters as mayavi
- [x] Refactor the `mesh` functions to avoid code duplication

master |PR
---------|------
image | image

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -814,9 +814,9 @@ def _compute_normals(mesh):
         point_normals=True,
         split_vertices=False,
         flip_normals=False,
-        consistent_normals=True,
+        consistent_normals=False,
         auto_orient_normals=False,
-        non_manifold_traversal=False,
+        non_manifold_traversal=True,
         inplace=True,
     )
 
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -49,14 +49,12 @@ class _Figure(object):
                  size=(600, 600),
                  shape=(1, 1),
                  background_color='black',
-                 smooth_shading=False,
                  off_screen=False,
                  notebook=False):
         self.plotter = plotter
         self.plotter_class = plotter_class
         self.display = display
         self.background_color = background_color
-        self.smooth_shading = smooth_shading
         self.notebook = notebook
 
         self.store = dict()
@@ -144,12 +142,11 @@ class _Renderer(_BaseRenderer):
 
     def __init__(self, fig=None, size=(600, 600), bgcolor='black',
                  name="PyVista Scene", show=False, shape=(1, 1),
-                 notebook=None, smooth_shading=False):
+                 notebook=None):
         from .renderer import MNE_3D_BACKEND_TESTING
         from .._3d import _get_3d_option
         figure = _Figure(show=show, title=name, size=size, shape=shape,
-                         background_color=bgcolor, notebook=notebook,
-                         smooth_shading=smooth_shading)
+                         background_color=bgcolor, notebook=notebook)
         self.font_family = "arial"
         self.tube_n_sides = 20
         self.shape = shape
@@ -245,7 +242,6 @@ class _Renderer(_BaseRenderer):
                 rgba=rgba, opacity=opacity, cmap=colormap,
                 backface_culling=backface_culling,
                 rng=[vmin, vmax], show_scalar_bar=False,
-                smooth_shading=self.figure.smooth_shading,
                 interpolate_before_map=interpolate_before_map,
                 style=representation, line_width=line_width, **kwargs,
             )
@@ -308,7 +304,6 @@ class _Renderer(_BaseRenderer):
                 color=color,
                 cmap=colormap,
                 opacity=opacity,
-                smooth_shading=self.figure.smooth_shading
             )
             return actor, contour
 
@@ -341,7 +336,6 @@ class _Renderer(_BaseRenderer):
                 opacity=opacity,
                 cmap=cmap,
                 backface_culling=backface_culling,
-                smooth_shading=self.figure.smooth_shading,
             )
 
     def sphere(self, center, color, scale, opacity=1.0,
@@ -364,7 +358,6 @@ class _Renderer(_BaseRenderer):
                 self.plotter,
                 mesh=glyph, color=color, opacity=opacity,
                 backface_culling=backface_culling,
-                smooth_shading=self.figure.smooth_shading
             )
             return actor, glyph
 
@@ -392,7 +385,6 @@ class _Renderer(_BaseRenderer):
                     color=color,
                     show_scalar_bar=False,
                     cmap=cmap,
-                    smooth_shading=self.figure.smooth_shading,
                 )
         return tube
 
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -238,10 +238,6 @@ class _Renderer(_BaseRenderer):
                     colormap = colormap.astype(np.float64) / 255.
                 from matplotlib.colors import ListedColormap
                 colormap = ListedColormap(colormap)
-            if 'Normals' in mesh.point_arrays:
-                smooth_shading = False
-            else:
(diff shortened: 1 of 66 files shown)


GuillaumeFavelier:
The issue is that the `reprensentation` parameter (named `style` in PyVista) was sent to the private `renderer._add_mesh()` function instead of `renderer.mesh()`.

larsoner:
Looks like the normals are still not fixed, so the abstraction layer is still missing something

GuillaumeFavelier:
It appears that the normals are most of the time computed by VTK for the mayavi backend too. 

https://github.com/mne-tools/mne-python/blob/3d3bb9a6b05f9521973c43f62d0a36e266a02385/mne/viz/backends/_pysurfer_mayavi.py#L362-L367

Even though the abstraction layer supports receiving precomputed normals, they would need to be computed on-the-fly in most cases.

larsoner:
Ahh I see. So the question is, why can't we use `smooth_shading=True` in PyVista, but it has always worked fine in Mayavi? Do we need to use the same set of options?

GuillaumeFavelier:
> why can't we use smooth_shading=True in PyVista

I think this is related to an issue on macos during migration to pyvista 0.25/vtk 9 if i'm not mistaken

GuillaumeFavelier:
> Do we need to use the same set of options?

That's a good idea :)

larsoner:
> I think this is related to an issue on macos during migration to pyvista 0.25/vtk 9 if i'm not mistaken

Yes but the question is why do we hit it on PyVista but not Mayavi? Maybe we won't anymore now that we supply the normals in `brain`. I think it's worth trying to make sure all the options are the same, then setting `smooth_shading=True` in `PyVista` (when we don't supply normals) and see if it works. Or if you can't set the options the same in PyVista, do our own VTK-based computation that's equivalent to Mayavi (or fix how PyVista does it in `pyvista` itself).

GuillaumeFavelier:
In terms of visual quality, it's certainly better to have smooth shading :star_struck:

larsoner:
We now get on Linux the same sort of segfault we've been seeing on macOS:

https://travis-ci.org/github/mne-tools/mne-python/jobs/710809861

GuillaumeFavelier:
> We now get on Linux the same sort of segfault we've been seeing on macOS

Good news that it's consistent then :sweat_smile: ?

I'll try the custom VTK filter now.

GuillaumeFavelier:
I try first with a homemade private function and if it works I'll use PyVista [`compute_normals()`](https://docs.pyvista.org/core/filters.html?highlight=compute_normals#pyvista.PolyDataFilters.compute_normals) with the good parameters.

larsoner:
Sounds good. Although if it does turn out to be a parameters issue, we should definitely open a PyVista issue about it as segfaulting with default parameters is not good. Maybe they need to be triaged by VTK version or something -- and maybe there is an underlying VTK bug, again worth whittling down to a minimal example, as the VTK folks have generally been very responsive when I've brought them issues that are easily reproducible.

GuillaumeFavelier:
For reference, here are the parameters used for normal computation (for smooth shading) by the 3d backends ('x' for True, ' ' for False):

function | mayavi | pyvista
------------|----------|-----------
cell_normals | |
point_normals | x | x
split_vertices | | 
flip_normals | | 
consistent_normals | | x
auto_orient_normals | | 
non_manifold_traversal | | x

When using those values with PyVista (`smooth_shading=True`), Travis fails as @larsoner describes in https://github.com/mne-tools/mne-python/pull/8041#issuecomment-662580458 (version 7c6b9bf).

When forcing PyVista to use the same parameters as Mayavi, the CIs are happy (version 30af751) therefore I suppose that either `consistent_normals` or `non_manifold_traversal` (or both) cause the failures.

I can't reproduce those failures locally so I will push some test commits to clarify the situtation.

*Note: default values for PyVista are available in [the documentation](https://docs.pyvista.org/core/filters.html?highlight=compute_normals#pyvista.PolyDataFilters.compute_normals)*

larsoner:
Great! Worth opening an issue about the defaults being unsafe in PyVista?

GuillaumeFavelier:
> Worth opening an issue about the defaults being unsafe in PyVista?

When the situation is clarified here, I'll report it. I can't reproduce it locally after all.

larsoner:
This will be red until #8050 lands

larsoner:
Green (except for patch diff which we can probably ignore)!

GuillaumeFavelier:
I did my best playing with my `TTY` and `Xvfb` but although I'm on linux I can't reproduce this locally, I am now sure this does not happen consistently because I restarted https://github.com/mne-tools/mne-python/pull/8041#issuecomment-662580458 and the red job became green @larsoner.

2e6f434, 90c42fb and 7868c18 show that it's not because of the parameters of `compute_normals()`.

I did 52655ec to start from scratch because I was too focused on `mesh()` and `surface()` and then f5c8910 made me think that the issue is related to `sphere()` somehow: everything is green when I use the homemade `_compute_normals()` until I force `sphere()` to compute normals although it's not necessary (MacOS fails consistently from that though).

Now I tend to think that the segfault happens when the normals are actually *recomputed* ([normals are always computed](https://github.com/pyvista/pyvista/blob/0.25.3/pyvista/plotting/plotting.py#L1389-L1399) as long as `smooth_shading=True`) but once again I need to confirm it. I think a dedicated minimal test can help here.

larsoner:
> I think a dedicated minimal test can help here.

Feel free to add one. And also during long debugging sessions like this you can speed up iterations by changing `azure_pipelines.yml` and `.travis.yml` calls to `pytest` to isolate for example``pytest ... mne/viz/`. Then each run comes back in 15 minutes instead of 30

mne/viz/backends/_pyvista.py
```diff
@@ -339,7 +341,7 @@ def surface(self, surface, color=None, opacity=1.0,
                 opacity=opacity,
                 cmap=cmap,
                 backface_culling=backface_culling,
-                smooth_shading=self.figure.smooth_shading
+                smooth_shading=self.figure.smooth_shading,
```
larsoner, line 344: Shouldn't this just be `False` since they're either added or computed above?

GuillaumeFavelier: I remove it in the commit just after

tutorials/source-modeling/plot_background_freesurfer_mne.py
```diff
@@ -331,7 +331,7 @@ def imshow_mri(data, img, vox, xyz, suptitle):
 # sphere, a given vertex in the source (sample) mesh can be mapped easily
 # to the same location in the destination (fsaverage) mesh, and vice-versa.
 
-renderer_kwargs = dict(bgcolor='w', smooth_shading=False)
+renderer_kwargs = dict(bgcolor='w')
```
larsoner, line 334: This is the one place we actually don't want smooth shading. Is there any way to disable it on this example? It makes it clear that the mesh is decimated quite a bit. Or maybe we should just use wireframes to make the point? WDYT?

larsoner: Yeah I think wireframes could look good for the last plots:

https://mne.tools/dev/auto_tutorials/source-modeling/plot_background_freesurfer_mne.html#surface-decimation

GuillaumeFavelier: Using could be a bit messy. I will let this option then :+1:

GuillaumeFavelier:
I modified `mesh()` and `surface()` to be light interfaces over the `polydata()` function. I also removed `_add_polydata_actor()` in favor of `polydata()` too. What do you say @larsoner ?

larsoner:
Thanks @GuillaumeFavelier !

Issue #8040, opened 2020-07-22
conda environment.yml broken?


cbrnr:
Currently, all CIs using `conda` with our `environment.yml` are broken because of unsatisfiable dependencies. It has something to do with `traits`, which is available as 5.2.0, but some other package (`pyface`?) wants >=6.0.0. I tested this locally, but everything works fine, so maybe the problematic packages have already been updated upstream. I can't restart Travis jobs for some reason, maybe @sappelhoff can you do it in #8037?

Related issue: I noticed that `pyqt` 5.9.2 gets installed automatically as a dependency by `conda`, but then we overwrite it with a newer version of `PyQt5` available in `pip`. This could be problematic and might explain issues related to PyQt that pop up from time to time.

sappelhoff:
![image](https://user-images.githubusercontent.com/9084751/88157538-2c68e580-cc0b-11ea-8f9e-21ffabb76076.png)

https://travis-ci.org/github/mne-tools/mne-python/builds/710508815

cbrnr:
Now the first jobs have a different error:

```
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n
```

cbrnr:
There doesn't seem to be a `-n` argument (`pytest -m "${CONDITION}" --tb=short --cov=mne -n 1 ${USE_DIRS}`).

GuillaumeFavelier:
`-n` is used by the `pytest-xdist` plugin. It's not installed?

cbrnr:
Looks like not. Where is the requirements file that contains this package?

cbrnr:
@sappelhoff could you maybe also restart the failing Azure job?

sappelhoff:
done, but I can only re-run the failed ones. Can only the PR owner or an admin re-run? I think you and I should have same permissions @cbrnr

cbrnr:
The failed ones are OK for now. I didn't find the restart button in Travis (although I was logged in and I know where the button usually is - now there is just a "More options" button). I haven't figured out how to restart Azure jobs, I'm logged in and now I see only a "Cancel" button for the running job.

sappelhoff:
>  I haven't figured out how to restart Azure jobs, I'm logged in and now I see only a "Cancel" button for the running job.

you do it from the GitHub UI, go to "details" and then there is an expanded dropdown menu on the left

cbrnr:
[...]

PR #8042, opened 2020-07-22
Use pyface < 7

cbrnr:
Fixes (works around) #8040 hopefully.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - psutil
 - numexpr
 - traits>=4.6.0
-- pyface>=6
+- pyface>=6,<7
 - traitsui>=6
 - imageio>=2.6.1
 - pip:
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - psutil
 - numexpr
 - traits
-- pyface
+- pyface<7
 - traitsui
 - imageio
 - tqdm
--- a/environment.yml
+++ b/environment.yml
@@ -8,7 +8,7 @@ dependencies:
 - scipy
 - matplotlib
 - numba
-- pandas>=0.18
+- pandas
 - xlrd
 - scikit-learn
 - h5py
@@ -18,13 +18,15 @@ dependencies:
 - joblib
 - psutil
 - numexpr
-- traits>=4.6.0
+- traits
 - pyface>=6,<7
-- traitsui>=6
-- imageio>=2.6.1
+- traitsui
+- imageio
+- numpydoc
+- tqdm
+- spyder-kernels
 - pip:
   - mne
-  - https://github.com/numpy/numpydoc/archive/master.zip
   - imageio-ffmpeg>=0.4.1
   - vtk>=9.0.1
   - pyvista>=0.24
@@ -36,6 +38,4 @@ dependencies:
   - nilearn
   - neo
   - python-picard
-  - tqdm
   - PyQt5>=5.10
-  - spyder-kernels
--- a/environment.yml
+++ b/environment.yml
@@ -19,7 +19,7 @@ dependencies:
 - psutil
 - numexpr
 - traits
-- pyface>=6,<7
+- pyface
 - traitsui
 - imageio
 - numpydoc
--- a/environment.yml
+++ b/environment.yml
@@ -22,7 +22,6 @@ dependencies:
 - pyface
 - traitsui
 - imageio
-- numpydoc
 - tqdm
 - spyder-kernels
 - pip:


cbrnr:
This seems to work, but it would be better if we used the most recent `pyface` (7.0.1). The problem is that for some reason, `traits` is pinned to 5.2.0, but `pyface` 7.0.1 requires the latest `traits` 6.1.0 (which is available in the default channel).

environment.yml
```diff
@@ -18,13 +18,15 @@ dependencies:
...
 - numexpr
-- traits>=4.6.0
-- pyface>=6
-- traitsui>=6
-- imageio>=2.6.1
+- traits
+- pyface>=6,<7
+- traitsui
+- imageio
+- numpydoc
```
larsoner, line 25: We can just get rid of this actually, but that can be another PR. This one seems to be headed toward green (other than our recent pip-pre 3.8 segfaults that are unrelated)

cbrnr: What do you mean? Remove these four packages completely from the env?

larsoner: No just `numpydoc`

larsoner:
Do we know which package actually requires `pyface<7`?

environment.yml
cbrnr: OK, I'll remove `numpydoc`. What about the other packages I moved from `pip` to `conda`? OK or revert?

cbrnr:
> Do we know which package actually requires `pyface<7`?

No, I tried to let `conda` solve the conflict, but after working for 15 minutes it spat out a very long list of conflicts that didn't really help me (because every package seemed to conflict with several other packages).

larsoner:
Which OS? I tried on macOS and Linux and in both cases `conda update conda` followed by `conda create -n test -f environment.yml` from `master` worked, in both cases installing pyface 7.0.1. So maybe this was just a transient issue?

larsoner:
To check, maybe remove the `numpydoc` and the `<7` pin from `pyface` and see if it succeeds here?

environment.yml
```diff
@@ -18,13 +18,15 @@ dependencies:
...
+- tqdm
+- spyder-kernels
```
larsoner, line 27: Not sure if we can do this yet, have you checked with #8009 ?

larsoner: Looks like we probably can since #8009 mentions 1.9.2 and that's what's in conda now

larsoner: ... but I guess it should be:
```suggestion
- spyder-kernels>=1.9.2
```

cbrnr: Shouldn't `conda` always install the latest available version unless there is a conflict with other (installed) packages? Let's see if CIs come back happy - if we don't test for the behavior in #8009 I will manually test it.

larsoner: I think it does when creating from scratch but I'm not sure what it does if you do `conda env update -f environment.yml` or something similar. But maybe we don't need to worry too much about this case

cbrnr: Yes, I would only support creating the env from scratch, everything else will drive us directly into dependency hell.

cbrnr:
Nope, doesn't work. The error log mentions `qt4 = traitsui.qt4:toolkit` - do we require Qt4 (instead of Qt5) somewhere?

larsoner:
If you change it to `traits>=6` it should hopefully fail earlier. It's in win64:

https://anaconda.org/anaconda/traits

So some package must require it. I'll fire up the Windows VM for local debugging...

cbrnr:
Also, how did `conda create -n test -f environment.yml` (as you've mentioned [here](https://github.com/mne-tools/mne-python/pull/8042#issuecomment-662453071)) work? First, `-f` should be `--file`, but even then do I get an error because the file wants to install in an env called `mne`. That's why I have to use `conda env update --file environment.yml`.

larsoner:
Sorry I meant `conda env create` not `conda create`

cbrnr:
> Sorry I meant `conda env create` not `conda create`

Thanks, this works (pretty inconsistent argument naming I would say). It does give the same results as `conda env update --file environment.yml` though, which is `traits==5.2.0` on my macOS machine.

Why do you think this is related to Windows? The latest version is on conda for all platforms, and we get the issue on a Linux CI (and on my local macOS).

larsoner:
I thought it was installing 6 on macOS and Linux but it wasn't. Changing it to have `- traitsui>=6` and `- traits>=6` fixes things for me on Linux (`pytest mne/gui` didn't work before but does now). No idea why it tries to pull `traits==5.2.0` if you don't tell it to use 6+. For some reason it installs Python 3.6 though...

larsoner:
Adding a `python>=3.8` or even `python>=3.7` does create conflicts. Argh...

cbrnr:
This whole conda thing is a mess. We mix/overwrite with pip because the default channel lags behind recent versions. Should we consider moving to `conda-forge` as @hoechenberger has repeatedly suggested? It looks like such a situation can occur anytime with our env file with the defaults channel.

cbrnr:
(Moving to `conda-forge` also has the advantage that we can use `mamba` instead of `conda`, which is much faster and has more features).

larsoner:
> Should we consider moving to conda-forge as @hoechenberger has repeatedly suggested? It looks like such a situation can occur anytime with our env file with the defaults channel.

I've seen problems with `conda-forge` due to packages not being ready for prime time in other repos that use `conda-forge`, so I wouldn't think of this as something that will solve our problems. It might fix more problems than it creates, but it might also be the reverse.

In any case, this seems to be some `base` problem. Even this causes a conflict:
```
name: mne
channels:
- defaults
dependencies:
- python>=3.8
- traits>=6
- pyface>=7
- traitsui>=6
```
So maybe we should manually pin `pyface` until they sort it out, and raise an upstream issue.

cbrnr:
Agreed - can you raise the upstream issue? You already have a great minimal example. I can pin `pyface` to the old version in the meantime.

cbrnr:
It would still be worth giving `conda-forge` a try, I'd assume that it will solve more problem than it creates (especially the whole `pyqt`/`PyQt5` `conda`/`pip` business).

hoechenberger:
> It would still be worth giving `conda-forge` a try, I'd assume that it will solve more problem than it creates (especially the whole `pyqt`/`PyQt5` `conda`/`pip` business).

Yes this could definitely be worth a shot. The current approach with conda + pip leads to inconsistent environments that behave in the strangest manner if you try to install or update any conda packages. Not a fan.

larsoner:
Thanks @cbrnr

larsoner:
https://github.com/conda/conda/issues/10095

drammock:
by the way, it would have been fine to specify `spyder-kerneles>=1.9.2` as @larsoner suggested.  It just wouldn't actually fix the problem with macOS Catalina.  But no need to go back and do it; it's already WIP in #8020

PR #8049, opened 2020-07-23
FIX: smooth_mat logic in _Brain

### LINKED ISSUE #8048: BUG: pyvista error when ploting stc without hemi data
I have a pipeline where i compute some scalars on a set of labels which I map back to an `stc` for visualization using `stc = mne.labels_to_stc(labels, vals)`.

`stc.plot(...)` works with mayavi but will now fail with pyvista with the following error 

> ~/Projects/mne-python/mne-python-devel/mne/viz/_brain/_brain.py in set_time_point(self, time_idx)
>    1138 
>    1139                 # interpolate in space
> -> 1140                 smooth_mat = hemi_data['smooth_mat']
>    1141                 if smooth_mat is not None:
>    1142                     act_data = smooth_mat.dot(act_data)
> 
> KeyError: 'smooth_mat'

I think this is caused by some hemisphere checks where `hemi_data['smooth_mat']` is generated.

To replicate the issue just run `/examples/connectivity/plot_mne_inverse_envelope_correlation.py` using the pyvista backend and commenting out the following lines.
https://github.com/mne-tools/mne-python/blob/607fb4613fb5a80dd225132a4a53fe43b8fde0fb/examples/connectivity/plot_mne_inverse_envelope_correlation.py#L95-L96

I'm not sure if this is a bug in `mne.labels_to_stc` not setting something it should or a `_brain` bug or both.

GuillaumeFavelier:
This PR fixes the way the smoothing matrix is retrieved from internal data and solves the example described in https://github.com/mne-tools/mne-python/issues/8048#issue-664151155

mayavi | pyvista
----------|-----------
image | image

Closes #8048

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1136,7 +1136,7 @@ class _Brain(object):
                     time_actor.SetInput(time_label(self._current_time))
 
                 # interpolate in space
-                smooth_mat = hemi_data['smooth_mat']
+                smooth_mat = hemi_data.get('smooth_mat')
                 if smooth_mat is not None:
                     act_data = smooth_mat.dot(act_data)
 


GuillaumeFavelier:
Let me know if this solves your issue @bloyl

bloyl:
Thanks @GuillaumeFavelier for the quick response.

This does indeed solve my issue.

PR #8050, opened 2020-07-23
MRG, MAINT: Cleaner resource use and test tols

larsoner:
Testing on my Windows machine I found that some of the tolerances were too strict (minor linalg differences when reg=0 in beamformer or for gamma_map). Also I noticed that there were un-freed resources in `test_what.py`, which I think is causing this in `master` [on Azure](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=8305&view=logs&j=c0130b29-789d-5a3c-6978-10796a508a7f&t=5f931690-334a-58bc-2c61-e2793b49abfb):
```
mne\io\tests\test_utils.py .                                             [ 41%]
##[error]Cmd.exe exited with code '-1073741819'.
```
This fixes it by being safer about `fiff_open`.


larsoner:
Actually I think the Azure error is really https://github.com/numpy/numpy/issues/16913. I'll do a full run here with `-vvs` just to confirm, but it does seem to be the case that a simple `np.dot` is the problem [in this Azure run](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=8324&view=logs&j=c0130b29-789d-5a3c-6978-10796a508a7f&t=5f931690-334a-58bc-2c61-e2793b49abfb).

PR #8054, opened 2020-07-24
Better default legend in plot_compare_evokeds() for list of Evokeds

hoechenberger:
When passing a list of Evoked's to `viz.plot_compare_evokeds()`, previously we would assign them sequential numbers (1 … N) as labels, which is … not necessarily too helpful.

But since by default we're also storing useful information about the condition(s) making up the Evoked in `Evoked.comment`, I thought we could make use of this to create more useful labels.

So now, for example, instead of getting this:
Screenshot 2020-07-23 at 14 41 16

we'll be getting this:
Screenshot 2020-07-24 at 10 55 53

without having to pass a dict to `plot_compare_evokeds`.

### DIFF
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -1926,9 +1926,12 @@ def plot_compare_evokeds(evokeds, picks=None, colors=None,
     ----------
     evokeds : instance of mne.Evoked | list | dict
         If a single Evoked instance, it is plotted as a time series.
+        If a list of Evokeds, the contents are plotted with their
+        ``.comment`` attributes used as condition labels. If no comment is set,
+        the index of the respectiv Evoked the list will be used instead,
+        starting with `1` for the first Evoked.
         If a dict whose values are Evoked objects, the contents are plotted as
-        single time series each and the keys are used as condition labels.
-        If a list of Evokeds, the contents are plotted with indices as labels.
+        single time series each and the keys are used as labels.
         If a [dict/list] of lists, the unweighted mean is plotted as a time
         series and the parametric confidence interval is plotted as a shaded
         area. All instances must have the same shape - channel numbers, time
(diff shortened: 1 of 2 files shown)


agramfort:
You should make sure comment is not None maybe

hoechenberger:
> What do you do if some evoked have comments and some have none? Mix comments with integers, or fall back to integers for all of them?

I think I would just mix it... make good use of all the info we have :)

agramfort:
Sounds good 👌

hoechenberger:
We're now using `.comment` if not-empty, and assign the list index + 1 as a label otherwise. For example, in the following figure I had removed `.comment` of the second Evoked before plotting:

Screenshot 2020-07-25 at 13 06 10

hoechenberger:
Rebased on `master` so documentation will hopefully build now.

I'm not sure yet about adding a test (i.e., how to test this best); I can look into this later today or tomorrow.

mne/viz/evoked.py
```diff
@@ -2129,8 +2132,20 @@ def plot_compare_evokeds(evokeds, picks=None, colors=None,
     # build up evokeds into a dict, if it's not already
     if isinstance(evokeds, Evoked):
         evokeds = [evokeds]
+
     if isinstance(evokeds, (list, tuple)):
-        evokeds = {str(idx + 1): evk for idx, evk in enumerate(evokeds)}
+        evokeds_copy = evokeds.copy()
+        evokeds = dict()
+        for evk_idx, evk in enumerate(evokeds_copy, start=1):
+            try:
```
agramfort, line 2140: Use hasattr and avoid the try please   Thx

hoechenberger: Done in 21b41c6

mne/viz/evoked.py
```diff
@@ -1926,9 +1926,12 @@ def plot_compare_evokeds(evokeds, picks=None, colors=None,
     ----------
     evokeds : instance of mne.Evoked | list | dict
         If a single Evoked instance, it is plotted as a time series.
+        If a list of Evokeds, the contents are plotted with their
+        ``.comment`` attributes used as condition labels. If no comment is set,
+        the index of the respectiv Evoked the list will be used instead,
+        starting with `1` for the first Evoked.
```
larsoner, line 1932: /home/circleci/project/mne/viz/evoked.py:docstring of mne.viz.plot_compare_evokeds:8: WARNING: py:obj reference target not found: 1
```suggestion
        starting with ``1`` for the first Evoked.
```

larsoner:
Thanks @hoechenberger

PR #8055, opened 2020-07-24
MRG, FIX: Pin PyQt5 for HiDPI bug

larsoner:
Until https://github.com/pyvista/pyvistaqt/issues/20 and https://gitlab.kitware.com/vtk/vtk/-/issues/17953 are fixed, we should pin PyQt5 < 5.14 on macOS.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
 numpy
 scipy
 matplotlib
-pyqt5
+pyqt5>=5.10,<5.14; platform_system == "Darwin"
+pyqt5>=5.10; platform_system == "Darwin"
 pyqt5-sip
 sip
 scikit-learn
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,7 +1,8 @@
 numpy
 scipy
 matplotlib
-pyqt5
+pyqt5>=5.10,<5.14; platform_system == "Darwin"
+pyqt5>=5.10; platform_system == "Darwin"
```
rob-luke, line 5: @larsoner should this be != ?

larsoner: Whoops, yes! Can you open a quick PR?

larsoner: Actually I can push a quick commit to fix it

rob-luke: Sorry I should have just opened a pr to start with.

PR #8056, opened 2020-07-24
MRG, MAINT: Clean up resource use

larsoner:
Clean up ResourceWarning in our codebase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -142,15 +142,15 @@ script:
         pip install -e .;
         python mne/tests/test_evoked.py;
       fi;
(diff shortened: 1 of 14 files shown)


larsoner:
Also

- removes `--showlocals` from pytest default config -- it makes the tracebacks super long even when `--tb=short` is used, and there is no way to disable it from the command line, so make it opt in.
- adds `-vv` to the runs so that the name of the test currently executing is printed (should help identify segfaults) -- previously the output was unusable due to `--showlocals` on error when combined with `-vv`, but we no longer have that problem
- fixes a matplotlib `master` deprecation involving `plt.colorbar`

PR #8059, opened 2020-07-25
[MRG] Bug fix - resample would break crop

bloyl:
resampling a raw would convert `raw._last_samps` to a list, which would break `raw.crop` which expected an `np.array`

This bug didn't exist in the last release, so I don't feel like this needs an entry in `latest.inc`. but if you think I should I can add it.


larsoner:
Thanks @bloyl

PR #8063, opened 2020-07-27
MRG, ENH: Add loose-as-dict for mixed source spaces

### LINKED ISSUE #2163: API for mixed source space inverses
So with Alan's work, we now have support for mixed source spaces. The forward modeling once you have a mixed source space seems straightforward (already in place). However, with the inverse, there are some use cases that need to be sorted out: Ideally, orientation constraints should be supported for the cortical surface source space and loose solutions for the deep structure(s). What should the API be?

Thoughts?

larsoner:
Closes #2163.

- [x] Check that it actually produces something reasonable
- [x] Add tests for bad inputs
- [x] Add tests that orientation priors are correct
- [x] Add tests for fixed surface + free vol (no idea if it works properly currently as some priors are actually zeroed out)
- [ ] ~~Maybe add some viz of volume?~~

### DIFF
--- a/mne/forward/forward.py
+++ b/mne/forward/forward.py
@@ -1004,7 +1004,7 @@ def _triage_loose(src, loose, fixed='auto'):
         if key != 'surface' and this_loose != 1:
             raise ValueError(
                 'loose parameter has to be 1 or "auto" for non-surface '
-                f'source spaces, got loose[{key:r}] = {this_loose}')
+                f'source spaces, got loose["{key}"] = {this_loose}')
         if not 0 <= this_loose <= 1:
             raise ValueError(
                 f'loose ({key}) must be between 0 and 1, got {this_loose}')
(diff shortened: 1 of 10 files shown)


larsoner:
[Output](https://21463-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_mixed_source_space_inverse.html#plot-the-surface) looks pretty reasonable so far

mne/forward/forward.py
```diff
@@ -967,48 +968,104 @@ def _select_orient_forward(forward, info, noise_cov=None, copy=True):
...
+        loose = float(loose)
+        loose = {key: loose for key in src_types}
+    loose_keys = set(loose.keys())
+    if loose_keys != src_types:
+        raise ValueError(
+            f'loose, if dict, must have keys {sorted(src_types)} to match the '
+            f'source space, got {sorted(loose_keys)}')
+    # if fixed is auto it can be ignored, if it's False it can be ignored,
+    # only really need to care about fixed=True
+    if fixed is True:
```
agramfort, line 992: we agree that fixed only applies to the surface?

larsoner: surface+discrete -- that's how it is in `master` and I agree we should only allow it for those (not volume)

agramfort:
needs more work here then?

>

larsoner:
Pushed a little commit with a test that pos+ori is not totally off

PR #8064, opened 2020-07-27
MRG, ENH: Volume rendering

### LINKED ISSUE #7648: ENH: Add volumetric rendering to _Brain
It would be nice to make MixedSourceEstimates and VolSourceEstimates plottable with _Brain instead of just surfaces.

Random ideas:

- Slicer is probably a good resource for ideas, and [context preserving](https://www.slicer.org/wiki/Modules:VolumeRendering-Documentation-3.5) looks cool
- [VTK example](https://lorensen.github.io/VTKExamples/site/Python/Medical/MedicalDemo4/) produces:

    Screenshot from 2020-04-21 12-24-59

- A random [ray-tracing lib, MIT license](https://github.com/Gjacquenot/PyRayTracer)
- [A recent issue about it](https://discourse.vtk.org/t/volume-rendering-with-vtk-cpu-based-ok-gpu-based-not-visible/475/8)
- [5-year-old post from VTK folks](https://blog.kitware.com/volume-rendering-improvements-in-vtk/)
- [5-year-old post, using Python](https://pyscience.wordpress.com/2014/11/16/volume-rendering-with-python-and-vtk/)

Alternatively we could go with isocontours, but seems harder to control/configure

larsoner:
Closes #7648

- [x] Vector initialization
- [x] Vector updates with time-clicking and clims
- [x] Volume initialization (should be unified / not rely on vector-ness, just happen in `set_time_point` like everything else)
- [x] Volume updates (moving scalar setting to `set_time_point` should fix this)
- [x] Volume updates with time-clicking
- [x] Actually make it look good
- [x] Make it work for non-vector `stc.plot`
- [x] Volume-only plotting
- [x] Fix bug with pos_lims for volumetric plotting
- [x] Fix bug with volumetric plotting on macOS + VTK9 (https://github.com/pyvista/pyvista/issues/850)
- [x] Add colorbar for volume-only (needs appropriate actor, which we don't have for vol -- maybe fake one?)
- [x] Add terrain interaction
- [x] Volume updates with clims (currently requires changing the time point for some reason)
- [x] Fix bugs with multi-views and `hemi='split'` with vectors, actors, etc.
- [x] Deal with `offset` / `surface` issues, esp. when `hemi='split'` and/or `hemi='lh' / 'rh'` only
- [x] Add GFP
- [x] Add picking of max point from the volumes like we have for left/right hemi already
- [x] Clicking on volume: needs to be maximum along the ray
- [x] Clicking: needs to remove any spheres in ray path, if no spheres, then add new points
- [x] Add horizontal layout
- [x] Add QSplitter for configurable sizing
- [x] Add `show_traces=<float>` support to specify the fraction to dedicate to the time viewer
- [x] Fix actor transfer mapping for volume-only case (maybe use the `color_tf` directly?)
- [x] Divergent volume
- [x] Fix existing broken test
- [x] Sort out how to add actors to scenes (do we need different meshes? different actors? neither?)
- [x] Add tests for point picking of volumes and mixed
- [x] Add tests for divergent, view_layout, volume_options

@GuillaumeFavelier I think I have this at a point where I've worked out the transformations and data injection:

Screenshot from 2020-07-27 17-02-27

If you `mne.minimum_norm.write_inverse_operator('mixed-inv.fif', inverse_operator)` the inv operator from the modified `plot_mixed_source_space_inverse.py`, you can iterate with this much quicker script:

<details>

```
import os.path as op
import mne
from mne.minimum_norm import read_inverse_operator, apply_inverse

# Set dir
subject = 'sample'
data_dir = op.join(data_path, 'MEG', subject)
fname_evoked = data_dir + '/sample_audvis-ave.fif'
fname_cov = data_dir + '/sample_audvis-shrunk-cov.fif'
fname_trans = data_dir + '/sample_audvis_raw-trans.fif'
condition = 'Left Auditory'
evoked = mne.read_evokeds(fname_evoked, condition=condition,
                          baseline=(None, 0))
noise_cov = mne.read_cov(fname_cov)
inv = read_inverse_operator('mixed-inv.fif')
stc_vec = apply_inverse(evoked, inv, pick_ori='vector')
    stc_vec.plot(views='rostral', hemi='both', src=inv['src'],
                 trans=fname_trans, initial_time=0.180,
                 subjects_dir=subjects_dir)
```

</details>

Can you take over and fit this into the `_Brain` + PyVista framework better? I have a checklist at the top for how I think it should go but feel free to update and change and push commits as you want.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ before_install:
     # Always install these via pip so we get the latest possible versions (testing bugfixes)
     - pip install --upgrade -r requirements_testing.txt codecov
     - if [ "${DEPS}" != "minimal" ]; then
-        pip install nitime;
+        pip install nitime pytest-xdist;
       fi
     # Don't source mne_setup_sh here because changing PATH etc. can't be done in a script
     - if [ "${DEPS}" != "minimal" ]; then
@@ -143,7 +143,7 @@ script:
         python mne/tests/test_evoked.py;
       fi;
     - echo 'pytest -m "${CONDITION}" --tb=short --cov=mne -vv ${USE_DIRS}'
-    - pytest -m "${CONDITION}" --tb=short --cov=mne -vv ${USE_DIRS}
+    - pytest -m "${CONDITION}" --tb=short --cov=mne -vv -n 1 ${USE_DIRS}
     # run the minimal one with the testing data
     - if [ "${DEPS}" == "minimal" ]; then
         export MNE_SKIP_TESTING_DATASET_TESTS=false;
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ before_install:
     # Always install these via pip so we get the latest possible versions (testing bugfixes)
     - pip install --upgrade -r requirements_testing.txt codecov
     - if [ "${DEPS}" != "minimal" ]; then
-        pip install nitime pytest-xdist;
+        pip install nitime;
       fi
     # Don't source mne_setup_sh here because changing PATH etc. can't be done in a script
     - if [ "${DEPS}" != "minimal" ]; then
@@ -143,7 +143,7 @@ script:
         python mne/tests/test_evoked.py;
       fi;
     - echo 'pytest -m "${CONDITION}" --tb=short --cov=mne -vv ${USE_DIRS}'
-    - pytest -m "${CONDITION}" --tb=short --cov=mne -vv -n 1 ${USE_DIRS}
+    - pytest -m "${CONDITION}" --tb=short --cov=mne -vv ${USE_DIRS}
     # run the minimal one with the testing data
     - if [ "${DEPS}" == "minimal" ]; then
         export MNE_SKIP_TESTING_DATASET_TESTS=false;
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -27,6 +27,8 @@ Changelog
 
 - Add :class:`mne.MixedVectorSourceEstimate` for vector source estimates for mixed source spaces, by `Eric Larson`_
 
+- Add mixed and volumetric source estimate plotting using volumetric ray-casting to :meth:`mne.MixedSourceEstimate.plot` and :meth:`mne.VolSourceEstimate.plot_3d` by `Eric Larson`_
+
 - Add :meth:`mne.MixedSourceEstimate.surface` and :meth:`mne.MixedSourceEstimate.volume` methods to allow surface and volume extraction by `Eric Larson`_
 
 - Add :meth:`mne.VectorSourceEstimate.project` to project vector source estimates onto the direction of maximum source power by `Eric Larson`_
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -596,6 +596,7 @@ numpydoc_xref_aliases = {
     'VolSourceEstimate': 'mne.VolSourceEstimate',
     'VolVectorSourceEstimate': 'mne.VolVectorSourceEstimate',
     'MixedSourceEstimate': 'mne.MixedSourceEstimate',
+    'MixedVectorSourceEstimate': 'mne.MixedVectorSourceEstimate',
     'SourceEstimate': 'mne.SourceEstimate', 'Projection': 'mne.Projection',
     'ConductorModel': 'mne.bem.ConductorModel',
     'Dipole': 'mne.Dipole', 'DipoleFixed': 'mne.DipoleFixed',
--- a/examples/inverse/plot_mixed_source_space_inverse.py
+++ b/examples/inverse/plot_mixed_source_space_inverse.py
@@ -136,8 +136,9 @@ src = inverse_operator['src']
 initial_time = 0.1
 stc_vec = apply_inverse(evoked, inverse_operator, lambda2, inv_method,
                         pick_ori='vector')
-brain = stc_vec.plot(hemi='both', src=inverse_operator['src'],
-                     initial_time=initial_time, subjects_dir=subjects_dir)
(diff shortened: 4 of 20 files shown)


larsoner:
Actually @GuillaumeFavelier if you want a more realistic image, use an oct6 surface and spacing=7. volume, these are more like what people would actually use. The example says the spacing higher just for speed.

GuillaumeFavelier:
I would be glad to help with this. I'm not sure what you mean by "oct6 surface" though. Also we have a `plot_volume_source_estimates()` function, is there a way to reuse that in this case or it's not relevant?

larsoner:
Not relevant, it's an orthoview / slice viewer in matplotlib. We could do something like this in VTK as well but it would be an entirely different function and probably only for volumes.

Oct6 refers to the argument to setup_source_space, it determines how densely the mesh is sampled

agramfort:
I just tried and personally I find it hard to see where the volume rendered
activity maps to in the brain.

the orthoview is much easier to use to see where things happen and it can
make publication ready figures
while here I would say no as it is.

grumpy Alex

larsoner:
> I just tried and personally I find it hard to see where the volume rendered activity maps to in the brain.

Indeed it looks bad currently, but I haven't tried at all to make it look good. Be patient :)

FWIW MIP (maximum intensity projection) mapping is commonly used -- even in stc_vol.plot() -- and it's useful, so I think it's likely we can come up with something that's helpful here.

agramfort:
please prove me wrong :)

>

larsoner:
Pushed a commit to enable maximum intensity projection, now I find it pretty easy to see where the activity comes from

larsoner:
(still a lot of work to do to make it prettier and work better, but it should at least give you some hope for the future instead of despair :) )

larsoner:
Okay I find it much more usable / understandable now. Feel free to try again @agramfort. Updated example [here](https://21475-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_mixed_source_space_inverse.html#plot-the-mixed-source-estimate), with updated image (looks better with pos=7. or smaller, this uses 10.):

larsoner:
Updated the [existing example](https://21488-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_mixed_source_space_inverse.html#plot-the-mixed-source-estimate):



And added volume-only mode and now use it the [LCMV example](https://21488-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d) (cc @britta-wstnr ):

agramfort:
ok it does look nice... ;)

>

larsoner:
[Example 1](https://21501-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d):



[Example 2](https://21501-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_mixed_source_space_inverse.html#plot-the-mixed-source-estimate):

drammock:
😍

larsoner:
[Fixed](https://21505-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d) two-sided volumes/data:



I think I just need to add some tests, fix the failing test (shouldn't add lh spheres to rh-only views and vice-versa), and then add tests to try to cover these new lines.

The sticking point will be figuring out if we can re-use actors and meshes in different renderers. @GuillaumeFavelier any ideas? Do we need to use vtkAssembly or something? I'm tempted just to try adding actors to multiple renderers/views to see if it works.

GuillaumeFavelier:
> The sticking point will be figuring out if we can re-use actors and meshes in different renderers

In my experiments, just adding an existing `vtkActor` to a different `vtkRenderer` of the same `vtkRenderWindow` works but from my research in https://github.com/mne-tools/mne-python/pull/8018, this could lead to unexpected behaviours (maybe inconsistencies while removing or modifying?) and the recommended plan of action is to let each renderer manage its own list of actors and mappers.

I would say that the safe route is to share the mesh itself.

AFAIK, `vtkAssembly` would help manage the properties of those actors but not solve the duplication of the underlying data.

larsoner:
Okay tests added, ready to go from my end @GuillaumeFavelier . Want to take a pass at the code, and once you're happy we can see if @agramfort wants to try and/or look ?

GuillaumeFavelier:
Yes, let's do this :+1:

larsoner:
I'm seeing segfaults on a lot of runs but not locally. Can you replicate?

larsoner:
Segfaults replicated -- it was from accidentally `_TimeViewer(brain)`ing twice. Added a check to ensure it won't happen again, hopefully that fixes the segfaults along the way...

agramfort:
I would like to test myself before we merge. Can you tell me when you have won against CIs?

larsoner:
`Ultraslow` run failure was spurious, restarted. Should be ready for you @agramfort

larsoner:
(codecov just complains that the patch diff is only 93% and the target is 95, which I think is okay. But I might push another commit just to see if I can touch a few more of the lines)

larsoner:
New thumbnails

https://21517-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/index.html#source-modeling



https://21517-1301584-gh.circle-artifacts.com/0/dev/auto_examples/index.html#inverse-problem-and-source-analysis

larsoner:
@GuillaumeFavelier already gave +1 for merge so I'll go ahead and merge

PR #8066, opened 2020-07-29
MRG: Manually calculate channel location for NIRS data

rob-luke:
#### Reference issue
Follows https://github.com/mne-tools/mne-nirs/pull/146/files/9e103b071bafc988f93cab543f68571ed8ae5cf2#diff-97adcb285086e704b219a0e18ede4576. I noticed I was unable to exactly recreate the channel locations as we had this hack in MNE to deal with bad channel locations stored in nirx files.

Previously "channel" locations were read from the nirx files, and these are the midpoints between light source and detector. But these weren't coded correctly for detectors that were very close to sources, so we just took the source location in that instance. Now I calculate the channel location as the midpoint between source and detector. The result is basically the same, except the result is now consistent even for short channels.

#### Additional information
Channels aren't defined for all vendors. So it makes sense to calculate it ourselves in a consistent manner and do this across all file types.

Just some extra info on notation.  
Source: Physical device that emits light.  
Detector: Physical device that measures light.  
Channel: Common construct used in papers to describe where we are measuring, usually the mid point between source and detector.

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -249,15 +249,10 @@ class RawNIRX(BaseRaw):
             det = int(requested_channels[ch_idx2, 1]) - 1
             info['chs'][ch_idx2 * 2]['loc'][6:9] = det_locs[det, :]
             info['chs'][ch_idx2 * 2 + 1]['loc'][6:9] = det_locs[det, :]
-            # Store channel location
-            # Channel locations for short channels are bodged,
-            # for short channels use the source location.
-            if det + 1 in short_det:
-                info['chs'][ch_idx2 * 2]['loc'][:3] = src_locs[src, :]
-                info['chs'][ch_idx2 * 2 + 1]['loc'][:3] = src_locs[src, :]
-            else:
-                info['chs'][ch_idx2 * 2]['loc'][:3] = ch_locs[ch_idx2, :]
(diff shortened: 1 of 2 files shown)


rob-luke:
Trying to trigger a full circle build so I can post the before and after figure here.

rob-luke:
Before PR: https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#view-location-of-sensors-over-brain-surface. 
After PR: https://21481-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#view-location-of-sensors-over-brain-surface

Before PR: channels (orange) are missing for short channels (the small lines)




After PR: most channels (orange) are the same as above, but now we also have channels for the short lines.

rob-luke:
@larsoner @agramfort can you please review

FYI: @Swy7ch @kylemath

larsoner:
> Trying to trigger a full circle build so I can post the before and after figure here.

For future reference, CircleCI builds all examples with `[circle full]`, front-page examples with `[circle front]`, and (relevant here) by default *any changed examples*. So if you want a few specific examples to build, you just modify them in some way, ideally by making some tiny wording improvement but if the example is already perfect then you can even change whitespace in some PEP8-compliant way.

larsoner:
Thanks @rob-luke !

PR #8068, opened 2020-07-29
MRG, MAINT: Unify 90% similar code

larsoner:
`plot_source_estimates` and `plot_vector_source_estimates` share 90%+ of their code. This unifies the two approaches into a single `_plot_stc` function.

This will make things like #8064 much easier, as instead of making deep changes in two places, you can make them in one.


agramfort:
@GuillaumeFavelier i let you merge if happy

GuillaumeFavelier:
Thank you @larsoner !

Issue #8067, opened 2020-07-29
ICA.find_bads_ecg() accepts "method", find_bads_eog() accepts "measure"


hoechenberger:
`ICA.find_bads_ecg()` has a kwarg `method : {'ctps', 'correlation'}`; in contrast, `find_bads_eog()` uses `measure : 'zscore' | 'cor'`

- The kwarg for selection of the method to use is once called `method`, and once `measure`
- Correlation is once selected by passing `'correlation'`, and once via `'cor'`
- The specification of valid values is once done via `key : {val1, val2}`, and once via `key : val1 | val2`

How can we make things more consistent?

agramfort:
I would make it so that find_bads_eog matches the API of find_bads_ecg. So rename measure to method and require correlation and not cor. But I would write it as val1 | val2

hoechenberger:
Turns out `find_bads_ecg` has **both**, a `measure` and a `method` kwarg which control different thing. Will therefore only slightly adjust the docstrings.

PR #8072, opened 2020-07-30
Make ICA.find_bads_* kwarg values more consistent

hoechenberger:
#### Reference issue
Closes GH-8067.

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -1067,7 +1067,7 @@ class ICA(ContainsMixin):
                       measure="zscore"):
         """Compute ExG/ref components.
 
-        See find_bads_ecg, find_bads, eog, and find_bads_ref for details.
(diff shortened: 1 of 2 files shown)


mne/preprocessing/ica.py
```diff
@@ -1200,7 +1200,7 @@ def find_bads_ecg(self, inst, ch_name=None, threshold=None, start=None,
             If True, data annotated as bad will be omitted. Defaults to True.
 
             .. versionadded:: 0.14.0
-        measure : {'zscore', "cor"}
+        measure : 'zscore' | 'cor'
```
drammock, line 1203: Not going to change this to "correlation"? I thought that was part of the plan?

agramfort: +1 to change to correlation with deprecation warning

hoechenberger: Done, without deprecation warning – has been introduced after 0.20

PR #8079, opened 2020-08-03
MRG: Add reading of custom MNE info to SNIRF

rob-luke:
#### Reference issue
I am adding a SNIRF writer at https://github.com/mne-tools/mne-nirs/pull/151

The SNIRF standard does not include some info that MNE has such as sex, middle name, last name (SNIRF has only SubjectID). 

To allow writing an MNE NIRS recording and reading it back without any loss, I have added non standard (but allowed) custom metadata tags to the SNIRF writer. In this PR I add reading of these extra values.

#### Additional information

- [x] Add reading of sex, middle name last name
- [x] Add reading of dig points 

I also noticed that unknown sex was not being correctly encoded in the NIRX reader.

### DIFF
--- a/mne/io/snirf/_snirf.py
+++ b/mne/io/snirf/_snirf.py
@@ -11,6 +11,7 @@ from ...annotations import Annotations
 from ...utils import logger, verbose, fill_doc, warn
 from ...utils.check import _require_version
 from ..constants import FIFF
+from .._digitization import _make_dig_points, _format_dig_points
 
 
 @fill_doc
@@ -156,13 +157,13 @@ class RawSNIRF(BaseRaw):
                 ln = dat.get('/nirs/metaDataTags/lastName')[0].decode('UTF-8')
                 subject_info['last_name'] = ln
             if 'middleName' in dat.get('nirs/metaDataTags/'):
-                ln = dat.get('/nirs/metaDataTags/middleName')[0].decode('UTF-8')
-                subject_info['middle_name'] = ln
+                m = dat.get('/nirs/metaDataTags/middleName')[0].decode('UTF-8')
+                subject_info['middle_name'] = m
             if 'sex' in dat.get('nirs/metaDataTags/'):
-                sex = dat.get('/nirs/metaDataTags/LengthUnit')[0].decode('UTF-8')
-                if sex in {'M', 'Male', '1', 1, 'm'}:
+                s = dat.get('/nirs/metaDataTags/LengthUnit')[0].decode('UTF-8')
+                if s in {'M', 'Male', '1', 1, 'm'}:
                     subject_info['sex'] = FIFF.FIFFV_SUBJ_SEX_MALE
-                elif sex in {'F', 'Female', '2', 2, 'f'}:
+                elif s in {'F', 'Female', '2', 2, 'f'}:
                     subject_info['sex'] = FIFF.FIFFV_SUBJ_SEX_FEMALE
             # End non standard name reading
             # Update info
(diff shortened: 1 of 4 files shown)


rob-luke:
@agramfort and @larsoner could you please review. I would appreciate feedback if you think the method of saving `raw.info['dig']` is appropriate. Could you please review in parallel with https://github.com/mne-tools/mne-nirs/pull/151

mne/io/snirf/_snirf.py
```diff
@@ -175,6 +194,30 @@ def natural_keys(text):
...
+                        extra_pos.append(dict(
+                            kind=FIFF.FIFFV_POINT_EEG,  # as in read_raw_nirx
+                            r=diglocs[idx, :],
+                            ident=len(extra_pos) + 1,
+                            coord_frame=FIFF.FIFFV_COORD_HEAD,
+                        ))
+                info['dig'] = _make_dig_points(nasion=nasion, lpa=lpa,
+                                               rpa=rpa, hpi=hpi)
+                for ex in _format_dig_points(extra_pos):
+                    info['dig'].append(ex)
```
larsoner, line 219:
```suggestion
                info['dig'].extend(_format_dig_points(extra_pos))
```

rob-luke:
Thank you both. I think this is ready to merge now.

rob-luke:
Thanks @larsoner 

> Just noticed the coord_frame -- are they actually in the head frame? ...  If not, UNKNOWN is a safer choic

Good catch. I have now made the default UNKNOWN and also added a [custom field to the SNIRF file](https://github.com/mne-tools/mne-nirs/pull/151/commits/cc66cc241e1e58c69da56247fa5e6c3da218f375) that encodes the coord frame, this is used if available.

> Also, I realized that you can probably simplify

Wow that is nicer. Thanks.

Ready for review again :)

rob-luke:
A final thought. Do you think we should preface each of the custom metadata items with MNE_, so instead of `nirs/metaDataTags/coordFrame` I would use `nirs/metaDataTags/MNE_coordFrame`?

larsoner:
>  Ok I have reverted this as _make_dig_points required coord_frame to be a string with the coord frame name, and I was encoding the integer value. I didnt want to put if coord == 4; c="head" elif coord==0; c="unknown" elif... as this seemed like more to maintain. Thoughts?

Use `mne.transforms._frame_to_str`

https://github.com/mne-tools/mne-python/blob/3545c3f26d4c1cffbe4b3b925eb059fb01663967/mne/transforms.py#L42

> added a custom field to the SNIRF file that encodes the coord frame, this is used if available.

I think `MNE_coordFrame` is pretty explicit. `FIFF_coordFrame` would be another choice since we are really relying on their constants. But I like `MNE_coordFrame` because I think it'll be easier for other people to figure out what it means.

rob-luke:
> Use mne.transforms._frame_to_str

Thank you 💯 

> But I like MNE_coordFrame because I think it'll be easier for other people to figure out what it means

Ok I will make this change now.

rob-luke:
Ok this is passing locally. Lets hope the CIs agree. Thanks for the help

larsoner:
Any way to avoid the coverage decrease? Update some testing files + tests?

rob-luke:
The tests over at https://github.com/mne-tools/mne-nirs/pull/151 hit all the lines.

I could export a file using our writer and commit that to testing data. But that feels a bit inelegant to me given the write then read process hits everything, and it's committing a whole file for just a few lines. We could also move the reader to MNE-nirs to keep them together. Or could we do a 'from mne_nirs import write_raw_snirf'?

larsoner:
One option would be to add `mne-nirs` to `requirements_testing.txt`, then decorate a test with `requires_version('mne_nirs')`, and write the data that has the relevant information. It becomes a little bit redundant with `mne-nirs` tests probably, in which case maybe do the bare minimum here to not take a coverage hit / smoke test style, and have more comprehensive tests there?

rob-luke:
> It becomes a little bit redundant with mne-nirs tests probably

Yes this is what I feel.

I suggest that in MNE we test that we can read data as produced by the official snirf converter, and from vendors (already merged). And that in MNE-NIRS we make sure we test the MNE specific additions (that's what's in this pr: sex, coord frame, middle name).

But let me know if I misunderstood your comment above. I think importing MNE-nirs is my least preferred option, at that point we could just move the code to MNE.

rob-luke:
Im going to try and manually hit a few more lines in the test. Please hold the line

mne/io/snirf/tests/test_snirf.py
```diff
@@ -89,6 +90,44 @@ def test_snirf_against_nirx():
     assert_allclose(raw.get_data(), raw_orig.get_data())
 
 
+@requires_h5py
+@requires_testing_data
+def test_snirf_nonstandard(tmpdir):
```
rob-luke, line 95: Ok I have tried to manually hit some additional lines to increase the coverage. But I can't think of a nice way to hit the dig points without lots of manual code.

larsoner: Hah! I considered suggesting that you mock/create the fields by copying and modifying, but didn't want to make a bunch of extra work for you. But you did it anyway. :)

This seems fine to me, especially if there are complementary tests in `mne-nirs` for round-trip-ness.

Perhaps at some point we can convince @agramfort to include the writer in MNE-Python by doing a comparison of the number of lines required to mock the fields vs lines required to just include the writer :)

rob-luke: Great. Yeah I should have just done this from the start, was being lazy.

> Perhaps at some point we can convince agramfort to include the writer in MNE-Python

I'm ok with maintaining things in MNE-NIRS, particularly since you both kindly review things there on request too. In my mind I am starting to have a clear line for what's appropriate where. Roughly epoching style analysis goes in MNE, and the NIRS specific and MRI style analysis goes in MNE-NIRS. Its no extra effort for me to switch between the two repos and its had the benefit of me learning more about CIs and package management.

larsoner:
Thanks @rob-luke !

PR #8082, opened 2020-08-03
MRG, ENH: Add flatmap

### LINKED ISSUE #8081: ENH: Flatmaps for cortex
See the WIP FreeSurfer PR where there is an implementation we can work with:

https://github.com/nipy/PySurfer/pull/269

The tricky part is that the vertices of the flatmap do not directly map onto the vertices of the cortex. I'm not sure why it is this way. I wonder for us if a better implementation would actually be to resample the flatmaps themselves (in the spherical space maybe?) to make them have the same effective vertices or something.

cc @johnsam7 this is probably a prerequisite for being able to do it for cerebellum

I'll probably take a look to see how difficult this would be

larsoner:
`plot_visualize_stc.py` on my system with this:

Screenshot from 2020-08-03 11-41-55

- [x] Reading function, plotting works
- [x] Interaction works
- [x] Wait for #8064 to be merged and rebase
- [x] Add checks that isn't used with vector or volume
- [x] Add some simple tests for `brain`
- [x] Add doc about how to get flat maps
- [x] Fix window-sizing bug

Closes #8081


larsoner:
cc @johnsam7 this should help prepare us to visualize cortex + cerebellum flatmaps

agramfort:
we should document or point to external doc on how to compute the flatmaps. Last time I tried It required manual work.

larsoner:
I have no idea how actually. I just know fsaverage has them. For now maybe it's enough just to tell people to morph?

agramfort:
anyone has experience with this here?

larsoner:
There are tutorials here:

- https://surfer.nmr.mgh.harvard.edu/fswiki/FreeSurferOccipitalFlattenedPatch
- https://openwetware.org/wiki/Beauchamp:FreeSurfer

I'll add a note that it's available for fsaverage and otherwise you need to do something like the above while adding tests soon

agramfort:
ok

>

larsoner:

larsoner:
@GuillaumeFavelier do you want to take a look?

larsoner:
CircleCI failure is just the one fixed by https://github.com/mne-tools/mne-python/pull/8109

mne/viz/_brain/_brain.py
```diff
@@ -409,6 +418,9 @@ def add_data(self, array, fmin=None, fmid=None, fmax=None,
         clim : dict
             Original clim arguments.
         %(src_volume_options_layout)s
+        colorbar_kwargs : dict | None
```
GuillaumeFavelier, line 421: Do we have a plan with `colorbar_kwargs`? I don't see it used in this PR.

larsoner: Indeed I don't use it in this example, locally I use it to set the tick and title sizes. Do you think I should update an example?

GuillaumeFavelier: Yes, it would be great to have it used once at least.

larsoner: https://21662-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_visualize_stc.html

agramfort:
what is weird is that you can still rotate flat maps. Once you have
started rotating it's not clear how to reset the view.

is there a possible easy fix?

larsoner:
It would require writing a new VTK interactor, so no not really. We should probably add a keyboard command to _Brain that resets to the original views

agramfort:
>
> We should probably add a keyboard command to _Brain that resets to the
> original views
>

+1. 'r' key?

larsoner:
Sure, I would say it should reset the time index and views -- basically get everything back to how it was when the viz was opened

agramfort:
+1

larsoner:
Added it as an item in #7162

PR #8083, opened 2020-08-03
MRG, ENH: Add preprocessing regression

### LINKED ISSUE #7636: Gratton & Coles Eye Movement Correction Procedure
If your issue is a usage question, please consider asking on our [Gitter channel](https://gitter.im/mne-tools/mne-python) or on our [mailing list](https://mail.nmr.mgh.harvard.edu/mailman/listinfo/mne_analysis) instead of opening an issue.

#### Describe the problem
Implement a Gratton and Coles EMCP regression procedure for EEG analysis of epoched data

#### Describe your solution
Have a working version that takes Epoch objects as input and outputs corrected Epochs objects 

ICA, but nice feature to test against existing tools and use for certain lab groups

#### Additional context
Currently implemented in [https://github.com/kylemath/deepeeg](https://github.com/kylemath/deepeeg) as part of deep learning feature extraction process

larsoner:
Takeover of #7637. cc @kylemath @cbrnr I tried to rework the API a bit to make it more general, and to allow fitting the coefficients on one segment of data and applying them to another. Let me know what you think.

- [x] Add method
- [x] Add example
- [x] Add tests

Closes #7637
Closes #7636

### DIFF
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1758,7 +1758,6 @@
   pages = {171--173}
 }
 
-
 @article{GrattonEtAl1983,
 	title = {A new method for off-line removal of ocular artifact},
 	volume = {55},
@@ -1772,4 +1771,29 @@
 	month = apr,
 	year = {1983},
 	pages = {468--484}
+}
+
+@book{OppenheimEtAl1999,
+	address = {Upper Saddle River, N.J},
+	edition = {2 edition},
+	title = {Discrete-{Time} {Signal} {Processing}},
+	isbn = {978-0-13-754920-7},
+	language = {English},
+	publisher = {Prentice Hall},
+	author = {Oppenheim, Alan V. and Schafer, Ronald W. and Buck, John R.},
+	month = jan,
+	year = {1999}
+}
+
+
+@book{CrochiereRabiner1983,
+	address = {Englewood Cliffs, N.J},
+	edition = {1 edition},
+	title = {Multirate {Digital} {Signal} {Processing}},
+	isbn = {978-0-13-605162-6},
+	language = {English},
+	publisher = {Pearson},
+	author = {Crochiere, Ronald E. and Rabiner, Lawrence R.},
+	month = dec,
+	year = {1983}
 }
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1757,3 +1757,19 @@
   pmcid = {PMC4610715},
   pages = {171--173}
 }
+
+
+@article{GrattonEtAl1983,
+	title = {A new method for off-line removal of ocular artifact},
+	volume = {55},
+	issn = {0013-4694},
+	doi = {10.1016/0013-4694(83)90135-9},
+	language = {en},
+	number = {4},
+	urldate = {2020-08-03},
+	journal = {Electroencephalography and Clinical Neurophysiology},
+	author = {Gratton, Gabriele and Coles, Michael G. H and Donchin, Emanuel},
+	month = apr,
+	year = {1983},
+	pages = {468--484}
+}
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1774,7 +1774,7 @@
 }
 
 @book{OppenheimEtAl1999,
-	address = {Upper Saddle River, N.J},
+	address = {Upper Saddle River, NJ},
 	edition = {2 edition},
 	title = {Discrete-{Time} {Signal} {Processing}},
 	isbn = {978-0-13-754920-7},
@@ -1787,7 +1787,7 @@
 
 
 @book{CrochiereRabiner1983,
-	address = {Englewood Cliffs, N.J},
+	address = {Englewood Cliffs, NJ},
 	edition = {1 edition},
 	title = {Multirate {Digital} {Signal} {Processing}},
 	isbn = {978-0-13-605162-6},
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -21,8 +21,8 @@ def regress_artifact(inst, picks=None, picks_artifact='eog', betas=None,
         The instance to process.
     %(picks_good_data)s
     picks_artifact : array-like | str
-        Channel picks to use as predictor / explanatory variables capturing
-        the artifact ofinterest (default is "eog").
+        Channel picks to use as predictor/explanatory variables capturing
+        the artifact of interest (default is "eog").
     betas : ndarray, shape (n_picks, n_picks_ref) | None
         The regression coefficients to use. If None (default), they will be
         estimated from the data.
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -7,7 +7,7 @@ import numpy as np
 from ..epochs import BaseEpochs
 from ..io.pick import _picks_to_idx
 from ..io.base import BaseRaw
-from ..utils import _check_preload, _validate_type, verbose
+from ..utils import _check_preload, _validate_type, _check_option, verbose
 
 
 @verbose
@@ -71,7 +71,8 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
         estimate = True
     else:
         estimate = False
-        assert betas.shape == betas_shape
+        betas = np.asarray(betas, dtype=float)
+        _check_option('betas.shape', betas.shape, (betas_shape,))
     for pi, pick in enumerate(picks):
         this_data = inst._data[..., pick, :]  # view
         orig_shape = this_data.shape
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -44,14 +44,14 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
     original data in two calls like (here for a single-condition ``epochs``
     only):
 
-        >>> epochs_no_ave = epochs.copy().subtract_evoked()
-        >>> _, betas = mne.preprocessing.regress(epochs_no_ave)
-        >>> epochs_clean, _ = mne.preprocessing.regress(epochs, betas=betas)
+        >>> epochs_no_ave = epochs.copy().subtract_evoked()  # doctest:+SKIP
+        >>> _, betas = mne.preprocessing.regress(epochs_no_ave)  # doctest:+SKIP
+        >>> epochs_clean, _ = mne.preprocessing.regress(epochs, betas=betas)  # doctest:+SKIP
 
     References
     ----------
     .. footbibliography::
-    """
+    """  # noqa: E501
     _check_preload(inst, 'regress')
     _validate_type(inst, (BaseEpochs, BaseRaw), 'inst', 'Epochs or Raw')
     picks = _picks_to_idx(inst.info, picks, none='data')
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -23,7 +23,7 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
     picks_ref : array-like
         Picks to use as the reference channels.
     betas : ndarray, shape (n_picks, n_picks_ref) | None
-        The regression coeffients to use. If None, they will be estimated
+        The regression coefficients to use. If None, they will be estimated
         from the data.
     copy : bool
         If True (default), copy the instance before modifying it.
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -11,8 +11,8 @@ from ..utils import _check_preload, _validate_type, _check_option, verbose
 
 
 @verbose
-def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
-            verbose=None):
+def regress_artifact(inst, picks=None, picks_artifact='eog', betas=None,
+                     copy=True, verbose=None):
     """Regress artifacts using reference channels.
 
     Parameters
@@ -20,11 +20,12 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
     inst : instance of Epochs | Raw
         The instance to process.
     %(picks_good_data)s
-    picks_ref : array-like
-        Picks to use as the reference channels.
+    picks_artifact : array-like | str
+        Channel picks to use as predictor / explanatory variables capturing
+        the artifact ofinterest (default is "eog").
     betas : ndarray, shape (n_picks, n_picks_ref) | None
-        The regression coefficients to use. If None, they will be estimated
-        from the data.
+        The regression coefficients to use. If None (default), they will be
+        estimated from the data.
     copy : bool
         If True (default), copy the instance before modifying it.
     %(verbose)s
@@ -55,17 +56,17 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
     _check_preload(inst, 'regress')
     _validate_type(inst, (BaseEpochs, BaseRaw), 'inst', 'Epochs or Raw')
     picks = _picks_to_idx(inst.info, picks, none='data')
-    picks_ref = _picks_to_idx(inst.info, picks_ref, allow_empty=False)
-    if np.in1d(picks_ref, picks).any():
-        raise ValueError('ref_picks cannot be contained in picks')
+    picks_artifact = _picks_to_idx(inst.info, picks_artifact)
+    if np.in1d(picks_artifact, picks).any():
+        raise ValueError('picks_artifact cannot be contained in picks')
     inst = inst.copy() if copy else inst
-    ref_data = inst._data[..., picks_ref, :]
-    ref_data = ref_data - np.mean(ref_data, -1, keepdims=True)
+    artifact_data = inst._data[..., picks_artifact, :]
+    ref_data = artifact_data - np.mean(artifact_data, -1, keepdims=True)
     if ref_data.ndim == 3:
-        ref_data = ref_data.transpose(1, 0, 2).reshape(len(picks_ref), -1)
+        ref_data = ref_data.transpose(1, 0, 2).reshape(len(picks_artifact), -1)
     cov = np.dot(ref_data, ref_data.T)
     # process each one separately to reduce memory load
-    betas_shape = (len(picks), len(picks_ref))
+    betas_shape = (len(picks), len(picks_artifact))
     if betas is None:
         betas = np.empty(betas_shape)
         estimate = True
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -1,0 +1,85 @@
+# Authors: Eric Larson <larson.eric.d@gmail.com>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+
+from ..epochs import BaseEpochs
+from ..io.pick import _picks_to_idx
+from ..io.base import BaseRaw
+from ..utils import _check_preload, _validate_type, verbose
+
+
+@verbose
+def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
+            verbose=None):
+    """Regress artifacts using reference channels.
+
+    Parameters
+    ----------
+    inst : instance of Epochs | Raw
+        The instance to process.
+    %(picks_good_data)s
+    picks_ref : array-like
+        Picks to use as the reference channels.
+    betas : ndarray, shape (n_picks, n_picks_ref) | None
+        The regression coeffients to use. If None, they will be estimated
+        from the data.
+    copy : bool
+        If True (default), copy the instance before modifying it.
+    %(verbose)s
+
+    Returns
+    -------
+    inst : instance of Epochs
+        The processed data.
+    betas : ndarray, shape (n_picks, n_picks_ref)
+        The betas used during regression.
+
+    Notes
+    -----
+    To implement the method outlined in :footcite:`GrattonEtAl1983`,
+    remove the evoked response from epochs before estimating the
+    regression coefficients, then apply those regression coefficients to the
+    original data in two calls like (here for a single-condition ``epochs``
+    only):
+
+        >>> epochs_no_ave = epochs.copy().subtract_evoked()
+        >>> _, betas = mne.preprocessing.regress(epochs_no_ave)
+        >>> epochs_clean, _ = mne.preprocessing.regress(epochs, betas=betas)
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    _check_preload(inst, 'regress')
+    _validate_type(inst, (BaseEpochs, BaseRaw), 'inst', 'Epochs or Raw')
+    picks = _picks_to_idx(inst.info, picks, none='data')
+    picks_ref = _picks_to_idx(inst.info, picks_ref, allow_empty=False)
+    if np.in1d(picks_ref, picks).any():
+        raise ValueError('ref_picks cannot be contained in picks')
+    inst = inst.copy() if copy else inst
+    ref_data = inst._data[..., picks_ref, :]
+    ref_data = ref_data - np.mean(ref_data, -1, keepdims=True)
+    if ref_data.ndim == 3:
+        ref_data = ref_data.transpose(1, 0, 2).reshape(len(picks_ref), -1)
+    cov = np.dot(ref_data, ref_data.T)
+    # process each one separately to reduce memory load
+    betas_shape = (len(picks), len(picks_ref))
+    if betas is None:
+        betas = np.empty(betas_shape)
+        estimate = True
+    else:
+        estimate = False
+        assert betas.shape == betas_shape
+    for pi, pick in enumerate(picks):
+        this_data = inst._data[..., pick, :]  # view
+        orig_shape = this_data.shape
+        if estimate:
+            # subtract mean over time from every trial/channel
+            cov_data = this_data - np.mean(this_data, -1, keepdims=True)
+            cov_data = cov_data.reshape(1, -1)
+            betas[pi] = np.linalg.solve(cov, np.dot(ref_data, cov_data.T)).T[0]
+        # subtract weighted (demeaned) eye channels from channel
+        this_data -= (betas[pi] @ ref_data).reshape(orig_shape)
+    return inst, betas
--- a/mne/preprocessing/_regress.py
+++ b/mne/preprocessing/_regress.py
@@ -31,7 +31,7 @@ def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
 
     Returns
     -------
-    inst : instance of Epochs
+    inst : instance of Epochs | Raw
         The processed data.
     betas : ndarray, shape (n_picks, n_picks_ref)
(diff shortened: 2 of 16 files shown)


tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,121 @@
...
+
+eog_epochs = mne.preprocessing.create_eog_epochs(raw)
+eog_epochs.apply_baseline((None, None))
+raw.plot(events=eog_epochs.events)
+fig = eog_epochs.average(picks=plot_picks).plot(picks=plot_picks)
+fig.suptitle('EOG epochs')
+fig.tight_layout()
+
+eog_epochs_clean, _ = mne.preprocessing.regress(eog_epochs)
+fig = eog_epochs_clean.average(picks=plot_picks).plot(picks=plot_picks)
```
kylemath, line 91: good practice to repeat the raw data plotting here as well so the user can scroll through and compare the blinks to the VEOG and HEOG channels , it may be nice to reduce the channel numbers for these raw plots to make it easier to find the VEOG and HEOG and compare them to MEG channels

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,121 @@
...
+fig.suptitle('EOG epochs, EOG regressed')
+fig.tight_layout()
+
```
kylemath, line 94: This looks great, and looks like it is working well

mne/preprocessing/_regress.py
```diff
@@ -0,0 +1,86 @@
...
+@verbose
+def regress(inst, picks=None, picks_ref='eog', betas=None, copy=True,
+            verbose=None):
+    """Regress artifacts using reference channels.
+
+    Parameters
+    ----------
+    inst : instance of Epochs | Raw
+        The instance to process.
+    %(picks_good_data)s
```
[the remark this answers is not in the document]
larsoner: This is some docstring magic for things we use a lot, see how it renders

https://21544-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.regress.html#mne.preprocessing.regress

mne/preprocessing/_regress.py
```diff
@@ -0,0 +1,86 @@
...
+    picks_ref : array-like
+        Picks to use as the reference channels.
```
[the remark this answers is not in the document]
larsoner: ahh right. `picks_artifact` seems fine to me, @cbrnr any ideas for a better name?

mne/preprocessing/_regress.py
```diff
@@ -0,0 +1,86 @@
...
+    -------
+    inst : instance of Epochs
+        The processed data.
+    betas : ndarray, shape (n_picks, n_picks_ref)
+        The betas used during regression.
+
+    Notes
+    -----
+    To implement the method outlined in :footcite:`GrattonEtAl1983`,
+    remove the evoked response from epochs before estimating the
```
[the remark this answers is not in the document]
larsoner: in MNE these are synonyms / it would be redundant

https://mne.tools/dev/glossary.html#term-evoked

tutorials/preprocessing/plot_35_artifact_correction_regression.py
larsoner: Good call -- better?

https://21558-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_35_artifact_correction_regression.html#eog

mne/preprocessing/_regress.py
```diff
@@ -11,20 +11,21 @@
...
 
     Parameters
     ----------
     inst : instance of Epochs | Raw
         The instance to process.
     %(picks_good_data)s
-    picks_ref : array-like
-        Picks to use as the reference channels.
+    picks_artifact : array-like | str
+        Channel picks to use as predictor / explanatory variables capturing
```
cbrnr, line 24: I'd remove the spaces in `predictor/explanatory`.

mne/preprocessing/_regress.py
```diff
@@ -11,20 +11,21 @@
...
+        the artifact ofinterest (default is "eog").
```
cbrnr, line 25: space missing: `of interest`

cbrnr: I can review the PR next week, so if you can wait I'd be happy to try it out.

doc/references.bib
```diff
@@ -1757,3 +1757,43 @@ @article{KayserTenke2015
...
+	language = {English},
+	publisher = {Prentice Hall},
+	author = {Oppenheim, Alan V. and Schafer, Ronald W. and Buck, John R.},
+	month = jan,
+	year = {1999}
+}
+
+
+@book{CrochiereRabiner1983,
+	address = {Englewood Cliffs, N.J},
```
drammock, line 1790:
```suggestion
	address = {Englewood Cliffs, NJ},
```

drammock: (Or `N.J.`)

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+   :local:
+   :depth: 2
+
+We begin as always by importing the necessary Python modules and loading some
+data. In this case we use data from :ref:`brainstorm <tut-brainstorm-auditory>`
+because it has both EOG and ECG electrodes. We then crop it to 60 seconds,
+set some channel information, then process the data.
+
+That there are other corrections that are useful for this dataset that we will
+not apply here -- see :ref:`tut-brainstorm-auditory` for more information.
```
drammock, line 21:
```suggestion
Note that there are other corrections that are useful for this dataset that we
will not apply here -- see :ref:`tut-brainstorm-auditory` for more information.
```

drammock: Not sure if this is the right repair for what you were trying to say.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+#
+# Artifacts whose time waveforms are accurately reflected by some reference
+# signal can be removed by regression. Two examples of this are EOG artifacts
+# captured by EOG channels, and ECG signals captured by ECG electrodes.
+#
+#
+# EOG
+# ~~~
+#
+# First let's epoch our data the normal way, using just event type 1 (standard)
```
drammock, line 50:
```suggestion
# First let's epoch our data the normal way, using just event type 1 (only
# the "standard" trials)
```

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+fig = epochs_clean.average(picks=plot_picks).plot(**evo_kwargs)
+fig.suptitle('Auditory epochs, EOG regressed')
+fig.tight_layout()
+del epochs, epochs_clean
+
+###############################################################################
+# The effect is subtle in these evoked data. You can see that a bump toward
+# the end of the window has had its amplitude decreased.
+#
+# The effect is even clearer if we create epochs around our (autodetected) EOG
```
drammock, line 88:
```suggestion
# The effect is clearer if we create epochs around our (autodetected) EOG
```

drammock: I have a hard time agreeing that the effect is clear in the current plots. Can we zoom in on the second, larger blink (around 21 seconds) and show fewer MEG channels (so that the blink perturbation looks bigger)?  I'd also consider showing the 2 raw plots side-by-side.  I'm neutral on whether showing the two EOG epochs plots adds enough to be worth it; understanding the improvement relies on noticing the different y-axis limits, so I would either fix the limits across the two `mag` panels, or call attention to the changed y-axis limits in the text.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+                                                         betas=betas)
+fig = eog_epochs_clean.average(picks=plot_picks).plot(**evo_kwargs)
+fig.suptitle('EOG epochs, EOG regressed')
+fig.tight_layout()
+del eog_epochs, eog_epochs_clean, raw_clean  # save memory
+
+###############################################################################
+# ECG
+# ~~~
+# We can follow the same process with ECG:
```
drammock, line 117:
```suggestion
# We can follow the same process with ECG, although it turns out not to work as
# well. This is likely because the ECG signal is like a rotating dipole, and
# therefore the ECG electrode's time waveform does not reflect the same
# temporal dynamics that manifest at each MEG channel. Other approaches like
# :ref:`ICA <tut-artifact-ica>` or :ref:`SSP <tut-artifact-ssp>` may work
# better with this dataset.
```

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+raw.plot(events=ecg_epochs.events, **raw_kwargs)
+fig = ecg_epochs.average(picks=plot_picks).plot(**evo_kwargs)
+fig.suptitle('ECG epochs')
+fig.tight_layout()
+
+###############################################################################
+# But it does not remove as much of the signal, even from the reference
+# epochs, likely because the ECG signal is a rotating dipole, and the ECG
+# electrode time waveform does not capture the same effective time waveform
+# that each channel does:
```
drammock, line 130: (delete; incorporated above)

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,142 @@
...
+
+# Here we operate in place to save memory
+mne.preprocessing.regress_artifact(
+    ecg_epochs, picks_artifact='ecg', copy=False)
+fig = ecg_epochs.average(picks=plot_picks).plot(**evo_kwargs)
+fig.suptitle('ECG epochs, ECG regressed')
+fig.tight_layout()
```
drammock, line 137: same feedback as for EOG section: put raw plots side by side, and (if included) put ECG epochs side by side and call attention to y-axis limits.

drammock:
OK @larsoner, as requested I've had a go at the tutorial. once CIs are happy I'm +1 to merge this.

larsoner:
Thanks. Let's wait for @cbrnr to give it a try

drammock:
tutorial rendering: https://21604-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_35_artifact_correction_regression.html

kylemath:
Here is an idea, on mobile so can't edit myself, but we may want to explicate the application the raw data 

Normally the Gratton eye correction subtracts the evoked activity before computing the betas, then removes blinks from original data like on tutorial

For raw data this wouldn't be possible, and we should say that in the prose of the tutorial to make it clear why we don't subtract evoked , but warn that this can regress out brain activity as a result

larsoner:
> For raw data this wouldn't be possible, and we should say that in the prose of the tutorial to make it clear why we don't subtract evoked , but warn that this can regress out brain activity as a result

Actually the tutorial gets around this problem by using the betas computed from the epochs with the evoked subtracted to do the regression on the raw. So I think we're okay here. But maybe it needs to be clearer that we're doing this to avoid capturing brain activity of interest in the beta estimation (which would be a risk if we just processed the raw directly)?

drammock:
@kylemath I pushed a commit (2bbb906) to implement your suggestion. Can you see if it captures what you were hoping for?

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+
+Repairing artifacts with regression
+===================================
+
+This tutorial covers removal of artifacts using regression as in
+:footcite:`GrattonEtAl1983`.
+
+.. contents:: Page contents
+   :local:
+   :depth: 2
```
cbrnr, line 13: I would remove the TOC because this document has basically just two sections. If you want to keep it, I'd create a new section for the description of the preparatory steps (at the same level as `# Removing artifacts by regression`) to improve the structure.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+
+We begin as always by importing the necessary Python modules and loading some
+data. In this case we use data from :ref:`brainstorm <tut-brainstorm-auditory>`
```
cbrnr, line 16: Brainstorm (capital B).

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+because it has both EOG and ECG electrodes. We then crop it to 60 seconds,
+set some channel information, then process the data.
+
+Note that there are other corrections that are useful for this dataset that we
+will not apply here — see :ref:`tut-brainstorm-auditory` for more information.
```
cbrnr, line 21: I'd prefer putting the `see :ref:`tut-brainstorm-auditory` for more information` part in parens instead of after a dash.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+               eog=True, ecg=True, ref_meg=False).load_data()
+raw.info['bads'] = ['MLO52-4408', 'MRT51-4408', 'MLO42-4408', 'MLO43-4408']
+raw.filter(None, 40)
+decim = 12  # 2400 -> 200 Hz sample rate for epochs
+
+###############################################################################
+# Removing artifacts by regression
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Generally speaking, artifacts whose time waveforms on the sensors are
```
cbrnr, line 42: To me using "whose" implies that we speak about a person. Of course I'm not a native speaker, but would it be better to change this to "artifact waveforms that are accurately reflected by some reference signal"?

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+# accurately reflected by some reference signal can be removed by regression.
+# Two examples of this are blink artifacts (captured by EOG channels), and
```
cbrnr, line 44: Remove the comma before "and".

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+
+# get ready to plot
+plot_picks = ['meg', 'eog', 'ecg']
+evo_kwargs = dict(picks=plot_picks, spatial_colors=True,
+                  verbose='error')  # ignore warnings about spatial colors
+
+# plot original data (averaged across epochs)
+fig = epochs.average(picks=plot_picks).plot(**evo_kwargs)
+fig.suptitle('Auditory epochs')
+fig.tight_layout()
```
cbrnr, line 83: Here I get a warning:
```
UserWarning: This figure includes Axes that are not compatible with tight_layout, so results might be incorrect.
```

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -0,0 +1,173 @@
...
+# clean up
+del epochs, epochs_clean
+
+###############################################################################
+# The effect is subtle in these evoked data, but you can see that a bump toward
+# the end of the window has had its amplitude decreased.
+#
+# The effect is clearer if we create epochs around (autodetected) blink events,
+# and plot the average blink artifact before and after the regression (note
+# that the y-axis limits change between the left and right plots):
```
cbrnr, line 99: I think it would be more impressive if the y-axis limit didn't change. Can we set it to the original limit so that it is very obvious that the eye blinks are gone (we're not really interested in the MEG signals so why zoom in).

tutorials/preprocessing/plot_35_artifact_correction_regression.py
cbrnr: Overall, this example doesn't impress me much - as stated in the introduction, this approach does not work well with ECG artifacts. I've never seen anyone use the regression approach for ECG artifacts, so why do we show it here? I would remove this example completely.

larsoner:
> The first example relies on the assumption that eye blinks are somehow phase-locked to the auditory stimulus.

It does not. The method does not require the artifacts to be time locked, just to be present somewhere in the epoch windows. And actually it will work better if it's not time locked, otherwise too much of the artifacts will be removed by subtracting the evoked before estimating the betas.

cbrnr:
OK, I somehow thought that the weights are calculated on the evoked response instead of the concatenated epochs.

larsoner:
And actually if there aren't any EOG artifacts in your epochs of interest, then it probably still doesn't hurt to much to use the method because the betas will be very close to zero, but you also probably don't need to bother using it because EOG is not problematic in that case anyway.

So I think we should stick to the first example, but maybe explain this but better

I'm fine with removing the ECG and just telling people it doesn't work

As far as the bipolar stuff goes, I'md rather have that be a follow up PR unless it's necessary to properly reproduce the original 1983 method (I don't think it is but not 100% sure). This already works pretty well and is meant to replicate the published method

larsoner:
Actually, I think the sample dataset (and Neuromag data in general) might already use bipolar EOG (HEOG and VEOG each require two electrodes). So I'll mention in this in the example, too

larsoner:
Oh wait this uses CTF data, not sure about their system. But if we're scrapping ECG, then I probably will just switch back to using `sample` anyway (only used Brainstorm auditory to get an ECG signal). But it's likely that CTF EOG is also bipolar

larsoner:
> This brings me to my last point. Typically, horizontal and vertical eye movements are enhanced if we use corresponding (bipolar) EOG channels. This is not even mentioned in the tutorial, and even worse I think it's not even that easy to implement with the current API (correct me if I'm wrong).

I think it should be fairly easy if you create bipolar EOG channels however you need to for your dataset at the beginning. From that point forward, you can compute regression coefficients from these (raw->epochs->subtract_evoked, estimate) and apply them to your data (raw, epochs, or evoked if we want to implement that at some point). So this to me seems like something that should be part of an EEG-specific tutorial if those systems typically do not provide bipolar H/VEOG signals and you have to do some data mangling to get there.

@cbrnr I now at least mention this bipolar-ness in the example now, and think I addressed the other issues you brought up -- feel free to look at the updated one:

https://21673-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_35_artifact_correction_regression.html

I decided to stick with the CTF dataset because the sample dataset blink effects are even smaller so less helpful for the example.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -5,20 +5,37 @@
 Repairing artifacts with regression
 ===================================
 
-This tutorial covers removal of artifacts using regression as in
-:footcite:`GrattonEtAl1983`.
+This tutorial covers removal of artifacts using regression as in Gratton et al.
+1983 :footcite:`GrattonEtAl1983`.
```
cbrnr, line 9: Can you put 1983 in parens (i.e. `(1983)`)?

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -5,20 +5,37 @@
...
 
 .. contents:: Page contents
    :local:
    :depth: 2
 
+Generally speaking, artifacts that result in time waveforms on the sensors
```
cbrnr, line 15: ...~~time~~ waveforms...

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -5,20 +5,37 @@
...
+that are accurately reflected by some reference signal can be removed by
+regression. One example of this is blink artifacts captured by (typically
```
cbrnr, line 17: "is blink artifacts" sounds a bit strange - "are blink artifacts" is maybe better ("Blink artifacts captured by...are one example of this").

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -5,20 +5,37 @@
...
+
+Although ECG signals are typically well captured by ECG electrodes,
+regression-based removal usually does not perform very well.
+This is likely because the heart acts like a rotating dipole, and
+therefore the ECG channel time waveform (typically based on a bipolar
+ECG recording) does not reflect the same temporal dynamics that manifest at
+each MEG channel (obtained by sampling some component of the related magnetic
+vector field). Other approaches like :ref:`ICA <tut-artifact-ica>` or
+:ref:`SSP <tut-artifact-ssp>` will likely work better for ECG.
+
```
cbrnr, line 28: The introduction contains the word "typically" three times - can you use synonyms?

cbrnr:
Very nice, I made a few last comments. Regarding the Gratton method, I actually read the paper now. They do talk about bipolar EOG channels explicitly (i.e. HEOG and VEOG). They also compute separate regression coefficients for blinks and eye movement by restricting the time window within each epoch to where these artifacts actually appear. For eye blinks, only VEOG is needed, and our example using automated blink detection is actually as close as we can get to the original method (the only thing missing is separate regression on eye movement). Maybe we should mention this, but it doesn't need to be reflected in the example code.

With regard to the application to raw data, I'm not sure if the statement "If instead one computed regression weights from the raw data, it is likely that some brain signal would also get removed." is accurate. One the one hand, some amount of brain activity will also be removed by the Gratton method where regression coefficients are computed on epoched data with the evoked response subtracted. I think as long as there are large ocular artifacts in the EOG signals, the regression coefficients will mostly be affected by these signals, so averaging and subtracting an evoked response might not be necessary. In addition, even if you subtract the evoked response, you will still have non-phase-locked brain activity. After all, the original Gratton method was designed to clean an evoked response and not continuous data. In summary, I would not get into too many details and just mention that it is possible to apply the procedure to continuous data, but that this is not the Gratton method anymore.

larsoner:
> I'm not sure if the statement "If instead one computed regression weights from the raw data, it is likely that some brain signal would also get removed." is accurate. 

Agreed it might be overstating things, but it seems like at least the risk of removing the evoked response is higher when you don't subtract the evoked response before computing the coefficients. I think they make this point in the paper, so I'd like to leave some wording along these lines in there. I'll try rewording it to be more conservative

tutorials/preprocessing/plot_35_artifact_correction_regression.py
larsoner: I think it's helpful to make it as clear as possible we're talking about the time-domain signal here

cbrnr: OK

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -6,24 +6,24 @@
...
 that are accurately reflected by some reference signal can be removed by
-regression. One example of this is blink artifacts captured by (typically
-bipolar) EOG channels, which we will examine here.
+regression. Blink artifacts captured by bipolar EOG channels serve as a good
+example of this, so we will demonstrate this here.
 
-Although ECG signals are typically well captured by ECG electrodes,
-regression-based removal usually does not perform very well.
+Although ECG signals are well captured by bipolar ECG electrodes,
+regression-based removal usually of ECG artifacts does not work very well.
```
cbrnr, line 21: ...removal of ECG artifacts usually does not work very well.

tutorials/preprocessing/plot_35_artifact_correction_regression.py
```diff
@@ -145,11 +145,12 @@
...
-# We can also apply the regression directly to the raw data. To do this relies
-# on first computing the regression weights *from epoched data with the evoked
-# response subtracted out* (as we did above).  If instead one computed
-# regression weights from the raw data, it is likely that some brain signal
-# would also get removed.
+# We can also apply the regression directly to the raw data. To do this
+# following the Gratton method requires first computing the regression weights
+# *from epoched data with the evoked response subtracted out* (as we did
+# above). It is possible to instead compute regression weights directly from
+# the raw data, but it this could bias the evoked response more than computing
```
cbrnr, line 152: ..., but ~~it~~this could bias...

cbrnr:
In the last section (applying to continuous data), do you want to mention the possibility to use a separate calibration data set to estimate the regression weights? This is supported by the API, right? If so, I think it would be helpful to show an example of how to do that.

larsoner:
> This is supported by the API, right? If so, I think it would be helpful to show an example of how to do that.

I agree it's useful and thought about adding some text. But I don't know if we have any data to do it, and it's not part of the published method AFAIK, so I'd rather leave it to a motivated person in another PR.

larsoner:
@cbrnr pushed a commit to fix the wording errors, feel free to merge if you're happy

mne/utils/docs.py
```diff
@@ -376,6 +376,42 @@
     Frequency-domain window to use in resampling.
     See :func:`scipy.signal.resample`.
 """
+docdict['decim'] = """
+decim : int
+    Factor by which to subsample the data.
+
+    .. warning:: Low-pass filtering is not performed, this simply selects
+                 every Nᵗʰ sample, i.e., compresses the signal (see Notes).
```
cbrnr, line 384: Is `Nᵗʰ` intentional? I generally like Unicode, but here it might not be necessary. Also, it is not clear what `N` refers to (or maybe it is clear that it refers to `decim`, but then why not be explicit about it).

larsoner: Yes, the old docs said "this simply selects every nth sample" so this is just meant to be a minor improvement to that wording. I guess we could say `"simply subselects one sample for every `decim` samples" or something, but I'm not sure it's clearer. What wording would you find clearer?

cbrnr: Hm. Nth sample is good, maybe we should clarify that N corresponds to the decim argument. Also, maybe we should write Nth and not Nᵗʰ (looks a bit weird with certain fonts)?

kylemath:
thanks everyone, great discussion, and good find @cbrnr of the minor differences between the published method in their restriction of the computation to certain time points, this was also done out of memory need I believe,

Another difference I guess is that we compute the regression for all channels at once in a large regression with multiple Y's, in their method they do it stepwise

agramfort:
thx guys. Example
https://21677-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_35_artifact_correction_regression.html#sphx-glr-auto-tutorials-preprocessing-plot-35-artifact-correction-regression-py
looks
good.

I understand that this does some kind of orthogonal projection in the
temporal domain. It's
like an SSP projection but in the time domain. It means that you could get
more than one
set of betas to be able to remove more than one component. You could also
extrapolate
a few API things from info['proj'] to this time domain approach.

>

PR #8086, opened 2020-08-04
[MRG] Add GFPs to report

hoechenberger:
Add GFP traces to Report
- When plotting Evokeds, add a GFP trace to the sensor traces
- Also add a "GFP-only" comparison plot via `plot_compare_evokeds()`
- Slightly improves coverage of existing tests too

#### Additional information
I tested locally with some data, but let's see how CI artifacts come back. Any feedback greatly appreciated.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -149,7 +149,7 @@ Changelog
 
 - Improved documentation building instructions and execution on Windows by `Eric Larson`_ and `kalenkovich`_
 
-- When passing a list of `~mne.Evoked` objects to `~mne.viz.plot_compare_evokeds`, each evoked's ``.comment`` attribute will be used to label the trace. If ``.comment`` is empty, a 1-based index is assigned as the label by `Richard Höchenberger`_
+- When passing a list of `~mne.Evoked` objects to `~mne.viz.plot_compare_evokeds`, each evoked's ``.comment`` attribute will be used to label the trace. If ``.comment`` is empty, a 1-based index is assigned as the label by  `Richard Höchenberger`_
 
 Bug
 ~~~
@@ -282,8 +282,6 @@ Bug
 - Fix bug in :func:`mne.read_epochs_fieldtrip` when importing data without a ``trialinfo`` field by `Thomas Hartmann`_
 
 - Fix bug in :meth:`mne.preprocessing.ICA.plot_properties` where time series plot doesn't start at the proper tmin by `Teon Brooks`_
-
-- `~mne.Report` now applies baseline correction by default when rendering (whitended) evoked data, as the documentation always suggested; pass ``baseline=None`` to disable this behavior by `Richard Höchenberger`_
 
 API
 ~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -149,7 +149,7 @@ Changelog
 
 - Improved documentation building instructions and execution on Windows by `Eric Larson`_ and `kalenkovich`_
 
-- When passing a list of `~mne.Evoked` objects to `~mne.viz.plot_compare_evokeds`, each evoked's ``.comment`` attribute will be used to label the trace. If ``.comment`` is empty, a 1-based index is assigned as the label by  `Richard Höchenberger`_
+- When passing a list of `~mne.Evoked` objects to `~mne.viz.plot_compare_evokeds`, each evoked's ``.comment`` attribute will be used to label the trace. If ``.comment`` is empty, a 1-based index is assigned as the label by `Richard Höchenberger`_
 
 Bug
 ~~~
@@ -282,6 +282,8 @@ Bug
 - Fix bug in :func:`mne.read_epochs_fieldtrip` when importing data without a ``trialinfo`` field by `Thomas Hartmann`_
 
 - Fix bug in :meth:`mne.preprocessing.ICA.plot_properties` where time series plot doesn't start at the proper tmin by `Teon Brooks`_
+
+- `~mne.Report` now applies baseline correction by default when rendering (whitended) evoked data, as the documentation always suggested; pass ``baseline=None`` to disable this behavior by `Richard Höchenberger`_
 
 API
 ~~~
--- a/mne/tests/test_docstring_parameters.py
+++ b/mne/tests/test_docstring_parameters.py
@@ -14,6 +14,7 @@ from mne.utils import run_tests_if_main, requires_numpydoc, _pl
 public_modules = [
     # the list of modules users need to access for all functionality
     'mne',
+    'mne.baseline',
     'mne.beamformer',
     'mne.chpi',
     'mne.connectivity',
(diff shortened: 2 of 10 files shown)


larsoner:
> I tested locally with some data, but let's see how CI artifacts come back.

You'll need to modify some examples to see this. By default CircleCI only builds changed examples. So make some trivial wording improvement to the ones you want to see built. If you can't find anything to improve, just make some trivial whitespace changes. (Try not to make a change you'll have to revert because this causes extra CI cycles later, delaying CIs on other PRs.)

hoechenberger:
Thanks for the reminder, @larsoner! Will do.

hoechenberger:
I removed the GFP traces from the butterfly plots bc I don't think they're too useful there; but I'm keeping the `plot_compare_evokeds` GFP comparisons

hoechenberger:
Artifact:
https://21559-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/report_evoked.html#5

Entry in the navbar on the left is still missing

hoechenberger:
Report:
https://21560-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/report_evoked.html#5

larsoner:
The EEG GFP being a near-constant value makes me wonder if the projectors were not applied, as the avg ref projector should have made it look less constant

agramfort:
Looks good

The Pb of EEG is that the data are not baseline
Corrected. Historically mne would apply baseline on read and that’s why
Baseline is not None by default in read_evokeds. Report
Does not force a baseline.

hoechenberger:
> The EEG GFP being a near-constant value makes me wonder if the projectors were not applied, as the avg ref projector should have made it look less constant

No, the problem is in the data itself or somewhere else:

```python
import os
import mne

data_path = mne.datasets.sample.data_path(verbose=False)
infile = os.path.join(data_path, 'MEG', 'sample',
                      'sample_audvis-no-filter-ave.fif')

evoked = (mne.read_evokeds(infile, condition='Left Auditory', verbose=False)
          .pick_types(meg=False, eeg=True))
evoked.plot()

print(evoked.info['projs'])

gfp = evoked.data.std(axis=0) * 1e6
print(f'GFP range (uV): {gfp.min():.1f} — {gfp.max():.1f}')
```

Produces
Screenshot 2020-08-05 at 10 34 00



> The Pb of EEG is that the data are not baseline
> Corrected

Son of a gun, you're right!

```python
evoked.apply_baseline().plot()
gfp = evoked.data.std(axis=0) * 1e6
print(f'GFP range (uV): {gfp.min():.1f} — {gfp.max():.1f}')
```

Produces
Screenshot 2020-08-05 at 10 35 46

So… Should we apply baseline correction by default? What if the user would prefer not to use the default baseline period `[start of epoch, 0]`?

hoechenberger:
> So… Should we apply baseline correction by default? What if the user would prefer not to use the default baseline period `[start of epoch, 0]`?

`Report.__init__()` accepts a `baseline` kwarg which is intended for that use. Will add that to the tutorial.

hoechenberger:
So the Report docs say it should perform baseline correction on the interval `(None, 0)` by default, by in actuality by default it would set `baseline=None` and therefore not apply baseline correction. I have addressed this in 6ce0d45. Reports look much better now, locally.

doc/changes/latest.inc
```diff
@@ -283,6 +283,8 @@ Bug
 
 - Fix bug in :meth:`mne.preprocessing.ICA.plot_properties` where time series plot doesn't start at the proper tmin by `Teon Brooks`_
 
+- `~mne.Report` now applies baseline correction by default when rendering (whitended) evoked data, as the documentation always suggested; pass ``baseline=None`` to disable this behavior by `Richard Höchenberger`_
```
agramfort, line 286:
```suggestion
- `~mne.Report` now applies baseline correction by default when rendering (whitened) evoked data, as the documentation always suggested; pass ``baseline=None`` to disable this behavior by `Richard Höchenberger`_
```

I am not sure the report should do anything more than what was done during processing. It may give some impression of the data in the report that is not what will be done eg before source imaging. Report should not do any computation logic. If need be the sample data should be fixed.

agramfort: or report should have a baseline parameter exposed.

agramfort: actually you have one but it should not be (None, 0) by default. So I would revert the change can set it in the tutorial

hoechenberger: I've added a new example to the tutorial, showing how to use the `baseline` parameter and which effect it has on the data vizualization.

I then also realized that in the existing example for whitening, it's stated that the data is whitened and baseline-corrected; however it was only whitened. I have now added baseline correction there too.

hoechenberger:
generated reports:

- [without baseline correction](https://21590-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/report_evoked.html)
- [with baseline correction](https://21590-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/report_evoked_baseline.html)
- [whitened with baseline correction](https://21590-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/report_evoked_whitened.html)

tutorials/misc/plot_report.py
```diff
@@ -115,21 +115,39 @@
...
 report.save('report_evoked.html')
 
 ###############################################################################
-# To render whitened :class:`~mne.Evoked` files with baseline correction, add
-# the noise covariance file. This will display ERP/ERF plots for both the
-# original and whitened :class:`~mne.Evoked` objects, but scalp topomaps only
-# for the original.
+# You have probably noticed that the EEG recordings look particularly odd. This
+# is because by default, `~mne.Report` does not apply baseline correction
+# before rendering evoked data. To request baseline correction, pass a
```
larsoner, line 130: It's both a combination of how the Report works and how the dataset was created, i.e.:

"... evoked data, and the MNE sample evoked dataset was created without applying baseline correction."

hoechenberger: Right, the phrasing was a little unfortunate there. I tried to address this in 4248ff7, WDYT?

hoechenberger:
Oh wow, seems I screwed up the docstyle big time. Only checked that the rendered docs looked good :\

hoechenberger:
Good to go from my end.

mne/tests/test_docstring_parameters.py
```diff
@@ -14,6 +14,7 @@
 public_modules = [
     # the list of modules users need to access for all functionality
     'mne',
+    'mne.baseline',
```
larsoner, line 17: Pushed this to reflect adding it to `python_reference.rst`

larsoner:
Thanks @hoechenberger

mne/tests/test_docstring_parameters.py
hoechenberger: Thanks!

PR #8089, opened 2020-08-04
MRG, FIX: Fix bug with raw scrolling

larsoner:
Fix a bug with raw plotting where going to the end and back (when the number of displayed channels changes) caused an error. Introduced by #7811 within-release so no `latest.inc` update.


agramfort:
Thx @larsoner

PR #8090, opened 2020-08-04
MRG, ENH: Add single_volume option to setup_volume_source_space

larsoner:
It's nice to be able to create a volumetric source space using FreeSurfer's complete `aseg.mgz` definition, without the `Unknown` vertices, like:
```
src_to_fname = op.join(subjects_dir, 'fsaverage', 'bem',
                       'fsaverage-vol-5-aseg-src.fif')
aseg_fname = op.join(subjects_dir, 'fsaverage', 'mri', 'aseg.mgz')
volume_label = mne.get_volume_labels_from_aseg(aseg_fname)
volume_label.pop(volume_label.index('Unknown'))
assert len(volume_label) == 43
if not op.isfile(src_to_fname):
    src_to = mne.setup_volume_source_space(
        'fsaverage', 5., 'aseg.mgz', volume_label=volume_label,
        subjects_dir=subjects_dir, verbose=True)
```
This creates a source space with 46 entries. A problem with this is that there is a ton of redundant information: the source `rr`, whether used or unused for a give subvol, are part of each of the 46 spaces. So when you go to save this to disk, you get a 2GB write error.

A nice solution is to merge all of these into a single volume with `single_volume=True`, which makes `len(src) == 1`, and takes up only 200 MB of disk space. Now that we have `extract_label_time_course` support for volumes and other extraction tools, it's easy enough to later extract the individual subvolumes later.

FWIW I actually prefer this method to using a `inner_skull`-bounded surface, as it ensures it only includes source points that FreeSurfer considered to be part of the actual brain by `.pop`-ing the `'Unknown'` label. Not sure if it's worth updating an example or tutorial to mention (or use) this approach...

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -43,6 +43,8 @@ Changelog
 
 - Add support for omitting the SDR step in volumetric morphing by passing ``n_iter_sdr=()`` to `mne.compute_source_morph` by `Eric Larson`_
 
+- Add ``single_volume`` argument to :func:`mne.setup_volume_source_space` to facilitate creating source spaces with many volumes (e.g., all subvolumes of ``aseg.mgz``) by `Eric Larson`_
+
 - Add support for passing a string argument to ``bg_img`` in `mne.viz.plot_volume_source_estimates` by `Eric Larson`_
 
 - Add support for providing the destination surface source space in the ``src_to`` argument of :func:`mne.compute_source_morph` by `Eric Larson`_
--- a/mne/morph.py
+++ b/mne/morph.py
@@ -857,6 +857,11 @@ def _interpolate_data(stc, morph, mri_resolution, mri_space, output):
 ###############################################################################
 # Morph for VolSourceEstimate
 
+def _compute_r2(a, b):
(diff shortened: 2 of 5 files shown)


larsoner:
Pushing a commit to improve the `compute_source_morph` output so you can more easily see how the variance explained increases as a function of morph step used:
```
Computing nonlinear Symmetric Diffeomorphic Registration...
Optimizing translation:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
Optimizing rigid-body:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:   23.3 mm
    Rotation:      11.3°
    R²:            95.6%
Optimizing full affine:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    R²:            96.9%
Optimizing SDR:
    R²:            98.9%
```

agramfort:
thx @larsoner

PR #8091, opened 2020-08-04
MRG, BUG: Avoid warning about rank when its explicit

larsoner:
`rank` was only passed to one of two functions that could compute the rank, so you could get a warning about rank deficiency even if you've explicitly set the rank.

### DIFF
--- a/mne/minimum_norm/tests/test_inverse.py
+++ b/mne/minimum_norm/tests/test_inverse.py
@@ -53,6 +53,7 @@ fname_inv_meeg_diag = op.join(s_path,
 fname_data = op.join(s_path, 'sample_audvis_trunc-ave.fif')
 fname_cov = op.join(s_path, 'sample_audvis_trunc-cov.fif')
 fname_raw = op.join(s_path, 'sample_audvis_trunc_raw.fif')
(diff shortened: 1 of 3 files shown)


mne/minimum_norm/tests/test_inverse.py
```diff
@@ -1190,4 +1191,24 @@ def test_inverse_mixed_loose(mixed_fwd_cov_evoked):
...
+    raw.rename_channels(
+        {ch_name: f'{ch_name[:3]} {ch_name[3:]}' for ch_name in raw.ch_names})
+    fwd = mne.read_forward_solution(fname_fwd)
+    cov = mne.read_cov(fname_cov)
+    with pytest.warns(RuntimeWarning, match='rank as it exceeds.*302 > 67'):
+        inv = make_inverse_operator(raw.info, fwd, cov)
+    rank = (inv['noise_cov']['eig'] > 0).sum()
+    assert rank == 302
+    # should not warn
+    inv = make_inverse_operator(raw.info, fwd, cov, rank=dict(meg=67))
```
larsoner, line 1208: on `master` this warns about the data-driven rank calc 302 > 67 (which it shouldn't -- rank is passed explicitly)

PR #8100, opened 2020-08-05
MRG BUG/API: Fix bugs with KIT/Ricoh reader

larsoner:
@apesquita can you try this branch with your data? In particular, as @agramfort said, some simple dataset would be great. Ideally something like a high-SNR auditory evoked dataset with MRI-based N100 source localization and `plot_alignment` calls should tell us if we're doing things correctly.

Introduces a `standardize_names=None` that defaults to True in 0.21 but changes to False in 0.22. False means "use names if present in the file". We don't have any files like this, but it seems to work in local testing. It would be good to get some tiny test file to really test this, but in the meantime I've modified a test to modify an existing test file to mock it.

Todo:

- [x] Check that localization and plot_alignment are correct
- [x] Fix EEG scaling
- [x] Choose/check naming for systems


larsoner:
Actually it looks like @Kyung-minAn sent @agramfort some data for testing. So we don't need any data from you @apesquita , but it would be good to get feedback on how it's working for you!

apesquita:
Super! I am on leave for a few days. But I'll give this a go next week.
Many thanks. Ana

Kyung-minAn:
Good!! 
@larsoner and @agramfort, Thank you for developing and testing it.
If you need more dataset for testing, please tell me~

@apesquita, Thanks~ Ana!

larsoner:
Okay @Kyung-minAn with your auditory data I can obtain these plots:

Screenshot from 2020-08-06 08-59-56

Screenshot from 2020-08-06 09-00-41

Screenshot from 2020-08-06 08-55-58

Was the head tilted that far forward?

It's entirely possible my rough coreg is the cause of the activation being localized on the wrong side of the Sylvian fissure. I think this is probably okay. What do you think?

@Kyung-minAn would it be possible for you to record the tiniest file possible (e.g., 1 sec data) that includes coregistration? I already have a dataset that is 20 sec long. If this is as short as possible, then we can use that for testing already. But it would be better to have a smaller one if possible.

mne/io/kit/constants.py
```diff
@@ -150,6 +161,7 @@
 KIT.SYSTEM_UMD_2014_07 = 52  # UMD update to 16 bit ADC, July 4, 2014 -
 KIT.SYSTEM_UMD_2014_12 = 53  # UMD December 4, 2014 -
 KIT.SYSTEM_UMD_2019_09 = 54  # UMD September 3, 2019 -
+KIT.SYSTEM_YOKOGAWA_440 = 1001  # no idea
```
larsoner, line 164: @Kyung-minAn you can see the naming scheme we use lists the site and the month/year of installation of the system (at least I assume that's what this is -- @teonbrooks ? ). Do you have this information?

larsoner:
@Kyung-minAn actually never mind about the testing data -- looks like we already have some nice small files!

Kyung-minAn:
@larsoner 
I upload a tiny data I cut 1 s long. 
[020321_PureTone_MEG160-export_1s.zip](https://github.com/mne-tools/mne-python/files/5035304/020321_PureTone_MEG160-export_1s.zip)
If small files you have is not suitable for testing, I would be glad that you use this one.

Kyung-minAn:
@larsoner , Thank you for the test results.
The plots you uploaded look so nice~!!

Although the sources are localized outside of the Sylvian fissure, I think that your test is successful.
This is a plot when I performed coregistration for the same data. 
image

I think the localization error results from the noisy MRI data. 

I'm so excited for your progress!!
Thanks a lot!!!

Min

larsoner:
Pushed a commit to fix EEG gains, now we verify that the values we get from the MEG/EEG/misc channels match what's in the PDF. Thanks for the excellent test data!

mne/io/kit/constants.py
larsoner: @Kyung-minAn any thoughts on the naming here?

Kyung-minAn: @larsoner, I have no idea.
I think we would better to ask KIT for this information.

larsoner: For now I just went with `SYSTEM_YOKOGAWA`. We can always improve this later as it's just for internal use

larsoner:
Ready for review/merge from my end @agramfort

mne/io/kit/constants.py
teonbrooks: yep, I believe that these are assigned at time of installation or servicing

mne/io/kit/constants.py
```diff
@@ -167,6 +179,7 @@
     KIT.SYSTEM_UMD_2014_07: None,
     KIT.SYSTEM_UMD_2014_12: 'KIT-UMD-3',
     KIT.SYSTEM_UMD_2019_09: None,
+    KIT.SYSTEM_YOKOGAWA: None,
```
agramfort, line 182: @apesquita @Kyung-minAn can you help us here? what can you tell us about the MEG system name/version/location?

mne/io/kit/constants.py
```diff
@@ -202,6 +216,7 @@
     KIT.SYSTEM_UMD_2014_07: 'University of Maryland, 2014',
     KIT.SYSTEM_UMD_2014_12: 'University of Maryland, 2014-',
     KIT.SYSTEM_UMD_2019_09: 'University of Maryland, 2019-',
+    KIT.SYSTEM_YOKOGAWA: 'Yokogawa ??? -',
```
agramfort, line 219: @apesquita @Kyung-minAn can you help us here?

paulsowman:
@larsoner Hi, just ran a quick comparo of this branch on our KIT MEG160 and KIT MEG125

Certainly aligns the helmet and sensors, but the face area is filled in on the MEG160 (bottom lefthand image)?

master
Screen Shot 2020-08-11 at 2 32 17 pm

ricoh
Screen Shot 2020-08-11 at 2 31 38 pm

@Kyung-minAn have you tried this?

Best, P

mne/io/kit/constants.py
Kyung-minAn: Dear @teonbrooks & @larsoner 
Yes. I also believe that these are the information about the MEG systems which is installed in that date.
KIT installed several MEG systems including adult- and child-customized systems.
And there might be differences in the number of channels and/or sensor arrays.

I think that It's a good chance to check it in details.
I will contact with KIT and check it out.

@larsoner, You went well with 'SYSTEM_YOKOGAWA'. 
I wonder that the results would be changed or not, If you test with another system info (e.g. 'SYSTEM_UMD'?

Many thanks,
Min

mne/io/kit/constants.py
Kyung-minAn: @agramfort , I'm asking to KIT about this issue.
I will let you know as soon as I get this info.
Thank you for developing the reader~

larsoner:
@paulsowman can you `mne.io.write_info(raw.info)` and upload it somewhere for me to try? That way I can debug the helmet-over-face issue

larsoner:
@paulsowman actually don't bother, I can replicate with our own test KIT data if I make it generate a helmet on the fly. I'll try to fix it

larsoner:
Okay @paulsowman feel free to try again. On `master` this is what the following code does with `testing` data -- it looks nice and smooth, but there is already some misalignment because our built-in helmet is not a good match for the sensor locations:
```
python -c "import mne; raw = mne.io.read_raw_kit('test_as-raw.con'); mne.viz.plot_alignment(raw.info, show_axes=True, eeg=False, surfaces=[])"
```
Screenshot from 2020-08-11 11-23-08

If I force it to generate a helmet (like it does in your case), it gets ugly:

Screenshot from 2020-08-11 11-23-26

This PR fixes helmet generation, and actually just always does it for KIT now -- it's not as pretty as our smoothed surface, but it should more accurately represent the geometry:

Screenshot from 2020-08-11 11-22-20

Can you see if it fixes your "filled in face" problem?

mne/io/kit/constants.py
Kyung-minAn: @agramfort I contacted KIT for this issue.
I heard that It might take some days for the person in charge to check e-mail because it is the holiday season in Japan.

teonbrooks:
hey, I just tried this out some of the KIT data and nothing is broken, channel naming is the same regardless of the `standardize_name` value (guessing no names are stored for ours), helmet is fine, everything seems to be OK

agramfort:
Thanks everyone !

christianbrodbeck:
KIT UMD also seems to be working as expected (thanks @jpkulasingham for testing!)

mne/io/kit/constants.py
Kyung-minAn: Dear all,
I checked this naming scheme with KIT and Ricoh company.
Your expectation is right. They mentioned that It is installation sites (University of Maryland / Yokogawa system) and date of calibration.
Thank you for developing and testing.

mne/io/kit/constants.py
Kyung-minAn: @agramfort,
I asked KIT and Ricoh company about this information.
They provided information about each MEG system (name/version/location).
SYSTEM_UMD_2014_07 :

mne/io/kit/constants.py
Kyung-minAn: @agramfort 
I hope that the information written above would be helpful.

mne/io/kit/constants.py
agramfort: @Kyung-minAn what would be great would be to have a tiny file from each of these YOKOGAWA systems. Do you think it's possible?

Kyung-minAn: @agramfort 
Do you mean three Yokogawa systems?

larsoner: It would be good to have a file from each of the three Yokogawa systems if possible, yes. Then we can test that these values are used/set correctly by us

Kyung-minAn: @agramfort @larsoner 
I appreciate your great contributions.
The short data that I sent before is recorded from the child-customized Yokogawa MEG system (PQ1151R).
For two Yokogawa adult MEG systems, I will ask Ricoh company to provide the data.
Do you have any preference for the testing data? Would it be fine with each coordinated con file (1-s long) and MRI image?
Thanks a lot~!!

agramfort:
no need of MRI but 10s of data for each system would be useful.

>

larsoner:
I would say 1 sec is even better than 10 sec (less space for testing files)

Kyung-minAn:
@agramfort @larsoner 
Ricoh company sent me three con files from adult MEG system (1s long).

PQA160C/1001/Kanazawa (until 2017)
RICOH160-1/10020/Kanazawa (since 2018)
RICOH160-1/10021/Kanazawa (since August 2020)

The last file is the recent version of the adult MEG system. 
They changed sensor layouts during their upgrade.

[Yokogawa_Ricoh_sample_data_1s.zip](https://github.com/mne-tools/mne-python/files/5155385/Yokogawa_Ricoh_sample_data_1s.zip)

Thank you for your help.

agramfort:
see https://github.com/mne-tools/mne-python/pull/8187/files

Kyung-minAn:
@agramfort 
Thank you for your effort. I checked it.
For a child-customized Yokogawa MEG system (the data I sent initially, PQ1151R/903/Kanazawa), would it be better to be added in the Ricoh data reader as well?

agramfort:
can you share again the file PQ1151R/903/Kanazawa ? is it tiny?

>

apesquita:
@agramfort @Kyung-minAn 

Hello,

I am trying to catch-up. A lot has happened! Thanks so much!

1. I have this small file from the child-size system that @Kyung-minAn had shared before. Does this one work for your purposes?

[motor_task_tiny.zip](https://github.com/mne-tools/mne-python/files/5195542/motor_task_tiny.zip)

2. Which layout from the following should I choose for a child-size data? I am not sure if this is yet available or not. 
['CTF-275.lout', 'CTF151.lay', 'CTF275.lay', 'EEG1005.lay', 'EGI256.lout', 'KIT-125.lout', 'KIT-157.lout', 'KIT-160.lay', 'KIT-AD.lout', 'KIT-AS-2008.lout', 'KIT-UMD-3.lout', 'Neuromag_122.lout', 'Vectorview-all.lout', 'Vectorview-grad.lout', 'Vectorview-grad_norm.lout', 'Vectorview-mag.lout', 'biosemi.lay', 'magnesWH3600.lout']

Thanks for helping me catch up!

Ana P

Kyung-minAn:
@apesquita, Thank you for sharing sample data. 
The data was recorded from child-customized MEG system (PQ1151R/903/Kanazawa).

@agramfort, Thank you for checking the data. If you need more tiny data, please tell me. I will cut it for 1 s long. 

Many thanks,

agramfort:
doing

```
import mne
raw = mne.io.read_raw_kit('010409_Motor_task_coregist-export_tiny.con')
raw.info['bads'] = ['MEG 099']
raw.plot_sensors()
raw.plot_psd(sphere='auto')
```

gives

Screenshot 2020-09-10 at 10 43 57

you see that there is hope that layout is already found out properly.

Just a shift problem with plot_psd. Any idea @larsoner ?

larsoner:
> Just a shift problem with plot_psd. Any idea @larsoner ?

If the head was toward the right side of the sensors during recording, this is a feature, not a bug. The topomap plots (and sensor plots) now try to account for `info['dev_head_t']` and also the head size when deciding where and how large to draw the head circle relative to the sensors

agramfort:
but why is it different when you do plot_sensors?

>

larsoner:
Ahh I didn't see the difference. Yeah it seems like those two plots should have the same head-sensor relationship, and they don't. So I agree it's a bug

PR #8107, opened 2020-08-07
MRG, MAINT: Go back to OrderedDict

larsoner:
As discussed in #8014, reverts parts of #7635: dict ordering was a CPython implementation detail in 3.6, becoming an official language feature in 3.7, so to be safe let's use OrderedDict until we require 3.7+.

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -2,6 +2,7 @@
 #
 # License: BSD (3-clause)
 
+from collections import OrderedDict
 from datetime import datetime, timedelta, timezone
 import os.path as op
 import re
@@ -249,7 +250,7 @@ class Annotations(object):
             out_keys = ('onset', 'duration', 'description', 'orig_time')
             out_vals = (self.onset[key], self.duration[key],
                         self.description[key], self.orig_time)
-            return dict(zip(out_keys, out_vals))
+            return OrderedDict(zip(out_keys, out_vals))
         else:
             key = list(key) if isinstance(key, tuple) else key
             return Annotations(onset=self.onset[key],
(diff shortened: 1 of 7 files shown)


cbrnr:
When will this happen? Not in the next release?

larsoner:
When will what happen? Requiring 3.7? No not in the next release. EOL for 3.6 appears to be around Dec 2021, so probably not for a couple releases.

hoechenberger:
Bummer!

Do we actually support anything but CPython?

cbrnr:
I don't think so. If MNE runs on PyPy then that's good, but I wouldn't support it (and it's likely that many things just work out of the box because we're not using Cython or any other compiled modules).

hoechenberger:
Hm, just wondering, if we're not officially supporting & testing on anything but CPython, why even bother about reverting to OrderedDict…

drammock:
do we even have an "official" policy of what python implementations we support?  I've not seen one...

cbrnr:
I don't think we have, but Python 3.6 should be independent of the actual implementation. That said, dicts preserve insertion order as of 3.6, so I see your point...

cbrnr:
(To clarify: I'm :+1: on reverting to `OrderedDict` to stay independent of the Python implementation.)

larsoner:
I'd expect MNE to work on PyPy3 as it's pure python, though I've never tried.

I don't think it'll be too onerous for us to stick to the official language features that night also increase our python implementation compatibility. I'd vote for sticking to official language features as we drop support for old versions.

agramfort:
thx @larsoner !

PR #8108, opened 2020-08-07
MRG, BUG: Fix bug with Report

larsoner:
Fixes a problem with class naming when using `add_bem_to_section` directly instead of `render_bem=True`. Only the name on the left would be hidden by clicking the button, not the content itself.


agramfort:
thx @larsoner

PR #8116, opened 2020-08-11
MRG: Recreate our helmet graphic

larsoner:
I tried to recreate [our helmet](https://github.com/mne-tools/mne-python/blob/master/doc/_static/mne_helmet.png):

Using Mayavi I get this output for the new example:

Screenshot from 2020-08-11 13-56-44

If I switch to PyVista, I get:

Screenshot from 2020-08-11 13-56-58

Problems thus far:

1. The coincident topology is not resolved for mayavi or PyVista (if you rotate the scene, PyVista also cuts off parts of the yellow). Perhaps if `plot_evoked_field` could set the `polygon_offset` it could be fixed?
2. The coincident topology for the field lines is wrong for PyVista -- they should be on top.
3. The camera systems are not equivalent between PyVista and Mayavi.

@GuillaumeFavelier can you look into these when you get a chance? Feel free to push directly here


GuillaumeFavelier:
I use a fixed width for the tubes, it would be better to find a better value automatically.

The coincident topology still needs to be fixed.

larsoner:
Indeed [it is getting closer](https://21815-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_mne_helmet.html#sphx-glr-auto-examples-visualization-plot-mne-helmet-py):

larsoner:
For coindicent topology, can we have something like `polygon_offset=0` in some `mesh` call? Then it sets the appropriate Mayavi and PyVista properties under the hood (use polygon offset resolution + set polygon offset)? Assuming it fixes the problem then it's okay to make use of this in the example, and we can use it in `_Brain`, too. (I've noticed some coincident topology problems there, too, from time to time when rotating a brain.) We probably need to enable the polygon offset for all meshes for it to work (?).

GuillaumeFavelier:
I added `polygon_offset` in `mesh()` and `surface()` and updated the docstrings. It also replaces `_resolve_coincident_topology` in `_Brain`. I'm not sure what is the best strategy to resolve those issues. For now, I just set some fixed values that give the best result based on what I observe.

> We probably need to enable the polygon offset for all meshes for it to work (?)

That I don't know. I did some experiments locally and it does not seem necessary but it may be in the future once we have a definitive solution.

Also about Mayavi, I did not find any alternative to `SetRelativeCoincidentTopologyPolygonOffsetParameters`. Hera are some relevant default values that I found:

```py
...
'resolve_coincident_topology': 'polygon_offset',
'resolve_coincident_topology_': 1,
'resolve_coincident_topology_polygon_offset_faces': 1,
'resolve_coincident_topology_z_shift': 0.01,
...
```

GuillaumeFavelier:
This is the helmet I have with the new changes:

image

GuillaumeFavelier:
By letting the contour filter detect the range of scalar values, I get the same result:

image

larsoner:
- Can you try playing with the Mayavi parameters window (leftmost icon in the toolbar) to see if you can find the right button to tweak there to fix the coincident topology? If you can't get it to work I can try.
- Should we add a parameter for the line width of the contours? It looks like mayavi defaults to 1 and PyVista doesn't

GuillaumeFavelier:
1) I can try
2) For PyVista these are tubes now, I will render again using just the lines. To tweak that, there is the `width` parameter.

This is what I use:

```py
        # And the field lines on top
        renderer.contour(surface=surf, scalars=data, contours=n_contours,
                         vmin=-vmax, vmax=vmax, opacity=alpha,
                         colormap=colormap_lines, kind='tube',
                         width=0.0005)
```

larsoner:
Do the lines also require a `polygon_offset` tweak?

larsoner:
(The tubes presumably do not because they actually protrude from the surface; the lines OTOH will be coincident)

GuillaumeFavelier:
Seems like it does not matter it it's just a screenshot. The difference is more visible when you interact with the scene:

line | tube
---|----
image | image

> Do the lines also require a polygon_offset tweak?

Apparently they don't the modification.

larsoner:
Okay, feel free to choose either one based on which one you think looks better / more like the original. I guess the only remaining issue is the Mayavi coincident topology, right?

GuillaumeFavelier:
> I guess the only remaining issue is the Mayavi coincident topology, right?

Yes indeed

GuillaumeFavelier:
I have pretty much the same mapper options:

image

I tried `z-shift` but there is some noise remaining near the cell edges:

image

The best solution I found was to "Force translucent" the white surface (that way, the colors blend):

image

Result:

image

larsoner:
pip-pre errors/segfaults are https://github.com/numpy/numpy/issues/17198 and can be ignored for now

GuillaumeFavelier:
Update after #8182 :

Mayavi | PyVista
---|---
image | image

GuillaumeFavelier:
Also we should consider making PyVista the default backend. It's an item of #7162

agramfort:
>
> Also we should consider making PyVista the default backend. It's an item
> of #7162 <https://github.com/mne-tools/mne-python/issues/7162>
>
+10 !

can you start a PR for this?

GuillaumeFavelier:
@agramfort I opened https://github.com/mne-tools/mne-python/pull/8220 :)

mne/viz/_3d.py
```diff
@@ -304,7 +304,8 @@ def _set_aspect_equal(ax):
 
 @verbose
 def plot_evoked_field(evoked, surf_maps, time=None, time_label='t = %0.0f ms',
-                      n_jobs=1, fig=None, verbose=None):
+                      n_jobs=1, fig=None, vmax=None, n_contours=21,
```
agramfort, line 307: 21 was the mayavi default?

agramfort (approved):

mne/viz/_3d.py
GuillaumeFavelier: yes, IIRC

agramfort:
ok merge when green then !

>

mne/viz/_3d.py
GuillaumeFavelier: My bad, [it's actually 10](https://github.com/enthought/mayavi/blob/9c128635943e28513776dec68492ffd0268c28d1/tvtk/tools/mlab.py#L771), 21 was the number of contours used in the example. I will change that.

GuillaumeFavelier:
I pushed a commit to tweak the distance too:

Before | After
---|---
image | image

 WDYT @larsoner ?

agramfort:
perfect on my side !

larsoner:
Thanks @GuillaumeFavelier !

PR #8117, opened 2020-08-11
WIP, ENH: Add renderer icon

larsoner:
Uses https://github.com/pyvista/pyvistaqt/pull/48 to set the icon for the Brain renderer. Looks like this:

<details>
<summary>Original</summary>

Screenshot from 2020-08-11 14-12-56

</details>

EDIT: Final version:

Screen Shot 2020-08-12 at 10 29 25 AM

Todo:

- [x] Fix the "Unknown" label, probably also in pyvistaqt
- [x] Get bikeshedded to death on the icon
- [x] Figure out how to get a higher-res one to work
- [x] Wait for PyVista PR to be merged


larsoner:
Okay should be ready for review/comments on the icon. We need to wait for the pyvistaqt PR to be merged to be confident in the API, but we can decide if the icon is acceptable in the meantime.

drammock:
> Get bikeshedded to death on the icon

I'll start :) 

1. just like it is in your screenshot
2. just like it is in screenshot, only without the rainbow (a solid color "MNE", probably white or black)
3. a brain. But [the fontawesome brain](https://fontawesome.com/icons/brain?style=solid) is ugly, so we'd have to roll our own. Not volunteering.

I vote for 2

cbrnr:
Bikeshedding, let's gooo! Is there a (transparent) rectangle around MNE or is it just the text? All icons in your dock have a round or rectangular outline (same on macOS), so I'd vote for 1 with a rectangular transparent outline.

larsoner:
> Bikeshedding, let's gooo! Is there a (transparent) rectangle around MNE or is it just the text?

I had my mouse over the icon, that's what gnome shell does when you mouse over

> I'd vote for 1 with a rectangular transparent outline.

If I add a black circle and remove the contour lines (they don't render anyway) we get:

Screenshot from 2020-08-11 15-37-49

drammock:
> If I add a black circle and remove the contour lines

I can live with that version.

cbrnr:
Very nice! I like the version with the black circle!

agramfort:
+1  for black circle background

MNE is becoming fancy ;)

cbrnr:
I'd need an icon for MNELAB as well. I'd like to steal the black circle one, but it should say MNELAB, which is probably too wide for a good icon. Any suggestions?

larsoner:
Same design but just say "LAB" or "MNE\nLAB" would work

cbrnr:
> Same design but just say "LAB" or "MNE\nLAB" would work

"LAB" is actually a pretty good idea, because MNE is implied by its colors... at least for people in the know. I'll try that.

mne/icons/mne.qrc
```diff
@@ -10,5 +10,6 @@
     <file alias="clear.svg">clear-black-18dp.svg</file>
     <file alias="screenshot.svg">screenshot-black-18dp.svg</file>
     <file alias="movie.svg">movie-black-18dp.svg</file>
+    <file alias="mne-icon.png">mne-circle-black.png</file>
```
larsoner, line 13: @GuillaumeFavelier let me know if this fits into the existing Qt resource scheme well enough

larsoner:
I ended up having to make it a 256x256 PNG (~10kB) instead of SVG because Qt's SVG implementation does not (and will ever not it seems) support clip paths. Seems to render in HiDPI properly on my macOS machine:

Screen Shot 2020-08-12 at 10 29 25 AM

So now we just need to wait for the pyvistaqt PR to be merged to make sure we're using the correct API calls.

hoechenberger:
> * just like it is in screenshot, only without the rainbow (a solid color "MNE", probably white or black)

+1

hoechenberger:
I'd advise against the circular shape, considering the icon redesigns coming with Big Sur in a few months:



big-sur-macbook-pro

A circular icon would look outdated again by the end of this year.

drammock:
> A circular icon would look outdated again by the end of this year.

Only for mac users, and even then, only those who upgrade their OS or get a brand new machine.  Also the apps in the screenshot look like they're all built-in apps (?). There is no guarantee that other vendors' apps that are currently circular (Firefox, Rstudio, Slack, matplotlib figures...) will get changed to match Apple's latest favored shape.

hoechenberger:
> > A circular icon would look outdated again by the end of this year.
> 
> 
> 
> Only for mac users, 

Yes, macOS is the only system where that matters, because on all other systems things are a hell of a lot more inconsistent anyway 😁 

> and even then, only those who upgrade their OS or get a brand new machine. 

I merely wanted to point out that it might be a good idea to think about the "future-proofness" of the icon if we're doing a new design anyway. Having a new-style icon on an old macOS doesn't hurt bc things grew to become a little messy over the past years. This will change with Big Sur.

> Also the apps in the screenshot look like they're all built-in apps (?).

In this Screenshot, yes. Manufacturers of 3rd-Party apps have already begun to create new icons, though. 

> There is no guarantee that other vendors' apps that are currently circular (Firefox, Rstudio, Slack, matplotlib figures...) will get changed to match Apple's latest favored shape.

I bet you 20 dollars that all regularly updated commercial apps will have updated icons within a few weeks after Big Sur hs been released. That lots of FLOSS software, OTOH, is neither at the forefront of UI design nor accessibility is no secret...

drammock:
> it might be a good idea to think about the "future-proofness" of the icon

To me it sounds like coupling an aspect of our branding to Apple's branding.  Your line of argument ("we shouldn't have an icon that looks old-fashioned") will apply again, equally forcefully, next time Apple decides to change their preferred icon shape, which I think shows that this is not "future proofing". If anything, it puts us on the hook to change the icon again and again to keep up with Apple's design changes.

I don't want an ugly icon. But I don't think that "looking like I'm not part of the OS" is enough to make an icon ugly. Thus I see no need to visually communicate that our software is "part of the same family" of the software that comes standard with any of the OSes that we run on (regardless of how internally consistent those OSes are in their icon design, and regardless of what other "regularly updated commercial apps" do).

cbrnr:
If Apple really wanted to enforce their new rounded rectangular icon shapes, they could do it with existing icons of any shape (that's what many Android launchers do). Therefore, I don't think it is very important to preemptively try to follow Apple's latest guidelines as long as we have a nice icon. Also, it shouldn't be difficult to replace the black circle with a black rectangle with rounded corners. I'd be fine with doing that once Big Sur is out, but for now I'd stick to the circle (unless @larsoner wants to showcase some shape variations).

GuillaumeFavelier:
https://github.com/pyvista/pyvistaqt/pull/48 has been merged :)

mne/icons/mne.qrc
GuillaumeFavelier: This is good enough :+1:

Issue #8120, opened 2020-08-14
Load NIRS data recorded using NIRStar v15.3


DylanMannKrzisnik:
Hello MNE users,

Following recommendations from the mailing list, I'm opening an issue with regards to loading NIRS data collected from a NIRx system, using NIRScout software version 15.3. Loading data using ***mne.io.read_raw_nirx()*** works for data collected using NIRScout versions 15.0 and. 15.2, but not 15.3.

I've created a [repo](https://github.com/DylanMannKrzisnik/NIRS_sample_data) to host some sample data. In short, it consists of a finger-tapping block-design experiment with bilateral motor coverage using 8 sources and 4 detectors. As mentioned in the repo's README file:

    The following command serves to import data using MNE-Python:

    **raw_intensity = mne.io.read_raw_nirx("full_path_to_data/NIRS_sample_data")**

    However, this results in the following error message:

    **RuntimeError: MNE does not support this NIRStar version ("15.3")**

More info in the repo.

What would be the best way around this issue? If anyone has suggestions, my labmates and I can look into it. Just wanted to know if anyone has some advice. Let me know if anything is missing from the repo.

Regards,

Dylan Mann-Krzisnik

larsoner:
Have you tried first just hacking the code to allow a "15.3" string here:

https://github.com/mne-tools/mne-python/blob/master/mne/io/nirx/nirx.py#L110-L111

Try making that change then following the fNIRS preprocessing tutorial:

https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#sphx-glr-auto-tutorials-preprocessing-plot-70-fnirs-processing-py

That will tell us if things at least mostly work already.

Then someone will need to figure out what changed between 15.2 and 15.3 to see if we actually need to add test files, etc. If they only made irrelevant changes (e.g., changing fields we don't use or adding new fields we don't care about) then modifying the string in `master` might already be good enough.

cc @rob-luke in case you have other ideas

DylanMannKrzisnik:
Hi again,

Indeed this seems to work! By adding "\"15.3\"" to the iterable in
L110-L111 of nirx.py I was able to run the tutorial scripts, with some
modifications to adapt to the sample data.
I should have done that first before posting anything. However, I was weary
about the downstream effects of such a simple modification. This alludes to
your point about needing to know what changed between 15.2 & 15.3.

We'll do more research on our side to see what the potential problems could
[...]

PR #8121, opened 2020-08-14
MRG: Added support for NIRStar version 15.3

DylanMannKrzisnik:
#### Reference issue
Fixes issue #8120 

Added support for NIRStar version 15.3 when importing NIRx data using mne.io.read_raw_nirx().

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -108,7 +108,7 @@ class RawNIRX(BaseRaw):
 
         # Check that the file format version is supported
         if not any(item == hdr['GeneralInfo']['NIRStar'] for item in
-                   ["\"15.0\"", "\"15.2\""]):
+                   ["\"15.0\"", "\"15.2\"", "\"15.3\""]):
             raise RuntimeError('MNE does not support this NIRStar version'
                                ' (%s)' % (hdr['GeneralInfo']['NIRStar'],))
         if "NIRScout" not in hdr['GeneralInfo']['Device']:
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -107,10 +107,12 @@ class RawNIRX(BaseRaw):
         hdr.read_string(hdr_str)
 
         # Check that the file format version is supported
-        if not any(item == hdr['GeneralInfo']['NIRStar'] for item in
-                   ["\"15.0\"", "\"15.2\"", "\"15.3\""]):
+        if hdr['GeneralInfo']['NIRStar'] not in ['"15.0"', '"15.2"', '"15.3"']:
             raise RuntimeError('MNE does not support this NIRStar version'
                                ' (%s)' % (hdr['GeneralInfo']['NIRStar'],))
+        if hdr['GeneralInfo']['NIRStar'] in ['"15.3"']:
+            warn("Full MNE support for NIRStar version 15.3 currently under development.")        
+
         if "NIRScout" not in hdr['GeneralInfo']['Device']:
             warn("Only import of data from NIRScout devices have been "
                  "thoroughly tested. You are using a %s device. " %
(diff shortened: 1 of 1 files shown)


mne/io/nirx/nirx.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, fname, preload=False, verbose=None):
 
         # Check that the file format version is supported
         if not any(item == hdr['GeneralInfo']['NIRStar'] for item in
-                   ["\"15.0\"", "\"15.2\""]):
+                   ["\"15.0\"", "\"15.2\"", "\"15.3\""]):
```
cbrnr, line 111: We could make this more readable by writing ` ['"15.0"', '"15.2"', '"15.3"']` (enclose strings with single quotes, then you don't have to escape double quotes).

cbrnr: And wouldn't it be sufficient to write:

`if hdr['GeneralInfo']['NIRStar'] not in ['"15.0"', '"15.2"', '"15.3"']:`

rob-luke:
Thanks for opening this PR @DylanMannKrzisnik,

> I've contacted the NIRx support team and there are no changes to the file format which are relevant to our inquiry.

In my experience we need to verify this ourselves as the support team is entirely focused on the big two nirs software.

Additionally nirstar 15.3 was released with a new version of nirsite (the software for creating montage layouts). Are you using the new version 2020.07? According to the release notes this new version modified the available head models and coordinate layouts. So I think we need to validate the locations are still be read correctly in to MNE through the combined nirsite->nirstar procedure.

Could you provide a short test recording and also make sure to document the entire procedure as in https://github.com/mne-tools/mne-testing-data/pull/51 ? Given the recent changes in nirsite please make sure to take a screen shot (with exact coord visible like above) of the montage you design so we can validate the positions are read correctly in to MNE. If possible please grab all the same screenshots as in the example pr above. You can use any montage layout you want. Please make the recording just a few seconds and include a trigger or two.

I would take the test recording myself but I'm in lockdown for the next five weeks.

DylanMannKrzisnik:
Thank you @cbrnr and @rob-luke for the pointers,

I'm also not going back to the lab for the weeks to come. The data recordings we've acquired so far are a few minutes long, and from what I'm getting those would be too long. Maybe there's a way to truncate the recording durations to a few seconds.

I don't recall recently updating nirsite, the montage layouts we have been using were thus probably generated using previous versions of the software. I'd have to confirm whenever I gain access to our lab.

In any case I'll make sure to follow guidelines from [https://github.com/mne-tools/mne-testing-data/pull/51](https://github.com/mne-tools/mne-testing-data/pull/51) when time comes.

larsoner:
If everyone is months away from having suitable test data, one way forward is to `warn` for now if 15.3 is used saying support is currently experimental

DylanMannKrzisnik:
Should be ready to merge.

rob-luke:
I have a few other labs that can take a test recording for us then. I'd prefer to hold off a few days till they get the data to us, but up to you @larsoner

larsoner:
If it's a days - week(s) timeline then I'm fine with waiting. We release 0.21 in about a month, and it would be good to have at least "experimental" (read: working but not verifiable by proper test files) 15.3 support

larsoner:
... and if it's week(s) then merge this, but if it's days then waiting is not too painful. @DylanMannKrzisnik you okay with waiting for a bit?

rob-luke:
Ok got a response, It will be sent to me today.

mne/io/nirx/nirx.py
```diff
@@ -107,10 +107,12 @@ def __init__(self, fname, preload=False, verbose=None):
         hdr.read_string(hdr_str)
 
         # Check that the file format version is supported
-        if not any(item == hdr['GeneralInfo']['NIRStar'] for item in
-                   ["\"15.0\"", "\"15.2\""]):
+        if hdr['GeneralInfo']['NIRStar'] not in ['"15.0"', '"15.2"', '"15.3"']:
             raise RuntimeError('MNE does not support this NIRStar version'
                                ' (%s)' % (hdr['GeneralInfo']['NIRStar'],))
+        if hdr['GeneralInfo']['NIRStar'] in ['"15.3"']:
+            warn("Full MNE support for NIRStar version 15.3 currently under development.")
```
cbrnr, line 114: Two PEP8 warnings here that you need to fix:
```
./mne/io/nirx/nirx.py:114:80: E501 line too long (90 > 79 characters)
./mne/io/nirx/nirx.py:114:91: W291 trailing whitespace
```

cbrnr:
Superseded by #8122.

PR #8122, opened 2020-08-17
MRG: Use new nirscout test data location

rob-luke:
When I initially added the nirx test files I didn't know that the vendor (NIRX) has a different file format for each device (NIRScout, NIRSport1, NIRSport2). We will need test files for each device. I have moved the nirscout files to their own sub directory in https://github.com/mne-tools/mne-testing-data/pull/71. This PR points the tests to the new location

And now I have added tests for the new test data too.


larsoner:
To save some scrolling:
```
mne/viz/tests/test_3d.py:340: in test_plot_alignment
    info = read_raw_nirx(nirx_fname).info
mne/io/nirx/nirx.py:43: in read_raw_nirx
    return RawNIRX(fname, preload, verbose)
<decorator-gen-208>:21: in __init__
    ???
mne/io/nirx/nirx.py:76: in __init__
    raise RuntimeError('The path you specified does not exist.')
E   RuntimeError: The path you specified does not exist.
```

larsoner:
(Also this should be a FileNotFoundError probably)

rob-luke:
Modifies test to point to new data locations. Adds test for new 15.3 test recording. Supersedes #8121

agramfort:
ping when good to merge

>

rob-luke:
@larsoner and @agramfort can you please review. I think this is ready to merge (if last test comes back green).

One sub test of Travis CI failed due to unknown error so I have restarted it.

cbrnr:
Since you've added a test for 15.3, can we remove the warning in #8121 then?

rob-luke:
> Since you've added a test for 15.3, can we remove the warning in #8121 then?

Yes I think this supersedes 8121. If you're happy to merge this then you could probably close that. Or merge that then this.

PR #8125, opened 2020-08-17
MRG: Reset the viz settings in _TimeViewer

GuillaumeFavelier:
This PR reworks the interactor in `_Brain` when using `flatmaps` and adds a button to reset the initial time index and camera settings "Reset".

- [x] Use `vtkInteractorStyleRubberBand2D` with parallel projection
- [x] Find a suitable icon
- [x] Build the qt icon resources
- [x] Fix conflicts
- [x] `try/except` or `hasattr` for "RubberBand2D"
- [x] Check the MNE logo

It's an item of #7162
Related to https://github.com/mne-tools/mne-python/pull/8082#issuecomment-672948671


larsoner:
Could we use a 2D interactor instead like

https://vtk.org/doc/nightly/html/classvtkInteractorStyleRubberBand2D.html

GuillaumeFavelier:
> Could we use a 2D interactor

Yes indeed! In my experiments it works assuming that the camera uses parallel projection :)

GuillaumeFavelier:
These are the results for ["restore" on material.io](https://material.io/resources/icons/?search=restore&icon=restore&style=outline):

1 | 2 | 3 | 4 |5
--|---|---|----|---
image | image | image | image | image

**4** is already used for "Restore Scale". I put it in case we would swap for something else.

Alternatively:

cancel | flip | repeat
---------|-----|-----------
image | image | image

I searched "reset" and "init" without success.

GuillaumeFavelier:
I prefer **1** personally.

Any opinion @agramfort , @larsoner, @hoechenberger, @drammock ?

hoechenberger:
I think **1** could be a nice choice :)

agramfort:
+1 for 1

GuillaumeFavelier:
I'll go ahead with **1** then

mne/viz/backends/_pyvista.py
```diff
@@ -230,7 +230,15 @@ def scene(self):
         return self.figure
 
     def set_interaction(self, interaction):
-        getattr(self.plotter, f'enable_{interaction}_style')()
+        if interaction == "rubber_band":
+            for renderer in self.plotter.renderers:
+                renderer.enable_parallel_projection()
+            style = vtk.vtkInteractorStyleRubberBand2D()
```
larsoner, line 236: Can you open a PR to pyvista to add this as a mode? Then we can `try/except` or `hasattr` here

larsoner: (For example, this mode will suffer from the "click-and-drag-not-captured" bug, because the fix for this is in PyVista subclassing the VTK interactor style and dealing with button presses.)

GuillaumeFavelier: > Can you open a PR to pyvista to add this as a mode? 
Sure, will do.

I use middle-click to drag with this one. I was not aware of this bug.

larsoner: https://github.com/pyvista/pyvista/issues/116

larsoner: (It's possible we also need to capture middle clicks, in which case hopefully from https://github.com/pyvista/pyvista/pull/854 it's clear what you'd need to do to capture those, too)

GuillaumeFavelier:
@larsoner I opened https://github.com/pyvista/pyvista/pull/872

Also, could you check the that mne logo is exactly like you want? I had the following message while building it:

```
findfont: Font family ['sans-serif'] not found. Falling back to DejaVu Sans.
```

Maybe that was not a good idea but I tried [multiple packages for my linux distribution](https://wiki.archlinux.org/index.php/fonts#Sans-serif) without any success.

larsoner:
> Also, could you check the that mne logo is exactly like you want? I had the following message while building it:

The MNE logo in the resources is a PNG. I'm not sure why this would happen.

But FWIW you can get the [Primetime font listed in the logo generation script](https://github.com/mne-tools/mne-python/blob/master/logo/generate_mne_logos.py#L30) for free [here](https://www.dafont.com/primetime.font)

GuillaumeFavelier:
> The MNE logo in the resources is a PNG. I'm not sure why this would happen.

The logo is not tracked in `mne/icons` like the other icons and I built the Qt resources to integrate the new "Reset" icon. To fix the conflicts, I had to regenerate the resources. That's why I wanted to check that everything is okay.

larsoner:
But you shouldn't have gotten any font warning from running the icon resource generation from the mne logo because it's not a vector format, but rather bitmap. Did you also need to run something from logo/? Or did you just run the icons/ script? Can you got bisect to verify it's coming from the logo commit?

larsoner:
On a system without the Primetime font, I can reproduce with:
```
$ cd mne/logo
$ python generate_mne_logos.py 
findfont: Font family ['sans-serif'] not found. Falling back to DejaVu Sans.
```
But you should not have needed to call `generate_mne_logos.py` unless you were changing the logo. You should just be able to do:
```
$ cd ../mne/icons
$ python resources.py
```
which does not emit any warning.

... but then I realized that the logo actually was not saved as part of the git repo! So now I see why you had to do all of this. I'll push a commit that adds the PNG so that people don't have to do it in the future, and also regenerate the resources.

agramfort (approved):

larsoner:
Thanks @GuillaumeFavelier

PR #8134, opened 2020-08-20
MRG, BUG: Fix long name parsing

larsoner:
Using volumetric label extraction with HCPMMP1_combined (following https://github.com/mne-tools/mne-python/pull/7639#issuecomment-625907891) we run into a problem with `_get_lut` not reading long names. This fixes it.

I'm not 100% sure if FreeSurfer officially supports such long names, but at least `freeview` displays the names correctly when viewed with:
```
$ freeview -v HCPMMP1_combined+aseg.mgz:colormap=lut:lut=HCPMMP1_combinedColorLUT.txt
```
Within-release bugfix of custom LUT reading, so no need for `latest.inc` update.


agramfort:
thx @larsoner

PR #8135, opened 2020-08-20
MRG, ENH: Speed up vol label extraction

larsoner:
When using `extract_label_time_course` with volumes and `mri_resolution=True` (default) and `mode='mean'` (default), for large volumetric regions like `Left-Cerebral-White-Matter` that are part of the LUT you can end up exploding data into a high-res MRI space. If you have 10000 time points, the CSR dot product will be a crazy shape like `(100000, 10000)` and kill memory; and even if it succeeds, it then takes the mean across the first dimension (high-res voxels).

There is a nice shortcut by linearity where the upsampling matrix itself can just be averaged across vertices head of time, so that the CSR dot result is just shape `(1, 10000)`, and the downstream code then takes `np.mean(..., axis=0)` of this as a no-op.

I also made a change from `*` to `@` in the code to make it clear that it's actually a dot product. The `*` is always with a sparse matrix, so it was always matrix product -- the code change is just to make it explicit.

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -2954,12 +2954,16 @@ def _prepare_label_extraction(stc, labels, src, mode, allow_empty, use_sparse):
     for li, label in enumerate(labels):
         if use_sparse:
             assert isinstance(label, dict)
+            vertidx = label['csr']
             # This can happen if some labels aren't present in the space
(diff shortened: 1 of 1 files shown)


agramfort (approved):

agramfort:
nice @larsoner !

PR #8137, opened 2020-08-20
MRG, ENH: Add exp var to mxne

larsoner:
Adds explained variance logging to `mixed_norm` like we have for `apply_inverse`

Currently there is some bug where things do not match properly, hopefully will fix soon

Todo:

- [x] Get Alex's approval
- [x] Add to rap_music
- [x] Add exp var logging to gamma_map and rap_music

### DIFF
--- a/mne/beamformer/_rap_music.py
+++ b/mne/beamformer/_rap_music.py
@@ -8,7 +8,10 @@
 import numpy as np
 from scipy import linalg
 
+from ..forward import is_fixed_orient, convert_forward_solution
 from ..io.pick import pick_channels_evoked, pick_info, pick_channels_forward
+from ..inverse_sparse.mxne_inverse import _make_dipoles_sparse
+from ..minimum_norm.inverse import _log_exp_var
 from ..utils import logger, verbose, _check_info_inv
 from ..dipole import Dipole
 from ._compute_beamformer import _prepare_beamformer_input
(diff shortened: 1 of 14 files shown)


larsoner:
Okay finally tracked down the problem with gamma_map -- I think our outputs have been wrong in the `xyz_same_gamma` case for ~7 years due to an indexing bug. Not 100% sure because it's actually the first time I've looked at and worked with the code, but based on the tests that I've added that fail on `master` and pass here, and from looking at the logic, I think it's the case.

larsoner:
Azure Notebook failure is spurious and can be ignored

larsoner:
Okay, pushed a commit to add it to `_rap_music`, and in the process simplified the code a bit by ensuring that the forward is not in surface orientation (makes life much easier). Also added exp var logging to rap-music and gamma map.

agramfort:
it works with MxNE but with RAP-MUSIC running plot_rap_music.py the 2 dipoles I obtain have the same gof which is not expected.

looking at :

plt.figure(); plt.plot(np.array([d.gof for d in dipoles]).T)

compared to 

plot_dipole_amplitudes(dipoles)

is interesting as you see that deep dipoles are much less visible when you plot GOF than amplitudes.

besides the RAP-MUSIC probable bug LGTM !

thx @larsoner !

larsoner:
> the 2 dipoles I obtain have the same gof which is not expected.

I don't see this:
```
plt.figure()
fig, axes = plt.subplots(2)
axes[0].plot(np.array([d.amplitude for d in dipoles]).T)
axes[1].plot(np.array([d.gof for d in dipoles]).T)
```
Figure_2

Different amplitudes and different GOFs.

Let's do a quick sanity check to see if they're reasonable by doing `mne.fit_dipole` with their pos and ori (first should be equivalent; second should be similar but not equivalent because it hasn't had the field subtracted):
```
>>> bem = mne.read_bem_solution(data_path + '/subjects/sample/bem/sample-5120-5120-5120-bem-sol.fif')
>>> for dipole in dipoles:
...     dip, _ = mne.fit_dipole(evoked, noise_cov, bem, trans, pos=dipole.pos[0], ori=dipole.ori[0])
...     dip.plot()
```
Screen Shot 2020-09-20 at 7 01 20 AM

Screen Shot 2020-09-20 at 7 01 23 AM

Looks like they match to me.

mne/beamformer/_rap_music.py
```diff
@@ -69,41 +78,31 @@ def _apply_rap_music(data, info, times, forward, noise_cov, n_dipoles=2,
...
     for k in range(n_dipoles):
         subcorr_max = -1.
-        for i_source in range(G.shape[1] // n_orient):
-            idx_k = slice(n_orient * i_source, n_orient * (i_source + 1))
-            Gk = G_proj[:, idx_k]
-            if n_orient == 3:
-                Gk = np.dot(Gk, forward['source_nn'][idx_k])
-
+        source_idx, source_ori, source_pos = 0, [0, 0, 0], [0, 0, 0]
+        for i_source in range(G.shape[1]):
```
agramfort, line 85:
```suggestion
        for i_source in range(G_proj.shape[1]):
```
?

agramfort:
when I wrote `besides the RAP-MUSIC probable bug LGTM !` it was already done for RAP-MUSIC

agramfort:
thx @larsoner all good !

PR #8155, opened 2020-08-25
Remove meg=True from _pick_data_or_ica

### LINKED ISSUE #7836: Possible bug in _pick_data_or_ica
While setting `meg=True` in calls to `mne.pick_types` in calls that didn't explicitly specify this argument (#7823), I noticed that the internal function `_pick_data_or_ica` probably doesn't do the right thing.

Previously, the function picked:

```python
picks = pick_types(info, exclude=exclude, misc=True)
```

Since `meg=True` by default, I changed this to:

```python
picks = pick_types(info, meg=True, exclude=exclude, misc=True)
```

But this function is supposed to pick ICA channels (I think), and these should be of type `'misc'` (another assumption I'm not sure of). But if this is true, `meg` should be set to `False`.

cbrnr:
Fixes #7836.

### DIFF
--- a/mne/io/pick.py
+++ b/mne/io/pick.py
@@ -960,8 +960,7 @@ def _pick_aux_channels(info, exclude='bads'):
 def _pick_data_or_ica(info, exclude=()):
     """Pick only data or ICA channels."""
     if any(ch_name.startswith('ICA') for ch_name in info['ch_names']):
-        # FIXME: is meg=True really correct here?
-        picks = pick_types(info, exclude=exclude, misc=True, meg=True)
+        picks = pick_types(info, exclude=exclude, misc=True)
     else:
         picks = _pick_data_channels(info, exclude=exclude, with_ref_meg=True)
     return picks


agramfort (commented):
can you update/add a tiny test?

cbrnr:
That's a great question. I don't know what this function is supposed to do, and I don't think that our tests cover this function. Let's wait and see if the change broke anything, then I can think about a test.

agramfort:
ok

>

cbrnr:
The macOS job on Travis keeps timing out (>50 minutes), should we do anything about it? I know that @larsoner tried to optimize in the past, but apparently this job still takes too long.

agramfort:
is it only in this PR?

>

cbrnr:
> is it only in this PR?

I've seen it several times with differents PRs.

cbrnr:
After 50 minutes (the timeout), tests are only at 46%, so there must be something fundamentally wrong with the macOS machines Travis is using, because we can't speed up by 2x (assuming that test durations are distributed uniformly across our test suite). I'd move to Azure and abandon Travis completely unless there's a specific reason.

larsoner:
Travis OSX has always been slow, and Travis is usually our bottleneck. Moving to Azure for macOS would be great. PR welcome if you're up for it, otherwise I can take a crack at it next week

cbrnr:
@larsoner I created #8159 to remind ourselves of that.

cbrnr:
@larsoner do you have an idea for a test for the change in this PR? I could add something that works now and fails with the previous behavior, but I'm not sure if this is really what we want.

larsoner:
In principle I think any given object should have ICA channels or data channels but not both, so I actually can't think of a test case that should actually matter to users. In principle you could maybe cobble one together with `add_channels` but I wouldn't expect it to work properly.

cbrnr:
I agree, since this is a private function anyway it probably never matters to users. Your call @agramfort, let us know if you want a test, or otherwise you can go ahead and merge.

agramfort (approved):
ok let's merge and move on.

agramfort:
thx @cbrnr

Issue #8147, opened 2020-08-23
ICA.plot_overlay() and baseline correction


hoechenberger:
#### Describe the bug
@agramfort pointed out that in ICA overlay plots (i.e., visualization of Evoked before and after artifact removal via ICA), the displayed cleaned Evoked isn't baseline-corrected, even if the input **was** baseline-corrected.

<img width="614" alt="Screenshot 2020-08-23 at 17 54 34" src="https://user-images.githubusercontent.com/2046265/90982926-d0c3ac00-e56a-11ea-9b21-e6cf84152fd7.png">

I've tracked this down to an issue in `mne.viz.ica.plot_ica_overlay()`

https://github.com/mne-tools/mne-python/blob/4c44bd6ae90dd4d5c6c2e925d93f1b2fbf7bda6a/mne/viz/ica.py#L791-L793

The root of our problem is that `evoked_cln` is not baseline-corrected before it is being submitted to `_plot_ica_overlay_evoked()`. This in itself could be tolerated; however, since our input data (called `inst` here) **was already** baseline-corrected (by us, before invoking `plot_overlay()`), we end up with an overlay of a baseline-corrected and a non-baseline-corrected Evoked.

#### Expected results

By applying baseline correction to `evoked_cln` before plotting, I get the expected result:
```python
     evoked_cln = ica.apply(inst.copy(), exclude=exclude) 
     evoked_cln.apply_baseline()
     fig = _plot_ica_overlay_evoked(evoked=inst, evoked_cln=evoked_cln, 
                                    title=title, show=show) 
```

<img width="614" alt="Screenshot 2020-08-23 at 17 55 05" src="https://user-images.githubusercontent.com/2046265/90983029-7ecf5600-e56b-11ea-8c7f-4ff8f58794d5.png">

So now I thought: neat, that'll be an easy fix! Just check if the passed Evoked instance has been baseline-corrected; and if yes, also baseline-correct the cleaned Evoked.

But there is no `Evoked.baseline` attribute as there is for Epochs. Therefore it seems to be impossible to find out whether an Evoked has been baseline-corrected or not.

#### Proposed fix

I suggest we
- add a `.baseline` attribute to `mne.Evoked`, which is inherited from the Epochs used for averaging
- add a check to `mne.viz.ica.plot_ica_overlay()`:
    - if `inst.baseline` is set, apply the same baseline correction to `evoked_cln`,
    - otherwise, do not apply baseline correction to `evoked_cln`

[...]

PR #8156, opened 2020-08-25
MRG: Add .baseline attribute to mne.Evoked

hoechenberger:
Add `Evoked.baseline`. x-ref #8147

This is in an early proof-of-concept stage: I haven't even run / adjusted the tests yet…

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -193,7 +193,7 @@ Changelog
 
 - Use PyVista as the default backend for 3D visualization instead of Mayavi by `Guillaume Favelier`_
 
-- `~mne.Evoked` has gained a new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
+- `~mne.Evoked` has gained new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
 
 Bug
 ~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -218,6 +218,8 @@ Bug
 
 - Fix bug with `mne.SourceSpaces.export_volume` with ``use_lut=False`` where no values were written by `Eric Larson`_
 
+- Fix bug with logging in :meth:`mne.io.Raw.set_eeg_reference` and related functions by `Simeon Wong`_
+
 - Fix bug with :func:`mne.preprocessing.annotate_movement` where bad data segments, specified in ``raw.annotations``, would be handled incorrectly by `Luke Bloy`_
 
 - Fix bug with :func:`mne.compute_source_morph` when more than one volume source space was present (e.g., when using labels) where only the first label would be interpolated when ``mri_resolution=True`` by `Eric Larson`_
@@ -356,6 +358,8 @@ Bug
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
 - Fix bug in :func:`mne.set_eeg_reference` and related functions to set ``info['custom_ref_applied']`` to ``True`` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
+
+- Fix bug with :func:`mne.chpi.compute_chpi_amplitudes` and :func:`mne.chpi.filter_chpi` to optimize time window length by `Steven Bierer`_
 
 - Fix bug in `~mne.Epochs.apply_baseline` that would – in some situations – lead to an update of the ``baseline`` attribute, even if the data wasn't actually rescaled, by `Richard Höchenberger`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -357,6 +357,8 @@ Bug
 
 - Fix bug in :func:`mne.set_eeg_reference` and related functions to set ``info['custom_ref_applied']`` to ``True`` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
 
+- Fix bug in `~mne.Epochs.apply_baseline` that would – in some situations – lead to an update of the ``baseline`` attribute, even if the data wasn't actually rescaled, by `Richard Höchenberger`_
+
 API
 ~~~
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -195,7 +195,7 @@ Changelog
 
 - `~mne.Evoked` has gained new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
 
-- Trying to remove a baseline from already baseline-corrected data by passing ``baseline=None`` to `~mne.Epochs.apply_baseline` will now raise an exception, by `Richard Höchenberger`_
+- Trying to remove a baseline from already baseline-corrected data by passing ``baseline=None`` to `~mne.Epochs.apply_baseline` or `~mne.Evoked.apply_baseline` will now raise an exception, by `Richard Höchenberger`_
 
 Bug
 ~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -195,6 +195,8 @@ Changelog
 
 - `~mne.Evoked` has gained new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
 
+- Trying to remove a baseline from already baseline-corrected data by passing ``baseline=None`` to `~mne.Epochs.apply_baseline` will now raise an exception, by `Richard Höchenberger`_
+
 Bug
 ~~~
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -391,6 +391,8 @@ API
 
 - The ``threshold`` argument in :meth:`mne.preprocessing.ICA.find_bads_ecg` defaults to ``None`` in version 0.21 but will change to ``'auto'`` in 0.22 by `Yu-Han Luo`_
 
+- The function ``mne.preprocessing.mark_flat`` has been deprecated in favor of :func:`mne.preprocessing.annotate_flat` by `Eric Larson`_
+
 - The default argument ``meg=True`` in :func:`mne.pick_types` will change to ``meg=False`` in version 0.22 by `Clemens Brunner`_
 
 - Added :meth:`mne.io.Raw.get_montage`, which obtains the montage that an instance has, by `Adam Li`_
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -193,6 +193,8 @@ Changelog
 
 - Use PyVista as the default backend for 3D visualization instead of Mayavi by `Guillaume Favelier`_
 
+- `~mne.Evoked` has gained a new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
+
 Bug
 ~~~
 
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -204,6 +204,16 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         """Channel names."""
         return self.info['ch_names']
 
+    @property
+    def tmin(self):
+        """First time point."""
+        return self.times[0]
+
+    @property
+    def tmax(self):
+        """Last time point."""
+        return self.times[-1]
+
     @fill_doc
     def crop(self, tmin=None, tmax=None, include_tmax=True):
         """Crop data to a given time interval.
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -39,6 +39,7 @@ from .io.write import (start_file, start_block, end_file, end_block,
                        write_id, write_float, write_complex_float_matrix)
 from .io.base import TimeMixin, _check_maxshield
 
+
 _aspect_dict = {
     'average': FIFF.FIFFV_ASPECT_AVERAGE,
     'standard_error': FIFF.FIFFV_ASPECT_STD_ERR,
@@ -257,7 +258,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
 
         try:
             _check_baseline(self.baseline, tmin, tmax, self.info['sfreq'])
-        except ValueError:  # It no longer applies, wipe it out
+        except ValueError:  # It no longer applies, wipe it out.
             warn('Cropping removes baseline period, setting baseline=None')
             self.baseline = None
 
@@ -1205,6 +1206,8 @@ def write_evokeds(fname, evoked):
 def _write_evokeds(fname, evoked, check=True):
     """Write evoked data."""
     from .epochs import _compare_epochs_infos
+    from .dipole import DipoleFixed  # avoid circular import
+
     if check:
         check_fname(fname, 'evoked', ('-ave.fif', '-ave.fif.gz',
                                       '_ave.fif', '_ave.fif.gz'))
@@ -1242,7 +1245,7 @@ def _write_evokeds(fname, evoked, check=True):
             write_int(fid, FIFF.FIFF_LAST_SAMPLE, e.last)
 
             # Baseline
-            if e.baseline is not None:
+            if not isinstance(e, DipoleFixed) and e.baseline is not None:
                 bmin, bmax = e.baseline
                 bmin = e.times[0] if bmin is None else bmin
                 bmax = e.times[-1] if bmax is None else bmax
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -1050,7 +1050,7 @@ def read_evokeds(fname, condition=None, baseline=None, kind='average',
             # Don't touch an existing baseline.
             bmin, bmax = evoked.baseline
             bmin = None if bmin is None else f'{bmin:.3f}'
-            bmax = None if bmin is None else f'{bmax:.3f}'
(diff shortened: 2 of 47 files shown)


hoechenberger:
Maybe @larsoner wants to share some early comments: "That's good, keep on going" / "💩, you got it all wrong!"

cbrnr:
Can you make the plot as @agramfort suggested [here](https://github.com/mne-tools/mne-python/issues/8147#issuecomment-679081438)? Maybe this isn't necessary after all.

hoechenberger:
@cbrnr 
> Maybe this isn't necessary after all.

Sorry for misleadingly referencing  #8147 here: I just meant to imply that while working on  #8147, the idea that Evoked.baseline doesn't exist got sparked. I didn't mean to imply that this PR was a fix to #8147

mne/evoked.py
```diff
@@ -911,6 +928,10 @@ def read_evokeds(fname, condition=None, baseline=None, kind='average',
         list.
     %(baseline)s
         Defaults to ``None``, i.e. no baseline correction.
+
+        .. versionchanged:: 0.21
+           The ``baseline`` argument has been deprecated and will be removed
+           in MNE-Python 0.22. Evoked baselines are now read from disk.
```
larsoner, line 934:
```suggestion
           in MNE-Python 0.22. Evoked baselines are now read from disk.
           Use ``evoked.apply_baseline(baseline)`` to apply a custom
           baseline.
```

larsoner:
@hoechenberger 0.21 is getting closer. Need help pushing this one over the line? Looks like it's close already

hoechenberger:
@larsoner I will pick up working on this tomorrow, so I'm confident it can be included in the next release

hoechenberger:
Currently finishing this up, got I/O working.

But now I'm wondering:
For whatever reason, I (we?) decided to deprecate the `baseline` param of `read_evokeds()`. I'm not so sure anymore where this is a wise idea… imagine you have several Evokeds stored in a single file want to read them all and apply a custom baseline, you'd then have to do:

```python
evokeds = read_evokeds(...)
evokeds = [e.apply_baseline(...) for e in evokeds]
```

instead of just

```python
evokeds = read_evokeds(..., baseline=...)
```

But **then again**, one typically baseline-corrects on the Epochs level already, right? And the new `Evoked.baseline` attribute would tell one whether and which baseline correction has been applied.

Still, I wanted to ask again: do you think it's okay to deprecate the `baseline` argument of `read_evokeds()`?

larsoner:
I'm okay with reverting the deprecation. It's pretty low-burden to keep it around, and the code is super old so the deprecation pain is pretty high for little gain.

hoechenberger:
> I'm okay with reverting the deprecation. It's pretty low-burden to keep it around, and the code is super old so the deprecation pain is pretty high for little gain.

And what to do in cases where a baseline is stored on disk, and the user passes a `baseline` param? The param takes precedence, and we issue a warning?

Would you see any **advantages** in removing the param?

larsoner:
> And what to do in cases where a baseline is stored on disk, and the user passes a baseline param? The param takes precedence, and we issue a warning?

I don't see a need to issue a warning, we allow re-baselining. We can just say in the docstring that passing a baseline param is equivalent to doing `evoked.apply_baseline(baseline)` afterward.

> Would you see any advantages in removing the param?

Only that ideally there should be one clear way to do things, and all of our instances having an `apply_baseline` could be that way. But I'm not convinced pushing in this direction is worth it in this case, all things considered.

larsoner:
Failures are real:
```
FAILED mne/inverse_sparse/tests/test_gamma_map.py::test_gamma_map - RuntimeWa...
FAILED mne/inverse_sparse/tests/test_gamma_map.py::test_gamma_map_vol_sphere
FAILED mne/inverse_sparse/tests/test_mxne_inverse.py::test_mxne_inverse_standard
FAILED mne/inverse_sparse/tests/test_mxne_inverse.py::test_mxne_vol_sphere - ...
FAILED mne/utils/tests/test_logging.py::test_logging_options - AssertionError...
```

hoechenberger:
Thanks, @larsoner! I hope I've caught everything now. I've also amended the docstrings and tutorials a bit

hoechenberger:
Seems I've broken I/O again. Ugh…

hoechenberger:
@larsoner I wonder if my fix from 94ee33f should also be implemented for `Epochs.apply_baseline`?

larsoner:
Probably. I'd try changing it and locally run `mne/tests/test_epochs.py` to see if anything breaks, if not, then push

hoechenberger:
Any idea how to fix the documentation build error:

```
/home/circleci/project/mne/evoked.py:docstring of mne.read_evokeds:5: WARNING: py:obj reference target not found: mne.Evoked.baseline
```

Docstring looks like this, I thought this was valid?

```
    """Read evoked dataset(s).

    .. versionchanged:: 0.21
        If the read `~mne.Evoked` objects had been baseline-corrected before
        saving, this will be reflected in their `~mne.Evoked.baseline`
        attribute after reading.
```
@drammock Can you help me? :)

hoechenberger:
> Probably. I'd try changing it and locally run `mne/tests/test_epochs.py` to see if anything breaks, if not, then push

Added this in eb8f927, let's see if it passes all tests.

larsoner:
Links to properties don't work, just render as code with double-backticks

You need a rebase or merge as there is a conflict

doc/changes/latest.inc
```diff
@@ -193,6 +193,8 @@ Changelog
 
 - Use PyVista as the default backend for 3D visualization instead of Mayavi by `Guillaume Favelier`_
 
+- `~mne.Evoked` has gained a new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
```
hoechenberger, line 196:
```suggestion
- `~mne.Evoked` has gained new ``baseline``, ``tmin``, and ``tmax`` attributes for more consistency with `~mne.Epochs` by `Richard Höchenberger`_
```

hoechenberger:
Doc artifacts:

[Evoked.save](https://22264-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Evoked.html#mne.Evoked.save)
[write_evokeds](https://22264-1301584-gh.circle-artifacts.com/0/dev/generated/mne.write_evokeds.html#mne.write_evokeds)
[read_evokeds](https://22264-1301584-gh.circle-artifacts.com/0/dev/generated/mne.read_evokeds.html#mne.read_evokeds)

mne/tests/test_epochs.py
```diff
@@ -745,6 +745,18 @@ def test_epochs_baseline(preload):
...
+    # rescalrd.
+    baseline = (None, None)
+    epochs = mne.Epochs(raw, events, None, 0, 1e-3, baseline=baseline,
+                        preload=preload)
+    epochs.apply_baseline(None)
+
+    if preload:
+        assert epochs.baseline == baseline
+    else:
+        assert epochs.baseline is None
```
larsoner, line 758: I don't understand why the data rescaling depends on whether or not `preload` was used. It seems more like in the non-preload case that there should be some "pending baseline operation" that gets applied after data are loaded from disk so that these actually match.

In other words, after `epochs.apply_baseline(...)`, both the `epochs.baseline` attribute and the result of `epochs.get_data()` should be the same when `preload=True` as when `preload=False`.

mne/tests/test_epochs.py
```diff
@@ -1220,21 +1234,27 @@ def test_evoked_io_from_epochs(tmpdir):
     assert_allclose(evoked.data, evoked2.data, rtol=1e-4, atol=1e-20)
     assert_allclose(evoked.times, evoked2.times, rtol=1e-4,
                     atol=1 / evoked.info['sfreq'])
+    # assert evoked2.baseline == baseline XXX find I/O roundtrip
```
larsoner, line 1237: Cruft? Should it match now?

mne/tests/test_epochs.py
hoechenberger: > I don't understand why the data rescaling depends on whether or not `preload` was used. It seems more like in the non-preload case that there should be some "pending baseline operation" that gets applied after data are loaded from disk so that these actually match.
> 
> In other words, after `epochs.apply_baseline(...)`, both the `epochs.baseline` attribute and the result of `epochs.get_data()` should be the same when `preload=True` as when `preload=False`.

Yeah that's what I thought at first too, until I figured that wasn't the case :) I used the following MWE do investigate the issue on `master`:

```python
# %%
import mne
import os
import numpy as np

preload = True
                                    'sample_audvis_raw.fif')
events = mne.find_events(raw, stim_channel='STI 014', verbose=False)

# %%
baseline = (None, 0)
baseline2 = None

# Apply a baseline
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7, baseline=baseline,
                    preload=preload, verbose=False)

# Remove the baseline
epochs2 = epochs.copy().apply_baseline(baseline2)

print(epochs.baseline)
print(epochs2.baseline)

# %%
if np.allclose(epochs.get_data(),
               epochs2.get_data()):
    print('Data is identical')
else:
    print('Data differs')
```
## `preload=True` on `master`
When we use `preload=True`, the following output it produced:

```
(None, 0)
None
Data is identical
```
i.e., although the values in `.baseline` differ, the data is the same. We can find out why this is the case by looking at `Epochs.apply_baseline`:
https://github.com/mne-tools/mne-python/blob/ca3e787ca40059ee72b93e517adbe7506688c2a3/mne/epochs.py#L661-L671
When we create `epochs` in our code:
```python
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7, baseline=baseline,
                    preload=True, verbose=False)
```
we enter the `if` branch, and the call to `rescale()` will actually rescale the data according to our baseline `(None, 0)`. After that, the `.baseline` attribute will be set.

Now when we apply the "baseline" `None` to `epochs2`:
 ```python
epochs2 = epochs.copy().apply_baseline(baseline2)
```
we will follow the same branching as in the former case, **but** `rescale` won't actually do anything, as when `baseline is None` it will leave the input data unmodified:
https://github.com/mne-tools/mne-python/blob/ca3e787ca40059ee72b93e517adbe7506688c2a3/mne/baseline.py#L62-L66
Then, like in the first case, the `.baseline` attribute will be set to the new value.

**This means that in this case, even though no rescaling took place, the `.baseline` attribute of the preloaded data was altered.**

## `preload=False` on `master`
Let's test with `preload=False`:
```
(None, 0)
None
Data differs
```
So the `.baseline` attribute was updated, just like in the previous example. But `get_data()` now yields different results. Why? Because if we don't preload the data, obviously a the to `apply_baseline()` won't perform any rescaling; instead, it **only** updates the `.baseline` attribute. **When** we finally load the data, this baseline operation is carried out: a baseline of `(None, 0)` is applied to `epochs`, and no baseline correction (`baseline=None`) is applied to `epochs2`.

## Summary for `master`
- When using `preload=True`, followed by `apply_baseline(None)`, we may end up with an `Evoked` that has `Evoked.baseline = None` set, even though a baseline had been applied earlier. **I consider this a bug.**
- When using `preload=False`, `apply_baseline()` may be used to toggle on / off baseline correction until we finally decide to call `get_data()`. **I consider this expected behavior.**


## Behavior in this PR branch

### `preload=True`
```
(None, 0)
(None, 0)
Data is identical
```
The rescaling no-op of
```python
epochs2 = epochs.copy().apply_baseline(baseline2)
```
correctly did **not** alter the `.baseline` attribute, as the preloaded data was not altered.

### `preload=False`
```
(None, 0)
None
Data differs
```
Behavior here is identical to `master`: As long as the data has not been preloaded, baseline correction can be toggled, and will be applied upon data retrieval via `get_data()`.

## Summary for this PR branch
- When using `preload=True`, followed by `apply_baseline(None)`, `Evoked.baseline` stays unchanged, which correctly reflects that the data has not been altered by the `apply_baseline()` call.  **I consider this expected behavior.**
- Same behavior as on `master`: When using `preload=False`, `apply_baseline()` may be used to toggle on / off baseline correction until we finally decide to call `get_data()`. **I consider this expected behavior.**

## Summary
I believe I'm doing the Right Thing™️ in this PR.

mne/tests/test_epochs.py
hoechenberger: Good catch, I had temporarily disabled that for testing. Took care of it in fbf112f!

mne/evoked.py
```diff
@@ -168,7 +173,12 @@ def apply_baseline(self, baseline=(None, 0), verbose=None):
 
         .. versionadded:: 0.13.0
         """
+        _check_baseline(baseline, self.times[0], self.times[-1],
+                        self.info['sfreq'])
+        # Rescale actually only happens if baseline is not None.
         self.data = rescale(self.data, self.times, baseline, copy=False)
+        if baseline is not None:
+            self.baseline = baseline
```
agramfort, line 181: can you check that one cannot pass baseline None if self.baseline is not None? basically once a baseline has been used you cannot go back

mne/evoked.py
```diff
@@ -220,11 +251,32 @@ def crop(self, tmin=None, tmax=None, include_tmax=True):
...
+
         mask = _time_mask(self.times, tmin, tmax, sfreq=self.info['sfreq'],
                           include_tmax=include_tmax)
         self.times = self.times[mask]
         self._update_first_last()
         self.data = self.data[:, mask]
+
+        try:
+            _check_baseline(self.baseline, tmin, tmax, self.info['sfreq'])
+        except ValueError:  # It no longer applies, wipe it out.
```
agramfort, line 276: can you try to avoid the try except block? Exception are error prone. It can hide the actual issue.

mne/evoked.py
```diff
@@ -748,6 +804,8 @@ def __init__(self, data, info, tmin=0., comment='', nave=1, kind='average',
             raise ValueError('unknown kind "%s", should be "average" or '
                              '"standard_error"' % (self.kind,))
         self._aspect_kind = _aspect_dict[self.kind]
+        self.baseline = baseline
+        self.apply_baseline(self.baseline, verbose=verbose)
```
agramfort, line 808: apply_baseline should update self already

agramfort (commented):
here is a first batch of comments

mne/evoked.py
hoechenberger: I just copied existing code from the `Epochs` codebase :) But can try to figure out something better.

mne/evoked.py
hoechenberger: That's effectively what's happening, actually. If `baseline=None`, `rescale` won't do anything. I also just copied the behavior from `Epochs` here. I could re-arrange the code such that things become clearer:

```python
        if baseline is not None:
            self.data = rescale(self.data, self.times, baseline, copy=False)
            self.baseline = baseline
```
WDYT?

mne/tests/test_epochs.py
hoechenberger: There's still a UX issue IMHO if one
- reads a baseline-corrected Evoked or Epochs **without preloading** from disk
- then invokes `evoked.apply_baseline(None)`
- only after that loads the data

this should **not** be possible, as we cannot remove an existing baseline correction, as @agramfort pointed out in https://github.com/mne-tools/mne-python/pull/8156#discussion_r487387001

I guess we'll have to re-think the logic of this.

hoechenberger: … for example, what if you

1. read **not-baseline-corrected** Evoked without preloading from disk
1. then invoke `evoked.apply_baseline((None, 0))`
1. then wishes to invoke evoked.apply_baseline(None)
1. and only after that loads the data

At step number 3, we wouldn't be able to tell whether the none-preloaded data with a `baseline != None` had been read from disk as a baseline-corrected dataset, or whether what happened was what I show in this example

mne/evoked.py
hoechenberger: I've addressed this in 8ee3bb7

mne/evoked.py
hoechenberger: Yes, but I think this additional line 807 makes it more explicit that we're storing the `baseline` parameter in `self.baseline`. If you insist, I can remove it; personally I find it nicer the way it is now.

hoechenberger: I've tried to improve things in def384d

mne/tests/test_epochs.py
hoechenberger: We could simply **drop support** for `apply_baseline(None)`, this would make these problems go away.
Opinions? @agramfort?

mne/evoked.py
hoechenberger: I've tried to improve things in def384d

mne/tests/test_epochs.py
larsoner: > … for example, what if you

If it simplifies things, I would raise an error on step (3). Even though in theory it's possible for us to do what the users are asking for, the fact that it depends on `preload` or not is not good. Basically we want to discourage people from doing `.apply_baseline((None, 0)).apply_baseline(None)`, even if in some circumstances it should technically be possible to do so.

> We could simply drop support

-1 on this, in pipelines it's nice to be able to set a var at the top and then just do `apply_baseline(baseline)` and this would require people to modify their code to say `if baseline is not None:` instead.

I would just always prevent going from a `.apply_baseline(non_None)`ed data to a `.apply_baseline(None)`ed data, then we have consistent behavior that we can tell people, and hopefully it simplifies the code end.

hoechenberger: > I would just always prevent going from a `.apply_baseline(non_None)`ed data to a `.apply_baseline(None)`ed data, then we have consistent behavior that we can tell people, and hopefully it simplifies the code end.

Sounds good, will implemenent

hoechenberger:
I believe I have addressed all comments. Documentation build will probably still fail, but code tests should come back positive.

agramfort (approved):

hoechenberger:
I'm still fighting with the docstrings, I simply cannot get all tests to pass locally 😭

hoechenberger:
This is Good Go Merge from my end. Travis finished successfully on the last commit before those docstring wars.

larsoner:
Thanks for working through this @hoechenberger !

PR #8169, opened 2020-08-27
FIX : allow Path in parse_folder

### LINKED ISSUE #8168: Report.parse_folder() does not support Path
#### Describe the bug
`Report.parse_folder()` does doesn't handle `pathlib.Path` input correctly, while other MNE functions do.

#### Steps to reproduce
```python
import pathlib
import mne

p = pathlib.Path('foo')
report = mne.Report()
report.parse_folder(p)
```

#### Expected results
```
Embedding : jquery.js
Embedding : jquery-ui.min.js
Embedding : bootstrap.min.js
Embedding : jquery-ui.min.css
Embedding : bootstrap.min.css
Iterating over 0 potential files (this may take some time)
```

#### Actual results
```python
Embedding : jquery.js
Embedding : jquery-ui.min.js
Embedding : bootstrap.min.js
Embedding : jquery-ui.min.css
Embedding : bootstrap.min.css
Untitled-1 in 
      5 p = pathlib.Path('foo')
      6 report = mne.Report()
----> 7 report.parse_folder(p)

 in parse_folder(self, data_path, pattern, n_jobs, mri_decim, sort_sections, on_error, image_format, render_bem, verbose)

~/Development/mne-python/mne/report.py in parse_folder(self, data_path, pattern, n_jobs, mri_decim, sort_sections, on_error, image_format, render_bem, verbose)
   1445 
   1446         if self.title is None:
-> 1447             self.title = 'MNE Report for ...%s' % self.data_path[-20:]
   1448 
   1449         if not isinstance(pattern, (list, tuple)):

TypeError: 'PosixPath' object is not subscriptable
```

agramfort:
closes #8168

### DIFF
--- a/mne/report.py
+++ b/mne/report.py
@@ -1407,7 +1407,7 @@ class Report(object):
 
         Parameters
         ----------
-        data_path : str
+        data_path : str | pathlib.Path
             Path to the folder containing data whose HTML report will be
             created.
         pattern : str | list of str
@@ -1436,6 +1436,7 @@ class Report(object):
             .. versionadded:: 0.16
         %(verbose_meth)s
         """
+        data_path = str(data_path)
         image_format = _check_image_format(self, image_format)
         _check_option('on_error', on_error, ['ignore', 'warn', 'raise'])
         self._sort = sort_sections
--- a/mne/report.py
+++ b/mne/report.py
@@ -25,7 +25,7 @@ from .io import read_raw_fif, read_info
 from .io.pick import _DATA_CH_TYPES_SPLIT
 from .source_space import _mri_orientation
 from .utils import (logger, verbose, get_subjects_dir, warn,
-                    fill_doc, _check_option)
+                    fill_doc, _check_option, _validate_type)
 from .viz import (plot_events, plot_alignment, plot_cov, plot_projs_topomap,
                   plot_compare_evokeds)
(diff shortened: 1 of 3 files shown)


mne/report.py
```diff
@@ -1407,7 +1407,7 @@ def parse_folder(self, data_path, pattern='*.fif', n_jobs=1, mri_decim=2,
 
         Parameters
         ----------
-        data_path : str
+        data_path : str | pathlib.Path
```
hoechenberger, line 1410:
```suggestion
        data_path : str
```
So it seems in all other places of MNE we **don't** mention that we support `Path` / path-like objects, just `str`. I believe this should be changed – but in another PR.

mne/report.py
```diff
@@ -1436,6 +1436,7 @@ class construction.
             .. versionadded:: 0.16
         %(verbose_meth)s
         """
+        data_path = str(data_path)
```
hoechenberger, line 1439:
```suggestion
        _validate_type(fname, 'path-like', 'fname')
        data_path = str(data_path)
```

hoechenberger (commented):
Two nitpicks. Will add a changelog entry.

hoechenberger (approved):

hoechenberger:
Thanks @agramfort!

Issue #8173, opened 2020-08-30
Reading in Persyst (.lay/.dat) files into Raw


adam2392:
#### Describe the problem
I have some files that were output from the Persyst system, which are text-readable .lay/.dat file combination. I would like to read them into MNE Python Raw.

This would be useful since I think it's a major "EEG" software processing system, which many clinical centers have. In addition, it is text-readable, so nothing hard about it. 

#### Describe your solution
A `read_raw_persyst` function that takes in a `lay_fname` and reads in data into a `Raw` object. The `.dat` filename is usually encoded in the `lay_fname`. 

The first solution won't be "perfect", but some more iterations from Persyst users might be warranted.

#### Describe possible alternatives
There is an initial start I found on Github: https://github.com/epiSurg/python-layread/blob/master/layread.py

but they do not read it into a `Raw` object, which I think would be more desirable for robust data flow.

#### Additional context
I have some example files added here.
[Archive.zip](https://github.com/mne-tools/mne-python/files/5146962/Archive.zip)

I can help implement it if perhaps @agramfort  or @larsoner can point me in the right direction? I'm assuming it's just that one function `read_raw_persyst`.

agramfort:
seems indeed easy to add.

you should add these testing files in the mne-testing-data repo and then
add the reader in PR to mne-python

you can look at brainvision or nicolet readers for inspiration

>

larsoner:
I would look at https://github.com/mne-tools/mne-python/pull/6674 to see the list of files you might need to change. If there are no new channel types, you probably only need to change half as many (or even fewer).

PR #8176, opened 2020-08-31
[MRG, IO] Persyst reader into Raw object

adam2392:
#### Reference issue
Closes: #8173 

Adds a Persyst reader for .lay + .dat file combinations and unit tests.

Note:
- birthdate and measurement date is not a strict adherence sometimes when output by Persyst. I am not sure why... but things like `16-Nov-199` obviously probably stands for nov 16th, 1992 based on the data, but not sure why it gets output funky. I'm assuming these things can change with Persyst versions without our knowing.

With that being said though, the data format and such seems like it shouldn't change all of a sudden. 

#### Additional information
Updates the testing repo to: https://github.com/mne-tools/mne-testing-data/pull/73


adam2392:
This is ready for review/merge on my end.

I am not sure why that final CI test failed. It says it has to do with unpacking the tar file(?)

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+                continue
+
+            # make sure key are lowercase
+            if key is not None:
+                key = key.lower()
+
+            # FileInfo
+            if section == 'fileinfo':
+                # extract the .dat file name
+                if key.lower() == 'file':
```
agramfort, line 101: you already made the key lower above

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+        description = [''] * num_comments
+        for t_idx, (_description, (_onset, _duration)) in \
+                enumerate(comments_dict.items()):
+            # extract the onset, duration, description to
+            # create an Annotations object
+            onset[t_idx] = _onset
+            duration[t_idx] = _duration
+            description[t_idx] = _description
+            annot = Annotations(onset, duration, description)
+            self.set_annotations(annot)
```
agramfort, line 230: set_annotation in a loop seem weird

agramfort: are you sure you don't want to do this at the end of the for loop? note that set_annotations flushed previous annotations

mne/io/persyst/tests/test_persyst.py
```diff
@@ -0,0 +1,101 @@
...
+from mne.datasets.testing import data_path, requires_testing_data
+from mne.io import read_raw_persyst
+from mne.utils import run_tests_if_main
+
+fname_lay = op.join(
+    data_path(download=False), 'Persyst',
+    'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay')
+fname_dat = op.join(
+    data_path(download=False), 'Persyst',
+    'sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.dat')
```
agramfort, line 21: you should be calling the _test_raw_reader function on you reader to make sure it's ok.

mne/io/persyst/persyst.py
adam2392: Ah yes that makes sense. You're right. I think I might've indented at some point w/o realizing it.

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+            'n_samples': n_samples
+        }
+        # create Raw object
+        super(RawPersyst, self).__init__(
+            info, preload, filenames=[dat_fpath],
+            last_samps=[n_samples - 1],
+            raw_extras=[raw_extras], verbose=verbose)
+
+        # set annotations based on the comments read in
+        num_comments = len(comments_dict.keys())
```
larsoner, line 217:
```suggestion
        num_comments = len(comments_dict)
```

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+            record = np.fromfile(dat_file_ID, dtype=precision,
+                                 count=count)
+
+        # chs * rows
+        record = np.reshape(record, (n_chs, -1), 'F')
+        # calibrate to convert to uV
+        data[...] = record[idx, ...] * cals
+        # cast as float32; more than enough precision
+        # then multiply to get V
+        data = data.astype(np.float32) / 10. ** 6
```
larsoner, line 277: `mult` is unused, so this code is going to be broken when projectors are used. You probably want to let `_mult_cal_one` take care of this stuff.

Also if you set `info['chs'][k]['cal']` appropriately above, you don't need to do the `1e-6` mult (here done as a div by `10 ** 6`) yourself. Setting it to 1e6 or 1e-6 (can't remember which) should have an equivalent effect.

mne/io/persyst/tests/test_persyst.py
```diff
@@ -0,0 +1,107 @@
...
+    assert_array_equal(first_ch_data.squeeze(), data[0, :])
+
+
+@requires_testing_data
+def test_persyst_wrong_file():
+    """Test reading Persyst files when passed in wrong file path."""
+    with pytest.raises(FileNotFoundError, match='The path you'):
+        read_raw_persyst(fname_dat, preload=True)
+
+    out_dir = mne.utils._TempDir()
```
larsoner, line 84: This is the old way to do it. Better to make use of the pytest fixture:
```suggestion
    out_dir = mne.utils._TempDir()
    out_dir = str(tmpdir)
```

mne/io/persyst/tests/test_persyst.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # data should have been set correctly
+    assert not data.min() == 0 and not data.max() == 0
+
+    first_ch_data = raw.get_data(picks=[0], start=200, stop=400)
+    assert_array_equal(first_ch_data.squeeze(), data[0, :])
+
+
+@requires_testing_data
+def test_persyst_wrong_file():
```
larsoner, line 79:
```suggestion
def test_persyst_wrong_file(tmpdir):
```

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+
+        # takes ~8 min for a 1.5GB file
+        start_time = time.time()
+
+        # sections and subsections currently unused
+        keys, data, sections, subsections = _read_lay_contents(fname)
+
+        # these are the section headers in the Persyst file layout
+        fileinfo_dict = dict()
+        channelmap_dict = dict()
```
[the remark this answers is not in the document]
adam2392: Yes these are the order, but they're always in the increasing order that they are read in. No reason to suspect that the order would change unless the user explicitly modified the file, which shouldn't be done. WDYT? worth adding?

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,443 @@
...
+        sampletimes_dict = dict()
```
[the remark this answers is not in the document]
adam2392: Yeah sampletimes seems redundant, and I included it in here because it was a field, but you can obtain the same info via the `testtime` and `testdate`.

mne/io/persyst/persyst.py
adam2392: Okay I see, I'll adjust to multiply the info cal accordingly.

Re `mult_cal_one` sorry what is this? I've never worked w/ projectors :p

larsoner: Basically instead of multiplying by `cals` and sub-indexing yourself you should let `_mult_cal_one` do it. See how it's used in other readers by `git grep _mult_cal_one`.

agramfort (approved):
looks good !

@adam2392 you should also see then how mne-bids allows you to write such files in bids

adam2392:
This should be good to go assuming the CIs turn green and @larsoner is okay with my changes to address his comments :). Thanks!

adam2392:
Just wanted to confirm that this is "good to go", possibly waiting for some additional CI test from @larsoner in #8210?

larsoner:
Yes let's wait for #8210. I just got it working (I think!). Do you want me to push a merge commit once #8210 is in `master` so we can see if everything is good here?

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,407 @@
...
+            else:
+                raise RuntimeError(f'Unknown format: {datatype}')
+
+            # allow offset to occur
+            f.seek(0, os.SEEK_END)
+            n_samples = f.tell()
+            n_samples = n_samples // (dtype.itemsize * n_chs)
+
+            if verbose:
+                logger.debug(f'Loaded {n_samples} samples '
```
larsoner, line 195: No need to nest in `if verbose`, it magically gets accounted for
```suggestion
            logger.debug(f'Loaded {n_samples} samples '
```

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,407 @@
...
+        'middle_name': patient_dict.get('middle'),
+        'last_name': patient_dict.get('last'),
+        'sex': patient_dict.get('sex'),
+        'hand': patient_dict.get('hand'),
+        'his_id': patient_dict.get('id'),
+        'birthday': birthdate,
+    }
+
+    # Recode sex values
+    if subject_info['sex'] in ('m', 'male'):
```
larsoner, line 290: I would use a dict lookup to avoid all of these uncovered lines and make it easier to expand later:
```
sex_dict = dict(
    m=...,
    male=...,
    f=...,
    female=...,
)
subject_info['sex'] = sex_dict.get(subject_info['sex'], FIFF.FIFFV_SEX_UNKNOWN)
```

mne/io/persyst/persyst.py
```diff
@@ -0,0 +1,407 @@
...
+    # Recode hand values
+    if subject_info['hand'] in ('m', 'male'):
+        subject_info['hand'] = FIFF.FIFFV_SUBJ_HAND_RIGHT
+    elif subject_info['hand'] in ('f', 'female'):
+        subject_info['hand'] = FIFF.FIFFV_SUBJ_HAND_LEFT
+    elif subject_info['hand'] in ('a', 'ambidextrous'):
+        subject_info['hand'] = FIFF.FIFFV_SUBJ_HAND_AMBI
+    else:
+        # no handedness is set when unknown
+        subject_info.pop('hand')
```
larsoner, line 305: same here but I would do:
```
try:
    subject_info['hand'] = hand_dict[subject_info['hand']]
except KeyError:
    subject_info.pop('hand')

larsoner:
Looks like the channel with index 55 (name `'pol $tp9'`) has values all right around `[-0.01200273]`, was this a bad channel during the recording or something?

adam2392:
> Looks like the channel with index 55 (name `'pol $tp9'`) has values all right around `[-0.01200273]`, was this a bad channel during the recording or something?

That is a byproduct of some center's recording systems I work with. So a rule of thumb I've come to realize is that if there is a `$` sign, then it most likely means its a dead channel (not connected, or used as REF, etc.)

agramfort (approved):

larsoner:
Thanks @adam2392 !

PR #8179, opened 2020-08-31
MRG: Trimming Sleep Physionet recordings to avoid W imbalance

hubertjb:
In the [example for the Sleep Physionet dataset](https://mne.tools/stable/auto_tutorials/sample-datasets/plot_sleep.html), there is a large class imbalance in favour of wake (W). This is because the recordings in this dataset were collected over longer periods of time and thus there are long periods of awake time before and after the night. It is common practice in the literature to trim the beginning and end parts of the recordings of this dataset to reduce class imbalance (e.g., https://arxiv.org/abs/1703.04046, https://arxiv.org/abs/1610.01683, https://arxiv.org/abs/1710.00633).

Following discussion with @agramfort, this PR modifies the example to only keep the last 30 minutes before the first sleep event and the first 30 minutes after the last sleep event when building the training and testing sets.

### DIFF
--- a/tutorials/sample-datasets/plot_sleep.py
+++ b/tutorials/sample-datasets/plot_sleep.py
@@ -110,8 +110,8 @@ raw_train.plot(duration=60, scalings='auto')
 #
 # Moreover, the recordings contain long awake (W) regions before and after each
 # night. To limit the impact of class imbalance, we trim each recording by only
-# keeping 30 minutes of wake time before the first occurence and 30 minutes
-# after the last occurence of sleep stages.
+# keeping 30 minutes of wake time before the first occurrence and 30 minutes
+# after the last occurrence of sleep stages.
 
 annotation_desc_2_event_id = {'Sleep stage W': 1,
                               'Sleep stage 1': 2,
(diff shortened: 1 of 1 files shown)


tutorials/sample-datasets/plot_sleep.py
```diff
@@ -118,6 +123,13 @@
 events_train, _ = mne.events_from_annotations(
     raw_train, event_id=annotation_desc_2_event_id, chunk_duration=30.)
 
+# keep last 30-min wake events before sleep and first 30-min wake events after
+# sleep
+sleep_stage_inds = np.where(events_train[:, 2] > 1)[0]
+offset = int((30 * 60) / 30)
```
larsoner, line 129: The units here / reasons for the given numbers are difficult to parse. The 60 makes sense because it's sec/min, the first 30 makes sense because it's min, but then the last 30 is ... ? Probably making use of some informative variable names above (possibly having to do with the time between samples?) and reusing them here could fix this problem.

Also `int(x_int / y_int)` is less clean than `x_int // y_int`

agramfort (approved):

agramfort:
good to go from my end.

@larsoner merge if happy

hubertjb:
> good to go from my end.
> 
> @larsoner merge if happy

Much nicer solution, thanks!

larsoner:
Precision and recall for all classes other than Wake are better on this PR. It would be nice if there were some overall metric that would go up with this PR compared to master, instead of some going up and some going down. @agramfort I'll let you add the right one someday when you're bored :)

PR #8180, opened 2020-09-01
MRG: Test PSD topomap works with NIRS epochs

rob-luke:
#### Reference issue
Ticking off a few of the last issues from #7057 

Turns out psd topomap already worked due to previous code changes. So I just added a test to ensure it continues working.

### DIFF
--- a/mne/viz/tests/test_epochs.py
+++ b/mne/viz/tests/test_epochs.py
@@ -340,6 +340,14 @@ def test_plot_psd_epochs():
     plt.close('all')
 
 
+def test_plot_psdtopo_nirs(fnirs_epochs):
+    """Test plotting of PSD topography for nirs data."""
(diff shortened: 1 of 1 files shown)


rob-luke:
@larsoner @agramfort this is ready for review

rob-luke:
> Worth updating some tutorial?

Yeah I have just started exploring a more time/frequency style of analysis, I think a tutorial would be worthwhile once I've wrapped my head around the best use cases. But that might be a while (manuscript first 😉). Meanwhile, I could add a token `fnirs_epochs.plot_psd_topomap()` to the existing tutorial, but because the usage is no different to other tutorials I'd suggest its unnecessary . But let me know if you think I should put it in.

larsoner:
Nah, good enough :)

PR #8181, opened 2020-09-01
MRG: Split _timeviewer.py into multiple files

GuillaumeFavelier:
`_timeviewer.py` is getting big, around 1500 lines on `master` and will probably continue to grow as more features are added.

This PR moves some classes in their own files and keep the main `_TimeViewer` class:
* `MplCanvas` which manages `matplotlib` integration as a `QWidget`
* the viewer callbacks (ShowView, IntSlider, TimeSlider, SmartSlider, BumpColorbarPoints, UpdateColorbarScale)
* `_LinkViewer` which synchronize multiple instances of `_TimeViewer`

It's only refactoring but it may help with https://github.com/mne-tools/mne-python/issues/7959


GuillaumeFavelier:

agramfort (approved):
@larsoner merge if happy

larsoner:
@GuillaumeFavelier codecov gave some inline comments (awesome!), do you want to take a few minutes to see if you can hit a few of them?

GuillaumeFavelier:
Yes, I'll see what I can do :)

GuillaumeFavelier:
Locally I missed some lines related to matplotlib events in `mplcanvas.py`.

larsoner:
These can be difficult to hit, but we do have `_fake_click` you could use. But we can save that for another PR

larsoner:
(or you can just manually call the callbacks yourself if you want, since we probably have access to them)

larsoner:
Thanks @GuillaumeFavelier

PR #6731, opened 2019-09-03
MRG: Use _Brain in plot_source_estimates

GuillaumeFavelier:
This PR continues the work of #6460 and #6515. The goal is to make `stc.plot()` use the `_Brain` class which relies on the abstract 3d backend layer `_Renderer`. Technically it provides a `PyVista` alternative to the `Mayavi`-based `PySurfer` or any backend that respects the `_Renderer` contract in general. 

In practice, the work has started with the `plot_source_estimates()` function.

***

**ToDo**

*- Ensure consistency of results among all the backends*
*- Update the `BaseRenderer` API*
*- Ensure proper testing of the `_Brain` features*

- [x] Add support for a colorbar and text
- [x] Fix time value in `text_label`
- [x] Import `mesh_edges` and `smoothing_matrix` from `PySurfer`
- [x] Update `stc` tests to use `renderer` fixture 
- [x] Add `add_label()` and `show_view()` functions
- [x] Add a `screenshot()` function
- [x] Add a `add_foci()` function
- [x] Add a `add_text()` function
- [x] Update branch with #6819
- [x] Add a `save_image()` function.

**ToDo in a follow-up PR**

- Port `TimeViewer` from `PySurfer`
- Add `remove_data()`, `remove_labels()`  and `remove_foci()` functions

**API**

`__init__()`:
- [x] Add support for `hemi=split`
- [ ] Add support for `cortex` and `geo_kwargs` *(on user need/request)*
- [x] Add support for `figure` ~~(naming by `int` id is missing)~~

`add_data()`:
- [x] Add support for `transparent`
- [ ] Add support for `verbose` *(on user need/request)*

`add_label()`:
- [ ] Add support for `label_name` *(useful for `remove_label`)*

`add_foci()`
- [ ] Add support for `map_surface` *(on user need/request)*

**Bugs**

- [x] The camera is not set up correctly (see this [comment](https://github.com/mne-tools/mne-python/pull/6731#issuecomment-527885549))
- [x] The colorbar range is different than the reference (see this [comment](https://github.com/mne-tools/mne-python/pull/6731#issuecomment-527912117), fixed by [370](https://github.com/pyvista/pyvista/pull/370))
- [x] Rework of colorbar colormap is [needed](https://github.com/mne-tools/mne-python/pull/6731#issuecomment-527948272)
[...]

PR #8182, opened 2020-09-01
MRG: Improve lighting in 3D viz

GuillaumeFavelier:
This PR reworks lighting in 3d viz scenes:

master | PR
---------|-------
image | image
image | image
image | image
image | image
image | image

Related to https://github.com/mne-tools/mne-python/pull/6731#issuecomment-532625434

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -256,6 +256,8 @@ class _Renderer(_BaseRenderer):
 
         # Inspired from Mayavi's version of Raymond Maple 3-lights illumination
         lights = list(self.plotter.renderer.GetLights())
+        headlight = lights.pop(0)
+        headlight.SetSwitch(False)
         for i in range(len(lights)):
             if i < 3:
                 lights[i].SetSwitch(True)
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -245,15 +245,14 @@ class _Renderer(_BaseRenderer):
     def scene(self):
         return self.figure
 
-    def update_lighting(self):
-        def _to_pos(elevation, azimuth):
-            theta = azimuth * np.pi / 180.0
-            phi = (90.0 - elevation) * np.pi / 180.0
-            x = np.sin(theta) * np.sin(phi)
-            y = np.cos(phi)
-            z = np.cos(theta) * np.sin(phi)
-            return x, y, z
+    def _orient_lights(self):
+        lights = list(self.plotter.renderer.GetLights())
+        lights.pop(0)  # unused headlight
+        lights[0].SetPosition(_to_pos(45.0, -45.0))
+        lights[1].SetPosition(_to_pos(-30.0, 60.0))
+        lights[2].SetPosition(_to_pos(-30.0, -60.0))
 
+    def update_lighting(self):
         # Inspired from Mayavi's version of Raymond Maple 3-lights illumination
         lights = list(self.plotter.renderer.GetLights())
         headlight = lights.pop(0)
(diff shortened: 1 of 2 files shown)


agramfort:
is there a lighting result (eg from pysurfer) that you are aiming for?

GuillaumeFavelier:
The goal is to reproduce the lighting method used by Mayavi. Should I maybe update the table to integrate it?

larsoner:
Yes I have noticed that they are different. The right still doesn't quite look like Mayavi, though -- the contrast is too large. Maybe the ambient light level is higher in Mayavi?

agramfort:
I agree it requires some adjustment for mayavi like figure.

>

GuillaumeFavelier:
Hm... I checked the `ambient` and `diffuse` values and they are the same ([mayavi](https://github.com/enthought/mayavi/blob/master/tvtk/tools/mlab.py#L850-L851) against [pyvista](https://github.com/pyvista/pyvista/blob/master/pyvista/plotting/plotting.py#L1088)) :thinking: 

I'll keep looking

larsoner:
Difference in light follows camera vs not? Default light locations?

larsoner:
Maybe this?

https://github.com/nipy/PySurfer/blob/master/surfer/viz.py#L3496-L3502

GuillaumeFavelier:
It's trickier than it seems. It's technically 3 lights (of type `CameraLight`) but the first one (of type `HeadLight`) is actually disabled.

GuillaumeFavelier:
*Updated table after e23c884*

master | PR
---------|-------
image | image
image |  image
image |  image
image | image
image | image

agramfort:
I like the way it looks !

GuillaumeFavelier:
@larsoner I pushed 4df834b to integrate https://github.com/mne-tools/mne-python/pull/8182#issuecomment-684932447

mne/viz/backends/_pyvista.py
```diff
@@ -243,6 +245,35 @@ def subplot(self, x, y):
...
+        lights = list(self.plotter.renderer.GetLights())
+        lights.pop(0)  # unused headlight
+        lights[0].SetPosition(_to_pos(45.0, -45.0))
+        lights[1].SetPosition(_to_pos(-30.0, 60.0))
+        lights[2].SetPosition(_to_pos(-30.0, -60.0))
+
+    def update_lighting(self):
+        # Inspired from Mayavi's version of Raymond Maple 3-lights illumination
+        lights = list(self.plotter.renderer.GetLights())
+        headlight = lights.pop(0)
```
larsoner, line 258: I'm a bit concerned that there are two `.pop`'s going on. Is it worth adding the conditional:
```
assert len(lights) in (3, 4)
if len(lights) == 4:
    lights.pop(0)
```
both places?

larsoner: ohhh wait you don't actually remove the light from VTK. Never mind!

mne/viz/backends/_pyvista.py
```diff
@@ -646,6 +677,15 @@ def _rad2deg(rad):
     return rad * 180. / np.pi
 
 
+def _to_pos(elevation, azimuth):
+    theta = azimuth * np.pi / 180.0
+    phi = (90.0 - elevation) * np.pi / 180.0
+    x = np.sin(theta) * np.sin(phi)
+    y = np.cos(phi)
+    z = np.cos(theta) * np.sin(phi)
+    return x, y, z
```
larsoner, line 686: This should probably use `mne.transforms.sph_to_cart`

larsoner:
I'll merge this, let's take care of `sph_to_cart` deduplication in another PR when someone sees it and remembers to do it

larsoner:
Thanks @GuillaumeFavelier !

PR #8189, opened 2020-09-01
WIP: Include recordings from extended corpus in Sleep Physionet age dataset

hubertjb:
This PR updates the Sleep Physionet age dataset fetcher to allow downloading the recordings from the [extended version of the corpus](https://physionet.org/content/sleep-edfx/1.0.0/). As a result, there are now 153 recordings available instead of the 39 that were previously available.

There is one issue I wasn't entirely sure how to handle: there are a few missing subjects from the 83 subjects described on the Physionet page (subjects 39, 68, 69, 78 and 79), as well as three subjects with one out of two recordings missing (13, 36 and 52). I took the approach of throwing an error/warning if one of the requested subjects or recordings is missing. I'd be curious to know if @agramfort @larsoner had a preferred approach?


mne/datasets/sleep_physionet/age.py
```diff
@@ -94,7 +104,28 @@ def fetch_data(subjects, recording=[1, 2], path=None, force_update=False,
     path = data_path(path=path, update_path=update_path)
     params = [path, force_update, base_url]
 
-    _check_subjects(subjects, 20)
+    _check_subjects(
+        subjects, 83, missing=[39, 68, 69, 78, 79], on_missing=on_missing)
+
+    # Check for missing recordings
```
hubertjb, line 110: The following could go in a utils function if we keep the `on_missing` approach:

hubertjb:
I forgot to update the tests - I'll fix that and update the PR.

agramfort:
works like a charm for me !

mne/datasets/sleep_physionet/age.py
```diff
@@ -94,7 +104,28 @@ def fetch_data(subjects, recording=[1, 2], path=None, force_update=False,
...
+            pass
+    if 13 in subjects and 2 in recording:
+        msg = ('Requested recording 2 for subject 13, but it is not available '
+               'in corpus.')
+        if on_missing == 'error':
+            raise ValueError(msg)
+        elif on_missing == 'warning':
+            warn(msg)
+        else:
+            pass
```
larsoner, line 128: please do refactor to a `_on_missing(on_missing, msg)`. It looks like there are instances in forward, epochs, and montage that can be refactored to use the same pattern. I'd prefer to standardize around NumPy's naming (at least the subset we use) as raise/warn/ignore, at least in the docstring (the code can treat raise/error and warn/warning equivalently for painless backward compat).

hubertjb: Where would you put `_on_missing`? Would `utils/check.py` be the right place?

agramfort:
+1 for utils/check.py

>

hubertjb:
With this refactoring the check for valid values of `on_missing` is only done in `_on_missing()` and not in the functions that call it. For instance, this means `Epochs` will not throw an error if `on_missing='foo'` unless there's actually a missing epoch. Is that OK? Otherwise I'll add the check back.

mne/datasets/sleep_physionet/tests/test_physionet.py
```diff
@@ -81,16 +81,48 @@ def test_run_update_age_records(tmpdir):
...
+    with pytest.raises(
+            ValueError, match='This dataset contains subjects 0 to 82'):
+        age.fetch_data(
+            subjects=[subject], recording=[1], on_missing='raise', **params)
+    with pytest.warns(RuntimeWarning,
+                      match='This dataset contains subjects 0 to 82'):
+        age.fetch_data(
+            subjects=[subject], recording=[1], on_missing='warn', **params)
+    age.fetch_data(
+        subjects=[subject], recording=[1], on_missing='ignore', **params)
```
agramfort, line 98: if feel this test is going to take too long. Fetching 6 physionet subjects can take a while.

maybe use @pytest.mark.slowtest

@larsoner any better suggestion?

mne/datasets/sleep_physionet/tests/test_physionet.py
```diff
@@ -81,16 +81,48 @@ def test_run_update_age_records(tmpdir):
...
+
+
+@pytest.mark.parametrize('subject,recording', [(13, 2), (36, 1), (52, 1)])
```
agramfort, line 101: same here

mne/epochs.py
```diff
@@ -380,14 +381,12 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
     def __init__(self, info, data, events, event_id=None, tmin=-0.2, tmax=0.5,
                  baseline=(None, 0), raw=None, picks=None, reject=None,
                  flat=None, decim=1, reject_tmin=None, reject_tmax=None,
-                 detrend=None, proj=True, on_missing='error',
+                 detrend=None, proj=True, on_missing='raise',
```
agramfort, line 384: this is a silent API change. I would suggest to keep it as it is here but let the function interpret 'error' as 'raise' and same for warning or warn as I don't want to touch API of Epochs object. Here you are breaking backward compat.

agramfort: you actually accept error in the _on_missing function so don't change anything to Epochs API

mne/utils/check.py
```diff
@@ -700,3 +700,32 @@ def _suggest(val, options, cutoff=0.66):
...
     else:
         return ' Did you mean one of %r?' % (options,)
+
+
+def _on_missing(on_missing, msg):
+    """Raise error or print warning with a message.
+
+    Parameters
+    ----------
+    on_missing : 'raise' | 'warn' | 'ignore'
```
agramfort, line 710:
```suggestion
    on_missing : 'raise' | 'error' | 'warn' | 'warning' | 'ignore'
```

mne/epochs.py
larsoner: I prefer to standardize around raise, see discussion above

This won't break backward compatibility, we leave error as an alias, no need to remove it (almost no effort required to support the old error/warning and raise/warn)

mne/utils/check.py
larsoner: +1 on the current PR, -1 on this suggestion

mne/datasets/sleep_physionet/tests/test_physionet.py
larsoner: I would use monkeypatch to make it actually not bother fetching the data. @hubertjb if this does not make sense I can push a quick commit

hubertjb: Just in case - this test and the next one won't actually fetch the data, because the checks for valid subjects/recordings happen before any fetching happen, but most importantly, the subjects/recordings that are requested in the `@pytest.mark.parametrize` are not available anyway. Is the idea to use monkeypatching just in case a future change makes the tests actually fetch a file?

larsoner: I assumed (and @agramfort probably did, too) that they actually fetch data. If they don't then this is fine! Actually I would go a step further and `monkeypatch` our `utils._fetch_file` or so such that if it attempts to fetch a file in this part of the test it throws an error. Okay for me to push something like this?

hubertjb: OK, good. That works for me @larsoner !

agramfort: if it does not fetch and it's fast then no need to do anything more complicated. I assumed reading the code that it would fetch something

agramfort: you can maybe just assert that number of files is empty if ignore

agramfort (approved):

larsoner:
@hubertjb feel free to take care of @agramfort's last comment about asserting that the file list returned is empty and the merge conflict, or I can do it

larsoner:
Pushed a merge commit + empty path test. Will merge once CIs come back happy. Thanks in advance @hubertjb !

hubertjb:
Great, thanks @larsoner @agramfort !

larsoner:
Thanks @hubertjb !

agramfort:
great Thanks @hubertjb !

PR #8190, opened 2020-09-01
MRG, ENH: ECoG plotting on pial

### LINKED ISSUE #7787: Animate ECoG activity on top of the FreeSurfer pial surface of the brain
#### Describe the problem
In #7768 animation is being added to the `plot_ecog` example. @larsoner suggested that we add in the animation into the example. Currently, it's put together using `matplotlib.animation`,  but seems like mne-python has other animation funcs (e.g. topomap), so I think it would be nice to have a "3D brain animate" function sitting in mne-python.

#### Describe your solution
A  function similar to the topomap (https://mne.tools/dev/generated/mne.Evoked.html#mne.Evoked.animate_topomap) animate function should be possible.

I think a nice feature, would be a four view panel (saggital, coronal, axial and "custom"), where "custom" is looking at the activity from an optimally selected view that shows as many channels as possible on the surface of the brain. The other 3 views are traditional.

@larsoner suggested this might go into the 3D brain that @GuillaumeFavelier is developing? Not familiar w/ this, but happy to help where possible.

If the 3D brain is really nice... I would be interested in discussing how visualization of SEEG activity might look.

#### Additional context
See: https://github.com/mne-tools/mne-python/pull/7768#issuecomment-628797523

larsoner:
Closes #7787.

I don't work with ECoG data so I don't know how sensible this is. @adam2392 can you see if this is something that ECoG people would be interested in? Here is what the tutorial image looks like on my machine:

Screenshot from 2020-09-01 13-22-15

Todo:
- [x] Verify it's actually useful
- [x] Pick a better name for the function
- [x] Add tests

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -19,6 +19,8 @@ Enhancements
 
 - Add ``n_pca_components`` argument to :func:`mne.viz.plot_ica_overlay` by `Eric Larson`_ (:gh:`8351`)
 
+- Add :func:`mne.stc_near_sensors` to facilitate plotting ECoG data by `Eric Larson`_ (:gh:`8190`)
+
 - Add ``proj`` argument to :func:`mne.make_fixed_length_epochs` by `Eric Larson`_ (:gh:`8351`)
 
 - Speed up :class:`mne.decoding.TimeDelayingRidge` with edge correction using Numba by `Eric Larson`_ (:gh:`8323`)
--- a/doc/python_reference.rst
+++ b/doc/python_reference.rst
@@ -804,6 +804,7 @@ Source Space Data
    read_source_morph
    split_label
    stc_to_label
+   stc_near_sensors
    transform_surface_to
    vertex_to_mni
    write_labels_to_annot
--- a/mne/__init__.py
+++ b/mne/__init__.py
@@ -60,7 +60,7 @@ from .source_estimate import (read_source_estimate,
                               spatio_temporal_src_adjacency,
                               spatio_temporal_tris_adjacency,
                               spatio_temporal_dist_adjacency,
-                              extract_label_time_course)
+                              extract_label_time_course, stc_near_sensors)
 from .surface import (read_surface, write_surface, decimate_surface, read_tri,
                       read_morph_map, get_head_surf, get_meg_helmet_surf,
                       dig_mri_distances)
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -18,7 +18,8 @@ from .cov import Covariance
 from .evoked import _get_peak
 from .filter import resample
 from .io.constants import FIFF
-from .surface import read_surface, _get_ico_surface, mesh_edges
+from .surface import (read_surface, _get_ico_surface, mesh_edges,
+                      _project_onto_surface)
 from .source_space import (_ensure_src, _get_morph_src_reordering,
                            _ensure_src_subject, SourceSpaces, _get_src_nn,
                            _import_nibabel, _get_mri_info_data,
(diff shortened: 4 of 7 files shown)


larsoner:
I think some combination of this and sensor-plotting-on-brain would probably be the best. But this at least is maybe an incremental step

larsoner:
https://21939-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html#sphx-glr-auto-tutorials-misc-plot-ecog-py

adam2392:
Wow this is awesome! you are a magician @larsoner.

I think for me I haven't seen these kinds of mne-visualizations yet except for MEG, so the only issue I have is interpreting the cool viz. So correct me where I'm wrong. You are visualizing the gamma band activity mapped to the closest parietal lobe on both hemispheres (in this case it's the left hemisphere), at 0.680 seconds. There is a small window where you are smoothing with a factor of 10. 

I might be asking a stupid question, but what is being shown in the bottom time series plot? not sure what `GFP` stands for and the other things in the legend.

larsoner:
> I think for me I haven't seen these kinds of mne-visualizations yet except for MEG,

The real question to me is, is the method used for projecting the data to the brain sensible for ECoG? It's pretty basic, doesn't use forward/inverse at all... But I imagine this is what people mentally try to do when they see electrodes on the brain anyway. Hence why I think this plus an actual sensor overlay would be potentially very useful.

> You are visualizing the gamma band activity mapped to the closest parietal lobe 

The `Notes` say this, does it clarify it for you?

    This function projects the ECoG sensors to the pial surface, then the
￼    activation at each pial surface vertex is given by the sum of the
￼    activation from each sensor. Activation is weighted by the fractional
￼    ``distance`` from the sensor, such that the vertex where the sensor
￼    projects gets weight 1 and a sensor at ``distance`` meters away (or
￼    larger) gets weight 0.

> There is a small window where you are smoothing with a factor of 10.

Actually the `smoothing` here has no effect. It's spatial upsampling which is not used for these data.

> but what is being shown in the bottom time series plot? not sure what GFP stands for and the other things in the legend.

It's the time course of the vertex shown by the ball, in this case the time course of the gamma activation. GFP is [global field power](https://mne.tools/stable/glossary.html#term-GFP). You can try running the example locally to play with the options and clicking around, playing with spacebar, etc.

adam2392:
Ah okay makes a bit more sense! So just to clarify is this the projected activity of all the channels on the brain? The time course on the bottom shows that for just one region (i.e. the blue ball plotted onto the brain).  If say someone wants to only visualize the effect from a subset of channels, then one simply needs to "pick_channels" right?

### Do I think this is useful?
So the heatmap on the pial surface is a weighted (by distance) linear combination of all the channel's gamma activity. I think that is pretty useful because there are many recordings that are ECoG arrays on the brain cortex, and many publications essentially produce (crappier) versions of what you have here. The cool thing I can see is that any "metric" or feature can be visualized as such following the example. I would definitely use it.

The future of iEEG though with epilepsy surgery is moving towards SEEG (so depth electrodes), which obviously then become way harder to visualize. But one step at a time :p. This looks so cool!

larsoner:
> So just to clarify is this the projected activity of all the channels on the brain?

Each colorful point on the brain is the result of summing together channels that are (or really, the version of them projected to the nearest point on the pial surface is) within `distance` of that point, weighted as 1. if they have 0 distance and 0. if they have `distance` (or greater) distance. So any colorful point is the weighted sum of 1 or more electrodes, but generally it's not going to be more than probably a couple of channels.

> The time course on the bottom shows that for just one region (i.e. the blue ball plotted onto the brain). If say someone wants to only visualize the effect from a subset of channels, then one simply needs to "pick_channels" right?

Yes indeed, do this before `stc.plot` and it should plot the subset. Or just set your `distance` less than your minimum electrode spacing, then you'll see the individual points on the brain, and can click on them to make their time-courses be plotted, etc.

> The future of iEEG though with epilepsy surgery is moving towards SEEG (so depth electrodes), which obviously then become way harder to visualize. But one step at a time :p. This looks so cool!

Hopefully not too bad -- we have translucent volumetric rendering already:

https://mne.tools/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d

sEEG should be fairly easy by comparison :) We can for example, when we add the sensor plots, make sEEG electrodes translucent spheres. And/or we figure out some sensible way to interpolate/map the sEEG data into a volumetric space, then the viz is already done actually. Again I think a combination would probably be best...

> I would definitely use it.

Okay let's wait to see if @agramfort thinks it makes sense, too -- if so, I'll add tests and we can pick a name then we should be good to go.

larsoner:
Actually also @choldgraf if you have an opinion on this PR/viz method it would be great!

adam2392:
> * [ ]  Pick a better name for the function

For naming, I wonder if it should be explicit that this is for the pial surface of the brain? Also... I might be asking another stupid question, but what is `stc` stand for? I couldn't find anything from using mne documentation search. It seems to be a file type that contains source reconstructions?

larsoner:
Tests added, ready for review/merge from my end

larsoner:
> For naming, I wonder if it should be explicit that this is for the pial surface of the brain?

In theory at some point we might want to make the surface to project to optional, so I'd rather leave it. I actually can live with `stc_near_sensors` I guess.

> what is stc stand for? I couldn't find anything from using mne documentation search

Argh, our search needs to work a bit better! See:

https://mne.tools/stable/glossary.html#term-source-estimates-abbr.-stc

adam2392:
> > For naming, I wonder if it should be explicit that this is for the pial surface of the brain?
> 
> In theory at some point we might want to make the surface to project to optional, so I'd rather leave it. I actually can live with `stc_near_sensors` I guess.

What about something to inform ECoG vs SEEG? 

> Hopefully not too bad -- we have translucent volumetric rendering already:
> 
> https://mne.tools/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d
> 
> sEEG should be fairly easy by comparison :) We can for example, when we add the sensor plots, make sEEG electrodes translucent spheres. And/or we figure out some sensible way to interpolate/map the sEEG data into a volumetric space, then the viz is already done actually. Again I think a combination would probably be best...

Whoa, I wasn't aware of this! This looks pretty awesome too... I'm assuming there should be some way of getting the data structures to work w/ arbitrary SEEG feature signals (e.g. either the raw LFP, or some say spectral power, etc.), so that it can plot things in this volumetric rendering? If not, I can open another issue to discuss? Sorry if it's obvious and can :p. I'm just not as familiar w/ the data structures outside `Raw` and `Epochs` because the other things always seem to be in the realm of source loc.

If it can do so, perhaps I'll try it on a future patient and see if I run into any troubles.

larsoner:
> SEEG: would we have an analagous function for doing the volumetric rendering of SEEG?

Yeah we should be able to project sEEG electrode activation into a volume without too much work. Probably about the same number of lines as here.

> I'm assuming there should be some way of getting the data structures to work w/ arbitrary SEEG feature signals (e.g. either the raw LFP, or some say spectral power, etc.), so that it can plot things in this volumetric rendering

Yep, just get those signals then put them into an `EvokedArray` (if they aren't already in an `Evoked`) with a corresponding `info` as done in the example here.

choldgraf:
My 2 cents is that this is a super common visualization in ECoG / sEEG, although it has sort-of dubious physiological truth (most folks just assume there's a spherical activation around each electrode, some other kind of heuristic like that, rather than actual physiological modeling etc). However since this is so common I think it'd be a good feature for the ECoG/sEEG community!

larsoner:
In that case this is ready for review/merge @agramfort . Failure is just 3.8 pre segfault that's happening on `master`

agramfort (commented):
I am personally uneasy advocating a projection method for ECoG data without references to literature doing it the way we do. Physics is here ignored and I am sure there is a more standard way of doing this. If not we need to write a paper ;)

I could already find some work:

https://hal.inria.fr/hal-01278377v1

although this is more ambitious.

larsoner:
>  this is a super common visualization in ECoG / sEEG

@choldgraf any reference papers for "each electrode just projects to a sphere of radius R"? @agramfort won't accept this viz without seeing it in use in the community

adam2392:
> > this is a super common visualization in ECoG / sEEG
> 
> @choldgraf any reference papers for "each electrode just projects to a sphere of radius R"? @agramfort won't accept this viz without seeing it in use in the community

See figures:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5580037/

See EcoG:
http://www.fieldtriptoolbox.org/tutorial/human_ecog/

choldgraf:
The field trip ecog page has some visualizations like this I think, maybe they have a reference http://www.fieldtriptoolbox.org/tutorial/human_ecog/

agramfort:
ok in http://www.fieldtriptoolbox.org/tutorial/human_ecog/ the relevant
cells are 51 and 57.

if we do the same as fieldtrip with similar API I am happy

>

choldgraf:
@agramfort you are parisian we know you are never truly happy

choldgraf:
(sorry, I couldn't resist, I haven't had as much chance to troll you these days)

agramfort:
deep inside I am...

you know when is the best time to visit Paris? August as the Parisians are
not there ;)

>

larsoner:
Let's push this to 0.22 and hopefully we can tackle ECoG + sEEG in that release cycle

jeschj01:
@larsoner This addition looks very useful for my current project. We are trying to show heat maps based on observed change from electrical stimulation. Is 'stc_near_sensors' is not currently available?

larsoner:
@jeschj01 it's not in the `master` branch but it's part of my branch shown at the top of this page, `larsoner:ecog`. If you set up a dev/editable install you can try it out. Or you can overwrite your existing 0.20 install with something like:
```
pip install --upgrade --no-deps https://github.com/larsoner/mne-python/archive/ecog.zip
```

adam2392:
@larsoner jw if you were planning on also adding the SEEG capability sometime (what you linked for source viz looks pretty neat)? No rush of course, I know you're a busy man :). 

What are the steps needed to get it working on the volumetric viz? Just wondering to see if it's in my ability to even come close to implementing. I'm completely lost in mne-python's viz functions.

larsoner:
@agramfort feel free to merge if you're happy

agramfort:
@larsoner it needs a what's new entry then +1 for MRG

PR #8194, opened 2020-09-01
MRG, BUG: Fix sphere glyphs

### LINKED ISSUE #8192: BUG: Arrows gone in mixed norm plotting
Glyph in glass brain in : 

https://mne.tools/dev/auto_examples/inverse/plot_time_frequency_mixed_norm_inverse.html#sphx-glr-auto-examples-inverse-plot-time-frequency-mixed-norm-inverse-py

do not show up. I should see 2 sources.

_Originally posted by @agramfort in https://github.com/mne-tools/mne-python/pull/8188#issuecomment-685065216_

larsoner:
Closes #8192

In this PR on mayavi:

Screenshot from 2020-09-01 16-28-42

pyvista:

Screenshot from 2020-09-01 16-29-02

these look different but I think it might just be a subtle angle/perspective difference (?), here's a different angle:

Screenshot from 2020-09-01 16-29-55
Screenshot from 2020-09-01 16-30-29

@GuillaumeFavelier can you look and see if the difference is meaningful and if necessary push commits to fix it? Also might be a good idea to check the code for any other instances of `if/elif/elif` and change them to `if/elif/else` with an `assert` like I did here (preferably with a user-friendly `_check_option` so that the `assert` is just an internal sanity check).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -313,7 +313,7 @@ Bug
 
 - Fix bug with :meth:`mne.io.Raw.plot_psd` with ``average=False`` and multiple channel types where channel locations were not shown properly by `Eric Larson`_
 
-- Fix bug with :func:`mne.viz.plot_sparse_source_estimates` when using ``'sphere'`` mode by `Eric Larson`_
+- Fix bug with :func:`mne.viz.plot_sparse_source_estimates` when using ``'sphere'`` mode by `Eric Larson`_ and `Guillaume Favelier`_
 
 - Fix bug in FieldTrip reader functions when channels are missing in the ``info`` object by `Thomas Hartmann`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -313,6 +313,8 @@ Bug
 
 - Fix bug with :meth:`mne.io.Raw.plot_psd` with ``average=False`` and multiple channel types where channel locations were not shown properly by `Eric Larson`_
 
+- Fix bug with :func:`mne.viz.plot_sparse_source_estimates` when using ``'sphere'`` mode by `Eric Larson`_
+
 - Fix bug in FieldTrip reader functions when channels are missing in the ``info`` object by `Thomas Hartmann`_
 
 - Throw proper error when trying to import FieldTrip Epochs data with non-uniform time for trials by `Thomas Hartmann`_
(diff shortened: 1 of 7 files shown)


GuillaumeFavelier:
Many thanks for refactoring `quiver3d()`, it greatly helps! 

There is indeed a difference on how the glyph is centered.

*lower opacity on the glyph, center with a red sphere*:

mayavi | pyvista
---------|-----------
image | image

Of course we want the backends to be consistent. By default, I will assume that Mayavi is the desired result and modify `_pyvista.py` accordingly.

GuillaumeFavelier:
Using the `sphere` mode, I have second thoughts though...

Mayavi | PyVista
------|------
image | image

GuillaumeFavelier:
Ideally, I think we want the glyphs to behave like Mayavi in `cone` mode but like PyVista if the `sphere` mode is selected right? @agramfort , @larsoner

hoechenberger:
> Ideally, I think we want the glyphs to behave like Mayavi in `cone` mode but like PyVista if the `sphere` mode is selected right? @agramfort , @larsoner

I would very much say: **yes**

hoechenberger:
No wait. The center of the sphere should be at the exact location where the "red dot / origin" of the cone is. I cannot see in the pics whether for pyvista, the center of the sphere is shifted "to the outside" of the brain, away from the surface. This is not what we want: the center should rest on the surface.

GuillaumeFavelier:
For PyVista, the center of the spheres and the red dots are at the position. Does it answer your question? I can push a test commit and you can verify.

hoechenberger:
> For PyVista, the center of the spheres and the red dots are at the position. Does it answer your question? I can push a test commit and you can verify.

Great, that answers my question.

GuillaumeFavelier:
I pushed d2fc5fb for testing purpose only. You can take a look @hoechenberger, @larsoner. It could be reverted before merge.

x | Mayavi | PyVista
---|---|----
**cone** | image | image
**sphere** | image | image

GuillaumeFavelier:
I pushed 8d51365 to make the cone radius consistent too:

Mayavi | PyVista
---------|--------
image | image

larsoner:
That looks right to me. For quiver matplotlib has the "pivot" argument, but we should just document that it's always the tail for arrows and cones and center for sphere (and make sure that this is the case). Not sure which cylinder should be, probably tail.

hoechenberger:
> Not sure which cylinder should be, probably tail.

I would say tail, too, to be consistent with cones.

GuillaumeFavelier:
For `plot_sparse_source_estimates()` no `cylinder` mode is available but in general, when `quiver3d()` is used with `cylinder`, there is a `glyph_center` parameter to change the 'pivot'.

larsoner:
Ahh okay, so no changes needed then I think!

agramfort:
@larsoner and @GuillaumeFavelier merge if you're happy

thx

hoechenberger (approved):

PR #8198, opened 2020-09-02
MRG, ENH: Add on_missing to plot_events

### LINKED ISSUE #6905: plot_events throws ValueError if event_id contains additional entries
mne.viz.plot_events(events, event_id=event_id) produces 

ValueError: 17 from event_id is not present in events.

Is it possible to make this only a warning and let the code run through? In my experiments the stimuli are letters and not all letters are present in every session, and it is very tiresome creating a specific filtered event dict for every session.

larsoner:
Closes #6905

@AKMeunier @rderollepot can you see if this works for you?

Regarding being able to ignore the existing warning, you can now do `verbose='error'` if you really want to also ignore the `event 2 missing from event_id will be ignored` warnings.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -101,7 +101,9 @@ Changelog
 
 - Add :func:`mne.viz.centers_to_edges` to help when using :meth:`matplotlib.axes.Axes.pcolormesh` with flat shading by `Eric Larson`_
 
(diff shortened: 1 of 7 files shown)


agramfort (approved):

larsoner:
Failure is just 3.8 pre, I'll have to look into that separately...

rderollepot:
Just tried it and yes, it works well for me !

Sorry, I offered to work on it but I never finished what I started...

PR #8200, opened 2020-09-02
MRG, FIX: ch_type in plot_projs_topomap

### LINKED ISSUE #7993: plot_projs_topomap() of data with EEG projection creates "EEG-style" topoplots even for MEG data
#### Describe the bug
Plotting SSP projectors via `Raw.plot_projs_topomap()`, we noticed that -- in some situations -- the visualized MEG fields didn't look quite right, i.e., looked like circular EEG topomaps instead like the expected "helmet-shaped" topomaps. Turns out, this only happened when among `info['projs']` was also an EEG average reference projector.

#### Steps to reproduce
```python
import os.path as op
import mne

raw.copy().set_eeg_reference(projection=True).plot_projs_topomap()  # Add EEG avg ref projector
raw.copy().set_eeg_reference(projection=False).plot_projs_topomap()  # No EEG avg ref projector
```

#### Expected results
MEG topomap fields should follow the MEG helmet, and EEG fields should be circular.

#### Actual results
**With EEG avg ref projector:**
Screenshot 2020-07-10 at 18 31 14

**Without EEG avg ref projector:**
Screenshot 2020-07-10 at 18 31 21

cc @agramfort

larsoner:
Closes #7993

This bug was in this release so no need for `latest.inc` update. You can see the problem on `master` [here](https://mne.tools/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#repairing-ecg-artifacts-with-ssp):

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -342,6 +342,7 @@ def plot_projs_topomap(projs, info, cmap=None, sensors=True,
         spheres.append(this_sphere)
         outliness.append(these_outlines)
         ch_typess.append(ch_type)
+        del data, pos, this_sphere, these_outlines, ch_type
     del sphere
 
     # setup axes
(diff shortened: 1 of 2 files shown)


hoechenberger (approved):
Great! If the artifacts come back correctly, we can merge

larsoner:
[CircleCI renders](https://22023-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#repairing-ecg-artifacts-with-ssp) like this:

agramfort (approved):

PR #8209, opened 2020-09-03
FIX : events_from_annotation when annotations.orig_time is None and f…

### LINKED ISSUE #8201: Empty epochs from valid annotation when Raw cropped from beginning
While working with a cropped raw file I came across a case where `mne.Epochs` did not output what I expected. This happens when the `Raw` object contains an annotation toward the end of the recording and is cropped from the beginning, such that `raw.first_samp` is now close to the onset of the annotation. When calling `mne.Epochs`, the resulting epoch is empty (if I understand correctly, because the event now falls outside of the recording) although it should still be possible to extract this epoch.

#### Code to reproduce

```python
import mne
import numpy as np


# Create fake data
sfreq = 100
duration_s = 10
data = np.random.randn(1, sfreq * duration_s)
info = mne.create_info(ch_names=['EEG1'], ch_types=['eeg'], sfreq=sfreq)

# Add annotation toward the end
onset = [8]
duration = [1]
description = ['0']
annots = mne.Annotations(onset, duration, description)
raw = raw.set_annotations(annots)

# Crop start of raw
raw.crop(tmin=7)

# Extract epochs
events, event_ids = mne.events_from_annotations(raw)
epochs = mne.Epochs(
    raw, events, tmin=0, tmax=1, baseline=None, on_missing='warning')

# epochs is empty
print(epochs.get_data())
```

#### Expected results
I would have expected `mne.Epochs` to return a valid epoch that starts 1 s after the beginning of the cropped raw and ends 1 s after that.

#### Actual results
The resulting epochs object is empty.

#### Additional information

```
Platform:      Linux-5.4.0-45-generic-x86_64-with-glibc2.10
Python:        3.8.5 (default, Aug  3 2020, 16:08:50)  [GCC 7.3.0]
Executable:    /home/hubert/miniconda3/envs/braindecode/bin/python
CPU:           x86_64: 4 cores
Memory:        7.7 GB

mne:           0.20.7
numpy:         1.18.5 {blas=openblas, lapack=openblas}
scipy:         1.4.1
matplotlib:    3.3.0 {backend=module://ipykernel.pylab.backend_inline}

sklearn:       0.23.2
numba:         Not found
nibabel:       Not found
cupy:          Not found
pandas:        1.1.0
dipy:          Not found
mayavi:        Not found
pyvista:       Not found
vtk:           Not found
```

agramfort:
closes #8201

cc @hubertjb

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -335,7 +335,7 @@ Bug
 
 - Fix bug with :meth:`mne.preprocessing.ICA.plot_properties` where a :class:`mne.io.Raw` object with annotations would lead to an error by `Yu-Han Luo`_
 
-- Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when `orig_time` of annotations is None and `raw.first_time > 0`, by `Alex Gramfort`_
+- Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
 API
 ~~~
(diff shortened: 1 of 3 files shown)


larsoner:
Thanks @agramfort

PR #8210, opened 2020-09-03
MRG, BUG: Fix errors in IO/loading/projectors

larsoner:
I wanted to add a tiny test to #8176 to ensure `_mult_cal_one` is used properly, but I found out it's not used properly in a number of existing readers. Basically it means (I think) that for all readers for a non-preloaded `raw`:

1. For all readers, `raw.apply_proj().load_data().get_data()` was broken (fix WIP)
2. For all readers, `raw.apply_proj().pick(...)` was broken (now disallowed; we didn't allow this until this release cycle so not so bad)
3. For some readers (definitely EDF/GDF, maybe CNT, maybe BrainVision), `_mult_cal_one` was not used properly, so projectors were not applied when data were not preloaded

This sounds bad, but `(1)` hopefully isn't too common a practice, (2) is within-release and probably rare, and (3) affects formats where using projections are less common (and using a patterns involving not preloading that are hopefully uncommon).

### DIFF
--- a/mne/io/edf/tests/test_gdf.py
+++ b/mne/io/edf/tests/test_gdf.py
@@ -102,7 +102,9 @@ def test_gdf2_data():
     # gh-5604
     assert raw.info['meas_date'] is None
     _test_raw_reader(read_raw_gdf, input_fname=gdf2_path + '.gdf',
-                     eog=None, misc=None)
+                     eog=None, misc=None,
+                     test_scaling=False,  # XXX this should be True
+                     )
 
 
 @testing.requires_testing_data
--- a/mne/io/egi/tests/test_egi.py
+++ b/mne/io/egi/tests/test_egi.py
@@ -148,7 +148,9 @@ def test_io_egi():
 
     include = ['TRSP', 'XXX1']
     raw = _test_raw_reader(read_raw_egi, input_fname=egi_fname,
-                           include=include, test_rank='less')
+                           include=include, test_rank='less',
+                           test_scaling=False,  # XXX probably some bug
+                           )
 
     assert_equal('eeg' in raw, True)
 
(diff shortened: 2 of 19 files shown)


larsoner:
Okay, should be all good now @agramfort

mne/io/edf/tests/test_gdf.py
```diff
@@ -102,7 +102,9 @@ def test_gdf2_data():
     # gh-5604
     assert raw.info['meas_date'] is None
     _test_raw_reader(read_raw_gdf, input_fname=gdf2_path + '.gdf',
-                     eog=None, misc=None)
+                     eog=None, misc=None,
+                     test_scaling=False,  # XXX this should be True
```
agramfort, line 106: you mean it's not working yet on gdf data?

mne/io/egi/tests/test_egi.py
```diff
@@ -57,7 +57,10 @@
 def test_egi_mff_pause(fname, skip_times, event_times):
     """Test EGI MFF with pauses."""
     with pytest.warns(RuntimeWarning, match='Acquisition skips detected'):
-        raw = _test_raw_reader(read_raw_egi, input_fname=fname)
+        raw = _test_raw_reader(read_raw_egi, input_fname=fname,
+                               test_scaling=False,  # XXX probably some bug
```
agramfort, line 61: idem?

agramfort (commented):
shall we merge this one and try to fix the other readers in other PRs?

mne/io/edf/tests/test_gdf.py
larsoner: The values seem too large. Each of these requires some investigation. I think it should be done in another PR. It's possible nothing is broken but it would be good to understand why the EEG values are around 0.01 even after removing the mean or applying an average reference (for example)

agramfort (commented):
@larsoner merge this one if you want to move on

PR #8212, opened 2020-09-04
FIX: Fix bug with raw scrolling (#8089)

larsoner:
Let's see how much the CIs suffer trying to build an old PR. At least locally `pytest mne/viz/tests/test_raw.py` passes.


larsoner:
CircleCI won't rebuild and deploy, but that's fine. Nothing urgent there -- our docs can be out of date for a week. I'll release and push to PyPi

PR #8225, opened 2020-09-09
FIX: Improve optimization of "auto" cHPI time window

NeuroLaunch:
#### Reference issue
None

Optimizes window length for cHPI amplitude estimation by accounting for the line frequency and its harmonics. The change primarily helps separate higher HPI coil frequencies from nearby line harmonics, but it should also improve robustness for more standard HPI frequencies (e.g. 83 Hz).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -47,6 +47,8 @@ Changelog
 
 - Add support for mixed source spaces to :func:`mne.compute_source_morph` by `Eric Larson`_
 
+- Add support for volume and mixed source spaces to :func:`mne.stats.summarize_clusters_stc` by `Eric Larson`_
+
 - Add support for omitting the SDR step in volumetric morphing by passing ``n_iter_sdr=()`` to `mne.compute_source_morph` by `Eric Larson`_
 
 - Add ``single_volume`` argument to :func:`mne.setup_volume_source_space` to facilitate creating source spaces with many volumes (e.g., all subvolumes of ``aseg.mgz``) by `Eric Larson`_
@@ -350,6 +352,8 @@ Bug
 - Fix bug with :meth:`mne.preprocessing.ICA.plot_properties` where a :class:`mne.io.Raw` object with annotations would lead to an error by `Yu-Han Luo`_
 
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
+
+- Fix bug in :func:`mne.set_eeg_reference` and related functions to set ``info['custom_ref_applied']`` to ``True`` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
 
 - Fix bug with :func:`mne.chpi.compute_chpi_amplitudes` and :func:`mne.chpi.filter_chpi` to optimize time window length by `Steven Bierer`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -47,8 +47,6 @@ Changelog
 
 - Add support for mixed source spaces to :func:`mne.compute_source_morph` by `Eric Larson`_
 
-- Add support for volume and mixed source spaces to :func:`mne.stats.summarize_clusters_stc` by `Eric Larson`_
-
 - Add support for omitting the SDR step in volumetric morphing by passing ``n_iter_sdr=()`` to `mne.compute_source_morph` by `Eric Larson`_
 
 - Add ``single_volume`` argument to :func:`mne.setup_volume_source_space` to facilitate creating source spaces with many volumes (e.g., all subvolumes of ``aseg.mgz``) by `Eric Larson`_
@@ -353,7 +351,7 @@ Bug
 
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
-- Fix bug in :func:`mne.set_eeg_reference` and related functions to set ``info['custom_ref_applied']`` to ``True`` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
+- Fix bug with :func:`mne.chpi.computer_chpi_amplitudes` and :func:`mne.chpi.filter_chpi` to optimize time window length by `Steven Bierer`_
 
 API
 ~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -351,7 +351,7 @@ Bug
 
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
-- Fix bug with :func:`mne.chpi.computer_chpi_amplitudes` and :func:`mne.chpi.filter_chpi` to optimize time window length by `Steven Bierer`_
+- Fix bug with :func:`mne.chpi.compute_chpi_amplitudes` and :func:`mne.chpi.filter_chpi` to optimize time window length by `Steven Bierer`_
 
 API
 ~~~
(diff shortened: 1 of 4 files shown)


doc/changes/latest.inc
```diff
@@ -351,6 +351,8 @@ Bug
 
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
+- Fix bug with :func:`mne.chpi._setup_hpi_amplitude_fitting` to improve auto time window by `Steven Bierer`_
```
agramfort, line 354: we don't document the fixes at the level of private functions. You should list the public functions that are impacted by the fix.

agramfort (commented):
is there a way to have a test?

mne/chpi.py
```diff
@@ -443,19 +443,13 @@ def _setup_hpi_amplitude_fitting(info, t_window, remove_aliased=False,
...
-    t_window = float(t_window)
-    if t_window <= 0:
-        raise ValueError('t_window (%s) must be > 0' % (t_window,))
-    logger.info('Using time window: %0.1f ms' % (1000 * t_window,))
-    model_n_window = int(round(float(t_window) * info['sfreq']))
+    if info['line_freq'] is not None:
+        line_freqs = np.arange(info['line_freq'], info['sfreq'] / 3.,
+                               info['line_freq'])
+    else:
+        line_freqs = np.zeros([0])
```
larsoner, line 450: Won't this add an additional DC term in the model? Seems like it would be better to use just:
```suggestion
        line_freqs = np.empty(0)
```
i.e., an empty zero-element array. The concatenations and such should still work. The `logger.info` line might need to be adjusted, though

mne/chpi.py
```diff
@@ -469,14 +463,22 @@ def _setup_hpi_amplitude_fitting(info, t_window, remove_aliased=False,
...
+        if len(hpi_freqs):
+            all_freqs = np.concatenate((hpi_freqs, line_freqs))
+            delta_freqs = np.diff(np.unique(all_freqs))
+            t_window = max(5. / all_freqs.min(), 1. / delta_freqs.min())
+        else:
+            t_window = 0.2
+    t_window = float(t_window)
+    if t_window <= 0:
+        raise ValueError('t_window (%s) must be > 0' % (t_window,))
+    logger.info('Using time window: %0.1f ms' % (1000 * t_window,))
```
larsoner, line 480: In a test that uses `test_move_anon_raw`, this logging line should have its value changed IIRC, because the line freq is 60 and the first cHPI freq is 83. So your updated test could be something like:
```
with catch_logging() as log:
    compute_chpi_...(..., t_window='auto', verbose=True)
assert '43.25 ms' in log.getvalue()
```
or something similar.

mne/chpi.py
```diff
@@ -469,14 +463,22 @@ def _setup_hpi_amplitude_fitting(info, t_window, remove_aliased=False,
...
-                               info['line_freq'])
-    else:
-        line_freqs = np.zeros([0])
-    logger.info('Line interference frequencies: %s Hz'
-                % ' '.join(['%d' % lf for lf in line_freqs]))
-
+    # calculate optimal window length.
+    if isinstance(t_window, str):
+        if t_window != 'auto':
+            raise ValueError('t_window must be "auto" if a string, got %r'
```
larsoner, line 469: To make codecov happy, you can do nowadays:
```suggestion
        _check_option('t_window', t_window, ('auto',), extra='if a string')
```

doc/changes/latest.inc
NeuroLaunch: I can work with @larsoner to devise a test with a range of HPI frequencies.

mne/chpi.py
NeuroLaunch: I think that line was intended to do exactly what you suggest: give an empty array (np.zeros[0]) doesn't fill with a zero, but the resulting array has zero size).  Your way is more explicit.

larsoner: Argh you're right -- either way is fine by me

mne/chpi.py
```diff
@@ -837,8 +835,19 @@ def compute_chpi_amplitudes(raw, t_step_min=0.01, t_window='auto',
        sinusoidal amplitudes to MEG channels.
        It uses SVD to determine the phase/amplitude of the sinusoids.
 
-    The output is meant to be used with :func:`~mne.chpi.compute_chpi_locs`.
+    The number of fitted points ``n_pos`` will depend on the velocity of head
+    movements as well as ``t_step_max`` (and ``t_step_min`` from
+    :func:`mne.chpi.compute_chpi_amplitudes`).
```
larsoner, line 840: It looks like this paragraph should not have been moved (though the one below indeed should have)

mne/tests/test_chpi.py
```diff
@@ -222,6 +222,11 @@ def test_calculate_chpi_positions_vv():
     mf_quats = read_head_pos(pos_fname)
     raw = read_raw_fif(chpi_fif_fname, allow_maxshield='yes')
     raw.crop(0, 5).load_data()
+    # check "auto" t_window estimation at full sampling rate
+    with catch_logging() as log:
+        compute_chpi_amplitudes(raw, t_step_min=0.1, t_window='auto',
+                                tmin=0, tmax=2, verbose=True)
+    assert '83.3 ms' in log.getvalue()
```
larsoner, line 229: Took me a bit to figure out that this is 5 / 60 because it's the biggest interval (compared to 1 / 23 = 43.4 ms for the delta between 83 and 60; or 5 / 83 = 60.2 ms for the cHPI freqs).

The `5. / f` is motivated by wanting a stable amplitude estimate. The line noise amplitude should be very stable, so I actually wonder if the `5. / f` check should just apply to the cHPI frequencies, and the `1. / delta` to the concat/unique over line+cHPIs...

On the other hand, we do use this for filter_chpi (which does notching at the line freqs), and MaxFilter's default is bigger anyway (0.2), so I can live with it either way. I guess the longer window here is the more conservative choice, so on balance I think I'm +0.5 on the current approach.

mne/chpi.py
NeuroLaunch: If I recall, there seems to be no explicit `n_pos` in the code so this through me off.  I see there is a reference to head velocity so your comment makes sense.

mne/tests/test_chpi.py
NeuroLaunch: Sorry, I meant to write some comments but was thrown off by the merge conflicts (I only meant to save to my own fork, forgot the auto stuff).  Perhaps the test I was piggy-backing upon is not robust, as the resulting 83.3 ms time window is trivially determined from the 60 Hz line freq.  Maybe I can temporarily alter the HPI freqs (on a copy of the clipped raw) to apply the proper stresses?

larsoner: No this is good enough, `master` would not use this frequency, so it already demonstrates the change

larsoner: (well, that something was changed, not all the changes -- but good enough for this I think)

doc/changes/latest.inc
```diff
@@ -47,8 +47,6 @@ Changelog
 
 - Add support for mixed source spaces to :func:`mne.compute_source_morph` by `Eric Larson`_
 
-- Add support for volume and mixed source spaces to :func:`mne.stats.summarize_clusters_stc` by `Eric Larson`_
```
larsoner, line 50: Looks like a rebase error

doc/changes/latest.inc
```diff
@@ -353,7 +351,7 @@ Bug
 
 - Fix bug with :func:`mne.events_from_annotations(raw.annotations) <mne.events_from_annotations>` when ``orig_time`` of annotations is None and ``raw.first_time > 0``, by `Alex Gramfort`_
 
-- Fix bug in :func:`mne.set_eeg_reference` and related functions to set ``info['custom_ref_applied']`` to ``True`` for 'ecog' and 'seeg' channels in addition to 'eeg' by `Liberty Hamilton`_
```
larsoner, line 356: This removal also looks like a rebase error

NeuroLaunch: Still new at this.  Using vimdiff (meld not supported on Mac?), and probably misinterpreted a squiggly colored line.  I'll try once more, comparing directly to the current lines in master.

larsoner:
on macOS I use https://yousseb.github.io/meld/ and probably followed https://stackoverflow.com/questions/43317697/setting-up-and-using-meld-as-your-git-difftool-and-mergetool-on-a-mac#answer-43868889 to set it up

larsoner:
Thanks @NeuroLaunch !

PR #8243, opened 2020-09-11
MRG, ENH: Add compute_native_head_t

larsoner:
Alternative to #7709.
Closes #7709.

@adam2392 I think this is a better fix -- basically we expose a way to get the transformation that is applied when doing `raw.set_montage` so it  can be used later. This way, all of the coordinate frames (head and MRI) are correct according to our definitions, when possible.

@jeschj01 this simplifies a bit / shows you what I suggested that you do on gitter. Will be useful for #8190 and other ECoG/sEEG-related stuff.

### DIFF
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -308,7 +308,7 @@ class DigMontage(object):
 
         Returns
         -------
-        tr : instance of Transformation
+        tr : instance of Transform
             The transformation. In the special case where the ``coord_frame``
             during montage init was ``'mri'``, this will be a :term:`trans`
             from the MRI to head coordinate frame.
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -303,22 +303,6 @@ class DigMontage(object):
 
         return dig_names
 
-    def compute_native_head_t(self):
-        """Compute the transformation to the head coordinate frame.
-
-        Returns
-        -------
-        tr : instance of Transform
-            The transformation. In the special case where the ``coord_frame``
-            during montage init was ``'mri'``, this will be a :term:`trans`
-            from the MRI to head coordinate frame.
-
-        Notes
-        -----
-        .. versionadded:: 0.21
-        """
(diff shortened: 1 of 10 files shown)


larsoner:
Updated tutorial:

https://22251-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html

mne/channels/montage.py
```diff
@@ -302,6 +303,22 @@ def _get_dig_names(self):
...
+        tr : instance of Transform
+            The transformation. In the special case where the ``coord_frame``
+            during montage init was ``'mri'``, this will be a :term:`trans`
+            from the MRI to head coordinate frame.
+
+        Notes
+        -----
+        .. versionadded:: 0.21
+        """
+        return compute_native_head_t(self)
```
agramfort, line 320: shall we deprecate function then? so we have one way of doing it?

agramfort (commented):
ok for you @adam2392 ?

mne/channels/montage.py
larsoner: I'm fine with leaving both. But really I guess I should kill this method and use the function, and then add the function to `python_reference.rst`

agramfort (approved):

agramfort:
@larsoner merge if CIs are green

adam2392:
This LGTM! Just to clarify, this trans is basically used for any plotting funds so as to keep the positions in the montage in their native space? I saw you passed to the plotting of the sensors on piano surface.

larsoner:
It's to allow transforming between the head coordinate system and the MRI coordinate system. It just so happens that ECoG/sEEG people tend to have their coordinates in MRI space (or maybe MNI and they want to use fsaverage), and this PR is meant to unify that with the more general coordinate system framework in MNE.

Writing this makes me think I should take the time to explain coordinate system standards in MNE in some tutorial. Right now it's scattered across a number of tutorials and documentation elements...

PR #8244, opened 2020-09-11
maybe improve linkcheck?

drammock:
- updates a few URLs to explicitly use https
- changes xcode url, the old one was an appstore redirect that won't even load on non-macs


agramfort (approved):

larsoner:
Thanks @drammock

PR #8245, opened 2020-09-11
MRG, ENH: Tweak vol plot defaults

larsoner:
1. Change some defaults so that colors are more visible/true
2. Change one class so that setting the interpolation mode actually has an effect (for `vtkImageResample`, setting to nearest neighbor for example did nothing, now it does)
3. Change `roll` param for `axial` view to correct it

`master:`

master

PR:

PR

EDIT: Updated with silhouette:

### DIFF
--- a/mne/defaults.py
+++ b/mne/defaults.py
@@ -83,7 +83,8 @@ DEFAULTS = dict(
                       combine_xyz='fro', allow_fixed_depth=True),
     interpolation_method=dict(eeg='spline', meg='MNE', fnirs='nearest'),
     volume_options=dict(
-        alpha=None, resolution=1., surface_alpha=None, blending='mip'),
+        alpha=None, resolution=1., surface_alpha=None, blending='mip',
+        silhouette_alpha=None, silhouette_linewidth=2.),
 )
 
 
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -1934,7 +1934,7 @@ class _BaseVolSourceEstimate(_BaseSourceEstimate):
     @copy_function_doc_to_method_doc(plot_source_estimates)
     def plot_3d(self, subject=None, surface='white', hemi='both',
                 colormap='auto', time_label='auto', smoothing_steps=10,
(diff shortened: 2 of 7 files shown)


agramfort:
I find the brain a bit too dark now. anyone has an opinion?

larsoner:
My problem with the current defaults is that they really mess with the colors. To me it's more similar now to the matplotlib-produced plots like:

larsoner:
It might also be possible to use a silhouette filter to achieve something more like the nilearn glass brain

https://lorensen.github.io/VTKExamples/site/Cxx/PolyData/Silhouette/

I'll try on Monday

larsoner:
[Updated example](https://22277-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#sphx-glr-auto-tutorials-source-modeling-plot-beamformer-lcmv-py):

agramfort (approved):

larsoner:
Tweaked defaults. Results for [the CircleCI example](https://22293-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#sphx-glr-auto-tutorials-source-modeling-plot-beamformer-lcmv-py):



Now for the original dataset plotted at the top:

Screenshot from 2020-09-13 09-21-32

I think the silhouette gives a nice sense of the bounds of the volume source space without disrupting the colors as much as `master`. And it seems to work pretty well during interaction, too:

Peek 2020-09-13 09-08

Added parameters that can be used to tweak the silhouette alpha and linewidth, too.

larsoner:
@GuillaumeFavelier do you want to look? I ended up needing `N` mappers for the `N` renderers so had to do an `import vtk` in `_brain`. Not the nicest...

Also I noticed that, when click-and-dragging one view, *all* views go into low-quality/high-speed mode rather than just the view that's being manipulated. I wonder if this would be solved by using `N` mappers for the `N` renderers. Just to come back to this internal debate again :)

PR #8252, opened 2020-09-14
WIP: Read measurement date and birth date from SNIRF files

rob-luke:
Reads measurement date and birthdate for SNIRF files.

#### Additional information
See spec here https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsimetadatatagsmeasurementdate 

Parallel writer here https://github.com/mne-tools/mne-nirs/pull/166 with check to ensure date time does vary here https://github.com/mne-tools/mne-nirs/pull/166/files#diff-97adcb285086e704b219a0e18ede4576R47-R49

### DIFF
--- a/mne/io/snirf/_snirf.py
+++ b/mne/io/snirf/_snirf.py
@@ -4,6 +4,7 @@
 
 import re
 import numpy as np
+import datetime
 
 from ..base import BaseRaw
 from ..meas_info import create_info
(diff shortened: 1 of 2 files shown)


mne/io/snirf/_snirf.py
```diff
@@ -221,6 +222,41 @@ def natural_keys(text):
...
+            # Several formats have been observed so we try each in turn
+            for dt_code in ['%Y-%m-%d%H:%M:%SZ',
+                            '%Y-%m-%d%H:%M:%S']:
+                try:
+                    meas_date = datetime.datetime.strptime(
+                        str_datetime, dt_code)
+                    meas_date = meas_date.replace(tzinfo=datetime.timezone.utc)
+                    break
+                except ValueError:
+                    pass
```
hoechenberger, line 241: Can we be more specific about the error? Could line 238 be placed outside of the try/except block?

mne/io/snirf/_snirf.py
```diff
@@ -221,6 +222,41 @@ def natural_keys(text):
...
+            if meas_date is None:
+                warn("Extraction of measurement date from SNIRF file failed. "
+                     "The date is being set to January 1st, 2000, "
+                     "instead of {}".format(str_datetime))
```
hoechenberger, line 245: We're allowing use of f-strings now. Care to switch this to an f-string?

https://realpython.com/python-f-strings/#f-strings-a-new-and-improved-way-to-format-strings-in-python

hoechenberger (commented):
Be sure to add a changelog entry too :)

larsoner:
In principle need for `latest.inc` update since this is a within-release fix/enhancement of something @rob-luke added. However, looking at `latest.inc` I see SNIRF reading isn't mentioned, so a line about adding `mne.io.read_raw_snirf` would be great

mne/io/snirf/_snirf.py
rob-luke: Nice idea. Thanks, old habits die hard

mne/io/snirf/_snirf.py
rob-luke: If the date cant be extracted then its printed in the warning below. Do you think I should throw something in the except also? The dates stored in the files seem to have one of the two formats, so I didn't want errors being thrown all the time (although the first one is the actual correct according to the spec).

If I put the replace outside the try block and the date isn't matched, then it will try and set the timezone for none type. So I am not sure where outside the try to put it.

larsoner: You could `try/except/else`, it makes clearer that the failure is supposed to be just in the `strptime` step

larsoner: Pushed a commit for this plus fix for `flake`

larsoner (approved):

larsoner:
Thanks @rob-luke !

rob-luke:
Thanks @hoechenberger and @larsoner.

PR #8254, opened 2020-09-14
MRG, MAINT: Reduce memory usage of examples

### LINKED ISSUE #8241: BUG: VTK/PyVista not freeing memory with multiple figures
A lot of our `master` builds have failed lately, like:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/3597/workflows/fcb12f01-1f51-4d87-8697-5ac9e30f9411/jobs/22228

@GuillaumeFavelier I suspect it could be some recent-ish (within the last 2 months?) viz update. It was happening before the PyVista 0.26 release. Can you look to see if code from a month or two ago uses the same memory as it does now? Often it's `plot_visualize_stc.py` that breaks things, but in the example above it's `plot_forward.py`.

larsoner:
Closes #8241

The goal is to reduce memory usage -- we'll see what CircleCI reports here vs on `master`. But some relevant test timings locally on master:
```
45.42s total   mne/minimum_norm/tests/
37.23s total   mne/tests/test_source_space.py
21.81s total   mne/forward/tests/
 0.00s total   mne/forward/_make_forward.py
```
This PR:
```
47.28s total   mne/minimum_norm/tests/
37.83s total   mne/tests/test_source_space.py
20.52s total   mne/forward/tests/
 0.00s total   mne/forward/_make_forward.py
```
So it looks like removing the `preload=True` from the `read_source_space` code (which was causing memory usage jumps in the hundreds of MB range) seems to be okay

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1488,8 +1488,8 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         class_name = 'Epochs' if class_name == 'BaseEpochs' else class_name
         return '<%s | %s>' % (class_name, s)
 
-    @fill_doc
-    def crop(self, tmin=None, tmax=None, include_tmax=True):
+    @verbose
+    def crop(self, tmin=None, tmax=None, include_tmax=True, verbose=None):
         """Crop a time interval from the epochs.
 
         Parameters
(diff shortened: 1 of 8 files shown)


larsoner:
These are hopefully pretty straightforward memory reduction methods, let's see if things work in `master`.

PR #8260, opened 2020-09-15
MAINT: Speed up more [skip travis]

### LINKED ISSUE #8008: API: Confirm best defaults for LCMV/DICS before 0.21
In #7656 we set the default `make_lcmv(..., weight_norm='sqrtm')` because it's superior to the official `'unit-noise-gain'` in extensive simulation testing. But we should just double check with @britta-wstnr (busy right now AFAIK) that this is the default we should have for 0.21 before releasing.

larsoner:
Lots of little memory tweaks, hopefully once green+merge CircleCI builds properly.

Closes #8008

### DIFF
--- a/tutorials/source-modeling/plot_beamformer_lcmv.py
+++ b/tutorials/source-modeling/plot_beamformer_lcmv.py
@@ -205,18 +205,10 @@ del filters, filters_vec
 # The plots for the scalar beamformer show brain activity in the right temporal
 # lobe around 100 ms post stimulus. This is expected given the left-ear
 # auditory stimulation of the experiment.
-#
-# Volumetric rendering (3D)
-# ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 lims = [0.3, 0.45, 0.6]
 kwargs = dict(src=src, subject='sample', subjects_dir=subjects_dir,
               initial_time=0.087, verbose=True)
-
-brain = stc.plot_3d(
-    clim=dict(kind='value', pos_lims=lims), hemi='both',
-    views=['sagittal', 'coronal', 'axial'], size=(800, 300),
-    view_layout='horizontal', show_traces=0.4, **kwargs)
 
 ###############################################################################
 # On MRI slices (orthoview; 2D)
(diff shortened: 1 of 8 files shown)


tutorials/source-modeling/plot_beamformer_lcmv.py
```diff
@@ -69,6 +69,7 @@
 # Visualize averaged sensor space data
 evoked = epochs.average()
 evoked.plot_joint()
+evoked.crop(0.05, 0.15, verbose='error')  # ignore baseline
```
hoechenberger, line 72: I think you should move this down into the next section, as it could be confusing having it up here, under "Visualize data"

tutorials/source-modeling/plot_mne_solutions.py
```diff
@@ -30,6 +30,7 @@
 fname_cov = data_path + '/MEG/sample/sample_audvis-cov.fif'
 fwd = mne.read_forward_solution(fname_fwd)
 cov = mne.read_cov(fname_cov)
+evoked.crop(0.05, 0.15, verbose='error')  # ignore baseline
```
hoechenberger, line 33: General question: Is it actually a good idea to simply ignore the baseline when examining the inverse solution? Is it not useful to know what the baseline looks like? At the sensor level, it's extremely valuable information… and I always thought it helps me decide whether things are going well on the source level as well. So, if normally one should **also** look at the inverse baseline, we should explicitly state that we're merely cropping here to save some memory and suggest users "do not try this at home".

larsoner: There are many things it's useful to investigate at many stages, I don't think we need to show / include them in all examples

larsoner:
Updated wordings/placements and added a fix for #8008

hoechenberger (approved):

tutorials/source-modeling/plot_beamformer_lcmv.py
```diff
@@ -286,8 +279,8 @@
 stc_fs = morph.apply(stc)
 del stc
 
-brain = stc_fs.plot_3d(
-    src=src_fs, initial_time=0.085, subjects_dir=subjects_dir,
+stc_fs.plot(
+    src=src_fs, mode='stat_map', initial_time=0.085, subjects_dir=subjects_dir,
     clim=dict(kind='value', pos_lims=lims), verbose=True)
```
hoechenberger: What does this change achieve? Asking for a friend. 😁

larsoner: `plot` uses matplotlib 2D mode via `nilearn.plotting`, `plot_3d` uses 3D volumetric rendering via `_Brain`

hoechenberger: thanks!

PR #8262, opened 2020-09-15
MRG, ENH: Reduce memory usage

larsoner:
1. Add memory usage for SourceEstimate and SourceSpaces
2. Use native memory types for source space label info (no need to upcast to float, this requires 2x mem)
3. Add option to `prepare_inverse_operator` to avoid copying `src` as it can be hundreds of MB (this is pretty minor and I doubt users will care, mostly it helps internally with `apply_inverse`, so no `latest.inc` update)
4. Add "maximum intensity projection" to our glossary
5. Minor tweaks to `plot_visualize_stc.py` which is our latest [CircleCI failure culprit](https://app.circleci.com/pipelines/github/mne-tools/mne-python/3705/workflows/ad776bdf-33b0-4f7c-b08b-47c0c225403b/jobs/22337)

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -189,6 +189,8 @@ Changelog
 
 - Add ``component_order`` parameter to :class:`mne.decoding.CSP` which allows switching between ``mutual_info`` (default) and ``alternate`` (a simpler and frequently used option) by `Martin Billinger`_ and `Clemens Brunner`_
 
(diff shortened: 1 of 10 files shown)


larsoner:
I'm going to merge this to try to help CircleCI, but @hoechenberger @agramfort feel free to look and if you have comments I'll make a follow-up PR

Issue #8258, opened 2020-09-15
Restrict allowed "verbose" values, or make the verbose kwarg a keyword-only argument?


hoechenberger:
Call me stupid but I just spent 5 minutes trying to "debug" why this wasn't doing as I expected:

```python
epochs.apply_baseline(None, 0)
```

Obviously, the baseline period should be specified as a tuple… What happened here is that instead of throwing an error message, I ended up with Epochs without baseline correction. Signature of `apply_baseline` is:
```python
    @verbose
    def apply_baseline(self, baseline=(None, 0), verbose=None):
```
So `verbose` was simply set to `0`.

While you could always argue that of course you should check your results and that you cannot make a software fool-proof, I was still wondering if we couldn't easily prevent some of those "user bugs" from happening. In this particular case, either being more restrictive on the allowed values of `verbose`, or declaring `verbose` to be a keyword-only argument would have successfully mitigated my inaptness :)

```python
    def apply_baseline(self, baseline=(None, 0), *, verbose=None):
```

Thoughts?

cbrnr:
I like both suggestions: make `verbose` a kw-only argument and restrict its values.

hoechenberger:
There's a third (complementary) solution: at one point, we should start discussing adding type annotations.

cbrnr:
Let's try to move this point as far as possible into the future...

larsoner:
I'm fine with adding the `*` to functions where `verbose` is close to the other arguments. We can even just start with `apply_baseline` for 0.21 as I agree it's a likely candidate for this problem.

I'm also fine with making the `verbose` decorator more picky, but let's do that early in 0.22

hoechenberger:
> I'm fine with adding the `*` to functions where `verbose` is close to the other arguments. We can even just start with `apply_baseline` for 0.21 as I agree it's a likely candidate for this problem.

I can do this later tonight I believe.

agramfort:
remind me of

np.zeros(2, 3) :)

>

hoechenberger:
> np.zeros(2, 3) :)

I **literally** had to run this code to figure out what's wrong with it. Ouch.

PR #8263, opened 2020-09-15
WIP: verbose -> keyword-only arg in apply_baseline

hoechenberger:
#### Reference issue
Addresses part of #8258

@larsoner I didn't have much time tonight and this is how far I got. Feel free to bring this one home in time for 0.21; I probably won't have time to give it any more love before the weekend.

### DIFF
--- a/mne/externals/decorator.py
+++ b/mne/externals/decorator.py
@@ -161,7 +161,13 @@ class FunctionMaker(object):
 
     def make(self, src_templ, evaldict=None, addsource=False, **attrs):
         "Make a new function from a given template and update the signature"
-        src = src_templ % vars(self)  # expand name and signature
+        # Expand name and signature. Remove `*` argument if present
+        # (placed before keyword-only arguments in function signatures).
+        vars_ = vars(self)
+        if 'kwonlyargs' in vars_:
+            vars_['signature'] = vars_['signature'].replace('*, ', '')
+        src = src_templ % vars_
+
         evaldict = evaldict or {}
         mo = DEF.search(src)
         if mo is None:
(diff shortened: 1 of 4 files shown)


mne/externals/decorator.py
```diff
@@ -161,7 +161,13 @@ def update(self, func, **kw):
 
     def make(self, src_templ, evaldict=None, addsource=False, **attrs):
         "Make a new function from a given template and update the signature"
-        src = src_templ % vars(self)  # expand name and signature
+        # Expand name and signature. Remove `*` argument if present
+        # (placed before keyword-only arguments in function signatures).
+        vars_ = vars(self)
+        if 'kwonlyargs' in vars_:
+            vars_['signature'] = vars_['signature'].replace('*, ', '')
+        src = src_templ % vars_
```
larsoner, line 169: In general we try just to vendor this rather than make our own updates, so we should check to see if an updated version of https://github.com/micheles/decorator/blob/master/src/decorator.py will do what we want

hoechenberger: So this is what happens when you don't check a file's entire path, and only look at its base name. 🤦‍♂️

cbrnr: Which functionality from this package do we use that's not already in Python's standard library?

larsoner: IIRC mostly the `verbose` decorator needs it to preserve traceback information properly. It's possible it's outdated for Python 3.6+, not sure

hoechenberger: I've checked with the latest [`decorator`](https://github.com/micheles/decorator) release and the problem persists.

PR #8264, opened 2020-09-16
ENH: Add Evoked.tmin, Evoked.tmax

hoechenberger:
Add Evoked.tmin, Evoked.tmax properties for more consistency with Epochs.

This was originally a part of GH-8156.


larsoner:
Thanks @hoechenberger

PR #8266, opened 2020-09-16
MRG, MAINT: Lower memory again

larsoner:
Another round. This time I cut down `plot_mne_dspm_source_localization` to deduplicate its content with other examples/tutorials, and removed a bunch of `verbose='error'`s that we no longer need.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2
 jobs:
     build_docs:
       docker:
-        - image: circleci/python:3.8.1-buster
+        - image: circleci/python:3.8.5-buster
       steps:
         - checkout
         - run:
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ doc/auto_examples/
 doc/auto_tutorials/
 doc/modules/generated/
 doc/sphinxext/cachedir
+tutorials/misc/report.h5
 pip-log.txt
 .coverage*
 tags
--- a/doc/carousel.inc
+++ b/doc/carousel.inc
@@ -14,7 +14,7 @@
         <div class="carousel-inner">
 
           <div class="item active">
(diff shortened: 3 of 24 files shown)


agramfort (commented):
@drammock merge if you're happy. it touches your tutorial text

drammock (approved):

agramfort (approved):

larsoner:
Okay so it looks like it's not specific examples that are the problem. If I build with:
```
BUILD_DEV_HTML=1 mprof run --python sphinx -b html -nWT --keep-going -d _build/doctrees . _build/html
```
I get:

Figure_1

I'll try tacking on an `xvfb-run` (which uses MESA GL, just a never version than CircleCI), but this ever-increasing memory usage is not good.

larsoner:
Okay I think I know the culprit. If I run using the pattern `plot_m`, I get this:

Screenshot from 2020-09-17 19-28-25

You can see a bunch of deltas (good), but also a step at around 120 seconds (bad). The step is the problem. If I run `plot_montage`, I get:

Screenshot from 2020-09-17 19-37-54

Compare this to one of our "large memory" examples, `plot_mne_dspm`:

Screenshot from 2020-09-17 19-53-37

In other words, the `plot_montage` example allocates some memory that never gets freed. It creates probably 30 PyVista plots, and for some reason these must never get GC'ed properly. I'll look into it.

In the meantime I've pushed a commit here to stop building it by removing the `plot_` from the name, and we'll see if the `[circle full]` is green. If it is, I'll merge, and we'll see if `master` is green. Then hopefully tomorrow I can track down what's causing the memory to linger so that we can restore the `plot_`. But we can also do this sometime after release, so at least CircleCI will no longer be stopping us -- hopefully we can release tomorrow!

I'll also try modifying sphinx-gallery to capture this residual memory problem...

agramfort:
success !

https://app.circleci.com/pipelines/github/mne-tools/mne-python/3745/workflows/6bb7974f-e742-4249-815e-ce37558c16e6

thanks so much @larsoner !

>

Issue #8150, opened 2020-08-24
MAINT: Release 0.21


larsoner:
We should cut a release in a few weeks, tentatively scheduled for September 15th. Feel free to tag issues and PRs that realistically can and should make it if you have permissions, and if you don't, feel free to comment on ones that you think should be tagged for 0.21.

agramfort:
+10 !

dengemann:
September 15 ? @larsoner

larsoner:
Whoops, yes I meant Sept 15th :)

hoechenberger:
We need to draft an announcement. Also with the last release we had this small video clip demonstrating the new Time Viewer. We should do a similar demo of some notable features in 0.21 as well. Which are the most notable features?

hoechenberger:
Maybe:

- muscle artifact detection #7407
- improved diploe plotting #8188
- allow renaming of montage channels #7685
- volumetric atlas support #7639
- eSSS #7156
- STC flatmap #8082
- REST  #8132
- cov.plot_topomap #8197
- Persyst reading #8176
- Nihon Kohden reading #6017

cbrnr:
Maybe `mne.io.read_raw` #7809 is also worth mentioning as a QoL improvement.

hoechenberger:
From the chanelog:

- Add 'auto' option to mne.preprocessing.ICA.find_bads_ecg() to automatically determine the threshold for CTPS method by Yu-Han Luo
- Add a notebook 3d backend for visualization in jupyter notebook with mne.viz.set_3d_backend() by Guillaume Favelier
- Use PyVista as the default backend for 3D visualization instead of Mayavi by Guillaume Favelier

larsoner:
* Volumetric 3D rendering (#8064)

larsoner:
BTW we are not ready to release, I'm still doing battle with CircleCI

cbrnr:
@hoechenberger since we're already discussing the release, did you have time to work on an `mne-base` package in conda-forge by any chance? I think this would be pretty useful, because the standard package is really quite bloated if all you need is basic EEG analysis tools.

hoechenberger:
@cbrnr Yes, in https://github.com/conda-forge/mne-feedstock/pull/37
I stopped at some point, I believe there were some difficulties, but cannot remember any details! Would be happy to try picking this up again. But not before the weekend.

drammock:
[...]

PR #8269, opened 2020-09-16
MRG, STY: Highlight new contributors [skip travis]

larsoner:
1. Adds `[ci skip]` and `[skip github]` support to our GitHub CI
2. Moves new contributor lines to the top, selected by [looking at names.inc in this diff](https://github.com/mne-tools/mne-python/compare/6f34b24032d2337ea5782c724079563274b6ece7...b96a760bc133f31715fa383b0a5b5999ae9aa564)
3. Bolds new contributor names
4. Moves new contributor changes to the top of each section
5. Renames "changelog / bug / API" to "enhancements / bugs / API changes" as these are more human-readable

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   MacOS:
+    if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.event.head_commit.message, '[skip github]')"
     runs-on: macos-latest
     defaults:
       run:
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -3,9 +3,9 @@
    whats_new page will have a link to the function/class documentation.
 
 .. NOTE: there are 3 separate sections for changes, based on type:
-   - "Changelog" for new features
-   - "Bug" for bug fixes
-   - "API" for backward-incompatible changes
+   - "Enhancements" for new features
+   - "Bugs" for bug fixes
+   - "API changes" for backward-incompatible changes
 
 .. _current:
 
@@ -30,7 +30,7 @@ Enhancements
 
 - Modified :meth:`mne.Epochs.pick_types` to remove dropped channel types from ``reject`` and ``flat`` dictionaries **by new contributor** |Rahul Nadkarni|_
 
-- Add function :func:`mne.preprocessing.regress_artifact` to remove artifacts using linear regression **by new contributor** |Kyle Mathewson|_ ** and `Eric Larson`_
+- Add function :func:`mne.preprocessing.regress_artifact` to remove artifacts using linear regression **by new contributor** |Kyle Mathewson|_ and `Eric Larson`_
 
 - Add ``sources`` and ``detectors`` options for fNIRS use of :meth:`mne.viz.plot_alignment` allowing plotting of optode locations in addition to channel midpoint ``channels`` and ``path`` between fNIRS optodes **by new contributor** |Kyle Mathewson|_
 
@@ -176,7 +176,7 @@ Enhancements
 
 - Add support for reading and writing surfaces in Wavefront .obj format to the :func:`mne.read_surface` and :func:`mne.write_surface` by `Marijn van Vliet`_
 
-- Add tutorial on how to manually fix BEM meshes in Blender by `Ezequiel Mikulan`_ and `Marijn van Vliet`_
+- Add tutorial on how to manually fix BEM meshes in Blender by `Marijn van Vliet`_ and `Ezequiel Mikulan`_
 
 - :func:`mne.write_evokeds` will now accept :class:`mne.Evoked` objects with differing channel orders in ``info['bads']``, which would previously raise an exception by `Richard Höchenberger`_
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -12,12 +12,37 @@
 Current (0.21.dev0)
 -------------------
 
-Changelog
-~~~~~~~~~
+.. |Rahul Nadkarni| replace:: **Rahul Nadkarni**
+.. |Lau Møller Andersen| replace:: **Lau Møller Andersen**
+.. |Kyle Mathewson| replace:: **Kyle Mathewson**
+.. |Jan Sedivy| replace:: **Jan Sedivy**
+.. |Johann Benerradi| replace:: **Johann Benerradi**
+.. |Martin Schulz| replace:: **Martin Schulz**
+.. |Jeroen Van Der Donckt| replace:: **Jeroen Van Der Donckt**
+.. |Simeon Wong| replace:: **Simeon Wong**
+.. |Svea Marie Meyer| replace:: **Svea Marie Meyer**
+.. |Lx37| replace:: **Lx37**
+.. |Liberty Hamilton| replace:: **Liberty Hamilton**
+.. |Steven Bierer| replace:: **Steven Bierer**
+
+Enhancements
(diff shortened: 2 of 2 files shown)


larsoner:
Bolding the entire lines did not look good, so I changed to bolding the names and moving to the top:

https://22353-1301584-gh.circle-artifacts.com/0/dev/whats_new.html

larsoner:
We don't have a contributor list so we can't copy that. We could copy it for the release notes, though. @hoechenberger @cbrnr @sappelhoff @drammock were commenting over in https://github.com/mne-tools/mne-python/issues/8150, WDYT?

hoechenberger:
**Personally** I would simply to it sth like this:

> # What’s new
> ## Current (0.21.dev0)
> 
> **123 people have contributed code or documentation** to this release. Of these, the following **X were first-time contributors:**
> 
> - Firsty Firstman
> - Secondy Secondwoman
> - Thirdy Whatever
> 
> We're grateful to all contributors!
> 
> ### Enhancements
> - foo by bar
> - baz by xxx
> 
> ### Bugs (should ideally be "Bug Fixes"?!)
> - foo by bar
> - baz by xxx
> 
> ### API changes (should ideally come BEFORE Bug Fixes??)
> - foo by bar
> - baz by xxx
>

larsoner:
> 123 people have contributed code or documentation to this release. 

The problem with this is that it's going to be a pain to keep up to date. We have to ask people to check to see if they have contributed yet and if not, then update the number. And then there will be even more merge conflicts. So I'd rather not add this part. This seems like a final at-relase-time release notes thing.

The "new contributors" section by contrast is pretty easy to update on the fly (at least until we get tens or hundreds of new contributors per release :) )

Otherwise I'm happy with @hoechenberger's proposal and can implement it if others agree

hoechenberger:
> > 123 people have contributed code or documentation to this release.
> 
> The problem with this is that it's going to be a pain to keep up to date. We have to ask people to check to see if they have contributed yet and if not, then update the number. And then there will be even more merge conflicts. So I'd rather not add this part. This seems like a final at-relase-time release notes thing.
> 
> The "new contributors" section by contrast is pretty easy to update on the fly (at least until we get tens or hundreds of new contributors per release :) )

Ohhh ok I wasn't aware of this distinction, I was actually only thinking about "making a changelog for a release". 

Do you think we **could** include these metrics when we actually cut the release? :)

> 
> Otherwise I'm happy with @hoechenberger's proposal and can implement it if others agree

Thanks! I've edited my comment a couple minutes ago to add more colors (yes, colors!) i.e. emojis ;) Not sure if you've seen that! 😅

agramfort:
I know we tend to all have egos but I would not put the names first.
People looking at the doc
should see the most important things.

ym 2c

larsoner:
> Do you think we could include these metrics when we actually cut the release? :)

Yes we always do this, see step (5):

https://github.com/mne-tools/mne-python/wiki/How-to-make-a-release

larsoner:
> I know we tend to all have egos but I would not put the names first.

A compromise solution then is to do something like bold their name in the `latest.inc` document...

hoechenberger:
> I know we tend to all have egos but I would not put the names first.
> People looking at the doc
> should see the most important things.
> 
> ym 2c

I **do** agree, yet I think it's a bit of a tradeoff. We want to incentivise contributions by new developers in particular, and I think putting their name in a prominent place could really provide some motivational boost :) esp. since our release cycle is relatively long (i.e. it may take half a year until you "actually" get acknowledged by appearing in the "stable" changelog)

Maybe a compromise could be to add some kind of navigation / TOC that helps you jump to the desired section(s) immediately? I always missed something like that anyway!

agramfort:
putting in bold the names of new contributors in the list sounds a good idea

drammock:
I like the way @larsoner's bolding of "first time contributor Name Surname" looks in the changelog, but what I don't like is that some are right up top (enhancements) while others are at the top of later sections (bug, API) so they're offscreen / not as prominent.  At the same time, I think I agree with @agramfort that having a special section for just the names seems to foreground the people over the code changes a bit too strongly.  What about doing away with the separate sections for bug/api/enh, and having a color-coded tag (or emoji?) on each change indicating what kind of change it is, and then sorting the *whole list* (as @larsoner has done per-section) to put new contributors at the top?

larsoner:
So something like this? :)

https://22353-1301584-gh.circle-artifacts.com/0/dev/whats_new.html

Then we also bold them once we make the names list for the release. It seems easy to maintain and helps keep them prominent

agramfort:
do what you find best

>

hoechenberger:
> What about doing away with the separate sections for bug/api/enh, and having a color-coded tag (or emoji?) on each change indicating what kind of change it is, and then sorting the _whole list_ (as @larsoner has done per-section) to put new contributors at the top?

Re color codes I would be -1, re emojis I would be +5 :-)

```
🚀 – Enhancement
🐜 – Bug Fix
🧐 – API Change
🥇 – First-Time Contributor

- 🚀 Speed up reading of annotations in EDF+ files by new contributor Jeroen Van Der Donckt
- 🚀 Add head to mri and mri to voxel space transform details to Source alignment and coordinate frames tutorial, by Alex Rockhill
- 🐜 🥇 Fix bug by adding error message when trying to save complex stc data in a non.-h5 format mne.VolSourceEstimate.save() by new contributor Lau Møller Andersen
- 🐜 🥇 Fix bug with logging in mne.io.Raw.set_eeg_reference() and related functions by new contributor Simeon Wong
- 🧐 The default argument meg=True in mne.pick_types() will change to meg=False in version 0.22 by Clemens Brunner
```

drammock:
> the most important feature/distinction in the list of changes is the type of change

yeah, OK. I find that convincing. But 

> BUG means "things change immediately when you run your code, hopefully for the better", API means "things will change soon, change your code" and enhancement means "use this if you want but you don't need to worry about it"

seems to imply that Bugfixes should come *first* and Enhancements should come *last*? Regardless, here's where I stand on the bigger-picture issue:

- bolding names of new contributors in the *ongoing what's new page* is good
- sorting within each what's new section to put new contributors at the top is good
- collapsing all sections is probably bad
- adding a tally to the *release notes* that breaks out numbers for all contributors vs new contributors is good
- maintaining the bolding of names in the release notes is good
- adding a separate section just listing new contrib. names separate from their contributions... @agramfort seems to think it sends the wrong message ("people looking at the docs should see the most important things") but maybe he meant for what's new page, rather than the release notes announcement?  I think I agree it doesn't quite make sense for the ongoing what's new page, but I also think that *we get to decide what is important*, and I think honoring and encouraging new volunteers is important. Moreover, a sentence saying "we are grateful to all our new contributors this release cycle" and then listing the ~15-20 names (can be comma separated rather than one per line) is really not taking up that much space given how long the release notes usually are (see [the last release notes](https://mail.nmr.mgh.harvard.edu/pipermail/mne_analysis/2020-March/006652.html) here).

hoechenberger:
See how Pandas does it:

https://pandas.pydata.org/docs/whatsnew/v1.1.2.html

larsoner:
Seems to be essentially the same as scipy, add a character by their name in the list. I think we can do more than this.

jona-sassenhagen:
Good to see this :)

larsoner:
Is this PR a good enough step for 0.21? And we can do more extensive refactoring during the 0.22 cycle @drammock @hoechenberger ?

drammock (approved):

drammock:
Changes look good but for some reason there are asterisks showing up in the second item here: https://22353-1301584-gh.circle-artifacts.com/0/dev/whats_new.html

larsoner:
It's just an old rendering. Let's merge and see if it's better :)

PR #8274, opened 2020-09-17
WIP, MAINT: Try mamba

larsoner:
Let's see if it's faster/better/working


cbrnr:
Seems like it takes 8s longer now (as compared to some random previous PR). Maybe because it has to pull mamba from conda-forge, which is why we should probably try to move everything to conda-forge to see a speedup (which will also be necessary with micromamba I guess, so maybe let's do #8046 first?).

larsoner:
I'm going to close this for now -- it doesn't seem like the slow part is the initial `conda` / `miniconda` download and setup (and `conda` comes for free now on Azure essentially if we want it), so I don't think there is too much to gain

Issue #8140, opened 2020-08-21
 get_point_spread/get_cross_talk for labels


olafhauk:
#### Describe the problem
get_point_spread/get_cross_talk currently only output results for point sources, but no summaries across vertices within labels (e.g. mean/max/SVD).

#### Describe your solution
I'm planning to add options 'labels' and 'mode' to get_point_spread/get_cross_talk in order to summarise PSFs/CTFs across vertices for a list of labels (e.g. mean, sum, max, SVD). Default would be labels=None and mode=None. 
('mode' would match extract_label_time_course (https://mne.tools/stable/generated/mne.extract_label_time_course.html, and was suggested by Eric a while back)

olafhauk:
I have something like this in mind:

    def _get_psf_ctf(resmat, src, idx, func='psf', mode=None, n_comp=1,
                     norm=False):
        """Get point-spread (PSFs) or cross-talk (CTFs) functions.

        Parameters
        ----------
        resmat : array, shape (n_dipoles, n_dipoles)
            Forward Operator.
        src : Source Space
            Source space used to compute resolution matrix.
        idx : list of int | list of Label | list of lists of int
            Source for indices for which to compute PSFs or CTFs.
            Can be:

            - list of integers:
                Return PSFs/CTFs for all indices specified in idx.
            - list of Label:
                Take indices from specified labels.
            - list of lists of integers:
                Similar to previous option, but indices are specified as list
                of lists of integers instead of labels. For every outer list
                item, compute PSFs/CTFs for indices specified in inner lists.

        func : str ('psf' | 'ctf')
            Whether to produce PSFs or CTFs. Defaults to psf.
        mode: None | 'mean' | 'max' | 'svd'
            Compute summary of PSFs/CTFs across indices specified in 'idx'.

            Can be:
            - None (default):
                Output individual PSFs/CTFs for each specific vertex.
            - 'mean':
                Mean of PSFs/CTFs across vertices.
            - 'max':
                PSFs/CTFs with maximum norm across vertices. Returns the n_comp
                largest PSFs/CTFs.
            - 'svd':
                SVD components across PSFs/CTFs across vertices. Returns the n_comp
                first SVD components.

        n_comp: int
            Number of PSF/CTF components to return for mode='max' or mode='svd'.
        norm : bool
            Whether to normalise to maximum across all PSFs and CTFs (default:
            False). This will be applied before computing summaries as specified in
            'mode'.

        Returns
        -------
[...]

PR #8275, opened 2020-09-17
Resolution labels

olafhauk:
before submitting.

#### Reference issue
Added features and tests for #8140.

Added new features for PSFs/CTFs in labels to _get_psf_ctf() and updated API.
Added tests to test_resolution_matrix.py.

#### Additional information
This should now do what it promises in the documentation. I will now work on the plot example.


olafhauk:
get_point_spread() and get_cross_talk() now work for labels and compute mean/sum/SVD across vertices as promised in the documentation. Can someone please check that this makes sense while I start working on the plot example?

mne/minimum_norm/resolution_matrix.py
```diff
@@ -116,20 +325,64 @@ def get_point_spread(resmat, src, idx, norm=False):
...
+        corresponding summary measure will be computed across all PSFs/CTFs
+        available from idx.
+        Can be:
+
+        * list of integers : Compute PSFs/CTFs for all indices to source space
+          vertices specified in idx.
+        * list of Label : Compute PSFs/CTFs for source space vertices in
+          specified labels.
+
+    mode : None | 'mean' | 'max' | 'svd'
```
larsoner, line 340: This is very close to our `extract_label_time_course` API. This makes me think:

1. It's important to add a `Notes` saying why it differs (if at all)
2. Why can't we just use `extract_label_time_course(psf, labels, mode='max')`? I guess one reason is that we can't get `n_comp`, we're stuck with 1 in `extract_label_time_course`. So I can live with them being separate. Will the output be the same when `n_comp=1`? If so, please add a test verifying it.
3. We should probably rename `svd` to `pca` here for API consistency.
4. `flip` variants should be added -- the PSF/CTF are signed so flipping makes sense

mne/minimum_norm/resolution_matrix.py
```diff
@@ -116,20 +325,64 @@ def get_point_spread(resmat, src, idx, norm=False):
...
+    norm : None | 'max' | 'norm'
+        Whether and how to normalise the PSFs and CTFs. This will be applied
+        before computing summaries as specified in 'mode'.
+        Can be:
+
+        * None : Use un-normalized PSFs/CTFs (Default).
+        * 'max' : Normalize to maximum absolute value across all PSFs/CTFs.
+        * 'norm' : Normalize to maximum norm across all PSFs/CTFs.
+    return_svd_vars : bool
+        Whether or not to return the explained variances across the specified
```
larsoner, line 363: A more compact version I've found is

"If True (default False), return ..."

mne/minimum_norm/resolution_matrix.py
```diff
@@ -138,17 +391,60 @@ def get_cross_talk(resmat, src, idx, norm=False):
...
+        corresponding summary measure will be computed across all PSFs/CTFs
+        available from idx.
+        Can be:
+
+        * list of integers : Compute PSFs/CTFs for all indices to source space
+          vertices specified in idx.
+        * list of Label : Compute PSFs/CTFs for source space vertices in
+          specified labels.
+
+    mode : None | 'mean' | 'max' | 'svd'
```
larsoner, line 406: DRY -- use `%(psf_ctf_mode)s` here for entries that are the same and add entries to `mne/utils/docs.py`

mne/minimum_norm/tests/test_resolution_matrix.py
```diff
@@ -114,13 +116,71 @@ def test_resolution_matrix():
...
+    stc_psf, s_vars_psf = get_point_spread(
+        rm_mne, forward_fxd['src'], idx, mode=mode, n_comp=n_comp,
+        norm='norm', return_svd_vars=True)
+    stc_ctf, s_vars_ctf = get_cross_talk(
+        rm_mne, forward_fxd['src'], idx, mode=mode, n_comp=n_comp,
+        norm='norm', return_svd_vars=True)
+
+    assert_array_almost_equal(s_vars_psf, s_vars_ctf)
+
+    assert s_vars_psf[0] > s_vars_psf[1] > s_vars_psf[2]
```
larsoner, line 155: If it's a variance it should sum to 1.
```suggestion
    assert s_vars_psf[0] > s_vars_psf[1] > s_vars_psf[2]
    assert_allclose(s_vars_psf.sum(), 1.)
```

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,256 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+
+        funcs = u[:, :n_comp]
+
+        # if explained variances for SVD components requested
+        if return_svd_vars:
+
+            # explained variance of individual SVD components
+            s_comp = s[:n_comp]
+            s2_sum = np.sum(s * s)
+            s_var = 100. * s_comp * s_comp / s2_sum
```
larsoner, line 313:
```suggestion
            s2 = s * s
            s_var = 100 * s2[:n_comp] / s2.sum()
```

larsoner (commented):
Some initial comments

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,256 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+        Can be:
+
+        * None : Use un-normalized PSFs/CTFs (Default).
+        * 'max' : Normalize to maximum absolute value across all PSFs/CTFs.
+        * 'norm' : Normalize to maximum norm across all PSFs/CTFs.
+    return_svd_vars : bool
+        Whether or not to return the explained variances across the specified
+        vertices for individual SVD components. This is only valid if
+        mode='svd'.
+        Default return_svd_vars=False.
```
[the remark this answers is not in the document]
olafhauk: I'm glad you said that. I'll only include this in the public function.

mne/minimum_norm/resolution_matrix.py
```diff
@@ -116,20 +325,64 @@ def get_point_spread(resmat, src, idx, norm=False):
...
     Returns
     -------
-    stc: instance of SourceEstimate
-        PSFs as an stc object.
+    stc : instance of SourceEstimate
+        PSFs or CTFs as an STC object.
+        All PSFs/CTFs will be returned as successive samples in one STC object,
+        in the order they are specified in idx. PSFs/CTFs for labels are
+        grouped together.
+    return_svd_vars : 1D array
```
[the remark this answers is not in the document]
olafhauk: It's one variance per component. The SVD (and mean/sum) will be computed across all vertices from all labels (the docstring says "Compute summary of PSFs/CTFs across all indices specified in 'idx'."). If the user wants to get this for individual labels, they'll have to call the function for each label (there is no speed benefit doing it otherwise). I'll illustrate this in the example.

mne/minimum_norm/resolution_matrix.py
olafhauk: 1 and 2: I hadn't thought of the similarity to extract_label_time_course. 
One different is the n_comp option, as you say. 
I don't think "flip" is recommended for PSF/CTFs, as it's important to know how contributions from different vertices cancel out in the mean/sum/pca. For example, it would be informative to know if the average point-spread from one label to another one is zero. This should be the default. Can we add the 'flip' later (if really necessary)?
In extract_label_time_courses, flipping and averaging is done over time courses. I would need to do this across vertices (i.e. flip every PSF/CTF)? The time axis is not meaningful for PSFs/CTFs, which I think is a reason to have separate functions (plus above).
3. Will rename svd to pca.
4. See response to '1 and 2'.

mne/minimum_norm/resolution_matrix.py
olafhauk: Good.

mne/minimum_norm/resolution_matrix.py
olafhauk: That's my fun fact of the day. So I add docdict['psf_ctf_mode'] = """... with the docstring information to mne/utils/docs.py?

mne/minimum_norm/tests/test_resolution_matrix.py
olafhauk: Good point. Actually 100 as it's percent. 
Is there a preference for fraction/percent/variance? I think percent is the most intuitive.

mne/minimum_norm/resolution_matrix.py
olafhauk: Ok.

mne/minimum_norm/resolution_matrix.py
larsoner: Yes and then make sure any function that uses the `%(psf_ctf_mode)s` has either `@verbose` decorator (if it has a `verbose` kwarg) or `@fill_doc` (if it does not have `verbose` kwarg)

mne/minimum_norm/resolution_matrix.py
larsoner: > For example, it would be informative to know if the average point-spread from one label to another one is zero. This should be the default. Can we add the 'flip' later (if really necessary)?
In extract_label_time_courses, flipping and averaging is done over time courses. I would need to do this across vertices (i.e. flip every PSF/CTF)?

Yes but keep in mind `mean_flip` just flips the signs of the vertex values (time courses or PSF/CTF), and it's completely spatial -- it does not vary as a function of "time" or anything else. It's totally a property of the source space. `'mean_flip'` in `extract_label_time_course` (and here) is (would be) functionally equivalent to just having changed your `forward['source_nn']`, i.e., the source normal, to point in the opposite direction and doing your inverse using that, then doing the standard mean over whatever values you have in the label (time or PSF/CTF). To me this mode is especially useful because we recommend doing `mean_flip` over labels instead of `mean` in `extract_label_time_course` to combat this cancellation problem. Hence why I'm pushing for it.

But someone motivated to try it can always add it later :)

mne/minimum_norm/resolution_matrix.py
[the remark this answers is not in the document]
olafhauk: Following your comment I now think I should change it to be comparable to extract_label_time_courses, i.e. input a list of labels and output a list for labels (rather than summarise across all vertices from all labels). Labels can easily be added before applying this function, if required.
Do you agree?

mne/minimum_norm/resolution_matrix.py
olafhauk: I agree it makes total sense for activation time courses. However, the fact that it varies over vertices but not over time is a problem for PSFs/CTFs. It might make sense to "flip" individual PSFs/CTFs (e.g. such that they all correlated positively with each other), but not vertices within PSFs/CTFs. I'll need to think about this more.
I'm happy to add this later if required/desired, but for now I'll move on with the options that are already there.

larsoner: Sounds good

mne/minimum_norm/resolution_matrix.py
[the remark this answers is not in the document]
larsoner: Not sure, what I wrote was just what I naively expected based on looking at the code. I'd let the illustrative example that you're working on be the guide :)

olafhauk:
I have now added an example for PSFs/CTFs for labels, and implemented Eric's suggestions. I now need your input to continue.
I still haven't been able to test this using pyvista yet.

olafhauk:
@larsoner Just checking that you've seen this.

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,236 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+    pca_vars = []
+
+    for verts in verts_all:
+
+        # get relevant PSFs or CTFs for specified vertices
+        funcs = resmat[:, verts]
+
+        # normalise PSFs/CTFs if requested
+        if norm is not None:
+
```
larsoner, line 125: I would make this whole function much more compact. No need to put spaces between every operation/clause

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,236 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+            funcs = _normalise_psf_ctf(funcs, norm)
+
+        # summarise PSFs/CTFs across vertices if requested
+        pca_var = None  # variances computed only if return_pca_vars=True
+        if mode is not None:
+
```
larsoner, line 131: like this

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,236 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+            funcs, pca_var = _summarise_psf_ctf(funcs, mode, n_comp,
+                                                return_pca_vars)
+
+        # convert to source estimate
+        stc = SourceEstimate(funcs, vertno, tmin=0., tstep=1.)
+
```
larsoner, line 137: and this

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,236 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+        stcs.append(stc)
+
```
larsoner, line 139: etc.

mne/minimum_norm/resolution_matrix.py
```diff
@@ -63,51 +67,236 @@ def make_inverse_resolution_matrix(forward, inverse_operator, method='dSPM',
...
+        else:  # faster if only one required
+
+            maxidx = absvals.argmax()
+
+        funcs = funcs[:, maxidx]
+
+    elif mode == 'maxnorm':  # pick PSF/CTF with maximum norm
 
-    return stc
+        norms = np.sqrt(np.sum(funcs ** 2, axis=0))
```
larsoner, line 258:
```suggestion
        norms = np.linalg.norm(funcs, axis=0)
```

larsoner (commented):
Mostly minor gripes from me! However, the example uses 2.4 GB, which is quite a lot:

https://22922-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/sg_execution_times.html

We should reduce this. If you're not sure how, I can take a look. I recently reduced memory usage in the PSF/CTF code elsewhere so hopefully it won't be too bad.

One quick change would be to make it better would be to show just MNE instead of sLORETA. They look remarkably similar anyway, so you can just say this in the narrative doc instead of spending CPU time (and memory) computing something so similar.

olafhauk:
> 
> 
> Mostly minor gripes from me! However, the example uses 2.4 GB, which is quite a lot:
> 
> https://22922-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/sg_execution_times.html
> 
> We should reduce this. If you're not sure how, I can take a look. I recently reduced memory usage in the PSF/CTF code elsewhere so hopefully it won't be too bad.
> 
> One quick change would be to make it better would be to show just MNE instead of sLORETA. They look remarkably similar anyway, so you can just say this in the narrative doc instead of spending CPU time (and memory) computing something so similar.

The problem probably are the resolution matrices, which are n_vertices-by-v_vertices. I'll drop the sLORETA example (as you suggest). Currently, make_inverse_resolution_matrix computes the whole resolution matrix, which is required as input to get_point_spread(). I could extend get_point_spread() such that it takes inverse operator, forward solution and labels as input, and only computes the PSFs/CTFs for specified labels. This could also be a separate function get_point_spread_labels.

larsoner:
> The problem probably are the resolution matrices, which are n_vertices-by-v_vertices. 

Yes indeed. Last time I reduced memory by a factor of ~2 by making it operate column- or row-wise rather than making a copy (e.g., when normalizing). We might be able to do the same thing here.

Can you push a commit to remove the sLORETA example, then we'll see what CircleCI's memory usage is like, then I'll see if I can reduce memory usage?

olafhauk:
> 
> 
> > The problem probably are the resolution matrices, which are n_vertices-by-v_vertices.
> 
> Yes indeed. Last time I reduced memory by a factor of ~2 by making it operate column- or row-wise rather than making a copy (e.g., when normalizing). We might be able to do the same thing here.
> 
> Can you push a commit to remove the sLORETA example, then we'll see what CircleCI's memory usage is like, then I'll see if I can reduce memory usage?

I've just pushed the changes following your suggestions.
I could move some 'mode' options from _summarise_psf_ctf() to _normalise_psf_ctf() and use your column-wise normalisation trick. 
But this time we already use a subset of columns/rows of the resolution matrix (for labels), so the effect should be smaller.
The PCA might also swallow memory (not sure what to do about this).

mne/minimum_norm/tests/test_resolution_matrix.py
```diff
@@ -114,13 +117,86 @@ def test_resolution_matrix():
...
     idx = [1, 100, 400]
-    stc_psf = get_point_spread(rm_mne, forward_fxd['src'], idx, norm=True)
-    stc_ctf = get_cross_talk(rm_mne, forward_fxd['src'], idx, norm=True)
 
-    assert_array_almost_equal(stc_psf.data, stc_ctf.data)
+    # check various summary and normalisation options
+    for mode in [None, 'sum', 'mean', 'maxval', 'maxnorm', 'pca']:
+
+        n_comps = [1, 3]
+
```
larsoner, line 127: Still has unnecessary spacing that makes it harder to read

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+fname_fwd = data_path + '/MEG/sample/sample_audvis-meg-eeg-oct-6-fwd.fif'
+fname_inv = data_path + '/MEG/sample/sample_audvis-meg-oct-6-meg-fixed-inv.fif'
+
+# Read forward solution
+forward = mne.read_forward_solution(fname_fwd)
+# Convert to fixed source orientations
+mne.convert_forward_solution(
+    forward, surf_ori=True, force_fixed=True, copy=False)
+
+# Read inverse operator
```
drammock, line 50: not sure this comment is needed, given that the next line calls the `read_inverse_operator` function. That's the advantage of good naming choices in the API

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+inverse_operator = read_inverse_operator(fname_inv)
+
+# Compute resolution matrices for MNE
+lambda2 = 1. / 3.**2
+method = 'MNE'
+rm_mne = make_inverse_resolution_matrix(forward, inverse_operator,
+                                        method=method, lambda2=lambda2)
```
drammock, line 57:
```suggestion
rm_mne = make_inverse_resolution_matrix(forward, inverse_operator,
                                        method='MNE', lambda2=lambda2)
```
(now that you're not doing sLORETA too, no need to parametrize the method arg)

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+label_names = [label.name for label in labels]
+
+lh_labels = [name for name in label_names if name.endswith('lh')]
+
+# Get the y-location of the label
+label_ypos = list()
+for name in lh_labels:
+    idx = label_names.index(name)
+    ypos = np.mean(labels[idx].pos[:, 1])
+    label_ypos.append(ypos)
```
drammock, line 81:
```suggestion
# Get the y-locations of the labels
label_ypos = [np.mean(labels[label_names.index(name)].pos[:, 1])
              for name in lh_labels]
```
not sure if this suggestion is too cryptic to be worth the savings of lines/characters? Feel free to take it or leave it.

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+rh_labels = [label[:-2] + 'rh' for label in lh_labels]
+
+###############################################################################
+# Compute point-spread function summaries (PCA) for all labels
+# ------------------------------------------------------------
+#
+# We summarise the PSFs per label by their first five principal components, and
+# use the first component to evaluate label-to-label leakage below.
+
+# Compute first SVD component across PSFs within labels
```
drammock, line 96: acronym for singular value decomposition not defined. Please define it or stick with "PCA"

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
+"""
+===========================================================================
+Compute source leakage among labels and visualize it using a circular graph
+===========================================================================
+
+This example computes all-to-all pairwise leakage between 68 regions in
+source space based on MNE inverse solutions and a FreeSurfer cortical
+parcellation. Label-to-label leakage is estimated as the correlation among the
+labels' point-spread functions. It is visualized using a circular graph
```
drammock, line 9:
```suggestion
labels' point-spread functions (PSFs). It is visualized using a circular graph
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+
+###############################################################################
+# We can show the explained variances of principal components per label. Note
+# how they differ across labels, most likely due to their varying spatial
+# extent.
+
+with np.printoptions(precision=1):
+    for [name, var] in zip(label_names, pca_vars_mne):
+        print('%s: %.1f%%' % (name, var.sum()))
+        print(var)
```
drammock, line 114:
```suggestion
        print(f'{name}: {var.sum():.1f}% {var}')
```
this condenses the output to one line for each label, and uses the f-string syntax instead of the (deprecated) %-replacement.

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+psfs_mat = np.zeros([n_labels, n_verts])
+
+# Leakage matrix for MNE, get first principal component per label
+for [i, s] in enumerate(stcs_psf_mne):
+    psfs_mat[i, :] = s.data[:, 0]
+
+# Compute label-to-label leakage as Pearson correlation of PSFs
+leakage_mne = np.corrcoef(psfs_mat)
+# Sign of correlation is arbitrary, so take absolute values
+leakage_mne = np.abs(leakage_mne)
```
drammock, line 136:
```suggestion
# Compute label-to-label leakage as Pearson correlation of PSFs.
# The sign of the correlation is arbitrary, so take absolute values.
leakage_mne = np.abs(np.corrcoef(psfs_mat))
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+
+# Compute label-to-label leakage as Pearson correlation of PSFs
+leakage_lor = np.corrcoef(psfs_mat)
+# Sign of correlation is arbitrary, so take absolute values
+leakage_lor = np.abs(leakage_lor)
+
```
drammock, line 142:
```suggestion
```
assuming this was cruft from sLORETA?

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Save the plot order and create a circular layout
+node_order = list()
+node_order.extend(lh_labels[::-1])  # reverse the order
+node_order.extend(rh_labels)
```
drammock, line 146:
```suggestion
node_order = lh_labels[::-1] + rh_labels  # mirror label order across hemis 
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+###############################################################################
+# Most leakage occurs for neighbouring regions, but also for deeper regions
+# across hemispheres.
+#
+# Save the figure (optional)
+# --------------------------
+#
+# By default matplotlib does not save using the facecolor, even though this was
+# set when the figure was generated. If not set via savefig, the labels, title,
+# and legend will be cut off from the output png file::
```
drammock, line 168:
```suggestion
# Matplotlib controls figure facecolor separately for interactive display
# versus for saved figures. Thus when saving you must specify ``facecolor``,
# else your labels, title, etc will not be visible::
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+stc_lh = stcs_psf_mne[idx[0]]
+stc_rh = stcs_psf_mne[idx[1]]
+
+# Maximum for scaling across plots
+max_val = np.max([stc_lh.data, stc_rh.data])
+
+###############################################################################
+# Point-spread function for the lateral occipital label in the left hemisphere
+
+brain_lh = stc_lh.plot(subjects_dir=subjects_dir, subject='sample',
```
drammock, line 192: is there a way to disable/hide the various pyvista sliders when generating figures that aren't going to be interactive? I think that would be helpful here.  I thought there was a way, but looking through the stc.plot docstring I'm not finding it. @GuillaumeFavelier @larsoner

drammock: I'm also not sure how much is added by the `show_traces` panel here?  The brains tell a pretty convincing story.

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Authors: Olaf Hauk <olaf.hauk@mrc-cbu.cam.ac.uk>
+#          Martin Luessi <mluessi@nmr.mgh.harvard.edu>
+#          Alexandre Gramfort <alexandre.gramfort@inria.fr>
+#          Nicolas P. Rougier (graph code borrowed from his matplotlib gallery)
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import matplotlib.pyplot as plt
+
```
drammock, line 21:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+subjects_dir = data_path + '/subjects'
+fname_fwd = data_path + '/MEG/sample/sample_audvis-meg-eeg-oct-6-fwd.fif'
+fname_inv = data_path + '/MEG/sample/sample_audvis-meg-oct-6-meg-fixed-inv.fif'
+
+# Read forward solution
+forward = mne.read_forward_solution(fname_fwd)
+# Convert to fixed source orientations
+mne.convert_forward_solution(
+    forward, surf_ori=True, force_fixed=True, copy=False)
+
```
drammock, line 49:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Read inverse operator
+inverse_operator = read_inverse_operator(fname_inv)
+
```
drammock, line 52:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+###############################################################################
+# Read and organise labels for cortical parcellation
+# --------------------------------------------------
+#
+# Get labels for FreeSurfer 'aparc' cortical parcellation with 34 labels/hemi
+labels = mne.read_labels_from_annot('sample', parc='aparc',
+                                    subjects_dir=subjects_dir)
+n_labels = len(labels)
+label_colors = [label.color for label in labels]
+
```
drammock, line 70:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# First, we reorder the labels based on their location in the left hemi
+label_names = [label.name for label in labels]
+
```
drammock, line 73:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+node_order.extend(lh_labels[::-1])  # reverse the order
+node_order.extend(rh_labels)
+
+node_angles = circular_layout(label_names, node_order, start_pos=90,
+                              group_boundaries=[0, len(label_names) / 2])
+
+# Plot the graph using node colors from the FreeSurfer parcellation. We only
+# show the 200 strongest connections.
+fig = plt.figure(num=None, figsize=(8, 8), facecolor='black')
+
```
drammock, line 154:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+leakage_lor = np.abs(leakage_lor)
+
+# Save the plot order and create a circular layout
+node_order = list()
+node_order.extend(lh_labels[::-1])  # reverse the order
+node_order.extend(rh_labels)
+
+node_angles = circular_layout(label_names, node_order, start_pos=90,
+                              group_boundaries=[0, len(label_names) / 2])
+
```
drammock, line 150:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Compute label-to-label leakage as Pearson correlation of PSFs
+leakage_lor = np.corrcoef(psfs_mat)
+# Sign of correlation is arbitrary, so take absolute values
+leakage_lor = np.abs(leakage_lor)
+
+# Save the plot order and create a circular layout
+node_order = list()
+node_order.extend(lh_labels[::-1])  # reverse the order
+node_order.extend(rh_labels)
+
```
drammock, line 147:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Leakage matrix for MNE, get first principal component per label
+for [i, s] in enumerate(stcs_psf_mne):
+    psfs_mat[i, :] = s.data[:, 0]
+
+# Compute label-to-label leakage as Pearson correlation of PSFs
+leakage_mne = np.corrcoef(psfs_mat)
+# Sign of correlation is arbitrary, so take absolute values
+leakage_mne = np.abs(leakage_mne)
+
```
drammock, line 137:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# Evaluate leakage based on label-to-label PSF correlations
+# ---------------------------------------------------------
+#
+# Note that correlations ignore the overall amplitude of PSFs, i.e. they do
+# not show which region will potentially be the bigger "leaker".
+
+# get PSFs from Source Estimate objects into matrix
+psfs_mat = np.zeros([n_labels, n_verts])
+
```
drammock, line 128:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# Leakage matrix for MNE, get first principal component per label
+for [i, s] in enumerate(stcs_psf_mne):
+    psfs_mat[i, :] = s.data[:, 0]
+
```
drammock, line 132:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
```diff
@@ -0,0 +1,209 @@
...
+# ------------------------------------------
+#
+# We need a matching forward solution and inverse operator to compute
+# resolution matrices for different methods.
+
+data_path = sample.data_path()
+subjects_dir = data_path + '/subjects'
+fname_fwd = data_path + '/MEG/sample/sample_audvis-meg-eeg-oct-6-fwd.fif'
+fname_inv = data_path + '/MEG/sample/sample_audvis-meg-oct-6-meg-fixed-inv.fif'
+
```
drammock, line 43:
```suggestion
```

examples/inverse/plot_psf_ctf_label_leakage.py
olafhauk: Thanks - for the example I would prefer the more transparent option.

examples/inverse/plot_psf_ctf_label_leakage.py
olafhauk: I'm afraid I'm still stuck with mayavi.

drammock: ok, but the public documentation gets rendered with pyvista ([here is the version I reviewed](https://23151-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_psf_ctf_label_leakage.html#plot-psfs-for-individual-labels)).  So we may still want/need to tailor the pyvista output to be pedagogically optimal.

olafhauk:
> 
> 
> > > The problem probably are the resolution matrices, which are n_vertices-by-v_vertices.
> > 
> > 
> > Yes indeed. Last time I reduced memory by a factor of ~2 by making it operate column- or row-wise rather than making a copy (e.g., when normalizing). We might be able to do the same thing here.
> > Can you push a commit to remove the sLORETA example, then we'll see what CircleCI's memory usage is like, then I'll see if I can reduce memory usage?
> 
> I've just pushed the changes following your suggestions.
> I could move some 'mode' options from _summarise_psf_ctf() to _normalise_psf_ctf() and use your column-wise normalisation trick.
> But this time we already use a subset of columns/rows of the resolution matrix (for labels), so the effect should be smaller.
> The PCA might also swallow memory (not sure what to do about this).

@larsoner 
If func is a (large) array, and you want the maximum absolute value in a memory-efficient way, would you replace
maxval = np.abs(funcs).max()

by
maxval = [np.abs(f) for f in funcs].max()

?

larsoner:
> If func is a (large) array, and you want the maximum absolute value in a memory-efficient way, would you replace
> maxval = np.abs(funcs).max()
>
> by
> maxval = [np.abs(f) for f in funcs].max()

I would use `max(-funcs.min(), funcs.max())`, stays vectorized and makes no copies

larsoner:
Also if you rebase it should fix the CIs

mne/minimum_norm/resolution_matrix.py
```diff
@@ -201,25 +201,21 @@ def _summarise_psf_ctf(funcs, mode, n_comp, return_pca_vars):
     s_var = None  # only computed for return_pca_vars=True
 
     if mode == 'maxval':  # pick PSF/CTF with maximum absolute value
-
-        absvals = np.abs(funcs).max(axis=0)
+        absvals = np.array([abs(col).max() for col in funcs.T])
```
larsoner, line 204: You can use the same trick here
```suggestion
        absvals = np.maximum(-np.min(funcs, axis=0), np.max(funcs, axis=0))
```

olafhauk:
I don't think these errors come from my half.

larsoner:
Indeed they are unrelated. Good to go from your end @olafhauk ?

examples/inverse/plot_psf_ctf_label_leakage.py
larsoner: You can just pass `time_viewer=False` to each of these calls

mne/minimum_norm/resolution_matrix.py
```diff
@@ -116,20 +248,26 @@ def get_point_spread(resmat, src, idx, norm=False):
...
-        Vertex indices for which PSFs or CTFs to produce.
-    norm : bool
-        Whether to normalise to maximum across all PSFs (default: False).
+    %(pctf_idx)s
+    %(pctf_mode)s
+    %(pctf_n_comp)s
+    %(pctf_norm)s
+    %(pctf_return_pca_vars)s
+    %(verbose)s
+    %(verbose)s
```
larsoner, line 257: Should only be one of these

mne/minimum_norm/tests/test_resolution_matrix.py
```diff
@@ -8,13 +8,15 @@
...
+                           assert_array_equal, assert_allclose)
 
 import mne
 from mne.datasets import testing
 from mne.minimum_norm.resolution_matrix import (make_inverse_resolution_matrix,
                                                 get_cross_talk,
-                                                get_point_spread)
+                                                get_point_spread,
+                                                _vertices_for_get_psf_ctf)
+from mne.utils import run_tests_if_main
```
larsoner, line 19: Please don't add this back in, eventually it should go away (it's not useful really now that we use `pytest`)

olafhauk: I guess I should remove it from test_resolution_metrics.py as well?

larsoner: Yes feel free, I remove them as I come across them when making changes to a given file

olafhauk:
> 
> 
> Indeed they are unrelated. Good to go from your end @olafhauk ?

It's all looking good to me now. Thanks a lot for your help.

larsoner:
Thanks @olafhauk !

Issue #8219, opened 2020-09-08
Parsing dates in NIRx file for non-english locales


rob-luke:
#### Describe the bug
Parsing dates in NIRx files does not work reliably for non English files. 

#### Steps to reproduce
See #7891 and test `test_encoding()` in `test_nirx.py`.

#### Expected results
Date should be read correctly.

#### Actual results
Currently reading of date fails and the date is set to Jan 1st 2000

#### Additional information
To address this issue we need to know the date format for non English files. If non standard month naming is used (French) we need to know the non standard names used. If someone encounters this issue, it would be good to contact the vendor too.

cbrnr:
As discussed in #8283, we should probably set invalid dates to `None`.

Issue #8128, opened 2020-08-18
How to deal with anonymization if one of the dates slips out of range?


hoechenberger:
Working with the `somato` dataset, I ran into an issue related to the fact that the measurement date I'm intending to anonymize deviates from the date in `info['file_id']['secs']` such that, when I reach the desired adjustment for the measurement data, `info['file_id']['secs']` goes out-of-range, even though the measurement date would be within the valid range.

MWE:

```python
import mne
import os.path as op
from datetime import datetime

data_path = mne.datasets.somato.data_path()
raw_fname = op.join(data_path, 'sub-01', 'meg', 'sub-01_task-somato_meg.fif')

raw = mne.io.read_raw_fif(raw_fname, verbose='error')
print(f"Date in info['meas_date']: {raw.info['meas_date']}")
print(f"Date in info['file_id']:   {datetime.fromtimestamp(raw.info['file_id']['secs'])}")

# Works
print('\nAnonymizing with daysback -> 10 years')
daysback = 10 * 365
raw_anon = raw.copy().anonymize(daysback=daysback)
print(f"Date in info['meas_date']: {raw_anon.info['meas_date']}")
print(f"Date in info['file_id']:   {datetime.fromtimestamp(raw_anon.info['file_id']['secs'])}")

# Does not work
print('\nAnonymizing with daysback -> 90 years')
daysback = 90 * 365
raw_anon = raw.copy().anonymize(daysback=daysback)
print(f"Date in info['meas_date']: {raw_anon.info['meas_date']}")
print(f"Date in info['file_id']:   {datetime.fromtimestamp(raw_anon.info['file_id']['secs'])}")
```

Output:
```python
Date in info['meas_date']: 2007-07-05 11:17:11.172243+00:00
Date in info['file_id']:   1970-01-01 01:00:00

Anonymizing with daysback -> 10 years
Date in info['meas_date']: 1997-07-07 11:17:11.172243+00:00
Date in info['file_id']:   1960-01-04 01:35:47

Anonymizing with daysback -> 90 years
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Untitled-2 in 
     21 print('\nAnonymizing with daysback -> 90 years')
     22 daysback = 90 * 365
---> 23 raw_anon = raw.copy().anonymize(daysback=daysback)
     24 print(f"Date in info['meas_date']: {raw_anon.info['meas_date']}")
[...]

PR #8283, opened 2020-09-21
Gracefully handle invalid EDF dates

### LINKED ISSUE #8281: Read EDF files with dummy zero year/month/day
#### Describe the bug
EDF file that has dummy year/month/day of 0's cannot be read by mne.io.read_raw_edf() and exits with error "month must be in 1..12"

#### Expected results
I except the code still be able to read the data in the .edf file, and I would suggest that another dummy date is used instead of the 0's in the file, i propose using 1900-01-01.

Current code in mne/io/edf/edf.py:
```
# Recording ID
        meas_id = {}
        meas_id['recording_id'] = fid.read(80).decode('latin-1').strip(' \x00')

        day, month, year = [int(x) for x in
                            re.findall(r'(\d+)', fid.read(8).decode())]
        hour, minute, sec = [int(x) for x in
                             re.findall(r'(\d+)', fid.read(8).decode())]
        century = 2000 if year < 50 else 1900

```
Proposal (same as above, only an additional if clause):
```
# Recording ID
        meas_id = {}
        meas_id['recording_id'] = fid.read(80).decode('latin-1').strip(' \x00')

        day, month, year = [int(x) for x in
                            re.findall(r'(\d+)', fid.read(8).decode())]
        hour, minute, sec = [int(x) for x in
                             re.findall(r'(\d+)', fid.read(8).decode())]
        century = 2000 if year < 50 else 1900

        if any([day == 0, month == 0, year == 0]):
            year = 1900
            month = 1
            day = 1

```

If you think this is a solution you'd accept, I can also do a Pull request, although I am happy to hand it over to someone else too, as it is just a simple if clause add-on.

Best,
Wolfgang

cbrnr:
Fixes #8281.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -22,7 +22,7 @@ Bugs
 
 - Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)
 
-- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (`#8283 <https://github.com/mne-tools/mne-python/pull/8283>`_)
+- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
 
 
 API changes
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -20,9 +20,9 @@ Enhancements
 Bugs
 ~~~~
 
-- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly, by `Eric Larson`_
+- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 https://github.com/mne-tools/mne-python/pull/8285`_)
 
-- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ ([#8283](https://github.com/mne-tools/mne-python/pull/8283))
+- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (`#8283 https://github.com/mne-tools/mne-python/pull/8283`_)
 
 
 API changes
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -20,9 +20,9 @@ Enhancements
 Bugs
 ~~~~
 
-- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 https://github.com/mne-tools/mne-python/pull/8285`_)
+- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)
 
-- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (`#8283 https://github.com/mne-tools/mne-python/pull/8283`_)
+- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (`#8283 <https://github.com/mne-tools/mne-python/pull/8283>`_)
 
 
 API changes
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -22,6 +22,9 @@ Bugs
 
 - Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly, by `Eric Larson`_
 
+- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ ([#8283](https://github.com/mne-tools/mne-python/pull/8283))
+
+
 API changes
 ~~~~~~~~~~~
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -20,7 +20,7 @@ Enhancements
 Bugs
 ~~~~
 
-- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)
+- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (:gh:`8285`)
 
 - :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
 
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -543,9 +543,8 @@ def _read_edf_header(fname, exclude):
                                  tzinfo=timezone.utc)
         except ValueError:
             warn(f'Invalid date encountered ({year + century:04d}-{month:02d}-'
-                 f'{day:02d} {hour:02d}:{minute:02d}:{sec:02d}), setting it to'
-                 '1900-01-01 00:00:00.')
-            meas_date = datetime(1900, 1, 1, 0, 0, 0, tzinfo=timezone.utc)
+                 f'{day:02d} {hour:02d}:{minute:02d}:{sec:02d}).')
+            meas_date = None
 
         header_nbytes = _edf_str_int(fid.read(8))
 
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -538,8 +538,14 @@ def _read_edf_header(fname, exclude):
         hour, minute, sec = [int(x) for x in
                              re.findall(r'(\d+)', fid.read(8).decode())]
         century = 2000 if year < 50 else 1900
-        meas_date = datetime(year + century, month, day, hour, minute, sec,
-                             tzinfo=timezone.utc)
+        try:
+            meas_date = datetime(year + century, month, day, hour, minute, sec,
+                                 tzinfo=timezone.utc)
+        except ValueError:
+            warn(f'Invalid date encountered ({year + century:04d}-{month:02d}-'
+                 f'{day:02d} {hour:02d}:{minute:02d}:{sec:02d}), setting it to'
+                 '1900-01-01 00:00:00.')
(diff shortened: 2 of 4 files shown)


mne/io/edf/edf.py
```diff
@@ -550,8 +550,14 @@ def _read_edf_header(fname, exclude):
...
-        meas_date = datetime(year + century, month, day, hour, minute, sec,
-                             tzinfo=timezone.utc)
+        try:
+            meas_date = datetime(year + century, month, day, hour, minute, sec,
+                                 tzinfo=timezone.utc)
+        except ValueError:
+            warn(f'Invalid date encountered ({year + century:04d}-{month:02d}-'
+                 f'{day:02d} {hour:02d}:{minute:02d}:{sec:02d}), setting it to'
+                 '1900-01-01 00:00:00.')
+            meas_date = datetime(1900, 1, 1, 0, 0, 0, tzinfo=timezone.utc)
```
agramfort, line 560: we should have a consistent default date across readers.

pb appeared in nirs format recently and anon has already 2 different defaults 

can you have a look?

larsoner: I hope that we're using DATE_NONE for this (so in this case you'd just set it to `None`)

cbrnr: We're not. Should we? `DATE_NONE = (0, 2 ** 31 - 1)`, I'm not sure how the various readers or more importantly users would handle this value.

Currently, we have #8128 discussing an issue with anonymization where we have 1970-01-01, 2000-01-01, and the BIDS recommendation 1920-01-01. For NIRx, we have #8219 where invalid dates are set to 2000-01-01.

agramfort:
we just need to pick one. No preference.

>

cbrnr:
Is this some known issue (two jobs fail)?

```
  Cloning https://github.com/larsoner/python-pytest-harvest (to revision doctest) to /tmp/pip-req-build-b8_nwlwr
  WARNING: Did not find branch or tag 'doctest', assuming revision or ref.
ERROR: Command errored out with exit status 1: git checkout -q doctest Check the logs for full command output.
```

larsoner:
> Is this some known issue (two jobs fail)?

Fixed in `master`, will restart jobs

mne/io/edf/edf.py
larsoner: >  I'm not sure how the various readers or more importantly users would handle this value.

Users would just see `info['meas_date'] = None`, and this is all you would do in this part of the code. The idea of `DATE_NONE` having these weird int values is just an I/O round-trip FIF detail.

larsoner: (to me this is probably the correct thing to use if we don't know the meas date, it's our default for example in `create_info`)

cbrnr: Makes sense, I'll update the PR and we should probably do the same thing in other related PRs.

wolfgang-ganglberger: Hi, I opened the #8281 datetime 'bug'. 
From user's side, `info['meas_date'] = None` makes sense and agreed that it probably is the most correct thing to output. I suggested the dummy date because I have seen similar corrections in other parts of the code. 
Thank you for working on this issue so fast, really great!

cbrnr:
Azure jobs fail because they can't create the temp file. Is there a recommended mechanism that works?

agramfort:
@cbrnr try using our _TempDir() function?

cbrnr:
Thanks, this should fix the problem. FWIW, the problem with `tempfile.NamedTemporaryFile` was that I opened it for writing and in the same block I wanted to read that file. This works on Unix, but not on Windows (and I think it is ugly anyway).

mne/io/edf/edf.py
cbrnr: Thanks @wolfgang-ganglberger, the test should now be fixed so this change will be in the latest dev version soon.

agramfort (approved):
@cbrnr can you just add an entry in latest.inc ?

cbrnr:
> can you just add an entry in latest.inc ?

Sure. Have we already converged on a new format? I think we agreed to include the GitHub PR number, so my entry would look like this:

```
- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates (by `Clemens Brunner`_, [#8283](https://github.com/mne-tools/mne-python/pull/8283))
```

If you agree I could change the only other entry by @larsoner accordingly.

agramfort:
remember we use rst and not md. I am fine adding url to the PR

>

cbrnr:
> remember we use rst and not md. I am fine adding url to the PR

Right, so this would be:

```
- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates (by `Clemens Brunner`_, `#8283 https://github.com/mne-tools/mne-python/pull/8283`_)
```

agramfort:
LGTM !

>

doc/changes/latest.inc
```diff
@@ -20,7 +20,10 @@ Enhancements
 Bugs
 ~~~~
 
-- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly, by `Eric Larson`_
+- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)
+
+- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (`#8283 <https://github.com/mne-tools/mne-python/pull/8283>`_)
```
hoechenberger, line 25: ❤️❤️❤️ @ PR numbers.

hoechenberger (approved):

doc/changes/latest.inc
cbrnr: 😄  - @hoechenberger could you take care of backporting this PR after it is merged?

hoechenberger: Wow am I the backport guy now? 🧐 
Just kidding, will do, no Problem! 😉

larsoner: In our docs you can just do:
```
:gh:`8283`
```
and it should produce the link above because of

https://github.com/mne-tools/mne-python/blob/master/doc/sphinxext/gh_substitutions.py

larsoner: (see for example the bottom of entries in https://github.com/mne-tools/mne-python/blob/master/doc/overview/roadmap.rst#clustering-statistics-api and the resulting output https://mne.tools/dev/overview/roadmap.html

larsoner:
```suggestion
- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)

- :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
```

doc/changes/latest.inc
```diff
@@ -20,7 +20,10 @@ Enhancements
 Bugs
 ~~~~
 
-- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly, by `Eric Larson`_
+- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (`#8285 <https://github.com/mne-tools/mne-python/pull/8285>`_)
```
larsoner, line 23:
```suggestion
- Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (:gh:`8285`)
```

larsoner (commented):
And if we really want the `#` in front, you or I could update the substitution code easily to use that instead of the `gh-` lead-in. `gh-` was chosen as the lead-in to match SciPy's documentation of github links, but we could change it, especially since github-changelog-generator uses the `#` terminology (and so do we).

doc/changes/latest.inc
hoechenberger: > In our docs you can just do:
> 
> ```
> 
> :gh:`8283`
> 
> ```
> 
> and it should produce the link above because of
> 
> 
> 
> https://github.com/mne-tools/mne-python/blob/master/doc/sphinxext/gh_substitutions.py

Mind = blown!

🤯

@sappelhoff you seeing this stuff???? ****WANT****

cbrnr: Nice, thanks @larsoner for pointing out this gem! And thanks @hoechenberger for backporting, I do like the title "backporting guy" though 😆 !

cbrnr:
> And if we really want the # in front, you or I could update the substitution code easily to use that instead of the gh- lead-in.

I think this would be a good idea. Where do I change this?

cbrnr:
Found it, let's see if this works.

hoechenberger:
> > And if we really want the # in front, you or I could update the substitution code easily to use that instead of the gh- lead-in.
> 
> I think this would be a good idea. Where do I change this?

Can we have "GH-XXXX"? Seems more generic and, should we ever move to another platform, will help us distinguish

cbrnr:
I'd prefer #XXX because it looks more familiar and other platforms (e.g. GitLab) also use this notation. Actually, the platform is not relevant for users, so showing GH-XXX might raise questions why GH is important, and should we switch to GitLab using GL-XXX would be similar. Instead, if we use #XXX we don't include the specific platform at all.

cbrnr:
@larsoner could it be that the [current substitution mechanism](https://github.com/mne-tools/mne-python/blob/af047ec0bba3ea61d1765075628c555328a6341f/doc/sphinxext/gh_substitutions.py#L11) only works for issues?

larsoner:
See comment a few lines down:
```
 # issue/PR mode (issues/PR-num will redirect to pull/PR-num)
```
and try it:

https://github.com/mne-tools/mne-python/issues/8283

cbrnr:
Sorry I don't get it. I'd like to link to a PR, how could the substitution distinguish between issues and PRs if both use identical (but separate) ranges starting from 1? We already have some issue links in our roadmap, so we can't just change everything to point to PRs.

larsoner:
> Sorry I don't get it. I'd like to link to a PR, how could the substitution distinguish between issues and PRs if both use identical (but separate) ranges starting from 1

They are not separate really. They don't overlap. Try this link and see where it goes:

`https://github.com/mne-tools/mne-python/issues/8283`

larsoner:
Or directly try the output links

https://22434-1301584-gh.circle-artifacts.com/0/dev/whats_new.html

cbrnr:
> They are not separate really. They don't overlap.

🤯  TIL!

cbrnr:
All green 🚀 !

larsoner:
> @cbrnr try using our _TempDir() function?

IMO `_TempDir` use in tests should be discouraged, the pytest `tmpdir` fixture is more standard and cleaner. We can fix this later though

larsoner:
Thanks @cbrnr

larsoner:
@hoechenberger I'll backport this at the same time as #8287

PR #8287, opened 2020-09-22
BUG: Fix version on landing page

larsoner:
@rob-luke pointed out that our stable page lists 0.21.dev0 as the version on the right:

Screenshot from 2020-09-22 08-39-59

This refactors the landing page to use the `|version|` substitution. It also uses more built-in RST to render rather than quite as much `.. raw:: html` which actually makes things a bit cleaner I think (e.g., make our docs slightly better if we ever want to build latex).

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -508,21 +508,40 @@ img.institution {
   max-height: 42px;
   max-width: 52px;
 }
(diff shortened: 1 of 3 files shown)


larsoner:
I'll merge and backport

PR #8295, opened 2020-09-23
WIP: Plot Evoked in Report via joint_plot(), not separate time traces and topomaps

hoechenberger:

Do not plot time traces and topomaps separately anymore. The joint plots select the GFP peaks for plotting the topomaps, which is a much more useful visualization than what we currently have in `master`.

#### Additional information
I'm leaving the old code in the file (commented out) so we can easily repurpose it later on should we feel like it.

Todo:
- [ ] Touch tutorial so we can see rendered results
- [ ] Ensure non-MEG/EEG channel types are plotted too

### DIFF
--- a/mne/report.py
+++ b/mne/report.py
@@ -1917,7 +1917,6 @@ class Report(object):
 
         html = []
         for ei, ev in enumerate(evokeds):
-            global_id = self._get_id()
             kwargs = dict(show=False)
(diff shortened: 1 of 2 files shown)


hoechenberger:
- [ ] Touch tutorial so we can see rendered results
- [ ] Ensure non-MEG/EEG channel types are plotted too

PR #8296, opened 2020-09-23
WIP: Enable github codespaces

rob-luke:
As discussed with @larsoner at https://github.com/mne-tools/mne-nirs/pull/165#issuecomment-696710402. This enables a [GitHub codespaces](https://github.com/features/codespaces) python environment with mne.

#### Additional information
I have found this useful for editing code and running tests over at https://github.com/mne-tools/mne-nirs
I made a little gif of how it works below.

ezgif-3-746e1704a2c7

I just copied the config over here. It may need specific tweaks to work optimally with MNE. But I thought I would gauge interest before optimising. This is more a coding environment, whereas binder is a notebook. This may be a way to lower the barrier for new coders (or it may just be another complication with more acronyms)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,27 @@
+FROM continuumio/miniconda3
+
+# Options for common setup script
+ARG INSTALL_ZSH="true"
+ARG UPGRADE_PACKAGES="false"
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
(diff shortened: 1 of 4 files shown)


agramfort:
I don't see the codespaces button in https://github.com/mne-tools/mne-nirs for me

rob-luke:
I just checked the codespaces website from another device, and it seems it's on a limited pre-release (I hit the sign me up button and got access). We should hold this discussion until it's public

cbrnr:
Press the . key or replace `.com` with `.dev` in the URL.

Issue #8044, opened 2020-07-22
Undesireable output of the to_data_frame 


balandongiv:
#### Describe the bug
The idea is to produce the output with respect to set `time_format` parameter  under the `to_data_frame `method. However, setting the time_format to either  `None, 'timedelta', 'datetime' and 'ms'` produce an incremental plus 1 integer instead of the scaled time (e.g., 0, 0.001,0.002) for a signal that sampled in 1000Hz.

#### Steps to reproduce

* filename='S17_5.mff' # The mff file is accessible via download link provided below
```
raw = mne.io.read_raw_egi(filename, preload=True)
print(raw.to_data_frame(picks=['E1']))
```

Return the following

```
Reading 0 ... 300936  =      0.000 ...   300.936 secs...
 	time            E1
0            0  15372.021811
1            1  15375.733752
2            2  15371.825459
3            3  15367.826391
4            4  15363.235307

```

```
raw = mne.io.read_raw_egi(filename, preload=True)
print(raw.to_data_frame(picks=['E1'],time_format='ms'))
```

Return the following

```
          time            E1
0            0  15372.021811
1            1  15375.733752
2            2  15371.825459
3            3  15367.826391
4            4  15363.235307
```

```

raw = mne.io.read_raw_egi(filename, preload=True)
print(raw.to_data_frame(picks=['E1'],time_format='timedelta'))
```

Return the following
```
                  time            E1
0             00:00:00  15372.021811
1      00:00:00.001000  15375.733752
2      00:00:00.002000  15371.825459
3      00:00:00.003000  15367.826391
4      00:00:00.004000  15363.235307
```

#### Expected results

At least, the column time should look

`[0.    0.001 0.002 0.003 0.004 0.005 0.006 0.007 0.008 ]`

#### Additional information
System:

Windows 10
anaconda3
mne-0.21.dev0 
Python 3.7
[Link to download S17_5.mff ](https://drive.google.com/drive/folders/15lAjuEPC2iSbBKBSy8y11Q0Fs24cnSd1?usp=sharing)

hoechenberger:
So what you you're looking for is, essentially, support for `time_format='s'`?

drammock:
@hoechenberger is correct that this is not a bug. The problem here is that `time_format=None` is not the default, so you have to explicitly pass `None` in order to get the equivalent of `time_format='s'`.  Eventually we plan to change this to default to `None`, but it will need to go through a deprecation cycle.

```python
import os
import mne
[...]

PR #7206, opened 2020-01-14
MRG, FIX, ENH: overhaul to_data_frame method

drammock:
closes #7179 

- [x] minor refactor of `_check_pandas_index_arguments`
- [x] fix `to_data_frame()` docstring re: tuples vs lists
- [x] make `index=None` work as described
- [x] fix logic when setting title of columns
- [x] provide different options for time conversion: keep as float, or convert to integer milliseconds, pd.Timedelta, or (raw only) pd.Timestamp
- [x] make the datetime/timedelta change work with `long_format=True`
- [x] improve tests

cc @jona-sassenhagen @dengemann @choldgraf @larsoner @agramfort

PR #8298, opened 2020-09-23
MRG, API: prep for to_data_frame default argument change

drammock:
crossref #7206 (especially [this comment](https://github.com/mne-tools/mne-python/pull/7206#issuecomment-581697291))

The goal is to change default from `'ms'` to `None`. In this PR I'm only changing docstring, not raising a warning if the current default is passed.  I suppose I could change the default to `None` and have `None` temporarily mean `'ms'`, and add temporarily a valid argument `'s'` to get the behavior currently intended by `None`, and raise a warning if the value is `None` rather than `'s'` or `'ms'`?  But at least some people have learned that explicitly passing `None` is what they should do (e.g., #8044) so IMO raising a warning in that way would cause more confusion than it might avert.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -1596,7 +1596,8 @@ time_format : str | None
     remain as float values in seconds. If ``'ms'``, time values will be rounded
     to the nearest millisecond and converted to integers. If ``'timedelta'``,
     time values will be converted to :class:`pandas.Timedelta` values. {}
-    Defaults to ``'ms'``.
+    Default is ``'ms'`` in version 0.22, and will change to ``None`` in
+    version 0.23.
 """
 raw_tf = ("If ``'datetime'``, time values will be converted to "
           ":class:`pandas.Timestamp` values, relative to "


agramfort:
MRG?

>

drammock:
> MRG?

Yes, sorry, forgot to add that.

agramfort (approved):
@larsoner merge if happy

dengemann:
@drammock any phantasy how that might interact downstream with R stuff?

drammock:
> @drammock any phantasy how that might interact downstream with R stuff?

some MNE-R tests will definitely start failing once the default changes, which won't happen in MNE-Python stable for ~1 year.  Hopefully before then one of us will take a fresh look at MNE-R and its test suite...  the update to R 4.0 changed the `stringsAsFactors` default, which probably already breaks some tests.

dengemann:
Ok sounds good! I want to defnitely spend a few days on this before the end of the year. I'll ping you once I have an idea when that could  be. Perhaps we can coordinate a bit.

larsoner:
Thanks @drammock

PR #8306, opened 2020-09-25
MRG: Fix bug in nirs merge data

rob-luke:
The `_merge_nirs_data` function is used in topo plotting and averages channels that are close together for a nicer display. After a channel is merged it is then removed from the data. If the same channel was merged twice, then it was removed twice, but the second time it was removing legitimate data. This would then mean the size of data and pos was incorrect and a nice error was thrown in the following topo function (which made it easy for met to find this bug).  Here I just ensure each channel is only removed once.

#### Additional information
I think this is a candidate for back port. Whats involved in that?

I found this error because I am considering proposing a different distance for merging channels, but that's an enhancement and I will put in another PR. But it means others aren't likely to hit this bug unless they are playing with internal functions.

### DIFF
--- a/mne/channels/layout.py
+++ b/mne/channels/layout.py
@@ -977,6 +977,7 @@ def _merge_nirs_data(data, merged_names):
                 indices = np.append(indices, merged_names.index(sub_ch))
             data[idx] = np.mean(data[np.append(idx, indices)], axis=0)
             to_remove = np.append(to_remove, indices)
+    to_remove = np.unique(to_remove)
     for rem in sorted(to_remove, reverse=True):
         del merged_names[rem]
         data = np.delete(data, rem, 0)


agramfort:
is the codecov issue legit?

rob-luke:
Yeah we aren't hitting the lines. I just checked locally that if we make #8037 we will hit the lines.

rob-luke:
Im trying to mock hit the lines now. Real data would be better, but cant get in lab.

agramfort:
yes we can merge. I assume adding a test is difficult?

>

rob-luke:
I will try for a bit more to make a test. I will ping you when ready (or if I give up)

rob-luke:
Sorry but I cant make it hit the line with this 1e-10 parameter in #8037. I am very confident this is bug though. Happy to merge if you are

agramfort:
let's go then

>

drammock:
> I think this is a candidate for back port. Whats involved in that?

Once this PR is merged, fetch latest upstream and update your local copy of master branch. Then create local branch that tracks `upstream:maint/0.21`, cherry-pick the relevant commit, and push to `maint/0.21`.  If you don't have push access to upstream, you can do it as a PR (push local `maint/0.21` to your fork, and open a PR from `rob-luke/mne-python:maint/0.21` into `mne-tools/mne-python:maint/0.21`)

larsoner:
Backported

rob-luke:
Thanks @agramfort @larsoner @drammock

PR #8307, opened 2020-09-25
WIP: Use greater distance when defining overlapping nirs channels

rob-luke:
When you have a NIRS recording with lots of very close channels the topomap looks bad. We had code to handle overlapping NIRS channels, but it was for exactly overlapping channels. This extends what's considered overlapping channels to 1.5cm. In NIRS experiments the source and detectors are ideally placed 3cm apart, so I have taken half this value and it looks nice in my figures.

Example of topoplot with existing value (note the small black dots representing channels in little clusters)...

image

And with the new value

image

#### Additional information
1.5 cm seems to play nicely with the topographic plotting too. But Im open to other suggested values or perhaps a totally different approach?

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -147,9 +147,9 @@ def _average_fnirs_overlaps(info, ch_type, sphere):
     # Channels to be excluded from picks, as will be removed after merging
     channels_to_exclude = list()
 
-    if len(locs3d) > 1 and np.min(dist) < 1e-10:
+    if len(locs3d) > 1 and np.min(dist) < 0.015:
 
-        overlapping_mask = np.triu(squareform(dist < 1e-10))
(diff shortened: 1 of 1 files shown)


rob-luke:
The tests for this will fail due to the bug in #8306. I will rebase once that's addressed.

larsoner:
I wonder if a more general approach would be to allow smoothing the topomaps. Can you see if tweaking these lines to use Rbf interpolation with some suitable `smooth > 0`? Basically I think we could add `smooth=0` as a default, and values > 0 switch to using Rbf with the given `smooth` value.

larsoner:
https://github.com/mne-tools/mne-python/blob/master/mne/viz/topomap.py#L609-L615

rob-luke:
Great, I will play around and report back.

PR #8308, opened 2020-09-25
apply patch from Christian on ICA reader

### LINKED ISSUE #7657: The computation of the mixing matrix for ICA imported from EEGLAB ICA is broken
#### Describe the bug

Commit https://github.com/mne-tools/mne-python/pull/6938/commits/23c6aa9ac85411afba3dd3687787b1c112226546 introduced a bug in the PR https://github.com/mne-tools/mne-python/pull/6938 as described in the comment https://github.com/mne-tools/mne-python/pull/6938/commits/23c6aa9ac85411afba3dd3687787b1c112226546#r413113709

#### Steps to reproduce

Using an ICA with a number of components smaller than the number of channels, I get the following components with the code as patched with this commit:

image

and using seaborn clustermap on the correlation matrix between these components, I get:

image

#### Expected results

Now, if I revert this change, I get something more reasonable:

image

image

I unfortunately don't have the "gold truth" to compare with for the components at this time, but it already seems much more reasonable to me...

#### Proposal

I spotted that the issue here is the change from 

```
    ica.mixing_matrix_ = np.linalg.pinv(ica.pca_components_) @ eeg.icawinv
```
to
```
    ica._update_mixing_matrix()  # ==> self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)
```

(Also, shouldn't "u * s" be "u @ s"? It does not seem to make any difference in my specific use case, but conceptually, it should be matrix multiplication, not a broadcasting.)

Apparently, the way EEGLAB defined its variable `linalg.pinv(self.unmixing_matrix_) != np.linalg.pinv(ica.pca_components_) @ eeg.icawinv`. I find it very hard to align the interfaces of MNE and EEGLAB because the approaches to ICA are a bit different, the naming is different, and so on, which makes it quite confusing. For example, how `u @ s` relates exactly to EEGLAB `icawinv` is unclear to me at this point. 

This seems to work:
```
    if n_components < n_ch:
        u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
        ica.unmixing_matrix_ = u @ s
        ica.pca_components_ = v
    else:
        [...]
    ica.mixing_matrix_ = np.linalg.pinv(ica.pca_components_) @ eeg.icawinv
```
The down side is that calling `ica._update_mixing_matrix()` at a later point would introduce a bug. So, an additional (ugly) fix is to store somewhere the value of `eeg.icawinv` and make  `ica._update_mixing_matrix()` of `ica.method == "imported_eeglab" `, which would looks like:

```
def _update_mixing_matrix(self):
    if self.method == "imported_eeglab":
        self.mixing_matrix_ = np.linalg.pinv(self.pca_components_) @ self._eeglab_icawinv         
    else:
        self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)

def read_ica_eeglab(fname):
    [...]
    if n_components < n_ch:
        u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
        ica.unmixing_matrix_ = u @ s
        ica.pca_components_ = v
    else:
        [...]
    ica._eeglab_icawinv = eeg.icawinv  # Not great to add this attribute...
    ica._update_mixing_matrix()
```

FYI, simply doing
```
    if n_components < n_ch:
        u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
        # replace `ica.unmixing_matrix_ = u @ s` by
        ica.unmixing_matrix_ = np.linalg.pinv(np.linalg.pinv(ica.pca_components_) @ eeg.icawinv)
        ica.pca_components_ = v
    else:
        [...]
    ica._update_mixing_matrix() 
```
does not work. Don't ask me to explain why... Anyway, I can submit a PR with the proposed ugly changes and we can see if we can prettify it somehow... although, at this point, I am at a lost for a clean alignment of MNE and EEGLAB ICA interfaces, so I am not sure I can improve much on this.  In any case, I would argue that a working ugly code is preferable to an elegant broken one...

agramfort:
closes #7657

@christian-oreilly @marcpabst did I get it right?

PR:

Screenshot 2020-09-25 at 15 50 20

master:

Screenshot 2020-09-25 at 15 44 10

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -707,7 +707,11 @@ class ICA(ContainsMixin):
         self.current_fit = fit_type
 
     def _update_mixing_matrix(self):
-        self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)
+        if self.method == "imported_eeglab":
+            self.mixing_matrix_ = \
+                np.linalg.pinv(self.pca_components_) @ self._eeglab_icawinv
+        else:
+            self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)
 
     def _update_ica_names(self):
         """Update ICA names when n_components_ is set."""
@@ -2662,10 +2666,11 @@ def read_ica_eeglab(fname):
         # sphere = eye(urchans), so let's use SVD to get our square
         # weights matrix (u * s) and our PCA vectors (v) back
         u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
-        ica.unmixing_matrix_ = u * s
+        ica.unmixing_matrix_ = u @ s
         ica.pca_components_ = v
     else:
         ica.unmixing_matrix_ = eeg.icaweights
         ica.pca_components_ = eeg.icasphere
+    ica._eeglab_icawinv = eeg.icawinv
     ica._update_mixing_matrix()
     return ica


mne/preprocessing/ica.py
```diff
@@ -2662,10 +2666,11 @@ def read_ica_eeglab(fname):
         # sphere = eye(urchans), so let's use SVD to get our square
         # weights matrix (u * s) and our PCA vectors (v) back
         u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
-        ica.unmixing_matrix_ = u * s
+        ica.unmixing_matrix_ = u @ s
```
larsoner, line 2669: `s` is 1D so after this operation `unmixing_matrix_` will be 1D. I think this can't be correct.

Maybe you mean `u @ np.diag(s)`, but this is the same as `u * s`.

marcpabst:
Looks good to me. This is the "ground truth" from EEGLAB:
image

marcpabst:
I should add that with this file `n_components == n_ch:`.

agramfort:
the patch breaks the tests.

>

larsoner:
In particular:
```
mne/preprocessing/tests/test_ica.py:1230: in test_read_ica_eeglab
    ica = read_ica_eeglab(fname)
mne/preprocessing/ica.py:2675: in read_ica_eeglab
    ica._update_mixing_matrix()
mne/preprocessing/ica.py:712: in _update_mixing_matrix
    np.linalg.pinv(self.pca_components_) @ self._eeglab_icawinv
E   ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 32 is different from 25)
```

mne/preprocessing/ica.py
larsoner: For more info see previous comment 

https://github.com/mne-tools/mne-python/issues/7657#issuecomment-618040530

PR #8313, opened 2020-09-28
WIP, ENH: Be verbose if fitting ICA on rank-deficient data

hoechenberger:
@agramfort and I encountered a weird, seemingly inexplicable issue when using ICA for EOG and ECG artifact correction. Turned out – the only reason that lead us to "waste" several hours was simply that the input data was rank-deficient, because it had 3 SSP projection vectors in `.info`.

I have therefore opted to enhance `ICA.fit()` such that it would (hopefully) detect such cases in the future, and voice its concerns or even fail hard.

I have added a `check_rank=True` kwarg to `ICA.fit()`. If `True` (default), the user will get an extensive warning if the rank of the data is expected to be NOT full even after the PCA step. The error message contains concrete directions on what to do.

My idea is that this message, for now, remains a warning, and will change to an `Exception` with 0.23 the latest. In fact, I'd prefer if we immediately made it an exception, because I think we should fail hard in this case; but I assume other devs would be against this, therefore we should probably do a deprecation cycle.

If the user passes `check_rank=False`, we will still produce a warning, but it won't be as extensive. My reasoning behind this is that it helps ensure that one doesn't accidentally leave `check_rank =False` in the code after testing something and forgetting about it. Once `check_rank=True` raises an exception, the behavior difference between `check_rank=True` and `check_rank=False` will become more clear.

Todo:
- [ ] Amend ICA tutorial
- [ ] Amend changelog

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -60,6 +60,7 @@ from ..filter import filter_data
 from .bads import _find_outliers
 from .ctps_ import ctps
 from ..io.pick import pick_channels_regexp
+from ..rank import _compute_rank_int
 
 __all__ = ('ICA', 'ica_find_ecg_events', 'ica_find_eog_events',
            'get_score_funcs', 'read_ica', 'read_ica_eeglab')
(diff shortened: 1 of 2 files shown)


hoechenberger:
Voice call with @agramfort:

- don't introduce a new parameter to `ICA.fit`, but offer more options for `ICA.max_pca_components`, e.g. `'rank'` or `'auto'` instead
- maybe deprecate `max_pca_components=None`

Issue #8315, opened 2020-09-29
Accessing data after applying gradient compensation & picking doesn't work


hoechenberger:
#### Steps to reproduce
```python
import os.path as op
from mne.io import read_raw_fif

base_dir = op.join('/Users/hoechenberger/Development/mne-python/mne', 'io',
                   'tests', 'data')
raw_ctf_fname = op.join(base_dir, 'test_ctf_raw.fif')

raw = read_raw_fif(raw_ctf_fname, preload=False)
raw.apply_gradient_compensation(3)
raw.copy().pick_types(meg=True).get_data()
```

#### Expected results
```
Opening raw data file /Users/hoechenberger/Development/mne-python/mne/io/tests/data/test_ctf_raw.fif...
    Read 5 compensation matrices
    Range : 24000 ... 31200 =     10.000 ...    13.000 secs
Ready.
Current compensation grade : 0
Reading 0 ... 7200  =      0.000 ...     3.000 secs...
Compensator constructed to change 0 -> 3
Applying compensator to loaded data
```

#### Actual results
```python
Opening raw data file /Users/hoechenberger/Development/mne-python/mne/io/tests/data/test_ctf_raw.fif...
    Read 5 compensation matrices
    Range : 24000 ... 31200 =     10.000 ...    13.000 secs
Ready.
Current compensation grade : 0
Compensator constructed to change 0 -> 3
Traceback (most recent call last):
  File "/private/tmp/mwe.py", line 10, in <module>
    raw.copy().pick_types(meg=True).get_data()
  File "<decorator-gen-166>", line 21, in get_data
  File "/Users/hoechenberger/Development/mne-python/mne/io/base.py", line 823, in get_data
    data, times = self[picks, start:stop]
  File "/Users/hoechenberger/Development/mne-python/mne/io/base.py", line 772, in __getitem__
    data = self._read_segment(start=start, stop=stop, sel=sel,
  File "<decorator-gen-163>", line 21, in _read_segment
  File "/Users/hoechenberger/Development/mne-python/mne/io/base.py", line 375, in _read_segment
    mult = self._comp[idx] * cals
ValueError: operands could not be broadcast together with shapes (304,309) (1,304) 
```

#### Additional information
If the data is preloaded:
```python
raw = read_raw_fif(raw_ctf_fname, preload=True)
```
the problem does not occur.

Also if I skip the `pick_types()` it does work.

NumPy is 1.19.1.

larsoner:
[...]

PR #8316, opened 2020-09-29
WIP, ENH: Avoid ICA on rank-deficient data

hoechenberger:
- `ICA.max_pca_components` can now be set to `rank` to limit the number of components to the rank the data.
- This is also the new default.
- The previous default value, ``None``,   is still an accepted, but deprecated.
- ``None``, too, now   dictates to use the rank of the data. Previously, it implied to   use the number of channels, which caused problems with   rank-deficient data.
- If requested number of ICA or PCA components exceeds the rank of   the data, we will now raise an exception.
- `compute_rank()` and `_estimate_rank_meeg_signals()` have gained   a `ref_meg=False` kwarg to allow for rank calculation to include MEG reference  channels. This was required to make some tests work.

Implementing this helped discover a few existing test cases that never converged because they fed rank-deficient data to ICA.

Closes #8313.

**I need help:**

- `mne/viz/tests/test_ica.py::test_plot_ica_overlay` doesn't pass because of #8315 
- `mne/preprocessing/tests/test_ica.py::test_ica_full_data_recovery` doesn't pass because the difference between actual and expected test results is too large. It's still rather small but larger than the predefined limits. I have no idea why that is the case, and whether it might be safe to just make the test more lenient or not.

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -60,6 +60,7 @@ from ..filter import filter_data
 from .bads import _find_outliers
 from .ctps_ import ctps
 from ..io.pick import pick_channels_regexp
+from ..rank import _compute_rank_int
 
 __all__ = ('ICA', 'ica_find_ecg_events', 'ica_find_eog_events',
            'get_score_funcs', 'read_ica', 'read_ica_eeglab')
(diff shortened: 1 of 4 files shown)


hoechenberger:
Another call with @agramfort … will try yet another approach. Stay tuned ;)

PR #8323, opened 2020-09-30
MRG, ENH: Speed up time_delaying_ridge using Numba

larsoner:
I remembered that the edge correction was actually a painfully slow step in the Fourier method. Using Numba reduces the total example time from 7.2 to 6.1 sec on my system.


agramfort (approved):

agramfort:
thx @larsoner

PR #8326, opened 2020-09-30
MRG, FIX: Test round-trip, fix EEGLAB ICA

### LINKED ISSUE #7657: The computation of the mixing matrix for ICA imported from EEGLAB ICA is broken

### LINKED ISSUE #8325: BUG?: ICA can blow up components
@hoechenberger not 100% sure this is related to #8313 and #8316 since those seem to have to do with ICA taking too long rather than doing something totally wrong. To me this code should work / be a no-op:
```
import numpy as np
import mne

data = np.random.RandomState(0).randn(10, 100000)
info = mne.create_info(10, 1000., 'eeg')
raw.set_eeg_reference()
assert np.linalg.matrix_rank(raw.get_data()) == 9

ica = mne.preprocessing.ICA(
    max_pca_components=10, n_pca_components=10, max_iter=1, verbose=True)
raw_new = ica.fit(raw).apply(raw.copy())
np.testing.assert_allclose(raw.get_data(), raw_new.get_data())
```
gives an `AssertionError`:
```
Mismatched elements: 1000000 / 1000000 (100%)
Max absolute difference: 4.53435389
Max relative difference: 70047.35032413
 x: array([[ 1.481445,  0.391694,  1.081697, ...,  0.370815, -1.556122,
        -1.275782],
       [-0.766405,  1.279594, -0.02692 , ...,  0.312448, -1.409486,...
 y: array([[ 6.455399e-05,  6.455399e-05,  6.455399e-05, ...,  6.455399e-05,
         6.455399e-05,  6.455399e-05],
       [ 3.581095e-03,  3.581095e-03,  3.581095e-03, ...,  3.581095e-03,...
```
It seems like this should work even if you get the number of components wrong. It suggests that something is wrong with the mixing/unmixing step that when the number of components equals the number of channels you don't get an identity transform here.

It also fails for `max_pca_components=10, n_pca_components=9` but passes for any value of `max_pca_components` (8, 9, 10) so long as `max_pca_components=9`.

If I remove this call:
```
        self.unmixing_matrix_ /= np.sqrt(
            self.pca_explained_variance_[sel])[None, :]  # whitening
```
Then things work okay.

larsoner:
Closes #8325 
Closes #8308
Closes #7657

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -776,7 +776,6 @@ class ICA(ContainsMixin):
         self.current_fit = fit_type
 
     def _update_mixing_matrix(self):
-        # stabilize this by using row normalization
         self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)
 
     def _update_ica_names(self):
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -769,12 +769,14 @@ class ICA(ContainsMixin):
             self.n_iter_ = n_iter + 1  # picard() starts counting at 0
             del _, n_iter
         assert self.unmixing_matrix_.shape == (self.n_components_,) * 2
-        self.unmixing_matrix_ /= np.sqrt(
-            self.pca_explained_variance_[sel])[None, :]  # whitening
+        # norms = np.sqrt(self.pca_explained_variance_[:self.n_components_])
+        # norms[norms == 0] = 1.
+        # self.unmixing_matrix_ /= norms  # whitening
         self._update_mixing_matrix()
         self.current_fit = fit_type
 
     def _update_mixing_matrix(self):
+        # stabilize this by using row normalization
         self.mixing_matrix_ = linalg.pinv(self.unmixing_matrix_)
 
     def _update_ica_names(self):
@@ -2730,10 +2732,12 @@ def read_ica_eeglab(fname):
         # sphere = eye(urchans), so let's use SVD to get our square
         # weights matrix (u * s) and our PCA vectors (v) back
         u, s, v = _safe_svd(eeg.icaweights, full_matrices=False)
+        ica.pca_explained_variance_ = s * s
         ica.unmixing_matrix_ = u * s
         ica.pca_components_ = v
     else:
         ica.unmixing_matrix_ = eeg.icaweights
         ica.pca_components_ = eeg.icasphere
+        ica.pca_explained_variance_ = np.ones(ica.unmixing_matrix_.shape[1])
     ica._update_mixing_matrix()
     return ica
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -123,7 +123,7 @@ def _check_for_unsupported_ica_channels(picks, info, allow_ref_meg=False):
 
 @fill_doc
 class ICA(ContainsMixin):
-    u"""M/EEG signal decomposition using Independent Component Analysis (ICA).
+    u"""Data decomposition using Independent Component Analysis (ICA).
 
     This object estimates independent components from :class:`mne.io.Raw`,
     :class:`mne.Epochs`, or :class:`mne.Evoked` objects. Components can
@@ -139,11 +139,11 @@ class ICA(ContainsMixin):
         Number of principal components (from the pre-whitening PCA step) that
         are passed to the ICA algorithm during fitting. If :class:`int`, must
         not be larger than ``max_pca_components``. If :class:`float` between 0
-        and 1, the number of components with cumulative explained variance less
-        than ``n_components`` will be used. If ``None``, ``max_pca_components``
-        will be used. Defaults to ``None``; the actual number used when
-        executing the :meth:`ICA.fit` method will be stored in the attribute
-        ``n_components_`` (note the trailing underscore).
+        and 1, the number of components with cumulative explained variance
+        greater than ``n_components`` will be used. If ``None``,
+        ``max_pca_components`` will be used. Defaults to ``None``; the actual
+        number used when executing the :meth:`ICA.fit` method will be stored in
+        the attribute ``n_components_`` (note the trailing underscore).
     max_pca_components : int | float | None
         The number of principal components (from the pre-whitening PCA step)
         that are retained for later use (i.e., for signal reconstruction in
@@ -190,7 +190,7 @@ class ICA(ContainsMixin):
         set additional parameters. Specifically, if you want Extended Infomax,
         set method='infomax' and fit_params=dict(extended=True) (this also
         works for method='picard'). Defaults to 'fastica'. For reference, see
-        [1]_, [2]_, [3]_ and [4]_.
+        :footcite:`Hyvarinen1999,BellSejnowski1995,LeeEtAl1999,AblinEtAl2018`.
     fit_params : dict | None
         Additional parameters passed to the ICA estimator as specified by
         ``method``.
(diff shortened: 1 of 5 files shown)


larsoner:
On this branch I get the same plots as in #8308:

Screenshot from 2020-10-01 14-18-38

@christian-oreilly @marcpabst can you try it on your datasets to see if it makes sense? I was able to unify the code so that no branching is required, which is nice.

mne/preprocessing/tests/test_ica.py
```diff
@@ -143,12 +143,46 @@ def test_ica_simple(method):
...
     amari_distance = np.mean(np.sum(np.abs(transform), axis=1) /
                              np.max(np.abs(transform), axis=1) - 1.)
     assert amari_distance < 0.1
 
 
+@requires_sklearn
+@pytest.mark.parametrize('n_pca_components', (8, 9, 10))
+@pytest.mark.parametrize('max_pca_components', [
+    9,
+    pytest.param(0.9999, marks=pytest.mark.xfail(reason='XXX should work')),
```
larsoner, line 156: @hoechenberger this should pass for `n_pca_components=8` and `9` (and eventually probably error for `10` but it fails currently. This suggests that the `max_pca_components=0.9999` component picking does not work right.

mne/preprocessing/tests/test_ica.py
```diff
@@ -143,22 +150,57 @@ def test_ica_simple(method):
...
     amari_distance = np.mean(np.sum(np.abs(transform), axis=1) /
                              np.max(np.abs(transform), axis=1) - 1.)
     assert amari_distance < 0.1
 
 
+@requires_sklearn
+@pytest.mark.parametrize('n_pca_components', (8, 9, 10))
+@pytest.mark.parametrize('max_pca_components', [
+    9,
+    pytest.param(0.999999, marks=pytest.mark.xfail(reason='XXX should work')),
```
larsoner, line 163: With 10 channels and an avg ref proj this gives 8 components instead of 9; if we use 1. instead, it works sometimes and not others because of numerical precision issues in the PCA.

If we want this to work (and I think we do) then we should change the logic to be "greater than or equal to" the given variance instead of "equal to". I consider this a bugfix. See previous [discussion](https://github.com/mne-tools/mne-python/pull/8321#discussion_r498159641).

hoechenberger: Our current logic is "less than or equal to" :)

hoechenberger: But I agree we should change it. That must be the reason why scikit-learn does it that way too.

larsoner:
Okay this one is ready for review @agramfort @hoechenberger . We should wait to hear from @christian-oreilly and @marcpabst before merge, though

mne/utils/tests/test_numerics.py
```diff
@@ -421,12 +421,13 @@ def test_pca(n_components, whiten):
...
     assert_array_equal(X, X_orig)
     assert_allclose(X_skl, X_mne)
+    assert pca_mne.n_components_ == pca_skl.n_components_
     for key in ('mean_', 'components_',
                 'explained_variance_', 'explained_variance_ratio_'):
         val_skl, val_mne = getattr(pca_skl, key), getattr(pca_mne, key)
         assert_allclose(val_skl, val_mne)
     if isinstance(n_components, float):
-        assert 1 < pca_mne.n_components_ < n_dim
+        assert pca_mne.n_components_ == n_dim - 1
```
hoechenberger, line 430: I'm surprised this is `n_dim - 1`? I would have expected `n_dim` now, with the new adjustments?

hoechenberger (commented):
Added a question, otherwise LGTM.
We need to do the same thing for `n_components`, for sake of consistency.

mne/utils/tests/test_numerics.py
larsoner: This data is rank reduced, see a few lines above the comment:
```
X[:, -1] = np.mean(X[:, :-1], axis=-1)  # true X dim is ndim - 1
```

larsoner:
Argh, I went to update the documentation, now I'm confused why we need `n_components`, `n_pca_components`, and `max_pca_components`. I thought we'd only need two: that the last one determined signal reconstruction, but now it says `n_pca_components` does this. I'll revert my doc changes (other than the refs) and hopefully we can fix it later.

larsoner:
@hoechenberger I'm beginning to wonder if we need `max_pca_componetns` to be float. Really it should be enough to allow `n_components` to be a float I think. The instability of the inversion comes from dividing by `np.sqrt(self.pca_explained_variance_[:self.n_components_])`. So if `n_components` is set appropriately, it seems like there shouldn't be a problem.

hoechenberger:
@larsoner
> @hoechenberger I'm beginning to wonder if we need `max_pca_componetns` to be float. Really it should be enough to allow `n_components` to be a float I think. The instability of the inversion comes from dividing by `np.sqrt(self.pca_explained_variance_[:self.n_components_])`. So if `n_components` is set appropriately, it seems like there shouldn't be a problem.

All `max_pca_components` are submitted to ICA for fitting; of those, only `n_components` are retained.

If it is like you say and we can get rid of the instability of choosing inappropriate `max_pca_components` and produce sane results, then things should be okay… Still I'm a little worried. Also, `max_pca_components` being a float could be just another way to limit how many components to fit ICA on – would be consistent with `n_components` and `n_pca_components`

hoechenberger:
> @hoechenberger I'm beginning to wonder if we need `max_pca_componetns` to be float. Really it should be enough to allow `n_components` to be a float I think. The instability of the inversion comes from dividing by `np.sqrt(self.pca_explained_variance_[:self.n_components_])`. So if `n_components` is set appropriately, it seems like there shouldn't be a problem.

Yet another example of our API being way too complicated. We've had so many issues with this…

hoechenberger:
> Argh, I went to update the documentation, now I'm confused why we need `n_components`, `n_pca_components`, and `max_pca_components`. I thought we'd only need two (`n_pca_components` and `max_pca_components`, and that `max_pca_components` determined signal reconstruction and `n_pca_components` was what was passed to the ICA algorithm, but this is not at all what our docs say. I'll revert my doc changes (other than the refs) and hopefully we can fix it later.

Yet another example of our API being way too complicated. We've had so many issues with this…

larsoner:
Okay let me see if I can write the docs properly

hoechenberger:
Maybe to clarify, my understanding is:

- max_pca_components: PCA components to keep
- n_components: number of ICA components that are retained after fitting ICA with max_pca_components --> the docs say otherwise, so either the docs are wrong or I made a mistake reading the sources
- n_pca_components: number of components (ICA + PCA!!!!) to use for signal reconstruction -> maybe this kwarg is just poorly named

hoechenberger:
(edited my above post a couple times)

larsoner:
> n_components: number of ICA components that are retained after fitting ICA on data with dimensionality reduced to max_pca_components 

The code has this:
```
        sel = slice(0, self.n_components_)
        if self.method == 'fastica':
            from sklearn.decomposition import FastICA
            ica = FastICA(whiten=False, random_state=random_state,
                          **self.fit_params)
            ica.fit(data_transformed[:, sel])
```
So I think the docs are correct -- ICA is fitted on the first `n_components` of the PCA.

`n_pca_components` isn't even used until `apply` so it only has to do with the apply step. I'll add this to the docs.

larsoner:
Now I wonder if using `n_components` as float would have been enough and `max_pca_components` as float is overkill... I'll push a commit that shows setting `n_components` to a float also gives the behavior we want, even with `max_pca_components=n_chan`

hoechenberger:
> So I think the docs are correct -- ICA is fitted on the first `n_components` of the PCA.

Oh ok, I got this mixed up then.

hoechenberger (approved):
LGTM once CI is green

mne/preprocessing/ica.py
```diff
@@ -259,25 +264,55 @@ class ICA(ContainsMixin):
...
+
+    1. Unmixes the data with the ``unmixing_matrix_``.
+    2. Includes ICA components based on ``ica.include`` and ``ica.exclude``.
+    3. Re-mixes the data with ``mixing_matrix_``.
+    4. Restores any data not passed to the ICA algorithm, i.e., the PCA
+       components between ``n_components`` and ``n_pca_components``.
+
+    Thus ``max_pca_components`` and ``n_pca_components`` determine how many
+    PCA components will be kept when reconstructing the data when calling
+    :meth:`apply`. These paremeters can be used for dimensionality reduction of
```
hoechenberger, line 285:
```suggestion
    :meth:`apply`. These parameters can be used for dimensionality reduction of
```

hoechenberger:
Thanks @larsoner! I'll want to try this on some of our data now before we move forward with the API redesign, ok?

Issue #7727, opened 2020-05-04
Working with rank-deficient EEGLAB data


hoechenberger:
So a colleague of mine was handed a set of files that had been pre-processed in EEGLAB. The data was read via `mne.read_epochs_eeglab()`. When computing the noise covariance via `mne.compute_covariance()` and visualizing the results via `Covariance.plot()`, we noticed that the data looked a bit odd. Turns out that during preprocessing, artifact rejection via ICA had been carried out, and some ICs had been removed. So the data was now rank-deficient. However, MNE didn't "pick this up" – in fact, `mne.compute_rank()` assumed the rank was full even with `rank=None` to force an actual estimation. And according to `raw.info`, obviously, the rank was full as well.

So I'm wondering how to best deal with this now. We could certainly pass `rank=dict(eeg=true_rank)` to all functions that require a `rank` input. But this seems cumbersome and error-prone, and I'm not sure if there are other places where this situation could lead to errors? I would prefer just being able to do something like, `raw.rank = true_rank`; or, even better, the EEGLAB import should do something of that sort. A way to detect rank deficiency after ICA is simply by checking if the number of sensors is equal to the number of ICs in the `EEG` structure:

```python
data = scipy.io.loadmat('myfile.set')
n_comps, n_sensors = data['EEG']['icaweights'][0, 0].shape
if n_comps == n_sensors:
    print('Data has full rank')
else:
    print(f'Data has rank {n_comps}')
```
(Of course this only works if no (avg) reference has been set)

Any opinions on that?

cc @cbrnr

hoechenberger:
Also just looking at what MNE's `ICA.apply()` does, it also appears to me that a potential change of data rank is not "documented" anywhere. Is this not bound to lead to issues?

agramfort:
one intuition is that the EEGLAB data were stored to disk in single
precision and our rank detection code does not use
thresholds for this

>

hoechenberger:
> one intuition is that the EEGLAB data were stored to disk in single precision and our rank detection code does not use thresholds for this

So you think that, under normal circumstances, rank calculation should work? Because I also remember that in the `mne-study-template`, we had to switch to using `rank='info'`, because the default didn't work with some MEG datasets…

cbrnr:
[...]

PR #8328, opened 2020-10-01
WIP: Warn if ICA was performed on potentially rank-deficient data

hoechenberger:
#### Reference issue
Closes #8313, #8316, #7727

- Emits a warning if we have reason to believe ICA operated on rank-deficient data.
- plus some small cleanups

#### Additional information
WIP because working on this PR exposed that the default for `max_pca_components`, currently `n_channels`, might not be the best choice. Lots of tests fail now because we get the new warning, typically:

```
E   RuntimeWarning: It is likely that we operated on rank-deficient data, which is known to be problematic. The estimated rank of the data is 363, which is smaller than the number of retained PCA components (366).
E   
E   Rank-deficiency  can be caused through SSP projections or certain referencing schemes, e.g. an EEG average reference.It is STRONGLY advised to perform rank reduction using PCA prior to ICA in order to ensure proper operation of the ICA algorithm.
E   
E   Since the rank of your data was estimated to be 363, we recommend instantiating the ICA object via:
E   
E       ICA(..., max_pca_components=363)
E   
E   (or using an even smaller number).
```

This is because currently, I calculate rank with SSP projectors applied, so the rank of the data would effectively be `n_channels - len(info['projs'])`, which is obviously smaller than `n_channels` if there is any projection vectors stored.

Now this got me thinking, should the default of `max_pca_components` be changed from `n_channels` to `n_channels - len(info['projs'])`?

Applied SSP projectors leading to rank-deficient data and thereby screwing up ICA was exactly the problem @agramfort and I witnessed in the past.

### DIFF
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1324,6 +1324,15 @@
   year = {2010}
 }
 
+
+@article{Sabbagh2020,
+  author = {Sabbagh, David and Ablin, Pierre and Varoquaux, Gaël and Gramfort, Alexandre and Engemann, Denis A.},
+  title = {Predictive regression modeling with MEG/EEG: from source power to signals and cognitive states},
+  journal = {NeuroImage},
+  year = {2020},
+  doi = {10.1016/j.neuroimage.2020.116893}
+}
+
 @article{Sarvas1987,
   author = {Sarvas, Jukka},
   doi = {10.1088/0031-9155/32/1/004},
--- a/examples/inverse/plot_mne_cov_power.py
+++ b/examples/inverse/plot_mne_cov_power.py
@@ -5,7 +5,14 @@ Compute source power estimate by projecting the covariance with MNE
 
 We can apply the MNE inverse operator to a covariance matrix to obtain
 an estimate of source power. This is computationally more efficient than first
-estimating the source timecourses and then computing their power.
+estimating the source timecourses and then computing their power. This
+code is based on the code from :footcite:`Sabbagh2020` and has been useful to
+correct for individual field spread using source localization in the context of
+predictive modeling.
+
+References
+----------
+.. footbibliography::
 """
 # Author: Denis A. Engemann <denis-alexander.engemann@inria.fr>
 #         Luke Bloy <luke.bloy@gmail.com>
--- a/mne/cov.py
+++ b/mne/cov.py
@@ -719,9 +719,10 @@ def compute_covariance(epochs, keep_sample_mean=True, tmin=None, tmax=None,
         perform estimates using multiple methods.
         If 'auto' or a list of methods, the best estimator will be determined
         based on log-likelihood and cross-validation on unseen data as
-        described in [1]_. Valid methods are 'empirical', 'diagonal_fixed',
-        'shrunk', 'oas', 'ledoit_wolf', 'factor_analysis', 'shrinkage',
-        and 'pca' (see Notes). If ``'auto'``, it expands to::
+        described in :footcite:`EngemannGramfort2015`. Valid methods are
+        'empirical', 'diagonal_fixed', 'shrunk', 'oas', 'ledoit_wolf',
+        'factor_analysis', 'shrinkage', and 'pca' (see Notes). If ``'auto'``,
+        it expands to::
 
              ['shrunk', 'diagonal_fixed', 'empirical', 'factor_analysis']
 
@@ -803,9 +804,10 @@ def compute_covariance(epochs, keep_sample_mean=True, tmin=None, tmax=None,
       .. versionadded:: 0.16
     * ``'ledoit_wolf'``
         The Ledoit-Wolf estimator, which uses an
-        empirical formula for the optimal shrinkage value [2]_.
+        empirical formula for the optimal shrinkage value
+        :footcite:`LedoitWolf2004`.
     * ``'oas'``
-        The OAS estimator [5]_, which uses a different
+        The OAS estimator :footcite:`ChenEtAl2010`, which uses a different
         empricial formula for the optimal shrinkage value.
 
       .. versionadded:: 0.16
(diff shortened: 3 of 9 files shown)


hoechenberger:
@agramfort @larsoner @cbrnr 
Would like to hear your thoughts on this

hoechenberger:
If #8326 becomes a success, we can also close this PR…

larsoner:
> WIP because working on this PR exposed that the default for max_pca_components, currently n_channels, might not be the best choice. Lots of tests fail now because we get the new warning, typically:

Do we fit ICA on the data with projectors applied, or not? If it varies, then we can probably compute rank only taking into account projectors marked as "active" (i.e., applied to the data). I imagine most raw instances have not had the projectors applied, and hence shouldn't be counted when computing the rank.

hoechenberger:
> Do we fit ICA on the data with projectors applied, or not?

I'm not sure, would need to step through the code. Does `Raw/Epochs.get_data()` apply projectors before returning the data?

larsoner:
> Does Raw/Epochs.get_data() apply projectors before returning the data?

Only if `apply_proj` or `proj=True` was done, in which case `active=True` should be set in `info['projs']`. Hence why I think we need to make use of that property.

And you should be able to check this in a test that uses it just by doing a quick `np.linalg.matrix_rank` on the data, at least if it's MEG (or EEG) only.

hoechenberger:
Ok, sounds good!

hoechenberger:
@larsoner 
I've added handling for projectors, but haven't extensively tested.

Also needed to add an option to include EOG channels in rank computation via `compute_rank`, similarly to how we handle `ref_meg`. Probably will need to do the same with ECG too…

larsoner:
> Also needed to add an option to include EOG channels in rank computation via compute_rank, similarly to how we handle ref_meg. Probably will need to do the same with ECG too…

Why? If these channels are included in the ICA, they should be included in the rank calculation, no? We need the ICA to be stable, so if those channels are included in ICA they should affect the rank calculation.

hoechenberger:
> Why? If these channels are included in the ICA, they should be included in the rank calculation, no? We need the ICA to be stable, so if those channels are included in ICA they should affect the rank calculation.

Yes, exactly. By default on `master` they are not included!
I discovered that bc ICA tests that do have an EOG channel started throwing a warning about rank deficiency, because only the M/EEG channels were considered by the rank calculation.

My current approach of adjusting `_estimate_rank_meeg_signals` is probably not the best idea, though, because EOG is clearly not M/EEG? Or is it?

hoechenberger:
@larsoner I've reverted the "EOG changes" so you can see for yourself what the problem was

cbrnr:
I'm OK with issuing a warning if we detect rank deficiency. However, the rank needs to be computed on the same data that ICA uses (and not just by assuming certain types of channels). Furthermore, I have never used projectors, so I hope these do not interfere with ICA - if they do (e.g. if projectors are not only available but also active) we should also issue a warning.

hoechenberger:
> I'm OK with issuing a warning if we detect rank deficiency. However, the rank needs to be computed on the same data that ICA uses (and not just by assuming certain types of channels).

Yes, this is precisely what we're / I'm  trying to achieve here :)

> Furthermore, I have never used projectors, so I hope these do not interfere with ICA - if they do (e.g. if projectors are not only available but also active) we should also issue a warning.

If they're applied, each projector will reduce the rank of the data by 1.

cbrnr:
> If they're applied, each projector will reduce the rank of the data by 1.

Yes, but at least issue an info. I could imagine myself having applied projectors accidentally, or people might want to compute ICA with the original data even if they have projectors applied. That way, they can at least see what's going on.

hoechenberger:
> Yes, but at least issue an info. I could imagine myself having applied projectors accidentally, or people might want to compute ICA with the original data even if they have projectors applied. That way, they can at least see what's going on.

You mean something like
```
INFO Performing ICA on data with rank X (Y channels, Z projectors applied).
```
?

cbrnr:
Exactly!

hoechenberger:
@cbrnr Ok, that seems reasonable.

@larsoner 
For your convenience, without the EOG-picking adjustments in `compute_rank()`, the following test fails (see CI logs: estimated rank is 4, but number of channels was 5 – 4 MEG + 1 EOG):

```
__________________________ test_eog_channel[fastica] __________________________
mne\preprocessing\tests\test_ica.py:931: in test_eog_channel
    ica.fit(inst, picks=picks1)
<decorator-gen-384>:21: in fit
    ???
mne\preprocessing\ica.py:570: in fit
    warn(f'It is likely that we operated on rank-deficient data, '
mne\utils\_logging.py:348: in warn
    warnings.warn_explicit(
E   RuntimeWarning: It is likely that we operated on rank-deficient data, which is known to be problematic. The estimated rank of the data is 4, which is smaller than the number of retained PCA components (5).
E   
E   Rank-deficiency  can be caused through SSP projections or certain referencing schemes, e.g. an EEG average reference. It is STRONGLY advised to perform rank reduction using PCA prior to ICA in order to ensure proper operation of the ICA procedure.
E   
E   Since the rank of your data was estimated to be 4, we recommend instantiating the ICA object via:
E   
E       ICA(..., max_pca_components=4)
E   
E   (or using an even smaller number).
```

larsoner:
IIRC the only reason it's really named `_estimate_rank_meeg_signals` is because typically we only want data channels for noise cov + inversion. I don't see a problem with making it work for other channel types, too -- we just need to make sure our default scaling factors do something reasonable for EOG and ECG channels for example, and that we don't start accidentally returning EOG/ECG in places in the code that currently use that function. This should be fixable via some picking arguments of some sort.

hoechenberger:
Thanks @larsoner. Are you implying that EOG (and ECG, for that matter) are not data channels?

larsoner:
[Yes](https://mne.tools/dev/glossary.html#term-data-channels)

PR #8329, opened 2020-10-01
MRG: ENH, BUG: Minor ICA cleanups

hoechenberger:
Some cherry-picks from #8328 

- add more validation
- fix bug if `max_pca_components` is a float
- remove redundant `max_pca_components ` argument from `ICA._fit`
- handle `max_pca_components=1.0` like `max_pca_components=None`
- make test cases more independent by avoiding side effects through repeated `ICA.apply()`'ing to the same object


larsoner:
Thanks @hoechenberger

PR #8333, opened 2020-10-02
Add _check_option return value

### LINKED ISSUE #8204: Should mne.utils._check_option return the checked value?
Currently, `mne.utils._check_option` returns `True` if a value matches a list of allowed values or else raises a `ValueError`. I'm often using it as follows:

```python
_check_option('component_order', component_order, ['alternate', 'mutual_info'])
self.component_order = component_order
```

It would be useful if `_check_option` returned the value instead of `True`, because this would shorten my example to:

```python
self.component_order = _check_option('component_order', component_order, ['alternate', 'mutual_info'])
```

I don't really see a problem with returning the value, but maybe others rely on the boolean return value (although a quick search didn't turn up any assignment to the return value of this function).

cbrnr:
Fixes #8204. Instead of just returning `True` if the argument is valid, the function now returns the valid argument for further use. This can simplify some code constructs.

### DIFF
--- a/mne/channels/tests/test_montage.py
+++ b/mne/channels/tests/test_montage.py
@@ -649,7 +649,7 @@ def test_read_dig_polhemus_isotrak_error_handling(isotrak_eeg, tmpdir):
     fname = op.join(tmpdir, 'foo.bar')
     with pytest.raises(
         ValueError,
-        match="Allowed val.*'.hsp', '.elp' and '.eeg', but got '.bar' instead"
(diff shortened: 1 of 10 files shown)


agramfort (approved):

hoechenberger:
@cbrnr Travis is not happy

hoechenberger (approved):
LGTM once CI goes green

cbrnr:
@hoechenberger all green!

hoechenberger:
Thanks, @cbrnr!

PR #8335, opened 2020-10-02
WIP: Plot label time course with _TimeViewer

GuillaumeFavelier:
With this PR, `_TimeViewer` can plot label time course.

Multiple modifications are needed. The first one is to allow `src=None` in `extract_label_time_course()` and the second one
is to add a 'label' mode in `_TimeViewer`.

This PR is still a work in progress.

Preliminary result:

image

<details>
<summary>Code snippet</summary>

```py
import os
import mne
labels = mne.read_labels_from_annot(
    'fsaverage', 'HCPMMP1', 'lh', subjects_dir=subjects_dir)
stc = mne.read_source_estimate(fname_stc, subject='fsaverage')
brain = stc.plot(
    subject='fsaverage',
    subjects_dir=subjects_dir,
    cortex='low_contrast',
    size=600,
    background="white",
    time_viewer=True,
    show_traces=False,
)
aud_label = [label for label in labels if label.name == 'L_A1_ROI-lh'][0]
brain.add_label(aud_label, borders=False)
brain.time_viewer.show_traces = True
brain.time_viewer.configure_point_picking()
brain.time_viewer._configure_label_time_course()
```

</details>

Related to https://github.com/mne-tools/mne-python/issues/7959


larsoner:
> The first one is to allow src=None in extract_label_time_course()

We will need a widget to set the extraction mode. Only mean/max should be supported if `src=None` but if `src` is provided then `mean_flip`, `pca_flip` should be allowed, too. `src` is an argument of `stc.plot` nowadays:

https://mne.tools/stable/generated/mne.viz.plot_source_estimates.html#mne.viz.plot_source_estimates

larsoner:
Another idea -- optionally extract then color each label according to its extracted value. So if you pass a parcellation with 44 labels (like hcpmmp1_combined), what you see is the "parcellation activation levels", i.e., a set of 44 patches of cortex, each patch with all the same value.

Also we'll have to figure out how to make it work for volumes, but all the same considerations probably apply -- it's "just" another transformation of the data before setting the VTK object scalars.

larsoner:
- I would never show the label color, it'll block the activation -- if people want this they can do `brain.add_label` or `brain.add_parcellation`, probably with `borders=True` separately

GuillaumeFavelier:
I updated the branch with #8352 and added a parameter `traces` to `add_label()` which can detect compatibility with time viewer ...etc. I think it makes sense and the color in the plot matches the label's:

image

<details>
<summary>Code snippet</summary>

```py
import os
import mne
labels = mne.read_labels_from_annot(
    'fsaverage', 'HCPMMP1', 'lh', subjects_dir=subjects_dir)
stc = mne.read_source_estimate(fname_stc, subject='fsaverage')
brain = stc.plot(
    subject='fsaverage',
    subjects_dir=subjects_dir,
    cortex='low_contrast',
    size=600,
    background="white",
    time_viewer=True,
    show_traces=False,
)
aud_label = [label for label in labels if label.name == 'L_A1_ROI-lh'][0]
brain.add_label(aud_label, borders=True, traces=True)
```

</details>

> I would never show the label color, it'll block the activation

I can display the time course instead of the label then when `traces=True`?

GuillaumeFavelier:
Using `read_labels_from_annot()` to plot the labels of the 'HCPMMP1' parcellation, the clutter is huge.

image

It would be cool to just click on a patch to display its time course :thinking:

GuillaumeFavelier:
> It would be cool to just click on a patch to display its time course

Something similar to *vertex time course* but for label:

output

larsoner:
Awesome! I still think we shouldn't color in the label based on its label color, though. It loses information if you click the play button the brain display is static. Instead I'd probably put in a circle in the center (of mass) of the label that is the label color, but still display on the brain the activation color (filing the label).

larsoner:
(Also this keeps consistency with "sphere means this activation time course is displayed")

GuillaumeFavelier:
Result after discussing offline with @agramfort :

output

The color for the parcellation is fixed (apparently, green is recommended) and by clicking on the mesh, the patch is overlayed (alpha=0.6) and its time course is shown/hidden (toggled). The colors are reused from the internal cycle (grey shoud be removed).

A "time course mode" widget would be added toggling between "vertex" (default) and "label" plus more for the parameters.

What do you think @larsoner ? Worst case, we can discuss it offline too.

*Note: the code is so prototype, I did not push yet*

GuillaumeFavelier:
Or a different theme:

image

larsoner:
Some ideas:

- I like the white better. I would also try black (or dark gray) since white often means "maximally active" whereas we are just trying to delineate regions (similar to how the light/dark gray delinate gyri and sulci).
- Why fill the label when we could just color in that portion with `borders=True` and alpha 1? It would obscure even less of the activity.
- I would have the z-order be 1) brain 2) annotation overlay 3) activity 4) currently active labels, with annotation overlay and currently active labels all rendered with `borders=True`. That way you see the most activity but it's still clear what has been clicked on.

larsoner:
(these may or may not work that well, but I'd keep in mind the context where you want to create a movie and be able to see activation on the brain -- hence the motivation to not fill labels, keep activation overlay toward the top, etc.)

GuillaumeFavelier:
> I like the white better. I would also try black (or dark gray) 

I'll try them all and report in here.

> Why fill the label when we could just color in that portion with borders=True and alpha 1?

I tried this one and the result is not as smooth, there is a clear aliasing now that borders use tubes. I will see how to deactivate it.

GuillaumeFavelier:
More standard borders with different color for the parcellation:

white (1) | grey (0.6) | black (0)
-----------|---------------|-------------
image | image | image

I personally prefer grey because it's less contrast on the brain but I am not sure of the effect we are looking for @agramfort, @larsoner

agramfort:
I prefer white :)

larsoner:
I also prefer gray. So unfortunately @agramfort has been outvoted here :)

larsoner:
(specifically I would use _brain_color because this should work with whatever brain options are used)

agramfort:
democracy is tough sometimes :) 

I'll try to convince some friends to vote ;)

GuillaumeFavelier:
When the 'label' picking mode is enabled, another tool bar is created. The plan is to set all the time course extraction parameters in here:

image

The `mode` parameter is available but only `mean` works for now, I might need help to make the other methods work properly :sweat:

GuillaumeFavelier:
Here is the example I use locally to generate my screenshots:

```py
import os
import mne
labels = mne.read_labels_from_annot(
    'fsaverage', 'HCPMMP1', 'lh', subjects_dir=subjects_dir)
stc = mne.read_source_estimate(fname_stc, subject='fsaverage')
brain = stc.plot(
    subject='fsaverage',
    subjects_dir=subjects_dir,
    cortex='low_contrast',
    size=600,
    background="white",
    time_viewer=True,
    show_traces=True,
)
aud_label = [label for label in labels if label.name == 'L_A1_ROI-lh'][0]
brain.add_annotation('HCPMMP1', traces=True)
```

But unfortunately `src` is not provided in this case. So is there a different subject I could use where `src` is set for `Brain` *and* I could use a valid annotation to test the extra parameters of `extract_label_time_course`?

larsoner:
You should be able to use `fsaverage-5-src.fif` which should either be in `subjects/sample/bem` or obtained with `fetch_fsaverage()` (it should put it there) or created with `src = mne.setup_source_space('fsaverage', spacing='ico5')` (but this hopefully isn't necessary)

GuillaumeFavelier:
This is the new example I use and I was able to confim that the other extraction modes are available (`auto`, `mean_flip` and `pca_flip`):

```py
import os
import mne
sample_dir = os.path.join(data_path, 'subjects', 'fsaverage', 'bem')
sample_dir2 = os.path.join(data_path, 'MEG', 'sample')
fname_stc = os.path.join(sample_dir2, 'fsaverage_audvis-meg')
fname_src = os.path.join(sample_dir, 'fsaverage-ico-5-src.fif')
stc = mne.read_source_estimate(fname_stc, subject='fsaverage')
src = mne.read_source_spaces(fname_src)
brain = stc.plot(
    subject='fsaverage',
    subjects_dir=subjects_dir,
    cortex='low_contrast',
    size=600,
    background="white",
    time_viewer=True,
    show_traces=True,
    src=src,
)
brain.add_annotation('HCPMMP1', traces=True)
```

image

mne/viz/_3d.py
```diff
@@ -1892,7 +1893,11 @@ def _plot_stc(stc, subject, surface, hemi, colormap, time_label,
             kwargs["fmax"] = scale_pts[2]
             kwargs["clim"] = clim
             kwargs["volume_options"] = volume_options
-            kwargs["src"] = src_vol
+            if src_vol is not None:
+                kwargs["src"] = src_vol
+            else:
+                kwargs["src"] = orig_src
```
larsoner, line 1899: I suspect this is a bug, for example if it's a mixed source estimate (surface+vol). I think now that we use `src` for both surface and vol, it just juld be `kwargs['src'] = orig_src`, and the logic for extracting the volume parts should be moved to `Brain` init (and really it should become logic for splitting into volume and surface parts).

larsoner (commented):
Keep in mind that eventually we want this to work for volume source estimates, too. I don't think it should require too much extra work: in both cases, clicking should just pick a vertex (and it does in `master`), then we figure out which surface label or volume label this is a part of and extract the time course for that label (that is what this PR should implement). If we do it this way, everything should also "just work" for mixed (surface+vol) source estimates, too.

GuillaumeFavelier:
I think volumes require special treatment. Before enabling traces=True for the label,  I find the default viz of the annotation very messy:

image

I am barely able to notice the limits of the labels. I believe doing the opposite for volume and actually using tubes could help.

larsoner:
> I think volumes require special treatment. Before enabling traces=True, I find the default viz of the annotation very messy:

Maybe for the viz end, sure. I'm just thinking that how they work (the algorithm/logical flow) should be the same

GuillaumeFavelier:
The logic of `stc_surf`/`stc_vol` and the way we interact with `Brain` through `add_data` prevents me to completely remove `_triage_stc` (1). The best I can do is move the function to `Brain` for consistency.

(1) I added `stc` and `src` as parameters to `add_data` but in `plot_stc` those are used to extract the correct `kwargs` to send to `add_data`, it creates some kind of dependency cycle. I think a cleaner API would be to just send stc/src and let Brain handle the rest. Unfortunately, I think this breaks backend compatibilty.

larsoner:
> (1) I added stc and src as parameters to add_data but in plot_stc those are used to extract the correct kwargs to send to add_data, it creates some kind of dependency cycle. I think a cleaner API would be to just send stc/src and let Brain handle the rest. Unfortunately, I think this breaks backend compatibilty.

I don't see any need for a cycle, we can just take advantage of no-op calls. There should be one function that splits stc+src into surface+vol components, e.g. `stc_surf, src_surf, stc_vol, src_vol = _split_surf_vol(stc, src)`. And both `plot_source_estimate` and `Brain.add_data` can call it.

Consider the `mixed_stc` case (surf+vol). Currently `plot_source_estimate`'s call to `_split_surf_vol(stc, src)` would break this into all-non-None `stc_surf, src_surf, stc_vol, src_vol` parts and then make two calls to `brain.add_data`. On the first call to `brain.add_data` by `plot_source_estimate`, the `surf_stc.data` and `surf_src` are passed, `Brain` calls `_split_surf_vol` which yields just non-None `surf_stc` and `surf_src` (`vol_stc` and `vol_src` are None), and it adds just the surface data. On the second `brain.add_data` call by `plot_source_estimates`, `vol_stc.data` and `vol_src` are passed, `Brain` calls `_split_surf_vol` which also yields only non-None `vol_stc` and `vol_src`, and so only volume data is added.

And someone using PyVista-backend `add_data` can pass `add_data(mixed_stc, src=mixed_src)` and internally `add_data` will call `_split_surf_vol`, both surf stc+src and vol stc+src will come out of it, and so `add_data` will add data of both types. This would require allowing `brain.add_data` to take a SourceEstimate as the `data` argument, which seems fine. But the result here should be equivalent to doing `plot_source_estimate(mixed_stc, src=mixed_src)`

GuillaumeFavelier:
Using `stc` as data argument, I tried d880dca

GuillaumeFavelier:
This is the new UI to switch between vertex and label time course. It's called "Trace mode" for now:

output

There is still work to do to cover all of this.

GuillaumeFavelier:
The new version is more intuitive:

Annotation='aparc' | Annotation='None'
---|---
image | image

GuillaumeFavelier:
Several technical issues need to be addressed :
1) Azure failing with:

```
========================== Starting Command Output ===========================
C:\shells\bash.exe --noprofile --norc /d/a/_temp/97fffddd-84e1-42ad-a2b1-ac73004d6dfc.sh
C:/msys64/mingw64/bin/python.exe: No module named pip

##[error]Bash exited with code '1'.
Finishing: Install dependencies with pip
```

2) Travis failing with:

```
../../../virtualenv/python3.8.1/lib/python3.8/site-packages/numba/core/types/__init__.py:108

  /home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/numba/core/types/__init__.py:108: DeprecationWarning: `np.long` is a deprecated alias for `np.compat.long`. Use `np.compat.long` by itself, which is identical in behavior, to silence this warning. In the likely event your code does not need to work on Python 2 you can use the builtin ``int`` for which ``np.compat.long`` is itself an alias. If you specifically wanted the numpy scalar type, use `np.int_` here.

    long_ = _make_signed(np.long)

../../../virtualenv/python3.8.1/lib/python3.8/site-packages/numba/core/types/__init__.py:109

  /home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/numba/core/types/__init__.py:109: DeprecationWarning: `np.long` is a deprecated alias for `np.compat.long`. Use `np.compat.long` by itself, which is identical in behavior, to silence this warning. In the likely event your code does not need to work on Python 2 you can use the builtin ``int`` for which ``np.compat.long`` is itself an alias. If you specifically wanted the numpy scalar type, use `np.int_` here.

    ulong = _make_unsigned(np.long)

../../../virtualenv/python3.8.1/lib/python3.8/site-packages/pyvista/utilities/cell_type_helpers/cell_selector.py:106

  /home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/pyvista/utilities/cell_type_helpers/cell_selector.py:106: ResourceWarning: unclosed file <_io.TextIOWrapper name='/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/pyvista/utilities/cell_type_helpers/vtk_cell_types.txt' mode='r' encoding='UTF-8'>

    generate_all_cell_types()
```

3) Coverage: waiting for latest result

larsoner:
I have to make a PR to get rid of Windows `ci-helpers`, I can deal with the warnings there, too. In the meantime just listen to the CIs that work / local testing

GuillaumeFavelier:
Argh: `94.01%` so close

GuillaumeFavelier:
This was originally a new set of parameters for `add_label` and `add_annotation` and ended up being a new mode in Brain's UI.

Sorry for the amount of changes. If necessary, I can split it to simplify the review process.

larsoner:
I think it makes sense to add the parameters and interactivity at the same time, so this is fine

GuillaumeFavelier:
It seems that coverage has a cost:

```
206.57s total  mne/viz/_brain/tests/
```

larsoner:
Yikes, even with the `visible=False` trick in there? Is it possible some things are not properly being set as invisible, such as the widgets? Or maybe some of these can use `time_viewer=False` and/or `show_traces=False` without losing coverage?

GuillaumeFavelier:
I'll investigate now

GuillaumeFavelier:
Hiding the window does have an impact:

```
=========================== slowest 20 test modules ===========================
158.37s total  mne/preprocessing/tests/
145.74s total  mne/beamformer/tests/
144.79s total  mne/tests/test_epochs.py
128.19s total  mne/minimum_norm/tests/
 99.08s total  mne/io/fiff/
 70.01s total  mne/tests/test_rank.py
 53.33s total  mne/tests/test_source_estimate.py
 52.93s total  mne/channels/tests/
 50.20s total  mne/viz/tests/test_epochs.py
 46.58s total  mne/viz/tests/test_raw.py
 40.53s total  mne/viz/tests/test_3d.py
 38.54s total  mne/io/egi/
 33.05s total  mne/stats/tests/
 31.46s total  mne/viz/tests/test_evoked.py
 31.28s total  mne/tests/test_cov.py
 30.54s total  mne/viz/_brain/tests/             <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
```

I'll try showing it only when necessary (screenshot).

larsoner:
That's a great decrease!

GuillaumeFavelier:
False alarm, my enthusiasm got the best out of me and I did not compare the right test: 

```
=========================== slowest 20 test modules ============================

191.86s total  mne/viz/_brain/tests/
```

It's just a 15 seconds decrease, I'll come back to it after https://github.com/mne-tools/mne-python/pull/8495

GuillaumeFavelier:
Still not enough on [Travis PIP PRE](https://travis-ci.org/github/mne-tools/mne-python/jobs/743370308#L5823):

```
=========================== slowest 20 test modules ============================

178.20s total  mne/preprocessing/tests/
147.45s total  mne/beamformer/tests/
134.86s total  mne/minimum_norm/tests/
134.40s total  mne/viz/tests/test_3d.py
121.80s total  mne/viz/_brain/tests/
 87.81s total  mne/tests/test_epochs.py
 85.76s total  mne/viz/tests/test_3d_mpl.py
 76.66s total  mne/inverse_sparse/tests/
 76.43s total  mne/tests/test_rank.py
 72.51s total  mne/tests/test_source_space.py
 63.46s total  mne/io/fiff/
 54.44s total  mne/tests/test_morph.py
 51.41s total  mne/tests/test_report.py
 45.88s total  mne/io/fieldtrip/
 43.46s total  mne/forward/tests/
 33.07s total  mne/viz/tests/test_ica.py
 32.06s total  mne/tests/test_cov.py
 31.43s total  mne/io/egi/
 30.93s total  mne/commands/tests/
 30.53s total  mne/channels/tests/
```

GuillaumeFavelier:
Among the easy ways to decrease testing time, I realized we still test both 3d backends (`mayavi` and `pyvista`) except on Circle. I was thinking maybe we could create a dedicated CI job for `mayavi` and keep the others on the default. What do you think @agramfort, @larsoner ?

GuillaumeFavelier:
Another solution would be to migrate the Travis PIP PRE job to Github Actions.

larsoner:
> Another solution would be to migrate the Travis PIP PRE job to Github Actions.

Sure, go for it. The Travis runs have been a painful bottleneck lately. +1 for moving all three remaining jobs to GitHub actions or Azure. Travis [is going to start limiting open-source projects to 1000 minutes per month](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing). In this PR the three jobs took over 100 minutes, so that is only 10 PR/branch builds a month.

While we're thinking reconfiguring CIs, we could also try:

- https://github.com/TrueBrain/actions-flake8
- https://github.com/codespell-project/actions-codespell

They make annotations inline, which is nice. These could be removed from the Azure style run and moved to a new GitHub Actions style run.

GuillaumeFavelier:
I'm on it :+1: I'll start with Github Actions then. It should not be hard to then move anything to Azure

larsoner:
Sounds good -- please try to keep code DRY where possible.

Also I realized a big advantage of having most parts in `.sh` files is that if a PR changes those to tweak how the CI runs, all others can benefit immediately by the checkout+merge that CIs do, without having to rebase. When you change YAML files, these only are reflected in other PRs once they rebase on that change. So it's nice we're moving in a more-`.sh` direction, even though it risks hurting the readability of the YAML by creating more files that need to be viewed to see what's really going on. But I think your current ones for the actions are clear enough.

GuillaumeFavelier:
> try to keep code DRY where possible.

I'll keep that in mind. I intend to open a different PR though, this one has unrelated changes.

> all others can benefit immediately by the checkout+merge that CIs do, without having to rebase

Ooooh I see :open_mouth:

GuillaumeFavelier:
@hoechenberger, @drammock, @larsoner, @agramfort could you give it a try?

```py
import os
import mne
initial_time = 0.13
                 time_viewer=True,
                 show_traces=True,
                 size=600,
                 )
```

You should see some combo boxes to select the annotation and the extraction mode, what do think?

image

agramfort:
I have some rendering issues. We can look tomorrow with screen sharing.

>

GuillaumeFavelier:
Thanks for trying! 

> We can look tomorrow with screen sharing.

Let me know when you're available :)

larsoner:
Milestoning for 0.23 -- I think we'll want some time to test this out a bit

hoechenberger:
@GuillaumeFavelier and I tested this this morning, and I think it's now working for @agramfort too in a fresh env. So it could make sense to include this in 0.22 already.

hoechenberger:
@larsoner 
> I also prefer gray.

I think the contrast is sometimes a bit low so it's difficult to make out the parcels. Would be cool if the outline color were adjustable.

mne/viz/_brain/_brain.py
```diff
@@ -934,6 +967,75 @@ def _configure_point_picking(self):
             self._on_pick
         )
 
+    def _configure_trace_mode(self):
+        from ...label import _read_annot_cands
```
agramfort, line 971: I think there is a bug in _read_annot_cands

see:

```
ipdb> cc = 'lh.random30.annot'
ipdb> cc.lstrip('lh.').lstrip('rh.').rstrip('.annot')
'andom30'
```

it should return `random30`

GuillaumeFavelier: What do you think about:

https://github.com/mne-tools/mne-python/blob/1ab0863cdfae6e4951b1027b897fc470d3537849/mne/label.py#L1883-L1884

agramfort: works great !

agramfort (commented):
I just tried

```
import os
import mne
initial_time = 0.13
subject = 'sample'
src = mne.setup_source_space(subject, subjects_dir=subjects_dir,
                             spacing='oct5', add_dist=False)
                 time_viewer=True, src=src,
                 show_traces=True, hemi='both', size=600)
```

now picking annotation BA i get a crash:

ValueError: 2792/3732 left hemisphere stc vertices missing from the source space, likely mismatch

mmagnuski:
I agree with @hoechenberger - maybe if the gray was a bit whiter then everyone would be happy? :)

larsoner:
Fine fine fine -- maybe just white with an alpha of 0.5 or 0.75 or so is the way to go?

GuillaumeFavelier:
Here is a result with`color='white'` with `alpha=0.75`:

image

larsoner:
Works for me

GuillaumeFavelier:
> now picking annotation BA i get a crash:
> ValueError: 2792/3732 left hemisphere stc vertices missing from the source space, likely mismatch

This is caused by https://github.com/mne-tools/mne-python/blob/6222d1715e462469212ae49aa66c9a60d7faf712/mne/source_estimate.py#L2875-L2883

It seems there is a mismatch of vertex indices between `src` and `stc`. This is the example I use to debug @agramfort :

```py
import os
import mne
fname_fwd = os.path.join(sample_dir, 'sample_audvis-meg-eeg-oct-6-fwd.fif')
fwd = mne.read_forward_solution(fname_fwd)
src = fwd['src']
brain = stc.plot(
    subject='sample',
    subjects_dir=subjects_dir,
    cortex='low_contrast',
    size=600,
    hemi='lh',
    background="white",
    time_viewer=True,
    show_traces="label",
    src=src,
)
```

hoechenberger:
> Here is a result with`color='white'` with `alpha=0.75`:

I think that could be a good compromise! Thanks

agramfort:
indeed I made a mistake with:

import os
import mne
initial_time = 0.13
subject = 'sample'
src = mne.setup_source_space(subject, subjects_dir=subjects_dir,
                             spacing='oct5', add_dist=False)
                 time_viewer=True, src=src,
                 show_traces=True, hemi='both', size=600)

can you make sure dumb users like me don't make this mistake without noticing.
Basically preventing them to do what I did by making sure the src and
the stc.vertices
match.

larsoner:
> Basically preventing them to do what I did by making sure the src and the stc.vertices

I think this is exactly what the warning was meant to do, and the message reflects that, no?

agramfort:
AFAIK there is no warning so far. It's just crashing for me now

>

larsoner:
Sorry not warning, error message:

> ValueError: 2792/3732 left hemisphere stc vertices missing from the source space, likely mismatch

It is telling you that there is a mismatch between your STC vertices and those of your source space. Maybe "likely mismatch between stc and source space (either subject or source space spacing)" or something?

agramfort:
I mean it should crash earlier.

>

GuillaumeFavelier:
> I mean it should crash earlier.

I can refactor the check in `stc.plot` or `brain.add_data` for example

agramfort:
can you just add a check_stc_src function that just gets called early in
stc.plot ?

>

mne/source_estimate.py
```diff
@@ -2860,6 +2860,16 @@ def _temporary_vertices(src, vertices):
...
 
 
+def _check_stc_src(stc, src):
+    if stc is not None and src is not None:
+        for s, v, hemi in zip(src, stc.vertices, ('left', 'right')):
+            n_missing = (~np.in1d(v, s['vertno'])).sum()
+            if n_missing:
+                raise ValueError('%d/%d %s hemisphere stc vertices '
+                                 'missing from the source space, likely '
+                                 'mismatch' % (n_missing, len(v), hemi))
```
agramfort, line 2870: what you should check is that src contains a superset of the vertices in stc as it's all you need. You need in the src the vertices from stc but if you have more it's ok.

larsoner: I think that's what's already done here already with the (effectively) `(~np.in1d(stc.vertices[ii], src[ii]['vertno'])).sum()`, i.e., "how many of the stc vertices are not in the source space". So nothing to change here I think?

... except that it would be good to check not just the surface STC + surface source but also volume STC + volume source. The same check applies there. Personally I would just check that `stc.vertices` and `src` are the same length, then iterate directly over the `stc.vertices` and `src` and just say that `stc.vertices[{ii}]` has a problem. But if it's already split into surfaces and volume somewhere then we could allow it to be more specific.

agramfort:
indeed... I tried:

import os
import mne
initial_time = 0.13
subject = 'sample'
src = mne.setup_source_space(subject, subjects_dir=subjects_dir,
                             spacing='oct6', add_dist=False)
                 time_viewer=True, src=src,
                 show_traces=True, hemi='both', size=600)

does it work for you?

larsoner:
## Source checking

No but this substitution does:
```
src = mne.read_source_spaces(os.path.join(subjects_dir, subject, 'bem', f'{subject}-oct-6-src.fif'))
```
because:
```
>>> np.setdiff1d(src2[0]['vertno'], src[0]['vertno'])
array([ 9078, 28168, 43556, 44008, 66569, 81967, 91727])
>>> np.setdiff1d(src2[1]['vertno'], src[1]['vertno'])
array([  3867,  25320,  42320,  85943, 104085, 106015, 110146, 144951,
       146149])
```
There are slight differences in how MNE-C chooses the vertices versus how we do. So the error message works :)

I checked and this was the case even on `maint/0.9` from 2015, as far back as I could go on my standard python 3.8 install and still have `import mne` work. (I can't believe that I was even able to run code from back then!)

## Comments

This seems to work really nicely! Comments:

1. Got a warning:
    ```
    /home/larsoner/python/mne-python/mne/viz/_brain/_brain.py:2166: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. Use `int` by itself, which is identical in behavior, to silence this warning. If you specifically wanted the numpy scalar type, use `np.int_` here.
   show = np.zeros(n_vertices, dtype=np.int)
   ```
    Perhaps use `np.int64` probably here just to be explicit about it.

2. Vector source estimate plots raise an error:
    ```
    ValueError: x and y must have same first dimension, but have shapes (421,) and (3, 421)
    ```
    Probably this should work, but the only modes supported should be `mean` and `max`. But really it should reflect whatever the behavior `extract_label_time_course` has, e.g.:
    ```
    >>> mne.extract_label_time_course(stc_mixed, labels=mne.read_labels_from_annot('sample'), mode='mean', src=src_mixed)
    array([[[-0.10257054, -0.11464187, -0.11951974, ...,  0.11791901,
    ...
    ```
    To me this suggests that the code coupling / sharing between `extract_label_time_course` and the new Brain functionality is incorrect somehow.

3. When just a volume source space is used, the "annotations" bar just shouldn't even show up. Same when not passing `src`. These should only be enabled when they can actually be used. Eventually we can add volumes, but it should be a separate PR -- and even then, we don't want the `annotations` option to show up when no `src` is provided.

4. When using a mixed source space / mixed source estimate, label extraction for the surfaces should work but it does not:
    ```
    RuntimeError: Label time course plotting is not supported for volume.
    ```

5. Just like we have for the vertex the MNI coordinate, for labels we can easily compute the center-of-mass vertex (there is a [function for this](https://mne.tools/stable/generated/mne.Label.html#mne.Label.center_of_mass)) and then give the MNI coord of it.

6. The colors seem to be the `matplotlib` ones. At some point we should be able to toggle to use the colors that come with the label/parcellation, as they are meaningful (for HCPMMP they correspond to auditory/visual/motor-ness).

Points 1-4 should be fixed in this PR, (5) you can do in this PR if you want, (6) probably needs to be another PR (unless you think it's easy and want to take a stab at it). Here is the script I used for debugging, I think something like it should be the standard for checking that all our viz types still work, as all combos of standard/vec x surf/vol/mixed need to always work without error:

<details>

```
import os.path as op
import numpy as np
import mne

inv_surf = mne.minimum_norm.read_inverse_operator(op.join(
src_surf = mne.read_source_spaces(op.join(
    subjects_dir, 'sample', 'bem', 'sample-oct-6-src.fif'))
stc_surf = mne.minimum_norm.apply_inverse(
    evoked, inv_surf, pick_ori='vector')

inv_vol = mne.minimum_norm.read_inverse_operator(op.join(
src_vol = mne.read_source_spaces(op.join(
    subjects_dir, 'sample', 'bem', 'volume-7mm-src.fif'))
stc_vol = mne.minimum_norm.apply_inverse(evoked, inv_vol, pick_ori='vector')

# This is a hack to get a mixed stc but it works
stc_mixed = mne.MixedVectorSourceEstimate(
    np.concatenate([stc_surf.data, stc_vol.data]),
    stc_surf.vertices + stc_vol.vertices, stc_surf.tmin, stc_surf.tstep)
src_mixed = src_surf + src_vol

kwargs = dict(clim=dict(kind='value', lims=[6, 9, 12]), initial_time=0.1,
              subjects_dir=subjects_dir, hemi='split', size=(900, 450),
              subject='sample')

# surface
brain_surf_vec = stc_surf.plot(src=src_surf, **kwargs)
brain_surf = stc_surf.magnitude().plot(src=src_surf, **kwargs)

# volume
brain_vol_vec = stc_vol.plot_3d(src=src_vol, **kwargs)
brain_vol = stc_vol.magnitude().plot_3d(src=src_vol, **kwargs)

# mixed
brain_mixed_vec = stc_mixed.plot(src=src_mixed, **kwargs)
brain_mixed = stc_mixed.magnitude().plot(src=src_mixed, **kwargs)
```

</details>

I know somewhere we parametrize over `surf`/`mixed` at least in the `_brain` tests, can you check that all three types are checked properly for this new functionality? Perhaps we need to extend them to vector/3D vs 1D estimates as well...

GuillaumeFavelier:
I addressed 1-4 of https://github.com/mne-tools/mne-python/pull/8335#issuecomment-742531087 @larsoner 

> can you check that all three types are checked properly for this new functionality? Perhaps we need to extend them to vector

I updated the parameter list in `test_brain_traces`, feel free to take a look.

agramfort (approved):
works like a charm for me !

agramfort:
my only concern is about the red CI and its running time that keeps increasing...

GuillaumeFavelier:
>  running time that keeps increasing...

A good part of this PR is only testing

GuillaumeFavelier:
I tried restarting the failed job `Windows 3.8 pip pre`

mne/viz/_brain/_brain.py
```diff
@@ -1958,7 +2096,7 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
                 filepath = label
                 label = read_label(filepath)
                 hemi = label.hemi
-                label_name = os.path.basename(filepath).split('.')[1]
+                label_name = os.path.basename(filepath).split('.')[0]
```
larsoner, line 2099: Most are named things like `lh.BA1.label`, this seems wrong?

larsoner (commented):
1. With a vector surface STC, "pca_flip" is shown as a mode but it shouldn't be:
    ```
    ValueError: Invalid value for the 'mode' parameter when using a vector and/or volume source estimate. Allowed values are 'mean', 'max', and 'auto', but got 'pca_flip' instead.
    ```
2. When closing the standard surface (non-vector) window with some labels picked, I got a bunch of warnings of the form:
    ```
    /home/larsoner/python/pyvistaqt/pyvistaqt/window.py:23: RuntimeWarning: Could not find value: #1f77b4
      self.signal_close.emit()
    /home/larsoner/python/pyvistaqt/pyvistaqt/window.py:23: RuntimeWarning: Could not find value: #ff7f0e
      self.signal_close.emit()
    ...
    ```
    But I can't seem to replicate, so maybe it had to do with the error I got when trying to extract the vector one earlier.
3. When switching from point to label picking, you can actually see each point get removed very quickly, suggesting that a `.update`/`.render` is called after each removal. It should speed things up not to update until the end

I only see (1) as a blocker here, otherwise it seems to work well!

GuillaumeFavelier:
I am not able to reproduce 2) and I am aware of 3) Probably special treatment is required: the code to "remove all spheres" just removes one sphere at a time which leads to this effect.

> It should speed things up not to update until the end

I agree

mne/viz/_brain/_brain.py
```diff
@@ -1163,9 +1163,15 @@ def _set_label_mode(mode):
...
         self._label_mode_widget = QComboBox()
         self.tool_bar.addWidget(self._label_mode_widget)
+        if self._data["stc"] is not None:
+            vec = self._data["stc"]._data_ndim == 3
+        else:
+            vec = False
         for source in ["stc", "src"]:
             if self._data[source] is not None:
                 for mode in self.default_label_extract_modes[source]:
+                    if vec and mode == "pca_flip":
```
larsoner, line 1173: It's not just `pca_flip`, `mean_flip` also will not work. Instead of duplicating checks you should DRY with this somehow:

https://github.com/mne-tools/mne-python/blob/master/mne/source_estimate.py#L3085-L3092

larsoner:
> the code to "remove all spheres" just removes one sphere at a time which leads to this effect.

It seems like the correct fix is to add some param to PyVista to optionally not call `self.update` when doing `remove_actor` or so, WDYT? I assume it does this under the hood but haven't checked. In any case, someone must be telling the renderer to re-render, I doubt VTK does it automatically when an actor is removed...

GuillaumeFavelier:
> It seems like the correct fix is to add some param to PyVista to optionally not call `self.update`

This parameter already exists in [`remove_actor`](https://docs.pyvista.org/plotting/plotting.html?highlight=remove_actor#pyvista.BasePlotter.remove_actor), I updated the function in 0594376

GuillaumeFavelier:
I investigate the failure on `linux / conda / py3.8`:

First, I restart the job to confirm a consistent issue.

larsoner:
Thanks @GuillaumeFavelier !

larsoner:
@GuillaumeFavelier we have a problem:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/5365/workflows/9357aa95-93ca-45f4-9de3-a244e1531e7a/jobs/24188

larsoner:
FYI I hotfixed `maint/0.22` and `master` to just skip the brain-gc during the doc builds, e.g.:

https://github.com/mne-tools/mne-python/commit/0efe6ae6643123c99cd90a7228c8db04a3569c82

This should be reverted in whatever PR you open to fix this, and the PR run with `[circle full]`. That will tell us if things work, then we merge and backport.

GuillaumeFavelier:
For future reference, the log on Circle is:

<details>

```
generating gallery for auto_tutorials/source-modeling... [ 46%] plot_eeg_mri_coords.py
1082.3 s : 1.15 GB    
generating gallery for auto_tutorials/source-modeling... [ 53%] plot_eeg_no_mri.py
1135.8 s : 1.22 GB    
generating gallery for auto_tutorials/source-modeling... [ 60%] plot_fix_bem_in_blender.py
1160.2 s : 1.32 GB    
generating gallery for auto_tutorials/source-modeling... [ 66%] plot_forward.py
1164.3 s : 1.32 GB    
generating gallery for auto_tutorials/source-modeling... [ 73%] plot_mne_dspm_source_localization.py
1231.7 s : 1.32 GB    
Exception ignored in: <finalize object at 0x7f9fc23a9600; dead>
  File "/usr/local/lib/python3.8/weakref.py", line 566, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/usr/local/lib/python3.8/tempfile.py", line 819, in _cleanup
    _warnings.warn(warn_message, ResourceWarning)
ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmpyb2dzodj'>
generating gallery for auto_tutorials/source-modeling... [ 80%] plot_mne_solutions.py
1274.1 s : 1.26 GB    
generating gallery for auto_tutorials/source-modeling... [ 86%] plot_object_source_estimate.py

  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/events.py", line 110, in emit
    results.append(listener.handler(self.app, *args))
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/gen_gallery.py", line 450, in generate_gallery_rst
    generate_dir_rst(src_dir, target_dir, gallery_conf,
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/gen_rst.py", line 364, in generate_dir_rst
    intro, title, cost = generate_file_rst(
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/gen_rst.py", line 842, in generate_file_rst
    clean_modules(gallery_conf, fname)
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/scrapers.py", line 474, in clean_modules
    reset_module(gallery_conf, fname)
  File "/home/circleci/project/doc/conf.py", line 462, in __call__
    _assert_no_instances(Brain, 'running')  # calls gc.collect()
  File "/home/circleci/project/mne/utils/misc.py", line 351, in _assert_no_instances
    assert n == 0, f'{n} {when}:\n' + '\n'.join(ref)
AssertionError: 8 running:
cell: <cell at 0x7f9fd69dd130: Brain object at 0x7f9f7421ee50>
cell: <cell at 0x7f9fabffd6d0: Brain object at 0x7f9fc1d3a3a0>
cell: <cell at 0x7f9fc1fbbaf0: Brain object at 0x7f9fc1fbbd90>
cell: <cell at 0x7f9fc6ab56a0: Brain object at 0x7f9fd661b1c0>
cell: <cell at 0x7f9fc6be7640: Brain object at 0x7f9fc68b9a60>
cell: <cell at 0x7f9fc0a581f0: Brain object at 0x7fa01fd14610>
cell: <cell at 0x7f9fc1f25490: Brain object at 0x7f9fc5353280>
cell: <cell at 0x7f9fc0d1c160: Brain object at 0x7f9f741c1c10>

The above exception was the direct cause of the following exception:

  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/cmd/build.py", line 276, in build_main
    app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/application.py", line 278, in __init__
    self._init_builder()
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/application.py", line 337, in _init_builder
    self.events.emit('builder-inited')
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx/events.py", line 117, in emit
    raise ExtensionError(__("Handler %r for event %r threw an exception") %
sphinx.errors.ExtensionError: Handler <function generate_gallery_rst at 0x7fa0409f5940> for event 'builder-inited' threw an exception (exception: 8 running:
cell: <cell at 0x7f9fd69dd130: Brain object at 0x7f9f7421ee50>
cell: <cell at 0x7f9fabffd6d0: Brain object at 0x7f9fc1d3a3a0>
cell: <cell at 0x7f9fc1fbbaf0: Brain object at 0x7f9fc1fbbd90>
cell: <cell at 0x7f9fc6ab56a0: Brain object at 0x7f9fd661b1c0>
cell: <cell at 0x7f9fc6be7640: Brain object at 0x7f9fc68b9a60>
cell: <cell at 0x7f9fc0a581f0: Brain object at 0x7fa01fd14610>
cell: <cell at 0x7f9fc1f25490: Brain object at 0x7f9fc5353280>
cell: <cell at 0x7f9fc0d1c160: Brain object at 0x7f9f741c1c10>)

Extension error:
Handler <function generate_gallery_rst at 0x7fa0409f5940> for event 'builder-inited' threw an exception (exception: 8 running:
cell: <cell at 0x7f9fd69dd130: Brain object at 0x7f9f7421ee50>
cell: <cell at 0x7f9fabffd6d0: Brain object at 0x7f9fc1d3a3a0>
cell: <cell at 0x7f9fc1fbbaf0: Brain object at 0x7f9fc1fbbd90>
cell: <cell at 0x7f9fc6ab56a0: Brain object at 0x7f9fd661b1c0>
cell: <cell at 0x7f9fc6be7640: Brain object at 0x7f9fc68b9a60>
cell: <cell at 0x7f9fc0a581f0: Brain object at 0x7fa01fd14610>
cell: <cell at 0x7f9fc1f25490: Brain object at 0x7f9fc5353280>
cell: <cell at 0x7f9fc0d1c160: Brain object at 0x7f9f741c1c10>)
ERROR:mprof:Program resulted with a non-zero exit code: 2
make: *** [Makefile:45: html_stable-memory] Error 2

Exited with code exit status 2

CircleCI received exit code 2
```

</details>

PR #8340, opened 2020-10-04
BUG: Split file issue on dashes

### LINKED ISSUE #8339: automatically combining split files with dashes in the filename fails
Hi, 

I am not 100% this is a bug, or rather a feature request. For now, i could find a work-around, but if it is a bug, it might be relevant to more users in the future. 

####  Describe the bug
it seems that split MEG files (automatically split by the acquisition software) cannot be automatically rejoined when the filenames contains dashes (`-`). At least, when I tried to read a BIDS-formatted file, the combination of split files failed. 

#### Steps to reproduce
run `mne.io.read_raw_fif(infiles,preload=False)` on a file with dashes in it, e.g `sub-<sub_no>_ses-<ses_no>_task-<task_id>_raw.fif`

If necessary, I can upload two files as a MWE, but I suppose it will happen with any files as long as there is a dash in the filename. 

#### Expected results
I expected both file parts, so `...raw.fif` and `...raw-1.fif` to be automatically detected and read. 

#### Actual results
Only, the first part was read, the second one was ignored. 

#### Additional information
I think I located the problem to  `_get_next_fname` in `https://github.com/mne-tools/mne-python/blob/master/mne/io/open.py`. In this function, in line 83, MNE tries to check whether this is the first file or one of the subsequent parts, by locating the dashes in the filename. If `idx2 = -1`, then there are not dashes and the filename is the first part (`_raw.fif`), if `idx2 != -1, but the part of the filename between idx2 and idx is a number, then this is already a subsequent file (`...raw-[n].fif`), and a potential next filename (`..._raw-[n+1].fif`) is generated. 
However, when the filename contains dashes any other place (like it is common for BIDS files), this procedure fails, because when the filename of first part is analysed, dashes will be detected, and the next_file_name will be set to None. 

As a quick fix, to check whether I understood the problem, I used a different character for split files detection (`#` instead of `-`), and then dashes in the filename don't cause problems. 

Therefore, the solution would entail making the procedure robust against arbitrary dashes in the filename. I propose a fix as a pull request, not sure whether it is sufficiently robust.

eort:
#### Reference issue
Potentially fixes  #8339

#### Additional information
Local tests with a few variations on filenames worked, but I did not extensive testing. 

The most likely source for problems in this commit is that I did not understand what the function of former lines 90 and 91 were. For my proposed fix, these lines would not do anything, so I removed them.

Hope that's useful.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,6 +23,7 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix bug with reading split files that have dashes in the filename by `Eduard Ort`_ (:gh:`8339`)
 
 - Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (:gh:`8285`)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -12,6 +12,8 @@
 Current (0.22.dev0)
 -------------------
 
+.. |Eduard Ort| replace:: **Eduard Ort**
+
 Enhancements
 ~~~~~~~~~~~~
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -32,6 +32,10 @@ Bugs
 
 - :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
 
+- Fix bug with :class:`mne.preprocessing.ICA` where ``n_pca_components`` as a :class:`python:float` would give the number of components that explained less than or equal to the given variance. It now gives greater than the given number for better usability and consistency with :class:`sklearn.decomposition.PCA`. Generally this will mean that one more component will be included, by `Eric Larson`_ (:gh:`8326`)
+
+- Fix bug with :func:`mne.preprocessing.read_ica_eeglab` where full-rank data were not handled properly by `Eric Larson`_ (:gh:`8326`)
+
 - Fix bug with :ref:`somato-dataset` where the BEM was not included by `Eric Larson`_ (:gh:`8317`)
 
 - Fix missing documentation of :func:`mne.io.read_raw_nihon` in :ref:`tut-imorting-eeg-data` by `Adam Li`_ (:gh`8320`)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -26,6 +26,8 @@ Bugs
 
 - Fix bug with :func:`mne.viz.plot_source_estimates` when using the PyVista backend where singleton time points were not handled properly by `Eric Larson`_ (:gh:`8285`)
 
+- Fix bug when passing ``axes`` to plotting functions, :func:`matplotlib.pyplot.tight_layout` will not be called when the figure was created using a constrained layout by `Eric Larson`_ (:gh:`8344`)
+
 - Fix bug with compensated CTF data when picking channels without preload by `Eric Larson`_ (:gh:`8318`)
 
 - Fix bug when merging fNIRS channels in :func:`mne.viz.plot_evoked_topomap` and related functions by `Robert Luke`_ (:gh:`8306`)
@@ -33,6 +35,8 @@ Bugs
 - :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
 
 - Fix bug with :class:`mne.preprocessing.ICA` where ``n_pca_components`` as a :class:`python:float` would give the number of components that explained less than or equal to the given variance. It now gives greater than the given number for better usability and consistency with :class:`sklearn.decomposition.PCA`. Generally this will mean that one more component will be included, by `Eric Larson`_ (:gh:`8326`)
+
+- Fix bug with :class:`mne.preprocessing.ICA` where projections were not tracked properly by `Eric Larson`_
 
 - Fix bug with :func:`mne.preprocessing.read_ica_eeglab` where full-rank data were not handled properly by `Eric Larson`_ (:gh:`8326`)
 
@@ -44,6 +48,7 @@ Bugs
 
 - ``ICA.max_pca_components`` will not be altered by calling `~mne.preprocessing.ICA.fit` anymore. Instead, the new attribute ``ICA.max_pca_components_`` will be set, by `Richard Höchenberger`_ (:gh:`8321`)
 
+- Fix bug that `~mne.viz.plot_ica_overlay` would sometimes not create red traces, by `Richard Höchenberger`_ (:gh:`8341`)
 
 API changes
 ~~~~~~~~~~~
--- a/doc/changes/names.inc
+++ b/doc/changes/names.inc
@@ -325,3 +325,5 @@
 .. _Jeroen Van Der Donckt: https://github.com/jvdd
 
 .. _Steven Bierer: https://github.com/neurolaunch
+
+.. _Eduard Ort: https://github.com/eort
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -294,8 +294,9 @@ general neuroimaging concepts. If you think a term is missing, please consider
     whitening
         A linear operation that transforms data with a known covariance
         structure into "whitened data" which has a covariance structure that
-        is the identity matrix (i.e., it creates virtual channels that are
-        uncorrelated and have unit variance).
+        is the identity matrix. In other words it creates virtual channels that
+        are uncorrelated and have unit variance. This is also known as a
+        sphering transformation.
 
         The term "whitening" comes from the fact that light with a flat
         frequency spectrum in the visible range is white, whereas
--- a/mne/io/fiff/tests/test_raw_fiff.py
+++ b/mne/io/fiff/tests/test_raw_fiff.py
@@ -20,7 +20,7 @@ import pytest
 from mne.datasets import testing
 from mne.filter import filter_data
 from mne.io.constants import FIFF
-from mne.io import RawArray, concatenate_raws, read_raw_fif
+from mne.io import RawArray, concatenate_raws, read_raw_fif, base
 from mne.io.tag import _read_tag_header
 from mne.io.tests.test_raw import _test_concat, _test_raw_reader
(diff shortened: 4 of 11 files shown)


hoechenberger:
Thanks for your contribution, @eort! 

@larsoner Can you have a look?

larsoner:
Thanks for looking into this!

> Local tests with a few variations on filenames worked, but I did not extensive testing.

Can you modify this test to save and load a split file using a name that fails on `master` but passes on this PR:

https://github.com/mne-tools/mne-python/blob/master/mne/io/fiff/tests/test_raw_fiff.py#L360-L361

It looks like [Azure CI style check](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=9405&view=logs&jobId=3f85e25b-fb23-5eb7-8385-51966a33df73&j=3f85e25b-fb23-5eb7-8385-51966a33df73&t=58c65f82-1ac8-5386-c698-54af79b6d3df) is unhappy with:
```
./mne/io/open.py:86:1: W293 blank line contains whitespace
./mne/io/open.py:93:1: W293 blank line contains whitespace
```

eort:
> Can you modify this test to save and load a split file using a name that fails on master but passes on this PR:

Well, sort of. I don't think I can properly test it, because the internals of the test fif file seem to be different from the fif file I experienced the error. Specifically, in the test file the next file name seems to be coded in `FIFF.FIFF_REF_FILE_NAME`, whereas in my file it is coded in `FIFF.FIFF_REF_FILE_NUM`. So, with the test file I can't reproduce the problems I have experienced. 

As a very crude approach to bypass this issue, I can force `open.py` to ignore the `FIFF.FIFF_REF_FILE_NAME` field and instead always use the `FIFF.FIFF_REF_FILE_NUM` field. When doing so, I can create a case such that reading a split file on `master` fails, while it works with this PR. However, the even though the PR works on that particular part, it doesn't pass the entire [test](https://github.com/mne-tools/mne-python/blob/master/mne/io/fiff/tests/test_raw_fiff.py#L360-L361). I suspect this is because I had to disable the `FIFF.FIFF_REF_FILE_NAME` field, in order to test the PR. 

A better way of testing would be to use a different test file, one that doesn't code the next file name as *name* but as *number*. As far as I understand this would be coded somewhere in the `fif` file itself, right? Can this be done when saving split files? With some pointers on how to do that, I could try to fix that. 

> It looks like Azure CI style check is unhappy with

Yes, I fixed that.

larsoner (commented):
> Specifically, in the test file the next file name seems to be coded in FIFF.FIFF_REF_FILE_NAME, whereas in my file it is coded in FIFF.FIFF_REF_FILE_NUM

I pushed a test in 57588ee that uses `monkeypatch` to prevent writing the name. This fails on `master` but passes on this PR, so we should be good to go from that end.

Can you update `latest.inc` in the `BUG` section? Then we can merge and backport

eort:
>  Can you update latest.inc in the BUG section? Then we can merge and backport

Done. 

One more thing, that extra test that you pushed does not test BIDS-formatted filenames (files ending on `...-1_meg.fif`). I don't think this PR would pass this test because of line [83](https://github.com/mne-tools/mne-python/blob/master/mne/io/open.py#L83) not being a number, but `<number>_meg`. In fact, I adapted your new test to check a bids-formatted file and indeed failed under this PR. 

Should BIDS-formatted files be covered with this function? Or should users just use the mne_bids for reading bids files? If split bids files should also be properly read with this function, the PR needs some fixing.

doc/changes/latest.inc
```diff
@@ -23,6 +23,7 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix bug with reading split files that have dashes in the filename by `Eduard Ort`_ (:gh:`8339`)
```
larsoner, line 26: I see you're a new contributor (hooray!) -- you need to add your name to `names.inc` with a link:

https://github.com/mne-tools/mne-python/blob/master/doc/changes/names.inc

We also now do some reformatting to bold the line for new contributors, see:

https://github.com/mne-tools/mne-python/blob/master/doc/changes/0.21.inc#L15

larsoner:
> In fact, I adapted your new test to check a bids-formatted file and indeed failed under this PR.

Does it pass on `master` but fail on this PR? Then yes it needs to be fixed in this PR. If no, then we can fix it in a follow-up PR if need be.

I'm not sure how often people will run into this problem / how problematic it actually is, though. MNE-Python (and MNE-BIDS) write the `FIFF_REF_FILE_NAME` tag. How were the files produced that didn't have the `FIFF_REF_FILE_NAME` tag? I'm not sure the BIDS-formatted filename is even `FIFF_REF_FILE_NUM`-compatible -- I'd assume that if only that tag is available, the subsequent ones must be named `-N.fif`.

eort:
> Does it pass on master but fail on this PR? 

No, it also fails on master.  So, overall it seems that splitting on numbers neither works if the filename contains dashes (fixed by this PR), nor if the filename is BIDS-formatted (would require a new PR, that would also affect the current PR).

>  How were the files produced that didn't have the FIFF_REF_FILE_NAME tag?
They were written by the MEG machine/acquisition software (Neuromag/Elekta scanner). 

> I'm not sure the BIDS-formatted filename is even FIFF_REF_FILE_NUM-compatible -- I'd assume that if only that tag is available, the subsequent ones must be named -N.fif

Possible. I agree that it probably doesn't affect many users. The only reason I bumped into this problem was that I was aware of BIDS (so I tried to have my files BIDS-formatted), but not of `mne_bids`. With mne_bids, it all works just fine. In that case, would it then make sense to show a warning, urging participants to use mne_bids when the filepath appears to be BIDS-formatted?

larsoner:
> No, it also fails on master. So, overall it seems that splitting on numbers neither works if the filename contains dashes (fixed by this PR), nor if the filename is BIDS-formatted (would require a new PR, that would also affect the current PR).

We already have tests for BIDS-formatting splits. They work just fine if you save using MNE-Python (master and this PR) when using `split_naming='bids'`:

https://github.com/mne-tools/mne-python/blob/master/mne/io/fiff/tests/test_raw_fiff.py#L370-L381

So the "problem" is really with the Neuromag software, or maybe more so with our expectations for it. If a given piece of software *only* writes out the `FILE_NUM` tag when splitting, I think this means that the files that get created *must* be named `*.fif` and `*-1.fif`. Is this the behavior you see when you try saving a file with the acquisiton software?

If that's the case then I think things with MNE-Python and mne-bids are fine, but people using Elekta acquisition software should not try to use BIDS filenames if they end up with split files because the software cannot write them properly. Or if they do it and end up with non-BIDS filenames, they need to use MNE-Python (with `split_naming='bids'`) to do a save-load round trip to get the names to be BIDS compatible.

eort:
> I think this means that the files that get created must be named *.fif and *-1.fif. Is this the behavior you see when you try saving a file with the acquisiton software?

Yes, split-files are written with `-1.fif` appended to the file name.

>people using Elekta acquisition software should not try to use BIDS filenames....Or if they do it and end up with non-BIDS filenames, they need to use MNE-Python (with split_naming='bids') to do a save-load round trip to get the names to be BIDS compatible.

That makes sense. But ideally this would also be somehow communicated. Otherwise the situation might be a little confusing. 
if you try to read a neuromag/bids-formatted file, on the current main, there is no sign that there is a problem (unless you have a clear expectation how long the recording should be). On the PR, you will get `RuntimeWarning: Split raw file detected but next file /tmp/pytest-of-ede/pytest-11/test_split_numbers0/sub-1_ses-2_task-3_split-01_meg.fi-2.fif does not exist.` or something like this.

larsoner:
> neuromag/bids-formatted file, on the current main, there is no sign that there is a problem

Can you make an issue for this on mne-bids? Perhaps they should detect the missing file warning and probably turn it into an error.

larsoner:
Thanks @eort !

eort:
> Can you make an issue for this on mne-bids? Perhaps they should detect the missing file warning and probably turn it into an error.

Sure, but why on mne-bids? The issue occurs when using `io.read_raw_fif(filename)`. I haven't tested what would happen when mne-bids was used for such files.

larsoner:
> Sure, but why on mne-bids? The issue occurs when using io.read_raw_fif(filename)

Let's talk about a concrete example. What use case are you trying to address where you will get this warning?

From what I understand, if you:

1. If you write data with MNE-Python or MNE-BIDS, things are fine and you can use BIDS names.
2. If you write data with Neuromag software, they don't name it according to BIDS, which is reasonable (they don't give this option, and have been writing files for years before BIDS existed)
3. If you rename the second part of the split file, it won't read properly -- but I'm not sure people would expect this behavior.

The motivation behind making changes in mne-bids is that mne-bids is all about conforming to a spec and making sure everything is in the right place. MNE-Python is more flexible and allows you to (and has allowed you to for quite a while) have these split files missing, and you get a warning.

I guess we could add a `on_split_missing='warn'` argument to `read_raw_fif` or so, but I have a feeling that this wouldn't actually address the problem much unless we change the default to `on_split_missing='raise'`, which we could consider doing. People who want to ignore a split file can always set this to `warn` or `ignore`.

eort:
One more case:
4. If you change both parts of a split file, produced with neuromag, to a BIDS name. 

But yes, this is a very specific case, that hardly ever will occur. As far as I see, there is only a problem when users try to read files with mne-python that were recorded on neuromag systems and were manually renamed to adhere to BIDS. Now that the PR is merged, split file names won't produce correct file names and an error will be raised. That could already be sufficient, but 
a warning a la "Don't manually rename files to BIDS and try to read them with mne-python, but just use mne-bids" might be helpful, though super-niche. Perhaps it would be better to address such behavior in the docs?

> The motivation behind making changes in mne-bids is that mne-bids is all about conforming to a spec and making sure everything is in the right place. 

No, I get that and I completely agree. I just haven't tested anything with mne-bids. So, I wasn't aware that the same behaviour would be observed when using mne-bids. If the same issue is present in mne-bids, then indeed, split files would just silently be ignored, if they were produced by the niche-case described above. 

> I guess we could add a on_split_missing='warn' argument to read_raw_fif or so, but I have a feeling that this wouldn't actually address the problem much unless we change the default to on_split_missing='raise', which we could consider doing. People who want to ignore a split file can always set this to warn or ignore.

I guess that would generally be a useful warning, but given that you normally have clear expectations regarding the length of your files perhaps not essential.

PR #8341, opened 2020-10-05
MRG, BUG: Fix line color in ICA.plot_overlay

hoechenberger:
`ICA.plot_overlay()` didn't seem to work on my system with latest Matplotlib anymore (macOS, both the default and Qt5Agg backends). I would only see black traces, instead of red and black. Debugging showed that changig the colors of lines in a figure (in this case, from black to read) that has
been `show()`n was somehow part of the problem. Now we only `show()` once all plotting operations are done (We call `plt_show(show)` at the end of the plotting function), and this now works with the default macOS and the Qt5Agg backends for me.


hoechenberger:
This should probably also be backported.

PR #8344, opened 2020-10-05
MRG, BUG: Fix handling of axes with constrained_layout=True

larsoner:
We should only internally do `tight_layout(...)` calls when axes were not created using `constrained_layout=True`. Our current code gets a matplotlib warning about disabling the `constrained_layout` of the axes.

Also fixes a bug where we would raise as error instead of re-emit as a warning


hoechenberger (approved):

hoechenberger:
One Travis job timed out after successfully running all tests.

hoechenberger:
Thanks @larsoner 

Can you do the backporting also?

PR #8348, opened 2020-10-06
ENH: Remove 15-char limit for FIF

### LINKED ISSUE #8312: Why is the channel name length limited to 15 characters?
MNE automatically truncates the channel name length to 15 characters: 
```

<ipython-input-68-1c3debadf790>:1: RuntimeWarning: 4 channel names are too long, have been truncated to 15 characters:
['LFP_BS_STN_R_234', 'LFP_BS_STN_R_567', 'LFP_BS_STN_L_234', 'LFP_BS_STN_L_567']
  info = mne.create_info(ch_labels, fsample, ch_types='ecog')
```

In the context of clinical neurophysiology it becomes very handy to keep information like 

1. the cortical location 
2. electrode manufacturer 
3. hemisphere 

inside of the channel name. 

Is there a reason for limiting the length?

larsoner:
Closes #8312.

- [x] Get it working
- [x] Communicate with Brainstorm and FieldTrip folks about this (https://github.com/brainstorm-tools/brainstorm3/issues/361, https://github.com/fieldtrip/fieldtrip/issues/1596)
- [ ] Add constant to fiff-constants (gives Jukka a chance to make it not in the MNE_ namespace if MEGIN would use it) (https://github.com/mne-tools/fiff-constants/pull/29)
- [ ] Update PR with newer `fiff-constants` commit, check `'bads'` handling/replacement
- [ ] Add test where a file is written with long names, then `monkeypatch` to make the tag empty on read, ensure it can still be written and read
- [ ] Add to MNE-MATLAB

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -17,6 +17,8 @@ Enhancements
 
 - Speed up :class:`mne.decoding.TimeDelayingRidge` with edge correction using Numba by `Eric Larson`_ (:gh:`8323`)
 
+- Remove the 15-character limitation for channel names when writing to FIF format. If you need the old 15-character names, you can use something like ``raw.rename_channels({n: n[:13] for n in raw.ch_names}, allow_duplicates=True)``, by `Eric Larson`_ (:gh:`8346`)
+
 - The ``max_pca_components`` argument of `~mne.preprocessing.ICA` can now be a float between 0.0 and 1.0, allowing users to specify a relative amount of cumulative explained variance. This can be used to exclude "unimportant" principal components, thereby conducting a rank reduction of the data, by `Richard Höchenberger`_ (:gh:`8321`)
 
 - `~mne.preprocessing.ICA` has gained a new attribute ``max_pca_components_``, which will be set when calling `~mne.preprocessing.ICA.fit`, by `Richard Höchenberger`_ (:gh:`8321`)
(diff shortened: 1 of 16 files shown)


bloyl:
+1

I love this idea.

I've taken to storing ROI source waveforms in `raw` objects to facilitate additional processing. So I run into this often and this approach would eliminate a lot of note keeping i currently need to do myself.

Do you think it makes sense to allow users to specify the renaming map? I'm not sure if it would offer any benefit other then allowing users some control of the names as they would appear in non-mne-based readers (besa etc).

bloyl:
Any thing I can do to push this forward? I'd love to start using it.

larsoner:
The next step I think is to make sure Brainstorm and FieldTrip people are on board. Happy to let you take the lead on that if you're up for it

bloyl:
Is there an mne-matlab PR they could test against?

larsoner:
No, didn't want to bother making one until we got some buy-in, but if you think the other order makes more sense I can cook one up

bloyl:
I'll open github issues and see what they think.

I think this is just a hidden change to fif file internals, so will be backward compatible and provided they use mne-matlab (for reading and writing) I don't think they will even notice.

bloyl:
Seems like an mne-matlab PR would be sufficient to support both brainstorm and fieldtrip.

PR #8350, opened 2020-10-07
MRG, BUG: Force keywords for verbose in short sigs

### LINKED ISSUE #8258: Restrict allowed "verbose" values, or make the verbose kwarg a keyword-only argument?

larsoner:
Closes #8258 
Closes #8263

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -46,6 +46,8 @@ Bugs
 
 - Fix bug with :func:`mne.add_reference_channels` when :func:`mne.io.Raw.reorder_channels` or related methods are used afterward by `Eric Larson`_ (:gh:`8303`)
 
+- Fix bug where the ``verbose`` arguments to :meth:`mne.Evoked.apply_baseline` and :meth:`mne.Epochs.apply_baseline` were not keyword-only by `Eric Larson`_ (:gh:`8349`)
+
 - ``ICA.max_pca_components`` will not be altered by calling `~mne.preprocessing.ICA.fit` anymore. Instead, the new attribute ``ICA.max_pca_components_`` will be set, by `Richard Höchenberger`_ (:gh:`8321`)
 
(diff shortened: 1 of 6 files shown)


hoechenberger:
Great!!

hoechenberger (approved):

hoechenberger:
Rendered API doc is now even messier than before, though:

Screenshot 2020-10-07 at 19 33 06

https://22643-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Epochs.html

larsoner:
Argh, that's the Sphinx tuple rendering bug...

larsoner:
backported

hoechenberger:
> Argh, that's the Sphinx tuple rendering bug...

Yep, and now with the `*` for kw-only args it looks even more hideous :D

larsoner:
Looks like the sphinx bug was fixed 4 days ago so hopefully it lands in some pre-release soon

larsoner:
https://github.com/sphinx-doc/sphinx/pull/8265

hoechenberger:
Super good!!!

PR #8351, opened 2020-10-07
MRG, API: Deprecate max_pca_components

### LINKED ISSUE #7727: Working with rank-deficient EEGLAB data

### LINKED ISSUE #8327: ICA should not touch init parameter n_pca_components during fit
This is what currently happens in `ICA._fit()`, which is invoked by `ICA.fit()`:
https://github.com/mne-tools/mne-python/blob/4270859dc3ba2208bb3eed4f73403b5de26125b8/mne/preprocessing/ica.py#L737-L739

We should have `ICA.n_pca_components_` instead (currently doesn't exist).

### LINKED ISSUE #8337: Proceeding with ICA works
This is just a rough sketchpad to help organize and keep track a little of the ICA construction site, because the big picture is fading quickly for me…

- [ ]  <strike>We now support floats in `max_pca_components`,  and we will keep this parameter</strike>
- [ ] <strike>We will drop `ICA.n_pca_components`, as typically, users can achieve similar results by controlling `max_pca_components` and `n_components`. This will close #8327</strike>
- [x] Ensure that projectors are handled better and that `n_pca_components=None` work (#8343)
- [x] Selection of number of components via `n_components` as a float should follow the "greater than" approach, not "less or equal than" (#8336, #8326)
- [x] Even in cases where we don't warn, display the estimated rank of the and potentially applied projectors, as suggested by @cbrnr (#8328)
- [ ] Next PR:
    - Drop `ICA.max_pca_components` via deprecation
    - Keep `ICA.n_components` and `ICA.n_pca_components`, keep `apply(..., n_pca_components)` to select a sub-set of all PCs when reconstructing the signal
    - Warn if ~~`n_pca_components`~~ `n_components` is likely to operate on rank-deficient data (produce ~zero-valued principal components) by checking the variance division values used when computing mixing_matrix_ from the unmixing_matrix_ (#8328)
    - Make sure `n_pca_components` is not changed during fit (should become n_pca_components_; #8327)
- [ ] Next PR: make n_pca_components and apply_baseline keyword-only by fixing decorator
How would we handle the deprecation of `max_pca_components`? Will we even need a deprecation cycle, since while we're changing the API, we're not removing **functionality**, and we could emit a fat helpful error message if sb tries to use this param?

### LINKED ISSUE #8342: viz.plot_ica_overlay needs an n_pca_components kwarg
While `ICA.apply()` does have `n_pca_components`, `ICA.plot_overlay()` does not, making it more difficult to interactively explore the effect of using different numbers of components for signal reconstruction.

### LINKED ISSUE #8353: BUG: 1. and 1 treated differently in ICA
From #8351.

My proposed solution is to disallow `1.`. We have `None` for "select all components", so we don't actually need `1.`.

larsoner:
Todo:

- [x] Deprecate `max_pca_components`
- [x] Get `test_ica.py` working
- [x] Add `n_pca_components` to `plot_ica_overlay`
- [x] Make keyword-only after #8350 is in
- [x] Update `plot_40_artifact_correction_ica.py`
- [x] Add warning when `n_components` leads to unstable pseudoinversion
- [x] Update `doc/_static/diagrams/ica.dot`
- [x] Decide if `n_components` as float selects based on the variance remaining of the `n_pca_components` components. This is how `max_pca_components` worked (it was relative to that)

Closes #8337
Closes #8342
Closes #8327
Closes #8328 
Closes #8316
Closes #8313
Closes #7727
Closes #8353

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -54,7 +54,7 @@ Bugs
 
 - Fix bug with :class:`~mne.preprocessing.ICA` where ``n_components=None, n_pca_components=None`` could lead to unstable unmixing matrix inversion by making ``n_components=None`` also use the lesser of ``n_components=0.999999`` and ``n_components=n_pca_components`` by `Eric Larson`_ (:gh:`8351`)
 
-- The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_
+- The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_ (:gh:`8351`)
 
 API changes
 ~~~~~~~~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -50,6 +50,8 @@ Bugs
 
 - Fix bug that `~mne.viz.plot_ica_overlay` would sometimes not create red traces, by `Richard Höchenberger`_ (:gh:`8341`)
 
+- Fix bug with :class:`~mne.preprocessing.ICA` where ``n_components=None, n_pca_components=None`` could lead to unstable unmixing matrix inversion by making ``n_components=None`` also use the lesser of ``n_components=0.999999`` and ``n_components=n_pca_components`` by `Eric Larson`_ (:gh:`8351`)
+
 - The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_
 
 API changes
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -17,6 +17,8 @@ Enhancements
 
 - Add ``n_pca_components`` argument to :func:`mne.viz.plot_ica_overlay` by `Eric Larson`_ (:gh:`8351`)
 
+- Add ``proj`` argument to :func:`mne.make_fixed_length_epochs` by `Eric Larson`_ (:gh:`8351`)
+
 - Speed up :class:`mne.decoding.TimeDelayingRidge` with edge correction using Numba by `Eric Larson`_ (:gh:`8323`)
 
 Bugs
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -15,11 +15,9 @@ Current (0.22.dev0)
 Enhancements
 ~~~~~~~~~~~~
 
+- Add ``n_pca_components`` argument to :func:`mne.viz.plot_ica_overlay` by `Eric Larson`_ (:gh:`8351`)
+
 - Speed up :class:`mne.decoding.TimeDelayingRidge` with edge correction using Numba by `Eric Larson`_ (:gh:`8323`)
-
-- The ``max_pca_components`` argument of `~mne.preprocessing.ICA` can now be a float between 0.0 and 1.0, allowing users to specify a relative amount of cumulative explained variance. This can be used to exclude "unimportant" principal components, thereby conducting a rank reduction of the data, by `Richard Höchenberger`_ (:gh:`8321`)
-
-- `~mne.preprocessing.ICA` has gained a new attribute ``max_pca_components_``, which will be set when calling `~mne.preprocessing.ICA.fit`, by `Richard Höchenberger`_ (:gh:`8321`)
 
 Bugs
 ~~~~
@@ -52,7 +50,9 @@ Bugs
 
 - Fix bug that `~mne.viz.plot_ica_overlay` would sometimes not create red traces, by `Richard Höchenberger`_ (:gh:`8341`)
 
+- The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_
+
 API changes
 ~~~~~~~~~~~
 
-...
+- The ``max_pca_components`` argument of :class:`~mne.preprocessing.ICA` has been deprecated, use ``n_pca_components`` instead by `Eric Larson`_
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -122,6 +122,9 @@ def _check_for_unsupported_ica_channels(picks, info, allow_ref_meg=False):
                          % (_pl(chs), chs, types))
 
 
+_KNOWN_ICA_METHODS = ('fastica', 'infomax', 'picard')
+
+
 @fill_doc
 class ICA(ContainsMixin):
     u"""Data decomposition using Independent Component Analysis (ICA).
@@ -360,7 +363,7 @@ class ICA(ContainsMixin):
     """  # noqa: E501
 
     @verbose
-    def __init__(self, n_components=None, max_pca_components=None,
+    def __init__(self, n_components=None, *, max_pca_components=None,
                  n_pca_components=None, noise_cov=None, random_state=None,
                  method='fastica', fit_params=None, max_iter=200,
                  allow_ref_meg=False, verbose=None):  # noqa: D102
@@ -374,7 +377,7 @@ class ICA(ContainsMixin):
                  DeprecationWarning)
 
         if method != 'imported_eeglab':  # internal use only
-            _check_option('method', method, ['fastica', 'infomax', 'picard'])
+            _check_option('method', method, _KNOWN_ICA_METHODS)
         if method == 'fastica' and not check_version('sklearn'):
             raise ImportError(
                 'The scikit-learn package is required for method="fastica".')
@@ -2342,6 +2345,9 @@ def read_ica(fname, verbose=None):
     ica_init, ica_misc = [_deserialize(k) for k in (ica_init, ica_misc)]
     current_fit = ica_init.pop('current_fit')
     max_pca_components = ica_init.pop('max_pca_components')
+    method = ica_misc.get('method', 'fastica')
+    if method in _KNOWN_ICA_METHODS:
+        ica_init['method'] = method
     if ica_init['noise_cov'] == Covariance.__name__:
         logger.info('Reading whitener drawn from noise covariance ...')
 
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -150,12 +150,12 @@ class ICA(ContainsMixin):
             Will select the smallest number of components required to explain
             the cumulative variance of the data greater than ``n_components``.
             Consider this hypothetical example: we have 3 components, the first
-            explaining 70%, the second 20%, and the third the remaining 10% of
-            the variance. Passing 0.8 here (corresponding to 80% of explained
-            variance) would yield the first two components, explaining 90% of
-            the variance: only by using both components the requested threshold
-            of 80% explained variance can be exceeded. The third component, on
-            the other hand, would be excluded.
+            explaining 70%%, the second 20%%, and the third the remaining 10%%
+            of the variance. Passing 0.8 here (corresponding to 80%% of
+            explained variance) would yield the first two components,
+            explaining 90%% of the variance: only by using both components the
+            requested threshold of 80%% explained variance can be exceeded. The
+            third component, on the other hand, would be excluded.
         - ``None``
             ``n_pca_components`` or ``0.999999`` will be used, whichever
             results in fewer components. This is done to avoid numerical
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -154,7 +154,8 @@ class ICA(ContainsMixin):
         - ``None``
             ``n_pca_components`` or ``0.999999`` will be used, whichever
             results in fewer components. This is done to avoid numerical
-            stability problems when whitening.
+            stability problems when whitening, particularly when working
+            with rank-deficient data.
 
         Defaults to ``None``. The actual number used when executing the
         :meth:`ICA.fit` method will be stored in the attribute
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -147,10 +147,15 @@ class ICA(ContainsMixin):
             Must be greater than 1 and less than or equal to
             ``n_pca_components``.
         - :class:`float` between 0 and 1 (exclusive)
-            The number of components with cumulative explained variance
-            of the data greater than ``n_components`` will be used. For
-            example, if 0.8 is passed then the number of components will
-            explain greater than 80%% of the variance.
+            Will select the smallest number of components required to explain
+            the cumulative variance of the data greater than ``n_components``.
+            Consider this hypothetical example: we have 3 components, the first
+            explaining 70%, the second 20%, and the third the remaining 10% of
+            the variance. Passing 0.8 here (corresponding to 80% of explained
+            variance) would yield the first two components, explaining 90% of
+            the variance: only by using both components the requested threshold
+            of 80% explained variance can be exceeded. The third component, on
+            the other hand, would be excluded.
         - ``None``
             ``n_pca_components`` or ``0.999999`` will be used, whichever
             results in fewer components. This is done to avoid numerical
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -776,7 +776,19 @@ class ICA(ContainsMixin):
             self.n_iter_ = n_iter + 1  # picard() starts counting at 0
             del _, n_iter
         assert self.unmixing_matrix_.shape == (self.n_components_,) * 2
-        norms = np.sqrt(self.pca_explained_variance_[:self.n_components_])
+        norms = self.pca_explained_variance_
+        stable = norms / norms[0] > 1e-6
+        norms = norms[:self.n_components_]
+        if not stable[self.n_components_ - 1]:
+            max_int = np.where(stable)[0][-1] + 1
+            warn(f'Using n_components={self.n_components} (resulting in '
+                 f'n_components_={self.n_components_}) may lead to an '
+                 f'unstable mixing matrix estimation because the ratio '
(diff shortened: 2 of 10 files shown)


larsoner:
Okay this is ready for review/merge from my end

larsoner:
Travis failure is an unrelated conda-forge python 3.9-related bug that's hitting all PRs/branches

mne/preprocessing/ica.py
```diff
@@ -141,48 +144,44 @@ class ICA(ContainsMixin):
         are passed to the ICA algorithm during fitting. If :class:`int`, must
         not be larger than ``max_pca_components``. If :class:`float` between 0
```
agramfort, line 145: max_pca_components is still here

mne/preprocessing/ica.py
```diff
@@ -141,48 +144,44 @@ class ICA(ContainsMixin):
...
-        `~mne.preprocessing.ICA.fit`. Defaults to ``None``. When calling
-        `~mne.preprocessing.ICA.fit`, the attribute ``max_pca_components_``
-        will be set to the absolute number of retained PCA components.
 
         .. versionchanged:: 0.22
-            Added support for float to select components by cumulative
-            explained variance.
+           The number of components returned for a :class:`python:float`
+           is greater than the given variance level instead of less than or
+           equal to it.
```
agramfort, line 156: this sentence is quite mysterious I fear.

Maybe something like:

If :class:`python:float` it must be less than 1 and it corresponds to the threshold
of explained variance. For example if 0.8 is passed than the number of components
will explain at least 80% of the variance.

mne/preprocessing/ica.py
```diff
@@ -682,68 +692,50 @@ def _fit(self, data, fit_type):
...
+        elif self.n_pca_components is not None:
+            self.n_components_ = self.n_pca_components
+        else:  # Both None
+            self.n_components_ = len(pca.components_)
+            msg = 'Selecting all PCA components'
+        if isinstance(self.n_components_, float):
+            # XXX not 100% sure if this should be use_ev or use_ev[:n_pca],
+            # i.e., is n_components float the threshold for the explained
+            # variance remaining after subselecting n_pca_components, or is
+            # it from the original data
```
agramfort, line 720: ?

mne/preprocessing/ica.py
```diff
@@ -753,38 +745,50 @@ def _fit(self, data, fit_type):
...
+            _, W, _, n_iter = picard(
+                data[:, sel].T, whiten=False, return_n_iter=True,
+                random_state=random_state, **self.fit_params)
             self.unmixing_matrix_ = W
             self.n_iter_ = n_iter + 1  # picard() starts counting at 0
             del _, n_iter
         assert self.unmixing_matrix_.shape == (self.n_components_,) * 2
-        norms = np.sqrt(self.pca_explained_variance_[:self.n_components_])
+        norms = self.pca_explained_variance_
+        stable = norms / norms[0] > 1e-6
```
agramfort, line 780: add a comment why 1e-6

agramfort (commented):
@hoechenberger and @cbrnr you might want to review and test this branch before it gets merged

mne/preprocessing/ica.py
larsoner: This is the point at the top of the issue that needs resolving.

Concete example. On `master`
- Assumue data is 4 channels with PCA exp var ratios 0.4, 0.3, 0.2, 0.1
- Doing `max_pca_components=3, n_components=1.` (or None) will yield only 3 components to ICA (`n_components_=3`)
- Doing `max_pca_components=4, n_components=1.` (or None) will yield 4 components in the ICA (`n_components_=4`).

This is because the `n_components=1.` is relative *to the variance explained by the max_pca_components*, not of the entire data (if it were the entire data, the first case above would need to raise an error because you'd be requesting 4 ICA components but only using 3 PCA ones).

If we kill `max_pca_components`, what should `n_pca_components=3, n_components=1.` mean? Should it yield `n_components_=4`, or `n_components_=3`? In other words, should the float `n_components` be relative to:

1. The explained variance across just `n_pca_components` -- this is my preference; it's implemented here as it required fewer tweaks to the tests; and is consistent with the behavior before if we're thinking of the `n_pca_components` in init now taking over the role of `max_pca_components`
2. The total explained variance in your data?

The only problem I've seen with (1) is that it means that code that looked like:

- `max_pca_components=None, n_pca_components=3, n_components=1.`

in master will have `n_components_=4` and on this PR it will have `n_components_=3`. But everywhere in our code base where we set `n_pca_components=N` as an int we had `max_pca_components=N` as well, so I'm hoping this might qualify as a bug fix of sorts (it's weird to have more components in the ICA step than you plan on reconstructing in the PCA step).

agramfort:
> what should n_pca_components=3, n_components=1. mean?

should break. you cannot limit the number of PCA components and require all
the variance to be explained
for ICA.

larsoner:
> should break. you cannot limit the number of PCA components and require all the variance to be explained for ICA.

It depends on if you're talking about wanting the `n_components=1.` fraction of explained variance of the *N PCA components* (which is how it works in `master` with `max_pca_components_`) or the *total variance of your original data* (what you're saying/assuming). I can certainly code it this way, and if I were designing it from scratch this is probably how I would do it, but I suspect this is going to break more people's code than the way I have it coded now.

larsoner:
(It certainly breaks more tests, but those are quick enough to fix)

cbrnr:
Wait, are we really distinguishing `1` from `1.0`? That's a bit unexpected, so can we find a better solution for this edge case (wanting just 1 component should be rarely needed, whereas all components corresponding to a fraction of 1.0 should be the default).

larsoner:
> Wait, are we really distinguishing 1 from 1.0

Yes float and int are different, I can add a test that `n_components=1` will raise an error (it does already at least when selecting by exp var, I can make sure it does for the int case)

> whereas all components corresponding to a fraction of 1.0 should be the default

This is the default as `n_components=None` means "use all"

hoechenberger:
@larsoner 
> It depends on if you're talking about wanting the `n_components=1.` fraction of explained variance of the _N PCA components_ (which is how it works in `master` with `max_pca_components`) or the _total variance of your original data_ (what you're saying/assuming). I can certainly code it this way, and if I were designing it from scratch this is probably how I would do it, but I suspect this is going to break more people's code than the way I have it coded now.

If we're re-thinking these questions from scratch anyway, I'd prefer if `n_components: float` would refer to the fraction of **total** explained variance. So we'd need to check if `n_components: float <= n_pca_components`

I cannot really imagine anyone would want a fraction of a fraction?

hoechenberger:
> > Wait, are we really distinguishing 1 from 1.0
> 
> Yes float and int are different, I can add a test that `n_components=1` will raise an error (it does already at least when selecting by exp var, I can make sure it does for the int case)

I believe this would lead to a cleaner API, yes. Good idea.

cbrnr:
What if I do want to select just one component?

hoechenberger:
> What if I do want to select just one component?

Is that an actual use case or just a theoretical consideration?

cbrnr:
> Is that an actual use case or just a theoretical consideration?

It's theoretical, but I'm pretty sure someone will want to use it for a good reason.

Also, it is a pretty unusual behavior to treat `1` and `1.0` differently, because even though the types are different, `1 == 1.0`, and I'd prefer a solution where we can avoid this case.

hoechenberger:
But ok, allowing 1 as int probably won't hurt anyone either..

cbrnr:
> But ok, allowing 1 as int probably won't hurt anyone either..

If both 1 and 1.0 do they same thing, then yes.

larsoner:
> If both 1 and 1.0 do they same thing, then yes.

Can you open a separate issue about this? Let's solve it and discuss it elsewhere

agramfort:
but using 1 component for ICA does not make sense it should not be allowed
anyway

>

mne/preprocessing/ica.py
```diff
@@ -138,51 +141,58 @@ class ICA(ContainsMixin):
...
+        - :class:`float` between 0 and 1 (exclusive)
+            The number of components with cumulative explained variance
+            of the data greater than ``n_components`` will be used. For
+            example, if 0.8 is passed then the number of components will
+            explain greater than 80%% of the variance.
+        - ``None``
+            ``n_pca_components`` or ``0.999999`` will be used, whichever
+            results in fewer components. This is done to avoid numerical
+            stability problems when whitening.
+
```
larsoner, line 158: This is an important change/fix I think. The defaults of `n_pca_components=None` and `n_components=None` are still dangerous for rank-reduced data. This makes our defaults safe for rank-reduced data.

larsoner (commented):
- docs updated
- killed 1. (float) and 1 (int) to be safe
- updated behavior of `n_components=None` (see inline comment)
- made `n_components` float value operate on total explained variance, not relative to `n_pca_components`

larsoner:
And the updated docs/diagram (formerly missing the pre-whitening info):

https://22658-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_40_artifact_correction_ica.html#ica-in-mne-python

hoechenberger (commented):
Why do we insist on the float values referring to **more** than the specified explained variance? I'd prefer if we'd change this to select components that explain **at least** the specified variance…

I.e. I would change `def _exp_var_ncomp()` such that the current

```python
    n = min((cvar <= n).sum() + 1, len(cvar))
```

becomes
```python
    n = min((cvar < n).sum() + 1, len(cvar))
```
No?

(also I believe we can drop the `min()` now that we disallow the `1` and `1.0` cases)

mne/preprocessing/ica.py
```diff
@@ -138,51 +141,58 @@ class ICA(ContainsMixin):
...
+            ``n_pca_components``.
+        - :class:`float` between 0 and 1 (exclusive)
+            The number of components with cumulative explained variance
+            of the data greater than ``n_components`` will be used. For
+            example, if 0.8 is passed then the number of components will
+            explain greater than 80%% of the variance.
+        - ``None``
+            ``n_pca_components`` or ``0.999999`` will be used, whichever
+            results in fewer components. This is done to avoid numerical
+            stability problems when whitening.
```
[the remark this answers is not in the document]
larsoner: In theory it can also happen when working with data with one huge first component, but I'll reword it to make it clear it's usually due to rank deficiency.

larsoner:
> Why do we insist on the float values referring to more than the specified explained variance?

To [be consistent](https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/decomposition/_pca.py#L476-L478) with [sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html):

"If 0 < n_components < 1 and svd_solver == 'full', select the number of components such that the amount of variance that needs to be explained is greater than the percentage specified by n_components"

For floats you generally shouldn't rely on equivalence anyway so this is really in practice a tiny (hopefully meaningless in practice) distinction here anyway

mne/preprocessing/ica.py
```diff
@@ -138,51 +141,59 @@ class ICA(ContainsMixin):
...
+        are passed to the ICA algorithm during fitting:
+
+        - :class:`int`
+            Must be greater than 1 and less than or equal to
+            ``n_pca_components``.
+        - :class:`float` between 0 and 1 (exclusive)
+            The number of components with cumulative explained variance
+            of the data greater than ``n_components`` will be used. For
+            example, if 0.8 is passed then the number of components will
+            explain greater than 80%% of the variance.
```
hoechenberger, line 153:
```suggestion
        - :class:`float` between 0 and 1 (exclusive)
            Will select the smallest number of components required to explain
            the cumulative variance of the data greater than ``n_components`` .
            Consider this hypothetical example: we have 3 components, the first
            explaining 70%, the second 20%, and the third the remaining 10% of
            the variance. Passing 0.8 here (corresponding to 80% of explained
            variance) would yield the first two components, explaining 90% of
            the variance: only by using the the first two components the
            requested threshold of 80% explained variance can be exceeded. The
            third component, on the other hand, would be excluded.
```

hoechenberger (approved):
Proposal to make the docs a little easier to understand, otherwise looks super great and good to merge!

hoechenberger:
@cbrnr Would be great to get your feedback / approval too before merging. Thanks!

hoechenberger:
Friendly ping @cbrnr

cbrnr:
I'll take a look within the next hour or so!

hoechenberger:
Great, thank you

cbrnr:
Where can I find the updated rendered ICA diagram?

hoechenberger:
> Where can I find the updated rendered ICA diagram?

The doc artifact is at https://22659-1301584-gh.circle-artifacts.com/0/dev/index.html

doc/changes/latest.inc
```diff
@@ -52,7 +52,11 @@ Bugs
 
 - Fix bug that `~mne.viz.plot_ica_overlay` would sometimes not create red traces, by `Richard Höchenberger`_ (:gh:`8341`)
 
+- Fix bug with :class:`~mne.preprocessing.ICA` where ``n_components=None, n_pca_components=None`` could lead to unstable unmixing matrix inversion by making ``n_components=None`` also use the lesser of ``n_components=0.999999`` and ``n_components=n_pca_components`` by `Eric Larson`_ (:gh:`8351`)
+
+- The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_
```
cbrnr, line 57: Add the GH PR?

cbrnr (approved):
LGTM! I've tested this with one of my current data sets, and everything seems to work as before.

I have one comment that we could also discuss in a separate issue, but I wanted to bring it up here. Do we really need `n_pca_components` in `ICA.__init__`? As far as I can see, this parameter is neither used during initialization nor fitting. It is used in `ICA.apply`, which also has an `n_pca_components` parameter, which when set overrides the value passed in the initializer (I think). Wouldn't our API be much cleaner and easier to understand if we dropped `n_pca_components` from `__init__` and just kept it in `apply`? Or am I missing something and we actually need to pass it when constructing the `ICA` object? Again, we don't have to discuss this here, so feel free to merge this PR and discuss this in a separate issue.

hoechenberger:
@larsoner 
> I have one comment that we could also discuss in a separate issue, but I wanted to bring it up here. Do we really need `n_pca_components` in `ICA.__init__`? As far as I can see, this parameter is neither used during initialization nor fitting. It is used in `ICA.apply`, which also has an `n_pca_components` parameter, which when set overrides the value passed in the initializer (I think). Wouldn't our API be much cleaner and easier to understand if we dropped `n_pca_components` from `__init__` and just kept it in `apply`? Or am I missing something and we actually need to pass it when constructing the `ICA` object?

You're absolutely right, it's only used during `apply()`. @larsoner and I had a brief discussion somewhere (too lazy to look it up now) and figured it might be best to just keep in in `__init__()` so you could set it right upon `ICA` instantiation and can forget about it later.

But happy to open this discussion again!

doc/changes/latest.inc
hoechenberger, line 57:
```suggestion
- The ``ica.n_pca_components`` property is no longer be updated during :meth:`mne.preprocessing.ICA.fit`, instead ``ica.n_components_`` will be added to the instance by `Eric Larson`_ (:gh:`8351`)
```

hoechenberger (approved):

hoechenberger:
Thanks a bunch, @larsoner!

cbrnr:
This must have been the PR that closed the most issues at once! Thanks @larsoner!

hoechenberger:
Absolutely. I like how I kept opening issues and @larsoner would come along closing them. How **convenient** 💪☺️

cbrnr:
@larsoner any chance we can backport this? It contains a nice fix in `read_ica` which I originally hadn't noticed. Previously, `method` was set to the default `FastICA` (as in `ICA.__init__`), but now we're actually reading in the method used to compute a particular solution. This is nice, because I have a bunch of files where I used `picard`, and currently I do need to install `scikit-learn` just because we don't read the actual method.

If it's too big of a change for a backport that's OK, it's not that I can't just install `scikit-learn`.

cbrnr:
Although of course just because of this I'm running into the issue that [scikit-learn has not been released for Python 3.9 yet](https://github.com/scikit-learn/scikit-learn/issues/18621), so generating the wheel myself takes ages.

agramfort:
we do not backport API changes.

we should push a new MNE release in early 2021?

>

cbrnr:
Yes, if we do not have to stick to a 6 month release cycle I'd be happy if the next one was a bit sooner.

agramfort:
point taken :)

>

hoechenberger:
Yes, can we cut a new release in December already? These ICA changes are really… important. Also lots of @drammock's viz changes are a real game changer already!

agramfort:
if we have the bandwidth I am all for it !

>

Issue #8038, opened 2020-07-21
Additional support for MFF


ephathaway:
If your issue is a usage question, please consider asking on our [Gitter channel](https://gitter.im/mne-tools/mne-python) or on our [mailing list](https://mail.nmr.mgh.harvard.edu/mailman/listinfo/mne_analysis) instead of opening an issue.

#### Describe the problem
Please provide a clear and concise description of the problem.

On behalf of [Brain Electrophysiology Laboratory](https://belco.tech/), we would like to add support for all different flavors of MFF. The `mne.io.egi` module can be used to read in raw MFF files, but we would like to be able to read MFF segmented files (analogous to `Epochs` class) and MFF averaged files (analogous to `Evoked` class). We would also like to be able to convert `BaseRaw`, `Epochs`, or `Evoked` objects to continuous, segmented, and averaged MFF files respectively and write these as output files. Finally, we would like to be able to read and store a wider range of event track information from MFF files. MFF event track XML files contain several attributes for each event, such as label, duration, and response, that are not currently captured when reading in raw MFF files.

#### Describe your solution

In order to address these requests, we propose using [mffpy](https://github.com/BEL-Public/mffpy), a Python MFF reader/writer created by [BEL](https://belco.tech/) for handling MFF inputs and outputs. This Python package is capable of reading and writing each of the different flavors of MFF as well as parsing individual XML files within an MFF directory. Our general proposal would be to make mffpy a dependency of MNE and then restructure the `mne.io.egi` module to read in MFF files with mffpy and then convert them to the appropriate object (Raw, Epochs, or Evoked). We also want to add a module that takes Raw, Epochs, and Evoked objects and converts them to an mffpy `Writer` object so that they can be output as MFF files. As far as event tracks, we are able to fully parse event track XML files with mffpy. We may need to create a new type of MFF events class to store this information, as it appears that event info is currently stored in a simple array with relative start time and event type.

#### Describe possible alternatives
A clear and concise description of any alternative solutions or features you have considered.

#### Additional context

We have developers at BEL that would be happy to implement these features. We are also happy to answer any questions about mffpy and how it could increase the functionality of MNE.

ephathaway:
@jusjusjus

drammock:
> We may need to create a new type of MFF events class to store this information, as it appears that event info is currently stored in a simple array with relative start time and event type.

[...]

PR #8354, opened 2020-10-09
ENH: read mff evoked

ephathaway:
This PR begins to address issue #8038 by implementing a function to read averaged MFF files as `Evoked` objects. We do this through `mne.io.read_evokeds_mff`, which takes in an averaged MFF file and returns an `EvokedArray` object for each individual average specified. This function utilizes the [mffpy](https://github.com/BEL-Public/mffpy) library to read signals and info data from the MFF.

To get this PR ready to merge, there are a few things I need some direction on:

1. I see that test data are stored in a separate repository. How should I go about adding new test data to this repository? For now I have added my example averaged MFF to a new directory called `/example_mffs`.

2. I can't figure out how to add an entry to the changelog. It looks like this is not done directly in the `/doc/whats_new.rst`.

3. I am unclear on what information is contained in `Info['custom_ref_applied']`. Is this a simple 0 if there is no custom ref or an average ref and 1 for any other custom ref? We should be able to read this info from the info1.xml in the MFF directory.


larsoner:
> This PR begins to address issue #8038 by implementing a function to read averaged MFF files as Evoked objects.

Awesome!

> This function utilizes the mffpy library to read signals and info data from the MFF.

We already have some functions in [mne/io/egi/egimff.py](https://github.com/mne-tools/mne-python/blob/master/mne/io/egi/egimff.py) to do some reading. Did you give those a try? They might already give you most or all of what you need without adding an additional dependency on mffpy.

I'm not sure if our (MNE-Python's) plan is to migrate to mffpy at some point versus keep our own `egimff.py` functions. For example if you see that mffpy does things better than `egimff.py` I don't see a problem in principle with eventually adding an mffpy dependency, though so far we have avoided it for EGI/MFF (mffpy) and EDF (pyedflib) and similar things. So I guess my vote would be to try to use our internal functions assuming they work.

> I see that test data are stored in a separate repository. How should I go about adding new test data to this repository?

Take a look at the README to see if it's clear enough (if it's not we should fix it):

https://github.com/mne-tools/mne-testing-data#add-or-change-files-in-the-repo-for-use-with-mne

And some recent examples:

https://github.com/mne-tools/mne-python/issues/8173
https://github.com/mne-tools/mne-python/pull/8176
https://github.com/mne-tools/mne-testing-data/pull/73

> I can't figure out how to add an entry to the changelog. It looks like this is not done directly in the /doc/whats_new.rst.

That is where it used to live until a few months ago, now it's [doc/changes/latest.inc](https://github.com/mne-tools/mne-python/blob/master/doc/changes/latest.inc)

If there is some reference to `doc/whats_new.rst` for example in the contributing docs, we should update it.

> Is this a simple 0 if there is no custom ref or an average ref and 1 for any other custom ref?

Yes but I'm not sure we've been entirely consistent about this so far. Basically each electrode's location should be stored in `info['chs'][ii]['loc'][:3]` and the location of the reference electrode should be `info['chs'][ii]['loc'][3:6]`. Beyond this, only an average reference is allowed (unless `info['custom_ref_applied'] = True`, and if an average reference has been applied to the data it should have a corresponding projector in `info['projs']` with `active=True`. But these are details we can probably work out later on in the process.

ephathaway:
> We already have some functions in [mne/io/egi/egimff.py](https://github.com/mne-tools/mne-python/blob/master/mne/io/egi/egimff.py) to do some reading. Did you give those a try? They might already give you most or all of what you need without adding an additional dependency on mffpy.

I think ultimately relying on mffpy for the egi io could simplify the code quite a bit. I found it very easy to plug in the mffpy `Reader` class to do most of the work for reading averaged files. Our goal with mffpy is to support reading and writing of all pieces of an MFF (all binary and XML files) and we have implemented this in a way that makes it easy to get to the signal data and info stored in XML files.

In our conversation on this issue #8038, you suggested we submit a PR for reading of MFF averaged files that uses mffpy as a dependency so that you could see how it would be used and evaluate whether we want to use it for MFF io in general. This is why I went that route instead of using existing `egimff.py` functions. I am open to either way of doing, but I think that using mffpy could simplify things quite a bit, especially when we start looking at reading segmented files and event parsing.

environment.yml
```diff
@@ -39,3 +39,5 @@ dependencies:
   - python-picard
   - PyQt5>=5.10,<5.14; platform_system == "Darwin"
   - PyQt5>=5.10; platform_system != "Darwin"
+  - mffpy==0.5.4
```
larsoner, line 42: probably `>=` is better so we get any bug fixes

environment.yml
```diff
@@ -39,3 +39,5 @@ dependencies:
...
+  - cached_property_bel>=1.0.1
```
larsoner, line 43: What is this? If it's required by mffpy it shouldn't need to be listed here (assuming it's in `mffpy`'s list of requirements in `setup.py`)

example_mffs/test_egi_evoked.mff/pnsSet.xml
```diff
@@ -0,0 +1,47 @@
...
+            <gain>1</gain>
+            <defaultDisplayAmplitude>7.5</defaultDisplayAmplitude>
+            <highpassDisplay>10</highpassDisplay>
+            <lowpassDisplay>100</lowpassDisplay>
+            <notchDisplay>60</notchDisplay>
+            <color>0.0000,0.0000,0.0000,1.0000</color>
+            <positiveUp>false</positiveUp>
+        </sensor>
+    </sensors>
+</PNSSet>
```
larsoner, line 47: All of these should go in `mne-testing-data/EGI/test_egi_evoked.mff/`

mne/io/egi/egimff.py
```diff
@@ -9,15 +9,18 @@
 
 import numpy as np
 
+from mffpy import Reader
```
larsoner, line 12: This should be an optional dependency, which means you should nest the import inside the function that uses it (see how we treat other optional dependencies like nibabel, dipy, sklearn, etc.)

mne/io/egi/egimff.py
```diff
@@ -686,3 +693,161 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+        If no categories.xml file in MFF directory specified by `fname`.
+
+    See Also
+    --------
+    Evoked, EvokedArray, create_info
+
+    Notes
+    -----
+    Preloading is automatic because we use `EvokedArray` to construct the
+    evoked(s) objects.
```
larsoner, line 749:
```suggestion
    evoked(s) objects.

    .. versionadded:: 0.23
```

mne/io/egi/egimff.py
```diff
@@ -686,3 +693,161 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    ValueError
+        If no categories.xml file in MFF directory specified by `fname`.
+
+    See Also
+    --------
+    Evoked, EvokedArray, create_info
+
+    Notes
+    -----
+    Preloading is automatic because we use `EvokedArray` to construct the
```
larsoner, line 748: This is true of all Evoked instances (they are always loaded) so no need to note it here I think. Just the `.. versionadded` directive in `Notes` probably

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +292,37 @@ def test_io_egi_crop_no_preload():
     assert_allclose(raw._data, raw_preload._data)
 
 
+@pytest.mark.parametrize('idx, cond, signals, bads', [
```
larsoner, line 295: Prevent this from bombing on CIs that have minimal dependencies
```suggestion
@requires_version('mffpy')
@pytest.mark.parametrize('idx, cond, signals, bads', [
```

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +292,37 @@ def test_io_egi_crop_no_preload():
...
+    assert_equal(evoked_cond.comment, cond)
+    assert_equal(evoked_idx.comment, cond)
+    # Check signal data
+    data = np.loadtxt(signals, ndmin=2).transpose() * 1e-6  # convert to volts
+    assert_allclose(evoked_cond.data, data, atol=1e-6)
+    assert_allclose(evoked_idx.data, data, atol=1e-6)
+    # Check info
+    assert_equal(evoked_cond.info, evoked_idx.info)
+    assert_equal(evoked_cond.info['description'], cond)
+    assert_equal(evoked_cond.info['bads'], bads)
```
larsoner, line 318: for builtin types we now prefer to let pytest do nice inspection for us, and this is a list, so
```suggestion
    assert evoked_cond.info['bads'] == bads
```

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +292,37 @@ def test_io_egi_crop_no_preload():
...
+    evoked_cond = read_evokeds_mff(egi_mff_evoked_fname, condition=cond)
+    evoked_idx = read_evokeds_mff(egi_mff_evoked_fname, condition=idx)
+    assert_equal(evoked_cond.comment, cond)
+    assert_equal(evoked_idx.comment, cond)
+    # Check signal data
+    data = np.loadtxt(signals, ndmin=2).transpose() * 1e-6  # convert to volts
+    assert_allclose(evoked_cond.data, data, atol=1e-6)
+    assert_allclose(evoked_idx.data, data, atol=1e-6)
+    # Check info
+    assert_equal(evoked_cond.info, evoked_idx.info)
```
larsoner, line 316: I'm not sure this will check what you want it to. I think we usually use `assert object_diff(a, b) == ''` for this sort of thing

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +292,37 @@ def test_io_egi_crop_no_preload():
...
+    assert_equal(evoked_cond.info['description'], cond)
+    assert_equal(evoked_cond.info['bads'], bads)
+    assert_equal(len(evoked_cond.info['ch_names']), 259)
+    assert 'ECG' in evoked_cond.info['ch_names']
+    assert 'EMG' in evoked_cond.info['ch_names']
+    pick_eeg = pick_types(evoked_cond.info, eeg=True, exclude=[])
+    assert_equal(len(pick_eeg), 257)
+    assert_equal(evoked_cond.info['nchan'], 259)
```
larsoner, line 324:
```suggestion
    assert evoked_cond.info['nchan'] == 259
```
etc.

larsoner (commented):
> In our conversation on this issue #8038, you suggested we submit a PR for reading of MFF averaged files that uses mffpy

Ahh sorry, I forgot about that plan!

I left a few comments to get the ball rolling. I think the big thing is to update the testing data. Let me know if the instructions don't make sense and I can do it

environment.yml
ephathaway: That's right. It's a dependency of `mffpy`. Version 0.5.5 of `mffpy` will install this dependency automatically. This will hopefully be packaged tomorrow.

larsoner: Okay we don't release (new features) again until March. So we should be able to delete this line and add `mffpy>=0.5.5` and then merge this PR once it's ready and 0.5.5 is out

ephathaway:
> If there is some reference to `doc/whats_new.rst` for example in the contributing docs, we should update it.

Yeah the contribution guidelines reference `doc/whats_new.rst` in a couple sections:

https://mne.tools/dev/install/contributing.html#general-requirements
https://mne.tools/dev/install/contributing.html#github-workflow

doc/changes/latest.inc
```diff
@@ -16,6 +16,7 @@ Current (0.22.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :func:`mne.io.read_evokeds_mff` to enable reading of averaged MFFs by `Evan Hathaway`_ (:gh:`8354`)
```
agramfort, line 19: please add that it will require mffpy library

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,40 @@ def test_io_egi_crop_no_preload():
     assert_allclose(raw._data, raw_preload._data)
 
 
+@requires_version('mffpy', '0.5')
```
ephathaway, line 294: Technically we need `mffpy v0.5.5` for this to run, but `mne.utils.requires_version` can only check against the "loose version" (e.g. 0.5). Should we edit this function to be able to check against the full v0.5.5?

ephathaway:
@larsoner @agramfort I believe we have addressed all of your requested changes. `mffpy v0.5.5` is now packaged and appears to be working like it should. There will a couple of CIs failing before my last commit, so we shall see if those still fail with the last fix.

mne/io/egi/tests/test_egi.py
larsoner: I don't think this is a problem with `requires_version`, which just uses `check_version`, which just uses `LooseVersion` from `distutils.version`. I think it's a problem with the `mffpy` 0.5.5 reporting its version as 0.5.0:
```
$ pip install mffpy
Defaulting to user installation because normal site-packages is not writeable
Collecting mffpy
  Downloading mffpy-0.5.5-py3-none-any.whl (140 kB)
     |████████████████████████████████| 140 kB 3.1 MB/s 
...
larsoner@bunk:~$ python
>>> import mne
>>> mne.utils.check_version('mffpy', '0.5')
True
>>> mne.utils.check_version('mffpy', '0.5.5')
False
>>> mne.utils.check_version('mffpy', '0.5.4')
False
>>> import mffpy
>>> mffpy.__version__
'0.5.0'
```

larsoner: And just to show how it works if the version of a lib is correct:
```
>>> import pytest
>>> pytest.__version__
'6.1.1'
>>> mne.utils.check_version('pytest', '6.1.1')
True
>>> mne.utils.check_version('pytest', '6.1.2')
False
```

doc/changes/latest.inc
```diff
@@ -16,6 +16,7 @@ Current (0.22.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :func:`mne.read_evokeds_mff` to enable reading of averaged MFFs (requires mffpy library >= 0.5.5) by `Evan Hathaway`_ (:gh:`8354`)
```
larsoner, line 19: It looks like this is your first contribution (hooray)! Can you see how we've formatted `Eduard Ort`'s contribution? It creates a bold entry. You'll also need to add your name to `doc/changes/names.inc`, it's part of what's causing CircleCI to be unhappy.

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,187 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    Returns
+    -------
+    evoked : EvokedArray or list of EvokedArray
+        The evoked dataset(s); one EvokedArray if condition is int or str,
+        or list of EvokedArray if condition is None or list.
+
+    Raises
+    ------
+    ValueError
+        If `fname` has file extension other than '.mff'.
```
larsoner, line 735: This tries to link but it can't (NumpyDoc/autodoc do not make the params a target), leading to Sphinx emitting these warnings that get treated as errors:
```
/home/circleci/project/mne/io/egi/egimff.py:docstring of mne.read_evokeds_mff:11: WARNING: py:obj reference target not found: condition
/home/circleci/project/mne/io/egi/egimff.py:docstring of mne.read_evokeds_mff:48: WARNING: py:obj reference target not found: fname
/home/circleci/project/mne/io/egi/egimff.py:docstring of mne.read_evokeds_mff:51: WARNING: py:obj reference target not found: fname
/home/circleci/project/mne/io/egi/egimff.py:docstring of mne.read_evokeds_mff:54: WARNING: py:obj reference target not found: fname
```
I would make these double-backticks so that they get rendered as code

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,187 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+
+    Parameters
+    ----------
+    fname : str
+        File path to averaged MFF file. Should end in .mff.
+    condition : int or str | list of int or str | None
+        The index (indices) or category (categories) from which to read in
+        data. Averaged MFF files can contain separate averages for different
+        categories. These can be indexed by the block number or the category
+        name. If `condition` is a list or None, a list of Evoked objects is
```
larsoner, line 710: Same here for `condition`

mne/io/egi/tests/test_egi.py
```diff
@@ -29,6 +29,9 @@
 egi_pause_fname = op.join(egi_path, 'test_egi_multiepoch_paused.mff')
 egi_eprime_pause_fname = op.join(egi_path, 'test_egi_multiepoch_eprime.mff')
 egi_pause_w1337_fname = op.join(egi_path, 'w1337_20191014_105416.mff')
+egi_mff_evoked_fname = op.join(egi_path, 'test_egi_evoked.mff')
+egi_txt_evoked_cat1_fname = op.join(base_dir, 'test_egi_evoked_cat1.txt')
+egi_txt_evoked_cat2_fname = op.join(base_dir, 'test_egi_evoked_cat2.txt')
```
larsoner, line 34: These files should probably live in `mne-testing-data` as well. No reason to keep them in this repo when the relevant file that's used with them is in `mne-testing-data`

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,40 @@ def test_io_egi_crop_no_preload():
...
+    # Check signal data
+    data = np.loadtxt(signals, ndmin=2).transpose() * 1e-6  # convert to volts
+    assert_allclose(evoked_cond.data, data, atol=1e-6)
+    assert_allclose(evoked_idx.data, data, atol=1e-6)
+    # Check info
+    assert object_diff(evoked_cond.info, evoked_idx.info) == ''
+    assert evoked_cond.info['description'] == cond
+    assert evoked_cond.info['bads'] == bads
+    assert len(evoked_cond.info['ch_names']) == 259
+    assert 'ECG' in evoked_cond.info['ch_names']
```
larsoner, line 321: You can check for channel type existence using `in inst`, e.g.:
```suggestion
    assert 'ECG' in evoked_cond.info['ch_names']
    assert 'ecg' in evoked_cond
```
This assumes that the channel type is set properly to ECG_CH, so if it's not, we can add this later if we correctly detect the channel as type ECG at some point

mne/io/egi/tests/test_egi.py
ephathaway: Got it. I put them in `mne/io/egi/tests/data` originally because that is where the `test_egi.txt` signals live, but I can certainly move then to `mne-testing-data`.

mne/io/egi/tests/test_egi.py
ephathaway: Ah good call. I'll look into why `mffpy` is reporting the wrong version.

mne/io/egi/tests/test_egi.py
larsoner: That one is paired with the `test_egi.raw` file that also lives in this repo. We've abandoned the practice of adding files to the repo for testing unless they are tiny and standalone -- this one is coupled to the `testing` dataset so no benefit to having it here

mne/io/egi/tests/test_egi.py
ephathaway: Nice. Just to be clear, do you suggest having both assert statements?
```
assert 'ECG' in evoked_cond.info['ch_names']
assert 'ecg' in evoked_cond
```
or just the one?
```
assert 'ecg' in evoked_cond
```

larsoner: Both, they check different aspects of the info (names and channel types)

mne/io/egi/tests/test_egi.py
ephathaway: I figured out the issue on the `mffpy` side. Just need to merge that PR with the changes and then `mffpy` should be displaying the right version.

larsoner: And probably release a 0.5.6?

ephathaway: Yes, that's the plan.

ephathaway:
@larsoner I have been working on fixing these failing checks the last few days. I fixed a conflict with `cached_property_bel` and another dependency and I think I have finally got the changelog formatted correctly. There are still a few that I am not sure how to address. It seems that the Windows pipelines using pip are failing because of a missing pip installation and a code coverage command that is not found. And then the Travis CI is failing because of some kind of connection error when downloading the testing data.

drammock:
> It seems that the Windows pipelines using pip are failing because of a missing pip installation

a rebase should fix this one at least; see #8483 and #8490

ephathaway:
> a rebase should fix this one at least; see #8483 and #8490

Nice. Looks like I caught both of these in my last rebase. No checks failing so far this run! Crossing my fingers.

larsoner:
Travis failures are unrelated and fixed by #8491

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    ValueError
+        If no categories.xml file in MFF directory specified by ``fname``.
+
+    See Also
+    --------
+    Evoked, EvokedArray, create_info
+
+    Notes
+    -----
+    .. versionadded:: 0.23
```
larsoner, line 747: Let's not be so pessimistic about our timeline :)
```suggestion
    .. versionadded:: 0.22
```
(It's possible I suggested the wrong version previously, in which case sorry!)

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+        raise ValueError('fname must be an MFF file with extension ".mff".')
+    # Confirm the input MFF is averaged
+    mff = mffpy.Reader(fname)
+    if mff.flavor != 'averaged':
+        raise ValueError('%s is a %s MFF file. fname must be the path to an \
+                         averaged MFF file.' % (fname, mff.flavor))
+    # Check for categories.xml file
+    if 'categories.xml' not in mff.directory.listdir():
+        raise ValueError('categories.xml not found in MFF directory. \
+                         %s may not be an averaged MFF file.' % fname)
```
larsoner, line 761: Nowadays we use f-strings and I don't think we use continuation lines in strings like this so this would be more standard for us:
```suggestion
        raise ValueError('categories.xml not found in MFF directory. '
                         f'{fname} may not be an averaged MFF file.')
```
But it's okay as is

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    if condition is None:
+        categories = mff.categories.categories
+        condition = list(categories.keys())
+    elif not isinstance(condition, list):
+        condition = [condition]
```
larsoner, line 766: If someone passes `condition='foo'` they should get a single instance but `condition = ['foo']` they should get a  list, so
```suggestion
    return_list = True
    if condition is None:
        categories = mff.categories.categories
        condition = list(categories.keys())
    elif not isinstance(condition, list):
        condition = [condition]
        return_list = False
```

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    logger.info('Reading %d evoked datasets from %s ...'
+                % (len(condition), fname))
+    output = [_read_evoked_mff(fname, c, channel_naming=channel_naming,
+                               verbose=verbose).apply_baseline(baseline)
+              for c in condition]
+    return output if len(output) > 1 else output[0]
```
larsoner, line 772:
```suggestion
    return output if return_list else output[0]
```

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+        epoch = mff.epochs[category]
+    elif isinstance(condition, int):
+        # Condition is interpreted as epoch index
+        epoch = mff.epochs[condition]
+        category = epoch.name
+    else:
+        raise TypeError('Condition must be either int or str.')
+    assert category in categories.keys(), 'Condition "%s" not found in \
+                                           available conditions %s.'\
+                                           % (category, categories.keys())
```
larsoner, line 794: `assert` statements are skipped in optimized mode and should only be used for internal checks that should always be true (I think of them as "active comments" to remind about shapes, sizes, types, etc. that should be guaranteed)
```suggestion
    _check_option('condition', category, sorted(categories)
```
And shouldn't this go *above* the `if/else` immediately preceding it so that invalid category names are caught before `mff.epochs[category]` is attempted so that we can raise an informative error?

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+
+    # Read in signals from the target epoch
+    data = mff.get_physical_samples_from_epoch(epoch)
+    eeg_data, t0 = data['EEG']
+    if 'PNSData' in data.keys():
```
larsoner, line 799:
```suggestion
    if 'PNSData' in data:
```

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+
+    # Load metadata into info object
+    # Exclude info['meas_date'] because record time info in
+    # averaged MFF is the time of the averaging, not true record time.
+    ch_names = [channel_naming % (i + 1) for i in
+                range(mff.num_channels['EEG'])]
+    ch_names.extend(egi_info['pns_names'])
+    info = create_info(ch_names, mff.sampling_rates['EEG'], ch_types)
+    info['nchan'] = sum(mff.num_channels.values())
+    assert info['nchan'] == all_data.shape[0], 'Number of channels does \
```
larsoner, line 816: Same statement about `assert` as above

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,199 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    with mff.directory.filepointer('history') as fp:
+        history = mffpy.XML.from_file(fp)
+    for entry in history.entries:
+        if entry['method'] == 'Montage Operations Tool':
+            if 'Average Reference' in entry['settings']:
+                # Average reference has been applied
+                projector, info = setup_proj(info)
+            else:
+                # Custom reference has been applied that is not an average
+                info['custom_ref_applied'] = True
```
larsoner, line 867: A single-electrode / monopolar reference is the standard type. If this is how the data were recorded then there should be no avg ref proj and also `info['custom_ref_applied'] = False`. The `custom_ref_applied` is used for other cases, such as when the average of two electrodes is used as a reference for all channels.

Not sure if this is relevant for the "Montage Operations Tool", maybe being in this section means that it's a custom, non-average, *and non-monopolar* reference, in which case what's here makes sense.

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,63 @@ def test_io_egi_crop_no_preload():
...
+def test_io_egi_evokeds_mff(idx, cond, tmax, signals, bads):
+    """Test reading evoked MFF file."""
+    # Test reading all conditions from evokeds
+    evokeds = read_evokeds_mff(egi_mff_evoked_fname)
+    assert len(evokeds) == 2
+    # Test reading list of conditions from evokeds
+    evokeds = read_evokeds_mff(egi_mff_evoked_fname, condition=[0, 1])
+    assert len(evokeds) == 2
+    # Test invalid condition type
+    with pytest.raises(TypeError):
```
larsoner, line 311: Help both make clearer what's being tested, and more specific/targeted
```suggestion
    with pytest.raises(TypeError, match='some matching substring'):
```

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,63 @@ def test_io_egi_crop_no_preload():
...
+    # Test reading evoked data from single condition
+    evoked_cond = read_evokeds_mff(egi_mff_evoked_fname, condition=cond)
+    evoked_idx = read_evokeds_mff(egi_mff_evoked_fname, condition=idx)
+    for evoked in [evoked_cond, evoked_idx]:
+        assert evoked.comment == cond
+        assert evoked.nave == 3
+        assert evoked.tmin == 0.0
+        assert evoked.tmax == tmax
+    # Check signal data
+    data = np.loadtxt(signals, ndmin=2).transpose() * 1e-6  # convert to volts
```
larsoner, line 322:
```suggestion
    data = np.loadtxt(signals, ndmin=2).T * 1e-6  # convert to volts
```

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,63 @@ def test_io_egi_crop_no_preload():
...
+    assert_allclose(evoked_cond.data, data, atol=1e-6)
```
larsoner, line 323: For EEG data in volt units an `atol=1e-6` is a pretty bad tolerance level. Can this be `atol=1e-10` or `atol=1e-12` or something? It's going to depend a bit on the number of decimal places in the text file, but hopefully adjusting `atol` and/or `rtol=1e-4` (for example if the field width is 5 characters in the text file this would be a sensible choice) or something similar will help make it clearer the match is actually pretty good.

mne/io/egi/tests/test_egi.py
```diff
@@ -288,4 +291,63 @@ def test_io_egi_crop_no_preload():
...
+    assert evoked_cond.info['sfreq'] == 250.0
+    assert not evoked_cond.info['custom_ref_applied']
+
+
+@requires_version('mffpy', '0.5.7')
+@requires_testing_data
+def test_read_evokeds_mff_bad_input():
+    """Test errors are thrown when reading an invalid averaged MFF."""
+    # Test file that is not an MFF
+    with pytest.raises(ValueError):
```
larsoner, line 346: match

larsoner (commented):
Looks like it's almost there! Just a few minor remaining suggestions

mne/io/egi/egimff.py
ephathaway: Typically when applying a montage operation tool, the result is something other than a monopolar reference (there is no setting to reference all electrodes to a single electrode; to do this you would have to individually map every electrode to the single electrode reference). However, it technically is possible to rereference to a single electrode, and I am not sure if there is a good way to determine this from the `history.xml`. If we can't know this info for sure, should we just always make `info['custom_ref_applied`] = False`?

larsoner: No, in that case I think the safest assumption is to set `info['custom_ref_applied'] = True` as you do here. So I think we're okay here

mne/io/egi/egimff.py
ephathaway: Ah yeah. Good catch!

mne/io/egi/egimff.py
ephathaway: Nice. I prefer f-strings, so I will gladly use them :)

mne/io/egi/egimff.py
ephathaway: Good idea to check the `condition` param first. Since `condition` can be either a `str` with the category name or an `int` with the epoch index, I will use a different check for each of these cases.

mne/io/egi/egimff.py
ephathaway: I don't think this `assert` is doing a lot anyway since we ultimately get the value for both sides of the `==` from the binary signals file, so I vote we just take it out.

mne/io/egi/egimff.py
ephathaway: We do return a single instance if there is only one condition in the list.
```
return output if len(output) > 1 else output[0]
```

larsoner: I'm saying that behavior is wrong

larsoner: (it should only so this is they pass a str, if they pass a list with one str they should still get a list with one Evoked out)

ephathaway: Ah got it. I'll make that change.

mne/io/egi/egimff.py
```diff
@@ -686,3 +692,200 @@ def _read_segment_file(self, data, idx, fi, start, stop, cals, mult):
...
+    # Get nave from categories.xml
+    try:
+        nave = categories[category][0]['keys']['#seg']['data']
+    except KeyError:
+        warn(f'Number of averaged epochs not found for condition {category}. '
+             'nave will default to 1.', category=UserWarning)
+        nave = 1
+
+    # Let tmin default to 0
+    return EvokedArray(all_data, info, tmin=0., comment=category,
```
larsoner, line 879: I wonder if we should add a new `EvokedMFF` class to make it clear that this was read from an MFF file. It seems cleaner than reusing `EvokedArray`:
```
class EvokedMFF(Evoked):

def __init__(self, info, data, ...)
    self.data = data
    self.info = info
    ...
```
But maybe this is more work than it's worth.

larsoner (approved):
LGTM +1 for merge other than one possible additional idea/change. Let me know what you think

mne/io/egi/egimff.py
ephathaway: I think since this `read_evokeds_mff` function resides within this special `egi` library, there won't be much confusion about what file type we are reading from. My vote is that adding a new class would be more trouble than it's worth.

larsoner:
Sorry for the slow response here @ephathaway, I should have merged earlier. Looks like there are conflicts -- if you give "maintainers permission to edit" (there is some way to do this, at least when opening a PR) then I can take care of it. Otherwise, can you rebase or `merge` with upstream/master then we can merge?

ephathaway:
> Otherwise, can you rebase or `merge` with upstream/master then we can merge?

I rebased on top of `master` and fixed the conflicting files.

I ran the whole test suite for good measure and found that there are a number of tests in `mne/test/test_morph.py` that will only pass if `dipy>=1.3.0`. This should probably be addressed outside of this PR, but could we require `dipy>=1.3.0`, at least for those particular tests? I can draft up an issue if you like.

larsoner:
> This should probably be addressed outside of this PR

Agreed -- GitHub makes transferring comments to issues pretty easy so I did it: https://github.com/mne-tools/mne-python/issues/8551

larsoner:
Okay the `conda` failure is real:
```
ResolvePackageNotFound: 
  - mffpy[version='>=0.5.7']
```
This is because we recently transitioned to doing everything on conda-forge. The quick fix is to add to the end of `environment.yml`:
```
- pip:
  - mffpy>=0.5.7
```
Want to do that here? A better long-term solution is to make `mffpy` a `conda-forge` package. @hoechenberger would you be able to help with this? I am not too experienced with `conda-forge` setups.

hoechenberger:
@larsoner yes I can create a conda-forge package tomorrow morning!

ephathaway:
> @larsoner yes I can create a conda-forge package tomorrow morning! UTC+1

Awesome! I was just researching how to do this, but that's great if you can take care of it. Is this something we would need to update with each new release of `mffpy`?

hoechenberger:
@ephathaway I've opened a PR: https://github.com/conda-forge/staged-recipes/pull/13291

After this has been merged, it will take about an hour or so until the package shows up on all conde-forge mirrors.

Regarding keeping the package updated:
A bot will realise when a new version has been pushed to PyPI, update the conda-forge recipe, and open a PR. One can even set it up such that this PR gets merged automatically as long as CI passes.

I can add you as a co-maintainer for this recipe if you're interested!

hoechenberger:
Had to patch `setup.py`, because the current mffpy sdist is defunct. Opened a PR upstream.

hoechenberger:
It's on Conda-forge now:
https://anaconda.org/conda-forge/mffpy

larsoner:
@ephathaway look like one more rebase/merge should do it (the `testing` dataset was updated in `master`, so you can just discard your edits there). I would do the merge myself but it looks like edits-by-maintainers is disabled. Sorry for the hassle, hopefully it's the last one!

ephathaway:
> I can add you as a co-maintainer for this recipe if you're interested!

That would be great! Please do.

ephathaway:
@larsoner great! I rebased onto master. Tests are passing on my end.

agramfort (approved):
Readme.rst just needs to be updated with new optional dependency

agramfort:
thanks @ephathaway

larsoner:
Indeed, thanks for working through everything @ephathaway !

ephathaway:
@agramfort @larsoner for sure! Thanks for all the support.

PR #8361, opened 2020-10-13
WIP: Fix bugs in Brain

GuillaumeFavelier:
This PR disables the rendering call each time a mesh is added to the scene in favour of only one call when everything is setup.

Quoting https://github.com/pyvista/pyvista/pull/935#issue-500086898:

> There are 2 main reasons for this:
> - Huge speed gain when hiding and displaying meshes.
> - It looks much more professionnal when clicking a button to see all the changes at once.

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -22,6 +22,7 @@ import vtk
 
 from .base_renderer import _BaseRenderer
 from ._utils import _get_colormap_from_array, ALLOWED_QUIVER_MODES
+from ...fixes import _get_args
 from ...utils import copy_base_doc_to_subclass_doc, _check_option
 from ...externals.decorator import decorator
 
@@ -658,6 +659,10 @@ def _add_mesh(plotter, *args, **kwargs):
         smooth_shading = kwargs.pop('smooth_shading')
     else:
         smooth_shading = True
+    # disable rendering pass for add_mesh, render()
+    # is called in show()
+    if 'render' in _get_args(plotter.add_mesh):
+        kwargs['render'] = False
     actor = plotter.add_mesh(*args, **kwargs)
     if smooth_shading and 'Normals' in mesh.point_arrays:
         prop = actor.GetProperty()
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -661,7 +661,7 @@ def _add_mesh(plotter, *args, **kwargs):
         smooth_shading = True
     # disable rendering pass for add_mesh, render()
     # is called in show()
-    if 'render' in _get_args(plotter.add_mesh):
+    if 'render' not in kwargs and 'render' in _get_args(plotter.add_mesh):
         kwargs['render'] = False
     actor = plotter.add_mesh(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


mne/viz/backends/_pyvista.py
```diff
@@ -658,6 +659,10 @@ def _add_mesh(plotter, *args, **kwargs):
         smooth_shading = kwargs.pop('smooth_shading')
     else:
         smooth_shading = True
+    # disable rendering pass for add_mesh, render()
+    # is called in show()
+    if 'render' in _get_args(plotter.add_mesh):
```
larsoner, line 664: For future compat in case at some point we want to pass `render=True` for some reason
```suggestion
    if 'render' not in kwargs and 'render' in _get_args(plotter.add_mesh):
```

larsoner:
@GuillaumeFavelier locally on master with a script similar to:

<details>

```
import os.path as op
import mne
    # sample_dir, 'sample_audvis-meg-oct-6-meg-inv.fif'))
    evoked, inv, return_residual=True, pick_ori='vector')

initial_time = 0.1
# stc, kwargs = stc, dict(overlay_alpha=0.5, brain_alpha=0.5, vector_alpha=0.)
stc, kwargs = stc.magnitude(), dict(alpha=0.5, surface='white')
kwargs['src'] = inv['src']
meth = stc.plot_3d if isinstance(stc, mne.VolSourceEstimate) else plot
brain = meth(
    clim='auto', hemi='both', views=['sag', 'axi', 'cor'],
    view_layout='horizontal', smoothing_steps='nearest', verbose=True,
    time_viewer=False, size=(800, 300), background='k',
    **kwargs)
```

</details>

I see this "flashing" (sorry the window disappears -- I'm talking about the meshes and outlines appearing and disappearing):

Peek 2020-10-13 08-16


And on this branch and latest PyVista `master` it doesn't seem to be improved much. Do you see the same thing? Can you see if it can be improved?

Actually with this example script it's a bit worse -- the display doesn't update until I click on the window (ignore the "blue brain" in the background -- the window produced by the script above will appear on top of it, flash for a bit, then not display the final image until I click on the window to bring it to focus):

Peek 2020-10-13 08-26

GuillaumeFavelier:
Thanks for reporting, I can reproduce. I will investigate in this PR (already opened).

GuillaumeFavelier:
For me, 0372cd2 improved the situation greatly. In `plot_source_estimates()`, `show()` is called last, when everything else is set basically. Could you try it now @larsoner ?

larsoner (approved):
Awesome, works with both `time_viewer=True` and `False`!

larsoner:
@GuillaumeFavelier I'm +1 for backporting, any reason not to?

GuillaumeFavelier:
I see no problem here :+1:

larsoner:
Thanks @GuillaumeFavelier

larsoner:
Nope, not backporting -- the conflict due to the TimeViewer / Brain refactor was ugly

GuillaumeFavelier:
Ouch, my bad. I thought it was straightforward to cherry-pick this.

larsoner:
Actually I think I figured it out: 35662d07f8cb31c104ee36e94b9484ff05424b21

PR #8362, opened 2020-10-13
remove 1. instance of checking metadata [Fixes #8360]

### LINKED ISSUE #8360: ValueError: metadata must have the same number of rows as events
I am  trying to create epochs with metadata from raw, but run into an error with a piece of code that worked fine about 2 months ago (some previous version, likely 0.20.dev).
I updated to the latest master. 

I want to use a subset of events of interest, while my raw data contains more events.
Creating epochs without metadata works fine.

(I tried to comment out the size check in _check_metadata, but it breaks later than.)

>   File "/home/sh254795/anaconda3/lib/python3.7/site-packages/mne/utils/mixin.py", line 358, in _check_metadata
>     % (len(metadata), len(self.events)))
> 
> ValueError: metadata must have the same number of rows (38) as events (86)

Have there been any changes to this code that could explain the error?

Example: 

```
import os
import numpy as np
import mne
import pandas as pd

                                    'sample_audvis_raw.fif')
raw.crop(tmax=60).load_data()

# find all events

# define events of interest 
event_id = dict(A=4, B=3)

tmin = -0.1
tmax = 0.5

#%% create metadata for the events of interest

# count events of interest 
n_events = sum(events[:,2] == 3) + sum(events[:,2] == 4)

# create arbitrary metadata
METADATA = np.arange(0,n_events)
 
META_DF = pd.DataFrame(data=METADATA, columns = ['trial'])
   
# Epoch the data without metadata - FINE
print('creating epochs...')
epochs = mne.Epochs(raw, events, event_id, tmin, tmax)
      

# Epoch the data with metadata - BREAKS
print('creating epochs...')
epochs = mne.Epochs(raw, events, event_id, tmin, tmax,
                    metadata=META_DF)
```

SophieHerbst:
Fixes #8360

Might be too good to be true, but so far it works for me:
I just removed the first instance of 
self.metadata = metadata in epochs.py
It occurs again later, after the events of interest were selected.

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -411,7 +411,6 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
             # ensure metadata matches original events size
             self.selection = np.arange(len(events))
             self.events = events
-            self.metadata = metadata
             del events
 
             values = list(self.event_id.values())


agramfort:
I don’t think it is the right fix

cbrnr:
No, see discussion in #8360.

PR #8358, opened 2020-10-12
RFC: Create SECURITY.md

drammock:
this is related to our [community profile](https://github.com/mne-tools/mne-python/community)

- [x] direct security reports to group dev email address
- [x] the release process wiki should be edited to mention that  this file would need to get updated every time a release is made
- [ ] decide whether SECURITY.md is ignored or added to the manifest

PR #8365, opened 2020-10-13
MRG, MAINT: update issue templates

drammock:
This updates our issue templates, based on some new capabilities (see `config.yml`) and based on looking around at what MPL and sklearn are doing with their templates these days.  Once #8358 is merged, I think that we'll also automatically get a "report a security vulnerabililty" entry like MPL has (see here: https://github.com/matplotlib/matplotlib/issues/new/choose).

The one question I have is whether folks agree with `blank_issues_enabled: false` in the config...  the only thing I could think of is for `RFC` / discussion types of issues, which are typically only opened by devs, who presumably know how to abuse one of the templates for that purpose? (i.e., removing the labels)


larsoner:
> which are typically only opened by devs, who presumably know how to abuse one of the templates for that purpose? (i.e., removing the labels)

I will get slightly annoyed by having to ctrl-A-then-del instead of just clicking the blank button (as mentioned [last time](https://github.com/mne-tools/mne-python/pull/5413#issuecomment-410774981) we talked about this), but I can live with it if we think it will save us some problems. I can't remember if we've had a lot of "bad reports" that would make us want to avoid having this option. Or maybe it's just preemptive?

drammock:
> I can't remember if we've had a lot of "bad reports" that would make us want to avoid having this option. Or maybe it's just preemptive?

Subjectively it seemed to me that there were still a lot, but maybe it's confirmation bias. Regardless, I thought that in addition to the prominent text in the bug reports template, having direct links to gitter and the mailing list on the issue triage page might also help cut down on the usage question traffic.

The trickiness here is that when some users run code and encounter any kind of `Error` message, they think "bug" (hence the warning text in the bug report template).  Other users may think "I don't know if this is a bug in MNE-Python code or an error in my code, so I'll go ask" and they might choose "bug report" or something else...  if they choose something else (including "blank issue") then they won't ever see the warning text in the bug report template.

> I will get slightly annoyed by having to ctrl-A-then-del instead of just clicking the blank button

I can restore the blank template. But if we do that, would it make sense to do what MPL does, and add [an issue template explicitly labeld for usage questions](https://github.com/matplotlib/matplotlib/issues/new?labels=Community+support&template=questions.md) that basically says "don't open an issue"?

larsoner:
> I can restore the blank template. But if we do that, would it make sense to do what MPL does, and add an issue template explicitly labeld for usage questions that basically says "don't open an issue"?

This might be useful regardless of whether or not we add a blank one.

I say we get rid of the blank one (let's see how annoyed I get :) ) and also add the usage one that points to the correct resource(s) for those issues.

drammock:
CIs all green; this one is ready for review/merge.

larsoner:
Thanks @drammock . Should we backport?

drammock:
shouldn't be necessary to backport, it doesn't affect the docs website at all, only the GitHub repo (which users will see the same regardless of which MNE version they're using)

drammock:
...unless there's a point-release to stable, in which case maybe the manifest / `SECURITY.md` will cause a problem?  I'm not really familiar with the intricacies of packaging.

PR #8366, opened 2020-10-13
MRG, ENH: Linearize volume morph

larsoner:
Todo:

- [x] Add round-trip save-load for linearize case to ensure it's saved

### DIFF
--- a/mne/morph.py
+++ b/mne/morph.py
@@ -337,7 +337,8 @@ class SourceMorph(object):
     src_data : dict
         Additional source data necessary to perform morphing.
     vol_morph_mat : scipy.sparse.csr_matrix | None
-        The volumetric morph matrix, if linearize=True was used.
+        The volumetric morph matrix, if :meth:`compute_vol_morph_mat`
+        was used.
     %(verbose)s
 
     Notes
@@ -456,8 +457,8 @@ class SourceMorph(object):
         return out
 
     @verbose
-    def linearize_volume_morph(self, *, verbose=None):
-        """Linearize the volumetric morphing step.
+    def compute_vol_morph_mat(self, *, verbose=None):
+        """Compute the sparse matrix representation of the volumetric morph.
 
         Parameters
         ----------
@@ -467,15 +468,16 @@ class SourceMorph(object):
         -----
         For a volumetric morph, this will compute the morph for an identity
         source volume, i.e., with one source vertex active at a time, and store
-        the result as a sparse morphing matrix. This takes a long time
-        (minutes) to compute initially, but can be drastically faster to apply
-        to for STCs with many time points, as it converts the volume morph
-        step into a sparse matrix dot product.
+        the result as a :class:`sparse <scipy.sparse.csr_matrix>`
+        morphing matrix. This takes a long time (minutes) to compute initially,
+        but drastically speeds up :meth:`apply` for STCs, so it can be
+        beneficial when many time points or many morphs (i.e., greater than
+        the number of volumetric ``src_from`` vertices) will be performed.
 
-        When calling :meth:`save`, this linearization is saved with the
-        instance, so this only needs to be called once. This function does
-        nothing if the morph is already linearized, or if there is no
-        volume morph present.
+        When calling :meth:`save`, this sparse morphing matrix is saved with
+        the instance, so this only needs to be called once. This function does
+        nothing if the morph matrix has already been computed, or if there is
+        no volume morphing necessary.
 
         .. versionadded:: 0.23
         """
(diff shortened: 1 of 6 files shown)


larsoner:
I wasn't sure if the transformation would actually be linear but it appears to be in testing (hooray!). Added I/O round-trip tests. This one is ready to go from my end.

agramfort:
any suggestion where to document this in the narrative doc?

larsoner:
I'll add a `.. note::` to a tutorial. We don't want to actually use it because we don't do enough morphing to justify the minutes it takes to compute

larsoner:
Last commit is just to quote a more accurate speedup number (~400x instead of ~100x -- it was 438x in my local test), but this is what they [otherwise say](https://22711-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_morph_volume_stc.html#apply-morph-to-volsourceestimate):

> Volumetric morphing is much slower than surface morphing because the volume for each time point is individually resampled and SDR morphed. The mne.SourceMorph.compute_vol_morph_mat() method can be used to compute an equivalent sparse matrix representation by computing the transformation for each source point individually. This generally takes a few minutes to compute, but can be saved to disk and be reused. The resulting sparse matrix operation is very fast (about 100x faster) to apply. This approach is more efficient when the number of time points to be morphed exceeds the number of source space points, which is generally in the thousands. This can easily occur when morphing many time points and multiple conditions.

mne/morph.py
```diff
@@ -446,6 +456,55 @@ def apply(self, stc_from, output='stc', mri_resolution=False,
                 output=output)
         return out
 
+    @verbose
+    def compute_vol_morph_mat(self, *, verbose=None):
```
agramfort, line 460: the SourceMorph object is meant to be constructed from compute_source_morph and has only 2 public methods (apply and save). I would therefore expect the precomptuation of the vol morph mat to be exposed only via compute_source_morph.

larsoner: Indeed I had it that way originally actually, but this way is actually more functional: it allows people who have already saved volume morphs (from 0.21, etc.) to load them, do this computation, and then save without needing to call `compute_source_morph` will all the same parameters they did the first time.

I also don't see any problem with adding more methods to this object if we need them.

larsoner: (For example I just had 100 subjects I did this with, it motivated me to do it this way)

agramfort:
I think it should be exposed anyway in the function.

For new users who do not have existing files that would be enough?

if it's for you can you make compute_vol_morph_mat method private?

>

larsoner:
> I think it should be exposed anyway in the function.

I see that you're motivated to make it so that `compute_source_morph` contains all options one would need. I don't think this is ultimately sustainable -- as we find other things that we can add "after the fact" like this (other computational shortcuts, metrics, plotting functions, etc.) it will be unsustainable. I'd rather return the object in `compute_vol_morph_mat` so that you can do `compute_source_morph(...).compute_vol_morph_mat().save(...)` or so.

I would argue instead we should make it so that `compute_source_morph` contains everything needed to construct the "core morph", and anything else (computational shortcuts, plotting, etc.) can live in other methods.

If you're worried about discoverability, I'd prefer to add something to the `Notes` section of the constructor. This plus I think the huge note in the example is probably good enough for people to see it.

> if it's for you can you make compute_vol_morph_mat method private?

I imagine that most people who have been using a volume morph in actual analyses would want to do this actually. And being able to just iterate over a `glob.glob('*-vol-morph.h5')` and do a save-load round-trip is a nice thing to be able to do.

mne/morph.py
agramfort: I slept over it and I think we should expose a precompute parameter in compute_source_morph. Users don't have to understand the concepts of morph_mat  which I doubt many users understand. It's now completely hidden in the surface morphing API. So what I would suggest is:

`compute_source_morph(..., precompute='auto')`

where it's True by default for surface and False for volume.

then we can keep the compute_vol_morph_mat method public.

and what would be just great would be to automatically compute the vol_morph_mat on apply if you can infer if it's going to be faster to first precompute and then apply with a dot.

larsoner: > where it's True by default for surface and False for volume.

I'm not going to implement it quite this way because we never want to / cannot turn off precompute for surface -- we actually removed the code path that did it at some point because it wasn't faster unless you had a single time point, and even in that case it wasn't much faster. But I can leave it as `precompute : bool` in case at some point we want to add other options.

agramfort (approved):
good enough !

larsoner:
3.8 was just a timeout, merging!

Issue #8363, opened 2020-10-13
BUG: save_movie doesn't block inputs


larsoner:
When creating a bunch of movies in a script, windows pop up. If they get focus (e.g., on Linux where focus stealing isn't always handled properly) then doing something like hitting "f" will cause the "fly" (I assume?) callback to occur and the brain goes away, ruining the movie. It would be nice if we could block / skip / ignore both mouse and keyboard inputs to the window during the `save_movie` loop.

cc @GuillaumeFavelier

GuillaumeFavelier:
Would it work if you set `pyvista.OFF_SCREEN = True` to avoid having the windows opening in the first place?

larsoner:
I doubt this would work with `time_viewer=True` / traces

GuillaumeFavelier:
Alternatively, I was thinking about a dedicated offline thread with disabled interaction. In theory, this could build the movie in the background without blocking interactions on the main thread.

larsoner:
That sounds a lot more complicated code-wise (threads always are). I'm also not sure it will work because the rendering thread will for example need the camera in one fixed position and the user could be moving it, so somehow there have to be two states. I'm not sure it would really even end up being usable since the rendering thread will not release the GIL very often and could be CPU intensive.

To "disable interaction" I was really just thinking to override the event handlers/listeners of PyVista+VTK, and once rendering is complete (via try/finally) restore them. Ideally this would mean making some private PyVista attributes public, or just living with modifying some private things to be no-ops, and modifying them. At the VTK level I think we can just iterate over all renderer layers, [disable their interaction](https://github.com/pyvista/pyvista/blob/master/pyvista/plotting/plotting.py#L3879), then at the end, reenable their interaction -- so the question is what do we need to do to disable all PyVista shortcuts...

GuillaumeFavelier:
> so the question is what do we need to do to disable all PyVista shortcuts...

AFAIK those are bound to vtk KeyPressEvent

PR #8367, opened 2020-10-14
MRG: Toggle interaction with context manager

GuillaumeFavelier:
This PR adds a context manager to toggle interactivity during movie in Brain.

This is still a work in progress, I got some warnings in `save_movie` if the user interacts:

```
WARN| vtkInteractorStyleTrackballCamera (0x563558f5de30): no current renderer on the interactor style.
```

@larsoner this is what I tried so far

Closes https://github.com/mne-tools/mne-python/issues/8363

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -2538,18 +2538,19 @@ class Brain(object):
                     framerate=24, interpolation=None, codec=None,
                     bitrate=None, callback=None, time_viewer=False, **kwargs):
         import imageio
-        images = self._make_movie_frames(
-            time_dilation, tmin, tmax, framerate, interpolation, callback,
-            time_viewer)
-        # find imageio FFMPEG parameters
-        if 'fps' not in kwargs:
-            kwargs['fps'] = framerate
-        if codec is not None:
-            kwargs['codec'] = codec
-        if bitrate is not None:
-            kwargs['bitrate'] = bitrate
-
-        imageio.mimwrite(filename, images)
+        from ..backends._pyvista import _disabled_interaction
+        with _disabled_interaction(self._renderer):
+            images = self._make_movie_frames(
+                time_dilation, tmin, tmax, framerate, interpolation, callback,
+                time_viewer)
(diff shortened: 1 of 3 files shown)


larsoner:
Did you try `renderer.SetInteractive(False)`?

GuillaumeFavelier:
Yes I did (it's commented)  but neither mouse or key press is disabled.

larsoner:
Maybe https://public.kitware.com/pipermail/vtkusers/2005-March/029165.html ?

Or maybe we can set the `_shadow_renderer` to actually be interactive and it will steal all events?

GuillaumeFavelier:
Actually, it's not difficult to `RemoveAllObservers`(https://vtk.org/doc/nightly/html/classvtkObject.html#a624b110b0073b8152cc5a8f5dfa23580), the tricky part is to restore them.

> Or maybe we can set the _shadow_renderer to actually be interactive and it will steal all events?

I will try that now

GuillaumeFavelier:
`_shadow_renderer` did not work for me either.

GuillaumeFavelier:
This works from `pyvistaqt>0.2.0`

mne/viz/_brain/_brain.py
```diff
@@ -2538,18 +2538,19 @@ def _save_movie(self, filename, time_dilation=4., tmin=None, tmax=None,
...
-        if bitrate is not None:
-            kwargs['bitrate'] = bitrate
-
-        imageio.mimwrite(filename, images)
+        from ..backends._pyvista import _disabled_interaction
+        with _disabled_interaction(self._renderer):
+            images = self._make_movie_frames(
+                time_dilation, tmin, tmax, framerate, interpolation, callback,
+                time_viewer)
+            # find imageio FFMPEG parameters
```
larsoner, line 2546: From here on does not need to be nested (if `ffmpeg` nicely releases the GIL, which it might, we can interact which is nice)

larsoner (approved):

agramfort (approved):

larsoner:
Thanks @GuillaumeFavelier !

PR #8368, opened 2020-10-14
WIP : rank was not passed to compute_depth_prior

agramfort:
@hichamjanati can you see if it fixes your problem with MxNE?

@larsoner do you confirm it's a bug?


larsoner (approved):
Yes this looks like a bug. Should only affect sparse solvers (or very advanced users), though, since the defaults are different for MNE and beamformers.

PR #8372, opened 2020-10-15
MRG, DOC: Make clear which functions have been renamed

hoechenberger:
@SophieHerbst got lost because after an MNE update,`mne.channels.find_ch_connectivity` and `mne.channels.read_ch_connectivity` were gone. Turns out these functions were renamerd, and this change was mentioned in the changelog indeed, but apparently not explicitly enough. Searching the docs for the function name wouldn't bring up helpful results either. This PR explicitly lists all `*connectivity` functions that have been renamed to `*adjacency` in the changelog, so it should be easier to discover these changes.

### DIFF
--- a/doc/changes/0.21.inc
+++ b/doc/changes/0.21.inc
@@ -399,6 +399,18 @@ API changes
 
   1. Arguments to clustering functions, such as `mne.stats.permutation_cluster_test`, and
   2. Function names for defining adjacency, such as `mne.spatio_temporal_src_adjacency` replacing ``mne.spatio_temporal_src_connectivity``.
+  
+  The complete list of changed function names is:
+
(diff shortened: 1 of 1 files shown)


agramfort (approved):

larsoner:
For PRs that change doc only it's better to do `[skip travis] [skip github]` rather than `[skip ci]` because that way we get to see the doc build. I'll build locally and look to make sure there are no parsing errors

hoechenberger:
> For PRs that change doc only it's better to do `[skip travis] [skip github]` rather than `[skip ci]` because that way we get to see the doc build. I'll build locally and look to make sure there are no parsing errors

Damn, sorry, this is exactly what I thought `[skip ci]` would do. Noting this down.

larsoner:
Thanks @hoechenberger

PR #8373, opened 2020-10-15
fixes for new sklearn

agramfort:
decovered as working on sklearn master


larsoner (approved):
SciPy 1.1 was released over 2 years ago so we might consider also bumping our version req to 1.1 (then we don't need that `orth` fix at all)

agramfort:
floor is yours @larsoner :)

larsoner:
Backported, thanks @agramfort

Issue #8304, opened 2020-09-24
raw.plot() bottom cannot be >= top


vagechirkov:
Dear all,

I faced the following bug during plotting raw fif data:

```python

----> 3 mne.viz.plot_raw(raw)

<decorator-gen-145> in plot_raw(raw, events, duration, start, n_channels, bgcolor, color, bad_color, event_color, scalings, remove_dc, order, show_options, title, show, block, highpass, lowpass, filtorder, clipping, show_first_samp, proj, group_by, butterfly, decim, noise_cov, event_id, show_scrollbars, show_scalebars, verbose)

/home/cdm/miniconda3/envs/meg_spikes/lib/python3.6/site-packages/mne/viz/raw.py in plot_raw(raw, events, duration, start, n_channels, bgcolor, color, bad_color, event_color, scalings, remove_dc, order, show_options, title, show, block, highpass, lowpass, filtorder, clipping, show_first_samp, proj, group_by, butterfly, decim, noise_cov, event_id, show_scrollbars, show_scalebars, verbose)
    428 
    429     _prepare_mne_browse_raw(params, title, bgcolor, color, bad_color, inds,
--> 430                             n_channels)
    431 
    432     # plot event_line first so it's in the back

/home/cdm/miniconda3/envs/meg_spikes/lib/python3.6/site-packages/mne/viz/raw.py in _prepare_mne_browse_raw(params, title, bgcolor, color, bad_color, inds, n_channels)
    663     _set_window_title(params['fig'], title or "Raw")
    664     # most of the axes setup is done in _prepare_mne_browse
--> 665     _prepare_mne_browse(params, xlabel='Time (s)')
    666     ax = params['ax']
    667     ax_hscroll = params['ax_hscroll']

/home/cdm/miniconda3/envs/meg_spikes/lib/python3.6/site-packages/mne/viz/utils.py in _prepare_mne_browse(params, xlabel)
    661         right=1 - _inch_to_rel_dist(fig, r_border),
    662         bottom=_inch_to_rel_dist(fig, b_border, horiz=False),
--> 663         top=1 - _inch_to_rel_dist(fig, t_border, horiz=False)
    664     )
    665     # Main axes must be a `subplot` for `subplots_adjust` to work (so user can

/home/cdm/miniconda3/envs/meg_spikes/lib/python3.6/site-packages/matplotlib/figure.py in subplots_adjust(self, left, bottom, right, top, wspace, hspace)
   2383                                  "tight_layout: setting "
[...]

PR #7299, opened 2020-02-08
MRG, FIX: make labels in epochs.plot start at 0, fixes issue #7260

SophieHerbst:
Makes epochs.plot() start at 0 instead of 1 (issue #7260). 
Note: a bug with spyder (#6528) prevents me from testing what happens after closing the interactive plot. I checked that the indices returned for manually selected epochs start with 0, too. 

Closes #7260

PR #8381, opened 2020-10-16
MRG, VIZ, MAINT: Use new figure class for epochs.plot and ICA.plot_sources

drammock:
this is an element of #7751 

Basically working already, though needs testing and maybe some cosmetics.

### DIFF
--- a/mne/viz/raw.py
+++ b/mne/viz/raw.py
@@ -557,8 +557,10 @@ def _setup_channel_selections(raw, kind, order):
         selections_dict = _divide_to_regions(raw.info)
         keys = _SELECTIONS[1:]  # omit 'Vertex'
     else:  # kind == 'selection'
-        from ..io import RawFIF, RawArray
-        if not isinstance(raw, (RawFIF, RawArray)):
+        from ..channels.channels import _get_ch_info
+        (has_vv_mag, has_vv_grad, *_, has_neuromag_122_grad, has_csd_coils
+         ) = _get_ch_info(raw.info)
+        if not (has_vv_grad or has_vv_mag or has_neuromag_122_grad):
             raise ValueError("order='selection' only works for Neuromag "
                              "data. Use order='position' instead.")
         selections_dict = OrderedDict()
--- a/mne/viz/raw.py
+++ b/mne/viz/raw.py
@@ -7,23 +7,20 @@
 # License: Simplified BSD
 
 from functools import partial
-from collections import OrderedDict, defaultdict
+from collections import OrderedDict
 
 import numpy as np
 
 from ..annotations import _annotations_starts_stops
 from ..filter import create_filter
-from ..io.pick import (pick_types, _pick_data_channels, pick_info,
-                       pick_channels, _DATA_CH_TYPES_ORDER_DEFAULT)
-from ..utils import verbose, _ensure_int, _validate_type, _check_option
+from ..io.pick import pick_types, _pick_data_channels, pick_info, pick_channels
+from ..utils import verbose, _validate_type, _check_option
 from ..time_frequency import psd_welch
 from ..defaults import _handle_default
 from .topo import _plot_topo, _plot_timeseries, _plot_timeseries_unified
-from .utils import (_prepare_mne_browse, figure_nobar, plt_show,
-                    _get_figsize_from_config, _setup_browser_offsets,
-                    _compute_scalings, _handle_decim, _check_cov,
-                    _set_ax_label_style, _simplify_float, _check_psd_fmax,
-                    _set_window_title, _shorten_path_from_middle)
+from .utils import (plt_show, _compute_scalings, _handle_decim, _check_cov,
+                    _check_psd_fmax, _shorten_path_from_middle,
+                    _get_channel_plotting_order, _make_event_color_dict)
 
 _RAW_CLIP_DEF = 1.5
 
@@ -68,11 +65,8 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
 
     bad_color : color object
         Color to make bad channels.
-    event_color : color object | dict
-        Color(s) to use for events. For all events in the same color, pass any
-        matplotlib-compatible color. Can also be a `dict` mapping event numbers
-        to colors, but if so it must include all events or include a "fallback"
-        entry with key ``-1``.
+    %(event_color)s
+        Defaults to ``'cyan'``.
     scalings : 'auto' | dict | None
         Scaling factors for the traces. If any fields in scalings are 'auto',
         the scaling factor is set to match the 99.5th percentile of a subset of
@@ -280,17 +274,7 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
     # determine trace order
     ch_names = np.array(raw.ch_names)
     ch_types = np.array(raw.get_channel_types())
-    if order is None:
-        # for backward compat, we swap the first two to keep grad before mag
-        ch_type_order = list(_DATA_CH_TYPES_ORDER_DEFAULT)
-        ch_type_order = tuple(['grad', 'mag'] + ch_type_order[2:])
-        order = [pick_idx for order_type in ch_type_order
-                 for pick_idx, pick_type in enumerate(ch_types)
-                 if order_type == pick_type]
-    elif not isinstance(order, (np.ndarray, list, tuple)):
-        raise ValueError('order should be array-like; got '
-                         f'"{order}" ({type(order)}).')
-    order = np.asarray(order)
+    order = _get_channel_plotting_order(order, ch_types)
     n_channels = min(info['nchan'], n_channels, len(order))
     # adjust order based on channel selection, if needed
     selections = None
@@ -300,21 +284,8 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
         default_selection = list(selections)[0]
         n_channels = len(selections[default_selection])
 
-    # if event_color is a dict
-    if isinstance(event_color, dict):
-        event_color = {_ensure_int(key, 'event_color key'): value
-                       for key, value in event_color.items()}
-        default = event_color.pop(-1, None)
-        default_factory = None if default is None else lambda: default
-        event_color_dict = defaultdict(default_factory)
-        for key, value in event_color.items():
-            if key < 1:
-                raise KeyError('event_color keys must be strictly positive, '
-                               f'or -1 (cannot use {key})')
-            event_color_dict[key] = value
-    # if event_color is a string or other MPL color-like thing
-    else:
-        event_color_dict = defaultdict(lambda: event_color)
+    # handle event colors
+    event_color_dict = _make_event_color_dict(event_color, events, event_id)
 
     # handle first_samp
     first_time = raw._first_time if show_first_samp else 0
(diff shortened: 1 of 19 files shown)


hoechenberger:
Hello @drammock, maybe I'm doing something wrong – and I really haven't looked at the code changes here – but right-clicking on a channel name in a plot produced by `ICA.plot_sources(evoked)` should bring up a property plot, is that right? Right-clicking doesn't do anything for me on macOS, both with the default macOS and with the `Qt5Agg` backends. I surely must be missing something?

drammock:
at the moment only `ica.plot_sources(raw)` has been updated.  `ica.plot_sources(epochs)` and ``ica.plot_sources(evoked)` haven't yet been changed.

hoechenberger:
Thanks @drammock! It's beautiful!

What I've noticed:

- clicking on the Help button doesn't open the help window – once I right-click on a component name and the property plot opens, the help window opens behind the property window
- I get the macOS beach ball for a few seconds after doing the right click. I wonder if it would be possible to prevent this blocking, and if we could even print `processing …` or something similar to inform the user that we haven't crashed, but are merely busy?

drammock:
@jasmainak is it easy for you to test the autoreject package with this branch, to make sure that `epoch_colors` still works as expected?

drammock:
OK, now `raw.plot()`, `epochs.plot()`, `ica.plot_sources(raw)`, and `ica.plot_sources(epochs)` all use the new figure classes.  Assuming I didn't break anything with these last few commints, this is ready for review.  (sorry for the monstrous diff; it turned out to be easier to do ICA and epochs at the same time).  Notes:

1. here are the circle renderings of the tutorials I updated this time: [ica tutorial](https://23291-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_40_artifact_correction_ica.html) and [epochs tutorial](https://23291-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_20_visualize_epochs.html#plotting-epochs-as-time-series). Note that in the ICA tutorial I used `show_scrollbars=False` in all the `plot_sources()` calls, which I think is generally better for tutorials except that you lose the x-axis label.
2. bad ICA components now get grey color instead of red color, similar to bad channels in the raw and epoch plots.
3. I changed the keyboard shortcuts for scrolling through epochs in the time domain, to make them more consistent with `raw.plot()`.  Previously, in `epochs.plot()` → and ← would scroll a full window width at a time.  Now, I've set it so that → and ← scroll a single epoch, and now shift→ and shift← scroll a full window width at a time.  This is consistent with `raw.plot()` where → does 1/4 window width and shift→ does full window width.
4. The labeling of epochs at the bottom of the main window and along the bottom edge of the scrollbar are now *numbers from epochs.selection* instead of sequential indices. This overrides the work in #7299 so I'm keen to hear @SophieHerbst's opinion whether this is a welcome or unwelcome change.
5. I have deprecated the parameter `event_colors` in `epochs.plot()` in favor of the similarly-named `event_color` parameter. It turns out that `raw.plot()` has an `event_color` parameter that was generally more flexible than what used to be implemented for epochs. Now, they both have the same name, they work the same way, and they both do everything each of them used to be able to do. Here is the new docstring: https://github.com/mne-tools/mne-python/blob/cc4dff991ae5bf645bae32d130db88e77c98b780/mne/utils/docs.py#L98-L105
6. The horizontal scrollbar in the epochs plot now has divisions showing epoch boundaries.
7. clicking in the data axes of epochs plots is now more similar to raw plots. In particular, you must click on data traces to mark bad epochs (previously clicking between the data traces was good enough). Now, clicking between data traces places the green vertical line (like it does on raw plots) and right-clicking removes the green vertical line (again, like raw plots).
8. all three plot types allow right-click on channel names to bring up additional context plots. raw plots → channel location, epoch plots → epochs image for that channel, ica plots → a `plot_properties` window for that component.

of these, I think items 3, 4, 5, and 7 are **prominent (non-cosmetic) changes** which I think should get buy-in from at least a couple other devs before merge, and some of them probably deserve changelog entries.  Anything else you notice that seems weird  --- anything at all --- please mention it now.

larsoner:
> used show_scrollbars=False in all the plot_sources() calls, which I think is generally better for tutorials except that you lose the x-axis label.

We could allow `show_scrollbars='xaxis'` or something to specify just to keep the x axis. It would be better if the name were `decorate=True` then `decorate='xaxis'` would make more sense but I'm okay with a bit of a misnomer here (not worth deprecation cycle to make the name more general).

> This overrides the work in #7299

This appeared to just be about 0- versus 1-based indexing so I think it's fine / complementary

Otherwise all these API changes sound like improvements to me, +1 for all of them. I'll wait to do a thorough review until we see if others agree and no other major changes are needed.

@hoechenberger I see a heart, does that mean you're happy with all points?

hoechenberger:
> @hoechenberger I see a heart, does that mean you're happy with all points?

Yes I’m quite happy! Still got to actually try it out, but I do love the list of changes

drammock:
> We could allow `show_scrollbars='xaxis'` or something to specify just to keep the x axis. It would be better if the name were `decorate=True` then `decorate='xaxis'` would make more sense but I'm okay with a bit of a misnomer here (not worth deprecation cycle to make the name more general).

WDYT about just always showing an x-axis label, regardless of whether scrollbars are visible or hidden?

larsoner:
I'm not sure how that interacts with `zen` mode, but I never really use this so I'm happy to defer to your judgment on this

larsoner:
I have no strong opinion, up to you. I would say if you want to go with as much backward compat as possible add `show_scrollbars='xaxis'` option, if it's not a big concern then just change the behavior of `False`

agramfort (approved):
it works like a charm !!!! and with 2000 lines removed 🍾 

@drammock you are like 🎅 to me !

hoechenberger:
LOL I can imagine that scene!!
"Daddy, what's for Christmas?" – "Best present EVER! 2,000 SLOC removed, BOOYAA!!! 🤡" – "😱"

larsoner:
And longstanding functionality added (such as butterfly mode for epochs, right?) and waaaay more DRY code, etc. Bravo!

I'll take a look and try things Monday. @hoechenberger do you want to test it out? I don't use the epochs or ICA very much so I'm not the best person to do it.

larsoner:
Does this close #4850, or is more needed for that?

hoechenberger:
> @hoechenberger do you want to test it out? I don't use the epochs or ICA very much so I'm not the best person to do it.

Yes but won't be able to do it before tonight or tomorrow UTC+1

drammock:
> Does this close #4850, or is more needed for that?

No but it should be quite easy to do. Will add that.

mne/viz/raw.py
```diff
@@ -552,13 +552,16 @@ def _setup_channel_selections(raw, kind, order):
...
     _check_option('group_by', kind, ('position', 'selection'))
     if kind == 'position':
         selections_dict = _divide_to_regions(raw.info)
         keys = _SELECTIONS[1:]  # omit 'Vertex'
     else:  # kind == 'selection'
         from ..io import RawFIF, RawArray
-        if not isinstance(raw, (RawFIF, RawArray)):
+        if isinstance(raw, BaseEpochs):
+            pass
+        elif not isinstance(raw, (RawFIF, RawArray)):
```
drammock, line 564: @larsoner is there a clever way to detect whether an Epochs object contains Neuromag data?  I was never satisfied for Raw checking against RawFIF and RawArray, so if you have a better idea on that too I'd love to hear it.

larsoner: If this has to do with `selection` mode, I assume you are talking about Neuromag MEG data (as EEG can be treated like any other format). The robust way would be to check for coil_type of any of info['chs'] being one of the Neuromag types. Part of this sort of functionality might already live in find_layout somehow, because it needs to inspect channels to see if we have MEG channels that we can use a nice built-in layout for.

drammock:
Yep, it's for selection mode (which now works for epochs). I'll `git grep` for coil_type and see what I can find. Thanks.

Sent from ProtonMail mobile

drammock:
Ready from my end. @larsoner I think this now cleanly implements `group_by='selection'` for epochs.  @hoechenberger have you had a chance to test this out yet?  Here's a quick testing script:

```python
import os
import numpy as np
import mne
plt.ion()
                                    'sample_audvis_raw.fif')
raw.del_proj()
raw.crop(tmax=20)
raw.apply_proj()
ecg_proj_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                             'sample_audvis_ecg-proj.fif')
ecg_projs = mne.read_proj(ecg_proj_file)
raw.add_proj(ecg_projs)
sample_data_noise_cov_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                                          'ernoise-cov.fif')
noise_cov = mne.read_cov(sample_data_noise_cov_file)
event_dict = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3}
event_color = {1: 'r', 2: 'g', -1: 'y'}
epochs = mne.Epochs(raw, events, event_id=event_dict, preload=True, proj=False)
fig = epochs.plot(n_epochs=6, event_color=event_color, events=epochs.events,
                  event_id=event_dict, group_by='selection')
# for a treat (not useful, but needed to support autoreject package):
# epoch_colors = [np.random.choice(list('rgbcmyk'), epochs.info['nchan']
#                                  ).tolist() for _ in range(len(epochs))]
# fig = epochs.plot(epoch_colors=epoch_colors, butterfly=True)
```

hoechenberger:
> Here's a quick testing script:

How convenient 😍 Will rest momentarily.

hoechenberger:
> Will rest momentarily.

Test. Will **test** momentarily. Ouch, Freudian 😅

hoechenberger:
Hey @drammock,

I'm currently forced to use my old 2013 MacBook Air, which has a small screen and is rather slow, and when trying to resize the Raw browser, I can reproducibly make Python crash. The traceback would like something like the following:

```python
  File "/tmp/foo.py", line 24, in <module>
    fig = epochs.plot(n_epochs=6, event_color=event_color, events=epochs.events,
  File "/Users/richard/Development/mne-python/mne/epochs.py", line 1111, in plot
    return plot_epochs(self, picks=picks, scalings=scalings,
  File "/Users/richard/Development/mne-python/mne/viz/epochs.py", line 881, in plot_epochs
    fig = _browse_figure(**params)
  File "/Users/richard/Development/mne-python/mne/viz/_figure.py", line 2112, in _browse_figure
    fig = _figure(inst=inst, toolbar=False, FigureClass=MNEBrowseFigure,
  File "/Users/richard/Development/mne-python/mne/viz/_figure.py", line 2102, in _figure
    fig = figure(FigureClass=FigureClass, **kwargs)
  File "/Users/richard/miniconda3/envs/mne/lib/python3.8/site-packages/matplotlib/pyplot.py", line 687, in figure
    figManager = new_figure_manager(num, figsize=figsize,
  File "/Users/richard/miniconda3/envs/mne/lib/python3.8/site-packages/matplotlib/pyplot.py", line 315, in new_figure_manager
    return _backend_mod.new_figure_manager(*args, **kwargs)
  File "/Users/richard/miniconda3/envs/mne/lib/python3.8/site-packages/matplotlib/backend_bases.py", line 3493, in new_figure_manager
    fig = fig_cls(*args, **kwargs)
  File "/Users/richard/Development/mne-python/mne/viz/_figure.py", line 357, in __init__
    self.subplotpars.update(left=left, bottom=bottom, top=top, right=right)
  File "/Users/richard/miniconda3/envs/mne/lib/python3.8/site-packages/matplotlib/figure.py", line 205, in update
    raise ValueError('bottom cannot be >= top')
ValueError: bottom cannot be >= top
```

Here's a video:
https://youtu.be/c_ddbYydzgs

I'm using `Qt5Agg`. Not sure if there's anything you can do about this…

I even once wasn't able to open the Raw browser anymore even after restarting Python, it would instantly crash with the above error message. After deleting `~/.mne`, it started working again.

Will look at the new features now and report back

drammock:
sounds like you hit #8304

hoechenberger:
@drammock I'm now constantly running into this, cannot really test your new viz :( Never saw this issue before, but this is also the smallest screen size I've ever used with MNE: 1366 x 768 pixels

hoechenberger:
@drammock I will try to give it another shot tomorrow, going to bed now :) But if you feel ready to get this merged, don't wait for me!

hoechenberger:
Thanks for addressing #8304, @drammock! This seems to be fixed for me now. However, bad news is, instead of this traceback plus (sometimes) a segfault I now **only** receive the segfault 💩Could be an issue with PyQt5 on macOS, I don't know. The problem still seems to be related to resizing somehow. I'm wondering if it would be possible to disable redrawing while the window is being resized? You know, like inn the Windows 3.11 days? Because I have the suspicion that this might be (part of) the issue here...

hoechenberger:
But I don't intend to filibuster this PR just because my old computer is having some troubles. If I'm the only one experiencing this, maybe we can just go ahead and merge for now… I unfortunately won't be able to provide feedback / test, though. But we can always iterate later. (Will hopefully have my 2019 MBP again in a week or two)

larsoner:
> Could be an issue with PyQt5 on macOS, I don't know.

Exceptions thrown during a callback will cause PyQt to abort. This should in theory be handled by `matplotlib` -- IIRC they have some mechanisms in place to make it so that any PyQt event handler of theirs does not pass on the exceptions (this is why they get printed instead). So this is probably a bug with their implementation somewhere. It would be great if we could isolate this and open an issue, or fix the bug in their code.

@hoechenberger are you on the latest matplotlib version? And what version of PyQt5?

mne/viz/tests/test_raw.py
```diff
@@ -273,11 +260,11 @@ def test_plot_raw_selection(raw, new_mpl):
...
-    _click_ch_name_helper(fig, ch_index=1, button=1)  # mark good
-    assert sel_fig.lasso.ec[:, 0].sum() == 0          # all channels black
+    assert sel_fig.lasso.ec[:, 0].sum() == 0   # R of RGBA zero for all chans
+    _click_ch_name(fig, ch_index=1, button=1)  # mark bad
+    assert sel_fig.lasso.ec[:, 0].sum() == 1   # one channel red
+    _click_ch_name(fig, ch_index=1, button=1)  # mark good
+    assert sel_fig.lasso.ec[:, 0].sum() == 0   # all channels black
     # test lasso
     sel_fig._set_custom_selection()  # lasso empty → should do nothing
     sensor_ax = sel_fig.mne.sensor_ax
```
larsoner, line 270: Okay now it looks like you can fix the test below:

https://travis-ci.org/github/mne-tools/mne-python/jobs/742563175#L6462-L6469

```
mne/viz/tests/test_raw.py:280: in test_plot_raw_selection
    assert got == want
E   AssertionError: assert ['MEG 0121', 'MEG 0122', 'MEG 0123'] == ['MEG 0111', 'MEG 0112', 'MEG 0131', 'MEG 0132', 'MEG 0133']
E     At index 0 diff: 'MEG 0121' != 'MEG 0111'
E     Right contains 2 more items, first extra item: 'MEG 0132'
E     Full diff:
E     - ['MEG 0111', 'MEG 0112', 'MEG 0131', 'MEG 0132', 'MEG 0133']
E     + ['MEG 0121', 'MEG 0122', 'MEG 0123']
```

mne/viz/tests/test_raw.py
```diff
@@ -308,7 +295,7 @@ def test_plot_raw_ssp_interaction(raw):
     fig = raw.plot()
     # open SSP window
     _fake_click(fig, fig.mne.ax_proj, [0.5, 0.5])
-    assert len(plt.get_fignums()) == 2
+    assert len(plt.get_fignums()) == 2  # XXX this fails with minimal deps
```
larsoner, line 298: revert, doesn't seem to fail?

larsoner:
Pushed a commit to showcase `butterfly=True, group_by='selection'`. Feel free to remove this commit and put it elsewhere if the spot I chose isn't good enough. Other than that and the Travis failure, LGTM!

hoechenberger:
@larsoner 
I'm using Matplotlib 3.3.2, Qt 5.12.9, and PyQt 5.12.3

mne/viz/tests/test_raw.py
drammock: I had deleted that once, must have snuck back in during one of the many rebases

drammock:
@larsoner all green here, can we get this in so more folks can start testing it?

larsoner:
Yes, thanks @drammock !

agramfort:
awesome @drammock <https://github.com/drammock> !

>

hoechenberger:
🎈

larsoner:
@drammock looks like these are yours:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/4732/workflows/ffa530e7-f767-4af5-b1b3-ebe0b679c0f9/jobs/23476

Can you check and open a quick PR to fix?

PR #8384, opened 2020-10-19
FIX: fix for empty event durations

### LINKED ISSUE #8383: Problem with load eeglab file
Hi everyone!
I have problem with loading data. I have eeglab files (example: https://ufile.io/f/js83q) and after using mne.io.read_raw_eeglab
I get:
Reading rest1_1_trigg.fdt
```python-traceback
TypeError Traceback (most recent call last)
TypeError: only size-1 arrays can be converted to Python scalars

The above exception was the direct cause of the following exception:

ValueError Traceback (most recent call last)

<ipython-input-20-49d5d8853521> in <module>
----> 1 mne.io.read_raw_eeglab(os.path.join(folder_with_data, files_all[0]))

~.conda\envs\mne\lib\site-packages\mne\io\eeglab\eeglab.py in read_raw_eeglab(input_fname, eog, preload, uint16_codec, verbose)
219 .. versionadded:: 0.11.0
220 """
--> 221 return RawEEGLAB(input_fname=input_fname, preload=preload,
222 eog=eog, verbose=verbose, uint16_codec=uint16_codec)
223

<decorator-gen-197> in init(self, input_fname, eog, preload, uint16_codec, verbose)

~.conda\envs\mne\lib\site-packages\mne\io\eeglab\eeglab.py in init(self, input_fname, eog, preload, uint16_codec, verbose)
353
354 # create event_ch from annotations
--> 355 annot = read_annotations(input_fname)
356 self.set_annotations(annot)
357 _check_boundary(annot, None)

~.conda\envs\mne\lib\site-packages\mne\annotations.py in read_annotations(fname, sfreq, uint16_codec)
665
666 elif name.endswith('set'):
--> 667 annotations = _read_annotations_eeglab(fname,
668 uint16_codec=uint16_codec)
669

~.conda\envs\mne\lib\site-packages\mne\io\eeglab\eeglab.py in _read_annotations_eeglab(eeg, uint16_codec)
611 duration = np.zeros(len(onset))
612 if len(events) > 0 and hasattr(events[0], 'duration'):
--> 613 duration[:] = [event.duration for event in events]
614
615 return Annotations(onset=np.array(onset) / eeg.srate,

ValueError: setting an array element with a sequence.
```
I am using version 0.20.8 of mne and Python 3.8.5.
Can you help me with this problem?

mmagnuski:
@ociepkam - Try if this works for you.

#### Reference issue
Fixes #8383.

It seems some eeglab files contain empty durations - these are read as empty errays from matlab files and cause error when trying to set the `durations` array.

I'll add a test later.


ociepkam:
It works. Thank you!

mne/io/eeglab/eeglab.py
```diff
@@ -609,7 +609,12 @@ def _read_annotations_eeglab(eeg, uint16_codec=None):
     onset = [event.latency - 1 for event in events]
     duration = np.zeros(len(onset))
     if len(events) > 0 and hasattr(events[0], 'duration'):
-        duration[:] = [event.duration for event in events]
+        for idx, event in enumerate(events):
+            if isinstance(events[idx].duration, np.ndarray):
```
cbrnr, line 613: You could replace `events[idx].duration` with `event.duration`.

Also, why do you set `this_duration` to `np.nan` if `event.duration` is an ndarray (regardless of its contents)?

mmagnuski: > You could replace events[idx].duration with event.duration.

Yes, I have that waiting locally, I know its ugly. :)

> Also, why do you set this_duration to np.nan if event.duration is an ndarray (regardless of its contents)?

Good question - I don't know what is the default in our case. I can set it to 0 as well. I think I was biased by the contents of this file - some of the durations were `NaN` in this case. And reading works when the durations are NaN.
I don't remember how common are nan durations in eeglab, maybe that's something that changed in more recent versions.

mmagnuski:
@cbrnr Thanks for the comments, I'll return to this tomorrow.

mne/io/eeglab/eeglab.py
cbrnr: I was more referring to the fact that you assume a missing duration if the type of `event.duration` is an ndarray. Why is that? What is the value of the `ndarray` in this case and wouldn't it make more sense to condition on the value?

mmagnuski: the array is empty in this case, I don't know if there are circumstances when the duration is an array but is not empty. My guess is that this behavior is specific to reading matlab structures with loadmat - structures can have empty fields which are read as empty arrays. I can add a check to make sure that `len(event.duration) == 0`.

cbrnr: Yes, I think a check would be good. So this means that non-missing durations are never stored as an ndarray but a float?

cbrnr (approved):

agramfort:
@mmagnuski can you add a what's new? thx +1 for MRG

hoechenberger:
Can we add a test for this please? :)

mmagnuski:
@agramfort Sure!
@hoechenberger of course, by "I'll add a test later" I meant later in the time scope of this PR. :)

hoechenberger:
Great, looking forward!

mmagnuski:
Thanks for reminding @hoechenberger, I'm on low time budget recently, sorry for delays (but, hey, it's been only 18 days! ;) )
I'll get to this today.

hoechenberger:
@mmagnuski Great! And no worries, I can very much relate.

mmagnuski:
I don't understand why codecov is angry at me... :(

mne/io/eeglab/tests/test_eeglab.py
```diff
@@ -129,6 +129,23 @@ def test_io_set_raw_more(tmpdir):
...
+    io.savemat(empty_dur_fname,
+               {'EEG': {'trials': eeg.trials, 'srate': eeg.srate,
+                        'nbchan': eeg.nbchan,
+                        'data': 'test_negative_latency.fdt',
+                        'epoch': eeg.epoch, 'event': evnts,
+                        'chanlocs': eeg.chanlocs, 'pnts': eeg.pnts}},
+               appendmat=False, oned_as='row')
+    shutil.copyfile(op.join(base_dir, 'test_raw.fdt'),
+                    empty_dur_fname.replace('.set', '.fdt'))
+    read_raw_eeglab(input_fname=empty_dur_fname, preload=True)
```
larsoner, line 147: Something like this to go one step beyond the smoke test?
```suggestion
    raw = read_raw_eeglab(input_fname=empty_dur_fname, preload=True)
    assert np.isnan(raw.annotations.duration).any()
```

mmagnuski:
@larsoner - less smoke now and we can see the green CI fields. :)

mne/io/eeglab/tests/test_eeglab.py
```diff
@@ -129,6 +129,24 @@ def test_io_set_raw_more(tmpdir):
...
+               {'EEG': {'trials': eeg.trials, 'srate': eeg.srate,
+                        'nbchan': eeg.nbchan,
+                        'data': 'test_negative_latency.fdt',
+                        'epoch': eeg.epoch, 'event': evnts,
+                        'chanlocs': eeg.chanlocs, 'pnts': eeg.pnts}},
+               appendmat=False, oned_as='row')
+    shutil.copyfile(op.join(base_dir, 'test_raw.fdt'),
+                    empty_dur_fname.replace('.set', '.fdt'))
+    raw = read_raw_eeglab(input_fname=empty_dur_fname, preload=True)
+    assert (raw.annotations.duration == 0).all()
```
larsoner, line 148: Shouldn't some or all of these been `np.nan` based on the conditional in `eeglab.py` and the `ev.duration = np.array([], dtype='float')` above?

larsoner:
I killed most CIs as we can safely ignore them (I just changed `latest.inc` -- left CircleCI running just to be safe)

mne/io/eeglab/tests/test_eeglab.py
mmagnuski: No, the eeglab reader turns `np.nan` somewhere later into zeros. I can use zeros instead of nans if you think this would be better. But we may also decide to expose the nan durations in annotations at some time in the future.

mmagnuski:
Thanks for the corrections!

mne/io/eeglab/tests/test_eeglab.py
larsoner: Makes sense!

larsoner:
Thanks @mmagnuski !

mmagnuski:
Thanks @larsoner @cbrnr @agramfort @hoechenberger ! 🚀

PR #8389, opened 2020-10-19
MRG, API: Remove trans

larsoner:
I started working on `labels_to_stc` for VolSourceEstimate / volumetric atlases, and I realized that we don't actually need the `trans` argument of `extract_label_time_course` because we can always easily reconstruct the `src[0]['rr']` in MRI space using the `src[0]['shape']` and `src[0]['mri_src_t']`. This deprecates `trans` in `extract_label_time_course`, though I'm half tempted just to leave it in as an argument since it doesn't hurt anything to include it and it's possible at some point in the future we might want it.

It was introduced as an argument only in 0.21, which is sad, and a fast turnaround for deprecation. So maybe we should actually backport this so that people using 0.21.1 get the warning and just remove `trans` from their code sooner.


agramfort (approved):
I would not backport an API change

Issue #8385, opened 2020-10-19
Error during read of Nihon file


tharikjose:
#### Describe the bug
I am facing an error during the reading phase of a Nihon EEG file. The problem is during the reading of the .21E file. The EEG is not corrupted since it can be opened in the Nihon program.  

#### Steps to reproduce
import mne

raw = mne.io.read_raw_nihon('Nihon/Nihon_Salvo_em_Tres_Montagens/DA7521RO.EEG')

[MWE]: https://en.wikipedia.org/wiki/Minimal_Working_Example
[built-in datasets]: https://mne.tools/dev/overview/datasets_index.html
[sample]: https://mne.tools/dev/overview/datasets_index.html#sample

#### Expected results
Expected to get the raw data of the EEG file.

#### Actual results
Loading DA7521RO.EEG
Found 21E file, reading channel names.
Traceback (most recent call last):
  File "C:/Users/Sorrino/Documents/Faculdade/EESC/01 - Mestrado/Reading EEG/Reading  Nihon.py", line 6, in <module>
    raw = mne.io.read_raw_nihon('Nihon/Nihon_Salvo_em_Tres_Montagens/DA7521RO.EEG')
  File "C:\Users\Sorrino\anaconda3\envs\mne_dev\lib\site-packages\mne\io\nihon\nihon.py", line 46, in read_raw_nihon
    return RawNihon(fname, preload, verbose)
  File "<decorator-gen-218>", line 24, in __init__
  File "C:\Users\Sorrino\anaconda3\envs\mne_dev\lib\site-packages\mne\io\nihon\nihon.py", line 311, in __init__
    header = _read_nihon_header(fname)
  File "C:\Users\Sorrino\anaconda3\envs\mne_dev\lib\site-packages\mne\io\nihon\nihon.py", line 126, in _read_nihon_header
    _chan_labels = _read_21e_file(fname)
  File "C:\Users\Sorrino\anaconda3\envs\mne_dev\lib\site-packages\mne\io\nihon\nihon.py", line 119, in _read_21e_file
    _chan_labels[idx] = name.strip()
IndexError: list assignment index out of range

Process finished with exit code 1

#### Additional information
Platform:      Windows-10-10.0.18362-SP0
Python:        3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
Executable:    C:\Users\Sorrino\anaconda3\envs\mne\python.exe
[...]

PR #8396, opened 2020-10-21
Add fix to give a proper error message when an unsupported Nihon Kohden file is read.

fraimondo:
#### Reference issue

related #8385 

In #8385, the user wants to read an unsupported file 'EEG-1200A V01.00'. The problem is that it first tries to read the channel labels (which in this file version are more than in the supported versions) and it raises a `IndexError`.

This fixes this issue by extending the list of channel labels as needed. Now the error is as expected:

```ValueError: Not a valid Nihon Kohden EEG file (EEG-1200A V01.00)```

#### Additional information
Issue #8385 expects the data to be read. However, in order to have that kind of fix, more information on the internals of the data format are needed.

### DIFF
--- a/mne/io/nihon/nihon.py
+++ b/mne/io/nihon/nihon.py
@@ -56,6 +56,7 @@ _valid_headers = [
     'EEG-2100  V02.00',
     'DAE-2100D V01.30',
     'DAE-2100D V02.00',
+    # 'EEG-1200A V01.00',  # Not working for the moment.
 ]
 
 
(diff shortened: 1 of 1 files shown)


agramfort:
@fraimondo are suggesting the file is broken or just that it's not supported presently?

fraimondo:
Not supported presently.

There are some differences between this new format version and the previous ones. It is already implemented in brainstorm.

Check this: https://sourceforge.net/p/biosig/mailman/message/35746530/

I'll need some more time to implement this new sub format in MNE. I already have the .EEG and .EDF for testing. For the moment, this will fix the error message so it is coherent.

larsoner (approved):

agramfort:
thx @fraimondo

larsoner:
@agramfort can you backport?

agramfort:
go for it I have too little time

>

Issue #8380, opened 2020-10-16
TFR analysis with more flexibility


mehdikuchi:
#### Addition of the tapering windows for TFR analysis

*The library only contains Morelet wavelet and it does not include other wavelets. There is no way to use other tapering windows but the DPSS and Gaussian windows. Nevertheless, users can define their own wavelet series and apply them to their signals using the cwt function; this however, may degrade the integrity of the library.* 

#### Implementation 
*The library does include the Morelet function which returns the wavelet series for the Morelet wavelet at different scales. Other wavelets can also be defined with the same signature of this function. Then, users can compute their TFR using these wavelet series with the use of cwt function of the time frequency module.*

#### Possible alternatives
*The code can also adds an ultimate freedom the the end users by letting them define their own tapering window and the oscillatory signal. This way the function can replace the Morelet function.*

agramfort:
@mehdikuchi you want to contribute something new? if so please do but start with small things. Maybe just add a new wavelet function?

PR #8397, opened 2020-10-21
TFR analysis with more flexibility #8380

mehdikuchi:
#### Reference issue
TFR analysis with more flexibility #8380

The wavelet function is added to the time_frequency module 

#### Additional information
The function accepts a callable with one input which specifies the length of the window to be used.
The resulting wavelet series can be used for analysis using the cwt function of the library.


mne/time_frequency/tfr.py
```diff
@@ -99,6 +99,59 @@ def morlet(sfreq, freqs, n_cycles=7.0, sigma=None, zero_mean=False):
...
+
+        Parameters
+        ----------
+        sfreq : float
+            The sampling Frequency.
+        freqs : array
+            Frequency range of interest (1 x Frequencies).
+        n_cycles : float | array of float, default 7.0
+            Number of cycles. Fixed number or one per frequency.
+        window : Callable object
```
larsoner, line 113: I would make `window` anything that is accepted by `get_window`:

https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.get_window.html

And then below we just call `get_window(window, n)` and we don't have to do any sanity checking (SciPy will do it for us)

mne/time_frequency/tfr.py
```diff
@@ -99,6 +99,59 @@ def morlet(sfreq, freqs, n_cycles=7.0, sigma=None, zero_mean=False):
         Ws.append(W)
     return Ws
 
+def wavelet(sfreq, freqs, n_cycles=7.0, window=np.hanning):
```
larsoner, line 102: functions are mutable and mutable defaults are bad, along with the `get_window` change this can be changed to the immutable `'hann'`

mne/time_frequency/tfr.py
```diff
@@ -99,6 +99,59 @@ def morlet(sfreq, freqs, n_cycles=7.0, sigma=None, zero_mean=False):
...
+    """Compute wavelets for the given frequency range with the specified windowing function.
+
+        Parameters
```
larsoner, line 105: This is all indented too far by one level (Sphinx will probably complain)

larsoner (commented):
Other than these comments, it would be good to have some tests for this. And can we use the `wavelet` function internally in our `*_morlet` functions?

mne/time_frequency/tfr.py
mehdikuchi: > This is all indented too far by one level (Sphinx will probably complain)
I didn't realize what you mean. do you mean that the comments is too lengthy? I have just copied it from the morlet function.

larsoner:
```suggestion
    Parameters
```
same for all lines that follow

mehdikuchi:
> Other than these comments, it would be good to have some tests for this. And can we use the `wavelet` function internally in our `*_morlet` functions?

Since all the wavelet calculations are passed to _compute_tfr, so yes but I have to update the _compute_tfr function as well

larsoner:
I would expect that the `morlet` function could be updated to be just a few lines:

https://github.com/mne-tools/mne-python/blob/master/mne/time_frequency/tfr.py#L40-L100

mne/time_frequency/tfr.py
mehdikuchi: Thanks

mehdikuchi:
> I would expect that the `morlet` function could be updated to be just a few lines:
> 
> https://github.com/mne-tools/mne-python/blob/master/mne/time_frequency/tfr.py#L40-L100

 I think it would be better to have this new function wavelet as the main wavelet generator and then any user can decide what 
type of window they want to use. However that would require a bit more work.

mehdikuchi:
The styling of the code need improvement.

mne/time_frequency/tfr.py
```diff
@@ -100,6 +100,58 @@ def morlet(sfreq, freqs, n_cycles=7.0, sigma=None, zero_mean=False):
     return Ws
 
 
+def wavelet(sfreq, freqs, n_cycles=7.0, window="hann"):
+    """Compute wavelets for the given frequency range.
```
agramfort, line 104: what type of wavelet are you implementing here? we have morlet which is one type of wavelet. You have others like Morse etc...  There should be one way to building the different wavelet functions with a consitent API.

mehdikuchi: I was looking for a window to apply in my tfr analysis something like what fieldtrip uses. I didn't find anything similar to that functionality, therefore I decided to add the window with the same oscillatory signal used in Morlet. Perhaps, it Is better to change the name of the function

agramfort: I know that in neuroscience we tend to call an oscillatory signal a wavelet but to me it's really not good practice. A wavelet should be zero mean and of unit norm see: https://en.wikipedia.org/wiki/Wavelet

here you use Fourier atom with a window. How is it called in Fieldtrip?

one of problem if your window is not zero mean is that if you have a constant signal you will get a non-zero TFR.

larsoner: > here you use Fourier atom with a window. How is it called in Fieldtrip?

And in one special case -- when the window is Gaussian -- it's a Morlet wavelet. I don't know if there are other wavelets that are just a complex exponential (Fourier atom) multiplied by a window. But calling this function `wavelet` indeed seems misleading...

agramfort:
for morlet pay attention to the \kappa_\sigma in
https://en.wikipedia.org/wiki/Morlet_wavelet

it is what makes the wavelet zero mean.

>

mne/time_frequency/tfr.py
mehdikuchi: The fieldtrip uses 
function [freq] = ft_freqanalysis(cfg, data)
where cfg is a struct defining how to implement the analysis. When you check the code for its hanning window it just gets the signal multiplied by the Fourier atom and apply that window by multiplication.

mehdikuchi: > I know that in neuroscience we tend to call an oscillatory signal a wavelet but to me it's really not good practice. A wavelet should be zero mean and of unit norm see: https://en.wikipedia.org/wiki/Wavelet
> 
> here you use Fourier atom with a window. How is it called in Fieldtrip?
> 
> one of problem if your window is not zero mean is that if you have a constant signal you will get a non-zero TFR.

The fieldtrip uses
function [freq] = ft_freqanalysis(cfg, data)
where cfg is a struct defining how to implement the analysis. When you check the code for its hanning window it just gets the signal multiplied by the Fourier atom and apply that window by multiplication.

agramfort:
@mehdikuchi I suggest you reopen a fresh PR as you have many commits in https://github.com/mne-tools/mne-python/pull/8397/commits that do not belong here

larsoner:
> @mehdikuchi I suggest you reopen a fresh PR as you have many commits in https://github.com/mne-tools/mne-python/pull/8397/commits that do not belong here

Agreed and the merge conflicts are extensive now. Feel free to open a clean PR when you get a chance if you're still interested @mehdikuchi !

Issue #8364, opened 2020-10-13
Adding Source Visualization for SEEG data Using stc_near_sensors


adam2392:
## Describe the problem
In https://github.com/mne-tools/mne-python/pull/8190, @larsoner suggests that we can add 3D translucent brain visualization of a time series for depth electrode data (i.e. SEEG). 

#### Describe your solution
"if we improve stc_near_sensors to allow returning a VolSourceEstimate object then nothing new needs to be done at the viz end. In other words, we "just" need to come up with a suitable way to interpolate the sEEG sensor activity into a volume, and once that's done, stc.plot and stc.plot_3d should just work."

Seems like the work in sequence comprises of:
1. make `stc_near_sensors` allow to return a `VolSourceEstimate` object
2. make `stc_near_sensors` have an interpolation scheme for SEEG sensor data into a 3D volume (possibly say just via 1/r^2 propagation?)
3. possibly add SEEG example showing `stc.plot` and `stc.plot_3d` similar to how Eric has done in https://github.com/mne-tools/mne-python/pull/8190

Any other suggestions to the above are appreciated! Will try to tackle this soon.

#### Describe possible alternatives
TBD

#### Additional context
Logging this as an issue, to not forget the conversation Eric brought up.

adam2392:
> It occurred to me that for surfaces we don't need any sort of source space because we can just directly interpolate onto the surfaces. Due to how we visualize volumes, though, that's not going to be the case. So I think we'll want to add a `src` argument to `stc_near_sensors` that's required for volume mode (and we can make surface mode use it if it's provided).

Sorry do you mind elaborating a little bit on what you mean here? 

How should I best tackle making this PR in your opinion? It seems that I want to perhaps first document and code-trace what's going on in `source_estimate.py::stc_near_sensors()` and then separate code into `if/else` blocks based on whether it's surface (i.e. pial) or volumetric (i.e. 3 MRI slices) visualization?

larsoner:
> Sorry do you mind elaborating a little bit on what you mean here?

We need to get sensor data into the brain, i.e., into source space. One way to do this is on the surface (already implemented), and in this case it's fine just to interpolate directly onto the high-density FreeSurfer mesh (100k+ vertices per hemisphere) because `stc.vertices` *always refer to numbers in this high density mesh*.

[...]

PR #8393, opened 2020-10-20
MRG, BUG: Fix plot_alignment

larsoner:
Fixes a bug introduced by #7747

Issue #8382, opened 2020-10-19
interactive sensor-level field patterns browser


wmvanvliet:
Our interactive source estimate viewer is so awesome, it makes me wish for an equally awesome sensor level viewer. For this, the venerable proprietary "Xfit" program that ships with the MegIn scanner may serve as inspiration.

With MEG, the raw sensor timecourses are quite difficult to interpret: gradiometers and magnetometers measure the field in different directions. The choice made by Xfit to make the field patterns the central visualization is a good one. See a short screencast of Xfit in action, here:

https://gfycat.com/jointaromatichorsemouse

For MNE-Python, I envision an interface similar to the source viewer, but with the field patterns instead of the pysurfer brain. On the bottom, we have the GFP timecourse.

What would be super nice would be the ability to "drop a dipole". Similar to clicking a vertex in the source viewer, have the ability to fit a dipole at the currently selected time, and see the dipole timecourse.

As extra bonus, it would be great if we would select a subset of channels and view the field patterns generated by only those sensors, and fit a dipole using only those sensors.

agramfort:
maybe a challenge for @GuillaumeFavelier ?

really cool video. It helps a lot to demo.

GuillaumeFavelier:
> really cool video. It helps a lot to demo.

It looks cool! :+1: 

When I have the cycles this week, I will prototype something but I might need help to guide me though :sweat_smile:

larsoner:
Some aspects of this are a special / enhanced case of "showing sensor data with the brain" -- we discussed this before at some point but I can't remember where. Along those lines I would propose (probably in PR-by-PR order):

1. A `brain.add_sensor_data`, where MEG sensors and EEG sensors are shown as in `plot_alignment`, but get colors according to their activity level. This is basically "just" parts of `plot_alignment` with some color mapping and time-point-updating built on top.
2. Some sort of surface mapping option, that allows you to map MEG sensor data to the MEG helmet (what @wmvanvliet is talking about) and EEG sensor data to the scalp, each with field map lines. This is "just" the functionality of `plot_evoked_field` mapped into the `Brain` object (plus interactivity).
3. GUI options for toggling these sensors and surface mappings on the fly.

Each of these I'd expect just to be 200-300 lines of code. If this makes sense to everyone, @GuillaumeFavelier feel free to move these points to #7162, and then we can discuss what I think is the trickier part:

- Some way to "place a dipole" and do `mne.dipole_fit` with it

[...]

PR #8402, opened 2020-10-22
[WIP] Trying to allow plotting SEEG activity in volumetric source space

### LINKED ISSUE #8388: Plotting MNI transformed ECoG
For decoding analysis across patients it becomes relevant to mainly work with MNI transformed coordinates. Using this tutorial https://mne.tools/stable/auto_tutorials/misc/plot_ecog.html, instead of the native FreeSurfer file the 'fsaverage' MNI brain can thus be used:

```
sample_path = mne.datasets.sample.data_path()
subject = 'fsaverage'
subjects_dir = sample_path + '\\subjects'

```

I installed MNE according to https://mne.tools/dev/install/mne_python.html, and using the `fsaverage` and run through the tutorial https://mne.tools/stable/auto_tutorials/misc/plot_ecog.html. 
At the line `mne.viz.set_3d_view(fig, 200, 70, focalpoint=[0, -0.005, 0.03])` I get the following error: 
"AttributeError: module 'mne.viz.backends._notebook' has no attribute '_set_3d_view'" 

My mne.sys_info() looks as follows: 
(base) C:\Users\ICN_admin\Documents\WRITE_BERLIN_BIDS>python -c "import mne; mne.sys_info()"
Platform:      Windows-10-10.0.18362-SP0
Python:        3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
Executable:    C:\Users\ICN_admin\Anaconda3\python.exe
CPU:           AMD64 Family 23 Model 8 Stepping 2, AuthenticAMD: 64 cores
Memory:        127.9 GB

mne:           0.21.0
numpy:         1.19.2 {blas=D:\\a\\1\\s\\numpy\\build\\openblas_info, lapack=D:\\a\\1\\s\\numpy\\build\\openblas_lapack_info}
scipy:         1.5.2
matplotlib:    3.1.1 {backend=Qt5Agg}

sklearn:       0.22.1
numba:         0.45.1
nibabel:       3.0.1
cupy:          Not found
pandas:        0.25.1
dipy:          Not found
mayavi:        4.7.2
pyvista:       0.26.1 {OpenGL 4.5.0 NVIDIA 432.00 via GeForce RTX 2070 SUPER/PCIe/SSE2}
vtk:           9.0.1
PyQt5:         5.9.2

Here the complete error message: 

grafik

adam2392:
## Reference issue
Fixes: #8364 
Closes #8388

1. Adds a `plot_seeg.py` tutorial for working w/ SEEG data analogous to `plot_ecog.py`.

2. Attempts to refactor `stc_near_sensors` to show volumetric SEEG activity with either `stc.plot` or `stc.plot_3d`

#### Additional information
Opening up PR to start iterating.

Ideally want to allow SEEG activity to be visualized as such: https://mne.tools/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#volumetric-rendering-3d
Assumes user has:
- electrode locations
- raw SEEG activity

Apologies if the code is nowhere where it should be. Never worked with the source API on mne, but now am eager to learn it if it'll help produce these nice viz of SEEG.

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -17,6 +17,7 @@ from scipy.sparse import coo_matrix, block_diag as sparse_block_diag
 from .cov import Covariance
 from .evoked import _get_peak
 from .filter import resample
+from .io.constants import FIFF
 from .surface import (read_surface, _get_ico_surface, mesh_edges,
                       _project_onto_surface)
 from .source_space import (_ensure_src, _get_morph_src_reordering,
@@ -3176,8 +3177,7 @@ def extract_label_time_course(stcs, labels, src, mode='auto',
 
 @verbose
 def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
-                     project=True, subjects_dir=None, src=None,
-                     coord_frame='mri', verbose=None):
+                     project=True, subjects_dir=None, src=None, verbose=None):
     """Create a STC from ECoG and sEEG sensor data.
 
     Parameters
@@ -3199,11 +3199,10 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         surface projection.
     %(subjects_dir)s
     src : instance of SourceSpaces
-        The source space. If you want to use a SourceMorph, first
-        morph the STC to a new subject (see Examples).
+        The source space.
 
-        .. versionchanged:: 0.18
-           Support for :class:`~nibabel.spatialimages.SpatialImage`.
+        .. warning:: If a surface source space is used, make sure that
+                     ``surf='pial'`` was used during construction.
     %(verbose)s
 
     Returns
@@ -3232,12 +3231,13 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         The value is given by the value of the nearest sensor, up to a
         ``distance`` (beyond which it is zero).
 
-    .. versionadded:: 0.21
+    .. versionadded:: 0.22
     """
     from scipy.spatial.distance import cdist, pdist
     from .evoked import Evoked
     _validate_type(evoked, Evoked, 'evoked')
     _validate_type(mode, str, 'mode')
+    _validate_type(src, (None, SourceSpaces), 'src')
     _check_option('mode', mode, ('sum', 'single', 'nearest'))
 
     # create a copy of Evoked using ecog and seeg
@@ -3255,7 +3255,7 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     pos = [pos[idx] for idx in range(len(pos)) if idx not in nan_inds]
 
     # coord_frame transformation from native mne "head" to MRI coord_frame
-    trans, _ = _get_trans(trans, 'head', coord_frame, allow_none=True)
+    trans, _ = _get_trans(trans, 'head', 'mri', allow_none=True)
 
     # convert head positions -> coord_frame MRI
     pos = apply_trans(trans, pos)
@@ -3263,32 +3263,45 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     # read surface files
     subject = _check_subject(None, subject, False)
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
-    # surface coordinate points
-    rrs = [read_surface(op.join(subjects_dir, subject,
-                                'surf', f'{hemi}.pial'))[0]
-           for hemi in ('lh', 'rh')]
-    offset = len(rrs[0])
-    rrs = np.concatenate(rrs)
-    rrs /= 1000.  # convert to meters
-
+    if src is None:  # fake a full surface one
+        # surface coordinate points
+        rrs = [read_surface(op.join(subjects_dir, subject,
+                                    'surf', f'{hemi}.pial'))[0]
+               for hemi in ('lh', 'rh')]
+        src = SourceSpaces([
+            dict(rr=rr / 1000., vertno=np.arange(len(rr)), type='surf',
+                 coord_frame=FIFF.FIFFV_COORD_MRI)
+            for rr in rrs])
+        del rrs
+        keep_all = False
+    else:
+        keep_all = True
+    # ensure it's a usable one
+    klass = dict(
+        surface=SourceEstimate,
+        volume=VolSourceEstimate,
+        mixed=MixedSourceEstimate,
+    )
+    _check_option('src.kind', src.kind, sorted(klass.keys()))
+    klass = klass[src.kind]
+    rrs = np.concatenate([s['rr'][s['vertno']] for s in src])
+    if src[0]['coord_frame'] == FIFF.FIFFV_COORD_HEAD:
+        rrs = apply_trans(trans, rrs)
     # projection will only occur with surfaces
-    if project and src is None:
-        logger.info(
-            f'Projecting {len(pos)} sensors onto {len(rrs)} '
-            f'vertices: {mode} mode')
+    logger.info(
+        f'Projecting data from {len(pos)} sensor{_pl(pos)} onto {len(rrs)} '
+        f'{src.kind} vertices: {mode} mode')
+    if project and src.kind == 'surface':
         logger.info('    Projecting electrodes onto surface')
         pos = _project_onto_surface(pos, dict(rr=rrs), project_rrs=True,
                                     method='nearest')[2]
 
-    # project source activity up to the minimum
-    # distance between our channel coordinates
-    # measured in 'mm'. Note mne-python stores 'meters'
     min_dist = pdist(pos).min() * 1000
     logger.info(
-        f'    Projected sensors have minimum distance {min_dist:0.1f} mm')
+        f'    Minimum {"projected " if project else ""}intra-sensor distance: '
+        f'{min_dist:0.1f} mm')
 
-    # compute pairwise distance between brain mesh coordinates
-    # and the channel coordinates
+    # compute pairwise distance between source space points and sensors
     dists = cdist(rrs, pos)
     assert dists.shape == (len(rrs), len(pos))
 
@@ -3306,30 +3319,23 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         w[range_, idx] = vals
     missing = np.where(~np.any(w, axis=0))[0]
 
-    print('INHERE')
-    print('MINIMUM DISTANCE: ', min_dist)
-    print(distance)
-    print(dists.shape)
-    print(vertices.shape)
-    print(len(missing))
-    print(w.shape)
-    print(evoked.data.shape)
     if len(missing):
         warn(f'Channel{_pl(missing)} missing in STC: '
              f'{", ".join(evoked.ch_names[mi] for mi in missing)}')
 
-    data = w @ evoked.data
-    vertices = [vertices[vertices < offset],
-                vertices[vertices >= offset] - offset]
-
-    # either return a surface, or volume SourceEstimate
-    if src:
-        return VolSourceEstimate(
-            data, vertices=vertices, tmin=evoked.times[0],
-            tstep=1. / evoked.info['sfreq'],
-            subject=subject, verbose=verbose
-        )
+    nz_data = w @ evoked.data
+    if not keep_all:
+        assert src.kind == 'surface'
+        data = nz_data
+        offset = len(src[0]['vertno'])
+        vertices = [vertices[vertices < offset],
+                    vertices[vertices >= offset] - offset]
     else:
-        return SourceEstimate(
-            data, vertices, evoked.times[0], 1. / evoked.info['sfreq'],
-            subject=subject)
+        data = np.zeros(
+            (sum(len(s['vertno']) for s in src), len(evoked.times)),
+            dtype=nz_data.dtype)
+        data[vertices] = nz_data
+        vertices = [s['vertno'].copy() for s in src]
+
+    return klass(data, vertices, evoked.times[0], 1. / evoked.info['sfreq'],
+                 subject=subject, verbose=verbose)
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3210,12 +3210,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     stc : instance of SourceEstimate
         The surface source estimate. If src is None, a surface source
         estimate will be produced, and the number of vertices will equal
-         the number of pial-surface vertices that were close enough to
-         the sensors to take on a non-zero volue. If src is not None,
-         a surface, volume, or mixed source estimate will be produced
-         (depending on the kind of source space passed) and the
-         vertices will match those of src (i.e., there may be me
-         many all-zero values in stc.data).
+        the number of pial-surface vertices that were close enough to
+        the sensors to take on a non-zero volue. If src is not None,
+        a surface, volume, or mixed source estimate will be produced
+        (depending on the kind of source space passed) and the
+        vertices will match those of src (i.e., there may be me
+        many all-zero values in stc.data).
 
     Notes
     -----
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3222,6 +3222,7 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     For surface projections, this function projects the ECoG sensors to
     the pial surface (if ``project``), then the activation at each pial
     surface vertex is given by the mode:
+    
     - ``'sum'``
         Activation is the sum across each sensor weighted by the fractional
         ``distance`` from each sensor. A sensor with zero distance gets weight
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3259,11 +3259,9 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     # convert head positions -> coord_frame MRI
     pos = apply_trans(trans, pos)
 
-    # read surface files
     subject = _check_subject(None, subject, False)
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     if src is None:  # fake a full surface one
-        # surface coordinate points
         rrs = [read_surface(op.join(subjects_dir, subject,
                                     'surf', f'{hemi}.pial'))[0]
                for hemi in ('lh', 'rh')]
@@ -3317,7 +3315,6 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         w.fill(0)
         w[range_, idx] = vals
     missing = np.where(~np.any(w, axis=0))[0]
-
     if len(missing):
         warn(f'Channel{_pl(missing)} missing in STC: '
              f'{", ".join(evoked.ch_names[mi] for mi in missing)}')
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3239,11 +3239,9 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     _validate_type(mode, str, 'mode')
     _check_option('mode', mode, ('sum', 'single', 'nearest'))
 
-    # create a copy of Evoked using either ecog or seeg
-    if src is None:
-        evoked = evoked.copy().pick_types(meg=False, ecog=True)
-    else:
-        evoked = evoked.copy().pick_types(meg=False, seeg=True)
+    # create a copy of Evoked using ecog and seeg
+    evoked = evoked.copy().pick_types(meg=False, ecog=True,
+                                      seeg=True)
 
     # get channel positions that will be used to pinpoint where
     # in the Source space we will use the evoked data
@@ -3253,10 +3251,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     # coord_frame (e.g. 'mni_tal', 'mri', 'fs_tal', etc.)
     trans, _ = _get_trans(trans, 'head', coord_frame, allow_none=True)
 
+    # convert head positions -> native coord_frame MRI
+    pos = apply_trans(trans, pos)
+
     # read surface files
     subject = _check_subject(None, subject, False)
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
-    data, vertices = list(), list()
     rrs = [read_surface(op.join(subjects_dir, subject,
                                 'surf', f'{hemi}.pial'))[0]
            for hemi in ('lh', 'rh')]
@@ -3264,8 +3264,6 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     rrs = np.concatenate(rrs)
     rrs /= 1000.
 
-    # convert head positions -> native coord_frame MRI
-    pos = apply_trans(trans, pos)
     logger.info(
         f'Projecting {len(pos)} sensors onto {len(rrs)} vertices: {mode} mode')
     if project:
@@ -3294,6 +3292,8 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     data = w @ evoked.data
     vertices = [vertices[vertices < offset],
                 vertices[vertices >= offset] - offset]
+
+    # either return a surface, or volume SourceEstimate
     if src:
         return VolSourceEstimate(
             data, vertices=vertices, tmin=evoked.times[0], tstep=1. / evoked.info['sfreq'],
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3248,6 +3248,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     # in the Source space we will use the evoked data
     pos = evoked._get_channel_positions()
 
+    # remove nan channels
+    nan_inds = np.where(np.isnan(pos).any(axis=1))[0]
+    nan_chs = [evoked.ch_names[idx] for idx in nan_inds]
+    evoked.drop_channels(nan_chs)
+    pos = [pos[idx] for idx in range(len(pos)) if idx not in nan_inds]
+
     # coord_frame transformation from native mne "head" to MRI coord_frame
     trans, _ = _get_trans(trans, 'head', coord_frame, allow_none=True)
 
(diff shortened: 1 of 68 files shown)


mne/source_estimate.py
```diff
@@ -3196,6 +3196,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         If True, project the electrodes to the nearest ``'pial`` surface
         vertex before computing distances.
     %(subjects_dir)s
+    src : instance of SourceSpaces | instance of SourceMorph
+        The source space. Can also be a SourceMorph to morph the STC to
+        a new subject (see Examples).
```
larsoner, line 3201: I would not allow `SourceMorph` here, just `src`. People can morph ahead of time if they want.

mne/source_estimate.py
```diff
@@ -3231,9 +3237,18 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
-    pos = np.array([ch['loc'][:3] for ch in evoked.info['chs']])
+
+    # create a copy of Evoked using either ecog or seeg
+    if src is None:
+        evoked = evoked.copy().pick_types(meg=False, ecog=True)
+    else:
+        evoked = evoked.copy().pick_types(meg=False, seeg=True)
+    # get channel positions and coord_frame transformation
+    # XXX: how to allow MNI?
+    pos = evoked._get_channel_positions()
```
larsoner, line 3248: The channel positions should always be in the head coordinate frame. Then we use `trans` to get them to the MRI coordinate frame.

People with coordinates in MNI coords should just use `fsaverage` as their subject since fsaverage is in MNI space (i.e., MNI==fsaverage's MRI coordinate frame)

mne/source_estimate.py
```diff
@@ -3243,6 +3258,8 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     offset = len(rrs[0])
     rrs = np.concatenate(rrs)
     rrs /= 1000.
+
+    # convert head positions -> native coord_frame (e.g. mni_tal, or mri)
```
larsoner, line 3262: No, always `MRI`

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,137 @@
+"""
+.. _tut_working_with_seeg:
+
+======================
+Working with SEEG data
```
larsoner, line 5:
```suggestion
Working with sEEG data
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,137 @@
...
+======================
+
+MNE supports working with more than just MEG and EEG data. Here we show some
+of the functions that can be used to facilitate working with
+stereoelectroencephalography (SEEG) data.
```
larsoner, line 10:
```suggestion
stereoelectroencephalography (sEEG) data.
```

mne/source_estimate.py
```diff
@@ -3176,7 +3176,7 @@ def extract_label_time_course(stcs, labels, src, mode='auto',
 
 @verbose
 def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
-                     project=True, subjects_dir=None, verbose=None):
+                     project=True, subjects_dir=None, src=None, verbose=None):
     """Create a STC from ECoG sensor data.
```
larsoner, line 3180: ECoG and sEEG

mne/source_estimate.py
```diff
@@ -3231,9 +3237,18 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
     _validate_type(mode, str, 'mode')
     _check_option('mode', mode, ('sum', 'single', 'nearest'))
-    evoked = evoked.copy().pick_types(meg=False, ecog=True)
-    pos = np.array([ch['loc'][:3] for ch in evoked.info['chs']])
+
+    # create a copy of Evoked using either ecog or seeg
+    if src is None:
+        evoked = evoked.copy().pick_types(meg=False, ecog=True)
+    else:
+        evoked = evoked.copy().pick_types(meg=False, seeg=True)
```
larsoner, line 3245:
```suggestion
    evoked = evoked.copy().pick_types(ecog=True, seeg=True)
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,137 @@
...
+import mne
+from mne.viz import plot_alignment, snapshot_brain_montage
+
+print(__doc__)
+
+# paths to mne datasets - sample SEEG and FreeSurfer subject
+misc_path = mne.datasets.misc.data_path()
+sample_path = mne.datasets.sample.data_path()
+subject = 'sample'
+subjects_dir = sample_path + '/subjects'
```
larsoner, line 31: For this example, do not use `sample`. Use `fsaverage`. That way the ECoG example is "ECoG data in a subject's space projected to a surface" and the sEEG example is "sEEG data in MNI space projected to a volume". And people who have ECoG/sEEG x subject/MNI x surf/vol can pick the parts of the tutorials they need. And the narrative docs of the two examples should begin by stating this very clearly, e.g., this should have something like:
```
This example shows how to use:

- sEEG data
- channel locations in MNI space
- projection into a volume

For an example that involves ECoG data, channel locations in a subject-specific MRI, or projection into a surface, see :ref:`tut-working-with-ecog`.
```
And then `plot_ecog.py` can be updated with a similar list+ref.

larsoner (commented):
Looks like a good start!

adam2392:
I keep on getting this error, and I'm not sure how to trace this down. It seems that there is no data getting attached to `stc.data`? 

```
  File "/Users/adam2392/Documents/mne-python/tutorials/misc/plot_seeg.py", line 142, in <module>
    brain = stc.plot(src=src, mode='stat_map', initial_time=0.68,
  File "/Users/adam2392/Documents/mne-python/mne/source_estimate.py", line 1943, in plot
    return plot_volume_source_estimates(
  File "<decorator-gen-140>", line 24, in plot_volume_source_estimates
  File "/Users/adam2392/Documents/mne-python/mne/viz/_3d.py", line 2104, in plot_volume_source_estimates
    img = stc.as_volume(src, mri_resolution=False)
  File "/Users/adam2392/Documents/mne-python/mne/source_estimate.py", line 2106, in as_volume
    return _interpolate_data(data, src, mri_resolution=mri_resolution,
  File "/Users/adam2392/Documents/mne-python/mne/morph.py", line 823, in _interpolate_data
    vols[this_inuse] = stc.data[stc_slice]
ValueError: shape mismatch: value array of shape (0,2001) could not be broadcast to indexing result of shape (24303,2001)
```

mne/source_estimate.py
```diff
@@ -3231,19 +3238,32 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
+    evoked = evoked.copy().pick_types(meg=False, ecog=True,
+                                      seeg=True)
+
+    # get channel positions that will be used to pinpoint where
+    # in the Source space we will use the evoked data
+    pos = evoked._get_channel_positions()
+
+    # coord_frame transformation from native mne "head" to desired coord_frame
+    # coord_frame (e.g. 'mni_tal', 'mri', 'fs_tal', etc.)
+    trans, _ = _get_trans(trans, 'head', coord_frame, allow_none=True)
```
larsoner, line 3252:
```suggestion
    # coord_frame transformation from native mne "head" to MRI coord_frame
    trans, _ = _get_trans(trans, 'head', coord_frame, allow_none=True)
```

mne/source_estimate.py
```diff
@@ -3231,19 +3238,32 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
+
+    # convert head positions -> native coord_frame MRI
+    pos = apply_trans(trans, pos)
```
larsoner, line 3255:
```suggestion
    # convert head positions -> coord_frame MRI
    pos = apply_trans(trans, pos)
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+of the functions that can be used to facilitate working with
+stereoelectroencephalography (sEEG) data.
+
+This example shows how to use:
+
+- sEEG data
+- channel locations in MNI space
+- projection into a volume
+
+For an example that involves ECoG data, channel locations in a subject-specific MRI, or projection into a surface, see :ref:`tut-working-with-ecog`.
```
larsoner, line 18: PEP8
```suggestion
For an example that involves ECoG data, channel locations in a
subject-specific MRI, or projection into a surface, see
:ref:`tut-working-with-ecog`.
```

tutorials/misc/plot_ecog.py
```diff
@@ -8,6 +8,14 @@
...
 of the functions that can be used to facilitate working with
 electrocorticography (ECoG) data.
+
+This example shows how to use:
+
+- ECoG data
+- channel locations in subject's MRI space
+- projection onto a surface
+
+For an example that involves sEEG data, channel locations in a template-specific MRI, or projection into a volume, see :ref:`tut-working-with-seeg`.
```
larsoner, line 18:
```suggestion
For an example that involves sEEG data, channel locations in
MNI space, or projection into a volume, see :ref:`tut-working-with-seeg`.
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+# we can plot frequency band power on top of it.
+#
+# .. note:: These are not real electrodes for this subject, so they
+#           do not align to the cortical surface perfectly.
+
+# fig = plot_alignment(raw.info, subject=subject, subjects_dir=subjects_dir,
+#                      surfaces=['pial'], trans=trans, coord_frame=coord_frame)
+# mne.viz.set_3d_view(fig, 200, 70, focalpoint=[0, -0.005, 0.03])
+#
+# xy, im = snapshot_brain_montage(fig, montage)
```
larsoner, line 109: Delete, and mention this difference at the top

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+gamma_power_t = raw.copy().filter(30, 90).apply_hilbert(
+    envelope=True).get_data()
+alpha_power_t = raw.copy().filter(8, 12).apply_hilbert(
+    envelope=True).get_data()
+gamma_power = gamma_power_t.mean(axis=-1)
+alpha_power = alpha_power_t.mean(axis=-1)
+
+# choose a colormap range wide enough for both frequency bands
+_gamma_alpha_power = np.concatenate((gamma_power, alpha_power)).flatten()
+vmin, vmax = np.percentile(_gamma_alpha_power, [10, 90])
```
larsoner, line 124: Maybe we should just plot raw traces in this example?

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+
+###############################################################################
+# Alternatively, we can project the sensor data to the nearest locations on
+# the pial surface and visualize that:
+
+# sphinx_gallery_thumbnail_number = 4
+
+# setup a volume-based source space here
+vol_src = mne.setup_volume_source_space(subject, subjects_dir=subjects_dir,
+                                        pos=5., mri='aseg.mgz')
```
larsoner, line 134: you should not need to do this for fsaverage. Instead do:
```suggestion
# get standard fsaverage volume source space
fetch_fsaverage(subjects_dir=subjects_dir)  # downloads it if necessary
fname_src = op.join(subjects_dir, 'fsaverage', 'bem',
                   'fsaverage-vol-5-src.fif')
vol_src = mne.read_source_spaces(fname_src)
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+ch_pos = dict(zip(ch_names, ch_coords))
+# Ideally the nasion/LPA/RPA will also be present from the digitization, here
+# we use fiducials estimated from the subject's FreeSurfer MNI transformation:
+lpa, nasion, rpa = mne.coreg.get_mni_fiducials(
+    subject, subjects_dir=subjects_dir)
+lpa, nasion, rpa = lpa['r'], nasion['r'], rpa['r']
+
+###############################################################################
+# Now we make a :class:`mne.channels.DigMontage` stating that the sEEG
+# contacts are in the FreeSurfer surface RAS (i.e., MRI) coordinate system.
```
larsoner, line 61:
```suggestion
# contacts are in the FreeSurfer surface RAS (i.e., MRI) coordinate system
# for the given subject. Keep in mind that ``fsaverage`` is special in that
# it is already in MNI space.
```

mne/source_estimate.py
```diff
@@ -3196,6 +3197,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         If True, project the electrodes to the nearest ``'pial`` surface
         vertex before computing distances.
```
larsoner, line 3198:
```suggestion
        vertex before computing distances. Only used when doing a
        surface projection.
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,149 @@
...
+
+# sphinx_gallery_thumbnail_number = 4
+
+# setup a volume-based source space here
+vol_src = mne.setup_volume_source_space(subject, subjects_dir=subjects_dir,
+                                        pos=5., mri='aseg.mgz')
+
+evoked = mne.EvokedArray(gamma_power_t, raw.info)
+stc = mne.stc_near_sensors(evoked, trans, subject, subjects_dir=subjects_dir,
+                           src=vol_src, project=False, mode='nearest')
```
larsoner, line 138:
```suggestion
                           src=vol_src, mode='nearest')
```

larsoner (commented):
Would it help if I pulled your changes and fixed the problem you're having with not having any data?

larsoner:
> Do you think this is a problem with the seeg coordinates, or possibly something with the stc_near_sensors function?

I suspect that your `evoked.data.shape[1] == 0` so your `stc.data.shape[1] == 0` and the volume morphing and stc plotting functions don't handle this case properly.

adam2392:
Okay a few updates:

1. it turned out that channel coords for the `misc-data` was in 'mm', so needed to divide by 1000 before setting to montage. 
2. In addition, if channels have `nan` coordinates, we need to remove to make sure `pdist(pos)` doesn't return nan. 
3. I added a bunch of inline comments to help navigate a bit better.
4. I added a bunch of print statements to debug, but I can remove them once I figure out why the following error is occuring.

```
  File "/Users/adam2392/Documents/mne-python/tutorials/misc/plot_seeg.py", line 129, in <module>
    brain = stc.plot(src=vol_src, mode='stat_map', initial_time=0.68,
  File "/Users/adam2392/Documents/mne-python/mne/source_estimate.py", line 1943, in plot
    return plot_volume_source_estimates(
  File "<decorator-gen-140>", line 24, in plot_volume_source_estimates
  File "/Users/adam2392/Documents/mne-python/mne/viz/_3d.py", line 2104, in plot_volume_source_estimates
    img = stc.as_volume(src, mri_resolution=False)
  File "/Users/adam2392/Documents/mne-python/mne/source_estimate.py", line 2106, in as_volume
    return _interpolate_data(data, src, mri_resolution=mri_resolution,
  File "/Users/adam2392/Documents/mne-python/mne/morph.py", line 823, in _interpolate_data
    vols[this_inuse] = stc.data[stc_slice]
ValueError: shape mismatch: value array of shape (1638,2001) could not be broadcast to indexing result of shape (14629,2001)
```

It's trying to get 14629 vertices, which I'm guessing is coming from the `vol_src` for `fsaverage`, but I don't see anyway of limiting the amount of vertices there. In `stc_near_sensors`, we only grab 1638 vertices within our "epsilon-ball" parametrized by `distance=0.01`? Any idea how I should proceed? I don't think it seems to be a bug anymore on my end...

tutorials/misc/plot_seeg.py
adam2392: Okay took this part out, and just added a `raw.plot()` to show that plotting SEEG data raw.

adam2392:
Awesome @larsoner I just saw the recent changes and it works now. Thanks for the help! It looks very cool.

Regarding what is then plotted later, I see that it is kind of sparse but focused in the left frontal lobe region... I wonder if the example should talk a bit about it and what's the main orange region vs the blips that might appear? Even for me working on this example, I'm not sure what is being visualized per say without trying to go thru the docs. I imagine for a newcomer that would be useful too and then they would understand how to visualize a different set of contacts, or regions.

larsoner:
> In stc_near_sensors, we only grab 1638 vertices within our "epsilon-ball" parametrized by distance=0.01? Any idea how I should proceed? I don't think it seems to be a bug anymore on my end...

Basically if `src` is passed, the output should always have `stc.shape == (n_vertices, n_times)` where `n_vertices = sum(len(s['vertno']) for s in src)`, i.e., all source points should be there. It's just a lot of them will be zero.

I took a stab at simplifying / unifying the code. Feel free to take a look and see if it makes sense. This is the rendered tutorial now:

https://22913-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html

Locally the `plot_alignment` plot would not look correct because your branch did not have #8393, so I also did a `merge` for your branch with `upstream/master`. Locally if you just do `git pull` you should get all my changes + this merge and be able to continue making changes.

larsoner:
> I see that it is kind of sparse but focused in the left frontal lobe region...

Take a look at the `plot_alignment` plot locally. This is the CircleCI version:

sphx_glr_plot_seeg_001

If you look at it locally you can see that actually a lot of points are outside the scalp surface. Not sure why this would be the case if your points are in MNI space...

adam2392:
Just took a look and those coordinates are in the wrong space. I updated the misc-data repo:

https://github.com/mne-tools/mne-misc-data/pull/4

adam2392:
New ones rendered locally look accurate according to the test subject I was using to update misc-data.

Screen Shot 2020-10-23 at 10 43 57 AM
Screen Shot 2020-10-23 at 10 44 25 AM
Screen Shot 2020-10-23 at 10 44 35 AM

BarryLiu-97:
If you guys need sEEG data and its's electrodes MNI coordinates, I can offer one sample in .edf but it is too big. The least one is about 300MB.

BarryLiu-97:
I can cut it then export it in .set format or you know, .fif format.

adam2392:
> If you guys need sEEG data and its's electrodes MNI coordinates, I can offer one sample in .edf but it is too big. The smallest one is about 300MB.

Any chance you can crop it to like 3-5 second so it's essentially a few MB at most? and then make a PR into the `seeg/` folder here: https://github.com/mne-tools/mne-misc-data?

We can have multiple SEEG example files then perhaps? @larsoner would that be valuable?

larsoner:
> We can have multiple SEEG example files then perhaps? @larsoner would that be valuable?

I'm not sure what the added value would be. Let's finish this example first, then see if there is something missing that we cannot cover adequately with this dataset and the ECoG dataset.

BarryLiu-97:
> Any chance you can crop it to like 3-5 second so it's essentially a few MB at most? and then make a PR into the `seeg/` folder here: https://github.com/mne-tools/mne-misc-data?
Yeah, of course. Although I don't know how to make a PR into the `seeg/` folder, I'll give it a try. While as to sEEG data, some of its channels may not be valuable in analysis, for the reference electrode and the DC channel. And in my sample data, some of these DC channels are used to mark data. You want the original data directly from the machine, or I delete some useless channels and then mark the data.

BarryLiu-97:
> > Any chance you can crop it to like 3-5 second so it's essentially a few MB at most? and then make a PR into the seeg/ folder here: https://github.com/mne-tools/mne-misc-data?
Yeah, of course. Although I don't know how to make a PR into the seeg/ folder, I'll give it a try. While as to sEEG data, some of its channels may not be valuable in analysis, for the reference electrode and the DC channel. And in my sample data, some of these DC channels are used to mark data. You want the original data directly from the machine, or I delete some useless channels and then mark the data.

BarryLiu-97:
> Any chance you can crop it to like 3-5 second so it's essentially a few MB at most? and then make a PR into the `seeg/` folder here: https://github.com/mne-tools/mne-misc-data?

Yeah, of course. But as to sEEG data, some of its' electrodes are useless in analysis, like ref electrode, or EKG channels, DC channels. You want the data directly from the machine or I delete the useless channels and set the marker then give you?

BarryLiu-97:
OK, if you guys need it, just tell me.

adam2392:
Okay the new example looks good after applying the `talairach.xfm`, which I've stuck in the PR in `mne-misc-data`. Once that gets merged in and we update the md5sum, then the docs can be rendered here.

larsoner:
@adam2392 now you need to update `mne/datasets/utils.py` to use version 0.7 and hash 2b2f2fec9d1197ed459117db1c6341ee for MISC

mne/tests/test_source_estimate.py
```diff
@@ -1538,16 +1538,3 @@ def test_stc_near_sensors(tmpdir):
...
-
-    # and now with volumetric projection
-    src = read_source_spaces(fname_vsrc)
-    with catch_logging() as log:
-        stc_vol = stc_near_sensors(evoked, trans, 'sample', src=src,
-                                   subjects_dir=subjects_dir, verbose=True,
-                                   distance=0.033)
-    assert isinstance(stc_vol, VolSourceEstimate)
-    log = log.getvalue()
-    assert '4157 volume vertices' in log
```
larsoner: Looks like you killed the volumetric test I added?

adam2392: oops sorry merging gone wrong.

mne/tests/test_source_estimate.py
```diff
@@ -1538,3 +1538,16 @@ def test_stc_near_sensors(tmpdir):
...
+    with catch_logging() as log:
+        stc_vol = stc_near_sensors(evoked, trans, 'sample', src=src,
+                                   subjects_dir=subjects_dir, verbose=True,
+                                   distance=0.033)
+    assert isinstance(stc_vol, VolSourceEstimate)
+    log = log.getvalue()
+    assert '4157 volume vertices' in log
+
+
+run_tests_if_main()
```
larsoner, line 1553: This part I actually removed

larsoner: (This line and the one above it I mean)

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,162 @@
...
+    verbose='error')  # ignore missing electrode warnings
+stc = abs(stc)  # just look at magnitude
+clim = dict(kind='value', lims=np.percentile(abs(evoked.data), [10, 50, 75]))
+
+# XXX: what is the difference between plot and plot_3d? They look visually
+# pretty similar.
+# plot Nutmeg style
+stc.plot(vol_src, subject=subject, subjects_dir=subjects_dir, clim=clim)
+
+# plot 3D source (brain region) visualization
```
larsoner, line 152: Create a separate section in the rendered doc:
```suggestion
###############################################################################
# Plot 3D source (brain region) visualization:
```

tutorials/misc/plot_seeg.py
```diff
@@ -0,0 +1,162 @@
...
+evoked = mne.EvokedArray(raw.get_data(), raw.info)
+stc = mne.stc_near_sensors(
+    evoked, trans, subject, subjects_dir=subjects_dir, src=vol_src,
+    verbose='error')  # ignore missing electrode warnings
+stc = abs(stc)  # just look at magnitude
+clim = dict(kind='value', lims=np.percentile(abs(evoked.data), [10, 50, 75]))
+
+# XXX: what is the difference between plot and plot_3d? They look visually
+# pretty similar.
+# plot Nutmeg style
```
larsoner, line 149:
```suggestion
# plot Nutmeg style
```

larsoner (commented):
Anything else left to do from your end @adam2392 or are you happy with this now?

larsoner:
Look like the `fetch_fsaverage` needs to be called at the top, see the error:

https://22951-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html

It's not great that we got `CircleCI received exit code 0` from that build step...

pushed a commit to fix these

adam2392:
> Anything else left to do from your end @adam2392 or are you happy with this now?

No this looks really good to me so far and honestly really useful for someone in the iEEG world to use mne-python in conjunction w/ freesurfer + neuroimaging viz.

I updated the docstring of `stc_near_sensors`. lmk wyt?

mne/source_estimate.py
```diff
@@ -3208,7 +3208,11 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     Returns
     -------
     stc : instance of SourceEstimate
-        The surface source estimate.
+        The surface source estimate. If using volume ``src``, ``stc``
+        will have the same number of vertices as the passed in
+        SourceSpace object. If using surface, then ``stc`` will
+        have the number of vertices on the ``pial`` surface that
+        get activated.
```
larsoner, line 3215: This is actually not quite accurate.

If src is None, a surface source estimate will be produced, and the number of vertices will equal the number of pial-surface vertices that were close enough to the sensors to take on a non-zero volue.

If src is not None, a surface, volume, or mixed source estimate will be produced (depending on the kind of source space passed) and the vertices will match those of ``src`` (i.e., there may be me many all-zero values in ``stc.data``).

adam2392: Okay that makes more sense. I'll add that in.

adam2392:
```
/home/circleci/project/mne/source_estimate.py:docstring of mne.stc_near_sensors:55: WARNING: Unexpected indentation.
```

I think this is occurring in the addition I made under `Notes`. I think(?) I fixed it?

mne/source_estimate.py
```diff
@@ -3194,21 +3195,33 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
 
     Notes
     -----
-    This function projects the ECoG sensors to the pial surface (if
-    ``project``), then the activation at each pial surface vertex is given
-    by the mode:
-
+    For surface projections, this function projects the ECoG sensors to
+    the pial surface (if ``project``), then the activation at each pial
+    surface vertex is given by the mode:
```
larsoner, line 3224: You probably need to restore the newline here for rst to render properly. The error you saw was because of an extra space indentation that I pushed a fix for

mne/source_estimate.py
```diff
@@ -3222,7 +3222,6 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     For surface projections, this function projects the ECoG sensors to
     the pial surface (if ``project``), then the activation at each pial
     surface vertex is given by the mode:
-
```
larsoner, line 3211: In other words, this is probably going to break things

adam2392: Fixed.

adam2392:
Another small thing: do you happen to know how one would do this if you don't have BEM? My understanding is you only have BEM from flash MRI data. I've never worked w/ BEM and it never gets output from my freesurfer `recon-all`.

```
# get standard fsaverage volume (5mm grid) source space
fname_src = op.join(subjects_dir, 'fsaverage', 'bem',
                    'fsaverage-vol-5-src.fif')
vol_src = mne.read_source_spaces(fname_src)
```

E.g. what if the above someone wanted to just use the parcellated MRI? Is that possible?

larsoner (approved):
LGTM +1 for merge

Let's see if @drammock and/or @agramfort are happy. Here is the rendered example:

https://22968-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html

Just pushed a tiny commit to make the thumbnail the rendered movie.

larsoner:
> Another small thing: do you happen to know how one would do this if you don't have BEM? My understanding is you only have BEM from flash MRI data. I've never worked w/ BEM and it never gets output from my freesurfer recon-all.

This is another thing that we should make clearer somewhere in the docs. You only need a BEM if you want to do some forward modeling.

What you really probably want is to constrain your volumetric sources to be within the brain volume. One way to do this is by passing a BEM, but really this just uses the `inner_skull` surface. Instead of doing this, you can use the `brain` surface, which you can extract from the `brainmask.mgz` file using a few FreeSurfer commands -- this is done in the process of `mne watershed_bem`, so this is a convenient way to get this surface. This does not require FLASH scans -- it uses only the T1.mgz image.

Alternatively, you can tell it to use `aseg.mgz` or `aseg+aparc.mgz` with `single_volume=True`, which will only include volumetric sources that are in the brain volume / `brainmask.mgz`. Maybe open a separate issue about how to create and work with volumetric source spaces so that we can discuss this stuff there?

agramfort (approved):

BarryLiu-97:
Hi, is there any way that we can set different parts of depth electrodes into different colors, for example, which may be useful, if the contacts of this electrode is out of the skull, it's green, otherwise yellow.

larsoner:
> Hi, is there any way that we can set different parts of depth electrodes into different colors, for example, which may be useful, if the contacts of this electrode is out of the skull, it's green, otherwise yellow.

This is really an issue for / part of #8382

larsoner:
Thanks for working through this @adam2392 !

BarryLiu-97:
Hi, recently I've been thinking that showing some info about the SEEG electrodes may be very useful when you move your mouse to the electrode in the figure produced by plot_alignment, for example, the name, the MNI coordinates, and which brain area the electrode is in. I have figured out how
 to calculate the brain area of each electrode, but I don't know how to add it to this.

BarryLiu-97:
image

BarryLiu-97:
This is the info of the electrodes I calculated using python.
image

drammock:
@BarryLiu-97 please open a new issue for new feature requests. This PR is already merged and closed.

BarryLiu-97:
> please open a new issue for new feature requests. This PR is already merged and closed.

Got it!

PR #8405, opened 2020-10-22
FIX: focalpoint

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/8379#issuecomment-714558278 and fixes at least `plot_3d_to_2d.py`

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1085,7 +1085,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         renderer.tube(origin=origin, destination=destination)
 
     renderer.set_camera(azimuth=90, elevation=90,
-                        distance=0.6, focalpoint=(0., 0., 0.))
+                        distance=0.6)
     renderer.show()
     return renderer.scene()
 
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1085,7 +1085,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         renderer.tube(origin=origin, destination=destination)
 
     renderer.set_camera(azimuth=90, elevation=90,
-                        distance=0.6)
+                        distance=0.6, focalpoint=(0., 0., 0.))
     renderer.show()
     return renderer.scene()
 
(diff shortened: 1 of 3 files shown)


mne/viz/_3d.py
```diff
@@ -1085,7 +1085,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         renderer.tube(origin=origin, destination=destination)
 
     renderer.set_camera(azimuth=90, elevation=90,
-                        distance=0.6, focalpoint=(0., 0., 0.))
+                        distance=0.6)
```
larsoner, line 1088: The real question is what should the default actually be? I think (0., 0., 0.) is a reasonable choice for us. In all 3D plotting cases I can think of, (0., 0., 0.) is meaningful (center of some coordinate frame) so I would leave this as the default, and then just set it explicitly to `focalpoint=None` or `focalpoint=(...)` in places where we want it to be something different.

GuillaumeFavelier:
The [default values](https://github.com/mne-tools/mne-python/blob/58d13fc833f382b5f41b28611fd2a128b75b37b0/mne/viz/_3d.py#L3000) for `examples/connectivity/plot_sensor_connectivity.py` make sense so no need to change anything here:

master | focalpoint=None
--|--
image | image

mne/viz/_3d.py
GuillaumeFavelier: We want a fixed value for `plot_alignment`? The default uses the bounds for now.

larsoner: Yes for `plot_alignment` in particular it seems reasonable to use `(0., 0., 0.)` because it's the alignment in the given `coord_frame`

GuillaumeFavelier:
The artifacts for the fixed example are available here: https://22876-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_3d_to_2d.html#sphx-glr-auto-examples-visualization-plot-3d-to-2d-py

Also, `plot_alignment`'s focal point is on `(0, 0, 0)`.

larsoner:
Thanks @GuillaumeFavelier

PR #8412, opened 2020-10-23
fix: do not log about re-referencing when no such thing is done

sappelhoff:
This came up on gitter: https://gitter.im/mne-tools/mne-python?at=5f92ad30c990bb1c39277c8d

When users add a previously unknown reference channel using `add_reference_channels`, the channel is initialized as a vector of zeroes. No actual re-referencing takes place, however the log messages confusingly print: 

> EEG channel type selected for re-referencing Applying a custom EEG reference

This PR suppresses that confusing log message.


larsoner (approved):

sappelhoff:
Travis is really the odd one out in terms of speed :)

larsoner:
Travis seems to be totally broken today. We'll move on without it, we can trust the coverage of Azure. Thanks @sappelhoff

PR #8420, opened 2020-10-25
ENH, DOC: Add section to publication plot example for sensor montage insets

sappelhoff:
Came up in Gitter: [October 23, 2020 8:29 PM](https://gitter.im/mne-tools/mne-python?at=5f9320ecea6bfb0a9a53aaad)


drammock:
@sappelhoff we already have this: https://mne.tools/dev/auto_examples/visualization/plot_publication_figure.html#sphx-glr-auto-examples-visualization-plot-publication-figure-py

It seems to me like it would be better to modify / expand that rather than having two separate examples/tutorials with similar goals.

sappelhoff:
>  I see you're aware of the other example, since you mention it in yours. Is there a reason you think these should be separate documents? 

mostly because @larsoner proposed it in his comment on gitter that I linked to in my OP.

I am happy to adjust my example (merge it with the other). Let's hear Eric's reasoning first.

larsoner:
Either way is fine by me, happy to defer to @drammock's judgment on this

sappelhoff:
I tried adding the inset plot to the existing "publication plot" example, but I don't think it's very well suited for this case, because there are >200 channels and I feel like it overloads the figure.

I'll try to go @drammock's proposed route 2) now, and add a new section to the publication_plot example.

sappelhoff:
@drammock any clue how I can make my plots show up where I inserted the `plt.show()`?

artifact: [link](https://23072-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_publication_figure.html#sphx-glr-auto-examples-visualization-plot-publication-figure-py)

Screenshot: 
image

larsoner:
Plots are always scraped and closed at the end of every code block. So end the code block after the `plt.show`, and in the next code block, where you need the plot again, repeat the same plot call(s)

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+
+# plot some fake EEG data
+ax.plot(times, mean)
+ax.set(xlabel="Time (s)", ylabel="Amplitude (µV)")
+
+###############################################################################
+# So far so good. Now let's add the smaller figure within the figure to show
+# exactly, which sensors we used to make the timecourse.
+
+# For that, we use an "inset_axes" that we plot into our existing axes
```
drammock, line 236:
```suggestion
# For that, we use an "inset_axes" that we plot into our existing axes

# recreate the figure (only necessary for our documentation server)
fig, ax = plt.subplots(figsize=(4.5, 3))
ax.plot(times, mean)
ax.set(xlabel="Time (s)", ylabel="Amplitude (µV)")
```

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+axins = inset_locator.inset_axes(ax, width="30%", height="30%", loc=2)
+
+# The head outline with the sensor positions can be plotted using the
+# MNE :class:`mne.io.raw` object that is the origin of our data.
+# Specifically, that object already contains all the sensor positions,
+# and we can plot them using the ``plot_sensors`` method.
+raw.copy().pick_channels(to_plot).plot_sensors(
+    title="", axes=axins
+)
+plt.show()
```
drammock, line 246:
```suggestion
```

(shouldn't need this now)

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+# Scale the data from the MNE internal unit V to µV
+data *= 1e6
+
+# Take the mean of the channels
+mean = np.mean(data, axis=0)
+
+# make a figure
+fig, ax = plt.subplots(figsize=(4.5, 3))
+
+# plot some fake EEG data
```
drammock, line 228: I wouldn't say "fake" here

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+
+# That's quite a mess, but we know that we want to change the sensor dots,
+# and those are most certainly a "PathCollection" object.
+print(axins.collections)
+
+# Now we we found exactly what we needed. Sometimes this can take a bit of
+# experimentation.
+sensor_dots = axins.collections[0]
+
+# Let's shrink the sensor dots to finish our figure.
```
drammock, line 266:
```suggestion
# Recreate the figure once more, then shrink the sensor dots
# to finish our figure.
fig, ax = plt.subplots(figsize=(4.5, 3))
ax.plot(times, mean)
ax.set(xlabel="Time (s)", ylabel="Amplitude (µV)")
axins = inset_locator.inset_axes(ax, width="30%", height="30%", loc=2)
raw.pick_channels(to_plot).plot_sensors(title="", axes=axins)
```

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+sensor_dots.set_sizes([1])
+plt.show()
```
drammock, line 268:
```suggestion
```

examples/visualization/plot_publication_figure.py
```diff
@@ -165,3 +179,90 @@
...
+# exactly, which sensors we used to make the timecourse.
+
+# For that, we use an "inset_axes" that we plot into our existing axes
+axins = inset_locator.inset_axes(ax, width="30%", height="30%", loc=2)
+
+# The head outline with the sensor positions can be plotted using the
+# MNE :class:`mne.io.raw` object that is the origin of our data.
+# Specifically, that object already contains all the sensor positions,
+# and we can plot them using the ``plot_sensors`` method.
+raw.copy().pick_channels(to_plot).plot_sensors(
```
drammock, line 243: do we need the `.copy()` here?

examples/visualization/plot_publication_figure.py
sappelhoff:
```suggestion
# plot some EEG data
```

examples/visualization/plot_publication_figure.py
sappelhoff: because `.pick_channels` works in-place and would change our `raw`, which a user may not want

sappelhoff:
> Plots are always scraped and closed at the end of every code block. So end the code block after the plt.show, and in the next code block, where you need the plot again, repeat the same plot call(s)

I see, thanks. And thanks @drammock for the helpful suggestions!

sappelhoff:
I am happy with the example now: [artifact example link](https://23075-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_publication_figure.html#custom-timecourse-with-montage-inset)

except, for two things:

# the axis labels are clipped 


image


# `ax.set()` does not seem to print the axis labels in the later two plots :confused: 

Could this be because `plot_sensors` works with `ax.axis("off")`? However, locally the code works as expected and labels *are* being plotted.

image

larsoner:
@sappelhoff locally does running `PATTERN=publication make html_dev-pattern` in the `doc/` dir produce the same thing as CircleCI? If not, then it could be that CircleCI using latest `matplotlib` has exposed some regression or something we haven't taken into account from their release notes.

If it is broken on your system as well, then it's probably some bug with how Sphinx-Gallery scrapes the images

sappelhoff:
I get

```
Configuration error:
There is a programmable error in your configuration file:

  File "/home/stefanappelhoff/miniconda3/lib/python3.7/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/home/stefanappelhoff/miniconda3/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "/home/stefanappelhoff/Desktop/bids/mne-python/doc/conf.py", line 28, in <module>
    from mne.viz import Brain
ImportError: cannot import name 'Brain' from 'mne.viz' (/home/stefanappelhoff/miniconda3/lib/python3.7/site-packages/mne/viz/__init__.py)

```

I was running the example using `python -ui examples/visualization/plot_publication_figure.py` before

examples/visualization/plot_publication_figure.py
drammock: I agree that in an interactive session that's probably what one would do, and it's good for model code to reflect actual practice as much as possible. But for the doc-building server, memory usage can be an issue (it's more strict when building/deploying the full docs after merge, than it is for building in-progress PRs).  *Probably* this won't be an issue since you're already restricting to just EEG channels, but in general be aware when writing examples/tutorials to be conservative about memory.

In this case I'd be inclined to take one of two paths:

1. remove the copy and add a comment saying "pick_channels edits the raw object in place, so normally we'd make a copy first in case we needed to re-use the raw with all sensors intact in later code. Here, all we need it for is plotting these 5 sensors, so we won't bother"

2. leave the copy in, but explain why ("pick_channels edits the raw object in place, so we'll make a copy here so that our raw object stays intact for later analysis"). Although that's not *strictly* true for this tutorial, it's likely to be true in general.

Personally I lean toward # 2

larsoner:
@sappelhoff try rebasing on latest `master` first. It sounds like your branch is probably quite a bit behind.

That or when you do `python -c "import mne; print(mne)"` it does not point to the `..` directory, i.e., you do not have it installed in editable mode in your environment.

larsoner:
@sappelhoff there are potentially two issues.

I just saw your edit with sys_info. You did this from the `mne-python` directory, which has a `mne` directory within it, so it should *always* import that one as it's in the current directory. If you do this same command from `mne-python/doc`, which does not have a `mne/` subdirectory, you might get a different result, i.e., your doc build is not using the `mne-python` version because it's not installed properly in editable mode in your env.

However I doubt that's the problem. I think the problem is really that `which sphinx_build` will point to a `sphinx-build` in your `python3.7` directories somewhere. Doing `pip install --upgrade sphinx` (assuming your `pip` is 3.8) should make it so that `sphinx-build` properly points to a 3.8 sphinx install.

sappelhoff:
> However I doubt that's the problem. I think the problem is really that which sphinx_build will point to a sphinx-build in your python3.7 directories somewhere

nice catch! fixed now :-)

> If it is broken on your system as well, then it's probably some bug with how Sphinx-Gallery scrapes the images

yup, it's broken in my system as well :-/

examples/visualization/plot_publication_figure.py
sappelhoff: Thanks, I agree that it's good to better explain *why* a copy is being made.

sappelhoff:
@larsoner do you have an idea where I should start investigating?

drammock:
Personally I would start by removing mne, sphinx, and sphinx-gallery from your conda base environment, if that's acceptable given your setup/workflow. If they only exist in one place (in your mne environment), it should be easier to debug.

larsoner:
> yup, it's broken in my system as well :-/

> @larsoner do you have an idea where I should start investigating?

I would first try messing with `sphinx-gallery`'s scraping settings. One potential culprit is the force-setting of Agg backend:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/scrapers.py#L36

I would just short-circuit `return` that function before anything gets set. Then rerun the `PATTERN` line (be on latest `sphinx-gallery` master to ensure it runs it even though the example is stale, and inspect the output to ensure it actually ran).

Anther potential culprit is the line that actually scrapes the image:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/scrapers.py#L141-L149

But I doubt that's the problem.

Or you could take the alternative (probably easier) approach of doing `matplotlib.use('Agg', force=True)` at the top of the script and add a `fig.savefig('check.png'`)`, run it directlry with `python whatever.py`, and then look at the output PNG file.

sappelhoff:
the `matplotlib.use('Agg', force=True)` + `fig.savefig` approach works like a charm, so it's probably something else:

image

larsoner:
Do you have `matplotlibrc` or something similar on your system? Sphinx-Gallery calls `plt.rcdefaults()`, maybe that's the issue...?

sappelhoff:
> Do you have matplotlibrc or something similar on your system? Sphinx-Gallery calls plt.rcdefaults(), maybe that's the issue...?

I am now a bit confused by what's "the issue"

--> in the CI run with Circle, we see that plots are clipped and axis labels not plotted
--> when running locally in a clean mne-python install (with clean base conda env!) I see the same
--> however, when using `matplotlib.use('Agg', force=True)` + `fig.savefig` at the top and bottom of the example respectively, locally I do get a sound plot as expected (*unlike* in the CI, and when running without the savefig)

larsoner:
Or maybe `constrained_layout=True` could automagically fix things for us in this example at least?

sappelhoff:
> Or maybe constrained_layout=True could automagically fix things for us in this example at least?

That seems to provide a

> UserWarning: This figure was using constrained_layout==True, but that is incompatible with subplots_adjust and or tight_layout: setting constrained_layout==False.

sappelhoff:
> It can be undone easily enough with a final fig.subplots_adjust(bottom=0.2, left=0.2, top=0.9, right=0.9) or so

Thanks for investigating @drammock I added this fix to the example now.

> I notice that the call to plot_sensors is changing the figure margins for the containing subplot. It shouldn't do that.

agreed, that's weird and unexpected

sappelhoff:
The example is now fixed. The test failures appear to be due some (probably temporary) pip installation issue. Can probably be fixed by a re-run of the affected CIs.

drammock (approved):

drammock:
thanks for sticking with this one, @sappelhoff!

sappelhoff:
And thanks a lot for your help, both of you! I'm very happy that this helped to uncover another issue as well.

Issue #8421, opened 2020-10-25
No Annotations descriptions displayed in introductory tutorial figure


hoechenberger:
The Annotations in the `plot_20_events_from_raw` tutorial don't have a description (or vertical line) in the raw plot in[ `master`](https://mne.tools/dev/auto_tutorials/intro/plot_20_events_from_raw.html#converting-between-events-arrays-and-annotations-objects), but displayed correctly in [0.21](https://mne.tools/dev/auto_tutorials/intro/plot_20_events_from_raw.html#converting-between-events-arrays-and-annotations-objects)

`master`:

![](https://mne.tools/dev/_images/sphx_glr_plot_20_events_from_raw_002.png)

0.21:

![](https://mne.tools/stable/_images/sphx_glr_plot_20_events_from_raw_002.png)

cc @drammock

PR #8422, opened 2020-10-25
DOC: Better Annotations docs

hoechenberger:
Added some cross-references and we're now using `annotations_from_events` in the tutorial – a function which I only came across accidentally, as it wasn't mentioned anywhere in the docs.

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -64,6 +64,11 @@ def _check_o_d_s(onset, duration, description):
 class Annotations(object):
     """Annotation object for annotating segments of raw data.
 
+    .. note::
+       To convert events to `~mne.Annotations`, use
+       `~mne.annotations_from_events`. To convert existing `~mne.Annotations`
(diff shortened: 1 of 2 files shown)


agramfort (approved):
@drammock I let you merge if you're happy

drammock:
annotations are not showing up on the final plot in the tutorial. But that is not a result of this PR, it's probably a bug I introduced with the raw plotting overhaul. [stable](https://mne.tools/stable/_images/sphx_glr_plot_20_events_from_raw_002.png) [master & this PR](https://mne.tools/dev/_images/sphx_glr_plot_20_events_from_raw_002.png).

hoechenberger:
@drammock 
> annotations are not showing up on the final plot in the tutorial. But that is not a result of this PR, it's probably a bug I introduced with the raw plotting overhaul.

Yep, already opened an issue here: #8421

PR #8410, opened 2020-10-23
WIP: Use latest pyvista

GuillaumeFavelier:
This PR configures one Travis job to use the latest version of pyvista and pyvistaqt.

I'm not entirely sure about the syntax.

This is a work in progress

PR #8427, opened 2020-10-26
MNT: use brain_gc fixture in some tests

### LINKED ISSUE #8403: Use brain_gc in remaining tests
The new `brain_gc` fixture should be added in `test_plot_source_estimates`, `test_plot_sparse_source_estimates` and `test_notebook`.

Related to https://github.com/mne-tools/mne-python/pull/8379#issuecomment-714429361

GuillaumeFavelier:
This PR adds the `brain_gc` fixture in remaining 3d viz tests.

Closes #8403 

I ran the tests locally using master versions of `pyvista` and `pyvistaqt`. Maybe some adjustments are necessary.

### DIFF
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -318,9 +318,9 @@ def _check_skip_backend(name):
 @pytest.fixture()
 def renderer_notebook():
     """Verify that pytest_notebook is installed."""
-    from mne.viz.backends.renderer import _use_test_3d_backend
-    with _use_test_3d_backend('notebook'):
-        yield
+    from mne.viz.backends import renderer
+    with renderer._use_test_3d_backend('notebook'):
+        yield renderer
 
 
 @pytest.fixture(scope='session')
@@ -474,7 +474,7 @@ def download_is_error(monkeypatch):
 @pytest.fixture()
 def brain_gc(request):
     """Ensure that brain can be properly garbage collected."""
-    keys = ('renderer_interactive', 'renderer')
+    keys = ('renderer_interactive', 'renderer', 'renderer_notebook')
     assert set(request.fixturenames) & set(keys) != set()
     for key in keys:
         if key in request.fixturenames:
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -479,6 +479,7 @@ def brain_gc(request):
     for key in keys:
         if key in request.fixturenames:
             is_pv = request.getfixturevalue(key)._get_3d_backend() == 'pyvista'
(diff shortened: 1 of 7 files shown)


mne/viz/tests/test_3d.py
```diff
@@ -615,6 +616,7 @@ def test_plot_source_estimates(renderer_interactive, all_src_types_inv_evoked,
             these_kwargs = kwargs.copy()
             these_kwargs.pop('src')
             meth(**these_kwargs)
+    renderer_interactive.backend._close_all()
```
larsoner, line 619: Rather than do this, how about we just have both `renderer_interactive` and `brain_gc` call this after their `yield` steps? Calling it more than once should be fine, and having it in both means we shouldn't need to worry about order.

GuillaumeFavelier: Sure, I'll see if this solves the need to the intermediate calls too

GuillaumeFavelier:
There is indeed a difference between pinned and latest versions.

I'll come back to 908c1d2 when new versions are released. For now, some adjustments are necessary here.

GuillaumeFavelier:
> I'll see if this solves the need to the intermediate calls too

For me, it did the trick locally, thanks :+1: 

Actually there is another strategy: waiting for a new release of pyvista before merging this. Since it's not high priority, I think we can afford to wait.

larsoner:
I would 

1. combine this PR and #8410
2. make sure latest PyVista and PyVistaQt master are installed on CircleCI and the Travis (and Azure) pip pre runs
3. triage the `brain_gc` to only check if new enough `PyVista` and `PyVistaQt` are installed

There is a a slight problem, though, that PyVista is currently at 0.26.0 for some reason. I'll open a PR to fix this, it's weird.

mne/conftest.py
```diff
@@ -474,19 +474,25 @@ def download_is_error(monkeypatch):
...
     for key in keys:
         if key in request.fixturenames:
             is_pv = request.getfixturevalue(key)._get_3d_backend() == 'pyvista'
+            close_func = request.getfixturevalue(key).backend._close_all
             break
     if not is_pv:
         yield
         return
+    import pyvista
+    if LooseVersion(pyvista.__version__) <= LooseVersion('0.26.1'):
```
larsoner, line 488:
```suggestion
    if not check_version('pyvista', '0.26.1'):
```

larsoner: ... but things are green so let's just do this later. Thanks @GuillaumeFavelier !

PR #8435, opened 2020-10-27
[MRG] BUG: Do not allow baseline removal after preload

hoechenberger:

In `master` and 0.21 it's possible to preload `Epochs`, apply baseline correction, and (seemingly) remove it again via `apply_baseline(None)`. This should not be allowed: removing BL correction can only happen as long as the data has not been loaded.


agramfort (approved):

hoechenberger:
One test on Azure fails because the test environment cannot be set up correctly. All others pass, including Travis. All tests passed before merging the changelog from `master`. So this should be good to merge.

larsoner:
LGTM, feel free to merge and backport

agramfort:
@larsoner merge if you're ok with this one

larsoner:
Backported, thanks @hoechenberger

PR #8439, opened 2020-10-27
CI: Allow latest numpy

larsoner:
Let's see if we can remove our problematic version pinning. I can't remember why we needed it 2 months ago, but I'm sure the CIs will remind me (and I'll look back if they fail).


hoechenberger (approved):
All seems to pass, so I guess this is good to merge!

agramfort:
thx @larsoner

Issue #8216, opened 2020-09-05
apply_baseline then crop


virvw:
If your issue is a usage question, please consider asking on our [Gitter channel](https://gitter.im/mne-tools/mne-python) or on our [mailing list](https://mail.nmr.mgh.harvard.edu/mailman/listinfo/mne_analysis) instead of opening an issue.

#### Describe the problem
apply_baseline on epochs followed by a crop removes the former command apply_baseline
It is counter-intuitive that these two operations, in that specific order, would be dependent on each other
If crop first then baseline ---> obsviously non sense
If baseline first then cropt ---> why not possible currently?

For many statistical contrasts, one may wish to apply a baseline (e.g. [-500.0, 0.0]) and then crop epochs (e.g. to apply statistical tests highly sensitive to the number of samples and decimation - wink) 

Currently, the warning is not strong enough to realize this happens: the sequence of operations applied to the data is essentially not respected and unless one looks at the execution online (aka not run some parallel processing) one really easily misses this is ongoing on. 

#### Describe your solution
Could crop have an option apply_baseline = 'on' /  'off' prior to cropping?

#### Describe possible alternatives
1) Have an easy way to mask portions of the signal but this won't help lower the computational problem
2) bypass apply_baseline but then the info history won't be kept

larsoner:
Are you on latest `master`? It sounds like this was probably fixed in https://github.com/mne-tools/mne-python/pull/7629 due to issue https://github.com/mne-tools/mne-python/issues/7628

virvw:
This is my current setup

conda 4.8.3
Python 3.8.3
(base) vv221713@is235208:~$  python -c "import mne; mne.sys_info()"
Platform:      Linux-4.4.0-187-generic-x86_64-with-glibc2.10
Python:        3.8.3 (default, Jul  2 2020, 16:21:59)  [GCC 7.3.0]
Executable:    /home/vv221713/anaconda3/bin/python
CPU:           x86_64: 32 cores
Memory:        62.5 GB

mne:           0.20.7
numpy:         1.18.5 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.5.0
matplotlib:    3.2.2 {backend=Qt5Agg}

sklearn:       0.23.1
numba:         0.50.1
pandas:        1.0.5

virvw:
1) OK after simple mne reinstall with pip: 
 I confirm that version mne: 0.20.8  does not warn about not applying baseline. However, it does say "baseline off" 

[...]

PR #8444, opened 2020-10-28
ENH, BUG: Retain Epochs.baseline after cropping; replace baseline=(None, None) with (tmin, tmax)

hoechenberger:
- Partial cropping of an `Epochs` baseline will now correctly inform the user that baseline correction has been switched off. Previously, this warning would only be issued if the **entire** baseline period was removed
- If a `baseline` tuple containing one or two `None` values – e.g. ``(None, 0)``, ``(0, None)``, or ``(None, None)`` – is passed to `Epochs` or `Epochs.apply_baseline`, those `None` value(s) will be replaced with the actual **time** (i.e., `Epochs.tmin` and `Epochs.tmax`, respectively) when populating`Epochs.baseline` This makes it possible (or at least easier) to later figure out whether we're cropping away parts of an existing baseline.

This is work that's laying the ground for resolving #8216

Probably should also be backported?


agramfort (approved):

mne/epochs.py
```diff
@@ -557,6 +561,15 @@ def _check_consistency(self):
         assert all(isinstance(log, tuple) for log in self.drop_log)
         assert all(isinstance(s, str) for log in self.drop_log for s in log)
 
+    def _gen_baseline_attr(self, baseline):
```
agramfort, line 564: I would not make it a method but jsut a private function _check_baseline

larsoner:
> Partial cropping of an Epochs baseline will now correctly inform the user that baseline correction has been switched off.

I don't think this is the right approach. I think we should make it so that there is no longer a requirement that the baseline period be within `tmin, tmax` instead.

hoechenberger:
> > Partial cropping of an Epochs baseline will now correctly inform the user that baseline correction has been switched off.
> 
> 
> 
> I don't think this is the right approach. I think we should make it so that there is no longer a requirement that the baseline period be within `tmin, tmax` instead.

I'm not sure I understand?

What I want to change in a followup PR is if we crop BL-corrected Epochs, we would retain the baseline attribute and the repr would essentially tell you, "these Epochs were BL-corrected for the following BL-period, which doesn't exist anymore"

Is that where you would like to go too?

hoechenberger:
Just trying to make super small incremental steps to reduce the risk of introducing bugs

larsoner:
No, I want to make it so that we no longer require the two values in `epochs.baseline` to be in the interval `(tmin, tmax)`. Concretely I would:

1. Convert `None` entries in a `baseline` tuple to float values at some point. I'm thinking right now that this should probably be *when the value of None would change effective value*. In other words:
    ```
    epochs = mne.Epochs(..., tmin=-0.2, baseline=(None, 0))
    assert epochs.baseline == (None, 0)
    epochs.crop(-0.1, None)
    assert epochs.baseline == (-0.2, 0)  # because None is wrong, we need to insert the float
    ```
    An alternative would be to always convert `None` values in an `epochs.baseline` tuple to float. This is probably simpler, and might be okay to do as well.

    It's not clear to me yet if we need to place restrictions on whether or not data needs to be preloaded for this to work. For example, if data is not preloaded in the case above, in `_get_epoch_from_raw` or whatever, we actually have to be more careful, because we'll need to load from `min(bmin, tmin)` to `max(bmax, tmax)` rather than just loading from `tmin` to `tmax`.

2. Allow `epochs.baseline` values to be outside the range `(tmin, tmax)` (necessary for the above to work).

hoechenberger:
> No, I want to make it so that we no longer require the two values in `epochs.baseline` to be in the interval `(tmin, tmax)`. Concretely I would:
> 
> 1. Convert `None` entries in a `baseline` tuple to float values at some point. I'm thinking right now that this should probably be _when the value of None would change effective value_ [...]
>    An alternative would be to always convert `None` values in an `epochs.baseline` tuple to float. This is probably simpler, and might be okay to do as well.

This behavior is currently implemented in this PR. I like it because it's explicit, and when you look at the repr of the `Epochs`, you get explicit baseline values instead of e.g. `[None, 0]`:

```python
epochs = mne.Epochs(raw, events, baseline=(None, 0), preload=True)
assert epochs.baseline == (epochs.tmin, 0)

epochs.crop(0, None)
assert epochs.baseline is None
```

larsoner:
> This behavior is currently implemented in this PR. I like it because it's explicit, and when you look at the repr of the Epochs, you get explicit baseline values instead of e.g. [None, 0]:

I can live with this behavior. I can see advantages of each.

But I don't like the first part of the PR:

> Partial cropping of an Epochs baseline will now correctly inform the user that baseline correction has been switched off. Previously, this warning would only be issued if the entire baseline period was removed

I don't think we need it because in my plan we make it so that the baseline period is *not* removed and can always be reported properly.

hoechenberger:
> I don't think we need it because in my plan we make it so that the baseline period is _not_ removed and can always be reported properly.

Yes yes yes, if I'm getting you correctly, this is **exactly** what I had initially planned for a follow-up PR. We're talking about this intended behavior, right?

```python
baseline = (None, 0)
epochs = mne.Epochs(raw, events, baseline=baseline, preload=True)
epochs.crop(0, None)
assert epochs.baseline == baseline
```

larsoner:
> Yes yes yes, if I'm getting you correctly, this is exactly what I had initially planned for a follow-up PR. We're talking about this intended behavior, right?

As long as you really mean for the last line:
```
assert epochs.baseline == (epochs.times[0], 0)  # not (None, 0)!
```
Then yes this is the behavior I have in mind.

But why add this other, seemingly unrelated new behavior ("Partial cropping of an Epochs baseline will now correctly inform the user that baseline correction has been switched off.") in this PR if it's immediately going to be removed by the next PR, which will avoid the removal / "switch off" part? I don't think we should do this, i.e., it should be removed from this PR.

We should not backport any of this stuff BTW because we want months to test it out and make sure we're not breaking things.

larsoner:
Had to update my above comment, it shouldn't be `(epochs.times[0], 0)` but rather whatever the original first time was, i.e., `(-0.2, 0)`

hoechenberger:
> Had to update my above comment, it shouldn't be `(epochs.times[0], 0)` but rather whatever the original first time was, i.e., `(-0.2, 0)`

Yes this is what I had in mind. Sorry, got mixed up going those back-and-forth between our discussion and my code.

I agree we should probably do this all at once in this PR. I won't have time to work on it for a few days I'm afraid, but happy to pick it up again very soon.

doc/changes/latest.inc
```diff
@@ -31,6 +31,8 @@ Enhancements
 
 - Add sEEG source visualization using :func:`mne.stc_near_sensors` and sEEG working tutorial by `Eric Larson`_ and `Adam Li`_ (:gh:`8402`)
 
+- If a ``baseline`` tuple containing one or two ``None`` values – e.g. ``(None, 0)``, ``(0, None)``, or ``(None, None)`` – is passed to `~mne.Epochs` or `~mne.Epochs.apply_baseline`, the ``None`` value(s) will be replaced with the actual time (i.e., :attr:`~mne.Epochs.tmin` and :attr:`~mne.Epochs.tmax`, respectively) when populating :attr:`~mne.Epochs.baseline`, by `Richard Höchenberger`_ (:gh:`8442`)
```
sappelhoff, line 34: just chiming in that this would be great to have!

hoechenberger: will continue working on this soon.

hoechenberger:
I've tried to implement what we discussed, but I'm in need of a break and will have to look at this again later tonight with hopefully fresh(er) eyes. This just to say, still WIP, and needs a little more thinking

mne/epochs.py
```diff
@@ -147,8 +147,6 @@ def _save_part(fid, epochs, fmt, n_parts, next_fname, next_idx):
     # save baseline
     if epochs.baseline is not None:
         bmin, bmax = epochs.baseline
-        bmin = epochs.times[0] if bmin is None else bmin
-        bmax = epochs.times[-1] if bmax is None else bmax
```
hoechenberger, line 151: Not needed anymore because we replace `None` with `timin` and `tmax` early on now

hoechenberger:
Let's see if tests pass now.

I personally feel it's not so easy to intuitively grasp what I've done here just by looking at the altered code & tests. I will try to update one or more examples to demonstrate the new behavior.

hoechenberger:
(btw no clue how to fix the Circle issue. Doc building chokes on 

    :attr:`~mne.Epochs.baseline`

in `latest.inc`. I've tried a number of different things but cannot seem to get it to work. This, however, works just fine:

    :attr:`~mne.Epochs.tmin`

maybe because it's a property..?

larsoner:
Yes properties and attributes are documented differently by NumpyDoc and what works for one will not necessarily work for the other

drammock:
> Yes properties and attributes are documented differently by NumpyDoc and what works for one will not necessarily work for the other

@hoechenberger nowadays we don't even need the `:role:` specification for things within the MNE namespace. So if the property/attribute is linkable (which I think is determined by presence in `doc/python_reference.rst`), it should be possible as 
```
`~mne.Epochs.baseline`
```

hoechenberger:
@drammock 
Thanks or the hints! However I've tried using

> ```
> `~mne.Epochs.baseline`
> ```

and it's not woring either. I've also tried adjusting `doc/python_reference.rst` (although `Epochs` is already listed there), no luck. For now I'm just going to use 

    ``Epochs.baseline``

without linking :\

hoechenberger:
Ok it's actually not that complicated after all :) The altered cropping behavior is demonstrated in https://23393-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_10_epochs_overview.html#selection-in-the-time-domain

## Further changes
```python
# %%
import os
import mne

                                    'sample_audvis_raw.fif')
```

### `None`s in `baseline` are replaced with their respective times (`tmin`, `tmax`), and an info message is issued
```python
baseline = (None, None)
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7,
                    baseline=baseline)

print(f'\nbaseline is: {epochs.baseline}')
```
Output:
```
Not setting metadata
Not setting metadata
86 matching events found
Setting baseline interval to [-0.2996928197375818, 0.6992832460543575] sec   <-- NEW
Created an SSP operator (subspace dimension = 3)
3 projection items activated

baseline is: (-0.2996928197375818, 0.6992832460543575)  <-- NEW: None's were replaced
```

### Instantiating `Epochs` with a baseline that's out of the data range will now produce an info message and adjust the interval:
```python
baseline = (-1000, 0)  # Would raise on `master`
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7,
                    baseline=baseline)

print(f'\nbaseline is: {epochs.baseline}')
```
Output:
```
Not setting metadata
Not setting metadata
86 matching events found
Baseline interval [-1000.0, 0.0] sec is outside of epoch data [-0.2996928197375818, 0.6992832460543575] sec
Setting baseline interval to [-0.2996928197375818, 0.0] sec
Created an SSP operator (subspace dimension = 3)
3 projection items activated

baseline is: (-0.2996928197375818, 0.0)
```

At one point this made it easier for me to load `Epochs` with cropped ("invalid") baseline periods from disk. If this behavior is undesired, I will have to look into this again (I don't remember the details anymore)

larsoner:
> Instantiating Epochs with a baseline that's out of the data range will now produce an info message and adjust the interval:

Please do not do this. It does not reflect what was done to the data and is not part of the plan I keep pushing, see for example point (2) [above](https://github.com/mne-tools/mne-python/pull/8444#issuecomment-717932225).

hoechenberger:
> > Instantiating Epochs with a baseline that's out of the data range will now produce an info message and adjust the interval:
> 
> Please do not do this. It does not reflect what was done to the data and is not part of the plan I keep pushing, see for example point (2) [above](https://github.com/mne-tools/mne-python/pull/8444#issuecomment-717932225).

Ok I will rework this part.

larsoner:
... in particular, doing `Epochs(tmin=-0.2 tmax=0.7, baseline=(-0.3, 0.7)` should raise an error. If you do `epochs = Epochs(tmin=-0.2, tmax=0.7, baseline=(-0.2, 0.7)` it should be fine and even `epochs.crop(0, 0.5); assert epochs.baseline == (-0.2, 0.7)` should eventually work, but probably not in this PR (it's separable).

hoechenberger:
> even `epochs.crop(0, 0.5); assert epochs.baseline == (-0.2, 0.7)` should eventually work, but probably not in this PR

That's actually already implemented here 😅 seems I got priorities wrong. Will rework things.

larsoner:
If it makes sense to do it all as part of this PR it's fine by me. I don't know what's going to be easiest actually, as different parts of the code interact a lot.

hoechenberger:
> different parts of the code interact a lot.

Yep, that's been giving me a headache :) Thanks for taking the time to review and discuss approaches and my attempts, I'm learning as we move along…

hoechenberger:
@larsoner It passes your proposed tests:

```python
# %%
import os
import mne

                                    'sample_audvis_raw.fif')

# %%
# raises a ValueError
epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(-0.3, 0.7))

# %%
# Good
epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(-0.2, 0.7))

# %%
# Good
epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(-0.2, 0.7))

epochs.load_data().crop(0, 0.5)
assert epochs.baseline == (-0.2, 0.7)
```

hoechenberger:
Passes as well:

```python
# %%
import numpy as np

epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(-0.2, 0.7))

epochs.load_data().crop(0, 0.5)
epochs.save('/tmp/epo.fif', overwrite=True)
epochs = mne.read_epochs('/tmp/epo.fif')
np.testing.assert_allclose(epochs.baseline, (-0.2, 0.7))
```

hoechenberger:
And, lastly, of course:

```python
# %%
# Good
epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(None, 0))

assert epochs.baseline == (epochs.tmin, 0)
```

larsoner:
Should be able to look tomorrow

hoechenberger:
Another test:

```python
# %%
# Good
import numpy as np

epochs = mne.Epochs(raw, events=events,
                    tmin=-0.2, tmax=0.7,
                    baseline=(-0.2, 0.7))

epochs.load_data().crop(0, 0.5)
epochs.save('/tmp/epo.fif', overwrite=True)
epochs_loaded = mne.read_epochs('/tmp/epo.fif', preload=False)

epochs_loaded.load_data()
np.testing.assert_array_almost_equal(epochs.get_data(),
                                     epochs_loaded.get_data())
```

doc/changes/latest.inc
```diff
@@ -69,6 +69,9 @@ Enhancements
 
 - `mne.preprocessing.ICA.plot_sources` now displays an `mne.preprocessing.ICA.plot_properties` window when right-clicking on component names on the y-axis (:gh:`8381` by `Daniel McCloy`_)
 
+- If a ``baseline`` tuple containing one or two ``None`` values – e.g. ``(None, 0)``, ``(0, None)``, or ``(None, None)`` – is passed to `~mne.Epochs` or `~mne.Epochs.apply_baseline`, the ``None`` value(s) will be replaced with the actual time (i.e., :attr:`~mne.Epochs.tmin` and :attr:`~mne.Epochs.tmax`, respectively) when populating ``Epochs.baseline``, by `Richard Höchenberger`_ (:gh:`8442`)
```
larsoner, line 72: wrong format for these nowadays

mne/epochs.py
```diff
@@ -1353,6 +1361,16 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
                         'points ...' % (n_events, len(self._raw_times)))
+
+        # don't rescale if the baseline is invalid (due to cropping of the
+        # baseline period)
+        rescale_data = True
+        if self.baseline != _check_baseline(self.baseline, tmin=self.tmin,
+                                            tmax=self.tmax,
+                                            sfreq=self.info['sfreq'],
+                                            on_baseline_outside_data='adjust'):
+            rescale_data = False
```
larsoner, line 1372: Relying on exact float equality is dangerous. One time when `tmin` and `sfreq` are double floating point you could get one result, and once a save-load round trip happens and they are single floating point you might get another. Is there some way instead to actually keep track of when we actually need to apply a baseline?

When I originally went to work on this at one point, I considered adding a private `self._do_baseline = False` that was `True` only in the one case when it actually needed to be applied later rather than right away, which I think is only for non-preloaded Epochs. This way you can either 1) check that `not self._do_baseline` if a `crop` operation is requested that would remove the baseline period, or (better but harder) 2) we would make the raw data getter get data from the interval `min(bmin, tmin), max(bmax, tmax)`, then crop.

mne/epochs.py
```diff
@@ -1928,37 +1948,72 @@ def as_type(self, ch_type='grad', mode='fast'):
...
-        if baseline_tmax > tmax + tstep:
-            raise ValueError(
-                "Baseline interval (tmax = %s) is outside of epoch "
-                "data (tmax = %s)" % (baseline_tmax, tmax))
-        if baseline_tmin > baseline_tmax:
-            raise ValueError(
-                "Baseline min (%s) must be less than baseline max (%s)"
-                % (baseline_tmin, baseline_tmax))
+def _check_baseline(baseline, tmin, tmax, sfreq,
+                    on_baseline_outside_data='raise'):
```
larsoner, line 1952: Rather than do all the `tmin/tmax/sfreq` gymnastics, can we just pass `self.times`?

mne/epochs.py
```diff
@@ -2777,11 +2835,14 @@ def __init__(self, fname, proj=True, preload=True,
         drop_log = tuple(drop_log[:step])
 
         # call BaseEpochs constructor
+        # again, ensure we reataing the baseline period originally loaded from
```
larsoner, line 2838:
```suggestion
        # again, ensure we reatain the baseline period originally loaded from
```

mne/epochs.py
larsoner: ... if it seems very difficult to refactor using some `_do_baseline`, another alternative would be to have a `_baseline_isclose(a, b, sfreq)` that checked that the terms were close to within `1e-2 / sfreq` or so for example. This should be good enough to deal with floating point inaccuracies (time points should step by `1. / sfreq`, so doing `1e-2 / sfreq` allows for a lot of slop, which could be necessary for very long / hundreds-of-thousands-of-time-points epochs).

mne/epochs.py
hoechenberger: We now expect `self.times` instead of `tmin, tmax` (ad21ff9); getting rid of `sfreq` is not as trivial: I tried simply replacing it with `times[-1] - times[0]`, but this breaks a few tests. So I'm not touching that one for now…

mne/epochs.py
hoechenberger: > When I originally went to work on this at one point, I considered adding a private `self._do_baseline = False` that was `True` only in the one case when it actually needed to be applied later rather than right away, which I think is only for non-preloaded Epochs.

I have implemented this now in eca094a. And I have to admit I'm starting to get (have gotten?) lost a bit regarding all the side-effects this open-heart surgery might have. Fortunately we've got excellent test coverage, but I'm quite concerned I might be introducing bugs here. I think we should also manually test before finally merging this…

> This way you can either 1) check that `not self._do_baseline` if a `crop` operation is requested that would remove the baseline period, 

This is not necessary, as cropping only works on preloaded data, and loading applies baseline correction.

> or (better but harder) 2) we would make the raw data getter get data from the interval `min(bmin, tmin), max(bmax, tmax)`, then crop.

I haven't looked into this, I don't think I want to have this kind of complexity for now :)

mne/epochs.py
```diff
@@ -1502,8 +1514,13 @@ def __repr__(self):
...
             s += 'off'
         else:
-            s += '[%s, %s]' % tuple(['None' if b is None else ('%g' % b)
-                                     for b in self.baseline])
+            s += '[%s, %s] sec' % tuple(['None' if b is None else ('%g' % b)
+                                        for b in self.baseline])
+            if self.baseline != _check_baseline(
+                    self.baseline, times=self.times, sfreq=self.info['sfreq'],
+                    on_baseline_outside_data='adjust'):
+                s += ' (baseline period was cropped after baseline correction)'
```
hoechenberger, line 1522: I need to rethink this one, as we still have a float comparison here

hoechenberger: No never mind, I think this is okay in this specific case… emphasis on I **think**

hoechenberger: `_check_baseline` allows for up to `1./sfreq` slack, so this should be okay here.

mne/epochs.py
hoechenberger: Haha or maybe I didn't. Just tried again and all tests do pass. Maybe I actually tried with `times[-1] - times[0]` initially, instead of `times[1] - times[0]`. At any rate, dropped the sfreq param now in a2ea0b5

hoechenberger: I've reverted this change again, as it won't work with data that only has one time point (as used in `test_pick_seeg_ecog()`)

larsoner (commented):
Changes look surprisingly readable!

I would say what is missing are some tests that the data are not changed. For this you could both add to the existing tests, but also add some new basic/separate test, for example have 1 channel of "EEG" data be a 2 sec linear ramp from -1 to 1 with a single event at 1 sec and epoch with `Epochs(..., tmin=-0.2, tmax=0.2, baseline=(-0.1, 0.1))`, check the values during baseline and epoch are `allclose` with what you'd expect, then crop to `t>=0`, assert the values are still the same and the `.baseline` is the same, do I/O round trip, etc.

hoechenberger:
> Changes look surprisingly readable!

Yes, this actually made me a bit suspicious ;) Will add more tests as you suggested! Thanks for the review!

larsoner:
@hoechenberger it would be great to have this for 0.22 -- any chance you can add the tests quickly so we can have at least a week or two of this code in `master` before we release it in 0.22?

hoechenberger:
@larsoner I will work on this tomorrow if it's okay with @agramfort? Hopefully the tests will all pass and not reveal a major screwup ;)

hoechenberger:
@larsoner I've added some tests as suggested

larsoner (approved):
@agramfort feel free to review and merge if you're happy!

hoechenberger:
@cbrnr @drammock @sappelhoff @jasmainak 
I you have an existing processing pipeline / scripts at hand, it would be amazing if you could check out this branch and see if it breaks your stuff. I still don't trust it 100%, even though we have good test coverage now.

mne/epochs.py
```diff
@@ -1928,37 +1939,74 @@ def as_type(self, ch_type='grad', mode='fast'):
...
-                "Baseline min (%s) must be less than baseline max (%s)"
-                % (baseline_tmin, baseline_tmax))
+def _check_baseline(baseline, times, sfreq, on_baseline_outside_data='raise'):
+    """Check if the baseline is valid, and adjust it if requested.
+
+    ``None`` values inside the baseline parameter will be replaced with
+    ``times[0]`` and ``times[-1]``.
+
+    Parameters
+    ----------
```
agramfort, line 1949: some parameters are missing in docstring

agramfort (approved):
just a nitpick.

I did not review as the diff is hard but I tested locally with a bunch of scenario and it works great

mne/epochs.py
hoechenberger: I've added the rest

cbrnr:
I don't really have a pipeline for ERP processing because I mostly work with ERD/ERS. However, in one of our current studies we will actually analyze ERPs, but this will probably take too long to serve as a test for this PR. Therefore, I'll have to pass on this one.

hoechenberger:
Thanks anyway, @cbrnr!

agramfort:
@larsoner your call? merge?

larsoner:
Thanks @hoechenberger !

Issue #8445, opened 2020-10-28
plot_alignment should show transparent brain when using sEEG electrodes


kdoelling1919:
#### Describe the new feature or enhancement
Using plot_alignment to see where sEEG electrodes are located in the brain. But if you you use any brain surfaces such as `surfaces=["pial"]` the brain is fully opaque so you can't see where the electrodes are that are inside the brain. So it would be best to give the brain a bit of transparency when you have sEEG electrodes embedded in the brain.

#### Describe your proposed implementation
As far as I can tell, this is a single line change. I have already created a feature branch in my fork which changes the setting of opacity depending on whether there are sEEG electrodes present for plotting.
https://github.com/kdoelling1919/mne-python/blob/5b039cdfb3da64c89947f09a0bfd83481955f29c/mne/viz/_3d.py#L840

#### Describe possible alternatives
Another option might be to set transparency generally any time there are sensors to be plotted that are inside the brain volume. Sounds like a bit more work (and more knowledge about the surfaces than I currently have) but it would be more general.

#### Additional comments
Here is what the plot_alignment looks like in current master version of plot_alignment
![master](https://user-images.githubusercontent.com/13767931/97433444-d39d9d00-191d-11eb-97ee-45291e469d08.png)

Here is what it looks like in my branch.
![transparent](https://user-images.githubusercontent.com/13767931/97433453-d6988d80-191d-11eb-9d8c-ae12d667ab56.png)

If people agree this is useful beyond just for myself. I'll make the pull request.

Thank you!

agramfort:
@GuillaumeFavelier thoughts? any way to expose such options?

GuillaumeFavelier:
Yeah, I prefer the result with transparency

> any way to expose such options?

`plot_alignment` has so many parameters already, I would rather not add another one. I prefer the auto-detection approach. I think computing the bounds of the brain surface is enough as a compromise between precision and computational cost.

larsoner:
Currently we make the innermost surface opaque (alpha=1). It's probably fine instead just to always make the innermost surface translucent (alpha=0.75 or something). That way you can see stuff behind it anyway, which is nice. Would that work for your sEEG example?

These are the examples where `plot_alignment` is used:

https://mne.tools/dev/generated/mne.viz.plot_alignment.html#examples-using-mne-viz-plot-alignment

[...]

PR #8446, opened 2020-10-28
[WIP] Add transparency to plot_alignment 

kdoelling1919:
#### Reference issue
Issue #8445

Reduce maximum alpha to be 0.75 in plot_alignment to aid viewing of electrodes inside the brain and skull (mostly sEEG). Also edits a few tutorials to see the new outputs of circleCI

#### Additional information
The goal currently is to see how these look in the tutorial examples so that we can decide whether this is the best strategy or to use an autodetect (of whether sensors are inside the surface) method instead.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -16,6 +16,7 @@ Current (0.22.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Update ``surfaces`` argument in :func:`mne.viz.plot_alignment` to allow dict for transparency values, and set default for sEEG data to have transparency (by `Keith Doelling`_ (:gh:`8445`))
 
 - Add ``n_pca_components`` argument to :func:`mne.viz.plot_ica_overlay` by `Eric Larson`_ (:gh:`8351`)
 
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -438,7 +438,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         The subject name corresponding to FreeSurfer environment
         variable SUBJECT. Can be omitted if ``src`` is provided.
     %(subjects_dir)s
-    surfaces : str | list
+    surfaces : str | list | dict
         Surfaces to plot. Supported values:
 
         * scalp: one of 'head', 'outer_skin' (alias for 'head'),
@@ -447,6 +447,11 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
           'inner_skull')
         * brain: one of 'pial', 'white', 'inflated', or 'brain'
           (alias for 'pial').
+
+        Can be dict to specify alpha values for each surface. Use None
+        to specify default value. Specified values must be between 0 and 1.
+        Example:
+        ``surfaces={'brain': 0.4, 'outer_skull': 0.6, 'head': None}``
 
         Defaults to 'auto', which will look for a head surface and plot
         it if found.
@@ -593,9 +598,14 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
 
     if isinstance(surfaces, str):
         surfaces = [surfaces]
-        user_alpha = {}
     if isinstance(surfaces, dict):
         user_alpha = surfaces
+        # check that alpha values contain floats between 0 and 1 or None
+        check_values = [x is None or (isinstance(x, (float, int)) and x >= 0
+                                      and x <= 1) for x in user_alpha.values()]
+        if not all(check_values):
+            raise ValueError("If surfaces is dict, its values must contain "
+                             "floats between 0 and 1 or None")
     else:
         user_alpha = {}
     surfaces = list(surfaces)
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -950,8 +950,7 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         if other_bools[key] and len(picks):
             title = DEFAULTS["titles"][key] if key != 'fnirs' else 'fNIRS'
             if key != 'fnirs' or 'channels' in fnirs:
-                other_loc[key] = [
-                       info['chs'][pick]['loc'][:3] for pick in picks]
+                other_loc[key] = [info['chs'][pick]['loc'][:3] for pick in picks]
                 # deal with NaN
                 other_loc[key] = np.array([loc for loc in other_loc[key]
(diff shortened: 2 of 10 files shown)


mne/viz/_3d.py
```diff
@@ -448,6 +448,11 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
         * brain: one of 'pial', 'white', 'inflated', or 'brain'
           (alias for 'pial').
 
+        Can be dict to specify alpha values for each surface. Use None
+        to specify default value. Specified values must be between 0 and 1.
+        Example:
+        ``surfaces={'brain': 0.4, 'outer_skull': 0.6, 'head': None}``
+
         Defaults to 'auto', which will look for a head surface and plot
```
larsoner, line 456: you can get nicer rendering with a code block:
```suggestion
        to specify default value. Specified values must be between 0 and 1,
        for example::

            surfaces=dict(brain=0.4, outer_skull=0.6, head=None)

        Defaults to 'auto', which will look for a head surface and plot
```

mne/viz/_3d.py
```diff
@@ -593,6 +598,16 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
 
     if isinstance(surfaces, str):
         surfaces = [surfaces]
+    if isinstance(surfaces, dict):
+        user_alpha = surfaces
```
larsoner, line 602: Dicts are mutable and we modify this later so we need to copy so we don't change the user's variable
```suggestion
        user_alpha = surfaces.copy()
```

mne/viz/_3d.py
```diff
@@ -593,6 +598,16 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
...
+        # check that alpha values contain floats between 0 and 1 or None
+        check_values = [x is None or (isinstance(x, (float, int)) and x >= 0
+                                      and x <= 1) for x in user_alpha.values()]
+        if not all(check_values):
+            raise ValueError("If surfaces is dict, its values must contain "
+                             "floats between 0 and 1 or None")
```
larsoner, line 608:
```suggestion
        for key, value in user_alpha.items():
            _validate_type(key, str, f'surfaces key {repr(key)}')
            _validate_type(value, (None, 'numeric'), f'surfaces[{repr(key)}]'
            if value is not None:
                user_alpha[key] = float(value)
                if not 0 <= user_alpha[key] <= 1:
                    raise ValueError(
                        f'surfaces[{repr(key)}] ({value}) must be between 0 and 1')
```
This will be more specific and give TypeError or ValueError as appropriate

mne/viz/_3d.py
```diff
@@ -836,8 +853,9 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
     skull_alpha = dict()
     skull_colors = dict()
     hemi_val = 0.5
+    max_alpha = 1.0 if len(other_picks['seeg']) == 0 else 0.75
     if src is None or (brain and any(s['type'] == 'surf' for s in src)):
-        hemi_val = 1.
+        hemi_val = max_alpha
     alphas = (4 - np.arange(len(skull) + 1)) * (0.5 / 4.)
```
larsoner, line 859: I would adjust this line to its linspace equivalent:
```
np.linspace(1. / 2., 0, 5)[:len(skull) + 1]
```
And then make the appropriate adjustment in this PR, namely to swap out the 1 to obtain:
```
np.linspace(max_alpha / 2., 0, 5)[:len(skull) + 1]
```
In other words, all defaults should get less opaque, not just the outermost one (scalp).

But if you thought about this and tried it already and it doesn't look as good that's fine! But it would still be nice to change to the `np.linspace` line now that I figured out what the heck it's doing. It's much more readable...

larsoner (commented):
Looks good so far!

mne/viz/_3d.py
kdoelling1919: Great! will incorporate

mne/viz/_3d.py
kdoelling1919: I figured since surfaces is reassigned in the next lines it wouldn't matter. But agreed this is safer/cleaner. Adding!

mne/viz/_3d.py
kdoelling1919: great!

kdoelling1919:
Looks good except for now you can see one of the depth electrodes didn't really hit the brain...
https://23107-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html#sphx-glr-auto-tutorials-misc-plot-seeg-py

Other tutorials that used plot_alignment seem to be in tact.
https://23107-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_source_alignment.html#sphx-glr-auto-tutorials-source-modeling-plot-source-alignment-py

https://23107-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/plot_eeg_on_scalp.html#sphx-glr-auto-examples-visualization-plot-eeg-on-scalp-py

agramfort:
this one
https://23107-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html#sphx-glr-auto-tutorials-misc-plot-seeg-py
looks bad no?

>

kdoelling1919:
Here's what it looks in current version
https://mne.tools/dev/auto_tutorials/misc/plot_seeg.html#sphx-glr-auto-tutorials-misc-plot-seeg-py

The only difference is that I added the pial surface to show that it can be done with the transparency. It looks fine to me but maybe could be improved by changing the viewing angle?

agramfort (approved):
all good !

doc/changes/latest.inc
```diff
@@ -16,6 +16,7 @@ Current (0.22.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Update ``surfaces`` argument in :func:`mne.viz.plot_alignment` to allow dict for transparency values, and set default for sEEG data to have transparency (by `Keith Doelling`_ (:gh:`8445`))
```
larsoner, line 19:
```suggestion
- Update ``surfaces`` argument in :func:`mne.viz.plot_alignment` to allow dict for transparency values, and set default for sEEG data to have transparency (by `Keith Doelling`_ (:gh:`8445`)
```
But we can fix this later, we're probably reformatting all of these lines within this release anyway

larsoner:
Thanks @kdoelling1919 !

kdoelling1919:
great success! Thanks to everyone!

PR #8449, opened 2020-10-28
MRG, BUG: Fix errant writing of too large int (part 2/2)

### LINKED ISSUE #7897: 2GB .fif limit

larsoner:
Closes #7897

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -29,6 +29,8 @@ Enhancements
 
 - Speed up :class:`mne.decoding.TimeDelayingRidge` with edge correction using Numba by `Eric Larson`_ (:gh:`8323`)
 
+- Add :meth:`mne.Epochs.reset_drop_log_selection` to facilitate writing epochs with many ignored entries in their drop log by `Eric Larson`_ (:gh:`8449`)
+
 - Add sEEG source visualization using :func:`mne.stc_near_sensors` and sEEG working tutorial by `Eric Larson`_ and `Adam Li`_ (:gh:`8402`)
 
 Bugs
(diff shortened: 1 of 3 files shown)


agramfort:
thx @larsoner

PR #8450, opened 2020-10-28
FIX, VIZ: ensure figs appear

### LINKED ISSUE #8441: Raw.plot() issue
I found something maybe a bug.
Now I used the master branch in GitHub.  I use Pycharm and terminal to run codes. One of my examples uses **raw.plot()** to plot raw data.  While I ran codes in Pycharm's console, **raw.plot()** went well, a figure showed up. While I ran .py in terminal like **python test.py**, no figure showed up. And if I type **python** in the terminal and type the codes in this way, also no figures showed up. But if I used the stable mne, everything was OK.

drammock:
closes #8441

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -695,6 +695,7 @@ class MNEBrowseFigure(MNEFigure):
         # fig.lasso.alpha_other = 0.3
         # fig.lasso.linewidth_selected = 3
         # fig.lasso.style_sensors(inds)
+        # plt_show(fig=fig)
 
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
     # HELP DIALOG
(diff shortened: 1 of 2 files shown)


drammock:
@hoechenberger after this is merged and #8381 is rebased, this should also fix the issue you mentioned about the help button not working.

hoechenberger (approved):

mne/viz/_figure.py
```diff
@@ -1173,6 +1175,7 @@ def _create_selection_fig(self):
         # add event listeners
         radio_ax.buttons.on_clicked(fig._radiopress)
         fig.canvas.mpl_connect('lasso_event', fig._set_custom_selection)
+        #plt_show(fig=fig)
```
agramfort, line 1178: cleanup?

mne/viz/_figure.py
```diff
@@ -695,6 +695,7 @@ def _create_ch_location_fig(self, idx):
         # fig.lasso.alpha_other = 0.3
         # fig.lasso.linewidth_selected = 3
         # fig.lasso.style_sensors(inds)
+        # plt_show(fig=fig)
```
agramfort, line 698: cleanup?

hoechenberger: I think these lines were left here on purpose…

mne/viz/_figure.py
hoechenberger: this one ideed looks like cruft :)

mne/viz/_figure.py
drammock: indeed, those are dealt with in #8381

agramfort (approved):

PR #8454, opened 2020-10-30
fix setting xlabel and ylabel in example

jjnurminen:
This is a small fix for an example by @dengemann, where labels of a subplot weren't being set correctly.

### DIFF
--- a/examples/visualization/plot_topo_customized.py
+++ b/examples/visualization/plot_topo_customized.py
@@ -48,8 +48,8 @@ def my_callback(ax, ch_idx):
     two parameters, the axis and the channel or data index.
     """
     ax.plot(freqs, psds[ch_idx], color='red')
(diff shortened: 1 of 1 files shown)


sappelhoff (approved):
good catch! thanks :-)

dengemann:
I am surprised about how long these obviously wrong two lines could exist without correction :) very good catch. Thank you!

dengemann (approved):

agramfort (approved):

jjnurminen:
The original code is interesting, because it replaces the instance methods with strings, which I guess is completely legal.
On the other hand, it seems that one can put any arbitrary stuff in the callbacks (like `non_existent_function()` or `raise NoSuchException`) and never see any error messages. :)

larsoner:
Thanks @jjnurminen

Issue #8451, opened 2020-10-29
A function get_ch_positions for DigMontage


adam2392:
#### Describe the new feature or enhancement
If I have an existing montage and want to add fiducials, there is currently no way to do that without calling `make_dig_montage` again. But the montage itself does not provide a way to get the ch positions.

Currently the only way I can find that gets a montage is either you make it from scratch make_dig_montage or you get it from raw using raw.get_montage. But if you say already have an existing montage via raw.get_montage() it seems you would have to hack out the channel positions and then call make_dig_montage again? Since there is no public API for getting the channel positions, I wonder if there's an easier way around this?

#### Describe your proposed implementation
Possibly a `get_ch_positions()` public function for DigMontage, that one can then do:

```
montage  = raw.get_montage()
ch_positions = montage.get_ch_positions()
```

#### Additional comments
TBD

agramfort:
sorry I don't get the usecase. Can you clarify what you start from and what
you are aiming for?

>

adam2392:
> sorry I don't get the usecase. Can you clarify what you start from and what you are aiming for?
> […](#)

Sorry about that. I updated the description. I just want to add fiducials to an existing montage.

agramfort:
ok got it. You want an easy way to get the parameters ch_pos etc... required by make_dig_montage from a DigMontage instance.

montage.get_ch_pos?

but then shall we add get_nasion, get_lpa etc...

@larsoner thoughts?

larsoner:
> but then shall we add get_nasion, get_lpa etc...

Maybe a more generic `get_positions(kinds='all' | 'channels' | 'extra' | 'fiducials' | list of str)`?

agramfort:
I prefer this yes. You would get a dict as output that you can directly
pass with **kwargs in make_dig_montage

>

larsoner:
In that case we should just make it `get_positions()`, and doing `mon_2 = make_dig_montage(**mon.get_positions())` should have all attributes of `mon` and `mon_2` the same. If you want the channel positions you just use `mon.get_positions['ch_pos']`

larsoner:
... or maybe `get_params` or `get_parameters()` is better, since `coord_frame` is an argument of `make_dig_montage`. So `mon.get_parameters()['ch_pos']` if you want just channel positions. Or we use `get_positions()` and live it being a misnomer because of `mon.get_positions()['coord_frame']` being there

larsoner:
[...]

PR #8460, opened 2020-10-30
[MRG] Getting ch positions from DigMontage via get_positions()

adam2392:
#### Reference issue
Fixes: #8451 

Adds `DigMontage.get_positions()` that would allow a full roundtrip with `make_dig_montage`. More importantly it obtains the DigMontage ch positions, coordinate frame and fiducials.

Adds a unit test. 

#### Additional information
~~For some reason the coordinate frame changes when using the template.~~ 

~~Is it safe to say we should have `get_positions` always return things in `coord_frame=head`? If so, what's suggestion to fix the unit test with the template that transforms unknown coordframe to the head coord_frame?~~ Now obtains the original coordinate frame.

### DIFF
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -317,7 +317,7 @@ class DigMontage(object):
             right preauricular point (``rpa``),
             Head Shape Polhemus (``hsp``), and
             Head Position Indicator(``hpi``).
-            E.g. ..
+            E.g.::
 
                 {
                     'ch_pos': {'EEG061': [0, 0, 0]},
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -316,7 +316,8 @@ class DigMontage(object):
             left preauricular point (``lpa``),
             right preauricular point (``rpa``),
             Head Shape Polhemus (``hsp``), and
-            Head Position Indicator(``hpi``). E.g. ..
+            Head Position Indicator(``hpi``).
+            E.g. ..
 
                 {
                     'ch_pos': {'EEG061': [0, 0, 0]},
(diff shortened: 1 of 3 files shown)


mne/channels/tests/test_montage.py
```diff
@@ -1333,17 +1333,25 @@ def test_get_montage():
...
 
+    # the montage should fulfill a roundtrip with make_dig_montage
+    test2_montage = make_dig_montage(**montage.get_positions())
+    assert_object_equal(test2_montage.dig, test_montage.dig)
+
     # 2. now do a standard montage
     montage = make_standard_montage('mgh60')
     # set the montage; note renaming to make standard montage map
-    raw.set_montage(montage)
+    raw.set_montage(montage.copy())
```
larsoner, line 1343: You should not need to `.copy`, if attributes are not properly copied when setting we have a bug somewhere. Is this leftover debugging cruft?

mne/channels/montage.py
```diff
@@ -304,6 +305,49 @@ def _get_dig_names(self):
...
+            Head Shape Polhemus (``hsp``), and
+            Head Position Indicator(``hpi``).
+            E.g.::
+
+                {
+                    'ch_pos': {'EEG061': [0, 0, 0]},
+                    'nasion': [0, 0, 1],
+                    'lpa': [0, 1, 0],
+                    'rpa': [1, 0, 0],
+                    'hsp': [np.nan, np.nan, np.nan],
```
larsoner, line 327: This is a strange result, I would have expected `[]` or `None` here instead of `[np.nan, np.nan, np.nan]`

adam2392: Yes you're right. They end up getting returned as `None` from `_get_data_as_dict_from_dig()` if they are not set.

larsoner (approved):

agramfort (approved):

agramfort:
@larsoner have you seen this issue https://travis-ci.org/github/mne-tools/mne-python/jobs/740940832 before?

larsoner:
Travis bug is just #8475, thanks @adam2392 !

PR #8462, opened 2020-10-31
MRG, BUG: Fix volumetric morphing treatment of zooms

### LINKED ISSUE #8453: BUG: compute_source_morph does not work well with a rescaling
The tests added in #8447 exposed some bug with our `compute_source_morph` when dealing with the output of `scale_mri`. In principle the `compute_source_morph` should just detect and undo the scaling but it doesn't. This is true both of uniform and non-uniform scalings in `scale_mri`.

The thing to do is probably to first add a test that `compute_source_morph` for a volume to itself is nearly an identity transform, then test locally where things go wrong. Perhaps something about the order translation->rigid body->full affine is wrong, or we need to allow a greater exploration of the space (somehow) because for a scaled MRI the only one of these that should happen is the full affine (including the scale factors is only possible there), but the intermediate outputs in `verbose=True` mode suggests that there are translations and rotations being imposed...

larsoner:
Closes #8453

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -28,6 +28,10 @@ Enhancements
 
 - Speed up heavy use of :meth:`mne.SourceMorph.apply` for volumetric source spaces by use of the method :meth:`mne.SourceMorph.compute_vol_morph_mat` by `Eric Larson`_ (:gh:`8366`)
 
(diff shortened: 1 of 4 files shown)


larsoner:
This one is green other than the old `matplotlib` bug

agramfort:
thx @larsoner !

PR #8464, opened 2020-11-01
MRG, BUG: fix writing with bids naming when no split is needed.

agramfort:
@larsoner I am curious how you would approach this.

Basically currently if you use bids split naming convention even if you don't need to split it will add a _split-01 in the first file.

do we have a way to detect early if the file will be split?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -54,6 +54,8 @@ Bugs
 
 - :func:`mne.io.read_raw_edf` now supports EDF files with invalid recording dates by `Clemens Brunner`_ (:gh:`8283`)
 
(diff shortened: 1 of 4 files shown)


larsoner:
The easy solution is to os.rename/move when necessary after writing the first file. To be safe we could check that the split01 name is also available when in BIDS naming mode, and even "reserve" it for us by touching an empty file there, and removing this touched file if we don't end up needing it or overwrite it with the move when we make it.

The hard solution is to compute how much space we'll actually need to write and split (and choose names) ahead of time based on that. As we've seen via epoch bug reports once or twice per release, it's a bit difficult to get right. But we have learned a lot about those calculations in the meantime at least.

I'm not sure which of these I prefer.

agramfort:
@larsoner ok for the rename option. Can I get your help on this one in this PR?

larsoner:
Sure, will push a fix tomorrow

larsoner:
Done

agramfort (commented):
thx @larsoner !

agramfort:
thx @larsoner !

agramfort:
backported

PR #8466, opened 2020-11-02
don't call subplots_adjust when user provided axes

drammock:


larsoner (approved):

larsoner:
Can we simplify the example somehow now and remove the `subplots_adjust` call from it?

drammock:
> Can we simplify the example somehow now and remove the `subplots_adjust` call from it?

looking into it now, in fact.

drammock:
turns out we still need `subplots_adjust` because of the axis labels, but now we only need to specify `bottom`.  I also delayed introducing the axis labels until the final cell, so we avoid `subplots_adjust` in the 2 earlier cells.

larsoner (approved):

larsoner:
Thanks @drammock

PR #8467, opened 2020-11-02
MRG, MAINT: Simplify morphing code

larsoner:
While looking into scaling issues with the volumetric morph, I wanted to first better figure out / remember how our surface code works. I always trip over the code when I read it. This is an attempt to simplify it, hopefully it's more readable now. I removed the `dense` code path since we don't use it. Along the way I made the `mpl` STC plotting code a tiny bit more DRY.

No need for `latest.inc` or example updates because this just changes internal code bits. No need for test update because our existing tests seemed sensitive enough to errors (e.g., iterating one two many or too few times).


agramfort (approved):

larsoner:
Failure is just some matplotlib version bump problem, will fix separately

PR #8469, opened 2020-11-02
MRG, FIX: retain reference to lasso.selection without blocking

drammock:
This makes it so that the list of sensors returned by `plot_sensors(..., kind='select')` will continue to be updated by figure interactions even if `block=False`. It restores previous behavior that was broken by #7955.

### DIFF
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -2116,7 +2116,7 @@ class SelectFromCollection(object):
             inters = set(inds) - set(sels)
             inds = list(inters.union(set(sels) - set(inds)))
 
-        self.selection = np.array(self.ch_names)[inds].tolist()
+        self.selection[:] = np.array(self.ch_names)[inds].tolist()
         self.style_sensors(inds)
         self.canvas.callbacks.process('lasso_event')
 
@@ -2134,7 +2134,7 @@ class SelectFromCollection(object):
 
     def select_many(self, inds):
         """Select many sensors using indices (for predefined selections)."""
-        self.selection = np.array(self.ch_names)[inds].tolist()
+        self.selection[:] = np.array(self.ch_names)[inds].tolist()
         self.style_sensors(inds)
 
     def style_sensors(self, inds):


drammock:
@mmagnuski this works for me in local testing, can you confirm?

mmagnuski:
@drammock I can confirm this fixes the problem, thanks!

drammock:
the Style failure has already been fixed in b2b63fd from #8477 
ready for review/merge

larsoner:
Travis minimal fails, there is a fix in #8477

larsoner:
(but it will conflict with the changes here)

drammock:
> (but it will conflict with the changes here)

we can merge that one first, I don't mind rebasing.

mne/viz/tests/test_raw.py
```diff
@@ -279,15 +279,13 @@ def test_plot_raw_selection(raw, new_mpl):
...
     sensor_ax = sel_fig.mne.sensor_ax
     # Lasso with 1 mag/grad sensor unit (upper left)
     _fake_click(sel_fig, sensor_ax, (0, 1), xform='ax')
     _fake_click(sel_fig, sensor_ax, (0.65, 1), xform='ax', kind='motion')
     _fake_click(sel_fig, sensor_ax, (0.65, 0.7), xform='ax', kind='motion')
     _fake_click(sel_fig, sensor_ax, (0, 0.7), xform='ax', kind='release')
-    want = ['MEG 0121', 'MEG 0122', 'MEG 0123'] if new_mpl else \
-        ['MEG 0111', 'MEG 0112', 'MEG 0131', 'MEG 0132', 'MEG 0133']  # XXX old
+    want = ['MEG 0121', 'MEG 0122', 'MEG 0123']
     want = sorted(want)
```
larsoner, line 289: Looks like this should be reverted

https://travis-ci.org/github/mne-tools/mne-python/jobs/741413151#L6453-L6460

larsoner:
Thanks @drammock !

Issue #8432, opened 2020-10-26
How to apply ica on brainvision raw data


arthurlgk:
I am getting this error when trying to fit the data:

ValueError: picks (NoneNone, treated as "data") yielded no channels, consider passing picks explicitly

How to resolve this issue?

hoechenberger:
Hello @arthurlgk, we'd need a little more information how you got there.

- Which version of `mne-python` are you running?
- Could you share the code that leads to this problem?

Thanks!

arthurlgk:
I am currently using mne 0.21.0.
And the code is the following:
```
fname = "demo_3.vhdr"
bv_raw = mne.io.read_raw_brainvision(dirname+fname)
ica = ICA(n_components=63, random_state=97)
ica.fit(bv_raw)
```
Thank you.

hoechenberger:
@arthurlgk Ok, this seems a little odd to me (the behavior you're seeing, not your code, which looks ok) @cbrnr Maybe you have an idea?

@arthurlgk Could you (privately) share with me one of the datasets that produces this problem? So I can try look into this myself. My email address is richard.hoechenberger@gmail.com

arthurlgk:
@hoechenberger Sent. Did you receive it?

cbrnr:
Not sure, but maybe the channel types are not EEG/MEG? @arthurlgk can you post the output of `bv_raw.info`?

hoechenberger:
> @hoechenberger Sent. Did you receive it?

Thank you, I received the email. Will look into this within the next 2 hours.

hoechenberger:
@arthurlgk I just looked at the `.vhdr` file you included, and there's something off.

You file looks something like this:
```
...
Ch1=Channel 1,,1,
Ch2=Channel 2,,1,
Ch3=Channel 3,,1,
...
```

while it's supposed to look like this:
```
...
Ch1=FP1,,0.5,µV
Ch2=FP2,,0.5,µV
Ch3=F3,,0.5,µV
...
```
i.e., there are no units specified in your file. I'm not sure if that's even valid (but I'm not an expert on BV data). How did you generate those files?

hoechenberger:
@sappelhoff Maybe you would know which unit is the default if not explicitly specified in the `vhdr` file? µV, I suppose?

sappelhoff:
yep, seems to me like that.

[...]

PR #8434, opened 2020-10-27
[MRG] BUG: Fix BrainVision reading for certain vhdr files

hoechenberger:
Fixes #8432

I need advice on how to best add a test (specifically, where to add new testing data. Essentially, I can re-use an existing data set and simply drop the "µV" units from the VHDR file)

PR #8471, opened 2020-11-03
[MRG] Add info where certain BV headers might come from

hoechenberger:
Based on https://github.com/mne-tools/mne-python/issues/8432#issuecomment-720903265


agramfort:
would it make sense to cover this with a test?

hoechenberger:
> would it make sense to cover this with a test?

It's already covered, this was added via https://github.com/mne-tools/mne-python/pull/8434
We just weren't sure where those weird files would actually come from, but now @arthurlgk told us that they used PyCorder to obtain these headers.

So this PR is really just a small clarification to remind us in the future which software might produce files that require special handling.

agramfort:
thx @hoechenberger 

I had read too quickly...

PR #8476, opened 2020-11-03
Add units param to get_data methods in Raw and Epochs

### LINKED ISSUE #8473: FR: Add "unit" parameter to get_data method for Epochs

---

```Python
```

```python
"""
Parameters
----------
"""
```

sappelhoff:
closes #8473 

Let me start by making a docstr proposal --> once we are clear on the functionality I can give it a try and implement it.

Should we link to https://mne.tools/stable/overview/implementation.html?highlight=units#id20 from the docstr?

### DIFF
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -798,7 +798,8 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
 
     @verbose
     def get_data(self, picks=None, start=0, stop=None,
-                 reject_by_annotation=None, return_times=False, verbose=None):
+                 reject_by_annotation=None, return_times=False, units=None,
+                 verbose=None):
         """Get data in the given range.
 
         Parameters
@@ -815,6 +816,19 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
             'bad' are omitted. If 'NaN', the bad samples are filled with NaNs.
         return_times : bool
             Whether to return times as well. Defaults to False.
+        units : str | dict | None
+            Specify the unit(s) that the data should be returned in. If None
+            (default), the data is returned in the channel type specific
+            default units. If str, must be a unit that will be used to scale
+            the data from all channel types generally compatible with that
+            unit. For example ``units="uV"`` will scale all channel types
+            that use Volts to micro Volts. Finally, if a dict is supplied,
+            keys must be channel types, and values must be units to scale
+            the data of that channel type to. For example
+            ``dict(eeg="nV")`` will scale all EEG channels to nano Volts,
+            but all other channel types will remain in their channel type
+            specific default unit.
+
         %(verbose_meth)s
 
         Returns
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -819,7 +819,9 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         units : str | dict | None
             Specify the unit(s) that the data should be returned in. If None
             (default), the data is returned in the channel type specific
-            default units. If str, must be a unit that will be used to scale
+            default units, which are SI units (see :ref:`channel-units` and
+            :term:`data channels`).
+            If str, must be an SI unit that will be used to scale
             the data from all channel types generally compatible with that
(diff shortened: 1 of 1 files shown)


larsoner:
> Should we link to https://mne.tools/stable/overview/implementation.html?highlight=units#id20 from the docstr?

Yes, feel free to add a proper RST target name to that header and link to it with `:ref:`. And also maybe link to
```
:term:`data_channels`
```
here https://mne.tools/dev/glossary.html since we also talk about SI units there a bit.

sappelhoff:
These are the units we use in MNE (are these *all* the units we use?) ([ref](https://mne.tools/stable/overview/implementation.html?highlight=units#id20))

```

    Volts (eeg, eog, seeg, emg, ecg, bio, ecog)

    Teslas (magnetometers)

    Teslas/meter (gradiometers)

    Amperes*meter (dipole fits, minimum-norm estimates, etc.)

    Moles/liter (“molar”; fNIRS data: oxyhemoglobin (hbo), deoxyhemoglobin (hbr))

    Arbitrary units (various derived unitless quantities)
```

I am a bit confused how we would handle derived units like `Teslas/meter`, or `Moles/liter`

would we allow scaling to `Teslas/millimeter`? If yes, how?

mne/io/base.py
```diff
@@ -801,6 +802,19 @@ def get_data(self, picks=None, start=0, stop=None,
             'bad' are omitted. If 'NaN', the bad samples are filled with NaNs.
         return_times : bool
             Whether to return times as well. Defaults to False.
+        units : str | dict | None
+            Specify the unit(s) that the data should be returned in. If None
+            (default), the data is returned in the channel type specific
+            default units. If str, must be a unit that will be used to scale
```
larsoner, line 808: Something like this probably
```suggestion
            default units, which are SI units (see :ref:`channel-units` and :term:`data channels`).
            If str, must be a unit that will be used to scale
```

sappelhoff:
finally, what "multiples" and "submultiples" should we allow? All those that are in this list?

https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html#prefixes

That'd be based on the CMIXF-12 format for representing units in ASCII

larsoner:
> would we allow scaling to Teslas/millimeter? If yes, how?

Yes. Somewhere we already parse fractional units and figure out what the scale factors are to get to SI units. Let me dig it up...

larsoner:
> These are the units we use in MNE (are these all the units we use?) (ref)

No, but it's probably "good enough" for most people as a reference. The complete list is probably provided by the FIF spec at the moment, because we expand it as we need to:

https://github.com/mne-tools/mne-python/blob/master/mne/io/constants.py#L786-L836

I would not worry about supporting all of these. Let's just get the data units we're likely to need. We can always expand later.

> would we allow scaling to Teslas/millimeter? If yes, how?

Refactor code from here, put it in `mne/utils/numerics.py` or something similar, reuse in the test and in your code:

https://github.com/mne-tools/mne-python/blob/master/mne/tests/test_defaults.py#L25

I would even just start with this list of prefixes and `known_SI`. It should have 99% of what you need (hopefully).

hoechenberger:
Just wondering if we can also allow µV in addition to uV, or would that become messy?

sappelhoff:
> Just wondering if we can also allow µV in addition to uV, or would that become messy?

I would suggest to allow `u` as a common ASCII abbreviation of micro, as well as the two (different!) unicode characters `µ` ([U+00B5](https://www.compart.com/de/unicode/U+00B5) and `μ` ([U+03BC](https://www.compart.com/de/unicode/U+03BC))

Also brings up this comment again where I was a little too slow to chime in that U+03BC is preferred over U+00B5 by the unicode consortium: https://github.com/mne-tools/mne-python/pull/7497#issuecomment-603494677

hoechenberger:
Is there any way to easily figure out which unicode characterI'm producing on macOS via Option-M (µ)?

sappelhoff:
See screenshot:

image

The first one is the micro sign, the second one is small greek mu

hoechenberger:
wow… thanks!

sappelhoff:
I thought I'd get to this PR this weekend, but unfortunately that won't happen.

I'll come back to this in a few weeks - if in the meantime somebody wants to pick it up, please be my guest.

sappelhoff:
closing in favor of #9136

PR #8474, opened 2020-11-03
Disable mandatorily downloading the testing data while setting up the egi_mff_pns_fname path.

Aniket-Pradhan:
#### Additional information
Any additional information you think is important.

Discussed this over Gitter with @larsoner ;D

PR #8477, opened 2020-11-03
MRG, TST: Need to skip properly

larsoner:
Fixes a bug pointed out by @Aniket-Pradhan on Gitter, restores our CI config such that it should catch the problems fixed by #8474, and catch the bugs fixed here.

Locally things pass with `MNE_SKIP_TESTING_DATASET_TESTS=true` so hopefully all is green here and we can merge and backport.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ matrix:
 
         # Minimal (runs with and without testing data)
         - os: linux
-          env: DEPS=minimial
+          env: DEPS=minimal
                CONDA_DEPENDENCIES="numpy scipy matplotlib"
                MNE_DONTWRITE_HOME=true MNE_FORCE_SERIAL=true MNE_SKIP_NETWORK_TEST=1 MNE_SKIP_TESTING_DATASET_TESTS=true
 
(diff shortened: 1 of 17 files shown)


Aniket-Pradhan:
Could you also verify if you get the same bug for this: [`mne/conftest.py:459: in src_volume_labels`](https://github.com/mne-tools/mne-python/blob/master/mne/conftest.py#L459)

larsoner:
> Could you also verify if you get the same bug for this: mne/conftest.py:459: in src_volume_labels

I do not. And it shouldn't occur because of the:
```
@pytest.mark.parametrize(params=[testing._pytest_param()])
```
You see a failure locally?

Once Travis catches up it should show up in https://travis-ci.org/github/mne-tools/mne-python/jobs/741178458

agramfort (approved):

larsoner:
> Could you also verify if you get the same bug for this: mne/conftest.py:459: in src_volume_labels

Travis replicated this -- looks like it's because `nibabel` is missing. Adding a `requires_nibabel()` should fix this -- I'll push a commit then hopefully all will be green!

PR #8482, opened 2020-11-04
add picks param in fix_stim_artifact

agramfort:
needed for one study where the stim artifact is also in the EOG channel


larsoner:
Thanks @agramfort

PR #8486, opened 2020-11-05
MRG: Better split file handling in Report.parse_folder()

hoechenberger:
Better split file handling in Reports: don't "report" on any splits but the first.

cc @agramfort, wdyt?

PR #8491, opened 2020-11-06
[MRG] FIX #8486, Evoked and Epochs don't get split

hoechenberger:
This should fix the error on Circle

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -103,7 +103,7 @@ Bugs
 
 - Attempting to remove baseline correction from preloaded `~mne.Epochs` will now raise an exception by `Richard Höchenberger`_ (:gh:`8435`)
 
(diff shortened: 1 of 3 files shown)


agramfort (approved):

hoechenberger:
One Travis run times out, but the other one is genuinely failing. I'm not sure what's happening, I cannot reproduce. It appears there's a side-effect somewhere, altering `raw_fname`??

hoechenberger:
@larsoner any idea?

larsoner:
https://github.com/mne-tools/mne-python/blob/master/mne/tests/test_report.py#L499

is missing the `@testing.requires_testing_data` decorator

hoechenberger:
D'oh!

larsoner:
Thanks @hoechenberger 

In the future when there is an error on CircleCI, it's good practice to make some minor change to the example in the PR that is supposed to fix it. That way in the PR we can see that the example runs without issue. (CircleCI for PRs only builds changed examples by default.)

hoechenberger:
Got it!

PR #8442, opened 2020-10-28
[MRG] DOC: Add new changelog template to contributors' guide

hoechenberger:
Also touch two changelog entries to see what it looks like when rendered.

Closes #8436

PR #8492, opened 2020-11-06
[MRG] FIX: Support commas in BrainVision channel names, event types, and event descriptions

sappelhoff:
This came up in https://github.com/bids-standard/pybv/pull/53

In the BrainVision format channel names, event types, and event descriptions that contain a comma `,` are encoded such that the comma is written as `\1`.

I add a test here to show that MNE-Python does currently not adequately replace the `\1` characters to `,` during reading.

This is potentially broken for event type and event description as well, but I haven't looked into that yet.

Should we fix it?


mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -272,6 +273,33 @@ def test_ascii(tmpdir):
...
+            continue
+        new_lines.append(line)
+
+    # Write temporary vhdr file for the test
+    tmpvhdr = vhdr_path.replace('.vhdr', '_comma-test.vhdr')
+    with open(tmpvhdr, 'w') as fout:
+        fout.writelines(lines)
+
+    raw = read_raw_brainvision(tmpvhdr)
+    os.remove(tmpvhdr)
```
larsoner, line 297: no need for this, `pytest` should clean up on its own

larsoner (commented):
> Should we fix it?

Yes I think fixing the channel names and the event type and descriptions would be good

mne/io/brainvision/tests/test_brainvision.py
sappelhoff: even without a fixture or anything?

larsoner: Oh sorry I assumed you were using a `tmpdir` fixture. You definitely should use `tmpdir` and `shutil.copyfile` instead of writing anything to the `vhdr_path` directory

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -646,7 +684,7 @@ def test_automatic_vmrk_sfreq_recovery():
 @testing.requires_testing_data
 def test_event_id_stability_when_save_and_fif_reload(tmpdir):
     """Test load events from brainvision annotations when read_raw_fif."""
-    fname = op.join(str(tmpdir), 'bv-raw.fif')
+    fname = tmpdir / 'bv-raw.fif'
```
sappelhoff, line 687: locally, this test failed, and I have no idea why.

sappelhoff: that is --> it does not fail because of my cosmetic change here, but because original_event_id and event_id are not equal. See:

```
(Pdb) event_id
{'Comma,Type/CommaValue,1': 1, 'Event/254': 2, 'New Segment/': 3, 'Optic/O  1': 4, 'Response/R255': 5, 'Stimulus/S253': 6, 'Stimulus/S255': 7, 'SyncStatus/Sync On': 8}
(Pdb) original_event_id
{'Comma,Type/CommaValue,1': 10001, 'Event/254': 254, 'New Segment/': 99999, 'Optic/O  1': 2001, 'Response/R255': 1255, 'Stimulus/S253': 253, 'Stimulus/S255': 255, 'SyncStatus/Sync On': 99998}
(Pdb) 
```

larsoner: It somehow has to do with your changes to `test.vmrk` above. Did you mean to revert that?

sappelhoff: No, that is a permanent change I wanted to add to the test file :thinking: but looking at the code, I also don't see why that would be a problem:

```Python
    fname = tmpdir / 'bv-raw.fif'
    raw = read_raw_brainvision(vhdr_path, eog=eog)
    original_events, original_event_id = events_from_annotations(raw)

    raw.save(fname)
    raw = read_raw_fif(fname)
    events, event_id = events_from_annotations(raw)

    assert event_id == original_event_id
    assert_array_equal(events, original_events)
```

1. we read the BV file
1. we get events from it (which works - even for the newly added event, as shown in the other test)
1. we save it to FIF
1. we read back from FIF
1. we get events from the just-read FIF annotations
1. ... they are different?

larsoner: Probably based on `event_id='auto'` default and `_check_bv_annot`, somehow your new event makes `_check_bv_annot` return False rather than True

larsoner: ... to avoid needing to deal with this in this PR, I would do the same trick as above, namely don't modify the actual `test.mrk` file, but copy it line-by-line and add an event at the bottom

sappelhoff: ahhh got it @larsoner ! Because of these lines:

https://github.com/mne-tools/mne-python/blob/fa7064a8d4cf8857d66ca001a17fa66e7121d3bc/mne/annotations.py#L956-L963

basically, the first raw is a "brainvision raw" ... and the second raw is a "fif raw" ... and the annotation-reading with event_id="auto" depends on the kind of raw format.

So passing an event_id other than "auto" at the second events_from_annotations call should fix it.

But is that what we want? The test confuses me a bit with it's goal.

sappelhoff: yup, `events, event_id = events_from_annotations(raw, original_event_id)` fixes it

sappelhoff: oh weird, I didn't see your comments :) should have refreshed to save me some detective work

larsoner: Don't do that. The goal of the test is that "standard BV events get parsed as BV events regardless of whether it's RawBrainvision or RawFIF". Passing the event_id defeats this purpose. Hence my suggestion to use the `tmpdir` + rewrite + append approach in the one test where you actually want to check for comma support rather than modifying `test.mrk` on disk directly in this PR

sappelhoff: I see, and introducing non-standard events obviously breaks the thing, reasonably so. Thanks!

sappelhoff:
Would this PR be eligible for a backport? It may come in handy for the pybv package.

larsoner:
Sure but we just cut 0.21.1 yesterday so it might be a few weeks or so until we do a 0.21.2

agramfort (approved):

larsoner (approved):

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -272,6 +272,42 @@ def test_ascii(tmpdir):
...
+
+    new_lines = []
+    for line in lines:
+        for to_replace, replacement in replace_dict.items():
+            if to_replace in line:
+                new = line.replace(to_replace, replacement)
+                new_lines.append(new)
+                break
+        else:
+            new_lines.append(line)
```
sappelhoff, line 301: The tests seem to be failing on Windows. Perhaps the manipulation of the test file prior to the test run is not working correctly on Windows?

Maybe this suggestion can help us to be sure:

```suggestion
    new_lines = []
    nperformed_replacements = 0
    for line in lines:
        for to_replace, replacement in replace_dict.items():
            if to_replace in line:
                new = line.replace(to_replace, replacement)
                new_lines.append(new)
                nperformed_replacements += 1
                break
        else:
            new_lines.append(line)
    assert nperformed_replacements == len(replace_dict)
```

sappelhoff:
indeed, on Windows, the temporary file-replacement for testing is not working as expected:

The windows tests only fail with:

```
mne\io\brainvision\tests\test_brainvision.py:304: in test_ch_names_comma
    assert nperformed_replacements == len(replace_dict)
E   assert 1 == 2
E     +1
E     -2
```

sappelhoff:
@cbrnr could you please have a look at this one and merge if you are happy? 

Then I can switch on the corresponding test on pybv with `@requires_version("mne", "0.22")`

doc/changes/latest.inc
```diff
@@ -105,6 +105,8 @@ Bugs
 
 - :meth:`mne.Report.parse_folder` will now correctly handle split FIFF files by `Richard Höchenberger`_ (:gh:`8486`)
 
+- Fix bug where BrainVision channel names, event types, and event descriptions containing commas were incorrectly parsed, `Stefan Appelhoff`_ (:gh:`8492`)
```
cbrnr, line 108:
```suggestion
- Fix bug where BrainVision channel names, event types, and event descriptions containing commas were incorrectly parsed (:gh:`8492` by `Stefan Appelhoff`_)
```

cbrnr: See #8442.

sappelhoff: done!

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -577,32 +618,42 @@ def test_read_vmrk_annotations(tmpdir):
...
     """Test load brainvision annotations and parse them to events."""
+    # First we add a custom event that contains a comma in its description
+    for src, dest in zip((vhdr_path, vmrk_path, eeg_path),
+                         ('test.vhdr', 'test.vmrk', 'test.eeg')):
+        shutil.copyfile(src, tmpdir / dest)
+
+    # Commas are encoded as "\1"
+    with open(tmpdir / 'test.vmrk', 'a') as fout:
+        fout.write(r"Mk15=Comma\1Type,CommaValue\11,7800,1,0")
+        fout.write("\n")
```
cbrnr, line 631: Why the extra line?

cbrnr (changes requested):

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -272,6 +273,46 @@ def test_ascii(tmpdir):
...
+                new = re.sub(to_replace, replacement, line)
+                new_lines.append(new)
+                nperformed_replacements += 1
+                break
+        else:
+            new_lines.append(line)
+    assert nperformed_replacements == len(replace_dict)
+
+    with open(comma_vhdr, 'w') as fout:
+        fout.writelines(new_lines)
```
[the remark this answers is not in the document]
sappelhoff: indeed, thanks for the catch! fixed in f3ac8a1e7bdbb53ad4eade8ee6a999727bbad1d8

mne/io/brainvision/tests/test_brainvision.py
sappelhoff: to end a file with an empty new line, which is the default on Linux and MacOS, but strictly speaking it would not be necessary in this case.

cbrnr: No, I think the `\n` could be combined with the previous line.

mne/io/brainvision/tests/test_brainvision.py
```diff
@@ -277,8 +277,8 @@ def test_ch_names_comma(tmpdir):
     """Test that channel names containing commas are properly read."""
     # commas in BV are encoded as \1
     replace_dict = {
-        r"^Ch4=F4,": r"^Ch4=F4\\1foo,",
-        r"^4\s\s\s\s\sF4": r"^4     F4,foo ",
+        r"^Ch4=F4,": r"Ch4=F4\\1foo,",
+        r"^4\s\s\s\s\sF4": r"4     F4,foo ",
```
cbrnr, line 281: `r"4     F4,foo "` doesn't need to be a raw string.

sappelhoff:
```suggestion
        r"^4\s\s\s\s\sF4": "4     F4,foo ",
```

mne/io/brainvision/tests/test_brainvision.py
sappelhoff, line 631:
```suggestion
        fout.write(r"Mk15=Comma\1Type,CommaValue\11,7800,1,0\n")
```

cbrnr (approved):

cbrnr:
I'll merge once CIs come back green.

cbrnr:
Not sure what's up with the [failing test](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=10040&view=logs&j=c0130b29-789d-5a3c-6978-10796a508a7f&t=b81cfa0d-2d5e-5a59-e316-838f621b71a0). It stopped after `mne/tests/test_report.py::test_render_add_sections[mayavi] PASSED        [ 75%]`, but it doesn't say which test failed and why.

sappelhoff:
I don't know either. 🤷‍♂️ but it passed before and it passes locally

cbrnr:
I'll re-run and see if this magically resolves the issue (it often does).

cbrnr:
Now Travis fails at the same test (`mne/tests/test_report.py::test_render_add_sections[pyvista] Traceback (most recent call last):`)? I'd like to clarify this before merging. @larsoner do you have any idea?

GuillaumeFavelier:
> Now Travis fails at the same test

It's related to https://github.com/mne-tools/mne-python/pull/8494

I'm working on a fix.

sappelhoff:
@cbrnr I rebased on Guillaume's fix and CIs pass now :-)

cbrnr:
Thanks @sappelhoff!

PR #8495, opened 2020-11-09
MAINT: Adjust to pyvista 0.27

GuillaumeFavelier:
I noticed 2 side effects since the update to `pyvista 0.27`:

1) `AttributeError` for `iren`

```
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/vtkmodules/qt/QVTKRenderWindowInteractor.py", line 335, in __getattr__

AttributeError: BackgroundPlotter has no attribute named iren
```

2) Screenshots are out of our tolerance range

```
mne/viz/_brain/tests/test_brain.py:241: in test_brain_init
    assert_allclose(img.shape, want_size)
E   AssertionError: 
E   
E   Mismatched elements: 2 / 3 (66.7%)
E   Max absolute difference: 40.
E   Max relative difference: 0.13333333
E    x: array([340, 318,   3])
E    y: array([300., 300.,   3.])
```

This PR fixes those issues. It's still a work in progress.

---

Also closes https://github.com/mne-tools/mne-python/issues/8505


GuillaumeFavelier:
The issue with the screenshot seems to be related to the fact that `plotter.image()`, the function that returns the current render window as an image, uses `self.window_size` instead of the actual `vtkRenderWindow` size. And it appears that in the case of `BackgroundPlotter`, `window_size` and `ren_win.GetSize()` are actually different (mainly because of UI).

Related to: https://github.com/pyvista/pyvistaqt/issues/64

GuillaumeFavelier:
Locally those changes did the trick for me, waiting for CIs

mne/viz/backends/_pyvista.py
```diff
@@ -825,9 +831,14 @@ def _take_3d_screenshot(figure, mode='rgb', filename=None):
...
         warnings.filterwarnings("ignore", category=FutureWarning)
         _process_events(figure.plotter)
-        return figure.plotter.screenshot(
+        # FIX: https://github.com/pyvista/pyvista/pull/995
+        old_window_size = figure.plotter.window_size
+        figure.plotter.window_size = figure.store["window_size"]
+        img = figure.plotter.screenshot(
             transparent_background=(mode == 'rgba'),
             filename=filename)
+        figure.plotter.window_size = old_window_size
```
larsoner, line 840: try/except, or not necessary here?

GuillaumeFavelier: For `window_size`? It exists now for a while. I don't think it's necessary.

larsoner: I mean if `screenshot` throws an error then` plotter.window_size` is never set back to its own value. So:
```
figure.plotter.window_size = ...
try:
    img = ...
finally:
    figure.plotter.window_size = old_window_size
```

GuillaumeFavelier:
Oh no... Now that `imge()` relies on `ren_win`, it's not possible to upsample by overriding `self.window_size`

larsoner:
>  it's not possible to upsample by overriding self.window_size

I didn't even know this is something that we supported. Do we even expose an option for this? It might be nice to have in the future but I don't think it's critical

GuillaumeFavelier:
Bad choice of words, I meant "resize" instead of "upsample", my bad

GuillaumeFavelier:
Okay so once https://github.com/pyvista/pyvista/pull/999 is merged, I'll try again with `window_size` and hopefully even `Travis PIP PRE` will come back happy.

GuillaumeFavelier:
Nevermind, it's been closed in favour https://github.com/pyvista/pyvista/pull/998

GuillaumeFavelier:
Now I obtain:

```
________________________________ test_stc_mpl _________________________________
mne\viz\tests\test_3d.py:499: in test_stc_mpl
    stc.plot(subjects_dir=subjects_dir, time_unit='s', views='ven',
mne\source_estimate.py:653: in plot
    brain = plot_source_estimates(
<decorator-gen-139>:24: in plot_source_estimates
    ???
mne\viz\_3d.py:1803: in plot_source_estimates
    return _plot_mpl_stc(stc, spacing=spacing, **kwargs)
mne\viz\_3d.py:1529: in _plot_mpl_stc
    _smooth_plot(initial_time, params)
mne\viz\_3d.py:1437: in _smooth_plot
    facecolors = polyc._facecolors3d
E   AttributeError: 'Poly3DCollection' object has no attribute '_facecolors3d'
```

which does not seem to appear on `master`.

GuillaumeFavelier:
Locally, I can reproduce after upgrading to `matplotlib-3.3.3`.

GuillaumeFavelier:
Oh no...

image

mne/viz/_3d.py
```diff
@@ -1434,7 +1436,10 @@ def _smooth_plot(this_time, params):
     color_ave = np.mean(colors[faces], axis=1).flatten()
     curv_ave = np.mean(params['curv'][faces], axis=1).flatten()
     # matplotlib/matplotlib#11877
-    facecolors = polyc._facecolors3d
+    if LooseVersion(matplotlib_version) < '3.3.3':
+        facecolors = polyc._facecolors3d
+    else:
+        facecolors = polyc._facecolor3d
```
larsoner, line 1442: Now would be a good time to abandon the private attribute all together. I suggest we do `get_facecolor()` then `set_facecolor()`. It will perhaps be slower than in-place setting, but probably not a meaningful slowdown compared to the actual matplotlib 3D rendering speed.

Can you see if it works?

GuillaumeFavelier: With the naive:

```py
    facecolors = polyc.get_facecolor()
    # create colors here
    facecolors[:] = colors
    polyc.set_facecolor(facecolors)
```

Locally I see:

```
mne/viz/_3d.py:1530: in _plot_mpl_stc
    _smooth_plot(initial_time, params)
mne/viz/_3d.py:1437: in _smooth_plot
    facecolors = polyc.get_facecolor()
../anaconda3/envs/mne/lib/python3.8/site-packages/mpl_toolkits/mplot3d/art3d.py:817: in get_facecolor
    return self._facecolors2d
E   AttributeError: 'Poly3DCollection' object has no attribute '_facecolors2d'
```

larsoner: `polyc.set_facecolors(colors)` without any `get_facecolor` or `get_facecolors` call seems to at least not kill the tests.

larsoner:
I am trying to test the `matpltolib` backend with:
```
import os.path as op
import mne
import time
    evoked, inv, return_residual=True, pick_ori=None)

initial_time = 0.1
brain = stc.plot(
    clim='auto',
    backend='matplotlib',
    hemi='lh', views='lat', smoothing_steps=5,
    view_layout='horizontal', verbose=True,
    time_viewer=True, background='k')
```
But I get nothing on `master` or this PR:

Screenshot from 2020-11-12 08-04-47

Hence my complaints about the `matplotlib` backend being not properly tested (or even used?)...

larsoner:
In the meantime I pushed a tiny fix for `nearest` that I hit that we'll want as well

PR #8498, opened 2020-11-09
[DOC] add link to resampling FAQ to all resampling methods

sappelhoff:
Came up in Gitter: [November 9, 2020 6:40 PM](https://gitter.im/mne-tools/mne-python?at=5fa97f0fd37a1a13d6975c37)

(@cbrnr feel free to push to this branch if you want to make this PR bigger in scope.)

For "resampling", MNE automatically applies anti-alias filters where appropriate. This small doc addition will clarify the situation for many users.

I was unsure about it personally, and tried looking into the source code ... but even from there it's not immediately obvious that anti-alias filtering is done. It seems to me like most of the magic is happening here:

I am still not sure that the filtering is truly done - but it may happen in the frequency domain code somehow.

https://github.com/mne-tools/mne-python/blob/bb16a3e6ea99024e468fa84c2ec55c344b47df7e/mne/cuda.py#L291-L340

### DIFF
--- a/mne/filter.py
+++ b/mne/filter.py
@@ -2018,6 +2018,9 @@ class FilterMixin(object):
                  pad='edge', verbose=None):  # lgtm
         """Resample data.
 
+        If appropriate, an anti-aliasing filter is applied before resampling.
+        See :ref:`resampling-and-decimating` for more information.
+
         .. note:: Data must be loaded.
 
         Parameters
--- a/mne/filter.py
+++ b/mne/filter.py
@@ -2018,7 +2018,9 @@ class FilterMixin(object):
                  pad='edge', verbose=None):  # lgtm
         """Resample data.
 
-        If appropriate, an anti-aliasing filter is applied before resampling.
+        Through use of an FFT-based (fast Fourier transform) method for
+        resampling, aliasing artifacts are prevented as if an anti-aliasing
+        filter were applied beforehand.
         See :ref:`resampling-and-decimating` for more information.
 
         .. note:: Data must be loaded.
--- a/mne/filter.py
+++ b/mne/filter.py
@@ -2018,9 +2018,7 @@ class FilterMixin(object):
                  pad='edge', verbose=None):  # lgtm
         """Resample data.
 
-        Through use of an FFT-based (fast Fourier transform) method for
-        resampling, aliasing artifacts are prevented as if an anti-aliasing
-        filter were applied beforehand.
+        If appropriate, an anti-aliasing filter is applied before resampling.
         See :ref:`resampling-and-decimating` for more information.
 
         .. note:: Data must be loaded.
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1058,9 +1058,7 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                  verbose=None):  # lgtm
         """Resample all channels.
 
-        Through use of an FFT-based (fast Fourier transform) method for
-        resampling, aliasing artifacts are prevented as if an anti-aliasing
-        filter were applied beforehand.
+        If appropriate, an anti-aliasing filter is applied before resampling.
         See :ref:`resampling-and-decimating` for more information.
 
         .. warning:: The intended purpose of this function is primarily to
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1058,7 +1058,9 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                  verbose=None):  # lgtm
         """Resample all channels.
 
-        If appropriate, an anti-aliasing filter is applied before resampling.
+        Through use of an FFT-based (fast Fourier transform) method for
+        resampling, aliasing artifacts are prevented as if an anti-aliasing
+        filter were applied beforehand.
         See :ref:`resampling-and-decimating` for more information.
 
         .. warning:: The intended purpose of this function is primarily to
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1058,6 +1058,9 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                  verbose=None):  # lgtm
         """Resample all channels.
 
+        If appropriate, an anti-aliasing filter is applied before resampling.
+        See :ref:`resampling-and-decimating` for more information.
+
         .. warning:: The intended purpose of this function is primarily to
                      speed up computations (e.g., projection calculation) when
                      precise timing of events is not required, as downsampling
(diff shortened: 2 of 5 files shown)


larsoner:
> I am still not sure that the filtering is truly done - but it may happen in the frequency domain code somehow.

Yes it's effectively a brick-wall filter at `min(fs_old / 2., fs_new / 2.)` based on how the FFT and IFFT are done

cbrnr:
Here's an explanation of how FFT-based resampling works: https://dsp.stackexchange.com/a/52943 The answer also links to a more theoretical explanation, but the gist is that the method accounts for aliasing (so you don't need to manually lowpass filter before).

cbrnr:
So technically the statement that MNE adds an anti-aliasing filter is not correct because this is not a separate step. We could say that resampling as implemented by the FFT-based method prevents aliasing or something along these lines.

mne/filter.py
```diff
@@ -2018,6 +2018,9 @@ def resample(self, sfreq, npad='auto', window='boxcar', n_jobs=1,
                  pad='edge', verbose=None):  # lgtm
         """Resample data.
 
+        If appropriate, an anti-aliasing filter is applied before resampling.
+        See :ref:`resampling-and-decimating` for more information.
```
sappelhoff, line 2022: What do you think about this one @cbrnr?

```suggestion
        Through use of an FFT-based (fast Fourier transform) method for
        resampling, aliasing artifacts are prevented as if an anti-aliasing
        filter were applied beforehand.
        See :ref:`resampling-and-decimating` for more information.
```

cbrnr (approved):

mne/io/base.py
```diff
@@ -1058,6 +1058,9 @@ def resample(self, sfreq, npad='auto', window='boxcar', stim_picks=None,
                  verbose=None):  # lgtm
         """Resample all channels.
 
+        If appropriate, an anti-aliasing filter is applied before resampling.
+        See :ref:`resampling-and-decimating` for more information.
```
sappelhoff, line 1062:
```suggestion
        Through use of an FFT-based (fast Fourier transform) method for
        resampling, aliasing artifacts are prevented as if an anti-aliasing
        filter were applied beforehand.
        See :ref:`resampling-and-decimating` for more information.
```

mne/source_estimate.py
```diff
@@ -712,6 +712,9 @@ def resample(self, sfreq, npad='auto', window='boxcar', n_jobs=1,
                  verbose=None):
         """Resample data.
 
+        If appropriate, an anti-aliasing filter is applied before resampling.
+        See :ref:`resampling-and-decimating` for more information.
```
sappelhoff, line 716:
```suggestion
        Through use of an FFT-based (fast Fourier transform) method for
        resampling, aliasing artifacts are prevented as if an anti-aliasing
        filter were applied beforehand.
        See :ref:`resampling-and-decimating` for more information.
```

sappelhoff (commented):
Good, then I'll use this one everywhere:

```
        Through use of an FFT-based (fast Fourier transform) method for
        resampling, aliasing artifacts are prevented as if an anti-aliasing
        filter were applied beforehand.
        See :ref:`resampling-and-decimating` for more information.
```

mne/filter.py
```diff
@@ -2018,7 +2018,9 @@ def resample(self, sfreq, npad='auto', window='boxcar', n_jobs=1,
                  pad='edge', verbose=None):  # lgtm
         """Resample data.
 
-        If appropriate, an anti-aliasing filter is applied before resampling.
+        Through use of an FFT-based (fast Fourier transform) method for
+        resampling, aliasing artifacts are prevented as if an anti-aliasing
+        filter were applied beforehand.
```
larsoner, line 2023: I would revert this. A brick-wall frequency-domain filter is an anti-aliasing filter. You can think of it as being equivalent to a time-domain (circular) convolution with a sinc if you want.

sappelhoff: cc @cbrnr you want to argue against this?

Else, I'll do as Eric says and go with my previous suggestion

cbrnr: I guess the important thing is to let people know if they need to manually apply an anti-aliasing filter or not. Typically, most people think of a lowpass filter to do that, in this is not necessary here. I think I didn't like the previous sentence because you said that "an anti-aliasing filter is applied *before* resampling" - I would argue that it's not applied before, because it is part of the FFT/IFFT procedure.

sappelhoff: How about:

```suggestion
        To achieve anti-aliasing, this function applies a brick-wall
        frequency-domain filter during the FFT-based
        (fast Fourier transform) resampling.
```

larsoner: > So technically the statement that MNE adds an anti-aliasing filter is not correct because this is not a separate step.

Quoting the link posted earlier:

> Alternatively, you can see your approach as filtering with an aliasing filter (i.e. lowpass with cutoff 1/M) before downsampling.

The fact that it's not manually/separately applied *in the code* is an implementation detail. It's mathematically equivalent to doing it, so we should let people sleep better at night by saying that an anti-aliasing filter is applied to their data. To me this is similar to not needing to tell people that we do overlap-add frequency-domain convolution when we filter, we can just say that the output is the convolution of the signal with the FIR filter because it's equivalent. (And FWIW actually some years ago we didn't use overlap-add or direct convolution but instead a one-shot full frequency-domain convolution method, too. But people didn't need to care when we switched, and we only bothered documenting it because it was a speed improvement.)

There are things we can and probably should discuss somewhere about these methods -- such as the extent to which using a circular convolution with a sinc / freq-domain brick-wall is risky -- but I wouldn't equivocate about whether or not an anti-aliasing filter is applied. The natural question when reading the new text is, "so wait, no anti-aliasing filter is applied? So there can be aliasing?" It's just going to needlessly make people worry.

larsoner: +1 for the brick-wall language, it's explicit, technically correct, and should let people sleep :)

larsoner: (though technically the `window` argument allows things other than a strict brick-wall filter -- the default `'boxcar'` just so happens to be the brick wall)

cbrnr: I agree that these are really technical details at the algorithm level, but then at least remove the word "before" to let me sleep better at night :smile:.

Also, when people go inspect the code and don't find the anti-aliasing filter (because they are expecting that some filter is applied before resampling), it would be good to discuss the FFT-based method somewhere to explain that it incorporates an anti-aliasing filter. I'd be interested in this, e.g. I haven't bothered to get into the details such as how do we handle cases where signal lengths are prime (do we accept that resampling will be slow then, or would it be better to switch to a polyphase implementation you mentioned before), and so on.

Anyway, I let you decide on the wording @larsoner.

larsoner (approved):

agramfort (approved):

larsoner:
Thanks @sappelhoff !

PR #8500, opened 2020-11-09
MRG, DOC: epochs getitem clarification

drammock:
@sappelhoff does this accord with the behavior you're seeing?

### DIFF
--- a/mne/utils/mixin.py
+++ b/mne/utils/mixin.py
@@ -139,7 +139,7 @@ class GetEpochsMixin(object):
            Query-based indexing only works if Pandas is installed and
            ``self.metadata`` is a :class:`pandas.DataFrame`.
 
-               .. versionadded:: 0.16
+           .. versionadded:: 0.16
         """
         return self._getitem(item)
 
--- a/mne/utils/mixin.py
+++ b/mne/utils/mixin.py
@@ -92,49 +92,52 @@ class GetEpochsMixin(object):
         -----
         Epochs can be accessed as ``epochs[...]`` in several ways:
 
-            1. **integer or slice:** ``epochs[idx]`` will return an
-               `~mne.Epochs` object with a subset of epochs chosen by index
-               (supports single index and Python-style slicing).
+        1. **Integer or slice:** ``epochs[idx]`` will return an `~mne.Epochs`
+           object with a subset of epochs chosen by index (supports single
+           index and Python-style slicing).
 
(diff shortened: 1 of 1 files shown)


mne/utils/mixin.py
```diff
@@ -109,12 +109,18 @@ def __getitem__(self, item):
...
-               If conditions are separated by '/', selects every item
-               containing every list tag (e.g. ['audio', 'left'] selects
-               'audio/left' and 'audio/center/left', but not 'audio/right').
+               experimental conditions. The list of names is treated as an
+               inclusive-or condition; if *none* of the provided names match
+               any epoch labels, a ``KeyError`` will be raised.
+
+               If epoch labels are tags separated by '/', then providing
+               multiple tags as *separate list entries* (e.g.,
+               ``epochs[['audio', 'left']]``) will also act as an inclusive-or
```
agramfort, line 118: isn't that repeating what is just said above?

sappelhoff: I think @drammock means to say that `[["left", "audio"]]` would work for both kinds of epochs:

1. those that use the `/` style: `"audio/left"`
1. those that have just single keywords: `"audio"` for one epoch, `"left"` for another epoch

mne/utils/mixin.py
```diff
@@ -109,12 +109,18 @@ def __getitem__(self, item):
...
+               filter. If multiple tags are provided as *a single list entry*
+               (e.g., ``epochs['name_1/name_2']``), this selects every item
+               containing every list tag. For example, ``epochs['audio/left']``
+               selects 'audio/left' and 'audio/center/left', but not
+               'audio/right'.
```
sappelhoff, line 123: I think I would refer to "logical and" here, like you refer to the "logical inclusive-or" above.

sappelhoff (commented):
Thanks, LGTM overall.

Minor: Further up in the example, the events are referred to as "auditory", later then as "audio" --> would be good to make this consistent.

agramfort (approved):

drammock:
@sappelhoff OK for you?  Here's the rendered doc https://23418-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Epochs.html#mne.Epochs.__getitem__

(note that I pushed a fix for `versionadded` being indented too far, but that's not reflected in the rendered doc linked above)

sappelhoff (approved):
Thanks @drammock - lgtm!

PR #8506, opened 2020-11-12
update pymatreader to version 0.0.24

thht:
pymatreader received a fix to properly load mat files of version 4.


agramfort:
@larsoner are CI issues expected?

besides diff looks good

larsoner:
Yes there are currently four breaking changes made by packages (matplotlib, dipy, numpy, pyvista) in a 24 span :)

These are fixed in #8495

larsoner:
Thanks @thht

PR #8515, opened 2020-11-13
MRG, ENH: Improve MATLAB compat

### LINKED ISSUE #8513: Option to disable tqdm (crashes on Matlab/Windows10)
The Matlab-Python integration works beautifully for most applications now, however there are some glitches left, especially in the way the standard streams are handled on Windows. As known unsupported features, Mathworks lists `py.input` and `py.raw_input`: https://fr.mathworks.com/help/matlab/matlab_external/limitations-to-python-support.html
But I just found out that there are also other things that do not work correctly, like the tqdm progress bars.

Minimal example to reproduce the bug, suggested by @larsoner : 
```
clear all; close all; home
tqdm = py.importlib.import_module('tqdm');
cell(py.list(tqdm.tqdm(py.str('10 11 12').split())))
```

It works on WSL(Win10)/Ubuntu18 + Matlab 2018b + Python 3.6 + tqdm 4.51.0:
```
ans =
  1×3 cell array
    {1×2 py.str}    {1×2 py.str}    {1×2 py.str}
```

It fails on Windows10 + Matlab 2019a/2020a/2020b + Python 3.8 + tqdm 4.51.0:
```
>> cell(py.list(tqdm.tqdm(py.str('10 11 12').split())))
Error using utils>__getattr__ (line 169)
Python Error: AttributeError: 'NoneType' object has no attribute 'write'
Error in std>fp_write (line 329)
Error in std>print_status (line 336)
Error in std>display (line 1520)
Error in std>refresh (line 1386)
Error in std>__init__ (line 1092)
```

This is most likely due to a bug in MATLAB on Windows, therefore you could decide that it is 2x not your problem :)
But since there are tons of forum posts about broken tqdm progress bars in various non-tty contexts, I think it might be a relevant question even for non-matlab/non-windows users.
How complicated for you would it be to add a global flag to disable tqdm in MNE-Python? 
@larsoner suggested something like `MNE_TQDM='off'` on gitter. 

Note that I'm using CPython instead of Anaconda because the integration with Matlab is much smoother. Matlab+CPython really works out-of-the-box now. FYI: the notes I'm preparing as I move forward:
https://neuroimage.usc.edu/brainstorm/MnePython

As a side request: it would be great to add an option to skip the validation of the licenses when downloading, and not via `argv`. I see that there is an option to accept the license with a flag `--accept-brainstorm-license` on the command line, but this means calling python via a system call, which is not exactly what we want here (the python environment configured in Matlab is not necessarily the one available from the system path of the process)
And same goes for anything that uses `input()`, actually.

Thanks!
Francois

larsoner:
Closes #8513

@ftadel can you give this branch a shot? Marking as `backport-candidate` so that we can backport then release 0.21.2 to make things easier once we get everything working (hopefully quickly!).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -57,6 +57,8 @@ Enhancements
 
 - Further improved documentation building instructions and execution on Windows (:gh:`8502` by `kalenkovich`_ and `Eric Larson`_)
 
+- Add option to disable TQDM entirely with ``MNE_TQDM='off'`` (:gh:`8515` by `Eric Larson`_)
+
 - `mne.preprocessing.ICA.plot_sources` now displays an `mne.preprocessing.ICA.plot_properties` window when right-clicking on component names on the y-axis (:gh:`8381` by `Daniel McCloy`_)
 
 
--- a/mne/datasets/brainstorm/bst_auditory.py
+++ b/mne/datasets/brainstorm/bst_auditory.py
@@ -6,7 +6,7 @@ from functools import partial
 
 from ...utils import verbose
 from ..utils import (has_dataset, _data_path, _get_version, _version_doc,
-                     _data_path_doc)
+                     _data_path_doc_accept)
 
 has_brainstorm_data = partial(has_dataset, name='brainstorm.bst_auditory')
 
@@ -29,15 +29,15 @@ URL: http://neuroimage.usc.edu/brainstorm/DatasetAuditory
 
 @verbose
 def data_path(path=None, force_update=False, update_path=True, download=True,
-              verbose=None):  # noqa: D103
+              *, accept=False, verbose=None):  # noqa: D103
     return _data_path(path=path, force_update=force_update,
                       update_path=update_path, name='brainstorm',
-                      download=download, archive_name='bst_auditory.tar.gz')
+                      download=download, archive_name='bst_auditory.tar.gz',
+                      accept=accept)
 
 
-_data_path_doc = _data_path_doc.format(name='brainstorm',
-                                       conf='MNE_DATASETS_BRAINSTORM_DATA'
-                                            '_PATH')
+_data_path_doc = _data_path_doc_accept.format(
+    name='brainstorm', conf='MNE_DATASETS_BRAINSTORM_DATA_PATH')
 _data_path_doc = _data_path_doc.replace('brainstorm dataset',
                                         'brainstorm (bst_auditory) dataset')
 data_path.__doc__ = _data_path_doc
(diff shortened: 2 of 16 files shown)


larsoner:
CircleCI passed the download stage so I killed it as that's all I really wanted to check -- it's going to die until #8516 lands anyway

ftadel:
It seems to be working like a charm!
Now it's a bit frustrating not to be able to follow the progression of the task at all, but at least it works!
We'll explore replacement solutions when we have time, or when Matlab gets fixed.

ftadel:
Now, I have a problem with joblib:

The example I ran, from @dimitriospantazis:
```
clf = py.sklearn.pipeline.make_pipeline(py.sklearn.preprocessing.StandardScaler(), py.sklearn.svm.SVC(pyargs('kernel', 'linear', 'C', 1)));
time_decod = py.mne.decoding.SlidingEstimator(clf, pyargs('n_jobs', int32(1), 'scoring', 'roc_auc', 'verbose', false));
scores = py.mne.decoding.cross_val_multiscore(time_decod, X, y, pyargs('cv', int32(5), 'n_jobs', int32(1), 'verbose', false));
```

It crashes if I set `n_jobs` to 4:
```
>> scores = py.mne.decoding.cross_val_multiscore(time_decod, X, y, pyargs('cv', int32(5), 'n_jobs', int32(4), 'verbose', false));
Error using parallel>_print (line 876)
Python Error: AttributeError: 'NoneType' object has no attribute 'write'
Error in parallel>__call__ (line 1006)
Error in parallel>run (line 128)
Error in base>cross_val_multiscore (line 441)
```

Together with this Matlab error:
image

I haven't investigated this at all. Posting it just in case you already have an idea of the issue.
Maybe it won't be possible to run any parallel processing from Matlab?...

larsoner:
@ftadel I'm not sure, there are at least some hints here about interoperability (but it's using MATLAB from Python):

https://stackoverflow.com/questions/55424095/error-pickling-a-matlab-object-in-joblib-parallel-context

I think we'll need to come up with a minimal joblib example and bother the joblib folks with it. It's possible that some backend other than `loky` would work.

agramfort:
to fix joblib you can use this:

https://github.com/mne-tools/mne-python/blob/master/mne/parallel.py#L15

basically add MNE_FORCE_SERIAL to environment variables

>

ftadel:
> basically add MNE_FORCE_SERIAL to environment variables

That's a good start, thanks. At least we can make sure everything will work without parallelization issues.

> I think we'll need to come up with a minimal joblib example and bother the joblib folks

I'll work on this soon.
Thanks for your help!

When can we expect this PR to be in the "current version of MNE-Python that people get with `pip install mne`"?

larsoner:
@ftadel I can push out a 0.21.2 to PyPi on Monday

ftadel:
>  I can push out a 0.21.2 to PyPi on Monday

You're incredibly reactive!! Thanks :)

> I think we'll need to come up with a minimal joblib example and bother the joblib folks

https://github.com/joblib/joblib/issues/1131

larsoner:
0.21.2 is out

https://pypi.org/project/mne/0.21.2/

PR #8514, opened 2020-11-13
WIP: Migrate Travis jobs to GitHub Actions

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/8335#issuecomment-726761191 and migrates all remaining jobs on Travis to GitHub Actions.

ToDo:

- [x] Migrate Travis jobs
- [x] Remove travis conf file
- [x]  Try separate jobs to fix `compat / old` job
- [x] Add inline annotations for codespell and flake (suggested in https://github.com/mne-tools/mne-python/pull/8335#issuecomment-726761191)
- [x] Remove `TRAVIS_OS_NAME` variable
- [x] Add `[skip azp]` mechanism (suggested in https://github.com/mne-tools/mne-python/pull/8514#issuecomment-726797122)

This is still work in progress.

PR #8516, opened 2020-11-13
FIX: Circle failures

drammock:
Fixes the CircleCI failures introduced by #8381

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -1773,9 +1773,9 @@ class MNEBrowseFigure(MNEFigure):
         if 'raw' in (self.mne.instance_type, self.mne.ica_type):
             return self.mne.inst[:, start:stop]
         else:
-            if not self.mne.inst.preload:
-                self.mne.inst.load_data()
-            data = np.concatenate(self.mne.inst._data, axis=-1)[:, start:stop]
+            ix = np.searchsorted(self.mne.boundary_times, self.mne.t_start)
+            item = slice(ix, ix + self.mne.n_epochs)
+            data = np.concatenate(self.mne.inst.get_data(item=item), axis=-1)
             times = np.arange(len(self.mne.inst) * len(self.mne.inst.times)
                               )[start:stop] / self.mne.info['sfreq']
             return data, times
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -1773,6 +1773,8 @@ class MNEBrowseFigure(MNEFigure):
         if 'raw' in (self.mne.instance_type, self.mne.ica_type):
             return self.mne.inst[:, start:stop]
         else:
+            if not self.mne.inst.preload:
+                self.mne.inst.load_data()
             data = np.concatenate(self.mne.inst._data, axis=-1)[:, start:stop]
             times = np.arange(len(self.mne.inst) * len(self.mne.inst.times)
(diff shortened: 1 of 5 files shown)


larsoner:
You are too quick -- I should have checked all of my emails before writing about the CircleCI failure, sorry about that! :)

agramfort (approved):

mne/viz/_figure.py
```diff
@@ -1773,6 +1773,8 @@ def _load_data(self, start=None, stop=None):
         if 'raw' in (self.mne.instance_type, self.mne.ica_type):
             return self.mne.inst[:, start:stop]
         else:
+            if not self.mne.inst.preload:
+                self.mne.inst.load_data()
```
larsoner, line 1777: Better than this would probably be `get_data` geting just the part you need. It can drastically reduce memory requirements. I think it's worth going through the indexing gymnastics to figure it out.

In particular you probably need to do something like `mne.inst.drop_bad()` then look at `len(inst)` and `len(epochs.times)` and figure out which epochs you need, `get_data` for them, then concat, then subselect (trim beginning/end where necessary).

Then in your smoke test below you can `assert epochs._data is None` before and after the `plot` call

drammock: @larsoner I've done as you suggest. Only downside is that now the console prints 

```
Loading data for 5 events and 421 original time points ...
Loading data for 6 events and 421 original time points ...
Loading data for 7 events and 421 original time points ...
Loading data for 8 events and 421 original time points ...
```
every time you scroll or change the number of visible channels. I can live with it, but might be worth considering whether `epochs.get_data` should get a `verbose` kwarg (in a separate PR).

larsoner: You can also just do `with mne.use_log_level('error'):` or `with mne.use_log_level('warn')` here if you want. It does the same thing as `verbose` actually. And even adding `verbose` is a pretty trivial change so you can make it here.

But I'm fine with another PR, too -- I'll merge if you don't push a commit before Circle comes back green :)

larsoner: ... looks like that time has already passed

larsoner: ... and actually it looks like I should at least wait for Azure to come back happy so that the `viz` tests run

drammock:
[Azure failure](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=10177&view=logs&j=a846d25a-e32c-5640-1b53-e815fab94407&t=c40cce3b-0d9d-58ac-d4ba-617876a7e096&l=1402)
```
forrtl: error (200): program aborting due to control-C event
Image              PC                Routine            Line        Source             
libifcoremd.dll    00007FFD4A663B58  Unknown               Unknown  Unknown
KERNELBASE.dll     00007FFD7DDC7D33  Unknown               Unknown  Unknown
KERNEL32.DLL       00007FFD7FDD84D4  Unknown               Unknown  Unknown
ntdll.dll          00007FFD80DC1791  Unknown               Unknown  Unknown
mne/tests/test_epochs.py::test_concatenate_epochs Terminate batch job (Y/N)? 
##[error]The operation was canceled.
```

any idea what went wrong there?

larsoner:
That's just a timeout and we can ignore it. @GuillaumeFavelier is fixing it in #8514

larsoner:
Thanks @drammock

PR #8519, opened 2020-11-13
WIP: added colors option to grow_labels

### LINKED ISSUE #8517: Spherical searchlights as labels for all vertices in source space
I would like to use spherical searchlights as labels for all vertices in source space to compute PSFs/CTFs for sources with spatial extent (rather than point sources). I tried to use mne.grow_labels(), but it produces the error "NotImplementedError: Can't produce more than 1024 unique colors". 

Could one implement an option like '''colors=False''' to prevent the error message?

Possible workaround: One can use list comprehension to call grow_labels per vertex/seed, but I don't see a reason to limit it to 1024 colors.

olafhauk:
Closes #8517

Added ```colors``` option to mne.grow_label() to avoid error message for large number of seeds.


agramfort:
can you update a test?

mne/label.py
```diff
@@ -1615,9 +1619,14 @@ def grow_labels(subject, seeds, extents, hemis, subjects_dir=None, n_jobs=1,
...
 
-    # add a unique color to each label
-    colors = _n_colors(len(labels))
-    for label, color in zip(labels, colors):
+    if colors is None:
+        # add a unique color to each label
+        label_colors = _n_colors(len(labels))
+    else:
+        # use specified colors
+        label_colors = colors
```
larsoner, line 1627: There is no guarantee or check that `label_colors` is the right size. I think you had the right approach before, namely to do:
```
label_colors = np.empty((len(labels), 3))
label_colors[:] = colors
```
This guarantees that whatever `colors` is, it can broadcast and you will end up with `label_colors` of the right shape.

olafhauk: Fine by me, but _n_colors() (label.py) returns array, shape (n, 4).

larsoner: Then you can do `label_colors = np.ones((len(labels), 4)); label_colors[:, :3] = label_colors` if you want. I doubt anyone wants an alpha < 1

olafhauk: > 
> 
> Then you can do `label_colors = np.ones((len(labels), 4)); label_colors[:, :3] = label_colors` if you want. I doubt anyone wants an alpha < 1

I'm now allowing four columns and added info about RGB/alpha values to doc-string.

larsoner:
LGTM -- can you update `latest.inc` and remove WIP from the title if you're happy and we'll merge? Thanks @olafhauk

mne/label.py
```diff
@@ -1615,9 +1620,15 @@ def grow_labels(subject, seeds, extents, hemis, subjects_dir=None, n_jobs=1,
...
-    # add a unique color to each label
-    colors = _n_colors(len(labels))
-    for label, color in zip(labels, colors):
+    if colors is None:
+        # add a unique color to each label
+        label_colors = _n_colors(len(labels))
+    else:
+        # use specified colors
+        label_colors = np.empty((len(labels), 4))
+        label_colors[:] = colors
```
agramfort, line 1629: `label_colors = np.array(colors)`

seems easier and equivalent.

also do you test somewhere that `len(colors) == len(labels)` ?

larsoner: No it's not equivalent. The point of doing the `np.empty` is that

1. It allows broadcasting like `colors=0.` if you don't care about the color they can all be black, or `color=(1., 0, 0)` to make them all red, and 
2. It automatically makes sure `len(labels) == len(colors)`

agramfort: ok then. +1 for MRG if CIs are green.

thx @olafhauk !

olafhauk: ```colors``` can also be of shape (, 4) (one color for all labels). I amended doc-string and added checks/tests.

mne/source_space.py
```diff
@@ -3197,3 +3197,110 @@ def _get_src_nn(s, use_cps=True, vertices=None):
     else:
         nn = s['nn'][vertices, :]
     return nn
+
+
+def vertex_depth(inst, info=None, picks=None, trans=None, mode='dist',
```
larsoner, line 3202: Errant push of `vertex_depth`?

olafhauk: Oh no.

olafhauk: I never know what to do about the codecov errors. Otherwise I think it's fine now.

agramfort:
Thx @olafhauk !

PR #8521, opened 2020-11-13
MRG, MAINT: Refactor volume interpolation code

larsoner:
I am still annoyed by how slow the volumetric morphing is, so I decided to see if I could alleviate one bottleneck, namely the linear and zero-order interpolations. I never 100% understood Matti's volume interpolation code I adapted so many years ago, but based on how it works it should definitely be faster when we have lots of volumes to morph. Whether it's already the same speed as dipy and nibabel (scipy.ndimage) or faster with 1, 10, 100, or 1000 volumes remains to be determined, but I wouldn't be surprised if it were about the same speed with even just 1. It definitely will be with 1000.

In any case, to move toward being able to use the sparse matrix code I:

1. Refactored our `_add_interpolator` code to do a generic grid-to-grid interpolation
2. Added tests showing equivalences between `nibabel`, `dipy`, and our code for linear and nearest interpolation (it's a ton of combos but it takes only 0.3 sec on my machine)
3. Made the code `jit`-able but it didn't actually speed it up (thanks, vectorization!)
4. Fixed some minor bugs with our treatment of edge/border volume vertices in `_add_interpolator` (exposed by the nibabel/dipy equivalency tests), where some vertices could be interpolated to by a subset of the 8 possible vertices (e.g., 4 or 2 or even 1 of them) but it only added the interpolation if all could.

This should allow me to then speed up `SourceMorph._morph_vols`, but I'll do that in a subsequent PR. Even if it does not speed things up, this is already a nice refactoring + unit testing of our code.

### DIFF
--- a/mne/source_space.py
+++ b/mne/source_space.py
@@ -2302,109 +2302,140 @@ def _add_interpolator(sp):
     # Convert MRI voxels from destination (MRI volume) to source (volume
     # source space subset) coordinates
     #
-    vol_dims = s['vol_dims']
     combo_trans = combine_transforms(s['vox_mri_t'],
                                      invert_transform(s['src_mri_t']),
                                      'mri_voxel', 'mri_voxel')
-    del s
-    combo_trans['trans'] = combo_trans['trans'].astype(np.float32)
 
     logger.info('Setting up volume interpolation ...')
+    interp = _grid_interp(s['vol_dims'], (mri_width, mri_height, mri_depth),
+                          combo_trans['trans'])
+
+    # Compose the sparse matrices
+    for si, s in enumerate(sp):
+        # limit it columns that have any contribution from inuse
+        any_ = sparse.diags(
+            np.asarray(
+                interp[:, s['inuse'].astype(bool)].sum(1)
+            )[:, 0].astype(bool).astype(float)
+        )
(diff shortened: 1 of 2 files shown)


mne/source_space.py
```diff
@@ -2302,109 +2302,140 @@ def _add_interpolator(sp):
...
+
+    # Compose the sparse matrices
+    for si, s in enumerate(sp):
+        # limit it columns that have any contribution from inuse
+        any_ = sparse.diags(
+            np.asarray(
+                interp[:, s['inuse'].astype(bool)].sum(1)
+            )[:, 0].astype(bool).astype(float)
+        )
+        s['interpolator'] = any_ * interp
```
agramfort, line 2321:
```suggestion
        s['interpolator'] = any_ @ interp
```

to be more explicit

agramfort (commented):
just a nitpick. 

I don't understand the detail of what you did but I assume you checked carefully that there is no regression or behavior change...

thanks

agramfort (approved):

larsoner:
Yes this even fixes some minor bugs in master exposed by the new unit tests. And the existing tests were already pretty good at detecting when something went wrong as well.

agramfort:
+1 for MRG then

>

PR #7070, opened 2019-11-16
[MRG] Add spectro-spatial decomposition SSD example

dengemann:
This is a first example demonstrating spatial filtering with SSD (Nikulin et al 2011, NIMG), which essentially enhances the oscillatory signal of interest by removing background signal from surrounding frequencies.
The example is autonomous as is, but if we converge on ideas, I'd extend it to have a Transformer API like SPoC or CSP.

cc @agramfort @DavidSabbagh @pierreablin @wmvanvliet @britta-wstnr

PR #8527, opened 2020-11-15
[FIX] some post-merge fixes for SSD

dengemann:
This PR implements a few post-merge fixes to #7070 for issues pointed out by @agramfort [cc @vpeterson]

@agramfort I put inverse_transform -> NotImplemented and propose to call it ``apply_array`` as this API works with arrays, not instances. I will personally do not have the bandwidth to implement a proper ``apply`` as far more handling of channels types will be needed. Note that CSP has got no ``apply`` method and XDawn's apply is like ICA in supporting instances.

Perhaps @vpeterson wants look into extensions in the future. For now I suggest let's focus on what necessarily has to be fixed.

### DIFF
--- a/examples/decoding/plot_ssd_spatial_filters.py
+++ b/examples/decoding/plot_ssd_spatial_filters.py
@@ -39,7 +39,7 @@ freqs_noise = 8, 13
 
 ssd = SSD(info=raw.info,
           reg='oas',
-          sort_by_spectral_ratio=False,  # False for purpose of exmple.
+          sort_by_spectral_ratio=False,  # False for purpose of example.
           filt_params_signal=dict(l_freq=freqs_sig[0], h_freq=freqs_sig[1],
                                   l_trans_bandwidth=1, h_trans_bandwidth=1),
           filt_params_noise=dict(l_freq=freqs_noise[0], h_freq=freqs_noise[1],
@@ -68,7 +68,7 @@ psd, freqs = mne.time_frequency.psd_array_welch(
     ssd_sources, sfreq=raw.info['sfreq'], n_fft=4096)
 
 # Get spec_ratio information (already sorted).
-# Note that this is not necessay if sort_by_spectral_ratio=True (default).
+# Note that this is not necessary if sort_by_spectral_ratio=True (default).
 spec_ratio, sorter = ssd.get_spectral_ratio(ssd_sources)
 
 # Plot spectral ratio (see Eq. 24 in Nikulin 2011).
--- a/examples/decoding/plot_ssd_spatial_filters.py
+++ b/examples/decoding/plot_ssd_spatial_filters.py
@@ -1,6 +1,6 @@
 """
 ===========================================================
-Compute Sepctro-Spatial Decomposition (SDD) spatial filters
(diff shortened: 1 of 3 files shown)


examples/decoding/plot_ssd_spatial_filters.py
```diff
@@ -39,7 +39,7 @@
 
 ssd = SSD(info=raw.info,
           reg='oas',
-          sort_by_spectral_ratio=False,  # True is recommended here.
+          sort_by_spectral_ratio=False,  # False for purpose of exmple.
```
agramfort, line 42:
```suggestion
          sort_by_spectral_ratio=False,  # False for purpose of example.
```

examples/decoding/plot_ssd_spatial_filters.py
```diff
@@ -68,6 +68,7 @@
     ssd_sources, sfreq=raw.info['sfreq'], n_fft=4096)
 
 # Get spec_ratio information (already sorted).
+# Note that this is not necessay if sort_by_spectral_ratio=True (default).
```
agramfort, line 71:
```suggestion
# Note that this is not necessary if sort_by_spectral_ratio=True (default).
```

mne/decoding/ssd.py
```diff
@@ -253,12 +251,11 @@ def get_spectral_ratio(self, ssd_sources):
...
-    def apply(self):
-        """Not implemented, see ssd.inverse_transform() instead."""
-        # Exists because of _XdawnTransformer
-        raise NotImplementedError('See ssd.inverse_transform()')
+    def inverse_transform(self):
+        """Not implemented yet."""
+        raise NotImplementedError('inverse_transform is not yet available.')
 
-    def inverse_transform(self, X):
+    def apply_array(self, X):
```
agramfort, line 258: why not just apply?

dengemann: Well apply supports classically mne objects, not arrays. So apply would raise the expectation that it accepts instances.

dengemann:
@agramfort committed your suggestions and pushed another round of edits.

mne/decoding/ssd.py
```diff
@@ -253,19 +250,22 @@ def get_spectral_ratio(self, ssd_sources):
...
         """Remove selected components from the signal.
 
         This procedure will reconstruct M/EEG signals from which the dynamics
         described by the excluded components is subtracted
         (denoised by low-rank factorization).
         See :footcite:`HaufeEtAl2014b` for more information.
 
+        .. note:: Only NumPy arrays are supported
+           At this point, unlike in other classes with an apply method,
+           only NumPy arrays are supported but not instances of MNE objects.
```
agramfort, line 267:
```suggestion
        .. note:: Unlike in other classes with an apply method,
           only NumPy arrays are supported (not instances of MNE objects).
```

agramfort (approved):

dengemann:
@vpeterson ok with these changes?

vpeterson:
> @vpeterson ok with these changes?

indeed!

larsoner:
Thanks @dengemann @agramfort

PR #8529, opened 2020-11-16
MRG, ENH: Speed up morphing

larsoner:
1. Reduce time in `examples/inverse/plot_morph_volume_stc.py`, which is a pretty representative case, when doing `precompute=True` on my machine from 5:32 to 3:23
2. Fix bug introduced by #8462 where `as_volume` did not produce the correct output shape (with unit test added)
3. Make testing of `_grid_interp` a bit more complete

I also opened a dipy issue to see if they're interested in using these techniques to get speedups at their end: https://github.com/dipy/dipy/issues/2293


agramfort:
thx @larsoner

Issue #8512, opened 2020-11-13
Source depth for vertices


olafhauk:
#### Describe the new feature or enhancement
I would like to compute a metric of "source depth" for vertices in source space. This could be the Euclidean distance of a vertex to the nearest sensor/electrode or the nearest BEM surface.

#### Describe your proposed implementation
I've got something like this in mind (for source_space.py):

```
def vertex_depth(src, reference, mode='dist'):
    """Compute depth of vertices in source space.

    Parameters
    ----------
    src : instance of Source Space
        The vertices for which depth to be computed.
    reference : measurement info or BEM surfaces
        The locations with respect to which depth shall be computed.
    mode : 'dist'
        The distance metric to be computed.

    Returns
    -------
    vertex_depth: array, shape (n_vertices)
        The distance of each vertex with respect to the reference locations.
    """
```

#### Additional comments
I would need some advice getting relevant sensor and BEM locations with the appropriate transformations.

larsoner:
+1 for this. Thoughts:

1. Do we even need `mode`? Do people want something other than the minimum Euclidean distance to a sensor / any surface vertex? Maybe we can omit it to start and then see if people want other options, unless you already see a use case.
2. The API will probably need a `trans` argument to deal with head vs MRI coordinate frames.
3. I might just shorten `"reference"` to `"to"`, or `"other"` or something. "reference" is used a lot in MNE to talk about EEG reference options so it's better to avoid it here if there is a reasonable alternative.

@ckiefer0 @wmvanvliet @britta-wstnr recently needed something like this, so they might also have ideas.

olafhauk:
Great. ```mode``` could also include options like average distance or max distance.

wmvanvliet:
I've got some code for this: https://github.com/AaltoImagingLanguage/conpy/blob/master/conpy/forward.py#L27

It selects all vertices within a certain distance from any sensor. In order to make this selection, it has to compute the distance between each vertex and the closest sensor. Deals with transforms along the way.

olafhauk:
Thanks @wmvanvliet. I'll use your example and add you as co-author if that's ok. I'll start with distances to sensors, BEM surfaces can be included later.

larsoner:
Fixed by #8534

PR #8534, opened 2020-11-17
Source depth

olafhauk:
#### Reference issue
Implements #8512

Added function vertex_depth() to source_space.py to compute source depths for vertices.

#### Additional information
Used example from @wmvanvliet.


olafhauk:
@wmvanvliet can you please have a loot at vertex_depths() and check that I've understood your example right. I also mention two problems in the documentation (importing SourceSpace and Forward properly, and an issue with the exception). Thanks!

olafhauk:
Thanks @wmvanvliet. Now I can't figure out how to add vertex_depths to python_references.rst.

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+    verbose : bool | str | int | None
+        If not None, override default verbose level (see :func:`mne.verbose`
+        and :ref:`Logging documentation <tut_logging>` for more).
+
+    Returns
+    -------
+    depth : array of shape (,n_vertices)
+        The depths of source space vertices with respect to sensors.
+    """
+    from .forward import Forward
```
larsoner, line 3236:
```suggestion
    _validate_type(inst, (Forward, SourceSpaces), 'inst')
    from .forward import Forward
```

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+    if isinstance(inst, Forward):
+        info = inst['info']
+        src = inst['src']
+    elif isinstance(inst, SourceSpaces):
```
larsoner, line 3240:
```suggestion
    else:
        assert isinstance(inst, SourceSpaces)  # guaranteed above
```

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+        src = inst
+        if info is None:
+            raise ValueError('You need to specify an Info object with '
+                             'information about the channels.')
+    src = inst
```
larsoner, line 3245:
```suggestion
    src = inst
    _valididate_type(info, (Info,), 'info')
```

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+        The info structure that contains information about the channels with
+        respect to which to compute distances.
+    picks : array-like of int | None
+        Indices of sensors to include in distance calculations. If `None``
+        (default) then only MEG channels are used.
+    trans : str | instance of Transform | None
+        Either the full path to the head<->MRI transform ``*-trans.fif`` file
+        produced during coregistration, or the Transformation itself. If trans
+        is None, an identity matrix is assumed. Only needed when ``inst`` is a
+        source space in MRI coordinates.
```
larsoner, line 3223:
```suggestion
    %(picks_data)s
    %(trans_not_none)s
```

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+    mode : str
+        How to compute source depth. 'dist' computes Euclidean distance
+        between vertices and nearest sensors.
+    verbose : bool | str | int | None
+        If not None, override default verbose level (see :func:`mne.verbose`
+        and :ref:`Logging documentation <tut_logging>` for more).
```
larsoner, line 3229:
```suggestion
    %(verbose)s
```

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+        Indices of sensors to include in distance calculations. If `None``
+        (default) then only MEG channels are used.
+    trans : str | instance of Transform | None
+        Either the full path to the head<->MRI transform ``*-trans.fif`` file
+        produced during coregistration, or the Transformation itself. If trans
+        is None, an identity matrix is assumed. Only needed when ``inst`` is a
+        source space in MRI coordinates.
+    mode : str
+        How to compute source depth. 'dist' computes Euclidean distance
+        between vertices and nearest sensors.
```
larsoner, line 3226:
```suggestion
```
Remove until other modes are available / added

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+                  verbose=None):
+    """Compute source depths as distances between vertices and nearest sensor.
+
+    Parameters
+    ----------
+    inst : instance of Forward | instance of SourceSpaces
+        The object to select vertices from.
+    info : instance of Info | None
+        The info structure that contains information about the channels with
+        respect to which to compute distances.
```
larsoner, line 3215:
```suggestion
        respect to which to compute distances. Can be None if ``inst`` is
        a :class:`~mne.Forward` instance.
```

larsoner:
FYI @olafhauk my suggestions are often only approximate so they might or might not work... And also if you're on the "Files" view of the PR rather than the conversations one, you can click each suggested change in "add to batch" mode and make a single commit rather than several.

olafhauk:
> 
> 
> FYI @olafhauk my suggestions are often only approximate so they might or might not work... And also if you're on the "Files" view of the PR rather than the conversations one, you can click each suggested change in "add to batch" mode and make a single commit rather than several.

I guess I should have read this first... thx!

mne/source_space.py
```diff
@@ -3197,3 +3200,103 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+                else:
+                    break
+            else:
+                raise last_exp
+        src_trans = invert_transform(_ensure_trans(trans, 'head', 'mri'))
+        print('Transform!')
+    else:
+        src_trans = Transform('head', 'head')  # Identity transform
+
+    dev_to_head = _ensure_trans(info['dev_head_t'], 'meg', 'head')
```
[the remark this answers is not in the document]
olafhauk: I'm afraid I don't understand this one.

olafhauk:
@wmvanvliet @larsoner I get ""TypeError: tuple indices must be integers or slices, not tuple" for _get_trans() (l. 3234) with testing dataset but not sample dataset.

mne/source_space.py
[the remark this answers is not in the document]
larsoner: pushed

larsoner:
@olafhauk see if my commits make sense to you

olafhauk:
I'm ready for some criticisms.

mne/source_space.py
```diff
@@ -3197,3 +3199,90 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+
+        * None : Same as 'min' (Default).
+        * 'min' : Minimum Euclidean distance to nearest sensor.
+        * 'max' : Maximum Euclidean distance to sensors.
+        * 'mean' : Average Euclidean distance across sensors.
+    %(verbose)s
+
+    Returns
+    -------
+    depth : array of shape (,n_vertices)
```
agramfort, line 3231:
```suggestion
    depth : array of shape (n_vertices,)
```

mne/source_space.py
```diff
@@ -3197,3 +3199,90 @@ def _get_src_nn(s, use_cps=True, vertices=None):
     else:
         nn = s['nn'][vertices, :]
     return nn
+
+
+@verbose
+def vertex_depths(src, info=None, picks=None, trans=None, mode=None,
```
agramfort, line 3205: I would keep the API simple and flexible I would have returned an array which is of shape (n_vertices, n_channels) where n_channels is all channels present in info (making it mandatory) or limited to picks is not None. Trans should be required if MEG channels are present. Then mode could be removed from API. I think it's simple enough to do `np.mean(dist, axis=1)`

also maybe rename to compute_src_channel_dist ?

thoughts @olafhauk @larsoner ?

larsoner: > Trans should be required if MEG channels are present.

`trans` will be necessary if `src` is in MRI space (i.e., not from forward or inv), so I don't think it can go away. It's also very common to have to provide `trans` when providing a `src` so I think it's good to keep it.

> I would have returned an array which is of shape (n_vertices, n_channels)

Sounds good to me, indeed it's easy enough to post-process this to get what you need. And generally `n_channels` is an order of magnitude smaller than `n_vertices` so it shouldn't be too slow or memory hungry.

agramfort:
ok so let's make trans optional. We agree here. I had not realized it was
not necessary if src from fwd or inv

>

mne/source_space.py
```diff
@@ -3197,3 +3199,90 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+
+    Parameters
+    ----------
+    src : instance of SourceSpaces
+        The object to select vertices from.
+    info : instance of Info | None
+        If not None, the depth of each vertex will be computed as its minimum
+        distance to sensors specified by 'info' and 'picks'. If none, the depth
+        will be computed as the distance to the mean location (centre of
+        gravity) of all vertices in source space.
```
wmvanvliet, line 3217: I'd rather see the (distance to sensors / distance to centre of gravity) be implemented in two different functions. Switching between these modes based on whether info is `None` is too magical for my tastes. You can also see that now the function body is one giant if statement, which hints that it should really be two functions.

agramfort:
agreed. Is it really a use case to pass info as None here?

>

mne/source_space.py
olafhauk: I agree, also because distance to COG and distance to sensors are anticorrelated, which may be confusing. 
What about
compute_distance_from_sensors()
compute_distance_from_centre()
?

olafhauk:
> 
> agreed. Is it really a use case to pass info as None here?
> […](#)

I agree, also because distance to COG and distance to sensors are anticorrelated, which may be confusing.
What about
compute_distance_from_sensors()
compute_distance_from_centre()
?

mne/source_space.py
olafhauk: Happy to return all distances for post-processing.
vertex_depth() currently only takes src (not fwd). Let me know if you want me to change this.

agramfort:
what is the use for compute_distance_from_centre ?

keep the API with src not fwd. You can always pas fwd['src']

>

olafhauk:
> 
> what is the use for compute_distance_from_centre ? keep the API with src not fwd. You can always pas fwd['src']
> […](#)

Distance from centre would provide a measure of depth purely based on anatomy, like distance from origin in a sphere. I for example would use it together with distance from sensors to see how they explain variation of resolution metrics. But I might be the only user.

mne/tests/test_source_space.py
```diff
@@ -55,9 +60,48 @@
...
+    assert limits[0] * 2 > depths.min()  # meaningful choice of limits
+    assert_array_less(limits[0], depths)
+    assert_array_less(depths, limits[1])
+
+    if picks != 'eeg':
+        # this should break things
+        info['dev_head_t'] = None
+        with pytest.raises(ValueError,
+                           match='Transform between meg<->head'):
+            compute_distance_to_sensors(src, info, use_picks, trans)
```
agramfort, line 102: you don't seem to test when not passing a trans file which is optional in your API

olafhauk:
@wmvanvliet @larsoner @agramfort I tried to add a test for the -trans option, but I must be misunderstanding something about transformations for source spaces and forward solutions, or about the testing dataset. I added some comments to test_compute_distance_to_sensors().

I also keep getting error messages from test_import_nesting, e.g. "Found un-nested import(s) for ["scipy submodules: ['special', 'constants', 'spatial']"]".

agramfort:
you should import these scipy modules within the functions to reduce import
time of mne

>

mne/source_space.py
```diff
@@ -3197,3 +3198,65 @@ def _get_src_nn(s, use_cps=True, vertices=None):
...
+
+    assert isinstance(src, SourceSpaces)
+    _validate_type(info, (Info,), 'info')
+
+    # Load the head<->MRI transform if necessary
+    if src[0]['coord_frame'] == FIFF.FIFFV_COORD_MRI:
+        src_trans, _ = _get_trans(trans, allow_none=False)
+    else:
+        src_trans = Transform('head', 'head')  # Identity transform
+
```
agramfort, line 3236: @olafhauk to cover this line and test without passing a trans you should pass the src from a forward instance `fwd['src']` or `inv['src']`.

HTH

olafhauk: That's what I tried on ll. 104-6 in test_compute_distance_to_sensors() (see comments in code), but the tests fail.

agramfort:
@olafhauk I pushed a fix

would it be useful to add this to an example?

I thinking of looking at this next to https://mne.tools/stable/generated/mne.sensitivity_map.html ?

olafhauk:
> 
> 
> @olafhauk I pushed a fix
> 
> would it be useful to add this to an example?
> 
> I thinking of looking at this next to https://mne.tools/stable/generated/mne.sensitivity_map.html ?

Correlating sensitivity with source depth? Good idea, I'll give it a try, but in a new PR?

agramfort:
I don't like much adding code in the public API without some example or narrative doc. These functions tend to not be discovered by users.

olafhauk:
> 
> 
> I don't like much adding code in the public API without some example or narrative doc. These functions tend to not be discovered by users.

Just when all my tests have passed I get more homework... ;-)
I can add a whole-brain distribution of "source depth" and then correlate the sensitivity values with it? I'm expecting the source depth distribution to look similar to the sensitivity map, and the correlation to be high.

agramfort:
:)

just display it so there is a usage example in the doc

>

agramfort (approved):
looks good

https://23936-1301584-gh.circle-artifacts.com/0/dev/auto_examples/forward/plot_forward_sensitivity_maps.html#sphx-glr-auto-examples-forward-plot-forward-sensitivity-maps-py

wmvanvliet:
Cool!

PR #8536, opened 2020-11-17
MRG, VIZ, FIX: plot_sensors title and interactivity

### LINKED ISSUE #7850: viz.plot_sensors() only partially honors axes parameter
#### Describe the bug
I wanted to add sensor alignment plots via `Raw.plot_sensors()` to some subplots of a larger figure. While the sensors **are** getting plotted, I observe some very annoying side effects that imho qualify as bugs.

#### Steps to reproduce
```python
import mne

fname = data_path + '/MEG/sample/sample_audvis_raw.fif'
raw = mne.io.read_raw_fif(fname)

fig, ax = plt.subplots(2, 2)

ax[0, 0].plot(range(10))
ax[0, 0].set_title('Title A')

ax[1, 0].plot(range(10))
ax[1, 0].set_title('Title B')

raw.plot_sensors(ch_type='mag', axes=ax[0, 1])
raw.plot_sensors(ch_type='grad', axes=ax[1, 1])
```

#### Expected results
All subplots should appear in their designated axes.

#### Actual results
Figure_1

- First sensor plot doesn't have a title
- Second sensor plot's title appears as `suptitle` while it should be an axes title
- The entire figure appears to be truncated at the top, left, and bottom

#### Additional information
The behavior reminds me somewhat of #7595, which has been fixed by @drammock in #7706.

I'm using Matplotlibs Qt backend on macOS.

drammock:
closes #7850 

- `plot_sensors` now puts title as axes title instead of `fig.suptitle`
- the on_pick callback now won't act unless the click was in the correct axes

### DIFF
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -964,6 +964,9 @@ def plot_sensors(info, kind='topomap', ch_type=None, title=None,
 
 def _onpick_sensor(event, fig, ax, pos, ch_names, show_names):
     """Pick a channel in plot_sensors."""
+    if event.mouseevent.inaxes != ax:
+        return
+
     if event.mouseevent.key == 'control' and fig.lasso is not None:
         for ind in event.ind:
             fig.lasso.select_one(ind)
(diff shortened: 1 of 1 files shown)


drammock:
Ready for review/merge. Only CI failure is the XDAWN test on py3.6 (a test which I think @larsoner is about to remove anyway).

agramfort (approved):

larsoner:
Thanks @drammock

PR #8537, opened 2020-11-17
MRG, MAINT: deduplicate definition of FIFF constants

drammock:
expands the entries in `get_channel_type_constants()` to include coil types and units, so they can be used in `create_info`.

### DIFF
--- a/mne/io/edf/tests/test_edf.py
+++ b/mne/io/edf/tests/test_edf.py
@@ -27,12 +27,12 @@ from mne.io.edf.edf import _get_edf_default_event_id
 from mne.io.edf.edf import _read_annotations_edf
 from mne.io.edf.edf import _read_ch
 from mne.io.edf.edf import _parse_prefilter_string
-from mne.io.pick import channel_indices_by_type
+from mne.io.pick import channel_indices_by_type, get_channel_type_constants
 from mne.annotations import events_from_annotations, read_annotations
-from mne.io.meas_info import _kind_dict as _KIND_DICT
 
 
 FILE = inspect.getfile(inspect.currentframe())
+KIND_DICT = get_channel_type_constants()
 data_dir = op.join(op.dirname(op.abspath(FILE)), 'data')
 montage_path = op.join(data_dir, 'biosemi.hpts')  # XXX: missing reader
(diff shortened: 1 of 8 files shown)


agramfort (approved):
thx @drammock

larsoner:
Failures look real

drammock:
> Failures look real

yep. ran out of steam last night. working on them now.

mne/io/pick.py
```diff
@@ -15,51 +15,94 @@
                      _check_option)
 
 
-def get_channel_type_constants():
-    """Return all known channel types.
+def get_channel_type_constants(include_defaults=False):
```
drammock, line 18: @larsoner could you have a look at how I've altered this function?  As I understand it: sometimes we want the returned dict to have minimal entries (values that should *always* be present in `info['chs']` for a given channel type, regardless of the acquisition system used).  Other times, we want additional values because we use the returned dict to pre-populate an info dict for synthetic data (e.g., in `create_info`).  Does that understanding accurately reflect all of the relevant use cases for this kind of dictionary?  If so, does the approach I've taken here seem optimal?

mne/io/edf/tests/test_edf.py
```diff
@@ -27,12 +27,12 @@
...
 from mne.io.edf.edf import _read_ch
 from mne.io.edf.edf import _parse_prefilter_string
-from mne.io.pick import channel_indices_by_type
+from mne.io.pick import channel_indices_by_type, get_channel_type_constants
 from mne.annotations import events_from_annotations, read_annotations
-from mne.io.meas_info import _kind_dict as _KIND_DICT
 
 
 FILE = inspect.getfile(inspect.currentframe())
+KIND_DICT = get_channel_type_constants()
```
larsoner, line 35: I'd rather nest this in `test_edf_stim_ch_pick_up` -- if we ever inadvertently break `get_channel_type_constants`, it will break tests during the collection rather than the running phase, which is not so nice.

larsoner (commented):
Otherwise LGTM. Not 100% sure on the `include_defaults` stuff but what you did looks reasonable

drammock:
CIs all green, ready for review/merge.

larsoner:
Thanks @drammock

Issue #8518, opened 2020-11-13
MAINT: CI


larsoner:
A few things we should do for CIs:

1. Migrate away from Travis (#8514)
2. Add `[skip azp]` [support](https://github.com/mne-tools/mne-python/pull/8514#issuecomment-726797122)
3. Add [inline annotations](https://github.com/mne-tools/mne-python/pull/8514#issuecomment-726797122) for codespell and flake
4. Add 3.9 CI. For this I have no idea when VTK will ship a wheel, so I might just build a Linux one locally and upload it somewhere for us to use, assuming tests actually pass when using it.

larsoner:
cc @GuillaumeFavelier feel free to take care of 2/3/4 as part of #8514 if you want or we can do it in separate PRs. (1) is really the critical thing to alleviate CI pressure anyway, so maybe best to do that first.

GuillaumeFavelier:
I think we're done here too, feel free to re-open if I missed something.

Closing through https://github.com/mne-tools/mne-python/pull/8533, https://github.com/mne-tools/mne-python/pull/8514/

larsoner:
I think just https://github.com/marketplace/actions/ci-skip-action or something similar. Do you want to try?

GuillaumeFavelier:
I'm on it :)

We might be more interested in the `fail-safe` version (which is very verbose):

https://github.com/marketplace/actions/ci-skip-action#fail-safe

And AFAIK, only one `commit-filter` can be used

larsoner:
> And AFAIK, only one commit-filter can be used

It seems like it should be easy enough to modify their action to support `str` or `list of str`, want to look into that? From my experience it's actually kind of fun to make the actions...

GuillaumeFavelier:
Okay, I'll look into it :+1:

PR #8538, opened 2020-11-18
MAINT: Fix skipping with GitHub Actions

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/issues/8518#issuecomment-729641049 and uses https://github.com/mstachniuk/ci-skip to add skipping mechanic to our GitHub Actions jobs.

Related to https://github.com/mne-tools/mne-python/issues/8518

### DIFF
--- a/.github/workflows/codespell_and_flake.yml
+++ b/.github/workflows/codespell_and_flake.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   check-skip:
+    runs-on: ubuntu-20.04
     - uses: actions/checkout@v2
       with:
         fetch-depth: 0
--- a/.github/workflows/codespell_and_flake.yml
+++ b/.github/workflows/codespell_and_flake.yml
@@ -11,17 +11,16 @@ jobs:
   check_skip:
     runs-on: ubuntu-20.04
     outputs:
-      skip: ${{ steps.result_step.outputs.result }}
+      skip: ${{ steps.result_step.outputs.ci-skip }}
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: mstachniuk/ci-skip@master
+      - id: result_step
+        uses: mstachniuk/ci-skip@master
         with:
           commit-filter: '[skip ci];[ci skip];[skip github]'
           commit-filter-separator: ';'
-      - run: echo "::set-output name=result::${CI_SKIP}"
-        id: result_step
 
   style:
     needs: check_skip
--- a/.github/workflows/codespell_and_flake.yml
+++ b/.github/workflows/codespell_and_flake.yml
@@ -16,6 +16,11 @@ jobs:
       CODESPELL_SKIPS: 'doc/auto_*,*.fif,*.eve,*.gz,*.tgz,*.zip,*.mat,*.stc,*.label,*.w,*.bz2,*.annot,*.sulc,*.log,*.local-copy,*.orig_avg,*.inflated_avg,*.gii,*.pyc,*.doctree,*.pickle,*.inv,*.png,*.edf,*.touch,*.thickness,*.nofix,*.volume,*.defect_borders,*.mgh,lh.*,rh.*,COR-*,FreeSurferColorLUT.txt,*.examples,.xdebug_mris_calc,bad.segments,BadChannels,*.hist,empty_file,*.orig,*.js,*.map,*.ipynb,searchindex.dat,install_mne_c.rst,plot_*.rst,*.rst.txt,c_EULA.rst*,*.html,gdf_encodes.txt,*.svg'
     steps:
       - uses: actions/checkout@v2
+      - uses: GuillaumeFavelier/ci-skip@enh/separator
+        with:
+          fail-fast: true
+          commit-filter: '[skip ci];[ci skip];[skip github]'
+          separator: ';'
       - uses: actions/setup-python@v2
         with:
           python-version: '3.8'
--- a/.github/workflows/codespell_and_flake.yml
+++ b/.github/workflows/codespell_and_flake.yml
@@ -8,8 +8,18 @@ on:
       - '*'
 
 jobs:
+  check-skip:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: GuillaumeFavelier/ci-skip@enh/separator
+        with:
+          fail-fast: true
+          commit-filter: '[skip ci];[ci skip];[skip github]'
+          separator: ';'
+
   style:
-    if: "github.repository == 'mne-tools/mne-python' && !contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.event.head_commit.message, '[skip github]')"
+    needs: check-skip
     runs-on: ubuntu-20.04
     env:
       CODESPELL_DIRS: 'mne/ doc/ tutorials/ examples/'
@@ -18,12 +28,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: GuillaumeFavelier/ci-skip@enh/separator
-        with:
-          commit-filter: '[skip ci];[ci skip];[skip github]'
-          separator: ';'
       - uses: actions/setup-python@v2
-        if: ${{ env.CI_SKIP == 'false' }}
         with:
           python-version: '3.8'
           architecture: 'x64'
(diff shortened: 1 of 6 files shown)


.github/workflows/linux_pip.yml
```diff
@@ -27,6 +26,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - uses: GuillaumeFavelier/ci-skip@enh/separator
+        with:
+          fail-fast: true
+          commit-filter: '[skip ci];[ci skip];[skip github]'
+          separator: ';'
```
larsoner, line 33: So they already support this? Then PR to their README would be good

GuillaumeFavelier: That's my fork, I am testing in here before opening the PR :)

larsoner: Ahh missed that, cool!

GuillaumeFavelier: I opened https://github.com/mstachniuk/ci-skip/pull/4

GuillaumeFavelier:
This PR is ready for reviews @larsoner

.github/workflows/codespell_and_flake.yml
```diff
@@ -16,19 +16,30 @@ jobs:
...
         name: 'Install dependencies'
+        if: ${{ env.CI_SKIP == 'false' }}
       - uses: rbialon/flake8-annotations@v1
+        if: ${{ env.CI_SKIP == 'false' }}
         name: 'Setup flake8 annotations'
       - run: make flake
+        if: ${{ env.CI_SKIP == 'false' }}
         name: 'Run flake8'
       - uses: GuillaumeFavelier/actions-codespell@feat/quiet_level
+        if: ${{ env.CI_SKIP == 'false' }}
```
larsoner, line 42: Do we have to do it this way? No way to say "don't do anything after this step" and put this once at the top?

.github/workflows/codespell_and_flake.yml
```diff
@@ -16,19 +16,30 @@ jobs:
       CODESPELL_SKIPS: 'doc/auto_*,*.fif,*.eve,*.gz,*.tgz,*.zip,*.mat,*.stc,*.label,*.w,*.bz2,*.annot,*.sulc,*.log,*.local-copy,*.orig_avg,*.inflated_avg,*.gii,*.pyc,*.doctree,*.pickle,*.inv,*.png,*.edf,*.touch,*.thickness,*.nofix,*.volume,*.defect_borders,*.mgh,lh.*,rh.*,COR-*,FreeSurferColorLUT.txt,*.examples,.xdebug_mris_calc,bad.segments,BadChannels,*.hist,empty_file,*.orig,*.js,*.map,*.ipynb,searchindex.dat,install_mne_c.rst,plot_*.rst,*.rst.txt,c_EULA.rst*,*.html,gdf_encodes.txt,*.svg'
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: GuillaumeFavelier/ci-skip@enh/separator
```
larsoner, line 21: We also should wait for this to get pushed upstream before merging

.github/workflows/codespell_and_flake.yml
GuillaumeFavelier: This is the way I know and it's also consistent with Azure Pipelines. Except that on the latter, you can use `stage dependencies`. I kept looking for an alternative in GitHub Actions with no success so far.

larsoner: Can't we make two jobs: the first is the a job named `ci_skip`, and the second job is this job with a `needs: ci_skip` with a condition, like:

https://github.com/pgxn/docker-pgxn-tools/blob/main/.github/workflows/cicd.yml#L24-L25

GuillaumeFavelier: I'll try... But does it mean that the 'skip' job needs to fail for the rest to run?

GuillaumeFavelier: In that case, I think the condition might not be necessary

GuillaumeFavelier:
I tried by migrating `codespell_and_flake` workflow. It could be interesting to use this method if all the jobs were in a single workflow like Azure. Otherwise, I think it's a more verbose way to `fail-fast` with `ci-skip`.

I think it's worth coming back to this approach once [restarting each job individually](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234/65) is supported. What do you think @larsoner ?

.github/workflows/codespell_and_flake.yml
```diff
@@ -8,14 +8,28 @@ on:
...
 jobs:
+  check-skip:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: GuillaumeFavelier/ci-skip@enh/separator
+        with:
+          fail-fast: true
```
larsoner, line 19: This one `fail-fast` at the top of each file looks better to me than having N `if: ${{ env.CI_SKIP == 'false' }}` for our N steps in each file.

Are those the two alternatives? Or are you saying somehow `fail-fast` does not work?

GuillaumeFavelier: I agree it looks better. The drawback is that it doubles the number of jobs. Also if `fail-fast: true` is acceptable in the main job, we don't need to create an additional `check-skip` for that. I don't know if I'm clear so I'll push the other alternative for `codespell_and_flake` to demonstrate.

GuillaumeFavelier:
In 8efbdef we have another job (`check-skip`) that fails and the 'main' job is skipped.
In 580fa56, the 'main' job just `fail-fast`.

larsoner:
I think I prefer the two-jobs approach. I don't think it will add much overhead.

Also we don't want `fail-fast`. This makes it show up as failed, which is not good. How about we set some var and then use *both* `needs` and some `if` condition on that var as in the link I had pasted [above](https://github.com/pgxn/docker-pgxn-tools/blob/main/.github/workflows/cicd.yml#L24-L25)? That way the `check_skip` job can succeed + set var, and then the next job waits because of the `needs` and only runs if the `if` condition is met (where the `if` condition depends on the var set by the `check_skip` job).

GuillaumeFavelier:
In theory this could work if it was possible to share variables between jobs, but [it's not the case at the moment](https://github.community/t/sharing-a-variable-between-jobs/16967). So to make this work, we would need to write the result to a file as artifact and read it in the job that depends on it. I am not sure all of this is possible in a simple `if` statement.

Something like `yanjingzhu` described in https://github.community/t/sharing-a-variable-between-jobs/16967/2:

> As you metioned artifacts, we would suggest you use it . You can write the version into a file , upload it as artifact in one job and then download the artifacts in the other job , read the file content.

GuillaumeFavelier:
I'll try anyway, it's an interesting approach :)

larsoner:
> I am not sure all of this is possible in a simple if statement.

If it's not, then it's no better than the repeated `CI_SKIP` statement approach, as you'll need it in every step thereafter. :(

.github/workflows/codespell_and_flake.yml
```diff
@@ -8,14 +8,32 @@ on:
       - '*'
 
 jobs:
+  check_skip:
+    runs-on: ubuntu-20.04
+    outputs:
+      skip: ${{ steps.result_step.outputs.result }}
```
larsoner, line 14: https://github.com/mne-tools/mne-python/runs/1424512672?check_suite_focus=true !

GuillaumeFavelier: Apparently it works :)

GuillaumeFavelier:
Okay now it works for `codespell_and_flake`. I'll use this in the other workflows now

.github/workflows/codespell_and_flake.yml
larsoner: Looks like it works, now we just wait for this before merge right?

GuillaumeFavelier: It works indeed. I'll try to fix my `docker` issues on https://github.com/mstachniuk/ci-skip/pull/4 then

GuillaumeFavelier: https://github.com/mstachniuk/ci-skip/pull/4 has been merged! Also the syntax changed slightly, instead of `separator`, it's `commit-filter-separator`. I updated the PR in a5288cf

larsoner:
@GuillaumeFavelier feel free to merge whenever it's ready

PR #8539, opened 2020-11-18
fix example

drammock:
fixes example that was broken by #8536


examples/visualization/plot_eeglab_head_sphere.py
```diff
@@ -99,7 +99,6 @@
 fake_evoked.plot_sensors(sphere=(x, y, z, radius), axes=ax[1], show=False)
 
 # add titles
-fig.texts[0].remove()
```
drammock, line 102: no longer needed because `plot_sensors` doesn't use `fig.suptitle` anymore

larsoner:
Thanks @drammock

PR #8542, opened 2020-11-18
MRG, BUG: Fix support for KIT Berlin

### LINKED ISSUE #8535: Problem reading EEG channels of an old yokogawa KIT MEG system with the newest MNE version.
#### Describe the bug
I'm working with data obtained by a 160 channel yokogawa/KIT MEG system, which has 125 axial gradiometers, the others are EEG and trigger channels. This system is located at the PTB institute in Berlin. To import the ".con" file I use the  function: mne.io.read_raw_kit("filename.con"). Until recently, everything worked well, values of both EEG and MEG channels were imported as they should (I was using the MNE-Python version 0.20.0). Then I updated the MNE-Python to the latest 0.21.2, the MEG channels were correct, but all the EEG channels have at each time values of 0.. Then I tried again with the version 0.20.0 and everything worked fine. 

#### Steps to reproduce
This issue can be reproduced by running the "example.py" with different versions of MNE, which imports "data.con" and plots raw data. Both files can be downloaded on the following link: https://imfm2815001.fmf.uni-lj.si/nextcloud/index.php/s/bHK4weHmKBTrS4P 

#### Actual results
I attach two images of the function raw.plot(), each with its own version of MNE-Python on the same dataset. When i used the  MNE-Python version 0.20.0 the matplotlib version 3.1.1 (latest version gives errors) and when using MNE-Python version 0.21.2 the matplotlib version 3.3.2.

MNE=0 20 0
MNE=0 21 2

#### Additional information
Platform:      Linux-5.8.18-100.fc31.x86_64-x86_64-with-glibc2.10
Python:        3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0]
Executable:    /home/marhl/Applications/anaconda3/envs/mne/bin/python
CPU:           x86_64: 4 cores
Memory:        15.5 GB

mne:           0.21.2
numpy:         1.19.2 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.5.2
matplotlib:    3.3.2 {backend=Qt5Agg}

sklearn:       0.23.2
numba:         0.51.2
nibabel:       3.2.0
cupy:          Not found
pandas:        1.1.3
dipy:          1.3.0
mayavi:        4.7.2
pyvista:       0.27.2 {pyvistaqt=0.2.0, OpenGL 4.5 (Core Profile) Mesa 19.2.8 via Mesa DRI Intel(R) HD Graphics 620 (Kaby Lake GT2)}
vtk:           9.0.1
PyQt5:         5.15.1

I hope it is OK that I reported this issue here. Since this is my first Bug report, I hope I have written everything correct. Can I help you somehow with any additional informations, please let me know.

larsoner:
@UrbanM simplifying your example to get rid of all of the `scalings` stuff and doing:
```
python -ic "import mne; mne.io.read_raw_kit("data.con").plot()"
```
Screen Shot 2020-11-18 at 5 28 46 PM

You'll notice that the scale bar for EEG data is 40 uV. Your plot from version 0.20 had a pretty wild `400000.0 uV` scale bar which suggests things were wrong there, too (although at least they were non-zero!).

This almost certainly isn't correct, unless I got very lucky -- I need to dig into why your `exg_gains` are all zero. And if those are being read correctly as zero, what the actually correct fallback gain is to use. Can you browse these data in some KIT (or other) software and let me know what the correct voltage of some value in some EEG channel at some time is supposed to be? For example, E14 at 28.77 seconds seems to be the top of a fairly smooth peak.

I also have no idea if the names I gave to these constants are consistent, feel free to look at the diff and comment @UrbanM!

Closes #8535

### DIFF
--- a/mne/io/kit/constants.py
+++ b/mne/io/kit/constants.py
@@ -48,6 +48,7 @@ KIT.FLL_SETTINGS = {
     50: (2, 1, 1),  # Hanger Type #3
     60: (2, 1, 1),  # Hanger Type #3
     100: (3, 3, 3),  # Low Band Kapper Type
+    101: (3, 3, 3),  # Low Band Kapper Type
     120: (3, 3, 3),  # Low Band Kapper Type
     200: (4, 4, 3),  # High Band Kapper Type
     300: (2, 2, 2),  # Kapper Type
@@ -164,6 +165,7 @@ KIT.SYSTEM_UMD_2019_09 = 54  # UMD September 3, 2019 -
 KIT.SYSTEM_YOKOGAWA_2017_01 = 1001  # Kanazawa (until 2017)
 KIT.SYSTEM_YOKOGAWA_2018_01 = 10020  # Kanazawa (since 2018)
 KIT.SYSTEM_YOKOGAWA_2020_08 = 10021  # Kanazawa (since August 2020)
+KIT.SYSTEM_BERLIN = 124
 # Sensor layouts for plotting
 KIT_LAYOUT = {
     KIT.SYSTEM_AS: None,
@@ -184,6 +186,7 @@ KIT_LAYOUT = {
     KIT.SYSTEM_YOKOGAWA_2017_01: None,
     KIT.SYSTEM_YOKOGAWA_2018_01: None,
     KIT.SYSTEM_YOKOGAWA_2020_08: None,
+    KIT.SYSTEM_BERLIN: None,
 }
 # Sensor neighbor definitions
 KIT_NEIGHBORS = {
@@ -205,6 +208,7 @@ KIT_NEIGHBORS = {
     KIT.SYSTEM_YOKOGAWA_2017_01: None,
     KIT.SYSTEM_YOKOGAWA_2018_01: None,
     KIT.SYSTEM_YOKOGAWA_2020_08: None,
+    KIT.SYSTEM_BERLIN: None,
 }
 # Names displayed in the info dict description
 KIT_SYSNAMES = {
@@ -225,6 +229,7 @@ KIT_SYSNAMES = {
     KIT.SYSTEM_YOKOGAWA_2017_01: 'Yokogawa of Kanazawa (until 2017)',
     KIT.SYSTEM_YOKOGAWA_2018_01: 'Yokogawa of Kanazawa (since 2018)',
     KIT.SYSTEM_YOKOGAWA_2020_08: 'Yokogawa of Kanazawa (since August 2020)',
+    KIT.SYSTEM_BERLIN: 'Physikalisch Technische Bundesanstalt, Berlin/128-channel MEG System',  # noqa: E501
 }
 
 LEGACY_AMP_PARAMS = {
--- a/mne/io/kit/constants.py
+++ b/mne/io/kit/constants.py
@@ -48,7 +48,7 @@ KIT.FLL_SETTINGS = {
     50: (2, 1, 1),  # Hanger Type #3
     60: (2, 1, 1),  # Hanger Type #3
     100: (3, 3, 3),  # Low Band Kapper Type
-    101: (3, 3, 3),  # Low Band Kapper Type
+    101: (1, 3, 2),  # Berlin (DC, 200 Hz, Through)
     120: (3, 3, 3),  # Low Band Kapper Type
     200: (4, 4, 3),  # High Band Kapper Type
     300: (2, 2, 2),  # Kapper Type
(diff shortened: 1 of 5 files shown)


UrbanM:
@larsoner thank you very much for all your work. I'm sorry for this bad example, initially these channels in the system were made for measuring EEG, but here we don't really measure the EEG signal, but the voltage from the new OPM MEG sensors, that is the reason for these high amplitudes. We multiply these values with a gain factor to obtain the magnetic fields. Here is a picture from the MegLaboratory (a software provided from the manufacturer of the MEG device). In this program the value of the channel E14 at time 28.77s is -2.55V. I can ask the manager of the device if he has some old data where they really measured EEG signal, would that help?
Screenshot from 2020-11-19 10-10-04
Screenshot from 2020-11-19 10-31-29

larsoner:
@UrbanM is it okay for me to add the dataset you created to the `MNE-testing-data` dataset? It would be nice to add a little test that the value obtained for that channel is correct. Could you also look at some MEG channel at some time and give me a value in the manufacturer's software? That way I can spot check them, too.

mne/io/kit/constants.py
```diff
@@ -48,6 +48,7 @@
     50: (2, 1, 1),  # Hanger Type #3
     60: (2, 1, 1),  # Hanger Type #3
     100: (3, 3, 3),  # Low Band Kapper Type
+    101: (3, 3, 3),  # Low Band Kapper Type
```
larsoner, line 51: @UrbanM any idea which type you have? Or equivalently, does your acquisition/manufacturer software tell you somewhere what you highpass/lowpass/notch filter settings were during acquisition?

mne/io/kit/constants.py
```diff
@@ -164,6 +165,7 @@
 KIT.SYSTEM_YOKOGAWA_2017_01 = 1001  # Kanazawa (until 2017)
 KIT.SYSTEM_YOKOGAWA_2018_01 = 10020  # Kanazawa (since 2018)
 KIT.SYSTEM_YOKOGAWA_2020_08 = 10021  # Kanazawa (since August 2020)
+KIT.SYSTEM_BERLIN = 124
```
larsoner, line 168: @UrbanM any idea when the system was installed? I think we try to name these based on that

mne/io/kit/kit.py
```diff
@@ -760,6 +760,7 @@ def get_kit_info(rawfile, allow_unknown_format, standardize_names=None,
     is_exg = [ch['type'] in (KIT.CHANNEL_EEG, KIT.CHANNEL_ECG)
               for ch in channels]
     exg_gains /= 2 ** (adc_stored - 14)
+    exg_gains[exg_gains == 0] = ad_to_volt
```
larsoner, line 763: FYI with this change I get 2.55 V on that channel at that time, feel free to try it

UrbanM:
@larsoner I will try to provide a better sample dataset without wild EEG channels. I checked the value of the MEG channel "MEG 003" in the manufacturers software at time 28.770 s it is -8.89 pT. I also checked the value with the MNE-python with your "kit" branch, i get the value -8.88592475e-12.

Screenshot from 2020-11-19 21-47-28

mne/io/kit/kit.py
UrbanM: Thank you for finding the solution, I tried it, it is correct.

larsoner:
> @larsoner I will try to provide a better sample dataset without wild EEG channels.

The current dataset is fine from the MNE-Python testing end of things. So I'm fine with putting it in MNE-testing-data, but if you don't want it living permanently there another similar short dataset is fine.

mne/io/kit/constants.py
UrbanM: I contacted the manager of the device, I will give you the answer soon.

mne/io/kit/constants.py
UrbanM: I contacted the manager of the device, I will give you the answer soon.

UrbanM:
Yes of course, if this dataset is OK, feel free to put it in MNE-testing-data.

mne/io/kit/constants.py
UrbanM: @larsoner I contacted the system administrator. Regarding the date of first installation he said: "Lets set as the installation date 2008 as in this year the system got 128 SQUID channels, before it had only 96. In 2018 the software was upgraded."  
And also his suggestion, if that makes sense:
"My suggestion for naming in constants.py:
`# Known KIT systems`
`KIT.SYSTEM_EAGLE_TECHNOLOGY_PTB = 124 # PTB (since 2008, software upgrade in 2018)`
and
`KIT_SYSNAMES = {`
`KIT.SYSTEM_EAGLE_TECHNOLOGY_PTB: 'Eagle Technology MEG (KIT/Yokogawa style) at PTB (since 2008, software upgrade in 2018)'`
Please pass on this information if it makes sense for you!"

mne/io/kit/constants.py
UrbanM: There is no default highpass/lowpass/notch filter settings, for this exact dataset it is HPF: TrueDC, LPF200Hz, BEF:Through . And for regular SQUID measurements we mostly use HPF: 0.1, LPF200Hz, BEF:Through. I hope this makes sense.

larsoner:
@agramfort ready for review/merge from my end

agramfort (approved):
@larsoner shall we backport?

PR #8545, opened 2020-11-19
MRG, VIZ: equal aspect ratio for inset sensor maps

drammock:
this is an element of #7751


drammock:
renderings:

**before/after**
Screenshot_2020-11-19_18-06-52 Screenshot_2020-11-19_18-07-26

**before**
before


**after**
after

larsoner:
Does it maintain the aspect during resize? That would be magical...

drammock:
> Does it maintain the aspect during resize? That would be magical...

Yes! 🧙‍♂️

agramfort (approved):

larsoner:
Thanks @drammock -- I think this one is worth a `latest.inc` update in the `BUG` section -- feel free to add and merge!

PR #8546, opened 2020-11-19
MRG, VIZ, MAINT: use figure class for PSD plots

drammock:
This is an element of #7751

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -38,9 +38,16 @@ class MNEFigure(Figure):
     """Base class for 2D figures & dialogs; wraps matplotlib.figure.Figure."""
 
     def __init__(self, **kwargs):
+        from matplotlib import rcParams
         # figsize is the only kwarg we pass to matplotlib Figure()
         figsize = kwargs.pop('figsize', None)
         super().__init__(figsize=figsize)
+        # things we'll almost always want
+        defaults = dict(fgcolor=rcParams['axes.edgecolor'],
+                        bgcolor=rcParams['axes.facecolor'])
+        for key, value in defaults.items():
+            if key not in kwargs:
+                kwargs[key] = value
         # add our param object
         self.mne = MNEFigParams(**kwargs)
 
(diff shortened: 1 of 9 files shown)


drammock:
This is ready for review.  Renderings: [epochs PSD](https://23789-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_20_visualize_epochs.html#plotting-the-power-spectrum-of-epochs), [raw psd](https://23789-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/raw/plot_40_visualize_raw.html#plotting-spectral-density-of-continuous-data).  Both renderings have issues with the margins / hspace, which don't occur locally in interactive plotting, but do occur locally when running `make html_dev-pattern`.  @larsoner is sphinx-gallery calling `tight_layout` automatically?  (if so, can we tell it not to, either case-by-case or globally?)

agramfort (approved):

mne/viz/_figure.py
```diff
@@ -2097,6 +2093,41 @@ def _blit_vline(self, visible):
...
+        bottom = self._inch_to_rel(b_margin, horiz=False)
+        top = 1 - self._inch_to_rel(t_margin, horiz=False)
+        # AXES: make subplots
+        axes = [self.add_subplot(n_axes, 1, 1)]
+        for ix in range(1, n_axes):
+            axes.append(self.add_subplot(n_axes, 1, ix + 1, sharex=axes[0]))
+        self.subplotpars.update(left=left, bottom=bottom, top=top, right=right,
+                                hspace=0.4)
+        # save useful things
+        vars(self.mne).update(ax_list=axes)
```
larsoner, line 2119: Can you add a comment why you need to use `vars` rather than just doing `self.mne.ax_list = axes` or so?

larsoner (commented):
Just one curiosity / comment that will help maintainability, otherwise +1 for merge

mne/viz/_figure.py
drammock: no good reason. I was mimicking `MNEBrowseFigure` (in which there were a lot of things to add at the end of `__init__`, and so using `vars` there made it much more concise).  Here doing a simple assignment would be fine.

PR #8553, opened 2020-11-22
MRG, ENH: make plot alignment use defaults for colors

agramfort:
plot_alignment does not use the colors used in coreg. It should be consistent. This is just an example that leads to something like this:

figure_1

as opposed to master:

Figure_2

FYI this is obtained with this script:

```
import numpy as np
import mne
from mne.viz import set_3d_backend
from mne.viz._brain.view import views_dicts

set_3d_backend("pyvista")
raw_fname = data_path + '/MEG/sample/sample_audvis_raw.fif'
subjects_dir = data_path + '/subjects'
subject = 'sample'
trans = data_path + '/MEG/sample/sample_audvis_raw-trans.fif'
info = mne.io.read_info(raw_fname)
fig = mne.viz.plot_alignment(info, trans, subject=subject, dig=True,
                             meg=[],
                             # meg=['helmet', 'sensors'],
                             coord_frame='mri',
                             subjects_dir=subjects_dir,
                             surfaces=['head-dense'])


dists = mne.dig_mri_distances(info, trans, subject,
                              subjects_dir=subjects_dir)

images = []

views = ['frontal', 'lateral', 'medial']
views += ['axial', 'rostral', 'coronal']

for view in views:
    mne.viz.set_3d_view(fig, **views_dicts['both'][view])
    images.append(fig.plotter.screenshot())

images = np.concatenate(
    [np.concatenate(images[:3], axis=1),
     np.concatenate(images[3:], axis=1)],
    axis=0)

plt.title(f'Distance from {len(dists)} digitized points to head: '
          f'{1e3 * np.mean(dists):.2f} mm')
plt.imshow(images)
plt.axis('off')
plt.show()
```

thanks @GuillaumeFavelier for the help and the Neurospin group @virvw @SophieHerbst 

this should be used to produce our automatic report on coregistration.

thoughts @larsoner @drammock @jasmainak @SherazKhan ?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -29,6 +29,10 @@ Enhancements
 
 - Update ``surfaces`` argument in :func:`mne.viz.plot_alignment` to allow dict for transparency values, and set default for sEEG data to have transparency (:gh:`8445` by `Keith Doelling`_)
 
+- Add support for ``mri_fiducials='estimated'`` in :func:`mne.viz.plot_alignment` to allow estimating MRI fiducial locations using :func:`mne.get_mni_fiducials` (:gh:`8553` by `Eric Larson`_)
+
+- Update default values in :ref:`mne coreg` and :func:`mne.viz.plot_alignment` for clearer representation of MRI and digitized fiducial points (:gh:`8553` by `Alex Gramfort`_ and `Eric Larson`_)
+
 - Add ``n_pca_components`` argument to :func:`mne.viz.plot_ica_overlay` (:gh:`8351` by `Eric Larson`_)
 
 - Add :func:`mne.stc_near_sensors` to facilitate plotting ECoG data (:gh:`8190` by `Eric Larson`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -29,7 +29,7 @@ Enhancements
 
 - Update ``surfaces`` argument in :func:`mne.viz.plot_alignment` to allow dict for transparency values, and set default for sEEG data to have transparency (:gh:`8445` by `Keith Doelling`_)
 
-- Add support for ``mri_fiducials='estimated'`` in :func:`mne.viz.plot_alignment` to allow estimating MRI fiducial locations using :func:`mne.get_mni_fiducials` (:gh:`8553` by `Eric Larson`_)
+- Add support for ``mri_fiducials='estimated'`` in :func:`mne.viz.plot_alignment` to allow estimating MRI fiducial locations using :func:`mne.coreg.get_mni_fiducials` (:gh:`8553` by `Eric Larson`_)
 
 - Update default values in :ref:`mne coreg` and :func:`mne.viz.plot_alignment` for clearer representation of MRI and digitized fiducial points (:gh:`8553` by `Alex Gramfort`_ and `Eric Larson`_)
 
--- a/mne/defaults.py
+++ b/mne/defaults.py
@@ -49,10 +49,10 @@ DEFAULTS = dict(
                      markersize=4),
     coreg=dict(
         mri_fid_opacity=1.0,
-        dig_fid_opacity=0.3,
+        dig_fid_opacity=1.0,
 
-        mri_fid_scale=1e-2,
-        dig_fid_scale=3e-2,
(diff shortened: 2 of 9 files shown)


virvw:
@agramfort use transparent skin but filled colors for fiducials and digitized points so we get a sense of depth

virvw:
add centroid for the nasion and for the ear fiducials as those are typically recorded with good precision by the researchers

larsoner:
> add centroid for the nasion and for the ear fiducials as those are typically recorded with good precision by the researchers

I think these were originally large because @christianbrodbeck had the opposite sense, that there was some ambiguity in it (from a digitizer accuracy standpoint, probably). But I agree that with well behaved subjects you can more precise than we show here, and it shouldn't hurt to make them the same size as the MRI ones. I'd rather just shrink the markers rather than make a large translucent one and a small opaque one.

> use transparent skin

We tried this default in https://github.com/mne-tools/mne-python/pull/8446 and decided against it. So I think if people want this they should use the new `dict` support for `surfaces` by doing something like `surfaces=dict(head=0.7)`.

> filled colors for fiducials

Part of the reason the fids are translucent is because in theory the `bem/subject-fiducials.fif` that you mark in MRI space should align perfectly / overlap with the fiducials recorded using the digitizer, so we need some way to see two sets of points in the same space. This is also why the MRI fids are opaque and the digitizer fids are translucent (and much larger, but I agree this isn't really necessary, as above). One option would be to make the MRI fiducials some other shape like a square, and the digitized ones stay a circle. That way you can see both even if they line up. (I think it's okay to use squares even though they have sidedness / orientation -- we can align the sides with the XYZ/RAS of the MRI, so actually get a bit more information out of it.)

christianbrodbeck:
If I recall correctly the reason for making the digitizer fiducials large was mainly to be able to see them together with the MRI fiducials, and to see the even when, for example, they're sunk slightly under the MRI head surface. With template brains the expected alignment is less perfect.

drammock:
+1 for being consistent, and I think the yellowish head is more pleasant / easier to see than the gray one. Regarding fiducial size, etc, I think making them somewhat smaller would be OK, and I like the idea of MRI fids and digitizer fids being different geometric solids...  actually I think a cube for one and a regular octahedron for the other would be ideal --- if you pick the relative sizes right it would make it really easy to tell if they line up:

Screenshot_2020-11-22_12-35-57

agramfort:
@drammock do you want to lead this following my gist above?

larsoner:
+1 for square plus something else. I'm not sure VTK has an octahedron glyph, if not then cube+circle seems fine (cube for MRI, circle for dig) and can also have a nice clean overlap if their volumes are identical. This is also consistent with how dig points show up as circles anyway...

I would start by trying to make them equal-volume, leading to a cube-side-to-circle-radius ratio of:
```
>>> (3 / (4 * pi)) ** (1/3)
0.6203504908994001
```
This way perfect alignment means some parts of the circle will protrude, and the corners of the square will, too.

drammock:
> @drammock do you want to lead this following my gist above?

how urgent is this? it might take me some time to get up to speed on vtk etc.

larsoner:
Not that urgent, but I think I can probably do it quickly -- let me take a look

larsoner:
Okay I took a try with the cube. I agree octahedron would be better but it will be a pain to implement.
```
$ mne coreg -d subjects -s sample --fif MEG/sample/sample_audvis_raw.fif --trans MEG/sample/sample_audvis_raw-trans.fif
```
Gives:

Screenshot from 2020-11-23 16-15-31

Screenshot from 2020-11-23 16-15-38

And with "fit fid" you get a different alignment where the nasion is aligned, you can see they are on top of each other at least:

And `tutorials/source-modeling/plot_source_alignment.py` gives:

Screenshot from 2020-11-23 16-12-27

larsoner:
Okay got `octahedron` working, ready for review/merge from my end:

Screenshot from 2020-12-09 11-46-39

Screenshot from 2020-12-09 11-45-56

agramfort (commented):
LGTM

https://24039-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_source_alignment.html#sphx-glr-auto-tutorials-source-modeling-plot-source-alignment-py

agramfort:
thx @larsoner !

PR #8554, opened 2020-11-22
FIX:  qt5 macOS 11 compatibility

### LINKED ISSUE #8525: BUG: Big Sur compat
More of an FYI than anything else -- PyQt5 appears not to work on Big Sur. I've tried 5.12 and 5.15.1, and both basically hang as soon as you make a mayavi, PyVista, or matplotlib Qt5Agg window.

So those thinking about upgrading, don't do so if you want to be able to use Python. For matplotlib at least we can use the MacOSX backend in the meantime, but 3D plotting is now (hopefully temporarily) totally broken for people who have upgraded. :(

Or maybe I'm the only one with this issue. If someone else upgrades, feel free to try it out and let me know!

christianbrodbeck:
#### Reference issue
Fixes #8525 by setting the environment variable before launching GUIs.

#### TODO
 - [x] test on macOS 10

### DIFF
--- a/mne/gui/_backend.py
+++ b/mne/gui/_backend.py
@@ -4,6 +4,7 @@
 # License: BSD (3-clause)
 
 import os
+import sys
 
 from ..utils import warn, _check_pyqt5_version
 
@@ -35,6 +36,9 @@ def _check_backend():
     backend = _get_pyface_backend()
     if backend == 'qt4':
         _check_pyqt5_version()
+        # QT 5 macOS 11 compatibility:
+        if sys.platform == 'darwin' and 'QT_MAC_WANTS_LAYER' not in os.environ:
+            os.environ['QT_MAC_WANTS_LAYER'] = '1'
     else:
         msg = ("Using the currently selected Pyface backend %s is not "
                "recommended, and it might not work properly. We recommend "


mne/gui/_backend.py
```diff
@@ -35,6 +36,9 @@ def _check_backend():
     backend = _get_pyface_backend()
     if backend == 'qt4':
         _check_pyqt5_version()
+        # QT 5 macOS 11 compatibility:
+        if sys.platform == 'darwin' and 'QT_MAC_WANTS_LAYER' not in os.environ:
```
hoechenberger, line 40: I think we should check whether we're actually running on Big Sur, and if we are, we don't even need to check whether the environment variable exists -- we'll need to set it anyway in order to make things work

hoechenberger (changes requested):

mne/gui/_backend.py
cbrnr: Or even simpler, just set the variable if running on `darwin` &ndash; I just checked on macOS Mojave and my PyQt5 app still works after setting it.

christianbrodbeck: Thanks for testing @cbrnr!
@hoechenberger I am reluctant to silently overwrite an environment variable that the user already set, unless you are sure that the user would never want it to be a different value

cbrnr: I think it's highly unlikely that a user set this variable for something other than fixing the PyQt5 bug. In this case, the value must be 1, so if you do not want to overwrite an existing variable the problem isn't fixed. I would set the variable to 1 when running on macOS, and if that variable has already been set to a value other than 1 I'd issue a warning. After all, this is a workaround, which hopefully we won't have to use for long.

hoechenberger: But with the current implementation you're also changing the environment permanently (setting a new variable if it doesn't already exist, changing behavior)

In the Study Template we had to deal with a similar issue and solved it this way:
https://github.com/mne-tools/mne-study-template/blob/766cc0e2dcab1fecf61804e1b013ba37988f212c/run.py#L135-L148

Wonder if this is something we could re-use here too..?

cbrnr: @hoechenberger yes, I agree that we should restore the environment.

cbrnr: But in this case, restoring the env might be simpler because we only have to reset (either delete or reset the value) of one variable, right?

hoechenberger: Yes but then you'd have to special-case "did it exist before the user started MNE or not?"

larsoner: We have `modified_env` context manager in `mne.utils` that does all this magic for you

cbrnr: Yes, why not. If you want to re-use a more general version go ahead, but the example you linked doesn't seem to work:

1. Updating the `os.environ` dict doesn't remove any entries that don't exist anymore.
2. Setting the signal handlers requires a [function with two parameters](https://docs.python.org/3/library/signal.html#signal.signal), and you want to keep their original behavior plus restoring the environment - currently, no exception is raised for these signals.

hoechenberger: Oh wow, thank you, obviously this has never been properly tested!

hoechenberger: > 2\. and you want to keep their original behavior plus restoring the environment - currently, no exception is raised for these signals

No sure what you mean here exactly, we have `atexit` to restore the env when the script terminates :)

christianbrodbeck: So we would have to combine the platform conditional with `modified_env`, isn't that a bit overkill? I just wanted to avoid messing with a power user's setting, but did not think setting this variable would interfere with the average use case.

hoechenberger: Sorry I didn't mean to make this more complicated than it needs to be.

I'd also be finde with simply setting the env variable no matter what, and setting ourselves a reminder to revert this change as soon as new (fixed) MPL binaries have been released.

hoechenberger (approved):

mne/gui/_backend.py
larsoner: @hoechenberger FWIW I'm not sure why you have any `atexit`, no env vars are saved/written when doing `python`:
```
$ echo $foo

$ python
Python 3.8.6 (default, Sep 25 2020, 09:36:53) 
[GCC 10.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ['foo'] = 'bar'
>>> exit()
$ echo $foo

$
```

larsoner:
There might be other places that we want something like this -- for example in `matplotlib` code -- so I think we probably want a FAQ entry or so. And PyQt5 5.12.2 will have this on by default. But this already seems like a step in the right direction so I'll merge. Thanks @christianbrodbeck

mne/gui/_backend.py
hoechenberger: wait... what?! Wow. I thought those were changed permanently. Good to know!!

cbrnr: @hoechenberger if you do e.g. `signal.signal(signal.SIGINT, lambda: pass)` (like it's implemented right now except instead of `lambda: pass` you have `_restore_env` (a function with no parameters), you will (1) get an error because you need to pass a function with two parameters, and (2) even if you pass a function with the correct signature (e.g. `signal.signal(signal.SIGINT, lambda x, y: pass)`) you need to make sure that the previous behavior of `SIGINT` is retained (i.e. a `KeyboardInterruptException` should be thrown at the end to halt the program), otherwise the program will continue to run even if the user presses Ctrl + C.

PR #8558, opened 2020-11-23
MRG, BUG: Fix EDF header parsing

### LINKED ISSUE #8544: EDF/EDF+ headers parsed incompletely/incorrectly
When reading in EDF/EDF+/BDF files with MNE, there are a few issues with the current parser:

**Patient/Participant Info**

https://github.com/mne-tools/mne-python/blob/8b0b8a967b1a8e2fd88c47221f535d73ca7d6fb2/mne/io/edf/edf.py#L524-L530

According to the [EDF+ spec](https://www.edfplus.info/specs/edfplus.html#additionalspecs), participant info is stored in the format of "{id} {sex} {birthdate} {name}" (minus the braces), with those four fields being separated by single spaces (and not containing any spaces themselves). At present, MNE only parses the first two fields, and incorrectly identifies the second field as containing the participant name.

**Recording Info**

https://github.com/mne-tools/mne-python/blob/8b0b8a967b1a8e2fd88c47221f535d73ca7d6fb2/mne/io/edf/edf.py#L532-L534

Currently, EDF/BDF recording info isn't parsed at all by MNE, and the string read in by the above lines isn't part of the exported header info so the info also can't be parsed later. As per the EDF+ spec, it takes the format "Startdate {startdate} {admin_code} {technician} {equipment}", also separated by spaces but not containing any.

The 'startdate' field is particularly important to parse if available, because it contains the meas_date for the file with a 4-digit year (e.g. '02-AUG-2014'), whereas the normal 'meas_date' region of the header only retains the last two digits (e.g. '02.08.14'). For regular use this is fine, but it creates issues with the BIDS-EEG spec which recommends the dates of recordings should be shifted to a year before 1925 during anonymization. A solution would be to attempt to parse the start date in the recording info and use the actual 'meas_date' value as a fallback.

**Date**

https://github.com/mne-tools/mne-python/blob/8b0b8a967b1a8e2fd88c47221f535d73ca7d6fb2/mne/io/edf/edf.py#L536-L540

As per the EDF+ spec, the earliest supported year for EDF files is 1985. If a 4-digit year isn't specified in the recording info, the parser should assume that a year of < 85 is a 20xx year and not an 19xx one. This will probably never affect anyone, but it might as well match the official docs.

### LINKED ISSUE #8551: BUG: Some tests require dipy >= 1.3
There are a number of tests in `mne/test/test_morph.py` that will only pass if `dipy>=1.3.0`. This should probably be addressed outside of this PR, but could we require `dipy>=1.3.0`, at least for those particular tests?

larsoner:
1. Implements a variant of @a-hurst's code from #8544 (most of the PR)
2. Fixes dipy decorator (1 line)
3. Fixes str-as-int parsing from `gitter` (1 line)
4. Adds stub test with `XXX` for carrying over `edf_info['subject_info']` to `raw.info['subject_info']`, which we should probably do to the extent possible (but in another PR)

@a-hurst the one thing I didn't follow was (what is keeping this WIP from my end):

> My only other change was adding recording_info=meas_id, to the edf_info.update() call further down so the recording info gets carried over to the final Raw object.

Can you take a look, and clarify what needs to change in / be added to this PR? And let me know if you're happy with the GitHub URL I put for your name link?

@sappelhoff @cbrnr can you look to see if this makes sense in light of discussions in #8544?

Closes #8544
Closes #8551

### DIFF
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -510,7 +510,7 @@ def _parse_prefilter_string(prefiltering):
 
 
 def _edf_str_int(x, fid=None):
-    return int(x.decode().rstrip('\x00'))
+    return int(x.decode().split('\x00')[0])
 
 
 def _read_edf_header(fname, exclude):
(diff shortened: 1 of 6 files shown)


a-hurst:
Looks good, thanks for this! The part you're asking about refers to this section:

```python
        # Populate edf_info
        edf_info.update(
            ch_names=ch_names, data_offset=header_nbytes,
            digital_max=digital_max, digital_min=digital_min,
            highpass=highpass, sel=sel, lowpass=lowpass, meas_date=meas_date,
            n_records=n_records, n_samps=n_samps, nchan=nchan,
            subject_info=patient, physical_max=physical_max,
            physical_min=physical_min, record_length=record_length,
            subtype=subtype, tal_idx=tal_idx)
```
In my local copy, I added `recording_info=meas_id, ` after `subject_info=patient, ` so that the recording info actually gets passed to the final raw object (before it was just discarded). I'm not sure if that's the proper naming convention or not, but I couldn't find any info on whether MNE standardized the naming and location of that kind of extended metadata.

Also, if you want to do some more extensive testing of the new header parsing you can base it off the tests I wrote for mne-bids here: 

https://github.com/mne-tools/mne-bids/blob/18cc4f912495e8de8ab30d9c3077b19c923bebf4/mne_bids/tests/test_copyfiles.py#L116-L142

The MNE test .edf and .bdf files I looked at don't seem to have any of those header fields actually filled out, so manually adding the header info pre-test might be useful!

mne/io/edf/edf.py
```diff
@@ -522,27 +522,56 @@ def _read_edf_header(fname, exclude):
...
-        hour, minute, sec = [int(x) for x in
-                             re.findall(r'(\d+)', fid.read(8).decode())]
-        century = 2000 if year < 50 else 1900
+        meas_time = fid.read(8).decode('latin-1')
+        hour, minute, sec = [int(x) for x in meas_time.split('.')]
         try:
-            meas_date = datetime(year + century, month, day, hour, minute, sec,
+            meas_date = datetime(year, month, day, hour, minute, sec,
                                  tzinfo=timezone.utc)
+            print(meas_date)
```
cbrnr, line 572: I guess this needs to go.

mne/io/edf/tests/test_edf.py
```diff
@@ -399,13 +407,17 @@ def test_edf_annot_sub_s_onset():
...
-    tempdir = _TempDir()
+    tempdir = str(tmpdir)
     with open(edf_path, 'rb') as f:  # read valid test file
         edf = bytearray(f.read())
+
     # original date in header is 29.04.14 (2014-04-29) at pos 168:176
+    # but we also use Startdate if available,
+    # which starts at byte 88 and is b'Startdate 29-APR-2014 X X X'
     # create invalid date 29.02.14 (2014 is not a leap year)
+    edf[101:104] = b'FEB'
```
cbrnr, line 420: OK, but now we don't test an invalid date at pos 168:176 anymore, right? I'd prefer to test both possibilities.

mne/io/edf/tests/test_edf.py
```diff
@@ -420,4 +432,7 @@ def test_invalid_date():
         read_raw_edf(op.join(tempdir, "temp.edf"), preload=True)
 
 
-run_tests_if_main()
+def test_empty_chars():
+    """Test blank char support."""
+    # from gitter
```
cbrnr, line 437: This comment doesn't really help.

cbrnr (commented):
LGTM, let me know if you want me to take a look at the missing feature once you've implemented it.

sappelhoff (approved):
LGTM overall once Clemens' comments are taken into account

larsoner:
> In my local copy, I added recording_info=meas_id, after subject_info=patient, so that the recording info actually gets passed to the final raw object (before it was just discarded).

It gets passed into a dict that ends up living only in a private attribute `raw._raw_extras[0]`, so people should not rely on these values to exist (see extensive discussion [here](https://github.com/mne-tools/mne-python/issues/8349)), so I'm -1 on adding these until we decide to make some EDF attributes public.

larsoner:
@cbrnr made the tests more complete, should cover what you asked about

cbrnr:
@larsoner did you push the changes?

larsoner:
nope, forgot the `-f` so it didn't go through -- it's there now, though!

cbrnr (approved):

sappelhoff (approved):

PR #8560, opened 2020-11-23
VIZ, FIX: epochs.plot() scalebars missing

drammock:
This fixes missing scalebars (toggled by pressing `'s'`) on epochs plots, and adds a test that would have caught their missingness.


agramfort:
thx @drammock

PR #8567, opened 2020-11-24
DOC: update docstring for picks

drammock:
This adds a note to the docstring for most `picks` parameters, reminding users that explicitly passing a bad channel name or index will result in that bad channel being included.  It also slightly refactors the picks docdict entries to make it easier to see what is included in each, and fixes a tiny bug where a space is missing between a word and a `(`.

note that the docstrings for ICA viz functions that have picks parameters do not get this reminder, since they refer to components, not channels.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -429,27 +429,30 @@ axes : list of Axes | None
 """
 
 # Picks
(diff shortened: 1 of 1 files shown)


agramfort (approved):

larsoner:
Thanks @drammock

PR #8574, opened 2020-11-25
MRG, ENH: Remove 15-char limit for FIF

### LINKED ISSUE #8312: Why is the channel name length limited to 15 characters?

larsoner:
- [x] Get things working with the block-based method
- [x] Get other projects on board (taken care of in #8348)
- [x] Get https://github.com/mne-tools/fiff-constants/pull/33/ merged
- [x] Update PR with newer `fiff-constants` commit, check `'bads'` handling/replacement
- [x] Add test where a file is written with long names, then `monkeypatch` to make the tag empty on read, ensure it can still be written and read
- [x] Add to MNE-MATLAB

Closes #8348
Closes #8312

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -25,12 +25,14 @@ from ..utils import (verbose, logger, warn,
                      _get_stim_channel, _check_fname)
 from ..io.compensator import get_current_comp
 from ..io.constants import FIFF
-from ..io.meas_info import anonymize_info, Info, MontageMixin, create_info
+from ..io.meas_info import (anonymize_info, Info, MontageMixin, create_info,
+                            _rename_comps)
 from ..io.pick import (channel_type, pick_info, pick_types, _picks_by_type,
                        _check_excludes_includes, _contains_ch_type,
                        channel_indices_by_type, pick_channels, _picks_to_idx,
                        _get_channel_types, get_channel_type_constants,
                        _pick_data_channels)
+from ..io.tag import _rename_list
 from ..io.write import DATE_NONE
 from ..io._digitization import _get_data_as_dict_from_dig
 
@@ -471,13 +473,14 @@ class SetChannelsMixin(MontageMixin):
             warn(msg.format(", ".join(sorted(names)), *this_change))
         return self
 
-    @fill_doc
-    def rename_channels(self, mapping):
+    @verbose
+    def rename_channels(self, mapping, allow_duplicates=False, verbose=None):
         """Rename channels.
 
         Parameters
         ----------
-        %(rename_channels_mapping)s
+        %(rename_channels_mapping_duplicates)s
+        %(verbose_meth)s
 
         Returns
         -------
@@ -494,7 +497,7 @@ class SetChannelsMixin(MontageMixin):
         from ..io import BaseRaw
 
         ch_names_orig = list(self.info['ch_names'])
-        rename_channels(self.info, mapping)
+        rename_channels(self.info, mapping, allow_duplicates)
 
         # Update self._orig_units for Raw
         if isinstance(self, BaseRaw) and self._orig_units is not None:
@@ -1150,17 +1153,16 @@ class InterpolationMixin(object):
         return self
 
 
-@fill_doc
-def rename_channels(info, mapping):
+@verbose
+def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
     """Rename channels.
-
-    .. warning::  The channel names must have at most 15 characters
 
     Parameters
     ----------
     info : dict
         Measurement info to modify.
-    %(rename_channels_mapping)s
+    %(rename_channels_mapping_duplicates)s
+    %(verbose)s
     """
     _validate_type(info, Info, 'info')
     info._check_consistency()
@@ -1187,12 +1189,6 @@ def rename_channels(info, mapping):
     for new_name in new_names:
         _validate_type(new_name[1], 'str', 'New channel mappings')
 
-    bad_new_names = [name for _, name in new_names if len(name) > 15]
-    if len(bad_new_names):
-        raise ValueError('Channel names cannot be longer than 15 '
-                         'characters. These channel names are not '
-                         'valid : %s' % new_names)
-
     # do the remapping locally
     for c_ind, new_name in new_names:
         for bi, bad in enumerate(bads):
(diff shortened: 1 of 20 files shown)


larsoner:
@agramfort this is ready for review from my end. Once we're happy with this, I can make the corresponding changes in MNE-MATLAB.

Should we try to get this into 0.22, or make it an early 0.23 PR? I'm leaning toward the latter so we at least get a few months in of dev use

agramfort:
let's merge this after the release

larsoner:
Agreed. @bloyl would you be interested in doing the companion PR to MNE-MATLAB? My MATLAB brain has atrophied over the last many years of using Python...

larsoner:
Companion PR is in https://github.com/mne-tools/mne-matlab/pull/21, so these two are ready for review/merge from my end.

mne/channels/channels.py
```diff
@@ -1197,13 +1193,19 @@ def rename_channels(info, mapping):
...
     # do the remapping in info
     info['bads'] = bads
+    rename = dict()
     for ch, ch_name in zip(info['chs'], ch_names):
+        rename[ch['ch_name']] = ch_name
         ch['ch_name'] = ch_name
+    _rename_comps(info.get('comps', []), rename)  # .get b/c fwd info omits it
+    for proj in info['projs']:
+        proj['data']['col_names'] = _rename_list(
+            proj['data']['col_names'].copy(), rename)
```
agramfort, line 1208: this pattern is a bit misleading. The .copy() suggests it's inplace (which it is) but you sill affect the output.
If operation is inplace I would not return anything so it's not ambiguous

mne/evoked.py
```diff
@@ -1080,7 +1081,9 @@ def _read_evoked(fname, condition=None, kind='average', allow_maxshield=False):
                 raise ValueError('Number of channels and number of '
                                  'channel definitions are different')
 
+            rename_map = _read_extended_ch_info(chs, my_evoked, fid)
             info['chs'] = chs
+            _rename_list(info['bads'], rename_map)
```
agramfort, line 1086: here it's explicit and clear what it does

mne/forward/forward.py
```diff
@@ -918,18 +920,17 @@ def write_forward_meas_info(fid, info):
...
         write_int(fid, FIFF.FIFF_NCHAN, len(info['chs']))
-        for k, c in enumerate(info['chs']):
-            #   Scan numbers may have been messed up
-            c = deepcopy(c)
-            c['scanno'] = k + 1
-            write_ch_info(fid, c)
+        _write_ch_infos(fid, info['chs'], False, rename_map)
     if 'bads' in info and len(info['bads']) > 0:
         #   Bad channels
+        bads = _rename_list(info['bads'].copy(), rename_map)
```
agramfort, line 931: same here

mne/io/ctf_comp.py
```diff
@@ -67,13 +67,17 @@ def read_ctf_comp(fid, node, chs, verbose=None):
     chs : list
         The list of channels from info['chs'] to match with
         compensators that are read.
+    rename : dict | None
+        The channel renaming to use.
```
agramfort, line 71: as it's public API can you be more explicit? Something like 

    A dictionary mapping the old channel to a new channel name
    e.g. {'MLC11-606' : 'MLC11'}.

I took inspiration from rename_channels docstring  and used a CTF channel name

mne/io/meas_info.py
```diff
@@ -1085,6 +1080,7 @@ def read_meas_info(fid, tree, clean_bads=False, verbose=None):
         elif kind == FIFF.FIFF_MNE_KIT_SYSTEM_ID:
             tag = read_tag(fid, pos)
             kit_system_id = int(tag.data)
+    rename = _read_extended_ch_info(chs, meas_info, fid)
```
agramfort, line 1083: the name `rename` suggests it should be a bool. It took me a bit of time
to understand that what you call rename here corresponds to the `mapping`
parameter in `rename_channels`. Would ch_names_mapping be too long?

mne/io/meas_info.py
```diff
@@ -1434,6 +1430,47 @@ def read_meas_info(fid, tree, clean_bads=False, verbose=None):
...
+                if key == 'ch_name':
+                    rename[ch[key]] = data
+                ch[key] = data
+        _update_ch_info_named(ch)
+    # we need to return the "rename" map so that we can also rename the
+    # bad channels
+    return rename
+
+
+def _rename_comps(comps, rename_map):
```
agramfort, line 1466: here you call it rename_map

mne/io/meas_info.py
```diff
@@ -1600,12 +1637,14 @@ def write_meas_info(fid, info, data_type=None, reset_range=True):
         write_coord_trans(fid, info['dev_ctf_t'])
 
     #   Projectors
-    _write_proj(fid, info['projs'])
+    rename_map = _make_rename_map(info['chs'])
```
agramfort, line 1640:
```suggestion
    ch_names_mapping = _make_ch_names_mapping(info['chs'])
```

?

mne/io/meas_info.py
```diff
@@ -2341,3 +2393,38 @@ def _dict_unpack(obj, casts):
     n = len(obj[list(casts)[0]])
     return [{key: cast(obj[key][ii]) for key, cast in casts.items()}
             for ii in range(n)]
+
+
+def _make_rename_map(chs):
```
agramfort, line 2398:
```suggestion
def _make_ch_names_mapping(chs):
```

agramfort (commented):
works great !

just a few suggestions to make the code more self explanatory.

thx heaps @larsoner !

mne/io/ctf_comp.py
larsoner: I think it's mostly an accident that this is public -- I don't think users are ever meant to use it. I'd actually rather make it private (by deprecation if need be) rather than actually support people using it. It's misleading because they shouldn't ever need to use `ch_names_mapping` themselves, and if they do, they're not likely to do something good/correct with it. So I consider the vagueness here might actually be more of a feature than a bug...

agramfort:
it took me some time to understand that's why I suggested to rename to
ch_names_mapping
?

also is there a way to hide this for any public function?

larsoner:
> it took me some time to understand that's why I suggested to rename to ch_names_mapping

I'm not sure what you're referring to here -- I made this change...?

Pushed a change to preserve the API of read_ctf_comp. Failures are just `pip pre`

agramfort (approved):

agramfort:
thanks @larsoner !

PR #8575, opened 2020-11-26
MNT: Add brain_gc fixture to test_process_clim_plot

GuillaumeFavelier:
This PR adds the `brain_gc` fixture in `test_process_clim_plot` (I missed it in https://github.com/mne-tools/mne-python/pull/8427) and fixes the issues with memory:

```
_______________________________________________________________________________ ERROR at teardown of test_process_clim_plot[pyvista] _______________________________________________________________________________
mne/conftest.py:522: in brain_gc
    assert len(bad) == 0, 'VTK objects linger:\n' + '\n'.join(bad)
E   AssertionError: VTK objects linger:
E     vtkXOpenGLRenderWindow
E     vtkGenericRenderWindowInteractor
E   assert 2 == 0
E     +2
E     -0
```

It's also important for https://github.com/mne-tools/mne-python/pull/8335 (that's how I discovered the issue in the first place)

### DIFF
--- a/mne/viz/tests/test_3d.py
+++ b/mne/viz/tests/test_3d.py
@@ -373,10 +373,12 @@ def test_plot_alignment(tmpdir, renderer):
 @testing.requires_testing_data
 @requires_pysurfer
 @traits_test
-def test_process_clim_plot(renderer_interactive):
+def test_process_clim_plot(renderer_interactive, brain_gc):
     """Test functionality for determining control points with stc.plot."""
+    is_pyvista = renderer_interactive._get_3d_backend() == 'pyvista'
     sample_src = read_source_spaces(src_fname)
-    kwargs = dict(subjects_dir=subjects_dir, smoothing_steps=1)
+    kwargs = dict(subjects_dir=subjects_dir, smoothing_steps=1,
+                  time_viewer=False, show_traces=False)
 
     vertices = [s['vertno'] for s in sample_src]
     n_time = 5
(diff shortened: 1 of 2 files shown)


agramfort (approved):

mne/viz/tests/test_3d.py
```diff
@@ -391,6 +393,10 @@ def test_process_clim_plot(renderer_interactive):
     stc.plot(colormap='hot', clim='auto', **kwargs)
     stc.plot(colormap='mne', clim='auto', **kwargs)
     stc.plot(clim=dict(kind='value', lims=(10, 50, 90)), figure=99, **kwargs)
+    # figure has been set so the internal ref needs to be clean to avoid
+    # lingering VTK objects in memory
+    if is_pyvista:
+        renderer_interactive.backend._FIGURES.clear()
```
larsoner, line 399: I don't fully understand -- doesn't this indicate that the `close_all` that happens at the end of this test is somehow not working? How do we get a figure set/created in a way that our close-all fixture will not close it?

GuillaumeFavelier: We keep an internal reference to mimic `mayavi`'s behaviour when `figure` (int) is used.

larsoner:
To put it differently, whenever you add `brain_gc` to a test, any failures should be fixed in `_brain.py`, not in the test itself -- this is because `brain_gc` does a [`close_all`](https://github.com/mne-tools/mne-python/blob/master/mne/conftest.py#L507) sort of operation, and this should close all figures. If it doesn't, it suggests some problem with our `close_all` that should be fixed.

mne/viz/tests/test_3d.py
larsoner: See follow-up comment -- I think this means `close_all` needs to be updated, or that we need to update our fixture to call another `close_all` for the case of int-passing or something.

larsoner (approved):
LGTM feel free to merge once CIs are green @GuillaumeFavelier

PR #8576, opened 2020-11-27
MRG: Use overlay compositing

GuillaumeFavelier:
This PR is a prototype that represents an alternative to VTK alpha blending. It computes the final brain activation as one 'texture map' which is the result of alpha compositing of the overlays.

The pros:

* only one mesh/actor/mapper in memory for one Brain (no duplication of vertices/triangles/normals anymore)
* no visual artifacts or z-fighting caused by overlapped surfaces
* the list of overlays can be modified and updated on request especially to simulate z-order

The cons:

* the final result requires blending of all the overlays (which can be expensive). I count on some smart optimization of the `_alpha_over` function to mitigate this (I might need help here). It's also possible to use intermediate results to speedup the computation.

This is still very early results.

Closes https://github.com/mne-tools/mne-python/issues/7599


larsoner:
I would much rather let the GPU deal with all of this if possible because this does not scale as well as the GPU will. But if there are no suitable GPU solutions then I CPU is better than wrong GPU

agramfort:
would this be helpful too for notebook backend?

larsoner:
> But if there are no suitable GPU solutions then I CPU is better than wrong GPU

I'm starting to think there are no suitable GPU solutions. @GuillaumeFavelier before going too far here, maybe try a basic GPU and basic CPU example, totally outside `Brain`, where you add, say, N overlays both ways (N actors alpha blended for GPU, 1 actor with N values alpha blended in the CPU) to one hemisphere of data? Then we can see what the performance hit is like as N is varied.

Or if this compositing is 90% done anyway, you can do the same thing probably by adding STC data, then parcellation, then labels, etc., and profile directly with some repeated calls to `screenshot` / `Render` or so.

I think it's worth going through this exercise even if CPU is our only option at least so we know what we're getting into.

Bonus points if you do GPU without depth peeling, GPU with depth peeling, and CPU (no depth peeling, not needed)...

GuillaumeFavelier:
> would this be helpful too for notebook backend?

The artifacts in https://github.com/mne-tools/mne-python/issues/7599#issuecomment-733626871 are most likely also present with the `notebook` 3d backend.

>  try a basic GPU and basic CPU example

got it :+1: 

> Or if this compositing is 90% done anyway

This version is very early work and limited to `Brain`. Refactoring is needed to extend to label, annotation, ...

larsoner:
> This version is very early work and limited to Brain. Refactoring is needed to extend to label, annotation, ...

Could you create a `_LayeredMesh` class that has 

1. an `__init__` that takes `rr, tris` args
2. a method `add_overlay(scalars, vertices, smoothing_matrix, cmap, ..., zorder)` where `scalars` is shape `(len(vertices), n_time)` or `(len(rr), n_time)` for full-res data or `(len(rr), 1)` for static data (like an annotation)
3. a `set_index` method that changes the data index that is displayed

It makes the implementation at the `Brain` end very clean, simple, and readable. And if we decide to switch at some point from CPU to GPU compositing, it just requires changing `_LayeredMesh` to make multiple meshes on the GPU rather than a single mesh with CPU-based alpha blending. WDYT?

You could even start by making this class using the `master` method, namely by creating multiple meshes rather than doing CPU blending. Then once that works, you only need to change / update the rendering/blending parts in `_LayeredMesh` rather than anything in `_Brain`.

GuillaumeFavelier:
`mayavi` brain tests fail because there is no option for scalars as rgba texture.

GuillaumeFavelier:
Just my opinion but I somehow find the design more elegant and it feels like less "hacks" are needed.

I updated `add_data`, `add_label` and `add_annotation`.

GuillaumeFavelier:
The implementation is not exactly what you describe in https://github.com/mne-tools/mne-python/pull/8576#issuecomment-738016809 but it's flexible @larsoner 

Always possible to improve

larsoner:
The plot on [master](https://mne.tools/dev/auto_examples/inverse/plot_source_space_snr.html#sphx-glr-auto-examples-inverse-plot-source-space-snr-py):



Looks different from (and more correct than) [the one on this PR](https://23953-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_source_space_snr.html#sphx-glr-auto-examples-inverse-plot-source-space-snr-py), and I don't think it's just due to the scale bar difference -- note the odd yellow splotches:



Also there is a difference in the automatic scale bar (should be about -100 to 40 and is on `master`, in this PR it changes to -70 to 38).

So it seems like there is both a blending-logic bug and a data scaling bug?

mne/viz/_brain/_brain.py
```diff
@@ -47,6 +47,157 @@ def safe_event(fun, *args, **kwargs):
...
+
+    def _compute_over(self, B, A):
+        alpha_a = A[:, 3]
+        alpha_b = B[:, 3]
+        alpha_c = alpha_a + alpha_b * (1 - alpha_a)
+        a = A[:, :3]
+        b = B[:, :3]
+        c = (a.T * alpha_a).T + (b.T * alpha_b * (1 - alpha_a)).T
+        c = (c.T / alpha_c).T
+        return np.c_[c, alpha_c]
```
larsoner, line 127: This is equivalent but uses broadcasting and fewer copies:
```suggestion
        assert A.ndim == B.ndim == 2
        assert A.shape[1] == B.shape[1] == 4
        A_w = A[:, 3:]  # * 1
        B_w = B[:, 3:] * (1 - A_w)
        C = A.copy()
        C[:, :3] *= A_w
        C[:, :3] += B[:, :3] * B_w
        C[:, 3:] += B_w
        C[:, :3] /= C[:, 3:]
        return np.clip(C, 0, 1, out=C)
```

mne/viz/_brain/_brain.py
```diff
@@ -47,6 +47,157 @@ def safe_event(fun, *args, **kwargs):
...
+        self._vertices = vertices
+        self._triangles = triangles
+        self._normals = normals
+
+        self._polydata = None
+        self._actor = None
+        self._is_mapped = False
+
+        self._cache = None
+        self._overlays = dict()
```
larsoner, line 96: We still need to use `OrderedDict` until we switch to Python3.7 minimum (it was only an official language feature in 3.7, in 3.6 it works with standard Python but it was an implementation detail)

GuillaumeFavelier:
> So it seems like there is both a blending-logic bug and a data scaling bug

I would say data scaling for now because when I click on `Restore scaling`, I obtain a more satisfying result:

image

GuillaumeFavelier:
The artifacts are available on: 
https://23965-1301584-gh.circle-artifacts.com/0/dev/index.html

I'll work on coverage now

GuillaumeFavelier:

larsoner:
I guess unsurprisingly, CircleCI renders faster with this PR, which is neat. And even locally doing `PATTERN=dspm_source make html_dev-pattern` (which makes a movie) on `master` takes 23 sec and on this PR it's 22 sec, which is great! So it's either a bit faster, or in the noise for speed.

larsoner (approved):
LGTM, will merge once CIs are happy. Awesome @GuillaumeFavelier !

agramfort:
great !

>

PR #8577, opened 2020-11-28
MRG, BUG: fix rstrip in EDF

agramfort:
I got some data with some weird \x00

I could read the file fine with eeglab but not with pyedf or pyedflib

not sure if we should merge this.


sappelhoff:
Maybe you can ask @a-hurst, who seems to know a lot about EDF :)

mne/io/edf/edf.py
```diff
@@ -621,13 +621,13 @@ def _read_edf_header(fname, exclude):
         ch_names = _unique_channel_names(ch_names)
         orig_units = dict(zip(ch_names, units))
 
-        physical_min = np.array([float(fid.read(8).decode())
+        physical_min = np.array([float(fid.read(8).decode().rstrip('\x00'))
```
larsoner, line 624: I would do a `.split('\x00')[0]` instead. Really I think this is an old-school C-string null termination, i.e., read until 8 bytes or `\x00` is encountered, and `\x00` replicates this behavior.

My C is rusty, though, so take this with a grain of salt and if someone knows better feel free to correct me...

a-hurst:
@agramfort The way it's handled for parsing the ID info and recording info parts of the header is to decode the bytes explicitly as 'latin-1' ('ascii' works too) and just use rstrip without an argument:

https://github.com/mne-tools/mne-python/blob/3276fed0d03341c8a1484a0e951eec551f784541/mne/io/edf/edf.py#L526

The EDF standard explicitly says only ASCII characters are allowed in the header, so that approach is in keeping with the spec and not just for convenience.

mne/io/edf/edf.py
larsoner: ... really I would refactor `_edf_str_int` to be `_edf_str_cast` and do a `float` cast here. It uses the `.split` method and keeps the code more DRY

larsoner:
@a-hurst unfortunately I don't think a bare `rstrip` will fix it:
```
>>> b'\x00'.decode('latin-1').rstrip()
'\x00'
```
But a `.split('\x00')[0]` should I think

a-hurst:
Ah yep you're right: for some reason I thought the headers were padded with null bytes as well but they're padded with space characters. Your solution is the way to go.

agramfort:
thanks for taking over

>

PR #8584, opened 2020-11-30
Detect/correct invalid HP/LP settings in EDF/BDF/GDF

### LINKED ISSUE #8582: Bug in mne.io.read_raw_gdf()
#### Describe the bug
Lowpass and highpass cut off frequencies are swapped when using `mne.io.read_raw_gdf()` to load a gdf file from the BCI IV 2b dataset.

#### Steps to reproduce

1. Download BCI IV 2b dataset from [here](http://www.bbci.de/competition/iv/) (it requires a name and an email)
2. Load one of the gdf files using `mne.io.read_raw_gdf()`

#### Expected results
`raw_gdf.info` should return:
```
<Info | 7 non-empty values
 bads: []
 ch_names: EEG:C3, EEG:Cz, EEG:C4, EOG:ch01, EOG:ch02, EOG:ch03
 chs: 6 EEG
 custom_ref_applied: False
 highpass: 0.5 Hz
 lowpass: 100 Hz
 meas_date: 2006-03-07 09:23:36 UTC
 nchan: 6
 projs: []
 sfreq: 250.0 Hz
>
```

Also `raw_gdf.plot()` should look like the figure below
figure1

```
<Info | 7 non-empty values
 bads: []
 ch_names: EEG:C3, EEG:Cz, EEG:C4, EOG:ch01, EOG:ch02, EOG:ch03
 chs: 6 EEG
 custom_ref_applied: False
 highpass: 100.0 Hz
 lowpass: 0.5 Hz
 meas_date: 2006-03-07 09:23:36 UTC
 nchan: 6
 projs: []
 sfreq: 250.0 Hz
>
```
> `raw_gdf.plot()`
figure2

#### Additional information
I'm using the 0.22.dev0 version of mne on Ubuntu 18.04

cbrnr:
Fixes #8582. EDF/BDF/GDF files can contain invalid LP/HP filter settings (cutoff frequencies for HP can be greater than the one for LP). This PR (1) issues a warning and (2) sets HP/LP values for affected channels to `None`.

### DIFF
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -342,12 +342,20 @@ def _read_header(fname, exclude):
     ext = os.path.splitext(fname)[1][1:].lower()
     logger.info('%s file detected' % ext.upper())
     if ext in ('bdf', 'edf'):
-        return _read_edf_header(fname, exclude)
-    elif ext in ('gdf'):
-        return _read_gdf_header(fname, exclude), None
+        edf_info, orig_units = _read_edf_header(fname, exclude)
+    elif ext == 'gdf':
+        edf_info, orig_units = _read_gdf_header(fname, exclude), None
     else:
-        raise NotImplementedError(
-            'Only GDF, EDF, and BDF files are supported, got %s.' % ext)
+        raise NotImplementedError(f'Only GDF, EDF, and BDF files are supported'
+                                  f', got {ext}.')
+    hi, lo = edf_info['highpass'], edf_info['lowpass']
+    idx = hi > lo
+    if idx.any():
+        warn(f'Highpass cutoff frequency {hi} is greater than lowpass cutoff '
+             f'frequency {lo}. Setting both values to None.')
+        edf_info['highpass'][idx] = np.nan
+        edf_info['lowpass'][idx] = np.nan
+    return edf_info, orig_units
 
 
 def _get_info(fname, stim_channel, eog, misc, exclude, preload):
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -348,13 +348,6 @@ def _read_header(fname, exclude):
     else:
         raise NotImplementedError(f'Only GDF, EDF, and BDF files are supported'
                                   f', got {ext}.')
-    hi, lo = edf_info['highpass'], edf_info['lowpass']
-    idx = hi > lo
-    if idx.any():
-        warn(f'Highpass cutoff frequency {hi} is greater than lowpass cutoff '
-             f'frequency {lo}. Setting both values to None.')
-        edf_info['highpass'][idx] = np.nan
-        edf_info['lowpass'][idx] = np.nan
     return edf_info, orig_units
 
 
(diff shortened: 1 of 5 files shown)


cbrnr:
@apospraf loading files from BCI Competition IV 2B works as expected with this PR.

cbrnr:
[NumPy installation fails in one job](https://github.com/mne-tools/mne-python/pull/8584/checks?check_run_id=1475280098)?

larsoner (commented):
Want to add a quick test via `monkeypatch` (probably of `_read_edf_header`)? If not I can see if it's easy and push if so

cbrnr:
Please go ahead and add a test, I won't have time until tomorrow.

larsoner:
I don't think the current approach quite works properly, for example it doesn't handle `'0.0'` and `'0.000'` because they are strings
```
>>> '0.0' < '0.000'
True
```
Moreover sometimes `edf_info['highpass']` can have length zero and `edf_info['lowpass']` can have length `n_channels`, which fails the NumPy `<` comparison on latest NumPy dev. It seems safer to deal with these being reversed when actually setting `info['highpass']` and `info['lowpass']`, but maybe not. I'm pushing a commit with this approach and a test that asserts it works for the `monkeypatch` code at least.

agramfort (approved):
thx @cbrnr

mne/io/edf/edf.py
```diff
@@ -342,12 +342,13 @@ def _read_header(fname, exclude):
...
-        return _read_gdf_header(fname, exclude), None
+        edf_info, orig_units = _read_edf_header(fname, exclude)
+    elif ext == 'gdf':
+        edf_info, orig_units = _read_gdf_header(fname, exclude), None
     else:
-        raise NotImplementedError(
-            'Only GDF, EDF, and BDF files are supported, got %s.' % ext)
+        raise NotImplementedError(f'Only GDF, EDF, and BDF files are supported'
+                                  f', got {ext}.')
+    return edf_info, orig_units
```
cbrnr, line 351: If you're moving the check to `_get_info` (which I think is a great idea), then you can also revert my changes here and directly `return` without creating two names.

mne/io/edf/edf.py
```diff
@@ -473,6 +474,13 @@ def _get_info(fname, stim_channel, eog, misc, exclude, preload):
     if np.isnan(info['lowpass']):
         info['lowpass'] = info['sfreq'] / 2.
 
+    if info['highpass'] > info['lowpass']:
+        warn(f'Highpass cutoff frequency {info["highpass"]} is greater than '
+             f'lowpass cutoff frequency {info["lowpass"]}. '
+             'Setting both values to None.')
+        info['highpass'] = None
+        info['lowpass'] = None
```
cbrnr, line 482: Or should we set them to `0` and `info['sfreq'] / 2` instead? I guess this amounts to the same thing (where do we equal `None` to these values?), but it would be more explicit (maybe at least in the warning message).

larsoner:
@cbrnr all good now?

cbrnr:
Yes, perfect!

PR #8594, opened 2020-12-02
MRG: Fix DICS rank handling

wmvanvliet:
DICS handing of the `rank` parameter was completely broken. It was changed from a single number to something more complicated, but the code was not updated appropriately.

Still needs unit test to cover this case, but here at least is a fix for the code.


larsoner:
I wonder if fixing this would allow us to remove the `reduce_rank=True` from here:

https://mne.tools/dev/auto_examples/inverse/plot_dics_source_power.html?highlight=dics#compute-source-power-using-dics-beamformer

larsoner:
Spoiler alert: it doesn't. I guess this makes sense since the `reduce_rank` has to do with the 3-orientations rank as opposed to the `n_ch` rank / whitening.

wmvanvliet:
The computations have not changed. It's just that setting the `rank` parameter to anything other than `None` was broken.

larsoner (approved):
I just pushed a test and `latest.inc` update, LGTM +1 for merge

agramfort (approved):

wmvanvliet:
Thanks @larsoner! I really appreciate you picking this up.

PR #8596, opened 2020-12-02
MRG, MAINT: remove / consolidate redundant examples

### LINKED ISSUE #8284: MAINT: CircleCI builds take 2h
This is just for the "build" step. Installation plus artifacts and the like make it take about 2.5 hours total.

We should look into streamlining some examples. Maybe some can also be combined and turned into tutorials (many machine learning ones?).

drammock:
Closes #8284

I will try to organize this PR with 1 removal/consolidation per commit, and use the commit message to describe which remaining example/tutorial covers the same ground.


larsoner:
Looks good so far

drammock:
@larsoner this was all the low-hanging fruit as far as I can tell.  Any further consolidation will take some more ambitious rewriting of examples/tutorials.  Do you prefer to leave this open as I chip away at it, or merge this once CIs are green and come back to the effort in subsequent PRs?

larsoner:
Multiple PRs as you work on it is better, that way the build time can go down each time you come back to it :)

larsoner:
```
/home/circleci/project/doc/auto_examples/index.rst:35: WARNING: py:obj reference target not found: tut_creating_data_structures
/home/circleci/project/doc/auto_tutorials/io/plot_10_reading_meg_data.rst:315: WARNING: undefined label: _tut_creating_data-structures (if the link has no caption the label must precede a section header)
```

drammock:
Ready for review/merge.  Here are the major changes (mostly new sections in tutorials that used to be part of an example):

- [EEG reference: REST method](https://23912-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_55_setting_eeg_reference.html#using-an-infinite-reference-rest)
- [how to use multitaper instead of Welch on Raw objects](https://23912-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/raw/plot_40_visualize_raw.html#plotting-spectral-density-of-continuous-data)
- [adding plots to the mne-objects-from-arrays tutorial](https://23912-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/simulation/plot_creating_data_structures.html)
- [the NEO example is now all by itself](https://23912-1301584-gh.circle-artifacts.com/0/dev/auto_examples/io/plot_read_neo_format.html), separate from the data-structures-from-arrays examples (which were merged into the simulation tutorial in the prev. item)

tutorials/raw/plot_40_visualize_raw.py
```diff
@@ -122,6 +124,33 @@
 
 raw.plot_psd(average=True)
 
+###############################################################################
+# The `~mne.io.Raw.plot_psd` method estimates the spectrum using the Welch
+# method, which is efficient for long data spans. MNE-Python also implements
+# multitaper spectrum estimation as part of the ``time_frequency`` submodule,
+# though this requires doing the plotting manually using matplotlib, and taking
+# care to handle the different channel types separately:
```
larsoner, line 132: In general doing `psd_multitaper` on a raw instance can be quite slow. Even for `sample` which is just ~300 sec long it takes 12 sec to compute, and yields 83,000+ frequency points, which is... a lot. Should probably at least `.. note::` or warn people here that it will take a long time.

I wonder if instead of doing this we should actually implement welch + multitaper. In principle it's really just `psd_multitaper` called on `make_fixed_length_epochs(raw)`. It's even slower than `psd_multitaper(raw)` but it should produce a nicer result. At some point I'll see what adding a `multitaper=True` argument to `plot_psd` is like...

drammock: > In general doing `psd_multitaper` on a raw instance can be quite slow.

I know.  Personally I don't think this addition to the tutorial is all that valuable, but I figured it must have been in the original example for a reason.  I can remove it, or I can add a warning and throw in a `tmax` for good measure, so it's not doing the full 60 seconds. WDYT?

> I wonder if instead of doing this we should actually implement welch + multitaper. In principle it's really just `psd_multitaper` called on `make_fixed_length_epochs(raw)`. It's even slower than `psd_multitaper(raw)` but it should produce a nicer result. At some point I'll see what adding a `multitaper=True` argument to `plot_psd` is like...

Personally I'd like to see `psd_welch` and `psd_multitaper` both return a `Spectrum` class that has its own plotting method(s). Then add a `raw.to_spectrum(method='welch'|'multitaper')` method that returns that `Spectrum` object. Then you could do `raw.to_spectrum().plot()`.  But that's a lot of work :)

larsoner: > I can remove it, or I can add a warning and throw in a tmax for good measure, so it's not doing the full 60 seconds. WDYT?

For now maybe just remove it. We should just improve `plot_psd` instead.

> > I wonder if instead of doing this we should actually implement welch + multitaper. 
> Then add a raw.to_spectrum(method='welch'|'multitaper') method

This misses the idea that you can combine welch and multitaper, though. I'll try to make a PR to show it. Then again I'm not sure if in practice people actually do combine them, it's even slower than either one alone anyway :)

drammock: > This misses the idea that you can combine welch and multitaper, though. I'll try to make a PR to show it. Then again I'm not sure if in practice people actually do combine them, it's even slower than either one alone anyway :)

yes, sorry --- I understood the idea of combining them, I was just going off on one of my "how the API could be better" tangents. :)

I'll remove the addition now.

larsoner:
Thanks @drammock

Issue #8583, opened 2020-11-30
More control over plot settings


eort:
#### Describe the new feature or enhancement
Would it be possible to give users more control over (some of the) plotting parameters? 
in my case, I wanted to plot multiple ERPs in a single figure. WIth `plot_evoked`, I could extract all the axes and gather all the plots in a single figure, however, the layout was not really nice (a lot of whitespace). I wished to constrain the layout, but couldn't do so because in `plot_evoked` `set_tight_layout` is hardcoded to `True`, which is incompatible with a constrained layout.

#### Describe your proposed implementation
If I understand the structure correctly, there are front-end auxiliary plotting functions that users interact with and the actual plotting functions that are called in the background and do the plotting. As it seems, only a subset of the arguments that the plotting functions have are being used by those auxiliary functions. Would it be possible to pass those "hidden" arguments as `kwargs`? Like that the API would not be cluttered with secondary plotting parameters, but still allow for more flexibility? 

#### Additional comments
As mentioned before, I only had a look at the `plot_evoked` function. I don't know in how far something similar is also relevant for other plotting functions.

agramfort:
you tried to do the right thing

set_tight_layout

should probably not be used systematically.

regarding kwargs I prefer to avoid them as much as possible as
they are almost impossible to document and can lead to silent bugs

can you open a PR with the changes that are necessary to address your
needs and looking at the diff we see how to move forward?

>

drammock:
@eort I'm in the process of refactoring a lot of the 2D plotting functions (see #7751). One of the things I am doing is eliminating hard-coded calls to `tight_layout` whenever possible. Please @-mention me when you open a PR, I'd be happy to chime in or help out...  `plot_evoked` was actually next on my TODO list since it shares a lot of code with the `plot_psd` functions (which I just finished last week).

eort:
Sounds very nice, thanks!

> since it shares a lot of code with the plot_psd functions (which I just finished last week).

cool, I was having similar issues with `plot_psds` (it doesn't have a title keyword).

> Please @-mention me  when you open a PR
will do

drammock:
>  I was having similar issues with `plot_psds` (it doesn't have a title keyword).

[...]

PR #8597, opened 2020-12-02
eliminate hardcoded tight layout

eort:
#### Reference issue
Candidate implementation for #8583  

#### Limitations
- I have just sticked to `plot_epochs` and didn't look for other occurrences of this behaviour. 
- I haven't added a `.. versionadded:: 0.XX ` entry

@drammock


drammock:
sorry to be slow on this.  I've been pondering, and I'm hesitating to merge this right away because we're about to release a new stable version in a couple weeks.  I'd rather not add a new `set_tight_layout` parameter to these plotting methods, and then turn around and deprecate it in the next version (because I think the right fix is to *never* call `tight_layout`).

What I'd like to do is get the more permanent fix in place before we release. I've just added #8583 to our release milestone, to motivate me to get around to it very soon.  I'd like to leave this PR open, and merge it *after* the next release if we don't get around to implementing a proper fix before release time...  that way we could introduce the new (temporary) `set_tight_layout` parameter at the beginning of a cycle and then remove it before the next release, saving ourself the pain of having to deprecate it for six months before removing it.

PR #8603, opened 2020-12-04
VIZ, MAINT: Use fig class for evoked

drammock:
This adds some class hierarchy documentation to the docstring of `mne/viz/_figure.py`, and refactors the PSD figure class (now called MNELineFigure) so it can more easily be re-used for other line-based figures like `evoked.plot()`.  Then the following viz methods will be switched over to use the new class:

- evoked.plot()
- evoked.plot_white()
- dipole.plot()
- evoked.plot_joint()  (maybe, this one might be trickier)


larsoner (approved):

larsoner:
Maybe just merge this first, then change the others?

drammock:
sure, if you want.

PR #8605, opened 2020-12-04
MRG, BUG: find_ecg_events() didn't respect tstart when calculating average heart rate

hoechenberger:
MWE:
```python
import os
import mne
mne.set_log_level('error')

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file)

_, _, avg_hr = mne.preprocessing.find_ecg_events(raw=raw)
print(avg_hr)

_, _, avg_hr = mne.preprocessing.find_ecg_events(raw=raw, tstart=120)
print(avg_hr)
```

This PR:
```
61.78988748454362
63.91261368728932
```

`master`:
```
61.78988748454361
36.296157683228415
```

### DIFF
--- a/mne/preprocessing/tests/test_ecg.py
+++ b/mne/preprocessing/tests/test_ecg.py
@@ -79,11 +79,11 @@ def test_find_ecg():
     assert 'ECG-SYN' not in ecg_epochs.ch_names
 
 
-def test_find_ecg_events_tstart():
+@pytest.mark.parametrize('tstart', (0, 12))
+def test_find_ecg_events_tstart(tstart):
     """Ensure tstart is taken into account when calculating avg heart rate."""
     raw = read_raw_fif(raw_fname, preload=False)
     event_id = 999
-    tstart = raw.times[-1] / 2
     events, _, average_hr = find_ecg_events(raw=raw, event_id=event_id,
                                             tstart=tstart)
 
--- a/mne/preprocessing/tests/test_ecg.py
+++ b/mne/preprocessing/tests/test_ecg.py
@@ -1,5 +1,5 @@
 import os.path as op
-
+from numpy.testing import assert_allclose
 import pytest
 from mne.io import read_raw_fif
 from mne import pick_types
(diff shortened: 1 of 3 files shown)


mne/preprocessing/tests/test_ecg.py
```diff
@@ -79,4 +79,18 @@ def test_find_ecg():
...
+    raw = read_raw_fif(raw_fname, preload=False)
+    event_id = 999
+    tstart = raw.times[-1] / 2
+    events, _, average_hr = find_ecg_events(raw=raw, event_id=event_id,
+                                            tstart=tstart)
+
+    duration_in_sec = len(raw) / raw.info['sfreq'] - tstart
+    duration_in_min = duration_in_sec / 60
+    average_hr_expected = len(events) / duration_in_min
+    assert_allclose(average_hr, average_hr_expected)
```
agramfort, line 93: can you rather update an existing test or maybe use pytest factorize with tstart value?

hoechenberger: Didn't want to touch the (one) existing test because it seems to be mostly concerned with testing proper function on not-preloaded data (and then some bits I don't understand at all); I've now added pytest parametrization. Test function takes about 0.75 sec on my machine now

agramfort: I suggested this to reduce computation time of tests or test this all at the same place.

it turns out it's tested here 

https://github.com/mne-tools/mne-python/blob/971f631487975fa0108c86d8ce9203afc2def6b6/mne/preprocessing/tests/test_ecg.py#L41

can we factorize this to try? objective: keep testing time low and centralize tests in one place. thanks !

hoechenberger: Ok I have addressed this in 7477f68
Time for running the test function has not changed for me compared to `master`! (ignoring between-run variability)

hoechenberger: Slightly improved test coverage via ee58ebb, we're now ~150 ms slower than on `master`. I think this is okay?

agramfort:
thx @hoechenberger

PR #8606, opened 2020-12-04
ENH: Add preprocessing.annotate_ecg

hoechenberger:
We already have:
- `annotate_flat`
- `annotate_movement`
- `annotate_muscle_zscore`

This PR adds `annotate_ecg`.

Also makes it such that `find_ecg_events` now gracefully handles cases where no ECG events could be found.


hoechenberger:
API doc: https://23954-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.annotate_ecg.html#mne.preprocessing.annotate_ecg

Updated tutorial: https://23954-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_10_preprocessing_overview.html#heartbeat-artifacts-ecg

drammock (commented):
Nice addition!  I didn't scrutinize the code changes in detail yet.  My main feedback is that I wonder whether the doc changes are in the right place / have the right focus or scope.  I don't believe we need to demo *every* possible capability in these tutorials, and in my mind this tutorial is about "finding and visualizing common artifacts".  So showing how to detect heartbeats *and end up with annotations of them* is absolutely relevant, but I'm not sure we need 3 figures all showing various annotation capabilities. I would suggest limiting to 1 annotation plot, maybe move it to the end of the ECG section, and frame it like this: "if you want to do something more custom or specialized with the heartbeats than the create_ecg_epochs function allows, you can use the annotate_ecg function to just find the R-peak (or to get approximate intervals around the PQRST complex) and then use those annotations to do something more specialized (e.g., XXX)."  where you get bonus points if you can replace XXX with an example of why someone might need to do this.

Also, did  you check that you actually *need* to create a copy of raw here?  There aren't any other `raw.plot()` calls in the tutorial; having the annotations in there might not affect any of the subsequent plots / computations.

mne/preprocessing/ecg.py
```diff
@@ -219,6 +210,14 @@ def find_ecg_events(raw, event_id=999, ch_name=None, tstart=0.0,
                               thresh_value=qrs_threshold, l_freq=None,
                               h_freq=None, verbose=False)
 
+    if ecg_events.size == 0:
+        logger.info('Could not detect any ECG events.')
+        out = (ecg_events, idx_ecg, np.nan)
+        if return_ecg:
+            ecg = ecg[np.newaxis]
+            out += (ecg,)
+        return out
```
larsoner, line 219: Why does the code below not work nicely already? If we can make it support the 0-events case easily / with one or two conditionals it's better than short-circuiting, as it simplifies the logic and flow

larsoner: Pushed a commit to revert this

hoechenberger:
@drammock @larsoner 

I discussed the content of this PR with @agramfort and I'd like to put it on hold for now until we come up with a more general idea of a framework that will actually allow users to **make use** of these annotations. My proposal being that `create_[ecg,eog]_epochs()` should allow users to pass annotated raw data, and make provision of an `events` array optional.

Thinking more broadly, maybe even `mne.Epochs` should gain Annotations support.

But this needs to discussed, and I'll open an issue shortly.

PR #8610, opened 2020-12-04
MRG, ENH: Add DICS bias tests

larsoner:
@wmvanvliet @britta-wstnr I am adding DICS bias tests along the lines of what we have for LCMV and minimum norm. After this PR is all good, we can add orientation tests to check the `max-power-ori` calculations.

However, it looks like `noise_csd`-based whitening is broken for DICS. I added a simple `_cov_as_csd` that just packs our data and noise cov into CSD objects, and the results should in principle match what we get for LCMV. I copied the code and limits from the LCMV tests. DICS performs within the same limits as LCMV as long as no whitening is used. Once things are whitened, things break. You can see that only the `noise_cov=False` versions of the bias tests are uncommented, except one -- the uncommented one went from the peak localization limit around 40 (meaning 40% of vertices were correctly localized exactly to the correct vertex) to being ~1%.

Two things I fixed so far:

1. Bad channel handling -- if there are bads in `info` but CSDs have the bad channels, there is a picking problem.
2. I added the same `_restore_pos_semidef` "trick" that we use for LCMV. Basically if EEG channels are used, the massive scale differences become a problem and the "trick" restores `Cm` to being positive semidefinite. (I'm not 100% sure it's correct but it seems to at least be okay for LCMV.) However, even without this hack workaround / fix, the localization limits have to be ~1%, so there is still something broken here.

### DIFF
--- a/mne/beamformer/_dics.py
+++ b/mne/beamformer/_dics.py
@@ -17,7 +17,7 @@ from ..forward import _subject_from_forward
 from ..minimum_norm.inverse import combine_xyz, _check_reference, _check_depth
 from ..source_estimate import _make_stc, _get_src_type
 from ..time_frequency import csd_fourier, csd_multitaper, csd_morlet
-from ._compute_beamformer import (_check_proj_match, _prepare_beamformer_input,
+from ._compute_beamformer import (_prepare_beamformer_input,
                                   _compute_beamformer, _check_src_type,
                                   Beamformer, _compute_power,
                                   _restore_pos_semidef, _proj_whiten_data)
@@ -193,11 +193,11 @@ def make_dics(info, forward, csd, reg=0.05, noise_csd=None, label=None,
                         (freq, i + 1, n_freqs))
 
         Cm = csd.get_data(index=i)
+
         # XXX: Weird that real_filter happens *before* whitening, which could
         # make things complex again...?
         if real_filter:
             Cm = Cm.real
-
         # Whiten the CSD
         Cm = np.dot(whitener, np.dot(Cm, whitener.conj().T))
         _restore_pos_semidef(Cm)
--- a/mne/beamformer/_dics.py
+++ b/mne/beamformer/_dics.py
@@ -9,6 +9,7 @@
 import numpy as np
 
 from ..channels import equalize_channels
+from ..io.pick import pick_info, pick_channels
 from ..utils import (logger, verbose, warn, _check_one_ch_type,
                      _check_channels_spatial_filter, _check_rank,
                      _check_option, _validate_type)
@@ -18,7 +19,8 @@ from ..source_estimate import _make_stc, _get_src_type
 from ..time_frequency import csd_fourier, csd_multitaper, csd_morlet
 from ._compute_beamformer import (_check_proj_match, _prepare_beamformer_input,
                                   _compute_beamformer, _check_src_type,
-                                  Beamformer, _compute_power)
+                                  Beamformer, _compute_power,
+                                  _restore_pos_semidef)
 
 
 @verbose
@@ -165,7 +167,9 @@ def make_dics(info, forward, csd, reg=0.05, noise_csd=None, label=None,
     n_freqs = len(frequencies)
 
     _check_one_ch_type('dics', info, forward, csd, noise_csd)
-    info, fwd, csd = equalize_channels([info, forward, csd])
+    # remove bads so that equalize_channels only keeps all good
+    info = pick_info(info, pick_channels(info['ch_names'], [], info['bads']))
+    info, forward, csd = equalize_channels([info, forward, csd])
 
     csd, noise_csd = _prepare_noise_csd(csd, noise_csd, real_filter)
 
@@ -189,11 +193,14 @@ def make_dics(info, forward, csd, reg=0.05, noise_csd=None, label=None,
                         (freq, i + 1, n_freqs))
 
         Cm = csd.get_data(index=i)
+        # XXX: Weird that real_filter happens *before* whitening, which could
+        # make things complex again...?
         if real_filter:
             Cm = Cm.real
 
         # Whiten the CSD
         Cm = np.dot(whitener, np.dot(Cm, whitener.conj().T))
+        _restore_pos_semidef(Cm)
 
         # compute spatial filter
         n_orient = 3 if is_free_ori else 1
--- a/mne/beamformer/_dics.py
+++ b/mne/beamformer/_dics.py
@@ -20,7 +20,7 @@ from ..time_frequency import csd_fourier, csd_multitaper, csd_morlet
 from ._compute_beamformer import (_check_proj_match, _prepare_beamformer_input,
                                   _compute_beamformer, _check_src_type,
                                   Beamformer, _compute_power,
-                                  _restore_pos_semidef)
+                                  _restore_pos_semidef, _proj_whiten_data)
 
 
 @verbose
@@ -263,9 +263,7 @@ def _apply_dics(data, filters, info, tmin):
             logger.info("Processing epoch : %d" % (i + 1))
 
         # Apply SSPs
-        if info['projs']:
-            _check_proj_match(info['projs'], filters)
-            M = np.dot(filters['proj'], M)
+        M = _proj_whiten_data(M, info['projs'], filters)
 
         stcs = []
         for W in Ws:
--- a/mne/beamformer/_dics.py
+++ b/mne/beamformer/_dics.py
@@ -20,7 +20,7 @@ from ..time_frequency import csd_fourier, csd_multitaper, csd_morlet
 from ._compute_beamformer import (_prepare_beamformer_input,
                                   _compute_beamformer, _check_src_type,
                                   Beamformer, _compute_power,
-                                  _restore_pos_semidef, _proj_whiten_data)
+                                  _proj_whiten_data)
 
 
 @verbose
@@ -198,15 +198,13 @@ def make_dics(info, forward, csd, reg=0.05, noise_csd=None, label=None,
         # make things complex again...?
         if real_filter:
             Cm = Cm.real
-        # Whiten the CSD
-        Cm = np.dot(whitener, np.dot(Cm, whitener.conj().T))
-        _restore_pos_semidef(Cm)
 
         # compute spatial filter
         n_orient = 3 if is_free_ori else 1
         W, max_power_ori = _compute_beamformer(
             G, Cm, reg, n_orient, weight_norm, pick_ori, reduce_rank,
-            rank=rank, inversion=inversion, nn=nn, orient_std=orient_std)
+            rank=rank, inversion=inversion, nn=nn, orient_std=orient_std,
+            whitener=whitener)
         Ws.append(W)
         max_oris.append(max_power_ori)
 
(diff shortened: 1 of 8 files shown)


larsoner:
Okay turns out the problem was that the whitener was not applied to the data. I still think there is something fishy with the ordering where `real_filter` happens before the `whitener` is applied, but the whitener can be complex (?). But we should fix that in another PR. @britta-wstnr @wmvanvliet can you review? We should get this in for 0.22

mne/beamformer/_dics.py
```diff
@@ -189,11 +193,14 @@ def make_dics(info, forward, csd, reg=0.05, noise_csd=None, label=None,
                         (freq, i + 1, n_freqs))
 
         Cm = csd.get_data(index=i)
+
+        # XXX: Weird that real_filter happens *before* whitening, which could
+        # make things complex again...?
```
britta-wstnr, line 198: Agree, @wmvanvliet can you double check?

mne/beamformer/tests/test_dics.py
```diff
@@ -669,4 +674,52 @@ def test_tf_dics(_load_forward, idx):
...
+    data = cov['data'] + 1e-1 * _rand_csd(rng, info)
+    assert data.dtype == np.complex128
+    return CrossSpectralDensity(_sym_mat_to_vector(data), cov['names'], 0., 16)
+
+
+# Just test free ori here (assume fixed is same as LCMV if these are)
+@pytest.mark.parametrize(
+    'reg, pick_ori, weight_norm, use_cov, depth, lower, upper, real_filter', [
+        (0.05, None, 'unit-noise-gain-invariant', False, None, 26, 28, False),
+        (0.05, None, 'unit-noise-gain-invariant', True, None, 40, 42, False),
```
britta-wstnr, line 691: How did you pick `upper` and `lower` percentages? Copied from LCMV?

britta-wstnr (commented):
Mostly LGTM, @larsoner.
Just a question re the boundaries for the tests.

mne/beamformer/tests/test_dics.py
larsoner: Yep, all values here match LCMV. Really it should stay that way, so once we settle on whether or not to change when `real_filter` happens I'll add a comment mentioning that they should stay in sync. Or maybe I'll just define different subset tuples in LCMV and use them here to make sure they match both places... I'll see if that stays readable.

britta-wstnr: Yes, agreed, they should match. How did we come up for them for the LCMV again?

larsoner: They are just empirical -- they reflect the values that we produce and have produced for a while. The idea is that we should only have to change them when we find a bug -- and hopefully the values generally go up when we do this. And similarly, they help us see when we've broken something in our code because they are reasonably sensitivity to breaking changes.

britta-wstnr: Okay, that's what I thought. So they would presume the implementation to be correct ... :-) Well I guess for a lot we have checks that things actually _are_ correct as well. 
Speaking of which, once we decided upon the `real` orientation selection issue, I can push the external tests (against FieldTrip) for the DICS.

larsoner:
@britta-wstnr I added a comment, and while I was in there I added some orientation tests.

larsoner:
@britta-wstnr I also realized the problem with the no-reg solutions -- the `bias` test covariances being used were actually quite rank deficient because they were calculated using the product of the forward with its transpose for < 250 source locations but 366 channels. I added a tiny bit of reg directly to the covariance, then project with the projectors, which is probably much more realistic. Now those tests are *much* more stable (the ranges are only one or two percent instead of ~10)!

This also showed me that we could get rid of the positive-semidefinite "restoring" hack -- that was only needed for the bias tests because they were horribly rank deficient.

mne/beamformer/tests/test_lcmv.py
```diff
@@ -666,46 +667,55 @@ def test_localization_bias_fixed(bias_params_fixed, reg, weight_norm, use_cov,
...
-        (0.05, 'vector', 'nai', True, None, 35, 37),
-        (0.05, 'vector', None, True, None, 12, 14),
-        (0.05, 'vector', None, True, 0.8, 39, 43),
-        (0.05, 'max-power', 'unit-noise-gain-invariant', False, None, 17, 20),
-        (0.05, 'max-power', 'unit-noise-gain', False, None, 17, 20),
-        (0.05, 'max-power', 'nai', True, None, 21, 24),
-        (0.05, 'max-power', None, True, None, 7, 10),
-        (0.05, 'max-power', None, True, 0.8, 15, 18),
-        (0.05, None, None, True, 0.8, 40, 42),
+    'reg, pick_ori, weight_norm, use_cov, depth, lower, upper, lo, uo', [
```
britta-wstnr, line 673: Should we give `lo` and `uo` a more descriptive name, maybe?

mne/beamformer/tests/test_lcmv.py
```diff
@@ -666,46 +667,55 @@ def test_localization_bias_fixed(bias_params_fixed, reg, weight_norm, use_cov,
...
+                        noise_cov, pick_ori=pick_ori,
+                        weight_norm=weight_norm,
+                        depth=depth)
+    loc = apply_lcmv(evoked, filters).data
+    if pick_ori == 'vector':
+        ori = loc.copy() / np.linalg.norm(loc, axis=1, keepdims=True)
+    else:
+        # doesn't make sense for pooled (None) or max-power (can't be all 3)
+        ori = None
     loc = np.linalg.norm(loc, axis=1) if pick_ori == 'vector' else np.abs(loc)
```
britta-wstnr, line 715: Wait, now I am confused. Didn't we set out to test whether the chosen orientation matches the simulated orientation? Does this just test the orientation the leadfield is giving?

britta-wstnr:
Ah, great that this is resolved and we don't need weirdly high regularization parameters anymore, @larsoner!

mne/beamformer/tests/test_lcmv.py
larsoner: The "simulated data" here is just the leadfield, and the (free ori) leadfield is oriented as X0, Y0, Z0, X1, Y1, Z1, ... , XN, YN, ZN for the N+1 sources, so the correct orientation of the simulated data is just `np.eye(3)` repeating over and over again

britta-wstnr: But we are not testing the actual selected orientation in the scalar beamformer cases, or are we?

larsoner: The `max_power_ori` can't possibly work in this situation -- it selects a single orientation for each location, and we are simulating activity in each of the three cardinal directions. We probably should have a separate test for example that if we simulate data in the normal direction, the max_power_ori beamformer recovers the correct orientations. I'll see if it's easy to add, should be...

larsoner: (can't possibly work to get the orientations right anyway)

britta-wstnr: Okay, I see. If we want to check for `real` in the orientation selection for the DICS beamformers, then we would indeed need a test with a single orientation.

mne/beamformer/tests/test_lcmv.py
```diff
@@ -666,46 +670,111 @@ def test_localization_bias_fixed(bias_params_fixed, reg, weight_norm, use_cov,
...
+# orientation values
+@pytest.mark.parametrize(
+    'reg, weight_norm, use_cov, depth, lower, upper, lower_ori, upper_ori', [
+        (0.05, 'unit-noise-gain-invariant', False, None, 38, 40, 0.52, 0.54),
+        (0.05, 'unit-noise-gain', False, None, 38, 40, 0.52, 0.54),
+        (0.05, 'nai', True, None, 56, 57, 0.56, 0.58),
+        (0.05, None, True, None, 27, 28, 0.54, 0.56),
+        (0.05, None, True, 0.8, 42, 43, 0.54, 0.56),
+        # no reg
+        (0.00, None, True, None, 50, 51, 0.57, 0.58),
```
larsoner, line 737: @britta-wstnr I implemented the test, but it doesn't do too well. The mean absolute dot product for random unit vectors converges to 0.5, a perfect match will be 1.0, and these hover between 0.53-0.57. Maybe it's because there are too many sources active? Though the non-ori-selection / full vector solutions don't seem to have this problem.

It's also possible that the covariance matrix I'm passing is not the best, even though it's the product of the simulated (normal-orientation) source lead fields with itself transposed. But this will at least help us tell if we break something with our orientation selection.

larsoner:
@britta-wstnr I say we fix the `real_filter` in another PR. It can be backported easily to 0.22 if we need to. Are you happy with this one otherwise?

larsoner:
@britta-wstnr okay with this one? Would be good to get this in since it has some useful fixes.

mne/beamformer/tests/test_lcmv.py
britta-wstnr: Mh I am puzzled about this, but I am also not sure I fully understood how you simulated the data. But fine to just let it in passing for now, and we can investigate later.

mne/beamformer/_dics.py
britta-wstnr: @larsoner what do we do with this? Keep as is and investigate what it would change in new PR?

mne/beamformer/tests/test_lcmv.py
```diff
@@ -666,46 +670,111 @@ def test_localization_bias_fixed(bias_params_fixed, reg, weight_norm, use_cov,
...
+        (0.00, 'unit-noise-gain-invariant', True, None, 73, 75, 0.57, 0.58),
+        (0.00, 'unit-noise-gain', True, None, 73, 75, 0.57, 0.58),
+        (0.00, 'nai', True, None, 73, 75, 0.57, 0.58),
+    ])
+def test_orientation_max_power(bias_params_fixed, bias_params_free,
+                               reg, weight_norm, use_cov, depth, lower, upper,
+                               lower_ori, upper_ori):
+    """Test orientation selection for bias for max-power LCMV."""
+    # we simulate data for the fixed orientation foward and beamform using
```
britta-wstnr, line 746: I have a neat `GitHub Actions` box saying `foward ==> forward`. Do you see those as well @larsoner or is that a new feature that is still being rolled out?

mne/beamformer/tests/test_lcmv.py
```diff
@@ -666,46 +670,111 @@ def test_localization_bias_fixed(bias_params_fixed, reg, weight_norm, use_cov,
...
+                            weight_norm=weight_norm,
+                            depth=depth)
+    loc = apply_lcmv(evoked, filters).data
+    ori = filters['max_power_ori']
+    loc = np.abs(loc)
+    # Compute the percentage of sources for which there is no loc bias:
+    max_idx = np.argmax(loc, axis=0)
+    perc = (want == max_idx).mean() * 100
     assert lower <= perc <= upper
+    # Comute the dot products of our forward normals and
```
britta-wstnr, line 765: Again, `Github actions` flagging `Comute ==> Compute`

britta-wstnr (commented):
Just spelling errors, otherwise looks good to go!

larsoner:
> Keep as is and investigate what it would change in new PR?

Yeah that's what I'm thinking. If it's wrong, it's wrong on `master` and this PR at least does not make it worse.

Issue #8609, opened 2020-12-04
CI, MAINT: circle failing again


drammock:
Circle builds are failing on the MxNE example with this:

```
Traceback (most recent call last):
  File "/home/circleci/project/examples/inverse/plot_time_frequency_mixed_norm_inverse.py", line 126, in <module>
    brain = stc.plot('sample', 'inflated', 'rh', views='medial',
  File "/home/circleci/project/mne/source_estimate.py", line 653, in plot
    brain = plot_source_estimates(
  File "<decorator-gen-139>", line 24, in plot_source_estimates
  File "/home/circleci/project/mne/viz/_3d.py", line 1794, in plot_source_estimates
    return _plot_stc(
  File "/home/circleci/project/mne/viz/_3d.py", line 1984, in _plot_stc
    brain.setup_time_viewer(time_viewer=time_viewer,
  File "/home/circleci/project/mne/viz/_brain/_brain.py", line 416, in setup_time_viewer
    self._configure_sliders()
  File "/home/circleci/project/mne/viz/_brain/_brain.py", line 719, in _configure_sliders
    value=self._data['smoothing_steps'],
KeyError: 'smoothing_steps'
```

(see here: https://app.circleci.com/pipelines/github/mne-tools/mne-python/5139/workflows/937701e3-a227-414d-a747-7c65e415acbe/jobs/23925)

@GuillaumeFavelier knows what went wrong perhaps?

GuillaumeFavelier:
I'm on it!

GuillaumeFavelier:
It seems caused by https://github.com/mne-tools/mne-python/blob/5a62f4719f28733687b5dd1f8fb9674f9f641c2f/mne/viz/_3d.py#L1911-L1912

PR #8611, opened 2020-12-04
FIX: vertices

GuillaumeFavelier:
This PR fixes hemi logic in stc.

Closes https://github.com/mne-tools/mne-python/issues/8609

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1904,7 +1904,7 @@ def _plot_stc(stc, subject, surface, hemi, colormap, time_label,
         "volume_options": volume_options,
         "verbose": False,
     }
-    for hi, hemi in enumerate(hemis):
+    for hemi in hemis:
         if isinstance(stc, _BaseVolSourceEstimate):  # no surf data
             break
         vertices = stc.vertices[0 if hemi == 'lh' else 1]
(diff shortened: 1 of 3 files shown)


larsoner:
Can you add a unit test that would have caught this? Having to hear about these sorts of problems from CircleCI is not as nice as local `pytest mne/viz`

mne/viz/_3d.py
```diff
@@ -1907,8 +1907,8 @@ def _plot_stc(stc, subject, surface, hemi, colormap, time_label,
     for hi, hemi in enumerate(hemis):
```
drammock, line 1907: I think with this change you no longer need `hi` or `enumerate`

hoechenberger (approved):

hoechenberger:
@GuillaumeFavelier We got a segfault on Py3.8, not sure if related?

GuillaumeFavelier:
I'll investigate because it's close to the test I modified.

larsoner:
Restarted -- I think I've seen a spurious macOS test failure like this before. Probably something having to do with closing windows / garbage collection / Qt / VTK

agramfort (approved):

larsoner:
Thanks @GuillaumeFavelier ! Hopefully CircleCI goes back to green now

PR #8621, opened 2020-12-07
VIZ, MAINT: remove set_tight_layout from evoked plotting code

### LINKED ISSUE #8583: More control over plot settings

drammock:
closes #8583 
closes #8597 (supersedes)

removes `set_tight_layout` from our internal evoked plotting code. This was a problem for users who were providing their own axes to `evoked.plot()`. Also tweaks the figure margins that we set when axes are not provided, to approximate what tight_layout would have yielded.

This is a short-term fix; soon evoked plots will use our new figure classes anyway.

cc @eort

### DIFF
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -192,7 +192,7 @@ def _plot_legend(pos, colors, axis, bads, outlines, loc, size=30):
 def _plot_evoked(evoked, picks, exclude, unit, show, ylim, proj, xlim, hline,
                  units, scalings, titles, axes, plot_type, cmap=None,
                  gfp=False, window_title=None, spatial_colors=False,
-                 set_tight_layout=True, selectable=True, zorder='unsorted',
+                 selectable=True, zorder='unsorted',
                  noise_cov=None, colorbar=True, mask=None, mask_style=None,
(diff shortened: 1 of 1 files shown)


agramfort (approved):

eort:
Looks good!

larsoner:
Thanks @drammock and @eort !

Issue #7591, opened 2020-04-09
BUG, MAINT: Remaining dig issues


larsoner:
In the long term we should:

- [x] Remove `_read_dig_points`, it should make use of standardized functions (#8622)
- [ ] Make `read_dig_polhemus_isotrak` actually read channel names where possible (they are embedded in the file at least sometimes)
- [ ] Make `.elp` reading by `read_custom_montage` triage automatically between isotrak and BESA (currently just assumes BESA)

Here is a sample for (3), renamed `.log` instead of `.elp` to make GitHub accept it:

[channel_locations.log](https://github.com/mne-tools/mne-python/files/4456505/channel_locations.log)

PR #8622, opened 2020-12-07
MRG, MAINT: Remove _read_dig_points

larsoner:
- Remove `mne.io._digitization._read_dig_points` and `_set_dig_kit` in favor of `mne.io.kit.coreg._read_dig_kit` and `_set_dig_kit`. The first function just reads files based on names that typically (I'm assuming) come with KIT machines to work the same way as it did before with KIT coreg, etc. The second does transforms necessary to get in the correct coondinate system. This is similar to how things work with CTF, and makes `_digitization` cleaner. The `_kit` functions now only show up in `io/kit` and `gui/*kit*`.
- Add `on_header_missing` to `read_polhemus_fastscan` because some of our testing files and such do not produce this header, and it's possible there are files in the wild that the KIT people are using that have this feature.

Part of #7591. @christianbrodbeck can you look since this messes with KIT?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -77,6 +77,8 @@ Enhancements
 
 - Add option to disable TQDM entirely with ``MNE_TQDM='off'`` (:gh:`8515` by `Eric Larson`_)
 
+- Add option ``on_header_missing`` to :func:`mne.channels.read_polhemus_fastscan` (:gh:`8622` by `Eric Larson`_)
+
 - `mne.preprocessing.ICA.plot_sources` now displays an `mne.preprocessing.ICA.plot_properties` window when right-clicking on component names on the y-axis (:gh:`8381` by `Daniel McCloy`_)
 
 - :func:`mne.io.read_raw_edf`, :func:`mne.io.read_raw_bdf`, and :func:`mne.io.read_raw_gdf` now detect and handle invalid highpass/lowpass filter settings (:gh:`8584` by `Clemens Brunner`_)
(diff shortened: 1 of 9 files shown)


agramfort:
this goes a bit backwards with our earlier attempt with @massich to centralize the logic on dig 

why moving things to kit rather and consolidation dig specific file?

larsoner:
> why moving things to kit rather and consolidation dig specific file?

The idea is that:

1. *standard* handling of digitization data or things that are useful to many readers (file formats, converting from unknown to head, etc.) should live in `mne/channels`, and that
2. *raw-reader-specific mangling* can be handled in their own modules.

It's a better split between what is standard dig information and what is specific to a format. For example, before this PR I had no idea why `_read_dig_points` even needed to exist. In this PR the functionality is clearly separated (and code commented) to make clear that the `elp`/`mrk` etc that get passed to `read_raw_kit` assume for example that

1. Any `.txt` should be read like a Polhemus FastScan
2. Any `.elp` and `.hsp` should be read like a Polhemus Isotrak (and not BESA elp)
3. The outputs should be ordered nasion/lpa/rpa instead of lpa/nasion/rpa (Neuromag order) to work with KIT coreg functions.

To me this is analogous to how we handle CTF data and coordinate frames: https://github.com/mne-tools/mne-python/blob/master/mne/io/ctf/trans.py#L43-L114

agramfort (approved):

massich:
Our initial idea was to move it to a centralized place, and we used `kit` to get hold of the whole thing. Once kit was there we could factor out lots of things that were common, and some of them we could not. If I recall properly, this was what got left from the original `kit`. It never made it back to `kit` 'cos I always believed that I could manage to factor the whole thing out and bring back something smaller. And apply a fancy programming pattern to rule them all.

PR #8624, opened 2020-12-07
VIZ, ENH: allow show/hide annotations by label

### LINKED ISSUE #8578: Toggle Annotations visibility in plot_raw()
If working with heavily-annotated data, it can become very difficult to tell overlapping Annotations apart, and their descriptions may overlap too – see e.g. this example based on the `sample` data:

I'm now intending to add EOG, ECG, and artifact Annotations to this dataset, but I'm **already** struggling to keep track of what's going on 🧐 Of course I could shrink the displayed time span, but this may not always be a feasible approach.

I wondered if we could offer an option to toggle visibility for each annotation individually? When in Annotations mode, we already display a window with radio buttons to select which Annotation(s) to add / modify. Could we maybe simply add checkboxes – probably left of the radio buttons – to toggle visibility?

cc @drammock

drammock:
closes #8578

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -31,8 +31,6 @@ Current (0.22.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add :func:`mne.source_space.compute_distance_to_sensors` to compute distances between vertices and sensors (:gh:`8534` by `Olaf Hauk`_ and `Marijn van Vliet`_)
-
 - Add :func:`mne.read_evokeds_mff` to read averaged MFFs (requires mffpy >= 0.5.7) **by new contributor** |Evan Hathaway|_ (:gh:`8354`)
 
 - Add :class:`mne.decoding.SSD` for spatial filtering with spatio-spectral-decomposition (:gh:`7070` **by new contributor** |Victoria Peterson|_ and `Denis Engemann`_)
@@ -84,6 +82,10 @@ Enhancements
 - If a ``baseline`` tuple containing one or two ``None`` values – e.g. ``(None, 0)``, ``(0, None)``, or ``(None, None)`` – is passed to `~mne.Epochs` or `~mne.Epochs.apply_baseline`, the ``None`` value(s) will be replaced with the actual time (i.e., :attr:`~mne.Epochs.tmin` and :attr:`~mne.Epochs.tmax`, respectively) when populating ``Epochs.baseline`` (:gh:`8442` by `Richard Höchenberger`_)
 
 - `~mne.Epochs` will now retain the information about an applied baseline correction, even if the baseline period is partially or completely removed through cropping later on (:gh:`8442` by `Richard Höchenberger`_)
+
+- Add :func:`mne.source_space.compute_distance_to_sensors` to compute distances between vertices and sensors (:gh:`8534` by `Olaf Hauk`_ and `Marijn van Vliet`_)
+
+- Annotations can now be shown/hidden interactively in raw plots (:gh:`8624` by `Daniel McCloy`_)
 
 Bugs
 ~~~~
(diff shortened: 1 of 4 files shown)


drammock:
demo

drammock:
@larsoner @hoechenberger @cbrnr One question I had when implementing this is: what should happen if someone toggles an annotation label to be invisible, but that label is still "selected", and then they click-drag on the plot to create a new annotation?  Currently what happens is the span selector shows during dragging, but the new annotation disappears as soon as the mouse button is released.  I don't think that's ideal, but the only alternative I could think of was for all annotations with that label to become visible when the mouse button is either pressed or released, either of which seem kinda weird from a UX point of view.  Thoughts?

cbrnr:
I'd deselect it before hiding. It doesn't even have to be reselected when showing the type again.

agramfort:
as soon as I had an annot of type A then all annots of this type become
visible.

that's how I would do it

>

hoechenberger:
Using the `MacOSX` backend here. Running into two issues:
- in fullscreen mode (tiling: Raw plot left, Annotations options right), something goes wrong when adding a new Annotation label. This problem already existed in `master`
- something it off with the checkboxes. This also happens outside of fullscreen view.

ezgif com-video-to-gif

drammock:
@hoechenberger I don't know the details of how macOS's tiling works, but I wouldn't necessarily expect it to work smoothly, given that the figure is re-created each time a label is added.  Anything you can tell me about it that might help figure out what is going wrong?  The checkbox duplication and sizing is also weird; any ideas there?

Regarding how to handle adding an annotation when its label's visibility is turned off: after some reflection I think I prefer @agramfort's suggested fix (automatically switch visibility to true for that label) rather than @cbrnr's suggestion (automatically de-activate a label when it's made invisible).  The latter proposal forces us to handle a new situation: no annotations visible means no labels are selected (currently there is always one label selected), so we'd need to special-case what the SpanSelector should (not) do when no label is selected.  Certainly possible, but I think "auto-make-visible" a label is easier to implement, and about equal in terms of user expectations of what should happen in that corner case.  Ayone else want to weigh in?

cbrnr:
I see the problem and I'm fine with @agramfort's suggestion. However, this means that at least one annotation type is always visible. I think it would be useful to hide all annotations, because sometimes annotations get in the way of the actual time series (especially in a data set with many different annotations). I don't know the best solution, but having a separate show/hide annotations would be one option (which is relatively independent from this PR).

agramfort:
>
> I see the problem and I'm fine with @agramfort
> <https://github.com/agramfort>'s suggestion. However, this means that at
> least one annotation type is always visible.
>
why? to me you make them all invisible. It's only when you add one that all
annot of this type become visible

drammock:
> It's only when you add one that all annot of this type become visible

Yes, this is what I had in mind. I'll move forward with that approach.

larsoner:
@hoechenberger do you want to test it?

larsoner (approved):
Code changes LGTM but it would be good if a heavy annotation user could test it out

hoechenberger:
@larsoner Yes I will have a look shortly.

hoechenberger (approved):
Works great for me!

hoechenberger:
> > It's only when you add one that all annot of this type become visible
> 
> Yes, this is what I had in mind. I'll move forward with that approach.

This isn't working for me yet, not sure if it was supposed to :)

drammock:
> > > It's only when you add one that all annot of this type become visible
> > 
> > 
> > Yes, this is what I had in mind. I'll move forward with that approach.
> 
> This isn't working for me yet, not sure if it was supposed to :)

working locally but I haven't pushed yet. Decided to see if I could also get `useblit=True` to work on the spanselector.

drammock:
ok, I didn't manage to get `useblit=True` working for the span selector, but can do that in a separate PR. But I did get the checkboxes to stay vertically aligned with their corresponding buttons, and I got them to stop migrating rightward with each new label added :)  They still change size a little bit, but I minimized that as much as I could.  Ready for one final look @hoechenberger

hoechenberger (approved):
Perfect! I love it 😍

hoechenberger:
@drammock Can you add a changelog entry please? Otherwise, LGTM!!

hoechenberger:
Thanks a bunch, @drammock!

Issue #8625, opened 2020-12-08
plot_sensor_connectivity


BarryLiu-97:
When I ran plot_sensor_connectivity.py in cmd, the figure just appeared less than 1 sec, then it was gone.

agramfort:
matplotlib figures are non blocking by default so it appears and closes.

you need to use block parameter in matplotlib

GuillaumeFavelier:
Also, can you give more details on how you start the script? Because AFAIK, the 3d plots require an interactive python environment.

So `python -i ./examples/connectivity/plot_sensor_connectivity.py` or `run ./examples/connectivity/plot_sensor_connectivity.py` inside `ipython` works for me.

BarryLiu-97:
> Also, can you give more details on how you start the script? Because AFAIK, the 3d plots require an interactive python environment.
> 
> So `python -i ./examples/connectivity/plot_sensor_connectivity.py` or `run ./examples/connectivity/plot_sensor_connectivity.py` inside `ipython` works for me.

just type 'python plot_sensor_connectivity.py'

BarryLiu-97:
Ok, it works when I added -i

BarryLiu-97:
What does the -i mean?

BarryLiu-97:
And is there any way that we can see the channels' name in 3d?

BarryLiu-97:
OK, it shows the chan name when I use mayavi

GuillaumeFavelier:
From `python --help`:

```
-i     : inspect interactively after running script; forces a prompt even
         if stdin does not appear to be a terminal; also PYTHONINSPECT=x
```

It's normally enabled by default but there seems to be a regression for the `pyvista` 3d backend on the `master` branch. I'll fix it soon :)

BarryLiu-97:
Can Miyavi make respond to the mouse click? Like only when I click on the electrode, the name shows up.

GuillaumeFavelier:
This feature is not currently implemented.

BarryLiu-97:
But could Miyavi support this? Recently I was trying to do a similar task but don't know how to start.

GuillaumeFavelier:
You can try opening an issue on [mayavi](https://github.com/enthought/mayavi) to ask directly?

We are transitioning out from `mayavi` in favor of `pyvista` so I doubt such a feature will be added for the `mayavi` 3d backend.

BarryLiu-97:
Thanks!

larsoner:
[...]

PR #8626, opened 2020-12-08
FIX: point_labels

GuillaumeFavelier:
As @BarryLiu-97 pointed out in https://github.com/mne-tools/mne-python/issues/8625#issuecomment-740486402, there is a regression in the `pyvista` 3d backend making the point labels hardly visible. This PR fixes the issue.


agramfort:
the pb has been reported and fixed upstream?

GuillaumeFavelier:
A new parameter has been introduced about 2 months ago.

More details available on https://github.com/pyvista/pyvista/pull/910

agramfort:
just checking that we are not hacking and not fixing upstream

merge if you're ok

>

PR #8631, opened 2020-12-08
DOC: Clearer error message

larsoner:
From a conversation on gitter I realized this error message should be clearer about what actually needs to be done


agramfort (approved):

agramfort:
thx @larsoner

PR #8633, opened 2020-12-08
FIX: prevent raw._data overwrite from _add_exg and _add_chpi

### LINKED ISSUE #8587: BUG, ENH:  Overwriting of raw with certain tools in simulation/raw.py
#### Describe the bug
In `add_ecg()`, `add_eog()`, and `add_chpi()` functions in mne/simulation/raw.py, the newly added waveforms overwrite the existing raw data (e.g. as created with `simulate_raw()` and `add_noise()`). 

Aside from this bug, I would suggest enhancements related to dealing with the _raw.first_samp_ parameter and its behavior with event and head position information when creating simulated data.

#### Steps to reproduce
The following is an edited snippet from my own code:
```
raw_sim = simulate_raw(raw.info, source_simulator, src=src, bem=bem, head_pos=head_pos)
add_noise(raw_sim, cov=cov)    # this part is ok

add_chpi(raw_sim, head_pos=head_pos)    # this part erases the above data in 'raw_sim'
add_ecg(raw_sim)    # so does this
```

Additionally, the bug is evident in the simulation example at https://mne.tools/stable/auto_examples/simulation/plot_simulated_raw_data_using_subject_anatomy.html, and at least one other.

#### Expected and actual results
In the MNE documentation example linked above, the source activity and structured noise created in _raw_sim_ are missing after the EOG and ECG steps.  As a result, the brain map source estimations reflect only the ECG transients rather than a __summation__ of source waveforms, spatially structured noise, EOG, and ECG.

#### Additional information
I have identified the source of the bug described and plan to make my own PR with a simple fix.  However, I thought it would be a good opportunity to also make documentation revisions and code enhancements related to simulations, such as a fuller description of the output in the examples and inclusion of the _raw.first_samp_ information when adding events to the SourceSimulator object used in `simulate_raw()`.

jstout211:
#### Reference issue
Fixes #8587 

in mne/simulation/raw.py  - functions _add_exg and _add_chpi, the code set the raw._data to a variable "data".
The variable "data" was written over by the output of these functions.  This would leave the raw._data as the last operation performed. The fix is to sum the raw._data numpy array with the output from the functions.  (+= vs =).


agramfort:
thx @jstout211 

can you add an entry in the what's new page?
also I wondering if we could add a test that would fail on master and that now works?

mne/tests/test_simulation.py
```diff
@@ -0,0 +1,63 @@
...
+
+import os.path as op
+import copy
+import numpy as np
+
+import mne
+from mne.datasets import sample
+
+
+def test_simulation_cascade():
```
larsoner, line 21: Rather than creating a new file, could you put this in `mne/simulation/tests/test_raw.py`?

mne/tests/test_simulation.py
```diff
@@ -0,0 +1,63 @@
...
+    mne.simulation.add_eog(raw_cascade, random_state=0)
+    mne.simulation.add_ecg(raw_cascade, random_state=0)
+    mne.simulation.add_noise(raw_cascade, cov, random_state=0)
+
+    cascade_data = raw_cascade._data
+    serial_data = raw_eog._data + raw_ecg._data + raw_noise._data
+    # Should add chpi once fixed above
+
+    test = (cascade_data == serial_data)
+    assert test.all()
```
larsoner, line 63: Exact equality with float is often iffy and this will give a less informative error than:
```
assert_allclose(cascade_data, serial_data)
```

larsoner (commented):
Looks like a good start, can you add an entry to `doc/latest.inc` in the BUG section? It looks like you are a new contributor so please try to follow the format of other new contributors and put your entry at the top of the section -- thanks!

mne/tests/test_simulation.py
jstout211: Added test into the specified file and git rm-ed the mne/tests/test_simulation.py

jstout211:
@larsoner - I tried to add a test for the add_chpi - but the assert_close fails.  I didn't see the ability to set a seed number on that simulation.  I backed out of that modification, so it currently tests eyeblink, cardiac, and noise addition.  add_chpi does not have testing coverage even though the code was changed in the same manner within simulation/raw.py.

larsoner:
> I didn't see the ability to set a seed number on that simulation.

There is no randomness in the cHPI so nothing to seed. Can you add the code that you think should pass but does not and I'll see if I can push a commit to fix it?

jstout211:
> > I didn't see the ability to set a seed number on that simulation.
> 
> There is no randomness in the cHPI so nothing to seed. Can you add the code that you think should pass but does not and I'll see if I can push a commit to fix it?

@larsoner - I changed the test to use the chpi test dataset and added the chpi components to both signal routes.  It is not currently passing. 

image

doc/changes/names.inc
```diff
@@ -342,4 +342,6 @@
 
 .. _Hongjiang Ye: https://github.com/rubyyhj
 
+.. _Jeff Stout: https://megcore.nih.gov/index.php/Staff
```
larsoner, line 345: @jstout211 let me know if you want a different URL (or if this is not you!), I chose this from 1 minute of Google-fu

larsoner:
Your code more or less worked, once I changed to use `raw.copy()` and `raw.get_data()` rather than `deepcopy(raw)` and `raw._data` directly things worked. Not 100% sure why, maybe some weird numerical precision thing... things pass locally, let's see if CIs agree and then merge once you confirm or send a new URL above -- thanks in advance @jstout211

doc/changes/names.inc
larsoner: @jstout211 I'm going to merge this PR but feel free to comment here if the name or URL should be changed

PR #8636, opened 2020-12-09
FIX: fixed loop over norm PSF/CTF options

### LINKED ISSUE #8635: 'norm' in test_resolution_matrix()
#### Describe the bug
In lines 119 and 122 of test_resolution_matrix option norm is set to string 'norm' rather than to variable norm, i.e. it does not loop through different options for norm. 

#### Steps to reproduce
Just run the test.

#### Expected results
Should test different norm options.

#### Actual results
Only tests option 'norm'.

#### Additional information
I'll need to rewrite the corresponding section of the test.

olafhauk:
#### Reference issue
Fixes #8635.

Tests for norm in test_resolution_matrix() now loop over multiple options (not just 'norm').


agramfort (approved):

larsoner:
Thanks @olafhauk

PR #8639, opened 2020-12-09
ENH: Allow vector source estimate in psf/ctf

### LINKED ISSUE #8620: Source orientations in get_point_spread()/get_cross_talk()
get_point_spread()/get_cross_talk() can currently only deal with fixed orientations, even though the resolution matrix can be of dimension (n_orient_inv * n_dipoles, n_orient_fwd * n_dipoles). This is undesirable (one may be interested in PSFs/CTFs for different source orientations), and this is currently not dealt with properly. I would like to propose an additional option 'vector' for those functions in order to solve this.

The new API could look a bit like this. It's a bit hairy to describe the different combinations of n_orientations for forward solution and inverse operator and their consequencies for PSFs and CTFs. The default options will hopefully be easy to interpret.

```
def _get_psf_ctf(resmat, src, idx, func, mode, n_comp, norm, return_pca_vars,
                 vector=False, verbose=None):
    """Get point-spread (PSFs) or cross-talk (CTFs) functions.

    Parameters
    ----------
    resmat : array, shape (n_orient_inv * n_dipoles, n_orient_fwd * n_dipoles)
        Forward Operator.
    src : Source Space
        Source space used to compute resolution matrix.
    %(pctf_idx)s
    func : str ('psf' | 'ctf')
        Whether to produce PSFs or CTFs. Defaults to psf.
    %(pctf_mode)s
    %(pctf_n_comp)s
    %(pctf_norm)s
    %(pctf_return_pca_vars)s
    vector: Bool
        Whether to return PSF/CTF as vector source estimate (3 values per
        location) or source esimate object (1 intensity value per location).
        Only allowed to be True if corresponding dimension of resolution matrix
        is 3 * n_dipoles.
    %(verbose)s

    Returns
    -------
    stcs : instance of SourceEstimate | list of instances of SourceEstimate
    PSFs or CTFs as STC objects.
    All PSFs/CTFs will be returned as successive samples in STC objects,
    in the order they are specified in idx. STCs for different labels will
    be returned as a list.
    If resmat was computed with n_orient_inv==3 for CTFs or n_orient_fwd for
    PSFs then 3 functions per vertex will be returned as successive samples
    (i.e. one function per orientation).
    If vector=False (default) and resmat was computed with n_orient_inv==3 for
    PSFs or n_orient_fwd for CTFs, then the three values per vertex will be
    combined into one intensity value per vertex in a SourceEstimate object.
    If vector=True, PSFs or CTFs with 3 values per vertex (one per orientation)
    will be returned in a VectorSourceEstimate object.
    %(pctf_pca_vars)s
```

#### Additional comments
There could be a separate example to illustrate this option, or I could add something to the existing one (but it would probably require a few figures).

### LINKED ISSUE #10187: Size of resolution matrix (make_inverse_resolution_matrix) inconsistent with documentation
#### Describe the bug
It is unclear what the correct way to compute the resolution matrix is when using `mne.minimum_norm.make_inverse_resolution_matrix` with an inverse operator that has loose orientations. The function `_get_matrix_from_inverse_operator` is called (in line 59) to return a matrix that can be multiplied with the leadfield, and the documentation there states:

> For loose orientation constraint, the CTFs are computed for the normal component (pick_ori=‘normal’).

However, it does not seem like `pick_ori='normal'` is actually used. 

#### Steps to reproduce
```
import mne 

fname_inv = data_path + '/MEG/sample/sample_audvis-eeg-oct-6-eeg-inv.fif' # free source orientation
inverse_operator = mne.minimum_norm.read_inverse_operator(fname_inv)

fname_fwd = data_path + '/MEG/sample/sample_audvis-eeg-oct-6-fwd.fif'
forward = mne.read_forward_solution(fname_fwd)

method = "sLORETA" 
snr = 3.
lambda2 = 1. / snr ** 2

res_matrix = mne.minimum_norm.make_inverse_resolution_matrix(forward,inverse_operator,method=method,lambda2=lambda2)
```

#### Expected results
I expect the res_matrix to have dimensions nsources x nsources, where nsources is the number of sources in the source space. 

#### Actual results
The res_matrix has dimensions (3*nsources) x (3*nsources). 

#### Additional information
Output of `mne.sys_info()`: 
```
Qt WebEngine seems to be initialized from a plugin. Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute before constructing QGuiApplication.
Platform:      Linux-5.11.0-44-generic-x86_64-with-debian-bullseye-sid
Python:        3.7.9 (default, Aug 31 2020, 12:42:55)  [GCC 7.3.0]
Executable:    /home/katharina/snap/anaconda3/envs/py37cmp-gui/bin/python
CPU:           x86_64: 16 cores
Memory:        31.3 GB

mne:           0.20.7
numpy:         1.18.5 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.5.2
matplotlib:    3.2.2 {backend=Qt5Agg}

sklearn:       1.0.1
numba:         Not found
nibabel:       3.2.1
cupy:          Not found
pandas:        1.3.4
dipy:          1.1.1
mayavi:        Not found
pyvista:       Not found
vtk:           Not found
```

olafhauk:
#### Reference issue
Closes #8620
Closes #10187

Allows free orientations for get_point_spread()/get_cross_talk(), uses VectorSourceEstimate.

#### Additional information
Next step: Add illustration to existing example for PSFs/CTFs.


mne/minimum_norm/resolution_matrix.py
```diff
@@ -83,11 +84,25 @@ def _get_psf_ctf(resmat, src, idx, func, mode, n_comp, norm, return_pca_vars,
...
+    in the order they are specified in idx. STCs for different labels will
+    be returned as a list.
+    If resmat was computed with n_orient_inv==3 for CTFs or n_orient_fwd for
+    PSFs then 3 functions per vertex will be returned as successive samples
+    (i.e. one function per orientation).
+    If vector=False (default) and resmat was computed with n_orient_inv==3 for
+    PSFs or n_orient_fwd for CTFs, then the three values per vertex will be
+    combined into one intensity value per vertex in a SourceEstimate object.
+    If vector=True, PSFs or CTFs with 3 values per vertex (one per orientation)
+    will be returned in a VectorSourceEstimate object.
```
agramfort, line 105:
```suggestion
    stcs : instance of SourceEstimate | list of instances of SourceEstimate
        The PSFs or CTFs as STC objects. All PSFs/CTFs will be returned as
        successive samples in STC objects, in the order they are specified
        in idx. STCs for different labels willbe returned as a list.
        If resmat was computed with n_orient_inv==3 for CTFs or
        n_orient_fwd for PSFs then 3 functions per vertex will be returned
        as successive samples (i.e. one function per orientation).
        If vector=False (default) and resmat was computed with
        n_orient_inv==3 for PSFs or n_orient_fwd for CTFs, then the three
        values per vertex will be combined into one intensity value per
        vertex in a SourceEstimate object. If vector=True, PSFs or CTFs
        with 3 values per vertex (one per orientation) will be returned in
        a VectorSourceEstimate object.
```

mne/utils/docs.py
```diff
@@ -1087,6 +1087,13 @@
     in the order they are specified in idx. STCs for different labels will
     be returned as a list.
 """
+docdict['pctf_vector'] = """
+vector: Bool
```
agramfort, line 1091:
```suggestion
vector: bool
```

mne/utils/docs.py
```diff
@@ -1087,6 +1087,13 @@
...
+    Whether to return PSF/CTF as vector source estimate (3 values per
+    location) or source esimate object (1 intensity value per location).
+    Only allowed to be True if corresponding dimension of resolution matrix
+    is 3 * n_dipoles.
```
agramfort, line 1095:
```suggestion
    is 3 * n_dipoles. Defaults to False.
```

larsoner:
@olafhauk do you want help here? It would be good if we could take care of not just #8620 (and the related #10187) but also #8737

Also perhaps it's more consistent to have a `pick_ori='vector'` parameter/option rather than a separate `vector` one. That way with loose orientation you can choose `pick_ori='normal'` or `pick_ori='vector'` just like you do for `apply_inverse`.

pettetmw:
So in this case, `combine_xyz='spectral'` would use something like?:
```
res_mat = np.linalg.norm(np.einsum('svj,svk->vjk', M, G),  # vector dot prods
                           ord=2, axis=(1, 2))  # ord=2 spectral (largest s.v.)
```
where:
res_mat is p x p resolution matrix
M is p x n x 3 inverse operator matrix (whitened)
G is n x p x 3 leadfield gain matrix (whitened)
p is number of dipoles
n is number of sensors

larsoner (approved):
Marking for merge when green, thanks in advance @olafhauk !

PR #8643, opened 2020-12-10
FIX: scalar_bar

GuillaumeFavelier:
This PR fixes the issue following https://github.com/mne-tools/mne-python/pull/8576 about the scalar bar especially when `transparent=True`.

### DIFF
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -21,7 +21,8 @@ import numpy as np
 import vtk
 
 from .base_renderer import _BaseRenderer
-from ._utils import _get_colormap_from_array, ALLOWED_QUIVER_MODES
+from ._utils import (_get_colormap_from_array, _alpha_blend_background,
+                     ALLOWED_QUIVER_MODES)
 from ...fixes import _get_args
 from ...utils import copy_base_doc_to_subclass_doc, _check_option
 from ...externals.decorator import decorator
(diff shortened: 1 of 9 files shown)


larsoner (approved):

mne/viz/backends/_pyvista.py
```diff
@@ -874,20 +874,21 @@ def _process_events(plotter):
...
+        actor.GetMapper().SetScalarRange(*rng)
     if scalar_bar is not None:
-        scalar_bar.SetLookupTable(actor.GetMapper().GetLookupTable())
+        lut = scalar_bar.GetLookupTable()
+        if background_color is not None:
+            background_color = np.array(background_color) * 255
+            alphas = ctable[:, -1][:, np.newaxis] / 255.
+            use_table = ctable.copy()
+            use_table[:, -1] = 255.
+            ctable = (use_table * alphas) + background_color * (1 - alphas)
```
larsoner, line 889: On second thought, can you DRY this code with something like `ctable = _alpha_blend_background(ctable, background_color)`, maybe in `mne/viz/_3d.py` that you can do a nested import in these two places?
```
mne/viz/backends/_pysurfer_mayavi.py
mne/viz/backends/_pyvista.py
```
`git grep "alphas = "` showed me that these code paths are probably duplicated...

GuillaumeFavelier:
Okay now it's ready on my end.

larsoner:
Thanks @GuillaumeFavelier

PR #7258, opened 2020-01-27
FIX: fix sphere radius in topomap

mmagnuski:
@larsoner
That was the tiny bug that caused sphere not being of requested radius. The `sphere` variable was not passed to `_find_topomap_coords`.

Now it looks a bit better:

but the extrapolation is still not correct. I will check in a while, maybethe `sphere` is not passed also to some extrapolation functions?

### TODO:
- [x] change / fix previous border test
- [x] fix extrapolation

PR #8645, opened 2020-12-10
MRG, FIX: prettier "head" extrapolation

### LINKED ISSUE #8425: Default topomap interpolation a bit uglier than some time before
I noticed that current topomap interpolation with `extrapolate='head'` produces more artifacts at the edges than in the previous version. Take a look at this:
### before
image

### now
image

Now the topos look ok only in `extrapolate='local'` mode (so I've learnt to like the non-circular convex hull mask).
I think I may be partly responsible for this. I have a vague memory of adding more extrapolation points some time ago to `extrapolate='head'`, but I can't find the PR now.
I'll try to take a look at this in the upcoming week.

mmagnuski:
Fixes #8425


mne/viz/topomap.py
```diff
@@ -568,7 +568,10 @@ def _get_extra_points(pos, extrapolate, origin, radii):
     else:
         assert extrapolate == 'head'
         # return points on the head circle
-        angle = np.arcsin(distance / 2 / np.mean(radii))
+        angle = np.arcsin(distance / np.mean(radii))
+        n_pnts = np.round(2 * np.pi / angle)
+        angle = 2 * np.pi / n_pnts
+
         points_l = np.arange(0, 2 * np.pi, angle)
```
larsoner, line 575: At this point I would just do `np.linspace(0, 2 * np.pi, n_pts, endpoint=False)` -- it ensures you get exactly `n_pts` which you are not guaranteed from the `np.arange` operating on `float` due to precision issues, and you don't need to recompute `angle`.

mmagnuski:
Update:
I've increased the distance between extrapolation points and the channels:




and it looks good now. Below is the history:

### before ('box' interpolation)
image

### master ('head' with not-so-good extrapolation points placement)
image

### this PR ('head' with corrected extrapolation points placement)
topo now v2 cut

mmagnuski:
One test fails and I now remember why I increased number of extrapolation points. But I think this could be easily fixed by using `max(n_pnts, 8)` as the number of exrapolation points for example. I'll test that later.

larsoner:
I would go with 8 or 16 minimum

mmagnuski:
I just looked into the older PR where I changed the number of extrapolation points (#7258) and we had 12 extrapolation points with 5 channels back then. So I'll go with 12 as the min - no test change should be necessary then :)

mne/viz/topomap.py
```diff
@@ -568,9 +568,10 @@ def _get_extra_points(pos, extrapolate, origin, radii):
     else:
         assert extrapolate == 'head'
         # return points on the head circle
-        angle = np.arcsin(distance / 2 / np.mean(radii))
-        points_l = np.arange(0, 2 * np.pi, angle)
-        use_radii = radii * 1.1
+        angle = np.arcsin(distance / np.mean(radii))
+        n_pnts = max(12, int(np.round(2 * np.pi / angle)))
+        points_l = np.linspace(0, 2 * np.pi, n_pnts, endpoint=False)
+        use_radii = radii * 1.1 + distance
```
larsoner, line 574: @mmagnuski this `+ distance` was breaking the tests but I assume it was intentional, so I pushed a commit to adjust the test to be more lenient

mmagnuski:
@larsoner Thanks! And sorry for being slow, I saw the error and planned to fix the test.

larsoner:
No problem, it was a quick fix and I want to make sure this gets into 0.22, which we want to release today or tomorrow

mmagnuski:
Ok, I fixed the merge conflict. The "Fix bug where events could overflow when writing to FIF" was present twice in the log so I removed the second instance.

mne/viz/topomap.py
mmagnuski: :+1:

larsoner:
Thanks @mmagnuski !

mmagnuski:
Thanks @larsoner! :rocket:

PR #8647, opened 2020-12-11
Add regression test for EEGLAB data with a chanlocs struct

hoechenberger:


larsoner:
Usually we do this with whatever PR actually needs it

hoechenberger:
> Usually we do this with whatever PR actually needs it

I guess so, but this specific release I need for testing MNE-BIDS 😅 wanted to add a test to MNE later

hoechenberger:
Then again, I can already add a test to this PR. Will update this branch.

larsoner (approved):

agramfort (approved):

agramfort:
@hoechenberger please merge when all green. thx

Issue #8640, opened 2020-12-10
Tutorial figure is cropped


hoechenberger:
I have not checked if I'd get the same results running this locally, but in our rendered docs, I just came across one figure (but there's probably more than this one?) which appears to be cropped:

<img width="872" alt="Screen Shot 2020-12-10 at 10 25 36" src="https://user-images.githubusercontent.com/2046265/101752666-0cf72a00-3ad2-11eb-93cf-7859221c8859.png">

(Tutorial: https://mne.tools/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#repairing-ecg-artifacts-with-ssp)

I think this is something we should aim to get fixed before the next release.

compulsory ping @drammock 😅😁

hoechenberger:
Also we probably should ensure there's no line break in "PCA" 👌 Not sure if there's an easy way to do this though

mmagnuski:
Oh, and the EEG head is huge compared to the MEG heads. :smile:

hoechenberger:
> Oh, and the EEG head is huge compared to the MEG heads. 😄

![image](https://user-images.githubusercontent.com/2046265/101844065-dfe75d80-3b4b-11eb-9d0f-767c12a8d75f.png)

larsoner:
I think we got this in #8651, but if not feel free to reopen

mmagnuski:
@larsoner I think this PR fixed the head size inconsistency between MEG and EEG, but not the title being cropped.

larsoner:
Ahh okay, I'll reopen. Setting the milestone to 0.23 because I don't think it's a blocker for release

PR #8651, opened 2020-12-11
VIZ: Fix head size

drammock:
this makes it so that `plot_projs_topomap` uses the same sized head for EEG and MEG subplots within a figure (unless user passes their own axes). It's a quick fix to make the docs look nicer before release; was mentioned in #8640 

cc @hoechenberger @mmagnuski 

Not sure if there are other places in the codebase where we want to do this?

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -349,10 +349,7 @@ def plot_projs_topomap(projs, info, cmap=None, sensors=True,
     n_projs = len(projs)
     if axes is None:
         fig, axes, ncols, nrows = _prepare_trellis(
-            n_projs, ncols='auto', nrows='auto')
-        for ax in axes[1:]:
-            ax.sharex(axes[0])
-            ax.sharey(axes[0])
+            n_projs, ncols='auto', nrows='auto', sharex=True, sharey=True)
     elif isinstance(axes, plt.Axes):
         axes = [axes]
     _validate_if_list_of_axes(axes, n_projs)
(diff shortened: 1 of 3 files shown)


mmagnuski:
Looks good!

hoechenberger:
Doc artifact: https://24068-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/plot_50_artifact_correction_ssp.html#repairing-ecg-artifacts-with-ssp

Looks much better now!

@drammock The "compat" CI run fails:
```
2020-12-12T00:22:39.5146016Z =================================== FAILURES ===================================
2020-12-12T00:22:39.5146681Z ___________________________ test_plot_projs_topomap ____________________________
2020-12-12T00:22:39.5147493Z mne/viz/tests/test_topomap.py:111: in test_plot_projs_topomap
2020-12-12T00:22:39.5153346Z     plot_projs_topomap(projs, info=info, colorbar=True, **fast_test)
2020-12-12T00:22:39.5154340Z mne/viz/topomap.py:354: in plot_projs_topomap
2020-12-12T00:22:39.5155007Z     ax.sharex(axes[0])
2020-12-12T00:22:39.5156421Z E   AttributeError: 'AxesSubplot' object has no attribute 'sharex'
```

Do you think you could address the issue with the seemingly cropped figures, too?

drammock:
> @drammock The "compat" CI run fails:

argh, yeah, looks like using `sharex` after axes creation was only added in MPL 3.3.0.   I'll have to see if there's a way to do it at creation time that doesn't break other plot types (it uses shared _prepare_trellis code).

> Do you think you could address the issue with the seemingly cropped figures, too?

I haven't yet had a chance to dig into why this is happening in sphinx but not in interactive plotting (probably a backend difference?  I've at least ruled out the idea that sphinx-gallery was secretly calling tight_layout...).  Separate PR, but sadly I'm off for the next 5 days or so, so might not make the 0.22 release unless someone else looks into it.

PR #8658, opened 2020-12-14
MRG, BUG: Fix bugs with envcorr

### LINKED ISSUE #8649: Bug in envelope_correlation
Hi everyone,

I think there is a bug in the function `envelope_correlation()` in the case of `orthogonalize="pairwise"`.
In the original article Hipp et al, 2012 they describe the correlation as between the orthogonalized Y on X with X.
So cor(y_ort, x)
But in the code in line 119:
`corr[li] = np.dot(label_data_orth, data_mag_nomean[li])`
The code seems to calculate the orthogonalized Y on all possible X with Y.
Say we have 5 channels
Then it should be:
- C1_ort on C2 with C2
- C1_ort on C3 with C3
- C1_orth on C4 with C4
- etc
But what is being calculated in envelope_correlation() is C1_ort on C2 with C1.

I hope this is not too confusing, but I tried implementing a simple example using for loops and some of the source code to make sure I knew what timeseries are being orthogonalized to which one.

The code:
```
from mne.connectivity import envelope_correlation
from mne.filter import next_fast_len
import numpy as np
from scipy.stats import pearsonr
from scipy.signal import hilbert

n_epochs = 1
n_channels = 5
n_times = 100
data = np.random.rand(n_epochs, n_channels, n_times)
n_fft = next_fast_len(n_times)

data_hil = hilbert(data, N=n_fft, axis=-1)[..., :n_times]

corr = np.ones((n_epochs, n_channels, n_channels))

for epo_idx in range(data_hil.shape[0]):
    epoch_data = data_hil[epo_idx]
    data_mag = np.abs(epoch_data)
    data_conj_scaled = epoch_data.conj()
    data_conj_scaled /= data_mag
    for idx1 in range(data_hil.shape[1]):
        for idx2 in range(data_hil.shape[1]):
            # Get timeseries for each pair
            x = epoch_data[idx1]
            y = epoch_data[idx1]
            x_mag = np.abs(x)
            x_conj_scaled = epoch_data.conj()
            x_conj_scaled /= data_mag
            # Calculate orthogonalization
            y_orth_x = (y*x_conj_scaled[idx1]).imag
            # Estimate correlation
            cor = pearsonr(x_mag,y_orth_x)[0]
            corr[epo_idx,idx1,idx2] = cor

# Make it symmetric (it isn't at this point)
corr = np.abs(corr)
corr = (corr.T + corr) / 2.

ec = envelope_correlation(data_hil, combine=None, orthogonalize="pairwise")

print(np.allclose(corr, ec))
```

In addition to this I was also very confused with line 113 and 114:
```
# compute variances using linalg.norm (square, sum, sqrt) since mean=0
data_mag_std = np.linalg.norm(data_mag_nomean, axis=-1)
```
I tried using data_mag_std = np.std(data_mag, axis=-1) and got different results.
According to the documentation in numpy for np.std the operations should be (square, mean, sqrt) when mean = 0 and not sum. Using the sum you have to divide with N (and also sqrt afterwards)
```
std1 = np.std(data_mag, axis=-1)
std2 = np.linalg.norm(data_mag-np.mean(data_mag,axis=-1, keepdims=True),axis=-1)
print(np.allclose(std1,std2/np.sqrt(n_times)))
```

Fixing the std calculation and changing lines 119 and 120 to:
```
corr[li] = np.mean(label_data_orth*data_mag_nomean, axis=-1)
corr[li] /= data_mag_std
```

Seemed to fix the problem for me, but I didn't use the exact same implementation for my own data so it should probably be double-checked.

larsoner:
Closes #8649.

@SherazKhan @dengemann can you look? This suggests the code I adapted originally has some bugs. It doesn't drastically change results or anything, in general it looks a bit cleaner now.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -97,7 +97,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix orthogonalization of power envelopes in :func:`mne.connectivity.envelope_connectivity` (:gh:`8658` **by new contributor** |Avoide|_ and `Eric Larson`_)
+- Fix orthogonalization of power envelopes in :func:`mne.connectivity.envelope_correlation` (:gh:`8658` **by new contributor** |Qianliang Li|_ and `Eric Larson`_)
 
 - Fix a transpose issue of :func:`mne.decoding.CSP.plot_filters` (:gh:`8580` **by new contributor** |Hongjiang Ye|_)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -28,7 +28,7 @@ Current (0.22.dev0)
 
 .. |Hongjiang Ye| replace:: **Hongjiang Ye**
 
-.. |Avoide| replace:: **Avoide**
+.. |Qianliang Li| replace:: **Qianliang Li**
 
 
 Enhancements
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -28,6 +28,8 @@ Current (0.22.dev0)
 
 .. |Hongjiang Ye| replace:: **Hongjiang Ye**
 
+.. |Avoide| replace:: **Avoide**
+
 
 Enhancements
 ~~~~~~~~~~~~
@@ -95,6 +97,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix orthogonalization of power envelopes in :func:`mne.connectivity.envelope_connectivity` (:gh:`8658` **by new contributor** |Avoide|_ and `Eric Larson`_)
+
 - Fix a transpose issue of :func:`mne.decoding.CSP.plot_filters` (:gh:`8580` **by new contributor** |Hongjiang Ye|_)
 
 - Fix :func:`mne.io.read_raw_curry` to deal with Curry datasets that have channels that are listed in the labels file, but which are absent from the saved data file (e.g. 'Ref' channel).  Also now populates info['meas_date'] if possible (:gh:`8400` **by new contributor** |Tod Flak|_)
--- a/mne/connectivity/envelope.py
+++ b/mne/connectivity/envelope.py
@@ -117,7 +117,7 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
         data_conj_scaled /= data_mag
         if log:
             data_mag *= data_mag
-            np.log10(data_mag, out=data_mag)
+            np.log(data_mag, out=data_mag)
         # subtract means
         data_mag_nomean = data_mag - np.mean(data_mag, axis=-1, keepdims=True)
         # compute variances using linalg.norm (square, sum, sqrt) since mean=0
@@ -133,7 +133,7 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
                 np.abs(label_data_orth, out=label_data_orth)
                 if log:
                     label_data_orth *= label_data_orth
-                    np.log10(label_data_orth, out=label_data_orth)
+                    np.log(label_data_orth, out=label_data_orth)
                 label_data_orth -= np.mean(label_data_orth, axis=-1,
                                            keepdims=True)
                 label_data_orth_std = np.linalg.norm(label_data_orth, axis=-1)
--- a/mne/connectivity/envelope.py
+++ b/mne/connectivity/envelope.py
@@ -131,6 +131,9 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
             else:
                 label_data_orth = (label_data * data_conj_scaled).imag
                 np.abs(label_data_orth, out=label_data_orth)
+                # protect against invalid value -- this will be zero
+                # after (log and) mean subtraction
+                label_data_orth[li] = 1.
                 if log:
                     label_data_orth *= label_data_orth
                     np.log(label_data_orth, out=label_data_orth)
@@ -147,7 +150,6 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
             if absolute:
                 corr = np.abs(corr)
             corr = (corr.T + corr) / 2.
-            corr.flat[::corr.shape[0] + 1] = 0.
         corrs.append(corr)
         del corr
 
--- a/mne/connectivity/envelope.py
+++ b/mne/connectivity/envelope.py
@@ -13,7 +13,7 @@ from ..utils import verbose, _check_combine, _check_option
 
 @verbose
 def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
-                         verbose=None):
+                         log=False, absolute=True, verbose=None):
     """Compute the envelope correlation.
 
     Parameters
@@ -40,6 +40,18 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
         absolute values.
 
         .. versionadded:: 0.19
+    log : bool
+        If True (default False), square and take the log before orthonalizing
+        envelopes or computing correlations.
+
+        .. versionadded:: 0.22
+    absolute : bool
+        If True (default), then take the absolute value of correlation
+        coefficients before making each epoch's correlation matrix
+        symmetric (and thus before combining matrices across epochs).
+        Only used when ``orthogonalize=True``.
+
+        .. versionadded:: 0.22
     %(verbose)s
 
     Returns
@@ -53,6 +65,10 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
     -----
     This function computes the power envelope correlation between
     orthogonalized signals [1]_ [2]_.
+
+    .. versionchanged:: 0.22
+      Computations fixed for ``orthogonalize=True`` and diagonal entries are
+      set explicitly to zero.
 
     References
     ----------
@@ -99,6 +115,9 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
         data_mag = np.abs(epoch_data)
         data_conj_scaled = epoch_data.conj()
         data_conj_scaled /= data_mag
+        if log:
+            data_mag *= data_mag
+            np.log10(data_mag, out=data_mag)
         # subtract means
         data_mag_nomean = data_mag - np.mean(data_mag, axis=-1, keepdims=True)
         # compute variances using linalg.norm (square, sum, sqrt) since mean=0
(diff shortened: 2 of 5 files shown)


doc/changes/latest.inc
```diff
@@ -95,6 +97,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix orthogonalization of power envelopes in :func:`mne.connectivity.envelope_connectivity` (:gh:`8658` **by new contributor** |Avoide|_ and `Eric Larson`_)
```
larsoner, line 100: @Avoide can you let me know what name and URL you want to have in our [`whats_new` page](https://mne.tools/dev/whats_new.html#bugs)?

mne/connectivity/envelope.py
```diff
@@ -107,22 +126,28 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
         corr = np.empty((n_nodes, n_nodes))
         for li, label_data in enumerate(epoch_data):
             if orthogonalize is False:  # the new code
-                label_data_orth = data_mag
-                label_data_orth_std = data_mag_std
+                label_data_orth = data_mag[li]
+                label_data_orth_std = data_mag_std[li]
             else:
                 label_data_orth = (label_data * data_conj_scaled).imag
+                np.abs(label_data_orth, out=label_data_orth)
```
larsoner, line 133: First change: when orthogonalizing, use `abs` of the orthogonalized data rather than just `(data * data_conj_scaled).imag)`. Not 100% sure this is right but conceptually the thing we're correlating with is `abs(data)` so it seems to make sense to take the `abs` here, too...

I think the original operated on `np.log10(label_data_orth ** 2)` so an `abs` was implicitly taken by the squaring operation.

mne/connectivity/envelope.py
```diff
@@ -107,22 +126,28 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
...
+                    label_data_orth *= label_data_orth
+                    np.log10(label_data_orth, out=label_data_orth)
                 label_data_orth -= np.mean(label_data_orth, axis=-1,
                                            keepdims=True)
                 label_data_orth_std = np.linalg.norm(label_data_orth, axis=-1)
                 label_data_orth_std[label_data_orth_std == 0] = 1
             # correlation is dot product divided by variances
-            corr[li] = np.dot(label_data_orth, data_mag_nomean[li])
-            corr[li] /= data_mag_std[li]
+            corr[li] = np.sum(label_data_orth * data_mag_nomean, axis=1)
```
larsoner, line 142: Second change is what @Avoide talked about in the other issue, we need to correlate with the correct pairs. This (plus the `log=True, absolute=False`) makes our code match both what @Avoide suggested and what FieldTrip produces.

mne/connectivity/tests/test_envelope.py
```diff
@@ -15,19 +15,23 @@
...
+                # Get timeseries for each pair
+                x, y = epoch_data[ii], epoch_data[jj]
+                x_mag = np.abs(x)
+                x_conj_scaled = x.conj()
+                x_conj_scaled /= x_mag
+                # Calculate orthogonalization
+                y_orth_x = (y * x_conj_scaled).imag
+                y_orth_x_mag = np.abs(y_orth_x)
+                # Estimate correlation
+                corr[ii, jj] += np.abs(np.corrcoef(x_mag, y_orth_x_mag)[0, 1])
```
larsoner, line 32: @Avoide you might recognize some of this :)

mne/connectivity/envelope.py
Avoide: I am not sure about the log10, but they do square and take the log-transform:
Citation from their supplementary information:
_"The  signal  Y(t,f)  is  orthogonalized  in  the  complex  plane  with respect to X(t,f). This results in a positive number |Y_orth_X(t,f)| which is then squared and log-transformed, and correlated with [|X(t,f)|."_ [Source](https://static-content.springer.com/esm/art%3A10.1038%2Fnn.3101/MediaObjects/41593_2012_BFnn3101_MOESM9_ESM.pdf)

doc/changes/latest.inc
Avoide: You can put in:
Qianliang Li and [URL](https://www.dtu.dk/english/service/phonebook/person?id=126774)

agramfort:
@SherazKhan @dengemann it's your code originally. Can you take a minute to have a lookt? thanks

mne/connectivity/envelope.py
larsoner: log10 is the same as log up to a scale factor, and that scale factor goes away in the correlation so it doesn't really matter. But `np.log` is faster than `np.log10` so why not!

larsoner: ... but that text does make it clear that it should result in a positive number even before the square and log transform, so I think this is correct now (and will continue to be on the next push where I change `np.log10` to `np.log`)

dengemann:
Thank you everyone for looking into this. This work is currently not loaded into my working memory. What I can say though is that I always had a feeling since our working session at Biomag2018 that the MNE implementation does something different than the prototype.
I'll have a look at the code now.

mne/connectivity/envelope.py
```diff
@@ -107,22 +126,28 @@ def envelope_correlation(data, combine='mean', orthogonalize="pairwise",
...
+            corr[li] /= data_mag_std
             corr[li] /= label_data_orth_std
         if orthogonalize is not False:
             # Make it symmetric (it isn't at this point)
-            corr = np.abs(corr)
+            if absolute:
+                corr = np.abs(corr)
             corr = (corr.T + corr) / 2.
+            corr.flat[::corr.shape[0] + 1] = 0.
```
dengemann, line 150: Perhaps consider adding a comment for Python newcomers that this updates the diagonal efficiently.

dengemann:
@SherazKhan can you also take a look?

mne/connectivity/tests/test_envelope.py
```diff
@@ -15,19 +15,23 @@
...
-        label_data_orth = np.imag(label_data * (data.conj() / np.abs(data)))
-        label_data_orig = np.abs(label_data)
-        label_data_cont = np.transpose(
-            np.dstack((label_data_orig, np.transpose(label_data_orth,
-                                                     (1, 2, 0)))), (1, 2, 0))
-        corr_mats[index] = np.array([np.corrcoef(dat)
-                                     for dat in label_data_cont])[:, 0, 1:].T
-    corr_mats = np.transpose(corr_mats, (2, 0, 1))
-    corr = np.mean(np.array([(np.abs(corr_mat) + np.abs(corr_mat).T) / 2.
-                             for corr_mat in corr_mats]), axis=0)
```
larsoner, line 30: >  I always had a feeling since our working session at Biomag2018 that the MNE implementation does something different than the prototype.

@dengemann IIRC the code here was the prototype that I used to code a more efficient version in MNE, so it should have replicated it properly. Unfortunately I had to change this prototype code to do something different in order to get it to follow FieldTrip and the paper definitions...

dengemann: No problem, If we can come up with a more flexible set of options this is good for research on this method.
The comparisons in the previous discussion suggest that difference may be rather subtle.

mne/connectivity/envelope.py
SherazKhan: I think abs was missing in the original code, because of the squaring in the next step.

mne/connectivity/envelope.py
SherazKhan: We were using cor, but this is much faster

> Second change is what @Avoide talked about in the other issue, we need to correlate with the correct pairs. This (plus the `log=True, absolute=False`) makes our code match both what @Avoide suggested and what FieldTrip produces.

We were using np.corrcoef but this is much faster

mne/connectivity/tests/test_envelope.py
SherazKhan: I second, differences will be subtle.

mne/connectivity/tests/test_envelope.py
larsoner, line 32: @SherazKhan the actual critical change in the code below is from `data_mag_nomean[li]` which only used the `li`th label's data to correlate with all N labels `label_data_orth` series, and this code now uses `data_mag_nomean`, i.e., all `N` labels data to correlate with all N labels `label_data_orth` series.

It's probably easier to see if you look at the "simplified" version here. The old/`master` version of the code does the correlation with the `jj` original (magnitude) data that you could call `y_mag`:
```
                corr[ii, jj] += np.abs(np.corrcoef(np.abs(epoch_data[jj]), y_orth_x)[0, 1])
```
Instead of this correlation, with (magnitude) of the `ii`th data that is called `x_mag` here, which is now implemented in this PR, equivalently written as:
```
                corr[ii, jj] += np.abs(np.corrcoef(np.abs(epoch_data[ii]), y_orth_x)[0, 1])
```

SherazKhan: @larsoner indeed, this is more closer to the original Hipp version.

larsoner:
Updated example looks reasonable:

https://24108-1301584-gh.circle-artifacts.com/0/dev/auto_examples/connectivity/plot_mne_inverse_envelope_correlation.html#compute-the-degree-and-plot-it

dengemann:
> Updated example looks reasonable:

I agree the change looks reasonable.
But I am still baffled that the connectivity matrix looks so noisy, apparently lacking structure.
In the early examples we always saw some structure.

larsoner:
The one on `master` isn't much better -- I think it's mostly because now we `.crop(0, 60)` for speed. If we remove that it (takes much longer and) looks like this on this PR:

Screenshot from 2020-12-14 15-04-14

Screenshot from 2020-12-14 15-04-17

dengemann:
@larsoner yeah it's not a blocker, it was the same concern on master.

SherazKhan:
The example should look similar to the original Hipp NN 2012
image

SherazKhan:
The example should look similar to the original Hipp NN 2012
image

SherazKhan:
@dengemann and I was able to reproduce something similar using camcan data

image

dengemann:
@SherazKhan I think we should repeat those benchmarks soon :)

SherazKhan:
I did some testing on my data, please the following for comparison between various options:

image

larsoner:
@SherazKhan to my eye the clearest result is `imag abs`, which is now the default actually (and it's better than original imag, which is the default in `master` and is no longer available). Do you agree?

SherazKhan:
> @SherazKhan to my eye the clearest result is `imag abs`, which is now the default actually (and it's better than original imag, which is the default in `master` and is no longer available). Do you agree?

@larsoner I agree.

larsoner:
Thanks for looking, discussing, and testing @dengemann @SherazKhan !

PR #8662, opened 2020-12-15
MRG, BUG, ENH: Add window option

### LINKED ISSUE #8660: PSDs computed with psd_welch look different between 0.20 and 0.21
#### Describe the bug

I noticed that the output of ``` mne.time_frequency.psd_welch``` looks slightly different between versions, example for a sine wave:

Screen Shot 2020-12-14 at 21 39 50

#### Steps to reproduce
Here is a minimum working example with simulated data: 

```python
import mne
import numpy as np

# some simulated signal
frequency = 10
dt = 0.001
fs = 1 / dt
n_seconds = 10
time = np.arange(0, n_seconds, dt)
signal = np.cos(2 * np.pi * frequency * time)

info = mne.create_info(["signal1"], fs, ch_types="eeg")
raw = mne.io.RawArray(signal[np.newaxis], info)

# compute PSD
psd, freq = mne.time_frequency.psd_welch(raw, fmin=1, fmax=30, n_fft=4000)

# plot
fig, ax = plt.subplots()
ax.semilogy(freq, psd.T)
fig.show()
```

#### Expected results
I expected it to look the same between versions. Is this a desired change? I liked the old behavior. :-) Can it be achieved with the new version somehow? I checked the change log for 0.21 where optimization of ```psd_welch``` for long data segments is reported, which may have led to this?

#### Additional information
I tested various MNE versions and the change appears between 0.20.8 and 0.21.0 and persists to 0.21.2.

larsoner:
Closes #8660

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -83,6 +83,8 @@ Enhancements
 
 - Add option ``on_header_missing`` to :func:`mne.channels.read_polhemus_fastscan` (:gh:`8622` by `Eric Larson`_)
 
+- Add option ``window`` to :func:`mne.time_frequency.psd_welch` and related functions (:gh:`8862` by `Eric Larson`_)
+
 - `mne.preprocessing.ICA.plot_sources` now displays an `mne.preprocessing.ICA.plot_properties` window when right-clicking on component names on the y-axis (:gh:`8381` by `Daniel McCloy`_)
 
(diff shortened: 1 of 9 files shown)


agramfort (approved):

agramfort:
CIs complain though

cbrnr:
Beautiful - thanks @larsoner!

PR #8663, opened 2020-12-15
MRG, BUG: Fix alpha for volumes

larsoner:
Some simple code to plot with `hot_r` with `transparent=True` and `alpha=1.` (should yield an opaque volume where values are large) looks like this on `master`:

Screenshot from 2020-12-15 12-38-12

And on this PR it's correct:

Screenshot from 2020-12-15 12-37-18

The key was the opacity and color transfer functions both need to yield floats between 0 and 1, so we just need to divide both by 255 (ctable is `np.int8`) rather than doing whatever silly gymnastics we were doing before where the colors ranged from 0 to 255 and the alphas 0 to 1/255 (effectively).

While I was messing around with this I noticed `add_data` could not use `colormap='mne'` so I refactored our `get_cmap` calls. I also added a simple test to `test_brain` to make sure our `ctable` is correct.

@GuillaumeFavelier can you take a quick look? Not sure if this or #8335 should go in first, but I think they're going to conflict.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -137,6 +137,8 @@ Bugs
 
 - Fix bug with coordinate frames when performing volumetric morphs via :func:`mne.compute_source_morph` and :meth:`mne.SourceMorph.apply` that could lead to ~5 mm bias (:gh:`8642` by `Eric Larson`_)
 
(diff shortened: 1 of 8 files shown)


GuillaumeFavelier (commented):
Awesome! LGTM. And don't worry about the conflicts, I'll fix these in #8335 👍

GuillaumeFavelier (approved):

larsoner:
Thanks for the quick review @GuillaumeFavelier !

PR #8667, opened 2020-12-17
Vendor ga.js; remove plusone.js – not needed anymore

### LINKED ISSUE #8664: The MNE website is really slow to access from China
I don't whether this is a suitable problem to issue here. However, I can't find other ways to report this problem. The problem has last for a long time, and recently the speed is intolerably slow. I ask several friends to test from different regions in China, the problem persists. And when I access the MNE website by VPN, the speed improves a lot.

hoechenberger:
Attempts to fix #8664


drammock (commented):
I confess I'm not optimistic that this will work.  The vendored JS still includes calls to URLs within the google domain.  Maybe things are different now (I left ~5 years ago), but my experience living in china was that even though these calls are *supposed* to be asynchronous, website loading would stall as soon as the site got to a point where a google URL was being accessed --- fonts, search, analytics, etc (you could tell because the little browser statusbar at the bottom of the window would say what site it was trying to access).  I'm also not optimistic because it seems like it can't possibly be this easy to fool the great firewall, and if it were, I would expect it not to work for very long.

I'd rather have the conversation about whether we actually need website analytics, and if we do, then are there alternatives to google analytics that will not hamstring our website for users in China.  Here are at least two FOSS options: [matamoto](https://matomo.org/) and [OWA](http://www.openwebanalytics.com/). **EDIT:** another option (used by scikit-image) is https://plausible.io/, which is subscription-based, with the price tied to number of page hits.

Last point: Are we sure that including [Apache-2.0](https://choosealicense.com/licenses/apache-2.0/) licensed JS code in a [BSD 3-clause](https://choosealicense.com/licenses/bsd-3-clause/) project is OK?  Before we comb through the licenses in detail I wonder if someone already knows the answer / knows of a tool or lookup table that has the answer.

hoechenberger:
Closing as it doesn't fix the issue

larsoner:
> I'd rather have the conversation about whether we actually need website analytics, and if we do, then are there alternatives to google analytics that will not hamstring our website for users in China. Here are at least two FOSS options: matamoto and OWA.

@agramfort set up the google analytics I think -- thoughts? I mostly just use it for seeing how much traffic we get. At one point I tried to use some google dashboards to get the old versions of pages that show up in search results to point to their newer variants, but gave up for some reason. So from my end I'm fine with switching to something else. If we need to reenable it later I guess we can, just with a temporary outage in our usage numbers.

If there are trivial things to move to `_static` like fonts, maybe we should do that, too. This could probably easily be a build-time step in `conf.py` to copy them into `_static` if they are missing (and have them in `.gitignore`) so that whenever docs are built they show up, but they don't need to be committed to our repo.

drammock:
> If there are trivial things to move to `_static` like fonts, maybe we should do that, too.

Fonts are already non-Google and are stored in _static as WOFFs. I agree that the only thing I've used it for is really basic stuff (total users, users by country)... the kind of stuff that any analytics software will do. AFAICT the real power of Google's product comes when you are buying advertising from them, which we'll never do.

PR #8669, opened 2020-12-17
MRG, MAINT: add codemeta

drammock:
Boilerplate was generated using https://codemeta.github.io/codemeta-generator/ and the things that might change (release date, release version, dependencies, and authors) are update by the new script `tools/generate_codemeta.py` which can be run at release time.

### DIFF
--- a/tools/generate_codemeta.py
+++ b/tools/generate_codemeta.py
@@ -3,8 +3,8 @@ import subprocess
 
 # UPDATE THESE AS NEEDED WITH EACH RELEASE
 release_date = '2020-12-17'
-release_version = '0.22'
-dependencies = ['Python 3.6', 'NumPy 1.15.4', 'SciPy 1.1']
+release_version = '0.22.0'  # major.minor.patch; must include trailing zero
+dependencies = ('Python 3.6', 'NumPy 1.15.4', 'SciPy 1.1')
 # add to these as necessary
 compound_surnames = (
     'García Alanis',
@@ -27,7 +27,7 @@ def parse_name(name):
     name, email = name_and_email.split(' <')
     email = email.strip('>')
     email = '' if 'noreply' in email else email  # ignore "noreply" emails
-    name = ' '.join(name.split('.'))             # remove periods
+    name = ' '.join(name.split('.'))             # remove periods from initials
     # handle compound surnames
     for compound_surname in compound_surnames:
         if name.endswith(compound_surname):
@@ -74,7 +74,7 @@ codemeta_boilerplate = f'''{{
     "dateCreated": "2010-12-26",
     "datePublished": "2014-08-04",
     "dateModified": "{release_date}",
-    "downloadUrl": "https://github.com/mne-tools/mne-python/archive/v0.22.0.zip",
+    "downloadUrl": "https://github.com/mne-tools/mne-python/archive/v{release_version}.zip",
     "issueTracker": "https://github.com/mne-tools/mne-python/issues",
     "name": "MNE-Python",
     "version": "{release_version}",
--- a/tools/generate_codemeta.py
+++ b/tools/generate_codemeta.py
@@ -30,7 +30,7 @@ def parse_name(name):
     name = ' '.join(name.split('.'))             # remove periods
     # handle compound surnames
     for compound_surname in compound_surnames:
-        if compound_surname in name:
+        if name.endswith(compound_surname):
             ix = name.index(compound_surname)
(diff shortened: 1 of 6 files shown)


larsoner:
The authors look like something that could be easily scripted except for the ORCID ID.

drammock:
CIs all green, ready for review/merge

tools/generate_codemeta.py
```diff
@@ -0,0 +1,112 @@
+import os
+import subprocess
+
+# UPDATE THESE AS NEEDED WITH EACH RELEASE
+release_date = '2020-12-17'
+release_version = '0.22.0'  # major.minor.patch; must include trailing zero
```
larsoner, line 6: Can't these use `datetime.today().strptime` and `import mne; mne.__version__` and then assert that the version is  formatted as `X.Y.Z` (no `dev`) for example?

tools/generate_codemeta.py
```diff
@@ -0,0 +1,112 @@
...
+dependencies = ('Python 3.6', 'NumPy 1.15.4', 'SciPy 1.1')
```
larsoner, line 7: Can we parse from the README or `setup.py`? Hopefully one or the other and they are already DRY, but I'm not sure we've implemented that...

tools/generate_codemeta.py
```diff
@@ -0,0 +1,112 @@
...
+    "version": "{release_version}",
+    "description": "MNE-Python is an open-source Python package for exploring, visualizing, and analyzing human neurophysiological data. It provides methods for data input/output, preprocessing, visualization, source estimation, time-frequency analysis, connectivity analysis, machine learning, and statistics.",
+    "applicationCategory": "Neuroscience",
+    "developmentStatus": "active",
+    "referencePublication": "https://doi.org/10.3389/fnins.2013.00267",
+    "keywords": [
+        "MEG",
+        "EEG",
+        "fNIRS",
+        "eCoG",
```
larsoner, line 89: Usually ECoG (cap E) in our code

tools/generate_codemeta.py
```diff
@@ -0,0 +1,112 @@
...
+    'García Alanis',
+    'van Vliet',
+    'De Santis',
+    'Dupré la Tour',
+    'de la Torre',
+    'van den Bosch',
+    'Van Der Donckt',
+    'van der Meer',
+    'van Harmelen',
+    'Visconti di Oleggio Castello'
```
larsoner, line 19: Can you make a note in the releases wiki to check and update this if needed?

larsoner (commented):
Have some ideas for making things more DRY/reduce number of steps this adds to our processes.

Also can you write a corresponding line in the wiki for making a release for example:
```Markdown
4. PENDING PR 8669: Check `tools/generate_codemeta.py` against new names in `doc/changes/latest.inc` for compound surnames, then run `tools/generate_codemeta.py` to update `codemeta.json`
```
Choosing the number/location based on probably needing to come after changing the version, but before pushing a commit, and before making the GitHub release? I think this plus my inline suggestions would reduce the "added effort" to support this to be ~1 minute during the "making a release" stage, which would be great!

tools/generate_codemeta.py
drammock: The wiki is already updated (step 7):

> Edit tools/generate_codemeta.py to update the variables release_date, release_version, dependencies (if needed), and compound_surnames (if needed). Then from the tools folder run python/generate_codemeta.py to update the codemeta metadata for this release.

I'll edit it once I've incorporated your other suggestions (e.g., automating version and release date)

drammock: comments addressed, and wiki has been updated accordingly.

drammock:
CIs all green again, ready for second review.

larsoner (approved):
So the only thing we have to do from now on (other than make sure this code still works) is, during release:

>  7. Edit tools/generate_codemeta.py to update the compound_surnames variable (if needed, based on new contributors from the changelog). Then from the tools folder run python generate_codemeta.py to update the codemeta metadata for this release.

Sounds easy enough to me, +1 for merge

drammock:
just found a small bug, don't merge yet.

drammock:
OK, local testing in an env using mne-stable seems to work correctly now.  after running it I manually back-dated the `dateModified` field to the release date of 0.22, but when doing actual releases this should not be necessary.

agramfort (approved):
thanks @drammock ! merge if good to go from your end

PR #8670, opened 2020-12-17
[WIP, DOC] Frequency Interference Causing 1/f component of power spectrum example

alexrockhill:
This example is based on a conversation I had with John Mosher where I mistakenly thought the Beer-Lambert Law applied to 1-100 Hz oscillations recorded within the brain and detected at MEG/EEG sensors. This wavelength is far too long to be attenuated by brain, CSF, skull and skin, but if it did, a 1/f exponential decay would be observed in the power spectral density plot as is actually observed, likely leading to this misconception. An alternate, more plausible explanation is that the same time offset in phase of spatially nearby dipoles causes greater destructive interference for greater frequencies as shown in this example. I thought this example would be useful to explain this idea.

### DIFF
--- a/examples/forward/plot_frequency_interference.py
+++ b/examples/forward/plot_frequency_interference.py
@@ -2,13 +2,13 @@
 """
 .. _ex-frequency-interference:
 
-Frequency Interference and 1/f Power Spectra
-============================================
+Frequency Interference and the 1/f Component in Power Spectra
+=============================================================
 
 This example shows how the spatial offset of nearly phase-locked source time
 courses with lower frequency oscillations have much less destructive
 interference than those with higher frequencies leading to a 1/f power
-spectral density (psd) distribution. This phenomena is commonly
+spectral density (PSD) distribution. This phenomena is commonly
 misattributed to the skull acting as a lowpass filter.
 
 .. contents:: This tutorial covers:
@@ -34,7 +34,8 @@ from mne.simulation import simulate_evoked
 
 data_path = sample.data_path()
 subjects_dir = op.join(data_path, 'subjects')
-sample_dir = op.join(data_path, 'MEG', 'sample')
+subject = 'sample'
+sample_dir = op.join(data_path, 'MEG', subject)
 
 # Read evoked data
 fname_evoked = op.join(sample_dir, 'sample_audvis-ave.fif')
@@ -42,24 +43,33 @@ evoked = mne.read_evokeds(fname_evoked, condition=0, baseline=(None, 0))
 
 
 ##############################################################################
-# Visualize 1/f power spectral density in m/eeg data
+# Visualize 1/f power spectral density in M/EEG data
 # --------------------------------------------------
 #
-# The power decreases with increasing frequency, this is usually an
-# exponential decay.
+# The power decreases with increasing frequency, and is usually an
+# exponential decay. This is shown in the loglog plots below which have a
+# negative linear trend (exponential decay in normal cartesian coordinates).
 
 psds, freqs_psd = psd_array_multitaper(
-    evoked.data, evoked.info['sfreq'], fmin=2, fmax=50)
+    evoked.data, evoked.info['sfreq'], fmin=2, fmax=40)
 fig, axes = plt.subplots(1, 3, figsize=(10, 5))
-axes[0].plot(freqs_psd, psds[mne.pick_types(evoked.info, meg='mag')].T)
-axes[0].set_title('mag')
-axes[1].plot(freqs_psd, psds[mne.pick_types(evoked.info, meg='grad')].T)
-axes[1].set_title('grad')
-axes[2].plot(freqs_psd, psds[mne.pick_types(evoked.info, eeg=True)].T)
-axes[2].set_title('eeg')
-fig.suptitle('1/f Power Decay in Real Evoked')
-axes[0].set_ylabel('Power')
-axes[1].set_xlabel('Frequency (Hz)')
+axes[0].plot(
+    freqs_psd, 10 * np.log10(psds[mne.pick_types(evoked.info, meg='mag')]).T)
+axes[0].set_title('Magnetometers')
+axes[1].plot(
+    freqs_psd, 10 * np.log10(psds[mne.pick_types(evoked.info, meg='grad')]).T)
+axes[1].set_title('Gradiometers')
+axes[2].plot(
+    freqs_psd, 10 * np.log10(psds[mne.pick_types(evoked.info, eeg=True)]).T)
+axes[2].set_title('EEG')
+fig.suptitle('1/f Power Decay in Real Evoked Data')
+axes[0].set_ylabel('Power (dB)')
+axes[1].set_xlabel('Log10 of Frequency (Hz)')
+for ax in axes:
+    ax.set_xscale('log')
+    ax.set_xticks([5, 10, 40])
+    ax.set_xticklabels([5, 10, 40])
+
 fig.show()
 
 ##############################################################################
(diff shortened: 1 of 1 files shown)


alexrockhill:
Documentation artifact link: https://24198-1301584-gh.circle-artifacts.com/0/dev/auto_examples/forward/plot_frequency_interference.html#sphx-glr-auto-examples-forward-plot-frequency-interference-py

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+"""
+.. _ex-frequency-interference:
+
+Frequency Interference
+======================
+
+This example shows how the spatial offset of nearly phase-locked source time
+courses with lower frequency oscillations have much less destructive
+interference than those with higher frequencies leading to a 1/f power
+spectral density (psd) distribution. This phenomena is commonly
```
agramfort, line 11:
```suggestion
spectral density (PSD) distribution. This phenomena is commonly
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+subjects_dir = op.join(data_path, 'subjects')
+sample_dir = op.join(data_path, 'MEG', 'sample')
+
+# Read evoked data
+fname_evoked = op.join(sample_dir, 'sample_audvis-ave.fif')
+evoked = mne.read_evokeds(fname_evoked, condition=0, baseline=(None, 0))
+
+
+##############################################################################
+# Visualize 1/f power spectral density in m/eeg data
```
agramfort, line 45:
```suggestion
# Visualize 1/f power spectral density in M/EEG data
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+# --------------------------------------------------
+#
+# The power decreases with increasing frequency, this is usually an
+# exponential decay.
+
+psds, freqs_psd = psd_array_multitaper(
+    evoked.data, evoked.info['sfreq'], fmin=2, fmax=50)
+fig, axes = plt.subplots(1, 3, figsize=(10, 5))
+axes[0].plot(freqs_psd, psds[mne.pick_types(evoked.info, meg='mag')].T)
+axes[0].set_title('mag')
```
agramfort, line 55:
```suggestion
axes[0].set_title('Magnetometers')
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+axes[1].plot(freqs_psd, psds[mne.pick_types(evoked.info, meg='grad')].T)
+axes[1].set_title('grad')
```
agramfort, line 57:
```suggestion
axes[1].set_title('Gradiometers')
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+axes[2].plot(freqs_psd, psds[mne.pick_types(evoked.info, eeg=True)].T)
+axes[2].set_title('eeg')
```
agramfort, line 59:
```suggestion
axes[2].set_title('EEG')
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+fig.suptitle('1/f Power Decay in Real Evoked')
+axes[0].set_ylabel('Power')
+axes[1].set_xlabel('Frequency (Hz)')
+fig.show()
```
agramfort, line 63: usually to show the 1/f you plot in loglog as in:

image

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+phase_shift = 1 / stc.sfreq  # 1 ms shifted per mm distant
+max_dist = 30  # distance in mm of nearly in-phase sources
+
+
+def euclidean_dist(vert0, vert1):
+    return np.sum((vert1 - vert0) ** 2) ** 0.5
+
+
+fig, axes = plt.subplots(1, freqs.size, figsize=(10, 5))
+fig.suptitle('Source Time Courses')
```
agramfort, line 99:
```suggestion
fig.suptitle('PSD of Source Time Courses')
```

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,164 @@
...
+    axes[2].set_title(f'eeg {freq} Hz')
+    axes[2].set_ylim([0, 1e-11])
+
+
+for ax in axes_all.flatten():
+    ax.set_xlim([5, 50])
+
+
+axes_all[1, 0].set_ylabel('Power')
+axes_all[0, freqs.size // 2].set_xlabel('Frequency (Hz)')
```
agramfort, line 161: it shows that the decay is even stronger on the EEG so the skull does increase the interference no?

alexrockhill: Hmmm that does seem like the logical conclusion but it also isn't the main effect, but maybe worth being a bit more conservative with the language

alexrockhill:
I only used one 30 mm radius cortical patch for the last example and I think the idea is that many patches with particular frequency time offsets would cause the exponential decay when summed. I'm working on another section to add onto the end which I think will take it full circle to the first loglog plot.

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,178 @@
...
+spectral density (PSD) distribution. This phenomena is commonly
+misattributed to the skull acting as a lowpass filter.
+
+.. contents:: This tutorial covers:
+   :local:
+   :depth: 2
+
+"""
+# Authors: Alex Rockhill <aprockhill@mailbox.org>
+#          John Mosher   <John.C.Mosher@uth.tmc.edu>
```
jasmainak, line 20: have you asked John Mosher if he's okay with being added as an author here? :)

alexrockhill: Yes he said that was good

examples/forward/plot_frequency_interference.py
```diff
@@ -0,0 +1,178 @@
...
+.. _ex-frequency-interference:
+
+Frequency Interference and the 1/f Component in Power Spectra
+=============================================================
+
+This example shows how the spatial offset of nearly phase-locked source time
+courses with lower frequency oscillations have much less destructive
+interference than those with higher frequencies leading to a 1/f power
+spectral density (PSD) distribution. This phenomena is commonly
+misattributed to the skull acting as a lowpass filter.
```
jasmainak, line 12: there might actually be a few references you might want to put here

jasmainak:
I leave it up to @agramfort to decide but IMO this is more like an "opinion piece" that's better reserved for a blog or paper ;-) The fact that you can produce 1/f with a forward model tells me nothing about whether there could not be other processes which contribute to the 1/f and if it's the dominating effect. Furthermore, we don't really disprove the "lowpass" hypothesis here. It's more a consequence of the Math and the quasistatic approximation

alexrockhill:
Hmmm it seems like it gets a bit more complicated when you start combining patches (see last figure https://24211-1301584-gh.circle-artifacts.com/0/dev/auto_examples/forward/plot_frequency_interference.html#sphx-glr-auto-examples-forward-plot-frequency-interference-py) especially for MEG.

alexrockhill:
> I leave it up to @agramfort to decide but IMO this is more like an "opinion piece" that's better reserved for a blog or paper ;-) The fact that you can produce 1/f with a forward model tells me nothing about whether there could not be other processes which contribute to the 1/f and if it's the dominating effect. Furthermore, we don't really disprove the "lowpass" hypothesis here. It's more a consequence of the Math and the quasistatic approximation

Hmmm maybe we should write a paper with references and more formality first. Good point @jasmainak :)

jasmainak:
I can share some scripts with your privately that might help with getting a proper 1/f ;-) It depends on the size of your patch etc. Anyhow, feels a bit premature to me to push as an educational piece in MNE. +1 on paper :-)

PR #8675, opened 2020-12-18
MRG, BUG: Fix bug with super versus eq in eSSS

larsoner:
In general the eSSS basis / projectors just need to be a superset of the good MEG channels in the data. This is particularly important when using empty-room data for an experiment with multiple runs where the runs could have different automatic (or manual) bad channels set.


agramfort (approved):

agramfort:
@larsoner I let you merge and backport

PR #8683, opened 2020-12-29
MRG, CI: never skip codespell/flake

drammock:
per discussion on somewhere (Slack? gitter?) this removes the check-skip job for the GitHub workflow for codespell/flake, because 

1. style violations are easy to introduce, and
2. the style check is really fast anyway so it's fine to always run it even when you want to skip other workflows.


drammock:
CIs green, ready for review/merge

larsoner:
Thanks @drammock

PR #8688, opened 2020-12-30
MRG, VIZ: fix shortcuts and interactivity for raw.plot() with macosx backend

### LINKED ISSUE #8687: Some keyboard shortcuts are broken in raw.plot
On macOS, toggling the scale bar (`s`) in an open `raw.plot` window doesn't work immediately. The scale bar is only toggled after e.g. switching to zen mode back and forth.

The help window shows incorrect shortcuts to increase/decrease window length and visible channels. For example, the window lists ⌘ + → as a shortcut to show a longer window, but the correct shortcut is fn + →. The same is true for the other three shortcuts.

F11 doesn't enter fullscreen mode, probably because this key is mapped to showing the desktop by default. I don't think we need a fullscreen shortcut on macOS because users can click the green button, but if we want to keep a shortcut we should change it to something else (at least on macOS where ^⌘F seems to be the default for full screen).

@drammock another one for you? If you don't have time I can also take a look (but not immediately).

drammock:
closes #8687 

- scalebar toggle now works correctly with `macosx` backend (previously it showed, but did not hide, scalebars immediately)
- zen mode now works correctly with `macosx` backend (previously it would show/hide the scrollbars but not use the freed space)
- help dialog corrected: macOS uses `fn+→` not `⌘+→` (similar for other 3 arrow keys)
- help dialog no longer shows `F11` as a fullscreen key for macOS (it didn't work, and fullscreen is built into the OS, so we just don't show a fullscreen key now)

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -21,6 +21,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix zen mode and scalebar toggling for :meth:`raw.plot() <mne.io.Raw.plot>` when using the ``macosx`` matplotlib backend (:gh:`8688` by `Daniel McCloy`_)
+
 - Fix bug with :func:`mne.preprocessing.maxwell_filter` where the eSSS basis had to exactly match the good channels instead of being a superset (:gh:`8675` by `Eric Larson`_)
 
 API changes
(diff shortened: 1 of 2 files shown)


cbrnr:
Thanks @drammock, this is 💯! Works like a charm!

cbrnr (approved):

cbrnr:
Can you add a changelog entry?

cbrnr:
Thanks @drammock!

agramfort:
Shall we backport ?

PR #8689, opened 2020-12-30
MRG, VIZ: fix annotation redraw when changing window duration

### LINKED ISSUE #8686: Enlarging time range in raw.plot doesn't draw new annotations
Here's a simple example to reproduce:

```python
import numpy as np
import mne

fs = 500
n_chans = 10
data = np.random.randn(n_chans, fs * 100) * 5e-6
info = mne.create_info(n_chans, fs, "eeg")
raw.annotations.append(onset=[1, 5, 9, 12, 15, 19],
                       duration=[2, 5, 3, 3, 5, 3],
                       description=["A1", "A1", "B2", "C3", "A1", "B2"])
raw.plot(n_channels=n_chans)
```

Once the plot window is open, increase the time window and notice that annotations outside the initial time window are not drawn. Here's what the plot looks like after increasing the window 13 times (notice that annotations from 10s are not drawn, but the overview bar still shows them):

Screen Shot 2020-12-30 at 16 08 50

@drammock do you have time to take a look?

drammock:
closes #8686

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,6 +23,8 @@ Bugs
 ~~~~
 - Fix bug with :func:`mne.preprocessing.maxwell_filter` where the eSSS basis had to exactly match the good channels instead of being a superset (:gh:`8675` by `Eric Larson`_)
 
+- Fix bug with :meth:`raw.plot() <mne.io.Raw.plot>` where annotations didn't immediately appear when changing window duration (:gh:`8689` by `Daniel McCloy`_)
+
 API changes
 ~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


cbrnr (approved):
Indeed this fixes the issue! Thanks @drammock!

cbrnr:
The only thing missing before this should be merged is a changelog entry.

cbrnr:
Thanks @drammock!

larsoner:
FYI @drammock looks like you missed the `doc/changes/0.22.inc` update when backporting, I put it in https://github.com/mne-tools/mne-python/commit/133b41f6b884e4739f22978debcc94567fa563eb

PR #8695, opened 2021-01-04
MRG: Use caching in Github Actions and Azure Pipelines

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/issues/8685#issuecomment-753985292 and attempts to solve the recent failures related to testing data download.

ToDo:

- [x] Use `$HOME` for `path` variable
- [x] Use caching for Azure too

PR #8696, opened 2021-01-04
MRG, MAINT: Better downloading for testing and misc

larsoner:
1. ~~Move the variables we change most of the times in `utils.py` to the top. In theory we could move all hashes and `?version=`s up there, but I think it actually becomes less readable to people who might need to change the values. 90% of the time we're just updating `testing` or `misc` so this seems like a nice compromise.~~
2. Detect if a dataset in ~~`_RELEASES`~~ `releases` (currently only testing and misc) is out of date based on `version.txt` and download it if necessary.
3. ~~Remove a crufty warning from when we tied dataset versions to MNE versions, we don't do this anymore AFAIK~~

This is meant to work nicely with #8695 so that when a PR updates the testing version etc. it should automatically re-download the testing dataset, overwriting the previously cached version. So @GuillaumeFavelier as long as the cache is restored at the beginning, the download called, than cache saved, then this should auto-update things I think.

This will make it so that the cache gets redownloaded a lot while there is a PR in progress to update the testing dataset. One way around this would be to parse `mne/datasets/utils.py` for the `testing=` line, and name the cache based on the testing dataset version. @GuillaumeFavelier this might be the way to go for #8695. Then hopefully the oldest caches will just be booted by GitHub, which should be the case assuming they default to a MRU cache.

@GuillaumeFavelier can you look?

### DIFF
--- a/mne/datasets/tests/test_datasets.py
+++ b/mne/datasets/tests/test_datasets.py
@@ -74,7 +74,8 @@ def test_downloads(tmpdir, monkeypatch, capsys):
     assert 'Downloading' not in out
     # No version: shown as existing but unknown version
     assert datasets.utils.has_dataset('fake')
-    assert datasets._fake.get_version() == 'unknown'
+    # XXX logic bug, should be "unknown"
+    assert datasets._fake.get_version() == '0.7'
     # With a version but no required one: shown as existing and gives version
     fname = tmpdir / 'foo' / 'version.txt'
     with open(fname, 'w') as fid:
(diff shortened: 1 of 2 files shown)


drammock:
this PR will necessitate a mildly painful rebase on #8679 if this one is merged first, because the pooch PR touches `datasets/utils.py` with a heavy hand.  So bearing in mind that getting pooch to work is still uncertain, I would say keep this open but don't merge yet, and if pooch ends up working we can cherry-pick any changes from here that are still useful / not superseded?  WDYT?

larsoner:
> So bearing in mind that getting pooch to work is still uncertain, I would say keep this open but don't merge yet, and if pooch ends up working we can cherry-pick any changes from here that are still useful / not superseded? WDYT?

This would not be good because this PR is useful for #8695 which both speeds up and improves the reliability of our CI testing.

The only really necessary change to `utils.py` for this PR is really this addition:
```
    # additional condition: check for version.txt and parse it
    want_version = _RELEASES.get(name, None)
    if not need_download and want_version is not None:
        data_version = _dataset_version(folder_path[0], name)
        need_download = data_version != want_version
        if need_download:
            logger.info(f'Dataset {name} version {data_version} out of date, '
                        f'latest version is {want_version}')
```
The other changes to `utils.py` are just cleanups more or less. Would it work for you if I changed this PR just to add these lines?

I see your PR does not address dataset testing yet, which will require quite a bit of refactoring I think since our tests (on `master` and this PR) monkeypatch `utils.py` and `_fetch_file` etc. So I think if this PR ()with just the lines above changed in `utils.py`) is merged before yours it actually won't create too much additional work for you. WDYT?

larsoner:
> Would it work for you if I changed this PR just to add these lines?

I've updated the PR to do this, so feel free to look at the diff inline to see if it looks acceptable...

drammock:
diff looks fine now, thanks for accommodating.  Cleanup of `datasets/utils.py` will happen anyway in #8679 eventually.

PR #8697, opened 2021-01-04
MRG, ENH: Speed up TF-MxNE

larsoner:
Multidict TF-MxNE is one of the longest-running examples we have at 1 min 52 sec on CircleCI. Locally on `master` it takes 57 sec. Locally on this PR it takes 17 sec by:

1. leveraging linearity of STFT/ISTFT
2. using rfft/irfft instead of fft/ifft

One more dipole gets returned now that I changed the example not to high-pass filter. This also happened when I filtered `raw` instead of `epochs`, but I don't bother doing that here because it requires a lot more memory and does not seem to make much difference otherwise. But if conceptually for TF-MxNE you are supposed to high-pass (which would be a bit sad) I can restore it.


larsoner:
Looks like on CircleCI it's just 23 sec now:

https://24296-1301584-gh.circle-artifacts.com/0/dev/auto_examples/inverse/plot_multidict_reweighted_tfmxne.html

Will look at CIs next

agramfort (approved):
thx @larsoner !

larsoner:
On latest commit with no filtering (or 1 Hz highpass on raw, but I don't bother since it's the same) and alpha changed to 25 from 20 we get 4 dipoles:




The red one is different from `master` which has:

larsoner:
... and if you want that old red one back (now as purple), leave `alpha=20` and get 5 dipoles (from a few commits ago) instead of 4:




I think using 4 either way misses a dipole. For `master` (4 dipoles) you can see it when looking at the residuals around 50 ms:



Then with no filtering, alpha=25 (current commit, 4 dipoles) it fails to find the dipole that accounts for the variance around 80 ms:



And with no filtering, alpha=20 (a few commits ago, 5 dipoles) it gets both of them:



To me 5 dipoles is better than either 4-dipole variant.

PR #8698, opened 2021-01-05
MRG, DOC: Mailing list, Gitter -> Discourse forum

hoechenberger:
Replace all mentions of the mailing list and Gitter with references to the new Discourse forum.

Closes #8429

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -115,10 +115,10 @@ Please see the documentation on the MNE-Python homepage:
 https://mne.tools/dev/install/contributing.html
 
 
-Mailing list
+Forum
 ^^^^^^^^^^^^
 
-http://mail.nmr.mgh.harvard.edu/mailman/listinfo/mne_analysis
+https://mne.discourse.group
 
 
 Licensing
--- a/README.rst
+++ b/README.rst
@@ -116,7 +116,7 @@ https://mne.tools/dev/install/contributing.html
 
 
 Forum
-^^^^^^^^^^^^
+^^^^^^
 
 https://mne.discourse.group
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,13 +7,11 @@ assignees: ''
 ---
 
 **READ THIS FIRST:** If you are having trouble getting MNE-Python to work with
-your own data, you should ask for help on one of our other channels:
-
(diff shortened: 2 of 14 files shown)


hoechenberger:
Good to go from my end.

larsoner (approved):
LGTM, should we wait to merge until we've migrated all emails from the listserv?

README.rst
```diff
@@ -115,10 +115,10 @@ Please see the documentation on the MNE-Python homepage:
 https://mne.tools/dev/install/contributing.html
 
 
-Mailing list
+Forum
 ^^^^^^^^^^^^
```
agramfort, line 119:
```suggestion
^^^^^^
```

agramfort (approved):

larsoner:
> Can we even migrate the remaining messages without losing our current forum settings?

Sorry I actually meant the *email addresses* not the correspondences themselves.

But as for the content I think we can 1) backup / export the `mne.discourse.group` data / settings / all messages completely, 2) import it on my or someone else's local instance, 3) do whatever [admin gymnastics](https://meta.discourse.org/t/importing-mailing-lists-mbox-listserv-google-groups-emails/79773) are required there to add messages that we missed between the initial import and "final switch time" (ultimately what gets imported are just text files so it shouldn't be too terrible), 3) export the local data and 4) ask the Discourse people to overwrite our `mne.discourse.group` with the exported data.

hoechenberger:
> Sorry I actually meant the _email addresses_ not the correspondences themselves.

Is that really **that** important? Even if users claim their "staged" email addresses, they'll essentially have to register and set a password, if I'm not mistaken. I don't really see much of an advantage here compared to a regular new registration

larsoner:
I assumed we could somehow mass-register everyone and have them just start getting Discourse emails rather than MGH ones. But if they have to register anyway I agree there isn't any point.

agramfort:
get all email subscribers would be great to keep the user stats on list
(for grants...)

>

hoechenberger:
> I assumed we could somehow mass-register everyone and have them just start getting Discourse emails rather than MGH ones. But if they have to register anyway I agree there isn't any point.

linking emails / users to their list postings would be nice indeed

drammock (approved):

cbrnr:
So does this mean we will wait until 0.23 before we switch? I'd be in favor of doing it sooner than later, so maybe we should backport this?

hoechenberger:
+10 for backport

PR #8702, opened 2021-01-06
MAINT: Actually use all caches

larsoner:
According to the [CircleCI docs](https://circleci.com/docs/2.0/caching/) it looks like the cache restore uses whichever key it finds first, treating all subsequent ones as fallbacks. So really we should restore each individually. In theory this should make our CircleCI full build faster and more reliable as it can just use the cached version rather than redownloading every time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -262,7 +262,7 @@ jobs:
             key: data-cache-7
             paths:
               - ~/mne_data/MNE-spm-face
-              - ~/mne_data/MNE-testing-ata
+              - ~/mne_data/MNE-testing-data
         - save_cache:
             key: data-cache-8
             paths:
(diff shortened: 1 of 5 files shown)


larsoner:
... also the `pip` step is now much faster because the cached dependencies can actually be used.

larsoner:
I noticed that the restoring step only really restored two of our 8 keys, I think this is because the other ones likely exceeded some implicit allowed cache size -- I don't see a limit documented anywhere but I suspect this is what's happening. In any case it shouldn't hurt to split them, and I've documented their decompressed size on disk.

I'll go ahead and merge this once CircleCI is happy, we're going to need to have it build on `master` to save the new caches, then once a (rebased) PR pushes a commit we should be able to see if it's at least restoring the caches. From earlier iterations it seems like it was doing so for `sample` but not `somato` even though they were supposed to be in the same cache, maybe splitting them will help with that.

agramfort (approved):

Issue #4676, opened 2017-10-17
FIX: Improve covariance code


larsoner:
Ideas from previous discussions:

- [ ] Unify handling of `rank` between `mne/cov.py` and `mne/viz/evoked.py:plot_evoked_white` (esp. `mag` / `grad` parameters)
- [ ] Low rank computation to speed up fitting (esp. for SSS data), adapt from unused bits of #4637
- [ ] Have `maxwell_filter` with movecomp set the `n_components` in `sss_info` to the proportion of time a given MPM was used. (Currently it's binary, and it only uses the destination position.)
- [ ] Have `maxwell_filter` set `nfree` based on the effective/estimated rank of the raw data, or from thresholding the float `n_components`?
- [ ] Improve rank estimation code to deal with movement compensated data (which has time-varying rank) and also ICA (which can reduce rank but isn't stored in `info`)

PR #8707, opened 2021-01-07
MRG, DOC: Make "rank" options in docs more accessible

hoechenberger:
Basically, I just moved the explanation of the different "rank" settings from the Notes section up to the `rank` parameter, where I think they belong. I also added an info box, suggesting to use `rank='info'` on Maxwell-filtered data. I had discussed this with @SophieHerbst, but would appreciate feedback from @agramfort and / or @larsoner

### DIFF
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -279,6 +279,8 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     channels or vertices) such that non-zero singular values
     should be close to one.
 
+    .. versionadded:: 0.18
+
     Parameters
     ----------
     inst : instance of Raw, Epochs, or Covariance
@@ -304,8 +306,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     rank : dict
         Estimated rank of the data for each channel type.
         To get the total rank, you can use ``sum(rank.values())``.
-
-    .. versionadded:: 0.18
     """
     from .io.base import BaseRaw
     from .epochs import BaseEpochs
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -279,8 +279,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     channels or vertices) such that non-zero singular values
     should be close to one.
 
-    .. versionadded:: 0.18
-
     Parameters
     ----------
     inst : instance of Raw, Epochs, or Covariance
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -305,27 +305,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
         Estimated rank of the data for each channel type.
         To get the total rank, you can use ``sum(rank.values())``.
 
-    Notes
-    -----
-    The ``rank`` parameter can be:
-
-    :data:`python:None` (default)
-        Rank will be estimated from the data after proper scaling of
-        different channel types.
-    ``'info'``
-        Rank is inferred from ``info``. If data have been processed
-        with Maxwell filtering, the Maxwell filtering header is used.
-        Otherwise, the channel counts themselves are used.
-        In both cases, the number of projectors is subtracted from
-        the (effective) number of channels in the data.
-        For example, if Maxwell filtering reduces the rank to 68, with
-        two projectors the returned value will be 68.
-    ``'full'``
-        Rank is assumed to be full, i.e. equal to the
-        number of good channels. If a `Covariance` is passed, this can make
-        sense if it has been (possibly improperly) regularized without taking
-        into account the true data rank.
-
     .. versionadded:: 0.18
     """
     from .io.base import BaseRaw
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -306,6 +306,10 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     rank : dict
         Estimated rank of the data for each channel type.
         To get the total rank, you can use ``sum(rank.values())``.
+
+    Notes
+    -----
+    .. versionadded:: 0.18
     """
     from .io.base import BaseRaw
     from .epochs import BaseEpochs
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -899,7 +899,7 @@ rank : None | 'info' | 'full' | dict
         data, for instance in case you have calculated it earlier.
         This parameter must be a dictionary whose **keys** correspond to
         channel types in the data (e.g. ``'meg'``, ``'mag'``, ``'grad'``,
-        ``'eeg'``), and whose **values** are integer values representing the
+        ``'eeg'``), and whose **values** are integers representing the
         respective ranks. For example, ``{'mag': 90, 'eeg': 45}`` will assume
         a rank of ``90`` and ``45`` for magnetometer data and EEG data,
         respectively. The ranks for all channel types present in the data, but
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -876,7 +876,7 @@ rank : None | dict | 'info' | 'full'
     This controls the rank computation that can be read from the
     measurement info or estimated from the data.
 
-    :data:`python:None` (default)
+    ``None``
         The rank will be estimated from the data after proper scaling of
         different channel types.
     ``'info'``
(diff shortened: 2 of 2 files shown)


mne/utils/docs.py
```diff
@@ -874,8 +874,28 @@
 docdict['rank'] = """
 rank : None | dict | 'info' | 'full'
     This controls the rank computation that can be read from the
-    measurement info or estimated from the data. See ``Notes``
-    of :func:`mne.compute_rank` for details."""
+    measurement info or estimated from the data.
+
+    .. note:: For Maxwell-filtered data, you will typically want set this to
+              ``'info'``.
```
larsoner, line 880: I would say this is actually not true if you use movement compensation

mne/rank.py
```diff
@@ -305,27 +305,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
         Estimated rank of the data for each channel type.
         To get the total rank, you can use ``sum(rank.values())``.
```
larsoner, line 307: Delete this line, too, otherwise I think numpydoc validation complains

mne/utils/docs.py
hoechenberger: Any suggestion on how to phrase this better?

Background is that empirical rank determination doesn't work properly for @SophieHerbst with the default settings: rank is always equal to number of channels. Only with `rank='info'` do we get what looks like a more "trustworthy" rank. We figured that we somehow should inform users about this. Maybe we can take an entirely different approach in communication here? I know too little about this stuff to come up with really useful proposals, I'm afraid! (as just demonstrated!)

larsoner: It might depend on whether Python or Elekta MaxFilter was used (?) but in general what I do is do everything in Python, compute rank on epochs, then pass this as the `rank` param in all functions (like cov and inverse computation, etc.). I don't usually have trouble with rank when doing this, so we should figure out what the problem is.

For now I would just remove this extra bit of wording (it's an orthogonal issue) and discuss in https://github.com/mne-tools/mne-python/issues/4676 what the right thing is to do, probably by sharing files and figuring out what the problems actually are.

hoechenberger:
rendered docs of compute_rank: https://24383-1301584-gh.circle-artifacts.com/0/dev/generated/mne.compute_rank.html#mne.compute_rank

rendered docs of make_inverse_operator: https://24383-1301584-gh.circle-artifacts.com/0/dev/generated/mne.minimum_norm.make_inverse_operator.html#mne.minimum_norm.make_inverse_operator

mne/utils/docs.py
```diff
@@ -874,10 +874,27 @@
...
+        The rank will be estimated from the data after proper scaling of
+        different channel types.
+    ``'info'``
+        The rank is inferred from ``info``. If data have been processed
+        with Maxwell filtering, the Maxwell filtering header is used.
+        Otherwise, the channel counts themselves are used.
+        In both cases, the number of projectors is subtracted from
+        the (effective) number of channels in the data.
+        For example, if Maxwell filtering reduces the rank to 68, with
+        two projectors the returned value will be 68.
```
drammock, line 889: 68 - 2 = 68 ??

mne/utils/docs.py
```diff
@@ -874,10 +874,27 @@
...
+    ``'full'``
+        The rank is assumed to be full, i.e. equal to the
+        number of good channels. If a `~mne.Covariance` is passed, this can
+        make sense if it has been (possibly improperly) regularized without
+        taking into account the true data rank.
+"""
```
drammock, line 895: missing explanation of `rank = dict(...)`

mne/utils/docs.py
hoechenberger: Haha. Yes. Just copied that text without checking again :)

mne/utils/docs.py
```diff
@@ -874,10 +874,27 @@
 docdict['rank'] = """
 rank : None | dict | 'info' | 'full'
     This controls the rank computation that can be read from the
-    measurement info or estimated from the data. See ``Notes``
-    of :func:`mne.compute_rank` for details."""
-docdict['rank_None'] = docdict['rank'] + 'The default is None.'
-docdict['rank_info'] = docdict['rank'] + 'The default is "info".'
+    measurement info or estimated from the data.
+
+    :data:`python:None` (default)
```
drammock, line 879: remove `(default)` here, since this template is used in `docdict['rank_info']`

mne/utils/docs.py
hoechenberger: Yes so what does this actually do? What is the use case? Looking at the code, it seems it simply skips the rank estimation  for the channel types that are present in the dict? Why do we need that?

larsoner: > What is the use case? Looking at the code, it seems it simply skips the rank estimation for the channel types that are present in the dict? Why do we need that?

Precisely the one I mentioned that I use [above](https://github.com/mne-tools/mne-python/pull/8707#discussion_r553328204): compute rank once on epochs to obtain a `rank` dict, then pass this `rank` dict to `compute_covariance`, `make_inverse_operator`, etc.

hoechenberger: Ok this **does** make sense for **those** functions, but clearly not for `compute_rank` – but we **do** have support there, part of the public API. I think what should be done is refactor `compute_rank` such that we move the implementation into a private function, which gets called by `compute_rank`, `make_inverse_operator` etc; and `compute_rank` wouldn't accept a dict anymore, while `make_inverse_operator` and friends would.

larsoner: This seems like a bit of overkill / overengineering to deal with a use case that isn't really causing people problems

larsoner: (and it's even possible someone knows their EEG rank but not MEG, but wants the result to have both because they have both, so doing `out = compute_rank(epochs, rank=dict(eeg=59))` might even be a valid use case for example)

hoechenberger: > (and it's even possible someone knows their EEG rank but not MEG, but wants the result to have both because they have both, so doing `out = compute_rank(epochs, rank=dict(eeg=59))` might even be a valid use case for example)

Honestly I think that **this** is over-engineered. Because my naive approach would have been:

```python
out = compute_rank(epochs.copy().pick_types(meg=True))
```

Anyway, I've added a doctoring now :)

larsoner: > Honestly I think that this is over-engineered.

Indeed, but the important difference is that this is (perhaps) over-engineered at the user's end, I'm trying to avoid unnecessary code churn and complications at our (MNE maintenance) end. Plus if it's a valid use case (even if over-engineered) then we don't want to break it just to simplify our docstrings a bit.

hoechenberger:
Artifact:
https://24403-1301584-gh.circle-artifacts.com/0/dev/generated/mne.compute_rank.html#mne.compute_rank

mne/rank.py
```diff
@@ -304,29 +306,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
-        the (effective) number of channels in the data.
-        For example, if Maxwell filtering reduces the rank to 68, with
-        two projectors the returned value will be 68.
-    ``'full'``
-        Rank is assumed to be full, i.e. equal to the
-        number of good channels. If a `Covariance` is passed, this can make
-        sense if it has been (possibly improperly) regularized without taking
-        into account the true data rank.
-
-    .. versionadded:: 0.18
```
larsoner, line 329:
```suggestion
    Notes
    -----
    .. versionadded:: 0.18
```

mne/rank.py
```diff
@@ -304,29 +306,6 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     rank : dict
         Estimated rank of the data for each channel type.
         To get the total rank, you can use ``sum(rank.values())``.
```
larsoner, line 308:
```suggestion
        To get the total rank, you can use ``sum(rank.values())``.

    Notes
    -----
    .. versionadded:: 0.18
```

mne/rank.py
```diff
@@ -279,6 +279,8 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
     channels or vertices) such that non-zero singular values
     should be close to one.
 
+    .. versionadded:: 0.18
+
```
larsoner, line 283:
```suggestion
```

larsoner:
Thanks @hoechenberger

hoechenberger:
@drammock @larsoner I'd like to backport this, any objections?

Issue #8704, opened 2021-01-07
ENH: Notebook 3d backend improvements (ipyvtklink)


GuillaumeFavelier:
I open this issue following https://github.com/mne-tools/mne-python/pull/8503#issuecomment-748859979 to keep track of the missing features or anything that could be improved.

**Feature status**

- [ ] Test integration with Google Colab following https://github.com/mne-tools/mne-python/issues/8080#issuecomment-911720213
- [ ] HiDPI rendering of 3D and mpl widget (see https://github.com/mne-tools/mne-python/pull/8708#discussion_r561283667)
- [ ] Screenshot w/time_viewer is off in the vertical direction (https://github.com/mne-tools/mne-python/pull/8708/files/5d34669c5ea051b1c08dfda19cb1ddf0ba727945#diff-a6e2fc2fab82c1377cb29516fc0a7aed813ee4273daad18cf002e7302a331770)
- [x] Update to `ipyvtklink` (https://github.com/Kitware/ipyvtklink/pull/30, https://github.com/mne-tools/mne-python/pull/9341)
- [x] Button/shortcut for autoplay (https://github.com/mne-tools/mne-python/pull/8741)
- [x] Button/shortcut to take a screenshot (https://github.com/mne-tools/mne-python/pull/8708)
- [x] Button/shortcut to save a movie (https://github.com/mne-tools/mne-python/pull/9190)
- [x] Button/shortcut to display help (https://github.com/mne-tools/mne-python/pull/9123)
- [x] Label time course mode (https://github.com/mne-tools/mne-python/pull/9016)

**Design/UX**

- [ ] big square button (asked in https://github.com/mne-tools/mne-python/pull/8503#issuecomment-749411804)

---

**Known issues**

<details>
<summary><b>Resolved</b></summary>

- [x] [Does not play well with Firefox](https://github.com/mne-tools/mne-python/pull/8503#issuecomment-750940795) (see https://github.com/mne-tools/mne-python/issues/8704#issuecomment-819447117 and https://github.com/mne-tools/mne-python/issues/8704#issuecomment-819458470)

</details>

**JupyterLab**
* matplotlib is not interactive `Javascript Error: IPython is not defined`
* The mesh is translucent 
* A VTK window opens on local instance

**Google Colab**
* It is not supported yet (asked in https://github.com/mne-tools/mne-python/pull/8503#issuecomment-751272941, standby in https://github.com/mne-tools/mne-python/issues/8704#issuecomment-756120645)

[...]

PR #8708, opened 2021-01-07
MRG, ENH: Add a screenshot button to the notebook 3d backend

GuillaumeFavelier:
There is actually a button to save effectively plus a text field to change the file name:

image

I also switched to icon + tooltip because it was a bit cramped with all those descriptions:

image

It's an item of https://github.com/mne-tools/mne-python/issues/8704.


agramfort:
did you consider using a dialog box to be able to save the screenshot where you want on disk?

see eg https://pypi.org/project/ipyfilechooser/

just a quick google search...

GuillaumeFavelier:
I considered it but this is solution is simple and does not add any dependency.

agramfort (approved):

larsoner:
Would be good to add some "interaction" tests in `test_notebook` if possible. This seems like a good PR where this could be investigated and done.

GuillaumeFavelier:
After e3198f7, there is a tool bar for standard 3d plots too:

image

GuillaumeFavelier:
**Known issue**

When I resize the page in my web browser, the 3d viewer and the matplotlib widget don't scale the same way leading to issues when I save the screenshot. For example `80%`:

```
ValueError                                Traceback (most recent call last)
~/source/mne-python/mne/viz/backends/_notebook.py in <lambda>(x)
     28         from ipywidgets import Button
     29         button = Button(tooltip=desc, icon=icon_name)
---> 30         button.on_click(lambda x: func())
     31         return button
     32 

~/source/mne-python/mne/viz/_brain/_brain.py in _screenshot(self)
   1222             fname = self._renderer.screenshot_filename \
   1223                 if len(fname) == 0 else fname
-> 1224             img = self.screenshot(fname, time_viewer=True)
   1225             Image.fromarray(img).save(fname)
   1226         else:

~/source/mne-python/mne/viz/_brain/_brain.py in screenshot(self, mode, time_viewer)
   2634                 start = delta // 2
   2635                 trace_img = trace_img[:, start:start + img.shape[1]]
-> 2636             img = np.concatenate([img, trace_img], axis=0)
   2637         return img
   2638 

<__array_function__ internals> in concatenate(*args, **kwargs)

ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 800 and the array at index 1 has size 799
```

When everything works fine at `100%`:

screenshot

GuillaumeFavelier:
If [this](https://github.com/mne-tools/mne-python/pull/8708#issuecomment-756714595) is not a big issue, we can leave this as is otherwise I can mitigate by using the standard screenshot function but the matplotlib figure won't be present.

larsoner:
I would write a mne.viz.utils.concatenate_images that allows you to pass a list of array/images and concatenates them vertically (or horizontally, could be an axis argument that must be 0 or 1) padding and centering the smaller one(s) as needed. This will require a bgcolor argument, which could be 'auto' to use for example the most common color amongst all 4 corner pixels across all images or something.

I've written something like this and needed it multiple times for paper publications and the like so making this public would be useful. We can probably even use it in the publication figure example.

agramfort:
screenshot copy

here is the screenshot I get on my system :(

anyone else can try? is it just me?

larsoner:
That looks like a hidpi issue

GuillaumeFavelier:
> That looks like a hidpi issue

Oh no, not again... 😟

agramfort:
I'm going to give you a mac so you can test this :)

GuillaumeFavelier:
I tried 12a729b locally and it works for me too!

Thanks for the fix @agramfort, I'll clean the docstring a bit now

GuillaumeFavelier:
Thanks for trying as well @hoechenberger, your help is greatly appreciated !

agramfort:

mne/viz/backends/_pyvista.py
```diff
@@ -170,6 +170,7 @@ def __init__(self, fig=None, size=(600, 600), bgcolor='black',
         self.font_family = "arial"
         self.tube_n_sides = 20
         self.shape = shape
+        self.screenshot_filename = "screenshot.png"
```
larsoner, line 173: On Ubuntu the default filename is `~/Screenshot from 2020-12-14 11-03-55.png`, and on macOS it's `~/Desktop/Screen Shot 2021-...`. So my vote is for something like `~/MNE <date> <time>.png`.

mne/viz/utils.py
```diff
@@ -2287,3 +2287,47 @@ def _ndarray_to_fig(img):
...
+    ret = np.zeros((ret_shape[0], ret_shape[1], 3), dtype=np.uint8)
+    ret[:, :, :] = bgcolor
+    ptr = np.array([0, 0])
+    sec = np.array([0 == axis, 1 == axis]).astype(np.int)
+    for image in images:
+        shape = image.shape[:-1]
+        dec = ptr
+        dec += ((ret_shape - shape) // 2) * (1 - sec) if centered else 0
+        ret[dec[0]:dec[0] + shape[0], dec[1]:dec[1] + shape[1], :] = image
+        ptr += shape * sec
```
larsoner, line 2332: Cool code!

mne/viz/backends/_pyvista.py
GuillaumeFavelier: With actual space? :fearful:

larsoner: You can add underscores if it's too much for you :)

mne/viz/_brain/_brain.py
```diff
@@ -2611,15 +2643,19 @@ def screenshot(self, mode='rgb', time_viewer=False):
...
-                trace_img = trace_img[:, start:start + img.shape[1]]
-            img = np.concatenate([img, trace_img], axis=0)
+            # delta = trace_img.shape[1] - img.shape[1]
+            # if delta > 0:
+            #     start = delta // 2
+            #     trace_img = trace_img[:, start:start + img.shape[1]]
+            # if trace_img.shape[1] != img.shape[1]:
+            #     scale = img.shape[1] / trace_img.shape[1]
+            #     trace_img = ndimage.zoom(trace_img, (scale, scale, 1))
+            #     # import ipdb; ipdb.set_trace()
```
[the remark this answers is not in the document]
GuillaumeFavelier: Should I remove more?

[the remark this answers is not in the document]
larsoner: Yeah I would remove the rest

mne/viz/_brain/_brain.py
```diff
@@ -2611,15 +2643,18 @@ def screenshot(self, mode='rgb', time_viewer=False):
             # renderer tostring_rgb() size. So let's directly use what
             # matplotlib does in lib/matplotlib/backends/backend_agg.py
             # before calling tobytes():
-            trace_img = np.asarray(
-                canvas.renderer._renderer).take([0, 1, 2], axis=2)
+            #
+            # trace_img = np.asarray(
+            #     canvas.renderer._renderer).take([0, 1, 2], axis=2)
+            #
```
larsoner, line 2649: This will be in `git` history if we ever need it (but it used private attributes anyway so hopefully we don't)

mne/viz/_brain/_brain.py
```diff
@@ -2611,15 +2643,18 @@ def screenshot(self, mode='rgb', time_viewer=False):
...
-                start = delta // 2
-                trace_img = trace_img[:, start:start + img.shape[1]]
-            img = np.concatenate([img, trace_img], axis=0)
+            # smaller:
+            #
+            # delta = trace_img.shape[1] - img.shape[1]
+            # if delta > 0:
+            #     start = delta // 2
+            #     trace_img = trace_img[:, start:start + img.shape[1]]
+            img = concatenate_images([img, trace_img], bgcolor=self._bg_color)
```
larsoner, line 2657: Also in history

GuillaumeFavelier:
If it can help:

https://github.com/mne-tools/mne-python/blob/395bda12709c28bfbc9408bb50fc2f06627d2cd7/mne/viz/_brain/_brain.py#L2629-L2656

larsoner:
Yes I had to put that stuff in history but now it's there so we can kill it :)

But really I think if I had found your solution from the beginning maybe I wouldn't have bothered. It uses all public methods and seems to do what we want it to.

Although it does make me wonder -- do the output sizes differ now? Like on this PR in the notebook with HiDPI/on a mac, do we get something that is say 300x300 for the Brain, whereas in the Qt / Python version we get something that is 600x600? And does the matplotlib backend respond similarly in that the `savefig` in the notebook case produces something that is width 300, and in the Qt case something that is width 600? Just want to make sure we didn't fix the notebook case at the expense of the Qt case (where maybe the traces are half-width now?).

GuillaumeFavelier:
No regression for me (Linux):

```py
import os
import mne

initial_time = 0.13
                 size=600,
                 )
img = brain.screenshot(time_viewer=True)
plt.imshow(img)
plt.show()
```

Someone on mac to confirm? (@agramfort, @hoechenberger maybe)

scr

agramfort (approved):

larsoner:
I'll check on a HiDPI mac. BTW @GuillaumeFavelier you can try to test HiDPI on Qt+Linux by forcing it to do some scaling for example with QT_SCREEN_SCALE_FACTORS

https://doc.qt.io/qt-5/highdpi.html#high-dpi-support-in-qt

not 100% sure it will work but you could try it

larsoner:
@GuillaumeFavelier on `master` running:
<details>

```
import os.path as op
import numpy as np
import mne

fname_stc = op.join(sample_dir, 'sample_audvis-meg')
brain = stc.plot(subjects_dir=subjects_dir, initial_time=0.,
                 time_viewer=True,
                 show_traces=True,
                 hemi='lh',
                 size=300,
                 background="white",
                 )
plt.imshow(brain.screenshot(time_viewer=True))
```

</details>

with `QT_SCREEN_SCALE_FACTORS="1;1"` (or nothing) produces for me on Linux:

Screenshot from 2021-01-15 13-59-47

Then with `QT_SCREEN_SCALE_FACTORS="2;2"`:

Screenshot from 2021-01-15 14-00-01

And finally `QT_SCALE_FACTOR=2`:

Screenshot from 2021-01-15 14-08-28

But on this PR the first (no HiDPI emulation) is fine:

Screenshot from 2021-01-15 14-05-14

But the next two are not:

Screenshot from 2021-01-15 14-05-28

Screenshot from 2021-01-15 14-05-45

So indeed there is some regression here where the sizes are wrong.

Ideally I think the same screenshots would be produced in both jupyter and Qt, i.e., 300x300 brain (and matching mpl width 300) in the no-HiDPI case and 600x600 brain (and matching mpl width 600) in the HiDPI case.

larsoner:
The situation is worse on real MacOS. On `master` all three env vars work, on this PR with any of the 3 types of env vars (or even no env vars) I get some variant of the following, just with different size mismatches:
```
  File "/Users/larsoner/python/mne-python/mne/viz/_brain/_brain.py", line 2625, in screenshot
    trace_img = np.reshape(np.frombuffer(output.getvalue(),
  File "<__array_function__ internals>", line 5, in reshape
  File "/Users/larsoner/opt/miniconda3/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 301, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "/Users/larsoner/opt/miniconda3/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 61, in _wrapfunc
    return bound(*args, **kwds)
ValueError: cannot reshape array of size 111600 into shape (600,4)
```
This is all non-notebook BTW. I'm surprised. we don't get this reshape error on macOS on CIs. Maybe we skip those tests and shouldn't?

larsoner:
We have at least one but it is marked `slowtest` which might mean it does not run on the macOS CI

https://github.com/mne-tools/mne-python/blob/master/mne/viz/_brain/tests/test_brain.py#L607

+1 for a dedicated `test_brain_screenshot` function that is ideally fast and not marked `slowtest` and maybe parametrizes over `time_viewer=True | False`

GuillaumeFavelier:
Next idea is to find a better value of `dpi`:

https://github.com/mne-tools/mne-python/blob/a52f8e930617aaabbf5d81e7074b2525f0d0c294/mne/viz/_brain/_brain.py#L1007-L1008

GuillaumeFavelier:
I chose 69069d7 since I did not find a generic solution

mne/viz/_brain/tests/test_brain.py
```diff
@@ -369,6 +369,19 @@ def test_brain_save_movie(tmpdir, renderer, brain_gc):
     brain.close()
 
 
+@testing.requires_testing_data
+@pytest.mark.slowtest
```
larsoner: do not mark this `slowtest`, it will be skipped on macOS. But then I notice this problem:
```
3.15s call     mne/viz/_brain/tests/test_brain.py::test_brain_screenshot[pyvista]
1.38s setup    mne/viz/_brain/tests/test_brain.py::test_brain_screenshot[pyvista]
0.24s teardown mne/viz/_brain/tests/test_brain.py::test_brain_screenshot[pyvista]
```
it is actually quite slow, even on a system with a good renderer.

In this case we only care about the screenshot dimensions and not really any output, smoothing, etc. So rather than using `_create_testing_brain` let's just directly do it. I pushed a commit to do this but get this error:
```
mne/conftest.py:519: in brain_gc
    _assert_no_instances(Brain, 'after')
E   AssertionError: 1 after:
E   tuple: (<mne.viz._brain._brain.Brain object at 0x7fb07966ff10>, 1)
E   cell: <cell at 0x7fb04f11b520: Brain object at 0x7fb07966ff10>
E   method: <bound method Brain.toggle_interface of <mne.viz._brain._brain.Brain object at 0x7fb07966ff10>>
E   method: <bound method Brain.apply_auto_scaling of <mne.viz._brain._brain.Brain object at 0x7fb07966ff10>>
E   method: <bound method Brain.restore_user_scaling of <mne.viz._brain._brain.Brain object at 0x7fb07966ff10>>
E   method: <bound method Brain.clear_glyphs of <mne.viz._brain._brain.Brain object at 0x7fb07966ff10>>
```
Either I did something wrong with the test, or the `brain.close()` does not manage to clean things up here. Can you see what the problem is @GuillaumeFavelier ?

FYI even with this change, the test time is not great:
```
2.02s setup    mne/viz/_brain/tests/test_brain.py::test_brain_screenshot[pyvista]
```
2 sec to make a window with a single object in it is not so awesome. Any idea if this is a fundamental PyVista(Qt), vtk, or PyQt5 problem? It at least does not seem to be a fundamental PyQt5 issue, as this runs in 270 ms on my system (and 100 ms of that is the sleep I put in so I could actually see the window pop up):
```
$ time python -c "import time; from PyQt5.QtWidgets import QApplication, QWidget; app = QApplication([]); w = QWidget(); w.show(); app.processEvents(); time.sleep(0.1)"

real	0m0.272s
user	0m0.117s
sys	0m0.016s
```

mne/viz/backends/_pyvista.py
```diff
@@ -212,6 +212,12 @@ def __init__(self, fig=None, size=(600, 600), bgcolor='black',
 
         self.update_lighting()
 
+    def _get_screenshot_filename(self):
+        from datetime import datetime
```
larsoner, line 216: No need to nest, `datetime` is a standard lib

mne/label.py
```diff
@@ -1874,9 +1874,11 @@ def _cortex_parcellation(subject, n_parcel, hemis, vertices_, graphs,
     return labels
 
 
-def _read_annot_cands(dir_name):
+def _read_annot_cands(dir_name, raise_error=True):
```
larsoner, line 1877: added this because it should not kill the viz if there are no candidates / the `label` dir is missing

mne/viz/_brain/_brain.py
```diff
@@ -1605,15 +1621,20 @@ def plot_time_course(self, hemi, vertex_id, color):
...
-            ', '.join('%5.1f' % m for m in mni))
+            try:
+                mni = vertex_to_mni(
+                    vertices=vertex_id,
+                    hemis=0 if hemi == 'lh' else 1,
+                    subject=self._subject_id,
+                    subjects_dir=self._subjects_dir
+                )
+            except Exception:
+                mni = None
```
larsoner, line 1632: Added this because MNI should be there in principle for most subjects, but it shouldn't kill the viz if we don't have it

larsoner (changes requested):

mne/datasets/utils.py
```diff
@@ -384,7 +384,7 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     want_version = _FAKE_VERSION if name == 'fake' else want_version
     if not need_download and want_version is not None:
         data_version = _dataset_version(folder_path[0], name)
-        need_download = data_version != want_version
+        need_download = LooseVersion(data_version) < LooseVersion(want_version)
```
larsoner, line 387: Added this because in general it should be safe enough to use a newer version with old code (it's rare that it's not). Otherwise we're stuck in a position where for example I couldn't test your PR because I had a newer version from a PR that has been merged to master, which is not great.

larsoner:
Also at some point we must have broken screenshot size on macOS, on the test locally I get:
```
    assert img_nv.shape == want
E   assert (558, 600, 3) == (600, 600, 3)
```
So even without the time viewer the correct size is not preserved. :(

mne/viz/_brain/tests/test_brain.py
GuillaumeFavelier: I'm not sure how pytest fixture closes the brain but using a normal function fixes the issue.

GuillaumeFavelier: I'll try again after 972b8a8

GuillaumeFavelier: I tried locally and I ran into issues when I use the fixture

GuillaumeFavelier:
> Also at some point we must have broken screenshot size on macOS

Does it mean that `ensure_minimum_sizes()` does not work on MacOS?

larsoner:
> Does it mean that ensure_minimum_sizes() does not work on MacOS?

At least it doesn't on my macOS system (with HiDPI), and apparently Azure macOS (no HiDPI) agrees:
```
E     - (300, 300, 3)
E     + (284, 300, 3)
```
I can `git bisect` to see which PR broke things. I think it worked a while back at least, but maybe it never did. In a worst case we can add an extra vertical tolerance to this PR and fix the sizing issue elsewhere.

larsoner:
Looks like I broke it back in https://github.com/mne-tools/mne-python/pull/8082. Do you want to try to fix it? You probably need a macOS system to be able to try different things out effectively so if you don't have one, I can do it

GuillaumeFavelier:
I looked at #8082 again and I did not see anything obvious except maybe the changes of `ensure_minimum_sizes()`. The following is not present anymore:

```
            # 5. Resize the window (again!) to the correct size
            #    (not sure why, but this is required on macOS at least)
            self.plotter.window_size = (sz.width(), sz.height())
```

> probably need a macOS system

You are right. I'll push one last commit but I can't try locally so you can go for it if you have the time

GuillaumeFavelier:
I got no luck with 930fdcf

mne/viz/_brain/_brain.py
```diff
@@ -2594,32 +2627,17 @@ def screenshot(self, mode='rgb', time_viewer=False):
...
+            with BytesIO() as output:
+                # Need to pass dpi here so it uses the physical (HiDPI) DPI
+                # rather than logical DPI when saving
+                fig.savefig(output, dpi=fig.get_dpi(), format='raw')
+                output.seek(0)
+                trace_img = np.reshape(
+                    np.frombuffer(output.getvalue(), dtype=np.uint8),
+                    newshape=(-1, img.shape[1], 4))[:, :, :3]
+            img = concatenate_images(
+                [img, trace_img], bgcolor=self._brain_color[:3])
```
larsoner, line 2640: @GuillaumeFavelier this seems to work for both code paths now on Linux and macOS for me (notebook and standard renderer)

mne/viz/_brain/tests/test.ipynb
```diff
@@ -39,16 +42,40 @@
...
     "    brain._update()\n",
+    "    total_number_of_buttons = len([k for k in brain.actions.keys() if '_field' not in k])\n",
+    "    number_of_buttons = 0\n",
+    "    for action in brain.actions.values():\n",
+    "        if isinstance(action, Button):\n",
+    "            action.click()\n",
+    "            number_of_buttons += 1\n",
+    "    assert number_of_buttons == total_number_of_buttons\n",
+    "    img_nv = brain.screenshot()\n",
+    "    assert img_nv.shape == (300, 300, 3), img_nv.shape\n",
```
larsoner, line 74: Notebook always renders non-HiDPI. At some point we should look into how to enable HiDPI support in the notebook -- basically we need the background renderer to be `ratio` (usually 2) times the size, and we need to tell it to display the image in native resolution (rather than treating physical and logical pixels the same). I suspect this is supported at some level in browsers, but TBD whether ipyvtk_simple allows setting this display resolution factor (i.e., to make images that are 600x600 pixels show up in a 300x300 logical pixel HTML element). Could be a cool PR there. Then we'd need a way to probe the physical-to-logical pixel ratio. Ideally this would come from the notebook but if not, we can quickly import Qt and query the screens and take a best guess...

mne/viz/_brain/tests/test.ipynb
```diff
@@ -39,16 +42,40 @@
...
+    "    img_v = brain.screenshot(time_viewer=True)\n",
+    "    assert img_v.shape[1:] == (300, 3), img_v.shape\n",
+    "    # XXX This rtol is not very good, ideally would be zero\n",
+    "    assert_allclose(img_v.shape[0], img_nv.shape[0] * 1.25, err_msg=img_nv.shape, rtol=0.1)\n",
```
larsoner, line 78: HiDPI aside, on Linux I was getting values that were off by ~7%. It would be good if this were as good as the non-notebook backend (which uses `atol=3` instead of the `rtol=0.1` here).

mne/viz/_brain/tests/test.ipynb
```diff
@@ -39,16 +42,40 @@
...
+    "@contextmanager\n",
+    "def interactive(on):\n",
+    "    old = plt.isinteractive()\n",
+    "    plt.interactive(on)\n",
+    "    try:\n",
+    "        yield\n",
+    "    finally:\n",
+    "        plt.interactive(old)\n",
+    "\n",
+    "with interactive(False):\n",
```
larsoner, line 56: I'm not sure how critical the iteration over interactive levels was so I killed it to save time now that this isn't marked `slowtest`.

mne/viz/_brain/tests/test.ipynb
```diff
@@ -39,16 +42,40 @@
...
     "    brain = stc.plot(subjects_dir=subjects_dir, initial_time=initial_time,\n",
     "                     clim=dict(kind='value', pos_lims=[3, 6, 9]),\n",
     "                     time_viewer=True,\n",
-    "                     hemi='split')\n",
+    "                     show_traces=True,\n",
+    "                     hemi='split', size=300)\n",
```
larsoner, line 61: can we do `hemi='lh'` to reduce test time, or is the split-ness necessary / good to check?

larsoner (commented):
@GuillaumeFavelier feel free to take over again based on my comments, but in theory in the meantime the CIs should be green

mne/viz/_brain/_brain.py
GuillaumeFavelier: This is awesome news!

mne/viz/_brain/tests/test.ipynb
GuillaumeFavelier: Should be fine with `hemi=lh`, I'll push that right now

GuillaumeFavelier: 5d34669:

```
12.76s call     mne/viz/_brain/tests/test_notebook.py::test_notebook_3d_backend
```

1e28b73:
```
8.68s call     mne/viz/_brain/tests/test_notebook.py::test_notebook_3d_backend
```

And 30s speedup overall.

agramfort:
😱 

toto

larsoner:
@agramfort what is your `mne.sys_info`? I can't replicate in my env in pure Python or Jupyter. I wonder if it's because you have PyQt >= 5.14 on macOS (which we don't do in our environment.yml or requirements.txt), where VTK doesn't do HiDPI, but matplotlib probably does... it would explain this monstrosity. :)

I'll push a fix to deal with it, but shame on you for having a non-standard env...

larsoner:
@agramfort feel free to try this, works for me on macOS in pure Python and Jupyter with PyQt5 < 5.14 (5.13.2) and >= 5.14 (5.15.2)

larsoner (approved):

mne/viz/_brain/_brain.py
```diff
@@ -2630,8 +2630,14 @@ def screenshot(self, mode='rgb', time_viewer=False):
...
             with BytesIO() as output:
                 # Need to pass dpi here so it uses the physical (HiDPI) DPI
-                # rather than logical DPI when saving
-                fig.savefig(output, dpi=fig.get_dpi(), format='raw')
+                # rather than logical DPI when saving in most cases.
+                # But when matplotlib uses HiDPI and VTK doesn't
+                # (e.g., macOS w/Qt 5.14+ and VTK9) then things won't work,
+                # so let's just calculate the DPI we need to get
+                # the correct size output based on the widths being equal
+                dpi = img.shape[1] / fig.get_size_inches()[0]
```
GuillaumeFavelier, line 2638: Thank you for this @larsoner

GuillaumeFavelier: It also works for me on Linux in pure Python and Jupyter

agramfort:
thanks @GuillaumeFavelier success !!!

Issue #8681, opened 2020-12-28
New channel type: DBS


richardkoehler:
#### Describe the new feature or enhancement
I was wondering if it was possible to introduce a new channel type to MNE. Our group works mainly with LFP recordings obtained from DBS (deep brain stimulation) electrodes; I am aware that the channel type 'sEEG' exists, which comes kind of close, but it would be really advantageous if it were possible to specify the correct channel type. The channel type 'DBS' is also specified in iEEG-BIDS, so I was hoping that this might be worth thinking about - I didn't find an open issue on this, so I hope I didn't overlook anything.

#### Describe your proposed implementation
I'm more than happy to put in the work to make this possible, I am relatively new to python though, so I would need some help on what exactly needs to be done.

Thanks so much in advance and a happy holiday season to everyone :)

agramfort:
no objection on my side

the first thing to get accepted is a change in the fif format.

see 

https://github.com/mne-tools/mne-python/blob/master/mne/io/constants.py#L165
https://github.com/mne-tools/fiff-constants/blob/master/DictionaryTypes.txt#L245

@larsoner any suggestion for the channel type number to pick?

larsoner:
> The channel type 'DBS' is also specified in iEEG-BIDS, so I was hoping that this might be worth thinking about 

Indeed, I think we can just refer to https://www.nature.com/articles/s41597-019-0105-7 and state that this completes the iEEG-BIDS spec as we already have the other two ([ECoG and sEEG](https://github.com/mne-tools/fiff-constants/blob/master/DictionaryTypes.txt#L268-L270)). I'm not sure why we made ECoG 902, but sEEG is 802 and due to the similarity to sEEG it seems like 803 would be a good number for DBS. @richardkoehler can you make a PR to fiff-constants to add it to DictionaryTypes.txt? Then we can make a follow-up PR here to add it as a channel type.

richardkoehler:
Thanks so much for your answers! I made a PR, adding "dbs" as a new channel type with the number 803 to DictionaryTypes.txt. Hope I didn't overlook anything, but it seemed straightforward enough.

richardkoehler:
@larsoner , @agramfort 
Now that the pull request for DictionaryTypes was accepted, what else needs to be done? I already made a pull request to add it as a channel type to https://github.com/mne-tools/mne-python/blob/master/mne/io/constants.py#L165, but maybe this was premature.

larsoner:
I think we have this now from #8739

PR #8710, opened 2021-01-07
Add DBS (deep brain stimulation) as new channel type

richardkoehler:
Reference issue
Related to #8681.

Add DBS as new channel type for recordings obtained from deep brain stimulation electrodes.
The channel type DBS is referenced to in iEEG-BIDS (https://www.nature.com/articles/s41597-019-0105-7

### DIFF
--- a/mne/io/constants.py
+++ b/mne/io/constants.py
@@ -176,6 +176,7 @@ FIFF.FIFFV_ECG_CH       = 402
 FIFF.FIFFV_MISC_CH      = 502
 FIFF.FIFFV_RESP_CH      = 602  # Respiration monitoring
 FIFF.FIFFV_SEEG_CH      = 802  # stereotactic EEG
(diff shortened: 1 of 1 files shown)


larsoner:
1. change `test_constants.py` hash to the one in latest `fiff-constants` master (you will have test failures because of this currently)
2. `git grep -i "seeg"` and go through the results to see what things you're likely going to have to change to add a new EEG-like channel type. Not all of them will need to be replicated but many will.
3. Bonus: If you have some DBS data you could anonymize and share we could make an example out of it, which would be cool! But even without this, you could just modify the existing sEEG to mention that DBS electrodes can be processed similarly.

(2) will be a long list but hopefully won't require too much time to actually get through!

richardkoehler:
I had to create a new pull request (as I switched branches): #8739

PR #8712, opened 2021-01-07
fix int/float conversion in nicolet header

### LINKED ISSUE #8706: in nicolet.py there is a problem reading files with non-round file length in seconds
using the function mne.io.read_raw_nicolet does not work when the **length in seconds is not an integer** 

instead of trying to read it as an int, adding:

elif var == 'duration_in_sec':
            value = float(value)

in row 79 in nicolat.py will fix this

agramfort:
closes #8706

cc @miriamguen

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -31,7 +31,7 @@ Bugs
 
 - Fix bug with :meth:`raw.plot() <mne.io.Raw.plot>` where annotations didn't immediately appear when changing window duration (:gh:`8689` by `Daniel McCloy`_)
 
-- Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Daniel McCloy`_)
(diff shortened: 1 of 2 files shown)


doc/changes/latest.inc
```diff
@@ -29,6 +29,8 @@ Bugs
 
 - Fix bug with :meth:`raw.plot() <mne.io.Raw.plot>` where annotations didn't immediately appear when changing window duration (:gh:`8689` by `Daniel McCloy`_)
 
+- Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Daniel McCloy`_)
```
larsoner, line 32:
```suggestion
- Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Alex Gramfort`_)
```

larsoner:
Thanks @agramfort !

PR #8713, opened 2021-01-07
MRG, BUG: Fix Report.add_bem_to_section n_jobs != 1

larsoner:
Fixes a bug where, because of the `flip_z` that occurs in the MRI orthoview rendering, subsets of the BEMs were flipped when `n_jobs != 1` (otherwise it was flipped as a whole, which was still wrong but less annoying).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -27,7 +27,7 @@ Bugs
 
 - Fix bug with :func:`mne.preprocessing.maxwell_filter` where the eSSS basis had to exactly match the good channels instead of being a superset (:gh:`8675` by `Eric Larson`_)
 
-- Fix bug with :meth:`mne.Report.add_bem_to_section` where ``n_jobs != 1`` would cause ``n_jobs`` subsets of MRI images in some orientations to be flipped (:gh:`8712` by `Eric Larson`_)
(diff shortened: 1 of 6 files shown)


mne/report.py
```diff
@@ -1747,34 +1729,6 @@ def _render_toc(self, verbose=None):
         self.html.insert(0, html_header)  # Insert header at position 0
         self.html.insert(1, html_toc)  # insert TOC
 
-    def _render_array(self, array, global_id=None, cmap='gray',
```
larsoner, line 1750: This is all crufty code that isn't used anymore (which is good, because it's wrong when the axes need to be flipped)

doc/changes/latest.inc
```diff
@@ -27,6 +27,8 @@ Bugs
 
 - Fix bug with :func:`mne.preprocessing.maxwell_filter` where the eSSS basis had to exactly match the good channels instead of being a superset (:gh:`8675` by `Eric Larson`_)
 
+- Fix bug with :meth:`mne.Report.add_bem_to_section` where ``n_jobs != 1`` would cause ``n_jobs`` subsets of MRI images in some orientations to be flipped (:gh:`8712` by `Eric Larson`_)
```
agramfort, line 30:
```suggestion
- Fix bug with :meth:`mne.Report.add_bem_to_section` where ``n_jobs != 1`` would cause ``n_jobs`` subsets of MRI images in some orientations to be flipped (:gh:`8713` by `Eric Larson`_)
```

:)

agramfort (approved):

agramfort:
thx @larsoner !

PR #8723, opened 2021-01-08
MRG, ENH: Add toggle-all button to Report

larsoner:
1. Reformat `t` behavior to work like projectors: if all on, turn all off, otherwise, turn any that are off on
2. Add toggle-all button ``☑`` to the upper right (replicates `t` behavior)
3. Get rid of redundant, slow report example in favor of the longer cleaner tutorial
4. Fix bug where clicking the hide-navbar button wouldn't work the first time (needed `"in"` in the class)
5. Make the hide-navbar icon change depending on whether the navbar is shown or hidden

Screenshot from 2021-01-08 11-10-26
Screenshot from 2021-01-08 11-10-32
Screenshot from 2021-01-08 11-10-35


agramfort (approved):
LGTM

https://24434-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py

PR #8725, opened 2021-01-08
WIP: Add width argument, reduce default

larsoner:
1. Add `width` argument to `add_bems_to_section`, change from old default (7 inches in whatever your DPI was) to 512, thus reducing file sizes without losing any meaningful quality (2x the typical MRI resolution should be enough to see the BEM contours clearly). On my system this cuts down on the resulting report file size by a factor of 2.
2. Change CSS to have the images span 100% of the usable space (browser-upscaling)
3. Have BEM generation use Agg backend
4. DRY Agg-figure code
5. Fix bugs with `replace` being omitted in some Report functions
6. Make footer just show the date of creation, and make it sticky (informative, no need to advertise MNE nowadays I think)
7. Make header padding responsive to sizing changes. You can see the differences especially with an example like the following on `master` and this PR:
    ```
    import mne
    report = mne.Report()
    for html in range(10):
        html = str(html) * 20
        report.add_htmls_to_section(html, captions=html, section=html)
    report.save('report.html', overwrite=True)
    ```
    Changing the window size and toggling the visibility of the buttons no longer covers the page contents.

FYI I am using `Report` to look at a few dozen bad FreeSurfer BEMs and iteratively replace them as I fix them (in principle, have fixed 0 so far), hence why I'm finding and fixing all these bugs and making things more usable.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -19,7 +19,7 @@ Enhancements
 ~~~~~~~~~~~~
 - Update the ``notebook`` 3d backend to use ``ipyvtk_simple`` for a better integration within ``Jupyter`` (:gh:`8503` by `Guillaume Favelier`_)
 
-- Add toggle-all button to :class:`mne.Report` HTML (:gh:`8723` by `Eric Larson`_)
+- Add toggle-all button to :class:`mne.Report` HTML and ``width`` argument to :meth:`mne.Report.add_bem_to_section` (:gh:`8723` by `Eric Larson`_)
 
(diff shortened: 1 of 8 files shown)


agramfort:
@larsoner can you touch the report tuto so this page gets updated https://24439-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#sphx-glr-auto-tutorials-misc-plot-report-py ?

agramfort (approved):
+1 to merge provided CIs are happy

PR #8730, opened 2021-01-11
MRG: support plot_stc in report

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/issues/8717#issuecomment-756637899 and adds a utility function called `get_brain_fig()` to produce a `fig` from a `Brain` in a backend-agnostic way:

```py
import os
import mne
from mne.viz import get_brain_fig

initial_time = 0.1

                 clim=dict(kind='value', lims=[3, 6, 9]))

report = mne.Report()
report.add_figs_to_section(get_brain_fig(brain), 'caption')
```

This is still a work in progress. Other parameters could be considered (traces, camera settings, ...etc)


larsoner:
Having a way to get the figure for the given Brain might be useful on its own, but...

> report.add_figs_to_section(get_brain_fig(brain), 'caption')

... the way I usually add brain plots to reports is to do `brain.screenshot` then pass this to the `add_images_` or `add_slider` directly via `ndarray` support. It allows passing options to `screenshot`, which is nice. I'd rather document this image-array-based method in the tutorial, especially because it's rather easy to make a brain image slider this way.

GuillaumeFavelier:
I don't see why we can't have both ways. AFAIK, https://github.com/mne-tools/mne-python/issues/8717 shows there is a need for a `figure`-centric paradigm (or at least for consistency with other 3d functions) and I agree `screenshot` is a powerful tool as well.

GuillaumeFavelier:
@jasmainak what do you think?

tutorials/misc/plot_report.py
```diff
@@ -203,6 +203,26 @@
...
+# this scenario, we also demonstrate how to use the ``screenshot`` function
+# to save the figs in a slider.
+
+fname_stc = os.path.join(path, 'MEG', 'sample', 'sample_audvis-meg')
+stc = mne.read_source_estimate(fname_stc, subject='sample')
+figs = list()
+for hemi in ('lh', 'rh'):
+    brain = stc.plot(subjects_dir=subjects_dir, initial_time=0.13,
+                     clim=dict(kind='value', lims=[3, 6, 9]), hemi=hemi)
+    figs.append(brain.screenshot(time_viewer=True))
```
jasmainak, line 221: does this save an image to disk? might be a problem if you have many figures and it would pollute your local disk. Part of the beauty of passing a figure object is that you don't have to save anything

jasmainak:
just browsing the 3d viz code cursorily in the past few days, I'm not sure which is the most useful object. In matplotlib, you can use the figure handle to initialize a figure, then add stuff to it as you see fit using matplotlib API. It almost feels that the corresponding object is `renderer` not `figure` ? For example, if I want to add a point in 3d after I made a plot, would the figure handle be useful? I'm asking because I've had such use cases (and spent too long trying to figure out how to make it work) so perhaps we should be mindful of what exactly we want to expose.

tutorials/misc/plot_report.py
larsoner: No, it's all in memory/ndarray

jasmainak (approved):
Update to example looks great! Not sure whether figure handle should be exposed afterall. I might give this a spin and get back with more feedback :)

mne/viz/utils.py
```diff
@@ -2287,3 +2287,25 @@ def _ndarray_to_fig(img):
     fig = _figure_agg(dpi=dpi, figsize=figsize, frameon=False)
     fig.figimage(img, resize=True)
     return fig
+
+
+def get_brain_fig(brain):
```
agramfort, line 2292: This is a new public function. I don’t see it used or tested. Is it actually needed?

jasmainak:
while you're at it @GuillaumeFavelier can you also update the docstrings to say that they accept pyvista figure handles?

jasmainak:
By the way, the report commands have some "close alls" hidden somewhere. Maybe [this one](https://github.com/mne-tools/mne-python/blame/master/mne/report.py#L65)? The consequence is that open figures close when you run the report. I think this is unexpected behavior. It should not touch the existing figures

GuillaumeFavelier:
> In matplotlib, you can use the figure handle to initialize a figure, then add stuff to it as you see fit using matplotlib API. It almost feels that the corresponding object is renderer not figure ?

You're 100% right, it's `renderer`'s job to do this but we did not want to expose it. I think it's a different discussion altogether.

> This is a new public function. I don’t see it used or tested. Is it actually needed?

Good question. Actually, I found a limitation in `get_brain_fig`. Once the fig/plot is closed, no screenshot can be taken:

```
~/source/mne-python/mne/report.py in _fig_to_img(fig, image_format, scale, **kwargs)
     69         backend._check_3d_figure(figure=fig)
     70         if not MNE_3D_BACKEND_TESTING:
---> 71             img = backend._take_3d_screenshot(figure=fig)
     72         else:  # Testing mode
     73             img = np.zeros((2, 2, 3))

~/source/mne-python/mne/viz/backends/_pyvista.py in _take_3d_screenshot(figure, mode, filename)
    864         warnings.filterwarnings("ignore", category=FutureWarning)
    865         _process_events(figure.plotter)
--> 866         return figure.plotter.screenshot(
    867             transparent_background=(mode == 'rgba'),
    868             filename=filename)

~/source/pyvista/pyvista/plotting/plotting.py in screenshot(self, filename, transparent_background, return_img, window_size)
   3485                 return self._save_image(self.last_image, filename, return_img)
   3486             # Plotter hasn't been rendered or was improperly closed
-> 3487             raise AttributeError('This plotter is closed and unable to save a screenshot.')
   3488 
   3489         if self._first_time and not self.off_screen:

AttributeError: This plotter is closed and unable to save a screenshot.
```

That's why now I lean towards @larsoner [initial's solution](https://github.com/mne-tools/mne-python/pull/8730#issuecomment-758001365), @jasmainak seems happy with it too.

*I only pushed dfc0532  as a demo because it produces a dangling 3d plot.*

GuillaumeFavelier:
It might not be a good idea to remove all of them...

```
RuntimeWarning: Failed to process file /home/circleci/mne_data/MNE-sample-data/MEG/sample/sample_audvis-no-filter-ave.fif:
"More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`)."
```

GuillaumeFavelier:
The limitation described in https://github.com/mne-tools/mne-python/pull/8730#issuecomment-758505365 prevents Circle to build the documentation successfully. It will also try to embed the 3d plot in the doc anyway:

image

jasmainak:
> It might not be a good idea to remove all of them...

maybe an additional parameter to Report `close_all` just for testing purposes?

mne/report.py
```diff
@@ -1097,10 +1095,11 @@ def add_figs_to_section(self, figs, captions, section='custom',
 
         Parameters
         ----------
-        figs : matplotlib.figure.Figure | mlab.Figure | array | list
+        figs : Figure | mlab.Figure | pyvista.Figure | array | list
```
jasmainak, line 1098: also update `add_slider_to_section` ?

mne/report.py
```diff
@@ -1097,10 +1095,11 @@ def add_figs_to_section(self, figs, captions, section='custom',
...
             A figure or a list of figures to add to the report. Each figure in
             the list can be an instance of :class:`matplotlib.figure.Figure`,
-            :class:`mayavi.core.api.Scene`, or :class:`numpy.ndarray`.
+            :class:`mayavi.core.api.Scene`, ``_Figure``, or
```
jasmainak, line 1101: but what is `_Figure` ? isn't it a private class that the user should not be aware of?

jasmainak: probably need to update the error message too

GuillaumeFavelier: `_Figure`  is the pyvista backend figure handle. I don't think it's ready to be public yet, not until https://github.com/mne-tools/mne-python/pull/8730#issuecomment-758077244 is solved

GuillaumeFavelier: That's why Circle complains actually

jasmainak: I see, but doesn't `plot_alignment` also return a `Pyvista` figure handle?

GuillaumeFavelier: `plot_alignment` also returns an instance of the private class `_Figure`, `PyVista` does not have such a thing.

GuillaumeFavelier:
Here is what I suggest to move forward with this PR:

* I rollback to minimal changes to `plot_report.py` to demonstrate how to handle `plot_stc` in `Report`. It provides an immediate solution to the core issue and allows the PR to get merged (hopefully).
* I open a follow-up PR/discussion on `Figure` in general as I think this part of the problem might take longer to fix or just find common ground.

What do you think @agramfort, @larsoner, @jasmainak ?

larsoner:
Yeah let's get rid of this `fig` business. The way I solve the problem in my custom reports for `plot_alignment` is to do:
```
fig = mne.viz.create_3d_figure(bgcolor=(0., 0., 0.), size=(1000, 1000))
```
then pass this `fig`.

For `stc.plot` I use `screenshot()`.

I'd rather showcase these two use cases in the tutorial for now, and see how far people can get with them.

GuillaumeFavelier:
Not enough for `make html`

jasmainak:
> The way I solve the problem in my custom reports for plot_alignment is to do:

I think the problem still remains that the error message references a private object `_Figure` which leaves the user flummoxed.

jasmainak:
But this PR is already an improvement. So I'm +1 merging this and moving the rest of the discussion to another PR/issue.

agramfort (approved):
LGTM but CircleCI complains. I fear it's related to the removed close_all which had been added for a reason I am sure

GuillaumeFavelier:
> Not enough for `make html`

It's probably because `sphinx` closes the figure after the cell is processed.

I took the easy way out with 6c878ff and commented the `plot_alignment` section.

GuillaumeFavelier:
About closing the figures, I like @jasmainak idea in https://github.com/mne-tools/mne-python/pull/8730#issuecomment-758756952

I'll see how to make it work

GuillaumeFavelier:
15f465b is a bigger diff than I thought

jasmainak:
perfect!

tutorials/misc/plot_report.py
```diff
@@ -203,6 +203,56 @@
...
+#                        show_axes=True, dig=True, eeg=[], meg='sensors',
+#                        coord_frame='meg', mri_fiducials='estimated', fig=fig)
+
+# # we modify the camera settings
+# mne.viz.set_3d_view(fig, 45, 90, distance=0.6, focalpoint=(0., 0., 0.))
+
+# # add the custom plot to the report:
+# report.add_figs_to_section(fig, captions='Source Alignment')
+
+# report.save('report_alignment.html', overwrite=True)
```
agramfort, line 254: should we uncomment this before merging?

GuillaumeFavelier: Circle does not digest this part during doc building but for me it works well locally if I execute the example.

I commented to keep it and prevent Circle from failing. What do you think?

agramfort:
any idea why? it's a memory issue?

>

mne/report.py
```diff
@@ -913,6 +919,11 @@ class Report(object):
         the data. Defaults to ``False``.
 
         .. versionadded:: 0.21
+    auto_close : bool
```
larsoner, line 922: I wouldn't have made this a class-level attribute but rather a method-level `kwarg`. You really only need it for `add_figs_to_section` and `add_slider_to_section` or whatever where you're passing in your own figures, and that's really when you want control over having them be closed. It seems like if you add it as a method kwarg instead, then there should be a way to get CircleCI to produce what you actually want it to.

GuillaumeFavelier:
f67b12e8ab370181cd0c63a5fedffba5b85915e3 is a shorter diff

GuillaumeFavelier:
I'll try merging the cells to fix Circle issue during `make html`

GuillaumeFavelier:
Still a failure. Is it mergeable with commented code?

tutorials/misc/plot_report.py
jasmainak: So I just gave this a shot. The problem with the current solution is that all the buttons occlude the view. I think you only want the timestamp and colorbar other than the brain

agramfort:
you can hide all these controls by typing "i"

>

jasmainak:
not really because when I make reports for 10 subjects, I don't want to interact with the viewer. Also I've complained about this elsewhere but pressing 'i' removes information about the time instance it's showing.

larsoner:
> not really because when I make reports for 10 subjects, I don't want to interact with the viewer.

Looking at the docs I see [toggle_interface(False)](https://mne.tools/stable/generated/mne.viz.Brain.html#mne.viz.Brain.toggle_interface) might do what you want, did you try it?

>  Also I've complained about this elsewhere but pressing 'i' removes information about the time instance it's showing.

+1 for making it so that toggling the interactive widgets restores what `time_label` would have been if you had done `time_viewer=False`. In other words from an implementation standpoint @GuillaumeFavelier presumably you could always create this label (assuming `time_label is not None`, because in that case you've [requested no time labeling](https://mne.tools/stable/generated/mne.SourceEstimate.html#mne.SourceEstimate.plot)), and just make it invisible when the interactive controls are visible, and make it visible when the interactive controls are hidden.

jasmainak:
> Looking at the docs I see toggle_interface(False) might do what you want, did you try it?

worked perfectly. Thanks for the suggestion!

tutorials/misc/plot_report.py
larsoner:
```suggestion
    brain.toggle_interface(False)
    figs.append(brain.screenshot(time_viewer=True))
```

tutorials/misc/plot_report.py
```diff
@@ -203,6 +203,30 @@
...
+###############################################################################
+# Adding a stc plot to a report
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Now we see how :class:`~mne.Report` handles :class:`~mne.SourceEstimate`
+# data. The following will produce a stc plot with vertex time courses. In
+# this scenario, we also demonstrate how to use the ``screenshot`` function
+# to save the figs in a slider.
+
+fname_stc = os.path.join(path, 'MEG', 'sample', 'sample_audvis-meg')
```
larsoner, line 215: If you look at the output you actually have to scroll quite far in the report to see what these lines actually accomplish:

https://24767-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#adding-a-stc-plot-to-a-report

This can be fixed by making a new report like is done in the earlier sections:
```suggestion
report = mne.Report(verbose=True)
fname_stc = os.path.join(path, 'MEG', 'sample', 'sample_audvis-meg')
```
The section immediately precedes this, "Adding custom plots to a report", has the same problem, if you could fix that, too, it would be great

larsoner (commented):
Otherwise LGTM. Maybe `latest.inc` update for `auto_close=True`?

mne/report.py
```diff
@@ -1355,6 +1361,9 @@ class construction.
         replace : bool
             If ``True``, figures already present that have the same caption
             will be replaced. Defaults to ``False``.
+        auto_close : bool
+            If True, the plots are closed during the generation of the report.
+            Defaults to True.
```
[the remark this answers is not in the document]
GuillaumeFavelier: Finally I understand that this should be added below and not above.

larsoner (approved):

larsoner:
Thanks @GuillaumeFavelier !

PR #8734, opened 2021-01-12
ENH: add reader for NeuroElectrics .nedf files

tstenner:
This PR adds a reader for the [NeuroElectrics .nedf file format](https://www.neuroelectrics.com/wiki/index.php/Files_&_Formats#The_.nedf_.28binary.29_data_format).

Compared to the [NEPy](https://pypi.org/project/NEPy/) package, it's about 70 times faster (due to vectorized reading with numpy.fromfile) and correctly handles (at least) the NEDF 1.3 and 1.4 files.

Once the sample data file (https://github.com/mne-tools/mne-testing-data/pull/80) has been added, I'll add a unit test.


mne/io/nedf/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""NEDF file import module"""
+
+# Author: Tristan Stenner <nedf@nicht.dienstli.ch>
```
cbrnr, line 3: 😆

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+import numpy as np
+
+from .. import BaseRaw
+from ... import create_info, Annotations
+
+
+def parse_nedf_header(filename):
+    """
+    Read the header information from the first 10kB of an .nedf file
```
cbrnr, line 16:
```suggestion
    """Read header information from the first 10kB of an .nedf file.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Parameters
+    ----------
+    filename : str
+        Path to the .nedf file
```
cbrnr, line 21:
```suggestion
        Path to the .nedf file.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Returns
+    -------
+    info : dict
+        A dictionary with information from the header
```
cbrnr, line 26:
```suggestion
        A dictionary with header information.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    dt : numpy.dtype
+        structure of the binary EEG+accelerometer+trigger data in the file
```
cbrnr, line 28:
```suggestion
        Structure of the binary EEG/accelerometer/trigger data in the file.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    info = {}
+    # nedf files have some accelerometer channels sampled at 100Hz and
+    # several other channels sampled at 500Hz.
+    # The layout is
+    #  (100HzCh1S1, 100HzCh2S1, 100HzCh3S1),
+    # ((500HzCh1S1, 500HzCh2S1, …, 500HzChnS1),…,
+    #  (500HzCh1S2, 500HzCh2S2, …, 500HzChnS2), …
+    #  (500HzCh1S5,  500HzCh2S5, …, 500HzChnS5)),
```
cbrnr, line 37:
```suggestion
    #  (500HzCh1S5, 500HzCh2S5, …, 500HzChnS5)),
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    #  (100HzCh1S2, 100HzCh2S2, 100HzCh3S2) and so on
+
+    # dtype for the binary data block
+    dt = []
+    # dtype for a single EEG sample
+    datadt = []
```
cbrnr, line 43:
```suggestion
    dt = []  # dtype for the binary data block
    datadt = []  # dtype for a single EEG sample
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    with open(filename, 'rb') as f:
+        header = f.read(10240)
+
+    headerend = header.find(b'\0')
+    if headerend == -1:
+        raise RuntimeError('End of header null not found')
+    headerxml = ElementTree.fromstring(header[:headerend])
+    nedfversion = headerxml.findtext('NEDFversion', '')
+    if nedfversion not in ['1.3', '1.4']:
+        print('Unexpected NEDFversion, hope this works anyway')
```
cbrnr, line 53: Maybe use `warn` instead of `print`?

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+    if headerxml.find('STIMSettings'):
+        # 2* -> two stim samples per eeg sample
+        datadt.append(('stim', 'B', (2, nchantotal, 3)))
+
+    if 'AdditionalChannelStatus' in headerxml:
+        raise RuntimeError('Unexpected AdditionalChannelStatus')
+
+    if headerxml.findtext('stepDetails/DeviceClass', '') == 'STARSTIM':
+        print('Found Starstim, not sure how to handle this')
```
cbrnr, line 82: Again maybe use `warn`. Also, would it be better to abort in such a case? The message is not really reassuring and people might be worried that the result is not what they expect.

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    dt.append(('data', np.dtype(datadt), (5,)))
+
+    date = headerxml.findtext('StepDetails/StartDate_firstEEGTimestamp', '')
+    info['meas_date'] = datetime.datetime.utcfromtimestamp(int(date) / 1000)
+    return info, np.dtype(dt)
+
+
+def _read_nedf_eeg(filename: str):
+    """
+    Read header info and EEG data from an .nedf file
```
cbrnr, line 97:
```suggestion
    """Read header info and EEG data from an .nedf file
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Parameters
+    ----------
+    filename : str
+        Path to the .nedf file.
+
+    Returns
+    -------
+    eeg : array, shape (n_samples, n_channels)
+        Unscaled EEG data
```
cbrnr, line 107:
```suggestion
        Unscaled EEG data.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    info : dict
+        Information from the file header
```
cbrnr, line 109:
```suggestion
        Information from the file header.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    triggers : array, shape (n_annots, 2)
+        Start samples and values of each trigger
```
cbrnr, line 111:
```suggestion
        Start samples and values of each trigger.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    scale : float
+        Scaling factor for the EEG data
```
cbrnr, line 113:
```suggestion
        Scaling factor for the EEG data.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    info, dt = parse_nedf_header(filename)
+
+    # to quote the original matlab implementation:
```
cbrnr, line 117: Is the original implementation linked somewhere? Is it necessary to check for license compatibility, or did you write this code from scratch?

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+    eeg = eeg.reshape((-1, info['nchan']))
+    # convert sign if necessary
+    eeg[eeg > (1 << 23)] -= 1 << 24
+
+    triggers = data['data']['trig'].flatten()
+    triggerind = triggers.nonzero()[0]
+    triggers = np.stack((triggerind, triggers[triggerind])).T
+
+    # scale channels accordingly (here: to volts)
+    scale = 2.4 / (6.0 * 8388607)
```
cbrnr, line 135: What do these numbers mean?

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+        eeg, header, triggers, scale = _read_nedf_eeg(nedffile)
+        info = create_info(ch_names=header['ch_names'],
+                           sfreq=header['sfreq'],
+                           ch_types='eeg')
+
+        info['bads'] = []
+        info['meas_date'] = header['meas_date']
+
+        super().__init__(info, preload=(eeg * scale).T, filenames=nedffile)
+        self.set_montage('standard_1020')
```
cbrnr, line 154: Does this work for every file?

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+        triggers = header['events']
+        onsets = triggers[:, 0] / info['sfreq']
+        self.set_annotations(Annotations(onsets, 0, triggers[:, 1]))
+
+
+def read_raw_nedf(filename):
+    """
+    Read NeuroElectrics .nedf files.
```
cbrnr, line 163:
```suggestion
    """Read NeuroElectrics .nedf files.
```

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+
+    NEDF file versions starting from 1.3 are supported.
+
+    Parameters
+    ----------
+    filename : str
+        Path to the .nedf file
```
cbrnr, line 170:
```suggestion
        Path to the .nedf file.
```

mne/io/nedf/nedf.py
tstenner: I read through the EEGLAB plugin (https://www.neuroelectrics.com/resources/software/) to get an idea how the nedf files are structured, but other than that everything's written from scratch.

mne/io/nedf/nedf.py
tstenner: I have no idea, but it's the same scaling factor the EEGLAB plugin uses.

mne/io/nedf/nedf.py
tstenner: As far as I know the NeuroElectrics enobio/StarStim devices only have 10-10/10-20 caps, so until they sell caps with custom layouts it works.

mne/io/nedf/__init__.py
tstenner: I wasn't sure if I should put my work email address in there.

mne/io/nedf/nedf.py
cbrnr: OK, then either don't mention "the original matlab implementation" in the comment or provide a link so that people can check it out themselves.

mne/io/nedf/nedf.py
cbrnr: Who wrote that plugin? The company? Then maybe mention that these magic numbers come from the company itself.

mne/io/nedf/__init__.py
cbrnr: I like the domain! You can decide yourself which email address you want to share.

mne/io/nedf/nedf.py
tstenner: From my experience with NE, NEDFVersion 1.3 and 1.4 and especially the stimulation data underwent some changes, but it mostly works. For some versions of the recording software and stimulation settings, the data block sizes are unexpected, but that's impossible to miss.

mne/io/nedf/nedf.py
tstenner: Done.

mne/io/nedf/nedf.py
tstenner: Done.

mne/io/nedf/nedf.py
tstenner: Done.

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,181 @@
...
+        Start samples and values of each trigger.
+    scale : float
+        Scaling factor for the EEG data.
+    """
+    info, dt = parse_nedf_header(filename)
+
+    # to quote the NEDF EEGLAB plugin
+    # (https://www.neuroelectrics.com/resources/software/):
+    # "binary data will always start at byte 5120"
+    # newer versions have 10kB for the header
```
cbrnr, line 118: OK, but why is "binary data will always start at byte 5120" relevant if the code only uses 10kB? If 5120 is never used, it's better to remove this statement to avoid confusion.

tstenner: Agreed, it's removed in a new commit. I'll push it with the unit tests so the CI doesn't feel the need to test everything again.

larsoner:
I'll went ahead and merged the testing dataset so that you can add tests -- if we need more data or documentation for the testing file we can always add it later

larsoner:
0.113 hash when I downloaded it was ce114ad6d5e3dbed06119386e6b1ce0c

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,193 @@
...
+    hdrlen = 10240
+    with open(filename, mode='rb') as f:
+        info, dt = _parse_nedf_header(f.read(hdrlen))
+        data = np.fromfile(f, dtype=dt)
+
+    # convert uint8-triplet -> float32
+    eeg = data['data']['eeg'] @ [1 << 16, 1 << 8, 1.]
+    eeg = eeg.reshape((-1, info['nchan']))
+    # convert sign if necessary
+    eeg[eeg > (1 << 23)] -= 1 << 24
```
larsoner, line 138: It will be pretty easy to convert this to something that properly supports `preload=True | False` if you're on board with my comment below about using STIM channels rather than annotations. Let me know if you want help with this and I can push a commit

mne/io/nedf/nedf.py
```diff
@@ -0,0 +1,193 @@
...
+
+    triggers = data['data']['trig'].flatten()
+    triggerind = triggers.nonzero()[0]
+    triggers = np.stack((triggerind, triggers[triggerind])).T
```
larsoner, line 142: This I'm not so sure about. In most systems that have digitally sampled TTL trigger channels (which this is, right?) we expose these as `STIM` channels, as this allows using `find_events`. I think it makes sense to continue this practice, rather than convert these to annotations directly.

mne/io/nedf/tests/test_nedf.py
```diff
@@ -0,0 +1,91 @@
...
+    assert_allclose(data_end, .0176, atol=.01)
+    assert_allclose(raw.get_data('Fpz', 0, 100).mean(), .0185, atol=.01)
+
+    assert_allclose(raw.annotations.onset, [22.384, 38.238, 49.496, 63.15])
+    assert_equal(raw.info['meas_date'].year, 2019)
+    assert_equal(raw.ch_names[2], 'AF7')
+
+    for ch in raw.info['chs']:
+        assert_equal(ch['kind'], FIFF.FIFFV_EEG_CH)
+        assert_equal(ch['unit'], FIFF.FIFF_UNIT_V)
```
larsoner, line 91: Eventually we should add _test_raw_reader, it exposes bugs with processing usually
```suggestion
        assert_equal(ch['unit'], FIFF.FIFF_UNIT_V)

    _test_raw_reader(read_raw_nedf, eegfile, test_preloading=False)
```
But it might be best to use this after we support not preloading

larsoner (commented):
Looks good! See comments about STIM channels and potential `preload` support

mne/io/nedf/nedf.py
tstenner: Although a TTL adapter exists, I don't know anyone using it or if it's recorded here (or even if it can encode more than one bit of information). From my understanding, the recording software subscribes to a specific LSL outlet and writes the LSL events to the trigger channel with a duration of a single sample.

larsoner: I see. I'm still somewhat inclined to have this as a STIM channel and have people use `find_events` with it, as it's basically a "channel" sampled at the rate of the data (even if the LSL stream does not actually use TTL pulses).

@agramfort any opinion?

mne/io/nedf/nedf.py
tstenner: The reader design with the numpy dtype would be perfect for lazy loading / `mmap()`ing files, but there's not native int24 type and enumerating the annotations requires reading the data anyway so I skipped it.

Adding `preload` support would mean that all calls to `get_data()` call `_read_segment_file()`, for `preload=True` it returns the previously loaded data and for `preload=False` loads it from disk?

larsoner: > Adding preload support would mean that all calls to get_data() call _read_segment_file(), for preload=True it returns the previously loaded data and for preload=False loads it from disk?

From an implementation standpoint it means that we need to implement `_read_segment_file` properly, and never in the constructor bother reading the entire data -- the `BaseRaw` class handles using `_read_segment_file` to load all data when a user does `preload=True`. So the code here ultimately would be a bit simpler.

This is also part of why I'd like to avoid the `annotations` approach -- it requires reading all the data from disk (or at least sampling every Nth stim channel sample), which at least partially negates the benefits of `preload=False` (not needing to load the whole thing first).

> mmap()ing files

FWIW we actually have a way to do this with `preload=<string>` already in the base class, no need to implement this at the individual-reader level

tstenner: Sounds good :+1:

Does this load data and triggers in a single pass or one pass for data and triggers each?

larsoner: Single pass since it treats the trigger just like any other channel

larsoner: Do you want to take a stab at making the necessary changes or would it help if I did?

tstenner: I haven't worked with stim channels before, so I'd be glad about any help

larsoner:
Okay pushed commits to:

1. Add `_test_raw_reader`
2. Remove `assert_equal` in favor of plan pytest `assert a == b`
3. Remove `assert_raises` from NumPy in favor of `pytest.raises` with more targeted `match`es
4. Change handling of stim channels to create `STI 014` out of the stim data
5. Add non-preload support
6. Handle last samples properly (previous version discarded the last 3 samples of data because they didn't fit evenly into the 5-sample standard)

Feel free to take a look and see if it makes sense @tstenner ! Happy to talk through / explain any changes, or fix things that you see are wrong with what I did

larsoner:
Also @tstenner can you add your name/URL to doc/changes/names.inc and update doc/changes/latest.inc ? Otherwise LGTM +1 for merge

mne/io/nedf/nedf.py
larsoner: I undid this line about `set_montage`. If tomorrow they start selling different caps we have to go through a deprecation cycle to get rid of this `standard_1020` behavior or do something more complicated, and it's easy enough for users to do `read_raw_nedf(fname).set_montage('standard_1020')`.

One way I could think to keep this sort of line would be to have a check that if all of the EEG channels in `info` match the `standard_1020` names then we set it for them. But given discussions we've had about realistic versus pure-spherical electrode positions and the like, I don't mind people needing to add their own `set_montage` to be explicit about the montage they're using. It also allows potentially setting the head size, etc. -- sorts of things it might be good to have users think about / work through in some tutorials.

tstenner:
Much appreciated, it's been a while since I've seen the insides MNE-Python.

> Remove assert_equal in favor of plan pytest assert a == b

Ah, good to know. I saw a bunch of `assert_equal`s in the BrainVision reader tests so I adopted this style.

> Handle last samples properly (previous version discarded the last 3 samples of data because they didn't fit evenly into the 5-sample standard)

I'm not sure if it's needed. Most of our in-house readers ignore the last nsamples%5 samples, because it's at most 4 samples, i.e. 8 ms and there's always more data at the end than needed because the recording has to be stopped manually. I'd rather have a somewhat simpler implementation (without allocating space for and copying the whole data in 4/5 cases), but that's just my opinion.

larsoner:
> I'm not sure if it's needed. Most of our in-house readers ignore the last nsamples%5 samples, because it's at most 4 samples, i.e. 8 ms and there's always more data at the end than needed because the recording has to be stopped manually. I'd rather have a somewhat simpler implementation (without allocating space for and copying the whole data in 4/5 cases), but that's just my opinion.

Indeed it's probably overkill in most situations, but if we can read the data correctly rather than omitting some samples 80% of the time I think we might as well do it. This code's one job is to pull/convert the right set of samples from disk, it's a shame to do it for all but the last few :) And ultimately the code to handle it did not end up being too complex.

If efficiency is a concern we can always make the code more efficient in the future, but I'd rather do that later if we find the copy/concatenate pattern to be a limiting factor in real use cases.

larsoner:
Thanks @tstenner !

PR #8735, opened 2021-01-12
WIP: Reject out of source space picks

GuillaumeFavelier:
This PR rejects the picked points out of source space. The goal is to avoid flat time course at some points.

It's still a work in progress.


GuillaumeFavelier:
Closing since the experience is not satisfying at all (I'm lucky if I end up picking a vertex in the source space). 

I'll reopen if I find a better solution.

PR #8736, opened 2021-01-12
MRG, VIZ, FIX: sEEG picking in _prepare_topomap_plot()

### LINKED ISSUE #8733: ValueError in mne.viz.plot_evoked_joint
I am using SEEG data, and I want to plot evoke joint figure, while I've got this problem.
```
    mne.viz.plot_evoked_joint(evoke)
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\viz\evoked.py", line 1458, in plot_evoked_joint
    colorbar=False, **topomap_args_pass)
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\evoked.py", line 356, in plot_topomap
    nrows=nrows, ncols=ncols)
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\viz\topomap.py", line 1540, in plot_evoked_topomap
    _prepare_topomap_plot(evoked, ch_type, sphere=sphere)
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\viz\topomap.py", line 103, in _prepare_topomap_plot
    exclude='bads')
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\io\pick.py", line 446, in pick_types
    _check_meg_type(meg)
  File "D:\SEEG_Cognition\SEEG\lib\site-packages\mne\io\pick.py", line 348, in _check_meg_type
    % (allowed_types, meg))
ValueError: meg value must be one of ['grad', 'mag', 'planar1', 'planar2'] or bool, not seeg
```
my code is pretty simpe, just 
```
import mne

fpath = 'color_epoch.fif'
data = mne.read_epochs(fpath, preload=True)
evoke = data['red'].average()
mne.viz.plot_evoked_joint(evoke)
```
Here is the data's link.

https://drive.google.com/file/d/1zIinPymoWgHlKav594EHyekU8zjewVS-/view?usp=sharing

drammock:
closes #8733 

Not sure if this is a reasonable fix since I've never personally worked with sEEG, but based on general knowledge I'm assuming there's nothing special needed as far as combining sensor pairs?  Can someone confirm?

cc @BarryLiu97.  On this branch using your code from #8733 I get this figure:

Figure_1

### DIFF
--- a/mne/viz/tests/test_topo.py
+++ b/mne/viz/tests/test_topo.py
@@ -6,7 +6,6 @@
 #
 # License: Simplified BSD
 
-import warnings
 import os.path as op
 from collections import namedtuple
 
@@ -112,8 +111,7 @@ def test_plot_joint():
     # test sEEG (gh:8733)
     evoked.del_proj().pick_types('mag')  # avoid overlapping positions error
     mapping = {ch_name: 'seeg' for ch_name in evoked.ch_names}
-    with warnings.catch_warnings():  # unit has changed from T/m to V
-        warnings.simplefilter('ignore')
+    with pytest.warns(RuntimeWarning, match='The unit .* has changed from .*'):
         evoked.set_channel_types(mapping)
     evoked.plot_joint()
 
--- a/mne/viz/tests/test_topo.py
+++ b/mne/viz/tests/test_topo.py
@@ -6,6 +6,7 @@
 #
 # License: Simplified BSD
 
+import warnings
 import os.path as op
 from collections import namedtuple
 
@@ -107,6 +108,14 @@ def test_plot_joint():
     evoked.plot_joint(ts_args=dict(proj='reconstruct'),
                       topomap_args=dict(proj='reconstruct'))
     plt.close('all')
+
+    # test sEEG (gh:8733)
+    evoked.del_proj().pick_types('mag')  # avoid overlapping positions error
+    mapping = {ch_name: 'seeg' for ch_name in evoked.ch_names}
+    with warnings.catch_warnings():  # unit has changed from T/m to V
+        warnings.simplefilter('ignore')
+        evoked.set_channel_types(mapping)
+    evoked.plot_joint()
 
 
 def test_plot_topo():
--- a/mne/viz/tests/test_topo.py
+++ b/mne/viz/tests/test_topo.py
@@ -85,8 +85,8 @@ def test_plot_joint():
                                                       time_unit='ms'),
                       ts_args=dict(spatial_colors=True, zorder=return_inds,
                                    time_unit='s'))
-    pytest.raises(ValueError, evoked.plot_joint, ts_args=dict(axes=True,
-                                                              time_unit='s'))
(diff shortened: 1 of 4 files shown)


drammock:
All green, ready for review/merge.

BarryLiu97:
Great!

agramfort:
given the extrapolation artifacts wouldn't it make sense to restrict the topomap to the convex hull of the sensors?

mne/viz/tests/test_topo.py
```diff
@@ -108,6 +109,14 @@ def return_inds(d):  # to test function kwarg to zorder arg of evoked.plot
                       topomap_args=dict(proj='reconstruct'))
     plt.close('all')
 
+    # test sEEG (gh:8733)
+    evoked.del_proj().pick_types('mag')  # avoid overlapping positions error
+    mapping = {ch_name: 'seeg' for ch_name in evoked.ch_names}
+    with warnings.catch_warnings():  # unit has changed from T/m to V
+        warnings.simplefilter('ignore')
+        evoked.set_channel_types(mapping)
```
larsoner, line 117: Typically with these we just do:
```suggestion
    with pytest.warns(RuntimeWarning, match='unit changed')
        evoked.set_channel_types(mapping)
```
Does the same (plus a bit more, checking that we got the expected one) with fewer lines/imports

larsoner (commented):
Otherwise LGTM

And yes in the example something like `extrapolate='local'` (I think this is what does the convex + padding?) would look better

drammock:
Here's what the sample code from #8733 generates now:

Figure_1

agramfort:
I find it better

jasmainak:
Late to the party, does `plot_joint` even make sense for sEEG data? What can you tell from this figure? I think you should just throw an error.

You need proper support for sEEG and a separate function altogether. Each "stick" plotted in a separate subplot.

larsoner:
Even if suboptimal I think it should work. We can create other optimized functions for sEEG visualization that people can can prefer as well, but I don't see a good reason to raise an error here.

BarryLiu97:
> Late to the party, does `plot_joint` even make sense for sEEG data? What can you tell from this figure? I think you should just throw an error.
> 
> You need proper support for sEEG and a separate function altogether. Each "stick" plotted in a separate subplot.

This sample data is not very good for many reasons, like the experimental paradigm is not good enough. So it is normal to see nothing useful, but to sEEG, it is useful if the experimental paradigm is good.

jasmainak:
@BarryLiu97 do you have a screenshot or a figure from a publication to share which shows how topoplots are used in sEEG data analysis?

BarryLiu97:
> @BarryLiu97 do you have a screenshot or a figure from a publication to share which shows how topoplots are used in sEEG data analysis?

Sadly, I haven't seen any publication using figures like plot_evoke_joint. But in sEEG, it may make sense if  using brain regions as the standard to separate the subplots for the most important feature of sEEG is the ability to get the deep brain structures' information.

larsoner:
Thanks for the fixes @drammock !

PR #8744, opened 2021-01-13
MRG: Support for non-FIFF files in Report.parse_folder

hoechenberger:
Support all non-FIFF files we can read when parsing a folder using `Report.parse_folder()`. Requested via https://mne.discourse.group/t/naming-convention-in-mne-vs-mne-bids-welcome-to-the-thunderdome/2467

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,6 +23,8 @@ Enhancements
 
 - Speed up :func:`mne.inverse_sparse.tf_mixed_norm` using STFT/ISTFT linearity (:gh:`8697` by `Eric Larson`_)
 
+- `~mne.Report.parse_folder` now processes supported non-FIFF files by default, too (:gh:`8744` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix zen mode and scalebar toggling for :meth:`raw.plot() <mne.io.Raw.plot>` when using the ``macosx`` matplotlib backend (:gh:`8688` by `Daniel McCloy`_)
--- a/mne/tests/test_report.py
+++ b/mne/tests/test_report.py
@@ -41,6 +41,10 @@ trans_fname = op.join(report_dir, 'sample_audvis_trunc-trans.fif')
 inv_fname = op.join(report_dir,
                     'sample_audvis_trunc-meg-eeg-oct-6-meg-inv.fif')
 mri_fname = op.join(subjects_dir, 'sample', 'mri', 'T1.mgz')
+bdf_fname = op.realpath(op.join(op.dirname(__file__), '..', 'io',
+                                'edf', 'tests', 'data', 'test.bdf'))
+edf_fname = op.realpath(op.join(op.dirname(__file__), '..', 'io',
+                                'edf', 'tests', 'data', 'test.edf'))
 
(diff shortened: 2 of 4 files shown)


agramfort (approved):
perfect !

mne/tests/test_report.py
```diff
@@ -170,6 +174,43 @@ def test_render_report(renderer, tmpdir):
...
+    report = Report()
+    report.parse_folder(data_path=tempdir, render_bem=False, on_error='raise')
+
+    # Check correct paths and filenames
+    for fname in fnames_out:
+        assert (op.basename(fname) in
+                [op.basename(x) for x in report.fnames])
+        assert (''.join(report.html).find(op.basename(fname)) != -1)
+
+    assert_equal(len(report.fnames), len(fnames_out))
```
larsoner, line 200: We prefer plain `assert` to `assert_equal` nowadays usually -- `assert_equal` is less informative than plain `assert len(report.fnames) == len(fnames_out)`. Feel free to remove all uses in this file if you want, but at least let's not add more of them

mne/tests/test_report.py
```diff
@@ -170,6 +174,43 @@ def test_render_report(renderer, tmpdir):
...
+    assert_equal(len(report.html), len(report.fnames))
+    assert_equal(len(report.fnames), len(report))
+
+    report.data_path = tempdir
+    fname = op.join(tempdir, 'report.html')
+    report.save(fname=fname, open_browser=False)
+    with open(fname, 'rb') as fid:
+        html = fid.read().decode('utf-8')
```
larsoner, line 208: Why you do this and `''.join(report.html)` above? I would do one or the other once and use the result all places

doc/changes/latest.inc
```diff
@@ -23,6 +23,8 @@ Enhancements
 
 - Speed up :func:`mne.inverse_sparse.tf_mixed_norm` using STFT/ISTFT linearity (:gh:`8697` by `Eric Larson`_)
 
+- `~mne.Report.parse_folder` now processes supported non-FIFF files by default, too (:gh:`8744` by `Richard Höchenberger`_)
```
larsoner, line 26: The `~` will make it show up just as `parse_folder` which will be less easy to read for non-MNE-experts than the full qualified name (or at least `Report.parse_folder`)
```suggestion
- `mne.Report.parse_folder` now processes supported non-FIFF files by default, too (:gh:`8744` by `Richard Höchenberger`_)
```

larsoner (commented):
Just minor comments, otherwise LGTM!

hoechenberger:
@larsoner I have addressed your comments.

mne/tests/test_report.py
hoechenberger: I've replaced all calls to `assert_equal` with plain `assert` now

mne/tests/test_report.py
hoechenberger: I've dropped the `join` line.

larsoner (approved):
LGTM feel free to merge when CIs come back happy, thanks @hoechenberger !

agramfort:
maybe some CIs need to be restarted. Than let's merge

hoechenberger:
One CI run failing again, this is caused by an upstream problem:
https://github.com/jupyter/nbformat/issues/206

PR #8749, opened 2021-01-14
WIP: Prototype of comparative views

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/issues/8748#issuecomment-760081331 and investigate ways to improve perception of the electrodes.

- [x] Try `vtkPolyDataSilhouette` on the brain (from https://github.com/mne-tools/mne-python/issues/8748#issuecomment-760157228)

It's still work in progress.


GuillaumeFavelier:
This is how it looks:

image

https://24532-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_seeg.html#sphx-glr-auto-tutorials-misc-plot-seeg-py

GuillaumeFavelier:
A good part of https://github.com/mne-tools/mne-python/blob/05863aceab82e8de68a731336f5a1a6220dacd4f/mne/viz/_brain/_brain.py#L2097-L2110 could be reused

jasmainak:
looks promising! feel free to ping me when you've implemented this edge contrast thing. I'll give it a shot

GuillaumeFavelier:
I'll link this concept of comparative views to an open issue on pyvista because the current solution duplicates the scene:

https://github.com/pyvista/pyvista/issues/395

GuillaumeFavelier:
Another try:

image

jasmainak:
Not sure why the electrodes have transparency. So for sEEG, you want to know for example which electrodes are in the white matter and which are on the cortex. Can you tell from the current viz?

GuillaumeFavelier:
I remember having a discussion about that... AFAIK the transparency was added because it helping actually.

> which electrodes are in the white matter and which are on the cortex.

Interesting... I would compute the convex hull of the white matter surface and use a different color code to differentiate them afterwards (those who lie inside or not).

GuillaumeFavelier:
Do we have an easier way to differentiate them?

jasmainak:
I see ... so currently I have sliced MRI images (3 views) with the electrode locations in a pdf from my collaborators. Every time I see something interesting in the time series, I go back to the pdf to see where this electrode is in the pdf :)

I'm just thinking out loud, maybe this kind of plot is hard to interpret? I've seen folks put names next to the electrodes though, it helps with the interpretability.

jasmainak:
> different color code to differentiate them afterwards (those who lie inside or not).

might be worth giving a shot if it's easy!

drammock:
I think what our eye expects is that in a semi-transparent brain, deeper electrodes should be harder to see.  That is not really happening the way our eye expects because the brain is being rendered as a surface (so all electrodes beneath the surface are getting occluded by the same amount, instead of deeper ones being fainter.  So the "right" solution is to render the brain as a semi-transparent solid instead of a semi-transparent surface.  I'm guessing that even trying to fake that by computing the distance from electrode to surface in the direction of the viewport might be too expensive to do on the fly?  WDYT @GuillaumeFavelier ?

GuillaumeFavelier:
> what our eye expects is that in a semi-transparent brain, deeper electrodes should be harder to see.

Totally agree :)

Among the tools I know, there are the depth buffer, ray casting and rgba buffers.

In theory, it should be possible to factor in the updated dynamic values of the depth buffer to set the opacity (some kind or linear relation based on distance). It sounds a lot like simulating volume rendering and I expect it to be costly.

GuillaumeFavelier:
Following a totally different idea, there is also [physically based rendering](https://github.com/pyvista/pyvista/pull/823) which could potentially improve perception?

drammock:
> It sounds a lot like simulating volume rendering and I expect it to be costly.

that's what I feared

GuillaumeFavelier:
Maybe it does not need to dynamic though.

Maybe something as simple as decreasing the opacity of the electrode as it gets closer to the center of mass of the brain could help?

larsoner:
I'm actually not sure it will be so slow. You don't have to ray trace, it's close to being just a matrix multiplication or two I think based on the world-to-viewer (composite) matrix, *and* it only needs to be updated when this view changes (or the electrode position changes, which is never in our viz). I would see how hard it is to make something like alpha depend on camera-to-electrode distance for each electrode. I would be surprised if VTK didn't even implement this in C for us -- I think we "just" need to find the right function(s) and assemble a pipeline for that updates the scalars automatically.

In a worst case we can do it in Python with some view-changed callback but I doubt we need to do it. @GuillaumeFavelier if you look around and don't find anything like this I can look.

larsoner:
(Distance from the center of the brain might also be better, not sure.)

GuillaumeFavelier:
Using https://github.com/mne-tools/mne-python/pull/8749#issuecomment-762356076 and `_CheckInside`:

image

https://user-images.githubusercontent.com/18143289/105360505-f11a9400-5bf8-11eb-9992-ef317a9c0fd3.mp4

larsoner:
Nice! One idea for a plan would be:

1. PR 1: add this "edge highlighting" effect as an option in Brain.__init__. Want to try a PR adding it there and modifying some examples? The LCMV MIP volumetric rendering and vector source estimate examples in particular will probably benefit from it.
2. Tackle something like a `brain.add_evoked(...)` that allows you to add sensors whose color changes according to the "activation" levels (#8382). `plot_alignment` has all the code for how to draw the sensors, then hopefully it's "just" a matter of refactoring that, adding some `clim` handling, and mapping the scalars appropriately.
3. Add some support for `add_evoked(..., surf_maps=None)` where you can pass `surf_maps` like you can for `plot_evoked_field`. Instead of plotting sensors themselves it will plot the proper surface mapping (to scalp for EEG and to helmet for MEG, generally), this would finish #8382 I think
4. Deal with the inside/outside stuff and other API considerations for sEEG data. I'm less sure of an API but it seems like it will be useful to get sensor plotting working first so we have a general framework internally for handling sensors that we can leverage

jasmainak:
looks really nice! do you think we could now somehow add the channel names to the plot? One way to avoid clutter would be show the names of only first and last channel on a shaft (since everything else is consecutive). Then this plot begins to become usable!

larsoner:
> Do you think we could now somehow add the channel names to the plot?

+1 for doing this as a follow-up PR after PR-2 `brain.add_evoked` above. It needs its own separate API like `show_names=False | list` where `list` contains a list of the channel names to show next to the channels or something. Then it's up to the user to say which channel names they actually want shown, which I think is reasonable (in sEEG you should know which are the first/last in a shaft, or we can make some sEEG-specific function to extract these).

GuillaumeFavelier:
What exactly did you have in mind for **PR1** @larsoner? I opened https://github.com/mne-tools/mne-python/pull/8771#issue-559860733 to discuss it.

GuillaumeFavelier:
https://github.com/mne-tools/mne-python/pull/8803 should be **PR2**

PR #8752, opened 2021-01-15
MRG, VIZ: Fix title position in plot_sensors

cbrnr:
The title is out of bounds in figures created by `plot_sensors` (including `montage.plot()`).

Before:
before

After:
after

### DIFF
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -1068,8 +1068,10 @@ def _plot_sensors(pos, info, picks, colors, bads, ch_names, title, show_names,
         picker = partial(_onpick_sensor, fig=fig, ax=ax, pos=pos,
                          ch_names=ch_names, show_names=show_names)
         fig.canvas.mpl_connect('pick_event', picker)
-
-    _set_window_title(fig, title)
+    if axes_was_none:
+        _set_window_title(fig, title)
+    else:
+        ax.set(title=title)
     closed = partial(_close_event, fig=fig)
     fig.canvas.mpl_connect('close_event', closed)
     plt_show(show, block=block)
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -1046,7 +1046,7 @@ def _plot_sensors(pos, info, picks, colors, bads, ch_names, title, show_names,
         # Equal aspect for 3D looks bad, so only use for 2D
         ax.set(aspect='equal')
         if axes_was_none:
-            fig.subplots_adjust(left=0, bottom=0, right=1, top=0.95)
+            fig.subplots_adjust(left=0, bottom=0, right=1, top=1)
         ax.axis("off")  # remove border around figure
     del sphere
 
@@ -1069,7 +1069,7 @@ def _plot_sensors(pos, info, picks, colors, bads, ch_names, title, show_names,
                          ch_names=ch_names, show_names=show_names)
         fig.canvas.mpl_connect('pick_event', picker)
 
-    ax.set(title=title)
+    _set_window_title(fig, title)
     closed = partial(_close_event, fig=fig)
     fig.canvas.mpl_connect('close_event', closed)
     plt_show(show, block=block)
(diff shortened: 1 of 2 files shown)


hoechenberger (approved):
Care to add a changelog entry? :)

larsoner:
@drammock feel free to review and merge when you're happy since you've been working in the `viz` code a lot lately

cbrnr:
We could also move the title from the canvas to the window title bar (and maximize the montage plot again). What do you think?

hoechenberger:
> We could also move the title from the canvas to the window title bar (and maximize the montage plot again). What do you think?

Me likes!! 😍

cbrnr:
Ooops... will fix this mess...

drammock:
> We could also move the title from the canvas to the window title bar (and maximize the montage plot again).

There is a downside to this, in that window titles don't show up in our tutorial docs.  We're not super consistent about this throughout the codebase, but I've been leaning toward using the plot title for generic descriptive stuff like "Power spectral density (magnetometers)" or "sensor positions", and using the window title for the information that is more useful interactively (esp. during preprocessing), like the subject ID.  On that basis, I would lean toward leaving this as a plot title rather than a window title.

On the other hand, sensor location plots often get inset into data plots, so it might simplify things to *not* add axes titles to them by default?  In which case putting that info in a window title makes sense (as long as we're careful that it doesn't change the window title when the sensor plot is merely an inset).

cbrnr:
In general I agree, but in this case the title "Sensor positions (eeg)" really doesn't contain a lot of helpful information and takes up a lot of space. Either way, let me know how to proceed.

drammock:
> putting that info in a window title makes sense (as long as we're careful that it doesn't change the window title when the sensor plot is merely an inset)

I think this reasoning wins out for me.  Just reading the code, I think currently it *will* change the figure title even when the sensor plot is merely an inset axes....  yep, just tested with this:

```python
import os
import mne
plt.ion()
fig, ax = plt.subplots()
inax = ax.inset_axes((0.1, 0.1, 0.4, 0.4))

dp = mne.datasets.multimodal.data_path()
raw = mne.io.read_raw(os.path.join(dp, 'multimodal_raw.fif'))
mne.viz.plot_sensors(raw.info, axes=inax)
```

cbrnr:
Is there a way to check whether the axes are the sole axes in a figure or if they are inset?

drammock:
> Is there a way to check whether the axes are the sole axes in a figure or if they are inset?

I don't think so...  I think `ax.inset_axes()` is ultimately syntactic sugar around `fig.add_axes()`, so they end up as the same object type.  I think the right approach is to *only* set the figure window title in cases where our `plot_sensors()` function created the figure in the first place...  so you can leverage the `axes_was_none` variable for that I think?

cbrnr:
@drammock I think this should work now - can you check?

mne/viz/utils.py
```diff
@@ -1068,8 +1070,10 @@ def _plot_sensors(pos, info, picks, colors, bads, ch_names, title, show_names,
         picker = partial(_onpick_sensor, fig=fig, ax=ax, pos=pos,
                          ch_names=ch_names, show_names=show_names)
         fig.canvas.mpl_connect('pick_event', picker)
-
-    _set_window_title(fig, title)
+    if axes_was_none:
+        _set_window_title(fig, title)
+    else:
+        ax.set(title=title)
```
drammock, line 1076:
```suggestion
```

remove these.  We don't want to set an axes title when it's an inset axes.

mne/viz/utils.py
```diff
@@ -1045,8 +1045,10 @@ def _plot_sensors(pos, info, picks, colors, bads, ch_names, title, show_names,
 
         # Equal aspect for 3D looks bad, so only use for 2D
         ax.set(aspect='equal')
-        if axes_was_none:
+        if axes_was_none:  # we'll show the plot title as the window title
             fig.subplots_adjust(left=0, bottom=0, right=1, top=1)
+        else:  # make room for title in axes
+            fig.subplots_adjust(left=0, bottom=0, right=1, top=0.95)
```
drammock, line 1051:
```suggestion
```

remove these.  If `axes` is an inset, this affects the parent axes (making its ticks and labels go off-canvas). Here I think the same logic applies as with the window title: if we didn't create the figure ourselves, we probably shouldn't be doing any calls to `subplots_adjust()`

drammock (changes requested):

drammock (approved):

drammock:
thanks @cbrnr!

Issue #5269, opened 2018-06-07
Feature: Standard electrode position computation


sappelhoff:
I have recently finished a little pet project of mine, in which I compute all 345 standard EEG electrode positions of the 10-05 (5%) system. Although that's nothing new per se and the data are available e.g., from [Robert Oostenveld](http://robertoostenveld.nl/electrode/), I thought that it would be nice for MNE to have the actual computations, which then yield the data to be used as part of the code. Currently, there are only a bunch of data files in [`mne/channels/data`](https://github.com/mne-tools/mne-python/tree/master/mne/channels/data) which is not as transparent as the code would be.

... as I started to hint at on [gitter](https://gitter.im/mne-tools/mne-python) yesterday, this also relates to the issues about montages, layouts, and digitization, see #5190 #3987

What do you think?

This is the [repository](https://github.com/sappelhoff/eeg_positions) containing the computations I am talking about.

Here is a [short notebook](https://gist.github.com/sappelhoff/d6eba442088c0b3528345e53dad278b1) showing an initial integration with MNE.

And a plot:
![screenshot from 2018-06-07 12-23-45](https://user-images.githubusercontent.com/9084751/41094203-c0bb02da-6a4d-11e8-856e-ba1e86c8c03e.png)

larsoner:
I think it would be nice to generate as many standard montage positions on the fly as possible rather than storing them in text files. We could basically change it so that they are calculated on the fly, and we only use the text file ones for validation (essentially they become test files, in principle you could `git mv` them but it's probably not worth doing so).

sappelhoff:
What do you mean by `git mv`ing them?

Other than that I completely agree with your point, however it seems like a major undertaking to implement this all at once.

Perhaps we could start by implementing the code to calculate the standard 10-20, 10-10, and 10-05 positions. Then whenever [`read_montage()`](https://martinos.org/mne/dev/generated/mne.channels.read_montage.html) is called with e.g., `kind='standard_1010'`, we would calculate the positions on the fly ... but when `kind='biosemi16'`, we fall back on the files until someone implements an on-the-fly calculation for biosemi.

Another point is that for all positions that we calculate ourselves (as opposed to reading from a file) will not need to be rescaled into MNE standards --> which will solve the two important rescaling issues I referenced in the original post.

agramfort:
objection on my side. Can  you open a PR to see what it would imply in
terms of code change?

jona-sassenhagen:
Loving the idea.

[...]

PR #8753, opened 2021-01-15
[EEG Discussion] Add aliases for electrodes to standard montages

tczhangzhi:
Add montage: the international 10-20 system for 62 channels |ZHANG Zhi|

As we know, there are many BCI dataset collected based on different montage standard. Some of them are included in the current framework, but others are not. I'm trying to make the process of importing the latest benchmark datasets more easily. For this PR, I'm adding support for SEED datasets, which are formatted based on the international 10-20 system for 62 channels.

> [OFFICIAL DOCUMENT]
>
> In the "Preprocessed_EEG" folder, there are files containing downsampled, preprocessed and segmented versions of the EEG data in Matlab (.mat file). The data was downsampled to 200Hz. A bandpass frequency filter from 0-75Hz was applied. We extracted the EEG segments corresponding to the duration of each movie. There are totally 45 .mat (Matlab) files, one for per experiment. Each subject performed the experiment three times with an interval of about one week. Each subject file contains 16 arrays. 15 arrays contain segmented preprocessed EEG data of 15 trials in one experiment (eeg_1~eeg_15, channel×data). A array name labels contains the label of the corresponding emotional labels (-1 for negative, 0 for neutral and +1 for positive). The detailed order of the channels is included in the dataset. The EEG cap according to the international 10-20 system for 62 channels is shown below...

#### Reference issue
None

Add montage for SEED dataset, which follows `the international 10-20 system for 62 channels`. The name, order, location of each channel have been checked carefully according to the dataset setting.

#### Additional information
If you think it is not necessary to follow the latest databases and want to keep the framework clean, feel free to close this PR. I will open a new project to build a high-level tool based on `mne`, to load datasets such as, DEEP, SEED and so on.


cbrnr:
Aren't these locations contained in the already existing `standard_1020` and/or `standard_1005` montages?

mne/channels/data/montages/standard_1020-62.elc
```diff
@@ -0,0 +1,136 @@
...
+P6
+P8
+PO7
+PO5
+PO3
+POz
+PO4
+PO6
+PO8
+Cb1
```
sappelhoff, line 132: As far as I know Cb1 and Cb2 are not part of the 10-10, 10-20, or 10-05 systems - what are these?

cbrnr: See the plot in my comment below.

sappelhoff:
I am not sure whether this is MNE or the proposed coordinates here.

I still think we should aim for computing the standard positions ourselves (e.g., using https://github.com/sappelhoff/eeg_positions) if we need *template* positions.

Adding more and more "pre-computed" positions feels non-transparent (where did they come from, under which assumptions were they computed, ...) and also like a maintenance burden.

Note that this is not an argument against your PR @tczhangzhi ... just some personal opinion on the side :)

cbrnr:
@sappelhoff I agree. We already have a lot of redundant montages, so I'm +💯 on integrating your template! Do you know the corresponding issue/PR where we've discussed this?

mne/channels/data/montages/standard_1020-62.elc
sappelhoff: Ah okay, I am still confused though, the positions of Cb2 here for example should rather be something like POO10h or OI2 according to the 10-05 system. Perhaps these are some "special" positions (exceptions to the 10-05 rule)?


https://github.com/sappelhoff/eeg_positions#10-05-system

image

cbrnr:
Also, I can't even think of a good reason to include anything else than *template* locations. Measured locations differ from individual to individual and that's why these subject-specific montages have to be acquired for each subject independently.

sappelhoff:
There is a big list, as this keeps coming up often:

- https://github.com/mne-tools/mne-python/issues/7141
- https://github.com/mne-tools/mne-python/issues/7472
- https://github.com/mne-tools/mne-python/issues/5269

cbrnr:
Let's use #7472 for this discussion.

mne/channels/data/montages/standard_1020-62.elc
tczhangzhi: My bad, I should make the difference clear. The following figure illustrates the layout of standard 10-20 systems which is close to its (64+3)-channel version:
montage-standard1020-for64
And the second figure represents the standard 10-20 system for (62+3) channel (I find it during an experiment on SEED dataset to compare our method with the state-of-the-art, and I notice it is widely compared):
montange-standard1020-for62

tczhangzhi:
> I am not sure whether this is MNE or the proposed coordinates here.
> 
> I still think we should aim for computing the standard positions ourselves (e.g., using https://github.com/sappelhoff/eeg_positions) if we need _template_ positions.
> 
> Adding more and more "pre-computed" positions feels non-transparent (where did they come from, under which assumptions were they computed, ...) and also like a maintenance burden.
> 
> Note that this is not an argument against your PR @tczhangzhi ... just some personal opinion on the side :)

Yes, this is MNE~

mne/channels/data/montages/standard_1020-62.elc
cbrnr: It looks like CB1 is identical to POO7 in the 10-05 system (https://github.com/sappelhoff/eeg_positions/blob/master/data/images/1005.png)

sappelhoff: > It looks like CB1 is identical to POO7 in the 10-05 system

yes, it certainly looks like it. 

@tczhangzhi you could perhaps verify by reading this article: https://www.sciencedirect.com/science/article/abs/pii/S1388245700005277

tczhangzhi: k! thanks!

tczhangzhi: Yes, m afraid you guys are right! Learn a lot, thanks for your time~ Then, this PR is only a subset of standard-1005, and by renaming the channel and use `on_missing` parameter we can achieve the same thing. So I will close this PR~ If you have any other questions feel free to reopen it, thanks.

sappelhoff: Cool that this issue could be solved and good luck

mmagnuski:
Could we add an alias for CB1 and CB2 in the localization files if these are identical to POO7 and POO8?
This naming is sometimes used and with current montage files users will not get position for these channels. I was recently using an open dataset that included these channels (`CB1` and `CB2`) but did not contain their position (although it had positions for other channels). I was about to drop these channels when I read this discussion.

sappelhoff:
Sounds good to me, but I'd also like to verify which manufacturer (or group?) Introduced this naming and what it stands for ... Also when is it used, and is it truly identical to poo7 etc 

It'd be good to know that for sure, wdyt?

sappelhoff:
Seems to be in the 5% article...

mmagnuski:
@sappelhoff Good idea. The dataset I was referring to comes from PREDiCT database (http://predict.cs.unm.edu/). I can contact database maintainers and ask about Cb1 and Cb2 present in their data.

tczhangzhi:
> @sappelhoff Good idea. The dataset I was referring to comes from PREDiCT database (http://predict.cs.unm.edu/). I can contact database maintainers and ask about Cb1 and Cb2 present in their data.

Good idea, then the API would be more user friendly. But I am not sure whether it should be provided by the template. Recently, I found that the channel names provided in some data sets are all capitalized, and there are also problems such as Cb1 and Cb2 that use historical names. I tend to suggest the template keep consistent with the standard, and provide an `alias` parameter for user to custom (or even note it in the document and let the user know they need to custom channel list while using `mne.create_info` before `set_montage`)

cbrnr:
@sappelhoff when we derive 10-05 template locations using your implementation, can we define aliases? Most channels have multiple names, and we don't want to create separate templates for identical locations that only differ in their name.

sappelhoff:
sure, I think that'd be an important feature. Can we start a list on synonymously used names?

- Cbz, Cb1, Cb2 (apparently also CBz, CB1, CB2 with capital B)
    - stands for "cerebellar", in the 5% article, synonymous to Iz, I1, I2
    - needs to be verified, because the Chatrian paper has a slightly different definition ... and @cbrnr and I thought that they would be POO7 and POO8
    -  There is also a bunch of other CB* electrodes, see figs below

- T3/T4 are synonymous with T7/T8 (see: https://core.ac.uk/download/pdf/144965483.pdf)

- T5/T6 are synonymous with T9/T10 (see: https://core.ac.uk/download/pdf/144965483.pdf)

- A1/A2 seem to be placed **on** the ears? (see figure in https://core.ac.uk/download/pdf/144965483.pdf)
    - does somebody have more precise information on that?
    - if they are placed **on** the ears, these are different from LPA and RPA (for info, see: https://www.fieldtriptoolbox.org/faq/how_should_i_report_the_positions_of_the_fiducial_points_on_the_head/)


Also there is much more, looking at these two figures from

> Chatrian GE, Lettich E, Nelson PL. Ten percent electrode system for topo-graphic studies of spontaneous and evoked EEG activity. Am J EEGTechnol 1985;25:83±92.

image



- M1/M2 is probably TP9/TP10 ?
    - standing for "mastoid" ... and TP9 / TP10 are typically pretty much on the mastoid

- T1/T2 are probably another synonym for T9/T10 ??? (like T5/T6 ???)


- Chatrian also includes ocular electrode positions:
    - So1 and So2 (superior occipital)
    - Io1 and Io2 (inferior ...)
    - Lo1 and Lo2 (lateral ...)
    - --> these are different by subject (pretty much like A1 and A2), because everyone's eyes (and ears) are slightly different and difficult to measure consistently ... or (for example) you might end up with an electrode too near or too far from the outer canthus of an eye 

PS: I am reopening and renaming ... would be nice to transfer this discussion to an issue...

drammock:
A1 and A2 are reference electrodes attached to earlobes by clips

sappelhoff:
> A1 and A2 are reference electrodes attached to earlobes by clips

do you also know what the A stands for, and where the convention comes from?

mmagnuski:
@sappelhoff A separate issue for discussion would be good. I think your summary list of synonymous channels would be a great introductory post there.

drammock:
> do you also know what the A stands for, and where the convention comes from?

probably "auricule" but that's a guess.  IDK the origin of that convention.

sappelhoff:
Thanks Dan, and I opened a new issue @mmagnuski #8765

sappelhoff:
Thanks Dan, and I opened a new issue @mmagnuski #8765

PR #8758, opened 2021-01-16
Replace master -> main

### LINKED ISSUE #7901: Rename `master` branch

cbrnr:
Fixes #7901.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -391,16 +391,16 @@ jobs:
             name: Deploy docs
             command: |
               set -e;
-              if [ "${CIRCLE_BRANCH}" == "master" ] || [ "${CIRCLE_BRANCH}" == "maint/0.22" ]; then
+              if [ "${CIRCLE_BRANCH}" == "main" ] || [ "${CIRCLE_BRANCH}" == "maint/0.22" ]; then
                 git config --global user.email "circle@mne.com";
                 git config --global user.name "Circle CI";
                 cd ~/mne-tools.github.io;
-                git checkout master
+                git checkout main
                 git remote -v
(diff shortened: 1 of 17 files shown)


agramfort:
scikit-learn moved to "main" so I guess we can do it too now :)

larsoner:
Thanks @cbrnr !

PR #8767, opened 2021-01-20
MAINT: Include statsmodels

larsoner:
Statsmodels appears to be on the anaconda pre site, let's see...


larsoner:
Nope, packaging problems apparently...

PR #8769, opened 2021-01-21
MNT: Brain refactoring

GuillaumeFavelier:
This PR modifies `Brain` so that it uses the internal `_renderer` as much as possible instead of importing from `_pyvista` or `vtk`.

*No feature is added.*


larsoner (approved):
This is 100% copy-paste + fix calls more or less, right? No functionality should change? If so +1 LGTM should be a useful refactoring.

Also will this `_silhouette` refactor help with the (potential) upcoming PR to have edges shown for the glass brain to improve visibility?

GuillaumeFavelier:
> This is 100% copy-paste + fix calls more or less, right?

Yes. I also got rid of `self._iren` and `self.remove_picking_callbacks` since we don't really need it.

> Also will this _silhouette refactor help with the (potential) upcoming PR to have edges shown for the glass brain to improve visibility?

I don't know yet but at least it's more DRY.

PR #8770, opened 2021-01-21
make large rotation penalty optional in chpi function

bloyl:
@larsoner

The utility of `_fit_coil_order_dev_head_trans` allowed for computing the device head transforms when the ordering of the HPIs in `dig` didn't match the ordering in the device cords.

At some point the cost function was modified to strongly favor small rotations. While often what you want, it isn't always what you want. for instance when running a phantom to check that the HPIs are still working after a PM visit.

This PR just makes this penalty optional.


mne/chpi.py
```diff
@@ -410,7 +410,7 @@ def _fit_chpi_quat(coil_dev_rrs, coil_head_rrs):
     return quat, gof
 
 
-def _fit_coil_order_dev_head_trans(dev_pnts, head_pnts):
+def _fit_coil_order_dev_head_trans(dev_pnts, head_pnts, pen_large_rots=True):
```
larsoner, line 413: maybe just `bias=True` is clear enough?

larsoner (commented):
Should we expose this publicly somehow / somewhere, too? Ideally things wouldn't break for you if I later went through and modified this function (e.g., if I had pushed such a variable name in the future). If we made some public interface we could avoid the coverage drop / lack of testing. A simple test comparing that, for reasonable data, the two match is probably good enough.

bloyl:
`bias` is fine. 

I don't have a strong opinion on making this public. In reality its a utility method that probably belongs in `transforms` not in chpi.

larsoner:
Thanks @bloyl !

PR #8771, opened 2021-01-22
MRG: Brain silhouette

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/8749#issuecomment-764676327 and adds a `silhouette` parameter to `Brain.__init__()`.

master | PR
---|---
image | image

*What would be even more helpful for this kind of mesh is ambient occlusion.*

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -1535,7 +1535,7 @@ add_data_kwargs : dict | None
 """
 docdict['brain_kwargs'] = """
 brain_kwargs : dict | None
-    Additional arguments to brain.__init__ (e.g.,
+    Additional arguments to the :class:`mne.viz.Brain` constructor (e.g.,
     ``dict(silhouette=True)``).
 """
 docdict['views'] = """
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -1533,6 +1533,11 @@ add_data_kwargs : dict | None
     Additional arguments to brain.add_data (e.g.,
     ``dict(time_label_size=10)``).
 """
+docdict['brain_kwargs'] = """
+brain_kwargs : dict | None
+    Additional arguments to brain.__init__ (e.g.,
+    ``dict(silhouette=True)``).
+"""
 docdict['views'] = """
 views : str | list
     View to use. Can be any of::
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2511,7 +2511,8 @@ def plot_vector_source_estimates(stc, subject=None, hemi='lh', colormap='hot',
                                  time_unit='s', show_traces='auto',
                                  src=None, volume_options=1.,
                                  view_layout='vertical',
-                                 add_data_kwargs=None, verbose=None):
+                                 add_data_kwargs=None, brain_kwargs=None,
+                                 verbose=None):
     """Plot VectorSourceEstimate with PySurfer.
 
     A "glass brain" is drawn and all dipoles defined in the source estimate
@@ -2587,6 +2588,7 @@ def plot_vector_source_estimates(stc, subject=None, hemi='lh', colormap='hot',
     %(src_volume_options)s
     %(view_layout)s
     %(add_data_kwargs)s
+    %(brain_kwargs)s
     %(verbose)s
 
     Returns
@@ -2614,7 +2616,7 @@ def plot_vector_source_estimates(stc, subject=None, hemi='lh', colormap='hot',
         vector_alpha=vector_alpha, cortex=cortex, foreground=foreground,
         size=size, scale_factor=scale_factor, show_traces=show_traces,
         src=src, volume_options=volume_options, view_layout=view_layout,
-        add_data_kwargs=add_data_kwargs)
+        add_data_kwargs=add_data_kwargs, brain_kwargs=brain_kwargs)
 
 
 @verbose
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -1678,7 +1678,8 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
                           time_unit='s', backend='auto', spacing='oct6',
                           title=None, show_traces='auto',
                           src=None, volume_options=1., view_layout='vertical',
-                          add_data_kwargs=None, verbose=None):
+                          add_data_kwargs=None, brain_kwargs=None,
+                          verbose=None):
     """Plot SourceEstimate.
 
     Parameters
@@ -1772,6 +1773,7 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
     %(src_volume_options)s
     %(view_layout)s
     %(add_data_kwargs)s
+    %(brain_kwargs)s
     %(verbose)s
 
     Returns
@@ -1828,7 +1830,8 @@ def plot_source_estimates(stc, subject=None, surface='inflated', hemi='lh',
         stc, overlay_alpha=alpha, brain_alpha=alpha, vector_alpha=alpha,
         cortex=cortex, foreground=foreground, size=size, scale_factor=None,
         show_traces=show_traces, src=src, volume_options=volume_options,
-        view_layout=view_layout, add_data_kwargs=add_data_kwargs, **kwargs)
+        view_layout=view_layout, add_data_kwargs=add_data_kwargs,
+        brain_kwargs=brain_kwargs, **kwargs)
 
 
 def _plot_stc(stc, subject, surface, hemi, colormap, time_label,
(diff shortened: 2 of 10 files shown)


GuillaumeFavelier:
For `stc_vec`:

image

GuillaumeFavelier:
Result looks good with SSAO on `stc_surf`:

PR | PR + ssao
--|--
image | image

<details>
<summary>Code snippet</summary>

```py
    def _enable_ssao(self, dataset):
        bounds = np.asarray(dataset.GetBounds())
        basic_pass = vtk.vtkRenderStepsPass()
        scene_size = np.linalg.norm(bounds[:3] - bounds[3:])
        ssao = vtk.vtkSSAOPass()
        ssao.SetRadius(0.1 * scene_size)  # comparison radius
        ssao.SetBias(0.001 * scene_size)  # comparison bias
        ssao.SetKernelSize(128)  # number of samples used
        ssao.BlurOff()  # do not blur occlusion
        ssao.SetDelegatePass(basic_pass)
        self.plotter.renderer.SetPass(ssao)
```

</details>

But I ran into visual artifacts on `stc_vector`:

image

Reference: https://blog.kitware.com/ssao/

larsoner:
Is it the combination of depth peeling + SSAO?

GuillaumeFavelier:
> Is it the combination of depth peeling + SSAO?

I think there is a depth peeling step in `vtkRenderStepsPass` by default

larsoner:
So maybe you need to disable `depth_peeling` when using SSO? In other words, maybe you have two depth peelings going on and they don't get along...

larsoner:
(i.e., don't call `brain.enable_depth_peeling()` / ensure it's off at the PyVista level however you need to do it, and let SSO do the depth peeling it needs to render)

GuillaumeFavelier:
Depth peeling is disabled by default in pyvista. I confirm that I still have the visual artifacts when the basic `vtkRenderStepsPass` is used and SSAO disabled. So the bug is related either to the `vtkRenderStepsPass`, my usage of it or my config. Do you reproduce those?

GuillaumeFavelier:
Avoiding `vtkRenderStepsPass` altogether and re-creating a full `vtkSequencePass` as described in [discourse.vtk.org](https://discourse.vtk.org/t/ssao-peculiar-effect-with-translucent-geometry/4410/2) fixed it for me!

GuillaumeFavelier:
After  f722e74 :

x | master | PR
--|--|--
surf | image | image
vec | image | image
vol | image  | image

larsoner:
Wow, that looks great! It makes all the renderings look much better in a realistic sense.

In some cases it makes them less usable, such as the volumetric case -- but I think this is mostly because the brain opacity is too high in that case. Can you see what something like [this example](https://mne.tools/dev/auto_tutorials/source-modeling/plot_beamformer_lcmv.html#sphx-glr-auto-tutorials-source-modeling-plot-beamformer-lcmv-py) looks like? On `master` it's



And locally on your PR I see:

Screenshot from 2021-01-22 09-54-43

GuillaumeFavelier:
Oh yeah I reproduce this on multi-renderers as well. It's trivial to solve the silhouette but SSAO gives this weird empty plot :thinking:

GuillaumeFavelier:
I have the same behaviour with the standard `Plotter` on a minimal example. My guess is that it is a limitation of the multi-renderer design of pyvista since those apperead with EDL too. Maybe better isolation like a multi-plotter system could solve this :thinking: 

<details>
<summary>MWE</summary>

```py
import numpy as np
import vtk
import pyvista as pv


def _enable_ssao(plotter, dataset=None, idx=None):
    lightsP = vtk.vtkLightsPass()
    opaqueP = vtk.vtkOpaquePass()
    translucentP = vtk.vtkTranslucentPass()
    volumeP = vtk.vtkVolumetricPass()

    collection = vtk.vtkRenderPassCollection()
    collection.AddItem(lightsP)

    # opaque passes
    ssaoCamP = vtk.vtkCameraPass()
    ssaoCamP.SetDelegatePass(opaqueP)

    if dataset is None:
        ssaoP = vtk.vtkSSAOPass()
        ssaoP.SetRadius(30)
        ssaoP.SetBias(0.1)
        ssaoP.SetKernelSize(32)
        ssaoP.SetBlur(True)
    else:
        bounds = np.asarray(dataset.GetBounds())
        scene_size = np.linalg.norm(bounds[:3] - bounds[3:])
        ssaoP = vtk.vtkSSAOPass()
        ssaoP.SetRadius(0.1 * scene_size)  # comparison radius
        ssaoP.SetBias(0.001 * scene_size)  # comparison bias
        ssaoP.SetKernelSize(128)  # number of samples used
        ssaoP.BlurOff()  # do not blur occlusion
    ssaoP.SetDelegatePass(ssaoCamP)
    collection.AddItem(ssaoP)

    # translucent and volumic passes
    ddpP = vtk.vtkDualDepthPeelingPass()
    ddpP.SetTranslucentPass(translucentP)
    ddpP.SetVolumetricPass(volumeP)
    collection.AddItem(ddpP)

    # finally overlays
    overP = vtk.vtkOverlayPass()
    collection.AddItem(overP)

    sequence = vtk.vtkSequencePass()
    sequence.SetPasses(collection)

    camP = vtk.vtkCameraPass()
    camP.SetDelegatePass(sequence)

    if idx is None:
        plotter.renderer.SetPass(camP)
    else:
        plotter.renderers[idx].SetPass(camP)


p = pv.Plotter(shape=(2, 1))
p.subplot(0, 0)
p.add_mesh(pv.Sphere())
_enable_ssao(p)
p.subplot(1, 0)
p.add_mesh(pv.Cone())
_enable_ssao(p)
p.show()
```

</details>

GuillaumeFavelier:
I confirm that better isolation fixes this. I built a custom Qt Plotter that contains separated instances of `QtInteractor` and now it SSAO works:

image

GuillaumeFavelier:
I think this can move forward without SSAO. It makes it less impactful/realistic though.

SSAO will need `MultiPlotter` and this implies heavy refactoring of `_pyvista` and `Brain`.

larsoner:
Agreed, let's move forward with `silhouette`. SSAO is a separable issue.

Can you modify `plot_visualize_stc`, `plot_beamformer_lcmv.py`, and `plot_mixed_source_space_inverse.py` which should all benefit in principle from the silhouette? That way we can compare `main` and this PR

mne/viz/_3d.py
```diff
@@ -1898,6 +1898,7 @@ def _plot_stc(stc, subject, surface, hemi, colormap, time_label,
     if backend in ['pyvista', 'notebook']:
         kwargs["show"] = False
         kwargs["view_layout"] = view_layout
+        kwargs["silhouette"] = True
```
larsoner, line 1901: This looks like it defaults to `True`, but really it should default to `False` and probably be exposed in `SourceEstimate.plot` and related functions. However I start to worry abit about parameter explosion in `SourceEstimate.plot`, maybe we need a `brain_kwargs : dict | None` argument that allows for passing these sorts of things. Eventually `ssao` might be in there for example.

GuillaumeFavelier: Similarly to `add_data_kwargs` then

mne/utils/docs.py
```diff
@@ -1533,6 +1533,11 @@
     Additional arguments to brain.add_data (e.g.,
     ``dict(time_label_size=10)``).
 """
+docdict['brain_kwargs'] = """
+brain_kwargs : dict | None
+    Additional arguments to brain.__init__ (e.g.,
```
larsoner, line 1538:
```suggestion
    Additional arguments to the :class:`mne.viz.Brain` constructor (e.g.,
```

mne/viz/_brain/_brain.py
```diff
@@ -405,6 +410,18 @@ def __init__(self, subject_id, hemi, surf, title=None,
...
         self._annots = {'lh': list(), 'rh': list()}
         self._layered_meshes = {}
+        # default values for silhouette
+        self._silhouette = {
+            'color': self._bg_color,
+            'line_width': 2,
+            'alpha': alpha,
+            'decimate': 0.9,
+        }
+        if isinstance(silhouette, dict):
```
larsoner, line 420:
```suggestion
        _validate_type(silhouette, (dict, bool), 'silhouette')
        if isinstance(silhouette, dict):
```

larsoner (commented):
These all look good to me / better than `master`:







Other than some minor gripes LGTM +1 for merge

larsoner:
And don't forget `latest.inc` update. Maybe start a bullet-point list for `mne.viz.Brain` enhancements since I have a feeling there will be a few we will be able to group together for this release?

larsoner (approved):

PR #8774, opened 2021-01-22
WIP, BUG: Make plot_evoked() not plot GFP for single channel

hoechenberger:
Fixes #8772

PR #8775, opened 2021-01-23
MRG, BUG, DOC: derive GFP from average reference for EEG

### LINKED ISSUE #8772: plot_evoked(gfp=True) with single-channel data behaves incorrectly
#### Problem
Plotting the GFP of data with only a single sensor via `plot_evoked(gfp=True)` produces a GFP, although the GFP – describing the spatial standard deviation – should not exist.

#### Steps to reproduce
```python
import os
import mne
                                    'sample_audvis_raw.fif')
              'visual/right': 4}
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7, event_id=event_dict,
                    preload=True)
evoked = epochs['auditory/left'].average()

evoked.pick('EEG 001').plot(gfp=True)
```

#### Expected results
- either no GFP trace, or
- a GFP constantly at zero, or
- a warning, or
- an error message

#### Actual results
GFP is displayed as the absolute value of the sensor signal.
Screen Shot 2021-01-22 at 16 27 48

#### Additional information
Using `master`.

hoechenberger:
The reason for #8772 is simply that we always calculated the GFP incorrectly, i.e. without re-referencing sensor signals to average first. When doing that, I now correctly get a flat GFP (0 at all time points) for single-channel recordings:

Screen Shot 2021-01-23 at 09 43 44

This also means that since the introduction of this feature in 0.15 or so, our GFP plots haven't been correct in terms of absolute amplitude (which isn't too bad since it's difficult to judge the absolute amplitudes from the plot anyway)

Closes #8772, #8774

### DIFF
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -267,7 +267,7 @@ def _plot_evoked(evoked, picks, exclude, unit, show, ylim, proj, xlim, hline,
         raise RuntimeError('Currently only single axis figures are supported'
                            ' for interactive SSP selection.')
 
-    _check_option('gfp', gfp, [True, False, 'only', 'power', 'power-only'])
+    _check_option('gfp', gfp, [True, False, 'only', 'rms', 'rms-only'])
 
     scalings = _handle_default('scalings', scalings)
     titles = _handle_default('titles', titles)
@@ -428,7 +428,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
             # Set amplitude scaling
             D = this_scaling * data[idx, :]
             _check_if_nan(D)
-            gfp_only = gfp == 'only' or gfp == 'power-only'
+            gfp_only = gfp == 'only' or gfp == 'rms-only'
             if not gfp_only:
                 chs = [info['chs'][i] for i in idx]
                 locs3d = np.array([ch['loc'][:3] for ch in chs])
@@ -477,8 +477,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                 if gfp in [True, 'only']:
                     this_gfp = D.std(axis=0, ddof=0)
                     label = 'GFP'
-                elif gfp in ['power', 'power-only']:
-                    # RMS
+                elif gfp in ['rms', 'rms-only']:
                     this_gfp = np.linalg.norm(D, axis=0)
                     label = 'RMS'
 
@@ -680,26 +679,26 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
         The axes to plot to. If list, the list must be a list of Axes of
         the same length as the number of channel types. If instance of
         Axes, there must be only one channel type plotted.
-    gfp : bool | 'only' | 'power' | 'power-only'
-        Plot the global field power (GFP) or the root mean square (RMS) power
-        of the data. GFP is the standard deviation of signals across channels.
+    gfp : bool | 'only' | 'rms' | 'rms-only'
+        Plot the global field power (GFP) or the root mean square (RMS) of the
+        data. GFP is the standard deviation of signals across channels.
         This is equivalent to the RMS of an average-referenced signal.
 
         - ``True``
-            plot the GFP and the traces for all channels
+            Plot the GFP and the traces for all channels.
         - ``'only'``
-            plot the GFP, but omit the individual channel traces
-        - ``'power'``
-            plot the RMS and the traces for all channels
-        - ``'power-only'``
-            plot the RMS, but omit the individual channel traces
+            Plot the GFP, but omit the individual channel traces.
+        - ``'rms'``
+            Plot the RMS and the traces for all channels.
+        - ``'rms-only'``
+            Plot the RMS, but omit the individual channel traces.
 
         The color of the GFP/RMS trace will be green if
         ``spatial_colors=False``, and black otherwise.
 
         .. versionchanged:: 0.23
-           Plot GFP (and not RMS) by default. Add ``'power'`` and
-           ``'power-only'`` options.
+           Plot GFP (and not RMS) by default. Add ``'rms'`` and
+           ``'rms-only'`` options.
 
     window_title : str | None
         The title to put at the top of the figure.
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -479,7 +479,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                     label = 'GFP'
                 elif gfp in ['power', 'power-only']:
                     # RMS
-                    this_gfp = np.sqrt((D ** 2).mean(axis=0))
+                    this_gfp = np.linalg.norm(D, axis=0)
                     label = 'RMS'
 
                 gfp_color = 3 * (0.,) if spatial_colors is True else (0., 1.,
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -685,11 +685,14 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
         of the data. GFP is the standard deviation of signals across channels.
         This is equivalent to the RMS of an average-referenced signal.
 
-        - ``True``: plot the GFP and the traces for all channels
-        - ``'only'``: plot the GFP, but omit the individual channel traces
-        - ``'power'``: plot the RMS and the traces for all channels
-        - ``'power-only'``: plot the RMS, but omit the individual channel
-           traces
+        - ``True``
+            plot the GFP and the traces for all channels
+        - ``'only'``
+            plot the GFP, but omit the individual channel traces
+        - ``'power'``
+            plot the RMS and the traces for all channels
+        - ``'power-only'``
+            plot the RMS, but omit the individual channel traces
 
         The color of the GFP/RMS trace will be green if
         ``spatial_colors=False``, and black otherwise.
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -479,7 +479,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                         this_gfp = D.std(axis=0, ddof=0)
                         label = 'GFP'
                     else:
-                        this_gfp = np.linalg.norm(D, axis=0)
+                        this_gfp = np.sqrt((D ** 2).mean(axis=0))
                         label = 'RMS'
 
                 gfp_color = 3 * (0.,) if spatial_colors is True else (0., 1.,
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -474,9 +474,13 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                     line_list[-1].set_pickradius(3.)
 
             if gfp:  # 'only' or boolean True
+                # GFP is the standard deviation of the signal across all
+                # channels for each time point:
+                #
+                # GFP(t) = √{[∑ (xᵢ(t) - x̅(t))²] / N}
+                this_gfp = D.std(axis=0, ddof=0)
                 gfp_color = 3 * (0.,) if spatial_colors is True else (0., 1.,
                                                                       0.)
-                this_gfp = np.sqrt((D * D).mean(axis=0))
                 this_ylim = ax.get_ylim() if (ylim is None or this_type not in
                                               ylim.keys()) else ylim[this_type]
                 if gfp_only:
@@ -674,8 +678,11 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
         the same length as the number of channel types. If instance of
         Axes, there must be only one channel type plotted.
     gfp : bool | 'only'
-        Plot GFP in green if True or "only". If "only", then the individual
-        channel traces will not be shown.
+        Plot the global field power (GFP), i.e. the standard deviation of
+        signals across channels. If ``True``,  the traces for all channels will
+        be plotted, too; if ``'only'``, the individual channel traces will
+        not be shown. The color of the GFP trace will be green if
+        ``spatial_colors=False``, and black otherwise.
     window_title : str | None
         The title to put at the top of the figure.
     spatial_colors : bool
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -266,8 +266,8 @@ def _plot_evoked(evoked, picks, exclude, unit, show, ylim, proj, xlim, hline,
     if axes is not None and proj == 'interactive':
         raise RuntimeError('Currently only single axis figures are supported'
                            ' for interactive SSP selection.')
-    if isinstance(gfp, str) and gfp != 'only':
-        raise ValueError('gfp must be boolean or "only". Got %s' % gfp)
+
+    _check_option('gfp', gfp, [True, False, 'only', 'power', 'power-only'])
 
     scalings = _handle_default('scalings', scalings)
     titles = _handle_default('titles', titles)
@@ -428,7 +428,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
             # Set amplitude scaling
             D = this_scaling * data[idx, :]
             _check_if_nan(D)
-            gfp_only = (isinstance(gfp, str) and gfp == 'only')
+            gfp_only = gfp == 'only' or gfp == 'power-only'
             if not gfp_only:
                 chs = [info['chs'][i] for i in idx]
                 locs3d = np.array([ch['loc'][:3] for ch in chs])
@@ -473,12 +473,15 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                                 linewidth=0.5)[0])
                     line_list[-1].set_pickradius(3.)
 
-            if gfp:  # 'only' or boolean True
-                # GFP is the standard deviation of the signal across all
-                # channels for each time point:
-                #
-                # GFP(t) = √{[∑ (xᵢ(t) - x̅(t))²] / N}
-                this_gfp = D.std(axis=0, ddof=0)
+            if gfp:
+                if gfp in [True, 'only']:
+                    this_gfp = D.std(axis=0, ddof=0)
+                    label = 'GFP'
+                elif gfp in ['power', 'power-only']:
+                    # RMS
+                    this_gfp = np.sqrt((D ** 2).mean(axis=0))
+                    label = 'RMS'
+
                 gfp_color = 3 * (0.,) if spatial_colors is True else (0., 1.,
                                                                       0.)
                 this_ylim = ax.get_ylim() if (ylim is None or this_type not in
@@ -494,7 +497,7 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
                                          zorder=3, alpha=line_alpha)[0])
                 ax.text(times[0] + 0.01 * (times[-1] - times[0]),
                         this_gfp[0] + 0.05 * np.diff(ax.get_ylim())[0],
-                        'GFP', zorder=4, color=gfp_color,
+                        label, zorder=4, color=gfp_color,
                         path_effects=gfp_path_effects)
             for ii, line in zip(idx, line_list):
                 if ii in bad_ch_idx:
@@ -677,12 +680,24 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
         The axes to plot to. If list, the list must be a list of Axes of
         the same length as the number of channel types. If instance of
         Axes, there must be only one channel type plotted.
-    gfp : bool | 'only'
-        Plot the global field power (GFP), i.e. the standard deviation of
-        signals across channels. If ``True``,  the traces for all channels will
-        be plotted, too; if ``'only'``, the individual channel traces will
-        not be shown. The color of the GFP trace will be green if
+    gfp : bool | 'only' | 'power' | 'power-only'
+        Plot the global field power (GFP) or the root mean square (RMS) power
+        of the data. GFP is the standard deviation of signals across channels.
+        This is equivalent to the RMS of an average-referenced signal.
+
+        - ``True``: plot the GFP and the traces for all channels
+        - ``'only'``: plot the GFP, but omit the individual channel traces
+        - ``'power'``: plot the RMS and the traces for all channels
+        - ``'power-only'``: plot the RMS, but omit the individual channel
+           traces
+
+        The color of the GFP/RMS trace will be green if
         ``spatial_colors=False``, and black otherwise.
+
+        .. versionchanged:: 0.23
+           Plot GFP (and not RMS) by default. Add ``'power'`` and
+           ``'power-only'`` options.
+
     window_title : str | None
         The title to put at the top of the figure.
     spatial_colors : bool
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -698,7 +698,6 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
 
         .. versionchanged:: 0.23
            Plot GFP for EEG instead of RMS. Label RMS traces correctly as such.
-
     window_title : str | None
         The title to put at the top of the figure.
     spatial_colors : bool
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -267,7 +267,7 @@ def _plot_evoked(evoked, picks, exclude, unit, show, ylim, proj, xlim, hline,
         raise RuntimeError('Currently only single axis figures are supported'
                            ' for interactive SSP selection.')
 
-    _check_option('gfp', gfp, [True, False, 'only', 'rms', 'rms-only'])
+    _check_option('gfp', gfp, [True, False, 'only'])
 
     scalings = _handle_default('scalings', scalings)
     titles = _handle_default('titles', titles)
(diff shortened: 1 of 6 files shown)


hoechenberger:
Actually due to the no linearity in the term, even the relative changes in GFP were incorrect all along, if I'm not mistaken.

hoechenberger:
I discussed this with @agramfort and therefore want to provide literature evidence why GFP **must** use the average-referenced signals.

TL;DR: GFP is based on average-referenced signal.

When using GFP, commonly two papers are cited: 
- D. Lehmann, W. Skrandies, Reference-free identification of components of checkerboard-evoked multichannel potential fields, Electroencephalography and Clinical Neurophysiology, Volume 48, Issue 6, 1980, Pages 609-621, https://doi.org/10.1016/0013-4694(80)90419-8
- Dietrich Lehmann, Wolfgang Skrandies, Spatial analysis of evoked potentials in man—a review, Progress in Neurobiology,
Volume 23, Issue 3, 1984, Pages 227-250, https://doi.org/10.1016/0301-0082(84)90003-0

In the 1980 paper, the term "global field power" does not yet occur; instead, it is called the "root-mean square", and they introduce the term "power":

Screen Shot 2021-01-23 at 12 17 58

In the 1984 paper, the term "global field power" shows up, albeit only for the "reference-free" calculation; they mention how the perks of using the average-referenced signal in the paragraph below the formula:

Screen Shot 2021-01-23 at 12 32 58

In: Skrandies, W. Global field power and topographic similarity. Brain Topogr 3, 137–141 (1990), https://doi.org/10.1007/BF01128870, the GFP is explicitly named as being based on the average-referenced signal:

Screen Shot 2021-01-23 at 12 22 50

More recently, in a 2008 review tutorial by Murray MM, Brunet D, Michel CM. Topographic ERP analyses: a step-by-step tutorial review. Brain Topogr. 2008 Jun;20(4):249-64. doi: 10.1007/s10548-008-0054-5, the GFP is presented as average-reference based too:

Screen Shot 2021-01-23 at 12 24 12

Lastly, in a 2011 paper by Brunet et. al, GFP is also defined as average-reference-based (Brunet, Denis, Micah M. Murray, and Christoph M. Michel. "Spatiotemporal analysis of multichannel EEG: CARTOOL." Computational intelligence and neuroscience 2011,  https://doi.org/10.1155/2011/813870)


Screen Shot 2021-01-23 at 12 41 20

cbrnr:
What's the difference to #8774 (which I just merged)?

hoechenberger:
#8774 was not supposed to be merged anymore ;) Because I figured it doesn't address the underlying problem: the incorrect calculation of the GFP, which is addressed in this PR

hoechenberger:
Ok I reverted the merge of #8774 

@agramfort Your call on this one :)

hoechenberger:
Here a comparison between the approach currently used in `master` and the one implemented in this PR:

```python
# %%
import os
import numpy as np

import mne


                                    'sample_audvis_raw.fif')
              'visual/right': 4}
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7, event_id=event_dict,
                    preload=False)
evoked = epochs['auditory/left'].average()
evoked.pick_types(meg=False, eeg=True)

fig, ax = plt.subplots(figsize=(12, 7))

ax.plot(np.sqrt((evoked.data ** 2).mean(axis=0)),
        label='master', ls='--')
ax.plot(evoked.data.std(axis=0, ddof=0),
        label='avg-ref', ls='--')

ax.set_xlabel('Sample', fontsize='large', fontweight='bold')
ax.set_ylabel('Signal amplitude in V', fontsize='large', fontweight='bold')
ax.legend()
ax.set_title('Global Field Power', fontsize='x-large', fontweight='bold')
```

Screen Shot 2021-01-23 at 20 55 57

agramfort:
to add some info to the debate. In
https://neuroimage.usc.edu/brainstorm/Tutorials/Averaging I read:

The green line represents the global field power (*GFP*), ie the sum of the
square of all the sensors at each time point. This measure is sometimes
used to identify transient or stable states in ERP/ERF. You can hide it
with the display options menu *Extra > Show GFP*.

can you check in brainstorm, fieldtrip or eeglab how it's done?

Remark: this is no such things as common average reference on MEG as you
don't need a reference :)

>

hoechenberger:
> The green line represents the global field power (*GFP*), ie the sum of the square of all the sensors at each time point.

But this would be something entirely differently again! Unless there's a bug in their docs. They don't even take the root of the value, so it's actual **power**. GFP as outlined in all the publications I cited above is actually **not** a measure of power, as one can see by the resulting unit (µV, not µV^2).


> can you check in rainstorm, fieldtrip or eeglab how it's done?

I don't remember a built-in function in EEGLAB; I always had to use a plugin / additional toolbox. On the mailing list, they recommend to simply calculate the standard deviation across channels, as I'm doing in this PR. This automatically uses the average-referenced signal.

Re Fieldtrip, Brainstorm I can do the research.

We could also discuss simply offering different "GFP" calculation methods.

hoechenberger:
> Remark: this is no such things as common average reference on MEG as you don't need a reference :)

Yep I'm aware of that, however you could still subtract the mean 😅

hoechenberger:
@agramfort The [FT docs](https://www.fieldtriptoolbox.org/reference/ft_globalmeanfield/) are helpful regarding my previous-to-last comment (GFP not actually being a power):

>  FT_GLOBALMEANFIELD calculates global mean field amplitude or power of input data
> 
>  Use as
>    [gmf] = ft_globalmeanfield(cfg, data)
> 
>  The data should be organised in a structure as obtained from the
>  FT_TIMELOCKANALYSIS function. The configuration should be according to
>  FT_PREPROCESSING function. The configuration should be according to
> 
>    cfg.method    = string, determines whether the amplitude or power should be calculated (see below, default is 'amplitude', can be 'power')
>    cfg.channel   = Nx1 cell-array with selection of channels (default = 'all'),
>                             see FT_CHANNELSELECTION for details
> 
> This function calculates the global mean field power, or amplitude,
 as described in:
 Lehmann D, Skrandies W. Reference-free identification of components of
 checkerboard-evoked multichannel potential fields. Electroencephalogr Clin
 Neurophysiol. 1980 Jun;48(6):609-21. PubMed PMID: 6155251.
>
 > Please note that to calculate what is clasically referred to as Global
 Mean Field Power, cfg.method must be 'amplitude'. The naming implies a
 squared measure but this is not the case.


That aside, this is how it's calculated:

```Matlab
switch cfg.method
  case 'amplitude'
    dataout.avg = std(dataout.avg,1);
  case 'power'
    dataout.avg = std(dataout.avg,1).^2;
end
```
https://github.com/fieldtrip/fieldtrip/blob/3ba8125c1e809a8a9e6ed4f6f286b8fb3357d28a/ft_globalmeanfield.m#L95-L100

I'm not really sure what `dataout.avg` is **exactly**, but it's clear for the `amplitude` GFP (the one I'm implementing in this PR, which is what is classically known as GFP), they simply calculate the standard deviation, i.e. use an average-referenced signal.

hoechenberger:
@agramfort I'd propose to introduce an `Evoked.gfp()` method that gets parameters to control its behavior (return power or amplitude, use average-reference signal or not).

hoechenberger:
ERPLAB uses the standard deviation, too:

```Matlab
        MGFP_data   = std(bindata(:,:,j));
```
https://github.com/lucklab/erplab/blob/13ce4c384b52b472713c586e2454902fac57e818/functions/mgfperp.m#L77

hoechenberger:
Updated tutorial:
https://24697-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_eeg_erp.html#global-field-power-gfp

cc @drammock

tutorials/evoked/plot_eeg_erp.py
```diff
@@ -108,6 +110,59 @@
...
+
+###############################################################################
+# It is difficult to find out the absolute values of the GFP in the plots we
+# produced above. By passing ``gfp='only'``, the channel traces are removed,
+# and we can focus on the GFP exclusively. Although we don't plot any channel
+# traces, we still pass ``spatial_colors=True`` to get the GFP in a nice
+# black-and-gray color scheme – without this parameter, it would be
+# bright-green.
+
+evoked_car.plot(gfp='only', spatial_colors=True)
```
drammock, line 145:
```suggestion
# To plot the GFP by itself you can pass ``gfp='only'`` (this makes it easier
# to read off the GFP data values, because the scale is aligned):

evoked_car.plot(gfp='only')
```

[nitpick] nothing wrong with showing the green here IMO; it doesn't look that bad and it's informative to see it both ways.

More substantively, this makes me wonder why we don't add a second y scale on the right side of the axes that aligns to the zero value of the GFP trace (something for a separate PR).

tutorials/evoked/plot_eeg_erp.py
```diff
@@ -108,6 +110,59 @@
...
+# That's it!
+#
+# Finally, let's try to manually reproduce the GFP plot from above!
+
+fix, ax = plt.subplots()
+ax.plot(evoked_car.times, gfp * 1e6, color='black')
+ax.fill_between(evoked_car.times, gfp * 1e6, color='lightgray')
+ax.set_xlabel('Time (s)')
+ax.set_ylabel('GFP (µV)')
+ax.set_title('EEG')
```
drammock, line 164: [nitpick] more concise

```suggestion
# As stated above, the GFP is the population standard deviation of the signal
# across channels. To compute it manually, we can leverage
# the fact that `evoked.data() <mne.Evoked.data>` is a NumPy array:

gfp = evoked_car.data.std(axis=0, ddof=0)

# Reproducing the plot style from above:
fig, ax = plt.subplots()
ax.plot(evoked_car.times, gfp * 1e6, color='black')
ax.fill_between(evoked_car.times, gfp * 1e6, color='lightgray')
ax.set(xlabel='Time (s)', ylabel('GFP (µV)', title='EEG')
```

drammock (commented):
some suggestions for making the tutorial a little tighter. Mostly nitpicks except the point about fixing ylim for the two plots that we're meant to compare

tutorials/evoked/plot_eeg_erp.py
hoechenberger: > More substantively, this makes me wonder why we don't add a second y scale on the right side of the axes that aligns to the zero value of the GFP trace (something for a separate PR).

I would love either that, or we create a 2nd axis for the GFP. Additional scale is probably much more space-preserving.

hoechenberger:
Updated tutorial: https://24702-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_eeg_erp.html#global-field-power-gfp

hoechenberger:
Tutorial is looking good now!!
https://24703-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_eeg_erp.html#global-field-power-gfp

hoechenberger:
BESA [uses](ftp://www.besa.de/be/besa.de/demonstrations_and_tutorials/BESA%20Research%20Tutorial%201%20-%20Introduction%20to%20Source%20Analysis.pdf) "the sum of squares, normalized to 100%"

Screen Shot 2021-01-24 at 12 46 10

hoechenberger:
BrainVision Analyzer appears to offer "RMS" and "GFP", the latter being called "reference-free" – which I assume implies "average-referenced"

Screen Shot 2021-01-24 at 12 51 20

agramfort:
at least we should not call GFP what is actually a RMS. Besides there seems to be a need for both GFP and RMS. So it's not one or the other.

any thought on how to expose this?

hoechenberger:
> at least we should not call GFP what is actually a RMS. Besides there seems to be a need for both GFP and RMS. So it's not one or the other.
> 
> any thought on how to expose this?

How about:
```python
class Evoked(...):
    def get_field(field_type: Literal['gfp', 'rms', 'power] = 'gfp'):
        ...
```

agramfort:
I was more thinking in the plotting code. Also see that plot_field method
is something very different

>

hoechenberger:
> Also see that plot_field method is something very different

Yeah, I changed my proposal to `get_field_strength()` :)

> I was more thinking in the plotting code. 

```python
plot_evoked(...,
            show_field_strength: bool | 'only' = False,
            field_strength_type: 'gfp' | 'rms' | 'power' = 'gfp'):
```

cbrnr:
Is this specific to `Evoked`? I think you could compute GFP with `Epochs` or even `Raw` - or doesn't this make sense?

hoechenberger:
> Is this specific to `Evoked`? I think you could compute GFP with `Epochs` or even `Raw` - or doesn't this make sense?

You could do it, but I've never seen anyone doing it. I would assume that the data is simply too noisy?

hoechenberger:
Bumping this again.

@agramfort WDYT of this proposal:

> ```python
> plot_evoked(...,
>             show_field_strength: bool | 'only' = False,
>             field_strength_type: 'gfp' | 'rms' | 'power' = 'gfp'):
> ```

We would deprecate `gfp=False` map `gfp` to the new parameters during the deprecation cycle?

Any other EEG people here? @sappelhoff to the rescue!! (maybe? 🦄)

larsoner:
> We would deprecate gfp and map gfp to the new parameters during the deprecation cycle.

I would not introduce a new parameter just for this. I would just fix the behavior of `gfp=True` to do what GFP should actually be (std over sensors, as a bugfix) and allow a new option (enh) that actually gives the old behavior `gfp='power'` which just does a power calculation. Even though it's a bit of a misnomer, it's less code churn for people for what is probably in many cases a minor change in behavior.

sappelhoff:
I have never used GFP, but eric's comment https://github.com/mne-tools/mne-python/pull/8775#issuecomment-766951111 sounds reasonable to me. And thanks for investigating @hoechenberger, this reads a bit like a crime novel ... although perhaps not with a really bad crime as the focus :)

hoechenberger:
> I would not introduce a new parameter just for this. I would just fix the behavior of `gfp=True` to do what GFP should actually be (std over sensors, as a bugfix) and allow a new option (enh) that actually gives the old behavior `gfp='power'` which just does a power calculation. Even though it's a bit of a misnomer, it's less code churn for people for what is probably in many cases a minor change in behavior.

But how would you handle the `gfp='only'` case, then?

larsoner:
I would do `'only'`, and either `'power-only'` or `'only-power'`

hoechenberger:
Doc artifact:
https://24730-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Evoked.html#mne.Evoked.plot

Should we add RMS to the tutorial, too?

hoechenberger:
> Should we add RMS to the EEG tutorial, too?

Otherwise this is good to merge from my end.

hoechenberger:
@agramfort My RMS formula deviates from what FT uses. They calculate the power based on the average-referenced signal, while here I'm using our "old" formula which doesn't apply any re-referencing. WDYT?

mne/viz/evoked.py
```diff
@@ -473,10 +473,17 @@ def _plot_lines(data, info, picks, fig, axes, spatial_colors, unit, units,
...
                     line_list[-1].set_pickradius(3.)
 
-            if gfp:  # 'only' or boolean True
+            if gfp:
+                if gfp in [True, 'only']:
+                    this_gfp = D.std(axis=0, ddof=0)
+                    label = 'GFP'
+                elif gfp in ['power', 'power-only']:
+                    # RMS
+                    this_gfp = np.sqrt((D ** 2).mean(axis=0))
```
agramfort, line 482:
```suggestion
                    this_gfp = np.linalg.norm(D, axis=0)
```
it avoids allocating D**2

mne/viz/evoked.py
```diff
@@ -673,9 +680,27 @@ def plot_evoked(evoked, picks=None, exclude='bads', unit=True, show=True,
         The axes to plot to. If list, the list must be a list of Axes of
         the same length as the number of channel types. If instance of
         Axes, there must be only one channel type plotted.
-    gfp : bool | 'only'
-        Plot GFP in green if True or "only". If "only", then the individual
-        channel traces will not be shown.
+    gfp : bool | 'only' | 'power' | 'power-only'
```
cbrnr, line 683: The values for this parameter are pretty confusing. It seems like this parameter actually controls two more or less independent options:

1. Plot GFP (aka RMS) or GPF**2 (aka MS) or don't plot at all
2. If something is plotted in option 1, either plot individual traces (`True`) or not (`False`)

Also, it should be RMS and not RMS power - this wording is used in several locations.

hoechenberger: > The values for this parameter are pretty confusing. It seems like this parameter actually controls two more or less independent options:

Yes, that's why I wanted to introduce a second param, but I thought @larsoner, you, and @sappelhoff thought this was not a good idea (per https://github.com/mne-tools/mne-python/pull/8775#issuecomment-766951111)?

mne/viz/evoked.py
hoechenberger: Nice one!!!!

mne/viz/evoked.py
```diff
@@ -266,8 +266,8 @@ def _plot_evoked(evoked, picks, exclude, unit, show, ylim, proj, xlim, hline,
     if axes is not None and proj == 'interactive':
         raise RuntimeError('Currently only single axis figures are supported'
                            ' for interactive SSP selection.')
-    if isinstance(gfp, str) and gfp != 'only':
-        raise ValueError('gfp must be boolean or "only". Got %s' % gfp)
+
+    _check_option('gfp', gfp, [True, False, 'only', 'power', 'power-only'])
```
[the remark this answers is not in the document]
hoechenberger: Yes absolutely. I just tried to follow @larsoner's initial suggestion, hence also my confusion in https://github.com/mne-tools/mne-python/pull/8775#issuecomment-767173351

mne/viz/evoked.py
hoechenberger: > GPF**2 (aka MS) 

This is something we need to discuss, because there's again 2 options here.

The previous behavior (which was erroneously called "GFP") was to calculate the RMS. What **you** proposed -- squaring the actual GFP -- would **also** be **some kind** of RMS, but not the same as the one we had before. Because it would now be based on the avg-referenced signal (because GFP is based on the avg-referenced signal), while the previous approach would **not** re-reference anything.

cbrnr: Yes, but at that time I did not know that there would be two more options involving "only"...

mne/viz/evoked.py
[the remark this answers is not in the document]
hoechenberger: Done in 86fbe7e

mne/viz/evoked.py
hoechenberger: My original proposal can be found in https://github.com/mne-tools/mne-python/pull/8775#issuecomment-766947428

hoechenberger: @cbrnr Please note that in `master`, we already have `gfp=True` and `gfp='only'` to select between GFP + channels and GFP only. So adding `'rms'` and `'rms-only'` seems natural, although I do agree this is not a very clean API.

hoechenberger:
So @cbrnr is unhappy with the API, specifically that the `gfp` parameter combines two functionalities (**whether** to show the GFP and **how** to show it). My original proposal involved using **two** parameters to disentangle this.

I'd like to ask everyone again to share their opinion on how to proceed.

drammock:
> I'd like to ask everyone again to share their opinion on how to proceed

My preference is to treat this as a bugfix: change how you compute GFP and leave the API alone.

If people aren't happy with that then I'd deprecate `gfp` and add 2 new params: `show_field_strength` (True, False, 'only') and `field_strength_type` ('gfp', 'rms')

hoechenberger:
> My preference is to treat this as a bugfix: change how you compute GFP and leave the API alone.
> 
> If people aren't happy with that then I'd deprecate `gfp` and add 2 new params: `show_field_strength` (True, False, 'only') and `field_strength_type` ('gfp', 'rms')

I'm open two both approaches. My preference from a "clean API" POV is clearly the 2nd proposal; from a "let's stay backward-compatible" POV, proposal 1 wins. I'll implement whichever of the two you decide upon. :) (In fact, proposal 1 is already implemented)

cbrnr:
👍 for treating it as a bugfix and changing the API later if necessary.

larsoner:
-1 for the clean API. The benefit is that the names are ideal, but the drawback is the associated amount of code churn for people for a change that is unlikely to make a huge difference for them in practice. `gfp` has been there as on option for over 5 years so this could be a lot. There are times we force a lot of people to change a lot of code (e.g., with `montage` in the last few years) but this is to overcome some big strategic, usability, and maintainability issues -- abandoning `gfp` by contrast would be a change to improve our naming, which doesn't seem worth it to me.

I'm fine with the bugfix approach that leaves the user-facing `gfp` options as they are in `main` and just fixes the calculation. And if people want the old way back, we can add some new parameter like `gfp_mode` or `field_strength_type` that makes it clear that it changes what "gfp" does and makes it not really a gfp anymore (hence our naming ends up imperfect, but I can live with that).

hoechenberger:
Ok, in this case, this PR should be good to merge!

larsoner:
> Ok, in this case, this PR should be good to merge!

It looks like this PR still implements new options, and the consensus seems to be that it shouldn't do this and it should just fix the existing GFP calculation

agramfort:
GFP with the argument of average referencing does not make sense for MEG

that's why Brainstorm which is historically an MEG software also does not
center
the data. I think both options are legit

>

hoechenberger:
> GFP with the argument of average referencing does not make sense for MEG that's why Brainstorm which is historically an MEG software also does not center the data. I think both options are legit

Yep, I see it the same way.

hoechenberger:
> that's why Brainstorm which is historically an MEG software also does not center the data.

FT does center the data, however :) But I do agree this might not make too much sense for MEG.

hoechenberger:
Although -- and maybe this bit of information can help for reconciliation a little -- I have noticed in some (limited) testing that for MEG, the difference between RMS and GFP was barely noticeable to inexistent in the few plots I produced. Maybe this was just per coincidence, but this is what I saw. In fact, at first I thought I was running my tests with the wrong branch or something; then I picked only EEG channels, and suddenly I **did** see a large difference between RMS and GFP.

drammock:
> GFP with the argument of average referencing does not make sense for MEG

Maybe we should only center the data for non-MEG channel types.  That is even more like a bugfix: "GFP was computed correctly for MEG but incorrectly for EEG, now we triage depending on channel type"

cbrnr:
I'm not a MEG expert, but isn't MEG already reference-free? EEG is not, that's why something like average reference must be used. But if you already have reference-free data, subtracting the average again shouldn't hurt.

hoechenberger:
Ok so the consensus (?) seems to be:

- don't add new parameter values
- for EEG data, spit out GFP
- for MEG data, spit out RMS (old behavior)

Yes? (This is getting a bit tiresome 🤯)

larsoner:
Sounds good to me

cbrnr:
@hoechenberger yes, for now this is OK. If you have time I'd be interested if there is a difference in GFP for MEG data without and with average reference.

hoechenberger:
> . If you have time I'd be interested if there is a difference in GFP for MEG data without and with average reference.

Question is, which data should I use? So far I used the MNE `sample` subject, but this is probably by no means representative.

If sb has an idea and could provide me with a folder full of data files, I'd happily run this comparison. But I don't have time to start digging around & finding suitable datasets :)

hoechenberger:
One last question before I finalize this:

Do we keep labeling the trace for MEG as "GFP", or should we change this to "RMS"?
+1 for RMS from me, but I'm not sure if it would be confusing for users if after updating MNE the "GFP" trace is suddenly called "RMS"…

larsoner:
+1 for changing the name, in the docs for `gfp : ` you can mention that it's calculated differently for MEG and EEG

cbrnr (approved):

agramfort (approved):

agramfort:
@drammock @larsoner merge if happy

thx @hoechenberger

hoechenberger:
Fixing the tests now, will push shortly & you can merge once CI goes green

hoechenberger:
I couldn't get `viz/tests/test_evoked.py::test_plot_evoked_reconstruct` to pass locally, and I'm not sure why. Would appreciate any help.

larsoner:
@hoechenberger the values changed for EEG (rightly so) and so we can just adjust the limits in the test, will push a commit. Good to know that the test is sufficiently sensitive to changes in values :)

hoechenberger:
> will push a commit

Thank you, @larsoner!

hoechenberger:
Tutorials look good:
https://24771-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_20_visualize_evoked.html#plotting-signal-traces

https://24771-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_eeg_erp.html#global-field-power-gfp

API docs look good:
https://24771-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Evoked.html?highlight=evoked%20plot#mne.Evoked.plot

This is good to merge once green.

hoechenberger:
> Tutorials look good:
> https://24771-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_20_visualize_evoked.html#plotting-signal-traces

Whoops. No, It doesn't. The amplitude of the RMS is super big… the heck? *scratches head*

hoechenberger:
Seems `np.linalg.norm()` uses the Frobenius norm:
Screen Shot 2021-01-27 at 14 44 23

This doesn't calculate the mean of the squared values before taking their square-root – right? Is there another norm we could use? Otherwise I'll just go back to the original approach, `np.sqrt((D**2).mean(axis=0))`

larsoner:
you can use norm and divide by sqrt(n) or use the old formula, those two should be equivalent

hoechenberger:
@larsoner I reverted to the old formula because it's more intuitive to grasp when looking at the code. If you think using the approach based on the Frobenius norm would have strong advantages, I can switch to that too!

larsoner (approved):

hoechenberger:
All good now.
https://24776-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/evoked/plot_20_visualize_evoked.html#id1

larsoner:
Thanks @hoechenberger !

hoechenberger:
@larsoner Can you try to backport if you have a minute? Otherwise I can do it later today.

larsoner:
done

hoechenberger:
You're an angel <3 Or maybe a saint? who knows! Thank you!

PR #8777, opened 2021-01-23
enforce abs path

### LINKED ISSUE #8693: Several questions about source localization workflow
Hello,
I am a student working with MNE and ran into several issues while working on source localization. I was wondering if anyone had any suggestions.

Firstly, I created my BEM surfaces using freesurfer's mri_watershed, and used these to try to make a source space. However, it appears as though the BEM surfaces are not scaled or located properly with respect to the brain, and the brain sticks out the back of the head. Here is plot_alignment with the BEM surfaces and the pial surface:
Screen Shot 2021-01-01 at 2 53 48 AM
Here is the T1.mgz with the BEM inner skull in freeview:
Screen Shot 2021-01-01 at 3 10 15 AM
This is not as far off as the pial surface, but the BEM is still quite clearly misaligned, particularly in the back of the head, and this leads to many sources being discarded since they are outside the inner skull.
I was wondering if anyone had any ideas on what is going wrong here? Is this something I would have to fix on the freesurfer side, or is there a quick fix i can implement in my MNE python code?

My second question is more simple: I want to get an average source estimate for each brain region in the atlas I use for my volume source space. I was wondering if there was a way to do this automatically, or if there was a way to output the number of sources that correspond to each label region, so that using these numbers and the order of labels in the atlas I could just average groups of rows in the final source estimate matrix.

Thank You!

agramfort:
closes #8693

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1632,6 +1632,7 @@ def _prepare_env(subject, subjects_dir):
     _validate_type(subject, "str")
 
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
+    subjects_dir = op.abspath(subjects_dir)  # force use of an absolute path
     if not op.isdir(subjects_dir):
         raise RuntimeError('Could not find the MRI data directory "%s"'
                            % subjects_dir)
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1633,6 +1633,7 @@ def _prepare_env(subject, subjects_dir):
 
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     subjects_dir = op.abspath(subjects_dir)  # force use of an absolute path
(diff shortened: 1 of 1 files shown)


mne/bem.py
```diff
@@ -1632,6 +1632,7 @@ def _prepare_env(subject, subjects_dir):
     _validate_type(subject, "str")
 
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
+    subjects_dir = op.abspath(subjects_dir)  # force use of an absolute path
```
hoechenberger, line 1635: Nice! Should this maybe be combined with `expanduser` too?

hoechenberger (approved):

hoechenberger (approved):

hoechenberger:
Thanks @agramfort!

PR #8780, opened 2021-01-24
MRG, BUG: fix IO or read_ica_eeglab

### LINKED ISSUE #8755: Read ICA EEGLab: problems with AMICA decomposition
### (possible) Bug

`read_ica_eeglab()` appears to not be able to read in AMICA (note: *not* `runica`) decompositions from an EEGLab .set file.

The error upon trying to instantiate ICA:

```
  AssertionError                            Traceback (most recent call last)
 <ipython-input-10-a6f5a4e0501a> in <module>
       5 from mne import preprocessing
       6 
 ----> 7 ica = preprocessing.read_ica_eeglab('full_filepath')
       8 
       9 
 
 ~/opt/anaconda3/envs/mne/lib/python3.8/site-packages/mne/preprocessing/ica.py in read_ica_eeglab(fname)
    2754 
    2755     n_ch = len(ica.ch_names)
 -> 2756     assert eeg.icaweights.shape == (n_components, n_ch)
    2757     # When PCA reduction is used in EEGLAB, runica returns
    2758     # weights= weights*sphere*eigenvectors(:,1:ncomps)';
 
 AssertionError: 
```

### Steps to reproduce
 ICA was performed with EEGLab's [AMICA](https://github.com/japalmer29/amica/wiki/AMICA) (not runica) with PCA reduction to rank correct for interpolated channels.

Looking at the data in EEGLab, `ICAweights = n_components x n_pcs`, and `ICAsphere = n_pcs x n_channels`. I’m guessing `read_ica_eeglab` is assuming the `icaweights` matrix is already multiplied by `icasphere`.

### Expected results
`read_ica_eeglab` should correctly read in pca reduced IC decompositions from AMICA (and possibly other algorithms -- this might want to be tested, such as FASTICA).

### Minimal reproducible example
[61 channel recording (.set/.fdt)](https://github.com/mne-tools/mne-python/files/5822735/MNE_TestData.zip)

agramfort:
attempt to close #8755

my debug script using this branch.

```
from scipy import linalg

import mne
from mne.viz.topomap import _find_topomap_coords

ica = mne.preprocessing.read_ica_eeglab('ICAamica_MNE_test.set')
# ica.info = raw.info
# ica._update_ica_names()
# ica.plot_components()

unmixing = ica.icaweights @ ica.icasphere
mixing = linalg.pinv(unmixing)

pos = _find_topomap_coords(ica.info, picks=None)
plt.close('all')
mne.viz.plot_topomap(mixing[:, 7], pos=pos)
plt.show()
```

here are the ICA topos I get using EEGLAB

Screenshot 2021-01-24 at 15 20 30

It's quite different from what I get using ica.plot_components() or using mne.viz.plot_topomap manually
as above.

any idea @cbrnr ?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -49,7 +49,7 @@ Bugs
 
 - Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Alex Gramfort`_)
 
-- Fix bug with :func:`mne.preprocessing.read_ica_eeglab` with reading AMICA decompositions (:gh:`8780` by `Alex Gramfort`_ and `Eric Larson`_)
+- Fix bug with :func:`mne.preprocessing.read_ica_eeglab` when reading decompositions using PCA dimensionality reduction (:gh:`8780` by `Alex Gramfort`_ and `Eric Larson`_)
 
 - Fix bug with ``replace`` argument of :meth:`mne.Report.add_bem_to_section` and :meth:`mne.Report.add_slider_to_section` (:gh:`8723` by `Eric Larson`_)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -49,6 +49,8 @@ Bugs
 
 - Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Alex Gramfort`_)
 
+- Fix bug with :func:`mne.preprocessing.read_ica_eeglab` with reading AMICA decompositions (:gh:`8780` by `Alex Gramfort`_ and `Eric Larson`_)
+
 - Fix bug with ``replace`` argument of :meth:`mne.Report.add_bem_to_section` and :meth:`mne.Report.add_slider_to_section` (:gh:`8723` by `Eric Larson`_)
 
 - Allow sEEG channel types in :meth:`mne.Evoked.plot_joint` (:gh:`8736` by `Daniel McCloy`_)
--- a/mne/io/eeglab/eeglab.py
+++ b/mne/io/eeglab/eeglab.py
@@ -323,7 +323,7 @@ class RawEEGLAB(BaseRaw):
                             ' the .set file contains epochs.' % eeg.trials)
 
         last_samps = [eeg.pnts - 1]
-        info, eeg_montage, update_ch_names = _get_info(eeg, eog=eog)
+        info, eeg_montage, _ = _get_info(eeg, eog=eog)
 
         # read the data
         if isinstance(eeg.data, str):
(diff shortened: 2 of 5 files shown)


cbrnr:
Not really. If the output of AMICA has already been loaded into EEGLAB (potentially using [this helper function](https://github.com/japalmer29/amica/blob/master/loadmodout15.m)), the format should be identical to Infomax ICA. What do the topoplots look like when plotting with MNE?

agramfort:
Screenshot 2021-01-26 at 10 33 30

see this dropbox folder: https://www.dropbox.com/sh/6lcn5sph5ww8psm/AAD_tsYa7PsSwLMUedtyr0Oda?dl=0

larsoner:
This is what I came up with as a more minimal example of the problem:
```
import numpy as np
import mne
mat = mne.externals.pymatreader.read_mat('ICAamica_MNE_test.set')['EEG']
unmixing = mat['icaweights'] @ mat['icasphere']
mixing = np.linalg.pinv(unmixing, rcond=1e-3)
np.testing.assert_allclose(mat['icawinv'], mixing, rtol=1e-4)
```
It's the relationship that we rely on and it fails here. I do not see any other fields in the `mat` that can help us modify things. So someone will have to dig into the EEGLAB code to figure out why these are inconsistent, or what we should be using instead. Would perhaps make a good EEGLAB bug report, but I didn't make this `.set` and don't know how so it probably shouldn't be me!

In the meantime we can `pinv(icawinv)` and use this instead of `icaweights @ mat['icasphere']` in cases where it doesn't match, which is what I just pushed a commit to do. Now the `plot_components` above produces the following (along with a warning about a mismatch):

Screenshot from 2021-01-26 11-49-29

cbrnr:
Very nice @larsoner! Indeed someone more familiar with EEGLAB? @arnodelorme, do you know why `icaweights @ icasphere` is not equal to `pinv(icawinv)`? Or is it just an erroneous file that's causing this issue?

arnodelorme:
if you remove ICA components, then `icaweights` is not square and  `icaweights @ icasphere` is not longer equal to `pinv(icawinv)`. This is the most likely explanation.

agramfort:
thx @arnodelorme for the feedback and @larsoner for taking over.

can you update what's new and then +1 for MRG on my side

cbrnr:
@larsoner maybe mention why this happens in the warning? Otherwise LGTM!

doc/changes/latest.inc
```diff
@@ -49,6 +49,8 @@ Bugs
 
 - Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Alex Gramfort`_)
 
+- Fix bug with :func:`mne.preprocessing.read_ica_eeglab` with reading AMICA decompositions (:gh:`8780` by `Alex Gramfort`_ and `Eric Larson`_)
```
cbrnr, line 52: I don't think this is a problem specific to AMICA; all ICA decompositions with principal components removed should be affected.

agramfort (commented):
@cbrnr merge when green and happy.

thanks

cbrnr:
@agramfort one test fails: https://github.com/mne-tools/mne-python/pull/8780/checks?check_run_id=1775321151#step:12:3153

cbrnr:
Thanks @agramfort and @larsoner!

Issue #8773, opened 2021-01-22
Adding overlap keyword to make_fixed_length_epochs


adam2392:
#### Describe the new feature or enhancement
I am trying to use time-frequency analysis that is present in mne without converting to arrays first. There is no way to do this using the ``Raw`` object, and one must convert to ``Epochs``, with each epoch "representing" the window that will have tfr methods applied. E.g. I want to compute the multitaper FFT, and/or morlet wavelet transform to get an ``AverageTfr`` object for multiple frequency bands, alpha, beta, gamma, etc.

Normally, one would like some overlap of these windows. For example, if you do multitpaer FFT, then you usually use a sliding window with some overlap (e.g. 1 sec window with 0.5 sec step).

In https://mne.tools/stable/generated/mne.make_fixed_length_events.html#mne.make_fixed_length_events
there is an overlap keyword.

However, in `make_fixed_legnth_epochs` there is not. 
https://mne.tools/stable/generated/mne.make_fixed_length_epochs.html#mne.make_fixed_length_epochs

#### Describe your proposed implementation
Is it possible to add the `overlap` keyword argument here?

#### Describe possible alternatives
I'm still not quite sure what the optimal way to apply tfr analysis to a Raw object. Obviously one can always convert to arrays, but then they lose the structure present in the mne data structures. I would prefer staying with the data structures mne has (e.g. Epochs for TFR). If what I said up there is not the optimal way to construct TFR then please let me know.

#### Additional comments
I can update an existing example to show this functionality off if needed.

agramfort:
ok to add overlap param to make_fixed_length_epochs 

yet I am not sure what you aim to obtain for TFR on raw. You want a spectrogram?

adam2392:
Yep I'm essentially after a spectrogram.

agramfort:
one idea now we refactored the raw.plot would be to add plot of spectrogram
in each "horizontal" band of the plot.

I've seen this in certain software for sleep data.

did you try scipy.signal.spectrogram?

>

adam2392:
> one idea now we refactored the raw.plot would be to add plot of spectrogram in each "horizontal" band of the plot. 
This sounds interesting to me. Would you elaborate what the "horizontal band" mean? like a frequency band?

And do you mean we could add a compute_spectrogram function for `Raw`, where we can specify say a frequency band, and/or a list of frequencies with an `average` kwarg?

> I've seen this in certain software for sleep data. did you try scipy.signal.spectrogram?
[...]

PR #8781, opened 2021-01-24
[WIP] Allow all tfr_functions to operate on Raw object

adam2392:
#### Reference issue
Addresses tfr discussion in: #8773 

1. Allows `tfr` functions to operate on ``Raw`` object that returns and `AverageTFR` object.
2. ~~Adds an API to `AverageTFR` for plotting a spectrogram and averaging frequencies between two ranges~~

#### Additional information
~~API is draft just to show what it would be like. I have not implemented 2. yet, and can take that out if you think it is not necessary/useful.~~

Related discussion on discourse: https://mne.discourse.group/t/is-there-a-way-to-obtain-the-time-points-from-time-frequency-stft/2682/3


adam2392:
@agramfort this is something along what I would propose for what we disucced in #8773

adam2392:
@agramfort just wanted to check in with you to see if this is going in the right direction? The main point is to allow tfr functions to operate on `Raw` objects with a disclaimer in the docstring about size/runtime.

Then a few questions regarding unimplemented functions:
- `average_freqs`: the intention is to provide a way to average within the frequency range too for things like computing "gamma band power", or something like that.
- `plot_spectrogram`: the intention here is to provide a plotting mechanism for plotting channels X time with frequency power (possibly in dB) as the color

^ Are these two functionalities desirable in the `_BaseTFR` class?

agramfort:
>
> @agramfort <https://github.com/agramfort> just wanted to check in with
> you to see if this is going in the right direction? The main point is to
> allow tfr functions to operate on Raw objects with a disclaimer in the
> docstring about size/runtime.
>
> Then a few questions regarding unimplemented functions:
>
>    - average_freqs: the intention is to provide a way to average within
>    the frequency range too for things like computing "gamma band power", or
>    something like that.
>
>
for this usecase I would raw.filter(...).apply_hilbert(envelope=False)
it's much faster than an STFT and averaging. It's also much easier to
explain.


>
>    - plot_spectrogram: the intention here is to provide a plotting
>    mechanism for plotting channels X time with frequency power (possibly in
>    dB) as the color
>
> ^ Are these two functionalities desirable in the _BaseTFR class?
>

how would it look like? you would display minutes of data? would you have a
way to scroll?

adam2392:
>The main point is to allow tfr functions to operate on Raw objects with a disclaimer in the docstring about size/runtime. Then a few questions regarding unimplemented functions: - average_freqs: the intention is to provide a way to average within the frequency range too for things like computing "gamma band power", or something like that.
> for this usecase I would raw.filter(...).apply_hilbert(envelope=False) it's much faster than an STFT and averaging. It's also much easier to explain.

I might not be an expert in time-freq signal processing, but aren't there some advantages to using FFT like techniques and then averaging (https://mne.tools/stable/auto_examples/time_frequency/plot_time_frequency_simulated.html), or is Hilbert Transform envelope considered the best for showing things like frequency bands? 

I might be lacking some basic working knowledge here... Lmk if instead you would prefer I post a more general question on the Discourse.

> - plot_spectrogram: the intention here is to provide a plotting mechanism for plotting channels X time with frequency power (possibly in dB) as the color ^ Are these two functionalities desirable in the _BaseTFR class?
> how would it look like? you would display minutes of data? would you have a way to scroll?
 
It would just be a static heatmap

agramfort:
>
> I might not be an expert in time-freq signal processing, but aren't there
> some advantages to using FFT like techniques and then averaging (
> https://mne.tools/stable/auto_examples/time_frequency/plot_time_frequency_simulated.html),
> or is Hilbert Transform envelope considered the best for showing things
> like frequency bands?
>
do your own experiments if you have doubts.

> I might be lacking some basic working knowledge here... Lmk if instead you
> would prefer I post a more general question on the Discourse.
>
>
>    - plot_spectrogram: the intention here is to provide a plotting
>    mechanism for plotting channels X time with frequency power (possibly in
>    dB) as the color ^ Are these two functionalities desirable in the _BaseTFR
>    class?
>    how would it look like? you would display minutes of data? would you
>    have a way to scroll?
>
> It would just be a static heatmap
>

honestly I would encourage you to develop this tool for your own needs and
once doing
a few iterations on your side  see what's best for MNE-Python.

adam2392:
Okay sounds good! Thanks @agramfort for answering questions. I've updated the PR now to just reflect the title of allowing TFR funcs to operate on Raw objects. Lmk if this is good?

mne/time_frequency/tfr.py
```diff
@@ -872,6 +882,12 @@ def tfr_multitaper(inst, freqs, n_cycles, time_bandwidth=4.0,
     Notes
     -----
     .. versionadded:: 0.9.0
+
+    If the function is run on a ``Raw`` object, then it one should be
```
agramfort, line 886:
```suggestion
    If the function is run on a ``Raw`` object, then it should be
```

mne/time_frequency/tfr.py
```diff
@@ -728,6 +730,14 @@ def tfr_morlet(inst, freqs, n_cycles, use_fft=False, return_itc=True, decim=1,
     mne.time_frequency.tfr_array_multitaper
     mne.time_frequency.tfr_stockwell
     mne.time_frequency.tfr_array_stockwell
+
+    Notes
+    -----
+    If the function is run on a ``Raw`` object, then it one should be
```
agramfort, line 736:
```suggestion
    If the function is run on a ``Raw`` object, then it should be
```

agramfort (commented):
you miss a what's new entry and I would strongly recommend to update a tutorial on time frequency to document this option

mne/time_frequency/tfr.py
```diff
@@ -2173,13 +2189,17 @@ def combine_tfr(all_tfr, weights='nave'):
...
     from ..evoked import Evoked
-    if not isinstance(inst, (BaseEpochs, Evoked)):
-        raise TypeError('inst must be Epochs or Evoked')
+    from ..io import BaseRaw
+    if not isinstance(inst, (BaseEpochs, Evoked, BaseRaw)):
+        raise TypeError('inst must be Epochs or Evoked or Raw')
     if isinstance(inst, BaseEpochs):
         data = inst.get_data()
+    elif isinstance(inst, BaseRaw):
+        # make data into 3D to play nice with epochs x ch x time
```
agramfort, line 2201: should you also do this:

https://github.com/mne-tools/mne-python/blob/8bc55d7435ce047c37465f2b5c173517491d7002/mne/time_frequency/tfr.py#L2204

I am thinking that maybe you can use the same code path for Evoked and Raw? They basically both contain 2D data arrays

adam2392: Like what's done here: c14d080?

adam2392:
Added an extension to an existing tutorial. Lmk if you think this is good, or if I should add more?

agramfort:
@adam2392 can you share some rendered doc page when you have one on CI? thx

Issue #8782, opened 2021-01-25
Add a describe method to annotations


rob-luke:
#### Describe the new feature or enhancement
I think the new describe method for raw objects in #8760 is great. Could we add one for annotations too?

#### Describe your proposed implementation
When I get a new dataset I often want to know what annotations are available, how many occurrences of each, etc. Similar to `raw.describe()` could we return this as a table with this basic info?

#### Describe possible alternatives
People may already have their own snippets for this. If others think this is too basic to be required then lets close this issue.

PR #8760, opened 2021-01-17
Add raw.describe() method

cbrnr:
I think it would be useful to add a little `raw.describe()` method, which outputs some basic information for each channel contained in the object. Currently, I'm printing

- channel name
- channel type
- unit
- five number summary (min, q1, median, q3, max)

Do you think this is worth adding? If yes, this should be fleshed out of course (e.g. what do we want to print) and printed in a nice table-like manner.

PR #8783, opened 2021-01-25
Add describe method for annotations

rob-luke:
#### Reference issue
#8782

Adds `describe()` to annotations. Completely inspired by #8760

#### Additional information
Just an initial attempt, if people are interested I can expand. If its too basic then will close.


cbrnr:
I think the standard `repr` is sufficient for most use cases, e.g.

```
<Annotations | 30 segments: T0 (15), T1 (8), T2 (7)>
```

I'd rather have a `to_data_frame` method for analyzing annotations in more detail, because in contrast to `raw.describe` I think what measures are interesting depends a lot on the specific use case.

agramfort:
I agree with @cbrnr 
I would add a to_data_frame method to Annotations and then it seems it's just a groupby('description').count()

hoechenberger:
I'd like to second the idea of allowing DataFrame export, and potentially tweaking the string representation a bit

agramfort (approved):

agramfort:
thx @rob-luke

rob-luke:
Thanks for the improved approach.

PR #8757, opened 2021-01-16
MRG, MAINT: new website theme [skip github][skip azp][circle front]

drammock:
I've been poking away at the sphinx_pydata_theme off and on for the last few days.  Here's a WIP pull request that makes the switch.  All of you should feel free to jump in and push a commit or two if you see something I haven't fixed yet and know what the fix should be (if you don't know what the fix is, at least add it to the list below it so it doesn't get missed).

Once the circle build finishes you'll see there are some obvious remaining issues, but so far nothing that seems insurmountable. Here's a partial list:

- [x] ~~some CSS properties I've tried to override but aren't working (e.g., our "sidebar" callout boxes need a bgcolor or border to distinguish them from the main text)~~ *this turned out to be an issue with dev vs stable version of the theme: the CSS color variable names had changed*
- [x] ~~`navbar_align` property in conf.py isn't working~~ *only available in dev version of theme*
- [x] navbar has some weird spacing of items *opened https://github.com/pandas-dev/pydata-sphinx-theme/issues/302*
- [x] The homepage footer is offscreen / there is a bunch of whitespace after the carousel *this problem appears for @drammock locally but not on the circle build*
- [x] what to do with the homepage sidebar boxes (currently they aren't on the side and aren't boxed)
- [x] any page that still has a `.. contents::` directive should have it removed, since the theme includes page contents in the right sidebar
- [x] ~~haven't yet tried to incorporate sphinx_multiversion, so currently our version dropdown is missing.~~ *sphinx-multiversion is for **building** multiple versions all at once (which I don't think we want?) and in my local tests it failed to build anything older than 0.17.*

closes #8184
closes #8785

PR #8787, opened 2021-01-25
MAINT: Fix doc date

larsoner:
Back in https://github.com/mne-tools/mne-python/pull/7532 I broke the `Last updated` tag at the bottom of the website, and it has read "Last updated on 2020-03-27." ever since. With this PR I now get "Last updated 01/25/2021, 11:32 AM EST" and each person should get the equivalent datetime in their own timezone.

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -82,7 +82,7 @@ Or, switch to documentation for the <a href="https://mne.tools/stable">current s
         <li><a href="https://www.au.dk/"><img class="institution" src="{{ pathto('_static/institution_logos/Aarhus.png', 1) }}" title="Aarhus Universitet" alt="Aarhus Universitet"/></a></li>
         <li><a href="https://www.uni-graz.at/"><img class="institution" src="{{ pathto('_static/institution_logos/Graz.jpg', 1) }}" title="Karl-Franzens-Universität Graz" alt="Karl-Franzens-Universität Graz"/></a></li>
       </ul>
-      <p class="text-center text-muted small">&copy; Copyright 2012-2020, MNE Developers. Last updated on 2020-03-27.</p>
+      <p class="text-center text-muted small">{{ copyright }}</p>
     </div></div></div>
   </footer>
   <script src="https://mne.tools/versionwarning.js"></script>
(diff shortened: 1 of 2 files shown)


drammock:
I have a fix for this already in #8757 that just uses one line in `conf.py` 

https://github.com/mne-tools/mne-python/pull/8757/files?file-filters%5B%5D=.py#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R47

and one line in `_templates/layout.html` 

https://github.com/mne-tools/mne-python/pull/8757/files?file-filters%5B%5D=.html&hide-deleted-files=true#diff-ee53b5b8ace34aceb7fd64220d4892db846fdf385d75b096628272fb1d02da8dR44

larsoner:
> EDIT oh, I see you're actually doing the same thing, just with a bunch of (unnecessary?) javascript to put the last updated timestamp into the user's local time zone?

Yes the difference is that the javascript localizes to both the user's time zone and their local date expression, which is to me much nicer than UTC's very long formatting in time zone that is often not the viewer's

larsoner:
> UTC's very long formatting in time zone that is often not the viewer's

I guess the long formatting only really matters if you include the time. I think it's nice to have in there as a dev but maybe users don't need it. And I guess we could localize the date at least, but this is probably a bit overkill since the YYYY-MM-DD format is probably okay, and weird stuff like the site showing an updated-date that is ahead of your actual date should be rare. So I can just change this to be your solution for a quicker merge to master and backport if you prefer @drammock ...

drammock:
I'm starting to see the value of having the timestamp in there, for devs.  WDYT about YYYY-MM-DD hh:mm:ss (basically follow ISO 8601 but replace `T` with a space) and always render time in UTC (no timezones)?

larsoner:
Especially if it's for devs then having the timezone is useful, though. If I see some date in UTC I don't know the conversion immediately and I can never remember if it changes during the year due to daylight savings. Even if I teach myself the rules and internalize them then every dev needs to keep track of this if they want to actually know the date/time that something changed.

Another hint that this formatting is useful/standard is if you look at [this commit](https://github.com/mne-tools/mne-python/commit/baaf9e33f2b54f6a40454a8aa146577bde1d6f58) on GitHub and mouseover the date that says "4 hours ago", for example on my machine it says "Jan 25, 2021 8:04 AM EST" not UTC. I'm assuming this is localized to each viewer.

larsoner:
... looks like on GitHub this is done with:
```
<relative-time datetime="2021-01-25T13:04:35Z" class="no-wrap">Jan 25, 2021</relative-time>
```
and after the page loads something (presumably JavaScript?) adds a `title="Jan 25, 2021, 8:04 AM EST"` to get this mouseover effect. So we are really doing something similar to GitHub, just actually replacing the text itself rather than adding a `title`.

drammock (approved):
OK you've convinced me to accept the extra code for localized time.  FWIW I'm seeing `01/25/2021 10:50 AM CST`  which may be the "standard format" in the country I live in but is definitely not my preferred way of seeing dates/times (I prefer YYYY-MM-DD and 24hr times).

larsoner:
Looks like GitHub uses probably `{dateStyle: "medium", timeStyle: "long"}` because when I use this I get the same formatting as GitHub. If you don't like this formatting you can complain to your country's standards-setting organizations :)

PR #8789, opened 2021-01-25
[MRG] ENH EGI MFF reader:  populate info['dig']

christianbrodbeck:
#### Reference issue
Partly addresses #8038

Adds digitization points to `.info` for `read_raw_egi` and `read_evokeds_mff`

### DIFF
--- a/mne/io/egi/egimff.py
+++ b/mne/io/egi/egimff.py
@@ -12,6 +12,7 @@ import numpy as np
 from .events import _read_events, _combine_triggers
 from .general import (_get_signalfname, _get_ep_info, _extract, _get_blocks,
                       _get_gains, _block_r)
+from .._digitization import DigPoint
 from ..base import BaseRaw
 from ..constants import FIFF
 from ..meas_info import _empty_info, create_info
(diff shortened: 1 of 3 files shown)


mne/io/egi/tests/test_egi.py
```diff
@@ -115,6 +115,7 @@ def test_io_egi_mff():
                            test_scaling=False,  # XXX probably some bug
                            )
     assert raw.info['sfreq'] == 1000.
+    assert len(raw.info['dig']) == 132  # 128 eeg + 1 ref + 3 cardinal points
```
larsoner, line 118: `ref` should be the first of the digitized EEG points, and its position should show up in `info['chs'][ii]['loc'][3:6]` (first three elements are electrode loc, next three are ref loc). Can you implement and ensure those conventions are enforced here?

mne/io/egi/tests/test_egi.py
```diff
@@ -115,6 +115,11 @@ def test_io_egi_mff():
                            test_scaling=False,  # XXX probably some bug
                            )
     assert raw.info['sfreq'] == 1000.
+    assert len(raw.info['dig']) == 132  # 1 ref + 128 eeg + 3 cardinal points
+    assert raw.info['dig'][0]['ident'] == 129  # reference channel nr
+    ref_loc = raw.info['dig'][0]['r']
+    for i in pick_types(raw.info, eeg=True):
+        assert_equal(raw.info['chs'][i]['loc'][3:6], ref_loc)
```
christianbrodbeck, line 122: @larsoner is that what you had in mind?

mne/io/egi/tests/test_egi.py
```diff
@@ -115,6 +115,11 @@ def test_io_egi_mff():
                            test_scaling=False,  # XXX probably some bug
                            )
     assert raw.info['sfreq'] == 1000.
+    assert len(raw.info['dig']) == 132  # 1 ref + 128 eeg + 3 cardinal points
+    assert raw.info['dig'][0]['ident'] == 129  # reference channel nr
```
larsoner, line 119: reference channel usually gets ident=0, then all other channels start from 1
```
>>> import mne
>>> info = mne.io.read_info(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
>>> info['dig'][85]
<DigPoint |     EEG #0 : (2.4, 111.0, -35.0) mm    : head frame>
>>> info['dig'][86]
<DigPoint |     EEG #1 : (-37.4, 105.7, 73.3) mm   : head frame>
>>> info['chs'][315]['ch_name']
'EEG 001'
>>> np.round(info['chs'][315]['loc'][:6] * 1000, 1)
array([-37.4, 105.7,  73.3,   2.4, 111. , -35. ])
```

mne/io/egi/tests/test_egi.py
larsoner: yes this part is good!

mne/io/egi/tests/test_egi.py
christianbrodbeck: Hm, so that would mean losing the EGI channel numbering -- is that preferable? I'm not sure, but couldn't there be a situation where you'd be looking at the locations and wanted to know which channel that was? Or would you always be looking at the `info['chs']` when this is relevant?

christianbrodbeck: I.e., seeing `<DigPoint |     EEG #0 : (2.4, 111.0, -35.0) mm    : head frame>` might be misleading if it does not correspond to the channel named `E0`...

larsoner: Yes the idea is that we take each format and make it conform to the `info` spec as much as possible. If for some reason we need the original ordering then we should store it in a public attribute or something.

One more point is that the EEG `dig` entries with idents `1...N` should be in the same order as the EEG channels in the data/Raw instance. But maybe you have it this way already.

larsoner: Yes it's an unfortunate coincidence that they start their numbering at zero and this collides with our dig `repr` which always says `EEG #N` starting from zero, but I think we should live with it and not deviate from our standard behavior here

christianbrodbeck: > One more point is that the EEG dig entries with idents 1...N should be in the same order as the EEG channels in the data/Raw instance. But maybe you have it this way already.

This is the case with the exception that the reference seems to be always the last EEG channel in the EGI format (as well as the `mne` objects created by the current importer).

So did I understand correctly that we want to move the reference and call it #0 in `dig` but leave it last in the data/`chs`?

larsoner: Oh it actually shows up as a channel in the data, too (just with necessarily zero value)? Then I guess in that case it's okay to leave it as is. The alternative would be to duplicate the point but I don't think that's a good solution. We can reconfigure later if need be

christianbrodbeck: Yes it's in the data with all 0s.
```py
In [2]: raw = mne.io.read_raw('test_egi.mff')

In [5]: raw.ch_names.index('E129')
Out[5]: 128

In [10]: numpy.all(raw._data[128] == 0)
Out[10]: True
```

mne/io/egi/egimff.py
```diff
@@ -253,21 +254,57 @@ def _read_locs(filepath, chs, egi_info):
...
-        loc = chs[id[0]]['loc']
-        loc[0] = sensor.getElementsByTagName('x')[0].firstChild.data
-        loc[1] = sensor.getElementsByTagName('y')[0].firstChild.data
-        loc[2] = sensor.getElementsByTagName('z')[0].firstChild.data
-        loc /= 100.  # cm -> m
-    return chs
+        chs[id[0]]['loc'][:3] = loc
+    # Insert reference properly
+    if dig_reference is not None:
+        dig_points.insert(0, dig_reference)
```
larsoner, line 303: Based on our discussion I'd revert this insertion and just leave `ident` numbers 0 to N corresponding to E0-EN as you probably had it before. It breaks from our convention of putting ref first but I think for a good reason (there is actually an all-zero ref channel and it's at the end). Sorry for the back-and forth

christianbrodbeck: EGI actually starts numbering at `E1` (at least in the montages I've seen) so in the 128 channel net the reference is `E129`. Is it ok to have `dig` `ident` numbers 1-129 with 0 missing?

larsoner: Yeah that should be okay I think

larsoner:
Thanks @christianbrodbeck !

PR #8790, opened 2021-01-25
[MRG] Coregistration-GUI:  use *.mff as digitization source

christianbrodbeck:
Depends on #8789.

This adds the capability to use `*.mff` files directly as digitization source to the coregistration GUI. I started on it for verifying the result of #8789, but it could be a useful feature since it obviates the need to convert to FIFF before doing coregistration. The screen-shot is `test_egi.mff`. 

image

### DIFF
--- a/mne/gui/_file_traits.py
+++ b/mne/gui/_file_traits.py
@@ -13,7 +13,7 @@ import numpy as np
 from traits.api import (Any, HasTraits, HasPrivateTraits, cached_property,
                         on_trait_change, Array, Bool, Button, DelegatesTo,
                         Directory, Enum, Event, File, Instance, Int, List,
-                        Property, Str, ArrayOrNone)
+                        Property, Str, ArrayOrNone, BaseFile)
 from traitsui.api import View, Item, VGroup
 from pyface.api import DirectoryDialog, OK, ProgressDialog, error, information
 
@@ -21,7 +21,7 @@ from ._viewer import _DIG_SOURCE_WIDTH
 
 from ..bem import read_bem_surfaces
 from ..io.constants import FIFF
-from ..io import read_info, read_fiducials
+from ..io import read_info, read_fiducials, read_raw_egi
 from ..io.meas_info import _empty_info
 from ..io.open import fiff_open, dir_tree_find
 from ..surface import read_surface, complete_surface_info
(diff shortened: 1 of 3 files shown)


mne/gui/_file_traits.py
```diff
@@ -288,7 +303,12 @@ def _get_n_omitted(self):
 
     @cached_property
     def _get__info(self):
-        if self.file:
+        if not self.file:
+            return
+        elif self.file.endswith('.mff'):
```
larsoner, line 308: I would make the pattern `self.file.endswith('.fif') or self.file.endswith('.fif.gz'):` and in that case use `return read_info(...)`, else use `return read_raw(...).info`. This should make it work with any reader that supports digitization

larsoner (commented):
I like the idea of supporting formats other than FIF! I think we can probably make it general to all formats rather than just extending to EGI that should take about the same number of lines. I think that there are other formats (e.g., CTF) that would immediately benefit. Feel free to `pytest.mark.parametrize` over these files in a test to see if you want.

mne/gui/_file_traits.py
```diff
@@ -288,53 +305,57 @@ def _get_n_omitted(self):
...
+            return
+        else:
+            # check that all fiducial points are present
+            has_point = {FIFF.FIFFV_POINT_LPA: False,
+                         FIFF.FIFFV_POINT_NASION: False,
+                         FIFF.FIFFV_POINT_RPA: False}
+            for d in info['dig']:
+                if d['kind'] == FIFF.FIFFV_POINT_CARDINAL:
+                    has_point[d['ident']] = True
+            if not all(has_point.values()):
```
larsoner, line 339: It seems like this whole section could be simplified, DRY'd, and have fewer missed coverage lines with something like:
```
missing = [key for key in ('LPA', 'Nasion', 'RPA') if not 
           has_point[getattr(FIFF, f'FIFFV_POINT_{key.upper()}']
if missing:
    points = _fiducial_coords(info['dig'])
    if len(points == 3):
        _append_fiducials(info['dig'], *points.T)
    else:
        error(...)
        self.reset_traits(['file'])
        return
```

larsoner (commented):
Other than one small idea for more DRY code and needing to wait for the EGI PR then a rebase/merge here, LGTM! Also don't forget to update doc/changes/latest.inc

christianbrodbeck:
I added tests for MFF and CTF, are there other formats that contain digitizer info?

larsoner:
CI failures look unrelated, thanks @christianbrodbeck !

PR #8794, opened 2021-01-27
MRG, FIX: Fix Brain offset default

larsoner:
`offset` should only be applied for inflated surfaces, at least by default. The effect is subtle but present for `sample` (`main`/`PR):

Screenshot from 2021-01-27 10-39-26

Screenshot from 2021-01-27 10-39-08

But it's much more obvious for MRIs that do not align well to canonical directions:

Screenshot from 2021-01-27 10-36-15

Screenshot from 2021-01-27 10-32-32

There are other problems that we hit with not-well-aligned MRIs but this seems like a straightforward bugfix at least.

Also adds some `See Also`s and fixes a bug where `self._subjects_dir` could be `None` which eventually causes problems -- it should just be resolved at the beginning using `get_subjects_dir` and that path stored in the object attribute.


larsoner:
Okay pushed one more commit to make it so that `offset` is computed along the closest-canonical rigid-body transform based on `talxfm` when it's available (`main`/`PR`):

Screenshot from 2021-01-27 11-49-11

Screenshot from 2021-01-27 11-49-49

Next I think we should have a `aligned=True` arg in `show_view` that uses this rotation to figure out how to make `az, el` arguments relative to these canonical axes (and also changes the origin based on `rigid[:, :3]`). This is going to take some geometrical thinking so I hope you're up for doing it in another PR @GuillaumeFavelier :)

larsoner:
Okay fixed the alignment by adding `align=True` default to `show_view`:

Screenshot from 2021-01-27 13-10-25

mne/viz/backends/_pyvista.py
```diff
@@ -993,21 +999,25 @@ def _set_3d_view(figure, azimuth, elevation, focalpoint, distance, roll=None,
...
+    # We need to add the requested roll to the roll dictated by the
+    # transformed view_up
     if roll is not None:
-        figure.plotter.camera.SetRoll(roll)
+        figure.plotter.camera.SetRoll(figure.plotter.camera.GetRoll() + roll)
 
-    figure.plotter.renderer._azimuth = azimuth
-    figure.plotter.renderer._elevation = elevation
-    figure.plotter.renderer._distance = distance
-    figure.plotter.renderer._roll = roll
```
GuillaumeFavelier, line 1010: Thanks for removing this too!

GuillaumeFavelier (approved):
LGTM !

GuillaumeFavelier:
Thanks @larsoner !

PR #2126, opened 2015-05-20
FIX: Faster read speed

Eric89GXL:
This speeds up Hari's example for me by a factor of ~400. Turns out `.times` was being created really often and bogging things down.

Closes #2049.

PR #8801, opened 2021-01-28
MRG, ENH: Dont store raw.times

### LINKED ISSUE #8745: ica.get_sources seems to be broken
here is the script to reproduce:

```py
import mne
from mne.preprocessing import ICA

raw.pick_types(meg=False, eeg=True, exclude='bads', stim=True).load_data()

ica = ICA(n_components=0.95, method='fastica').fit(raw)

ica_sources = ica.get_sources(raw, start=20, stop=40)
data, times = ica_sources[:]

ica_sources.plot()
assert data.shape == times.shape
assert abs(times[0] - 20) < 1
```

I get:

```py
ValueError: shape mismatch: objects cannot be broadcast to a single shape
```

### LINKED ISSUE #8795: EDF Reader memory hungry
try `mprof run` on this script:

```
import mne
mne.set_log_level("WARNING")


def read_raws(paths, preload):
    return [mne.io.read_raw_edf(path, preload=preload) for path in paths]


paths = mne.datasets.sleep_physionet.age.fetch_data(
    subjects=range(1, 10), recording=[1])
paths, _ = zip(*paths)

raws = read_raws(paths, preload=False)
```

it produces

figure_1

it's very surprising that it requires so much memory even when preload=False.

see https://github.com/braindecode/braindecode/pull/188

cc @gemeinl

larsoner:
Go back to creating `raw.times` on the fly when needed. On `master` this code:

<details>

```
import mne
mne.set_log_level("WARNING")


def read_raws(paths, preload):
    import gc
    out = list()
    for path in paths:
        out.append(mne.io.read_raw_edf(path, preload=preload))
        gc.collect()
    return out

paths = mne.datasets.sleep_physionet.age.fetch_data(
    subjects=range(1, 20), recording=[1])
paths, _ = zip(*paths)
raws = read_raws(paths, preload=False)
```

</details>

Produces with `mprof run mem.py && mprof plot`:

Screenshot from 2021-01-28 08-03-30

On this PR it produces:

Screenshot from 2021-01-28 08-04-01

Moreover this code, which creates a ton of epochs (a variant of the case from #2126):

<details>

```
import numpy as np
import mne

def test():
    raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif', preload=True)
    n_ev = 10000
    events = np.zeros((n_ev, 3), int)
    events[:, 2] = 1
    events[:, 0] = raw.first_samp + np.linspace(0, len(raw.times), n_ev + 2)[1:-1].round().astype(int)
    epochs = mne.Epochs(raw, events, preload=False, proj=False, baseline=None)
    epochs.average()


if __name__ == '__main__':
    import time
    t0 = time.time()
    test()
    print(time.time() - t0)
```

</details>

Takes 19.6 sec on `main` and 1.9 seconds on this PR. This speedup is mostly due to avoiding recomputing `picks` over and over again. So this PR has increased memory and CPU efficiency, hopefully.

Closes #8795
Closes #8745

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -827,7 +827,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
                             ignore_chs=self.info['bads'])
 
     @verbose
-    def _detrend_offset_decim(self, epoch, verbose=None):
+    def _detrend_offset_decim(self, epoch, picks, verbose=None):
         """Aux Function: detrend, baseline correct, offset, decim.
 
         Note: operates inplace
@@ -842,12 +842,9 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
 
         # Baseline correct
         if self._do_baseline:
-            picks = pick_types(self.info, meg=True, eeg=True, stim=False,
-                               ref_meg=True, eog=True, ecg=True, seeg=True,
-                               emg=True, bio=True, ecog=True, fnirs=True,
-                               dbs=True, exclude=[])
-            epoch[picks] = rescale(epoch[picks], self._raw_times,
-                                   self.baseline, copy=False, verbose=False)
+            rescale(
+                epoch, self._raw_times, self.baseline, picks=picks, copy=False,
+                verbose=False)
 
         # Decimate if necessary (i.e., epoch not preloaded)
         epoch = epoch[:, self._decim_slice]
@@ -872,7 +869,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
             If False copies of data and measurement info will be omitted
             to save time.
         """
-        self._current = 0
+        self.__iter__()
 
         while True:
             try:
@@ -1378,10 +1375,12 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
                     return data[:, picks]
 
             # we need to load from disk, drop, and return data
+            detrend_picks = self._detrend_picks
             for ii, idx in enumerate(use_idx):
                 # faster to pre-allocate memory here
                 epoch_noproj = self._get_epoch_from_raw(idx)
-                epoch_noproj = self._detrend_offset_decim(epoch_noproj)
+                epoch_noproj = self._detrend_offset_decim(
+                    epoch_noproj, detrend_picks)
                 if self._do_delayed_proj:
                     epoch_out = epoch_noproj
                 else:
@@ -1397,6 +1396,8 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
             n_out = 0
             drop_log = list(self.drop_log)
             assert n_events == len(self.selection)
+            if not self.preload:
+                detrend_picks = self._detrend_picks
             for idx, sel in enumerate(self.selection):
                 if self.preload:  # from memory
                     if self._do_delayed_proj:
@@ -1407,7 +1408,8 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
                         epoch = self._data[idx]
                 else:  # from disk
                     epoch_noproj = self._get_epoch_from_raw(idx)
-                    epoch_noproj = self._detrend_offset_decim(epoch_noproj)
+                    epoch_noproj = self._detrend_offset_decim(
+                        epoch_noproj, detrend_picks)
                     epoch = self._project_epoch(epoch_noproj)
 
                 epoch_out = epoch_noproj if self._do_delayed_proj else epoch
@@ -1455,6 +1457,16 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
                 return data[:, picks]
         else:
             return None
+
+    @property
+    def _detrend_picks(self):
+        if self._do_baseline:
+            return pick_types(self.info, meg=True, eeg=True, stim=False,
+                              ref_meg=True, eog=True, ecg=True, seeg=True,
+                              emg=True, bio=True, ecog=True, fnirs=True,
+                              dbs=True, exclude=())
+        else:
+            return []
 
     @fill_doc
     def get_data(self, picks=None, item=None):
@@ -2367,9 +2379,10 @@ class EpochsArray(BaseEpochs):
                                   list(self.event_id.values())).sum():
             raise ValueError('The events must only contain event numbers from '
                              'event_id')
-        for ii, e in enumerate(self._data):
+        detrend_picks = self._detrend_picks
+        for e in self._data:
             # This is safe without assignment b/c there is no decim
-            self._detrend_offset_decim(e)
+            self._detrend_offset_decim(e, detrend_picks)
         self.drop_bad()
 
 
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -835,7 +835,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
 
         # Detrend
         if self.detrend is not None:
-            # We explitly detrend just data channels (not EMG, ECG, EOG which
+            # We explicitly detrend just data channels (not EMG, ECG, EOG which
             # are processed by baseline correction)
             use_picks = _pick_data_channels(self.info, exclude=())
             epoch[use_picks] = detrend(epoch[use_picks], self.detrend, axis=1)
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -31,10 +31,9 @@ from .io.tree import dir_tree_find
 from .io.tag import read_tag, read_tag_info
 from .io.constants import FIFF
 from .io.fiff.raw import _get_fname_rep
-from .io.pick import (pick_types, channel_indices_by_type, channel_type,
+from .io.pick import (channel_indices_by_type, channel_type,
                       pick_channels, pick_info, _pick_data_channels,
-                      _pick_aux_channels, _DATA_CH_TYPES_SPLIT,
-                      _picks_to_idx)
+                      _DATA_CH_TYPES_SPLIT, _picks_to_idx)
 from .io.proj import setup_proj, ProjMixin, _proj_equal
 from .io.base import BaseRaw, TimeMixin
 from .bem import _check_origin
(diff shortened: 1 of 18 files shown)


larsoner:
This also seems to make our code magically more robust because it fixes #8745 :)

cbrnr:
I still don't understand what's using so much memory in the first example. You're loading 19 data sets, so how come the script uses ~1200MB of RAM when `preload=False`? This is only expected if `preload=True`(a data set has roughly 60MB, so 19 * 60MB = 1140MB explains that).

In the second figure, what's still taking ~140MB? Are these the `times` arrays (19 * 7MB)?

mne/io/base.py
```diff
@@ -1482,7 +1484,9 @@ def ch_names(self):
     @property
     def times(self):
         """Time points."""
-        return self._times
+        out = np.arange(self.n_times) / float(self.info['sfreq'])
+        out.flags['WRITEABLE'] = False
+        return out
```
agramfort, line 1489: I am a bit worried about this. If a code accesses times in a loop it will recompute many times. This was like this years ago and we did it this way to avoid the issue. One way would be to cache the out (ie compute times the first time you access it and then keep it for future use).

what do you think @larsoner ?

larsoner: Indeed see links above I test that use case explicitly in this PR so at least our internal uses are not a problem with Epochs.

We could cache it but then we're in a bad place where we need to both avoid using it internally *and* keep track of it using `self._update_times`. So I'd rather not do it, and instead tell users if they get slowdowns from using `raw.times` themselves they should get it once. Internally we can make sure we don't use it any more than necessary.

larsoner:
> I still don't understand what's using so much memory in the first example ... a data set has roughly 60MB...

You have to be careful what you mean by "has" -- on disk, or in memory? I think what you're seeing is merely a coincidence between the size of the dataset / number of channels on disk, and what the number of time points occupies in memory.

For example `SC4011E0-PSG.edf` has size 51110528 on disk according to `ls -al`, and it has 7 channels and 8406000 time points according to MNE. If we were to preload this data, it would take 7 * 8406000 * 8 = 470736000 bytes. Storing an array of time points will similarly require 8406000 * 8 = 67248000 bytes in memory. So even with we preload no data, if we store the `.times` attribute internally this will require ~60 MB per file.

> In the second figure, what's still taking ~140MB? Are these the times arrays (19 * 7MB)?

This is just the cost of importing mne:
```
import mne
```
Has a memory profile

Screenshot from 2021-01-28 10-08-07

About 50 MB is from SciPy imports. Feel free to dig into other causes if you want, but there are probably better targets for optimization than this...

cbrnr:
Thanks for clarifying @larsoner! So if only `times` are stored what led to >1200MB of RAM consumption in the first example then?

mmagnuski:
@cbrnr 20 files * ~60 MB per file due to `.times`.

cbrnr:
😂 oh yeah, it's been a long day...

mmagnuski:
:) sure, I know how it is.

gemeinl:
@larsoner Computing the times as
`np.arange(0, n_times/sfreq, 1/sfreq)`
instead of 
`np.arange(n_times) / sfreq`
is a ~50% speedup for me.
See https://github.com/mne-tools/mne-python/issues/8795#issuecomment-769054164

larsoner:
Indeed it's faster:
```
In [2]: n_times = 1000000
In [3]: sfreq = 1000.
In [4]: %timeit np.arange(0, n_times) / sfreq
3.53 ms ± 51.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [5]: %timeit np.arange(0, n_times / sfreq, 1 / sfreq)
1.29 ms ± 20.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
But I always worry about numerical precision being a problem with float step and endpoint because of stuff like this:
```
In [6]: np.arange(0.1, 0.3, 0.1)
Out[6]: array([0.1, 0.2])
In [7]: np.arange(0.1, 0.4, 0.1)
Out[7]: array([0.1, 0.2, 0.3, 0.4])
```
I guess we could maybe just slice into it to get rid of any extra points. Then we need to think about whether or not accumulation error will be a problem -- in this example for instance it's pretty small so might not matter (?):
```
In [8]: x = np.arange(0, n_times / 2.5, 1 / 2.5)
In [9]: len(x)
Out[9]: 1000000
In [10]: x[-1]
Out[10]: 399999.60000000003
In [11]: y = np.arange(n_times) / 2.5
In [12]: len(y)
Out[12]: 1000000
In [13]: y[-1]
Out[13]: 399999.6
```
An in-between option option would be `linspace`:
```
In [14]: %timeit np.linspace(0, (n_times - 1)/ sfreq, n_times)
2.35 ms ± 27.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [15]: z = np.linspace(0, (n_times - 1) / 2.5, n_times)
In [16]: z[-1]
Out[16]: 399999.6
```
But this made me realize that the slowdown is really probably due to the copy of the memory, so I think this is ultimately my preference:
```
In [41]: %timeit x = np.arange(n_times, dtype=float); x/=sfreq;
2.27 ms ± 53 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
Plus with the changes here I don't think it will make too much difference to people, so I'd rather stick with the one we know works (`arange`) until we see a use case where getting another speed bump will be really helpful.

cbrnr:
In-place operators FTW! 🚀

larsoner:
Actually my favorite turns out to be `numba` version, because it shouldn't suffer from any precision loss and lands in between:
```
In [8]: %timeit np.arange(0, n / sfreq, 1 / sfreq)
429 µs ± 1.68 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [9]: @jit(nopython=True)
    ...: def arange_div(n, d):
    ...:     out = np.empty(n, np.float64)
    ...:     for i in range(n):
    ...:         out[i] = i / d
    ...:     return out
In [10]: %timeit _arange_div(n, 1 / sfreq)
710 µs ± 3.02 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [11]: %timeit np.linspace(0, n / sfreq, n, endpoint=False)
1.02 ms ± 6.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

cbrnr:
But we can't use numba, right? Can you list timings of all variants on the same computer?

larsoner:
We can optionally use it (have been for at least a year now). I'll have a fallback that uses `np.arange` and inplace division:
```
In [1]: import numpy as np
In [2]: from mne.utils import _arange_div
In [3]: n = 1000000
In [4]: sfreq = 2.5
In [5]: %timeit np.arange(n) / sfreq
3.5 ms ± 13.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [6]: %timeit x=np.arange(n, dtype=float); x/=sfreq
1.14 ms ± 14.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [7]: %timeit np.linspace(0, (n - 1) / sfreq, n)
997 µs ± 15.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [8]: %timeit _arange_div(n, 1 / sfreq)
712 µs ± 3.17 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [9]: %timeit np.arange(0, (n - 0.5) / sfreq, 1 / sfreq)
430 µs ± 1.53 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
To me this is the best compromise between speed, readability, and not having to worry about numerical precision issues. And in the future maybe we can switch to the float-spaced `arange` if we see this is still a critical bottleneck.

cbrnr:
Nice! Sometimes I wish Python was Julia though 😉...

mmagnuski:
@larsoner Cool comparisons, I like the numba to np.arange + inplace fallback too.
@cbrnr Haha, yes, but everytime I wished python was Julia (or R), later using Julia (or R) i wished it was python. :)

mne/epochs.py
```diff
@@ -1439,6 +1441,16 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
         else:
             return None
 
+    @property
+    def _detrend_picks(self):
+        if self._do_baseline:
+            return pick_types(self.info, meg=True, eeg=True, stim=False,
+                              ref_meg=True, eog=True, ecg=True, seeg=True,
+                              emg=True, bio=True, ecog=True, fnirs=True,
+                              dbs=True, exclude=())
```
agramfort, line 1450: can you reuse a constant listing all data channels to be more future poof?

agramfort (commented):
I am wondering if we could not just deprecate the .times attribute... it's not really useful. It's also historically broken as it does not take into account first_samp

thoughts? too big of a change?

larsoner:
I think it's too big of a change, it has been around for a long time and is probably used in lots of code and I don't see a big benefit maintenance / usability / etc. wise from removing it. I wouldn't let the risk of inefficiency this PR introduces be something to push us over the edge on it at least...

cbrnr:
I've never needed this attribute, but it's probably not necessary to deprecate it, because this PR greatly improves upon the previous implementation. I was surprised to learn that all raw objects always carry around this relatively large chunk of memory (which is useless for many/most use cases) even when `preload=False`, but now this is not the case anymore. I did get it right that now `times` are computed on demand, i.e. only when someone explicitly asks for this array, right? It's never attached to a `raw` object, so I'd say there's no reason to get rid of it.

larsoner:
And even with 1e6 samples it takes ~1 ms to create so I really doubt this will create a bottleneck for people actually processing data.

agramfort:
Fair enough!

larsoner:
Setting to WIP until I take care of the above comment about DRYing the code. There are a number of places in `epochs.py` that can benefit from this so I couldn't do it quickly.

larsoner:
This one is also ready @agramfort , failures are just our recent PyVista segfaults

sappelhoff:
@larsoner I couldn't find any docs on using numba optionally in MNE, if numba does speed up some mne operations, wouldn't it make sense to tell people that they can optionally install it to reap the benefits?

mne/preprocessing/ica.py
```diff
@@ -906,28 +908,19 @@ def _sources_as_raw(self, raw, add_channels, start, stop):
             raw._data = data
 
         # populate copied raw.
-        start, stop = _check_start_stop(raw, start, stop)
         if add_channels is not None:
             raw_picked = raw.copy().pick_channels(add_channels)
-            data_, times_ = raw_picked[:, start:stop]
+            data_, _ = raw_picked[:, start:stop]
```
agramfort, line 913: did you consider using raw_picked._getitem(..., return_times=False) here?

agramfort (approved):
just a question. Besides LGTM

mne/preprocessing/ica.py
larsoner: Nope, added that after this change -- will push a commit

cbrnr:
Thanks @larsoner! Great and very useful change!

PR #8803, opened 2021-01-28
WIP, ENH: Add sensors to Brain

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/issues/8382#issuecomment-712223093 and implements `add_sensors` to plot MEG and EEG sensors.

For now, I move code and experiment. Refactoring is necessary to avoid duplication between `plot_alignment()` and `add_sensors`.

This is still work in progress.


GuillaumeFavelier:
I play with it using:

```py
import os.path as op

import numpy as np
import nibabel as nib

import mne

raw_fname = op.join(data_path, 'MEG', 'sample', 'sample_audvis_raw.fif')
trans_fname = op.join(data_path, 'MEG', 'sample',
                      'sample_audvis_raw-trans.fif')
trans = mne.read_trans(trans_fname)
src = mne.read_source_spaces(op.join(subjects_dir, 'sample', 'bem',
                                     'sample-oct-6-src.fif'))

# Load the T1 file and change the header information to the correct units
t1w = nib.load(op.join(data_path, 'subjects', 'sample', 'mri', 'T1.mgz'))
t1w = nib.Nifti1Image(t1w.dataobj, t1w.affine)
t1w.header['xyzt_units'] = np.array(10, dtype='uint8')
t1_mgh = nib.MGHImage(t1w.dataobj, t1w.affine)

Brain = mne.viz.get_brain_class()
brain = Brain(subject_id="sample", hemi="both", surf="white", alpha=0.7,
              units="m")
brain.add_sensors(info=raw.info, trans=trans, coord_frame='meg')

mne.viz.plot_alignment(raw.info, trans=trans, subject='sample',
                       subjects_dir=subjects_dir, surfaces='head-dense',
                       show_axes=True, dig=True, eeg=[], meg='sensors',
                       coord_frame='meg', mri_fiducials='estimated',
                       fig=brain._renderer.scene())
```

I had to change `units` to make the brain scale correctly but I still think some transformation is missing.

image

larsoner:
brain.add_sensors should not have a `coord_frame` argument. It's always MRI because `brain` is always in MRI. This is probably our transform problem

GuillaumeFavelier:
Working on EEG now.

Where to fetch the sensors' color?

larsoner:
Same place as `plot_alignment`, it's some defaults.py / DEFAULTS entry

https://github.com/mne-tools/mne-python/blob/main/mne/viz/_3d.py#L1063-L1065

Eventually we'll want to add a parameter to allow controlling the sensor color on a channel-name basis, but not in the first PR (need to sort out the API). But if you can at least consider this use case it might help guide the initial design.

GuillaumeFavelier:
Another naive question, should I add support for projected EEG also? On what surface?

Also `%(trans)s` is not automagically detected in the docstring.

larsoner:
> Another naive question, should I add support for projected EEG also? On what surface?

Not in the first PR, we can always add it later.

> Also %(trans)s is not automagically detected in the docstring.

You either need `@verbose` decorator if there is a `verbose=None` kwarg or `@fill_doc` otherwise

GuillaumeFavelier:
Based on `plot_ecog`, I got something like the following in this early version:

image

This slider should fit somewhere nice when `time_viewer=True`, the default pyvista colormap is used and the range of scalar values is calculated on the fly. Glyph refactor is also required for this.

mne/viz/_3d.py
```diff
@@ -416,6 +417,50 @@ def plot_evoked_field(evoked, surf_maps, time=None, time_label='t = %0.0f ms',
...
+        if warn_meg:
+            warn('MEG sensors not found. Cannot plot MEG locations.')
+    else:
+        meg_rrs = apply_trans(trans,
+                              np.concatenate(meg_rrs, axis=0))
+        meg_tris = np.concatenate(meg_tris, axis=0)
+    return dict(rr=meg_rrs, tris=meg_tris)
+
+
+def _get_coord_frame_trans(coord_frame, info, trans):
```
alexrockhill, line 444: This seems confusing, could you just transform to `head` if in "meg", then transform to `mri` if in originally in "meg" (now "head" because it was just transformed) or "head"? That seems like it might be more straightforward

alexrockhill: Ah I guess you might need them multiple times so that might not work

PR #8805, opened 2021-01-28
MRG, ENH: Add warning about bad whitener conditioning

larsoner:
Attempt to address https://mne.discourse.group/t/strange-source-activity-with-lcmv-beamformer-in-volume-source-space-empty-room-covariance/2549

@SophieHerbst @hoechenberger can you see if this warns when you try your beamforming script? If not can you send me an evoked file to use? This was my testing script since I didn't have one:

<details>

```
import mne
cov = mne.read_cov('data-cov.fif')
noise_cov = mne.read_cov('noise-cov.fif')
fwd = mne.read_forward_solution('FORWARD_ico-4_MEGonly_CEREB-fwd.fif')
info = fwd['info']
info.update(comps=[], projs=[])

filters = mne.beamformer.make_lcmv(
    info, fwd, cov, noise_cov=noise_cov, rank=dict(meg=73), verbose=True)
```

</details>

And it now produces for me a nice warning, which I line-wrap here:
```
Computing rank from covariance with rank={'meg': 73}
Computing rank from covariance with rank={'meg': 73}
Making LCMV beamformer with rank {'meg': 73}
Computing inverse operator with 306 channels.
    306 out of 306 channels remain after picking
Selected 306 channels
Whitening the forward solution.
Computing rank from covariance with rank={'meg': 73}
    Setting small MEG eigenvalues to zero (without PCA)
/home/larsoner/Desktop/COV/rep.py:8: RuntimeWarning: The passed rank['meg']=73 exceeds the rank
estimated from the noise covariance (72) leading an increase in condition number by a
factor of 5e+12, this might amplify noise during whitening by a factor of 2e+06
  filters = mne.beamformer.make_lcmv(
Creating the source covariance matrix
Adjusting source covariance matrix.
Computing beamformer filters for 39825 sources
  File "/home/larsoner/Desktop/COV/rep.py", line 8, in <module>
    filters = mne.beamformer.make_lcmv(
  File "<decorator-gen-360>", line 22, in make_lcmv
  File "/home/larsoner/python/mne-python/mne/beamformer/_lcmv.py", line 176, in make_lcmv
    W, max_power_ori = _compute_beamformer(
  File "/home/larsoner/python/mne-python/mne/beamformer/_compute_beamformer.py", line 240, in _compute_beamformer
    raise ValueError(
ValueError: Singular matrix detected when estimating spatial filters. Consider reducing the rank of the forward operator by using reduce_rank=True.
```
The error is also there for me on `main`.

### DIFF
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -13,8 +13,9 @@ from .io.pick import (_picks_by_type, pick_info, pick_channels_cov,
                       _picks_to_idx)
 from .io.proj import make_projector
 from .utils import (logger, _compute_row_norms, _pl, _validate_type,
-                    _apply_scaling_cov, _undo_scaling_cov, _check_option,
-                    _scaled_array, warn, _check_rank, _on_missing, verbose)
+                    _apply_scaling_cov, _undo_scaling_cov,
+                    _scaled_array, warn, _check_rank, _on_missing, verbose,
+                    _check_on_missing)
 
 
 @verbose
@@ -318,9 +319,7 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
 
     rank = _check_rank(rank)
     scalings = _handle_default('scalings_cov_rank', scalings)
-    _validate_type(on_rank_mismatch, str, 'on_rank_mismatch')
-    _check_option('on_rank_mismatch', on_rank_mismatch,
-                  ('ignore', 'warn', 'raise'))
+    _check_on_missing(on_rank_mismatch, 'on_rank_mismatch')
 
     if isinstance(inst, Covariance):
         inst_type = 'covariance'
--- a/mne/rank.py
+++ b/mne/rank.py
@@ -13,8 +13,8 @@ from .io.pick import (_picks_by_type, pick_info, pick_channels_cov,
                       _picks_to_idx)
 from .io.proj import make_projector
 from .utils import (logger, _compute_row_norms, _pl, _validate_type,
-                    _apply_scaling_cov, _undo_scaling_cov,
-                    _scaled_array, warn, _check_rank, verbose)
+                    _apply_scaling_cov, _undo_scaling_cov, _check_option,
+                    _scaled_array, warn, _check_rank, _on_missing, verbose)
 
 
 @verbose
@@ -273,7 +273,7 @@ def _compute_rank_int(inst, *args, **kwargs):
 
 @verbose
 def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
-                 proj=True, tol_kind='absolute', check_passed=False,
+                 proj=True, tol_kind='absolute', on_rank_mismatch='ignore',
                  verbose=None):
     """Compute the rank of data or noise covariance.
 
@@ -299,7 +299,7 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
         If True, all projs in ``inst`` and ``info`` will be applied or
         considered when ``rank=None`` or ``rank='info'``.
     %(rank_tol_kind)s
-    %(check_passed)s
+    %(on_rank_mismatch)s
     %(verbose)s
 
     Returns
@@ -318,6 +318,9 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
 
     rank = _check_rank(rank)
     scalings = _handle_default('scalings_cov_rank', scalings)
+    _validate_type(on_rank_mismatch, str, 'on_rank_mismatch')
+    _check_option('on_rank_mismatch', on_rank_mismatch,
+                  ('ignore', 'warn', 'raise'))
 
     if isinstance(inst, Covariance):
         inst_type = 'covariance'
@@ -357,7 +360,7 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
             # special case: if whitening a covariance, check the passed rank
             # against the estimated one
             est_verbose = False
-            if not (check_passed and
+            if not (on_rank_mismatch != 'ignore' and
                     rank_type == 'estimated' and
                     ch_type == 'meg' and
                     isinstance(inst, Covariance) and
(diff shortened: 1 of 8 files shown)


mne/rank.py
```diff
@@ -376,29 +393,44 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
+                    this_rank, s = _estimate_rank_meeg_cov(
+                        data, pick_info(simple_info, picks), scalings, tol,
+                        return_singular=True, verbose=est_verbose)
+                    if ch_type in rank:
+                        ratio = s[this_rank - 1] / s[rank[ch_type] - 1]
+                        if ratio > 100:
+                            warn(f'The passed rank[{repr(ch_type)}]='
+                                 f'{rank[ch_type]} exceeds the rank estimated '
+                                 f'from the noise covariance ({this_rank}) '
+                                 'leading an increase in condition number by '
```
hoechenberger, line 417:
```suggestion
                                 f'leading an increase in condition number by '
```

for consistency

mne/rank.py
```diff
@@ -376,29 +393,44 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
+                                 f'a factor of {ratio:0.1g}, this might '
+                                 'amplify noise during whitening by a factor '
```
hoechenberger, line 419:
```suggestion
                                 f'amplify noise during whitening by a factor '
```
same

mne/rank.py
```diff
@@ -376,29 +393,44 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
+
+                    this_rank, s = _estimate_rank_meeg_cov(
+                        data, pick_info(simple_info, picks), scalings, tol,
+                        return_singular=True, verbose=est_verbose)
+                    if ch_type in rank:
+                        ratio = s[this_rank - 1] / s[rank[ch_type] - 1]
+                        if ratio > 100:
+                            warn(f'The passed rank[{repr(ch_type)}]='
+                                 f'{rank[ch_type]} exceeds the rank estimated '
+                                 f'from the noise covariance ({this_rank}) '
```
hoechenberger, line 416:
```suggestion
                                 f'from the noise covariance ({this_rank}), '
```

hoechenberger:
@larsoner Thanks, I couldn't test it so far, but I was just thinking: should we maybe even **raise** by default? Because this is a serious issue that will typically blow up your entire analysis, right?

larsoner:
This is just a heuristic, AFAIK it's not guaranteed to work. There might be realistic cases that trigger the warning, so I don't want to raise an error.

SophieHerbst:
Hi @eric, 
I tested it and I get to see the warning. 

Now, I am not sure I would have known what to do about it before our discussion. 
How about adding your recommendation of passing the rank of the matrix used to whiten?
(– and also add that to the documentation?) 
Also, I don't understand what you mean by "condition number". 

[While testing this I realized that the interactive console in VS code had all warnings disabled per default, so I might have missed them - thankfully fixed now with @hoechenberger's help.]

larsoner:
Reworded and added to `rank` arg docstring, see if that's better @SophieHerbst !

hoechenberger (approved):

SophieHerbst:
yep, it is clear now @larsoner

mne/utils/docs.py
```diff
@@ -1003,6 +1006,13 @@
         Support for reducing rank in all modes (previously only supported
         ``pick='max_power'`` with weight normalization).
 """
+docdict['check_passed'] = """
+check_passed : bool
+    If True, sanity-check passed MEG values against computed ones (only used
+    for covariances; default False).
```
agramfort, line 1012: this is not explicit enough.

mne/rank.py
```diff
@@ -376,29 +393,45 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
-                    rank[ch_type] = (inst['data'][picks] > 0).sum() - n_proj
+                    this_rank = (inst['data'][picks] > 0).sum() - n_proj
                 else:
                     data = inst['data'][picks][:, picks]
                     if proj:
                         data = np.dot(np.dot(proj_op, data), proj_op.T)
-                    rank[ch_type] = _estimate_rank_meeg_cov(
-                        data, pick_info(simple_info, picks), scalings, tol)
+
+                    this_rank, s = _estimate_rank_meeg_cov(
```
agramfort, line 408: can you avoid single letter var names like s here?

mne/rank.py
```diff
@@ -376,29 +393,45 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
...
+                        data, pick_info(simple_info, picks), scalings, tol,
+                        return_singular=True, verbose=est_verbose)
+                    if ch_type in rank:
+                        ratio = s[this_rank - 1] / s[rank[ch_type] - 1]
+                        if ratio > 100:
+                            warn(f'The passed rank[{repr(ch_type)}]='
+                                 f'{rank[ch_type]} exceeds the estimated rank '
+                                 f'of the noise covariance ({this_rank}) '
+                                 f'leading a potential increase in '
```
agramfort, line 417:
```suggestion
                                 f'leading to a potential increase in '
```

mne/utils/docs.py
agramfort: can we find a better name than check_passed ? will it error or warn?

larsoner: It will warn. I guess we could rename it `on_mismatch : str` with raise/warn(default)/ignore like we do with `on_missing`.

agramfort:
on_rank_mismatch?

as it's not used all the time on a rank function from the public API.

>

agramfort (approved):

larsoner:
Failures are just #8819 so I'll go ahead and merge

hoechenberger:
Thank you, @larsoner!!

Issue #8550, opened 2020-11-20
Unnecessary FileNotFoundError in Persyst Reader 


pmyers16:
#### Describe the bug
The Persyst reader unnecessarily throws a FileNotFound error when both the .lay and .dat
files have been moved to a new directory. The reader attempts to account for this by replacing
the dirname of the .dat file's path with the dirname of the .lay file, but still throws the error if 
the path to the .dat file as listed in the .lay file's "file" field does not exist. 

This issue does not occur if the File field is a relative path. I am not sure under what circumstances
the File field is an absolute path.

#### Steps to reproduce

1. Download the Persyst files from [mne testing data](https://github.com/mne-tools/mne-testing-data) and save both files to the same folder, say "C:/Users/<username>/persyst_data"

2. Change line 2 of the .lay file to an absolute path that does not actually exist, say "C:/Users/<username>/persyst_data/other_folder/sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.dat"

3. Run the following code:
`lay_fname = "C:/Users/<username>/persyst_data/sub-pt1_ses-02_task-monitor_acq-ecog_run-01_clip2.lay"`
`rawPersyst = mne.io.read_raw_persyst(lay_fname)`

#### Expected results
The code snippet should not fail since `read_raw_persyst` correctly identifies that the .dat file has moved directories along with the .lay file. The reader should succeed for **either** of the two following scenarios:

1. The path listed in the 'File' field in the .lay file exists
OR
2.  Both the .lay and .dat files have been moved to the same directory (i.e. the dirname in the 'File' field is ignored)

If neither of these two scenarios occurs, the error message should say this, listing both paths checked, not the path to the .lay file that does exist.
 
#### Actual results
The following FileNotFoundError: 

>   File "D:\ProgramFiles\Anaconda\envs\ezscalp\lib\site-packages\mne\io\persyst\persyst.py", line 39, in read_raw_persyst
>     return RawPersyst(fname, preload, verbose)
>   File "<decorator-gen-215>", line 24, in __init__
>   File "D:\ProgramFiles\Anaconda\envs\ezscalp\lib\site-packages\mne\io\persyst\persyst.py", line 101, in __init__
>     raise FileNotFoundError(error_msg)
> FileNotFoundError: The data path you specified does not exist for the lay path,

#### Additional information
Output of mne.sys_info(): 

[...]

PR #8311, opened 2020-09-26
Update persyst.py

infinitejest:
Fixes issue that annotation parser is not able to parse comments with same description

#### Reference issue
Fixes #8292 

#### What does this implement/fix?
Persyst annotation parser of Lay files is not able to parse when annotations have same description over different annotation time and duration. This fix creates lists of tuples of annotation with same description and then creates an annotation instance of each individual annotation including repeated instances. 

#### Additional information
OF note when try to create events from annotation objects events are not able to recognize multiple instances of annotation objects with same description.

PR #8806, opened 2021-01-28
[MRG] Fix persyst reader - adds test for multiple repeated Comments, allow commas and require lay and dat file to be in same directory

adam2392:
#### Reference issue
Closes: #8311
Closes: #8550 

Initial strategy was taken from #8311, and then I simplified the approach.
Co-authored-by: Andres Rodriguez <rodriguezandr@gmail.com>

Adds unit test for repeated comments, and comments with `,` characters.

In addition, cleans up the logic and explicitly requires `.lay` and `.dat` files to be in the same directory.

#### Additional information
Requires `mne-testing-data` to be updated to 0.114


adam2392:
See: https://github.com/mne-tools/mne-testing-data/pull/81

mne/datasets/utils.py
```diff
@@ -246,7 +246,7 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     path = _get_path(path, key, name)
     # To update the testing or misc dataset, push commits, then make a new
     # release on GitHub. Then update the "releases" variable:
-    releases = dict(testing='0.113', misc='0.8')
+    releases = dict(testing='0.114', misc='0.8')
```
larsoner, line 249: This has been released with hash 0e7fb1b37f924bf50ce5db9f96c67972

mne/io/persyst/tests/test_persyst.py
```diff
@@ -16,10 +17,10 @@
 from mne.utils import run_tests_if_main
 
 fname_lay = op.join(
-    data_path(download=False), 'Persyst',
+    data_path(download=True), 'Persyst',
```
larsoner, line 20: Should not need to change this and it will break things if you do, please revert

mne/io/persyst/tests/test_persyst.py
```diff
@@ -148,6 +149,26 @@ def test_persyst_standard():
     _test_raw_reader(read_raw_persyst, fname=fname_lay)
 
 
+@requires_testing_data
+def test_persyst_annotations():
+    """Test annotations reading in Persyst."""
+    out_dir = mne.utils._TempDir()
```
larsoner, line 155:
```suggestion
def test_persyst_annotations(tmpdir):
    """Test annotations reading in Persyst."""
    out_dir = str(tmpdir)
```

mne/io/persyst/persyst.py
```diff
@@ -74,7 +74,10 @@ def __init__(self, fname, preload=False, verbose=None):
         fileinfo_dict = OrderedDict()
         channelmap_dict = OrderedDict()
         patient_dict = OrderedDict()
-        comments_dict = OrderedDict()
+        comments_dict = defaultdict(list)
```
larsoner, line 77: The use of `OrderedDict` is intentional to a) make it clear it needs to be ordered and b) ordering only became an official language feature in 3.7 so it's safer to use OrderedDict while we still support 3.6 (other implementations like PyPy are not required to be ordered until then)

mne/io/persyst/persyst.py
```diff
@@ -108,8 +111,10 @@ def __init__(self, fname, preload=False, verbose=None):
             # Patient (All optional)
             elif section == 'patient':
                 patient_dict[key] = val
+            # Comments (turned into mne.Annotations)
             elif section == 'comments':
-                comments_dict[key] = val
+                comments_dict[key].append(val)
```
larsoner, line 116: then this can just be the following if you still want a one-liner
```suggestion
                comments_dict[key] = comments_dict.get(key, list()) + [val]
```
or split it into two lines if you want

adam2392:
Okay this looks good to me. Codecov get's hit I think cuz I touched `utils.py`?

doc/changes/latest.inc
```diff
@@ -80,6 +80,8 @@ Bugs
 
 - Fix bug with :ref:`mne make_scalp_surfaces` where ``--overwrite`` was not functional (:gh:`8800` by `Yu-Han Luo`_)
 
+- Fix bugs with `mne.io.read_raw_persyst` where multiple ``Comments`` with the same name are allowed, and ``Comments`` with a "," character are now allowed (:gh:`8311` and :gh:`8806` by `Adam Li`_ and `Andres Rodriguez`_)
```
larsoner, line 83: `Andres Rodriguez` is a new contributor (congrats!), so this should go to the top of the section and have new-contributor highlighting on the name

larsoner (commented):
Otherwise LGTM! I think codecov is just having a bad day

larsoner:
Thanks @adam2392 @infinitejest !

infinitejest:
I just realized my fix for the comments/annotations with commas did not make it to the persyst.py. It is just a matter to add a try statement to line 344 with following

```
 try: 
          time_sec, duration, state, var_type, text = line.split(',')
          status = 2
          key = text
          value = (time_sec, duration)
except: 
          status = 2
          key =  "Error importing annotation"
          value = (0,0)
```

I will work in the other issue that persyst.py only works on windows over another PR

PR #8807, opened 2021-01-28
MRG, ENH: Add support for other formats to browse_raw

larsoner:
1. Use `read_raw` rather than `read_raw_fif` in `browse_raw` to extend support to other formats
2. Add `.con` as a recognized KIT format name
3. Fix bug where `scalings='auto` would use the wrong sample range to compute auto scalings. It passed float times when it needed to pass int samples to `_read_segment`.


hoechenberger (approved):

hoechenberger:
Thanks, @larsoner!

PR #8808, opened 2021-01-28
MRG, MAINT: Deprecated param and pytest-qt

larsoner:
About a quarter of our macOS and Linux conda runs are dying because of `setIcon` calls in PyVista. I even hit it locally once in several tests. This PR tries to avoid this by using `qtbot.wait_exposed` in case it's some multi-window problem.

It also changes `fig.gca(projection='3d')` calls that are deprecated in latest matplotlib `master` to more modern calls that explicitly create axes.

Hopefully all comes back green and I can try restarting macOS and Linux a few times to see if it's more reliable.


GuillaumeFavelier:
Since `qtbot` is available, we could even use `addWidget`. 

I'm also pleasantly surprised that `waitExposed` works on `plotter` directly. In pyvistaqt, `app_window` is tested instead.

larsoner:
Well this appears not to work. :sob: 

I'll remove all the `qtbot` stuff and just leave the `gca` stuff in there for now.

GuillaumeFavelier:
Thank you for trying. I will exclude it from the possibilities.

It's still not clear to me what instruction causes the crash though. If it's only `set_icon`, we could disable it entirely.

I will investigate.

larsoner:
Sometimes it's in paintEvent, too

larsoner:
Feel free to merge this one if you're happy with it in the meantime though

GuillaumeFavelier (approved):
LGTM

GuillaumeFavelier:
Thanks @larsoner

PR #8817, opened 2021-02-01
fix mne.viz.plot_topomap with some missing grad in a pair

### LINKED ISSUE #8802: Issue with mne.viz.plot_topomap on sample dataset restricted to MEG channels
#### Describe the bug
The `mne.viz.plot_topomap` does not produce the expected result on the sample dataset when limiting to MEG channels. 

#### Steps to reproduce
Run the following script:
```
import os
import mne
import numpy as np

subjects_dir = os.path.join(data_path, "subjects")
data_dir = os.path.join(data_path, 'MEG', 'sample')
file_name = os.path.join(data_dir, 'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(file_name, preload=True, verbose=False)
raw.pick_types(meg='grad', eeg=False, eog=False, stim=True)
print('done')

# info of the loaded channels
info = raw.copy().pick_types(meg=True).info

X = raw.get_data(picks=['meg'])
print("Data shape:", X.shape)

# simulate random data
data = np.random.randn(203)

assert len(info['chs']) == data.size

# plot
mne.viz.plot_topomap(data, info)
```

#### Expected results
`plot_topomap` should create a topographic map as image.

#### Actual results
Got the following error:
```
ValueError: cannot reshape array of size 203 into shape (101,2,newaxis)
```
which intervenes when the function `_merge_grad_data(data, method)` is called.

#### Additional information
Platform:      macOS-10.15.7-x86_64-i386-64bit
Python:        3.7.4 (default, Aug 13 2019, 15:17:50)  [Clang 4.0.1 (tags/RELEASE_401/final)]
Executable:    /Users/cedricallain/anaconda3/bin/python
CPU:           i386: 8 cores
Memory:        16.0 GB

mne:           0.21.0
numpy:         1.17.2 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.3.1
matplotlib:    3.1.1 {backend=module://ipykernel.pylab.backend_inline}

sklearn:       0.21.3
numba:         0.45.1
nibabel:       Not found
cupy:          Not found
pandas:        0.25.1
dipy:          Not found
mayavi:        Not found
pyvista:       Not found
vtk:           Not found

agramfort:
closes #8802

### DIFF
--- a/mne/viz/tests/test_topomap.py
+++ b/mne/viz/tests/test_topomap.py
@@ -557,6 +557,19 @@ def test_plot_topomap_bads():
     plt.close('all')
 
 
+def test_plot_topomap_bads_grad():
+    """Test plotting topomap with bad gradiometer channels (gh-8802)."""
+    import matplotlib.pyplot as plt
+    data = np.random.RandomState(0).randn(203)
+    info = read_info(evoked_fname)
+    info['bads'] = ['MEG 2242']
+    picks = pick_types(info, meg='grad')
+    info = pick_info(info, picks)
+    assert len(info['chs']) == 203
+    plot_topomap(data, info, res=8)
+    plt.close('all')
+
+
 def test_plot_topomap_nirs_overlap(fnirs_epochs):
     """Test plotting nirs topomap with overlapping channels (gh-7414)."""
     fig = fnirs_epochs['A'].average(picks='hbo').plot_topomap()
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -90,6 +90,8 @@ Bugs
 
 - Fix bug with :ref:`mne make_scalp_surfaces` where ``--overwrite`` was not functional (:gh:`8800` by `Yu-Han Luo`_)
 
+- Fix bug with :func:`mne.viz.plot_topomap` when plotting gradiometers with a missing channel in a pair (:gh:`8817` by `Alex Gramfort`_)
+
 API changes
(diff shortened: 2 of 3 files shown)


mne/viz/tests/test_topomap.py
```diff
@@ -557,6 +557,19 @@ def test_plot_topomap_bads():
     plt.close('all')
 
 
+def test_plot_topomap_bads_grad():
+    """Test plotting topomap with bad gradiometer channels (gh-8802)."""
+    import matplotlib.pyplot as plt
```
larsoner, line 562: No need to nest these imports anymore due to pytest infrastructure, eventually we can move these to the top. But it can be another PR since this whole file suffers from our old conventions.

mne/viz/tests/test_topomap.py
```diff
@@ -557,6 +557,19 @@ def test_plot_topomap_bads():
...
+    data = np.random.RandomState(0).randn(203)
+    info = read_info(evoked_fname)
+    info['bads'] = ['MEG 2242']
+    picks = pick_types(info, meg='grad')
+    info = pick_info(info, picks)
+    assert len(info['chs']) == 203
+    plot_topomap(data, info, res=8)
+    plt.close('all')
```
larsoner, line 570: Same here, no need to `plt.close('all')` at the end, it happens automatically via pytest now

larsoner:
Thanks @agramfort

PR #8822, opened 2021-02-01
MRG, ENH: Add Maxwell basis computation

larsoner:
Requested on Discourse. I've used this function a number of times previously, seems reasonable to expose it for advanced users. Single-letter `S` follows the publication nomenclature.


drammock (approved):
looks reasonable though I'm a little unsure about the function name. What about something like "get_maxwell_basis" or "get_basis_from_info" or similar?

larsoner:
Renamed to `compute_maxwell_basis` based on the other names in `preprocessing.rst`

drammock:
Thanks @larsoner !

PR #8828, opened 2021-02-02
MNT: CIs

GuillaumeFavelier:
This PR:

1) continues to escape `%` in azure pipelines:

`
##[warning]%25 detected in ##vso command. In March 2021, the agent command parser will be updated to unescape this to %. To opt out of this behavior, set a job level variable DECODE_PERCENTS to false. Setting to true will force this behavior immediately. More information can be found at https://github.com/microsoft/azure-pipelines-agent/blob/master/docs/design/percentEncoding.md
`

2) investigates failling viz test on `MacOS` and `Linux`

PR #8829, opened 2021-02-02
MRG, ENH: Speed up import by caching indented params

### LINKED ISSUE #8826: MNE-Python import time
It seems to me like it's taking longer to `import mne` from version to version. Currently, importing takes a little over one second. I don't know if that's OK (probably) or if we should invest a little time to see if we can easily cut down on import time. I've already profiled the import with:

```
python -X importtime -c "import mne" 2> mne_import.log
```

`tuna` (`pip install tuna`) can be used to view the output interactively:

```
tuna mne_import.log
```

Here's a screenshot of the output for a rough overview (you have to run `tuna` yourself for an interactive version):
Screen Shot 2021-02-02 at 08 50 01

It seems like a lot of time is spent in `mne.io`, which maybe isn't necessary IDK.

Do you think this is worth pursuing or is a second of import time OK?

larsoner:
So far 940 ms to 560 ms on my system by caching the indented versions of our docdict. It says 325 sec of this is importing numpy for me so not a ton left to be gained. I think there might be some other gains to be made for example by nesting scipy imports, though, so I'm at least going to try that.

Closes #8826

### DIFF
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -13,6 +13,7 @@ import logging
 import os.path as op
 import warnings
 
+from .docs import fill_doc
 from ..externals.decorator import FunctionMaker
 
 
@@ -79,6 +80,7 @@ def verbose(function):
     Examples
     --------
     You can use the ``verbose`` argument to set the verbose level on the fly::
+
         >>> import mne
         >>> cov = mne.compute_raw_covariance(raw, verbose='WARNING')  # doctest: +SKIP
         >>> cov = mne.compute_raw_covariance(raw, verbose='INFO')  # doctest: +SKIP
@@ -88,7 +90,6 @@ def verbose(function):
     """  # noqa: E501
     # See https://decorator.readthedocs.io/en/latest/tests.documentation.html
     # #dealing-with-third-party-decorators
-    from .docs import fill_doc
     try:
         fill_doc(function)
     except TypeError:  # nothing to add
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -36,6 +36,8 @@ Enhancements
 
 - Reduce memory consumption of `mne.io.Raw` and speed up epoching when thousands of events are present for `mne.Epochs` (:gh:`8801` by `Eric Larson`_)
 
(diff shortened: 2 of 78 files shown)


mne/utils/_logging.py
```diff
@@ -88,7 +90,6 @@ def verbose(function):
     """  # noqa: E501
     # See https://decorator.readthedocs.io/en/latest/tests.documentation.html
     # #dealing-with-third-party-decorators
-    from .docs import fill_doc
```
larsoner, line 91: @cbrnr this is where I just did `return function` to see how much time the `verbose` decorator was eating up. Turns out almost all of it was in `fill_doc` hence why I target it in this PR.

cbrnr:
I'm getting 1152ms vs. 557ms on my machine, so already half the time! 👍

cbrnr:
Down to 533ms with the latest changes!

larsoner:
I ended up nesting all `scipy` imports, and improving (I think) our `utils/linalg.py` code by using caches to store the functions, which should be fast enough for our cases. Ready for review/merge from my end

cbrnr:
497ms! LGTM but what about failing CIs?

cbrnr:
Also, this should really be mentioned (and highlighted) in the changelog! Cutting import time in half is 🔥!

larsoner:
If we (reasonably) assume anyone doing any analysis will `import numpy` we could claim the "MNE import overhead" decreased from 700 ms to 200 ms :)

cbrnr:
True! It's 482ms for me, of which `numpy` uses 130ms, which leaves 352ms. On `main` its 1102ms, minus `numpy` that's 972ms. So that's almost a 3x speed increase! Super nice, thanks for implementing this idea so quickly @larsoner!

cbrnr:
@larsoner there are still errors with `test_module_nesting`.

drammock:
Also if we're nesting scipy now, there should be an update to the contributor guide about which imports to nest

cbrnr:
Is the segfault legit?

larsoner:
Not relevant for this PR it's just #8828

cbrnr (approved):

agramfort (approved):

agramfort:
thx @larsoner

PR #8831, opened 2021-02-02
FIX: only delete visible annotation spans

drammock:
issue raised on discourse: https://mne.discourse.group/t/annotation-editing-deletion-and-undo/2650/2

previously, right-clicking to remove an annotation would remove *all* annotation spans that overlap at the mouse position, *regardless of whether they were visible or hidden by the UI*.  Now it will remove only *visible* spans under the mouse cursor, meaning it is possible to remove only a specific span by hiding other overlapping ones first.

### DIFF
--- a/mne/viz/tests/test_raw.py
+++ b/mne/viz/tests/test_raw.py
@@ -539,6 +539,21 @@ def test_plot_annotations(raw):
     assert len(fig.mne.annotation_texts) == 1
 
 
+@pytest.mark.parametrize('hide_which', (0, 1))
+def test_remove_annotations(raw, hide_which):
+    # test right-click doesn't remove hidden annotation spans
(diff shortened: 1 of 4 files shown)


mne/viz/tests/test_raw.py
```diff
@@ -539,6 +539,21 @@ def test_plot_annotations(raw):
     assert len(fig.mne.annotation_texts) == 1
 
 
+@pytest.mark.parametrize('hide_which', (0, 1))
+def test_remove_annotations(raw, hide_which):
```
larsoner, line 543: Worth adding a `None` condition where then `len(raw.annotations) == 0`?

larsoner (approved):

larsoner:
Thanks @drammock !

datalw:
I just tested it btw, it worked for me very well! Thanks a lot!

PR #8834, opened 2021-02-03
ENH: Allow automated compilation of metadata

hoechenberger:
This PR adds` mne.epochs.make_metadata()`, which allows for the automated construction of metadata for use in Epochs. 

The users passes an events array, an event name -> event ID mapping, a time period and a sampling frequency. The function will then

- iterate over all events specified in the mapping,
- determine which events fall into the specified time period (relative to the current – time-locked – event), and
- will add them to the metadata in two cells: one contains a boolean value indicating whether the event in question is present in the respective time period / Epoch, and one with a float value of the event time in seconds, relative to the time-locked event.

It will then return the constructed metadata and a matching subset of the events array. These values can then be readily used to construct Epochs of arbitrary durations, while ensuring all of them will be assigned the desired metadata (which may or may not contain events from a different time range around the time-locked event)

Example:

```python
import os
import mne

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False).crop(tmax=120)
tmax = 1
tmin = -0.5
sfreq = raw.info['sfreq']

all_events = mne.find_events(raw, stim_channel='STI 014')
            'visual/right': 4, 'face': 5, 'button': 32}

metadata, events = mne.epochs.make_metadata(
    events=all_events, event_id=event_id, tmin=tmin, tmax=tmax, sfreq=sfreq)

epochs = mne.Epochs(raw, events, tmin=tmin, tmax=tmax, event_id=event_id,
                    preload=True, metadata=metadata)
epochs.metadata
```
Screen Shot 2021-02-03 at 12 36 34

Conditionally selecting a subset of Epochs is easy thanks to Pandas query strings:

```python
epochs['`visual/right` and button']
```
```
<Epochs |  7 events (all good), -0.499488 - 1.00064 sec, baseline [-0.499488, 0] sec, ~21.4 MB, data loaded, with metadata,
 'auditory/left': 2
 'button': 3
 'face': 1
 'visual/right': 1>
```

```python
epochs['`visual/right` and not `visual/left`']
```
```
<Epochs |  80 events (all good), -0.499488 - 1.00064 sec, baseline [-0.499488, 0] sec, ~210.3 MB, data loaded, with metadata,
 'auditory/left': 38
 'button': 3
 'face': 1
 'visual/right': 38>
```

This sort-of addresses #8360 as well.

I'd like to hear your opinions on this.

I've specifically tried to create a function that produces very generic data to ensure maximum versatility. I believe this set of metadata can also serve as a skeleton / starting point for users, as they're free to manipulate the data further before adding it to their Epochs.

cc @SophieHerbst


hoechenberger:
WIP: Handling repeated events within the same time period

agramfort (commented):
@hoechenberger can you make / update an example that gets rendered in the doc so people can get a quick way to get started?

I am personally super excited about this. It makes certain things suddenly so much easier to do !

hoechenberger:
I will add it to a tutorial @agramfort

mne/epochs.py
```diff
@@ -2052,6 +2052,115 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+    # This follows the approach taken in mne.Epochs
+    start_sample = int(round(tmin * sfreq))
+    stop_sample = int(round(tmax * sfreq)) + 1
+
+    # Make indexing easier
+    # We create the DataFrame before subsetting the events so we end up with
+    # indices corresponding to the original event indices. Not used for now,
+    # but might come in handy sometime later
+    events_df = pd.DataFrame(events, columns=('sample', 'prev_id', 'id'))
+    id_to_name_map = dict(zip(event_id.values(), event_id.keys()))
```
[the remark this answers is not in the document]
drammock: Missed changing the last `)` to `}` here

drammock:
> I am personally super excited about this. It makes certain things suddenly so much easier to do !

Agree, this looks like a very useful addition

mne/epochs.py
```diff
@@ -2052,6 +2052,115 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
     return perc
 
 
+def make_metadata(*, events, event_id, tmin, tmax, sfreq):
```
larsoner, line 2055: I would not lead with `*`, it makes it harder to use and decreases the likelihood of what I think would be rare errors. In this specific case, this order follows that of Epochs already.

In the general case, I think the principle currently applied here would generalize to "never allow people to use positional arguments in new functions and methods, they should all start with `*`", which I don't think is what we want. I think we should use `*` sparingly to prevent bad errors where people can get in trouble mixing up `verbose` with something else like we did in `apply_baseline` for example.

mne/epochs.py
```diff
@@ -2052,6 +2052,115 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+    Returns
+    -------
+    metadata : pandas.DataFrame
+        The pre-assembled metadata.
+    events : array, shape (n, 3)
+        The events corresponding to the generated metadata, i.e. one
+        time-locked event per row.
+    """
+    _check_pandas_installed()
+    import pandas as pd
```
larsoner, line 2098:
```suggestion
    pd = _check_pandas_installed()
```

larsoner (commented):
Looks like a useful addition to me. Probably worth adding to an example or two to see how it's useful. Maybe an epochs tutorial and/or metadata tutorial, then also maybe the reaction times example if it actually helps there?

mne/epochs.py
hoechenberger: Oh, that is much nicer indeed ☺️

mne/epochs.py
hoechenberger, line 2110:
```suggestion
    id_to_name_map = {v: k for k, v in event_id.items())}
```

mne/epochs.py
hoechenberger: I wanted to make at least part of the params kw-only because this function is so new and we might want to change the keyword order as it evolves. Thought this would make for a more painless transition then. But I agree that making **all** params kw-only is probably a little bit too much.

mne/epochs.py
```diff
@@ -2052,6 +2052,114 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
 
 
+def make_metadata(events, event_id, tmin, tmax, sfreq):
+    """Generate a default set of `~mne.Epochs` metadata.
+
+    This function generated metadata based on all events falling into the time
+    interval ``[tmin, tmax]``. Only events specified in ``event_id`` will be
+    considered.
+
+    The function will also return a subset the events array that corresponds to
```
agramfort, line 2062:
```suggestion
    The function will also return a subset of the events array that corresponds to
```

agramfort (commented):
besides some documentation I think it's good to go. I've already seen a successful use of this code for data with reaction times.

hoechenberger:
@larsoner @drammock @agramfort 
This is ready for review.

I suggest you start with the tutorial I created to get an idea of how the various parameters work:
https://25350-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_40_autogenerate_metadata.html

mne/epochs.py
```diff
@@ -2064,6 +2064,280 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+    from every time period. To keep the **last occurrence** of a certain
+    event instead, pass ``keep_last``.
+
+    Lastly, you may specify hierarchical event descriptors (HED) in
+    ``keep_first`` and ``keep_last`` to keep the time of the first and last
+    occurrence, respectively, of an entire group of different events.
+    For example, if you have the following hierarchical event names:
+    ``stim/a``, ``stim/b`` and pass ``keep_first='stim'``, the time of
+    **either** of the two events will be extracted, depending on which of the
+    two occurred first in the current time period.
```
drammock, line 2092: I know you said I should look at the tutorial first, but I'm not doing that: I should be able to read the docstring and basically grasp how this works.  I tried that and failed, so what follows is me thinking out loud about what it might mean, in order to hopefully point the way to a better explanation / organization of this info.

1. The very first line says it generates Epochs metadata, so I'm surprised to realize it doesn't take an Epochs object.
2. I can't tell: what exactly is "default" about the metadata that it generates?
3. are `tmin`/`tmax` relative to something?  since you're not passing Raw or Epochs, I guess they must be relative to `time_locked_events`?
4. OK, I'm starting to understand I think:  You provide an events array, specify which Event IDs you intend to later use to make epochs (this is `time_locked_events`, right?), and you specify an `event_id` dict and a `tmin`/`tmax` to determine what *other* events to care about, and then this function does some magic and returns a smaller events array (to pass to epochs constructor) and a dataframe with some info about all those other events in it.
5. Early on it says it uses *all* events but later you talk about only using the first (or last) event between `tmin` and `tmax`.  So it doesn't actually use all events?
6. Not clear to me yet whether `keep_first` and `keep_last` are mutually exclusive...  I suppose I could `keep_first='buttonpress'` and `keep_last='screen_flip'` or something?

mne/epochs.py
```diff
@@ -2064,6 +2064,280 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+        different to include more or fewer events in the metadata.
+    sfreq : float
+        The sampling frequency during data acquisiton.
+    time_locked_events : list of str | str | None
+        This can be a subset of event names (keys of ``event_id``). Metadata
+        rows will only be created for the specified event(s), instead of for
+        all events present in ``event_id``. If ``None`` (default), create
+        one row for each event in ``event_id``.
+    keep_first : str | list of str | None
+        Specify hierarchical event descriptors(s) (HED) of events from which
```
drammock, line 2122: do these *have to be* HEDs? I can imagine a case of multiple buttonpresses where I only want the first one (but there is no `button/left` or `button/right`, just one `button`)

mne/epochs.py
```diff
@@ -2064,6 +2064,280 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+          values indicating whether that specific event occurred within the
+          given time period
+
+        - if applicable, additional columns for ``keep_first`` and
+          ``keep_last`` event names, if HEDs were specified; the values will
+          be strings indicating which event was matched through the HED
+          pattern
+
+        - columns named ``[event_name]_time``, whose values indicate the time
+          of event occurrence in seconds, relative to the time-locked event
```
drammock, line 2153: I think a digested version of this info ought to be in the introductory description.  i.e., something like

> This function mimics the epoching process (constructs time windows around events of interest) and collates information about other events that occurred within those time windows. The information is returned as a pandas dataframe suitable for use as epochs metadata: one row per time-locked event, and columns indicating presence/absence and latency of each ancillary event type.

drammock: writing that makes me wonder why the presence/absence column is needed; couldn't `np.nan` in the latency column be enough information to indicate absence?

drammock (commented):
I've reviewed only the docstring so far.

hoechenberger:
Thank you for your review, @drammock! I will try to address your comments very soon.

mne/epochs.py
hoechenberger: > writing that makes me wonder why the presence/absence column is needed; couldn't `np.nan` in the latency column be enough information to indicate absence?

This is a fair point and I think eventually it's a bit a question of preference.

- To me, it's foremost a semantic consideration: The names of the boolean presence/absence columns are the exact same as the event names (i.e., those in `event_id` or `keep_first`/`keep_last`). I wouldn't feel comfy putting latencies in there as it wouldn't be clear that these **are** latencies, at least not without using multiindex columns, which are not supported by Epochs metadata.

- A second argument pertains to indexing. Look at this example:

```python
> print(epochs.metadata[['response/right', 'response/right_time']].head())

   response/right  response/right_time
0           False                  NaN
1           False                  NaN
2            True             0.508789
3           False                  NaN
4           False                  NaN
```

Now, assume we want to use `DataFrame.query()` to retrieve all rows where we have a `response/right` event. If `response/right` is a boolean, it's straightforward:

```python
> print(epochs.metadata
        .query('`response/right`')
        [['response/right', 'response/right_time']]
        .head())

    response/right  response/right_time
2             True             0.508789
8             True             0.484375
10            True             0.473633
12            True             0.458984
13            True             0.401367
```

If, however, the only way to figure whether an event occurred is by checking whether its value (latency) is `NaN`, things get messy, as there is no simple way to do this check using the `query()` method. Instead, one can abuse the fact that `NaN` doesn't equal itself:
```python
> print(epochs.metadata
        .query('`response/right_time` == `response/right_time`')
        [['response/right', 'response/right_time']]
        .head())

     response/right  response/right_time
2              True             0.508789
8              True             0.484375
10             True             0.473633
12             True             0.458984
13             True             0.401367
```
which works, but is pretty ugly.

mne/epochs.py
hoechenberger: I have tried to address this in 9560140

Is you probably can tell, I'm struggling a little due to the complex nature of all of this 😅
But is it clearer to you now?

mne/epochs.py
hoechenberger: > 4\. OK, I'm starting to understand I think:  You provide an events array, specify which Event IDs you intend to later use to make epochs (this is `time_locked_events`, right?), and you specify an `event_id` dict and a `tmin`/`tmax` to determine what _other_ events to care about, and then this function does some magic and returns a smaller events array (to pass to epochs constructor) and a dataframe with some info about all those other events in it.

Yes, that's precisely what happens 😅

> 6\. Not clear to me yet whether `keep_first` and `keep_last` are mutually exclusive...  I suppose I could `keep_first='buttonpress'` and `keep_last='screen_flip'` or something?

Yes, your example is entirely possible with the current implementation! The function just ensures that users cannot do `keep_first='button', keep_last='button'`.

hoechenberger: > 1\. The very first line says it generates Epochs metadata, so I'm surprised to realize it doesn't take an Epochs object.

I've tried to address this one now too.

hoechenberger: > 2\. I can't tell: what exactly is "default" about the metadata that it generates?

I had tried to outline this in the `Returns` section for the `metadata` return value, did you not see this or do you think this info is placed or written badly there?

mne/epochs.py
hoechenberger: > I think a digested version of this info ought to be in the introductory description. i.e., something like
> 
> > This function mimics the epoching process (constructs time windows around events of interest) and collates information about other events that occurred within those time windows. The information is returned as a pandas dataframe suitable for use as epochs metadata: one row per time-locked event, and columns indicating presence/absence and latency of each ancillary event type.

I've adopted this proposal now :)

mne/epochs.py
```diff
@@ -2064,6 +2064,289 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
 
+def make_metadata(events, event_id, tmin, tmax, sfreq,
+                  time_locked_events=None, keep_first=None, keep_last=None):
+    """Generate metadata from events for use with `~mne.Epochs`.
+
+    This function mimics the epoching process (constructs time windows around
+    events of interest) and collates information about other events that
+    occurred within those time windows. The information is returned as a
+    :class:`pandas.DataFrame` suitable for use as `~mne.Epochs` metadata: one
+    row per time-locked event, and columns indicating presence/absence and
```
drammock, line 2075: For consistency: 

```suggestion
    row per event of interest, and columns indicating presence/absence and
```

mne/epochs.py
```diff
@@ -2064,6 +2064,289 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+    latency of each ancillary event type.
+
+    Notably, the time window used for metadata generation need not correspond
+    to the time period used to create the `~mne.Epochs`; it may well be much
+    shorter or longer, if desired.
```
drammock, line 2080: I'd put this in `Notes` section, and maybe give there a very short example of why you'd want to do this.

mne/epochs.py
```diff
@@ -2064,6 +2064,289 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+
+    Parameters
+    ----------
+    events : array, shape (m, 3)
+        The events array.
+    event_id : dict
+        The mapping from event names (keys) to event IDs (values). Metadata
+        will only be generated for the events listed here. To restrict the
+        events for which metadata **rows** are created, pass the
+        ``time_locked_events`` parameter.
```
drammock, line 2093:
```suggestion
        A mapping from event names (keys) to event IDs (values). Only events
        included in ``event_id`` will be incorporated as columns of the
        returned metadata :class:`pandas.DataFrame`.
```

mne/epochs.py
```diff
@@ -2064,6 +2064,289 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+    tmax : float
+        End of the time interval for metadata generation in seconds, relative
+        to the time-locked event of the respective time period.
+    sfreq : float
+        The sampling frequency during data acquisiton.
+    time_locked_events : list of str | str | None
+        This can be a subset of event names (keys of ``event_id``). Metadata
+        rows will only be created for the specified event(s), instead of for
+        all events present in ``event_id``. If ``None`` (default), create
+        one row for each event in ``event_id``.
```
drammock, line 2113:
```suggestion
        Event types around which to create the time windows / for which to
        create *rows* in the returned metadata :class:`pandas.DataFrame`. If
        provided, the string(s) should be keys in ``event_id``. If ``None``
        (default), create rows for all event types present in ``event_id``.
```

drammock (commented):
another docstring-only review...  I'm still trying to make sure we get the API right, before worrying about the rest (tutorial, implementation).  My proposed re-writes of the `event_id` and `time_locked_events` parameter descriptions have led me to believe that the parameter names are wrong.  I think `event_id` should be something like "column_events" and `time_locked_events` should be something like "row_events".  WDYT?  Personally I think it will make it easier to understand / predict what the function will do.

I'm not convinced that what I'm proposing is 100% clear...  for example, in the initial docstring paragraph I'm realizing that "event of interest" is used to mean "row event" but it could easily be misinterpreted as "column event".

I also think we need some attention to detail when discussing events vs *event types*...  i.e., if you pass a single string to `time_locked_events` you will not get a single-row dataframe, but a dataframe with as many rows as there are occurrences of that *event type* in the events array.

hoechenberger:
@drammock, thank you for your repeated extensive reviews, I really appreciate it. I know this is boring and somewhat painful stuff.

> I think `event_id` should be something like "column_events" and `time_locked_events` should be something like "row_events". WDYT?

In the current implementation, `time_locked_events` (which will be the ones whose names will be listed in the `event_name` column) are actually necessarily as **subset** of ``event_id``. Introducing the terms `row_events` and `column_events`, I'd be worried this could be perceived as some sort of row-vs-column dichotomy, making users believe a `row` event cannot be a `column` event too… but maybe I'm totally overthinking this?

> I also think we need some attention to detail when discussing events vs _event types_..

Absolutely, I realized that too. So to get the semantics clear: an **event** consists of an **event ID** and an associated **event name**. I'll try to rework the docstring to strictly follow this pattern.

hoechenberger:
> I also think we need some attention to detail when discussing events vs _event types_... i.e., if you pass a single string to `time_locked_events` you will not get a single-row dataframe, but a dataframe with as many rows as there are occurrences of that _event type_ in the events array.

… the entire event **type** thing is further complicated by the fact that we support selecting entire – subsets? groups? types? kinds? – of events whose names are hierarchical event descriptors (HEDs).

So when we have event names `stim/a`, `stim/b`, and `stim/c` – what's `stim`? Currently I'm just awkwardly phrasing sentences that try to avoid giving `stim` a name, e.g.:

           The ``keep_first``
           parameter can be used to specify subsets of HEDs, effectively
           creating a new event that is the union of the matching HED pattern,
           and keeping the very first event of this set.

I'm afraid nobody will be able to read this once and understand what the heck is supposed to happen 🤯

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+
+# topoplot at t=0.05 s, averaged from 0.05-0.05 = 0 to 0.05+0.05 = 0.1 s
+resp_erp_incorrect.plot_topomap(times=0.05, average=0.05, size=3,
+                                title='Avg. topography 0–100 ms after '
+                                      'incorrect responses')
+
+###############################################################################
+# We can see a strong negative deflection immediately after incorrect
+# responses, compared to correct responses. The topoplot, too, leaves no doubt:
+# what we're looking at is the ERN.
```
agramfort, line 412:
```suggestion
# what we're looking at the ERN.
```

mne/epochs.py
```diff
@@ -2064,6 +2064,289 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
     return perc
 
 
+def make_metadata(events, event_id, tmin, tmax, sfreq,
+                  time_locked_events=None, keep_first=None, keep_last=None):
```
agramfort, line 2068:
```suggestion
                  epochs_event_id=None, keep_first=None, keep_last=None):
```
I find it more explicit that it's the event_id that will be used to create epochs.
Also it could be a dict or a list. If it's dict you can just call list on it.

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# the initial event named "response/left" and "response/right"
+print(metadata['response'])
+
+###############################################################################
+# We're facing a similar issue with the stimulus events, and now there are not
+# only two, but **four** different types: ``stimulus/compatible/target_left``,
+# ``stimulus/compatible/target_right``, ``stimulus/incompatible/target_left``,
+# and ``stimulus/incompatible/target_right``. Even more, because in the present
+# paradigm stimuli were presented in rapid succession, sometimes multiple
+# stimulus events occurred within the 1.5 second time period were using to
```
agramfort, line 200: were?

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# generate our metadata. See for example:
+
+metadata.loc[metadata['stimulus/compatible/target_left'] &
+             metadata['stimulus/compatible/target_right'],
+             :]
+
+###############################################################################
+# This can easily lead to lead to confusion during later staged of processing,
```
agramfort, line 208:
```suggestion
# This can easily lead to confusion during later staged of processing,
```

agramfort (commented):
LGTM besides these questionable nitpicks.

now I'd be curious to have someone with a fresh eye on this
as we have discussed together so maybe I get things because
of our discussions.

mne/epochs.py
hoechenberger: Addressed in caa36e9

mne/epochs.py
hoechenberger: @drammock suggested to change this too, and proposed to use the terms "row_events" (instead of time_locked_events) and "column_events" (instead of event_id).

Not sure which proposal is better, but you two independently suggesting a name change makes it clear **something** should be changed :)

hoechenberger: > Also it could be a dict or a list. If it's dict you can just call list on it.

Not sure if we need to support the `dict` case, as we're returning a new `event_id` dict anyway, so there's no need for users to create it manually

mne/epochs.py
```diff
@@ -2062,6 +2062,298 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+        else:
+            return list(x)
+
+    row_events = _ensure_list(row_events)
+    keep_first = _ensure_list(keep_first)
+    keep_last = _ensure_list(keep_last)
+
+    if set(keep_last) & set(keep_first):
+        raise ValueError('The event names in keep_first and keep_last must be '
+                         'mutually exclusive')
```
agramfort, line 2211: say what you got to have a more explicit error message

mne/epochs.py
```diff
@@ -2062,6 +2062,298 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+
+    # Event names
+    metadata[metadata.columns[0]] = ''
+
+    # First set of columns: boolean values indicating whether the
+    # respective event occurred in that epoch
+    start_idx = 1
+    stop_idx = start_idx + len(event_id.keys())
+    for col in metadata.columns[start_idx:stop_idx]:
+        metadata[col] = False
```
agramfort, line 2272: `metadata.iloc[:, start_idx:stop_idx] = False`

should be enough

mne/epochs.py
```diff
@@ -2062,6 +2062,298 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+
+    # Second set of columns: keep_first and keep_last
+    start_idx = stop_idx
+    stop_idx = start_idx + len(keep_first_cols + keep_last_cols)
+    for col in metadata.columns[start_idx:stop_idx]:
+        metadata[col] = None
```
agramfort, line 2278: iloc too here?

mne/epochs.py
```diff
@@ -2062,6 +2062,298 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+
+    # Last set of columns: event times
+    start_idx = stop_idx
+    for col in metadata.columns[start_idx:]:
+        metadata[col] = np.nan
```
agramfort, line 2283: idem

mne/epochs.py
```diff
@@ -2062,6 +2062,298 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+        epoch_start_sample = time_locked_event['sample'] + start_sample
+        epoch_stop_sample = time_locked_event['sample'] + stop_sample
+        events_in_epoch = events_df.loc[
+            (events_df['sample'] >= epoch_start_sample) &
+            (events_df['sample'] <= epoch_stop_sample), :]
+
+        assert not events_in_epoch.empty
+
+        # Store the metadata
+        for _, epochs_event in events_in_epoch.iterrows():
```
agramfort, line 2304: I think itertuples is faster then iterrows but It probably does not matter here

mne/tests/test_epochs.py
```diff
@@ -2890,6 +2890,75 @@ def assert_metadata_equal(got, exp):
...
+    keep_last = [] if keep_last is None else keep_last
+    event_names = sorted(set(event_id.keys()) | set(keep_first) |
+                         set(keep_last))
+
+    for event_name in event_names:
+        assert event_name in metadata.columns
+        assert f'{event_name}_time' in metadata.columns
+
+    # Check the time-locked event's metadata
+    for _, row in metadata.iterrows():
```
agramfort, line 2941: itertuples?

agramfort (commented):
here is a last batch of reviews @hoechenberger !

hoechenberger:
@agramfort I've addressed your comments, except for the adoption of `df.itertuples()` (instead of `iterrorws()`) in the tests, because we have column names with forward slashes, which won't be valid  `NamedTuple` keys

agramfort (approved):
@drammock @cbrnr ok for you?

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+
+###############################################################################
+# Keeping only the first events of a group
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The metadata now contains 400 rows – one per stimulation – and the same
+# number of columns as before. Great!
+#
+# We have two types of responses in our data: ``response/left`` and
+# ``respnse/right``. We would like to map those to "correct" and "incorrect".
```
cbrnr, line 167:
```suggestion
# ``response/right``. We would like to map those to "correct" and "incorrect".
```

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# To make this easier, we can ask `~mne.epochs.make_metadata` to generate an
+# entirely **new** column that refers to the first response observed during the
+# given time interval. This works by passing a subset of the hierarchical
+# event descriptions (HEDs) used to name events via the ``keep_first``
+# parameter. For example, in the case of the HEDs``response/left`` and
```
cbrnr, line 172:
```suggestion
# parameter. For example, in the case of the HEDs ``response/left`` and
```

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# ``respnse/right``, we could pass ``keep_first='response'`` to generate a new
```
cbrnr, line 173:
```suggestion
# ``response/right``, we could pass ``keep_first='response'`` to generate a new
```

cbrnr (commented):
Besides my minor corrections, I've noticed that I cannot scroll the wide tables horizontally (macOS, Chrome). There's no horizontal scrollbar...

hoechenberger:
> I've noticed that I cannot scroll the wide tables horizontally (macOS, Chrome)

In Safari, I have no scrollbar in the "dataframe element" either; instead, the table is visible in full width, which is just as annoying:

cbrnr:
You're right, I didn't see the horizontal scrollbar at the bottom because I expected it to be in the frame (just like in the other output frames).

Also, I forgot the most important comment: the tutorial is super well written! Really nice addition!

drammock:
I think the dataframe problem relates to our website theme update.  I'll look into it, after I finish reviewing this PR

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# include information about events from only a fraction of the epochs duration;
+# or it might include events that occurred well outside a given epoch.
+#
+# Let us look at a concrete example. In the Flankers task of the ERP CORE
+# dataset, participants were required to respond to visual stimuli by
+# pressing a button. We're interested in looking at the visual evoked responses
+# (ERPs) of trials with correct responses. Based on literature studies, we
+# decide that responses later than 1500 ms after stimulus onset are to be
+# considered invalid, because they don't capture the neuronal processes of
+# interest. We can approach this  in the following way with the help of
```
drammock, line 90: might be nice to cite one or two sources to justify our cutoff of 1.5 sec.

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# We can see that the generated table has 802 rows, each one corresponding to
+# an individual event in ``all_events``. The first column, ``event_name``,
+# contains the name of the respective event around which the metadata of that
+# specific column was generated – we'll call that the "time-locked event",
+# because we'll assign it time point zero.
+#
+# Unless you have a very wide screen, you'll most likely have to scroll
+# horizontally to look at the rest of the columns that were automatically
+# created.
+#
```
drammock, line 120:
```suggestion
```

This will get fixed in the site theme.

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+# The names of columns 2 to 7 correspond to the event names specified in the
+# ``all_event_id`` dictionary. These columns contain boolean values – ``True``
+# or ``False`` – indicating whether that specific event occurred during the
```
drammock, line 123:
```suggestion
# ``all_event_id`` dictionary. These columns contain boolean values — ``True``
# or ``False`` — indicating whether that specific event occurred during the
```

em-dashes, not en-dashes

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+
+###############################################################################
+# Time to calculate the ERPs for correct  and incorrect responses.
+# For visualization, we'll only look at sensor ``FCz``, which is known to show
+# the ERN nicely in the given paradigm. We'll also create a topoplot to get an
+# impression of the average scalp potentials measured in the first 100 ms after
+# an incorrect response.
+
+resp_erp_correct = epochs['response_correct == True'].average()
+resp_erp_incorrect = epochs['response_correct == False'].average()
```
drammock, line 397: comparing to `bool` with `==`?  Why not `epochs['response_correct']` and `epochs['~response_correct']`

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+
+mne.viz.plot_compare_evokeds({'Correct Response': resp_erp_correct,
+                              'Incorrect Response': resp_erp_incorrect},
+                             picks='FCz', show_sensors=True,
+                             title='ERPs at FCz, time-locked to response')
+
+# topoplot at t=0.05 s, averaged from 0.05-0.05 = 0 to 0.05+0.05 = 0.1 s
```
drammock, line 404:
```suggestion
# topoplot of average field from time 0.0-0.1 s
```

drammock (commented):
First, I agree with @cbrnr that this is a very well-written tutorial. Most of my comments there are nitpicks. I have once again offered a lot of feedback on the function docstring, and I have three bigger-picture comments to make here:

1. I think the tutorial is pushing the limits of being too long, but I think it's fine to ignore that for now, and come back later for a second pass of tightening up the prose (I can do it after the sprint, for example, when I'm less busy).

2. I'm still not convinced that the metadata needs to have separate boolean columns in addition to latency.  I think your main objection before was that it was awkward to get back just the non-nan latencies, but I think it's pretty easy with `metadata['event_name'].notna()`, or if you want the whole row, `metadata[metadata['event_name'].notna()]`. Is there some other reason to hold on to the boolean columns?

3. I'm still not sure we've found the best/clearest parameter names.  Two other ideas for `time_locked_events`/`row_events` that I've had during this review are `tzero_events` and `epoching_events` (or maybe just `epoch_events`).

drammock:
incidentally, codecov makes it almost impossible to review source changes now:

Screenshot_2021-03-01_09-37-22

I wonder if we should disable these in-line coverage annotations?  Seems like they're helpful if there's just 1 or 2 of them, and they're crippling when there are more than that.

larsoner:
Agreed codecov has been a painful, feel free to open a PR or push a commit to `main` to disable inline comments

hoechenberger:
> 2\. I'm still not convinced that the metadata needs to have separate boolean columns in addition to latency.  I think your main objection before was that it was awkward to get back just the non-nan latencies, but I think it's pretty easy with `metadata['event_name'].notna()`, or if you want the whole row, `metadata[metadata['event_name'].notna()]`. Is there some other reason to hold on to the boolean columns?

As soon as we move away from the pure pandas domain, over to MNE Epochs land, indexing (which happens via `DataFrame.query()` internally) becomes more awkward, at least that's what I thought. I just googled a bit more, and according to [this posting on SO](https://stackoverflow.com/a/46021212) one might be able to use `'foo.notnull()'` in a query string. I haven't tested this yet.

My other argument would still stand, though:
If you provide `keep_first` or `keep_last` referring to a set of multiple event types, you need a way to easily expose **which** event type actually **was** the first or last. For example, say you have `stim/a` and `stim/b` in your data, and provide `keep_first='stim'`, this will create the columns `stim` and `stim_time`. Assuming a time window where `stim/1` occurred before `stim/2`, we'd store `'2'` in the `stim` column, making it very easy to find out which stimulus type is associated with the latency in `stim_time`. If `stim` contained only a boolean value, one would have to manually look at `stim/1_time` and `stim/2_time` in order to figure things out.

> 3\. I'm still not sure we've found the best/clearest parameter names.  Two other ideas for `time_locked_events`/`row_events` that I've had during this review are `tzero_events` and `epoching_events` (or maybe just `epoch_events`).

I'm also not fully convinced by the choices we've made. I'd be okay with leaving the parameters names as they are for now, and maybe coming up with better names at a later time, hopefully before the next MNE release :)

cbrnr:
@drammock pressing "a" toggles comments - just like in our raw browser 😄!

cbrnr:
> I'm also not fully convinced by the choices we've made. I'd be okay with leaving the parameters names as they are for now, and maybe coming up with better names at a later time, hopefully before the next MNE release :)

It might be worth coming up with good names now, because later means a painful deprecation cycle.

hoechenberger:
> It might be worth coming up with good names now, because later means a painful deprecation cycle.

YEs that's why I suggested, hopefully before the next release ;) Mhh!! thinking, thinking … 🤯

cbrnr:
It's probably best to decide on the names in this PR. That said, I don't have better suggestions right now.

drammock:
Let's settle the API now. Candidate parameter names for "thing that determines which events to construct intervals around / to use as the basis of each row" are:

- `row_events` 🤖 
- `epoch_events` 🧠 
- `epoching_events` 🦒 
- `epoch_event_id` 🐸 
- `tzero_events` ⌚ 
- `time_locked_events` 🔒 

Vote with your reaction! I'll react once for each option just to populate the buttons for everyone, but FYI my vote is for `epoch_events`

drammock:
> As soon as we move away from the pure pandas domain, over to MNE Epochs land, indexing (which happens via `DataFrame.query()` internally) becomes more awkward, at least that's what I thought.

```python
import os
import numpy as np
import pandas as pd
import mne

kiloword_data_folder = mne.datasets.kiloword.data_path()
kiloword_data_file = os.path.join(kiloword_data_folder,
                                  'kword_metadata-epo.fif')
epochs = mne.read_epochs(kiloword_data_file)
epochs = epochs[:5]  # for simplicity
epochs.metadata['foo'] = [np.nan, 0.3, 0.77, np.nan, 0.51]
# this works:
epochs['foo.notna()']
```
@hoechenberger  this works the way you want it to, I think: `epochs['foo.notna()']` returns just the epochs for which metadata column "foo" had a non-null entry.  The same operation on the metadata is a little more verbose to type, but still not that complicated to my eye:
```python
epochs.metadata[epochs.metadata['foo'].notna()]
```

hoechenberger:
> Let's settle the API now. Candidate parameter names for "thing that determines which events to construct intervals around / to use as the basis of each row" are:

Seems that `epoch_events` won 👍

hoechenberger:
@drammock 
> @hoechenberger this works the way you want it to, I think: `epochs['foo.notna()']` returns just the epochs for which metadata column "foo" had a non-null entry.

Perfect! I **thought** I had tried exactly this and ran into some issues, but maybe I did something wrong. Will try again :)

We still need to decide upon this part, though

@hoechenberger 
> If you provide `keep_first` or `keep_last` referring to a set of multiple event types, you need a way to easily expose **which** event type actually **was** the first or last. For example, say you have `stim/a` and `stim/b` in your data, and provide `keep_first='stim'`, this will create the columns `stim` and `stim_time`. Assuming a time window where `stim/1` occurred before `stim/2`, we'd store `'1'` in the `stim` column, making it very easy to find out which stimulus type is associated with the latency in `stim_time`. If `stim` contained only a boolean value, one would have to manually look at `stim/1_time` and `stim/2_time` in order to figure things out.

tutorials/epochs/plot_40_autogenerate_metadata.py
hoechenberger: I know this is bad practice, but I felt this was "more readable" for the `False` part. But since we shouldn't teach bad practices, I fully embrace your proposal :)

drammock:
>  say you have `stim/a` and `stim/b` in your data, and provide `keep_first='stim'`, this will create the columns `stim` and `stim_time`.

Agreed.

> Assuming a time window where `stim/a` occurred before `stim/b`, we'd store `'a'` in the `stim` column, making it very easy to find out which stimulus type is associated with the latency in `stim_time`.

Yep.

> If `stim` contained only a boolean value, one would have to manually look at `stim/a_time` and `stim/b_time` in order to figure things out.

In my proposal `stim` would never contain just a boolean value, it would contain either a latency or `nan`.  But I see now the misunderstanding:  I have been advocating for getting rid of the boolean columns *that are associated with `stim/a`, `stim/b`, etc*, not the ones assocated with `keep_first` or `keep_last`.  I agree that if you use `keep_first` or `keep_last` then you would need to generate an extra column to keep track of which of the various `stim/*` events you were recording the latency for.  So in my view, if all you have is `stim/a` and `stim/b` and you pass `keep_first='stim'`, then you end up with 4 columns:

```
    stim/a   stim/b   stim   stim_time
0     0.3      0.4      a       0.3
0     0.3      0.2      b       0.2
...
```

That is not a great result because the HED-derived column has `_time` but the other ones don't.  So I would revise the above proposal to be:

```
    stim/a   stim/b   which_stim   stim
0     0.3      0.4         a        0.3
0     0.3      0.2         b        0.2
...
```

OR 

```
    stim/a_time   stim/b_time   stim   stim_time
0       0.3           0.4         a       0.3
0       0.3           0.2         b       0.2
...
```

I like the first version because it adds an affix in fewer cases, which means the dataframes are not as wide and there is less typing required to access the columns.  But I could live with either option.

tutorials/epochs/plot_40_autogenerate_metadata.py
hoechenberger: I thought em-dashes are typically used without spaces around them, though? Which is the main reason for me to use en-dashes when I want to insert (grammatical) parantheses without having to use `(` and `)`

mne/epochs.py
```diff
@@ -2062,6 +2062,301 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+           If you are planning to attach the generated metadata to
+           `~mne.Epochs` and intend to include only events that fall inside
+           your epochs time interval, pass the same ``tmin`` and ``tmax``
+           values as you use for your epochs.
+
+    tmax : float
+        End of the time interval for metadata generation in seconds, relative
+        to the time-locked event of the respective time window.
+    sfreq : float
+        The sampling frequency during data acquisiton.
```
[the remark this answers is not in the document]
hoechenberger: WDYT about something like

```
The sampling frequency of the data belonging to the provided events array.
```

I think that has less potential for confusion :)

mne/epochs.py
```diff
@@ -2062,6 +2062,301 @@ def _drop_log_stats(drop_log, ignore=('IGNORED',)):
...
+
+        .. note::
+           If you are planning to attach the generated metadata to
+           `~mne.Epochs` and intend to include only events that fall inside
+           your epochs time interval, pass the same ``tmin`` and ``tmax``
+           values as you use for your epochs.
+
+    tmax : float
+        End of the time interval for metadata generation in seconds, relative
+        to the time-locked event of the respective time window.
```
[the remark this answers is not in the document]
hoechenberger: Absolutely, I was thinking about that too, but wasn't sure whether this is "valid". Will try and see what Sphinx tell me!

tutorials/epochs/plot_40_autogenerate_metadata.py
drammock: em-dash is the correct mark to use for parenthetical comments/asides when you don't want to use `(` and `)`.  Whether to put spaces around the em-dash is, as far as I know, up to the typesetter.  I often see them both with and without surrounding spaces, and I personally think *with* spaces is easier to read.

tutorials/epochs/plot_40_autogenerate_metadata.py
drammock: you can use `not response_correct` instead of `~response_correct` if readability is a concern.

mne/epochs.py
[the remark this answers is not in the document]
drammock: `The sampling frequency of the data from which the events array was extracted`?  I dunno, your last suggestion is about as good as this one IMO. Go with whichever you like.

mne/epochs.py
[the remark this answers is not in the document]
drammock: I think spinx will let you.  I wonder if @agramfort will let you though :)

hoechenberger:
> Agreed codecov has been a painful, feel free to open a PR or push a commit to `main` to disable inline comments

I suppose this is related to some tests permanently failing, no? Once we fix those, all should be fine again … I hope

tutorials/epochs/plot_40_autogenerate_metadata.py
hoechenberger: Interesting, I've never seen em-dashes **with** spaces in English books. But yeah, maybe a question of the typesetters' preferences indeed 👌

drammock:
> for `keep_first` event types we'd introduce a `name_first` column, and for `keep_last` a `name_last` column accordingly

brilliant 💯

hoechenberger:
Since I just realized I'm not sure whether to pick `stim_first` & `stim_last` or `first_stim` & `last_stim` as column names, and because voting is so much fun – whaddy'all think?

- 🎉  `stim_first` & `stim_last`
-  🚀  `first_stim` & `last_stim`

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+
+Auto-generating ``Epochs`` metadata
+===================================
+
+This tutorial shows how to auto-generate metadata for `~mne.Epochs`, based on
+events via `mne.epochs.make_metadata`.
+
+We are going to use data from the :ref:`erp-core-dataset` (derived from
+:footcite:`Kappenman2021`). This is EEG data from a single participant
+performing an active visual task (Flankers task).
```
hoechenberger, line 12:
```suggestion
performing an active visual task (Eriksen flanker task).
```

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -0,0 +1,449 @@
...
+raw data. The data is pretty clean and contains very few blinks, so there's no
+need to apply sophisticated preprocessing and data cleaning procedures.
+"""
+
+from pathlib import Path
+import matplotlib.pyplot as plt
+import mne
+
+
+data_dir = Path(mne.datasets.erp_core.data_path())
```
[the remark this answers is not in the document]
hoechenberger: Is that an open question or are you asking me to do something? :) (I wouldn't know WHAT to do, exactly! But can have a look at the Circle config to hopefully figure things out)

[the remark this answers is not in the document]
drammock: It's kind of an open question...  I thought we had something in place that pre-populated the circle system with all the datasets in advance, before any of the tutorials were run?  Or maybe that would be dumb because on most PR runs circle only builds 0-2 tutorials?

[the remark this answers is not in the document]
larsoner: > I thought we had something in place that pre-populated the circle system with all the datasets in advance

Just the ones that are used according to a suitable regex. You need to add `erp_core` to [`tools/circleci_download.sh`](https://github.com/mne-tools/mne-python/blob/main/tools/circleci_download.sh). For any new dataset my plan is to have people look at https://github.com/mne-tools/mne-python/pull/8867 to see the minimal set of files (9 including the tutorial that uses the dataset) that need to be changed to have a complete new-dataset PR, so it would be good to look to see if you've changed the appropriate set of files @hoechenberger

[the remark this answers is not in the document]
hoechenberger: Got it, will have a look.

hoechenberger:
Sorry I have to bug you all again, but when implementing the discussed changes and playing through the tutorial several times, I now feel the urge to add a slight modification.

In the beginning, we had two columns per stimulus:
-  one named `stimulus`, containing booleans indicating whether the stimulus was present in the current time window;
- and the other named `stimulus_time` with the latency.

We then decided to drop the boolean columns due to their redundancy, as absence can be indicated by `NaN` in the latency columns too. All good so far.

But we also opted to rename to drop the `_time` suffix from the latency columns' names. So what used to be `stimulus_time` is now simply `stimulus`. And this is what bothers me now.
I now have code looking like this:

```python
epochs['response < 1.5 & first_response == "left"']
```

And I think this is semantically less … pleasant than:

```python
epochs['response_time < 1.5 & first_response == "left"']
```

So here is my simple question: **Can we have the `_time` back** (or, if you will, `_latency` or whichever suffix might seem appropriate)? We could even make the suffix user-configurable if we cannot reach a clear consensus here. But I'd love to have my columns containing "time" named accordingly.

Looking forward to hearing your thoughts!

😇🦄🥸

drammock:
> I think this is semantically less … pleasant

it doesn't bother me as much as it seems to bother you. For me, downsides that outweigh it are:

1. not being able to see my whole dataframe in the terminal because the column names are all 5 characters longer than I think they need to be.
2. having to type the extra underscore & word in a context where tab completion is unavailable (i.e., within a query string)

If users really want a suffix there is [pandas.DataFrame.add_suffix](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.add_suffix.html).  If we really can't reach agreement I would settle for a `suffix='_time'` parameter so I can specify `suffix=''`.

agramfort:
I have no opinion :)

>

cbrnr:
I agree with @drammock. It's pretty clear without `_time` and the benefits are more important to me.

agramfort:
let's remove the _time then and l merge during the day :)

>

hoechenberger:
Whoops, day is almost over, sorry ;) I'll finish it up within the next hours. Thanks for your feedback!

hoechenberger:
Doc artifacts:

Tutorial:
https://25600-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_40_autogenerate_metadata.html#sphx-glr-auto-tutorials-epochs-plot-40-autogenerate-metadata-py

API docs:
https://25600-1301584-gh.circle-artifacts.com/0/dev/generated/mne.epochs.make_metadata.html

tutorials/epochs/plot_40_autogenerate_metadata.py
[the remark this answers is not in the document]
hoechenberger: This is resolved now. Data should be downloaded by Circle before the tutorial is run.

agramfort:
thx @hoechenberger

drammock:
Many thanks for this!  I think it will be really useful for a lot of people and I appreciate your patience with all the API back-and-forth.

PR #8838, opened 2021-02-03
MAINT: Dont fail on coverage

### LINKED ISSUE #8776: Codecov coverage
I noticed that Codecov frequently reports very low coverage (e.g. #8760 currently shows Codecov jobs in red). Often, when I push additional (minor) changes, coverage suddenly increases although almost nothing changed. Therefore, I normally ignore Codecov reports, but this is of course not the idea of this action. Is there anything we can do to make Codecov more reliable? Or am I doing something wrong?

larsoner:
I don't like that PRs have their status set as failed even when CIs are still running. And even once all are done running, codecov reporting can be iffy. This PR sets codecov just to be informational which will make it always green. This will mean we have to manually check the coverage reports, but we already have to do that when we're interested since we can't fully trust the output.

Closes #8776

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,6 +8,7 @@ coverage:
   status:
     patch:
       default:
+        informational: true
         target: 95%
         if_no_uploads: error
         if_not_found: success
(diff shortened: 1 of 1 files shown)


cbrnr:
Fixes #8776?

larsoner:
Indeed and it seems to be behaving properly so in it goes!

PR #8840, opened 2021-02-04
WIP, ENH: Add FSL betsurf BEM generation

### LINKED ISSUE #7232: Inconsistent surfaces extracted using watershed
Dear MNE expert, 

I am working with a EEG dataset and I wish to run source on that. Unfortunately the T1w are not FLASH and I know that in the absence of FLASH images, MNE offers a somewhat less accurate
solution based on the watershed algorithm (see Jas et al., 2018). 

Indeed for some subjects in my dataset I got the following error when I am going to run `make_bem_model`. 

`RuntimeError: Surface inner skull is not completely inside surface outer skull`

For other subject instead the surfaces are simply wrong (see attached), even though `make_bem_model` hadn't give me any error 

Screen Shot 2020-01-20 at 14 13 39

Screen Shot 2020-01-20 at 14 13 04

I wan wondering whether if you have some tips on how to fix them. 

Thanks

Dave

larsoner:
Closes #7232

Gives us another way to compute meshes. They are not perfect but work for at least some subjects. For example the defaults (use `fsl` for everything) for two subjects are not perfect, but if we pass `--brainmask brainmask.mgz` to use the FreeSurfer-computed (and possibly manually fixed) brainmask then we're at least as good as `watershed_bem` at least for the skull surfaces (white is the original brain surface from the marching cubes):

Screenshot from 2021-02-04 16-23-26

Screenshot from 2021-02-04 16-18-20

Screenshot from 2021-02-04 16-18-04

Screenshot from 2021-02-04 16-19-54

I don't know why the nose gets killed in all of these but it's hopefully okay?

Without supplying `--brainmask` it fails on some pretty spectacularly like `sample`, probably because the T1 is not normalized how `bet` expects:

Screenshot from 2021-02-04 16-14-42

It takes about a minute if you use flirt, but if you pass `-t` to use the FreeSurfer registration it takes about 20 seconds.

Need to document and add `skimage` as a dependency for this, at least when using `brainmask`, because it uses their marching cubes algorithm. Eventually we could add things like Gaussian smoothing (scipy.ndimage) before marching cubes and/or Lapalacian filtering of the tesselated meshes (trimesh) but so far it doesn't seem to help much here.

Also need to add a test (maybe speed up the `fsaverage` one) that uses `-t` / `talairach`, and one that uses `brainmask`.


agramfort:
@larsoner I think we cannot decouple this new feature from the BEM solver as meshing issues and meshes too close from each other will blow up the numerics in the BEM. I don't think we have any safeguard for this in the BEM code. Basically looking at the images is not enough to assess if it's good enough for source imaging.

my 2c

larsoner:
> Basically looking at the images is not enough to assess if it's good enough for source imaging.

Maybe good enough for MEG-only / 1-layer, i.e., only care about the inner skull surface being more correct than `mri_watershed`?

But I see what you're saying. I know there is some code floating around for ensuring meshes have some sufficient spacing. What would you consider the minimum for our BEM solver? 1mm? 3mm?

agramfort:
>
>
> Basically looking at the images is not enough to assess if it's good
> enough for source imaging.
>
> Maybe good enough for MEG-only / 1-layer, i.e., only care about the inner
> skull surface being more correct than mri_watershed?
>
but for 1 layer the code we have is already quite good no?

> But I see what you're saying. I know there is some code floating around
> for ensuring meshes have some sufficient spacing. What would you consider
> the minimum for our BEM solver? 1mm? 3mm?
>
rule of thumb is that the triangles edge sizes should be comparable to the
distance between meshes and between inner surface and the dipoles in source
space

larsoner:
> but for 1 layer the code we have is already quite good no?

Often the `inner_skull` produced by FreeSurfer based on `brainmask.mgz` has some "bumps" where it shouldn't. You can see it in some of the above where the white (which often came from `brainmask.mgz` and was turned into `brain.surf` using marching cubes in Python) is bumpier than the red `inner_skull` surface produced by `betsurf`. In these examples it's subtle, for some subjects it's much more pronounced.

Arguably people should correct their `brainmask.mgz` not to have these problems, but this requires some finesse and FreeSurfer skills that I think are a bit much to ask of people.

This is not my area of expertise, though, it's entirely possible that these bumps don't matter or are somehow correct/reasonable, though it doesn't look like it to me by eye, and doesn't match what FieldTrip + SPM produces for the inner skull (ever) or what betsurf produces (in most cases).

> rule of thumb is that the triangles edge sizes should be comparable to the distance between meshes and between inner surface and the dipoles in source space

Nice, this gives a good way of quantifying how well these meshes would be usable in our source localization. And I have a bunch (~100 or so?) of adult MRIs with FLASH so I could actually compare FLASH/watershed/betsurf/FieldTrip+SPM (and anything else) directly several methods in principle. Maybe I'll just start with my own MRI. I am "three" in the images above, but I'm sure you recognized me from the high-fidelity outer_skin surface betsurf produced :)

@agramfort concretely would it make sense to start in a separate PR with some metrics to quantify 3-layer BEM mesh quality in terms of 1) mesh-to-mesh spacing, 2) within-mesh neighbor vertex spacing, 3) and inner-skull-to-white-surface spacings?

agramfort:
+1 to have a function that summarizes a BEM quality. Then we can call this
function on many subjects
and see how it compares.

>

larsoner:
Yikes that's a lot of conflicts :fearful: 

I'll close for now but hopefully I'll find time to come back to this at some point!

Issue #8825, opened 2021-02-02
ENH: Need tools for fixing mesh topology


yh-luo:
#### Describe the bug
Someone else's high-resolution head was used for the selected subject.

`lh.seghead` in FreeView looks like:
![subject-346-freeview](https://user-images.githubusercontent.com/29604854/106544018-d5d94e00-6541-11eb-8f85-c4a56f560d1c.png)
 
But the high-resolution head rendered in `mne coreg` gui looks like:
(Notice the differences of the top-left corner of the goggle, the ears, and the nose bridge.)
I can confirm that it is from other subject's `head-dense.fif` (`lh.seghead`) via FreeView.
![subject-346-coreg](https://user-images.githubusercontent.com/29604854/106544075-ef7a9580-6541-11eb-9520-567da8d19a94.png)

The high-resolution head also did not align to the head. No rotations were made, the head looks like:
![subject-346-coreg-head](https://user-images.githubusercontent.com/29604854/106544365-5c8e2b00-6542-11eb-8d1f-0f3e3d124321.png)

---
FreeView showed that `lh.seghead` and `outer_skin.surf` were aligned.
From the above images, it looks like someone else's dense head was used. The specified subject is **subject-346**, the error messages indicated that **subject-102** dense head was used instead?

subject-102 is the first subject in the subject dropdown dialog.
It seems to me that `mne coreg` gui renders the first subject and then update with the selected subject. Maybe the bug result from that process?
```python
mne coreg -s subject-346
Exception occurred in traits notification handler for object: <mne.gui._file_traits.SurfaceSource object at 0x7f3db86bcb30>, trait: file, old value: subjects//subject-102/bem/subject-102-head-dense.fif, new value: subjects//subject-346/bem/subject-346-head-dense.fif
Traceback (most recent call last):
  File "/home/ntubabylab/anaconda3/envs/mne/lib/python3.8/site-packages/traits/trait_notifiers.py", line 522, in _dispatch_change_event
    self.dispatch(handler, *args)
  File "/home/ntubabylab/anaconda3/envs/mne/lib/python3.8/site-packages/traits/trait_notifiers.py", line 484, in dispatch
    handler(*args)
  File "/home/ntubabylab/anaconda3/envs/mne/lib/python3.8/site-packages/mne/gui/_file_traits.py", line 175, in read_file
    bem = read_bem_surfaces(self.file, verbose=False)[0]
  File "<decorator-gen-98>", line 22, in read_bem_surfaces
[...]

PR #4585, opened 2017-09-19
MRG: Move datasets

larsoner:
WIP because I need to update some hashes, we need to finish uploading files, and I need to make sure the `brainstorm` dataset changes actually work.

PR #8841, opened 2021-02-05
MRG, ENH: write_head_bem for writing head surface files

yh-luo:
#### Reference issue
Discussed in #8825 .

- [X] `write_head_bem` for customized head surface
- [X] let `read_bem_surfaces` read in defective head surfaces, but warn
~~`fix_all_defects` (all credits to @christian-oreilly , just merge the codes into MNE-Python)~~
- [X] DOC
  - [X] `write_head_bem`
  - [X] `on_defects`
~~`fix_all_defects`~~
~~A tutorial/example to illustrate how to fix the topological defects~~

- [ ] `write_head_bem` for` mne make_scalp_surfaces`
- [X] Update: Editing BEM surfaces in Blender
- [ ] FAQ: My watershed BEM meshes look incorrect
- [X] `latest.inc`

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1231,7 +1231,8 @@ def _extract_volume_info(mgz):
 # Read
 
 @verbose
-def read_bem_surfaces(fname, patch_stats=False, s_id=None, verbose=None):
+def read_bem_surfaces(fname, patch_stats=False, s_id=None, on_defects='raise',
+                      verbose=None):
     """Read the BEM surfaces from a FIF file.
 
     Parameters
@@ -1244,6 +1245,7 @@ def read_bem_surfaces(fname, patch_stats=False, s_id=None, verbose=None):
         If int, only read and return the surface with the given s_id.
         An error will be raised if it doesn't exist. If None, all
         surfaces are read and returned.
+    %(on_defects)s
     %(verbose)s
 
     Returns
@@ -1267,7 +1269,7 @@ def read_bem_surfaces(fname, patch_stats=False, s_id=None, verbose=None):
         raise ValueError('surface with id %d not found' % s_id)
     for this in surf:
         if patch_stats or this['nn'] is None:
-            _check_complete_surface(this)
+            _check_complete_surface(this, incomplete=on_defects)
     return surf[0] if s_id is not None else surf
 
 
@@ -1554,6 +1556,32 @@ def write_bem_surfaces(fname, surfs, overwrite=False, verbose=None):
             _write_bem_surfaces_block(fid, surfs)
             end_block(fid, FIFF.FIFFB_BEM)
             end_file(fid)
+
+
+@verbose
+def write_head_bem(fname, rr, tris, on_defects='raise', overwrite=False,
+                   verbose=None):
+    """Write a head surface to a fiff file.
+
+    Parameters
+    ----------
+    fname : str
+        Filename to write.
+    rr : array, shape=(n_vertices, 3)
+        Coordinate points.
+    tris : ndarray of int, shape (n_tris, 3)
+        Triangulation (each line contains indices for three points which
+        together form a face).
+    %(on_defects)s
+    overwrite : bool
+        If True (default False), overwrite the file.
+    %(verbose)s
+    """
+    fname = _check_fname(fname, overwrite=overwrite, name='fname')
+    surf = _surfaces_to_bem([dict(rr=rr, tris=tris)],
+                            [FIFF.FIFFV_BEM_SURF_ID_HEAD], [1], rescale=False,
+                            incomplete=on_defects)
+    write_bem_surfaces(fname, surf, overwrite=True)
 
 
 def _write_bem_surfaces_block(fid, surfs):
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1246,6 +1246,8 @@ def read_bem_surfaces(fname, patch_stats=False, s_id=None, on_defects='raise',
         An error will be raised if it doesn't exist. If None, all
         surfaces are read and returned.
     %(on_defects)s
+
+        .. versionadded:: 0.23
     %(verbose)s
 
     Returns
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1579,11 +1579,10 @@ def write_head_bem(fname, rr, tris, on_defects='raise', overwrite=False,
         If True (default False), overwrite the file.
     %(verbose)s
     """
-    fname = _check_fname(fname, overwrite=overwrite, name='fname')
     surf = _surfaces_to_bem([dict(rr=rr, tris=tris)],
                             [FIFF.FIFFV_BEM_SURF_ID_HEAD], [1], rescale=False,
                             incomplete=on_defects)
-    write_bem_surfaces(fname, surf, overwrite=True)
+    write_bem_surfaces(fname, surf, overwrite=overwrite)
 
 
 def _write_bem_surfaces_block(fid, surfs):
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -1569,8 +1569,8 @@ def write_head_bem(fname, rr, tris, on_defects='raise', overwrite=False,
     ----------
     fname : str
         Filename to write.
-    rr : array, shape=(n_vertices, 3)
-        Coordinate points.
+    rr : array, shape (n_vertices, 3)
+        Coordinate points in the MRI coordinate system.
     tris : ndarray of int, shape (n_tris, 3)
         Triangulation (each line contains indices for three points which
         together form a face).
--- a/tutorials/source-modeling/plot_fix_bem_in_blender.py
+++ b/tutorials/source-modeling/plot_fix_bem_in_blender.py
@@ -162,10 +162,8 @@ coords, faces = mne.read_surface(op.join(bem_dir, 'outer_skin.surf'))
 # Make sure we are in the correct directory
 if 'flash' in bem_dir:
     bem_dir = op.join(subjects_dir, 'sample', 'bem')
-# Backup the original head file
-shutil.copy(op.join(bem_dir, 'sample-head.fif'),
-            op.join(bem_dir, 'sample-head_orig.fif'))
 
+# Remember to backup the original head file in advance!
 # Overwrite the original head file
 mne.write_head_bem(op.join(bem_dir, 'sample-head.fif'),
                    coords, faces, overwrite=True)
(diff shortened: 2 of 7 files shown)


larsoner:
Let's add the defect-fixing code in another PR. I have a compact version floating around I can make a quick PR with.

The one thing I would add to this PR is see if you can make use of this `write_head_bem` here:

https://github.com/mne-tools/mne-python/blob/main/mne/commands/mne_make_scalp_surfaces.py#L130-L134

That plus a `latest.inc` update and I'd say +1 for merge

agramfort:
coming late to the party but why adding a public function here? Maybe we can keep this private and just use it for the end goal of topological fixing?

larsoner:
If people have a `-head.fif` that is broken they need some public function to read it (we have this already), then they can fix it, but then they need some public function to write it. You might think that they could keep the original surface but they can't since other fields will be inconsistent and need to be updated. This happens automagically for them in the new function because it eventually calls `complete_surface_info` under the hood.

Also it allows people to use their own head triangulations with just `rr, tris` if they have them from some other source, which I think is nice.

larsoner:
... for example if someone wants to use Blender to fix their head mesh, they'll need to use this function afterward.

agramfort:
adding some public functions without documentation, some example is for me
bad practice. I am not saying no I am saying it's not done here in the
right order.

>

larsoner:
Agreed -- we don't have a good dataset to *use* it in an example, but @yh-luo you could mention it in this example:

https://github.com/mne-tools/mne-python/blob/6ae3b22033c745cce5cd5de9b92da54c13c36484/tutorials/source-modeling/plot_fix_bem_in_blender.py

And probably add a new entry to the FAQ to mention that you can load, edit (e.g., in blender), then save using this new function, perhaps a new entry or added to this section:

https://github.com/mne-tools/mne-python/blob/baaf9e33f2b54f6a40454a8aa146577bde1d6f58/doc/overview/faq.rst#my-watershed-bem-meshes-look-incorrect

yh-luo:
Yes I'm planning on an example to use these functions, thus WIP.
> [ ] A tutorial/example to illustrate how to fix the topological defects


> Also it allows people to use their own head triangulations with just rr, tris if they have them from some other source, which I think is nice.

That's what I thought too. I think these functions can also be used to fix `inner_skull.surf`, etc. I used Blender to fix these surfaces, it takes time to do so. If mne-python can help with that process that would be nice.

I'm working on this but holidays in my country are coming. I plan to finish this in in mid- to end-February 🙂

larsoner:
> I think these functions can also be used to fix inner_skull.surf, etc.

Sounds like you are talking about "topology fixing functions", which I think we should save for a second PR. It's really an orthogonal issue and it's going to be more annoying with multiple new dependencies, etc. And I have more compact versions of the topology-fixing code sitting around.

If we follow my PR plan, this PR will just be for `write_bem_head` and the topology-check keyword argument. So adding just to the FAQ and a note to the Blender tutorial I think would be enough here.

yh-luo:
Then todos for me:

- `write_head_bem`  for `mne make_scalp_surfaces`
- update [Editing BEM surfaces in Blender](https://mne.tools/stable/auto_tutorials/source-modeling/plot_fix_bem_in_blender.html?highlight=blender)
- FAQ [My watershed BEM meshes look incorrect](https://mne.tools/stable/overview/faq.html#my-watershed-bem-meshes-look-incorrect)
- `latest.inc`

larsoner:
Sounds good to me!

yh-luo:
**Updates**
- Better leave `mne make_scalp_surfaces` that way. When converting the FreeSurfer surface file `lh.seghead` to `-head.fif`, the surface need to be rescaled. Then, it will be decimated to create medium and sparse head, which do not and should not need to be rescaled. The parameter `rescale` need to be passed in `write_head_bem` in order to be used in `mne make_scalp_surfaces`. I don't think that's a good idea to further confuse the end-user if they will only interact with the computed and correctly-scaled `-head.fif`. Also `msg` will be passed in `_surfaces_to_bem` to hint the usage of `--force`. I can chain the warning but I don't think that's worth the effort.

<details>
<summary>Related codes in `mne make_scalp_surfaces`</summary>

L112:
```python
    surf = mne.bem._surfaces_to_bem(
        [surf], [mne.io.constants.FIFF.FIFFV_BEM_SURF_ID_HEAD], [1],
        incomplete=incomplete, extra=msg)[0]
```

L330:
```python
        dec_surf = mne.bem._surfaces_to_bem(
            [dict(rr=points, tris=tris)],
            [mne.io.constants.FIFF.FIFFV_BEM_SURF_ID_HEAD], [1], rescale=False,
            incomplete=incomplete, extra=msg)
```
</details>

- FAQ [My watershed BEM meshes look incorrect](https://mne.tools/stable/overview/faq.html#my-watershed-bem-meshes-look-incorrect) is already very detailed. I think the updated tutorial will be a better place to illustrate the usage of `mne.write_head_bem`. Current version: https://25045-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_fix_bem_in_blender.html#sphx-glr-auto-tutorials-source-modeling-plot-fix-bem-in-blender-py
- I also noticed that the `.surf` files are moved into `bem/flash/` (previously in `bem/`), which broke the tutorial using the current version of `sample` dataset. The change occurred in #4585. This PR will fix that too.

tutorials/source-modeling/plot_fix_bem_in_blender.py
```diff
@@ -129,9 +138,68 @@
...
+
+# We use the same surface to simulate the fixed version here
+# Read in the .obj file
+coords, faces = mne.read_surface(
+    op.join(conv_dir, 'sample-head.obj'))
+
+# Remember to backup the original head file in advance!
+# Overwrite the original head file
+mne.write_head_bem(op.join(bem_dir, 'sample-head.fif'),
+                   coords, faces, overwrite=True)
```
larsoner, line 201: I would not actually overwrite these files, I'd leave these in a comment like
```suggestion
# This could be accomplished for example with::
#
#     >>> mne.write_head_bem(op.join(bem_dir, 'sample-head.fif'),
#                            coords, faces, overwrite=True)
```

tutorials/source-modeling/plot_fix_bem_in_blender.py
```diff
@@ -129,9 +138,68 @@
...
+
+# For consistency only
+coords = surf['rr']
+faces = surf['tris']
+
+# Write the head as an .obj file for editing
+mne.write_surface(op.join(conv_dir, 'sample-head.obj'),
+                  coords, faces, overwrite=True)
+
+# We use the same surface to simulate the fixed version here
```
larsoner, line 193:
```suggestion
# Usually here you would go and edit your meshes.
#
# Here we just use the same surface as if it were fixed
```

tutorials/source-modeling/plot_fix_bem_in_blender.py
```diff
@@ -129,9 +138,68 @@
...
+#
+# For EEG forward modeling, it is possible that ``outer_skin.surf`` would be
+# manually edited. In that case, remember to save the fixed version of
+# ``-head.fif`` from the edited surface file for coregistration.
+
+# Load the fixed surface
+coords, faces = mne.read_surface(op.join(bem_dir, 'outer_skin.surf'))
+
+# Make sure we are in the correct directory
+if 'flash' in bem_dir:
```
larsoner, line 163: I wouldn't use this conditional, people can be expected to have up-to-date datasets. So just use the correct path for the current version of the dataset

mne/bem.py
```diff
@@ -1556,6 +1560,32 @@ def write_bem_surfaces(fname, surfs, overwrite=False, verbose=None):
...
+    %(on_defects)s
+    overwrite : bool
+        If True (default False), overwrite the file.
+    %(verbose)s
+    """
+    fname = _check_fname(fname, overwrite=overwrite, name='fname')
+    surf = _surfaces_to_bem([dict(rr=rr, tris=tris)],
+                            [FIFF.FIFFV_BEM_SURF_ID_HEAD], [1], rescale=False,
+                            incomplete=on_defects)
+    write_bem_surfaces(fname, surf, overwrite=True)
```
larsoner, line 1586:
```suggestion
    write_bem_surfaces(fname, surf, overwrite=overwrite)
```

tutorials/source-modeling/plot_fix_bem_in_blender.py
```diff
@@ -129,9 +138,68 @@
             op.join(bem_dir, 'inner_skull_orig.surf'))
 
 # Overwrite the original surface with the fixed version
+# In real study you should provide the correct metadata using ``volume_info=``
 mne.write_surface(op.join(bem_dir, 'inner_skull.surf'), coords, faces,
                   overwrite=True)
```
larsoner, line 143: Can you comment out this overwriting as well?

tutorials/source-modeling/plot_fix_bem_in_blender.py
```diff
@@ -129,9 +138,68 @@
...
+coords, faces = mne.read_surface(op.join(bem_dir, 'outer_skin.surf'))
+
+# Make sure we are in the correct directory
+if 'flash' in bem_dir:
+    bem_dir = op.join(subjects_dir, 'sample', 'bem')
+
+# Remember to backup the original head file in advance!
+# Overwrite the original head file
+mne.write_head_bem(op.join(bem_dir, 'sample-head.fif'),
+                   coords, faces, overwrite=True)
```
larsoner, line 169: Let's comment out this one, too

tutorials/source-modeling/plot_fix_bem_in_blender.py
yh-luo: `write_surface` and `overwrite` was used from the top to the bottom of the tutorial. I understand the danger of overwriting the files, should I modify the previous codes too? I will push a commit that way. If not I can just revert it.

larsoner (approved):
LGTM, @agramfort can you look to see if the use case makes more sense with the tutorial?

mne/bem.py
```diff
@@ -1556,6 +1560,31 @@ def write_bem_surfaces(fname, surfs, overwrite=False, verbose=None):
...
+@verbose
+def write_head_bem(fname, rr, tris, on_defects='raise', overwrite=False,
+                   verbose=None):
+    """Write a head surface to a fiff file.
+
+    Parameters
+    ----------
+    fname : str
+        Filename to write.
+    rr : array, shape=(n_vertices, 3)
```
agramfort, line 1572:
```suggestion
    rr : array, shape (n_vertices, 3)
```

mne/bem.py
```diff
@@ -1556,6 +1560,31 @@ def write_bem_surfaces(fname, surfs, overwrite=False, verbose=None):
...
+        Coordinate points.
```
agramfort, line 1573:
```suggestion
        Coordinate points in the head coordinate system.
```

agramfort (commented):
besides these nitpicks LGTM

thx @yh-luo

mne/bem.py
larsoner: In this case I think it should be the MRI coordinate system (part of the `bem/*` files)

yh-luo: Yes, it's the MRI coordinate system.

agramfort:
thx @yh-luo

Issue #6801, opened 2019-09-21
can't raw.filter MISC channels


ossadtchi:
Hello!

With raw.filter I can not filter the instance of the raw created with pick_channels using names of MISC channels. It complains about not finding the channels to work with. Is it intended and one is supposed to filter only EEG or MEG channels? I would love to have a "parallel" code handling the two types of channels!

works:
```
raw_grad = raw.copy().pick_channels(names_grad);
raw_grad.filter(1, 45., h_trans_bandwidth='auto', filter_length ='auto', phase = 'zero');
```
does not work:
```
raw_msc = raw.copy().pick_channels(names_msc);
raw_msc.filter(0.5, 5., h_trans_bandwidth='auto', filter_length ='auto', phase = 'zero');
```
THANKS!!

agramfort:
`raw.filter` filters only data channels (EEG, MEG). You need to pass picks parameter manually to the  filter method. Please reopen if it does not work.

ossadtchi:
Thanks, Alex, but if I understood you correctly 
neither 
```
raw_misc.filter(0.5, 5., h_trans_bandwidth='auto', filter_length='auto',
                            phase = 'zero', picks = names_misc);
```
nor
```
raw_misc.filter(0.5, 5., h_trans_bandwidth='auto', filter_length='auto',
                             phase = 'zero', picks = np.array([0,1,2])); ```

appear to work and still give the same error. :(

**Comment by agramfort** [comment]
can you share a minimal code snippet to replicate the issue using mne sample data?

**Comment by ossadtchi** [comment]
Thanks! Yes, here it is. Should be OK with any raw fif file. 
```
tmin, tmax = 0, 50  

raw = mne.io.read_raw_fif(raw_fname) #, add_eeg_ref=False)
raw.crop(tmin, tmax).load_data()

# make a list of names for the grads and misc
picks_grad = mne.pick_types(raw.info, stim=False, meg = 'grad');
names_grad  = [raw.info['ch_names'][n] for n in picks_grad];
picks_misc = mne.pick_types(raw.info, stim=False, meg = False, misc = True);
names_misc  = [raw.info['ch_names'][n] for n in picks_misc];
raw_grad = raw.copy().pick_channels(names_grad);
raw_misc = raw.copy().pick_channels(names_misc);

# works:
raw_grad.filter(1, 45., h_trans_bandwidth='auto', filter_length='auto',
                            phase = 'zero');
# fails:
[...]

PR #8844, opened 2021-02-07
MRG, FIX: resampling MISC channels using raw.resample

enricovara:
Fixes an issue related to #6801 in `raw.resample`.

### Issue Description
Resampling MISC channels using `raw.resample` (mne.io.Raw.resample) by passing `stim_picks='misc'`, `stim_picks=None` or `stim_picks=stim_ch_name` was not possible.

### Examples
Using:
`raw_data.resample(Fs, npad='auto', window='boxcar', stim_picks='misc', n_jobs=1, events=None, pad='reflect_limited')`
yields an error:
> File "", line 24, in resample
> File "{envpath}/lib/python3.8/site-packages/mne/io/base.py", line 1166, in resample
> if len(stim_picks) > 0:
> TypeError: len() of unsized object

Using `stim_picks=ch_names` where ch_names is a list containing the misc channel names doesn't work either:
`raw_data.resample(Fs, npad='auto', window='boxcar', stim_picks=ch_names, n_jobs=1, events=None, pad='reflect_limited')`
yields:
> File "", line 24, in resample
> File "{envpath}/lib/python3.8/site-packages/mne/io/base.py", line 1168, in resample
> data_chunk[stim_picks], n_new, data_chunk.shape[1])
> IndexError: arrays used as indices must be of integer (or boolean) type

Using `stim_picks=None` called
`stim_picks = pick_types(self.info, meg=False, ref_meg=False, stim=True, exclude=[])`
behind the scenes, excluding MISC channels.

An unfriendly which required  importing `pick_types` from `mne.io.pick` was to pass `stim_picks=pick_types(raw_data.info, misc=True, exclude=[])`.

### Solution
This PR updates `mne/io/base.py` to allow flexibility for setups which record stimuli/auxiliary channels as `MISC` rather than `stim` without causing retro-compatibility issues. This also makes `raw.resample` behave consistently with other similar functions like `raw.filter`. 

### Additional information
If the `stim_picks` parameter is passed with anything but `None` it is now updated by
`stim_picks = _picks_to_idx(self.info, stim_picks, exclude=(), with_ref_meg=False)`
before it is updated by `stim_picks = np.asanyarray(stim_picks)` as before.

Another solution would have been to update `raw.resample` more aggressively by replacing the `stim_picks` parameter with a `picks` parameter that would function like the `picks` parameter in `raw.filter`. While more consistent, this would force existing users of `raw.resample` to review their code upon updating mne.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -66,7 +66,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix bug with `mne.io.Raw.resample` to allow passing ``stim_picks='misc'`` (:gh:`6801` **by new contributor** |Enrico Varano|_ and `Eric Larson`_)
+- Fix bug with `mne.io.Raw.resample` to allow passing ``stim_picks='misc'`` (:gh:`8844` **by new contributor** |Enrico Varano|_ and `Eric Larson`_)
 
 - Fix bugs with `mne.io.read_raw_persyst` where multiple ``Comments`` with the same name are allowed, and ``Comments`` with a "," character are now allowed (:gh:`8311` and :gh:`8806` **by new contributor** |Andres Rodriguez|_ and `Adam Li`_)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -20,6 +20,8 @@ Current (0.23.dev0)
 
 .. |Matt Sanderson| replace:: **Matt Sanderson**
 
+.. |Enrico Varano| replace:: **Enrico Varano**
+
 Enhancements
 ~~~~~~~~~~~~
 - Add dbs as new channel type for deep brain stimulation (DBS) recordings (:gh:`8739` **by new contributor** |Richard Koehler|_)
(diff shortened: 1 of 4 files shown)


enricovara:
@larsoner 
Ad requested in #6801

mne/io/base.py
```diff
@@ -1152,6 +1152,10 @@ def resample(self, sfreq, npad='auto', window='boxcar', stim_picks=None,
         if stim_picks is None:
             stim_picks = pick_types(self.info, meg=False, ref_meg=False,
                                     stim=True, exclude=[])
+        else:
+            stim_picks = _picks_to_idx(self.info, stim_picks, exclude=(),
+                                       with_ref_meg=False)
+
         stim_picks = np.asanyarray(stim_picks)
```
larsoner, line 1159: We don't need this line anymore because in either part of the conditional above it's going to be an `ndarray` of int afterward

larsoner (commented):
LGTM! All that seems to be missing is a test that would fail on `master` but pass on this PR. I would add something like this to `mne/io/fiff/tests/test_raw_fiff.py`:
```
def test_resample_stim():
    """Test stim_picks argument."""
    data = np.ones((2, 1000))
    info = create_info(2, 1000., ('eeg', 'misc'))
    raw = RawArray(data, info)
    raw.resample(500., stim_picks='misc')
```
Or whatever recreates your issue in just a few lines and is very fast to execute.

enricovara:
@larsoner Thank you very much for the guidance!

doc/changes/latest.inc
```diff
@@ -64,6 +66,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix bug with `mne.io.Raw.resample` to allow passing ``stim_picks='misc'`` (:gh:`6801` **by new contributor** |Enrico Varano|_ and `Eric Larson`_)
```
larsoner, line 69:
```suggestion
- Fix bug with `mne.io.Raw.resample` to allow passing ``stim_picks='misc'`` (:gh:`8844` **by new contributor** |Enrico Varano|_ and `Eric Larson`_)
```

larsoner (commented):
Just one last minor fix then I'll merge, thanks @enricovara !

PR #8846, opened 2021-02-08
MRG, BUG: Raw.rename_channels() should update Raw._orig_units too

hoechenberger:
In `main`, calling `rename_channels()` on a `Raw` instance would not
adjust `Raw._orig_units`.

While you could argue that the original units were only specified for
the original channel names, I would say that this makes the units
simply useless after channels have been renamed ;) Therefore I
consider the current behavior a bug.

This PR makes it so that if `Raw._orig_units` exists, its keys are
adjusted such that they stay in sync with the new channel names.

Discovered while working on data with @MaelysSolal

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -116,7 +116,7 @@ Bugs
 
 - When creating `~mne.Epochs`, we now ensure that ``reject_tmin`` and ``reject_tmax`` cannot fall outside of the epochs' time interval anymore (:gh:`8821` by `Richard Höchenberger`_)
 
-- :meth:`raw.rename_channels() <mne.Raw.rename_channels>` now also applies the new channel names to ``raw._orig_units``, if present (:gh:`8846` by `Richard Höchenberger`_)
+- :meth:`raw.rename_channels() <mne.io.Raw.rename_channels>` now also applies the new channel names to ``raw._orig_units``, if present (:gh:`8846` by `Richard Höchenberger`_)
 
 API changes
 ~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


agramfort (approved):

doc/changes/latest.inc
```diff
@@ -116,6 +116,8 @@ Bugs
 
 - When creating `~mne.Epochs`, we now ensure that ``reject_tmin`` and ``reject_tmax`` cannot fall outside of the epochs' time interval anymore (:gh:`8821` by `Richard Höchenberger`_)
 
+- :meth:`raw.rename_channels() <mne.io.Raw.rename_channels>` now also applies the new channel names to ``raw._orig_units``, if present (:gh:`8846` by `Richard Höchenberger`_)
```
larsoner, line 119: I would not reference private properties in changelogs, users shouldn't use or care about them, and mentioning them makes it seem like maybe they should or maybe could even make use of them somehow. Instead it's better to say that you've fixed a bug in accounting for the original units or whatever the downstream, public-API or behavior result is

mne/channels/channels.py
```diff
@@ -483,11 +483,33 @@ def rename_channels(self, mapping):
             .. versionchanged:: 0.20
                Return the instance.
 
+            .. versionchanged:: 0.23
+               For `~mne.io.Raw` instances, ``_orig_units`` will be adjusted
```
larsoner, line 487: Same here, don't document private attribute changes in public docstrings

larsoner:
Thanks @hoechenberger

PR #8852, opened 2021-02-10
FIX: paintEvent

GuillaumeFavelier:
Locally I was able to reproduce https://github.com/mne-tools/mne-python/pull/8757#issuecomment-774343270 and this small fix did the trick for me. Let's see how it is on Circle.

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -571,7 +571,6 @@ class Brain(object):
         'Left': Decrease camera azimuth angle
         'Right': Increase camera azimuth angle
         """
-        from ..backends._utils import _qt_disable_paint
         if self.time_viewer:
             return
         if not self._data:
@@ -679,10 +678,7 @@ class Brain(object):
             self._configure_status_bar()
 
             # show everything at the end
-            with _qt_disable_paint(self.plotter):
-                with self._ensure_minimum_sizes():
-                    self.show()
-            self._update()
+            self.show()
 
     @safe_event
     def _clean(self):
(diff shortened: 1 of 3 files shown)


GuillaumeFavelier:
It works in here too:

image

https://25098-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/stats-source-space/plot_stats_cluster_spatio_temporal.html#sphx-glr-auto-tutorials-stats-source-space-plot-stats-cluster-spatio-temporal-py

GuillaumeFavelier:
I think it's worth restarting the CIs, even though `processEvents()` is not called directly, I don't want to introduce regressions.

mne/viz/_brain/_brain.py
```diff
@@ -682,6 +682,7 @@ def setup_time_viewer(self, time_viewer=True, show_traces=True):
             with _qt_disable_paint(self.plotter):
                 with self._ensure_minimum_sizes():
                     self.show()
+            self._update()
```
larsoner, line 685: This will only happen if we call `setup_time_viewer`, which only happens if `time_viewer=True` (effectively) is used, right? This suggests that if `time_viewer=False` then the bug will still be there. It seems like we Probably need an equivalent change in very similar code in `_pyvista.py`, or some other fix, or maybe this belongs at the `Brain` level?

GuillaumeFavelier: You're totally right. And this code in Brain can be removed since `self.show()` calls `self._renderer.show()`. I'll update on `_pyvista`

GuillaumeFavelier: ... and cannot be removed since `ensure_minimum_size` has a different version for Brain.

GuillaumeFavelier:
This ready on my end.

larsoner:
Thanks @GuillaumeFavelier

PR #8856, opened 2021-02-10
"resp" channel is for respiration, not responses

### LINKED ISSUE #8853: Confusion around "resp" channel type
In `channels.py`, it is claimed that the `resp` channel type is for certain response triggers:

https://github.com/mne-tools/mne-python/blob/19fb479de44865d5ce79ad5af539a06081fae0c5/mne/channels/channels.py#L704-L706

This seems to be supported by code treating BTi data:
https://github.com/mne-tools/mne-python/blob/19fb479de44865d5ce79ad5af539a06081fae0c5/mne/io/bti/bti.py#L1188-L1189

But the following comment in `constants.py` suggests it's a resp-iratory, not a resp-onse channel:
https://github.com/mne-tools/mne-python/blob/19fb479de44865d5ce79ad5af539a06081fae0c5/mne/io/constants.py#L177

Now … what gives?

x-ref https://github.com/mne-tools/mne-bids/issues/700

cc @agramfort

hoechenberger:
Fixes #8853

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -120,7 +120,7 @@ Bugs
 
 - When creating `~mne.Epochs`, we now ensure that ``reject_tmin`` and ``reject_tmax`` cannot fall outside of the epochs' time interval anymore (:gh:`8821` by `Richard Höchenberger`_)
 
-- :func:`read_raw_bti() <mne.io.read_raw_bti>` erroneously treated response channels as respiratory channels (:gh:`8855` by `Richard Höchenberger`_)
+- `~mne.io.read_raw_bti` erroneously treated response channels as respiratory channels (:gh:`8855` by `Richard Höchenberger`_)
 
 API changes
 ~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


doc/changes/latest.inc
```diff
@@ -120,6 +120,8 @@ Bugs
 
 - When creating `~mne.Epochs`, we now ensure that ``reject_tmin`` and ``reject_tmax`` cannot fall outside of the epochs' time interval anymore (:gh:`8821` by `Richard Höchenberger`_)
 
+- :func:`read_raw_bti() <mne.io.read_raw_bti>` erroneously treated response channels as respiratory channels (:gh:`8855` by `Richard Höchenberger`_)
```
larsoner, line 123: FYI if you can live with it being shortened to `read_raw_bti` (no parens) then you can just do
```suggestion
- `~mne.io.read_raw_bti` erroneously treated response channels as respiratory channels (:gh:`8855` by `Richard Höchenberger`_)
```
And the `:func:` is optional because of how we set our default role in `doc/conf.py`

larsoner (approved):
LGTM +1 for merge

doc/changes/latest.inc
hoechenberger: Right, I KNEW that once, but have forgotten again :D thanks for the reminder!

drammock: Adding the parens automatically is a sphinx configuration variable FWIW.  So if people are using the more convoluted syntax just because they want parens, we could enable auto-parens-addition globally.

PR #8857, opened 2021-02-10
MRG, MAINT: Clean up BEM surface code

larsoner:
In preparation for writing some BEM quality code, I wanted to have an `_ensure_bem_surfaces` function. This should simplify our handling of BEM surfaces, and in the process simplified `plot_alignment` by populating the `'surfs'` entry of the spherical ConductorModel dict. This ended up being a bugfix where on `main` you could do `plot_alignment(..., bem=sphere, surfaces='brain')` and have no surfaces show up if you had a 1-layer sphere model. This now raises an error because it seems like a user mistake. In the case where someone wants a FreeSurfer brain surface they can do `surfaces='pial'` explicitly if they want, but I assume this use case is rare.

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -8,6 +8,7 @@
 # The computations in this code were primarily derived from Matti Hämäläinen's
 # C code.
 
+from collections import OrderedDict
 from functools import partial
 import glob
 import os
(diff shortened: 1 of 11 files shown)


mne/bem.py
```diff
@@ -457,19 +448,17 @@ def _check_surfaces(surfs, incomplete='raise'):
...
     """Check that the coordinate limits are reasonable."""
     sizes = surf['rr'].max(axis=0) - surf['rr'].min(axis=0)
     if (sizes < 0.05).any():
-        raise RuntimeError('Dimensions of the surface %s seem too small '
-                           '(%9.5f mm). Maybe the the unit of measure is '
-                           'meters instead of mm' %
-                           (_surf_name[surf['id']], 1000 * sizes.min()))
+        raise RuntimeError(
+            f'Dimensions of the surface {_bem_surf_name[surf["id"]]} seem too '
+            f'small ({1000 * sizes.min():9.5f}). Maybe the the unit of measure'
```
agramfort, line 460:
```suggestion
            f'small ({1000 * sizes.min():9.5f}). Maybe the unit of measure'
```

agramfort (approved):
nitpick then +1 for MRG

PR #8858, opened 2021-02-11
MRG: Change resp channel unit to Volts

hoechenberger:
Respiratory channels contain an analog signal, monitoring the expansion of a chest strap / belt.
The signal is measured in Volts.

Scaling factors (µV, mV, V) depend on the specific amplifier system used, so we cannot provide a
sophisticated default scaling (and leave it at 1).

cc @MaelysSolal, @agramfort

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -72,6 +72,8 @@ Enhancements
 
 - Add :func:`mne.write_head_bem` to support writing head surface files (:gh:`8841` by `Yu-Han Luo`_)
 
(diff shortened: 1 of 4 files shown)


agramfort (approved):
nice !

hoechenberger:
Updated tutorial rendering seems to look good:

https://25124-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/sample-datasets/plot_sleep.html?highlight=plot_sleep#read-the-psg-data-and-hypnograms-to-create-a-raw-object

PR #8839, opened 2021-02-04
MRG: Do not overwrite times_in in spectral_connectivity

apadee:
#### Reference issue
Fixes #8824.


#### What does this implement/fix?
The `mne.spectral.spectral_connectivity` method now correctly reads and uses the time axis when given a `mne.Epochs` data object.  
It used to rewrite the time axis from the data with a generic one. This caused the time interval given as `tmin` and `tmax` arguments to not be aligned with the data.  
A warning has been added if either `tmin` or `tmax` bound falls outside of the scope of the data.  
A test to check for this issue has been added. 

#### Additional information
Prior to the fix, this used to raise a `ValueError` as shown below:

```
(mnedev) ap@cookie:/opt/mne-python/mne/tests$ pytest test_spectral_connectivity.py 
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /opt/mne-python, configfile: setup.cfg
plugins: cov-2.11.1, timeout-1.4.2, harvest-1.10.1
collected 1 item                                                                                                                                                                                                                             

test_spectral_connectivity.py F                                                                                                                                                                                                        [100%]


========================================================================================================== slowest 20 test modules ===========================================================================================================
0.40s total    mne/tests/test_spectral_connectivity.py


================================================================================================================== FAILURES ==================================================================================================================
_________________________________________________________________________________________________________ test_spectral_connectivity _________________________________________________________________________________________________________
test_spectral_connectivity.py:40: in test_spectral_connectivity
    ret = spectral_connectivity(epochs, method='pli', mode='multitaper',
<decorator-gen-446>:24: in spectral_connectivity
    ???
../connectivity/spectral.py:785: in spectral_connectivity
    n_signals, indices_use) = _prepare_connectivity(
../connectivity/spectral.py:948: in _prepare_connectivity
    mask = _time_mask(times_in, tmin, tmax, sfreq=sfreq)
../utils/numerics.py:493: in _time_mask
    raise ValueError('No samples remain when using tmin=%s and tmax=%s %s'
E   ValueError: No samples remain when using tmin=-0.9 and tmax=-0.2 (original time bounds are [0.0, 2.001282051803185])
```

PR #8861, opened 2021-02-11
MRG, BUG: Fix handling of times

larsoner:
Follow-up to #8839 to fix CircleCI failures

- We need to be careful about inferring times in `spectral_connectivity` because data can for example arrive in `(ndarray, SourceEstimate)` pairs and in that case we should get the times from the SourceEstimate, not the ndarray.
- if there is a mismatch, we really only need to warn about it once, not on every epoch.
- if there is a mismatch, print at least some of the arrays so that users can see the mismatch.
- when using `mne.use_log_level('warn', add_frames=3):` actually have the frames printed so you can get an output that allows you to track the lines like:
```
/home/larsoner/python/mne-python/examples/connectivity/plot_mne_inverse_psi_visual.py:95: RuntimeWarning: time scales of input time series do not match
  psi, freqs, times, n_epochs, _ = phase_slope_index(
┌connectivity.spectral:794
├connectivity.spectral:952
├connectivity.spectral:499
└utils._logging:360 :           time scales of input time series do not match
/home/larsoner/python/mne-python/examples/connectivity/plot_mne_inverse_psi_visual.py:95: RuntimeWarning: time scales of input time series do not match
  psi, freqs, times, n_epochs, _ = phase_slope_index(
┌connectivity.effective:115
├connectivity.spectral:835
├connectivity.spectral:499
└utils._logging:360 :           time scales of input time series do not match
```
- no need for latest.inc update as it's a within-cycle bugfix

### DIFF
--- a/examples/connectivity/plot_mne_inverse_psi_visual.py
+++ b/examples/connectivity/plot_mne_inverse_psi_visual.py
@@ -91,7 +91,7 @@ fmax = 20.
 tmin_con = 0.
 sfreq = epochs.info['sfreq']  # the sampling frequency
 
-psi, freqs, times, n_epochs, _ = phase_slope_index(
+psi, _, _, _, _ = phase_slope_index(
     comb_ts, mode='multitaper', indices=indices, sfreq=sfreq,
     fmin=fmin, fmax=fmax, tmin=tmin_con)
 
--- a/mne/connectivity/spectral.py
+++ b/mne/connectivity/spectral.py
@@ -476,11 +476,14 @@ def _check_method(method):
 
 
 def _get_and_verify_data_sizes(data, sfreq, n_signals=None, n_times=None,
-                               times=None):
+                               times=None, warn_times=True):
     """Get and/or verify the data sizes and time scales."""
     if not isinstance(data, (list, tuple)):
         raise ValueError('data has to be a list or tuple')
     n_signals_tot = 0
+    # Sometimes data can be (ndarray, SourceEstimate) groups so in the case
+    # where ndarray comes first, don't use it for times
+    times_inferred = False
     for this_data in data:
         this_n_signals, this_n_times = this_data.shape
         if n_times is not None:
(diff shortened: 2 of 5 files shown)


larsoner:
Merging to make CIs happy but if people have comments I can make a follow-up PR

examples/connectivity/plot_mne_inverse_psi_visual.py
```diff
@@ -91,7 +91,7 @@
 tmin_con = 0.
 sfreq = epochs.info['sfreq']  # the sampling frequency
 
-psi, freqs, times, n_epochs, _ = phase_slope_index(
+psi, _, _, _, _ = phase_slope_index(
```
drammock, line 94: no need to go back and change, but this would have been a great place for `*_`:
```suggestion
psi, *_ = phase_slope_index(
```

mne/connectivity/spectral.py
```diff
@@ -829,8 +837,9 @@ def spectral_connectivity(data, method='coh', indices=None, sfreq=2 * np.pi,
 
         # check dimensions and time scale
         for this_epoch in epoch_block:
-            _get_and_verify_data_sizes(
-                this_epoch, sfreq, n_signals, n_times_in, times_in)
+            _, _, _, warn_times = _get_and_verify_data_sizes(
```
drammock, line 840: here too:
```suggestion
            *_, warn_times = _get_and_verify_data_sizes(
```

drammock (commented):
thanks for fixing this.  changes look fine, just two unimportant style comments

examples/connectivity/plot_mne_inverse_psi_visual.py
larsoner: This is @agramfort's preferred way of doing it as if the api changes it forces a change in your code. I would prefer actually `psi = phase_slope_index(...)[0]` in most cases if I didn't agree with the idea of making sure the API is stable

PR #8862, opened 2021-02-11
MNT: Migrate VTK Widgets

GuillaumeFavelier:
This PR moves the VTK widgets into a dedicated Qt interface for the standalone solution and into Jupyter Widgets for notebooks.

Here is the latest sketch (*it may change during development*):

image

For the approach I chose, I'll just build the dock and move one widget at a time. So both old and new ways will live in parallel for a while. I expect working on refactoring and coverage at end. For now, it's a work in progress.

It's an item of https://github.com/mne-tools/mne-python/issues/7162


GuillaumeFavelier:
The previous interface relies on sliders heavily but I forgot that `QSlider`s use integer by default and I don't seem to make the `pyvistaqt.DoubleSlider` work.

larsoner:
Don't use sliders in Qt, use QDoubleSpinBox es

larsoner:
(for time you probably do want a silder -- I would actually oversample the `.times` by a factor of 10 or so to give it a float-like behavior)

GuillaumeFavelier:
That's the easy way out of this for sure :grin:

GuillaumeFavelier:
In the previous interface, each renderer had an orientation widget. On the sketch, there is only one widget for that. The simple solution is to duplicate those. A complicated one would be to command the click on a renderer to update the widget.

larsoner:
How about a SpinBox representing the raveled plotter number to control? Then at a later date we can connect some on_click action to update this SpinBox with whichever one was clicked, then you have interactive mouse-based control and direct indexing control.

GuillaumeFavelier:
> How about a SpinBox representing the raveled plotter number to control?

This works too!

GuillaumeFavelier:
For now, this is how it looks:

image

The time controller is not a slider and the orientation widget is not migrated yet. Also, I find the clim/scale controller hard to use.

GuillaumeFavelier:
Now everything is migrated. I must admit, it is a cleaner 3D viewer:

image

The time widget is still not a slider tough.

GuillaumeFavelier:
Once the case of the time widget is closed, I think it will be a good time to double-check the UI/UX and fix the (hopefully) minor details. Then I'll move on to test/coverage. I'm wondering how `LinkViewer` will behave with this interface :thinking: 

And finally same treatment for the `notebook` backend.

GuillaumeFavelier:
I had modified the time callback back at the time because VTK sliders handle float values but `QSlider` support time index natively:

image

GuillaumeFavelier:
It's now ready for UI/UX reviews! Meanwhile I'll work on fixing the tests.

GuillaumeFavelier:
I struggle with `test_screenshot` now. I tried hiding the `self.dock_widget` and playing with `_ensure_minimum_size()` without success.

GuillaumeFavelier:
Apparently, the very convenient signal `textActivated` requires `Qt 5.14`:

image

GuillaumeFavelier:
I'll try `currentTextChanged` instead:

Reference:
https://doc.qt.io/qt-5/qcombobox.html#currentTextChanged

GuillaumeFavelier:
Next step is the notebook dock.

hoechenberger:
@GuillaumeFavelier I haven't run this locally, so my suggestions are only based on the latest screenshot you posted:

- All `clim` sliders should have individual labels
- `clim` should be replaced with a proper word / words :) "Color Limits" or something? Or simply "Limits"?
- The time slider is too narrow. Do you think you could, by any chance, place it either between the 3D and Matplotlib figure; or below the Matplotlib figure inside a rectangular widget that only appears upon hovering?
- In the toolbar should be some whitespace between the dropdown menus and their respective labels
- The "Help" button should probably be the rightmost button, or even be right-aligned? (not sure if either of those options would look great, but it might be worth a shot!)
- I would love to be able to increase the width of the traces in the Matplotlib figure; or maybe we should even change the defaults: I find them to narrow & shallow by default, which might be related to my not-so-perfect eyesight

larsoner:
I'm not too worried about the time slider width. You have a full-width matplotlib window you can click inside if you want more precision

hoechenberger:
> I'm not too worried about the time slider width. You have a full-width matplotlib window you can click inside if you want more precision

Good point, but this is not a Qt widget and therefore not properly accessible, or is it? it would be great being able to operate the UI in a, say, standardized way. Can we maybe capture left and right arrow keypresses and map them to time slider movements?

Also, I think your observation would actually justify removing the narrow time slider (let's just call it a stub!), which I would be in full support of.

larsoner:
> Good point, but this is not a Qt widget and therefore not properly accessible, or is it?

The time slider should have full accessibility. Would having it wider help with accessibility? I'm not familiar with the requirements there or what would help.

> Can we maybe capture left and right arrow keypresses and map them to time slider movements?

The time slider above should already have left/right arrow support once you've brought it to focus / tabbed to it. But we could also implement this in the render and matplotlib windows.

hoechenberger:
> The time slider above should already have left/right arrow support once you've brought it to focus / tabbed to it.

I actually didn't think of this. This is neat. In this case, I'd like to retract my proposal to change / move the time slider for now. :)

agramfort:
- autoscaling does not seem to work anymore
- toggling interactions makes the window move
- toggling interactions off an on resets the window layout I manually adjust
- maybe the renderer widget could be hidden if there is only one?

GuillaumeFavelier:
With da37eef, the `notebook` interface of `Brain` has a new face:

image

As you can see, some UI tweaks are still necessary. Plus, the big majority of the callbacks needs to be updated to be compatible.

GuillaumeFavelier:
Thank you all for trying, I know it's not yet stable so I'll do my best to consider your comments and improve the overall experience :+1:

GuillaumeFavelier:
After da37eef, the dock has changed too, does it help with some of your comments @hoechenberger ?

image

hoechenberger:
> With [da37eef](https://github.com/mne-tools/mne-python/commit/da37eef806ed2b26e5928cbe2cd52728c750db0d), the `notebook` interface of `Brain` has a new face:

Amazing work, @GuillaumeFavelier!



> After [da37eef](https://github.com/mne-tools/mne-python/commit/da37eef806ed2b26e5928cbe2cd52728c750db0d), the dock has changed too, does it help with some of your comments @hoechenberger ?

Yes! Although I'm wondering … what does the `f` in `f{min,mid,max,scale}` mean, anyway?
They all pertain to the colormap range / scale, right? `fmin, fmax` are already used elsewhere consistently in MNE to refer to minimum and maximum frequencies, respectively…

Would it make sense to have something like this instead:

Screen Shot 2021-02-22 at 12 18 53

GuillaumeFavelier:
> Would it make sense to have something like this instead:

This is a fresh new take on this **clim**, I like it :smile:

hoechenberger:
> This is a fresh new take on this **clim**, I like it 😄

😅🥳

GuillaumeFavelier:
Using `QGroupBox` I got the following @hoechenberger :

image

The name has not changed yet.

larsoner:
Looks great!

I would call it `colormap` or `clim` instead of `colors`, though, it matches better with our usual terminology and options

hoechenberger:
Very nice indeed! 😍 Is there maybe even a collapsible group box widget?

hoechenberger:
> Looks great!
> 
> 
> 
> I would call it `colormap` or `clim` instead of `colors`, though, it matches better with our usual terminology and options

Yes although I'd make it two words :) Color Map? But I'll leave this decision to them native speakers hrhr

larsoner:
It's always one word in my experience, see also matplotlib's usage

https://matplotlib.org/stable/tutorials/colors/colormaps.html

GuillaumeFavelier:
Result after 01ba379:

image

GuillaumeFavelier:
So If I understood correctly you would also move the "Restore Scaling" button from the tool bar to call it "Reset" in the colormap group box right @hoechenberger ?

GuillaumeFavelier:
Also I think you're right with the help button of tool bar:

> The "Help" button should probably be the rightmost button, or even be right-aligned?

To achieve this in a cleaner way, I would like to move the annotation and extraction mode into a group in the dock as well.

GuillaumeFavelier:
Also by going all out with `QGroupBox`:

image

drammock:
> Also by going all out with `QGroupBox`

That looks great!

hoechenberger:
Great! Can the Renderer spin box become a Dropdown-Menü (haha autocorrect, how adorable 🥰), WDYT?

GuillaumeFavelier:
It's possible too :+1: . No strong opinion here.

GuillaumeFavelier:
After 9b77d4f:

image

GuillaumeFavelier:
I can finally move the help button :sweat_smile:

GuillaumeFavelier:
And after  ccebb51 @hoechenberger :

image

GuillaumeFavelier:
I still think the slider is the best type of widget for the `scale` controller of the Colormap group. The 'snap' feature does not really makes sense otherwise 🤔

hoechenberger:
could you share a MWE to reproduce the exact figures you're sharing here? :)

GuillaumeFavelier:
Whoops, my bad I forgot to share it:

```py
import os
import mne
initial_time = 0.13
                 hemi='split',
                 views=['lat', 'med'],
                 )
```

GuillaumeFavelier:
Update after 77670d4:

image

hoechenberger:
Thanks @GuillaumeFavelier!

- The playback speed spinbox uses an unsuitable step with (jumps straight to 1)
- The playback speed, renderer, and smoothing widgets are too wide, their content is "lost" inside. Centering the text could MAYBE help a little, but I think they're just way too wide.
- I agree that the spinboxes for the cmap don't really serve their purpose well, and sliders might do a better job here
- the time slider widget needs a label and at least during interaction it should display its current value
- widget interaction blocks until the 3D rendering has completed, leading to a seemingly unresponsive UI at times. if possible, 3D rendering should be aborted and restarted with the current values if UI widget interaction continues

All in all I'm very happy to see the progress you've been making here!! Amazing work!

hoechenberger:
> * the time slider widget needs a label and at least during interaction it should display its current value

I think you could also put a label below it, i.e.,

```
|----------------o---|
Current time: 0.123 s
```

hoechenberger:
- I cannot navigate through all widgets by pressing TAB. In fact, currently can only jump from one spinbox to another. Could you look into this is you have a quiet minute or two? :)

GuillaumeFavelier:
In terms of API, I added a parameter `compact` which is `True` by default that decides whether the label is added at top with text aligned or on the left. This is the result with `compact=False` for `playback_speed`, `renderer` and `smoothing`:

image

GuillaumeFavelier:
And after 6f866f5:

image

GuillaumeFavelier:
>  The playback speed spinbox uses an unsuitable step width (jumps straight to 1)

I changed the default to `0.01`.

> The playback speed, renderer, and smoothing widgets are too wide, their content is "lost" inside. Centering the text could MAYBE help a little, but I think they're just way too wide.

I introduced `compact` for that.

> I agree that the spinboxes for the cmap don't really serve their purpose well, and sliders might do a better job here

This one is tricky to solve at the moment see https://github.com/mne-tools/mne-python/pull/8862#issuecomment-777506656

> the time slider widget needs a label and at least during interaction it should display its current value

There is a `vtkActor` for that in the 3D scene, do you suggest to move it to the dock?

 > widget interaction blocks until the 3D rendering has completed, leading to a seemingly unresponsive UI at times. if possible, 3D rendering should be aborted and restarted with the current values if UI widget interaction continues

Which interaction in particular caused this? For smoothing for example, it's a CPU intensive task for high values :thinking: 

> I cannot navigate through all widgets by pressing TAB. In fact, currently I can only jump from one spinbox to another.

This is low priority, I'll see what I can do. Worst case scenario, I'll add it to https://github.com/mne-tools/mne-python/issues/7162 :+1:

hoechenberger:
> This one is tricky to solve at the moment see [#8862 (comment)](https://github.com/mne-tools/mne-python/pull/8862#issuecomment-777506656)
> 
> > the time slider widget needs a label and at least during interaction it should display its current value
> 
> There is a `vtkActor` for that in the 3D scene, do you suggest to move it to the dock?

No, I suggest you add an additional display of the time, nearby the slider. If I use a slider, I expect it to give me some kind of immediate feedback, or at least to have anchor labels, otherwise it's hard to know what I'm manipulating when using the slider

> > widget interaction blocks until the 3D rendering has completed, leading to a seemingly unresponsive UI at times. if possible, 3D rendering should be aborted and restarted with the current values if UI widget interaction continues
> 
> Which interaction in particular caused this? For smoothing for example, it's a CPU intensive task for high values 🤔

I think I had this with smoothing, yes

GuillaumeFavelier:
I realized after reading your comment again that I misunderstood:

> The playback speed, renderer, and smoothing widgets are too wide, their content is "lost" inside. Centering the text could MAYBE help a little, but I think they're just way too wide.

I think you refer to the content and not the label. So I tried the following:

image

I could not find a satisfying way to achieve the same for `QComboBox` unfortunately.

hoechenberger:
I actually thought of something like this :)

Screen Shot 2021-02-23 at 14 42 08

GuillaumeFavelier:
Interesting...

But where is the current time? What are `-1` and `0.5`?

hoechenberger:
> What are `-0.1` and `0.5`?

The anchors, i.e. `inst.tmin` and `inst.tmax`


> But where is the current time?

If it can be displayed as a tooltip during slider interaction, I'm happy… is that possible?

larsoner:
> > But where is the current time?

> If it can be displayed as a tooltip during slider interaction, I'm happy… is that possible?

I would make a horizontal layout with 5 columns, left limit (left justified), spring, current time (centered), spring, right limit (right justified).

hoechenberger:
> I would make a horizontal layout with 5 columns, left limit (left justified), spring, current time (centered), spring, right limit (right justified).

If that works space-wise, it could be a good pitch indeed :)

agramfort:
colormap relates to jet, viridis etc.
color limits or colorbar is more appropriate I feel.

the auto scaling seems to work a bit differently now. Is it me?
the scale parameter also seems slow to manipulate.

>

GuillaumeFavelier:
I don't know what spring is, @hoechenberger do you think you could sketch something :sweat_smile: ?

hoechenberger:
> I don't know what spring is, @hoechenberger do you think you could sketch something 😅 ?

Just a variable horizontal space that takes up all the available space :)

GuillaumeFavelier:
Thank you, I would call that stretch xD

> the auto scaling seems to work a bit differently now. Is it me?

Could be, I changed many thing in this PR

hoechenberger:
> color limits or colorbar is more appropriate I feel.

Good point, +1 for Color Limits or maybe Color Range

larsoner:
Regarding spin box -> slider, you can fake a continuous slider just like you do for time for example by having for example 10000 (really INT_MAX would work but Qt might not like it) int values and just dividing the result by 10000, now you have a float between 0 and 1 (and I doubt anyone will notice the discretization). I would make a new class `QFloatSlider` to do this since it's a useful class.

Once you do this, you could do the same layout as FreeView with label/slider/value for our min, mid, and max:

Screenshot from 2021-02-23 09-00-08

You could actually do the same thing with time.

> Good point, +1 for Color Limits or maybe Color Range

Someday we might want to add support for actually changing the colormap itself. I would call it "Color mapping" since it captures all current and potential future aspects.

hoechenberger:
> How about a SpinBox representing the raveled plotter number to control? T

@GuillaumeFavelier added a dropdown menu for this yesterday or so, which I believe makes **much** more sense here. A spinbox for renderer selection feels wrong to me :)

GuillaumeFavelier:
Update after 659579c7465ed6e4810948beaf6f324b29dfb68b:

image

GuillaumeFavelier:
Meanwhile on `notebook`:

image

hoechenberger:
Great progress, @GuillaumeFavelier! I appreciate your taking the time to kind of ingest, process, and produce something from all the sometimes contrary opinions and ideas shared here!

When it comes to UI design and accessibility, I usually take inspiration from how Apple does things in iOS and macOS; I recently attended a Clubhouse session where lots of blind folks joined and shared how they organize their daily lives, and the people "on the stage" univocally agreed that iOS was a total game changer. So, coming back to macOS, I just opened the accessibility settings and there they have sliders that look like this:

Screen Shot 2021-02-23 at 16 14 34

i.e. the label / descriptor appears left of the slider – which isn't possible in our UI due to space constraints, but I think that's alright. However what I believe **would** make sense is to move the labels you have now added to appear **below** the slider.

Further suggestions:

- the time slider precision (number of digits after the comma) should be equivalent as to what is displayed in the 3D figure
- I think we should have a label above the time slider, for example `Time Point (s)`, so
    - you know what the slider controls
    - you know the unit of the displayed values

I go crazy if I see axes with units missing, sorry about that, must be a result of the many engineering classes I took ;)

GuillaumeFavelier:
Update after 2a542c9f7938a5b3cfadd4be7fe4359be2a9b388:

image

larsoner:
Looks great!

Now I think the critical usability thing will be the float slider support I mentioned above as it's more or less the same between the macOS and FreeView implementations. However I prefer the FreeView one because it has a float entry box to the right which is nice because you can specify the actual value you want. I would implement this float-slider+entry for both the time slider and the fmin/fmid/fmax sliders. I think for both of these you can label above (Time, fmin/fmid/fmax) and have limits at the bottom. 

This to me is the mockup that is both most usable and compact (numbers on right are actually entry like in Freeview, and update when you do the sliders):

```
     Time (s)    
---------o------  1.1
-0.2         0.5

   min/mid/max
--o-------------  3.0
--------o-------  6.0
------------o---  9.0
1.0        15.0
```

GuillaumeFavelier:
Very clear mockup, many thanks @larsoner :wink: 

The float slider is next on my list :+1:

I noticed that you switched the labels `min/mid/max` to the top? This is different than the tuple label/slider/value you describe in https://github.com/mne-tools/mne-python/pull/8862#issuecomment-784224151

I'll go with that first and we can still update afterwards if needed

larsoner:
I think top is better for consistency with time and to give more horizontal room for the slider, that way there is more room to control and adjust it

hoechenberger:
> I think top is better for consistency with time and to give more horizontal room for the slider, that way there is more room to control and adjust it

Fair point, however I think it's better to have the label close to the respective slider it belongs to. Could a compromise be that we move the labels to the top as you suggested, but to the top **of each slider individually**? As we currently have it with the Time Slider?

larsoner:
I'm not sure what tangible usability benefit this actually brings -- they are always ordered and they can't actually go past each other so really we could just call it "limits" and it's not even very ambiguous. There is a tangible cost, namely the vertical space.

hoechenberger:
Let's see how it looks with your suggestion and then iterate :)

GuillaumeFavelier:
Update after 4f26de0, the tricky part is to maintain consistency between the backends:

PyVista | Notebook
--|--
image | image

larsoner:
>  the tricky part is to maintain consistency between the backends:

I don't think they have to be absolutely identical. For example if `ipywidgets` makes a scrollbar on the slider, this is bad. So let's make the slider full-width, and put the text entry box below it.

Qt looks great, though! One suggestion would be to make the sliders have stretch factor 3 and number boxes stretch factor 1 so that the sliders take up 3/4 of the space.

The "scale" should really probably just be a slider, with the right extreme being 2x, and the left extreme being 0.5x or so (and in between values being in between those two). Entering a text value is not so useful since it's really for gross adjustments. How about this:
```
==o==========   3.0
======o======   6.0
==========o==   9.0
---------------------
======o====== rescale
```
People will very quickly learn that "rescale" will scale all limits based on whether (and how far) they click to the left or right of the center point of "rescale". The long `----` is a horizontal divider, which I think would also help with this. Or maybe "all" is a better name than "rescale", not sure. Either way, a tooltip for the slider and/or text to the right stating what it does would help.

Thanks for sticking with the iterations, I think we're getting close!

GuillaumeFavelier:
> I don't think they have to be absolutely identical.

I know but in terms of code, having an API to achieve the same result on both backends has great value for maintenance I think.

>  if ipywidgets makes a scrollbar on the slider, this is bad

I would rather find a way so that the slider layout is flexible. I'm not so good with `ipywidgets` CSS rules.

I'll update the `Color Limits` group accordingly :+1: (scale, stretch, separator, tooltip)

GuillaumeFavelier:
> I would rather find a way so that the slider layout is flexible.

It's fixed in c52a386

PyVista | Notebook
--|--
image | image

GuillaumeFavelier:
> I know but in terms of code, having an API to achieve the same result on both backends has great value for maintenance I think.

It's called `_AbstractDock` for now and it must be implemented by `Brain`'s `_Renderer` (otherwise it's considered an abstract class).

larsoner:
Sounds good, let me know when I should look. And really even without the spring adjustments I mentioned we could still merge, then @hoechenberger or I could start looking into UI tweaks. It doesn't have to be perfect, just roughly as usable as what we have now is good enough and I think we're there already.

If you agree then what's left to be done? Make CIs happy?

GuillaumeFavelier:
The issues I know are:

* on the `pyvista` backend:
   * auto-scaling need to be checked to avoid regression
   * toggling interactions makes the window move
   * toggling interactions off an on resets the window layout adjusted manually
   
* on the `notebook` backend:
   * the 3d viewer might need to be resized so that the dock can fit nicely 
   * the combo boxes are broken or I can't make the default widget work
   * not all the widgets of the dock scale with the layout's width (`e.g.` the internal `VBox`es)
   * the sliders can trigger *a lot* of events/updates
   * the button widget does not expand to take all the space available in the the dock layout
   * `toggle_interface` just hide the dock, the renderer won't expand to take all the space available
   * the `matplotlib` canvas remains stubbornly aligned on the left
   
Minimum things before merge would be: CIs, testing and coverage.

The features not yet implemented would be: scale, stretch, separator and tooltip (reference: https://github.com/mne-tools/mne-python/pull/8862#issuecomment-785178468)

larsoner:
> The features not yet implemented would be: scale, stretch, separator and tooltip 

Let's just put these in an issue, these are all aesthetic and easily doable by me you @hoechenberger or someone else once you have the basic framework in place. So +1 from me to stop tweaking the display characteristics and work on getting the bugs above ironed out instead.

GuillaumeFavelier:
fb8e1acb366bfd127618c66d0e0be09b82ec1e2c fixed the main blocker on `notebook`

hoechenberger:
+5 on what @larsoner said

GuillaumeFavelier:
About the CIs, the recurring issue seems to be:

```
______________________________ test_gdf2_birthday ______________________________
mne/io/edf/tests/test_gdf.py:74: in test_gdf2_birthday
    assert raw._raw_extras[0]['subject_info']['age'] == 44
E   assert 45 == 44
E     +45
E     -44
```

but I saw it in other PRs, I don't think it's related.

larsoner:
I'll look into it

hoechenberger (approved):

GuillaumeFavelier:
I found no regression with `Auto-Scale` @agramfort, it works in this PR exactly like in `master` (5ea4bda51244d36b5af8bf198af4672e66fab6ee)

After auto-scaling on `time=0.040s`:

PR | master
--|--
image | image

GuillaumeFavelier:

larsoner:
@GuillaumeFavelier can you push a commit without the skip? It would be nice to see CIs pass (other than the pip pre ones at least)

mne/viz/_brain/_brain.py
```diff
@@ -920,169 +883,337 @@ def _configure_scalar_bar(self):
...
+                self._configure_label_time_course()
+            self._update()
 
-        # set the slider style
-        self._set_slider_style()
+        # setup label extraction parameters
+        def _set_label_mode(mode):
+            if self.traces_mode != 'label':
+                return
+            import copy
```
larsoner, line 1141: don't nest builtin imports, move to top

mne/viz/_brain/_brain.py
```diff
@@ -2723,7 +2805,8 @@ def screenshot(self, mode='rgb', time_viewer=False):
         screenshot : array
             Image pixel values.
         """
-        img = self._renderer.screenshot(mode)
+        with self._ensure_screenshot_size():
+            img = self._renderer.screenshot(mode)
```
larsoner, line 2809: This feels like a hack. `brain.screenshot` should not be something that changes the window size. The window should be sized properly at the beginning like it is on `master`.

mne/viz/backends/_notebook.py
```diff
@@ -10,58 +10,201 @@
...
-        self.actions = dict()
+        self.dock_width = 300
+        self.dock = None
+        self.actions = None
+        self.widgets = None
         kwargs["notebook"] = True
         super().__init__(*args, **kwargs)
 
+    def _add_widget(self, layout, widget):
+        from ipywidgets import HBox
```
larsoner, line 22: Rather than nesting all `ipywidgets` imports, can we avoid importing `mne.viz.backends._notebook` elsewhere in the code until we need to do so (i.e., nest its import to delay trying to import `ipywidgets`)?

mne/viz/backends/_notebook.py
```diff
@@ -10,58 +10,201 @@
 
 class _Renderer(_PyVistaRenderer):
```
larsoner, line 11: This inhertance is a bit weird, because it overrides a bunch of dock-related stuff from `_PyVistaRenderer`. How about:

1. `_PyVistaRenderer`, which handles rendering using PyVista. It is a subclass of `_AbstractRenderer` and `_AbstractDock`.
2. `_QtRenderer`, which handles dock-related stuff, subclasses both `_PyVistaRenderer` and `_QtDock`, and lives in a new `_qt.py` file (with `_QtDock`).
3. `_IPyRenderer`, which handles ipywidgets dock-related stuff, subclasses both `_PyVistaRenderer` and `_IPyDock`, the latter providing the dock methods, and lives in this `_notebook.py` file (with _IPyDock).

And really I'm not sure if we even need separate `_QtDock` and `_IPyDock` classes, these could just implement the dock methods directly on `_QtRenderer` and `_IPyRenderer`. Up to you...

Either way, this has the advantage that all Qt imports can be done at the top of `_qt.py`, and all ipywidgets can be done at the top of `_notebook.py`, etc. But more importantly, it becomes more readable and more maintainable because the hierarchy is clearer -- the split in files and classes reflects the split in functionality.

mne/viz/backends/float_slider.py
```diff
@@ -0,0 +1,52 @@
...
+# Authors: Guillaume Favelier <guillaume.favelier@gmail.com
+#          Eric Larson <larson.eric.d@gmail.com>
+#
+# License: Simplified BSD
+
+
+def float_slider_class():
+    """Return the QFloatSlider class."""
+    from PyQt5.QtCore import pyqtSignal
+    from PyQt5.QtWidgets import QSlider
```
larsoner, line 12: Shouldn't need to be nested anymore after the above refactoring

larsoner (commented):
Some high-level ideas, I haven't played with the code yet.

GuillaumeFavelier:
The current failures on CIs are:

```
__________________________________ test_plot __________________________________
mne\time_frequency\tests\test_tfr.py:522: in test_plot
    cbar = fig.get_axes()[0].CB  # Fake dragging with mouse.
E   AttributeError: 'AxesSubplot' object has no attribute 'CB'
___________________________ test_plot_epochs_clicks ___________________________
mne\viz\tests\test_epochs.py:174: in test_plot_epochs_clicks
    assert first_ch in fig.mne.info['bads']
E   AssertionError: assert 'MEG 0113' in []
____________________________ test_plot_ica_sources ____________________________
mne\viz\tests\test_ica.py:233: in test_plot_ica_sources
    assert_array_equal(ica.exclude, [0])
E   AssertionError: 
E   Arrays are not equal
E   
E   (shapes (0,), (1,) mismatch)
E    x: array([], dtype=float64)
E    y: array([0])
___________________________ test_plot_raw_selection ___________________________
mne\viz\tests\test_raw.py:258: in test_plot_raw_selection
    assert sel_fig.lasso.ec[:, 0].sum() == 1   # one channel red
E   assert 0.0 == 1
E     +0.0
E     -1
_________________________ test_plot_raw_child_figures _________________________
mne\viz\tests\test_raw.py:332: in test_plot_raw_child_figures
    assert len(fig.mne.child_figs) == 1
E   assert 0 == 1
E     +0
_____________________________ test_plot_tfr_topo ______________________________
mne\viz\tests\test_topo.py:268: in test_plot_tfr_topo
    _fake_click(fig, fig.axes[0], (0.08, 0.65))
```

I'm not sure it's related either.

mne/viz/_brain/_brain.py
GuillaumeFavelier: It is one :sweat: It's all I came up after investing a day into this. I did fd03c18 to move on.

mne/viz/_brain/_brain.py
GuillaumeFavelier: Got it!

mne/viz/backends/_notebook.py
GuillaumeFavelier: What  do you think of ?

* In `pyvista`:
  `_PyVistaRenderer` <- `_AbstractRenderer`
   `_Renderer` <- `_PyVistaRenderer`, `_AbstractDock`, `_AbstractToolBar` 
   
 * In `notebook`:
  `_Renderer` <- `_PyVistaRenderer`, `_AbstractDock`, `_AbstractToolBar`

GuillaumeFavelier: I followed it in 4228bd263e3f306c8d4d2e15cdedb0c1fa2e4520

GuillaumeFavelier: Oh I forgot to split in different files though

GuillaumeFavelier: With different files to separate implementations:

* In `pyvista`:
   `_PyVistaRenderer` <- `_AbstractRenderer`
   `_QtDock` <- `_AbstractDock` (actual implementation)
   `_QtToolBar` <- `_AbstractToolBar` (actual implementation)
   `_Renderer` <- `_PyVistaRenderer`, `_QtDock`, `_QtToolBar` 
   
 * In `notebook`:
    `_IpyDock` <- `_AbstractDock` (actual implementation)
    `_IpyToolBar` <- `_AbstractToolBar` (actual implementation)
    `_Renderer` <- `_PyVistaRenderer`, `_IpyDock`, `_IpyToolBar`

mne/viz/backends/_notebook.py
larsoner: still to be un-nested

mne/viz/backends/_pyvista.py
```diff
@@ -855,6 +856,183 @@ def _silhouette(self, mesh, color=None, line_width=None, alpha=None,
             prop.SetLineWidth(line_width)
 
 
+class _Renderer(_PyVistaRenderer, _AbstractDock, _AbstractToolBar):
```
larsoner, line 859: I still think this should live in a `_qt.py` to mirror `_notebook.py`

mne/viz/backends/abstract_tool_bar.py
```diff
@@ -0,0 +1,38 @@
...
+#
+# License: Simplified BSD
+
+from abc import ABC, abstractmethod
+
+
+class _AbstractToolBar(ABC):
+    @abstractmethod
+    def _tool_bar_load_icons(self):
+        pass
```
larsoner, line 14: Can we try a variant of https://stackoverflow.com/questions/9202723/excluding-abstractproperties-from-coverage-reports#answer-41284681 ?

mne/viz/backends/float_slider.py
larsoner: still true, and even better this can live in the new `_qt.py`

larsoner (commented):
Other than still wanting some of the `_pyvista.py` code moved to `_qt.py` and wanting some imports un-nested this LGTM! Let me know if you want me just to make those changes so you can work on something more interesting. Also I want to take a stab at fixing the screenshot size issue, that's a regression I'd rather not put in `main`.

GuillaumeFavelier:
I'm not sure to understand how splitting into different files will fix the issue with Qt :thinking: 

Now I'll push commits to move some code around and un-nest the imports then I'll leave you in charge @larsoner 

> Also I want to take a stab at fixing the screenshot size issue, 

You're more than welcome to do so.

GuillaumeFavelier:
Also a naive remark but I'm confused with the naming convention for files. Why `abstract_dock.py` is okay but `qt.py` or `pyvista.py` is not :confused:

larsoner:
> I'm not sure to understand how splitting into different files will fix the issue with Qt ￼

It won't, they're separate issues. We both need to fix the screenshot stuff, and separately I think...

> Also a naive remark but I'm confused with the naming convention for files. Why abstract_dock.py is okay but qt.py or pyvista.py is not ￼

... that right now you have:

1. Qt-dock-plus-PyVista-renderer, i.e., a `_Renderer` in `_pyvista.py` and
2. ` IPython-dock-plus-PyVista-renderer in `_notebook.py`

My proposal is to move the Qt dock stuff -- and the Qt-based `_Renderer` -- to a new `_qt.py`, which then will have the same structure as the IPython dock and IPython-based `_Renderer` live in `_notebook.py`. make sense?

GuillaumeFavelier:
> I'm confused with the naming convention for files

I realize I was not clear enough here, I'm talking about the underscore character before the file name. What is the condition to add one?

> My proposal is to move the Qt dock stuff -- and the Qt-based _Renderer -- to a new _qt.py

I get it but why put everything qt related in one file? This is not exactly what I expected by splitting :sweat_smile: 

I implemented https://github.com/mne-tools/mne-python/pull/8862#discussion_r583579492 in d47018ed066821d554a36434bf382f7cef9452fc (I know CIs won't appreciate).

Feel free to revert or modify @larsoner

larsoner:
> I get it but why put everything qt related in one file? This is not exactly what I expected by splitting sweat_smile

The idea is that you have 3D-rendering-related stuff in `_pyvista.py`, Qt-GUI-related stuff in `_qt.py`, and Notebook-GUI-related stuff in `_notebook.py`. To me this is the clearest separation of functionality. So if we're going to have notebook-related stuff in `_notebook.py`, we should have Qt-related stuff in a `_qt.py` for consistency. Perhaps it becomes clearer if you imagine if we added a third or fourth option for GUI abstraction...?

And as a side effect, by waiting to import the Qt or IPy parts in the code with a nested import, it means that we can un-nest our Qt and IPy imports in `_qt.py` and `_notebook.py`. Not only does this clean the code, in the future it might make it easier to use qtpy or something since we just change stuff in one place (at the top of the file) instead of a bunch of places (at all our nested-qt-import points).

larsoner:
>  I'm talking about the underscore character before the file name. What is the condition to add one?

Usually we add them when we want something to be private. Technically we don't have to add them here because `_brain` is already a private namespace, but no harm in having them

larsoner:
Having now implemented the `_qt.py` and `_notebook.py`, I can say that I do like it -- one you're living in "Qt land" and in the other you're living in "ipywidgets land", it seems pretty natural. Have a look and see if you agree...

Now onto the annoying resizing problem!

mne/viz/_brain/_brain.py
larsoner: The problem is that the left area is too big, so it forces the size to be too large. I'm going to try putting it inside a scrollable area. This is how freeview works. Also we need it for sustainability anyway as we add more controls.

larsoner: Screenshot from 2021-02-26 10-14-57

larsoner: All better:

Screenshot from 2021-02-26 13-41-28

will push a commit shortly

larsoner:
@GuillaumeFavelier sizing issue should be fixed, take a look at the code and `_qt.py` and `_notebook.py` etc. and make sure I didn't do something bad in terms of the hierarchy.

The sizing fix was QScrollArea + `scroll.setMinimumSize(scroll.sizeHint().width, 0)`, i.e., make sure it always occupies its minimum width, then our `_ensure_minimum_sizes` and event processing takes care of the rest (i.e., resizing the window for us to accommodate the brain + dock). I also fixed the movie generation, no need to toggle the dock when capturing a movie anymore!

agramfort:
screenshot now only takes a copy of the brain (no time series).

I am saying why not but is it a conscious choice?

for the me the scale control for clim does not behave smoothly.

I can show you by sharing my screening my screen on monday

>

larsoner:
> screenshot now only takes a copy of the brain (no time series).

This behavior should change depending on whether you do `time_viewer=True` or (default) `False`

agramfort:
with time viewer True:
pouet2

with time viewer False:
pouet

using this code

```
import os
import mne
initial_time = 0.13
time_viewer = False
# time_viewer = True
                 hemi='split',
                 views=['lat', 'med'],
                 time_viewer=time_viewer
                 )
```

GuillaumeFavelier:
How do you take the screenshot @agramfort ?

GuillaumeFavelier:
I pushed a commit to fix:

```
AttributeError: module 'mne.viz.backends._qt' has no attribute '_close_3d_figure'
AttributeError: module 'mne.viz.backends._qt' has no attribute '_set_3d_view'
```

GuillaumeFavelier:
> one you're living in "Qt land" and in the other you're living in "ipywidgets land"

I admit this separation is simpler :bow_and_arrow: and I like `_abstract.py` too

GuillaumeFavelier:
Locally by using `Brain` in `ipython`, I obtain the size I request for the screenshot but during testing with `pytest`, `test_brain_time_viewer` still fails with: 

```
===================================================================================================== FAILURES =====================================================================================================
_________________________________________________________________________________________ test_brain_time_viewer[pyvista] __________________________________________________________________________________________
mne/viz/_brain/tests/test_brain.py:506: in test_brain_time_viewer
    assert_allclose(img.shape, want_shape)
E   AssertionError: 
E   
E   Mismatched elements: 1 / 3 (33.3%)
E   Max absolute difference: 282.
E   Max relative difference: 0.94
E    x: array([300, 582,   3])
E    y: array([300., 300.,   3.])
```

Reproduced in https://github.com/mne-tools/mne-python/runs/1989903394#step:13:3644 for example

GuillaumeFavelier:
I pushed a commit to fix the screenshot taken *from the interface*. I'm surprised it was not reported sooner.

hoechenberger:
> I'm surprised it was not reported sooner.

Sorry 😢

larsoner:
> Locally by using Brain in ipython, I obtain the size I request for the screenshot but during testing with pytest, test_brain_time_viewer still fails with:

This is because we toggle the interface off earlier in the test, which makes the 3D area take up the entire window width, which I'd consider expected behavior (though in the future we could consider shrinking the window). Toggling the interface back on before doing the screenshot fixes it locally, will push a fix.

GuillaumeFavelier:
I pushed a commit to fix `AttributeError` from `_set_3d_title` and `_take_3d_screenshot` as well.

larsoner:
Okay, let's get this in and iterate further on the interface as necessary. Thanks a ton @GuillaumeFavelier , this is great!

hoechenberger:
@GuillaumeFavelier You're a wizard 🧙🏻‍♂️ Thank you!

PR #8863, opened 2021-02-11
WIP, ENH: Make "verbose" decorator provide help for static type checkers like Pylance

hoechenberger:
The `@verbose` decorator used to "eat up" the parameters of the decorated
functions, leading to an awfully useless experience with Pylance in VS Code.

Thanks to @cbrnr's persitence, a Pylance dev pointed us to the correct
solution, which is implemented in this PR:
See https://github.com/microsoft/pylance-release/issues/346#issuecomment-777490280 

### `main`:
Screen Shot 2021-02-11 at 19 49 42

### this PR:
Screen Shot 2021-02-11 at 19 49 23

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -74,6 +74,8 @@ Enhancements
 
 - The signal of ``resp`` (respiratory) channels is now assumed to be in the unit Volt  (:gh:`8858` by `Richard Höchenberger`_)
 
+- Static type checkers like Pylance (comes with VS Code) now display the parameters of many more functions correctly, largely improving overall usability for VS Code users (:gh:`8862` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix bug with `mne.connectivity.spectral_connectivity` where time axis in Epochs data object was dropped. (:gh:`8839` **by new contributor** |Anna Padee|_)
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -689,6 +689,8 @@ numpydoc_xref_ignore = {
     # unlinkable
     'mayavi.mlab.pipeline.surface',
     'CoregFrame', 'Kit2FiffFrame', 'FiducialsFrame',
+    # type hint used in @verbose decorator
+    '_FuncT',
 }
 numpydoc_validate = True
 numpydoc_validation_checks = {'all'} | set(error_ignores)
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -123,6 +123,7 @@ nitpick_ignore = [
     ("py:class", "an object providing a view on D's values"),
     ("py:class", "a shallow copy of D"),
     ("py:class", "(k, v), remove and return some (key, value) pair as a"),
+    ("py:class", "_FuncT"),  # type hint used in @verbose decorator
 ]
 for key in ('AcqParserFIF', 'BiHemiLabel', 'Dipole', 'DipoleFixed', 'Label',
             'MixedSourceEstimate', 'MixedVectorSourceEstimate', 'Report',
@@ -689,8 +690,6 @@ numpydoc_xref_ignore = {
     # unlinkable
     'mayavi.mlab.pipeline.surface',
     'CoregFrame', 'Kit2FiffFrame', 'FiducialsFrame',
-    # type hint used in @verbose decorator
-    '_FuncT',
 }
 numpydoc_validate = True
 numpydoc_validation_checks = {'all'} | set(error_ignores)
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -12,6 +12,7 @@ import sys
 import logging
 import os.path as op
 import warnings
+from typing import Any, Callable, TypeVar
 
 from .docs import fill_doc
 from ..externals.decorator import FunctionMaker
@@ -49,7 +50,13 @@ _filter = _FrameFilter()
 logger.addFilter(_filter)
 
 
-def verbose(function):
+# Provide help for static type checkers: see
+# https://github.com/microsoft/pyright/blob/8ca2c14a7a71a5f19db2082ecb531cb0f76b484d/docs/type-stubs.md#cleaning-up-generated-type-stubs  # noqa
+# point 3
+_FuncT = TypeVar('_FuncT', bound=Callable[..., Any])
+
+
+def verbose(function) -> Callable[[_FuncT], _FuncT]:
     """Verbose decorator to allow functions to override log-level.
 
     Parameters
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -50,13 +50,12 @@ _filter = _FrameFilter()
 logger.addFilter(_filter)
 
 
-# Provide help for static type checkers: see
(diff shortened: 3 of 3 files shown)


cbrnr:
Now if only those type hints could be disabled...

hoechenberger:
> Now if only those type hints could be disabled...

Aren't they actually the most important feature? They're currently useless with MNE – something I want to look into soon.
In some other Python projects, and also in TypeScript (typed JavaScript), the typing information has saved me from numerous bugs already.

cbrnr:
Long story, but no. I don't like that they are being forced onto people. If Pylance doesn't have an option to disable them I'll not going to use it. Pulling types from docstrings would be nice though because that's what we already have.

hoechenberger:
> If Pylance doesn't have an option to disable them I'll not going to use it.

There's still Jedi for you then, I guess :)

hoechenberger:
> Pulling types from docstrings would be nice though because that's what we already have.

I'm actually surprised this isn't already happening, I thought this was implemented … I will investigate.

agramfort (approved):

hoechenberger:
@drammock @larsoner Any idea how to deal with the doc building problem?

```
/home/circleci/project/mne/utils/_logging.py:docstring of mne.utils._logging.verbose:: WARNING: py:class reference target not found: _FuncT
/home/circleci/project/mne/utils/_logging.py:docstring of mne.utils._logging.verbose:: WARNING: py:class reference target not found: _FuncT
```

hoechenberger:
> > Pulling types from docstrings would be nice though because that's what we already have.
> 
> I'm actually surprised this isn't already happening, I thought this was implemented … I will investigate.

Seems it's currently not supported :(
https://github.com/microsoft/pylance-release/issues/854

cbrnr:
But the good news is that apparently it is supported by PyCharm and Jedi!

hoechenberger:
Not sure what happened, but suddenly it's not working properly for me anymore. Please refrain from merging for now, I need to investigate.

larsoner:
My guess is that make Circle happy you need to add `_FuncT` to [numpydoc_xref_ignore](https://github.com/mne-tools/mne-python/blob/main/doc/conf.py#L665)

larsoner:
Regarding it working then not working anymore, it might depend on what branch you were on when you opened vs code. Maybe you weren't on this branch the most recent time you opened it?

hoechenberger:
Should be working now.

@cbrnr, @agramfort, care to give it a shot?


@larsoner When doing my initial testing, I somehow managed to sneak a typo somewhere (didn't end up in the final commit) which effectively broke the decorator, thereby mysteriously fixing the doctoring rendering 💩 Software is strange sometimes.

hoechenberger:
@MaelysSolal This change could be of interest to you: way fewer (misleading!) red wiggly lines in VS Code.

cbrnr (approved):
Very nice, works as it should. Also, useless type hints ("unknown") are not displayed anymore either.

cbrnr:
BTW, if you look at the docstring of `mne.time_frequency.tfr_multitaper`, what does the tilde mean in references such as `~mne.time_frequency.tfr_array_multitaper`?

hoechenberger:
> Also, useless type hints ("unknown") are not displayed anymore either.

I'm still seeing those though! You sure you didn't accidentally use Jedi?

> what does the tilde mean in references

In the rendered docs, `~mne.foo.bar.baz` would only show up as `baz` (or only as `foo.bar.baz`?? in any case, the `mne` part will be skipped)

hoechenberger:
> > Also, useless type hints ("unknown") are not displayed anymore either.
> 
> I'm still seeing those though! You sure you didn't accidentally use Jedi?

It's because I had `python.analysis.typeCheckingMode` set to `basic`. Setting it to `off` (default) removes the unhelpful type hints.

--> So this seems good to merge!

hoechenberger:
> So this seems good to merge!

No, wait, doc build is still failing.

@larsoner @drammock I've added '_FuncT'` to `numpydoc_xref_ignore`, yet I'm still seeing 
```
/home/circleci/project/mne/utils/_logging.py:docstring of mne.utils._logging.verbose:: WARNING: py:class reference target not found: _FuncT
/home/circleci/project/mne/utils/_logging.py:docstring of mne.utils._logging.verbose:: WARNING: py:class reference target not found: _FuncT
```
Any idea how to resolve this?

cbrnr:
> It's because I had `python.analysis.typeCheckingMode` set to basic. Setting it to off (default) removes the unhelpful type hints.

Nice!

> In the rendered docs, `~mne.foo.bar.baz` would only show up as `baz` (or only as `foo.bar.baz`?? in any case, the `mne` part will be skipped)

What kind of convention is this? Neither VS Code nor PyCharm render these correctly, IIRC we used to write these things differently in the past. We should consider not using this syntax if there is an alternative that can be rendered. But anyway, this is another issue.

hoechenberger:
> What kind of convention is this? Neither VS Code nor PyCharm render these correctly, IIRC we used to write these things differently in the past. We should consider not using this syntax if there is an alternative that can be rendered. But anyway, this is another issue.

I think it's MNE-specific.

The MNE docstrings are a nightmare in VS Code for me too :( Mostly unusable. At least I get function signatures now with this PR…

cbrnr:
Thanks @hoechenberger!

larsoner:
>> In the rendered docs, ~mne.foo.bar.baz would only show up as baz

> What kind of convention is this?

Aside from import-time population using `docdict`, our docstrings follow standard reStructuredText syntax (following the NumpyDoc standard as well) meant to be rendered by Sphinx. The `~` prefix in an interpreted text role [shortens the path to just the last element](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax).

larsoner:
> @larsoner @drammock I've added '_FuncT' to numpydoc_xref_ignore, yet I'm still seeing

Did you try my suggestion [above](https://github.com/mne-tools/mne-python/pull/8863#issuecomment-777800926):

    My guess is that make Circle happy you need to add _FuncT to numpydoc_xref_ignore

It's a little bit better than `nitpick_ignore` if it works (more targeted / clearer why the error is happening).

larsoner:
Ahh based on your commit it looks like you did. So it's coming from Sphinx directly, not numpydoc. `nitpick_ignore` should be fine for this case then!

cbrnr:
Thanks @larsoner - I wonder if we have to use this convention because according to the definition:

(1) This does not work with all cross-reference roles, but is domain specific.
(2) In HTML output, the link’s `title` attribute (that is e.g. shown as a tool-tip on mouse-hover) will always be the full target name.

An alternative would be to use :role:`title <target>` instead (but I'm not sure if this is better supported by IDEs).

larsoner:
We've used it for years now without issue, I don't think it's worth using the more verbose `title <target>` pattern. Our domain is Python, probably the most widely used Sphinx domain, and it's supported there.

cbrnr:
I agree it's not worth it, it seems like IDEs don't render such links with roles correctly in any case.

PR #8867, opened 2021-02-12
MRG, DOC, ENH: new frequency-tagging tutorial

dominikwelke:
we propose a new tutorial for basic analysis of frequency-tagging data.
it comes with an example dataset from visual stimulation with inverting checkerboards.

#### Reference issue
no fixes, but loosely related to discussion about a new spectra class (#7671)

#### authors
@dominikwelke  @kalenkovich


agramfort:
thx @dominikwelke !

how big are the files? can we add this to the doc build to see it rendered on circleci?

dominikwelke:
hi @agramfort, thanks for the quick response!

> how big are the files? 

its about 40mb. i published the data on osf and the script fetches them from there

> can we add this to the doc build to see it rendered on circleci?

locally it builds fine. how do i add it to the doc build?

agramfort:
if you call your tutorial plot_ssvep.py it will be automatically rendered on the website.
Can you try here while fixing the conflicting file?

thanks

agramfort:
@dominikwelke can you rename your tutorial to plot_ssvp.py ?

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,780 @@
...
+Basic analysis of an SSVEP/vSSR dataset
+=======================================
+
+In this tutorial we compute the frequency spectrum and quantify signal-to-noise
+ratio (SNR) at a target frequency in EEG data recorded during fast periodic
+visual stimulation (FPVS).
+Extracting SNR at stimulation frequency is a simple way to quantify frequency
+tagged responses in MEEG (a.k.a. steady state visually evoked potentials,
+SSVEP, or visual steady-state responses, vSSR in the visual domain,
+or auditory steady-state responses, aSSR in the auditory domain).
```
rob-luke, line 12: The a is capitalised in ASSRs.
```suggestion
or auditory steady-state responses, ASSR in the auditory domain).
```

rob-luke (commented):
This is awesome. And I think its great that although you are doing VSSEPs you included terms such as ASSR, this should improve search engine hits.

I think their might be some differences in terminology between the visual and auditory fields in the topic of steady state responses. I had never heard of frequency tagging, so I already learnt a lot from this tutorial. Another example is that in ASSRs we typically use[ F-tests and Hotelling T^2 tests](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3387309/), whereas you seem to use a t-test. Once the tutorial is building in the CI do you mind if I suggest some additional terms in the text that might help ASSR researchers find this great tutorial?

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,780 @@
...
+###############################################################################
+# Calculate power spectral density (PSD)
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# We use Welch's method for frequency decomposition, since it is really fast.
+# We chose a frequency resolution of 0.1 hz.
+# You could compare it with, e.g., multitaper to get an impression of the
+# influence on SNR. All the other methods implemented in MNE can be used as
+# well.
+
+tmin = 0.
```
rob-luke, line 134: From the description above it seems that a change in the stimulus occurs at 0s (a switch between 12 and 15 hz). Im not sure about VSSEPs, but for ASSRs this would elicit a neural change response. So we would typically exclude the first ~1s of the epoch as this will have broadband frequency components generated by the change response. Perhaps it would be informative to plot the average epoch to show if this is the case or not? This is 100% curiosity from me about the differences between visual and auditory responses.

rob-luke:
FYI @jundurraga

tutorials/time-freq/plot_ssvep.py
dominikwelke: this is really a toy dataset: 
i presented checkerboards that invert b/w fields with a fix frequency, either 12 or 15 hz depending on the trial (this elicits a steady-state response). there is also an ITI of a second or so.

yes, i agree that excluding the first second would probably improve the signal quality. i didnt do it with these data, explicitly, but towards the end of the tutorial i simulated shorter trials and, in the very first seconds of the trials snr seems lower

tutorials/time-freq/plot_ssvep.py
dominikwelke: > This is awesome. And I think its great that although you are doing VSSEPs you included terms such as ASSR, this should improve search engine hits.
> 
> I think their might be some differences in terminology between the visual and auditory fields in the topic of steady state responses. I had never heard of frequency tagging, so I already learnt a lot from this tutorial. Another example is that in ASSRs we typically use[ F-tests and Hotelling T^2 tests](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3387309/), whereas you seem to use a t-test. Once the tutorial is building in the CI do you mind if I suggest some additional terms in the text that might help ASSR researchers find this great tutorial?

thanks for the nice feedback and the suggestions @rob-luke ! 
happy to include the additional stats as soon as the the tutorial builds successfully.

agreed, the range of terminology with this methodology seem alomost unlimited :) 
it's one of the goals of the tutorial to include as many of these keywords as possible and keep it general purpose.
in terms of basic analysis it doesnt really matter which sensory domain you used to elicit a steady-state/frequency-tagged response

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+standard.
+
+OUTLINE:
+
+- We will visualize both the PSD and the SNR spectrum of our epoched data.
+
+- We will extract SNR at stimulation frequency for all trials and channels.
+
+- We will show, that we can statistically separate 12hz and 15 hz responses
+in our data.
```
larsoner, line 32:
```suggestion
- We will show, that we can statistically separate 12hz and 15 hz responses
  in our data.
```

dominikwelke:
thank you @larsoner , i was unable to track this problem..

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+            abs(np.subtract(windowed_freqs, 12.)))].mean(axis=1)
+
+plt.figure()
+plt.plot(window_starts, np.array(window_snrs))
+plt.title('Time resolved 12hz SNR - %is sliding window' % window_length)
+plt.legend(['individual trials'])
+plt.ylabel('average SNR')
+plt.xlabel('t0 of analysis window [s]')
+plt.axhline(1, ls='--', c='r')
+plt.show()
```
larsoner, line 777: The global-state-based mechanics are less preferred nowadays compared to something class-based like:
```
fig, ax = plt.subplots(1)
ax.plot(...)
ax.set(title=..., ylabel=..., xlabel=...)
ax.legend(...)
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+
+##############################################################################
+# Well.. turns out this was a bit too optimistic ;)
```
larsoner, line 780: The runtime on CircleCI is over a minute. Ideally an example like this would be < 30 sec. Do you see places where it could be cut down a bit?

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+We use a simple example dataset with frequency tagged visual stimulation:
+N=2 participants observed checkerboards patterns inverting with a constant
+frequency of either 12Hz of 15Hz. 10 trials of 20s length each.
+32ch wet EEG was recorded.
+
+Data can be downloaded at https://osf.io/7ne6y/
+Data format: BrainVision .eeg/.vhdr/.vmrk files organized according to BIDS
+standard.
+
+OUTLINE:
```
larsoner, line 25: Instead of doing this, I would just make these three the second-level headings (the title of the example being the first-level heading):
```
Visualizing PSD and SNR spectrum
--------------------------------

Data preprocessing
^^^^^^^^^^^^^^^^^^

...

Estimating SNR at stimulation frequency
---------------------------------------

...

Statistically separating 12 and 15 Hz responses
-----------------------------------------------

...
```

Usually in RST you would then do a `.. contents:: local` or so, but it turns out we are about to update our website theme to automatically add this, e.g., the right bar here:

https://25174-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/machine-learning/plot_sensors_decoding.html

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+DATA:
+
+We use a simple example dataset with frequency tagged visual stimulation:
+N=2 participants observed checkerboards patterns inverting with a constant
+frequency of either 12Hz of 15Hz. 10 trials of 20s length each.
+32ch wet EEG was recorded.
+
+Data can be downloaded at https://osf.io/7ne6y/
+Data format: BrainVision .eeg/.vhdr/.vmrk files organized according to BIDS
+standard.
```
larsoner, line 23: The data download location is probably not needed. The data format specs like being in BIDS could probably live in `datasets_index.rst` (where an entry is needed anyway, e.g., with a link ref like `ssvep-dataset`) so that this tutorial can focus on the processing mechanics and justification:

https://github.com/mne-tools/mne-python/blob/main/doc/overview/datasets_index.rst

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+Extracting SNR at stimulation frequency is a simple way to quantify frequency
+tagged responses in MEEG (a.k.a. steady state visually evoked potentials,
+SSVEP, or visual steady-state responses, vSSR in the visual domain,
+or auditory steady-state responses, ASSR in the auditory domain).
+
+DATA:
+
+We use a simple example dataset with frequency tagged visual stimulation:
+N=2 participants observed checkerboards patterns inverting with a constant
+frequency of either 12Hz of 15Hz. 10 trials of 20s length each.
```
larsoner, line 18: 10 trials of 20s is probably a detail for `datasets_index.rst`

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
+"""
+=======================================
+Basic analysis of an SSVEP/vSSR dataset
+=======================================
+
+In this tutorial we compute the frequency spectrum and quantify signal-to-noise
+ratio (SNR) at a target frequency in EEG data recorded during fast periodic
+visual stimulation (FPVS).
```
larsoner, line 8:
```suggestion
visual stimulation (FPVS) at 12 Hz and 15 Hz in different trials
(see :ref:`ssvep-dataset` for more information).
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# to the average power in its neighboring bins.
+# This procedure has two advantages over using the raw PSD:
+#
+# * it normalizes the spectrum and accounts for 1/f power decay.
+#
+# * power modulations which are not very narrow band will disappear.
+#
+#
+
+###############################################################################
```
larsoner, line 125:
```suggestion
#
```
No need to separate continuous text blocks

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# super constant, or frequency bands are very narrow).
+#
+# The function below does what we want.
+#
+
+def snr_spectrum(psd, noise_n_neighborfreqs=1, noise_skip_neighborfreqs=1):
+    """
+    Parameters
+    ----------
+    psd - np.array
```
larsoner, line 168: numpydoc standard
```suggestion
    psd : ndarray
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# plotting PSD on a log scale).
+# Hence SNR values must be positive and can minimally go towards 0.
+# You can nicely see this at 50hz with the artifact induced by notch
+# filtering the line noise:
+# The PSD spectrum shows a prominent negative peak and average SNR approaches
+# 0.
+#
+
+
+###############################################################################
```
larsoner, line 296:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+#
+# Let's have a look at the trials with 12 hz stimulation, for now.
+#
+
+# define stimulation frequency
+stim_freq = 12.
+
+###############################################################################
+# get index for the stimulation frequency (12hz)
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
larsoner, line 323: too many ^'s, should match title. Title should also have capital G

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# It is very easy - even unintended - to end up double-dipping or p-hacking.
+# So if you want to work with an ROI or individual channels, ideally select
+# them *a priori* - before collecting or looking at the data - and preregister
+# this decision so people will believe you.
+# If you end up selecting an ROI or individual channel for reporting *because
+# this channel or ROI shows an effect*, e.g. in an explorative analysis, this
+# is also fine but make it transparently and correct for multiple comparison.
+#
+
+##############################################################################
```
larsoner, line 440:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# So that's it, we hope you enjoyed our little tour through this example
+# dataset.
+#
+# As you could see, frequency-tagging is a very powerful tool that can yield
+# very high signal to noise ratios and effect sizes that enable you to detect
+# brain responses even within a single participant and single trials of only
+# a few seconds duration.
+#
+
+##############################################################################
```
larsoner, line 546:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# Using Welch's method as implemented in MNE makes it very easy to change
+# the amount of data that is actually used in the spectrum
+# estimation. We can tweak both the length of the actual trial, and the length
+# of the Welch window.
+#
+# Here we employ this to show you some features of frequency
+# tagging data that you might or might not have already intuitively expected:
+#
+
+##############################################################################
```
larsoner, line 561:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+#
+# This should be easy to understand: in longer recordings there is simply
+# more signal (one second of additional stimulation adds, in our case, 12
+# cycles of signal) while the noise is (hopefully) stochastic and not locked
+# to the stimulation frequency.
+# In other words: with more data the signal term grows faster than the noise
+# term.
+#
+
+##############################################################################
```
larsoner, line 622:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,786 @@
...
+# a longer Welch window results in higher average SNR values;
+# a shorter window results in lower values but is more robust /
+# there is a higher confidence in the mean.
+#
+# Yet, one should also not pick too small a window: in our data we can see
+# that the very short Welch windows < 2-3s are not great -
+# here we've hit the noise bottom.
+#
+
+##############################################################################
```
larsoner, line 730:
```suggestion
```

tutorials/time-freq/plot_ssvep.py
kalenkovich: I'll take care of it.

tutorials/time-freq/plot_ssvep.py
kalenkovich: Fixed.

tutorials/time-freq/plot_ssvep.py
kalenkovich: Fixed.

kalenkovich:
> Another example is that in ASSRs we typically use[ F-tests and Hotelling T^2 tests](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3387309/), whereas you seem to use a t-test. 

Hey, @rob-luke! Could you please tell us what kind of questions would you try answering with an F-test or a Hotelling T²? This would help us incorporating these tests into the tutorial.

From a brief look at [the paper you cited](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3387309/) I gathered that you would use an F-test to compare the power at the stimulation frequency to the sum of the power at the neighboring frequencies. This would tell you whether there is a peak above the noise. I don't think that we need this here. Considering the size of the peaks, it seems a bit gratuitous to test what we already know to be true. That said, if this is a common test done in practice we should probably add it anyways. Is it common to test for the presence of the peak this way? How would one interpret a non-significant value? Did I understand correctly that an F-test would be used to test for the presence of the peak?

The use of the Hotelling T² is somewhat less intuitive: a) it treats complex and real part of the Fourier coefficients as two variables, b) it compares their average squares to the squares of their averages (total power vs. stimulus-locked power). Again, as far as I can tell, this is used to check if there is a peak at a given frequency bin. Is that correct?

rob-luke:
> this is really a toy dataset

Its a **great** toy dataset. But this example is also going to become the basis of peoples analysis so its important to make sure its correct. You've addressed one common mistake people make of including neighbouring noise frequencies using `noise_skip_neighborfreqs`. The other common mistake I see when reviewing papers is people including the initial transient response in their evaluation of the steady state component. Ideally you could plot the average epoch and illustrate how long the transient response duration, and remove this from the start of the epoch by adjusting tmin. Or at least mention it, particuarly given you explore trial length at the end of the document.

> Hey, rob-luke! Could you please tell us what kind of questions would you try answering with an F-test or a Hotelling T²? This would help us incorporating these tests into the tutorial.

I don't suggest we add these tests to this tutorial. It would be a major addition. Honestly I think I just created noise for you here. I was just so excited to see a great steady state tutorial that I started thinking out loud. At most you could add a comment saying "This tutorial doesn't address determining the presence of a neural response, an F-test or Hotelling T² would be more appropriate for these purposes." But only if you want.

> Is it common to test for the presence of the peak this way? How would one interpret a non-significant value? Did I understand correctly that an F-test would be used to test for the presence of the peak? ... the Hotelling T² is somewhat less intuitive. as far as I can tell, this is used to check if there is a peak at a given frequency bin. Is that correct?

You're interpretation of the tests is correct, they are used to detect the presence of a neural response. These tests are very widely used in the auditory field. Most commonly they are used in objective audiometry or newborn hearing screening to objectively determine if someone can hear a sound. In this case, the absence of a response indicates that someone may have some hearing loss and should be refered for further testing (or a cable isnt plugged in, or they have auditory neuropathy). An F-test compares the frequency of interest to neigbouring frequencies to determine an SNR and presence of a response, but can be biased due to the 1/f structure of the EEG signal. The HT2 test uses the variability within the frequency bin across trials as the noise estimate to determine the SNR and overcomes the 1/f issue. But having read your tutorial again I assume you are coming from a more BCI perspective.


> Do you see places where it could be cut down a bit?

I think you could remove the bonus exercises. You could also remove the notch filter as you dont examine this frequency.

kalenkovich:
Thank you for the explanation, @rob-luke!

> Ideally you could plot the average epoch and illustrate how long the transient response duration, and remove this from the start of the epoch by adjusting tmin. Or at least mention it, particuarly given you explore trial length at the end of the document.

We'll have to check whether we can get a decent ERP from a single participant in order to assess the early visual response duration. We can mention it, of course. Also, we can see what cutting a bit at the beginning of a trial does to SNR. My guess is that it won't matter for 20s-long epochs but will have an effect on, say, 5s-long ones.

> I don't suggest we add these tests to this tutorial. It would be a major addition. Honestly I think I just created noise for you here. I was just so excited to see a great steady state tutorial that I started thinking out loud. At most you could add a comment saying "This tutorial doesn't address determining the presence of a neural response, an F-test or Hotelling T² would be more appropriate for these purposes." But only if you want.

Adding a note with a reference to an article makes total sense.

> Do you see places where it could be cut down a bit?

I think there was a misunderstanding. This was just a technical question about the tutorial taking too long to build, not too long to read 😄

agramfort:
See rendered tutorial https://25218-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/time-freq/plot_ssvep.html#sphx-glr-auto-tutorials-time-freq-plot-ssvep-py

i see a few formatting issues

dominikwelke:
> (...) including the initial transient response in their evaluation of the steady state component. 

just a note: at the moment we use welch's method with a 10s window and little overlap for the spectrum estimation. hence with 20s trials the first half of each trial (potentially including the transient response) contributes only 50% to the final SNR measure. 
i would maybe even increase the overlap, which would further decrease the relative influence of an impaired segment.

of course we could also use plain fft and the full trial for the frequency spectrum. 
using welch rather reflects our personal decision for a bit more robustness vs. the highest possible snr value - correct me if i'm wrong but in my experience these values are not very informative in themselves anyway, not across participants..
but i wouldnt even know another paper that uses welch from the top of my head :)


> The runtime on CircleCI is over a minute..

@rob-luke 's suggestion to cut parts of the bonus exercise would probably work. this part takes the longest, computation wise, since it a series of loops with psd estimation + snr extraction cylces for each single bar in the boxplots.

also i think it wouldn't hurt, since there is some redundancy in there..
@rob-luke @larsoner : any opinion on what exactly to cut, and what to keep here?

tutorials/time-freq/plot_ssvep.py
agramfort: +1 to link to datasets_index page.

btw it also does not render well.

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+    montage_style,
+    head_size=0.095)  # head_size parameter default = 0.095
+raw.set_montage(montage, verbose=False)
+
+# Set common average reference
+raw.set_eeg_reference('average', projection=False, verbose=False)
+
+# Apply notch filter to remove line-noise
+notch = np.arange(raw.info['line_freq'], raw.info['lowpass'] / 2,
+                  raw.info['line_freq'])
```
agramfort, line 85:
```suggestion
notch = np.arange(raw.info['line_freq'], raw.info['sfreq'] / 2,
                  raw.info['line_freq'])
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+# Hence, we need to set some parameters for this baseline - how many
+# neighboring bins should be taken for this computation, and do we want to skip
+# the direct neighbors (this can make sense if the stimulation frequency is not
+# super constant, or frequency bands are very narrow).
+#
+# The function below does what we want.
+#
+
+def snr_spectrum(psd, noise_n_neighborfreqs=1, noise_skip_neighborfreqs=1):
+    """
```
agramfort, line 162: missing docstring header line

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+    Parameters
+    ----------
+    psd : ndarray
+        containing psd values as spit out by mne functions. must be 2d or 3d
+        with frequencies in the last dimension
+    noise_n_neighborfreqs - int
```
agramfort, line 168:
```suggestion
    noise_n_neighborfreqs : int
```
to be fixed everywhere in docstring

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+        number of neighboring frequencies used to compute noise level.
+        increment by one to add one frequency bin ON BOTH SIDES
+    noise_skip_neighborfreqs - int
+        set this >=1 if you want to exclude the immediately neighboring
+        frequency bins in noise level calculation
+
+    Returns
+    -------
+    snr - np.array
```
agramfort, line 177: can you provide the shape ?

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+i_bin_12hz = np.argmin(abs(np.subtract(freqs, stim_freq)))
+# could be updated to support multiple frequencies
+
+# for later, we will already find the 15 hz bin and the 1st and 2nd harmonic
+# for both.
+i_bin_24hz = np.argmin(abs(np.subtract(freqs, 24)))
+i_bin_36hz = np.argmin(abs(np.subtract(freqs, 36)))
+i_bin_15hz = np.argmin(abs(np.subtract(freqs, 15)))
+i_bin_30hz = np.argmin(abs(np.subtract(freqs, 30)))
+i_bin_45hz = np.argmin(abs(np.subtract(freqs, 45)))
```
agramfort, line 334: you should use 

https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+
+###############################################################################
+# get indices for the different trial types
```
agramfort, line 337:
```suggestion
# Get indices for the different trial types
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,767 @@
...
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+i_trial_12hz = np.where(epochs.events[:, 2] == event_id['12hz'])
+i_trial_15hz = np.where(epochs.events[:, 2] == event_id['15hz'])
+
+###############################################################################
+# get indices for the EEG channels forming the ROI
```
agramfort, line 344:
```suggestion
# Get indices for the EEG channels forming the ROI
```

rob-luke:
> i wouldnt even know another paper that uses welch from the top of my head

My recollection is that this affects the false positive rate when detecting the presence of responses. But I'm not 100% on this. However, if FFT is the common approach then I'd stick to it for a demo. But if you want to keep as is, is it just a one line change to switch to FFT? If so, it might be informative to demonstrate how both could be done (just in a comment, don't repeat the whole analysis). 

> these values are not very informative in themselves anyway, not across participants

Supra-threshold responses (in the auditory field) across participants are related to modulation perception [[1](https://www.sciencedirect.com/science/article/pii/S0378595515000544)] and speech reception [[2](https://journals.lww.com/ear-hearing/Abstract/2020/05000/Neural_Modulation_Transmission_Is_a_Marker_for.12.aspx)].

larsoner:
> @rob-luke 's suggestion to cut parts of the bonus exercise would probably work. this part takes the longest, computation wise, since it a series of loops with psd estimation + snr extraction cylces for each single bar in the boxplots.

> also i think it wouldn't hurt, since there is some redundancy in there.. @rob-luke @larsoner : any opinion on what exactly to cut, and what to keep here?

I would say give it a try and see if it speeds things up

tutorials/time-freq/plot_ssvep.py
kalenkovich: I think it would work differently when there is not an exact frequency in `freqs`. If the frequencies are [..., 1.44, 1.66, 1.88, ...] and the stimulation frequency is 1.7 Hz, we would like to find the index of 1.66, not 1.88 as `searchsorted` would do.

agramfort: combine it with np.round?

tutorials/time-freq/plot_ssvep.py
dominikwelke: see https://github.com/mne-tools/mne-python/pull/8867/commits/54998daff64e1ef4d8a1cfb41289069280ae514f

tutorials/time-freq/plot_ssvep.py
dominikwelke: see https://github.com/mne-tools/mne-python/pull/8867/commits/54998daff64e1ef4d8a1cfb41289069280ae514f

tutorials/time-freq/plot_ssvep.py
dominikwelke: included in https://github.com/mne-tools/mne-python/pull/8867/commits/e0bbb2b609efc7f28224b86f3f6d125578184142

dominikwelke:
there were some new commits:

**re: ci runtime**
we have removed one of the sections towards the end that had become obsolete anyway. this should reduce runtime quite a bit..

**re: transient response**
@rob-luke  we now exclude the first second of each trial from the main analysis. 
this seems to be often done, also in VSSR.

here is a plot of the ERP from the roi channels. 
grafik
grafik
this is the 12hz case, but 15hz looks quite similar.. 
only 10 trials but there is a strong VEP at the beginning, and the 12hz modulation on top seems to be already there..

**re: switching to FFT**
we switched to using plain FFT. i did it as a special case of `psd_welch`.. i hope i got this correct, or is there some subtlety in MNE's implemenation that i missed?!

**adding paper references**
we also included some references to basic review papers. @rob-luke i usually used the picton 2003 paper as a basic review for assr. please let us know if there is a better one from your perspective!

**re: reference to F- and t2 tests**
>"This tutorial doesn't address determining the presence of a neural response, an F-test or Hotelling T² would be more appropriate for these purposes."
added.

**general structure**
please feel free to comment whether the general structure and flow makes sense to you.

tutorials/time-freq/plot_ssvep.py
dominikwelke: agreed, the info should go there..

please give me a hand on which files i have to change. i saw the final [doc page](https://mne.tools/dev/overview/datasets_index.html#datasets) where it should go, but wasnt able to reverse engineer the source of the texts :)

also i didnt implement the dataset version correctly, yet.
so far it defaults to 0.7, according to `_dataset_version(path, name)` in `utils.py`
how do you usually handle this?

larsoner: > please give me a hand on which files i have to change.

[doc/overview/datasets_index.rst](https://github.com/mne-tools/mne-python/blob/main/doc/overview/datasets_index.rst)

> how do you usually handle this?

If you can add a `version.txt` in the actual archive for the data, that's best. But for now if you've already created them and don't want to upload a new version you can add a conditional here:

https://github.com/mne-tools/mne-python/blob/main/mne/datasets/utils.py#L162-L165

dominikwelke: alright thanks, i moved the info to `dataset_index.rst` and added a version.txt to the data repository

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,695 @@
+"""
+==========================================================
```
larsoner, line 2:
```suggestion
.. _tut-ssvep:

==========================================================
```

See CircleCI:
```
/home/circleci/project/doc/overview/datasets_index.rst:428: WARNING: undefined label: tut-ssvep
```

larsoner (commented):
For the other CircleCI error:
```
/home/circleci/project/doc/overview/datasets_index.rst:416: WARNING: py:func reference target not found: mne.datasets.ssvep.data_path
```
You need to add your new function to `doc/datasets.rst`

tutorials/time-freq/plot_ssvep.py
kalenkovich: I am not sure how we could use rounding. We could round the stimulation frequency to the nearest whole multiple of the frequency step but this seems like an overly complicated way to look for the value we need. Do you see a simpler solution?

dominikwelke:
> You need to add your new function to `doc/datasets.rst`

this file doesnt exist in our branch yet. it's very new, right?
do we need to rebase to the dev status first to solve this, or is there another way?

larsoner:
You can locally merge with master and push. I can take care of it if you're unsure of the commands

larsoner:
Sorry, `main` not `master`

dominikwelke:
> You can locally merge with `main` and push. I can take care of it if you're unsure of the commands

thanks, i hope it worked..

larsoner:
The CIs will tell us :)

Also the linux 3.9 and `pip pre` Azure Windows job will both fail for reasons unrelated to your code, those are safe to ignore

agramfort:
rendrred links

https://25281-1301584-gh.circle-artifacts.com/0/dev/overview/datasets_index.html#ssvep-dataset
https://25281-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/time-freq/plot_ssvep.html

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+
+##############################################################################
+# Plot PSD and SNR spectra
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# Now we will plot grand average PSD (in blue) and SNR (in red) ± sd
+# for every frequency bin.
+# PSD is plotted on a log scale.
+#
+
+fig, axes = plt.subplots(2, 1, sharex='all', sharey='none', dpi=300)
```
agramfort, line 268:
```suggestion
fig, axes = plt.subplots(2, 1, sharex='all', sharey='none', figsize=(8, 5))
```
dpi is too different between displays.

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+rng = range(np.where(np.floor(freqs) == 1.)[0][0],
```
agramfort, line 269:
```suggestion
freq_range = range(np.where(np.floor(freqs) == 1.)[0][0],
```
rng is what we use everywhere for random_state (random number generator)

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+i_bin_36hz = np.argmin(abs(np.subtract(freqs, 36)))
+i_bin_15hz = np.argmin(abs(np.subtract(freqs, 15)))
+i_bin_30hz = np.argmin(abs(np.subtract(freqs, 30)))
+i_bin_45hz = np.argmin(abs(np.subtract(freqs, 45)))
+
+###############################################################################
+# Get indices for the different trial types
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+i_trial_12hz = np.where(epochs.events[:, 2] == event_id['12hz'])
```
agramfort, line 363:
```suggestion
i_trial_12hz = np.where(epochs.events[:, 2] == event_id['12hz'])[0]
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+i_trial_15hz = np.where(epochs.events[:, 2] == event_id['15hz'])
```
agramfort, line 364:
```suggestion
i_trial_15hz = np.where(epochs.events[:, 2] == event_id['15hz'])[0]
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+                               exclude='bads', selection=roi_vis)
+
+###############################################################################
+# Apply the subset, and check the result
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# Now we simply need to apply our selection and yield a result. Therefore,
+# we typically report grand average SNR over the subselection.
+#
+
+snrs_target = snrs[i_trial_12hz, :, i_bin_12hz][0][:, picks_roi_vis]
```
agramfort, line 385:
```suggestion
snrs_target = snrs[i_trial_12hz, :, i_bin_12hz][:, picks_roi_vis]
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+#
+# Let's check out, whether this was a good decision!
+#
+# Here we will plot average SNR for each channel location as a topoplot.
+# Then we will do a simple paired T-test to check, whether average SNRs over
+# the two sets of channels are significantly different.
+#
+
+# get average SNR at 12hz for ALL channels
+snrs_12hz = snrs[i_trial_12hz, :, i_bin_12hz][0]
```
agramfort, line 406:
```suggestion
snrs_12hz = snrs[i_trial_12hz, :, i_bin_12hz]
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+topo_pos_grave = []
+[topo_pos_grave.append(
+    montage.positions[ch][:2]) for ch in epochs.info['ch_names']]
+topo_pos_grave = np.array(topo_pos_grave)
+
+# plot SNR topography, eventually
+fig, ax = plt.subplots(1)
+mne.viz.plot_topomap(snrs_12hz_chaverage, topo_pos_grave, vmin=1., axes=ax)
+
+print("sub 2, 12hz trials, SNR at 12hz")
```
agramfort, line 425:
```suggestion
print("sub 2, 12 Hz trials, SNR at 12 Hz")
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+print("average SNR (all channels): %f" % snrs_12hz_chaverage.mean())
+print("average SNR (occipital ROI): %f" % snrs_target.mean())
+
+tstat_roi_vs_scalp = \
+    ttest_rel(snrs_target.mean(axis=1), snrs_12hz.mean(axis=1))
+print("12 hz SNR in occipital ROI is significantly larger than 12 hz SNR over "
```
agramfort, line 431:
```suggestion
print("12 Hz SNR in occipital ROI is significantly larger than 12 Hz SNR over "
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+# Statistical analysis
+# --------------------
+# Back from this little detour into open science, let's move on and
+# do the analyses we actually wanted to do:
+#
+# We will show that we can easily discriminate the brains responses in the
+# trials with different stimulation frequencies.
+#
+# In the frequency and SNR spectrum plot above, we had all trials mixed up.
+# Now we will extract 12 and 15 hz SNR as well as SNR of the 1st and 2nd
```
agramfort, line 465:
```suggestion
# Now we will extract 12 and 15 Hz SNR as well as SNR of the 1st and 2nd
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+labels = ['12hz trials', '15hz trials']
+x = np.arange(len(labels))  # the label locations
+width = 0.6  # the width of the bars
+res = dict()
+
+# loop to plot SNRs at stimulation frequencies and harmonics
+for i, f in enumerate(freq_plot):
+    # extract snrs
+    stim_12hz_tmp = \
+        snrs_roi[i_trial_12hz, np.argmin(abs(np.subtract(freqs, f)))][0]
```
agramfort, line 487: you don't need to use np.subtract everywhere. Just use `-` you work with array here

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+###############################################################################
+# As you can easily see there are striking differences.
+# Let's verify this using a series of two-tailed paired T-Tests.
+#
+
+# Compare 12 hz and 15 hz SNR in trials after averaging over channels
+
+tstat_12hz_trial_stim = \
+    ttest_rel(res['stim_12hz_snrs_12hz'], res['stim_12hz_snrs_15hz'])
+print("12 hz Trials: 12 hz SNR is significantly higher than 15 hz SNR"
```
agramfort, line 519: please check Hz vs hz everywhere

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+
+# get average SNR at 12hz for ALL channels
+snrs_12hz = snrs[i_trial_12hz, :, i_bin_12hz][0]
+snrs_12hz_chaverage = snrs_12hz.mean(axis=0)
+
+# create a standard montage
+montage = mne.channels.make_standard_montage('easycap-M1', head_size=0.095)
+
+# add xyz coordinates for all channels
+montage.positions = montage._get_ch_pos()
```
agramfort, line 413: you cannot use private functions / methods in a public tutorial

agramfort (commented):
a few comments

rob-luke:
Were the epochs actually 20 s and the stimulus frequencies exactly 12 and 15 Hz? For ASSRs its not recommended to use integer second epochs because the line noise adds constructively, and this then means that to ensure integer number of modulation cycles you end up with a stimulus frequency of something like 12.125 Hz. However, its common to then talk about the stimuli using the rounded number. The dataset is great either way, but I just wondered if the rounded numbers in the dataset description were correct or approximated?

And then on that note... I don't think the notch filter is required and only adds unnecessary complication to the tutorial. You don't analyse the content of the frequency bin later (because there is no stimulation at that rate), so the notch filter is just superfluous. I ran the analysis locally without it and it made no difference to the remaining results.

Also the legend in the final plot is confusing. Could you add a legend item for each trial or make the color scheme continuous?

tutorials/time-freq/plot_ssvep.py
dominikwelke: done

tutorials/time-freq/plot_ssvep.py
dominikwelke: I guess there is a non-private way to get xyz coordinates of channels for use with `viz.plot_topomap`, but i didnt find it. could you give me a hint?

larsoner: Can you avoid getting the positions yourself and instead use the `mask` parameter of `plot_topomap`?

dominikwelke: ok, i just realized that i can pass an `info` object as position argument.. much easier

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,698 @@
...
+        np.ones(noise_n_neighborfreqs),
+        np.zeros(2 * noise_skip_neighborfreqs + 1),
+        np.ones(noise_n_neighborfreqs)))
+    averaging_kernel /= averaging_kernel.sum()
+
+    # Calculate the mean of the neighboring frequencies by convolving with the
+    # averaging kernel.
+    mean_noise = np.apply_along_axis(
+        lambda psd_: np.convolve(psd_, averaging_kernel, mode='valid'),
+        axis=-1, arr=psd
```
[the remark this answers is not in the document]
dominikwelke: well, you can pass this function a 1d PSD array / a list (single channel, singel trial data) 
i guess this would create problems with `convolve2d`?

dominikwelke:
> Were the epochs actually 20 s and the stimulus frequencies exactly 12 and 15 Hz? For ASSRs its not recommended to use integer second epochs because the line noise adds constructively, and this then means that to ensure integer number of modulation cycles you end up with a stimulus frequency of something like 12.125 Hz. However, its common to then talk about the stimuli using the rounded number. The dataset is great either way, but I just wondered if the rounded numbers in the dataset description were correct or approximated?

we used exactly 12.0 and 15.0 Hz stimulation (as accurate as the display gets) and exactly 20s trial duration.
thanks for the hint, though, I'll keep it in mind for potential future collections :)
 
> And then on that note... I don't think the notch filter is required and only adds unnecessary complication to the tutorial. You don't analyse the content of the frequency bin later (because there is no stimulation at that rate), so the notch filter is just superfluous. I ran the analysis locally without it and it made no difference to the remaining results.
> 
> Similarly, the low pass filter is not required and removing it simplifies the tutorial and does not change results.

true, both were removed

> Also the legend in the final plot is confusing. Could you add a legend item for each trial or make the color scheme continuous?

changed to greyscale

tutorials/time-freq/plot_ssvep.py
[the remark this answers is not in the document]
kalenkovich: `convolve2d` won't work with 3D-arrays, unfortunately. We could use `scipy.ndimage.filters.convolve` or `scipy.ndimage.filters.correlate` but they don't have the `mode='valid'` option which further complicates the code.

larsoner:
So now in addition to pip pre we have https://github.com/conda-forge/pyface-feedstock/pull/27 causing problems on Linux and Windows conda builds. Looks like we should just rely on macos + linux pip + windows pip.

kalenkovich:
> So now in addition to pip pre we have [conda-forge/pyface-feedstock#27](https://github.com/conda-forge/pyface-feedstock/pull/27) causing problems on Linux and Windows conda builds. Looks like we should just rely on macos + linux pip + windows pip.

Sorry for being a bit slow: we should just ignore the failed Linux and Windows conda builds, right? In addition to

> Also the linux 3.9 and pip pre Azure Windows job will both fail for reasons unrelated to your code, those are safe to ignore

larsoner:
Yes for now, hopefully it'll be cleared up in the next couple of days!

larsoner:
@rob-luke @agramfort the latest rendering is here:

https://25353-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/time-freq/plot_ssvep.html

@dominikwelke can you add a conditional like https://github.com/mne-tools/mne-python/blob/main/tools/circleci_download.sh#L90-L92 ? It would get rid of the download inside the example rendering.

tutorials/time-freq/plot_ssvep.py
dominikwelke: this one is still open.. 
@larsoner I am not sure i entirely got your suggestion.. was it about reordering the sections, or about adding internal links, or both?

dominikwelke:
> @dominikwelke can you add a conditional like https://github.com/mne-tools/mne-python/blob/main/tools/circleci_download.sh#L90-L92 ? It would get rid of the download inside the example rendering.

the conditional alone didnt work.. the ci sais 'no such file'
i guess the dataset has to be listed somewhere else too. can you direct me to the spot?

tutorials/time-freq/plot_ssvep.py
larsoner: The output currently looks like:

Screenshot from 2021-02-24 12-57-27

Notice that you have a custom/non-standard OUTLINE: section in the middle column, *and* the standard Sphinx-generated outline on the right, which is not the same. The sphinx one is automatically generated from the second-level headings in the file (the only first-level heading being the title of the example).

A better way to structure this is to make it so that whatever outline you want to show is the one automatically generated by Sphinx based on the first-level headings -- i.e., change your first level headings to have the content you want. And where you currently have `OUTLINE:` in your RST you can just put:
```
.. contents:: local
   :depth: 1
```
And it should make a nice little TOC for you. Having one inline is still useful even with the "right TOC" because in this theme below a specific width the right bar disappears:

Screenshot from 2021-02-24 13-00-26

If this still isn't clear I can push a quick commit to do it

larsoner:
> the ci sais InputError: (...) no such file or directory

Fun! We've hit another release-related error, as PyQt5 5.15.3 came out 3 hours ago:

https://pypi.org/project/PyQt5/#history

I'll make a hotfix in `main` to pin us to !=5.15.3 and hopefully that fixes things for you

tutorials/time-freq/plot_ssvep.py
dominikwelke: adressed in 03ac0512c1344d72d0b2a7a77f187e1bb6e01187

hope this worked

larsoner:
> hope this worked

If you have all doc-building dependencies installed (e.g., with `pip install -r requirements_doc.txt`) and do:
```
$ cd doc
$ PATTERN=ssvep make html_dev-pattern
...
$ make show
```
it will build the docs plus your specific example. Locally when I do this on your branch and wait a few minutes for the built to complete I see:

Screenshot from 2021-02-24 15-49-15

So clearly I got the syntax wrong. :) Changing it to:
```
.. contents:: Outline
   :depth: 2
```
and re-running `PATTERN=ssvep make html_dev-pattern` produces:

Screenshot from 2021-02-24 15-53-24

dominikwelke:
> If you have all doc-building dependencies installed 

it was running but stopped working when the theme was changed (`no theme named 'pydata_sphinx_theme' found (missing theme.conf?)`) 
i'll update my environment
 
> So clearly I got the syntax wrong. :) Changing it to:
> 
> ```
> .. contents:: Outline
>    :depth: 2
> ```

great. i actually wondered whether this would be correct, but i blindly trusted you :)
changed

rob-luke:
> do you have a good reference for this @rob-luke ?

Don't bother unless you can find one easily. Youve changed it to 15.0 now, so its clear to the reader. Thanks for the extra info. Its a great addition to the tutorials.

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,677 @@
...
+data_path = mne.datasets.ssvep.data_path()
+bids_fname = data_path + '/sub-02/ses-01/eeg/sub-02_ses-01_task-ssvep_eeg.vhdr'
+
+raw = mne.io.read_raw_brainvision(bids_fname, preload=True, verbose=False)
+raw.info['line_freq'] = 50.
+
+# Set montage
+montage = mne.channels.make_standard_montage(
+    'easycap-M1',
+    head_size=0.095)  # head_size parameter default = 0.095
```
rob-luke, line 76: If this is just the default value can it be removed?
```suggestion
montage = mne.channels.make_standard_montage( 'easycap-M1)
```

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,677 @@
...
+# The function below does what we want.
+#
+
+def snr_spectrum(psd, noise_n_neighbor_freqs=1, noise_skip_neighbor_freqs=1):
+    """Compute SNR spectrum from PSD spectrum using convolution.
+
+    Parameters
+    ----------
+    psd : ndarray, shape ([n_trials, n_channels,] n_frequency_bins)
+        Data object containing PSD values. Works with arrays as spit out by
```
rob-luke, line 177:
```suggestion
        Data object containing PSD values. Works with arrays as produced by
```

agramfort (approved):
Ok to merge when CIs are green (except the ones already red in main)

Thanks a lot @dominikwelke !

dominikwelke:
ok, all CIs are green @agramfort @larsoner @rob-luke @kalenkovich 

il push a last commit with the changelog

doc/changes/latest.inc
```diff
@@ -90,6 +92,8 @@ Enhancements
 
 - :func:`mne.time_frequency.EpochsTFR.average` now allows different ways of averaging, such as "median", or callable functions (:gh:`8879` by `Adam Li`_)
 
+- New Tutorial for analyzing frequency-tagging data (:gh: `8867` **by new contributor** |Dominik Welke|_ and `kalenkovich`_)
```
larsoner, line 95: Can you move this to the top of the `Enhancements` section? We list the changes by new contributors at the top of each section
```suggestion
- New Tutorial for analyzing frequency-tagging data (:gh:`8867` **by new contributor** |Dominik Welke|_ and `kalenkovich`_)
```
Also the space after the `:gh:` is causing problems for CircleCI

tutorials/time-freq/plot_ssvep.py
```diff
@@ -0,0 +1,675 @@
...
+We will visualize both the power-spectral density (PSD) and the SNR
+spectrum of the epoched data,
+extract SNR at stimulation frequency,
+plot the topography of the response,
+and statistically separate 12 Hz and 15 Hz responses in the different trials.
+Since the evoked response is mainly generated in early visual areas of the
+brain the statistical analysis will be carried out on an occipital
+ROI.
+
+.. contents:: local
```
larsoner, line 38:
```suggestion
.. contents:: Outline
```

larsoner (commented):
Other than two minor remaining issues, LGTM +1 for merge!

https://25558-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/time-freq/plot_ssvep.html

kalenkovich:
> Other than two minor remaining issues, LGTM +1 for merge!
> 
> https://25558-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/time-freq/plot_ssvep.html

Committed your suggestions.

agramfort:
@dominikwelke can you merge main branch here to make sure CIs are green?

dominikwelke:
> @dominikwelke can you merge main branch here to make sure CIs are green?

@agramfort  all CIs are green now

agramfort:
thx a lot @dominikwelke and @kalenkovich !

PR #8870, opened 2021-02-13
MRG: selection.read_selection() ➡️ channels.read_vectorview_selection()

### LINKED ISSUE #8865: mne.read_selection() returning funny channels
#### Describe the bug

trying to select channels from an EEG dataset.

#### Steps to reproduce

```
l = ['Vertex', 'temporal', 'parietal', 'occipital', 'frontal']

ll = mne.read_selection(name=l, info=r.info)
```
returns a list with **Neuromag** channels although there are only EEG channels (EGI) in the dataset.

@hoechenberger

hoechenberger:
Fixes #8865

@larsoner There is a bug in `deprecated_alias`:
```python
/Users/hoechenberger/Development/mne-python/mne/utils/docs.py:2551: DeprecationWarning: Function 
read_neuromag_selection is deprecated; read_selection has been deprecated in favor of
read_neuromag_selection and will be removed in 0.24
```
In the sentence before the semicolon, it refers to the new function name, not to the deprecated one.

#### Additional information
I moved the API section of the changelog up above the bugs section, and added a new section
"Deprecations" directly below, now placing "Bugs" last. I think this is better for most users as the
number of API changes and deprecations is typically small, compared to the endless number
of bugs we fix. If API changes & deprecations are listed after the bugs, I think they're easy to miss.
Of course, totally up for discussion.


agramfort:
maybe vectorview is a better name than neuromag ?

hoechenberger:
Sure, why not!

hoechenberger:
Closed & reopened to restart CI.

doc/changes/latest.inc
```diff
@@ -76,6 +76,16 @@ Enhancements
 
 - Static type checkers like Pylance (comes with VS Code) now display the parameters of many more functions correctly, largely improving overall usability for VS Code users (:gh:`8862` by `Richard Höchenberger`_)
 
+- Add ``match_alias`` parameter to :meth:`mne.io.Raw.set_montage` and related functions to match unrecognized channel location names to known aliases (:gh`8799` **by new contributor** |Zhi Zhang|_)
```
larsoner, line 79: New contributor line needs to be at the top of the section

doc/changes/latest.inc
```diff
@@ -76,6 +76,16 @@ Enhancements
...
+
+API changes
+~~~~~~~~~~~
+- None
+
+Deprecations
```
larsoner, line 85: Did we decide to make a new section called `Deprecations`? Usually this is what `API changes` is for I think -- ENH/BUG/API has been the standard section naming and order for all our releases. Let's not reorganize/rename/reorder in this PR, but if this important to you we can discuss and then redo all of the lists.

mne/tests/test_read_neuromag_selection.py
```diff
@@ -22,30 +22,30 @@ def test_read_selection():
...
 
     ch_names_new = [ch.replace(' ', '') for ch in ch_names]
     raw_new = read_raw_fif(raw_new_fname)
     for i, name in enumerate(sel_names):
-        sel = read_selection(name, info=raw_new.info)
+        sel = read_vectorview_selection(name, info=raw_new.info)
         assert ch_names_new[i] in sel
 
-    pytest.raises(TypeError, read_selection, name, info='foo')
+    pytest.raises(TypeError, read_vectorview_selection, name, info='foo')
```
larsoner, line 48: if you want to go the extra mile, I usually modernize `pytest.raises` calls in tests that I'm updating/changing. That way someday in the distant future all of them will be updated

mne/channels/channels.py
```diff
@@ -1762,3 +1766,180 @@ def combine_channels(inst, groups, method='mean', keep_stim=False,
...
+
+        * ``'Vertex'``
+        * ``'Left-temporal'``
+        * ``'Right-temporal'``
+        * ``'Left-parietal'``
+        * ``'Right-parietal'``
+        * ``'Left-occipital'``
+        * ``'Right-occipital'``
+        * ``'Left-frontal'``
+        * ``'Right-frontal'``
```
larsoner, line 1871: All of this seems more appropriate for a `Notes` section (I know you copy-pasted, but since all lines are new/changed it's not a bad time to do it anyway)

drammock: note also that I think these bullet points are indented too far, they will end up rendered as if they were bullets within a blockquote (as often happens in our docs)

doc/changes/latest.inc
agramfort: Yes we only have API section

mne/channels/channels.py
hoechenberger: > All of this seems more appropriate for a `Notes` section

Personally, I think this is a bit of an anti-pattern: without knowing the names of the selections, users cannot use the function. Forcing them to scroll down to the Notes section doesn't seem very user-friendly, so I'd rather leave it here.



> note also that I think these bullet points are indented too far

Thanks, I will have a look.

doc/changes/latest.inc
hoechenberger: > Did we decide to make a new section called `Deprecations`?

Nope, this is a new proposal here :)

> Let's not reorganize/rename/reorder in this PR, but if this important to you we can discuss and then redo all of the lists in a separate PR.

Will revert this here and make a new proposal in a separate PR

mne/channels/channels.py
larsoner: > Personally, I think this is a bit of an anti-pattern: without knowing the names of the selections, users cannot use the function. Forcing them to scroll down to the Notes section doesn't seem very user-friendly, so I'd rather leave it here.

I still think these should be moved out of the top, for the following reasons:

- The top should be reserved for a hopefully short explanation of the function. That way the parameter list at the top is close to the definition list below it and people don't have to scroll a ton (or back and forth) to see them.
- The argument that some critical parameters should have their options described at the top does not generalize well. Imagine applying this logic to something like `maxwell_filter` where you arguably need to know about what options are available for half a dozen arguments and what those options mean to effectively use it.

To me `Notes` is a reasonable place for it, but another option would be to move them to a non-bulleted list in the description of the parameter `name`. This fits with the general pattern that parameter options should be described (compactly) within the parameter definition list itself, or in Notes for details / if there are too many options.

hoechenberger: > To me `Notes` is a reasonable place for it, but another option would be to move them to a non-bulleted list in the description of the parameter `name`. This fits with the general pattern that parameter options should be described (compactly) within the parameter definition list itself, or in Notes for details / if there are too many options.

I like the idea of moving it to the parameter description. To me, Notes is more like implementation details a normal user shouldn't have to worry about

hoechenberger: I've addressed this in f470d86

hoechenberger:
I believe all comments have been addressed.

mne/channels/channels.py
```diff
@@ -1762,3 +1766,172 @@ def combine_channels(inst, groups, method='mean', keep_stim=False,
...
+    By default, the selections used in ``mne_browse_raw`` are supported.
+    Additional selections can be added by specifying a selection file (e.g.
+    produced using ``mne_browse_raw``) using the ``fname`` parameter.
+
+    The ``name`` parameter can be a string or a list of string. The returned
+    selection will be the combination of all selections in the file where
+    (at least) one element in name is a substring of the selection name in
+    the file. For example, ``name=['temporal', 'Right-frontal']`` will produce
+    a combination of ``'Left-temporal'``, ``'Right-temporal'``, and
+    ``'Right-frontal'``.
```
larsoner, line 1859: This belongs with the parameter description or in Notes

mne/channels/channels.py
```diff
@@ -1762,3 +1766,172 @@ def combine_channels(inst, groups, method='mean', keep_stim=False,
...
+                        % (type(info),))
+    else:  # info is None
+        spacing = 'old'
+
+    # use built-in selections by default
+    if fname is None:
+        fname = op.join(op.dirname(__file__), '..', 'data', 'mne_analyze.sel')
+
+    if not op.isfile(fname):
+        raise ValueError('The file %s does not exist.' % fname)
```
larsoner, line 1902: If you want to save coverage here you could use
```suggestion
    fname = _check_fname(fname, must_exist=True, overwrite='read')
```

agramfort (approved):

doc/changes/latest.inc
```diff
@@ -78,6 +80,10 @@ Enhancements
 
 - Support new EEGLAB file format (:gh:`8874` by `Clemens Brunner`_)
 
+API changes
+~~~~~~~~~~~
+- ``mne.read_selection`` has been deprecated in favor of `mne.read_vectorview_selection`. ``mne.read_selection`` will be removed in MNE-Python 0.24 (:gh:`8870` by `Richard Höchenberger`_)
+
```
larsoner, line 86: Still not quire right, see previous comment:

> ENH/BUG/API has been the standard section naming and order for all our releases. Let's not reorganize/rename/reorder in this PR

larsoner: Oh wait I think it's okay, I must have been looking at an old commit!

larsoner, line 86: Actually no it's still in the wrong spot

hoechenberger: OUCH! Arrgh, sorry about that, will fix.

hoechenberger: Should be good now.

larsoner:
thanks @hoechenberger !

PR #8874, opened 2021-02-16
Fix EEGLAB import

### LINKED ISSUE #8864: ValueError: Could not find EEG array in the .set file.
#### Describe the bug
Dear MNE developers
Thank you for developing this wonderful tool. 

I am faced with the error when I tried to use read_raw_eeglab() to load the .set file made by EEGLAB
What I did in EEGLAB is just loading the data by Quick 30 and save it in '.set(with '.fdt')' format only. 
I think this is very weird, so I guess this is bug.

The script I made is below.
```
subject_name = 'DA'
script_path = os.path.abspath('./Python/EEG_analysis')
data_path = os.path.abspath('./output_play/' + subject_name) 
os.path.exists(data_path + '/'+ subject_name + '.set') # The print out result is 'True'
raw = mne.io.read_raw_eeglab(input_fname = data_path + '/'+ subject_name + '.set')
```
#### Actual results
```
ValueError                                Traceback (most recent call last)
<ipython-input-17-1519c0d84577> in <module>
----> 1 raw = mne.io.read_epochs_eeglab(input_fname = data_path + '/'+ subject_name + '_play_merged.set')

~/.pyenv/versions/anaconda3-5.3.1/lib/python3.7/site-packages/mne/io/eeglab/eeglab.py in read_epochs_eeglab(input_fname, events, event_id, eog, verbose, uint16_codec)
    277     epochs = EpochsEEGLAB(input_fname=input_fname, events=events, eog=eog,
    278                           event_id=event_id, verbose=verbose,
--> 279                           uint16_codec=uint16_codec)
    280     return epochs
    281 

<decorator-gen-198> in __init__(self, input_fname, events, event_id, tmin, baseline, reject, flat, reject_tmin, reject_tmax, montage, eog, verbose, uint16_codec)

~/.pyenv/versions/anaconda3-5.3.1/lib/python3.7/site-packages/mne/io/eeglab/eeglab.py in __init__(self, input_fname, events, event_id, tmin, baseline, reject, flat, reject_tmin, reject_tmax, montage, eog, verbose, uint16_codec)
    449                  uint16_codec=None):  # noqa: D102
    450         eeg = _check_load_mat(input_fname, uint16_codec)
--> 451 
    452         if not ((events is None and event_id is None) or
    453                 (events is not None and event_id is not None)):

~/.pyenv/versions/anaconda3-5.3.1/lib/python3.7/site-packages/mne/io/eeglab/eeglab.py in _check_load_mat(fname, uint16_codec)
     63             'mne-python developers for more information.')
     64     if 'EEG' not in eeg:
---> 65         raise ValueError('Could not find EEG array in the .set file.')
     66     else:
     67         eeg = eeg['EEG']

ValueError: Could not find EEG array in the .set file.
```

#### Additional information
I am using the latest version of EEGLAB(2021), so there might be some minor change in the .set file. This might be why I am faced with this error.

Thank you in advance.

cbrnr:
EEGLAB files may contain all fields in the root of the file instead of within an `EEG` struct. This PR adds support for this format. Fixes #8864.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -76,6 +76,8 @@ Enhancements
 
 - Static type checkers like Pylance (comes with VS Code) now display the parameters of many more functions correctly, largely improving overall usability for VS Code users (:gh:`8862` by `Richard Höchenberger`_)
 
+- Support new EEGLAB file format (:gh:`8874` by `Clemens Brunner`_)
+
 Bugs
 ~~~~
 - Fix bug with `mne.connectivity.spectral_connectivity` where time axis in Epochs data object was dropped. (:gh:`8839` **by new contributor** |Anna Padee|_)
@@ -103,7 +105,7 @@ Bugs
 - Fix bug with :func:`mne.stats.permutation_cluster_1samp_test` and related clustering functions when ``adjacency=None`` and ``out_type='indices'`` (:gh:`#8842` by `Eric Larson`_)
 
 - Fix bug with :func:`mne.viz.plot_alignment` where plotting a sphere model could ignore the ``brain`` argument (:gh:`8857` by `Eric Larson`_)
-- 
+-
 - Fix bug with ``replace`` argument of :meth:`mne.Report.add_bem_to_section` and :meth:`mne.Report.add_slider_to_section` (:gh:`8723` by `Eric Larson`_)
 
 - Fix compatibility bugs with :mod:`mne_realtime` (:gh:`8845` by `Eric Larson`_)
--- a/mne/datasets/utils.py
+++ b/mne/datasets/utils.py
@@ -251,7 +251,7 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     path = _get_path(path, key, name)
     # To update the testing or misc dataset, push commits, then make a new
     # release on GitHub. Then update the "releases" variable:
-    releases = dict(testing='0.115', misc='0.8')
+    releases = dict(testing='0.116', misc='0.8')
     # And also update the "md5_hashes['testing']" variable below.
     # To update any other dataset, update the data archive itself (upload
     # an updated version) and update the md5 hash.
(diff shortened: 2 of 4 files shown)


doc/changes/latest.inc
```diff
@@ -103,7 +105,7 @@ Bugs
 - Fix bug with :func:`mne.stats.permutation_cluster_1samp_test` and related clustering functions when ``adjacency=None`` and ``out_type='indices'`` (:gh:`#8842` by `Eric Larson`_)
 
 - Fix bug with :func:`mne.viz.plot_alignment` where plotting a sphere model could ignore the ``brain`` argument (:gh:`8857` by `Eric Larson`_)
-- 
+-
```
hoechenberger, line 108:
```suggestion

```

hoechenberger:
@cbrnr Do you think you could add a test?

cbrnr:
Do we have a small test file or can we use/modify an existing one?

hoechenberger:
> Do we have a small test file or can we use/modify an existing one?

Could you maybe modify a file "on the fly" during a test such that it doesn't have the `EEG` struct (i.e., copy the contents of the `EEG` struct to the "root" of the HDF5 file), and simply try to read the metadata / info (i.e., do not perform a preload)? Would this work?

cbrnr:
Yes, but given that this will be the new default format we should probably have a real file that we can use. Maybe someone with access to EEGLAB can save one of our existing files (e.g. `test_raw.set`) in the new format?

hoechenberger:
@cbrnr btw if you pull in / rebase onto `main`, CI should now pass thanks to #8875

cbrnr:
@hoechenberger I've created a small test file based on `test_raw.set` - if this gets merged I can add a test.

cbrnr:
Preloading doesn't work as expected with the new format (that's why I can't test it). Previously, we supported `preload=False` only for the combined .set/.fdt format (I think). Now that fields are not contained inside `EEG`, it might be possible to load everything but `data` though with the new one-file format. It looks like `pymatreader` does support a `variable_names` argument, so we could implement `preload=False` if we knew which variables are required/mandatory (there is no option to load all variables except `data` unfortunately). I'd say this is something for another PR.

hoechenberger (approved):

hoechenberger:
@cbrnr flake is still complaining; other than that, LGTM

hoechenberger:
Thanks @cbrnr!

Can you backport this?

cbrnr:
Yes, will do!

cbrnr:
@hoechenberger done! For some reason, I had to make a separate commit for the changelog entry. Can you please check if everything is OK (and if so add a "backported" label here)?

hoechenberger:
It seems CI failed on `maint/0.22`, I just restarted Circle

cbrnr:
The only thing that looks like it could be an error is that I forgot to update the hash of `misc`. @hoechenberger are you looking into this?

hoechenberger:
> The only thing that looks like it could be an error is that I forgot to update the hash of `misc`. @hoechenberger are you looking into this?

Nope I'm currently busy with something else, and yes it appears to be something related to the hash! :)

cbrnr:
Alright, I'll fix it then!

PR #8884, opened 2021-02-17
MRG, BUG: Fix verbose

### LINKED ISSUE #8872: Verbose not working with the Epochs.drop_bad method
#### Describe the bug
The `verbose` argument in not working in the [`drop_bad()`](https://mne.tools/dev/generated/mne.Epochs.html?highlight=drop_bad#mne.Epochs.drop_bad) method.

#### Steps to reproduce
```python
import os
import mne


                                    'sample_audvis_raw.fif')
epochs = mne.Epochs(raw, events, tmin=-0.3, tmax=0.7)

epochs.verbose = False

epochs.drop_bad(verbose=True)
```

#### Expected results
The `verbose` argument set to `True` overwrites the `Epochs.verbose` attribute and prints the number of bad epochs dropped.
```
86 events found
Event IDs: [ 1  2  3  4  5 32]
Not setting metadata
Not setting metadata
86 matching events found
Created an SSP operator (subspace dimension = 3)
3 projection items activated
Loading data for 86 events and 601 original time points ...
1 bad epochs dropped
```

#### Actual results
The number of bad epochs dropped is not printed.
```
86 events found
Event IDs: [ 1  2  3  4  5 32]
Not setting metadata
Not setting metadata
86 matching events found
Created an SSP operator (subspace dimension = 3)
3 projection items activated
```

#### Additional information
It seems that in the source code of [`drop_bad()`](https://github.com/mne-tools/mne-python/blob/main/mne/epochs.py#L1184-L1224) the argument `verbose` which defaults to `None` is not used.

larsoner:
Closes #8872

@HanBnrd can you try it with your code?

Basically the fix is that for any methods that have `verbose`, we need to propagate `verbose` to any other method calls. We don't need to do this with functions, but we do with methods since when they are called with `verbose is None` they resort to using `self.verbose`.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -96,6 +96,8 @@ Bugs
 
 - Fix bug with :meth:`raw.plot() <mne.io.Raw.plot>` where ``scalings='auto'`` did not compute scalings using the full range of data (:gh:`8806` by `Eric Larson`_)
 
+- Fix bug with :meth:`mne.io.Raw.load_data` and :meth:`mne.Epochs.drop_bad` where ``verbose`` logging was not handled properly (:gh:`8884` by `Eric Larson`_)
+
 - Fix bug with :func:`mne.io.read_raw_nicolet` where header type values such as num_sample and duration_in_sec where not parsed properly (:gh:`8712` by `Alex Gramfort`_)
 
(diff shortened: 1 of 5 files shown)


HanBnrd:
Well done!

Just tried it on `Raw.load_data()` and `Epochs.drop_bad()`, `verbose` now works with your fix.

Thanks!

agramfort (approved):

PR #8888, opened 2021-02-19
Test abstract BaseEpochs class

### LINKED ISSUE #8871: Improve Epochs docs (and more)
Based on my question in the [forum](https://mne.discourse.group/t/computing-tfr-drops-bad-epochs/2691), I suggest to clarify the `preload` parameter to make it clear that preloading does *not* depend on the status of `raw.preload` (which I assumed was the case).

I'd change the `preload` description as follows:

> Load all epochs ~~from disk~~ into memory when creating the object or wait before accessing each epoch (more memory efficient but can be slower).

In addition, I'd add the following note:

> Epochs are preloaded based on the value of the `preload` argument. Even if the `raw` argument has already been preloaded, the returned `Epochs` object will **not** be preloaded by default. This means that bad epochs (e.g. epochs with not enough data) will not be dropped immediately, so the number of good epochs displayed may change once the epochs are preloaded.

@hoechenberger @larsoner @agramfort @drammock since this is such a widely used object I'd be grateful if you commented and edited my suggestion if necessary.

Furthermore, I noticed that `preload` is used inconsistently across `BaseEpochs`, `Epochs`, `EpochsArray`, and `EpochsFIF` classes:

- `BaseEpochs`: Doesn't have a `preload` parameter but `preload_at_end` - what's the reason to use a different parameter?
- `Epochs`: `preload=False`
- `EpochsArray`: Has neither `preload` nor `preload_at_end` parameters.
- `EpochsFIF`: `preload=True`

I think it would be great to double-check these inconsistencies - there are probably good reasons, but if not we should discuss changing accordingly.

Finally, @hoechenberger suggested to discuss `preload=None` as a default, meaning that this would re-use its `raw.preload` value.

cbrnr:
Fixes #8871.

One benefit of declaring `BaseEpochs` abstract is that the following code won't work anymore:

```python
import numpy as np
import mne


fs, n_chan = 256, 4
info = mne.create_info(n_chan, fs, "eeg")
data = np.random.rand(3, n_chan, 2 * fs + 1)
events = np.array([[100, 0, 1], [1000, 0, 1], [1200, 0, 2]])
epochs = mne.BaseEpochs(info, data, events, tmin=-1, tmax=1)
```

However, we'll have to wait and see if this breaks anything.

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -2826,7 +2826,7 @@ class EpochsFIF(BaseEpochs):
             # we also retain original baseline without re-applying baseline
             # correction (data is being baseline-corrected when written to
             # disk)
-            epoch = BaseEpochs(
+            epoch = super().__init__(
                 info, data, events, event_id, tmin, tmax,
                 baseline=None,
                 metadata=metadata, on_missing='ignore',
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -11,6 +11,7 @@
 #
 # License: BSD (3-clause)
 
+from abc import ABC, abstractmethod
 from collections import Counter
 from copy import deepcopy
 import json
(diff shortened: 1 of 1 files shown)


cbrnr:
Apparently, we're instantiating `BaseEpoch` in several places, e.g.https://github.com/mne-tools/mne-python/blob/14182158075792f3c0f5dc901c327aa2af7feaf3/mne/epochs.py#L2829. These must be replaced by `super().__init__()` calls. I don't know. Is it worth pursuing?

cbrnr:
OK, I've decided it's not worth it.

cbrnr:
Such a shame that this PR didn't live up to its unique number 😢...

hoechenberger:
Let's try to make #9999 special 🤗

PR #8891, opened 2021-02-20
Fix: Wrong channel-adjacency-matrix for Neuromag122

### LINKED ISSUE #8890: BUG: Channel-Adjacency for Neuromag122
#### Describe the bug
When I try to get the channel-adjacency-matrix for Gradiometer from **Neuromag122**-Data with `find_ch_adjacency`
, a matrix with dimensions 306x306 is given (when read from ./mne/channels/data/neighbors/neuromag122_neighb.mat).

And when I try to compute the adjacency with `mne.channels.channels._compute_ch_adjacency`, I get the following result:

```
File "C:\Users\marti\anaconda3\envs\mne\lib\site-packages\mne\channels\channels.py", line 1426, in _compute_ch_adjacency
xy = _find_topomap_coords(info, picks, sphere=HEAD_SIZE_DEFAULT)
File "C:\Users\marti\anaconda3\envs\mne\lib\site-packages\mne\channels\layout.py", line 624, in _find_topomap_coords
pos = _auto_topomap_coords(
File "C:\Users\marti\anaconda3\envs\mne\lib\site-packages\mne\channels\layout.py", line 723, in _auto_topomap_coords
raise ValueError('The following electrodes have overlapping positions,'
ValueError: The following electrodes have overlapping positions, which causes problems during visualization:
MEG 001, MEG 002, MEG 003, [...]
```

#### Steps to reproduce
```
import mne
data_path = "./neuromag122_test_file-raw.fif"
info = mne.io.read_info(data_path)
# Trying the load the channel-adjacency
ch_adjacency = mne.channels.find_ch_adjacency(info, 'grad')
print(ch_adjacency[0].shape)
# Trying to bypass mne.channels.read_ch_adjacency()
mne.channels.channels._compute_ch_adjacency(info, 'grad')
```

#### Expected results
A 122x122 Matrix

#### Actual results
A 306x306 Matrix and a Value Error

#### Additional information
I uploaded an [anonymized raw-file](https://github.com/mne-tools/mne-python/files/6016694/neuromag122_test_file-raw.zip) from our Neuromag122 for reproduction of the issue.

marsipu:
#### Reference issue
Fixes #8890.

This disables reading the channel-adjacency for Neuromag122-Data from a precomputed file in .../channels/data/neighbors and fixes the computation of the matrix.


marsipu:
I don't understand why some checks fail.
How do I know, if they are relevant?

marsipu:
Do you think, this is sufficient as a solution?

larsoner:
> How do I know, if they are relevant?

In this case those two builds can be ignored, the failure is unrelated and @drammock plans to look into it soon.

> Do you think, this is sufficient as a solution?

Maybe we should wait a couple of days to hear from the upstream issue https://github.com/fieldtrip/fieldtrip/issues/1676 , and then merge this if we can't get an updated file, WDYT?

marsipu:
Yes sure, that would interest me too (maybe there are several versions of Neuromag122?). Thank you for looking into it

larsoner:
... as mentioned in the issue, let's get rid of the `.mat` file and the associated lines that are currently commented out. That plus a `doc/changes/latest.inc` and we should be good to go here!

marsipu:
Ok:)

larsoner:
@marsipu I realized that we are missing some sort of test. This is a little bit tricky because we don't have a Neuromag 122 test file. Do you have or could you create some tiny raw (you could crop it to 1 sec and save it in MNE) file that we could [add to the testing dataset](https://github.com/mne-tools/mne-testing-data#add-or-change-files-in-the-repo-for-use-with-mne)? I would probably put it in [misc](https://github.com/mne-tools/mne-testing-data/tree/master/misc). Then we could for example add a tiny test that connectivity can be computed and the result is a sparse array of shape `(122, 122)`.

marsipu:
I'll do it soon

marsipu:
Oh, I rebased the branch but I didn't think about the commits then being part of the pull request. Is this a problem, should I revert the new commits from main?

larsoner:
> Oh, I rebased the branch but I didn't think about the commits then being part of the pull request. Is this a problem, should I revert the new commits from main?

Hmmm, if you rebase properly then the commits from `main` should not even show up in the PR. Moreover, they should come before your commits instead of after them. For example one commit that shows up in this PR now:

cdd332b9cf89867c3b7e37fd7aefc8072ccf250c

Is actually a different commit actually in `main`:

0de802aab6d858132e92c4dd74534ca0299f960f

How did you do the rebase?

One option is that, if your latest commit is actually meant to be b2ef4a0, you can do `git reset --hard b2ef4a0`. Then something like this should work:
```
git fetch upstream
git rebase upstream/main
git push origin -f neuromag122_adjacency  # WARNING: Will force push
```

marsipu:
Thank you for your help, I started the rebase in a terminal with
```
git fetch upstream
git rebase upstream/main
```
but then pushed the changes with PyCharm (because of the GitHub credentials I already have there). Maybe that combination was the problem.

marsipu:
I added the test-file to mne-testing and updated the version in mne, just the hash-update is missing.

marsipu:
I added the test-file to mne-testing and updated the version in mne, just the hash-update is missing.

marsipu:
> Then we could for example add a tiny test that connectivity can be computed and the result is a sparse array of shape `(122, 122)`.
I added a first attempt of a test, not sure if it is at the right spot (I haven't acquired a lot of experience with testing yet)

mne/channels/tests/test_channels.py
```diff
@@ -344,6 +344,17 @@ def test_find_ch_adjacency():
...
 
+@testing.requires_testing_data
+def test_neuromag122_adjacency():
+    """Test computing the adjacency matrix of Neuromag122-Data"""
+    nm122_fname = op.join(testing.data_path(), 'misc',
+                          'neuromag122_test_file-raw.fif')
+    raw = read_raw_fif(nm122_fname, preload=True)
+    conn, ch_names = find_ch_adjacency(raw.info, 'grad')
+    assert_equal(conn.getnnz(), 1564)
+    assert_equal(len(ch_names), 122)
```
larsoner, line 355: We can use plain `assert` and pytest does nice things with it. We tend to reserve `assert_*` functions for array comparisons nowadays.
```suggestion
    assert conn.getnnz() == 1564
    assert len(ch_names) == 122
    assert conn.shape == (122, 122)
```

mne/datasets/utils.py
```diff
@@ -251,7 +251,7 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     path = _get_path(path, key, name)
     # To update the testing or misc dataset, push commits, then make a new
     # release on GitHub. Then update the "releases" variable:
-    releases = dict(testing='0.116', misc='0.8')
+    releases = dict(testing='0.117', misc='0.8')
```
larsoner, line 254: you also need to update the hash below

larsoner (commented):
Looks like it should work after a hash update!

mne/datasets/utils.py
marsipu: Oh I thought I had already updated the release. Thanks

marsipu:
Ok, I don't know why it was not changed in the conversation, but I already changed the testing-version in 2e2f76c before. so I marked the conversation as resolved

mne/channels/tests/test_channels.py
```diff
@@ -344,6 +344,18 @@ def test_find_ch_adjacency():
     assert ch_names[0] == 'MEG 001'
 
 
+@testing.requires_testing_data
+def test_neuromag122_adjacency():
+    """Test computing the adjacency matrix of Neuromag122-Data"""
```
larsoner, line 349: pydocstyle is unhappy
```suggestion
    """Test computing the adjacency matrix of Neuromag122-Data."""
```

marsipu:
I rebased the branch onto upstream/main again to resolve the conflict in latest.inc.
Is there something else I can do?

larsoner:
Thanks @marsipu !

Issue #8880, opened 2021-02-16
Implementation of stft() function as stft_array and tfr_stft


adam2392:
#### Describe the new feature or enhancement
Currently `stft` operates on arrays, but is not the same API as wavelet, stockwell and multitaper. In addition, it does not have a `tfr_stft` analog , which operates on mne-python data structures.

The proposed PR would refactor `stft` to be named `stft_array` and have similar API to the `tfr_array` functions, and also implement an analogous `tfr_stft` function.

See discourse discussion: https://mne.discourse.group/t/is-there-a-way-to-obtain-the-time-points-from-time-frequency-stft/2682

The `stft` is a nice function since it is super fast, and can operate on `Raw` object data within seconds. 

#### Describe your proposed implementation
Tbd

#### Additional comments
related: https://github.com/mne-tools/mne-python/pull/8781. If this is accepted, then one can use `tfr_stft` to compute spectrograms of the data for certain frequency bands (e.g. gamma power shown as channels X time over a whole Raw dataset).

larsoner:
> The proposed PR would refactor stft to be named stft_array and have similar API to the tfr_array functions, and also implement an analogous tfr_stft function.

I would propose instead that we deprecate `stft` (have it call `_stft` that has all the current `stft` code in it) and make a new `stft_array` that has the right API (but calls `_stft` under the hood). I don't think we ever really meant for `stft` to be public, right @agramfort ?

adam2392:
And should I still include `tfr_stft`?

larsoner:
Yeah that would make sense to fit with the existing API

agramfort:
ok to move replace stft to _stft and create new public functions tfr_stft
and tfr_array_stft that just calls _stft?
that's what you had in mind?

>

PR #8892, opened 2021-02-22
[WIP] Adding functionality for tfr_array_stft and tfr_stft

adam2392:
#### Reference issue
Closes: #8880 

Need to implement unit tests.

- Implements `tfr_array_stft` and `tfr_stft` which operate on numpy arrays and mne-python objects respectively.

- Deprecates `stft` in favor of `tfr_array_stft`.

#### Additional information
No parallelizations across channels because felt that could be saved for a later PR.

### DIFF
--- a/mne/time_frequency/_stft.py
+++ b/mne/time_frequency/_stft.py
@@ -52,6 +52,7 @@ def tfr_stft(inst, window_size, step_size=None, average=True, verbose=None):
     if average:
         # how many epochs are we averaging
         nave = len(data)
+        power = np.mean(power, axis=0)
         out = AverageTFR(info, power, times, freqs, nave, method='stft-power')
     else:
         if isinstance(inst, BaseEpochs):
@@ -92,7 +93,7 @@ def tfr_array_stft(data, sfreq, window_size, step_size=None,
 
     Returns
     -------
-    X : array, shape (n_channels, wsize // 2 + 1, n_step)
+    X : array, shape (n_epochs, n_channels, wsize // 2 + 1, n_step)
         STFT coefficients for positive frequencies with
         ``n_step = ceil(T / tstep)``.
     freqs : array, shape (wsize // 2 + 1,)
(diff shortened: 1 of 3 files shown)


agramfort:
thinking about it we have an STFT from the TF-MxNE solver. This stft implementation uses
a certain convention for edge which is ok for TF-MxNE but maybe non-standard. I would
maybe suggest to use the scipy implementation these new public functions and then make the
current stft function private and just used Tf-MxNE.

what do you think @larsoner ?

larsoner:
I think eventually we want to replace the TF-MxNE version anyway, so I don't think there is a big difference. One advantage of our version is that it preserves power properly, whereas the SciPy version doesn't (it does on round trip, just not between time and time-frequency representations), there is an issue about it:

https://github.com/scipy/scipy/issues/13489

So I actually don't mind using our internal version to start and then hopefully eventually just changing its internal calls to use `stft`.

agramfort:
the round trip is quite import for TF-MxNE solver but not really otherwise.

if the scipy code much faster? if so I would really encourage to use this
one.

>

larsoner:
> if the scipy code much faster? if so I would really encourage to use this one.

I haven't checked but, having looked at your code, I wouldn't expect there to be large differences between the two

adam2392:
Should I leave things as is and add a note to change things once scipy implementation has been fixed?

mne/time_frequency/_stft.py
```diff
@@ -1,10 +1,140 @@
...
+            'data must be 3D with shape (n_epochs, n_channels, n_times), '
+            f'got {data.shape}')
+
+    n_epochs, n_channels, n_signals = data.shape
+
+    power_list = []
+
+    for epochidx in range(n_epochs):
+        # compute STFT
+        power = _stft(data[epochidx, ...], window_size=window_size,
```
agramfort, line 119: I doubt the stft returns power by default. It should output complex valued data. Can you check?

also I would see if I can preallocate the power_list as a numpy array before the for loop to avoid doubling the RAM usage.

agramfort (commented):
I would really encourage you @adam2392 to do a benchmark with the scipy code in terms of speed and result quality.

mne/time_frequency/_stft.py
```diff
@@ -1,10 +1,140 @@
...
+from ..io.pick import _pick_data_channels, pick_info
 from ..fixes import _import_fft
-from ..utils import logger, verbose
+from ..utils import logger, verbose, deprecated, fill_doc, _validate_type
+from .tfr import AverageTFR, _get_data
 
 
+@fill_doc
+def tfr_stft(inst, window_size, step_size=None, average=True, verbose=None):
+    """Compute STFT Short-Term Fourier Transform using a sine window.
```
mmagnuski, line 13: I think you meant Short-**Time** Fourier Transform here.

PR #8896, opened 2021-02-22
MRG, ENH: Allow channel-specific annotations

### LINKED ISSUE #5507: Event markers per channel
#### Describe the problem
Events and annotations can only be specified for all channels. 

#### Describe your solution
A way to assign an event to a specific channel, along with plotting and analysis routines that handle that would be extremely helpful for removing specific segments of specific bad channels, or for any decoding/classification of channel-wise EEG features. Maybe an extra column in the events array which specifies the channel would be a good way to input this information. 

#### Additional context
I understand this is a large request, but id like to start a discussion here and see if anyone else is interested in this feature.

larsoner:
Closes #5507

Removes some heavy, seemingly overkill parametrization of the `test_annotations.py` where the parameters were only used in a single function for example. Also makes an `overwrite` entry in `docdict`.

Todo:

- [x] Make `rename_channels` change the names in `annotations`
- [x] Make `drop_channels` drop channel names, and channel-specific annotations that no longer apply

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -145,12 +145,13 @@ docdict['reject_by_annotation_raw'] = docdict['reject_by_annotation_all'] + """
     Has no effect if ``inst`` is not a :class:`mne.io.Raw` object.
 """
 docdict['annot_ch_names'] = """
-ch_names : array-like, shape (n_annotations,) | None
-    1D object array of tuples of channel names associated with the annotations.
-    Empty entries are assumed to be associated with all channels (i.e., with
-    the time slice itself). For example, this creates three annotations,
-    associating the first with all channels, the second with two channels,
-    and the third with a single channel::
+ch_names : list | None
+    List of lists of channel names associated with the annotations.
+    Empty entries are assumed to be associated with no specific channel,
+    i.e., with all channels or with the time slice itself. None (default) is
+    the same as passing all empty lists. For example, this creates three
+    annotations, associating the first with the time interval itself, the
+    second with two channels, and the third with a single channel::
 
         Annotations(onset=[0, 3, 10], duration=[1, 0.25, 0.5],
                     description=['Start', 'BAD_flux', 'BAD_noise'],
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -145,16 +145,16 @@ docdict['reject_by_annotation_raw'] = docdict['reject_by_annotation_all'] + """
     Has no effect if ``inst`` is not a :class:`mne.io.Raw` object.
 """
 docdict['annot_ch_names'] = """
-ch_names : array-like | None
-    Array-like of lists of channel names associated with the annotations.
-    Empty entries or entries with None are assumed to be associated with all
-    channels (i.e., with the time slice itself). For example, to create two
-    annotations, associating the second with two channels could be done
-    by using ``ch_names`` like::
+ch_names : array-like, shape (n_annotations,) | None
+    1D object array of tuples of channel names associated with the annotations.
+    Empty entries are assumed to be associated with all channels (i.e., with
+    the time slice itself). For example, this creates three annotations,
+    associating the first with all channels, the second with two channels,
+    and the third with a single channel::
 
         Annotations(onset=[0, 3, 10], duration=[1, 0.25, 0.5],
                     description=['Start', 'BAD_flux', 'BAD_noise'],
-                    ch_names=[[], ['MEG0111', 'MEG2563']])
+                    ch_names=[[], ['MEG0111', 'MEG2563'], ['MEG1443']])
 """
 
 # General plotting
@@ -1418,6 +1418,12 @@ on_missing : str
     %s channels have missing coordinates.
 
     .. versionadded:: 0.20.1
+""" % (_on_missing_base,)
+docdict['on_missing_ch_names'] = """
+on_missing : str
+    %s entries in ch_names are not present in the raw instance.
+
+    .. versionadded:: 0.23.0
 """ % (_on_missing_base,)
 docdict['rename_channels_mapping_duplicates'] = """
 mapping : dict | callable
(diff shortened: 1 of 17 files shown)


mne/utils/docs.py
```diff
@@ -144,6 +144,18 @@
...
+ch_names : array-like | None
+    Array-like of lists of channel names associated with the annotations.
+    Empty entries or entries with None are assumed to be associated with all
+    channels (i.e., with the time slice itself). For example, to create two
+    annotations, associating the second with two channels could be done
+    by using ``ch_names`` like::
+
+        Annotations(onset=[0, 3, 10], duration=[1, 0.25, 0.5],
+                    description=['Start', 'BAD_flux', 'BAD_noise'],
+                    ch_names=[[], ['MEG0111', 'MEG2563']])
```
cbrnr, line 157: You describe creating two annotations, but the example shows three.

Also, would it make sense to also support channel indexes as an alternative to channel names? That way, ranges could be more easily specified. But I guess there's always the `pick_*` route, which maybe should be mentioned in another example if we don't want to support it directly.

larsoner: I think working with channel names will generally be easier for people, especially once picking and reordering gets involved.

This does remind me, though, that I need to fix `rename_channels` to git into `annotations.ch_names` now, though.

agramfort: Agreed. Indices are more error prone

cbrnr: @larsoner but the example is still confusing/wrong, could you clarify?

larsoner:
Okay I think this is good to go now

mne/utils/docs.py
```diff
@@ -144,6 +144,18 @@
 docdict['reject_by_annotation_raw'] = docdict['reject_by_annotation_all'] + """
     Has no effect if ``inst`` is not a :class:`mne.io.Raw` object.
 """
+docdict['annot_ch_names'] = """
+ch_names : array-like, shape (n_annotations,) | None
+    1D object array of tuples of channel names associated with the annotations.
```
cbrnr, line 149:
```suggestion
    1D array of tuples of channel names associated with the annotations.
```

cbrnr:
LGTM +1 for merge when green!

mne/utils/docs.py
larsoner: I think "object array" is more specific and hence useful here, like saying "float array"

cbrnr: Ah, I didn't know you were referring to a string array. Object is a very generic term. I know numpy and pandas use these to refer to arrays that contain objects, but mostly these are strings. Therefore, I'd use string array here.

cbrnr: Even more precise would be list of list of strings (where list could be replaced by array).

cbrnr: `Sequence[Union[None, Sequence[str]]]` would be the correct type hint I guess?

cbrnr: Or maybe even `Sequence[Optional[Sequence[str]]]`...

larsoner: > I know numpy and pandas use these to refer to arrays that contain objects, but mostly these are strings

It's not a string array, it's an object array -- they are different. An ndarray with dtype `str` gets its values stored in memory by NumPy. An ndarray with dtype object just holds pointers to some other Python objects. I'm used to these being called "object array"s but I'll reword things to make this hopefully clearer.

cbrnr: Yes, that's very numpy-specific. A sequence of string sequences would be accurate and easy to understand.

cbrnr: List is also OK!

hoechenberger (approved):
Good stuff!

hoechenberger:
@larsoner The CI error seems unrelated to me!

hoechenberger:
Thank you, @larsoner!

jgcaffari1:
Is this only for the most recent version of python? I do not seem to have this functionality even though I am up to date.

hoechenberger:
The latest development version and the upcoming release of MNE-Python 0.24 only support Python 3.7 and newer.

jgcaffari1:
I have updated everything and the annotations recognize ch_names as input, but they do not actually annotate specific channels.   Do the channels need to have specific names as strings?

drammock:
> I have updated everything and the annotations recognize ch_names as input, but they do not actually annotate specific channels. Do the channels need to have specific names as strings?

so far this feature has no effect when *plotting* (we plan to do that, but it's not done yet).  Is that what you're noticing?

jgcaffari1:
That is what I am noticing.  Would I be able to see the channel specific annotations if I save and open in another program?

drammock:
> Would I be able to see the channel specific annotations if I save and open in another program?

I think SigViewer has the ability to display channel-specific annotations.  I don't know if SigViewer can read our `Annotations` data structure (but you can export our annotations to CSV).  @cbrnr would know more...

jgcaffari1:
Thank you.

cbrnr:
Yes, SigViewer should be able to import events from a CSV file with columns position, duration, channel, type, and name. A channel value of -1 means all channels. However, I just tried this with a dummy file and the import didn't work (but there was no error either), so I guess this functionality is currently broken 😢 .

jackz314:
Just wondering if there has been any progress on displaying per-channel annotations?

larsoner:
Not that I've seen

drammock:
> Just wondering if there has been any progress on displaying per-channel annotations?

@britta-wstnr good one for the fall sprint?

aplzr:
Chiming in to say that I would very much appreciate this feature as well.

aplzr:
I just realized that [visualization of channel-specific annotations has actually been implemented in mne-qt-browser](https://github.com/mne-tools/mne-qt-browser/pull/202), so I'm posting here again for the benefit of everyone that didn't find the respective issues and pull requests, but _did_ find this pull request (as I had initially).

- Install `mne-qt-browser` (`python -m pip install mne-qt-browser`)
- Activate its qt backend
    ```python
    from mne.viz import set_browser_backend
    set_browser_backend("qt")
    ```
- Plot a `RawArray` that has channel-specific annotations

mneqtspindleannot

PR #8956, opened 2021-02-24
MRG, BUG: Fix bug with empty dim

larsoner:
When cHPI is off, we should still return arrays of the correct dimensionality.


agramfort (approved):
not sure why CIs complain

thx @larsoner

PR #8959, opened 2021-02-25
MRG, ENH: macOS 11 compatiblity for STC plots

hoechenberger:
Sets `QT_MAC_WANTS_LAYER=1` for STC plots via PyVista / pyvistaqt.

Was required on my system with macOS 11.2.1, Qt 5.12.9, pyqt 5.12.3 (installed from `conda-forge`)

Previous fix via #8554 didn't cover STC plots.

cc @cbrnr, @GuillaumeFavelier

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -86,6 +86,8 @@ Enhancements
 
 - Reading and writing FIFF files whose filenames end with ``_meg.fif.gz``, ``_eeg.fif(.gz)``, and ``_ieeg.fif(.gz)`` doesn't emit a warning anymore; this improves interobaility with BIDS-formatted datasets (:gh:`8868` by `Richard Höchenberger`_)
 
+- macOS 11 compatibility for PyVista `~mne.SourceEstimate` plots (:gh:`8959` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix bug with `mne.connectivity.spectral_connectivity` where time axis in Epochs data object was dropped. (:gh:`8839` **by new contributor** |Anna Padee|_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -86,7 +86,7 @@ Enhancements
 
 - Reading and writing FIFF files whose filenames end with ``_meg.fif.gz``, ``_eeg.fif(.gz)``, and ``_ieeg.fif(.gz)`` doesn't emit a warning anymore; this improves interobaility with BIDS-formatted datasets (:gh:`8868` by `Richard Höchenberger`_)
 
-- macOS 11 compatibility for PyVista `~mne.SourceEstimate` plots (:gh:`8959` by `Richard Höchenberger`_)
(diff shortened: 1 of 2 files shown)


cbrnr:
It's strange that conda-forge is still stuck at this old PyQt version (5.12 was released 2 years ago). Versions >=5.15.2 don't have this problem anymore.

hoechenberger:
> It's strange that conda-forge is still stuck at this old PyQt version

It seems compilation of Qt 5.15 is a blocker :\
https://github.com/conda-forge/pyqt-feedstock/issues/98

hoechenberger:
@GuillaumeFavelier please merge if you think this is okay :)

mne/viz/backends/renderer.py
```diff
@@ -112,6 +114,11 @@ def set_3d_backend(backend_name, verbose=None):
         _reload_backend(backend_name)
         MNE_3D_BACKEND = backend_name
 
+    # Qt5 macOS 11 compatibility
+    if (backend_name == 'pyvista' and sys.platform == 'darwin' and
```
larsoner, line 118: I actually don't think the backend name matters, mayavi will have the same problem

hoechenberger: I've dropped the `pyvista` check in 77f36c9

doc/changes/latest.inc
```diff
@@ -86,6 +86,8 @@ Enhancements
 
 - Reading and writing FIFF files whose filenames end with ``_meg.fif.gz``, ``_eeg.fif(.gz)``, and ``_ieeg.fif(.gz)`` doesn't emit a warning anymore; this improves interobaility with BIDS-formatted datasets (:gh:`8868` by `Richard Höchenberger`_)
 
+- macOS 11 compatibility for `~mne.SourceEstimate` plots (:gh:`8959` by `Richard Höchenberger`_)
```
drammock, line 89: I think the changelog entry should explicitly mention that an env var is being created. Even if right now this only affects MNE, it can potentially affect other programs in future so we should communicate it prominently.

hoechenberger: My understanding is that these variables will not be visible to any other applications - only to the current Python interpreter and all subprocesses it might spawn.

hoechenberger:
```shell
(base) ➜  ~ echo ${FOO}; python -c "import os; os.environ['FOO'] = 'bar'"; echo ${FOO}


(base) ➜  ~
```

larsoner: Substitute "other programs" with "other Python modules" and the point makes sense though

hoechenberger: > "other Python modules"

Oh wow, so it's not even global for everything running through the same Python interpreter? Crazy :)

larsoner: No, I meant Dan's point stands -- it will affect those

hoechenberger: I've improved the change log entry. Feel free to merge if you're happy!

drammock:
Tx @hoechenberger !

PR #8963, opened 2021-02-26
Fix indexing dipoles read from a bdip file

wmvanvliet:
When reading dipoles from a `*.bdip` file, not all attributes of the `Dipole` object were proper numpy arrays. This only becomes apparent when you try to index the `Dipole` and get indexing errors. A quick PR that adds a unit test and fix for this.


agramfort (approved):

hoechenberger:
Thanks, @wmvanvliet!

PR #8965, opened 2021-02-26
MRG: "GFP" trace in STC time viewer is actually RMS

hoechenberger:
I've also rearranged the respective formula such that it becomes immediately obvious that we're dealing with RMS values.

See #8775 for our lengthy discussion on GFP vs RMS naming, albeit for sensor-level data.


hoechenberger:
Rendering looks good:
https://25463-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/source-modeling/plot_mne_dspm_source_localization.html#visualization

mne/viz/_brain/_brain.py
```diff
@@ -1130,13 +1130,15 @@ def _configure_vertex_time_course(self):
...
 
-        # plot the GFP
+        # plot RMS of the activation
         y = np.concatenate(list(v[0] for v in self.act_data_smooth.values()
                                 if v[0] is not None))
-        y = np.linalg.norm(y, axis=0) / np.sqrt(len(y))
-        self.gfp, = self.mpl_canvas.axes.plot(
-            self._data['time'], y,
-            lw=3, label='GFP', zorder=3, color=self._fg_color,
+        rms = np.sqrt((y ** 2).mean(axis=0))
```
agramfort, line 1136: Previous code was more efficient as did not need to allocate y**2

agramfort (commented):
Besides lgtm

mne/viz/_brain/_brain.py
hoechenberger: Yes, but does it really matter?
Can of course revert if you prefer!

agramfort:
Well maybe not but it’s bad practice I would say. It advertised not optimal code for no reason

hoechenberger:
@agramfort Ok I've addressed this, also in `viz/evoked.py`

agramfort (approved):
Thx @hoechenberger 👍

larsoner:
FYI this had CI failures beyond just pip and pip pre, fixed by #8971

hoechenberger:
Thanks for taking care of this, @larsoner

PR #8972, opened 2021-03-01
FIX: Brain lights

GuillaumeFavelier:
This PR fixes the 3 light system:

PR | master
--|--
image | image

The patch is just shared to all renderers. Such a function is already implemented on pyvista master so the function `update_lighting()` will be replaced by `enable_3_lights()` in the future.

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -525,9 +525,6 @@ class Brain(object):
         if surf == 'flat':
             self._renderer.set_interaction("rubber_band_2d")
 
-        if hemi == 'rh' and hasattr(self._renderer, "_orient_lights"):
-            self._renderer._orient_lights()
-
     def _setup_canonical_rotation(self):
         from ...coreg import fit_matched_points, _trans_from_params
         self._rigid = np.eye(4)
--- a/mne/viz/backends/_pyvista.py
+++ b/mne/viz/backends/_pyvista.py
@@ -262,35 +262,24 @@ class _Renderer(_BaseRenderer):
     def scene(self):
         return self.figure
 
-    def _orient_lights(self):
-        lights = list(self.plotter.renderer.GetLights())
-        lights.pop(0)  # unused headlight
-        lights[0].SetPosition(_to_pos(45.0, -45.0))
-        lights[1].SetPosition(_to_pos(-30.0, 60.0))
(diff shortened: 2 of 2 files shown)


larsoner:
Perhaps a separate bug, it looks like the lighting is asymmetric:

1. the back/posterior of the left hemisphere and the front/anterior of the right hemisphere are illuminated, and
2. the front/anterior of the left hemisphere and the back/posterior of the right hemisphere are dark

Are the lights set up to illuminate equivalently the left and right? Are they attached to the camera, or to the world?

GuillaumeFavelier:
> Are the lights set up to illuminate equivalently the left and right?

I think we use the `_orient_lights()` function for that:

https://github.com/mne-tools/mne-python/blob/2c485d57331543652b16a29167c86a4b5b0ee128/mne/viz/_brain/_brain.py#L528-L529

https://github.com/mne-tools/mne-python/blob/2c485d57331543652b16a29167c86a4b5b0ee128/mne/viz/backends/_pyvista.py#L265-L270

GuillaumeFavelier:
> Are they attached to the camera, or to the world?

`Light 0` is a Headlight, `Light 1` and `Light 2` are CameraLights.

And according to VTK doc:

> A Headlight is always located at the camera and is pointed at the camera's focal point. The renderer is free to modify the position and focal point of the camera at any time.

> A CameraLight is also attached to the camera, but is not necessarily located at the camera's position. CameraLights are defined in a coordinate space where the camera is located at (0, 0, 1), looking towards (0, 0, 0) at a distance of 1, with up being (0, 1, 0). CameraLight uses the transform matrix to establish this space.

Reference: https://vtk.org/doc/nightly/html/classvtkLight.html#a1e0175706ef066ca837726425bdcc6d7

larsoner:
Okay here it was on your commit:

Screenshot from 2021-03-01 12-43-03

And here it is on mine that I just pushed -- the lighting is uniform left/right which I think is important (otherwise you end up with the asymmetry I mention above):

Screenshot from 2021-03-01 13-00-01

Okay for you @GuillaumeFavelier ?

agramfort (approved):

GuillaumeFavelier:
Yes, all good 👌

larsoner:
Thanks @GuillaumeFavelier !

PR #8975, opened 2021-03-01
MRG, ENH: Better float slider spacing and interaction

larsoner:
1. Kill `QLineEdit` in favor of `QDoubleSpinBox` with a sensible step. Automagically fixes spacing issues and makes things more compact and usable (default step is 1/20th of the limits, which works well in testing). Also use `setKeyboardTracking(False)` to avoid needing to update until enter is pressed.
2. Add `mousePressEvent` handling to QFloatSlider so that when you click somewhere specific on the slider, you jump to that point.
3. Use `QFloatSlider` for the time axis so that non-integral time points can be used, which is the way it was before and is especially useful when using non-nearest-neighbor interp (doesn't seem to hurt nearest-neighbor use cases because you still see it "snap" on changes, and has the advantage of jump-to-location on click).

@GuillaumeFavelier can you try it and see what you think?

`main`:

https://user-images.githubusercontent.com/2365790/109566835-6f990a00-7ab2-11eb-8f18-8b3e6b5b7b65.mp4

this PR:

https://user-images.githubusercontent.com/2365790/109566849-745dbe00-7ab2-11eb-962d-9db0be6e2367.mp4


GuillaumeFavelier:
I hope you don't mind @larsoner, I pushed a commit to port `_named_layout` to `_notebook` as well

larsoner:
Yes that's great!

Next I think I'll change `scale` just to be `-` and `+` buttons. In theory we could do a slider where you click to the right or left and it rescales by the proportion away from the midpoint you are (i.e., basically the way we had it before) but this will require custom widgets in both backends and I'm not sure it ends up being that much more usable than buttons that you can just click somewhat rapidly. For example if the `+` increases by (and `-` decreases by the inverse of) 1.2x then you've doubled the limits in 4 clicks, which I think will be pretty good for interaction.

GuillaumeFavelier (approved):
I restarted Azure but it seems to be stuck at the dependencies step. I don't think it's related.

hoechenberger (approved):
Awesome! If @GuillaumeFavelier is happy, this should be good to merge

PR #8978, opened 2021-03-02
MAINT: Revert conda azure change

larsoner:
For some unknown reason #8973 caused timeouts in the windows 3.7 conda run. This PR reverts the conda commands to be in the YAML which is a bit annoying but should at least work. Probably some bash vs powershell issue...


larsoner:
... and now dipy.org is down so we have to ignore CircleCI failing to retrieve their intersphinx inventory. :)

PR #8980, opened 2021-03-02
MRG, DOC: fix link to install instructions

drammock:
fixes a doc link that is currently resolving (wrongly) to https://docs.python.org/3/install/index.html


larsoner:
Thanks @drammock !

PR #8988, opened 2021-03-03
MRG, FIX: Fix rebuilding of all files

larsoner:
The date changes that resolve the datetime locally are useful for end-users but bad for developers because it means that a sphinx config value changes on every run, which means that all files must be rewritten. On this PR CircleCI gets the full date, but locally we get a config var that just changes once a year, which seems like a good compromise.


drammock (approved):

larsoner:
Things look good on CircleCI, work locally, and flake is happy so in it goes

PR #8989, opened 2021-03-04
FIX: Handling bad marker coils when reading KIT data #8981

### LINKED ISSUE #8981: Handling bad marker coils when reading KIT data
Hello,

We are attempting to read .con files from KIT system, and mne.io.read_raw_kit() runs into an error when there are missing marker coils. This seems to be caused by retrieving coordinates for markers (e.g. 'hpi_1') when they don't exist (line 716 in mne/io/kit/kit.py).

We propose the following solution, which checks whether a marker coil exists before retrieving its values:

    # nasion, lpa, rpa, HPI in native space
    #elp = [dig.pop(key) for key in (
    #    'fidnz', 'fidt9', 'fidt10',
    #    'hpi_1', 'hpi_2', 'hpi_3', 'hpi_4')]
    elp = []
    for key in (
        'fidnz', 'fidt9', 'fidt10',
        'hpi_1', 'hpi_2', 'hpi_3', 'hpi_4'):
        if key in dig:
            elp.append(dig.pop(key))

@JD-Zhu @paulsowman

JD-Zhu:
Fixes #8981.

When reading .con files from KIT system, an error occurs if there are missing marker coils. The proposed solution here resolves this problem by checking whether a marker coil exists before retrieving its coordinates.


JD-Zhu:
@agramfort I think this is ready for review. 

It's my first contribution - sorry about the commit messages, I didn't realise the tags should be included in each commit, rather than just the PR title.

mne/io/kit/kit.py
```diff
@@ -720,9 +720,12 @@ def get_kit_info(rawfile, allow_unknown_format, standardize_names=None,
...
-                'fidnz', 'fidt9', 'fidt10',
-                'hpi_1', 'hpi_2', 'hpi_3', 'hpi_4')]
+            elp = []
+            for key in (
+                    'fidnz', 'fidt9', 'fidt10',
+                    'hpi_1', 'hpi_2', 'hpi_3', 'hpi_4'):
+                if key in dig:
+                    elp.append(dig.pop(key))
             if 'hpi_5' in dig and dig['hpi_5'].any():
                 elp.append(dig.pop('hpi_5'))
```
agramfort, line 730: it seems this if just below is exactly to catch the issue of missing hpi_5. Maybe we can use the same logic for all hpi's ?

JD-Zhu: Yes I think that's what it's doing, but not sure why it was checking specifically for hpi_5 and not the others.

Also, based on the additional condition `dig['hpi_5'].any()`, it seems like [0, 0, 0] is considered as invalid coordinates. Again not sure if this is something specific to hpi_5 or should be checked for all marker coils. I don't think we should treat [0, 0, 0] as invalid in general though.

drammock:
> It's my first contribution - sorry about the commit messages, I didn't realise the tags should be included in each commit, rather than just the PR title.

Hooray for your first contribution!! FYI, it's not crucial to put tags like `sty` or `doc` in every commit message if the messages are descriptive (like yours are). When we merge PRs we squash all commits into one anyway. Your commit messages are better than many of mine, which often just say "missed one" or "try again" 😆

agramfort:
I don't have enough test KIT files so maybe we should just take your fix
here and it seems
the less likely to introduce a regression

>

mne/io/kit/kit.py
larsoner: > Yes I think that's what it's doing, but not sure why it was checking specifically for hpi_5 and not the others.

I added this because there was one dataset I think was from @paulsowman that had the first four only. But I suppose it makes sense that you might be missing any one (or up to two?) of the five.

JD-Zhu:
Thanks @larsoner

Indeed we have another dataset with two missing coils. The present fix cannot handle that situation yet, as it would fail at the assertion a few lines below:

            assert elp.shape in ((7, 3), (8, 3))

In the case of two missing coils, the shape becomes (6, 3). I tried removing this assertion temporarily, but it seems to run into more problems later on.

________________________________

larsoner:
That's an assert statement I put in to make sure that, if someone changed the behavior, they look at the downstream impact. Let me know if you need help assessing it. But I think checking that at least 3 are available, and maybe indexing of the relevant ones (I might have coded it to assume the last coil was the missing one?) will be necessary

JD-Zhu:
> That's an assert statement I put in to make sure that, if someone changed the behavior, they look at the downstream impact. Let me know if you need help assessing it. But I think checking that at least 3 are available, and maybe indexing of the relevant ones (I might have coded it to assume the last coil was the missing one?) will be necessary

Thanks for your help @larsoner
I changed the assert to allow a minimum of 3 marker coils (+ 3 fiducials), and did the same for a similar check in mne\io\kit\coreg.py (line 151). This seems to work fine without further issues.
For the indexing of relevant marker coils, I think this is already provided by their names (hpi_1, hpi_2, etc)?

JD-Zhu:
I can't figure out why there is a merge conflict. When I pull from upstream, it says "already up to date".

JD-Zhu:
One of the tests failed - I can't figure out what it means / how to fix it. Would appreciate any advice!

mne/io/kit/kit.py
larsoner: It looks like this special case is no longer needed. All `hpi_*` should be treated the same way.

While unifying it, we can try getting rid of the `.any` and see if it breaks anything. Maybe `np.isfinite(dig['hpi_5']).all()` is a better check.

mne/io/kit/kit.py
```diff
@@ -720,14 +720,17 @@ def get_kit_info(rawfile, allow_unknown_format, standardize_names=None,
...
             elp = np.array(elp)
             hsp = np.array(hsp, float).reshape(-1, 3)
-            assert elp.shape in ((7, 3), (8, 3))
+            assert elp.shape in ((6, 3), (7, 3), (8, 3))
```
larsoner, line 733: Now that it's possible for a user (as opposed to only a dev making code chainges) to run into a situation where this could be violated, this should become a proper check and raise, e.g.:
```suggestion
            if elp.shape not in ((6, 3), (7, 3), (8, 3)):
                raise RuntimeError(f
                    f'Fewer than 3 HPI coils found, got {len(elp) - 3}')
```

mne/io/kit/kit.py
JD-Zhu: Agreed and implemented. Doesn't seem to break anything.

mne/io/kit/kit.py
JD-Zhu: Implemented the suggested change.

doc/changes/names.inc
```diff
@@ -363,3 +363,5 @@
 .. _Enrico Varano: https://github.com/enricovara/
 
 .. _Dominik Welke: https://github.com/dominikwelke/
+
+.. _Judy D Zhu: https://github.com/JD-Zhu
```
agramfort, line 367: missing empty line at the end of the file

agramfort (commented):
diff LGTM

can you merge again the main branch to see if it fixes the CIs. The +1 for merge when green

JD-Zhu:
Still 1 failed test. Any idea what the issue might be?

agramfort:
it does not seem related. Let's wait for us to fix this other issue
and then we'll merge this one. thx

>

JD-Zhu:
Sure :smiley:

doc/changes/names.inc
larsoner, line 367: @JD-Zhu can you add a line to the top of the `Bugs` section of `doc/changes/latest.inc` about this change? Follow the format for new contributors involving bolding, etc.

JD-Zhu: In the description, should I reference the function that led to the error (mne.io.read_raw_kit), or the actual (internal?) function I modified (get_kit_info in mne/io/kit/kit.py)? I referenced the former - please let me know if that's incorrect. Thanks.

larsoner: We only refer to public-facing functions in `latest.inc`, so that's perfect!

(`get_kit_info` isn't really supposed to be public-facing or used by people even though it's not prefixed by an `_` like it probably should be.)

larsoner:
Thanks @JD-Zhu !

JD-Zhu:
Thanks @larsoner @agramfort @drammock 
Happy to contribute!

PR #8996, opened 2021-03-04
VIZ: fix pick index when right-click on channel name in ica plot

### LINKED ISSUE #8995: using picks with ica.plot_sources breaks interactive properties plot
If i specify `picks` to `ica.plot_sources` If I then click on the ica_name to see the properties the indexing is incorrect and the properties for the wrong IC is plotted

For example in the image below I am not plotting **ICA001**, now if I right click on the **ICA004** to plot it's properties I get a window with the correct IC name however the topo and presumably the rest of the the properties plot is for the wrong IC (ICA003) in this case.

image

cc @drammock

drammock:
closes #8995 

cc @bloyl

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -7,6 +7,11 @@
    - "Bugs" for bug fixes
    - "API changes" for backward-incompatible changes
 
+.. NOTE: changes from first-time contributors should be added to the TOP of
+   the relevant section (Enhancements / Bugs / API changes), and should look
+   like this (where xxxx is the pull request number):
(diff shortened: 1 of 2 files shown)


hoechenberger (commented):
Nice!

- Could you add a changelog entry?
- ideally, this should be back ported too

bloyl:
Works like a charm. Thanks.

I'm not sure if this has been in a release yet (I might be wrong on that) so backporting etc might not be needed.

hoechenberger (approved):

hoechenberger:
Thank you, @drammock!

PR #8997, opened 2021-03-05
WIP: Prototype of MultiPlotter

GuillaumeFavelier:
I have been experimenting with the new PyVistaQt class MultiPlotter to bring proper SSAO and/or EDL to MNE on https://github.com/mne-tools/mne-python/pull/8771#issuecomment-765449184 (among other features like multiple toolbars, menu bars and multi-renderers). And I don't think it's reasonable to do that in one go.

This PR is far from being mergeable but I believe it could drive smaller PRs that can help with 3d backend refactoring in general. For instance:
- [ ] Separate the concepts of "plotter" and 3d "viewer"
- [ ] Get rid of `self.plotter` in `Brain` and use `self._renderer` instead
- [ ] Get rid of `self.plotter` in `_Renderer` and use `self.figure` instead
- [ ] Create proper `Renderer` bindings for `add_mesh` and `add_actor`
- [ ] Generic way to pick a 3d viewer
- [ ] Generic way to subplot
- [ ] Generic way to iterate through the 3d viewers

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -476,10 +476,11 @@ class Brain(object):
                                        fig=figure)
 
         self.plotter = self._renderer.plotter
+        self.viewer = self._renderer.viewer
         if self.notebook:
             self.window = None
         else:
-            self.window = self.plotter.app_window
+            self.window = self.plotter._window
             self.window.signal_close.connect(self._clean)
 
         self._setup_canonical_rotation()
@@ -628,9 +629,10 @@ class Brain(object):
             self.status_bar = None
             self.interactor = None
         else:
-            self.main_menu = self.plotter.main_menu
             self.status_bar = self.window.statusBar()
-            self.interactor = self.plotter.interactor
+            # XXX:WIP
+            # self.main_menu = self.plotter.main_menu
+            # self.interactor = self.plotter.interactor
 
         # Derived parameters:
         self.playback_speed = self.default_playback_speed_value
@@ -672,14 +674,16 @@ class Brain(object):
             self.mpl_canvas.show()
         self.toggle_interface()
         if not self.notebook:
-            self._configure_playback()
-            self._configure_menu()
+            # XXX:WIP
+            # self._configure_playback()
+            # self._configure_menu()
             self._configure_status_bar()
-
-            # show everything at the end
-            with _qt_disable_paint(self.plotter):
-                with self._ensure_minimum_sizes():
-                    self.show()
+            
+            # XXX:WIP
+            # # show everything at the end
+            # with _qt_disable_paint(self.plotter):
+            #     with self._ensure_minimum_sizes():
+            #         self.show()
             self._update()
 
     @safe_event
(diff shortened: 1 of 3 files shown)


larsoner:
I would add to this list, if we haven't done it already:

- [ ] Avoid using PyVista's toolbar and menu bar and create/use our own as needed

GuillaumeFavelier:
Done in here on this PR:

https://github.com/mne-tools/mne-python/blob/f20f8161e4596318619b665860ccb02a2c6628c2/mne/viz/backends/_pyvista.py#L82-L83

GuillaumeFavelier:
Update after dd0d67ddc03e34fa8efd5fd07efe5e65166b9149 (with experimental `margin=False`):

image

I see the main challenges being 1) ensuring correct size and 2) managing picking nicely for all the viewers

larsoner:
Can you check to make sure volumetric rendering looks okay? Just want to make sure it plays nicely with the ambient occlusion etc.

GuillaumeFavelier:
There are some issues with the colorbar:

```
BackgroundPlotter has no attribute named scalar_bar
```

SSAO=on | SSAO=off
--|--
image | image

<details>

```py
import os.path as op
import mne
    # sample_dir, 'sample_audvis-meg-oct-6-meg-inv.fif'))
    evoked, inv, return_residual=True, pick_ori='vector')

initial_time = 0.1
# stc, kwargs = stc, dict(overlay_alpha=0.5, brain_alpha=0.5, vector_alpha=0.)
stc, kwargs = stc.magnitude(), dict(alpha=0.5, surface='white')
kwargs['src'] = inv['src']
meth = stc.plot_3d if isinstance(stc, mne.VolSourceEstimate) else stc.plot
brain = meth(
    clim='auto', hemi='both', # views=['sag', 'axi', 'cor'],
    view_layout='horizontal', smoothing_steps='nearest', verbose=True,
    time_viewer=True, size=(800, 600), background='k',
    show_traces=False,
    **kwargs)
```

</details>


and `plot_beamformer_lcmv`:

image

GuillaumeFavelier:
It's better after #9121 

image

GuillaumeFavelier:
This PR now requires https://github.com/pyvista/pyvistaqt/pull/90. 

I plan to move the `mplcanvas` to a bottom dock and use the new `resize_content()` function to replace `ensure_minimum_size`.

GuillaumeFavelier:
In my experiments, the `mplcanvas` as a dock has several advantages:
* it can be made separate dynamically with the default dock features (meaning that the "separate" particular case can be removed)
* it does not mess with the central widget layout meaning that the `resize_content()` function is enough the ensure the viewer dimensions (important for screenshot). I have to demonstrate that by updating testing.
* the size of the dock can be handled separately

This is how it looks locally:

image

Probably the margins can be nullified.

larsoner:
Agreed docking is useful, can that be a first PR, separate from the MultiPlotter stuff? It seems orthogonal, easier to review, and will get in sooner than if we tie it into this PR

GuillaumeFavelier:
Circle does not seem ready yet. I'll investigate locally:

```
RuntimeWarning: No 3D backend found. Resorting to matplotlib 3d
```

PR #9000, opened 2021-03-05
WIP, ENH: Button to change the background color

GuillaumeFavelier:
This PR adds a color picker in the tool bar. It's still work in progress.

https://user-images.githubusercontent.com/18143289/110133645-d416da80-7dcc-11eb-8de6-9cc781d2cd75.mp4

ToDo:
- [ ] Update matplotlib canvas too
- [ ] Add notebook support
- [ ] test/coverage

It's an item of #7162

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1368,9 +1368,16 @@ class Brain(object):
                 callback=partial(_save_image, img=img)
             )
 
+    def _set_background(self, val):
+        self.plotter.background_color = val
+
     def _configure_tool_bar(self):
         self._renderer._tool_bar_load_icons()
         self._renderer._tool_bar_initialize()
+        self._renderer._tool_bar_add_color_picker(
+            name="background",
+            func=self._set_background,
+        )
         self._renderer._tool_bar_add_button(
             name="screenshot",
             desc="Take a screenshot",
--- a/mne/viz/backends/_abstract.py
+++ b/mne/viz/backends/_abstract.py
@@ -465,6 +465,10 @@ class _AbstractToolBar(ABC):
         pass
 
     @abstractmethod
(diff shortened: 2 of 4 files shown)


GuillaumeFavelier:
You might be interested @hoechenberger since you suggested the idea :)

hoechenberger:
Awesome, I'd totally forgotten I'm slightly creative sometimes! 😅 will play with it ASAP!!!!

hoechenberger:
@GuillaumeFavelier It's great!

But it's actually **much more** than I'd actually meant to propose 😅
See, my idea was to have a button which would essentially toggle between a "light" and a "dark" background – with the idea that, at some time in the future, the button press could switch the **entire** GUI theme between light and dark.

The color picker is a great idea, but I think it's over-engineered for **my** use cases, and won't play as nicely with the idea to add a GUI theme switching functionality. See what I mean..?

What do others think?

GuillaumeFavelier:
Whoops, I misunderstood then. I can close if it's not needed :+1:

GuillaumeFavelier:
The light/dark theme is on the todo list, I'll link your comment to have the details :)

hoechenberger:
Sorry about this 😕

larsoner:
@hoechenberger if you want to get a start on the light/dark theme, I would look into:

1. Look up how light/dark theme of the OS can be detected by Qt (probably some standard, platform-independent command nowadays?)
2. Look how to switch Qt between two themes (probably via stylesheets?)

This is probably a common issue nowadays so some googling + quick changes might get us most of the way there. I don't think we need to get it working on all OSes and both Qt/notebook at once, even if it just works on macOS+Qt it's a good start.

FWIW I don't think we will want to change the background color of the render window based on the theme, just the controls. Otherwise the screenshots/images people produce will differ across machines, which is not good.

PR #9014, opened 2021-03-09
MNT: menu_bar & status_bar

GuillaumeFavelier:
This PR disables the default menu bar created by PyVistaQt and uses the `_QtMenuBar` interface instead.

It also helps with https://github.com/mne-tools/mne-python/pull/8997

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -624,11 +624,9 @@ class Brain(object):
         # Direct access parameters:
         self.tool_bar = None
         if self.notebook:
-            self.main_menu = None
             self.status_bar = None
             self.interactor = None
         else:
-            self.main_menu = self.plotter.main_menu
             self.status_bar = self.window.statusBar()
             self.interactor = self.plotter.interactor
 
@@ -667,13 +665,13 @@ class Brain(object):
         self._configure_picking()
         self._configure_tool_bar()
         self._configure_dock()
+        self._configure_menu()
         if self.notebook:
             self._renderer.show()
             self.mpl_canvas.show()
         self.toggle_interface()
         if not self.notebook:
             self._configure_playback()
-            self._configure_menu()
             self._configure_status_bar()
 
             # show everything at the end
@@ -712,10 +710,10 @@ class Brain(object):
         if getattr(self.plotter, 'picker', None) is not None:
             self.plotter.picker = None
         # XXX end PyVista
-        for key in ('plotter', 'main_menu', 'window', 'tool_bar',
+        for key in ('plotter', 'window', 'dock', 'tool_bar', 'menu_bar',
                     'status_bar', 'interactor', 'mpl_canvas', 'time_actor',
-                    'picked_renderer', 'act_data_smooth', '_iren',
-                    'actions', 'widgets', 'geo', '_hemi_actors', '_data'):
+                    'picked_renderer', 'act_data_smooth',
+                    'actions', 'widgets', 'geo', '_data'):
             setattr(self, key, None)
 
     @contextlib.contextmanager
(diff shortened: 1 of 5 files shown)


GuillaumeFavelier:
I'll work on a `_QtStatusBar` interface too.

GuillaumeFavelier:
It's good to go on my end. I'm waiting for the CIs.

mne/viz/_brain/_brain.py
```diff
@@ -1463,24 +1458,27 @@ def _configure_shortcuts(self):
...
-        self.status_msg = QLabel(self.default_status_bar_msg)
-        self.status_progress = QProgressBar()
-        self.status_bar.layout().addWidget(self.status_msg, 1)
-        self.status_bar.layout().addWidget(self.status_progress, 0)
+        if self.notebook:
+            return
+        self._renderer._status_bar_initialize(window=self.window)
+        self.status_msg = self._renderer._status_bar_add_label(
+            self.default_status_bar_msg, stretch=1)
+        self.status_progress = self._renderer._status_bar_add_progress_bar()
```
larsoner, line 1481: Rather than all this special casing, I would make these part of QtDock / IPythonDock / AbstractDock and just have the IPythonDock be a no-op when necessary

larsoner: (Really the name maybe should be QtGUI / IPythonGUI / AbstractGUI, but the idea is that any special-casing based on backend functionality can be made just based on which backend class is called into rather than using `self.notebook` in `_brain.py`, other than the very first one when you use the `notebook` arg to decide which GUI class to use)

GuillaumeFavelier: It leaves some room for future implementation but at the moment, neither the status bar of the menu bar is implemented for the notebook backend.

larsoner: > It leaves some room for future implementation but at the moment, neither the status bar of the menu bar is implemented for the notebook backend.

Indeed -- the GUI backends will implement things differently. The extreme case of "differently" means "not at all" in some cases, and that's okay.

The more important aspect to me is simplifying code flow for maintenance purposes. Eventually there should be one place in `brain` where `self.notebook` matters, and that's at the beginning. Then everything else that depends on the backend lives in that backend's class. Then people who work on `Brain` don't have to see spaghetti code that sometimes defers to the class and other times uses `self.notebook` to decide what to do. The interface is just that you call to the right method of the GUI wrapper class, whose interface is nicely defined in `_abstract.py`. And if what you need is not there, you modify the abstract + classes as necessary. Make sense?

larsoner: ... or to put it in quantitative terms:
```
$ git grep "if self\.notebook"
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:            if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:            if self.notebook and self._renderer.figure.display is not None:
mne/viz/_brain/callback.py:        if self.notebook:
mne/viz/_brain/callback.py:        if self.notebook:
mne/viz/_brain/mplcanvas.py:        if self.notebook:
mne/viz/_brain/mplcanvas.py:        if self.notebook:
mne/viz/backends/_pyvista.py:        if self.notebook:
```
All of these -- or at least the ones in `_brain`, `callback`, and `mplcanvas` -- should go away, other than one in `_brain.py` during `Brain` instantiation.

If you're convinced, do you want to see how feasible it is to do this in this PR? If not, you or I (if you're not interested in it) could take a stab at it in another one.

PR #9015, opened 2021-03-09
MRG, ENH: viz.plot_drop_log() gains special-handling for subject=None

hoechenberger:
Now omits displaying the subject name in the title if it's `None` (would display literal `None` before, causing confusion to some!)

Also changes the title of the drop log plot to make it clearer what is being shown.


mne/epochs.py
```diff
@@ -1243,7 +1243,7 @@ def drop_log_stats(self, ignore=('IGNORED',)):
         return _drop_log_stats(self.drop_log, ignore)
 
     @copy_function_doc_to_method_doc(plot_drop_log)
-    def plot_drop_log(self, threshold=0, n_max_plot=20, subject='Unknown',
+    def plot_drop_log(self, threshold=0, n_max_plot=20, subject='Unknown subj',
```
hoechenberger, line 1246: This is for consistency with `viz.plot_drop_log()`

agramfort (approved):

hoechenberger:
Demo:
https://25716-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#adding-custom-plots-to-a-report

mne/report.py
```diff
@@ -2086,7 +2086,10 @@ def _render_epochs(self, epo_fname, image_format, data_path):
         """Render epochs."""
         global_id = self._get_id()
         epochs = read_epochs(epo_fname)
-        kwargs = dict(subject=self.subject, show=False)
+        kwargs = dict(
+            subject=self.subject if self.subject is not None else self.subject,
```
drammock, line 2090: What does this change do exactly? Why is the conditional needed?

drammock (commented):
I know this is already merged, but see comment below

hoechenberger:
@drammock 
The conditional allows users to entirely disable displaying the subject ID in the plot title.

[semi-unrelated] For historical reasons (I guess?), `viz.plot_drop_log()` defaults to the subject ID `Unknown subj`, and `Epochs.plot_drop_log()` defaulted to `Unknown`; I've now made it such that the defaults are the same in both: `Unknown subj`

[actual argument starts here] There was, however, no way to entirely **disable** displaying the subject ID, e.g. if you didn't have one. In this case, you'd either have to go with the default of `Unknown`/`Unknown subj`, which is … not great; or you could try to pass `subject=None` or an empty string, `subject=''`. Let's see what happens there (running on `maint/0.22`):
```python
# %%
import os
import mne

path = mne.datasets.sample.data_path()
fname_raw = os.path.join(path, 'MEG', 'sample', 'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(fname_raw, verbose=False).crop(tmax=60)
            'visual/right': 4, 'face': 5, 'buttonpress': 32}

# create some epochs and ensure we drop a few, so we can then plot the drop log
reject = dict(eeg=150e-6)
epochs = mne.Epochs(raw=raw, events=events, event_id=event_id,
                    tmin=-0.2, tmax=0.7, reject=reject, preload=True)


epochs.plot_drop_log()
epochs.plot_drop_log(subject=None)
epochs.plot_drop_log(subject='')
```

Screen Shot 2021-03-12 at 16 31 51
Screen Shot 2021-03-12 at 16 32 14

As you can see, `None` simply gets converted to a string, so this doesn't work. But even if we pass an empty string, the result is not good: we still get the colon and space which separate the (now non-existent) subject ID from the percentage value of dropped epochs.

Now running the above with the changes made in this PR, the first two plots (default arguments and `subject=None`) would look like this:
Screen Shot 2021-03-12 at 16 30 52
which I believe is more desirable.

You could of course argue that one could've just special-cased `subject=''` instead of adding support for `None`, but the `None` case feels cleaner to me personally.

Does that make sense?

drammock:
sure, but what *doesn't* make sense is why 

```py
subject=self.subject if self.subject is not None else self.subject
```

is different/better than

```py
subject=self.subject
```

hoechenberger:
> sure, but what _doesn't_ make sense is why
> 
> ```python
> subject=self.subject if self.subject is not None else self.subject
> ```
> 
> is different/better than
> 
> ```python
> subject=self.subject
> ```

... oh. yes this doesn't make any sense

hoechenberger:
I suppose this is some leftover / cruft honestly. This should be simplified. I will do a followup PR to get this fixed. Thanks for catching this!

PR #9016, opened 2021-03-09
MNT: notebook conditionals in Brain & _update

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/9014#discussion_r590341165 and refactors some conditionals out of `Brain`.

It also helps with https://github.com/mne-tools/mne-python/pull/8997


larsoner:
Looks like it's improved but not complete:
```
$ git grep "if self\.notebook"
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/_brain.py:        if self.notebook:
mne/viz/_brain/callback.py:        if self.notebook:
mne/viz/_brain/callback.py:        if self.notebook:
mne/viz/_brain/mplcanvas.py:        if self.notebook:
mne/viz/_brain/mplcanvas.py:        if self.notebook:
mne/viz/backends/_pyvista.py:        if self.notebook:
```
Or are these too difficult to refactor?

GuillaumeFavelier:
I take the time to do it properly. Some of them are not obvious, for example:

https://github.com/mne-tools/mne-python/blob/0f0fa02352a864713fe6f116ba05d0ca36349bc8/mne/viz/_brain/_brain.py#L667-L678

and

https://github.com/mne-tools/mne-python/blob/0f0fa02352a864713fe6f116ba05d0ca36349bc8/mne/viz/_brain/mplcanvas.py#L32-L48

I might have an idea for `window`, `interactor` and `_screenshot`.

Those in `callback.py` are there by design. To remove them would require something like `IpyWidget` and `QtWidget` I think.

larsoner:
Looks like great progress! It also seems like the last few `if self.notebook`s in `Brain` could be pushed into the `self._renderer` methods, since they involve using either the notebook screenshot filename + notebook size or Qt file dialog + Qt interactor size.

I looked at `mplcanvas.py`, maybe the two `_renderer` flavors could also each have a `_matplotlib_canvas()` function? Then the ipy matplotlib canvas + config lives in the _notebook.py file, and the Qt5Agg canvas + config lives in the _qt.py file.

GuillaumeFavelier:
I lack inspiration to get rid of `self.interactor` completely so I tried b819d02. This can probably be improved.

larsoner:
I could 8 `.notebook`s in `_brain.py`. I do think most of them can still be refactored into the Qt and notebook files, as well as the conditionals in `mplcanvas.py` (probably by making an AbstractMplCanvas). Do you want to do it in this PR, or another PR (for you or me)?

mne/viz/backends/_notebook.py
```diff
@@ -171,6 +171,26 @@ def _tool_bar_add_text(self, name, value, placeholder):
...
         pass
 
+    def _tool_bar_add_screenshot_button(self, name, desc, func):
+        def _screenshot():
+            from PIL import Image
+            fname = self.actions[f"{name}_field"].value
+            fname = self._get_screenshot_filename() \
+                if len(fname) == 0 else fname
+            img = func()
+            Image.fromarray(img).save(fname)
```
larsoner, line 181: This code is not DRY. We already have a `_ndarray_to_fig(img).savefig(filename)` in `_pysurfer_mayavi`, maybe we should instead have a `_save_ndarray_img(fname, img)` somewhere that is used by `_notebook.py`, `_qt.py`, and `_pysurfer_mayavi.py`?

GuillaumeFavelier: I did it in d6a6fc7

GuillaumeFavelier:
I am waiting for the CIs, hopefully the failure on MacOS is just random. I tried to refactor the last bit of:

https://github.com/mne-tools/mne-python/blob/62e29ed11801e635d1694e477f0c3cd946d428b2/mne/viz/_brain/_brain.py#L659-L664

into:

https://github.com/mne-tools/mne-python/blob/98bbc62f74b34c0afac9d1a12856326f6cf51daa/mne/viz/_brain/_brain.py#L658-L665

But this part requires double-checking.

This is ready for reviews @larsoner

larsoner:
Awesome!
```
$ git grep "\.notebook"
mne/externals/tqdm/_tqdm/__init__.py:    """See tqdm.notebook.tqdm for full documentation"""
...
mne/externals/tqdm/_tqdm/notebook.py:    A shortcut for `tqdm.notebook.tqdm(xrange(*args), **kwargs)`.
mne/viz/_brain/_brain.py:        self.notebook = (_get_3d_backend() == "notebook")
mne/viz/_brain/tests/test.ipynb:    "    assert brain.notebook\n",
mne/viz/backends/_pyvista.py:        self.notebook = notebook
mne/viz/backends/_pyvista.py:        if not self.notebook:
mne/viz/backends/_pyvista.py:        if self.notebook:
mne/viz/backends/_pyvista.py:            if not self.notebook:
mne/viz/backends/_pyvista.py:            if not self.notebook and hasattr(plotter_class, 'set_icon'):
```
I would get rid of the `brain.notebook` property altogether at this point. In the `ipynb` test I would check for notebook-ness some other way (correct `isinstance` of something?)

mne/viz/backends/_qt.py
```diff
@@ -280,8 +317,158 @@ def _status_bar_add_progress_bar(self, stretch=0):
...
+        )
+        FigureCanvasQTAgg.updateGeometry(self.canvas)
+        self.manager = None
+        self.axes = self.fig.add_subplot(111)
+        self.axes.set(xlabel='Time (sec)', ylabel='Activation (AU)')
+        self.brain = brain
+        self.time_func = brain.callbacks["time"]
+        for event in ('button_press', 'motion_notify') + extra_events:
+            self.canvas.mpl_connect(
+                event + '_event', getattr(self, 'on_' + event))
```
larsoner, line 360: Not DRY. Let's refactor somehow.

For example, you could either have this code in the abstract class `__init__`, and have that `__init__` call the abstract method `self._create_mpl_canvas(...)`, which for this class does the necessary FigureCanvasQtAgg gymnastics, and for the _IpyMplCanvas does the IPython stuff

larsoner (approved):
Let's merge once CIs come back happy, thanks in advance @GuillaumeFavelier !

PR #9023, opened 2021-03-10
MRG, ENH: Slider triggers callback during dragging

hoechenberger:
The slider in the reports in `main` only triggers an update callback when the dragging action has ended. This PR makes
it such that the callback is emitted during slider movement, so one doesn't have to release the mouse button to see the results.

Also enables animations for clicking action.

This PR:

https://user-images.githubusercontent.com/2046265/110623479-2f661580-819d-11eb-86df-5c38b0cd0787.mov

`main`:

https://user-images.githubusercontent.com/2046265/110623434-1c534580-819d-11eb-8889-931c228ae1d0.mov

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -105,6 +105,8 @@ Enhancements
 
 - `~mne.Epochs` metadata can now be generated automatically from events using `mne.epochs.make_metadata` (:gh:`8834` by `Richard Höchenberger`_)
 
+- Interactions with sliders in `mne.Report` will now continuously update the linked content (it was updated only on mouse button release before) (:gh:`9023` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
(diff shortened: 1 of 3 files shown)


hoechenberger:
Artifact:
https://25684-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_report.html#adding-custom-plots-to-a-report

larsoner:
I don't see a reason to make it optional, I consider it a bug that it never worked this way (it's way more standard user interaction). I doubt it will slow anything down for people in any meaningful way. The only affect on size should be the few extra lines of JavaScript

hoechenberger:
@agramfort 
> does it affect the size of the generated html files?

If a slider is used, the images it can display will consequently always be embedded in the HTML file. This part has not changed with this PR. So it's like @larsoner said: the only change to the generated HTML this PR produces is in the JS, so I suppose we're actually **saving** a few bytes now. 👌

larsoner:
Thanks @hoechenberger !

PR #9026, opened 2021-03-11
MAINT: Disable inline comments

larsoner:
It turns out you can disable the comments by pressing "a", but given how often they are false alarms and how much noise they add, I think it's better just to disable them.


drammock (approved):

drammock:
thanks @larsoner

PR #9028, opened 2021-03-11
MAINT: Remove pyface exclusion

larsoner:
After https://github.com/conda-forge/pyface-feedstock/pull/27 we shouldn't need to avoid this version of `pyface` anymore. In fact there is no real reason to list these since the `mayavi` dependency will already pull them in. Now `environment.yml` is closer to `requirements.txt`, too, which is nice.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -18,9 +18,6 @@ dependencies:
 - joblib
 - psutil
 - numexpr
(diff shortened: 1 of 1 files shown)


agramfort (approved):
thx @larsoner

PR #9034, opened 2021-03-12
MRG, MAINT: Better notebook tests

larsoner:
- [x] Add `nbexec` fixture that can be used to run tests in an IPython kernel
- [x] Encapsulate existing tests in separate tests
- [ ] ~~Read `test_3d.py` and `test_brain.py` tests in `test_notebook.py` and use `ipytest` to run them (probably), need to figure out how to run only the notebook tests, using the notebook backend. Probably by replacing `renderer_interactive` with `renderer_notebook` using regex or so...~~


mne/viz/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+    fname = op.join(data_path(download=False),
+                    'NIRx', 'nirscout', 'nirx_15_2_recording_w_overlap')
+    raw_intensity = read_raw_nirx(fname, preload=False)
+    raw_od = optical_density(raw_intensity)
+    raw_haemo = beer_lambert_law(raw_od)
+    evts, _ = events_from_annotations(raw_haemo, event_id={'1.0': 1})
+    evts_dct = {'A': 1}
+    tn, tx = -1, 2
+    epochs = Epochs(raw_haemo, evts, event_id=evts_dct, tmin=tn, tmax=tx)
+    return epochs
```
larsoner, line 49: From here up in this file is `git mv` / old code

larsoner:
@GuillaumeFavelier can you take a look and see if you agree that `test_notebook.py` is a better testing infrastructure than `test.ipynb`? To me it will be easier to add stuff this way, but it has the disadvantage that you can't directly run the notebook.

GuillaumeFavelier:
I'm fine not running in the notebook and updating `test.ipynb` directly is not convenient at all. I'll take a look and test locally

GuillaumeFavelier:
Similarly to https://github.com/mne-tools/mne-python/pull/9190#issuecomment-808170719, the notebook tests were skipped so I pushed b08647c to fix that.

GuillaumeFavelier:
The tests are green for me locally and on Azure `Notebook` job:

```
...
mne/viz/_brain/tests/test_notebook.py::test_notebook_alignment[notebook] Coverage.py warning: --include is ignored because --source is set (include-ignored)
PASSED [ 17%]
mne/viz/_brain/tests/test_notebook.py::test_notebook_interactive[notebook] Coverage.py warning: --include is ignored because --source is set (include-ignored)
PASSED [ 17%]
mne/viz/_brain/tests/test_notebook.py::test_notebook_button_counts[notebook] Coverage.py warning: --include is ignored because --source is set (include-ignored)
PASSED [ 18%]
...
```

Is it still **WIP**? What is missing?

larsoner:
> Is it still WIP? What is missing?

Ready for merge except tests still aren't passing on Windows:
```
c:\miniconda\lib\asyncio\events.py:501: in add_reader
    raise NotImplementedError
E   NotImplementedError
```
I think we probably just need to skip if the platform is Windows because it's probably not supported in some part of the stack. But I think we can rely on it working in practice for people. I'll push a commit to skip there

larsoner:
Feel free to rebase #9190 and incorporate tests there!

PR #9040, opened 2021-03-15
DOC: Break stuff (just for demo, do not merge!)

### LINKED ISSUE #9030: Adding new connectivity indexes
The simplest way to check the relation between two signals is using correlation or a similar a similar approach. However, there are connectivity indexes that don't need to make the assumption of linearity between the interaction of the signals. These kind of indexes can be developed based on the information theory.  

New functions in the Connectivity Estimation group would be nice to be incorporated. 

Mutual information is a good option as a non-directed measure.
Transfer entropy seems to be a good option for a directed measure (causal) - although this most likely is too complex to handle myself alone  

#### Additional comments
A review about connectivity indexes
Bastos André M., Schoffelen Jan-Mathijs. A Tutorial Review of Functional Connectivity Analysis Methods and Their Interpretational Pitfalls. Frontiers in Systems Neuroscience  

Transfer entropy paper
Thomas Schreiber. Measuring Information Transfer. https://arxiv.org/abs/nlin/0001042v1

A nice article about information theory in neuroscience
Nicholas M. Timme and Christopher Lapish. A Tutorial for Information Theory in Neuroscience. Eneuro

hoechenberger:
#### Reference issue
Example: Fixes #9030.

I broke stuff.

### DIFF
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -85,7 +85,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
     Attributes
     ----------
     info : dict
-        Measurement info.
+        Measurement info. Hello!
     ch_names : list of str
         List of channels' names.
     nave : int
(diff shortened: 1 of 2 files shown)


mne/evoked.py
```diff
@@ -85,7 +85,7 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
     Attributes
     ----------
     info : dict
-        Measurement info.
+        Measurement info. Hello!
```
rob-luke, line 88: 👋

Issue #8906, opened 2021-02-23
use bibtex in plot_source_power_spectrum_opm.py


drammock:
convert references in `examples/time_frequency/plot_source_power_spectrum_opm.py` to use footcite / footbibliography

matteoanelli:
I and @timonmerk can do this one.

neurogima:
Hi, I'm going to work on this

hoechenberger:
Closed via #9050

PR #9050, opened 2021-03-15
DOC: use bibtex in plot_source_power_spectrum_opm.py (#8906)

matteoanelli:
#### Reference issue
DOC: #8906.

Use bibtex in plot_source_power_spectrum_opm.py.


examples/time_frequency/plot_source_power_spectrum_opm.py
```diff
@@ -21,6 +21,7 @@
 
 Preprocessing
 -------------
+
```
britta-wstnr, line 24: You should not add an extra line here

hoechenberger (approved):
LGTM

PR #9062, opened 2021-03-15
FIX saving fif file with the wrong extension

### LINKED ISSUE #8510: fif file extension not required when writing FIFF files
I tried a `raw.save('test.h5')` sort of by accident and while it complained that I wasn't using standard `_raw.fif` conventions it happily saved a fif file with an h5 extension. It took me awhile to realize that this wasn't actually an h5 file (I had hoped the h5 support had been extended).

I'm not sure this is a _bug_, as there isn't really the expectation that it should have written an h5 file but it was a bit confusing.

Is there a reason we don't require a `.fif` ending for files that are always written as fiff?

seems like it would be simple to use the `endings_err` parameter to `check_fname` 
https://github.com/mne-tools/mne-python/blob/1fa17ddf895b42c08aea3120d127572834ac5c96/mne/utils/check.py#L35

vagechirkov:
Fixes #8510.

Restrict saving options of the fif file to .fif and .fif.gz and add the test for this error raising.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -115,7 +115,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix bug with restricting :func:`mne.io.base.save` saving options to .fif and .fif.gz extensions (:gh:`9062` by |Valerii Chirkov|_)
+- Fix bug with restricting :func:`mne.io.Raw.save` saving options to .fif and .fif.gz extensions (:gh:`9062` by |Valerii Chirkov|_)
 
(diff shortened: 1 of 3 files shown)


larsoner:
@vagechirkov since this is a (useful) change in behavior we should probably add a line to `doc/changes/latest.inc`. Can you put a new entry in the `Bugs` section?

larsoner (approved):

agramfort (approved):
+1 for MRG when CIs are green

thx @vagechirkov

larsoner:
Local CI passes so in it goes, thanks @vagechirkov !

PR #9065, opened 2021-03-15
Ica overlay tmax documentation

### LINKED ISSUE #8987: Document correctly that plot_ica_overlay() for Raw only plots the first 3 seconds of data
See https://mne.discourse.group/t/ica-plot-overlay-pan-forward-and-missing-data/2781

sfc-neuro:
#### Reference issue
Fixes #8987.

Fixed the documentation to say tmax = None defaults to 3s

### DIFF
--- a/mne/viz/ica.py
+++ b/mne/viz/ica.py
@@ -735,7 +735,7 @@ def plot_ica_overlay(ica, inst, exclude=None, picks=None, start=None,
     start : int
         X-axis start index. If None from the beginning.
     stop : int
-        X-axis stop index. If None to the end.
+        X-axis stop index. If None to 3.0 s.
     title : str
         The figure title.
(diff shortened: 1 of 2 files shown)


mne/viz/ica.py
```diff
@@ -735,7 +735,7 @@ def plot_ica_overlay(ica, inst, exclude=None, picks=None, start=None,
     start : int
         X-axis start index. If None from the beginning.
     stop : int
```
agramfort, line 737: you should also fix the docstring at it can be int | None

same for start above.

sfc-neuro: Thanks!

agramfort (approved):

agramfort:
thx @sfc-neuro

Issue #9060, opened 2021-03-15
Time-domain asymmetric dynamical connectivity - data structure storage and basic OLS algorithms


adam2392:
#### Describe the new feature or enhancement
This is related to a goal of mine to add generic "connectivity" function to MNE for the code-sprint.

Right now, there are a decent amount of "spectral connectivity" (possibly over time) metrics derived from raw EEG/MEG/iEEG data.There are alternative characterizations, such as dynamical system perspective, which models the data as:

```
x(t+1) = Ax(t)
```

where A is possibly an asymmetric matrix characterized by its eigenvalues. Note this is different then just looking at the "pearson correlation" (or covariances) over time, since correlation matrices are PSD and symmetric.

Since a linear dynamical system representation of the MEEG data is useful, it would be nice to have a "data structure" to store such a representation and also expose basic algorithms for estimating this model from the raw data. I'm opening this issue to have a discussion to see if this is feasible to add.

#### Describe your proposed implementation
One can leverage [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html) to add an option to say `Raw`, or `Epochs`.

One can add regularization as a result. 

#### Describe possible alternatives
In a related viewpoint, this is related to what is known as Dynamic Mode Decomposition (DMD), which is also useful, but the representation of the data is different, because there one is interested in not only the `A` matrix, but also the "DMD modes".  This is related to the linear Koopman operator in dynamical systems theory; DMD is a finite approximation of Koopman.

Example paper:
- https://arxiv.org/pdf/1409.5496.pdf

#### Additional comments
Example papers using ``A`` matrix:
- https://pubmed.ncbi.nlm.nih.gov/29060480/
- https://www.biorxiv.org/content/10.1101/862797v4.full

In addition https://arxiv.org/pdf/1802.08334.pdf talks about the convergence of such OLS estimates for A, giving some sort of theoretical guarantees on how "estimable" an A matrix is from MEEG data.

#### Comparisons to spectral connectivity
Spectral connectivity generally are defined as some correlation measure over frequency domain, and is generally symmetric. Here, I am proposing a ``time-domain`` and possibly ``asymmetric` "connectivity" derivative of the raw MEEG data.

adam2392:
@alexrockhill

alexrockhill:
Is there a replication we can do from one of these paper with real data available to make sure it works properly?

adam2392:
[...]

PR #9076, opened 2021-03-15
[WIP] Enhancement of iEEG datasets with a real-clinical seizure Ecog dataset

adam2392:
#### Reference issue
Addresses part of #9060 

In collaboration with @alexrockhill 

Implements a `mne.datasets.epilepsy` dataset directory for housing "clinically" relevant datasets. In this case, there is a subject ECoG that has a seizure midway. Adds `sub-pt01` from https://openneuro.org/datasets/ds003029.

Once this is added, then we can provide the dataset to @larsoner who can upload to OSF. Then a tutorial can be rendered showing e.g. the gamma band power during seizure onset.

#### Next Steps
In a next PR, we intend on demonstrating how to compute a linear dynamical system from this dataset.

#### Additional information
Waiting to add also a tutorial.


alexrockhill:
If these were in `fif`, then the tutorial would be simplified as far as reading in channel locations etc. ...

larsoner:
Yes also but I think there is value in going from the native format. People who collect the data will have to figure out how to manipulate their data, and it's nice to have the example show that part a bit.

alexrockhill:
Like cross-reference how to load in brainvision files ;) I guess I'll write it keeping it in brainvision

It's just not very high-level because `mne-bids` isn't a dependency. Is that okay to add to have `mne-bids` parse the channel locations for you or no?

larsoner:
> It's just not very high-level because mne-bids isn't a dependency.

I think it's okay to add it as a dependency just needed for the example (i.e., not a dependency for MNE-Python in environment.yml or requirements.txt) if it simplifies things. We just add it to `requirements_doc.txt` and I think CircleCI should even deal with it properly.

adam2392:
@larsoner is the storage of this dataset as is what you need from us? https://www.dropbox.com/sh/bsufh5npukzqe59/AAC4G3RilD-sSA3jH1SL_qhpa?dl=0

Should I delete everything except the compressed file? Should I rename anything accordingly?

larsoner:
I can upload `MNE-Clinical.zip` (probably after putting it in a `MNE-Clinical-data` dir and recompressing as `.tar.gz`) to OSF if that's what you need for the example

adam2392:
> I can upload `MNE-Clinical.zip` (probably after putting it in a `MNE-Clinical-data` dir and recompressing as `.tar.gz`) to OSF if that's what you need for the example

I think this is almost ready: https://www.dropbox.com/sh/bsufh5npukzqe59/AAC4G3RilD-sSA3jH1SL_qhpa?dl=0

@alexrockhill working on some issues related to our alignment (these electrode coordinates might be crap :/) in https://github.com/mne-tools/mne-python/pull/9087, and afterwards we can upload.

That PR can be the main one to merge.

alexrockhill:
Is this good to go then?

adam2392:
@larsoner i just subbed in the final changes in the epilepsy dataset. Are you able to upload to osf now? 

Did you want us to rename this to like epilepsy-seizure? 

We were thinking we can add additional epilepsy datasets through openneuro in the future.

larsoner:
URL: https://osf.io/z4epq/download?revision=1
MD5: ffb139174afa0f71ec98adbbb1729dea

larsoner:
(renamed `MNE-epilepsy-ecog-data`)

larsoner:
Closing in favor of #9087 where we'll update the paths, etc.

PR #9079, opened 2021-03-15
DOC Bibtex fix for spatial_resolution.py

### LINKED ISSUE #8925: use bibtex in spatial_resolution.py
convert references in `mne/minimum_norm/spatial_resolution.py` to use footcite / footbibliography

NeuroLaunch:
#### Reference issue
Example: Fixes #8925.

Adds footcite and footbibliography support for `resolution_metrics()` in _minimum_norm/spatial_resolution.py_.

#### Collaborator
@cmista

### DIFF
--- a/mne/minimum_norm/spatial_resolution.py
+++ b/mne/minimum_norm/spatial_resolution.py
@@ -68,11 +68,11 @@ def resolution_metrics(resmat, src, function='psf', metric='peak_err',
     For details, see :footcite:`MolinsEtAl2008` :footcite:`HaukEtAl2019`.
 
     .. versionadded:: 0.20
-    
+
     References
     ----------
     .. footbibliography::
-    """    
+    """
     # Check if input options are valid
     metrics = ('peak_err', 'cog_err', 'sd_ext', 'maxrad_ext', 'peak_amp',
                'sum_amp')
@@ -327,8 +327,3 @@ def _rectify_resolution_matrix(resmat):
                     (shape[0], shape[1], resmat.shape[0], resmat.shape[1]))
 
     return resmat
-
-###############################################################################
-# References
-# ----------
-# .. footbibliography::
--- a/mne/minimum_norm/spatial_resolution.py
+++ b/mne/minimum_norm/spatial_resolution.py
@@ -65,19 +65,10 @@ def resolution_metrics(resmat, src, function='psf', metric='peak_err',
 
     Notes
     -----
-    For details, see [1]_ [2]_.
+    For details, see :footcite:`MolinsEtAl2008` :footcite:`HaukEtAl2019`.
 
     .. versionadded:: 0.20
-
-    References
-    ----------
-    .. [1] Molins A, Stufflebeam S M, Brown E N, Hämäläinen M S (2008).
-           Quantification of the benefit from integrating MEG and EEG data in
-           minimum l2-norm estimation. Neuroimage, 42(3):1069-77.
-    .. [2] Hauk O, Stenroos M, Treder M (2019). "Towards an Objective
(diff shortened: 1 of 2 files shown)


mne/minimum_norm/spatial_resolution.py
```diff
@@ -65,19 +65,14 @@ def resolution_metrics(resmat, src, function='psf', metric='peak_err',
 
     Notes
     -----
-    For details, see [1]_ [2]_.
+    For details, see :footcite:`MolinsEtAl2008` :footcite:`HaukEtAl2019`.
```
sappelhoff, line 68: the footcite directive is documented here: https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#role-footcite

I think you could do something like this here:

```suggestion
    For details, see `:footcite:`MolinsEtAl2008,HaukEtAl2019`.
```

mne/minimum_norm/spatial_resolution.py
```diff
@@ -65,19 +65,14 @@ def resolution_metrics(resmat, src, function='psf', metric='peak_err',
...
 
     .. versionadded:: 0.20
-
+
```
sappelhoff, line 71: you added some trailing whitespace here: That's undesirable (invisible, useless characters) --> a good solution would be to look into your editor settings. For most editors you can have an automatic removal of trailing whitespace :-)

mne/minimum_norm/spatial_resolution.py
```diff
@@ -332,3 +327,8 @@ def _rectify_resolution_matrix(resmat):
                     (shape[0], shape[1], resmat.shape[0], resmat.shape[1]))
 
     return resmat
+
+###############################################################################
+# References
+# ----------
+# .. footbibliography::
```
sappelhoff, line 334: why is this necessary?

drammock: agree, this is not needed.  in a *code* file the references section goes within the docstring where the reference is cited.  In a *tutorial* file you want something like this just once at the very end.

NeuroLaunch: That was a remnant, so we'll get that out.

mne/minimum_norm/spatial_resolution.py
NeuroLaunch: Just saw this, we'll update it on next commit.

drammock (approved):

rob-luke (changes requested):
I think two references were missed at 

https://github.com/mne-tools/mne-python/blob/dbdf70f280bf7ae93a61b0c3248bcb7d15aba0a4/mne/minimum_norm/spatial_resolution.py#L47

NeuroLaunch:
> 
> 
> I think two references were missed at
> 
> https://github.com/mne-tools/mne-python/blob/dbdf70f280bf7ae93a61b0c3248bcb7d15aba0a4/mne/minimum_norm/spatial_resolution.py#L47

Indeed they were (we text searched, but not for dual citations like that).  I'll fix that now and re-push.

rob-luke (approved):

sappelhoff:
@NeuroLaunch your commits look like your `.gitconfig` is not configured :thinking: 

Can you take a look at `~/.gitconfig` (if you are on Mac or Linux) on Windows, this is probably `C:\Users\username\.gitconfig`

The first lines in the file should look somewhat like this:

```
[user]
        email = stefan.appelhoff@mailbox.org
        name = Stefan Appelhoff
```

If you use an email that you also used to register in GitHub, the GitHub User Interface will be better able to link your contributions to your account.

NeuroLaunch:
Good catch, @sappelhoff !  I had a typo.  It explains some odd behavior with my authentication (despite my ssh setup).

PR #9080, opened 2021-03-15
Correct references in hf_sef.py and limo.py #8916 #9080

### LINKED ISSUE #8916: use bibtex in hf_sef.py
convert references in `mne/datasets/hf_sef/hf_sef.py` to use footcite / footbibliography

nordme:
Fixes #8916. Updates reference in hf_sef.py


mne/datasets/hf_sef/hf_sef.py
```diff
@@ -16,7 +16,8 @@ def data_path(dataset='evoked', path=None, force_update=False,
               update_path=True, verbose=None):
     u"""Get path to local copy of the high frequency SEF dataset.
 
-    Gets a local copy of the high frequency SEF MEG dataset [1]_.
+    Gets a local copy of the high frequency SEF MEG dataset
+    :footcite: `NurminenEtAl2017`.
```
sappelhoff, line 20: there is a space here that shouldn't be there:
```suggestion
    :footcite:`NurminenEtAl2017`.
```

mne/datasets/hf_sef/hf_sef.py
```diff
@@ -42,10 +43,6 @@ def data_path(dataset='evoked', path=None, force_update=False,
     path : str
         Local path to the directory where the HF-SEF data is stored.
 
-    References
-    ----------
-    .. [1] Nurminen, J., Paananen, H., Mäkelä, J. (2017): High frequency
-           somatosensory MEG dataset. https://doi.org/10.5281/zenodo.889234
```
sappelhoff, line 48: I think this should rather be 
```
References
----------
.. footbibliography::
```

mne/datasets/hf_sef/hf_sef.py
```diff
@@ -98,3 +95,9 @@ def data_path(dataset='evoked', path=None, force_update=False,
 
     _do_path_update(path, update_path, key, name)
     return destdir
+
+###############################################################################
+# References
+# ----------
+#
+# .. footbibliography::
```
sappelhoff, line 103: I think this needs to be removed and should be reflected in lines 45-48

mne/datasets/limo/limo.py
```diff
@@ -242,3 +236,9 @@ def load_data(subject, path=None, force_update=False, update_path=None,
     epochs.info['bads'] = missing_chans  # missing channels are marked as bad.
 
     return epochs
+
+###############################################################################
+# References
+# ----------
+#
+# .. footbibliography::
```
sappelhoff, line 244: same comment as above

mne/datasets/limo/limo.py
```diff
@@ -86,12 +86,6 @@ def data_path(subject, path=None, force_update=False, update_path=None,
     and prompt the user to save the 'datasets' path to the mne-python config,
     if it isn't there already.
 
-    References
-    ----------
-    .. [1] Guillaume, Rousselet. (2016). LIMO EEG Dataset, [dataset].
-           University of Edinburgh, Centre for Clinical Brain Sciences.
-           https://doi.org/10.7488/ds/1556.
-    .. [2] https://datashare.is.ed.ac.uk/handle/10283/2189?show=full
```
sappelhoff, line 94: same comment as above :)

sappelhoff:
One more comment: You could consider editing the title of your PR. Currently it's `Solve #8916`, which doesn't tell a reviewer much --> at least not unless there is a reviewer who exactly memorizes each issue by its number :wink: 

What works better is to be as explicit as possible in your PR titles. For example:

`DOC: Use footcite in hf_sef.py  and limo.py`

That way reviewers know that it's about documentation, it's about 2 files, and it's about using a sphinx feature to cite refs.

mne/datasets/hf_sef/hf_sef.py
drammock: actually I think this is correct.  We want the references moved to the bottom of the example (previously they were at the top, which is non-standard)

mne/datasets/limo/limo.py
drammock: this too is correct.

mne/datasets/hf_sef/hf_sef.py
sappelhoff: Maybe I don't get it ... is this file an example? Looks like a normal .py module file to me

larsoner: Indeed this is a standard module file so I think it should be moved up to the docstring

drammock: ah, oops.  I didn't look closely enough.

nordme:
Thanks for the review!! @sappelhoff Hopefully this last commit should fix the references to appear at the bottom of the docstrings.

larsoner:
Something tragic happened with the commit history here :) Can you look @nordme ?

sappelhoff:
yup, looks like a familiar problem :| (merging some commits in here that don't belong here) @nordme let me know if you need help

sappelhoff:
I think you should go back to the last time this branch was in order. As far as I can see in the [history](https://github.com/mne-tools/mne-python/pull/9080/commits), this is the commit: 123ece68f8b9aaf8620df9b2f41454d64031cc86

Following these instructions may be good: https://stackoverflow.com/a/21718540/5201771

Afterwards you checkout your `main` branch and bring it up to date ... to then merge the master branch into your `fix_2` branch (the branch with which you opened this PR)

nordme:
Ok, should be all better!! Lmk if there are any other comments. Thanks, @sappelhoff

sappelhoff (approved):

drammock (approved):

drammock:
thanks @nordme!

PR #9087, opened 2021-03-15
[MRG, ENH] Update plot_ecog tutorial to use real data, visualize seizures

alexrockhill:
#### Reference issue
Example: Addresses part of #9060.

Show seizure data from patient using John's Hopkins epilepsy monitoring unit data.

#### Additional information
cc @libertyh @adam2392


mne/datasets/__init__.py
```diff
@@ -25,6 +25,7 @@
 from . import refmeg_noise
 from . import ssvep
 from . import erp_core
+from . import epilepsy
```
agramfort, line 28: I fear that epilepsy is too generic. Can you add ecog to the name? unless we think we will put in this dataset all epilepsy related data?

tutorials/misc/plot_ecog.py
```diff
@@ -183,6 +157,87 @@ def animate(i, activity):
...
+# matter and there is a true 0 point.
+cmap = 'RdBu_r'
+
+raw_notched = raw.copy().notch_filter([60, 120])
+ts_data = raw_notched.get_data()
+# Create a dictionary of seizure-related events from raw.annotations
+# These will be used to update the animation and show the most relevant
+# annotation
+events = dict()
+for t in np.arange(len(raw.annotations)):
```
larsoner, line 186: This looks a lot like [mne.events_from_annotations](https://mne.tools/stable/generated/mne.events_from_annotations.html)

tutorials/misc/plot_ecog.py
```diff
@@ -183,6 +157,87 @@ def animate(i, activity):
...
+# to seizure onset that are marked in the annotations. This will be done
+# very similarly to the high gamma visualization
+
+# For this plot we will use a different angle (from left side)
+fig = plot_alignment(raw.info, subject='fsaverage', subjects_dir=subjects_dir,
+                     surfaces=['pial'], trans=None, coord_frame='head')
+mne.viz.set_3d_view(fig, -360, 90)
+
+xy, im = snapshot_brain_montage(fig, raw.info)
+
```
larsoner, line 171: We should separate these into two sections so that the animation lands in its own independent section
```suggestion
###############################################################################

```

larsoner (commented):
Just a couple of preliminary comments

alexrockhill:
Ok, I think this is okay for review now

tutorials/misc/plot_ecog.py
```diff
@@ -183,20 +166,118 @@ def animate(i, activity):
...
+# the title. The seizure eventually becomes generalized and then subsides.
+
+# colormap to view spectral power - RdBu_r works well for directional
+# data such as voltage time series, where positive and negative values
+# matter and there is a true 0 point
+cmap = 'RdBu_r'
+
+# Apply a notch filter to remove 60 Hz line noise and harmonics
+raw_notched = raw.copy().pick_types(
+    meg=False, ecog=True).notch_filter([60, 120])
```
drammock, line 185: after @larsoner's addition of downsampling, this would need to change (otherwise second notch is above Nyquist), but it looks like he added a notch filter up above anyway so we shouldn't need it here at all

```suggestion
    meg=False, ecog=True)
```

tutorials/misc/plot_ecog.py
```diff
@@ -11,102 +11,81 @@
...
+bids_path = BIDSPath(root=bids_root, subject='pt1', session='presurgery',
+                     task='ictal', datatype='ieeg', extension='vhdr')
 
-montage = mne.channels.make_dig_montage(
-    ch_pos, coord_frame='mri', nasion=nasion, lpa=lpa, rpa=rpa)
-print('Created %s channel positions' % len(ch_names))
+# then we'll use it to load in the sample dataset
+# Here we use a format (iEEG) that is only available in MNE-BIDS 0.7+, so it
+# will emit a warning on versions <= 0.6
+raw = read_raw_bids(bids_path=bids_path, verbose=False)
```
drammock, line 59: to get it to run locally I still need
```py
with warnings.catch_warnings(record=True):
    warnings.simplefilter("ignore")
    raw = read_raw_bids(bids_path=bids_path, verbose=False)
```
but @larsoner's changes should make it work on CircleCI now

tutorials/misc/plot_ecog.py
```diff
@@ -183,20 +166,118 @@ def animate(i, activity):
...
+raw_notched = raw.copy().pick_types(
+    meg=False, ecog=True).notch_filter([60, 120])
+
+# Apply a gentle high pass filter to get rid of drift
+raw_notched.filter(l_freq=0.1, h_freq=None)
+
+# Downsample the data to make the animation more reasonable
+sfreq = 50  # Hz
+raw_notched.resample(sfreq)
+ts_data = raw_notched.get_data()
```
drammock, line 193: @larsoner added an initial downsample to 200 Hz, but I see here you're downsampling to 50.  Could we just go straight to 50 up above instead of doing it twice?

tutorials/misc/plot_ecog.py
```diff
@@ -183,20 +166,118 @@ def animate(i, activity):
...
+# We want the mid-point (0 uV) to be white, so we will scale from -vmax to vmax
+# so that negative voltages are blue and positive voltages are red
+ts_data = raw.get_data()
+vmax = np.percentile(ts_data, 90)
+paths = ax.scatter(*xy_pts.T, c=np.zeros(len(xy_pts)), s=40,
+                   cmap=cmap, vmin=-vmax, vmax=vmax)
+ax.set_xlim([0, im.shape[0]])
+ax.set_ylim([im.shape[1], 0])
+tsl = slice(start_sample - sfreq, start_sample + sfreq)
+tpaths = tax.plot(raw_notched.times[tsl], ts_data[:, tsl].T)
```
drammock, line 245: after fixing the notch filter freqs I'm still getting errors on this line when running locally: 

```
  File "/opt/mne/python/tutorials/misc/plot_ecog.py", line 248, in <module>
    tpaths = tax.plot(raw_notched.times[tsl], ts_data[:, tsl].T)
  File "/opt/miniconda3/envs/mnedev/lib/python3.8/site-packages/matplotlib/axes/_axes.py", line 1743, in plot
    lines = [*self._get_lines(*args, data=data, **kwargs)]
  File "/opt/miniconda3/envs/mnedev/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 273, in __call__
    yield from self._plot_args(this, kwargs)
  File "/opt/miniconda3/envs/mnedev/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 399, in _plot_args
    raise ValueError(f"x and y must have same first dimension, but "
ValueError: x and y must have same first dimension, but have shapes (0,) and (100, 86)
```

tutorials/misc/plot_ecog.py
adam2392: We wanted to characterize the "gamma band" power, which has frequencies from "30-90" traditionally. If still doesn't make sense to do it here, let us know.

tutorials/misc/plot_ecog.py
adam2392: Used the wrong `raw` vs `raw_notched`. It should be fixed now.

agramfort:
can you see why doc build fails? can you share when you have it the URL to the rendered artifact? thx @alexrockhill !

adam2392:
Rn it's getting hit with the following error:

```
ERROR: HTTP error 404 while getting https://github.com/mne-tools/mne-bids/archive/main.zip
ERROR: Could not install requirement https://github.com/mne-tools/mne-bids/archive/main.zip (from -r requirements_doc.txt (line 10)) because of HTTP error 404 Client Error: Not Found for url: https://codeload.github.com/mne-tools/mne-bids/zip/main for URL https://github.com/mne-tools/mne-bids/archive/main.zip
```

tutorials/misc/plot_ecog.py
```diff
@@ -183,20 +166,118 @@ def animate(i, activity):
...
+sl = slice(start_sample, ts_data.shape[1], 10)
 
 evoked = mne.EvokedArray(
     gamma_power_t[:, sl], raw.info, tmin=raw.times[sl][0])
-stc = mne.stc_near_sensors(evoked, trans, subject, subjects_dir=subjects_dir)
+src = mne.read_source_spaces(
+    op.join(subjects_dir, 'fsaverage', 'bem', 'fsaverage-ico-5-src.fif'))
+trans = None  # identity transform
+stc = mne.stc_near_sensors(evoked, trans, 'fsaverage', src=src,
+                           subjects_dir=subjects_dir)
```
drammock, line 274: running locally now I see on this line:
```
  File "<decorator-gen-283>", line 24, in stc_near_sensors
  File "/opt/mne/python/mne/source_estimate.py", line 3301, in stc_near_sensors
    evoked.drop_channels(nan_chs)
  File "/opt/mne/python/mne/channels/channels.py", line 920, in drop_channels
    return self._pick_drop_channels(idx)
  File "/opt/mne/python/mne/channels/channels.py", line 943, in _pick_drop_channels
    pick_info(self.info, idx, copy=False)
  File "<decorator-gen-8>", line 24, in pick_info
  File "/opt/mne/python/mne/io/pick.py", line 539, in pick_info
    raise ValueError('No channels match the selection.')
ValueError: No channels match the selection.
```

tutorials/misc/plot_ecog.py
drammock: ah, right, so up above we need at least 200 Hz sfreq.  And I assume that's still too high for the animation to work smoothly / compute in a reasonable amount of time?  In that case let's just change the code comment above (see other comment)

tutorials/misc/plot_ecog.py
```diff
@@ -183,20 +166,118 @@ def animate(i, activity):
...
+# matter and there is a true 0 point
+cmap = 'RdBu_r'
+
+# Get a copy of the filtered data earlier and only get the ecog channels
+raw_notched = raw.copy().pick_types(meg=False, ecog=True)
+
+# Apply a gentle high pass filter to get rid of drift
+raw_notched.filter(l_freq=0.1, h_freq=None)
+
+# Downsample the data to make the animation more reasonable
```
drammock, line 189:
```suggestion
# Downsample again, to compute the animation faster
```

adam2392:
> Rn it's getting hit with the following error:
> 
> ```
> ERROR: HTTP error 404 while getting https://github.com/mne-tools/mne-bids/archive/main.zip
> ERROR: Could not install requirement https://github.com/mne-tools/mne-bids/archive/main.zip (from -r requirements_doc.txt (line 10)) because of HTTP error 404 Client Error: Not Found for url: https://codeload.github.com/mne-tools/mne-bids/zip/main for URL https://github.com/mne-tools/mne-bids/archive/main.zip
> ```

According to this: https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch Not sure why this would raise an error. Any advice @drammock ?

tutorials/misc/plot_ecog.py
adam2392: I don't actually get that error... I somehow get to this, but I just assumed that was cuz my Pycharm IDE is glitchy showing all the figures:

```
  File "/Users/adam2392/Documents/mne-python/tutorials/misc/plot_ecog.py", line 279, in <module>
    brain.show_view(azimuth=-20, elevation=60)
TypeError: show_view() got an unexpected keyword argument 'azimuth'
```

alexrockhill: Hmm, this worked for me last time, but double checking now

alexrockhill:
The `requirement-doc.txt` requirements won't work because we use the 0.7dev of mne-bids right? Maybe we have to wait to merge for a new release?

larsoner:
CircleCI timed out when rendering:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/7050/workflows/f8b16fdd-5f2b-40c3-925d-095038002124/jobs/26076

This means either memory usage was too high or it took too long. Locally `PATTERN=plot_ecog make html_dev-pattern` should complete your example running within about 30 seconds, and with < 1.5 GB of memory used. Let me know if you want help optimizing this

tutorials/misc/plot_ecog.py
libertyh: Right, to compute gamma band power we need the higher sampling rate, but the point of downsampling here was because of the animation. I suppose this could happen for the gamma power data too, once it's computed, and that could save space/memory. However, we should not downsample to 50 Hz *before* computing the gamma power.

libertyh:
> CircleCI timed out when rendering:
> 
> https://app.circleci.com/pipelines/github/mne-tools/mne-python/7050/workflows/f8b16fdd-5f2b-40c3-925d-095038002124/jobs/26076
> 
> This means either memory usage was too high or it took too long. Locally `PATTERN=plot_ecog make html_dev-pattern` should complete your example running within about 30 seconds, and with < 1.5 GB of memory used. Let me know if you want help optimizing this

Is this because the whole seizure animation itself is quite long? The seizure is longer than 30 seconds in real time, so either we'd have to chop it and make it shorter or significantly downsample (which would also make it harder to see). I'm not 100% positive this is the issue but I could imagine that that's it.

larsoner:
I suggested to @alexrockhill just to show a little bit of time before onset, and a little after (maybe 5 before, 5 after). Do you think that would work @libertyh ?

tutorials/misc/plot_ecog.py
alexrockhill: Fixed it

libertyh:
> I suggested to @alexrockhill just to show a little bit of time before onset, and a little after (maybe 5 before, 5 after). Do you think that would work @libertyh ?

That would work for reducing memory, but in looking at the clinical seizure pattern, there are a lot of things that are interesting at longer time scales. For example, near the onset (https://snipboard.io/QkfacP.jpg), middle (https://snipboard.io/JRPXpV.jpg), and offset (https://snipboard.io/uyHAhO.jpg) of the seizure, the patterns look very different. Perhaps long term it would be interesting to have an input or similar for the user to specify a time onset and offset, and provide some suggestions for these time points? 

That's not to say this isn't a good solution for now, but I do think for people involved in seizure localization research it's important to see how the pattern changes across space and time, and 10 seconds isn't enough for this particular dataset.

larsoner:
Makes sense. Unfortunately I think we'll kill CircleCI if we try to do all of this in the example :) Maybe keeping the animations short and copy-pasting/adapting the explanatory text above into the example in some form would work. The idea I think can be just to give people an idea of the ways they can look at the data here, and say in practice they should look at longer stretches of time and/or different segments of data (and probably with less downsampling, etc.).

alexrockhill:
Yeah it just seems like this is too much for the docs to handle with it actually being useful but at least the dataset is up there and users can change the code and actually run it if they want to

larsoner:
FWIW I don't think this error was on the brain plot but rather on the animations before that, because it died in the `matplotlib_scraper`, not the `brain_scraper`

alexrockhill:
> FWIW I don't think this error was on the brain plot but rather on the animations before that, because it died in the `matplotlib_scraper`, not the `brain_scraper`

That one is two seconds at 50 Hz so I don't think it can get any shorter without being cut completely...

larsoner:
> That one is two seconds at 50 Hz so I don't think it can get any shorter without being cut completely...

I wonder if having 100 frames is problematic for creating a GIF for example. This run:
```
    - ../tutorials/misc/plot_ecog.py:                                                                  258.97 sec   727.5 MB
```
is painfully long but maybe we have to live with it :) Let's see how it looks when it finishes rendering!

larsoner:
Locally when trying to run your example I get:
```
Animation size has reached 31485150 bytes, exceeding the limit of 31457280.0. If you're sure you want a larger animation embedded, set the animation.embed_limit rc parameter to a larger value (in MB). This and further frames will be dropped.
WARNING: /home/larsoner/python/mne-python/tutorials/misc/plot_ecog.py failed to execute correctly: Traceback (most recent call last):
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/scrapers.py", line 333, in save_figures
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/scrapers.py", line 131, in matplotlib_scraper
    image_rsts.append(_anim_rst(anim, image_path, gallery_conf))
  File "/home/larsoner/python/sphinx-gallery/sphinx_gallery/scrapers.py", line 190, in _anim_rst
    anim.save(image_path, writer=writer, dpi=use_dpi)
  File "/home/larsoner/python/matplotlib/lib/matplotlib/animation.py", line 1170, in save
    for data in zip(*[a.new_saved_frame_seq() for a in all_anim]):
  File "/home/larsoner/python/matplotlib/lib/matplotlib/animation.py", line 1170, in <listcomp>
    for data in zip(*[a.new_saved_frame_seq() for a in all_anim]):
  File "/home/larsoner/python/matplotlib/lib/matplotlib/animation.py", line 1726, in new_saved_frame_seq
    return itertools.islice(self.new_frame_seq(), self.save_count)
ValueError: Indices for islice() must be None or an integer: 0 <= x <= sys.maxsize.
```
Something must be wrong with the number of samples being requested... want me to push a commit with an attempted fix?

tutorials/misc/plot_ecog.py
```diff
@@ -169,34 +151,133 @@ def animate(i, activity):
...
+tax.plot([sfreq, sfreq], [ts_data.min(), ts_data.max()], color='k')
+fig.colorbar(paths, ax=ax)
+title = ax.set_title('iEEG voltage over time', size='large')
+
+# this will be a much longer animation because the seizure is quite long
+sl = slice(start_sample, ts_data.shape[1])
+show_power = ts_data[:, sl]
+anim = animation.FuncAnimation(fig, animate, init_func=init,
+                               fargs=(show_power, events),
+                               frames=show_power.shape[1] - sfreq,
```
larsoner: This is the problem, it converts `frames` from an int to a float and for some reason the behaviors are different. Maybe it means "seconds" when you use it that way? I don't see it in their docs, but changing it to an int seemed to fix it

adam2392:
Got hit w/ a weird error in build-docs:

```
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/scrapers.py", line 333, in save_figures
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/scrapers.py", line 131, in matplotlib_scraper
    image_rsts.append(_anim_rst(anim, image_path, gallery_conf))
  File "/home/circleci/.local/lib/python3.8/site-packages/sphinx_gallery/scrapers.py", line 190, in _anim_rst
    anim.save(image_path, writer=writer, dpi=use_dpi)
  File "/home/circleci/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 1161, in save
    writer.grab_frame(**savefig_kwargs)
  File "/usr/local/lib/python3.8/contextlib.py", line 120, in __exit__
    next(self.gen)
  File "/home/circleci/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 253, in saving
    self.finish()
  File "/home/circleci/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 354, in finish
    self.cleanup()
  File "/home/circleci/.local/lib/python3.8/site-packages/matplotlib/animation.py", line 390, in cleanup
    raise subprocess.CalledProcessError(
subprocess.CalledProcessError: Command '['convert', '-size', '110x110', '-depth', '8', '-delay', '2.0', '-loop', '0', 'rgba:-', '/home/circleci/project/doc/auto_tutorials/misc/images/sphx_glr_plot_ecog_004.gif']' returned non-zero exit status 1.
```

larsoner:
I don't really understand this since locally I can't reproduce the problem. I might try running CircleCI with SSH to see if I can fix it. I'll try to do it tonight sometime...

larsoner:
Turns out it was dying (I think) because one of the animations had no frames, and it just gave that cryptic error about no files found. I pushed a commit that allows things to build for me locally, hopefully it works with CircleCI this time. It's still not as fast as I'd like (~1.5 min on my laptop) but might be good enough given that it produces three animations (most time is spent in the raw time one I think).

larsoner:
Okay it at least runs now:

https://26177-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html

@adam2392 @alexrockhill @libertyh WDYT?

adam2392:
Wow looks awesome. The only thing I can think of is that the last two plots are super fast. If we amp up the delay by a bit between frames would that hit the CI too hard?

tutorials/misc/plot_ecog.py
```diff
@@ -139,64 +141,159 @@
...
+    event_idx = np.where(events[:, 0] - raw.first_samp == center)[0]
+    if len(event_idx):
+        title.set_text(inv_event_id[events[event_idx[0], 2]])
+    return artists
+
+
+anim = animation.FuncAnimation(
+    fig, animate_raw, init_func=lambda: artists,
+    frames=ts_data.shape[1] - 2 * half_width - 1, blit=True,
+    interval=1000 / sfreq)
```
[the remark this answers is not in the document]
adam2392: Trying now with half the framerate

tutorials/misc/plot_ecog.py
```diff
@@ -139,64 +141,159 @@
                                 ['Gamma', 'Alpha']):
     ax.imshow(im)
     ax.set_axis_off()
-    sc = ax.scatter(*xy_pts.T, c=band_power, s=200,
+    sc = ax.scatter(*xy_pts.T, c=band_power, s=100,
                     cmap=cmap, vmin=vmin, vmax=vmax)
     ax.set_title(f'{band} band power', size='x-large')
+    ax.set_xlim([0, im.shape[0]])
+    ax.set_ylim([im.shape[1], 0])
 fig.colorbar(sc, ax=axs)
```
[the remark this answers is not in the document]
adam2392: Done

tutorials/misc/plot_ecog.py
```diff
@@ -11,114 +11,116 @@
...
+# To make the example run much faster, we will start 1 seconds before the
+# seizure onset event and use 4 seconds of seizure
+onset_events = events[events[:, 2] == event_id['onset']]
+start = (onset_events[0, 0] - raw.first_samp) / raw.info['sfreq']
+raw.crop(start - 0.5, start + 3.5)
 
-###############################################################################
-# Now that we have our montage, we can load in our corresponding
-# time-series data and set the montage to the raw data.
+# And then downsample. This is just to save time in this example, you should
```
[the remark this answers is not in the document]
adam2392: Done I think? :p

tutorials/misc/plot_ecog.py
```diff
@@ -11,114 +11,116 @@
...
+from mne_bids import BIDSPath, read_raw_bids
 
 import mne
 from mne.viz import plot_alignment, snapshot_brain_montage
 
 print(__doc__)
 
 # paths to mne datasets - sample ECoG and FreeSurfer subject
-misc_path = mne.datasets.misc.data_path()
+bids_root = mne.datasets.epilepsy_ecog.data_path()
```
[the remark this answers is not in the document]
adam2392: Done

alexrockhill:
Wow, thanks for all the improvements @larsoner! And of course for getting it to pass Circle! I had a think about the first two animations and had an idea to try, we can just fix what works too but I'll mock it up if anyone wants chat.

alexrockhill:
What do you think? https://26243-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html?highlight=plot_ecog @libertyh, I iterated off of the ideas you started, does it look okay? The cropping looks weird and didn't look that way when I ran it locally but I'll see if I can fix it

adam2392:
https://26245-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html#sphx-glr-auto-tutorials-misc-plot-ecog-py

cropping looks better now. I fixed some headers in the docs.

jasmainak:
animations look cool. But can we pause one moment to ask how does this example help someone new to MNE interested in analyzing sEEG be bombarded with so much code. I've seen this same syndrome in the stats module and I see folks struggling all the time. Can we simplify? Less code, more interleaving of text and code.

What does the TFR plot tell you?

Will there be similar big code blocks also for sEEG example? Can we develop something that's a bit more generalizable rather than hacking MNE code to work for one dataset?

alexrockhill:
> animations look cool. But can we pause one moment to ask how does this example help someone new to MNE interested in analyzing sEEG/ECoG be bombarded with so much code. I've seen this same syndrome in the stats module and I see folks struggling all the time. Can we simplify? Less code, more interleaving of text and code.
> 
> What does the TFR plot tell you?
> 
> Will there be similar big code blocks also for sEEG example? Can we develop something that's a bit more generalizable rather than hacking MNE code to work for one dataset?

It's a bit hacked together and we were looking at using layout but it wasn't easy to encapsulate some of this code within the existing mne layout code. I think this is a good start and in future PRs we can pull in more code into generalized functions but that's just my opinion, if you want to abstract specific parts, I'd be happy to give it a shot.

As for the balance of code and comments, I agree that it would be nice to have less code but I'm not sure it's the best to put in the code base right now, it could probably be refactored to use layout. I would say that the plots are very informative, maybe more in a copy paste kind of way for the code, because it does show the activity pretty clearly and because the views are idiosyncratic to the surgical plan, I think it has to be a bit code heavy. I tried to pare it down but still captures the essence of the seizure and its relation to the anatomy so I think that is very informative.

jasmainak:
okay to do this iteratively in future PRs!

jasmainak:
> What does the TFR plot tell you?

But what about this? :)

alexrockhill:
https://26267-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html

larsoner:
From talking to @alexrockhill I propose:

1. Roll back to the commit that I had with the `scatter` animation with gamma power, time-traces, and brain plot
2. Remove the time trace one
3. @alexrockhill makes a PR to add support to `plot_topo` or something similar to make his cool trace-over-brain-image plot
4. in that same PR, he could add that trace-over-brain plot to the ECoG example, with animation/interactive updating

That way what we merge into `main` now is much smaller (to make @jasmainak happier) and we still soon get nice time-overlay-plotting support. Okay with you @jasmainak @adam2392 @libertyh ?

jasmainak:
sounds like a great plan. I'd say it will also make your future user happy, not just @jasmainak ;-)

tutorials/misc/plot_ecog.py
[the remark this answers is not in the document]
adam2392: Oh oops only added to `datasets_reference.rst`. Will add there too.

larsoner:
@alexrockhill do you want to push a commit to work toward my suggestion + @drammock's comments above?

alexrockhill:
> @alexrockhill do you want to push a commit to work toward my suggestion + @drammock's comments above?

Yeah sorry I was busy learning to dig clams this weekend with my family... Still interesting even though don't eat meat. Will push in a minute or two

alexrockhill:
I'm not sure how `doc/python_reference.rst` should be changed. Isn't it just the `datasets` as a submodule?

tutorials/misc/plot_ecog.py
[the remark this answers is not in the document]
larsoner: This is now taken care of above in `datasets.rst`, indeed nothing left to do here @alexrockhill

tutorials/misc/plot_ecog.py
```diff
@@ -11,192 +11,343 @@
...
-# We can then plot the locations of our electrodes on our subject's brain.
-# We'll use :func:`~mne.viz.snapshot_brain_montage` to save the plot as image
-# data (along with xy positions of each electrode in the image), so that later
-# we can plot frequency band power on top of it.
+# Explore the electrodes on a template brain
+# ------------------------------------------
 #
-# .. note:: These are not real electrodes for this subject, so they
-#           do not align to the cortical surface perfectly.
+# Our electrodes are shown in ``mni_tal`` space, which corresponds to
```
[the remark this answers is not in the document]
alexrockhill: Maybe this isn't the best place to go into how to do the CVS morph to transform to `fsaverage` space so this could just be left out. Seems more like a developer detail.

tutorials/misc/plot_ecog.py
```diff
@@ -11,192 +11,343 @@
...
-                                ['Gamma', 'Alpha']):
-    ax.imshow(im)
-    ax.set_axis_off()
-    sc = ax.scatter(*xy_pts.T, c=band_power, s=200,
-                    cmap=cmap, vmin=vmin, vmax=vmax)
-    ax.set_title(f'{band} band power', size='x-large')
-fig.colorbar(sc, ax=axs)
+# You can save a movie like the one on our documentation website with:
+# brain.save_movie(time_dilation=1, interpolation='linear', framerate=5,
+#                  time_viewer=True)
```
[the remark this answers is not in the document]
larsoner: Points 1/2/4/5 are really bugs with Brain so I don't think we can/should fix them here. For the blockiness it's a choice to show the discretization of `src` but we could change the steps to `5` for simplicity. For 6/7 @alexrockhill you could update the narrative doc to mention that it's the one with highest-amplitude activation in the gamma band, and maybe change the view shown (but it's always kind of hard to see so we might just need to live with 7 as is)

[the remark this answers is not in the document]
larsoner: (although maybe for `2` changing the upper crop time limit or increasing `framerate` a tiny bit might help...)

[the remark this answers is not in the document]
drammock: - Ah right, the blockiness is because this is `stc_near_sensors` not an MNE/dSPM/etc map. That makes sense.
- is it really not possible to pick the sphere+trace color?  If not, can we change the colormap to something (magma/inferno perhaps?) that will contrast with the default MPL blue of the sphere?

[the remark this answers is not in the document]
larsoner: I'd rather change the colormap to `inferno` or something similar.

I don't think there is a way with Brain currently to set which points get a trace, since currently it's meant to be done interactively (and have some sensible defaults). For a programmatic case you could currently do `brain.add_foci` then extract the time course manually and plot it but this is overkill for an example. We could probably make public the functions that internally are used to add a dot + trace but not in this PR.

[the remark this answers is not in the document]
drammock: I'm fine with changing colormap (and maybe which brain view(s) are shown) to make the sphere easier to see.

alexrockhill:
Ok, I think that's good https://26452-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html?highlight=ecog

tutorials/misc/plot_ecog.py
```diff
@@ -11,192 +11,169 @@
...
+# Our electrodes are shown after being morphed to fsaverage brain so we'll use
+# this fsaverage brain to plot the locations of our electrodes. We'll use
+# :func:`~mne.viz.snapshot_brain_montage` to save the plot as image data
+# (along with xy positions of each electrode in the image), so that later
+# we can plot frequency band power on top of it.
 
-# Convert from a dictionary to array to plot
-xy_pts = np.vstack([xy[ch] for ch in raw.info['ch_names']])
+fig = plot_alignment(raw.info, subject='fsaverage', subjects_dir=subjects_dir,
+                     surfaces=['pial'], coord_frame='mri')
```
adam2392, line 108:
```suggestion
                     surfaces=['pial'], coord_frame='mni_tal')
```

alexrockhill: This causes an error... you can't pass that into `coord_frame`, it's already in `fsaverage` because we're using `subject = 'fsaverage'`

adam2392: Ah I see you're right.

drammock:
> Ok, I think that's good https://26452-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html?highlight=ecog

I'm getting confused here:

1. The [discussion above](https://github.com/mne-tools/mne-python/pull/9087#discussion_r599019570) converged on keeping the trace and changing the colormap (the whole point of the colormap change was to make the sphere marking the electrode shown in the trace more prominent).  Now the colormap has changed but the trace is gone?  
2. the speed of the animation is a lot faster. Why?  The resulting flickering effect is ironic given that flickering can induce seizure in some viewers.  I'm not susceptible to seizure in those circumstances but I do find it quite hard to watch now.

alexrockhill:
> > Ok, I think that's good https://26452-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html?highlight=ecog
> 
> I'm getting confused here:
> 
> 1. The [discussion above](https://github.com/mne-tools/mne-python/pull/9087#discussion_r599019570) converged on keeping the trace and changing the colormap (the whole point of the colormap change was to make the sphere marking the electrode shown in the trace more prominent).  Now the colormap has changed but the trace is gone?
> 2. the speed of the animation is a lot faster. Why?  The resulting flickering effect is ironic given that flickering can induce seizure in some viewers.  I'm not susceptible to seizure in those circumstances but I do find it quite hard to watch now.

It sounded like there were more issues with the time course than information from seeing it, I agree that I'm not sure what it adds. It's also hard to see all in one html page on my display at least so now with the smaller size and no time course it all fits on one screen which I like. It didn't seem like there was a strong argument for having a time course, rather just addressing how to fix bugs with the display, is there a good reason to have it?

libertyh:
Just looked at the most recent changes to this, and looks cool! I know I've been out for a bit. A couple of things:

1. In the "gamma power over time" plot, it's unclear to me what the viridis colormap is supposed to be showing. What is the purpose of the different colored traces? Are they ordered by electrode device or similar? It seems distracting to me if the color isn't meaningful.
2. How do we interpret the gamma power itself on the plot? Looks like there are a few posterior electrodes in the temporal lobe that have one kind of "spike" -- is this meaningful or just an artifact? Are the traces plotted upside down (is positive power down or up?)
3. For the animation, I think maybe this was mentioned above, but is it possible to put the electrodes back on top of the plot (they can be small) so it's obvious where the interpolated data is coming from?

alexrockhill:
It's rather long, so perhaps we should turn up the smoothing but leave the time dilation at none. I think the increased flickering is due to changing the colormap boundaries but I like that the way it is uses the whole color range.

alexrockhill:
> Just looked at the most recent changes to this, and looks cool! I know I've been out for a bit. A couple of things:

There's a few planned changes but we want to start with some rendered plots and animations that don't push the limits of Circle.

> 
> 1. In the "gamma power over time" plot, it's unclear to me what the viridis colormap is supposed to be showing. What is the purpose of the different colored traces? Are they ordered by electrode device or similar? It seems distracting to me if the color isn't meaningful.

They are indeed just colored by index. The one thing it really helps with is contrast since the line plots overlap a bit. However, if the colors are all random, it looks quite bad which is why this is the best I could think of. Not sure taking out the colors would be an improvement if that's what you meant...

> 2. How do we interpret the gamma power itself on the plot? Looks like there are a few posterior electrodes in the temporal lobe that have one kind of "spike" -- is this meaningful or just an artifact? Are the traces plotted upside down (is positive power down or up?)

Yeah, they are upside down, I'll flip them over. I'm not sure if the spike is meaningful, maybe @adam2392 knows.

> 3. For the animation, I think maybe this was mentioned above, but is it possible to put the electrodes back on top of the plot (they can be small) so it's obvious where the interpolated data is coming from?

Hmmm, I'm not sure how to do this, perhaps with `brain.add_data`. I'll have to look into it.

adam2392:
Re spike: It looks like we're not just cropping to around the seizure, and it's showing a condensed version of the entire trace, so I presume that spike is just some time during the seizure, which is pretty normal to see huge variation in frequency power.

alexrockhill:
Looks pretty good https://26457-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html?highlight=ecog

drammock:
Flickering in the movie is much better now, thanks.  For the "gamma power over time" I'm thinking about ways to solve the overlap issue: can you make it as big as the previous image that showed just the electrodes? And/or zoom in slightly?  If you try zooming, I think it would be OK to have some of the extreme anterior/posterior electrodes go out of frame, as long as there were still *some* electrodes visible that weren't undergoing seizure.

alexrockhill:
The best way I can think of addressing that is the second view so I added that back

alexrockhill:
Hmmm, Sphinx renders fine for me for the second plot on the right locally but is all messed up [here](https://26459-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/misc/plot_ecog.html), any suggestions? cc @larsoner

alexrockhill:
So I looked into what needed to be done to animate the time series as had been done in the updated example and it looks like `evoked.animate_topomap` already exists so an `evoked.animate_topo` could be added but I think this is a bit of a special case using long evoked that wouldn't necessarily be helpful generally where evoked time courses are usually not nearly as long. Not sure if a separate PR to add animation to a topo would be useful, in my opinion, the way I had it previously is the most helpful to a user; it's not too much code although it is a bit code heavy and it conveys the most information about the seizure. I'm fine merging this version too, I just don't think it's as informative as the animated line plots and the animated bar plots. In my opinion, it's really nice to have a bit of code to show how things work as long as it's well-written-- if everything is too encapsulated, the user doesn't learn about the components in a way that makes the code more flexible for them to use.

There is still this issue with the plotting that seems like a backends issue, unless there is a solution to that issue, the cropping + zooming in might have to go.

larsoner:
My vote is to remove the crop+zoom for now and add it later. Same with animations, let's make incremental improvements in subsequent PRs to `plot_topo` / `find_layout` etc., then add animation etc. I agree these things are useful -- maybe we want an `animate_topo` or something -- but let's do those in smaller chunks rather than all at once here.

alexrockhill:
> My vote is to remove the crop+zoom for now and add it later. Same with animations, let's make incremental improvements in subsequent PRs to `plot_topo` / `find_layout` etc., then add animation etc. I agree these things are useful -- maybe we want an `animate_topo` or something -- but let's do those in smaller chunks rather than all at once here.

I addressed @libertyh's comment (good suggestion, thanks) and hopefully fixed the plotting so I think this is okay unless you want to remove it.

alexrockhill:
Animated gist for the record https://gist.github.com/alexrockhill/fa7fd568eed5fb8996961a7e1564ce0e

larsoner:
Thanks @libertyh @alexrockhill @adam2392 for working through this, it's a great improvement!

Looking forward to the next round with some `plot_topo` / `find_layout` gymnastics @alexrockhill :)

Issue #3525, opened 2016-08-17
"Morph" dipole to MNI space


agramfort:
we have a vertex_to_mni  function for source space based inverse methods

it would be great to have the same thing for dipole fits. E.g.

dipole.to_mni() -> returns new Dipole instance where dipole.pos is in MNI coordinates.

thoughts?

Eric89GXL:
Probably requires something like #3527

Eric89GXL:
@agramfort do you have an API in mind? I'm thinking:
```
dip.to_mni(subject)
```
Where `subject` is a FreeSurfer subject name.

Unfortunately the dipole fits are not necessarily constrained to a cortical surface, so I think we need to use volumetric / MRI coregistration instead of surface-based coregistration to get there. Agreed?

Eric89GXL:
... oh wait, I think each FreeSurfer subject already has an MNI transform. So this should be very simple actually, assuming I have the API right.

agramfort:
yes it MNE would be great to we could overlay this with the FSL MNI brain
template

dip.to_mni works for me

vagechirkov:
I would like to try to solve this issue (sprint 2021) :)

sappelhoff:
closed in #9091

PR #9091, opened 2021-03-15
ENH: dipole to MNI space

vagechirkov:
Reference issue #3525.

Adding this method to Dipole should make it clear how to convert dipoles coordinates to MNI space.

This is already mentioned in the [tutorial](https://github.com/mne-tools/mne-python/blob/7168e683e319f99a3c57ec68f7f0e844eafcb463/tutorials/source-modeling/plot_dipole_fit.py#L54), although I think it is more convenient this way.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add :func:`mne.Dipole.to_mni` for more convenient  dipole.pos to mni conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
+- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` by |Valerii Chirkov|_)
 
 - Update citations in maxwell.py (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,6 +39,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :func:`mne.Dipole.to_mni` for more convenient  dipole.pos to mni conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
+
 - Update citations in maxwell.py (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
 - New Tutorial for analyzing frequency-tagging data (:gh:`8867` **by new contributor** |Dominik Welke|_ and `kalenkovich`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,7 +39,7 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` by |Valerii Chirkov|_)
+- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
 - Update citations in maxwell.py (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
--- a/mne/dipole.py
+++ b/mne/dipole.py
@@ -268,6 +268,27 @@ class Dipole(object):
             show_all, ax, block, show, scale=scale, color=color, fig=fig,
             title=title)
 
+    def to_mni(self, subject, mri_head_t, subjects_dir=None,
+               verbose=None):
+        """Convert pos from head coordinate system to MNI ones.
+
+        Parameters
+        ----------
+        %(subject)s
+        mri_head_t : instance of Transform
+            MRI<->Head coordinate transformation.
+        %(subjects_dir)s
+        %(verbose)s
+
+        Returns
+        -------
+        coordinates : array, shape (n_pos, 3)
+            The MNI coordinates (in mm) of pos.
+        """
+        from .source_space import head_to_mni
+        return head_to_mni(self.pos, subject, mri_head_t,
+                           subjects_dir=subjects_dir, verbose=verbose)
+
     def plot_amplitudes(self, color='k', show=True):
         """Plot the dipole amplitudes as a function of time.
 
--- a/mne/dipole.py
+++ b/mne/dipole.py
@@ -276,15 +276,13 @@ class Dipole(object):
         Parameters
         ----------
         %(subject)s
-        trans : str | None
-            The head<->MRI transform filename. Must be provided unless BEM
-            is a sphere model.
+        %(trans)s
         %(subjects_dir)s
         %(verbose)s
 
         Returns
         -------
-        coordinates : array, shape (n_pos, 3)
+        pos_mni : array, shape (n_pos, 3)
             The MNI coordinates (in mm) of pos.
         """
         from .source_space import head_to_mni
--- a/mne/dipole.py
+++ b/mne/dipole.py
@@ -271,7 +271,7 @@ class Dipole(object):
     @verbose
     def to_mni(self, subject, mri_head_t, subjects_dir=None,
                verbose=None):
-        """Convert pos from head coordinate system to MNI ones.
+        """Convert dipole location from head coordinate system to MNI coordinates.
 
         Parameters
         ----------
--- a/mne/dipole.py
+++ b/mne/dipole.py
@@ -29,7 +29,7 @@ from .forward._compute_forward import (_compute_forwards_meeg,
 from .surface import (transform_surface_to, _compute_nearest,
                       _points_outside_surface)
 from .bem import _bem_find_surface, _bem_surf_name
-from .source_space import _make_volume_source_space, SourceSpaces
+from .source_space import _make_volume_source_space, SourceSpaces, head_to_mni
 from .parallel import parallel_func
 from .utils import (logger, verbose, _time_mask, warn, _check_fname,
                     check_fname, _pl, fill_doc, _check_option, ShiftTimeMixin,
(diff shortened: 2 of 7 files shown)


drammock:
I know this is still marked as draft, but FYI these are some of the issues on the CI so far:

```
WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.Dipole.to_mni':
  PR01: Parameters {'subject', 'subjects_dir', 'verbose'} not documented
  PR02: Unknown parameters {'%(subject)s', '%(subjects_dir)s', '%(verbose)s'}
  PR07: Parameter "%(subject)s" has no description
  PR07: Parameter "%(subjects_dir)s" has no description
  PR07: Parameter "%(verbose)s" has no description
```

mne/dipole.py
```diff
@@ -268,6 +268,27 @@ def plot_locations(self, trans, subject, subjects_dir=None,
             show_all, ax, block, show, scale=scale, color=color, fig=fig,
             title=title)
 
+    def to_mni(self, subject, mri_head_t, subjects_dir=None,
```
wmvanvliet, line 271:
```suggestion
    @verbose
    def to_mni(self, subject, mri_head_t, subjects_dir=None,
```

wmvanvliet: You need the `@verbose` decorator if you have a parameter in your function signature.

mne/dipole.py
```diff
@@ -268,6 +268,28 @@ def plot_locations(self, trans, subject, subjects_dir=None,
             show_all, ax, block, show, scale=scale, color=color, fig=fig,
             title=title)
 
+    @verbose
+    def to_mni(self, subject, mri_head_t, subjects_dir=None,
+               verbose=None):
+        """Convert pos from head coordinate system to MNI ones.
```
agramfort, line 274:
```suggestion
        """Convert dipole location from head coordinate system to MNI coordinates.
```

mne/dipole.py
```diff
@@ -268,6 +268,28 @@ def plot_locations(self, trans, subject, subjects_dir=None,
...
+
+        Parameters
+        ----------
+        %(subject)s
+        mri_head_t : instance of Transform
+            MRI<->Head coordinate transformation.
```
agramfort, line 280: you should factorize the docstring as this is what we call `trans` everywhere. See fit_dipole function that writes:

```
    trans : str | None
        The head<->MRI transform filename. Must be provided unless BEM
        is a sphere model.
```

I would rename to `trans`

mne/dipole.py
```diff
@@ -268,6 +268,28 @@ def plot_locations(self, trans, subject, subjects_dir=None,
...
+        %(subjects_dir)s
+        %(verbose)s
+
+        Returns
+        -------
+        coordinates : array, shape (n_pos, 3)
```
agramfort, line 286: I would call it pos_mni

mne/dipole.py
agramfort: Although I see that we have mri_head_t in head_to_mni function.

I think users are more used to the word `trans` in all tutorials etc.

mne/dipole.py
```diff
@@ -269,15 +269,16 @@ def plot_locations(self, trans, subject, subjects_dir=None,
...
+    def to_mni(self, subject, trans, subjects_dir=None,
                verbose=None):
         """Convert dipole location from head coordinate system to MNI coordinates.
 
         Parameters
         ----------
         %(subject)s
-        mri_head_t : instance of Transform
-            MRI<->Head coordinate transformation.
+        trans : str | None
```
agramfort, line 279: if you follow the general API we are following with trans, it can be `str | None or instance of Transform`

see how it's done elsewhere

agramfort: is there a way to factorize the docstring using the shared docdict?

vagechirkov: I think I did if finally :)

mne/dipole.py
```diff
@@ -268,6 +268,27 @@ def plot_locations(self, trans, subject, subjects_dir=None,
...
 
+    @verbose
+    def to_mni(self, subject, trans, subjects_dir=None,
+               verbose=None):
+        """Convert dipole location from head coordinate system to MNI coordinates.
+
+        Parameters
+        ----------
+        %(subject)s
+        %(trans)s
```
agramfort, line 279:
```suggestion
        %(trans_not_none)s
```

look at the text on mne/utils/docs.py
to see what it will be replaced with.

vagechirkov: Alright, got it!

mne/tests/test_dipole.py
```diff
@@ -140,6 +141,13 @@ def test_dipole_fitting(tmpdir):
                                    rank='info')  # just to test rank support
     assert isinstance(residual, Evoked)
 
+    # Test conversion of dip.pos to MNI coordinates.
+    dip_mni_pos = dip.to_mni('sample', fwd['mri_head_t'],
+                             subjects_dir=subjects_dir)
+    head_to_mni_dip_pos = head_to_mni(dip.pos, 'sample', fwd['mri_head_t'],
```
agramfort, line 147: can you try making one call with the str input passing the file name? find the -trans.fif file in the testing data.
just to make sure we test the different input types

doc/changes/latest.inc
```diff
@@ -39,6 +39,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :func:`mne.Dipole.to_mni` for more convenient  dipole.pos to mni conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
```
agramfort, line 42:
```suggestion
- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to mni conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
```

agramfort: also mni -> MNI

vagechirkov: Thank you!

vagechirkov:
```suggestion
- Add :func:`mne.Dipole.to_mni` for more convenient  dipole.pos to mni conversion (:gh:`9043` by |Valerii Chirkov|_)
```

agramfort (approved):
thx @vagechirkov 

+1 for MRG when CI are green

doc/changes/latest.inc
```diff
@@ -39,7 +39,7 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
+- Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` by |Valerii Chirkov|_)
```
larsoner, line 42: Let's keep the `**by new contributor**` in here -- it's better to have it both places I think

vagechirkov: Alright!

mne/dipole.py
```diff
@@ -268,6 +268,28 @@ def plot_locations(self, trans, subject, subjects_dir=None,
...
+        %(trans_not_none)s
+        %(subjects_dir)s
+        %(verbose)s
+
+        Returns
+        -------
+        pos_mni : array, shape (n_pos, 3)
+            The MNI coordinates (in mm) of pos.
+        """
+        from .source_space import head_to_mni
```
larsoner, line 288: No need to nest this, it can go at the top. We only nest our own imports to avoid circular import problems (which shouldn't be the case here because `.source_space` is already imported from above)

doc/changes/latest.inc
larsoner: Looks like it's not there in the current version?

vagechirkov: Yes, sorry, I've fixed this

mne/tests/test_dipole.py
```diff
@@ -146,7 +146,7 @@ def test_dipole_fitting(tmpdir):
                              subjects_dir=subjects_dir)
     head_to_mni_dip_pos = head_to_mni(dip.pos, 'sample', fwd['mri_head_t'],
                                       subjects_dir=subjects_dir)
-    assert_array_equal(dip_mni_pos, head_to_mni_dip_pos)
+    assert_allclose(dip_mni_pos, head_to_mni_dip_pos, rtol=1e-3, atol=0)
```
drammock, line 149: @larsoner I suggested this change to @vagechirkov because the arrays were coming out as *nearly* identical (all elements different, but only by ~0.00001 mm or so).  But I'm curious why they were different to begin with (presumably `fname_trans` and `fwd['mri_head_t']` are not identical?)  Could this be due to loading one from disk (making it 32-bit precision)?

larsoner: Could be 32-bit precision, or it could be that in one place it's a head->MRI and somewhere it gets converted to MRI->head (or vice-versa) using `linalg.inv`, which can also lose precision (especially if it's after a 32-bit conversion).

drammock: what do you think is a sensible tolerance here?  since we're talking MNI space (millimeters) I'd be inclined for something like `atol=1e-3, rtol=0` or similar?  that ensures they're within one micrometer of each other

larsoner: That would work, too. The given tolerance is probably also fine given the `rtol=1e-3` and knowledge that the dipole positions are not too wacky.

vagechirkov: Thank you for the explanation! But I am not sure that I understand the difference between ```atol``` and ```rtol``` in this case.

larsoner: `atol` would allow you to specify absolute-deviation-from-correct, e.g. 0.001 mm or so like @drammock said. `rtol` is relative, so it's like a percentage-from-correct error. In this case it's not going to make a big difference because our values are all within a similar range (probably not off by more than a couple orders of magnitude) and none (presumably) get close to zero.

larsoner:
Thanks @vagechirkov !

mne/tests/test_dipole.py
vagechirkov: Alright, thank you!

PR #9092, opened 2021-03-16
split up code cells in example

### LINKED ISSUE #9004: split code block in examples/preprocessing/plot_virtual_evoked
right now, because all plots come from a single code block, they are plotted at the top of the example in a group of 4 (and consequently the plots are really small).  By splitting the 4 plotting calls into different code blocks, they will plot larger / be easier to see & compare, without increasing run time of the example.  Code blocks can be split with a line of 79 `#` marks (adding a bit of explanatory text too is usually a good idea)

eort:
#### Reference issue
Fixes #9004

- the code was split into four code blocks, so that the figures are printed in separate cells. 
- a few lines of explanation were added

### DIFF
--- a/examples/preprocessing/plot_virtual_evoked.py
+++ b/examples/preprocessing/plot_virtual_evoked.py
@@ -26,14 +26,30 @@ data_path = sample.data_path()
 fname = data_path + '/MEG/sample/sample_audvis-ave.fif'
 evoked = mne.read_evokeds(fname, condition='Left Auditory', baseline=(None, 0))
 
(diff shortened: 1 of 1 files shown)


rob-luke:
Render: https://25909-1301584-gh.circle-artifacts.com/0/dev/auto_examples/preprocessing/plot_virtual_evoked.html#sphx-glr-auto-examples-preprocessing-plot-virtual-evoked-py

agramfort (approved):

agramfort:
thx @eort !

PR #9096, opened 2021-03-16
Fixed length epochs overlap

### LINKED ISSUE #8773: Adding overlap keyword to make_fixed_length_epochs

sfc-neuro:
#### Reference issue
Fixes #8773.

Adds compatibility for the overlap parameter in the function make_fixed_length_epochs

#### Additional information
Tests were also added to test the new feature inspired by the pre-existing ones written for make_fixed_length_events.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -45,7 +45,7 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add parameter ``overlap=0.`` to :func:`mne.epochs.make_fixed_length_epochs` to allow creating overlapping fixed length epochs (:gh:`9096` **by new contributor** |Silvia Cotroneo|_)
+- Add ``overlap`` parameter to :func:`mne.epochs.make_fixed_length_epochs` to allow creating overlapping fixed length epochs (:gh:`9096` **by new contributor** |Silvia Cotroneo|_)
 
 - Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
@@ -222,4 +222,3 @@ API changes
 - ``mne.read_selection`` has been deprecated in favor of `mne.read_vectorview_selection`. ``mne.read_selection`` will be removed in MNE-Python 0.24 (:gh:`8870` by `Richard Höchenberger`_)
 
 - ``mne.beamformer.tf_dics`` has been deprecated and will be removed in MNE-Python 0.24 (:gh:`9122` by `Britta Westner`_)
-
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,8 +39,12 @@ Current (0.23.dev0)
 
 .. |Matteo Anelli| replace:: **Matteo Anelli**
 
+.. |Silvia Cotroneo| replace:: **Silvia Cotroneo**
+
 Enhancements
 ~~~~~~~~~~~~
+- Add parameter ``overlap=0.`` to :func:`mne.epochs.make_fixed_length_epochs` to allow creating overlapping fixed length epochs (:gh:`9096` **by new contributor** |Silvia Cotroneo|_)
+
 - Add :meth:`mne.Dipole.to_mni` for more convenient  dipole.pos to MNI conversion (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
 
 - Update citations in maxwell.py (:gh:`9043` **by new contributor** |Valerii Chirkov|_)
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3678,11 +3678,7 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
         Raw data to divide into segments.
     duration : float
         Duration of each epoch in seconds. Defaults to 1.
-    overlap : float
-        The overlap between epochs. Must be ``0 <= overlap < duration``.
-        Default is 0.
 
-        .. versionadded:: 0.23
     %(preload)s
     %(reject_by_annotation_epochs)s
 
@@ -3690,6 +3686,11 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
     %(proj_epochs)s
 
         .. versionadded:: 0.22.0
+     overlap : float
+        The overlap between epochs. Must be ``0 <= overlap < duration``.
+        Default is 0.
+
+        .. versionadded:: 0.23.0
     %(verbose)s
 
     Returns
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3668,8 +3668,8 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
 
 @verbose
 def make_fixed_length_epochs(raw, duration=1., preload=False,
-                             reject_by_annotation=True, proj=True,
-                             verbose=None, overlap=0.):
+                             reject_by_annotation=True, proj=True, overlap=0.,
+                             verbose=None):
     """Divide continuous raw data into equal-sized consecutive epochs.
 
     Parameters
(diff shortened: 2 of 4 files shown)


mne/epochs.py
```diff
@@ -3667,7 +3667,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
 
 
 @verbose
-def make_fixed_length_epochs(raw, duration=1., preload=False,
+def make_fixed_length_epochs(raw, duration=1., overlap=0., preload=False,
```
agramfort, line 3670: @sfc-neuro you might break the code of people by inserting a new param in the middle. We usually but it just before verbose.

mne/epochs.py
```diff
@@ -3678,6 +3678,9 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
         Raw data to divide into segments.
     duration : float
         Duration of each epoch in seconds. Defaults to 1.
+    overlap : float
+        The overlap between epochs. Must be ``0 <= overlap < duration``.
+        Default is 0.
```
agramfort, line 3683: you need to add the `versionadded` directive to document when this parameter appeared in a release.

mne/viz/ica.py
```diff
@@ -44,11 +44,11 @@ def plot_ica_sources(ica, inst, picks=None, start=None,
...
-    start : int
-        X-axis start index. If None, from the beginning.
-    stop : int
-        X-axis stop index. If None, next 20 are shown, in case of evoked to the
-        end.
+    start : int | None
+        X-axis start index. If None (default), from the beginning.
+    stop : int | None
+        X-axis stop index. If None (default), next 20 are shown, in case of
+        evoked to the end.
```
agramfort, line 51: these commits are unrelated. You should have started a branch from main rather than from your preview PR.

mne/epochs.py
sfc-neuro: Thanks, added

mne/epochs.py
sfc-neuro: Thanks, moved.

mne/epochs.py
```diff
@@ -3678,6 +3678,11 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
...
+
+        .. versionadded:: 0.23
```
agramfort, line 3685: docstring entries must match the order of the parameters. So it should be moved just before verbose

mne/viz/ica.py
sfc-neuro: I think it should be fixed in the next pull request! Thanks for pointing it out

mne/epochs.py
sfc-neuro: Thanks!

mne/tests/test_epochs.py
```diff
@@ -3254,6 +3254,22 @@ def test_make_fixed_length_epochs():
...
     assert len(epochs) > len(epochs_annot)
 
+    # overlaps
+    epochs = make_fixed_length_epochs(raw, duration=1, preload=True)
+    assert len(epochs) > 10
+    epochs_ol = make_fixed_length_epochs(raw, duration=1, overlap=0.5,
+                                         preload=True)
+    assert len(epochs_ol) > 20
+    epochs_ol_2 = make_fixed_length_epochs(raw, duration=1, overlap=0.9,
+                                           preload=True)
```
agramfort, line 3264:
```suggestion
    epochs_ol_2 = make_fixed_length_epochs(raw, duration=1, overlap=0.9)
```

mne/tests/test_epochs.py
```diff
@@ -3254,6 +3254,22 @@ def test_make_fixed_length_epochs():
     assert len(epochs_annot) > 10
     assert len(epochs) > len(epochs_annot)
 
+    # overlaps
+    epochs = make_fixed_length_epochs(raw, duration=1, preload=True)
+    assert len(epochs) > 10
+    epochs_ol = make_fixed_length_epochs(raw, duration=1, overlap=0.5,
+                                         preload=True)
+    assert len(epochs_ol) > 20
```
agramfort, line 3262:
```suggestion
    epochs_ol = make_fixed_length_epochs(raw, duration=1, overlap=0.5)
    assert len(epochs_ol) > 20
```

mne/tests/test_epochs.py
```diff
@@ -3254,6 +3254,22 @@ def test_make_fixed_length_epochs():
     assert len(epochs_annot) > 10
     assert len(epochs) > len(epochs_annot)
 
+    # overlaps
+    epochs = make_fixed_length_epochs(raw, duration=1, preload=True)
```
agramfort, line 3258:
```suggestion
    epochs = make_fixed_length_epochs(raw, duration=1)
```

sfc-neuro: I'll comment here bc it is the same for the suggestions before. If I do not specify the preload, the `assert len(epochs)` fails because the len(epochs) seem to only be working with preloaded data.

I thought I should keep it for this reason, what do you think? Could I change the check in a more optimized way?

sfc-neuro: Edit: I changed it to check the length of the events in the epochs objects to avoid preloading but still performing an analogous check.

Now the tests are not anymore on the same style of the previous ones in the same file (`mne/tests/test_epochs.py`)  that create epochs with `preload=True` and asserting `len(epochs)> n`.

@agramfort do you think they should also be modified too for better readability and optimization?

agramfort (approved):
can we just add a what's new entry?

thx @sfc-neuro !

agramfort:
@sfc-neuro can you quickly push a what's new entry so we can move on here? thanks

doc/changes/latest.inc
```diff
@@ -39,8 +39,12 @@ Current (0.23.dev0)
 
 .. |Matteo Anelli| replace:: **Matteo Anelli**
 
+.. |Silvia Cotroneo| replace:: **Silvia Cotroneo**
+
 Enhancements
 ~~~~~~~~~~~~
+- Add parameter ``overlap=0.`` to :func:`mne.epochs.make_fixed_length_epochs` to allow creating overlapping fixed length epochs (:gh:`9096` **by new contributor** |Silvia Cotroneo|_)
```
hoechenberger, line 46:
```suggestion
- Add ``overlap`` parameter to :func:`mne.epochs.make_fixed_length_epochs` to allow creating overlapping fixed length epochs (:gh:`9096` **by new contributor** |Silvia Cotroneo|_)
```

mne/epochs.py
```diff
@@ -3685,6 +3685,11 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
     %(proj_epochs)s
 
         .. versionadded:: 0.22.0
+    overlap : float
+        The overlap between epochs. Must be ``0 <= overlap < duration``.
+        Default is 0.
```
hoechenberger, line 3690:
```suggestion
        The overlap between epochs, in seconds. Must be
        ``0 <= overlap < duration``. Default is 0, i.e., no overlap.
```

hoechenberger (approved):
LGTM once CI is green

wmvanvliet:
Thanks @sfc-neuro!

PR #9097, opened 2021-03-16
DOC: Add HEDs to glossary and reference in tutorial and epochs.py  #8974.

### LINKED ISSUE #8974: Add HED to glossary
In MNE, we can describe events using hierarchical event descriptors (HED), such as `stim/a`, `stim/b` etc. They should be mentioned in the glossary and potentially a reference to the original publication should be added to an Epochs/events tutorial

Publication:
http://sccn.ucsd.edu/~nima/downloads/HED_globalsip_paper.pdf

anaradanovic:
#### Reference issue
Fixes #8974.

I added the citation for hierarchical event descriptors in the glossary, added this citation to the tutorial (plot_40_autogenerate_metadata.py ) and to epochs.py. 
Also fixed a couple typos in the description in plot_40_autogenerate_metadata.py.

### DIFF
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -169,6 +169,16 @@ general neuroimaging concepts. If you think a term is missing, please consider
         it is a one-dimensional time series capturing the spatial variability
         of the signal across sensor locations.
 
+    HED
+    hierarchical event descriptors
+        Hierarchical event descriptors (abbr. ``HED``) are tags that can be
+        used to describe many different types of EEG experiment events (for 
+        example, stimulus/circle/red/left and stimulus/square/blue/left).
+        These are organized hierarchically such that the first labels are fixed
+        and corresponding descriptors can be extended to describe any event
+        with any level of detail. HED tags make it easy to search through and
+        analyze variations of events of the same type.
+
     HPI
     cHPI
     head position indicator
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -171,13 +171,12 @@ general neuroimaging concepts. If you think a term is missing, please consider
 
     HED
     hierarchical event descriptors
-        Hierarchical event descriptors (abbr. ``HED``) are tags that can be
-        used to describe many different types of EEG experiment events (for 
-        example, stimulus/circle/red/left and stimulus/square/blue/left).
-        These are organized hierarchically such that the first labels are fixed
-        and corresponding descriptors can be extended to describe any event
-        with any level of detail. HED tags make it easy to search through and
-        analyze variations of events of the same type.
+        Hierarchical event descriptors (abbr. ``HED``) are tags that use
+        keywords separated by '/' to describe different types of EEG
+        experimental events (for example, stimulus/circle/red/left and
+        stimulus/circle/blue/left). These tags can be used to group
+        experimental events and select event types for analysis.
+        
 
     HPI
     cHPI
--- a/doc/glossary.rst
+++ b/doc/glossary.rst
@@ -172,11 +172,10 @@ general neuroimaging concepts. If you think a term is missing, please consider
     HED
     hierarchical event descriptors
         Hierarchical event descriptors (abbr. ``HED``) are tags that use
-        keywords separated by '/' to describe different types of EEG
+        keywords separated by '/' to describe different types of
         experimental events (for example, stimulus/circle/red/left and
         stimulus/circle/blue/left). These tags can be used to group
         experimental events and select event types for analysis.
-        
 
     HPI
     cHPI
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -141,6 +141,15 @@
   year = {1994}
 }
 
+@inproceedings{BigdelyShamloEtAl2013,
+  author={Bigdely-Shamlo, Nima and Kreutz-Delgado, Kenneth and Robbins, Kay and Miyakoshi, Makoto and Westerfield, Marissa and Bel-Bahar, Tarik and Kothe, Christian and Hsi, Jessica and Makeig, Scott},
+  booktitle={2013 IEEE Global Conference on Signal and Information Processing},
+  pages={1--4},
+  title={Hierarchical event descriptor (HED) tags for analysis of event-related EEG studies},
+  organization={IEEE},
+  year={2013},
+}
+
 @article{BlankertzEtAl2008,
   author = {Blankertz, Benjamin and Tomioka, Ryota and Lemm, Steven and Kawanabe, Motoaki and Müller, Klaus-Robert},
   doi = {10.1109/MSP.2008.4408441},
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -142,12 +142,13 @@
 }
 
 @inproceedings{BigdelyShamloEtAl2013,
-  author={Bigdely-Shamlo, Nima and Kreutz-Delgado, Kenneth and Robbins, Kay and Miyakoshi, Makoto and Westerfield, Marissa and Bel-Bahar, Tarik and Kothe, Christian and Hsi, Jessica and Makeig, Scott},
-  booktitle={2013 IEEE Global Conference on Signal and Information Processing},
-  pages={1--4},
-  title={Hierarchical event descriptor (HED) tags for analysis of event-related EEG studies},
-  organization={IEEE},
-  year={2013},
+  author = {Bigdely-Shamlo, Nima and Kreutz-Delgado, Kenneth and Robbins, Kay and Miyakoshi, Makoto and Westerfield, Marissa and Bel-Bahar, Tarik and Kothe, Christian and Hsi, Jessica and Makeig, Scott},
+  doi = {10.1109/GlobalSIP.2013.6736796},
+  booktitle = {2013 IEEE Global Conference on Signal and Information Processing},
+  pages = {1--4},
+  title = {Hierarchical event descriptor (HED) tags for analysis of event-related EEG studies},
+  organization = {IEEE},
+  year = {2013},
 }
 
 @article{BlankertzEtAl2008,
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -2185,7 +2185,7 @@ def make_metadata(events, event_id, tmin, tmax, sfreq,
 
     .. versionadded:: 0.23
 
-     References
+    References
     ----------
     .. footbibliography::
     """
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -2107,9 +2107,10 @@ def make_metadata(events, event_id, tmin, tmax, sfreq,
         (default), rows are created for **all** event types present in
         ``event_id``.
     keep_first : str | list of str | None
-        Specify subsets of :term:`hierarchical event descriptors` (HEDs)
-        matching events of which  the **first occurrence** within each
-        time window shall be stored in addition to the original events.
+        Specify subsets of :term:`hierarchical event descriptors` (HEDs,
+        :footcite:`BigdelyShamloEtAl2013`) matching events of which the
+        **first occurrence** within each time window shall be stored in
+        addition to the original events.
 
         .. note::
            There is currently no way to retain **all** occurrences of a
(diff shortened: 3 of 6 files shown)


doc/glossary.rst
```diff
@@ -169,6 +169,16 @@ general neuroimaging concepts. If you think a term is missing, please consider
...
 
+    HED
+    hierarchical event descriptors
+        Hierarchical event descriptors (abbr. ``HED``) are tags that can be
+        used to describe many different types of EEG experiment events (for 
+        example, stimulus/circle/red/left and stimulus/square/blue/left).
+        These are organized hierarchically such that the first labels are fixed
+        and corresponding descriptors can be extended to describe any event
+        with any level of detail. HED tags make it easy to search through and
+        analyze variations of events of the same type.
```
agramfort, line 180: @hoechenberger I am uneasy on this. HED from the eeglab folks is more than the tagging system we use with the `/`

At beast we support a subset of HED I think

doc/references.bib
```diff
@@ -141,6 +141,15 @@ @article{BergScherg1994
   year = {1994}
 }
 
+@inproceedings{BigdelyShamloEtAl2013,
+  author={Bigdely-Shamlo, Nima and Kreutz-Delgado, Kenneth and Robbins, Kay and Miyakoshi, Makoto and Westerfield, Marissa and Bel-Bahar, Tarik and Kothe, Christian and Hsi, Jessica and Makeig, Scott},
```
britta-wstnr, line 145: The spell check wrongly detects the name Hsi to be misspelled. You could add "Hsi" to the `ignore_words.txt, however, we @rob-luke @hoechenberger wonder whether we should actually parse the `bibtext`-files for spell checking.
@drammock ?

rob-luke: Could we add `*.bib` here...

https://github.com/mne-tools/mne-python/blob/5e63884ceff9bc7658f56c0699d5bf115d20723a/.github/workflows/codespell_and_flake.yml#L15

larsoner: Agreed we should add `references.bib` to CODESPELL_SKIPS

https://github.com/mne-tools/mne-python/blob/main/Makefile#L8

doc/glossary.rst
hoechenberger: I **thought** the forward slashes were pretty much what made up the first revision of HED. I need to have a look at the paper again, but I believe they didn't even specify pre-defined terms – this only came in later revisions of HED

doc/references.bib
larsoner: ... in both places, I didn't realize we had it duplicated. That's not great :( But for now yes let's add to both

drammock:
@anaradanovic is it clear from above what the additional change needed is here?  We want our spellchecker to skip `doc/references.bib`, and it needs to be added in 2 places;

1. https://github.com/mne-tools/mne-python/blob/main/.github/workflows/codespell_and_flake.yml#L15
2. https://github.com/mne-tools/mne-python/blob/main/Makefile#L8

doc/glossary.rst
sappelhoff: > At best we support a subset of HED I think

I agree, this is what I raised in https://github.com/mne-tools/mne-python/issues/8974#issuecomment-790737767

We should read the HED (1st generation) paper (so the very first HED paper) carefully to check, and adjust the proposed docs here to include this as a caveat.

Importantly, HED is a "standard" that has evolved in the past (and continuous to evolve). At its current stage it goes way way beyond what we do in MNE.

doc/references.bib
sappelhoff, line 145: adding the DOI here.

Also note that the remaining references have a space between `author`, the `=`, and the content like `{....}`, whereas you do: `author={....}` So a nitpicky reviewer may ask you to be consistent :-)



```suggestion
  author={Bigdely-Shamlo, Nima and Kreutz-Delgado, Kenneth and Robbins, Kay and Miyakoshi, Makoto and Westerfield, Marissa and Bel-Bahar, Tarik and Kothe, Christian and Hsi, Jessica and Makeig, Scott},
  doi={10.1109/GlobalSIP.2013.6736796},
```

sappelhoff (commented):
I re-read the initial HED paper ([1st generation HED tags from 2013](http://sccn.ucsd.edu/~nima/downloads/HED_globalsip_paper.pdf)), and I don't think saying that "we use HED in MNE" is helpful, .. even saying that we use a "subset of HED" is probably a bit misleading.

To give an example of an HED string from the paper ... this is a "comma separated list of HED tags", and the way HED (1st generation) is supposed to be used:

```
Stimulus/Feedback,
Stimulus/Visual/Uniform Color/Red,
Stimulus/Visual/Shape/Ellipse/Circle/Height/2-deg,
Stimulus/Visual/Shape/Ellipse/Circle/Width/2-deg,
Stimulus/Visual/Background/Uniform Color/Black
```

This is kind of an event description with lots of metadata on the event.

In MNE, we'd rather have something like the following for the same case:

`circle/red`

... and then `circle/blue`, ... `rectangle/red`, ... for different conditions.

---

Another important point is that the HED standard has progressed significantly (it's at 3rd generation) and now encompasses metadata specification beyond events and up to the scale of recording information, or even info about the room where a session was recorded.

Yet another unfortunate point is, that the 1st generation HED tags (which were still **relatively** basic) already DID have a specific vocabulary which they refer to in the paper (ref 7: https://sccn.ucsd.edu/wiki/HED)

and in MNE, we have no support / validation of that vocabulary. That vocabulary can be seen here: http://www.hedtags.org/display_hed.html

click on "select version" and pick `1.3` to get the 1st generation HED tags. See the following screenshot for a preview:

image


Where do we go from here?

larsoner:
Yikes, sounds like we don't really support even the first version of HED. Thanks for taking a deep dive into it @sappelhoff . Maybe we should say we support something HED-like, or offer very basic support for parsing HED tags?

> Yet another unfortunate point is, that the 1st generation HED tags (which were still relatively basic) already DID have a specific vocabulary which they refer to in the paper (ref 7: https://sccn.ucsd.edu/wiki/HED)

If it would help people we could add a parser for this. Maybe `mne.events.validate_hed(event_id)` or so. But nobody has brought this up so far so maybe we don't need it...

My vote would be to scale back our claims in the documentation to align with the above information from @sappelhoff .

anaradanovic:
Thanks for the feedback everyone!!!

I just some questions for clarification :

-Should they just be described as event tags/descriptors for events, allowing for simple parsing during data analysis? 
-Do we want to still reference the paper in the tutorial and epochs file? I'm iffy on this, considering it is the same term that is used in that paper but if MNE doesn't support it it may be misleading if someone reads the paper but doesn't find the same functionality in MNE. Perhaps saying that it's related to [reference]?
-Still adding the skip to the spellchecker, yes?

agramfort:
for me there are 2 different issues:
1) document / cross-reference with HED (easy to do here)
2) add better support for HED in MNE (harder)

let's stick to 1) here. We can mention HED insert some more details as given
by Stefan and cross-refeence with EEGLAB doc materials.

>

hoechenberger:
> Thanks for the feedback everyone!!!
> 
> I just some questions for clarification :
> 
> -Should they just be described as event tags/descriptors for events, allowing for simple parsing during data analysis?
> -Do we want to still reference the paper in the tutorial and epochs file? I'm iffy on this, considering it is the same term that is used in that paper but if MNE doesn't support it it may be misleading if someone reads the paper but doesn't find the same functionality in MNE. Perhaps saying that it's related to [reference]?
> -Still adding the skip to the spellchecker, yes?

Hello, I'm sorry that this blew up so much, we didn't anticipate this when we added the issue to the Sprint Todo! Thank you for sticking around nevertheless, it's highly appreciated! 👌

My personal opinion is that we could very well add "HED" to the glossary, briefly describe what it is, and state that MNE supports something somewhat similar by allowing you to group experimental events via forward slashes.

Then in the rest of the docs, we should probably avoid the term HED, but I'm not sure which phrasing we should use ideally. Maybe just something like, "event type grouping"?

agramfort:
ok for me

the way I explain the / syntax in MNE is a list of tags. So we "tag" the
events with certain keywords
that allows smart selection.

>

hoechenberger:
Tags sounds good! We could use the term "nested tags" maybe... "tags can be nested"

agramfort:
why do you say nested?

I use A/B or B/A for mne it's the same

>

hoechenberger:
isn't that some kind of nesting? 🧐

anaradanovic:
maybe nesting implies some sort of hierarchical order/dependencies? I could say a "list of tags separated using a '/' ", making clear that it's not a list datatype.

agramfort:
for me it's not nested as one does not belong to another. It's not
hierarchical

>

anaradanovic:
> for me it's not nested as one does not belong to another. It's not hierarchical
> […](#)

I agree! I'll do as I suggested ^^ and see how the wording comes out. :)

hoechenberger:
Ok, great!

anaradanovic:
Let me know how you feel about the new wording! I tried to integrate how I understood how these are used in MNE. I changed the reference to "inspired by []" because I wasn't sure how to say "we use this term but it's not really near the same functionality"

Let me know your thoughts!!! I'm happy to take any further suggestions :)

anaradanovic:
I already see I missed an extra line space... I can fix that on the next round if there are any other changes to be made :)

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -147,18 +147,18 @@
...
+# value pertains only the first (or, in this specific example: the only)
 # response, regardless of side (left or right). To indicate **which** event
 # type (here: response side) was matched, a second column is added:
-# ``first_recponse``. The values in this column are the event types sans
-# the string used for matching, as it is already encoded as the column name,
-# i.e. in our example, it we expect it to only contain ``'left'`` and
-# ``'right'``.
+# ``first_response``. The values in this column are the event types without the
+# string used for matching, as it is already encoded as the column name, i.e.
+# in our example, it we expect it to only contain ``'left'`` and ``'right'``.
```
britta-wstnr, line 161:
```suggestion
# in our example, we expect it to only contain ``'left'`` and ``'right'``.
```

tutorials/epochs/plot_40_autogenerate_metadata.py
```diff
@@ -430,3 +430,8 @@
 fig.suptitle('ERN (Difference Wave)', fontweight='bold')
 
 fig
+
+###############################################################################
+# References
+# ----------
```
drammock, line 436: Error: `/home/circleci/project/doc/auto_tutorials/epochs/plot_40_autogenerate_metadata.rst:1841: WARNING: Title level inconsistent`

```suggestion
# ~~~~~~~~~~
```

doc/glossary.rst
```diff
@@ -169,6 +169,15 @@ general neuroimaging concepts. If you think a term is missing, please consider
         it is a one-dimensional time series capturing the spatial variability
         of the signal across sensor locations.
 
+    HED
+    hierarchical event descriptors
+        Hierarchical event descriptors (abbr. ``HED``) are tags that use
+        keywords separated by '/' to describe different types of EEG
```
drammock, line 175:
```suggestion
        keywords separated by '/' to describe different types of
```

doc/glossary.rst
```diff
@@ -169,6 +169,15 @@ general neuroimaging concepts. If you think a term is missing, please consider
...
+        experimental events (for example, stimulus/circle/red/left and
+        stimulus/circle/blue/left). These tags can be used to group
+        experimental events and select event types for analysis.
+
```
drammock, line 179:
```suggestion
```

drammock (approved):

mne/epochs.py
```diff
@@ -2183,6 +2184,10 @@ def make_metadata(events, event_id, tmin, tmax, sfreq,
     before or after an epoch, e.g. during the inter-trial interval.
 
     .. versionadded:: 0.23
+
+     References
```
drammock, line 2188:
```suggestion
    References
```

drammock:
thanks @anaradanovic !!

hoechenberger:
🎉

PR #9099, opened 2021-03-16
ICA max_iter new defaults

### LINKED ISSUE #8982: Set ICA's max_iter to a higher default
Currently, `ICA.max_iter` is set to 200 by default. This is one of the few MNE defaults I **always** change when working with data from our lab, and I've seen quite a few support requests indicating that ICA didn't converge within those 200 iterations too, the most recent one here: https://mne.discourse.group/t/cannot-find-ecg-artifacts-during-ica/2778/11?u=richard

I'd like to suggest to increase the `max_iter` default to 1000 to ensure that non-convergence will really be a rather rare exception for most users. I don't think there's many valid use cases for working with non-converged ICA decompositions anyway, or are there? Every time I've encountered the warning about non-converged ICA, the decomposition looked … flaky.

WDYT?

cc @cbrnr

kimcoco:
#### Reference issue
Example: Fixes #8982.

I changed the default max_iter values for different ICA methods - as proposed in the issue. Also added a test function.


hoechenberger (commented):
Great job so far! I only fixed two formatting issues.

@britta-wstnr, @rob-luke, @wmvanvliet and I just discussed the default of the new parameter and we agreed that maybe using `'auto'` would be a better choice than than `None`, as the latter might lead to confusion. Do you think you could change from `None` to `'auto'`? Than it should be good to merge!

mne/preprocessing/ica.py
```diff
@@ -195,9 +195,10 @@ class ICA(ContainsMixin):
         Additional parameters passed to the ICA estimator as specified by
         ``method``.
     max_iter : int
```
agramfort, line 197:
```suggestion
    max_iter : int | None
```

mne/preprocessing/ica.py
```diff
@@ -195,9 +195,10 @@ class ICA(ContainsMixin):
...
-        Maximum number of iterations during fit. Defaults to 200. The actual
-        number of iterations it took :meth:`ICA.fit` to complete will be stored
-        in the ``n_iter_`` attribute.
+        Maximum number of iterations during fit. Defaults to 1000 for ``'fastica'``
```
agramfort, line 198:
```suggestion
        Maximum number of iterations during fit. If None, it will default to 1000 for ``'fastica'``
```

mne/preprocessing/tests/test_ica.py
```diff
@@ -214,6 +214,20 @@ def test_ica_noop(n_components, n_pca_components, tmpdir):
...
+                         ("infomax", 500), ("picard", 500)])
+def test_ica_max_iter_(method, max_iter_default):
+    """Test that ICA.max_iter is set to the right defaults."""
+    _skip_check_picard(method)
+
+    # check that default max_iter comes out for no input
+    ica = ICA(n_components=3, method=method)
+    assert ica.max_iter == max_iter_default
+    # check that user input comes out unchanged
+    ica = ICA(n_components=3, method=method, max_iter=2000)
```
agramfort, line 227: how fast is this test? any chance to make it faster using super high tolerance?

mne/preprocessing/ica.py
kimcoco: @agramfort, @hoechenberger - Thank you for the suggestions! As you can see above, Richard and other mentors discussed whether to make the default 'auto', and I already implemented that locally (just haven't pushed the changes yet); do you think the None option would be better?

mne/preprocessing/tests/test_ica.py
kimcoco: No ICA is fitted in this little test, it just asserts that the value in the class parameter is set right. So timing is fast (0.3 s), and I wouldn't know how to speed it up, as the values in n_components or max_iter don't make a difference at this point...(?)

mne/preprocessing/ica.py
agramfort: +1 for 'auto' and disabling None option.

mne/preprocessing/tests/test_ica.py
agramfort: ok

larsoner:
One critical point from the issue:

> Needs a deprecation cycle.

So we should have `max_iter=None` in this PR / version 0.23 which behaves as `max_iter=200`, which will change to `max_iter='auto'` in 0.24. @kimcoco feel free to take a stab at this, or I can push commits (deprecations can be annoying).

drammock:
also see here: https://github.com/mne-tools/mne-python/pull/9099/checks?check_run_id=2123567105#step:7:10
```
Error: ./mne/preprocessing/ica.py:421:80: E501 line too long (80 > 79 characters)
```

hoechenberger:
> So we should have `max_iter=None` in this PR / version 0.23 which behaves as `max_iter=200`,

Oh that is a good idea, we were already brainstorming how to differentiate between the default and a manually set `max_iter=200`. Introducing `None` for one version fixes this problem.

hoechenberger:
@larsoner I'd like to point out that @kimcoco already implemented a deprecation warning, but you're right we should slightly revise things.

@kimcoco Feel free to approach me tomorrow!

hoechenberger (commented):
Looking good so far!

hoechenberger (approved):

agramfort (approved):

agramfort:
@kimcoco can you see what's wrong with CIs?

hoechenberger:
@kimcoco We forgot to filter the new warning in one more test:
`mne/viz/tests/test_topomap.py::test_plot_topomap_nirs_ica `

Once you add the filter, everything should go green!

hoechenberger:
@kimcoco I've just added the required change, let's hope CI goes green now!

hoechenberger:
The one CI failure is not real, all tests passed. Merging.

hoechenberger:
Thank you so much, @kimcoco!

larsoner:
Looks like this broke several examples:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/7138/workflows/e22614ba-538f-4b99-86fc-15623a5f87f7/jobs/26172

@kimcoco @hoechenberger could you make a quick PR to fix these? If you're working on something else I can make one

hoechenberger:
Oh, I guess we should've run `[circle full]`. I will have a look

PR #9109, opened 2021-03-16
BUG: Fixing Report multiprocessing

### LINKED ISSUE #9101: BUG: parse_folder() with n_jobs>1 for mne.Report causes attribute-error
#### Describe the bug
When changing the n_jobs-parameter for the parse_folder-method of mne.Report to a value greater than 1, an error is thrown.

#### Steps to reproduce
```python
import mne

path = mne.datasets.sample.data_path(verbose=False)
report = mne.Report(verbose=True)
report.parse_folder(path, pattern='*raw.fif', render_bem=False, n_jobs=2)
report.save('test.html', overwrite=True)
```

#### Expected results
No error.

#### Actual results
```python
  File "<decorator-gen-368>", line 22, in parse_folder
  File "C:\Users\marti\PycharmProjects\mne-python\mne\report.py", line 1620, in parse_folder
    r = parallel(p_fun(self, fname, info, cov, baseline, sfreq, on_error,
  File "C:\Users\marti\PycharmProjects\mne-python\mne\parallel.py", line 128, in run
    return fun(*args, **kwargs)
  File "C:\Users\marti\anaconda3\envs\mnedev\lib\site-packages\joblib\parallel.py", line 1054, in __call__
    self.retrieve()
  File "C:\Users\marti\anaconda3\envs\mnedev\lib\site-packages\joblib\parallel.py", line 933, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
  File "C:\Users\marti\anaconda3\envs\mnedev\lib\site-packages\joblib\_parallel_backends.py", line 542, in wrap_future_result
    return future.result(timeout=timeout)
  File "C:\Users\marti\anaconda3\envs\mnedev\lib\concurrent\futures\_base.py", line 440, in result
    return self.__get_result()
  File "C:\Users\marti\anaconda3\envs\mnedev\lib\concurrent\futures\_base.py", line 389, in __get_result
    raise self._exception
AttributeError: 'Report' object has no attribute 'projs'
```

#### Additional information
The error seems to be coming from line 362 in report.py. The report-object doesn't have the _projs_-attribute as if somehow lost attributes when being passed to the new process in __iterate_files_ (line 1620). Other attributes as _data_path_ and _title_ seem to survive.

marsipu:
#### Reference issue
Fixes #9101.

This fixes the bug appearing when calling Report().parse_folder() with n_jobs>1 by adding _projs_ to the list of pickled attributes in _get_state_params().

Solution suggested by @larsoner

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -199,6 +199,8 @@ Bugs
 
 - Fix :func:`mne.read_dipole` yielding :class:`mne.Dipole` objects that could not be indexed (:gh:`8963` by `Marijn van Vliet`_)
 
+- Fix bug when setting n_jobs > 1 in :func:`mne.Report.parse_folder` (:gh:`9109` by `Martin Schulz`_)
+
 API changes
 ~~~~~~~~~~~
 - ``mne.read_selection`` has been deprecated in favor of `mne.read_vectorview_selection`. ``mne.read_selection`` will be removed in MNE-Python 0.24 (:gh:`8870` by `Richard Höchenberger`_)
--- a/mne/report.py
+++ b/mne/report.py
@@ -1659,7 +1659,7 @@ class Report(object):
         # Note: self._fname is not part of the state
         return (['baseline', 'cov_fname', 'fnames', 'html', 'include',
                  'image_format', 'info_fname', 'initial_id', 'raw_psd',
-                 '_sectionlabels', 'sections', '_sectionvars',
+                 '_sectionlabels', 'sections', '_sectionvars', 'projs',
                  '_sort_sections', 'subjects_dir', 'subject', 'title',
                  'verbose'],
                 ['data_path', 'lang', '_sort'])
--- a/mne/tests/test_report.py
+++ b/mne/tests/test_report.py
@@ -627,7 +627,7 @@ def test_split_files(tmpdir, split_naming):
 
 
 def test_survive_pickle(tmpdir):
-    """Testing functionality of Report-Object after pickling"""
+    """Testing functionality of Report-Object after pickling."""
     tempdir = str(tmpdir)
     raw_fname_new = op.join(tempdir, 'temp_raw.fif')
     shutil.copyfile(raw_fname, raw_fname_new)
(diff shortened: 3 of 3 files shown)


marsipu:
test_report returns 6 errors out of 23, which all seem to related to:
```python
ModuleNotFoundError: No module named 'tvtk.custom.renderer
```
And they also appear in the main branch.

larsoner:
Locally I can't replicate that issue, at least on `main`. Can you try:
```
pip uninstall mayavi
pip install --no-cache mayavi
```
and see if it helps?

marsipu:
Yes, that helped (although I just seem to have replaced the same version of mayavi (4.7.2))

larsoner:
If you updated `vtk` in between it breaks Mayavi...

>No everything passes (except one Deprecation-Warning:

Maybe `pip install --upgrade pyreadline`? I have 2.1 and don't see this deprecation warning

marsipu:
> 
> 
> If you updated `vtk` in between it breaks Mayavi...
> 
> > No everything passes (except one Deprecation-Warning:
> 
> Maybe `pip install --upgrade pyreadline`? I have 2.1 and don't see this deprecation warning

pyreadline is already 2.1 and the deprecation-warning stays. But I use Windows, maybe that makes a difference?
By the way, what method would you recommend to keep the development-conda-environment up to date?
`conda env  update -n mnedev -f environment.yml`?
And/or something pip-related?

larsoner:
@marsipu can you add an entry in `doc/changes/latest.inc` to mention this fix? Then I think we can merge. We could in principle add a test but it's often a problem to do so with parallel processing. One potential test would be to pickle and unpickle and ensure `.projs` is there -- this should fail on `master` and pass on this PR I think.

marsipu:
I added a test which just tests `parse_folder()` and `save()` of Report after pickling the object, which fails on `main`. What do you think?

Thanks.

larsoner (approved):
LGTM +1 for merge, thanks @marsipu !

agramfort (approved):

agramfort:
thx @marsipu

hoechenberger:
Great job!!

PR #9074, opened 2021-03-15
Bibtex for limo.py

nordme:
Fixes #8918. Adds bibtex to limo.py

PR #9117, opened 2021-03-16
fix errant merge (references outside docstring)

drammock:
#9074 was merged too soon, it needed a fix.  This is that fix.

@sappelhoff can you review?


sappelhoff (approved):
lgtm

Issue #9111, opened 2021-03-16
Add montage.add_estimated_fiducials


larsoner:
New method of `montage` that, if the montage is in `MRI` / FreeSurfer surface RAS coordinates, will estimate the fiducials (using mne.coreg.get_mni_fiducials) and add them to the montage. For example this docstring:
```
"""Estimate fiducials based on FreeSurfer parameters

Parameters
---------
subject : ...
subjects_dir ...
```
It should help iEEG folks get their montages to be MNE/FIF compatible without requiring a bunch of changes in MNE. For example they would have:
```
montage = ...  # wherever they get their electrodes in MRI (FS Surface RAS) coords
montage.add_estimated_fiducials(subject, subjects_dir)  # adds LPA/Nasion/RPA based on FS files
trans = montage.get_native_head_t()  # head<->MRI trans can now be computed properly
```
Then when they `set_montage`, MNE internally converts to all electrode locations to head coords as usual (for now at least), and iEEG folks pass `trans` to functions that need them (like `plot_alignment`) and now all coordinate frames are properly defined. Still future compatible with the dream of MNE supporting coordinate frames other than head for EEG electrodes (since it just allows us to properly define the Neuromag coordinate frame), but gives us a bit more leeway to punt on that for now.

Discussed with @libertyh as a possible partial solution for iEEG folks.

adam2392:
I can do this really quickly. 

Does this function just call: `mne.coreg.get_mni_fiducials`? 

Is there a unit-test in mind we can apply here?

larsoner:
Excellent!

> Does this function just call: mne.coreg.get_mni_fiducials?

Yes, and adds it to the montage in the right way (whatever that is).

> Is there a unit-test in mind we can apply here?

You can test it on fsaverage, it should match (to numerical precision) the built-in fsaverage trans .fif file we have. You can test it on sample, and it should match to some correlation limit (0.98+ I'm guesssing) the manually-created trans for `sample`.

PR #9118, opened 2021-03-16
[MRG; ENH] Add "add_estimated_fiducials" to DigMontage

adam2392:
#### Reference issue
Closes: #9111 

Will enhance: #9087 and `plot_seeg` related tutorials in the future. Necessary ingredients for dataset though:

- FreeSurfer subject-specific files
- electrode coordinates in FS subject-specific T1w

Add `add_estimated_fiducials` function to `DigMontage`.

Implements unit test... todo

#### Additional information
From BIDS convos of mne sprint chat.


adam2392:
@larsoner I am not sure how to best follow the unit tests you specified in https://github.com/mne-tools/mne-python/issues/9111#issuecomment-800559393

I kind of... pseudocoded it out here, but not familiar w/ all the testing datasets in MNE. Would you be able to provide some more explicit guidance here on what exactly do I load and compare?

larsoner:
Probably something close to this:

https://github.com/mne-tools/mne-python/blob/main/mne/tests/test_coreg.py#L267-L281

You want to load the actual manually-created `sample` `-trans.fif`, and compare it to the one you get if you use your new function. You can do this second part for example by creating an arbitrary 1-channel montage for `sample` (the channel position does not matter) with the MRI coordinate frame, then use your estimation function, and `montage.get_native_head_t()` and see if the rotation angle (`quat_to_rot`) is small and the translation is small.

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
         )
         return positions
 
+    def add_estimated_fiducials(self, subject, subjects_dir):
```
rob-luke, line 350: The %(subject)% isn't getting expanded.
Add a verbose argument, add @verbose before this line.

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
...
+        and then use ``montage.get_native_head_t()`` to get the
+        head <-> MRI transform.
+        """
+        from mne.coreg import get_mni_fiducials
+
+        # get coordframe and fiducial coordinates
+        montage_bunch = _get_data_as_dict_from_dig(self.dig)
+
+        # estimate LPA, nasion, RPA from FreeSurfer fsaverage
+        fids_mri = get_mni_fiducials(subject, subjects_dir)
```
rob-luke, line 373: And pass the verbose through here.

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
         )
         return positions
 
+    def add_estimated_fiducials(self, subject, subjects_dir):
+        """Estimate fiducials based on FreeSurfer ``fsaverage`` subject.
+
+        Parameters
+        ----------
+        %(subject)s
+        %(subjects_dir)s
```
rob-luke, line 356:
```suggestion
        %(subjects_dir)s
        %(verbose)s
```

rob-luke:
The failing ci is due to the doc strings. I made some suggestions about how this could be fixed, there may well be a better fix.

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
...
+        Notes
+        -----
+        Since MNE uses the FIF data structure, it relies on the ``head``
+        coordinate frame. Any coordinate frame can be transformed
+        to ``head`` if the fiducials (i.e. LPA, RPA and Nasion) are
+        defined. One can use this function to estimate those fiducials
+        and then use ``montage.get_native_head_t()`` to get the
+        head <-> MRI transform.
+        """
+        from mne.coreg import get_mni_fiducials
```
larsoner, line 367: All our own imports should be relative (not absolute), even if they're nested
```suggestion
        from ..coreg import get_mni_fiducials
```

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
...
+
```
larsoner, line 368: First step of the function should be to check the coord frame. If it's not MRI, raise an error. The description above should make it clear this will only work for montages in MRI (FreeSurfer surface RAS) coordinates.

mne/channels/montage.py
```diff
@@ -348,6 +347,40 @@ def get_positions(self):
...
+        fids_mri = get_mni_fiducials(subject, subjects_dir)
+
+        # set the coordinate frame based on the coordinate frame
+        # XXX: is this step right? It for example would set mri -> mni_tal
+        for fid in fids_mri:
+            fid['coord_frame'] = montage_bunch.coord_frame
+        _validate_type(item=fids_mri, types=list, item_name='dig')
+
+        # add those digpoints to montage
+        self.dig.extend(fids_mri)
```
larsoner, line 382: Might be better to insert these at the beginning, I think that's usually where we put fiducials in the dig list.

mne/channels/tests/test_montage.py
```diff
@@ -1458,4 +1458,29 @@ def test_plot_montage():
     plt.close('all')
 
 
+@testing.requires_testing_data
+def test_montage_add_estimated_fiducials():
+    data_path = testing.data_path(download=True)
+
```
larsoner, line 1464: It's not a good idea to add `download=True` in tests. And in any case this is already defined up on line 48 (or so), so:
```suggestion
```

mne/channels/montage.py
larsoner: Agreed these edits should work

adam2392:
Is there a way to link the docstring to an existing tutorial? 

E.g.

https://mne.tools/stable/auto_tutorials/source-modeling/plot_source_alignment.html?highlight=freesurfer%20coordinates#coordinate-frame-definitions

If a user isn't understanding why one needs the `mri` coord_frame. I know I asked that question until I read Eric's comment.

adam2392:
https://26055-1301584-gh.circle-artifacts.com/0/dev/generated/mne.channels.DigMontage.html#mne.channels.DigMontage.add_estimated_fiducials

drammock:
> Is there a way to link the docstring to an existing tutorial?

Crossrefs to tutorials are probably most appropriate in a `See also` section, but could also be added inline to the docstring intro paragraph.  To link to that tutorial you can use its target name:

https://github.com/mne-tools/mne-python/blob/1026ad762371b0a802c67af89e163434ac9640b6/tutorials/source-modeling/plot_source_alignment.py#L3

adam2392:
If this CI works, will add the changelog. don't wanna hit the CIs again :p

larsoner:
Feel free to push, I'll probably go through and clear out the GitHub Actions log soon anyway...

larsoner (approved):
LGTM +1 for merge once CIs come back happy (someday)

mne/channels/montage.py
```diff
@@ -348,6 +347,54 @@ def get_positions(self):
...
+
+        Parameters
+        ----------
+        %(subject)s
+        %(subjects_dir)s
+        %(verbose)s
+
+        See Also
+        --------
+        :ref:`tut-plot_source_alignment`
```
larsoner, line 368: This is how it is in `plot_source_alignment.py`
```suggestion
        :ref:`plot_source_alignment`
```

larsoner: (will push a commit for this plus a couple of tiny changes)

adam2392:
https://26126-1301584-gh.circle-artifacts.com/0/dev/generated/mne.channels.DigMontage.html#mne.channels.DigMontage.add_estimated_fiducials

larsoner (approved):

larsoner:
Thanks @adam2392 !

doc/changes/latest.inc
```diff
@@ -127,7 +127,9 @@ Enhancements
 
 - `mne.viz.plot_drop_log` and :meth:`mne.Epochs.plot_drop_log` now omit displaying the subject name in the title if ``subject=None`` is passed (:gh:`9015` by `Richard Höchenberger`_)
 
-- :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103` by |Rotem Falach|_)`
+- Add :func:`mne.channels.DigMontage.add_estimated_fiducials` which will add LPA, RPA and Nasion fiducial points to the ``DigMontage`` object in ``mri`` coordinate frame (:gh:`9118` by `Adam Li`_)
+
+  - :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103` by |Rotem Falach|_)`
```
agramfort, line 132: indent pb

mne/channels/montage.py
```diff
@@ -348,6 +348,60 @@ def get_positions(self):
...
 
+    @verbose
+    def add_estimated_fiducials(self, subject, subjects_dir=None,
+                                verbose=None):
+        """Estimate fiducials based on FreeSurfer ``fsaverage`` subject.
+
+        This takes a montage with the ``mri`` coordinate frame,
+        corresponding to the FreeSurfer RAS (xyz in the volume) T1w
+        image of the specific subject. It will call
+        :func:`mne.coreg.get_mni_fiducials` to estimate LPA, RPA and
```
agramfort, line 359: I don't like to have stuff from coreg module to be imported outside. coreg should be GUI code only I think.

mne/channels/montage.py
```diff
@@ -348,6 +348,60 @@ def get_positions(self):
...
+        Notes
+        -----
+        Since MNE uses the FIF data structure, it relies on the ``head``
+        coordinate frame. Any coordinate frame can be transformed
+        to ``head`` if the fiducials (i.e. LPA, RPA and Nasion) are
+        defined. One can use this function to estimate those fiducials
+        and then use ``montage.get_native_head_t()`` to get the
+        head <-> MRI transform.
+        """
+        from ..coreg import get_mni_fiducials
```
agramfort, line 386: yes nested import is required exactly for this I think.

agramfort (commented):
sorry for coming late to the party :(

doc/changes/latest.inc
larsoner: I fixed this in `main`, it was my error during the mege...

mne/channels/montage.py
larsoner: Agreed we could move it to `source_space.py`, it's where we deal with coordinate frames most of the time anyway

agramfort:
>
> Agreed we could move it to source_space.py, it's where we deal with coordinate frames most of the time anyway
>
>
+1

can you do the PR?

PR #9122, opened 2021-03-17
Deprecate time-frequency DICS beamformer

britta-wstnr:
Deprecating the time-frequency DICS beamformer `tf_dics`. 
This functionality is prone to producing bad results, e.g. when choosing a very short integration window for the covariance matrices. With today's computational power, better options are available, e.g. source localizing epochs and subsequently doing time-frequency analysis in source space directly, or *hilbert beamforming*.
Plus, the functionality can easily be replicated by the user if absolutely needed. 

#### Additional information
I propose to later add an example for hilbert beamforming as a better "replacement" for the `dics_tf` example, together with @kimcoco.

@wmvanvliet @larsoner @agramfort : OK to deprecate?


larsoner:
Just needs a line in the API section of `latest.inc` then should be good to go I think!

agramfort (approved):
+1 for MRG after an entry in what's new in the API session

agramfort:
thx @britta-wstnr !

Issue #7975, opened 2020-07-07
to_data_frame() method for EpochsTFR objects


eioe:
Are there plans to implement a `to_data_frame()` method for `EpochsTFR` and `AverageTFR` objects?

I would like to export/save the according data for further analysis in R. 
I just recently learned about the `to_data_frame()` methods for `Epoch` and `Evoked` objects and find them terrifically useful. Would be great to get the same luxury for tfr data. 

I'm also thankful for hints regarding major obstacles that speak against it or if there are already similar implementations that I'm not aware of. 
Thanks!

larsoner:
I think it would be fine to implement this for the TFR objects, do you want to give it a shot?

eioe:
I will see what I achieve. However, I'm still quite new to MNE and have little experience with PRs into professionnally mainatained projects. So this will take a while and it's likely that I will come back for help. 

If someone else feels like implementing this earlier, please go for it. :)

larsoner:
> However, I'm still quite new to MNE and have little experience with PRs into professionnally mainatained projects. So this will take a while and it's likely that I will come back for help.

No problem! We're here to help people work through the contributing process, so feel free to take a stab at an initial implementation, open a pull request, and we can iterate from there. It can help to read the [contributing guide](https://mne.tools/dev/install/contributing.html#overview-of-contribution-process) to help get oriented with the overall process.

eioe:
Thanks! Yes, I worked through the guide and set up a local dev enviroment. 
I'll come back when I hit a wall.

dengemann:
Sounds cool! Keep me in the loop. We can then add simultaneously a related vignette to mne-r: https://mne.tools/mne-r

eioe:
@drammock & @dengemann 
as discussed earlier, I would like to use the sprint to finally close this issue. 
Can you add it to the Sprint Program please and assign me? 
Btw the same was suggested earlier in #6619.

agramfort:
go for it

>

dengemann:
done @eioe

drammock:
closed by #9124

PR #9124, opened 2021-03-17
ENH: add to_data_frame method to EpochsTFR

eioe:
This is related to issue #7975 

EpochsTFR and AverageTFR do not allow to extract data as a (pandas) data frame which in turn would be handy for export to (e.g.) R. 

To build this functionality, I use the `to_data_frame()` method of the `BaseEpochs` class as template. 
Major differences: 
1. TFR data is 4 dimensional. 
2. `EpochsTFR` did not yet have the `selection` attribute which should be used for the `epoch` index in the resulting data frame. 
3. The method in `BaseEpochs` gets scaled (according to the channel type) before it's written out. For TFR data this seems to be more complicated because it should also be taken into account whether any/which baselining procedure was applied to get meaningful units/scaling. 

I implemented 1. and 2. `EpochsTFR` and `AverageTFR` now both inherit a `to_data_frame`method from `_BaseTFR` which has similar way of functioning as the methods for `Epochs` and `Evoked`. Currently, I'm skipping the scaling for the TFR classes and leave it to the users to interpret/label the output as they should know best what state the data is in.


mne/time_frequency/tfr.py
```diff
@@ -2138,6 +2140,67 @@ def average(self, method='mean'):
...
+    @fill_doc
+    def to_data_frame(self, picks=None, index=None,
+                      scalings=None, copy=True, long_format=False,
+                      time_format='ms'):
+        """Export data in tabular structure as a pandas DataFrame.
+
+        Channels are converted to columns in the DataFrame. By default,
+        additional columns "time", "frequency", "epoch" (epoch number),
+        and "condition" (epoch event description) are added, unless ``index``
+        is not ``None``(in which case the columns specified in ``index`` will
```
britta-wstnr, line 2152:
```suggestion
        is not ``None`` (in which case the columns specified in ``index`` will
```
missing whitespace

agramfort:
don't forget to add some tests

mne/time_frequency/tfr.py
```diff
@@ -2138,6 +2140,67 @@ def average(self, method='mean'):
...
+        be used to form the DataFrame's index instead).
+
+        Parameters
+        ----------
+        %(picks_all)s
+        %(df_index_epo)s
+            Valid string values are 'time', 'freq', and 'condition'.
```
britta-wstnr, line 2159:
```suggestion
            Valid string values are ``'time'``, ``'freq'``, and ``'condition'``.
```

eioe:
I updated the original post to better reflect what happened in the mean time. 
Summary: 
- `EpochsTFR` now also has the attributes `selection` and `drop_log` (which by default it grabs from the `Epochs` instance that's used for the tf decomposition)
- `EpochsTFR` and `AvergaeTFR` now have a `to_data_frame_method` 
- all new functionalities should be covered by tests

mne/time_frequency/tfr.py
```diff
@@ -1012,6 +1017,82 @@ def save(self, fname, overwrite=False, *, verbose=None):
...
+        index = _check_pandas_index_arguments(index, valid_index_args)
+        time_format = _check_time_format(time_format, valid_time_formats)
+        # get data
+        times = self.times
+        picks = _picks_to_idx(self.info, picks, 'all', exclude=())
+        if isinstance(self, EpochsTFR):
+            data = self.data[:, picks, :, :]
+        else:
+            data = self.data[picks, :, :]
+            data = np.expand_dims(data, axis=0)
```
drammock, line 1066: slightly simpler:

```suggestion
            data = self.data[np.newaxis, picks]  # add singleton "epochs" axis
```

mne/time_frequency/tfr.py
```diff
@@ -1012,6 +1017,82 @@ def save(self, fname, overwrite=False, *, verbose=None):
...
+        -------
+        %(df_return)s
+        """
+        # check pandas once here, instead of in each private utils function
+        pd = _check_pandas_installed()  # noqa
+        # arg checking
+        if isinstance(self, EpochsTFR):
+            valid_index_args = ['time', 'freq', 'epoch', 'condition']
+        else:
+            valid_index_args = ['time', 'freq']
```
drammock, line 1055: (nitpick) this is slightly more DRY

```suggestion
        valid_index_args = ['time', 'freq']
        if isinstance(self, EpochsTFR):
            valid_index_args.append(['epoch', 'condition'])
```

mne/time_frequency/tfr.py
```diff
@@ -2064,12 +2176,30 @@ def __init__(self, info, data, times, freqs, comment=None, method=None,
...
             n_epochs = len(data)
             events = _gen_events(n_epochs)
+        if selection is None:
+            n_epochs = len(data)
+            selection = np.arange(n_epochs)
+        if drop_log is None:
+            drop_log = tuple(
+                () if k in selection else ('IGNORED',)
+                for k in range(max(len(events),
+                               max(selection) + 1)))
```
drammock, line 2186: this is a little unwieldy.  can you pull `max(len(events), max(selection) + 1)` out as a variable with an intelligible name?

mne/time_frequency/tfr.py
```diff
@@ -2064,12 +2176,30 @@ def __init__(self, info, data, times, freqs, comment=None, method=None,
...
+        else:
+            drop_log = drop_log
+        # check consistency:
+        assert len(selection) == len(events)
+        assert len(drop_log) >= len(events)
+        assert len(selection) == sum(
+            (len(dl) == 0 for dl in drop_log))
+        # TODO Check for consistency with .metadata?
```
drammock, line 2194: is this cruft or a valid `TODO`?

mne/time_frequency/tfr.py
```diff
@@ -2298,7 +2428,8 @@ def write_tfrs(fname, tfr, overwrite=False, *, verbose=None):
     ----------
     fname : str
         The file name, which should end with ``-tfr.h5``.
-    tfr : AverageTFR instance, or list of AverageTFR instances
+    tfr : AverageTFR instance, list of AverageTFR instances, or EpochsTFR \
+        instance
```
drammock, line 2432:
```suggestion
    tfr : AverageTFR | list of AverageTFR | EpochsTFR
```

mne/time_frequency/tfr.py
```diff
@@ -2349,7 +2482,8 @@ def read_tfrs(fname, condition=None):
 
     Returns
     -------
-    tfrs : list of instances of AverageTFR | instance of AverageTFR
+    tfrs : list of instances of AverageTFR | instance of AverageTFR | \
+        instance of EpochsTFR
```
drammock, line 2486:
```suggestion
    tfr : AverageTFR | list of AverageTFR | EpochsTFR
```

drammock (approved):
Sorry it took a while to get around to reviewing this. It looks good!  I've made only minor tweaks/suggestions.

mne/time_frequency/tfr.py
eioe: I guess this can be ignored. It was a note to myself to check back whether you (mne maintainers) want to see such checks. 
I looked it up, and there are also no consistency checks against (the shape of) `metadata` for other properties of the TFR classes (e.g., `events`) or for `selection` in the `(Base)Epochs` class. So I assume this is/was not seen necessary. 
```suggestion
```

mne/time_frequency/tfr.py
eioe:
```suggestion
            n_epochs_prerejection = max(len(events), max(selection) + 1)
            drop_log = tuple(
                () if k in selection else ('IGNORED',)
                for k in range(n_epochs_prerejection))
```
better? 🙂

mne/time_frequency/tfr.py
```diff
@@ -1049,10 +1049,9 @@ def to_data_frame(self, picks=None, index=None, long_format=False,
         # check pandas once here, instead of in each private utils function
         pd = _check_pandas_installed()  # noqa
         # arg checking
+        valid_index_args = ['time', 'freq']
         if isinstance(self, EpochsTFR):
-            valid_index_args = ['time', 'freq', 'epoch', 'condition']
-        else:
-            valid_index_args = ['time', 'freq']
+            valid_index_args.append(['epoch', 'condition'])
```
drammock, line 1054: I broke your tests; I think this will fix them:

```suggestion
            valid_index_args.extend(['epoch', 'condition'])
```

drammock:
thanks @eioe!

mne/time_frequency/tfr.py
eioe: 🙏 thx

cbrnr:
Looks very nice, great job @eioe!

eioe:
Thanks @cbrnr , I also open #9275 for review then.

cbrnr:
Yes, still on my agenda!

PR #9129, opened 2021-03-17
TST: See if cancellation works [skip circle] [skip azp]

larsoner:


larsoner:
I've seen this working in practice so I'll close

PR #9130, opened 2021-03-17
Add optipng to gitignore

HanBnrd:
Adds `optipng.exe` to the gitignore for Windows users.


larsoner:
Thanks @HanBnrd !

PR #9135, opened 2021-03-17
adding add_reference_channels as a method of Raw/Epochs/Evoked

### LINKED ISSUE #8979: Add add_reference_channels() method to Raw, Epochs, and Evoked
Currently there's `mne.add_reference_channels()`, which takes an instance of `Raw`, `Epochs`, or `Evoked`, and adds reference channels.

It would be great to have this functionality exposed as a **method** too, i.e., `Raw.add_reference_channels()`, `Epochs.add_reference_channels()`, and `Evoked.add_reference_channels()`, just as is the case e.g. with `set_eeg_reference()`

The respective `Raw`/`Epochs`/`Evoked` instance should be modified in-place, and the modified instance should be returned to allow for method chaining.

This would need to be added to `UpdateChannelsMixin` in `mne/channels/channels.py`

Overlapping parts of the method and  `mne.add_reference_channels()` docstrings should be added to the `docdict` to avoid duplication.

Falach:
#### Reference issue
Fixes #8979

add_reference_channels is now also a method of Raw/Epochs/Evoked

#### Additional information
I also implement a little test that checks mainly that the object has changed (the return value is not a copy), there are more tests already at the original function location (mne/io/reference).
Let me know if the docdict is used as you expected.

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1072,6 +1072,7 @@ class UpdateChannelsMixin(object):
             assert all(len(r) == self.info['nchan'] for r in self._read_picks)
         return self
 
+    @fill_doc
     def add_reference_channels(self, ref_channels):
         """ Add reference channels to data that consists of all zeros.
 
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1075,14 +1075,13 @@ class UpdateChannelsMixin(object):
     def add_reference_channels(self, ref_channels):
         """ Add reference channels to data that consists of all zeros.
 
-        Adds reference channels to data that were not included during recording.
-        This is useful when you need to re-reference your data to different
-        channels. These added channels will consist of all zeros.
+        Adds reference channels to data that were not included during
+        recording. This is useful when you need to re-reference your data
+        to different channels. These added channels will consist of all zeros.
 
         Parameters
         ----------
         %(ref_channels)s
-
         Returns
         -------
         inst : instance of Raw | Epochs | Evoked
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1082,6 +1082,7 @@ class UpdateChannelsMixin(object):
         Parameters
         ----------
         %(ref_channels)s
+
         Returns
         -------
         inst : instance of Raw | Epochs | Evoked
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1074,7 +1074,7 @@ class UpdateChannelsMixin(object):
 
     @fill_doc
     def add_reference_channels(self, ref_channels):
-        """ Add reference channels to data that consists of all zeros.
+        """Add reference channels to data that consists of all zeros.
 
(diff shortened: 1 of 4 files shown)


mne/channels/channels.py
```diff
@@ -1072,6 +1072,25 @@ def add_channels(self, add_list, force_update_info=False):
...
+        """ Add reference channels to data that consists of all zeros.
+
+        Adds reference channels to data that were not included during
+        recording. This is useful when you need to re-reference your data
+        to different channels. These added channels will consist of all zeros.
+
+        Parameters
+        ----------
+        %(ref_channels)s
+        Returns
```
britta-wstnr, line 1085:
```suggestion

        Returns
```

britta-wstnr: Add a blank line here.

mne/channels/channels.py
```diff
@@ -1072,6 +1072,26 @@ def add_channels(self, add_list, force_update_info=False):
             assert all(len(r) == self.info['nchan'] for r in self._read_picks)
         return self
 
+    def add_reference_channels(self, ref_channels):
```
drammock, line 1075:
```suggestion
    @fill_doc
    def add_reference_channels(self, ref_channels):
```

britta-wstnr: (Also put that below in the other file!)

drammock: this will also need to be imported at the top of the file:
`from ..utils import fill_doc`

mne/channels/tests/test_channels.py
```diff
@@ -366,6 +366,31 @@ def test_drop_channels():
     pytest.raises(ValueError, raw.drop_channels, 5)  # must be list or str
 
 
+def test_add_reference_channels():
+    """ Test if there is a new reference channel that consist of all zeros"""
+    raw = read_raw_fif(raw_fname, preload=True)
+    raw_original_channels_number = len(raw.ch_names)
```
agramfort, line 372:
```suggestion
    n_raw_original_channels = len(raw.ch_names)
```
I find it more explicit. A var that starts with n_ is a count (int). A var that starts with raw_ suggests it's a raw instance.

mne/channels/tests/test_channels.py
```diff
@@ -366,6 +366,31 @@ def test_drop_channels():
...
+    epochs = Epochs(raw, read_events(eve_fname))
+    epochs.load_data()
+    epochs_original_shape = epochs._data.shape[1]
+    evoked = epochs.average()
+    evoked_original_channels_number = len(evoked.ch_names)
```
agramfort, line 377:
```suggestion
    n_evoked_original_channels = len(evoked.ch_names)
```

agramfort (commented):
besides the nitpicks LGTM

thx @Falach

agramfort (approved):

agramfort:
thx a lot @Falach !

hoechenberger:
Great!!!

PR #9136, opened 2021-03-17
MRG: Add `units` parameter to get_data for Raw

### LINKED ISSUE #8473: FR: Add "unit" parameter to get_data method for Epochs

HanBnrd:
#### Reference issue
Partially addresses #8473:
- [x] Add `units` to *Raw.get_data*
- [ ] ~~Add `units` to *Epochs.get_data*~~

Adds a `units` parameter to the [`get_data()`](https://github.com/mne-tools/mne-python/blob/main/mne/io/base.py#L799-L891) function of `base.py`.

With `units` being either `None`, `str` or `dict`.

#### Additional information
Continuation of #8476.


larsoner:
I would do it in a new `mne/io/tests/test_raw.py:test_get_data_units` function.

HanBnrd:
I think that's ready for review @sappelhoff @larsoner. It does not completely close #8473 because it's only implemented for Raw and not Epochs, but I won't have much time to continue working on it unfortunately.

Doc: https://26384-1301584-gh.circle-artifacts.com/0/dev/generated/mne.io.Raw.html#mne.io.Raw.get_data

mne/io/base.py
```diff
@@ -831,13 +848,80 @@ def get_data(self, picks=None, start=0, stop=None,
...
         start = 0 if start is None else start
         stop = min(self.n_times if stop is None else stop, self.n_times)
         if len(self.annotations) == 0 or reject_by_annotation is None:
-            return self._getitem(
+            getitem = self._getitem(
                 (picks, slice(start, stop)), return_times=return_times)
+            if return_times:
+                data, times = getitem
+                return data * ch_factors[:, None], times
+            return getitem * ch_factors[:, None]
```
sappelhoff, line 924: `None` is valid, and works --> but `np.newaxis` is more explicit IMHO: https://stackoverflow.com/a/945313/5201771

```suggestion
                return data * ch_factors[:, np.newaxis], times
            return getitem * ch_factors[:, np.newaxis]
```

mne/io/base.py
```diff
@@ -847,8 +931,8 @@ def get_data(self, picks=None, start=0, stop=None,
         if len(onsets) == 0:
             data, times = self[picks, start:stop]
             if return_times:
-                return data, times
-            return data
+                return data * ch_factors[:, None], times
+            return data * ch_factors[:, None]
```
sappelhoff, line 935: same as above

```suggestion
                return data * ch_factors[:, np.newaxis], times
            return data * ch_factors[:, np.newaxis]
```

mne/io/base.py
```diff
@@ -888,8 +972,8 @@ def get_data(self, picks=None, start=0, stop=None,
             data, times = self[picks, start:stop]
 
         if return_times:
-            return data, times
-        return data
+            return data * ch_factors[:, None], times
+        return data * ch_factors[:, None]
```
sappelhoff, line 976:
```suggestion
            return data * ch_factors[:, np.newaxis], times
        return data * ch_factors[:, np.newaxis]
```

mne/io/base.py
```diff
@@ -831,13 +848,80 @@ def get_data(self, picks=None, start=0, stop=None,
...
         """
         picks = _picks_to_idx(self.info, picks, 'all', exclude=())
+
+        # Convert into the specified unit
+        ch_factors = np.ones(len(picks))
+
+        si_units = _handle_default('si_units')
+        si_units_splitted = {key: si_units[key].split('/') for key in si_units}
+        prefixes = {'': 1e0, 'd': 1e1, 'c': 1e2, 'm': 1e3, 'µ': 1e6, 'u': 1e6,
+                    'n': 1e9, 'p': 1e12, 'f': 1e15}
```
sappelhoff, line 858: `prefixes` looks like something general that may be reused in the future, so I would put it to `defaults.py`, and get them with `_handle_default('unit_prefixes'`).

mne/io/base.py
```diff
@@ -831,13 +848,80 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            ch_types = self.get_channel_types(picks=picks, unique=True)
+            unit_ch_type = list(set(ch_types) & set(si_units.keys()))
+            if len(unit_ch_type) > 1:
+                raise ValueError('"units" cannot be str if there is more than '
+                                 'one channel type with a unit '
+                                 f'{unit_ch_type}.')
+            units = {unit_ch_type[0]: units}  # make the str argument a dict
+
+        if units is None:
+            pass
```
sappelhoff, line 872: I think I'd have expected something like this:

```python
if isinstance(units, None):
    pass
elif isinstance(units, str):
    ...
elif isinstance(units, dict):
    ...
else:
    raise ValueError()
```

Why are you doing:

```python
if isinstance(units, str):
    ...

if isinstance(units, None):
    pass
elif isinstance(units, dict):
    ...
else:
    raise ValueError()
```    

?

Also, you could save yourself one "else" and one "if" clause there by using `_validate_type`, see: https://github.com/mne-tools/mne-python/blob/bbcba957bcef31af9b49262ff1435a3c45c2e1df/mne/utils/check.py#L355-L365

Like this:

```python
_validate_type(units, types=(None, str, dict), item_name="units")
if str ...
elif dict ...

... no else necessary because we know it's NONE


```

sappelhoff (approved):
very clean diff, thanks a lot for working on this @HanBnrd. I have made a few comments, but nothing that absolutely must be done (only minor improvements).

I'll leave the final review to @larsoner

mne/io/base.py
HanBnrd: Thanks @sappelhoff 

> Why are you doing

This is the way I found to avoid having duplicate code in str and dict cases. What it's currently doing is if it's a string I create an appropriate dict from it. Then, if it was None I do nothing, elif it's a dict of units (either created or provided) I iterate over it to get conversion coefficients, else it's another type so not supported.

I guess I could also do instead:
```python
_validate_type(None, str, dict)
if str:
   convert to dict

# And after that:
if dict:  # originally provided or create from str
    loop over it and get coefficients
# if not it's necessarily None so nothing to do here so no need for an else
```

sappelhoff: I see, thanks. Yes, what you propose looks like less code to me that achieves the same thing while being just as readable.

But let's hear what @larsoner has to say before pushing any commits.

mne/io/tests/test_raw.py
```diff
@@ -582,3 +582,61 @@ def test_describe_df():
                               -9.643554807784935e-13,
                               1.928710961556987e-12,
                               4.146728567347522e-11]))
+
+
+def test_get_data_units():
```
agramfort, line 587: I would suggest to put the conversion logic in a dedicated private function that you can test with a test that is very fast to run.

I would do:

```
# put here a pytest parametrize.
def test_convert_unit(orig_unit, dest_unit, factor):
     factor_ = _get_unit_convert_factor(orig_unit, dest_unit)
     assert factor == factor_
```

doing this will make it obvious what is tested and if it's correct.

then you can have one test that just calls get_data

mne/io/base.py
larsoner: The `if str:` then `if dict` code seems clean / readable and efficient to me

hoechenberger:
Hello, what's the status of this? 😊

HanBnrd:
> 
> 
> Hello, what's the status of this? 😊

Hi @hoechenberger, I didn't have much time this week but will get back on it this weekend :)

hoechenberger:
> Hi @hoechenberger, I didn't have much time this week but will get back on it this weekend :)

Perfect! Do not hesitate to ask for help if you get stuck!

mne/io/tests/test_raw.py
```diff
@@ -582,3 +583,66 @@ def test_describe_df():
...
+    """Test the "units" argument of get_data method."""
+    # Test the unit conversion function
+    assert _get_scaling('eeg', 'uV') == 1e6
+    assert _get_scaling('grad', 'fT/cm') == 1e13
+    assert _get_scaling('csd', 'uV/cm²') == 1e2
```
HanBnrd, line 593: Hi @agramfort, I put the conversion code in a new private function and added a few lines to test it, is that alright?

mne/io/base.py
```diff
@@ -1881,6 +1928,61 @@ def _convert_slice(sel):
         return sel
 
 
+def _get_scaling(ch_type, target_unit):
+    """Return the scaling factor based on the channel type and a target unit.
+
+    Parameters
+    ----------
+    ch_type: str
```
sappelhoff, line 1936:
```suggestion
    ch_type : str
```

mne/io/base.py
```diff
@@ -1881,6 +1928,61 @@ def _convert_slice(sel):
...
+        The channel type.
+    target_unit: str
```
sappelhoff, line 1938:
```suggestion
    target_unit : str
```

mne/io/base.py
```diff
@@ -1881,6 +1928,61 @@ def _convert_slice(sel):
...
+        The target unit for the provided channel type.
+
+    Returns
+    -------
+    scaling: int
```
sappelhoff, line 1943: This is probably a float in most cases, because the prefixes (like `1e6`) are float types

```suggestion
    scaling : float
```

sappelhoff (approved):
some minor comments, but provided that the CIs pass, I am +1 for merging. Thanks a lot @HanBnrd

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
             'bad' are omitted. If 'NaN', the bad samples are filled with NaNs.
         return_times : bool
             Whether to return times as well. Defaults to False.
+        units : str | dict | None
+            Specify the unit(s) that the data should be returned in. If None
```
hoechenberger, line 821:
```suggestion
            Specify the unit(s) that the data should be returned in. If ``None``
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            (default), the data is returned in the channel type specific
```
hoechenberger, line 822:
```suggestion
            (default), the data is returned in the channel-type-specific
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            default units, which are SI units (see
+            :ref:`units` and :term:`data channels`). If str, must be an SI unit
```
hoechenberger, line 824:
```suggestion
            :ref:`units` and :term:`data channels`). If a string, must be an SI unit
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            that will be used to scale the data from all channels of the type
+            associated with that unit. This only works if the data contains one
+            channel type that has a unit (unitless channel types are left
+            unchanged). For example if there are only EEG and stim channels
```
hoechenberger, line 828:
```suggestion
            unchanged). For example if there are only EEG and STIM channels,
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            ``units='uV'`` will scale to micro Volts EEG channels and stim
```
hoechenberger, line 829:
```suggestion
            ``units='uV'`` will scale EEG channels to micro-Volts while STIM
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            channels will be unchanged. Finally, if a dict is provided, keys
```
hoechenberger, line 830:
```suggestion
            channels will be unchanged. Finally, if a dictionary is provided, keys
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            must be channel types, and values must be units to scale the data
+            of that channel type to. For example
+            ``dict(grad='fT/cm', mag='fT')`` will scale the corresponding types
+            accordingly but all other channel types will remain in their
```
hoechenberger, line 834:
```suggestion
            accordingly, but all other channel types will remain in their
```

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
...
+            channel type specific default unit.
```
hoechenberger, line 835:
```suggestion
            channel-type-specific default unit.
```

hoechenberger (commented):
Few nitpicks, otherwise LGTM

mne/io/base.py
```diff
@@ -816,6 +817,22 @@ def get_data(self, picks=None, start=0, stop=None,
             'bad' are omitted. If 'NaN', the bad samples are filled with NaNs.
         return_times : bool
             Whether to return times as well. Defaults to False.
+        units : str | dict | None
+            Specify the unit(s) that the data should be returned in. If
+            ``None`` (default), the data is returned in the
+            channel-type-specific default units, which are SI units (see
+            :ref:`units` and :term:`data channels`). If a string, must be an SI
+            unit that will be used to scale the data from all channels of the
```
HanBnrd, line 825: I feel that we should change the wording here @sappelhoff, because one may understand that they need to use SI units (as defined with their SI prefixes). What do you think? Any suggestions on how to make this clearer?

mne/io/base.py
```diff
@@ -831,13 +848,43 @@ def get_data(self, picks=None, start=0, stop=None,
...
+        # Loop over the dict to get channel factors
+        if isinstance(units, dict):
+            for ch_type, ch_unit in units.items():
+                # Get the scaling factors
+                scaling = _get_scaling(ch_type, ch_unit)
+                if scaling != 1:
+                    ch_types = self.get_channel_types(picks=picks)
+                    indices = [i_ch for i_ch, ch in enumerate(ch_types)
+                               if ch == ch_type]
+                    ch_factors[indices] = ch_factors[indices] * scaling
```
agramfort, line 875:
```suggestion
                    ch_factors[indices] *= scaling
```

mne/io/base.py
```diff
@@ -888,8 +935,8 @@ def get_data(self, picks=None, start=0, stop=None,
             data, times = self[picks, start:stop]
 
         if return_times:
-            return data, times
-        return data
+            return data * ch_factors[:, np.newaxis], times
+        return data * ch_factors[:, np.newaxis]
```
agramfort, line 939: this code will now generate a new copy of the data on return even if no specific scaling is required. Given how much the IO code has been optimized over the years I think this is not OK. Basically no need to multiply by np.ones if ch_factors has not been set.

mne/io/base.py
```diff
@@ -1881,6 +1928,61 @@ def _convert_slice(sel):
...
+    si_units = _handle_default('si_units')
+    si_units_splitted = {key: si_units[key].split('/') for key in si_units}
+    prefixes = _handle_default('prefixes')
+    prefix_list = list(prefixes.keys())
+
+    # Check that the provided unit exists for the ch_type
+    unit_list = target_unit.split('/')
+    si_unit_list = si_units_splitted[ch_type]
+    if len(unit_list) != len(si_unit_list):
+        raise ValueError(f'{target_unit} is not a valid unit.')
```
agramfort, line 1957: this message can be cryptic. Can you say what units would be valid?

mne/io/base.py
```diff
@@ -1881,6 +1928,61 @@ def _convert_slice(sel):
...
+    for i, unit in enumerate(unit_list):
+        valid = [prefix + si_unit_list[i]
+                 for prefix in prefix_list]
+        if unit not in valid:
+            raise ValueError(f'{target_unit} is not a valid unit.')
```
agramfort, line 1962: idem. Maybe something like you try to get V from a gradiometer of something

mne/io/tests/test_raw.py
agramfort: can you improve a bit with eg:

```
    assert _get_scaling('eeg', 'uV') == 1e6
    assert _get_scaling('eeg', 'dV') == 1e1
    assert _get_scaling('eeg', 'pV') == 1e12
    assert _get_scaling('mag', 'fT') == 1e15
    assert _get_scaling('grad', 'T/m') == 1
    assert _get_scaling('grad', 'T/mm') == 1e-3
    assert _get_scaling('grad', 'fT/m') == 1e15
    assert _get_scaling('grad', 'fT/cm') == 1e13
    assert _get_scaling('csd', 'uV/cm²') == 1e2
    ```

hoechenberger: >     assert _get_scaling('eeg', 'dV') == 1e1

`dV`?? Have you been drinking? 😅😊 Is this actually being used? 😱

agramfort:
no but it should be tested as it's advertised as supported :)

>

hoechenberger:
> no but it should be tested as it's advertised as supported :)

Heh. Neat :) 😁

mne/io/base.py
HanBnrd: See 9e5ea86

agramfort (approved):
+1 to MRG if CIs are green

larsoner:
Does it make sense to refactor `mne/tests/test_defaults.py:test_si_units` to make use of the new conversion code? I suspect there are some redundancies now

larsoner:
(also our CIs are failing on all PRs due to a matplotlib issue so feel free to ignore the raw plotting errors)

larsoner (approved):
LGTM +1 for merge assuming CIs are happy (aside from matplotlib failures)!

HanBnrd:
> Does it make sense to refactor mne/tests/test_defaults.py:test_si_units to make use of the new conversion code? I suspect there are some redundancies now

@larsoner, I deleted a good chuck of it as most of the tests were already done in _get_scaling

HanBnrd:
> LGTM +1 for merge assuming CIs are happy (aside from matplotlib failures)!

Should I merge main again once the CI is fixed?

larsoner:
In theory you shouldn't need to, if we just restart the CIs at that point they should get the fix because they build a merged-with-main version of the PR. But you could also merge with `main` in your branch and it would work without needing to manually restart

hoechenberger:
Thank you, @HanBnrd!!

Issue #8149, opened 2020-08-23
Custom montage object not loading to NIRS info


Isamaoz:
Hello,

I am trying to classify EEG and NIRS data using MNE. The datasets I am using were converted into MATLAB format; therefore, I created my data structures from scratch (raw data, montage files, epochs). I did not have any problems loading and classifying the EEG data, but I realised that raw.set_montage() does not support NIRS channels. When I tried to set the montage file to the info object, the coordinates of the sources and detectors are converted into NaNs. 

Do you know if there is an alternative for me to set a montage using NIRS? Thank you!

agramfort:
ping @rob-luke

rob-luke:
Hi @Isamaoz,

Can you provide a minimum working example of how you are setting the montage?

I haven't added added NIRS support to set_montage(), so it doesn't surprise me it doesn't work. Taking a quick look at the code I think the following line is just skipping the nirs channels. But the whole code seems quite EEG specific.

https://github.com/mne-tools/mne-python/blob/master/mne/channels/montage.py#L700

I won't be able to add support for this before 0.21. @Isamaoz are you able to take make an attempt at a PR to fix this issue?

HanBnrd:
Hi @larsoner, this issue also is closed by #9141 🙂

larsoner:
Indeed, thanks!

PR #9141, opened 2021-03-18
WIP: Adds support for fNIRS data to set_montage and new standard artinis montage

### LINKED ISSUE #9093: ENH: Add montage for Artinis OctaMon and Brite23
We would like to add a default montage for the Artinis OctaMon fNIRS device. This would enable to have spacial locations useful for some pre-processing functions (such as the modified Beer Lamber law) and having visualization on a brain map.

octamon

Adding a file in [mne/channels/data/montages/](https://github.com/mne-tools/mne-python/tree/main/mne/channels/data/montages), modifying [`set_montage()`](https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.set_montage) (in Raw but also other data types) as and probably the loading functions.

#### Additional comments
Not sure what file format is best to use (csd/sfp/txt/elc), would you have recommendations?

rob-luke:
#### Reference issue
Add montage for Artinis OctaMon and Brite23 Fixes #9093

1. This adds support for fNIRS data to set_montage
2. Add a new standard montage for the artinis brand device as used in @HanBnrd's lab

#### Additional information
This PR has had feature creep and now contains a suggestion for importing data as csv files, to support the reading of unstructured data this PR now contains new code to check the naming of channels, etc. I have suggested reverting this PR back to its original goal of adding fNIRS support to `.set_montage()` and providing two sample montage files.


mne/channels/data/montages/artinis-brite23.txt
```diff
@@ -0,0 +1,58 @@
+MNI Coordinates
```
rob-luke, line 1: Once we are sure the elc file is correct we can delete this. Must be done before merge.

mne/channels/data/montages/artinis-octamon.txt
```diff
@@ -0,0 +1,35 @@
+MNI Coordinates
```
rob-luke, line 1: Once we are sure the elc file is correct we can delete these. Must be done before merge.

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -74,3 +74,56 @@
...
+###############################################################################
+# Fake some data
+# --------------
+#
+# Here we just create some fake data with the correct names for an
+# artinis octomon system
+
+import mne
+from mne.channels import make_standard_montage
+from mne.channels.tests.test_standard_montage import _simulate_artinis_octomon
```
rob-luke, line 90: Obviously this is a big NO. But just wanted to share this small example with Johann as I cant attend sprint in his timezone today.

rob-luke:
Hi @HanBnrd,

I cant make the sprint in your time zone today so I have tidied up our fork a little so I could share my thoughts with you. I have added a temporary change to one of the docs so we can discuss some questions I had. You can view the render of the doc change here: https://26156-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_30_reading_fnirs_data.html#view-location-of-sensors-over-brain-surface

Now to my questions to you...

#### Is this the API/workflow that you were thinking about?
We have ended up with this workflow...
```python
raw_intensity = load_some_data()
montage = make_standard_montage("artinis-octamon")
raw_intensity.set_montage(montage)
# ... processing as usual
```
will this work for you? If not, what's an alternative API we could use?


#### Locations look wrong
Currently the sources and detectors in 3d look like...
image
From my understanding the sources and detectors should be over the frontal cortex, and the arrangement should be in a dual cross configuration. So this looks close (its frontal and two nice distinct crosses), but its getting squished in to the brain.

I think we have a coordinate system mismatch going on somewhere, either in the raw files or in the loading, storage, or plotting.

Could you dig in to this a bit further? If we can get the sensor positions plotting correctly we can then go back and clean up our functions, make them more generic, add tests etc. @larsoner has helped me with coord system issues before, but he isn't on your timezone either. You might need to ask around in the available mentors room for someone who can help.

mne/channels/data/montages/artinis-brite23.txt
rob-luke: For those playing along at home...

In MNE we use the terminology source (S) and detector (D). These artinis files use the terminology transmitter (Tx) and receiver (Rx).

HanBnrd:
Hi @rob-luke,
> **Is this the API/workflow that you were thinking about?**

Yes the API looks good to me :)

> Locations look wrong

At the moment I'm having a look at `info['dig']` to see if it may come from that, because at the moment we just modify the `info['chs'][i]['loc']`.
Having a look at `_set_montage()` it is updating in place 3 things as far as I understand:
- `info['chs'][i]['loc']`
- `info['dig']`
- `info['dev_head_t']` (which seems to be only useful for MEG so not our case)

rob-luke:
For reference here is a picture of the location of optodes as provided by the vendor.

HanBnrd:
It's probably not an issue related to `info['dig']` but I guess it's better to update it as well.

mne/channels/montage.py
```diff
@@ -831,14 +862,22 @@ def _backcompat_value(pos, ref_pos):
...
+        # XXX info['dig'] modified in place
         info['dig'] = _format_dig_points(digpoints, enforce_order=True)
 
         if mnt_head.dev_head_t is not None:
+            # XXX info['dev_head_t'] modified in place
             info['dev_head_t'] = Transform('meg', 'head', mnt_head.dev_head_t)
 
+        fnirs_picks = _picks_to_idx(info, 'fnirs', allow_empty=True)
+        if len(fnirs_picks) > 0:
+            info = _set_montage_fnirs(info, montage)
```
larsoner, line 874: To avoid the pain of the next person errantly using `montage` instead of `mnt_head`, after conversion above we could `del montage` so that it's no longer in the namespace as a strong hint that the montage-in-head should be used by the subsequent code.

HanBnrd:
This is how it looks now with the fix thanks to @larsoner's help :)
artinis

mne/channels/montage.py
```diff
@@ -699,6 +729,7 @@ def _set_montage(info, montage, match_case=True, match_alias=False,
 
     if isinstance(montage, DigMontage):
         mnt_head = _get_montage_in_head(montage)
+        del montage
```
HanBnrd, line 732: Is that what you meant @larsoner?

rob-luke:
Nice! I had a feeling it was the coord frame 🤔 Now that things are correct we just need to make the code nice and add tests, docs etc.

But before that, can you try this on your local data and confirm it helps to load your labs files, that's why we are doing this :)

HanBnrd:
> Now that things are correct we just need to make the code nice and add tests, docs etc.
But before that, can you try this on your local data and confirm it helps to load your labs files, that's why we are doing this :)

Yes, I'll do that tomorrow :)

HanBnrd:
https://26236-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_30_reading_fnirs_data.html

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -74,3 +74,56 @@
              recommended.
 
 """  # noqa:E501
+
+
+
+
```
larsoner, line 80:
```
Error: ./tutorials/io/plot_30_reading_fnirs_data.py:81:1: E303 too many blank lines (4)
```
```suggestion

###############################################################################
```
But really you could make this first part a part of the block-comment above.

HanBnrd:
Hi @rob-luke,

So it works with data from the actual OctaMon. I used a file from an experiment where I exported data in real time from OxySoft (Artinis' software) into a CSV thanks to TCP.
See the experiment interface [here](https://gitlab.com/HanBnrd/workload-bci).

```python
import mne
import numpy as np

from mne.channels import make_standard_montage
from mne.channels.tests.test_standard_montage import _simulate_artinis_octamon


data = []
f = open(PATH + 'raw.csv', 'r', encoding='utf8')
f.readline()  # skip headers
for line in f:
    data.append(line.split(',')[1:])  # first col is timestamp
data = np.array(data)
data = np.swapaxes(data, 0, 1)

ch_names = ['D1_S1 hbo', 'D1_S1 hbr', 'D1_S2 hbo', 'D1_S2 hbr',
            'D1_S3 hbo', 'D1_S3 hbr', 'D1_S4 hbo', 'D1_S4 hbr',
            'D2_S5 hbo', 'D2_S5 hbr', 'D2_S6 hbo', 'D2_S6 hbr',
            'D2_S7 hbo', 'D2_S7 hbr', 'D2_S8 hbo', 'D2_S8 hbr']
ch_types = ['hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr']
sfreq = 10.  # Hz
info = mne.create_info(ch_names=ch_names, ch_types=ch_types, sfreq=sfreq)
raw = mne.io.RawArray(data, info, verbose=True)

raw.set_montage('artinis-octamon')

raw.plot_sensors()

subjects_dir = mne.datasets.sample.data_path() + '/subjects'
mne.datasets.fetch_fsaverage(subjects_dir=subjects_dir, verbose=True)

fig = mne.viz.create_3d_figure(size=(800, 600), bgcolor='white')
fig = mne.viz.plot_alignment(raw.info, show_axes=True,
                             subject='fsaverage', coord_frame='mri',
                             trans='fsaverage', surfaces=['brain', 'head'],
                             fnirs=['channels', 'pairs',
                                    'sources', 'detectors'],
                             dig=True, mri_fiducials=True,
                             subjects_dir=subjects_dir, fig=fig)
mne.viz.set_3d_view(figure=fig, azimuth=70, elevation=100, distance=0.4,
                    focalpoint=(0., -0.01, 0.02))
```
Unfortunately I cannot share the data used here with the above code.

Also, once there are loaders for oxy3 and oxy4 files (the proprietary format) it'll be more straight forward hopefully.
2d
3d

I'll try to see if I can make some artificial evoked topo like [this](https://mne.tools/dev/auto_tutorials/preprocessing/plot_70_fnirs_processing.html).

HanBnrd:
```python
epochs = mne.make_fixed_length_epochs(raw, duration=5.)
epochs.baseline = (None, 1)
times = np.arange(0., 4.9, 1.)
topomap_args = dict(extrapolate='local')
epochs.average(picks='hbo').plot_joint(times=times, topomap_args=topomap_args)
```
Evoked plot working fine :)
image

HanBnrd:
Just so we have this somewhere:
The vendor confirmed that the montages are for adult's head.

tutorials/io/plot_30_reading_fnirs_data.py
HanBnrd: I was thinking we could probably make a separate tutorial, I'm currently trying to find an OctaMon and Brite 23 sample dataset that could be shared publicly.

HanBnrd: Something like: "How to set an fNIRS montage"

rob-luke: Setting a montage for fNIRS is exactly the same as for other modalities. Why does it need its own example? I suggest it does not. Instead we should point people to the existing tutorial on setting montages?

rob-luke:
@HanBnrd I tweaked the documentation as we discussed with @drammock. I will probably touch this a few more times today to get the linking working etc so that is ready for you when you wake up. I will post a render here when I think its building correctly for you to check.

tutorials/io/plot_30_reading_fnirs_data.py
rob-luke: We discussed in person. Marking as resolved.

rob-luke:
Ok @HanBnrd I am done for the day. You can find the latest render at : https://26356-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_30_reading_fnirs_data.html

Please edit as you wish. There are also some tests that need to be updated or extended. I will leave this to you now. Ping me when you want a review.

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -59,18 +60,140 @@
...
+# either hbo or hbr, and wavelength is specified in nm.
+ch_names = ['D1_S1 hbo', 'D1_S1 hbr', 'D1_S2 hbo', 'D1_S2 hbr',
+            'D1_S3 hbo', 'D1_S3 hbr', 'D1_S4 hbo', 'D1_S4 hbr',
+            'D2_S5 hbo', 'D2_S5 hbr', 'D2_S6 hbo', 'D2_S6 hbr',
+            'D2_S7 hbo', 'D2_S7 hbr', 'D2_S8 hbo', 'D2_S8 hbr']
+
+ch_types = ['hbo', 'hbr', 'hbo', 'hbr',
+            'hbo', 'hbr', 'hbo', 'hbr',
+            'hbo', 'hbr', 'hbo', 'hbr',
+            'hbo', 'hbr', 'hbo', 'hbr']
```
HanBnrd, line 117: @rob-luke @larsoner @drammock, I'd like to add a file in mne/channels/data/montages/ that would contain the source detector mapping for fNIRS montages since we have them in the vendor provided file (https://github.com/mne-tools/mne-python/pull/9141/commits/e0f8cc64fbdd67e7169af324e6eb4af23099f657) and work on a function to load them.
Something like "optode-channels.txt" that would contain the mappings and wavelengths for each fNIRS montage.

The way I see it is like this:
```python
# Like this to get optical density channels
ch_names, ch_types = default_fnirs_channels('artinis-octamon', type='od')

# Or this to get concentration changes channels
ch_names, ch_types = default_fnirs_channels('artinis-octamon', type='hb')

# And then
info = mne.create_info(ch_names=ch_names, ch_types=ch_types, sfreq=sfreq)

raw.set_montage('artinis-octamon')
```
This will save user's time when loading the data before setting the montage.

rob-luke: I don't think we should do this. My thoughts are:

1. Artinis is now exporting their data in SNIRF format so this will only help for legacy files.
2. Artinis are currently writing an MNE reader for their non snirf data that should be ready soon, that would make this suggestion redundant.
2. legacy datasets should be converted to snirf and then imported in to MNE. The society for fNIRS provides tools to convert data to snirf. If your dataset can't be converted using their tools then we should open an issue or PR fix there.
5. In this pr we provide an example of how to load legacy data, but we state that it's not supported due to the lack of standardisation in how people use csvs to store fNIRS data. If we start writing more functions to load these file formats then we will need to provide support for them. 
3. I don't want to encourage loading data in this way, as the community is trying to consolidate on the agreed snirf format.
4. It's extra code to maintain and doesn't really reduce the lines of code in a script much.

However, if everyone else wants this feature then I suggest it's implemented in another PR as I think it will have some complications (mainly setting the wavelength fields consistently), and this pr is nicely self contained as is.

mne/channels/tests/test_standard_montage.py
```diff
@@ -73,3 +78,58 @@ def test_standard_superset():
...
+                'D1_S3 hbo', 'D1_S3 hbr', 'D1_S4 hbo', 'D1_S4 hbr',
+                'D2_S5 hbo', 'D2_S5 hbr', 'D2_S6 hbo', 'D2_S6 hbr',
+                'D2_S7 hbo', 'D2_S7 hbr', 'D2_S8 hbo', 'D2_S8 hbr']
+    ch_types = ['hbo', 'hbr', 'hbo', 'hbr',
+                'hbo', 'hbr', 'hbo', 'hbr',
+                'hbo', 'hbr', 'hbo', 'hbr',
+                'hbo', 'hbr', 'hbo', 'hbr']
+    sfreq = 10.  # Hz
+    info = create_info(ch_names=ch_names, ch_types=ch_types, sfreq=sfreq)
+    raw = RawArray(data, info, verbose=True)
```
HanBnrd, line 100: I modified the test to create from a numpy array instead of using the modified NIRx so it is less confusing.

mne/channels/tests/test_standard_montage.py
```diff
@@ -73,3 +78,58 @@ def test_standard_superset():
...
+    trans_octamon = compute_native_head_t(montage_octamon)
+    montage_brite = make_standard_montage("artinis-brite23")
+    trans_brite = compute_native_head_t(montage_brite)
+    fif = op.join(op.dirname(__file__), '..', '..', 'data', 'fsaverage',
+                  'fsaverage-trans.fif')
+    fsaverage = read_trans(fif)
+    assert_array_almost_equal(list(trans_octamon.values())[2],
+                              list(fsaverage.values())[2])
+    assert_array_almost_equal(list(trans_brite.values())[2],
+                              list(fsaverage.values())[2])
```
HanBnrd, line 135: @larsoner is that what you meant when you mentioned comparing to fsaverage? They are not the same however... Not sure how I should fix that.

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -60,17 +61,135 @@
...
+
+raw.set_montage('artinis-octamon')
+
+# View the position of optodes in 2D to confirm the positions are correct.
+raw.plot_sensors()
+
+
+###############################################################################
+# To validate the positions were loaded correctly it is also possible
+# to view the location of the sources (black), detectors (red),
```
HanBnrd, line 164: @rob-luke, the OctaMon has 1 detector in the middle surrounded by the 4 sources for each side, on the 3D plot we see red point in middle surrounded by black points, is that how it is supposed to be (red=detector and black=source) and do you have the same colour matching for NIRx?

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -60,17 +61,135 @@
...
+# :ref:`tut-sensor-locations`, :ref:`plot_source_alignment`, and
+# :ref:`ex-eeg-scalp`.
+#
+# Below is an example of how to load the optode positions for an Artinis
+# OctaMon device. However, many fNIRS researchers use custom optode montages,
+# in this case you can generate your own .elc file (see `example file
+# <https://github.com/mne-tools/mne-python/blob/main/mne/channels/data
+# /montages/standard_1020.elc>`_) and load that instead.
+
+raw.set_montage('artinis-octamon')
```
HanBnrd, line 156: @rob-luke, I removed
```python
# To load a custom montage use:
# raw.set_montage('/path/to/custom/montage.elc')
```
as it is not possible to set a montage from a custom file (even with other modalities than fNIRS).
Maybe that should be a separate PR if that's really needed.

mne/channels/tests/test_standard_montage.py
HanBnrd: https://github.com/mne-tools/mne-python/pull/9141/checks?check_run_id=2156410557#step:12:3319

tutorials/io/plot_30_reading_fnirs_data.py
HanBnrd: https://26372-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_30_reading_fnirs_data.html#applying-standard-sensor-locations-to-imported-data

rob-luke: Can you please check. Colors should be consistent.

HanBnrd: pyvista

I think that's okay actually because the NIRx has 8 sources and 16 detectors according to `ch_names`.

mne/channels/tests/test_standard_montage.py
rob-luke: THIS WILL NOT WORK. You must keep the naming structure as I described. All the downstream processing and MNE functions expect this.

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -60,17 +61,135 @@
...
+
+# In MNE the naming of channels MUST follow this structure of
+# `S#_D# type` or `S#_D# wavelength`, where # is replaced
+# by the appropriate source and detector number, type is
+# either hbo or hbr, and wavelength is specified in nm.
+
+ch_names = ['D1_S1 hbo', 'D1_S1 hbr', 'D1_S2 hbo', 'D1_S2 hbr',
+            'D1_S3 hbo', 'D1_S3 hbr', 'D1_S4 hbo', 'D1_S4 hbr',
+            'D2_S5 hbo', 'D2_S5 hbr', 'D2_S6 hbo', 'D2_S6 hbr',
+            'D2_S7 hbo', 'D2_S7 hbr', 'D2_S8 hbo', 'D2_S8 hbr']
```
rob-luke, line 111: THIS WILL NOT WORK. You must keep the naming structure as I described. All the downstream processing and MNE functions expect this.

tutorials/io/plot_30_reading_fnirs_data.py
rob-luke: In this picture red is a source. You can tell because the short detectors are black.

tutorials/io/plot_30_reading_fnirs_data.py
HanBnrd: That's the issue indeed because of this:
```python
source, detector = ch.split(' ')[0].split('_')
```
Will fix it.

HanBnrd: Regex is probably better actually.

rob-luke:
> it is not possible to set a montage from a custom file 

I thought the entire point of this PR was to add generic support for adding custom montages. Otherwise we have just written an Artinis reader, which we did not want to do because they are already working on that. 

Just a sanity check for me: can you confirm that for the Brite23 the researcher has no choice of optode placement, and that the montage is completely fixed?

I have a few additional concerns:
1. This method will not work for loading raw or optical density data. As it does not set the wavelength field in the info correctly. I have added a comment to this effect in the example. 
2. Is there any way we can validate that we are indeed reading in your artinis files correctly? We dont need to add it to official tests, but if you could just confirm that you have locally validated the channel orders etc are correct against another software.

-------------------

The more I think about it, the more I think that adding support for people to load random CSV files is going to be a nightmare. Johann and I are very much on the same mind wavelength, but even we are getting mixed up. I would really rather just point people to the society for fnirs format and tell them to first convert their data to the official file format SNIRF. If someones dataset is not compatible with the SNIRF converters, our effort would be better placed assisting in that project, rather than duplicating all their effort here. If there was no other way to get your data in to MNE then I would support loading CSVs. However, there is a conscious and coordinated approach by the fNIRS society to standardise the file format and provide tools to convert data to this format. FYI @agramfort @larsoner @drammock 

It's my opinion that by providing a detailed example of how to load data in this way we are implicitly saying we support this approach of loading random T/CSVs (even if we state the opposite in the text). To dissuade people from even attempting to load CSV files could we remove the example and instead just include a small note in this tutorial saying ..  
"Loading of legacy data not in the above mentioned formats is not currently supported. However, advanced users may load raw data (see tutorial: Creating MNE-Python data structures from scratch) and use the method set_montage()."  We can keep the current example in the tests for advanced users.

fNIRS montages are more complex than EEG. In fNIRS its not just about the location of the optode, but its also dependent on the order in which the light sources were turned on, not all sensors are recording meaningful data at all times. This adds an extra layer of complication and is why I build in a strict naming structure for the ch_names and ordering.

tutorials/io/plot_30_reading_fnirs_data.py
rob-luke: sorted was a nice solution too

mne/channels/tests/test_standard_montage.py
```diff
@@ -73,3 +78,59 @@ def test_standard_superset():
...
             assert_allclose(c_1005[key], value, atol=1e-4, err_msg=key)
+
+
+def _simulate_artinis_octamon():
+    """Simulate artinis OctaMon channel data from numpy data.
+
+    This is to test data that is imported with missing or incorrect montage
+    info. This data can then be used to test the set_montage function.
+    """
+    data = np.random.normal(size=(16, 100))
```
rob-luke, line 89: Yes! This is much better.

HanBnrd:
> I thought the entire point of this PR was to add generic support for adding custom montages. Otherwise we have just written an Artinis reader, which we did not want to do because they are already working on that.

I don't know about how advanced they are in their reader project but it's not sure whether it will include default locations (maybe it'll require to have digitized locations). This is hard to tell because there is no PR open about this, do you have more info about it? In any case if they plan to use default locations, our set_montage will be a good starting point for them, and enable in the meantime to load data manually from csv.

> Just a sanity check for me: can you confirm that for the Brite23 the researcher has no choice of optode placement, and that the montage is completely fixed?

It is physically possible to move the optodes (like some EEG caps) but it is clearly indicated on the cap where to put each one. So no problem here.

> 1. This method will not work for loading raw or optical density data. As it does not set the wavelength field in the info correctly. I have added a comment to this effect in the example.

I didn't realise the mbll was using *info['chs'][#]['loc']* (I thought it was parsing the channel names). I can modify *_set_montage_fnirs* so that it works for raw data.

Overall I think that being able to load data from numpy arrays is very useful. This is quite easy to get back on your feet with EEG because pretty much every cap uses the 10-20 system but much more difficult with fNIRS (because of the source-detector distances required which prevents this 10-20 angle based positioning). Having a way to get positions with fNIRS will be very helpful.

This PR would enable to have the functions to apply fNIRS montages (with example with Artinis montages), and make it much easier to implement partial support for other devices which don't have loaders in MNE and are not in SNIRF (which is still the case for most devices). To have this partial support (= loading from eg. csv and having channel locations) it will just require to add an MNI coordinate elc file for that specific device.

rob-luke:
Just to be clear @HanBnrd, I am not suggesting to abandon this PR, I am just suggesting that we point most users to use SNIRF and remove the tutorial as it is implicitly indicates support for this approach and I think will result in endless support requests. (even more so than EEG due to the extra complexity of fNIRS data).

I already get several emails a week for support on the nirx reader (I ask them all to raise issues but they rarely do), and that is quite robust. You are a very advanced programmer and not the typical user.

HanBnrd:
Yes sure :)

> as it is not possible to set a montage from a custom file (even with other modalities than fNIRS).

Oh actually it kinda is but with extra steps, by creating a montage with *[make_dig_montage](https://mne.tools/stable/generated/mne.channels.make_dig_montage.html#mne.channels.make_dig_montage)* instead of *make_standard_montage*.

tutorials/io/plot_30_reading_fnirs_data.py
rob-luke: > Oh actually it kinda is

So maybe we can add in a small comment like before that would be something like `raw.set_montage(make_dig_montage("path/to/file))`, or whatever the correct is.

HanBnrd: I'm afraid it'll add a lot of comments as it does not work directly from an elc file with this function, it requires arrays, so there is some manual parsing of the elc to do in that case (*_read_elc* is a private method). Should we just point to the *make_dig_montage* doc maybe?

rob-luke: Lets skip for now 👍 leave it out.

rob-luke:
@HanBnrd let's take a pause on adding new features and wait for everyone to specify the scope of this PR. And it will be more efficient if you discuss what you want to add in advance, I think your latest commit might be better suited as an improvement to the general nirs functions (and maybe a different PR, smaller well contained PRs are easier to review)...

https://github.com/mne-tools/mne-python/blob/bbcba957bcef31af9b49262ff1435a3c45c2e1df/mne/preprocessing/nirs/nirs.py#L67

HanBnrd:
> @HanBnrd let's take a pause on adding new features and wait for everyone to specify the scope of this PR. And it will be more efficient if you discuss what you want to add in advance, I think your latest commit might be better suited as an improvement to the general nirs functions (and maybe a different PR, smaller well contained PRs are easier to review)...

Yes, good point, I got carried away by tests for this *set_montage* 😅 Thanks for your time on that PR @rob-luke by the way.

I'll add all the related issues in separate issues once we've got more info on what should be part of this PR and what not.

mne/io/meas_info.py
```diff
@@ -774,6 +777,29 @@ def _check_consistency(self, prepend_error=''):
...
+                    int(suffix)
+                except ValueError:
+                    raise RuntimeError(
+                        'Bad info: info["chs"][%d]["ch_name"] has an invalid '
+                        'suffix %s' % (idx, suffix))
+            s, d = name.split('_')
+            if not re.match('^S[0-9]+$', s) or not re.match('^D[0-9]+$', d):
+                raise RuntimeError(
+                    'Bad info: info["chs"][%d]["ch_name"] has an invalid '
+                    'name %s, should follow the S#_D# format' % (idx, name))
```
rob-luke, line 801: There is already a function that check the naming of nirs channels. It goes a step further and also checks that the channels are all paired by source detectors. However, it only works now for raw and optical density data, not haemoglobin. I think it would be better to extend the existing function rather than have two different name checking functions.

The existing function takes a `freqs` argument and checks that each channel pair uses these frequencies. We could change freqs to `pair_vals` and extend the function to take either frequency pairs (`[760, 850]`) OR haemoglobin pairs (`[hbo, hbr]`)

I think this could be done in a small PR self contained PR and merged quite quickly (at least before this big PR). Johann would you like me to do it today (while your timezone is dark) or would you like to?

https://github.com/mne-tools/mne-python/blob/bbcba957bcef31af9b49262ff1435a3c45c2e1df/mne/preprocessing/nirs/nirs.py#L67

rob-luke: @HanBnrd

HanBnrd: Sure, feel free to do this if that's okay for you, I don't have much time to work on it until the end of the week.

rob-luke: Ok I will give it a go. But I hope you have time to give it a real critical review, time for you to get revenge!!

drammock:
> If someones dataset is not compatible with the SNIRF converters, our effort would be better placed assisting in that project, rather than duplicating all their effort here. If there was no other way to get your data in to MNE then I would support loading CSVs.

Caveat: I have no exposure to the NIRS community so I'm trusting what @rob-luke says here. If indeed there is an (open source?) community project to make converters to snirf for all legacy text formats, then I agree that we should NOT show an example on the MNE-Python website of how to skip that step and work directly with the CSVs.

rob-luke:
> If indeed there is an (open source?) community project to make converters to snirf for all legacy text formats

The primary tool for this is the [Homer3 converter](https://github.com/fNIRS/snirf_homer3). HOMER is the established tool in the fNIRS field, they have access to a wide range of measurement devices (and wrote the original `.nirs` data specification which preceded SNIRF). The other tool that can read most data formats and export SNIRF is the [NIRS Toolbox](http://huppertlab.net/nirs-toolbox-2/). There is also a [snirf validator](https://github.com/huppertt/snirfvalidator) (I haven't tested),  and a [python snirf](https://github.com/huppertt/python_snirf) library (I haven't tested). Currently these tools are spread across different repositories, but they are currently trying to move everything in to the [society for fNIRS official GitHub organisation](https://github.com/fNIRS). I plan to have a meeting with the SNIRF developers soon to discuss how the BIDS and SNIRF specification can align better, and can ask on the progress of a generic converter. If I get the impression that their efforts to centralise the conversion to SNIRF is a year away then we can add an example. However, if their efforts are progressing well then we don't need the tutorial.

But to keep things moving I think we try and merge this PR without the tutorial and instead with the short comment "Loading of legacy data not in the above mentioned formats is not currently supported. However, advanced users may load raw data (see tutorial: Creating MNE-Python data structures from scratch) and use the method set_montage()." We can keep the current example in the tests for advanced users.. Given its already written we can always add the tutorial if needed in future.

But im like 75/25 convinced with my own argument here, happy to be dissuaded to keep the tutorial if its common consensus.

HanBnrd:
I agree with @rob-luke.

> "Loading of legacy data not in the above mentioned formats is not currently supported. However, advanced users may load raw data (see tutorial: Creating MNE-Python data structures from scratch) and use the method set_montage()."

I would also add a link to the ch_name format for fNIRS (`"S#_D# wavelength"`) if there is a resource about it, or maybe just a warning here saying what the channel name format must be.

drammock:
> I plan to have a meeting with the SNIRF developers soon to discuss how the BIDS and SNIRF specification can align better, and can ask on the progress of a generic converter. If I get the impression that their efforts to centralise the conversion to SNIRF is a year away then we can add an example.

Thanks for providing links to the various efforts.  I'm going to change my answer now.  It's my impression that there's nothing *for Python* that is close to complete/mature/documented.  Therefore I think showing example code to read from a CSV is a worthwhile service to the NIRS community, as long as it comes with strong wording that this is general guidance and they'll need to figure out how to adapt it to their particular CSV format (as we discussed on Discord last week).  In fact I think the wording can be stronger, and say "If you have access to MATLAB, use HOMER3 to convert to snirf first, then use our snirf reader. If you can't use MATLAB, here is a general schematic recipe for going straight from CSV to MNE-Python; you will likely have to adapt this depending on the system from which your CSV originated."

mne/channels/montage.py
```diff
@@ -725,6 +726,41 @@ def _get_montage_in_head(montage):
...
+    """
+    # Modify info['chs'][#]['loc'] in place
+    num_ficiduals = len(montage.dig) - len(montage.ch_names)
+    picks = _picks_to_idx(info, 'fnirs', exclude=[], allow_empty=True)
+    info._check_nirs_ch_names()
+    for ch_idx in picks:
+        ch = info['chs'][ch_idx]['ch_name']
+        name, suffix = ch.split(' ')
+        if suffix not in ['hbo', 'hbr']:
+            info['chs'][ch_idx]['loc'][9] = int(suffix)
```
HanBnrd, line 746: This will make it work with intensity or OD if I'm not wrong @rob-luke, is that right?

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -61,16 +62,158 @@
...
+.. warning:: This method is not supported. You should convert your data
+             to the `SNIRF <https://github.com/fNIRS/snirf>`_
+             format using the tools provided by the Society
+             for functional Near-Infrared Spectroscopy, and then load it
+             using :func:`mne.io.read_raw_snirf`.
+
+             This method will only work for data that has already been
+             converted to oxyhaemoglobin and deoxyhaemoglobin. It will not work
+             with raw intensity data or optical density data.
+
```
HanBnrd, line 77: Related to the previous comment we could then delete this part if I'm not wrong.
```suggestion
```

HanBnrd:
## Checklist
*Made this checklist to recap what is left to do:*
- [ ] I need to try set_montage on Brite 23 data as well
- [ ] @rob-luke, do you agree that set montage now supports intensity and OD (in addition to Hb)? [Lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-213d577ac140576c72995b52dae928fa209c69041ef736d0b1636ad3e4becedcR745-R746) and [lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-6f0870e641c47408196eed98425a9d97656eacfeeceba1807f3a515e974b7689R74-R77)
- [ ] @larsoner, is `del montage` okay? [Lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-213d577ac140576c72995b52dae928fa209c69041ef736d0b1636ad3e4becedcR797)
- [ ] @larsoner, I added tests to compare `compute_native_head_t(montage)` to fsaverage but they are failing. I would need some help on this please. I'm not sure to what extent they should be equal. [Lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-895975b207874912e552b134dc9e169e2c4da119d21126fb85125d74cdf44088R126-R136) & [failing tests](https://github.com/mne-tools/mne-python/pull/9141/checks?check_run_id=2159192155#step:12:3291)

*(Feel free to comment on what's missing and I'll edit it)*

tutorials/io/plot_30_reading_fnirs_data.py
HanBnrd: Typo [L111](https://github.com/mne-tools/mne-python/pull/9141/files#diff-6f0870e641c47408196eed98425a9d97656eacfeeceba1807f3a515e974b7689R111) also I think.

rob-luke:
> do you agree that set montage now supports intensity and OD (in addition to Hb)

Dunno, add tests to check.

mne/io/meas_info.py
rob-luke: Ok I have expanded `_check_channels_ordered()` to cover both raw, od, and hb data. It has a different API to your function, but provides the same functionality and more (also checks that the pairings are correct).

https://github.com/mne-tools/mne-python/blob/2a8d56b453ccf4e7b3b22a80bdd42d4ac604474e/mne/preprocessing/nirs/nirs.py#L80

But what might be more interesting is the tests...

https://github.com/mne-tools/mne-python/blob/2a8d56b453ccf4e7b3b22a80bdd42d4ac604474e/mne/preprocessing/nirs/tests/test_nirs.py#L206-L363

mne/channels/montage.py
rob-luke: I don't know why the montage setting function would be changing information in the data about light wavelength. That doesn't seem intuitive to me.

Instead you could check that the data structure is correct as is done in the other fNIRS functions and throw an error if its wrong.

See 
https://github.com/mne-tools/mne-python/blob/2a8d56b453ccf4e7b3b22a80bdd42d4ac604474e/mne/preprocessing/nirs/_optical_density.py#L31

or
https://github.com/mne-tools/mne-python/blob/2a8d56b453ccf4e7b3b22a80bdd42d4ac604474e/mne/preprocessing/nirs/_beer_lambert_law.py#L37-L38

rob-luke: Except for hb data it would be 

https://github.com/mne-tools/mne-python/blob/70709f5069ddb413c2719f4d2b41fd6bfea2349a/mne/preprocessing/nirs/_beer_lambert_law.py#L59-L60

mne/channels/tests/test_standard_montage.py
```diff
@@ -73,3 +78,59 @@ def test_standard_superset():
...
+                'hbo', 'hbr', 'hbo', 'hbr',
+                'hbo', 'hbr', 'hbo', 'hbr']
+    sfreq = 10.  # Hz
+    info = create_info(ch_names=ch_names, ch_types=ch_types, sfreq=sfreq)
+    raw = RawArray(data, info, verbose=True)
+
+    return raw
+
+
+def test_set_montage_artinis():
```
rob-luke, line 105: A few additional tests I would like to see:
* What happens if there is less channels in the data than locations in the montage?
* What happens if the channel names don't correspond to the montage?
* What happens if there are more channels in the data than in the montage?
All these things should throw a meaningful error. I think most of this should be picked up by the existing set_montage code, but as fNIRS acts quite differently we should check that reasonable error text is thrown.

mne/io/meas_info.py
rob-luke: Ok after reading the `_set_montage_fnirs` code above I think you can just use the existing `_check_channels_ordered`. But let me know if you think there is an additional use case for this.

mne/channels/tests/test_standard_montage.py
rob-luke: Also we need to test that the set_montage function operates correctly on amplitude, od, and hb data.

HanBnrd:
I tried it on Brite 23 and it is a bit shifted unfortunately. Maybe related to the failing comparison of `compute_native_head_t(montage)` to fsaverage.
brite23

mne/io/meas_info.py
HanBnrd: Hi @rob-luke, I checked to see how I could use *_check_channels_ordered()* instead of *_check_nirs_ch_names()* and it appears to be a bit difficult because it takes a *raw* object while *_set_montage()* only works with *info* objects. I think that would require an API change for *_check_channels_ordered()* or *_set_montage()* which is probably not a good idea in this PR, what do you think we should do?

rob-luke: Good catch @HanBnrd, totally agree with your suggested modification and have opened it in #9280. Thanks for sticking through these gritty parts, the fNIRS support is still evolving 😉.

HanBnrd: Thanks @rob-luke, I'm working on the tests today and will do the switch to use the new API after that 🙂

rob-luke: Cheers, I know its a bit of a grind

mne/channels/montage.py
```diff
@@ -725,6 +726,45 @@ def _get_montage_in_head(montage):
...
+                                      _channel_chromophore,
+                                      _check_channels_ordered)
+    # Modify info['chs'][#]['loc'] in place
+    num_ficiduals = len(montage.dig) - len(montage.ch_names)
+    picks = _picks_to_idx(info, 'fnirs', exclude=[], allow_empty=True)
+    freqs = np.unique(_channel_frequencies(info))
+    if freqs.size > 0:
+        _check_channels_ordered(info, freqs)
+    else:
+        _check_channels_ordered(info, np.unique(_channel_chromophore(info)))
```
rob-luke, line 748: save one line?
```suggestion
    freqs = np.unique(_channel_frequencies(info))
    if freqs.size > 0:
        picks = _check_channels_ordered(info, freqs)
    else:
        picks = _check_channels_ordered(info, np.unique(_channel_chromophore(info)))
```

HanBnrd:
Hi @rob-luke, I've added all the tests, I'll let you review when you have time 🙂

>- [ ]  @larsoner, is `del montage` okay? [Lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-213d577ac140576c72995b52dae928fa209c69041ef736d0b1636ad3e4becedcR797)
>- [ ]  @larsoner, I added tests to compare `compute_native_head_t(montage)` to fsaverage but they are failing. I would need some help on this please. I'm not sure to what extent they should be equal. [Lines](https://github.com/mne-tools/mne-python/pull/9141/files#diff-895975b207874912e552b134dc9e169e2c4da119d21126fb85125d74cdf44088R126-R136) & [failing tests](https://github.com/mne-tools/mne-python/pull/9141/checks?check_run_id=2159192155#step:12:3291)
>- [ ]  The montage with Brite 23 is slightly shifted. It may be an issue?

There are still the above points to be addressed but it may be helpful to have a chat with @larsoner for this. Maybe during the next office hour on Discord if you're planning to be there @larsoner?

mne/channels/tests/test_standard_montage.py
```diff
@@ -73,3 +82,156 @@ def test_standard_superset():
...
+    trans_octamon = compute_native_head_t(montage_octamon)
+    montage_brite23 = make_standard_montage('artinis-brite23')
+    trans_brite = compute_native_head_t(montage_brite23)
+    fif = op.join(op.dirname(__file__), '..', '..', 'data', 'fsaverage',
+                  'fsaverage-trans.fif')
+    fsaverage = read_trans(fif)
+    assert_array_almost_equal(list(trans_octamon.values())[2],
+                              list(fsaverage.values())[2])
+    assert_array_almost_equal(list(trans_brite.values())[2],
+                              list(fsaverage.values())[2])
```
HanBnrd, line 148: Tests failing because compute_native_head_t and fsaverage are different

larsoner: I think we can live with them being not identical, but they should match to some small angle and translation difference. You can get these from the trans matrices, if you're not sure how I can try to push a commit to do it

mne/channels/montage.py
larsoner: yes, sorry for the (very) slow response!

mne/channels/tests/test_standard_montage.py
HanBnrd: Thanks @larsoner, yes if you have time I would be happy to have a bit of help on it as I'm not very familiar with this.

larsoner: Pushed. I'm surprised the translation is > 1 cm. It doesn't seem like this should happen if they are using the standard locations in MNI coordinates for the fiducials. Do these montages have the following locations for the fiducials?
```
>>> import mne
>>> import pprint
>>> pprint.pprint(mne.coreg.get_mni_fiducials('fsaverage'))
[<DigPoint |        LPA : (-80.6, -29.1, -41.3) mm  : MRI (surface RAS) frame>,
 <DigPoint |     Nasion : (1.5, 85.1, -34.8) mm     : MRI (surface RAS) frame>,
 <DigPoint |        RPA : (84.4, -28.5, -41.3) mm   : MRI (surface RAS) frame>]
```
I wonder if what Artinis defines as LPA/RPA differs from how they were defined for `fsaverage-fiducials.fif`. Maybe we should just make a note that this is probably the case and move on. I think for analyses it should be fine as long as you use the `compute_native_head_t(...)` result as your `trans` in plotting functions rather than using `'fsaverage'`.

HanBnrd: Hi @larsoner, thank you very much.

The Artinis montages have the following locations for the fiducials:
- For OctaMon: LPA = (-82.58, -20.09, -43.10), Nz = (0.96, 83.56, -48.63) and RPA = (80.25, -19.67, -43.88)
- For Brite 23: LPA = (-81.41, -17.18, -45.56), Nz = (-4.62, 82.33, -45.74) and RPA = (79.66, -18.72, -45.89)

It seems that when using *compute_native_head_t* as `trans` this solves the shifting problem observed with Brite 23. This is what I get now with
```python
fig = mne.viz.plot_alignment(raw.info, show_axes=True,
                             subject='fsaverage', coord_frame='mri',
                             trans=trans, surfaces=['brain', 'head'],
                             fnirs=['channels', 'pairs',
                                    'sources', 'detectors'],
                             dig=True, mri_fiducials=True,
                             subjects_dir=subjects_dir, fig=fig)
```
image
compared to [the previous comment](https://github.com/mne-tools/mne-python/pull/9141#issuecomment-815970429).

If I understand well, the diamonds are the fsaverage MRI fiducials and the spheres are the Artnins montage fiducials?
Does it mean that the relative positioning of the 3 fiducials is not the same for Artinis and fsaverage, and therefore cannot be solved by a simple translation and rotation of the whole montage?

If the user must not use *fsaverage* for `trans`, is there anything else that should change in the above code for *plot_alignment* (eg. maybe `subject`, `coord_frame` and `mri_fiducials`)?

larsoner: > Does it mean that the relative positioning of the 3 fiducials is not the same for Artinis and fsaverage,

Yes exactly, and given that they are different for two the two Artinis montages, it seems like they did *not* transform these points to MNI space. If they had done that, I don't see why they wouldn't have had the same points for their own two montages.

> and therefore cannot be solved by a simple translation and rotation of the whole montage?

It doesn't mean this. `trans` takes care of the rot+trans we need.

> is there anything else that should change in the above code for plot_alignment (eg. maybe subject, coord_frame and mri_fiducials)?

Nope, what you've done with `trans=compute_native_head_t(mon), subject='fsaverage'` should be enough to get things to align properly, and whether you show `mri_fiducials` and which `coord_frame` you use are just viz options.

HanBnrd:
Hi, I think this is ready for review @rob-luke, @larsoner, @drammock.
Feel free to mark it as ready if you're happy with it @rob-luke 🙂

HanBnrd:
Merging this PR will close #8149 I believe.

larsoner:
I'll take a look after @rob-luke gets a chance

rob-luke:
I'll get a chance early in the work week

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -61,16 +62,143 @@
...
+             and then load it using :func:`mne.io.read_raw_snirf`.
 
+fNIRS measurements often have a non-standardised format that is not supported
+by MNE and cannot be converted easily into SNIRF. However if it is possible to
+store this data in a legacy csv or tsv format, we show here a way to load it
+even though it is not officially supported by MNE due to the lack of
+standardisation of the file format (the naming and ordering of channels, the
+type and scaling of data, and specification of sensor positions varies between
+each vendor). You will likely have to adapt this depending on the system from
+which your CSV originated.
```
rob-luke, line 81: sometimes its csv, other times CSV. which is correct?

rob-luke (commented):
If we are to demonstrate how to generate raw and info from unsuported data, should we somewhere put in a description of the mandatory formatting? I described the mandatory rules at d42951dcd0f4eb89ffc376befeba5a35a6c680b1

And I assume the issue with the missaligned brite23 montage is fixed?

And I cant see a render of the docs in the jobs list, is it missing or have I lost my mind?

mne/channels/montage.py
```diff
@@ -725,6 +726,45 @@ def _get_montage_in_head(montage):
...
+def _set_montage_fnirs(info, montage):
+    """Set the montage for fNIRS data.
+
+    This needs to be different to electrodes as each channel has three
+    coordinates that need to be set. For each channel there is a source optode
+    location, a detector optode location, and a channel midpoint that must be
+    stored. This function modifies info['chs'][#]['loc'] and info['dig'] in
+    place.
+    """
+    from ..preprocessing.nirs import (_channel_frequencies,
```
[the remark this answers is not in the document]
HanBnrd: This is because when I tried to import those functions at the top, I had circular import issues.

mne/channels/montage.py
```diff
@@ -862,6 +903,7 @@ def _backcompat_value(pos, ref_pos):
 
         for name, use in zip(info_names, info_names_use):
             _loc_view = info['chs'][info['ch_names'].index(name)]['loc']
+            # XXX info['chs'][#]['loc'] modified in place
```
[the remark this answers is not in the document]
HanBnrd: It is not directly related but I wrote those comments to pinpoint all the lines where the in-place modifications were happening as it wasn't always obvious at first glance for example for this line 907.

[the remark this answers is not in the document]
HanBnrd: Should I remove them or is it okay?

tutorials/io/plot_30_reading_fnirs_data.py
HanBnrd: All changed to capital letters as they are acronyms.

HanBnrd:
Hi @rob-luke, thanks for reviewing.

> If we are to demonstrate how to generate raw and info from unsuported data, should we somewhere put in a description of the mandatory formatting? I described the mandatory rules at d42951d

It was in an in-line comment but I just made it a warning now.

> And I assume the issue with the missaligned brite23 montage is fixed?

Yes it is solved by not using `trans='fsaverage'` with Artinis montages (cf. [this comment](https://github.com/mne-tools/mne-python/pull/9141#discussion_r614724189)).

> And I cant see a render of the docs in the jobs list, is it missing or have I lost my mind?

I think it was because of the checks timing out, hopefully this will work now. Will paste the link here when it's built.

mne/channels/montage.py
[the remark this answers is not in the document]
rob-luke: Thanks.

rob-luke:
Ok your turn @larsoner 😃

HanBnrd:
> > And I cant see a render of the docs in the jobs list, is it missing or have I lost my mind?
>
> I think it was because of the checks timing out, hopefully this will work now. Will paste the link here when it's built.

Actually CircleCI is not running it seems, I'm not sure where to find the built docs.

HanBnrd:
Hi @rob-luke, an important point related to https://github.com/mne-tools/mne-python/pull/9141#issuecomment-803489405: I just realised that it is actually possible to read any elc file (see https://github.com/mne-tools/mne-python/issues/9309#issuecomment-824746709) and our PR will make it work for fNIRS too:
```python
montage = read_custom_montage('path/to/artinis-brite23.elc', coord_frame='mri')
raw.set_montage(montage)
```
*(This is probably what has been tried in #8149)*

We just need to have `coord_frame='mri'` so that we can compute the head_t later for *plot_alignment* (because plot_alignment cannot use 'fsaverge' otherwise it's shifted):
```python
trans = compute_native_head_t(montage)
```
Should we write something about this in the example?

I think you can add this in the PR description, and also that our PR will close #8149.

mne/channels/montage.py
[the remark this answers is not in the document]
drammock: such comments are useful, but it would be clearer if it didn't start with `XXX` as this is usually a signal for "something to be fixed later / eventually".  Can you change them to say "next line modifies info['chs'][#]['loc'] in-place" or similar?

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -61,16 +62,143 @@
...
-             recommended.
+.. warning:: This method is not supported and users are discoraged to use it.
+             You should convert your data to the
+             `SNIRF <https://github.com/fNIRS/snirf>`_ format using the tools
+             provided by the Society for functional Near-Infrared Spectroscopy,
+             and then load it using :func:`mne.io.read_raw_snirf`.
 
+fNIRS measurements often have a non-standardised format that is not supported
+by MNE and cannot be converted easily into SNIRF. However if it is possible to
+store this data in a legacy csv or tsv format, we show here a way to load it
```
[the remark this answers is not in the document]
drammock: this suggestion leaves the previous sentence beginning dangling... be sure to remove the "however if it is possible to" part

drammock:
I'm not sure why CircleCI isn't running here.  The first thing I'd try is resovling the merge conflict on `latest.inc` and see if that triggers it to run.

larsoner:
This recently happened to @cbrnr , I think because he revoked some permissions for CircleCI on his user (?). You could look at your settings.

Also see if you can view https://app.circleci.com/pipelines/github/mne-tools/mne-python/7826/workflows/c965a22a-19cf-402a-b92f-633247c4b948/jobs/27243, I hope it asks you to log in or for permission, and then you can view it. This might fix things

HanBnrd:
Doc:
- [*make_standard_montage*](https://27704-1301584-gh.circle-artifacts.com/0/dev/generated/mne.channels.make_standard_montage.html#mne.channels.make_standard_montage)
- [Tutorial](https://27704-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/30_reading_fnirs_data.html#loading-legacy-data-in-csv-or-tsv-format)

HanBnrd:
Docs:
- [Tutorial](https://27707-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/30_reading_fnirs_data.html#loading-legacy-data-in-csv-or-tsv-format)
- [*make_standard_montage*](https://27707-1301584-gh.circle-artifacts.com/0/dev/generated/mne.channels.make_standard_montage.html#mne.channels.make_standard_montage)

tutorials/io/30_reading_fnirs_data.py
```diff
@@ -61,16 +62,146 @@
...
+fig = mne.viz.plot_alignment(
+    raw.info, trans=trans, subject='fsaverage', subjects_dir=subjects_dir,
+    surfaces=['brain', 'head'], coord_frame='mri', dig=True, show_axes=True,
+    fnirs=['channels', 'pairs', 'sources', 'detectors'], fig=fig)
+mne.viz.set_3d_view(figure=fig, azimuth=90, elevation=90, distance=0.5,
+                    focalpoint=(0., -0.01, 0.02))
+
+
+###############################################################################
+# Storing of optode locations
```
drammock, line 197: this whole section seems almost like a "note" that applies to all the other sections.  Maybe this info belongs at the very top of the tutorial, before the NIRx section?

tutorials/io/30_reading_fnirs_data.py
```diff
@@ -61,16 +62,146 @@
...
+
+# sphinx_gallery_thumbnail_number = 2
+
+###############################################################################
+# First, we generate an example CSV file which will then be loaded in to MNE.
+# This step would be skipped if you have actual data you wish to load.
+# We simulate 16 channels with 100 samples of data and save this to a file
+# called fnirs.csv.
+
+pd.DataFrame(np.random.normal(size=(16, 100))).to_csv("fnirs.csv")
```
drammock, line 95: I wonder if it is really necessary to write out the CSV and then re-load it?  Would it be just as good to say something like "here we'll make some fake data using `pd.DataFrame`; in practice you'd use `pd.read_csv` instead to load your file"?

drammock (commented):
this is looking very good @HanBnrd. Just 2 comments.

tutorials/io/30_reading_fnirs_data.py
rob-luke: I think that some users won't know how to read a csv. And this will demonstrate it.

drammock: fair enough.

larsoner:
@HanBnrd please click "Ready for review" if this is good to go, and let me know

It sounds like @rob-luke and @drammock are happy so if it is good to go I'll give it one more look and hopefully merge!

HanBnrd:
I just pushed one more commit implementing @drammock's [suggestion](https://github.com/mne-tools/mne-python/pull/9141#discussion_r618688787) and @rob-luke's [comment](https://github.com/mne-tools/mne-python/issues/9309#issuecomment-825214207).

Here is the new tutorial: https://27748-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/30_reading_fnirs_data.html

That's ready for review now @larsoner.

larsoner:
Great, thanks @HanBnrd @rob-luke !

HanBnrd:
Thanks everybody! 🎉

PR #9146, opened 2021-03-18
DOC: Add description about ´coord_frame´ parameter in mne.channels.make_dig_montage

### LINKED ISSUE #7093: DOC: creating custom montages without fiducials
Currently, it is possible to create custom montages with mne.channels.Montage. This now gives a
DeprecationWarning with the redirection to use DigMontage instead. 

As far as I understood it, this function requires reference points on the head, which I don't have. I only have standard coordinates for channels, for custom EEG caps. Below is the example use case which still works, but will not anymore in the next version. It would be great if this functionality would still be preserved. Or is it also currently possible with DigMontage somehow?

I also use the same type of workflow to create custom montages for working with ECoG grids, which also lack the reference points.

```python
import mne
import pandas as pd

df = pd.read_csv('channels.txt')
ch_names = df.name.to_list()
pos = df[['x','y','z']].values
montage = mne.channels.Montage(pos=pos, ch_names=ch_names, 
                               kind='motor-cap', selection=range(64))
montage.plot()
```
channel file: [channels.txt](https://github.com/mne-tools/mne-python/files/3871332/channels.txt)

Desired output:
Screenshot from 2019-11-20 12-50-04

catalinamagalvan:
#### Reference issue
Example: Fixes #7093.

Add notes about valid arguments for ´coord_frame´ parameter and its use for creating custom montages.

### DIFF
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -121,8 +121,8 @@ def make_dig_montage(ch_pos=None, nasion=None, lpa=None, rpa=None,
 
     Notes
     -----
-    Valid ``coord_frame`` arguments are 'meg','mri','mri_voxel','head',
-    'mri_tal','ras', 'fs_tal','ctf_head','ctf_meg','unknown'. For custom
+    Valid ``coord_frame`` arguments are 'meg', 'mri', 'mri_voxel', 'head',
+    'mri_tal', 'ras', 'fs_tal', 'ctf_head', 'ctf_meg', 'unknown'. For custom
     montages without fiducials this parameter has to be set to 'head'.
     """
     _validate_type(ch_pos, (dict, None), 'ch_pos')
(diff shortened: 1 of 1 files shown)


mne/channels/montage.py
```diff
@@ -118,6 +118,12 @@ def make_dig_montage(ch_pos=None, nasion=None, lpa=None, rpa=None,
     read_dig_egi
     read_dig_fif
     read_dig_polhemus_isotrak
+
+    Notes
+    -----
+    Valid ``coord_frame`` arguments are 'meg','mri','mri_voxel','head',
+    'mri_tal','ras', 'fs_tal','ctf_head','ctf_meg','unknown'. For custom
+    montages without fiducials this parameter has to be set to 'head'.
```
drammock, line 126:
```suggestion
    Valid ``coord_frame`` arguments are 'meg', 'mri', 'mri_voxel', 'head',
    'mri_tal', 'ras', 'fs_tal', 'ctf_head', 'ctf_meg', 'unknown'. For custom
    montages without fiducials this parameter has to be set to 'head'.
```

drammock (approved):
CI failures are unrelated to this PR; so merging. Thanks @catalinamagalvan!

PR #9149, opened 2021-03-18
Dark-mode to PyVista-Brain (pair-programmed with @GuillaumeFavelier)

marsipu:
This follows PR #9126.

Add a theme-parameter to Brain with PyVista-Renderer accepting 'dark', 'light' and 'auto'. Dark-Mode depends on [qdarkstyle](https://github.com/ColinDuquesnoy/QDarkStyleSheet) and the automatic detection works with [darkdetect](https://github.com/albertosottile/darkdetect).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -147,7 +147,6 @@ Enhancements
 
 - :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103`, :gh:`9175` by |Rotem Falach|_ and `Richard Höchenberger`_)
 
-
 - Add parameter ``theme`` to :class:`mne.viz.Brain` for optional Dark-Mode (:gh:`9149` by `Martin Schulz`_, `Guillaume Favelier`_)
 
 Bugs
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -148,6 +148,8 @@ Enhancements
 - :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103`, :gh:`9175` by |Rotem Falach|_ and `Richard Höchenberger`_)
 
 
+- Add parameter ``theme`` to :class:`mne.viz.Brain` for optional Dark-Mode (:gh:`9149` by `Martin Schulz`_, `Guillaume Favelier`_)
+
 Bugs
 ~~~~
 - Fix bug with :func:`mne.viz.plot_evoked_topo` where ``ylim`` was only being applied to the first channel in the dataset (:gh:`9162` **by new contributor** |Ram Pari|_ )
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -149,6 +149,8 @@ intersphinx_mapping = {
     'dipy': ('https://dipy.org/documentation/1.2.0.', None),
     'mne_realtime': ('https://mne.tools/mne-realtime', None),
     'picard': ('https://pierreablin.github.io/picard/', None),
+    'qdarkstyle': ('https://github.com/ColinDuquesnoy/QDarkStyleSheet', None),
+    'darkdetect': ('https://github.com/albertosottile/darkdetect', None)
 }
 
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -149,8 +149,7 @@ intersphinx_mapping = {
     'dipy': ('https://dipy.org/documentation/1.2.0.', None),
     'mne_realtime': ('https://mne.tools/mne-realtime', None),
     'picard': ('https://pierreablin.github.io/picard/', None),
-    'qdarkstyle': ('https://github.com/ColinDuquesnoy/QDarkStyleSheet', None),
-    'darkdetect': ('https://github.com/albertosottile/darkdetect', None)
+    'qdarkstyle': ('https://qdarkstylesheet.readthedocs.io/en/latest', None)
 }
 
 
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -254,6 +254,11 @@ class Brain(object):
         Color of the foreground (will be used for colorbars and text).
         None (default) will use black or white depending on the value
         of ``background``.
+    theme : str | path-like
+        Can be "auto" (default), "light", or "dark" to use darkdetect
+        (required for auto mode) or qdarkstyle (required for dark mode)
+        to style the widgets. Automatic detection does not yet work on
+        Linux. Can also be a path-like to a custom style sheet.
     figure : list of Figure | None | int
         If None (default), a new window will be created with the appropriate
         views. For single view plots, the figure can be specified as int to
@@ -365,7 +370,7 @@ class Brain(object):
 
     def __init__(self, subject_id, hemi, surf, title=None,
                  cortex="classic", alpha=1.0, size=800, background="black",
-                 foreground=None, figure=None, subjects_dir=None,
+                 foreground=None, theme='auto', figure=None, subjects_dir=None,
                  views='auto', offset='auto', show_toolbar=False,
                  offscreen=False, interaction='trackball', units='mm',
                  view_layout='vertical', silhouette=False, show=True):
@@ -472,6 +477,7 @@ class Brain(object):
                                        fig=figure)
 
         self._renderer._window_initialize(self._clean)
+        self._renderer._window_set_theme(theme)
         self.plotter = self._renderer.plotter
 
         self._setup_canonical_rotation()
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -291,10 +291,9 @@ class Brain(object):
        brain's silhouette to display. If True, the default values are used
        and if False, no silhouette will be displayed. Defaults to False.
     theme : str | path-like
-        Can be "auto" (default), "light", or "dark" to use darkdetect
-        (required for auto mode) or qdarkstyle (required for dark mode)
-        to style the widgets. Automatic detection does not yet work on
-        Linux. Can also be a path-like to a custom style sheet.
+        Can be "auto" (default), "light", or "dark" or a path-like to a
+        custom stylesheet. For Dark-Mode and automatic Dark-Mode-Detection,
+        `qdarkstyle` respectively `darkdetect` is required.
     show : bool
         Display the window as soon as it is ready. Defaults to True.
 
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -293,7 +293,8 @@ class Brain(object):
     theme : str | path-like
         Can be "auto" (default), "light", or "dark" or a path-like to a
         custom stylesheet. For Dark-Mode and automatic Dark-Mode-Detection,
-        :mod:`qdarkstyle` respectively :mod:`darkdetect` is required.
+        :mod:`qdarkstyle` respectively and `darkdetect
+        <https://github.com/albertosottile/darkdetect>`__ is required.
     show : bool
         Display the window as soon as it is ready. Defaults to True.
 
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -422,6 +422,8 @@ class Brain(object):
         self._size = size if len(size) == 2 else size * 2  # 1-tuple to 2-tuple
         subjects_dir = get_subjects_dir(subjects_dir)
 
+        self.theme = theme
+
         self.time_viewer = False
         self._hemi = hemi
         self._units = units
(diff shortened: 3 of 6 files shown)


mne/viz/_brain/_brain.py
```diff
@@ -254,6 +254,11 @@ class Brain(object):
         Color of the foreground (will be used for colorbars and text).
         None (default) will use black or white depending on the value
         of ``background``.
+    theme : str | path-like
+        Can be "auto" (default), "light", or "dark" to use darkdetect
+        (required for auto mode) or qdarkstyle (required for dark mode)
+        to style the widgets. Automatic detection does not yet work on
+        Linux. Can also be a path-like to a custom style sheet.
```
larsoner, line 261: I would open an upstream issue with `darkdetect` for this if they don't have one. In the meantime I would use:
```
gsettings get org.gnome.desktop.interface gtk-theme
```
if it ends in `-dark` assume it's dark. It at least [works on my machine](https://blog.codinghorror.com/the-works-on-my-machine-certification-program/)™ :)

mne/viz/_brain/_brain.py
```diff
@@ -365,7 +370,7 @@ class Brain(object):
 
     def __init__(self, subject_id, hemi, surf, title=None,
                  cortex="classic", alpha=1.0, size=800, background="black",
-                 foreground=None, figure=None, subjects_dir=None,
+                 foreground=None, theme='auto', figure=None, subjects_dir=None,
```
larsoner, line 373: Let's put it at the end, or maybe at least just before `show=True`

mne/viz/backends/_qt.py
```diff
@@ -435,6 +436,24 @@ def _window_ensure_minimum_sizes(self, sz):
             self._process_events()
             self._process_events()
 
+    def _window_set_theme(self, theme):
+        if theme == 'auto':
+            import darkdetect
```
larsoner, line 441: Make sure optional dependencies stay optional, e.g.
```suggestion
            try:
                import darkdetect
            except Exception:
                detected_theme = None
            else:
                detected_theme = darkdetect.theme()
            theme = 'light' if detected_theme is None else deteced_theme.lower()
```
Something similar should be done for `qdarkstyle`

mne/viz/_brain/_brain.py
marsipu: @cbrnr opened an issue

GuillaumeFavelier: This one: https://github.com/albertosottile/darkdetect/issues/7

GuillaumeFavelier:
@marsipu I pushed a commit to reorder `theme` and take https://github.com/mne-tools/mne-python/pull/9149#discussion_r596966663 into account

marsipu:
> 
> 
> @marsipu I pushed a commit to reorder `theme` and take [#9149 (comment)](https://github.com/mne-tools/mne-python/pull/9149#discussion_r596966663) into account

Nice @GuillaumeFavelier . I also thought we maybe could shorten the documentation-text a bit more like this?

GuillaumeFavelier:
> we maybe could shorten the documentation-text a bit more like this?

Sounds good :+1:

mne/viz/_brain/_brain.py
```diff
@@ -290,6 +290,10 @@ class Brain(object):
        and ``decimate`` (level of decimation between 0 and 1 or None) of the
        brain's silhouette to display. If True, the default values are used
        and if False, no silhouette will be displayed. Defaults to False.
+    theme : str | path-like
+        Can be "auto" (default), "light", or "dark" or a path-like to a
+        custom stylesheet. For Dark-Mode and automatic Dark-Mode-Detection,
+        `qdarkstyle` respectively `darkdetect` is required.
```
GuillaumeFavelier, line 296: CircleCI seems to have some issues with the new modules:

```
/home/circleci/project/mne/viz/_brain/_brain.py:docstring of mne.viz._brain._brain.Brain:97: WARNING: py:obj reference target not found: qdarkstyle
/home/circleci/project/mne/viz/_brain/_brain.py:docstring of mne.viz._brain._brain.Brain:97: WARNING: py:obj reference target not found: darkdetect
```

Probably the docstring needs to be updated:


```suggestion
        :mod:`qdarkstyle` respectively :mod:`darkdetect` is required.
```

And the list `intersphinx_mapping` in `doc/conf.py` too.

marsipu:
Somehow when I test it locally, the :mod:-phrases doesn't turn into a link leading to the websites. I can't find the issue

larsoner:
The `:mod:` should not be necessary I think. The problem is probably more that these modules are not in our `intersphinx` inventory in `doc/conf.py`.

https://qdarkstylesheet.readthedocs.io/en/latest/reference/qdarkstyle.html

I don't see a Sphinx documentation page for `darkdetect` so I would just make that one code with double-backticks rather than try to link.

larsoner:
(or you could do a custom HTML link to the darkdetect pypi or GitHub page if you want)

doc/conf.py
```diff
@@ -149,6 +149,8 @@
     'dipy': ('https://dipy.org/documentation/1.2.0.', None),
     'mne_realtime': ('https://mne.tools/mne-realtime', None),
     'picard': ('https://pierreablin.github.io/picard/', None),
+    'qdarkstyle': ('https://github.com/ColinDuquesnoy/QDarkStyleSheet', None),
```
larsoner, line 152: I think these need to point to and `objects.inv` created by Sphinx, so I think this URL needs to exist:

But this one does:

https://qdarkstylesheet.readthedocs.io/en/latest/objects.inv

```suggestion
    'qdarkstyle': ('https://qdarkstylesheet.readthedocs.io/en/latest', None),
```
I don't think `darkdetect` has a Sphinx-built site so I don't think intersphinx will work, I would just use an anonymous RST URL for that...

mne/viz/_brain/_brain.py
```diff
@@ -290,6 +290,10 @@ class Brain(object):
        and ``decimate`` (level of decimation between 0 and 1 or None) of the
        brain's silhouette to display. If True, the default values are used
        and if False, no silhouette will be displayed. Defaults to False.
+    theme : str | path-like
+        Can be "auto" (default), "light", or "dark" or a path-like to a
+        custom stylesheet. For Dark-Mode and automatic Dark-Mode-Detection,
+        :mod:`qdarkstyle` respectively :mod:`darkdetect` is required.
```
larsoner, line 296:
```suggestion
        :mod:`qdarkstyle` respectively and
        `darkdetect <https://github.com/albertosottile/darkdetect>`__
        is required.
```

doc/conf.py
marsipu: Thank you

marsipu:
The failed test in `mne/decoding/tests/test_transformer.py::test_scaler` is probably unrelated, right?

GuillaumeFavelier:
This is what I saw:

```
=================================== FAILURES ===================================
_________________________________ test_scaler __________________________________
mne/decoding/tests/test_transformer.py:67: in test_scaler
    assert_allclose(X * stds[:, np.newaxis] + means[:, np.newaxis],
E   AssertionError: 
E   Not equal to tolerance rtol=1e-12, atol=1e-20
E   mean
E   Mismatched elements: 23576 / 23576 (100%)
E   Max absolute difference: 2.00587641e-11
E   Max relative difference: 42.17865626
E    x: array([[[-1.005809e-13, -1.005809e-13, -1.005809e-13, ...,
E            -1.005809e-13, -1.005809e-13, -1.005809e-13],
E           [ 2.151313e-13,  2.151313e-13,  2.151313e-13, ...,...
E    y: array([[[ 2.454723e-12,  1.490368e-12, -4.383434e-13, ...,
E            -1.402699e-12, -2.367054e-12, -4.383434e-13],
E           [-1.681645e-12, -8.432133e-12, -6.503422e-12, ...,...
```

I don't think it's related and I don't see it happening on `master`(82b3459a35c1c61c3c2d1ab724aaf9ee831d791d). Maybe you can rebase @marsipu ?

GuillaumeFavelier (approved):
I tried it locally and it works as expected, I'm +1 for merge once CIs come back happy :+1: 

light | dark | custom™
--|--|--
image | image | image

drammock:
> This is what I saw:

I've seen that on another PR too, agree it's unrelated.

larsoner:
Indeed on https://github.com/albertosottile/darkdetect/pull/8 and this branch after installing `qdarkstyle` I get:

Screenshot from 2021-03-19 13-28-55

mne/viz/backends/_qt.py
```diff
@@ -435,6 +436,40 @@ def _window_ensure_minimum_sizes(self, sz):
...
+        if theme == 'dark':
+            try:
+                import qdarkstyle
+            except ModuleNotFoundError:
+                logger.info('For Dark-Mode "qdarkstyle" has to be installed! '
+                            'You can install it with `pip install qdarkstyle`')
+                stylesheet = None
+            else:
+                stylesheet = qdarkstyle.load_stylesheet()
+        elif isfile(theme):
```
larsoner, line 465: I would rather have this be `elif theme != 'light'`. That way someone passing `theme='my_themem.css'` will not just get a light theme by default, they will get a FileNotFoundError because they passed a typo `my_themem.css` instead of `my_theme.css` or whatever.

The only downside is that someone cannot use a file named `'light'` or `'dark'` which seems fine

marsipu: Ok

hoechenberger:
CIs are green – @marsipu @GuillaumeFavelier is this good to merge?

hoechenberger (approved):

marsipu:
We also talked about inverting the icon-pixmaps with the theme once and haven't done it yet. Should we add this to this PR too?

larsoner:
Yeah I think that's reasonable assuming it's not too difficult

doc/changes/latest.inc
```diff
@@ -148,6 +148,8 @@ Enhancements
 - :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103`, :gh:`9175` by |Rotem Falach|_ and `Richard Höchenberger`_)
 
 
+- Add parameter ``theme`` to :class:`mne.viz.Brain` for optional Dark-Mode (:gh:`9149` by `Martin Schulz`_, `Guillaume Favelier`_)
```
larsoner, line 151: There is an extra newline above that should not be there

mne/viz/backends/_qt.py
```diff
@@ -466,6 +498,25 @@ class _Renderer(_PyVistaRenderer, _QtDock, _QtToolBar, _QtMenuBar,
...
+                    ' `pip install qdarkdetect`')
+        theme = 'light'
+    else:
+        detected_theme = darkdetect.theme()
+        if detected_theme is not None:
+            theme = detected_theme.lower()
+        else:
+            theme = 'light'
+
+    return theme
```
larsoner, line 517: One problem with this flow is that, because `'auto'` is the default (which I think is the right choice), anybody without `darkdetect` who uses the `Brain` class  / `stc.plot` for anything without explicitly setting `theme` will get this `logger.info` message saying that no dark mode detection can be done. I'd rather make things simpler and do:
```suggestion
def _detect_theme():
    try:
        import darkdetect
        return darkdetect.theme().lower()
    except Exception:
        return 'light'
```
We already document in the parameter that they should have `darkdetect` for things to work properly so I don't think we need to `logger` them about it. This code is also robust to any issue with `darkdetect` we might encounter (at the cost of failing silently when it doesn't work, which I think is an acceptable tradeoff for non-essential aesthetic UI functionality)

larsoner (commented):
Can you add a `brain_kwargs=None` to `SourceEstimate.plot` and `VectorSourceEstimate.plot`, and `VolSourceEstimate.plot_3d`? See for example how we handle `add_data_kwargs`, the difference being that `brain_kwargs` should be passed to the Brain constructor.

marsipu:
> 
> 
> Can you add a `brain_kwargs=None` to `SourceEstimate.plot` and `VectorSourceEstimate.plot`, and `VolSourceEstimate.plot_3d`? See for example how we handle `add_data_kwargs`, the difference being that `brain_kwargs` should be passed to the Brain constructor.

I am sorry, I can't follow you: I already see the parameter `brain_kwargs` with `default=None` documented and established to pass kwargs onto the renderer (mne.viz._3d.py::1846 and mne.viz._3d.py::1857) for all mentioned plot-methods. Or what do you mean?

larsoner:
Oh I didn't realize we already had that param, never mind!

larsoner:
Thanks @marsipu @GuillaumeFavelier !

agramfort:
😎

hoechenberger:
super good!

GuillaumeFavelier:
Awesome work @marsipu :ok_hand:

PR #9150, opened 2021-03-18
[BUG] Tfr misc issues

### LINKED ISSUE #8013: BUG: TFR.plot misc issues

eort:
#### Reference issue
Example: Fixes #8013 .

- function returns a list of figures rather than a single figure, for every pick that is being plotted
- the title is set for each figure separately, depending on the arguments of `title` and `combine` in following way:
    - if `title` is `None` or a `String`, set the title of every figure accordingly
    - if `title` is `'auto'` and `combine` is `None`, set the title of every figure to the channel name that is being plotted
    - if `title` is `'auto'` and `combine` is `not None`, set the title to `Avg/Rms` of  <number of picks> sensors
- the documentation now better reflects what is actually done in the function
- fixing bugs related to `plot()`, triggered a bug in `plot_joint()`, or actually the test script of `plot_joint()`. The doc of `plot_joint()` specify only `mean` and `rms` being valid keywords for `combine`, however, in https://github.com/mne-tools/mne-python/blob/f20c60ec23481e8c9888fb9f7da1544378dd2489/mne/time_frequency/tests/test_tfr.py#L551
the function is also tested for `combine is None`. I removed that particular test situation
- some minor issues in the doc string were fixed

#### What else could be done? 
-  When using a `mask`, `mask_cmap` is currently set to `gray`. Perhaps that is just personal style preference, but I find it extremely not pretty. My suggestions would be to use the same colormap as the actual plot is using. Especially if the mask is shown with transparency and not contour
- passing a list of strings as title
- the way the `default` value is specified in the docstring varies across keywords. Is that something that should be unified?
- specify the output as either a list of figures or a figure (if it is only one)
- with default for picks=None, a lot figures will be opened (200+ for only gradiometers in the data). Perhaps there is a warning or something that can be raised before it happens? (I just almost killed my laptop because of it ;) )
- a more elaborate unit test?
 
#### Unrelated issues 
- the `contour` value to `mask_style` does not seem to work
- the colorbar units are missing
- https://github.com/mne-tools/mne-python/blob/a6abf3649bdf754c86bb983f393f690791816048/mne/viz/utils.py#L1819
this function is a bit weird that it only works if the title argument is not None. I'd think it would make more sense to remove the `title` keyword and just assume that if you call that  function, you'll get an output as if `title` was `None`. 

I hope I haven't forgotten anything important :)


cbrnr:
I think returning a list of figures is counter-intuitive. Users will expect the `plot` method to actually produce a plot (just like all other MNE functions). Therefore, returning just one figure with multiple axes is probably the better solution.

eort:
That was also my first hunch, but things get quite convoluted when trying to make reasonably pretty figures. I talked to @sappelhoff about it, and we agreed that it might be easiest to let user take care of subplots in a single figure by using the `axes` keyword. 
I mean, it certainly is possible to swap out figures for subplots, but as far as I can see, it will take quite a bit of digging, and i am not sure the potential gain is worth the effort.

mmagnuski:
@cbrnr 
> Therefore, returning just one figure with multiple axes is probably the better solution.

Returning multpile figures when multiple channels are selected is used frequently in mne (for example in `plot_properties` or the function that produces epoch images IIRC) and is much simpler. One multi-axes figure can be a possible future improvement to keep in mind.

mne/time_frequency/tests/test_tfr.py
```diff
@@ -548,7 +549,7 @@ def test_plot_joint():
 
     topomap_args = {'res': 8, 'contours': 0, 'sensors': False}
 
-    for combine in ('mean', 'rms', None):
+    for combine in ('mean', 'rms'):
```
mmagnuski, line 552: combine is None by default, so we should test this value too.

mmagnuski: The docs say " If None, plot one figure per selected channel." That is the intended behavior, although as you describe, not what the function actually does.

mne/time_frequency/tfr.py
```diff
@@ -1250,34 +1254,44 @@ def _plot(self, picks=None, baseline=None, mode='mean', tmin=None,
...
-            else:
-                ax = axes[idx]
-                fig = ax.get_figure()
+
+        # make sure there are as many axes as there will be channels to plot
+        if isinstance(axes, list) or isinstance(axes, np.ndarray):
+            figs_and_axes = [(ax.get_figure(), ax) for ax in axes]
+        elif isinstance(axes, plt.Axes):
+            figs_and_axes = [(ax.get_figure(), ax) for ax in [axes]]
+        elif axes is None:
```
mmagnuski, line 1282: if `axes` is None and `combine` is not None then there should be only one figure

mne/time_frequency/tfr.py
```diff
@@ -1289,22 +1303,22 @@ def _plot(self, picks=None, baseline=None, mode='mean', tmin=None,
                 yscale=yscale, mask=mask, mask_style=mask_style,
                 mask_cmap=mask_cmap, mask_alpha=mask_alpha)
 
-            if title is None:
-                if combine is None or len(tfr.info['ch_names']) == 1:
-                    title = tfr.info['ch_names'][0]
+            if title == 'auto':
+                if len(tfr.info['ch_names']) == 1:
```
mmagnuski, line 1307: why not leave the `combine is None or ` in the if statement? Now do the same in the else part checking for `if combine is None`.

mne/time_frequency/tfr.py
```diff
@@ -1289,22 +1303,22 @@ def _plot(self, picks=None, baseline=None, mode='mean', tmin=None,
...
+                subtitle = title
+            fig.suptitle(subtitle)
 
-            plt_show(show)
-            # XXX This is inside the loop, guaranteeing a single iter!
-            # Also there is no None-contingent behavior here so the docstring
-            # was wrong (saying it would be collapsed)
-            return fig
+        plt_show(show)
+        return [fig for (fig, ax) in figs_and_axes]
```
mmagnuski, line 1321: You've created 'figs` list earlier on - you can return it here.

mne/time_frequency/tests/test_tfr.py
eort: Are you sure? This looks like `mean` is the default and further down the docs it lists only `mean` and `rms` as possible values. https://github.com/mne-tools/mne-python/blob/f20c60ec23481e8c9888fb9f7da1544378dd2489/mne/time_frequency/tfr.py#L1314

mne/time_frequency/tfr.py
eort: Yes, that would happen here as well. Earlier (new line 1256), `n_picks` is set to `1` if `combine` is not `None`. So that here only a single figure would be created.

mne/time_frequency/tfr.py
eort: Good point. I'll change that back!

mne/time_frequency/tfr.py
eort: Actually, earlier I created a list of (fig, axes) tuples to simplify the the indexing in the loop. So, at the return, I select only the figs, and drop the axes

cbrnr:
I'm fine with returning a list of figures as long as all figures are also displayed.

mmagnuski:
@cbrnr by default they should be, this is how it works in the functions I referred to.

mne/time_frequency/tfr.py
mmagnuski: ok!

mne/time_frequency/tests/test_tfr.py
mmagnuski: Ah, yeas, you are right, I was looking at `tfr.plot()`, not `plot_joint()`

mne/time_frequency/tfr.py
mmagnuski: Right, the `figs` list is only present in one branch.

mmagnuski: The only issue might be that now you always return list of figures, while previously a single figure was returned, so the change may break code that relies on the previous behavior (plotting one channel for example). But its easy to fix so I think it should be ok.

larsoner (commented):
@eort can you add a line to the `BUG` section mentioning this fix, and that a list of figures is now always returned instead of a single figure handle? Otherwise LGTM +1 for merge!

larsoner:
@mmagnuski are you happy here?

mmagnuski:
I am good generally. Although codecov claims that 75% of the diff is covered, so it might be worthwhile to add a test here (for example for the tfr.plot() figure / axis behavior, how titles are created and for the returned list of fig length).

eort:
> can you add a line to the BUG section mentioning this fix, and that a list of figures is now always returned instead of a single figure handle? 

will do.

> so it might be worthwhile to add a test here

I'll have a look whether some more settings can be tested!

larsoner (approved):
LGTM +1 for merge, thanks @eort !

mne/time_frequency/tests/test_tfr.py
```diff
@@ -521,6 +521,7 @@ def test_init_EpochsTFR():
 
 
 def test_plot():
+
```
larsoner, line 524:
```
mne/time_frequency/tests/test_tfr.py:525 in public function `test_plot`:
        D201: No blank lines allowed before function docstring (found 1)
```

eort:
Not sure what exactly is the problem with those older (python?) versions, but it seems like extracting the title text from the figure works differently across versions. 

An alternative to how it is done now, we could use this here: `figs[0]._suptitle().get_text()`, which also works on my system. But it has the drawback of using private functions of matplotlib.

larsoner:
If it's just for the test I would do:
```
try:
    title = figs[0].get_figure().texts[0].get_text()
except Exception:  # old matplotlib version
    title = figs[0]._suptitle().get_text()
assert title == 'title'
```

eort:
Ok. Is there a way to test that without having to push to this PR?

larsoner:
You could install locally an env with an older version of matplotlib. If you want to see how the CIs do it, it's here:

https://github.com/mne-tools/mne-python/blob/main/.github/workflows/compat_old.yml#L36

So something like this should work:
```
conda create -n old numpy=1.15.4 scipy=1.1 matplotlib=3.0.3 pandas=0.23.4 scikit-learn=0.20.2
conda activate old
pip install -r requirements_testing.txt
pytest mne/time_frequency/tests/test_tfr.py -k plot
```

eort:
turns out, the old-compatible version works also for the recent tests. 

Oh, and sorry for the messing up the commit! I think I included previous changes that I didn't have locally yet into the commit

larsoner:
> Oh, and sorry for the messing up the commit! I think I included previous changes that I didn't have locally yet into the commit. I think d0ed820 is all you need.

What's important is that the diff on GitHub looks correct. Does it?

If not, you can push more commits to fix it. Or if you're certain you don't want your last two commits, you could do while on this branch:
```
git reset --hard d0ed820
git push origin -f tfr_misc_issues
```

larsoner (approved):
LGTM, will merge once CIs come back happy, thanks @eort !

PR #9151, opened 2021-03-18
Fix mne.viz.plot_compare_evokeds did not check type of combine

matteoanelli:
It adds the type check of the `combine` parameter and integrates the respective test. 

#### Additional information
Before this fix, the user could pass, for example, a list. Hence, creating unexpected behavior from the plot.

To point out that the documentation correctly specifies that combine must be None, str, or callable(). However, while I was experimenting with this function I found out that there were no specific checks implemented.


agramfort (commented):
can you also add a what's new entry?

thx heaps !

mne/viz/evoked.py
```diff
@@ -2225,6 +2225,9 @@ def plot_compare_evokeds(evokeds, picks=None, colors=None,
     if show_sensors is None:
         show_sensors = (len(picks) == 1)
 
+    if not callable(combine):
+        _validate_type(combine, (None, str), 'combine',
+                       'None, str or callable')
```
hoechenberger, line 2230: Don't we need to test / allow for `callable` here, too?

hoechenberger: @larsoner How would one check for `callable` using `_validate_type`?

hoechenberger: @matteoanelli I just talk to @larsoner and we'll most probably want to enhance `_validate_type()` to accept `callable`. We should do that in a separate PR I would suggest.

Let me know if/when you're available and we can look into this together. otherwise I can start working on it on my own. Thanks!

hoechenberger: Oh it actually seems that adding the string `'callable'` might be sufficient! I will test this now.

mne/viz/tests/test_evoked.py
```diff
@@ -385,6 +385,11 @@ def test_plot_compare_evokeds():
     plot_compare_evokeds(evoked_dict, cmap=('cmap title', 'inferno'),
                          linestyles=['-', ':', '--'])
     plt.close('all')
+    # test combine
+    match = 'must be an instance of None, callable, or str'
```
rob-luke, line 389: Can the test check for the item_name that you set to combine too?

mne/viz/evoked.py
matteoanelli: I tested the function and it looks like it is working with 'callable' as a string. However, it may be a good thing to add an explanation of what 'item_name', 'type_name' do inside '_validate_type()'.

mne/viz/tests/test_evoked.py
sappelhoff: @matteoanelli and I don't understand what you mean by that @rob-luke - can you clarify, please? :-)

matteoanelli:
I updated the mne.viiz.plot_compare_evokeds as you suggested. I tested using 'callable' as str. It is properly working in my opinion. I had to do a minor change from your proposed solution to not check if callable() twice.

In addition in mne.utils.check._validate_type I added 'callable' to the possible values. In addition, I added a brief description of what 'item_name', 'type_name' do.

Finally, I propose I minor change in the output messages. 
Before, for example, with _validate_type("ciao", types=int) it would have given the following message "None must be an instance of.."  since 'item_name'=None. I think it can cause some misunderstanding to the function users.
In what I propose it would be "Item must be an instance of.." if 'item_name'=None.

drammock:
CI failures are unrelated:
```
_________________________________ test_scaler _________________________________
mne\decoding\tests\test_transformer.py:67: in test_scaler
    assert_allclose(X * stds[:, np.newaxis] + means[:, np.newaxis],
E   AssertionError: 
E   Not equal to tolerance rtol=1e-12, atol=1e-20
E   mean
E   Mismatched elements: 23576 / 23576 (100%)
E   Max absolute difference: 2.00587641e-11
E   Max relative difference: 42.17865626
E    x: array([[[-1.005809e-13, -1.005809e-13, -1.005809e-13, ...,
E            -1.005809e-13, -1.005809e-13, -1.005809e-13],
E           [ 2.151313e-13,  2.151313e-13,  2.151313e-13, ...,...
E    y: array([[[ 2.454723e-12,  1.490368e-12, -4.383434e-13, ...,
E            -1.402699e-12, -2.367054e-12, -4.383434e-13],
E           [-1.681645e-12, -8.432133e-12, -6.503422e-12, ...,...
```

mne/viz/tests/test_evoked.py
drammock: I think the idea is that the `match` string should be checking for `'combine'` as well, e.g.;
```suggestion
    match = 'combine must be an instance of None, callable, or str'
```

mne/utils/check.py
```diff
@@ -388,7 +394,8 @@ def _validate_type(item, types=None, item_name=None, type_name=None):
                 type_name[-1] = 'or ' + type_name[-1]
                 type_name = ', '.join(type_name)
         raise TypeError('%s must be an instance of %s, got %s instead'
-                        % (item_name, type_name, type(item),))
+                        % ('Item' if item_name is None else item_name,
+                           type_name, type(item),))
```
drammock, line 398: can we make this an f-string? something like
```py
        _item_name = 'Item' if item_name is None else item_name
        raise TypeError(f"{_item_name} must be an instance of {type_name}, "
                        f"got {type(item)} instead")
```

hoechenberger:
Hello, what's the status here? 😊

matteoanelli:
There was a conflict in the  doc/changes/latest.inc, now everything should be ok.

hoechenberger (approved):

hoechenberger:
Thanks, @matteoanelli!!

PR #9156, opened 2021-03-18
MRG: Add tutorial for make_fixed_length_epochs function

nordme:
This PR adds a tutorial for calling mne.make_fixed_length_epochs().

The first draft of this tutorial covers calling the function, warns the reader against using for event-related analysis, and finishes with a brief demonstration of fixed length epoch use, in this case using fixed epochs as a basis for examining sensor space connectivity.


tutorials/epochs/make_fixed_length_epochs.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
```
larsoner, line 1: You need to name the file `plot_*` to get it to be rendered by CircleCI:

https://26238-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/make_fixed_length_epochs.html#sphx-glr-auto-tutorials-epochs-make-fixed-length-epochs-py

Looks already quite polished so it will be nice to see the CircleCI result!

tutorials/epochs/make_fixed_length_epochs.py
```diff
@@ -0,0 +1,135 @@
...
+the signal.
+"""
+
+import os
+import mne
+import matplotlib.pyplot as plt
+from mne.preprocessing import compute_proj_ecg
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+import numpy as np
+
```
agramfort, line 25:
```suggestion
import os
import numpy as np
import mne
from mne.preprocessing import compute_proj_ecg
import matplotlib.pyplot as plt
from mpl_toolkits.axes_grid1 import make_axes_locatable

```

nordme: Thanks, @larsoner and @agramfort for the reviews! Alex, what is the significance of the import reordering?

agramfort: it should reflect the hierarchy of dependencies between packages. First you put the packages from the standard library, then numpy, then matplotlib as it depends on numpy, then mne as it depends on the above.

it's not necessary but it's useful convention.

hoechenberger (approved):

hoechenberger:
LGTM once we're green

PR #9160, opened 2021-03-19
[DOC] Stc near sensors doc

### LINKED ISSUE #8479: stc_near_sensors mode has incongruent docstring
#### Proposed documentation ehancement
I think(?) `stc_near_sensors` only enables nearest, single and sum, but it also says there is a "zero" and "weighted" option? 

Once it's cleared up what sort of activation `mode`s are available and when, I can make a PR to fix up the docstring.

Reference: https://mne.tools/dev/generated/mne.stc_near_sensors.html?highlight=stc_near#mne.stc_near_sensors

eort:
#### Reference issue
Closes #8479.

- remove zero as value for `mode`
- add single as value for `mode`
- in Notes, change `weighted` to `single`

#### Additional information
I find it a little confusing that `sum` is a valid value for that keyword, but does not appear in the code. I guess it is just the default behavior? But the default value is not specified.

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3227,7 +3227,8 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     mode : str
         Can be "sum" to do a linear sum of weights, "nearest" to
         use only the weight of the nearest sensor, or "single" to
-        do a distance-weight of the nearest sensor. See Notes.
+        do a distance-weight of the nearest sensor. Default is "sum".
+        See Notes.
     project : bool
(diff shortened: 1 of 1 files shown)


agramfort (approved):

sappelhoff (approved):
> I find it a little confusing that sum is a valid value for that keyword, but does not appear in the code. I guess it is just the default behavior? But the default value is not specified.

I also guess it's the default behavior ... and then "single" and "nearest" add something to that (or don't).

you could add to the docstr that "sum" is the default for `mode`, but either way, this looks fine.

drammock:
thanks @eort!

PR #9163, opened 2021-03-19
[MRG, BUG, ENH] Add axes for plot mne.channels.layout

alexrockhill:
Add axes plot to Layout, see example:

```
import numpy as np
import mne

box = (0, 100, 0, 100)
pos = np.random.random((10, 4))
layout = mne.channels.Layout(box, pos, ['test'] * 10, range(10), 'test')
layout.plot(show_axes=True)
```


agramfort (approved):

larsoner:
Thanks @alexrockhill !

PR #9165, opened 2021-03-19
MAINT: Move function

larsoner:
Requested by @agramfort in https://github.com/mne-tools/mne-python/pull/9118#discussion_r596933282. It's purely a code move.

Unfortunately `mne.coreg.get_mni_fiducials` is public in the `mne.coreg` namespace:

https://mne.tools/dev/generated/mne.coreg.get_mni_fiducials.html

I'm not sure it's worth it to move its API page to a different namespace, but we could. @agramfort do you prefer we leave it, or move it to `mne.` or `mne.source_space.` or somewhere else?

### DIFF
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -20,6 +20,7 @@ import re
 import numpy as np
 
 from ..defaults import HEAD_SIZE_DEFAULT
+from ..source_space import get_mni_fiducials
 from ..viz import plot_montage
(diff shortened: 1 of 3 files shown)


agramfort (commented):
git grep 

tell me that mne.coreg.get_mni_fiducials is still used in a few places

larsoner:
From talking to @agramfort I think he agreed that it's okay to leave the other instances since they are part of the existing public API, so this PR just helps remove some nested imports at least

agramfort (approved):

PR #9166, opened 2021-03-19
MRG, FIX, ENH: First_samp support and simulation tests

### LINKED ISSUE #9120: ENH, DOC:  First_samp support for SourceSimulator class
Currently, simulating raw data with a `SourceSimulator `object (in _simulation/raw.py : simulate_raw()_) causes extra time points at the beginning if the source events are offset in time, even when a `first_samp` argument is included to nominally account for that offset.  This could happen, for example, if an experimental raw object and associated event list are used to instantiate _simulate_raw()_.  Effectively, the `first_samp` argument only changes the simulated raw object's absolute timing attributes (.first_samp, .last_samp, .first_time, etc), but doesn't affect how sources are actually added to it.  This means that if the first event has an arbitrarily large time stamp, then the simulator object (and therefore the resulting raw object) will be arbitarily long.

The documentation should also be updated.  For example, presently there is no indication that a `SourceSimulator `object is a valid argument for the _simulate_raw()_ constructor.

I propose to add `first_samp `support to the `SourceSimulator `class, so that event waveforms added to an instant of the class (via _.add_data(_)) are appropriately time stamped.  This can be implemented by simply subtracting the `first_samp ` value from the integer event times (which are expressed as absolute samples) when adding event data.  The `first_samp `value would be passed as an argument at the time the source object is instantiated; it would default to 0 (which would not influence existing API usage) and event times less than the value would cause an error.

I am inclined to assert that the `first_samp` argument for _simulate_raw()_ be the same as the `first_samp` attribute of the simulator object.  In any case, a suitable test function should make sure that added events in the raw object occur at the intended time samples.  (Presently there is no test for using source simulators with _simulate_raw()_.)

The alternative way to simulate raw data with _simulate_raw()_ is with a `SourceEstimate `object or an iterable of `SourceEstimate `objects.  The `first_samp` argument shouldn't affect how the raw data is created with this approach, as the STCs are simply concatenated in time.

Rather than changing the SS class definition, MNE users could be made aware of the current `first_samp` limitations via an instructive example.

#### Additional comments
With @cmista for Code Sprint 2021; I have mentioned the proposed change to @larsoner

NeuroLaunch:
#### Reference issue
Fixes #9120.

Modified _mne.simulation.raw.simulate_raw()_ and _mne.simulation.source.SourceSimulator_ to properly interpret `first_samp `parameter.  Adding events to a `SourceSimulator` would not account for any sample offsets, such that the resulting raw data created with _simulate_raw()_ would have shifted event-aligned waveforms regardless of the `first_samp `argument.

#### Additional information
In order to properly effect the sample offset change, I reinterpreted the `duration` argument  of the `SourceSimulator `constructor to more intuitively match the first and last sample indices.

This PR also corrects a small inconsistency with HPI coils whereby an arbitrary entry was made if the first HPI data point wasn't at time = 0.


mne/simulation/source.py
```diff
@@ -336,28 +339,43 @@ class SourceSimulator(object):
...
-        return int(self.duration / self._tstep)
+        if self._duration is not None:
+            return int(self._duration / self._tstep)
+        ls = np.max(self._last_samples) if len(self._last_samples) > 0 \
+            else self.first_samp
+        return ls - self.first_samp + 1  # will be at least 1
+
+    @property
+    def first_samp(self):
+        return self._first_sample
```
agramfort, line 374: rather than adding a property I would just rename `self._first_sample` to `self.first_samp`

mne/simulation/tests/test_raw.py
```diff
@@ -428,6 +430,55 @@ def test_simulate_round_trip(raw_data):
         simulate_raw(raw.info, stc, None, None, None, forward=fwd)
 
 
+@pytest.mark.slowtest
+@testing.requires_testing_data
+def test_sourcesim_events(raw_data):
+    """Test of SourceSimulator usage with testing data."""
```
agramfort, line 436: can you try to use a more informative description. I am sure we have other tests. If this one is specific to a feature (first_samp?) than let's be explicit. If it's a minor change on an other test, I would just update an other test. Typically I would have started by adding a first_samp in an other test and checked that it was actually tacken into account. I am worried about the time of the mne test suite which is growing too fast

NeuroLaunch: I think this warrants its own test.  While I was motivated by the first_samp addition, there was apparently no existing check for the timing of SourceSim events as they appear in the resulting simulated raw object.  That said, I will look more closely at consolidating it into an existing test function (like the one for chpi coils).

NeuroLaunch: I simplified the SourceSim test a bit then consolidated it into the raw_bem test function in the same module, as it uses the same testing data.  It takes a fraction of the time that the previous test components take, so that's a plus.  Thanks.  I'll move this from draft to a proper PR.

mne/simulation/raw.py
```diff
@@ -149,7 +153,8 @@ def simulate_raw(info, stc=None, trans=None, src=None, bem=None, head_pos=None,
         See Notes for details.
 
         .. versionchanged:: 0.18
-           Support for tuple, and iterable of tuple or SourceEstimate.
+           Support for tuple, iterable of tuple or SourceEstimate,
+           or SourceSimulator.
```
agramfort, line 157: the stc param type description should be changed accordingly. So far it just says:

`stc : iterable | SourceEstimate`

mne/simulation/source.py
```diff
@@ -394,39 +407,42 @@ def add_data(self, label, waveform, events):
...
         Returns the stim channel data according to the simulation parameters
-        which should be added through function add_data. If both start_sample
+        which should be added through the add_data method. If both start_sample
         and stop_sample are not specified, the entire duration is used.
 
         Parameters
         ----------
         start_sample : int
-            First sample in chunk. Default is 0.
+            First sample in chunk. Default is the .first_samp value.
```
agramfort, line 429:
```suggestion
            First sample in chunk. Default is the value of the `first_samp` attribute.
```

mne/simulation/source.py
```diff
@@ -453,11 +469,10 @@ def get_stc(self, start_sample=0, stop_sample=None):
...
```
agramfort, line 472: idem

mne/simulation/tests/test_raw.py
```diff
@@ -366,6 +368,31 @@ def test_simulate_raw_bem(raw_data):
         diffs = np.sqrt(np.sum((locs - fits) ** 2, axis=-1)) * 1000
         med_diff = np.median(diffs)
         assert med_diff < tol, '%s: %s' % (bem, med_diff)
+    # also test event timings with SourceSim
```
agramfort, line 371:
```suggestion
    # also test event timings with SourceSimulator
```

doc/changes/latest.inc
```diff
@@ -146,6 +146,9 @@ Enhancements
 - :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103` by |Rotem Falach|_)
 
 
+- Add first_samp support for raw simulations with `mne.simulation.simulate_raw` and `mne.simulation.SourceSimulator` (:gh:`9166` by `Steven Bierer`_)
+
+
```
agramfort, line 151:
```suggestion
- :func:`mne.io.anonymize_info` now anonymizes also sex and hand fields when ``keep_his`` is ``False`` (:gh:`9103` by |Rotem Falach|_)

- Add first_samp support for raw simulations with `mne.simulation.simulate_raw` and `mne.simulation.SourceSimulator` (:gh:`9166` by `Steven Bierer`_)

```

mne/simulation/raw.py
hoechenberger, line 157:
```suggestion
           Support for tuple, iterable of tuple or `~mne.SourceEstimate`,
           or `~mne.simulation.SourceSimulator`.
```

mne/simulation/raw.py
```diff
@@ -271,8 +276,10 @@ def simulate_raw(info, stc=None, trans=None, src=None, bem=None, head_pos=None,
     logger.info('Setting up raw simulation: %s position%s, "%s" interpolation'
                 % (len(dev_head_ts), _pl(dev_head_ts), interp))
 
+    if isinstance(stc, SourceSimulator) and stc.first_samp != first_samp:
+        logger.info('SourceSimulator first_samp does not match argument.')
```
hoechenberger, line 280: Shouldn't we warn here? Or is it really so "unimportant"?

mne/simulation/source.py
```diff
@@ -327,6 +327,9 @@ class SourceSimulator(object):
     duration : float | None
         Time interval during which the simulation takes place in seconds.
         If None, it is computed using existing events and waveform lengths.
+    first_samp : int
+        First sample from which the simulation takes place, as an integer.
+        Comparable to the first_samp property of raw objects. Default is 0.
```
hoechenberger, line 332:
```suggestion
        Comparable to the ``first_samp`` property of `~mne.io.Raw` objects.
        Default is 0.
```

mne/simulation/source.py
hoechenberger, line 429:
```suggestion
            First sample in chunk. Default is the value of the ``first_samp`` attribute.
```

mne/simulation/source.py
hoechenberger, line 472:
```suggestion
            First sample in chunk. Default is the value of the ``first_samp`` attribute.
```

mne/simulation/source.py
```diff
@@ -453,11 +469,10 @@ def get_stc(self, start_sample=0, stop_sample=None):
         Parameters
         ----------
         start_sample : int
-            First sample in chunk. Default is 0.
+            First sample in chunk. Default is the .first_samp value.
         stop_sample : int | None
-            The stop sample of the returned stc. This sample is not part of the
-            output to follow slicing semantics. If None, then all samples past
-            start_sample is returned.
+            The final sample of the returned stc. If None, then all samples
```
hoechenberger, line 474:
```suggestion
            The final sample of the returned STC. If ``None``, then all samples
```

mne/simulation/source.py
```diff
@@ -453,11 +469,10 @@ def get_stc(self, start_sample=0, stop_sample=None):
...
+            past start_sample are returned.
```
hoechenberger, line 475:
```suggestion
            past ``start_sample`` are returned.
```

hoechenberger:
The ``stc`` paramter types for `simulate_raw()` need to be updated (originally mentioned by @agramfort)
I have already applied a number of other small suggestions & changes

NeuroLaunch:
Sorry - I 've been traveling and I just saw these comments.  Thank you for the additional suggestions and changes, @agramfort and @hoechenberger.  I'll look over the current code state and resolve anything outstanding.

mne/simulation/raw.py
NeuroLaunch: I'm not sure of the convention MNE has for generating a warning.  There could be valid reasons that a user would originally add events to SourceSimulator using one first_samp value but then declare a different first_samp value in simulate_raw();  e.g. to align the raw data with another raw or reference waveform.  Still figured some sort of indicator in the log was warranted.

hoechenberger:
Perfect! Please ping me once you're done :)

mne/simulation/source.py
```diff
@@ -336,28 +340,38 @@ class SourceSimulator(object):
...
+        return self.n_times * self._tstep
 
     @property
     def n_times(self):
         """Number of time samples in the simulation."""
-        return int(self.duration / self._tstep)
+        if self._duration is not None:
+            return int(self._duration / self._tstep)
+        ls = np.max(self._last_samples) if len(self._last_samples) > 0 \
+            else self.first_samp
```
agramfort, line 369:
```suggestion
        ls = self.first_samp
        if len(self._last_samples) <= 0:
             ls = np.max(self._last_samples)
```
is more readable I think

mne/simulation/source.py
```diff
@@ -453,11 +471,11 @@ def get_stc(self, start_sample=0, stop_sample=None):
         Parameters
         ----------
         start_sample : int
```
agramfort, line 473:
```suggestion
        start_sample : int | None
```

mne/simulation/source.py
```diff
@@ -489,16 +510,17 @@ def get_stc(self, start_sample=0, stop_sample=None):
             wf_stop = self._last_samples[ind[i]]
 
             # Recover the indices of the event that should be in the chunk
-            waveform_ind = np.in1d(np.arange(wf_start, wf_stop),
-                                   np.arange(start_sample, stop_sample))
+            waveform_ind = np.in1d(np.arange(wf_start, wf_stop + 1),
+                                   np.arange(start_sample, stop_sample + 1))
```
agramfort, line 514: why these +1? with zero-based indexing it should not happen

NeuroLaunch: The stop_sample is interpreted here as the (0-based) index of the last sample, analogous to the .last_samp property of raw objects.  So the +1 is necessary to make sure the chosen samples span the start to stop sample, inclusive.  e.g. np.arange(0, 3 + 1) = array([0, 1, 2, 3]).

NeuroLaunch: > Perfect! Please ping me once you're done :)

@hoechenberger, I am OK with the changes if you and @agramfort are.  Thank you both for your patience.

hoechenberger: Yes I agree that we need +1 here, since according to the docstring, we're working with a closed interval here

mne/simulation/tests/test_source.py
```diff
@@ -357,23 +356,50 @@ def test_source_simulator(_get_fwd_labels):
...
                        'should match'):
         ss.add_data(mylabels[0], wfs[:2], events)
+    with pytest.raises(ValueError, match='duration must be None or'):
+        ss = SourceSimulator(src, tstep, tstep / 2)
+
+    # Verify first_samp functionality.
+    ss = SourceSimulator(src, tstep)
+    offset = 50
+    for i in range(3):  # events are offset, but first_samp = 0
+        events[i][:, 0] += offset
```
agramfort, line 388:
```suggestion
        events[i, 0] += offset
```

mne/simulation/source.py
```diff
@@ -452,12 +471,12 @@ def get_stc(self, start_sample=0, stop_sample=None):
 
         Parameters
         ----------
-        start_sample : int
-            First sample in chunk. Default is 0.
+        start_sample : int | None
+            First sample in chunk. Default is the value of the ``first_samp``
+            attribute.
```
agramfort, line 476:
```suggestion
            First sample in chunk. If ``None`` the value of the ``first_samp``
            attribute is used. Defaults to ``None``.
```

agramfort (commented):
just 2 remaining nitpicks than +1 for MRG

thx @NeuroLaunch

hoechenberger (approved):
LGTM once CI comes back green

agramfort (approved):

mne/simulation/tests/test_source.py
```diff
@@ -357,23 +356,50 @@ def test_source_simulator(_get_fwd_labels):
...
                        'should match'):
         ss.add_data(mylabels[0], wfs[:2], events)
+    with pytest.raises(ValueError, match='duration must be None or'):
+        ss = SourceSimulator(src, tstep, tstep / 2)
+
+    # Verify first_samp functionality.
+    ss = SourceSimulator(src, tstep)
+    offset = 50
+    for i in range(3):  # events are offset, but first_samp = 0
+        events[i, 0] += offset
```
hoechenberger, line 388:
```suggestion
        events[i, :, 0] += offset
```

agramfort (approved):

hoechenberger:
Thank you, @NeuroLaunch!

mne/simulation/source.py
```diff
@@ -507,11 +530,13 @@ def get_stc(self, start_sample=0, stop_sample=None):
...
         """Iterate over 1 second STCs."""
-        # Arbitrary chunk size, can be modified later to something else
+        # Arbitrary chunk size, can be modified later to something else.
         # Loop over chunks of 1 second - or, maximum sample size.
         # Can be modified to a different value.
-        n_times = self.n_times
-        for start_sample in range(0, n_times, self._chk_duration):
-            stop_sample = min(start_sample + self._chk_duration, n_times)
+        last_sample = self.last_samp
+        for start_sample in range(self.first_samp, last_sample + 1,
```
larsoner, line 537: There is a failure in CircleCI here:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/7473/workflows/c20954f6-74e6-4e15-b084-d0b8c26badc2/jobs/26531

Can you look into it @NeuroLaunch ?

mne/simulation/source.py
```diff
@@ -336,28 +340,39 @@ class SourceSimulator(object):
         The number of time samples of the simulation.
     """
 
-    def __init__(self, src, tstep=1e-3, duration=None):
+    def __init__(self, src, tstep=1e-3, duration=None, first_samp=0):
+        if duration is not None and duration < tstep:
+            raise ValueError('duration must be None or >= tstep.')
+        self.first_samp = first_samp
```
larsoner, line 346: I would do:
```
self.first_samp = _ensure_int(first_samp, 'first_samp')
```
using `_ensure_int` from `..utils`. It will ensure this is an integer and raise an error if it's not.

hoechenberger:
Whoops, seems I merged too soon :<

mne/simulation/tests/test_source.py
NeuroLaunch: I'm not sure the events[i, 0] is correct!  The events array was (by a previous contributor to test_source.py) created as a list of separate event arrays... I'm a bit out of the loop now - was this change already reverted?

PR #9171, opened 2021-03-20
MAINT: reduce tutorial memory usage

drammock:
#9156 added a tutorial that is very memory-hungry, and CircleCI can't handle building it.  This PR reduces build time / memory consumption on my machine from 167.38 sec  / 3563.6 MB before this PR to  23.90 sec / 244.5 MB after this PR.  Along the way I tightened up the text and code in a few places.


drammock:
circle is happy, so in it goes!

PR #9161, opened 2021-03-19
ENH: add _repr_html_ method in the classes Info and BaseRaw

vagechirkov:
Reference issue #6846.

Reuse `report.py` code in order to produce html reports within the Info class.

Here is the output of in the IPython notebook:

image

Issue #6846, opened 2019-09-30
measurement Info shoudl have a _reprt_html method to be printed in a notebook


agramfort:
and the MNE report should be using the _repr_html method when available to produce HTML rendering of mne objects

larsoner:
I would wait to work on this until we can use https://github.com/sphinx-gallery/sphinx-gallery/issues/540 to see the outputs

agramfort:
this sphinx-gallery PR will help but it's quite orthogonal to me.

>

larsoner:
Anybody want a fun quick one for 0.21? @hoechenberger maybe?

hoechenberger:
> Anybody want a fun quick one for 0.21? @hoechenberger maybe?

Maybe I can squeeze it in somewhere, but no promises :)

vagechirkov:
```Python
import os
import mne

from IPython.display import display, HTML

sample_data_folder = mne.datasets.sample.data_path()
sample_data_raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                                    'sample_audvis_filt-0-40_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file)

string = raw.info.__repr__()


template_red = '<span1 style="color: red;  margin-left: 150px;">{}</span1>'
template_blue = '<span1 style="color: blue;">{}</span1>'
info_html_string = '<div class="MathJax_Display" style="text-align: left; background-color:rgba(201, 76, 76, 0.3);">'
info_html_string += "<style> span1 {border-radius: 5px;  font-family:fantasy;} </style>"
info_html_string += '<span1 style="color: green; margin-left: 250px; font-size: large;">INFO</span1><br>'

for line in string.split('\n')[1:-1]:
    line_parts = line.split(':')
    for n, l in enumerate(line_parts):
        if n == 0:
            info_html_string += template_red.format(l + ':')
        else:
            info_html_string += template_blue.format(l)
    info_html_string += "<br>"

display(HTML(info_html_string))
```

@hoechenberger @agramfort is it something that you expect in the ```_repr_html``` method?

agramfort:
can you share a screen shot?

>

vagechirkov:
> can you share a screen shot?
> […](#)

The design is random :)

![image](https://user-images.githubusercontent.com/42982039/111637402-ac2a6c80-87f9-11eb-9917-db76dbc11e5d.png)

hoechenberger:
[...]

PR #9172, opened 2021-03-20
ENH: Extend _repr_html_ method for Raw

vagechirkov:
Reference issue #6846 and PR #9161

Add more information in the `_repr_html_` output.

image

### DIFF
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -64,7 +64,7 @@ raw_template = Template("""
     </tr>
     <tr>
         <th>Measurement time range</th>
-        <td>{{tmin}} to {{tmax}} (HH:MM:SS)</td>
+        <td>{{tmin}} – {{tmax}} (H:MM:SS)</td>
     </tr>
 </table>
 """)
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -2,7 +2,6 @@ from ..externals.tempita import Template
 
 
 info_template = Template("""
-<h4>{{caption}}</h4>
 <table class="table table-hover">
     <tr>
         <th>Measurement date</th>
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -15,6 +15,13 @@ info_template = Template("""
         <td>{{info['experimenter']}}</td>
         {{else}}<td>Unknown</td>{{endif}}
     </tr>
+        <th>Participant</th>
+        {{if info['subject_info'] is not None}}
+            {{if 'his_id' in info['subject_info'].keys()}}
+            <td>{{info['subject_info']['his_id']}}</td>
+            {{endif}}
+        {{else}}<td>Unknown</td>{{endif}}
+    </tr>
     <tr>
         <th>Digitized points</th>
         {{if info['dig'] is not None}}
@@ -63,8 +70,8 @@ raw_template = Template("""
         <td>{{', '.join(filenames)}}</td>
     </tr>
     <tr>
-        <th>Measurement time range</th>
-        <td>{{tmin}} – {{tmax}} (H:MM:SS)</td>
+        <th>Duration</th>
+        <td>{{duration}} (HH:MM:SS)</td>
     </tr>
 </table>
 """)
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -65,7 +65,7 @@ raw_template = Template("""
     </tr>
     <tr>
         <th>Measurement time range</th>
-        <td>{{u'%0.2f' % tmin}} to {{u'%0.2f' % tmax}} sec.</td>
+        <td>{{tmin}} to {{tmax}} (hours:minutes:seconds)</td>
     </tr>
 </table>
 """)
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -60,6 +60,10 @@ info_template = Template("""
 raw_template = Template("""
 {{info_repr[:-9]}}
     <tr>
+        <th>Filenames</th>
+        <td>{{filenames}}</td>
+    </tr>
+    <tr>
         <th>Measurement time range</th>
         <td>{{u'%0.2f' % tmin}} to {{u'%0.2f' % tmax}} sec.</td>
     </tr>
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -65,7 +65,7 @@ raw_template = Template("""
     </tr>
     <tr>
         <th>Measurement time range</th>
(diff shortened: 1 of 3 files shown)


agramfort:
I would just show the basenames of the filenames and represent the
measurement time range in hour / min / seconds

>

vagechirkov:
> I would just show the basenames of the filenames and represent the measurement time range in hour / min / seconds
> […](#)

Here is an example:

image

mne/io/base.py
```diff
@@ -1651,9 +1651,17 @@ def __repr__(self):  # noqa: D105
     def _repr_html_(self, caption=None):
         if not isinstance(caption, str):
             caption = 'Raw'
+        basenames = [os.path.basename(f) for f in self._filenames]
+        m, s = divmod(self.first_time, 60)
+        h, m = divmod(m, 60)
+        tmin = '{:d}:{:02d}:{:02d}'.format(int(h), int(m), int(s))
```
hoechenberger, line 1657: Could you turn this into an f-string please?

mne/io/base.py
```diff
@@ -1651,9 +1651,17 @@ def __repr__(self):  # noqa: D105
...
+        m, s = divmod(self._last_time, 60)
+        h, m = divmod(m, 60)
+        tmax = '{:d}:{:02d}:{:02d}'.format(int(h), int(m), int(s))
```
hoechenberger, line 1660: Same: could you turn this into an f-string please?

hoechenberger:
Can we get rid of the "Raw" title? I believe it's unnecessary.

Screen Shot 2021-03-20 at 13 09 12

hoechenberger:
"Measurement date" should ideally include the time too :)

mne/data/html_templates.py
```diff
@@ -59,9 +59,13 @@
...
 raw_template = Template("""
 {{info_repr[:-9]}}
+    <tr>
+        <th>Filenames</th>
+        <td>{{' , '.join(filenames)}}</td>
+    </tr>
     <tr>
         <th>Measurement time range</th>
-        <td>{{u'%0.2f' % tmin}} to {{u'%0.2f' % tmax}} sec.</td>
+        <td>{{tmin}} to {{tmax}} (hours:minutes:seconds)</td>
```
hoechenberger, line 68: let's make this more consise:
```suggestion
        <td>{{tmin}} to {{tmax}} (HH:MM:SS</td>
```

mne/data/html_templates.py
```diff
@@ -59,9 +59,13 @@
 
 raw_template = Template("""
 {{info_repr[:-9]}}
+    <tr>
+        <th>Filenames</th>
+        <td>{{' , '.join(filenames)}}</td>
```
hoechenberger, line 64:
```suggestion
        <td>{{', '.join(filenames)}}</td>
```

vagechirkov:
> "Measurement date" should ideally include the time too :)

image

hoechenberger:
Much better!

I haven't looked at the generated HTML code, but I have the impression that there's either another line 
break or empty row now before the first one actually containing data:

Screen Shot 2021-03-20 at 13 34 58


Do you think this could be improved easily?

mne/data/html_templates.py
```diff
@@ -59,9 +59,13 @@
...
 raw_template = Template("""
 {{info_repr[:-9]}}
+    <tr>
+        <th>Filenames</th>
+        <td>{{', '.join(filenames)}}</td>
+    </tr>
     <tr>
         <th>Measurement time range</th>
-        <td>{{u'%0.2f' % tmin}} to {{u'%0.2f' % tmax}} sec.</td>
+        <td>{{tmin}} to {{tmax}} (HH:MM:SS)</td>
```
hoechenberger, line 68: Just tried locally and I think this looks better
```suggestion
        <td>{{tmin}} – {{tmax}} (H:MM:SS)</td>
```

mne/data/html_templates.py
```diff
@@ -59,9 +59,13 @@
 
 raw_template = Template("""
 {{info_repr[:-9]}}
+    <tr>
+        <th>Filenames</th>
+        <td>{{', '.join(filenames)}}</td>
+    </tr>
     <tr>
         <th>Measurement time range</th>
```
hoechenberger, line 67: Do you think we could change this to simply "Measurement time"? Would be slightly more concise … @agramfort WDYT?

vagechirkov:
> Much better!
> 
> I haven't looked at the generated HTML code, but I have the impression that there's either another line
> break or empty row now before the first one actually containing data:
> 
> Screen Shot 2021-03-20 at 13 34 58
> 
> Do you think this could be improved easily?

I think I've fixed this now :) 

image

hoechenberger:
Cool, looking good!

agramfort:
regarding measurement time range I would rather show the duration eg (6min 13s or 8h 2min 28s). For example a sleep recording can last 10h hours and it would be spanned over different days.

hoechenberger:
> regarding measurement time range I would rather show the duration eg (6min 13s or 8h 2min 28s). For example a sleep recording can last 10h hours and it would be spanned over different days.

So you're proposing

```
Duration      12:34:56 (hh:mm:ss)
```
?

I like this.

Also wondering if we should maybe include `info['subject_info']['his_id']` as well, if present:
```
Participant      Mrs X
```

agramfort:
Yes and yes 👍

vagechirkov:
image

hoechenberger (approved):

agramfort (approved):
LGTM provided you manage to make CIs green

agramfort:
Thx @vagechirkov

hoechenberger:
This is a real usability improvement, thanks a bunch!!

vagechirkov:
> This is a real usability improvement, thanks a bunch!!

Do you think it makes sense to create similar HTML-based representations for other MNE-Python objects?

hoechenberger:
> > This is a real usability improvement, thanks a bunch!!
> 
> 
> 
> Do you think it makes sense to create similar HTML-based representations for other MNE-Python objects?


I was actually thinking about exactly this, yes. Most relevant for me personally would be Epochs and Evoked; but having such a nice representation for more objects eventually could make total sense!!

vagechirkov:
> > > This is a real usability improvement, thanks a bunch!!
> > 
> > 
> > Do you think it makes sense to create similar HTML-based representations for other MNE-Python objects?
> 
> I was actually thinking about exactly this, yes. Most relevant for me personally would be Epochs and Evoked; but having such a nice representation for more objects eventually could make total sense!!

Alright, I can do this. I would create a new issue for several objects to keep progress :)

agramfort:
+1 for evoked and epochs

PR #9173, opened 2021-03-20
avoid tiny numbers passed to scikit-learn  Scaler that detects otherw…

agramfort:
…ise feature as constant in version >= 1

avoid skip from https://github.com/scikit-learn/scikit-learn/issues/19726


larsoner:
Currently -1 on this solution, let's see if I can convince @agramfort in the sklearn thread...

PR #9103, opened 2021-03-16
[MRG] anonymize sex and hand when keep_his is false

Falach:
#### Reference issue
Fixes #8130

#### What does this implement/fix?
when keep_his = false it's also anonymize hand and sex

#### Additional information
I updated the relevant tests as well

PR #9175, opened 2021-03-20
BUG: Only use valid FIFF values for anonymization

hoechenberger:
GH-9103 improved anonymization when `keep_his=False`; however, it would break I/O roundtrips because handedness
and sex would be set to values that aren't supported by the FIFF specs.

This PR changes the behavior such that anonymization sets `sex = 0` (unknown) and entirely removes `hand`, as FIFF
doesn't specify an "unknown" handedness.

Also adds more testing coverage for Raw & Epochs I/O roundtrip after anonymization.

cc @Falach

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -143,7 +143,7 @@ Enhancements
 
 - Add :func:`mne.channels.DigMontage.add_estimated_fiducials` which will add LPA, RPA and Nasion fiducial points to the ``DigMontage`` object in ``mri`` coordinate frame (:gh:`9118` by `Adam Li`_)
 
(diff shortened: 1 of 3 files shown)


hoechenberger:
@larsoner Circle still won't start for me, I've opened a ticket; hopefully they'll be able to resolve this soon.

rob-luke (approved):
Thanks for this. I hit this issue with MNE-BIDS yesterday too, so great that its addressed so quickly.

rob-luke:
Also FYI: with this fork all MNE-BIDS tests passed locally on my machine.

hoechenberger:
@rob-luke Glad to hear it!! :)

agramfort (approved):

agramfort:
Thx @hoechenberger

PR #9176, opened 2021-03-21
SSD decoding example

### LINKED ISSUE #9116: Add SSD to the Decoding (MVPA) tutorials
#### Proposed documentation enhancement
Add an example of how SSD can be used within a machine-learning pipeline. At the moment we have an [example](https://mne.tools/stable/auto_examples/decoding/plot_ssd_spatial_filters.html#sphx-glr-auto-examples-decoding-plot-ssd-spatial-filters-py) on how SSD enhances oscillatory brain activity, but not how it can be integrated within a decoding pipeline. The examples can then be added in the `tutorials/machine-learning/plot_sensors_decoding.py`, which are the moment SSD is missing there. 

My idea is to show how SSD can be beneficial before applying CSP or SPoC. A brief explanation on the formulation of SSD can be added too, as done for CSP. 

What do you think  @dengemann @larsoner @sappelhoff @drammock ??

vpeterson:
#### Reference issue
Fixes #9116.

An example of how `mne.decoding.ssd` can be used for enhancing decoding performance. A motor imagery dataset is used. Two examples are implemented, showing how SSD can be applied before data epoching, i.e. outside the decoding pipeline, and after data epoching, i.e. within the decoding pipeline. A note on how SSD can also improve the neurophysiological interpretation of the solution is done at the end of this example.

In addition, since SSD was missing at the `tutorials/machine-learning/plot_sensors_decoding.py`, a section briefly describing SSD was added.
#### Additional information
I have my personal concerns regarding the use of SSD before data epoching. Although SSD is an unsupervised spatial filtering approach, there are some parameters that are learned during its implementation. I followed what I understood was done [here](https://www.sciencedirect.com/science/article/abs/pii/S1053811914005503). Let me know what are your thoughts, I might be missing something. 

I open this PR as Draft


vpeterson:
For some weird reason, the documentation cannot be built here although my computer is working. It says something regarding memory. Any idea @drammock ???

tutorials/machine-learning/plot_sensors_decoding.py
```diff
@@ -240,6 +240,25 @@
...
 #
+# Spatio-Spectral Decomposition (SSD)
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# Spatio-Spectral Decomposition (:class:`mne.decoding.SSD`)
+# :footcite:`NikulinEtAl2011` maximizes the signal power at a peak frequency
+# while simultaneously minimizing it at the neighboring, surrounding frequency
+# bins.
+#
+# SSD is an unsupervised spatial filtering approach that can be used as a
+# pre-processing approach for data dimentionality reduction while the 1/f noise
```
agramfort, line 251:
```suggestion
# pre-processing approach for data dimensionality reduction while the 1/f noise
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,255 @@
...
+# as before, filter data
+raw_ssd.filter(freq_ssd[0], freq_ssd[1], fir_design='firwin',
+               skip_by_annotation='edge')
+###############################################################################
+# SSD + CSP + LDA pipeline
+# ^^^^^^^^^^^^^^^^^^^^^^^^^
+
+# SSD can be applied either BEFORE or AFTER data epoching, each approach has
+# it owns advantages and disadvantages. Here both approaches are going to be
+# implemented.
```
agramfort, line 103: I would argue that doing it outside the pipeline can cause a potential overfit, although maybe minimal
as it's unsupervised. I would not advertise this.

agramfort (commented):
thanks @vpeterson !

can you share the rendered example pages? how long does it take to run our system?
I fear it's too much to ask for the CI system that builds the doc.

examples/decoding/plot_decoding_ssd_csp_eeg.py
vpeterson: Exactly!! that is my maaaain concern!!! I am showing this because for other non-machine learning applications may have sense to do it before epoching, but I wouldn't be happy if this can be misleading and give the wrong message regarding how a nested ML pipeline should be trained and evaluated. So, maybe remove it? or add a warning?

drammock:
> For some weird reason, the documentation cannot be built here although my computer is working. It says something regarding memory. Any idea @drammock ???

You can run locally as `PATTERN=plot_decoding_ssd_csp_eeg html_dev-pattern-memory` to find out how much time/RAM it takes to run the example.  I can never remember what the limit is for CircleCI (@larsoner knows I'm sure)...  I do know that the limit for building *merged* PRs is even lower than for *open / work-in-progress* PRs so some memory optimization is definitely needed.

larsoner:
The rough limit is < 1.5 GB for running one example while building locally. < 1 GB is even better. There are usually ways to trim an example to get to this point via some combination of:

1. Cropping raw
2. Resampling raw or decimating epochs
3. Using fewer epochs
4. Picking fewer channels to process

without losing the point of the example. If you try some of these but can't fix things, if you can get the example to a state where it works on your machine and makes the point(s) it should make, I or @drammock could try the tricks we know to trim it and push a commit once the mem usage is down.

vpeterson:
> The rough limit is < 1.5 GB for running one example while building locally. < 1 GB is even better. There are usually ways to trim an example to get to this point via some combination of:
> 
> 1. Cropping raw
> 2. Resampling raw or decimating epochs
> 3. Using fewer epochs
> 4. Picking fewer channels to process
> 
> without losing the point of the example. If you try some of these but can't fix things, if you can get the example to a state where it works on your machine and makes the point(s) it should make, I or @drammock could try the tricks we know to trim it and push a commit once the mem usage is down.

Great, thank you!!! I will do that, I think the `for` analyzing the impact in n_components in SSD can be trimmer to the half

vpeterson:
> > For some weird reason, the documentation cannot be built here although my computer is working. It says something regarding memory. Any idea @drammock ???
> 
> You can run locally as `PATTERN=plot_decoding_ssd_csp_eeg make html_dev-pattern-memory` to find out how much time/RAM it takes to run the example. I can never remember what the limit is for CircleCI (@larsoner knows I'm sure)... I do know that the limit for building _merged_ PRs is even lower than for _open / work-in-progress_ PRs so some memory optimization is definitely needed.

I check this and the memory was 515.7 MB ...so it should work, right?

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
+"""
+.. _ex-decoding-ssd-csp-eeg:
+
+============================================================================
+Improving motor imagery decoding from EEG using Spatio Spectra Decomposition
+============================================================================
+
+Improving the decoding of Motor Imagery from EEG data when SSD is applied
```
rob-luke, line 8: SSD and CSP both need to be defined before acronyms are used. SSD is in the title, but it should be in the text too. And CSP does not seem defined anywhere.

agramfort:
here are the rendered doc pages:

https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/machine-learning/plot_sensors_decoding.html
https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_examples/decoding/plot_ssd_spatial_filters.html#sphx-glr-auto-examples-decoding-plot-ssd-spatial-filters-py
https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_examples/decoding/plot_decoding_ssd_csp_eeg.html#sphx-glr-auto-examples-decoding-plot-decoding-ssd-csp-eeg-py

I don't know why there is no plot in the second one.

I also find that the error bars are quite big.

thoughts?

vpeterson:
> here are the rendered doc pages:
> 
> https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/machine-learning/plot_sensors_decoding.html
> https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_examples/decoding/plot_ssd_spatial_filters.html#sphx-glr-auto-examples-decoding-plot-ssd-spatial-filters-py
> https://26474-1301584-gh.circle-artifacts.com/0/dev/auto_examples/decoding/plot_decoding_ssd_csp_eeg.html#sphx-glr-auto-examples-decoding-plot-decoding-ssd-csp-eeg-py
> 
> I don't know why there is no plot in the second one.
> 
> I also find that the error bars are quite big.
> 
> thoughts?

thank you for posting the docs files, I forgot to do it yesterday. The STD is also big for CSP. The reason why that could be happening is because this dataset has only 45 trials. Then, I am taking in the CV the 20% for testing (i.e. only 9 trials are held out). I can try other configurations, but then we have the risk of overfitting. I will check anyways

@dengemann here is the example

agramfort:
sorry I did not look carefully but it is enough to have these 45 trials to
make your point here experimentally? I am convinced in theory

>

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+from mne.io import concatenate_raws, read_raw_edf
+from mne.datasets import eegbci
+from mne.decoding import CSP
+from mne.decoding import SSD
+
+###############################################################################
+# Set parameters and read data
+
+# avoid classification of evoked responses by using epochs that start 1s after
+# cue onset.
```
britta-wstnr, line 38: Not sure I understand this, is cue onset at -2 seconds? Then maybe mention this (you cut from -1. to 4. below).

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+
+raw_fnames = eegbci.load_data(subject, runs)
+raw = concatenate_raws([read_raw_edf(f, preload=True) for f in raw_fnames])
+eegbci.standardize(raw)  # set channel names
+montage = make_standard_montage('standard_1005')
+raw.set_montage(montage)
+
+# strip channel names of "." characters
+raw.rename_channels(lambda x: x.strip('.'))
+# for the sake of comparison, we save the raw object before filteing
```
britta-wstnr, line 52:
```suggestion
# for the sake of comparison, we save the raw object before filtering
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+# pipeline methods
+lda = LinearDiscriminantAnalysis()
+csp = CSP(n_components=4, reg=None, log=True, norm_trace=False)
+pipe_csp = Pipeline([('CSP', csp), ('LDA', lda)])
+# this is the accuracy we would like to improve by adding SSD into the pipeline
+scores_csp = cross_val_score(pipe_csp, epochs_data, labels, cv=cv,
+                             n_jobs=1).mean()
+###############################################################################
+# SSD enhances SNR, thus what is considered 'noise' should be defined
+# Typically the 2 Hz sourronunding frequencies are taken
```
britta-wstnr, line 89:
```suggestion
# Typically bands of 2 Hz surrounding the frequencies of interest are taken
```
This needs to be reworded, I am not sure what is meant. I inserted what I think it means - but might be wrong.

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+freq_noise = [6, 14]
+# when applying SSD, data should be broader band filtered
+# a broader band of the freq_signal is defined
```
britta-wstnr, line 92:
```suggestion
# when applying SSD, data should be filtered in a broader band than for CSP,
# thus, a broader bandwidth is defined for freq_signal
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+
+# copy raw data
+raw_ssd_transformed = raw_ssd.copy()
+# define SSD filter parameters
+filt_params_signal = dict(l_freq=freq_signal[0], h_freq=freq_signal[1],
+                          l_trans_bandwidth=4, h_trans_bandwidth=4)
+filt_params_noise = dict(l_freq=freq_noise[0], h_freq=freq_noise[1],
+                         l_trans_bandwidth=4, h_trans_bandwidth=4)
+
+# Since we are working with a high electrode counting dataset (64 channels),
```
britta-wstnr, line 116:
```suggestion
# Since we are working with a data set with many electrodes (64 channels),
```

britta-wstnr: (Although it probably depends highly on the context whether 64 electrodes is considered many ... ;-)

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+# it is interesting to see the impact of the number of component in SSD, i.e.,
+# the impact in data dimensionality reduction before applying CSP.
+# The minimum n_components is 4, since 4 CSP components were selected
```
britta-wstnr, line 119:
```suggestion
# The minimum n_components is 4, since 4 CSP components were selected, and
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+# the maximum is the number of channels, here 64.
+steps = 4
+n_components = np.arange(4, 65, steps)
+scores_ssd_csp = np.zeros((len(n_components,)))
+std_ssd_csp = np.zeros((len(n_components,)))
+for n, n_comp in enumerate(n_components):
+    print('RUNNING n_components_' + str(n_comp))
+    # set to true return_filtered so as the transformed signal will be filtered
+    # at the desired frequency band.
```
britta-wstnr, line 128:
```suggestion
    # set return_filtered to True, so the transformed signal will be filtered in
    # the desired frequency band.
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+ax.set_xlabel('number of components SSD')
+ax.set_ylabel('classification accuracy')
+ax.set_xticks(np.arange(0, len(n_components), 2))
+ax.set_xticklabels(np.arange(4, 65, steps * 2))
+ax.set_title('Impact of SSD before epoching')
+
+ax.legend()
+# The results show how SSD can be used to improve decoding performance.
+# It is clear that there is a maximum achieved accuracy value, and then
+# the performance of the decoding model decreases as the n_components increases
```
britta-wstnr, line 163: This can maybe be described/explained a bit better. When is the maximum value achieved and why?

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+
+# In this context, data should be first epoched before calling pipeline.
+epochs_ssd = Epochs(raw_ssd, events, event_id, tmin, tmax,
+                    proj=True, picks=np.arange(0, n_comp, 1),
+                    baseline=None, preload=True)
+epochs_ssd_cropped = epochs_ssd.copy().crop(tmin=1., tmax=2.)
+labels = epochs_ssd.events[:, -1] - 2
+epochs_ssd_data = epochs_ssd_cropped.get_data()
+
+# Since SSD will be applied in epoched data, it is very likely that the
```
britta-wstnr, line 176:
```suggestion
# Since SSD will be applied on epoched data, it is very likely that the
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+# optimal number of SSD components be different from the value found before.
```
britta-wstnr, line 177:
```suggestion
# optimal number of SSD components will be different from the value found 
# before.
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,256 @@
...
+data_new = ssd.apply(epochs_data)
+csp.fit(data_new, labels)
+# Plot topographies.
+pattern_epochs = EvokedArray(data=csp.patterns_[:4].T,
+                             info=raw_ssd.info)
+pattern_epochs.plot_topomap(units=dict(mag='A.U.'), time_format='')
+
+# As it can be seen from the topographycal plots, the CSP patternes learned
+# after SSD was applied, enhances the left vs. right event related
+# (de)/synchronization patterns.
```
britta-wstnr, line 252:
```suggestion
# As it can be seen from the topographical plots, the CSP patterns that were
# learned after SSD was applied enhance the left vs. right event related
# (de-)synchronization patterns.
```

britta-wstnr (commented):
I did a pass on the text and comments and made a few suggestions.

examples/decoding/plot_decoding_ssd_csp_eeg.py
britta-wstnr: I agree - and a short motivation of why SSD would work for this could be added as well (it gets explained at the end of the example, but would be nice to motivate at the beginning already).

vpeterson:
> I did a pass on the text and comments and made a few suggestions.

thank you @britta-wstnr !

examples/decoding/plot_decoding_ssd_csp_eeg.py
vpeterson: Oh ! thank you, that shouldn't be there! I cropped the signal between 1 and 2 secs after the visual cue.

examples/decoding/plot_decoding_ssd_csp_eeg.py
vpeterson: Sure! I totally missed that. Please, check how does it look now.

vpeterson:
> sorry I did not look carefully but it is enough to have these 45 trials to make your point here experimentally? I am convinced in theory
> […](#)

In theory, yes, using 45 trials should be enough for training a 2 classes problem, but for the practical viewpoint, we might be tight. In order to reduce variance, I have increased the number of random splits from 5 to 10.

vpeterson:
Hi ! I have modified the decoding example following your suggestions/comments. 
[Here ](https://26588-1301584-gh.circle-artifacts.com/0/dev/auto_examples/decoding/plot_decoding_ssd_csp_eeg.html#sphx-glr-auto-examples-decoding-plot-decoding-ssd-csp-eeg-py ) you can find the new version of it. Thank you!

agramfort:
I am not sure I agree with the statements in this tutorial.

When I see https://26588-1301584-gh.circle-artifacts.com/0/dev/_images/sphx_glr_plot_decoding_ssd_csp_eeg_001.png
I see that SSD improves a bit when you overfit by using the full data
(SSD outside of the pipeline).

When you do it inside the pipeline
https://26588-1301584-gh.circle-artifacts.com/0/dev/_images/sphx_glr_plot_decoding_ssd_csp_eeg_002.png
I see no reall difference between CSP and SSD+CSP.

to me the take home message of this example should be "do not do SSD
outside of the pipeline it is not statistically OK"

vpeterson:
> I am not sure I agree with the statements in this tutorial. When I see https://26588-1301584-gh.circle-artifacts.com/0/dev/_images/sphx_glr_plot_decoding_ssd_csp_eeg_001.png I see that SSD improves a bit when you overfit by using the full data (SSD outside of the pipeline). When you do it inside the pipeline https://26588-1301584-gh.circle-artifacts.com/0/dev/_images/sphx_glr_plot_decoding_ssd_csp_eeg_002.png I see no reall difference between CSP and SSD+CSP. to me the take home message of this example should be "do not do SSD outside of the pipeline it is not statistically OK"

I totally understand your concern @agramfort , which in fact are my concerns too.  Although I personally think that all pre-processing methods should be considered within the pipeline, to avoid overfitting, this was not the intended main message of this tutorial. So, if this is what is being caught, we can just avoid the use of SSD before data epoching and try to find a better way to show the potential of SSD when used within the pipeline?

agramfort:
>
> I totally understand your concern @agramfort
> <https://github.com/agramfort> , which in fact are my concerns too.
> Although I personally think that all pre-processing methods should be
> considered within the pipeline, to avoid overfitting, this was not the
> intended main message of this tutorial. So, if this is what is being
> caught, we can just avoid the use of SSD before data epoching and try to
> find a better way to show the potential of SSD when used within the
> pipeline?
>

+1 to consider this unconvincing. Ok for you?

vpeterson:
> I totally understand your concern @agramfort <https://github.com/agramfort> , which in fact are my concerns too. Although I personally think that all pre-processing methods should be considered within the pipeline, to avoid overfitting, this was not the intended main message of this tutorial. So, if this is what is being caught, we can just avoid the use of SSD before data epoching and try to find a better way to show the potential of SSD when used within the pipeline?
> +1 to consider this unconvincing. Ok for you?

This morning I realized that the window size was too small for running SSD within the pipeline. I haven't finalized yet the new example version, but this is the output I got now. Are you suggesting closing this PR or not including the example of SSD outside the pipeline?

Figure_1
Figure_2

agramfort:
if we run this for more subjects and average can it be even clearer? I find
it still not super obvious sorry

>

vpeterson:
> if we run this for more subjects and average can it be even clearer? I find it still not super obvious sorry
> […](#)

Sure, I can do that, it might take more memory but hopefully, the docfiles would be able to be compiled anyway. Anyhow, in the best-case scenario, we would achieve something similar to Figure 4b of [Haufe et al.](https://www.sciencedirect.com/science/article/abs/pii/S1053811914005503) paper.  I think that the last attached Figure (SSD after epoching) is, more or less,  already in that shape, isn't it?. I can also connect tomorrow to Discord to further discuss this if you are available @agramfort ! thank you!

agramfort:
yes exactly like this figure. I see it's obtained by averaging 80 subjects.
Let's see
if you can replicate this figure without limitation of computation time on
your machine
and then we try to find a compromise on the example

ok?

>

vpeterson:
Hi @agramfort. I have run the script over the first 100 subjects available in the dataset. Then, I picked up those subjects whose decoding performance was around the same value (to avoid large std values due to differences in performance). This is what I got for those 10 subjects when SSD is applied after epoching.

Figure_ssd_10subjects

I think these results better show now that SSD can be used before decoding for improving performance. As also reported by the authors, overall classification performance reaches its maximum for a small SSD number of components (here between 6 and 12), and then classification decreases as the number of components increases. 

Also, for one of those subjects, I plot the topographical map of the CSP patterns. This is how the patterns look without SSD 
Figure_csp_patterns

and these are the CSP patterns for the same subject when SSD was applied before learning CSP. 
Figure_ssd_csp_patterns

in which, since the experiment is done base on the alpha band, we can see how the patterns now better enhance the occipital region (which could be more related to the visual stimulus rather than the MI process)

Let me know if this makes more sense now. Anyhow, I understand if you consider this example is not strong enough and the PR should then be closed. 

Thank you,

agramfort:
this was obtained using a proper pipeline? without overfitting?
if so you would update the tutorial demoing this? how long does the example
take to run?

>

vpeterson:
> this was obtained using a proper pipeline? without overfitting? if so you would update the tutorial demoing this? how long does the example take to run?
> […](#)

yes, it is obtained with the non-overfitting pipeline. But on my computer takes 20 minutes to run everything. I understand that is not doable for online docfiles creation, right? I will commit the changes anyway so as to keep track of that here too.

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
+"""
+.. _ex-decoding-ssd-csp-eeg:
+
+============================================================================
+Improving motor imagery decoding from EEG using Spatio Spectra Decomposition
+============================================================================
+
+The Spatio Spectra Decomposition (SSD) is a supervised spatial filtering
```
agramfort, line 8:
```suggestion
The Spatio Spectra Decomposition (SSD) is a unsupervised spatial filtering
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+algorithm capable of extracting components primarily related to the signal of
+interest. It can be used as a pre-processing step for the analysis of neuronal
```
agramfort, line 10:
```suggestion
algorithm that can be used as a pre-processing approach for data dimensionality
reduction while the 1/f noise in the neural data is reduced :footcite:`HaufeEtAl2014b`.
It is useful extract to capture induced activity during motor imagery.
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+activity. In this example SSD will be used in the context of Motor Imagery
+decoding from EEG data. It will be applied before extracting features with
```
agramfort, line 12:
```suggestion
activity. SSD will be applied before extracting features with
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
+from sklearn.model_selection import ShuffleSplit, cross_val_score
+
+from mne import (Epochs, pick_types, events_from_annotations,
+                 EvokedArray)
+from mne.channels import make_standard_montage
+from mne.io import concatenate_raws, read_raw_edf
+from mne.datasets import eegbci
+from mne.decoding import CSP
+from mne.decoding import SSD
```
agramfort, line 37:
```suggestion

from mne.decoding import SSD
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+###############################################################################
+# Set parameters and read data
+
+# epochs from 0.5 to 2.5 s after the cue onset will be used
+tmin, tmax = 0.5, 2.5
+event_id = dict(left=2, right=3)  # Motor imagery: left vs right hand
+runs = [4, 8, 12]
+Subjects_set = [2, 7, 31, 34, 42, 56, 60, 62, 85, 100]
```
agramfort, line 45:
```suggestion
subjects = [2, 7, 31, 34, 42, 56, 60, 62, 85, 100]
```
why? it's not a set it's more a list. Also in Python variables names (unless they are constants) should be in snake_case (no capital letters which are reserved for class names).

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+NSubject = len(Subjects_set)
```
agramfort, line 46:
```suggestion
n_subjects = len(subjects)
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+# the impact in data dimensionality reduction before applying CSP.
+# The minimum n_components is 4, since 4 CSP components were selected
+# the maximum is the number of channels, here 64.
+steps = 4
+n_components = np.arange(4, 65, steps)
+scores_ssd_csp_e = np.zeros((NSubject, len(n_components)))
+std_ssd_csp_e = np.zeros((NSubject, len(n_components)))
+
+for s in range(NSubject):
+    subject = Subjects_set[s]
```
agramfort, line 70:
```suggestion
for s_idx, subject in enumerate(subjects):
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+    cv = ShuffleSplit(10, test_size=0.2, random_state=42)
+    # pipeline methods
+    lda = LinearDiscriminantAnalysis()
+    csp = CSP(n_components=4, reg=None, log=True, norm_trace=False)
+    pipe_csp = Pipeline([('CSP', csp), ('LDA', lda)])
+    # this is the accuracy we would like to improve by adding SSD into the
+    # pipeline
+    scores_csp[s] = cross_val_score(pipe_csp, epochs_data, labels, cv=cv,
+                                    n_jobs=1).mean()
+    std_csp[s] = cross_val_score(pipe_csp, epochs_data, labels, cv=cv,
```
agramfort, line 107:
```suggestion
    std_csp[s_idx] = cross_val_score(pipe_csp, epochs_data, labels, cv=cv,
```

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+                                 n_jobs=1).std()
+###############################################################################
+    # SSD + CSP + LDA pipeline
+    # ^^^^^^^^^^^^^^^^^^^^^^^^
+
```
agramfort, line 112:
```suggestion
    # SSD + CSP + LDA pipeline
    # ^^^^^^^^^^^^^^^^^^^^^^^^

```

examples/decoding/plot_decoding_ssd_csp_eeg.py
agramfort: avoid single letter variable names

examples/decoding/plot_decoding_ssd_csp_eeg.py
```diff
@@ -0,0 +1,237 @@
...
+ax.set_xticks(np.arange(0, len(n_components), 2))
+ax.set_xticklabels(np.arange(4, 65, steps * 2))
+ax.set_ylim(0.6, 1.1)
+ax.set_title('Impact of SSD as a function of n_components')
+ax.legend()
+
+# SSD acts as a regularizer. Here we used it as a dimensionality reduction tool
+# for improving discriminative power in the alpha-band. On average, the best
+# performance was achieved between 6 and 12 components, value that lies
+# between the range found by the authors in :footcite:`HaufeEtAl2014b`.
```
agramfort, line 165:
```suggestion
# between the range found by the authors in :footcite:`HaufeEtAl2014b`.

```

examples/decoding/plot_decoding_ssd_csp_eeg.py
vpeterson: oh, I was just thinking in "set" as a "dataset/subset", but sorry I forgot about the naming convention here!

vpeterson (commented):
adding review changes

drammock:
we do have 1 or 2 tutorials that we simply show the code but don't execute them during our doc build.  This is achieved by naming the tutoral `something.py` instead of `plot_something.py`.  Is that a suitable option here @agramfort?  Or is the result of the 20-minute build still not convincing enough to be worth showcasing?

PR #9179, opened 2021-03-22
FIX: Brain locale

GuillaumeFavelier:
This small PR fixes the issue with `Brain` locale.

master | PR
--|--
image |image

It's an item of #7162

### DIFF
--- a/mne/viz/backends/_qt.py
+++ b/mne/viz/backends/_qt.py
@@ -10,7 +10,7 @@ from functools import partial
 
 import pyvista
 
-from PyQt5.QtCore import Qt, pyqtSignal
+from PyQt5.QtCore import Qt, pyqtSignal, QLocale
 from PyQt5.QtGui import QIcon
 from PyQt5.QtWidgets import (QComboBox, QDockWidget, QDoubleSpinBox, QGroupBox,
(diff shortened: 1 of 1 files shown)


GuillaumeFavelier:
Feel free to try it @hoechenberger, @agramfort :+1:

hoechenberger:
Perfect, thanks!

hoechenberger (approved):

larsoner:
Am I remembering correctly that the colorbar has the same problem on your system? Did this fix that as well, or do we need a fix on the VTK end for that?

GuillaumeFavelier:
> the colorbar has the same problem on your system?

Yes I just noticed. I don't know a fix for that except playing with my env variables.

larsoner:
Can you try doing `QLocale.setDefault(QLocale.c())` at the first import of `Qt` (before any other VTK windows / applications are created? It's not the right solution to change the global state like this (at least not without setting it back), but it would at least tell us if it's perhaps the Qt global state change breaking things in VTK:

https://gitlab.kitware.com/vtk/vtk/-/issues/17660

Also can you see if you have this problem in a VTK / PyVista scalarbar example that does not use Qt? It would help us isolate the problem to VTK+Qt instead of VTK alone as well

PR #9180, opened 2021-03-22
ENH: Traces dock widget

GuillaumeFavelier:
This PR follows https://github.com/mne-tools/mne-python/pull/8997#issuecomment-802891396 and brings support for docked traces:

This is how it looks by default for me on `i3`:

image

And the dock widgets can be detached/resized/reattached:

image

*Locally, `Brain` testing works but I don't know how `macOS` CIs will react* :pray:

<details>

```py
import os
import mne

initial_time = 0.096
                 surface="white",
                 size=600,
                 background="white")

```
</details>

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -654,7 +654,7 @@ class Brain(object):
         self._configure_help()
         # show everything at the end
         self.toggle_interface()
-        self._renderer._window_show()
+        self._renderer.show()
 
         # sizes could change, update views
         for hemi in ('lh', 'rh'):
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -413,7 +413,7 @@ class Brain(object):
         if len(size) not in (1, 2):
             raise ValueError('"size" parameter must be an int or length-2 '
                              'sequence of ints.')
-        self._size = size if len(size) == 2 else size * 2  # 1-tuple to 2-tuple
+        size = size if len(size) == 2 else size * 2  # 1-tuple to 2-tuple
         subjects_dir = get_subjects_dir(subjects_dir)
 
         self.time_viewer = False
@@ -466,7 +466,7 @@ class Brain(object):
             offset = (surf == 'inflated')
         offset = None if (not offset or hemi != 'both') else 0.0
 
-        self._renderer = _get_renderer(name=self._title, size=self._size,
+        self._renderer = _get_renderer(name=self._title, size=size,
                                        bgcolor=background,
                                        shape=shape,
                                        fig=figure)
@@ -654,7 +654,7 @@ class Brain(object):
         self._configure_help()
         # show everything at the end
         self.toggle_interface()
-        self._renderer._window_show(self._size)
+        self._renderer._window_show()
 
         # sizes could change, update views
         for hemi in ('lh', 'rh'):
@@ -717,7 +717,7 @@ class Brain(object):
             self.visibility = value
 
         # update tool bar and dock
-        with self._renderer._window_ensure_minimum_sizes(self._size):
+        with self._renderer._window_ensure_minimum_sizes():
             if self.visibility:
                 self._renderer._dock_show()
                 self._renderer._tool_bar_update_button_icon(
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -478,8 +478,7 @@ class Brain(object):
                                        bgcolor=background,
                                        shape=shape,
                                        fig=figure)
-
-        self._renderer._window_initialize(self._clean)
+        self._renderer._window_close_connect(self._clean)
         self._renderer._window_set_theme(theme)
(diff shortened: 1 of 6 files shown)


GuillaumeFavelier:
Maybe `show_traces="separate"` is not necessary anymore and should be removed.

GuillaumeFavelier:
I think the result is better when the margins are disabled for the traces widget :)

margin=True | margin=False
--|--
image | image

GuillaumeFavelier:
It's ready for reviews @larsoner, @agramfort

larsoner:
From a quick test this is my experience / thoughts:

1. This will be a great enhancement!

2. The matplotlib widget has a minimum width that is too big -- I should be able to resize it smaller than this:

    https://user-images.githubusercontent.com/2365790/111998156-1d1b9e00-8af2-11eb-8eab-144dc0075e61.mp4

3. The resize-boundary-grabber (what you can click on to actually resize the widget) once it has been removed appears to be 1 pixel. Maybe this is related to the removal of some border or padding (?) but it makes it very difficult to resize

4. Removing the control and matplotlib widgets is easy because they have titles. The "brain plots widget" should also have a title so it can just as easily be moved

GuillaumeFavelier:
`2)` should be fixed after aa23887828ca6a70d64c697d9306897196c386d8, I modified `_window_ensure_minimum_sizes` so that the minimum size for the mplcanvas is more flexible. I did not manage to make it work without the margins but maybe it will help with `3)` in the end.

About `4)`, I still have to find a way to do that :)

Locally, the notebook test was failing so I fixed it in here. I don't know why it was not detected.

```
E   ~/source/mne-python/mne/viz/_brain/_brain.py in _configure_tool_bar(self)
E      1232     def _configure_tool_bar(self):
E      1233         self._renderer._tool_bar_load_icons()
E   -> 1234         self._renderer._tool_bar_set_theme(self.theme)
E      1235         self._renderer._tool_bar_initialize()
E      1236         self._renderer._tool_bar_add_screenshot_button(
E   
E   AttributeError: '_Renderer' object has no attribute '_tool_bar_set_theme'
```

larsoner:
> About 4), I still have to find a way to do that :)

Could you embed the plotter in a group box, and make that group box what's movable?

larsoner:
> I did not manage to make it work without the margins but maybe it will help with 3) in the end.

Still 1 pixel for me, can you replicate? If so, could you try the margin stuff suggested here:

https://stackoverflow.com/questions/60279502/floating-qdockwidget-resize-change-size-of-the-resize-handler-grip-on-the-bord

If not, I can work on it, since it might actually be the fault of my Ubuntu theme:

https://forum.qt.io/topic/60122/increase-floating-qdockwidget-resize-margin-size

GuillaumeFavelier:
> Could you embed the plotter in a group box

I know how to do this

> make that group box what's movable?

But not that. I will make some research :sweat_smile: 

> Still 1 pixel for me, can you replicate?

With `i3`, I resize with the keyboard. I should switch to my KDE to replicate :+1:

larsoner:
> But not that. I will make some research ￼

I'd assume that however you're currently making the matplotlib window movable, you "just" need to do that same thing with a new groupbox widget, then make the plotter a child of that group box. Hopefully it's not more complicated than this, but if it is, I apologize :) If it ends up being really complicated maybe it's not worth pursuing.

GuillaumeFavelier:
> however you're currently making the matplotlib window movable

Ah this came basically for free with `QDockWidget`, it's part of the default features:

https://doc.qt.io/qt-5/qdockwidget.html#DockWidgetFeature-enum

larsoner:
So add a new GroupBox to that widget, put the plotter in that group box, and maybe it works?

hoechenberger:
> And the dock widgets can be detached/resized/reattached:

Haven't tried this out yet, but hopefully one cannot "accidentally" detach the docks? Because that's something I hate :D These multi-window designs are … idk, is anyone even still using that? Not even GIMP comes with that stuff anymore by default! 😅

larsoner:
From trying it out, I find it works pretty well. You can move/detach/reattach easily.

GuillaumeFavelier:
I tried in 8d836cd but it was some sort of frankenstein, it's way too hacky and not stable in my opinion :sweat_smile: 

https://user-images.githubusercontent.com/18143289/112320591-6fee8480-8caf-11eb-960b-ee071bcc0b31.mp4

larsoner:
Those interactions seem okay. Why not make the `QDockWidget` the central widget?

larsoner:
Ohhhhh I see. You need a central widget and things go around it. I can live with the brain plots being this central widget :)

larsoner:
In that case we should be good to go here, then?

larsoner (approved):

GuillaumeFavelier:
Ready on my end. Do you want to take care of `3)` in a different PR maybe?

agramfort (approved):
provided CIs are green

larsoner:
Fixed the grip issue

larsoner:
Thanks @GuillaumeFavelier !

PR #9184, opened 2021-03-22
MRG: Additional checking of fNIRS data before processing

rob-luke:
#### Reference issue
Further preparation for #9141 

Now that we are going to have people creating their own data structures it will become more important to thoroughly check the ordering and encoding of fNIRS metadata.

#### Additional information
I will add checking for hbo/hbr type next, but wanted to thoroughly test the existing function before I expand it.


rob-luke:
Render: https://26465-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_30_reading_fnirs_data.html#storing-of-optode-locations

rob-luke:
I think this is good to go (assuming CIs come back green, they do locally). @larsoner @agramfort @drammock could you please review.

tutorials/io/plot_30_reading_fnirs_data.py
```diff
@@ -68,9 +68,37 @@
...
+             ``["S11_D2 hbo", "S11_D2 hbr", "S1_D2 hbo", "S1_D2 hbr"]``
+             is acceptable, but
+             ``["S11_D2 hbo", "S1_D2 hbo", "S11_D2 hbr", "S1_D2 hbr"]``
+             is not. Further, the order of type must be maintained for all
+             fNIRS channels so the following is not valid
+             ``["S11_D2 hbo", "S11_D2 hbr", "S1_D2 hbr", "S1_D2 hbo"]``.
+
+             For raw amplitude measurements and for optical density data
+             the wavelength information must be stored in
+             ``info["chs"][ii]["loc"][9]``
```
larsoner, line 98: Everything before this (channel names and ordering) has a public API that allows it to be changed easily. But this part -- about `info['chs'][ii]['loc']` -- does not, and exists in an info field that we say [shouldn't be modified by people](https://mne.tools/dev/generated/mne.Info.html#mne.Info). So this I'd rather add somewhere else as it doesn't seem right for a user-facing tutorial.

larsoner: ... so really I think I would recommend:

1. Move this text in its entirety to somewhere else `meas_info.py` maybe? Or `mne/preprocessing/*nirs*`?
2. Reword this to be much shorter, warning people that the channel names and ordering are set properly by MNE reading and preprocessing functions and shouldn't be changed by the user by `raw.rename_channels`, `raw.reorder_channels`, etc.

rob-luke: > as it doesn't seem right for a user-facing tutorial

Thanks @larsoner, I agree that users shouldn't be editing or creating fields like this and have expressed that in #9141.

I will remove the tutorial section from this PR and leave these details for discussion in  #9141. And keep this PR as just backend improvements.

rob-luke:
Done. This PR is now only additional back end checks for fNIRS data. @larsoner could you please review again.

larsoner (approved):
LGTM and I think a `latest.inc` change is not necessary since this seems mostly like internal refactoring / tests that shouldn't affect users (just devs implementing new things), feel free to merge once CIs come back happy @rob-luke

rob-luke:
Thanks @larsoner Ill merge if all green.

PR #9185, opened 2021-03-23
MRG: Test fnirs channel structures for chromophore data

rob-luke:
#### Reference issue
Preparation for #9141 

Adds checking of fNIRs channel naming structure for data after transformation to chromophore.

Previously all testing was done on the raw data or after transformation to optical density. But with possibility of people creating their own data structures without the readers I have added additional error checking for the hbo/hbr data type.

#### Additional information
This is the last addition I have planned in preparation for #9141

### DIFF
--- a/mne/preprocessing/nirs/tests/test_nirs.py
+++ b/mne/preprocessing/nirs/tests/test_nirs.py
@@ -176,7 +176,7 @@ def test_fnirs_channel_naming_and_order_readers(fname):
     raw_names_reversed = raw.copy().ch_names
     raw_names_reversed.reverse()
     raw_reversed = raw.copy().pick_channels(raw_names_reversed, ordered=True)
-    with pytest.raises(ValueError, match='not ordered correctly'):
+    with pytest.raises(ValueError, match='not ordered .* frequencies'):
         _check_channels_ordered(raw_reversed, freqs)
     # So if we flip the second argument it should pass again
     _check_channels_ordered(raw_reversed, [850, 760])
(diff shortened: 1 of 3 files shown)


rob-luke:
@larsoner could you please review. This should be the last improvement to these checks, I think I have covered most edge cases now (famous last words).

mne/preprocessing/nirs/tests/test_nirs.py
```diff
@@ -306,3 +310,54 @@ def test_fnirs_channel_naming_and_order_custom_optical_density():
     # and this is how you would fix the ordering, then it should pass
     raw.pick(picks=[0, 3, 1, 4, 2, 5])
     _check_channels_ordered(raw, [760, 850])
+
+
+def test_fnirs_channel_naming_and_order_custom_chroma():
+    """Ensure fNIRS channel checking on manually created data."""
+    data = np.random.normal(size=(6, 10))
```
larsoner, line 317: I know it doesn't currently matter since you don't use the data for anything, but let's use a seeded random var in case we later change the test to use the data (since we always try to do it, it's reasonable to assume it has been done here):
```suggestion
    data = np.random.RandomState(0).randn(6, 10)
```
Or if you want it to be simpler you can just use `np.zeros(...)`

larsoner (commented):
Otherwise LGTM

I'll push a commit for my suggestion since it's almost trivial, then we can merge

larsoner:
Thanks @rob-luke !

rob-luke:
Thanks @larsoner

PR #9187, opened 2021-03-23
MRG: Support µV in EDF files (in addition to uV)

### LINKED ISSUE #9186: Reading EDF file yields incorrect scaling
Reported on the forum:
https://mne.discourse.group/t/eeg-data-in-edf-not-converting-in-volt/

The user provided an EDF file which specifies in its header that the data is in µV. When reading this file with MNE-Python, the data is treated as V.

Sumalyo:
This shows a warning if edf file data is given in Volts as mentioned in the issue discussions. Though it interprets the data in Volts, it informs the data about a possible problem in scaling. Fixes #9186.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -19,6 +19,8 @@ Current (0.23.dev0)
 
 .. |New Contributor| replace:: **New Contributor**
 
+.. |Sumalyo Datta| replace:: **Sumalyo Datta**
+
 .. |Anna Padee| replace:: **Anna Padee**
 
 .. |Richard Koehler| replace:: **Richard Koehler**
@@ -157,6 +159,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix bug with :func:`mne.io.read_raw_edf` where new symbols for interpreting data in microvolt was added (:gh:`9187` **by new contributor** |Sumalyo Datta|_ )
+
 - Fix bug with :func:`mne.viz.plot_compare_evokeds` did not check type of combine. (:gh:`9151` **by new contributor** |Matteo Anelli|_)
                                         
 - Fix bug with :func:`mne.viz.plot_evoked_topo` where ``ylim`` was only being applied to the first channel in the dataset (:gh:`9162` **by new contributor** |Ram Pari|_ )
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -159,7 +159,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix bug with :func:`mne.io.read_raw_edf` where new symbols for interpreting data in microvolt was added (:gh:`9187` **by new contributor** |Sumalyo Datta|_ )
+- Fix bug with :func:`mne.io.read_raw_edf` where µV was not correctly recognized (:gh:`9187` **by new contributor** |Sumalyo Datta|_)
 
 - Fix bug with :func:`mne.viz.plot_compare_evokeds` did not check type of combine. (:gh:`9151` **by new contributor** |Matteo Anelli|_)
                                         
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -1452,4 +1452,3 @@ def _get_annotations_gdf(edf_info, sfreq):
         desc = events[2]
 
     return onset, duration, desc
-
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -615,6 +615,9 @@ def _read_edf_header(fname, exclude):
                 continue
             if unit == 'uV':
                 edf_info['units'].append(1e-6)
+            if unit == 'V':
+                warn('Unit given in file is Volts(V), please check file and/or adjust processing if necessary')
+                edf_info['units'].append(1)
             elif unit == 'mV':
                 edf_info['units'].append(1e-3)
             else:
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -616,11 +616,11 @@ def _read_edf_header(fname, exclude):
             if unit == 'uV':
                 edf_info['units'].append(1e-6)
             if unit == 'V':
-                warn('Unit given in file is Volts(V), please check file and/or adjust processing if necessary')
                 edf_info['units'].append(1)
             elif unit == 'mV':
                 edf_info['units'].append(1e-3)
             else:
+                warn('Unit not recognized, treated as Volts(V)')
                 edf_info['units'].append(1)
         edf_info['units'] = np.array(edf_info['units'], float)
 
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -1452,4 +1452,4 @@ def _get_annotations_gdf(edf_info, sfreq):
         desc = events[2]
 
     return onset, duration, desc
-    
+
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -613,14 +613,14 @@ def _read_edf_header(fname, exclude):
         for i, unit in enumerate(units):
             if i in exclude:
                 continue
-            if unit == 'uV':
-                edf_info['units'].append(1e-6)
-            if unit == 'V':
+            elif unit == 'V':
                 edf_info['units'].append(1)
             elif unit == 'mV':
                 edf_info['units'].append(1e-3)
+            elif unit == 'uV':
+                edf_info['units'].append(1e-6)
             else:
-                warn('Unit not recognized, treated as Volts(V)')
+                warn(f'Unit "{unit}" not recognized, no scale used')
                 edf_info['units'].append(1)
         edf_info['units'] = np.array(edf_info['units'], float)
 
--- a/mne/io/edf/edf.py
+++ b/mne/io/edf/edf.py
@@ -613,14 +613,11 @@ def _read_edf_header(fname, exclude):
         for i, unit in enumerate(units):
             if i in exclude:
                 continue
-            elif unit == 'V':
-                edf_info['units'].append(1)
+            if unit in ('uV', 'µV', 'µV'):
+                edf_info['units'].append(1e-6)
             elif unit == 'mV':
                 edf_info['units'].append(1e-3)
-            elif unit == 'uV':
-                edf_info['units'].append(1e-6)
             else:
-                warn(f'Unit "{unit}" not recognized, no scale used')
                 edf_info['units'].append(1)
         edf_info['units'] = np.array(edf_info['units'], float)
 
@@ -1455,3 +1452,4 @@ def _get_annotations_gdf(edf_info, sfreq):
         desc = events[2]
 
     return onset, duration, desc
+    
--- a/doc/changes/names.inc
+++ b/doc/changes/names.inc
@@ -383,3 +383,5 @@
 .. _Ram Pari: https://github.com/ramkpari
 
 .. _Erica Peterson: https://github.com/nordme
+
+.. _Sumalyo Datta: https://github.com/Sumalyo


cbrnr:
I think this doesn't quite do what we want. We would like to issue a warning if we detect an unknown unit and simply assume it's in V. Currently, we only detect mV and uV and assume everything else is in V. Your specific file contains units ".V", so I suggest to:

1. Add `V` to the list of supported units (and hence issue no warning).
2. If the unit is not `mV`, `uV` or `V`, issue a warning but treat it as `V`.

Sumalyo:
@cbrnr  Sorry, I misunderstood the issue, so we issue no warning if the unit is `V` and show a warning if we encounter any other unit telling the user it's being interpreted as `Volts(V)`. I will make this change. Hence something like:
image
Please let me know if this is what you want, I would like to work on this issue. Thank you.

cbrnr:
Yes, that's exactly what I'd do!

Sumalyo:
I have made these changes in the PR, please review @cbrnr 
https://github.com/mne-tools/mne-python/pull/9187/commits/29073dd80ab47874f931f73f6f9c6195c1f05630

Sumalyo:
@larsoner @cbrnr I have made relevant changes to fix #9186 and show warnings when an unknown unit is encountered while loading the edf header and would be read as Volts in MNE by default; this seems to fail some tests; what should I do to fix that?

cbrnr:
Some of our tests apparently use unsupported units. I'll take a look.

Sumalyo:
Thank you @cbrnr

mne/io/edf/edf.py
```diff
@@ -615,9 +615,12 @@ def _read_edf_header(fname, exclude):
                 continue
             if unit == 'uV':
                 edf_info['units'].append(1e-6)
+            if unit == 'V':
+                edf_info['units'].append(1)
```
cbrnr, line 619: Move these two lines to the top such that the order is `V`, `mV`, and `uV`. You need to create `if`/`elif`/`else` branches in order for this to work (previously you used another `if` clause).

cbrnr:
Also, I think the `else` clause should take care of all unknown units and say just that (unknown unit). We can't assume that it is Volts, so the warning could just be:

```
warn(f'Unit "{unit}" not recognized.')
```

@larsoner it is pretty common that EDF/BDF files contain units other than V/mV/uV - should we really issue a warning? I'm thinking that maybe an info would be better suited.

mne/io/edf/edf.py
Sumalyo: I hope this would fix that issue
image

Sumalyo:
@cbrnr should I add something like 
`logger.info('f,Unit "{unit}" not recognized as V/mV/uV. No scaling applied to data. Adjust preprocessing accordingly if necessary')`

mne/io/edf/edf.py
cbrnr: Yes, this looks good! You _could_ start with an `if` after the `continue` statement, but an `elif` is also fine!

cbrnr:
Let's wait for @larsoner. I'd either leave the warning from your most recent commit, or just replace `warn` with `logger.info` (I'd prefer the latter).

Sumalyo:
@cbrnr Thank you

cbrnr:
Yes, the remaining tests still fail because these data sets contain unknown units (e.g. `G` and `Boolean`). Currently, the tests fail because now a warning is raised. If we demote the warning to an info message all will be fine.

larsoner:
>  it is pretty common that EDF/BDF files contain units other than V/mV/uV - should we really issue a warning? I'm thinking that maybe an info would be better suited.

To me `warn` should be used when we really want people to see something and think about it and make sure things are okay. `info` should be used when we want to give people extra information that might be useful but is not necessarily essential because they might not ever see it (e.g., with a common default MNE_LOGGING_LEVEL=warning). So I think we need to decide which of these two "might have the wrong units" lands in for EDF/BDF.

One option is to keep track of *which* channels we don't know the units of. If the channel is a *data* channel (EEG, MEG, etc.) then we `warn`. If it's something else (STIM, MISC, etc.), then we `info`. To me this makes sense since the data channels are the ones for which getting units correct is probably most critical for people's analyses. Maybe whichever channels have `Boolean` (maybe a stim channel) and `G` (gravitational constant?) won't trip the warning with this approach...

cbrnr:
In principle, that's a good idea. However, `read_raw_edf` by default assigns type `eeg` to all channels except for those named `status` or `trigger`. There is an optional `misc` parameter, which allows users to pass channel names that should be interpreted as type `misc`.

So practically, this means that we cannot really distinguish between different channel types.

larsoner:
Would the right way to use the currently problematic files be to pass `misc=...` to the `read_raw_edf`? If so then I think warning that we don't know the units of the channel with a suitable message (including a suggestion that if they are not EEG type, to pass them as `misc`) is helpful and still a good thing to do.

cbrnr:
Yes, , since we're assuming `eeg` by default it makes sense to warn if we encounter anything else than `V`, `mV` or `uV`. @Sumalyo this will require adding a `misc` parameter to `_read_edf_header` in order to distinguish these cases. Let me know if you need help implementing this.

Sumalyo:
@cbrnr I understand that currently that `misc` is only dealt with in method `_get_info` but thereafter it is never passed on in `_read_header` and subsequently `_read_edf_header` , which accounts for the units. I am working on this trying to check if data with unknown units, is from a `misc` channel if not then issuing a warning and a suggestion in that case. Should I also add a `loger.info` if a `misc` channel is processed?

cbrnr:
Great @Sumalyo! No, I don't think that a `logger.info` is necessary.

Sumalyo:
@cbrnr ok sure. I will only issue the suggestion and warning if the channel index/name is not in `misc` and it still has units other than `V` , `mV` and `uV`

Sumalyo:
@cbrnr I have tried out something with a branch in my fork. Could this work??
https://github.com/Sumalyo/mne-python/pull/1/commits/e37f62260fe3e654fe5d2b885931d10566dc7d48

Sumalyo:
@cbrnr I decided to test the modifications on my system with the file provided in the forum and observed something interesting when I opened it up on hex editor. As given below:
hex_forum
the file apparently has units `µV` instead of `uV` and thus the appropriate scaling could be achieved with a fix like this:
fix_dump
which would resolve the scaling issue referred to in the forum posts for the file.
However, I implemented the full fix and tested it out by **corrupting** the edf file to have the units of channel 1 and 2 as `.V` :
hex_faulty
and tested my fix by reading the edf file and issuing a warning if the channels are not marked as misc. I wrote a small python script to test all cases: reading a legitimate edf file (passing a file from another experiment with all proper units); reading faulty files with and without `misc` (issuing a warning when appropriate) and then reading the file in the forum without `misc` (using the `µV` unit fix)
cmd_dump


@cbrnr @larsoner Please help me understand what I should include in this PR further; as issuing the warning with channels not marked misc causes CI checks to fail. (like `test_edf.py::test_bdf_dat`)
I have included all relevant scripts and supporting files in https://github.com/Sumalyo/mne-python/tree/reportedf/mne/edf_test
the python code used for reading the edf is :
`from mne.io import concatenate_raws, read_raw_edf`
`print("Legitimate EDF")`
`raw = read_raw_edf('Subject01_1.edf', preload=True, stim_channel='auto', verbose=True)`
`print("Faulty EDF")`
`raw = raw = read_raw_edf('fault.edf', preload=True, stim_channel='auto', verbose=True)`
`print("Faulty EDF with Misc.")`
`raw = read_raw_edf('fault.edf', preload=True, stim_channel='auto', verbose=True,misc=[0,1])`
`print("EDF File in Forum")`
`raw = read_raw_edf('subject 2 trial 1(5).edf',preload=True, stim_channel='auto', verbose=True)`
`

cbrnr:
You're right, the sample file actually contains `µV` as units when viewed with 8859-1 encoding (previously I just opened it with ASCII). I agree that adding `µV` to our supported units is a good idea. Actually, I would add both the Greek mu and the micro symbol just to be on the safe side (https://en.wikipedia.org/wiki/Micro-#Symbol_encoding_in_character_sets).

Also, if you want to determine if a given string is equal to any one out of several values, you could do it like this:

```
elif unit in ('uV', 'µV'):
```

Sumalyo:
@cbrnr yes,sure I just added a quick fix (in a hurry) with the `or` statement, I will code it with `in` and also add the greek symbol for microvolt. So should I do away with the warning (for the CI checks to pass) or should I keep that?

cbrnr:
We could actually make this really simple now. If you add `µV` to the list of supported units, we don't need to issue a warning because this fixes the issue with the sample file. Of course it would be possible to be extra careful and warn if any unsupported units are encountered (when `misc` is not set), but IMO this complicates the loading process (e.g. we'd have to adapt some of our own tests to use the `misc` parameter).

So if this is also OK with @larsoner I'd really like to keep it simple (meaning that all the `misc` business can be removed again and the fix is basically changing one line of code related to microvolts).

agramfort:
+1 to make it in the end

elif unit in ('uV', 'µV'):

mne/io/edf/edf.py
```diff
@@ -324,7 +324,7 @@ def _read_segment_file(data, idx, fi, start, stop, raw_extras, filenames,
     return tal_data
 
 
-def _read_header(fname, exclude):
+def _read_header(fname, exclude,misc=[]):
```
hoechenberger, line 327: You [shouldn't](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) use a mutable object as a default argument. Either make it `None`, or an empty tuple.

mne/io/edf/edf.py
```diff
@@ -334,6 +334,10 @@ def _read_header(fname, exclude):
     exclude : list of str
         Channel names to exclude. This can help when reading data with
         different sampling rates to avoid unnecessary resampling.
+    misc : list or tuple
```
hoechenberger, line 337: list and tuple should be treated equivalently ([Duck typing](https://en.wikipedia.org/wiki/Duck_typing)), so simply listing one of them here should suffice.

mne/io/edf/edf.py
```diff
@@ -342,7 +346,7 @@ def _read_header(fname, exclude):
     ext = os.path.splitext(fname)[1][1:].lower()
     logger.info('%s file detected' % ext.upper())
     if ext in ('bdf', 'edf'):
-        return _read_edf_header(fname, exclude)
+        return _read_edf_header(fname, exclude,misc)
```
hoechenberger, line 349:
```suggestion
        return _read_edf_header(fname, exclude, misc)
```

mne/io/edf/edf.py
```diff
@@ -355,7 +359,7 @@ def _get_info(fname, stim_channel, eog, misc, exclude, preload):
     eog = eog if eog is not None else []
     misc = misc if misc is not None else []
 
-    edf_info, orig_units = _read_header(fname, exclude)
+    edf_info, orig_units = _read_header(fname, exclude,misc)
```
hoechenberger, line 362:
```suggestion
    edf_info, orig_units = _read_header(fname, exclude, misc)
```

mne/io/edf/edf.py
```diff
@@ -520,7 +524,7 @@ def _edf_str(x):
     return x.decode('latin-1').split('\x00')[0]
 
 
-def _read_edf_header(fname, exclude):
+def _read_edf_header(fname, exclude,misc=[]):
```
hoechenberger, line 527: Same here: don't use mutable objects as default arguments

hoechenberger (changes requested):

mne/io/edf/edf.py
Sumalyo: @hoechenberger as of now the issue stands just to support `µV ` as a valid unit for microvolt to resolve the scaling problem referred to in #9186 , and I plan to do away with the `misc` implementation entirely as it makes the loading process complicated (and also fails a lot of CI tests which pass data with other without `misc` . However, if `misc` is to be implented finally I will make the requested changes in code (implementation, style and documentation) . Currently waiting for @larsoner s opinion, to resolve #9186 if only adding support for an extra symbol would do the job, and the `misc` parts can be ignored entirely.

cbrnr: @agramfort already gave his OK, so you could already implement/revert to the much simpler case. @larsoner can then directly merge if he's happy when he comes online.

cbrnr: And don't forget to add your contribution to the changelog!

Sumalyo:
@cbrnr I guess https://github.com/mne-tools/mne-python/pull/9187/commits/34e736171e3ffe047316bfdfb09fff342b365598 seems to pass all tests, this would fix the issue. I have added support for greek mu and legacy micro symbols checking with `if` and  `in`. @hoechenberger I guess now everything seems fine. I seem to have some issue when updating the changelog to include this PR, please suggest a way to fix that conflict. Thank you

cbrnr:
@Sumalyo you need to rebase your branch onto the latest `main`, resolve any conflicts, and then force-push:

```
git fetch upstream
git rebase upstream/main
git push --force
```

hoechenberger:
Or, easier (IMHO):

Simply pull upstream/main, resolve conflicts, commit, and push (without force)

larsoner:
> So if this is also OK with @larsoner I'd really like to keep it simple (meaning that all the misc business can be removed again and the fix is basically changing one line of code related to microvolts).

My sense is that a warning if the units of an EEG channel are unknown could still be useful, but I'm willing to assume YAGNI and take care of it in another PR if someone actually hits a units problem!

doc/changes/latest.inc
```diff
@@ -157,6 +159,8 @@ Enhancements
 
 Bugs
 ~~~~
+- Fix bug with :func:`mne.io.read_raw_edf` where new symbols for interpreting data in microvolt was added (:gh:`9187` **by new contributor** |Sumalyo Datta|_ )
```
cbrnr, line 162:
```suggestion
- Fix bug with :func:`mne.io.read_raw_edf` where µV was not correctly recognized (:gh:`9187` **by new contributor** |Sumalyo Datta|_)
```

Sumalyo: Thanks for the suggestion, I have committed the changes

hoechenberger (approved):

mne/io/edf/edf.py
```diff
@@ -613,7 +613,7 @@ def _read_edf_header(fname, exclude):
         for i, unit in enumerate(units):
             if i in exclude:
                 continue
-            if unit == 'uV':
+            if unit in ('uV', 'µV', 'µV'):
```
agramfort, line 616: why 2 times µV ?

besides LGTM

Sumalyo: as @cbrnr suggested that there are two special symbols for microvolt as per the wiki page https://en.wikipedia.org/wiki/Micro-#Symbol_encoding_in_character_sets the greek mu symbol and another legacy micro symbol used to denote the same. Just as a precaution I have implemented both for units to denote data in microvolts.

agramfort (approved):

agramfort:
thx @Sumalyo !

drammock:
merged too fast @agramfort the doubled µ is a mistake

mne/io/edf/edf.py
drammock: they're supposed to be two different codepoints (μ `03BC` greek small letter mu; µ `00B5` micro sign) but copy/paste from the GH website suggests they're the same (both micro sign).  I suggest either:

1. make sure they're actually different codepoints and add a comment to this line explaining why it looks redundant, or
2. change them to `'\u03BCV'` and `\u00B5V'` and add a comment explaining why using codepoints

Sumalyo: @drammock I think something like this  `if unit in ('\u03BCV','\u00B5V','uV'):` might work then, what should I do now to make this change

drammock: you'll need to make a new PR based off of current `main`. Pushing more commits to this branch won't work because it's already merged.

Sumalyo:
@hoechenberger @drammock @cbrnr I have opened another PR https://github.com/mne-tools/mne-python/pull/9196 to resolve this, extremely sorry for the mess. Please let me know if it's okay now. Thank you.

Issue #9177, opened 2021-03-21
Jupyter Notebook representations of MNE objects


vagechirkov:
Add HTML rendering of the MNE objects using `_repr_html_` method.

TODO list:

- [x] `Info` #6846 #9161
        
    <details>

    ![image](https://user-images.githubusercontent.com/42982039/111919350-d7c38600-8a89-11eb-8198-02c228652748.png)

    </details>

- [x] `BaseRaw` #9161 #9172 

    <details>

    ![image](https://user-images.githubusercontent.com/42982039/111919303-9d59e900-8a89-11eb-80d9-2730cdd69e03.png)

    </details>

- [x] `Epochs` #9189 

    <details>

    ![image](https://user-images.githubusercontent.com/42982039/112514530-b4545000-8d95-11eb-90de-f4c6892a126f.png)

    </details>

- [ ] `Evoked`
- [ ] `SourceEstimate`

hoechenberger:
I took the liberty to add `SourceEstimate` to the list, too :)

mmagnuski:
This is cool! :)

PR #9189, opened 2021-03-23
MRG, ENH: add _repr_html_ method in the class BaseEpochs

vagechirkov:
Reference issue #9177

Add `_repr_html_` for BaseEpochs:

image

(updated image)

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -55,6 +55,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :meth:`mne.io.base.BaseRaw._repr_html_` to make HTML based representation of the Epochs object in IPython Notebook (:gh:`9174` by |Valerii Chirkov|_)
+
 - Speed up :func:`mne.viz.plot_ica_properties` by refactoring (:gh:`9174` **by new contributor** |Valerii Chirkov|_)
 
 - Add ``apply_function`` method to epochs and evoked objects (:gh:`9088` **by new contributor** |Erica Peterson|_ and `Victoria Peterson`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -55,7 +55,7 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
-- Add :meth:`mne.io.base.BaseRaw._repr_html_` to make HTML based representation of the Epochs object in IPython Notebook (:gh:`9174` by |Valerii Chirkov|_)
+- Add HTML representation for `~mne.Epochs` in Jupyter Notebooks (:gh:`9174` by |Valerii Chirkov|_)
 
 - Speed up :func:`mne.viz.plot_ica_properties` by refactoring (:gh:`9174` **by new contributor** |Valerii Chirkov|_)
 
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -75,3 +75,24 @@ raw_template = Template("""
     </tr>
 </table>
 """)
+
+epochs_template = Template("""
+<table class="table table-hover">
+    <tr>
+        <th>N events</th>
+        <td>{{'%s events' % len(epochs.events)}}</td>
+    </tr>
+    <tr>
+        <th>Time range</th>
+        <td>{{'%g: %g sec' % (epochs.tmin, epochs.tmax)}}</td>
+    </tr>
+    <tr>
+        <th>Baseline</th>
+        <td>{{baseline}}</td>
+    </tr>
+    <tr>
+        <th>Events</th>
+        <td>{{events}}</td>
+    </tr>
+</table>
+""")
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -92,7 +92,7 @@ epochs_template = Template("""
     </tr>
     <tr>
         <th>Time range</th>
-        <td>{{f'{epochs.tmin:.3f} – {epochs.tmax:.3f} sec',}}</td>
+        <td>{{f'{epochs.tmin:.3f} – {epochs.tmax:.3f} sec'}}</td>
     </tr>
     <tr>
         <th>Baseline</th>
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -83,16 +83,20 @@ epochs_template = Template("""
         <td>{{'%s events' % len(epochs.events)}}</td>
     </tr>
     <tr>
+        <th>Events</th>
+        {{if events is not None}}
+        <td>{{events}}</td>
+        {{else}}
+        <td>Not available</td>
+        {{endif}}
+    </tr>
+    <tr>
         <th>Time range</th>
-        <td>{{'%g: %g sec' % (epochs.tmin, epochs.tmax)}}</td>
+        <td>{{'%.3f: %.3f sec' % (epochs.tmin, epochs.tmax)}}</td>
     </tr>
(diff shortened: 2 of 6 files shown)


mne/epochs.py
```diff
@@ -1599,6 +1600,25 @@ def __repr__(self):
         class_name = 'Epochs' if class_name == 'BaseEpochs' else class_name
         return '<%s | %s>' % (class_name, s)
 
+    def _repr_html_(self):
+        if self.baseline is None:
+            baseline = 'off'
+        else:
+            baseline = tuple(['None' if b is None else ('%g' % b)
+                              for b in self.baseline])
+            baseline = '%s: %s sec' % baseline
```
agramfort, line 1609:
```suggestion
            baseline = '%.3f: %.3f sec' % baseline
```

mne/epochs.py
```diff
@@ -1599,6 +1600,25 @@ def __repr__(self):
...
+            if self.baseline != _check_baseline(
+                    self.baseline, times=self.times, sfreq=self.info['sfreq'],
+                    on_baseline_outside_data='adjust'):
+                baseline = ' (baseline period was cropped after baseline\
+                    correction)'
```
agramfort, line 1614: not sure it's necessary to print if you display also the tmin and tmax

mne/epochs.py
```diff
@@ -1599,6 +1600,25 @@ def __repr__(self):
...
+        events = ''
+        for k, v in sorted(self.event_id.items()):
```
agramfort, line 1616: you need to be careful here as 

`event_id : int | list of int | dict | None`

agramfort (commented):
the report should also be updated accordingly to use the template for -epo.fif files.

thx @vagechirkov

mne/epochs.py
vagechirkov: Alright! Should I also update [this part](https://github.com/mne-tools/mne-python/blob/e430158cee998776551efffbb343e4e42a45ae37/mne/epochs.py#L1594) of the `__repr__` method?

agramfort:
diff looks good

can you touch the plot_report file and maybe another one so we see in the doc artifact how it looks?

also don't forget to update what's new.

thx @vagechirkov

mne/data/html_templates.py
```diff
@@ -75,3 +75,28 @@
...
+        <th>Events</th>
+        {{if events is not None}}
+        <td>{{events}}</td>
+        {{else}}
+        <td>Not available</td>
+        {{endif}}
+    </tr>
+    <tr>
+        <th>Time range</th>
+        <td>{{'%.3f: %.3f sec' % (epochs.tmin, epochs.tmax)}}</td>
```
hoechenberger, line 95:
```suggestion
        <td>{{f'{epochs.tmin:.3f} – {epochs.tmax:.3f} sec',}}</td>
```

mne/data/html_templates.py
```diff
@@ -75,3 +75,28 @@
     </tr>
 </table>
 """)
+
+epochs_template = Template("""
+<table class="table table-hover">
+    <tr>
+        <th>N events</th>
+        <td>{{'%s events' % len(epochs.events)}}</td>
```
hoechenberger, line 83:
```suggestion
        <th>Number of events</th>
        <td>{{len(epochs.events)}}</td>
```

mne/epochs.py
```diff
@@ -1599,6 +1600,29 @@ def __repr__(self):
         class_name = 'Epochs' if class_name == 'BaseEpochs' else class_name
         return '<%s | %s>' % (class_name, s)
 
+    def _repr_html_(self):
+        if self.baseline is None:
+            baseline = 'off'
+        else:
+            baseline = tuple(['None' if b is None else ('%.3f' % b)
+                              for b in self.baseline])
```
hoechenberger, line 1608: Actually, we shouldn't have ``None`` values in `Epochs.baseline` anymore these days, as we're replacing them with the actual times. Did you come across a situation where this was not the case? Because then it's a bug we need to fix.

In any case: you can (should) drop special-casing `None` here

mne/epochs.py
```diff
@@ -1599,6 +1600,29 @@ def __repr__(self):
...
+            baseline = '%s: %s sec' % baseline
```
hoechenberger, line 1609:
```suggestion
            baseline = f'{baseline[0]} – {baseline[1]} sec'
```

mne/epochs.py
```diff
@@ -1599,6 +1600,29 @@ def __repr__(self):
...
+            for k, v in sorted(self.event_id.items()):
+                events += '%s: %i<br>' % (k, sum(self.events[:, 2] == v))
+        elif isinstance(self.event_id, list):
+            events = ''
+            for k in self.event_id:
+                events += '%s: %i<br>' % (k, sum(self.events[:, 2] == k))
+        elif isinstance(self.event_id, int):
+            events = '%i: %i' % (self.event_id, len(self.events[:, 2]))
+        else:
+            events = None
```
hoechenberger, line 1622: Do you think you could

- assign `sum(self.events[:, 2] == v)`, `sum(self.events[:, 2] == k)`, `len(self.events[:, 2])` to a variable
- use this variable and `k` in an f-string, i.e. `events = f'{k}: {num_of_events}'`
?

hoechenberger (changes requested):
Thanks for your work on this! I left a few comments; looks good to me generally; I'd like to encourage you to make use of f-strings; also special-casing `None` in `Epochs.baseline` should not be necessary.

mne/data/html_templates.py
```diff
@@ -75,3 +75,28 @@
...
+        <th>Events</th>
+        {{if events is not None}}
+        <td>{{events}}</td>
+        {{else}}
+        <td>Not available</td>
+        {{endif}}
+    </tr>
+    <tr>
+        <th>Time range</th>
+        <td>{{f'{epochs.tmin:.3f} – {epochs.tmax:.3f} sec',}}</td>
```
hoechenberger, line 95:
```suggestion
        <td>{{f'{epochs.tmin:.3f} – {epochs.tmax:.3f} sec'}}</td>
```

mne/epochs.py
vagechirkov: Alright, I will update this part! I have no examples where the baseline is `None`.

mne/epochs.py
vagechirkov: Alright!

vagechirkov:
> diff looks good
> 
> can you touch the plot_report file and maybe another one so we see in the doc artifact how it looks?
> 
> also don't forget to update what's new.
> 
> thx @vagechirkov

As far as I understood there are no epochs files in the example dataset. Here is the output of the report after I created one:

image

doc/changes/latest.inc
```diff
@@ -55,6 +55,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add :meth:`mne.io.base.BaseRaw._repr_html_` to make HTML based representation of the Epochs object in IPython Notebook (:gh:`9174` by |Valerii Chirkov|_)
```
hoechenberger, line 58:
```suggestion
- Add nice HTML based representation for `~mne.Epochs` in Jupyter Notebooks (:gh:`9174` by |Valerii Chirkov|_)
```

hoechenberger (commented):
Looks very clean! Just one small suggestion.

hoechenberger (approved):
All good for me once CI goes green!

hoechenberger:
Thanks, @vagechirkov!

Issue #8060, opened 2020-07-26
ENH: Add support for writing other formats


larsoner:
@rob-luke wants to add support for writing SNIRF files. We could support this internally easily via file-extension `*.snirf` triage. WDYT @agramfort ?

There are also some potentially low-hanging fruit in terms of supporting writing to other formats using external libraries, at least BrainVision (pybv) EDF (pyEDF, maybe edflib?) and CTF (pyCTF). I don't see too much work in supporting these since in theory we "just" hook up our structures to the right places to their writers, then rely on them to write the correct data. I don't see this as too much additional maintenance burden for us since most of the heavy lifting will be done by those other libraries.

Thoughts?

agramfort:
I don't know these external packages: if they are released regularly on
pypi, if they are unit tested / maintained, if license is ok etc.
I just fear that users will complain to us if these packages break.

can we first try with a write_raw_fnirs in mne-nirs and maybe reconsider in
a few months after seeing how smoothly this goes?

it's just once it's in MNE it tends to stay for a long time and we'll have
to maintain it.

>

rob-luke:
> can we first try with a write_raw_fnirs in mne-nirs and maybe reconsider in a few months after seeing how smoothly this goes?

Sure I will put this in mne-nirs

sappelhoff:
Just fyi: I maintain **[pybv](https://github.com/bids-standard/pybv/)** with help from @choldgraf, and @cbrnr (at least he made a release and a few PRs :slightly_smiling_face: ) ... @hoechenberger has made the corresponding conda-forge package work, and apart from that there is one more non-MNE contributor :-)

We have a BSD-3-Clause License, a pypi and conda forge package, and unit tests. However, we do not regularly release, because the writer currently already supports all features that the contributor and maintainer team need.

Note: some parts of the code in pybv stem from @palday's "philistine" package. As far as I can see it, @palday also maintains [pyedf](https://github.com/bids-standard/pyedf/)

Yet with regards to EDF writing, the development looks more active in [pyedflib](https://github.com/holgern/pyedflib) ... with the drawback that that writer is not pure python, it depends on the [edflib C library](https://gitlab.com/Teuniz/EDFlib)

cbrnr:
[...]

PR #9192, opened 2021-03-24
Add functionality to export to EEGLAB .set

jackz314:
#### Reference issue
N/A

This adds the ability to export Raw or Epochs instances as EEGLAB set files. As well as the ability to import channel location data from separate CSV files or another file (using `read_raw`).

#### Additional information
The exported to set function (`save_set`) was based on mnelab's [`write_set`](https://github.com/cbrnr/mnelab/blob/4bcee31bac8677fe880edfb35396428e6100e950/mnelab/io/writers.py#L19). I added channel location export and events/epochs data export so that Epochs exports can be read by EEGLAB.

The channel location expansion function (xyz to all, used for `save_set`) referenced MATLAB's [cart2sph](https://www.mathworks.com/help/matlab/ref/cart2sph.html), as well as EEGLAB's [relevant functions](https://github.com/sccn/eeglab/blob/develop/functions/sigprocfunc/convertlocs.m).

It has been roughly tested on a few data sets on both EEGLAB and MNE.


jackz314:
Seems like the tests are failing due to the lack of docstrings. Should I add them?

mmagnuski:
Welcome on your first mne PR, @jackz314!
Thanks for contributing, beinging able to save files in .set format would be a nice addition (at least from my perspective).

> Seems like the tests are failing due to the lack of docstrings. Should I add them?

Yes, we follow the numpy docstring format. Take a look around in the codebase - you will see the pattern. The docstrings are required for all public functions.
Adding a functionality like that also requires writing tests that cover > 90% of the added code.
Also - make sure you follow the PEP8 code style. 
You will find [more guidelines here](https://mne.tools/stable/install/contributing.html#mne-python-coding-conventions).

agramfort:
thx @jackz314 for the work

we should discuss how to approach this. So far we have been reluctant to support writing for something different from .fif as it's very likely that the save to something else will not persist to disk all the informations in the mne instance.

Also you add a lot of public functions to mne and we are support careful about this. Should i had started this PR I would have tried to support writing without any new public function: just use .save method and detect the file extension to know what to do.

hoechenberger:
Nice job, @jackz314!
Like @agramfort said, ideally all functions except for `save_set()` should be private.

Also I'm wondering if instead of adding `save_set()`, we should consider adding a parameter to `save()`, allowing to save in EEGLAB format. Or it could even auto-pick "EEGLAB .set" if the output file extension is `.set`.
But this really needs a bit of discussion (and, most certainly, much more testing…)

jackz314:
Thanks guys. I'm currently writing tests on this, but my expertise in both EEGLAB and MNE is limited, so I would probably need help from other people to test this more thoroughly - As I'm writing these tests, I'm already discovering multiple bugs in my code. I agree maybe merging `save_set` with `save` would be better. 

The only two user-facing functions would probably be `save_set` and `import_eeg_chan_location_from_csv`. `get_eeglab_full_cords` and `cart_to_eeglab_full_coords` doesn't seem to be useful outside of conversion between eeglab and mne, so I can change them to private.

As for information loss, unfortunately, it seems like there will be some unavoidable info loss, so far from my limited testing, things like re-reference, baseline, and high/low pass freq info are lost when converting between the two. For re-referencing and hi/lo pass, the modified data is preserved, but in EEGLAB there's no way of knowing what kind of re-referencing is done, and when converting back to MNE, there's no way of getting the filtering parameters.

Info lost is definitely not ideal, but in a lot of cases it's tolerable (as long as data is intact and as expected), so I think maybe some warning could be given to users who try to export as set files, that way they are aware of the info lost, but can still benefit from the conversion. I'm currently working with a lab where EEGLAB is heavily used, but they also want to use Python as well, since some things like ML is easier on Python, this is why I wrote these functions, and I think other people in similar situations would benefit from this as well.

hoechenberger:
> import_eeg_chan_location_from_csv

This one should probably also be integrated into [`read_custom_montage()`](https://mne.tools/stable/generated/mne.channels.read_custom_montage.html)

jackz314:
> > import_eeg_chan_location_from_csv
> 
> This one should probably also be integrated into [`read_custom_montage()`](https://mne.tools/stable/generated/mne.channels.read_custom_montage.html)

I just realized that channel location data in MNE is called Montage, I will integrate it into `read_custom_montage()`, but it should probably be a separate PR.

hoechenberger:
> I will integrate it into `read_custom_montage()`, but it should probably be a separate PR.

This actually sounds like a good idea!!

jackz314:
I also would like to know where I should put the tests for these functions. Currently, I put the tests for Epochs' save set to `tests/test_epochs.py`, and I plan on putting the tests for Raw's save set to `io/tests/test_raw.py`, would this be appropriate? Or should I put both of them into `io/eeglab/tests/test_eeglab.py` since they are related to EEGLAB?

jackz314:
I noticed that channel location data is actually stored in the format of -y, x, z instead of x, y, z. Is there a particular reason for this? Additionally, is this a standard that applies to all location data or specific to some formats like EEGLAB set files?

hoechenberger:
Hello @jackz314,

I briefly talked with @agramfort about this effort and we had the feeling that adding support for writing to new formats could end up as opening Pandora's box: it would be very challenging to ensure that no data is lost (or even to quantify as to **which** data is lost, specifically) during an I/O roundtrip.

We felt that maybe adding this functionality would be better suited in an external package, which could very well be maintained under the MNE umbrella (`mne-tools` GitHub organization), but could be released separately from MNE, which would make a lot of sense as this would allow to push out bugfixes on an independent schedule.

Another idea I had and which is kind of contradictory to what I said above is adding a new method for exports that possibly lead to data loss. This is essentially what GIMP does: "save/save as" saves to their native format; "export" converts the data, possibly losing information. So I would vote for calling a method that exports to non-FIFF files `export()`.

I'd love to hear your and the other developers' thoughts!

jackz314:
Personally, I think this feature would be helpful even if there is some data loss, like I mentioned above, I think it would be better to provide warnings to users who use this function (and maybe on relevant docs as well) about potential data loss than to not provide the feature at all. As you said, I think using a separate and distinct name like `export` instead of `save` or `save_set` could further make sure that the user knows what they are doing.

Some data will be lost unfortunately due to some features being implemented very differently in EEGLAB and MNE (e.g., reference channel/method, filtering configuration, etc.). However, since the core data and some other fields like events/trials/epochs/annotations (the events "family") and channel info can mostly be exchanged without data loss, exporting will still be helpful even if other fields might not survive the i/o cycle. 

I'm leaning more towards your second idea, because I feel like since other functions that read from EEGLAB are already maintained in the main MNE package, any changes to EEGLAB that requires any updates to MNE's I/O should occur naturally together. I also think that maintaining a separate package just for exporting to EEGLAB seems a bit overkill and shouldn't be necessary, especially since reading from EEGLAB is already in MNE. Still, I'm also fine with starting a new package, if the consensus later agrees that this feature should be in a separate package, I'd be happy to port my code over.

hoechenberger:
Pinging @larsoner, @jasmainak, @adam2392, and @cbrnr

agramfort:
i hear you Jack but it's a non-trivial maintenance cost and I anticipate a
lot of complaints because some data was lost during
IO round trip and user "did not read the warning".

it also means a lot of CI time for testing as suddenly we need to duplicate
many IO tests to guarantee nothing breaks
when people save to .set files.

adam2392:
Don't we have a "pybv" package for brainvision? Perhaps we have: 

Pyedf (I think there's pyedflib but writing was non trivial)

Pyeeglab?

Maybe "raw.export" just calls these underlying packages? Unit testing should primarily be in those sub packages to address @agramfort objection?

agramfort:
+1 for a pyeeglab package now regarding the export I am not convinced

hoechenberger:
@adam2392 
> Maybe "raw.export" just calls these underlying packages? Unit testing should primarily be in those sub packages to address @agramfort objection?

Yeah that sounds like a possibility. `Raw.export()` would just be a few lines then, passing `self` to one of those packages and emitting a warning – that's about it.

> the only reason I even use eeglab is cuz their ICA is more advanced then mne.

It's off-topic alright, but could you elaborate? Do you mean the interactive exploration of components..?

adam2392:
They have this nice feature which plugs into a database of features I think that can automatically estimate ICA labels in case you have 100s of subjects and can't feasibly run manual ICA, they actually have a cool feature that will match it against known labels to try to estimate if it's eye blink, heart beat, etc. 

Sorry off topic yeah :p.

hoechenberger:
> They have this nice feature which plugs into a database of features I think that can automatically estimate ICA labels in case you have 100s of subjects and can't feasibly run manual ICA, they actually have a cool feature that will match it against known labels to try to estimate if it's eye blink, heart beat, etc.
> 
> Sorry off topic yeah :p.

Ah, that one. Yes we should add this to MNE too. Actually shouldn't be **too** difficult, I believe!

drammock:
we already have ICA components via template matching: https://mne.tools/dev/auto_tutorials/preprocessing/plot_40_artifact_correction_ica.html#selecting-ica-components-using-template-matching

Is the missing piece then that in EEGLAB there's a built-in library of templates, whereas here you have to provide the template from one of your subjects?

cbrnr:
I agree with @agramfort that people will expect full IO roundtrip, so we definitely should treat FIFF differently as our main format. We could of course add an `export` method, which could rely on the presence of external packages. For example, I'm using `pyEDFlib` and `pybv` in MNELAB.

Ideally, we could create a package dedicated to IO, which would contain readers and writers for various formats. Then we could even outsource existing readers, and people could use that package if they just want to load data but not process them with MNE.

jackz314:
I think maintaining separate packages but include the features in MNE sounds like a good idea that would be beneficial to both users and maintainers. I also agree with @cbrnr that ideally, we could have separate packages that do IO for formats outside of MNE. 

Although, I can also see the interoperability between these packages and MNE potentially becoming a problem (e.g., MNE changed some data structure, but this isn't updated in the separate packages, etc.)

I guess for the next step, we should discuss how exactly will this package set up look like?

agramfort:
technically eeglab, fieldtrip and brainstorm can load MNE .fif files

we maintain mne-matlab https://github.com/mne-tools/mne-matlab
for this. So I would argue that we have the responsibility to be able
the read the files from other toolboxes and the work is shared.

my 2c

>

drammock:
> I can also see the interoperability between these packages and MNE potentially becoming a problem

I think that if we write our tests well, this will not be a serious concern, we'll catch most problems before they end up in a released version

agramfort:
well it means you need to be an expert with eeglab file format and we will
not be able to test
the matlab readers on our CI. A full test would be to write with MNE, read
with eeglab, save back with eeglab
and read again with MNE to check it's all good

wdyt?

>

drammock:
Couldn't the CI of the new I/O package test this? Is the issue how to run MATLAB and eeglab in a CI?

agramfort:
I am just anticipating issues but maybe it's doable... if someone has
the energy to lead
this go for it. It's useful for the community for sure.

cbrnr:
I think it would be a very useful package, but it will require quite a lot of energy and effort to get it flying. I also agree with @agramfort that it will be a considerable maintenance effort, but we already have that with our various readers, so outsourcing them would not really create more work once the package exists.

hoechenberger:
> I also agree with @agramfort that it will be a considerable maintenance effort, but we already have that with our various readers, so outsourcing them would not really create more work once the package exists.

… and on the plus side, that I/O package could be released on a different schedule than MNE in order to add new features or fix bugs.

jackz314:
So where should this new IO package (pyeeglab?) be hosted and maintained? I also agree that testing for the EEGLAB side of things would be quite difficult, I myself find it more difficult/tedious to do even simple tests across MNE and EEGLAB, compared to just testing on MNE.

agramfort:
>
> So where should this new IO package (pyeeglab?) be hosted and maintained?
>

maybe we start on your github?


> I also agree that testing for the EEGLAB side of things would be quite
> difficult, I myself find it more difficult/tedious to do even simple tests
> across MNE and EEGLAB, compared to just testing on MNE.
>

yes I can imagine this

mmagnuski:
I also agree that it is a good idea to place the .set writer outside of mne.
BTW it is possible to run matlab tests on Circle or Travis CI (among others) but I've never tested that myself.
Here are some resource that might be useful:
https://www.mathworks.com/help/matlab/matlab_prog/continuous-integration-with-matlab-on-ci-platforms.html
https://circleci.com/developer/orbs/orb/mathworks/matlab
https://docs.travis-ci.com/user/languages/matlab/

jackz314:
It seems the name [`pyeeglab`](https://pypi.org/project/PyEEGLab/) is taken :( Any name suggestions? I'm considering something like `py2eeglab`.

agramfort:
I would favor

eeglab2py

meaning : gowing from eeglab to python

>

jackz314:
> I would favor eeglab2py meaning : going from eeglab to python

But shouldn't export be going from python to EEGLAB? And now I feel like either name wouldn't be suitable if in the future we decide to also include reader functions in that package (py<->eeglab).

hoechenberger:
How about `eeglabio`? 

```python
import eeglabio


eeglabio.read_epochs(...)
eeglabio.save(...)
```

jackz314:
> How about `eeglabio`?

That sounds good, I think I'll use this one unless there are more conflicts.

mmagnuski:
`eeglabio` sounds Italian :)
Assuming the main point of the package would be to export from mne to set files (because the other way around is already present in mne [unless we want to move that code to the new package]) - we could use `mne2set` or `mne2eeglab`.

jackz314:
I think `eeglabio` is fine, I'm using it instead of the other more specific ones to make sure future compatibilities are easier.

jackz314:
I started the `eeglabio` package at [https://github.com/jackz314/eeglabio](https://github.com/jackz314/eeglabio), it's also published on [testpypi](https://test.pypi.org/project/eeglabio). The code in this PR is also updated to use that package. Let me know what you guys think.

larsoner (changes requested):
Requesting changes just to move discussion to #8060, and then update here as necessary

larsoner:
(also, thanks for being patient while we sort out the larger questions surrounding export support in MNE!)

mne/epochs.py
```diff
@@ -1810,6 +1810,56 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+
+        if fmt == "set":
+            from eeglabio.epochs import export_set
+            # load data first
+            self.load_data()
+
+            # remove extra epoc and STI channels
+            chs_drop = \
+                [ch for ch in ['epoc', 'STI 014'] if ch in self.ch_names]
+            self.drop_channels(chs_drop)
```
cbrnr, line 1845: Please don't change the object in-place. You could either make a copy, or even better pick the channels you want in `self.get_data()` a few lines down.

jackz314: Fixed.

mne/epochs.py
```diff
@@ -1840,12 +1840,12 @@ def export(self, fname, fmt="auto"):
             self.load_data()
 
             # remove extra epoc and STI channels
-            chs_drop = \
-                [ch for ch in ['epoc', 'STI 014'] if ch in self.ch_names]
-            self.drop_channels(chs_drop)
+            drop_chs = {'epoc', 'STI 014'}
```
cbrnr, line 1843: You're never adding anything to `drop_chs` so I think it would be better if this was just a list instead of a set. Same thing in `mne/io/base.py`.

jackz314: I used a set so that maybe lookups will be faster, but you're probably right, with only two items right now, lists are probably better.

mne/epochs.py
```diff
@@ -1810,6 +1810,56 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+    def export(self, fname, fmt="auto"):
+        """Export Epochs to EEGLAB's .set format.
+
+        Parameters
+        ----------
+        fname : str
+            Name of the export file.
+        fmt : str
+            Format of the export. Defaults to auto, which will infer
+            format from fname's extension
```
drammock, line 1822: thinking ahead to brainvision exporting:  doesn't brainvision write *three* files (.eeg, .vmrk, .vhdr)?  How will this API accommodate that?  I'd rather get it right now than have to change it later.

mne/epochs.py
```diff
@@ -1810,6 +1810,56 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+
+        Notes
+        -----
+        Channel locations are expanded to the full EEGLAB format
+        For more details see eeglabio.utils.cart_to_eeglab
```
drammock, line 1827:
```suggestion
        Channel locations are expanded to the full EEGLAB format.
        For more details see ``eeglabio.utils.cart_to_eeglab``.
```

Would be ideal if we could cross-link to eeglabio docs via intersphinx; do you know how to set that up @jackz314?

mne/epochs.py
```diff
@@ -1810,6 +1810,56 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+                # swap x (1) and y (0)
+                cart_coords[:, [0, 1]] = cart_coords[:, [1, 0]]
+            else:
+                cart_coords = None
+
+            export_set(fname, self.get_data(picks=pick_chs),
+                       self.info['sfreq'], self.events, self.tmin, self.tmax,
+                       pick_chs, self.event_id, cart_coords)
+        else:
+            raise ValueError("Format not supported (%s)" % fmt)
```
drammock, line 1861: maybe a `NotImplementedError` is more appropriate here?  I can also see the logic of making a distinction between known formats not yet supported (`NotImplementedError`) and unrecognized formats (`ValueError`).

drammock (commented):
comments about `BaseEpochs.export()` docstring also apply to `BaseRaw.export()` docstring.  One other comment is to consider using our docdict since the param descriptions will be similar/identical between the two instance methods.

mne/epochs.py
```diff
@@ -1810,6 +1810,56 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+        if fmt == "auto":
+            fmt = op.splitext(fname)[1]
+            if fmt:
+                fmt = fmt[1:]
+            else:
+                raise ValueError("Couldn't infer format from filename "
+                                 "(no extension found)")
+
+        if fmt == "set":
+            from eeglabio.epochs import export_set
```
drammock, line 1838: one more comment: see how we do it with `_check_pandas_installed`, we should do something similar here.

mne/epochs.py
cbrnr: I'd accept all three extensions for BrainVision.

mne/epochs.py
jackz314: I haven't done it before, I can look into it but some help would be appreciated.

mne/epochs.py
jackz314: I think `NotImplementedError` would be appropriate for other to be implemented cases in `elif` statements, but `ValueError` in `else` will catch the unknown formats. 

Perhaps it's appropriate to have something similar to [`read_custom_montage`](https://github.com/mne-tools/mne-python/blob/176680a238a7c4f132fdd9198fc7bbe7797d764f/mne/channels/montage.py#L1106) considering the BrainVision case you mentioned before.

mne/epochs.py
drammock: Here is the basics: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

For configuring our end, see `doc/conf.py` around line 129 or so

For the eeglabio end, you'll need to create the `objects.inv` file, which means the eeglabio docs need to also be built with sphinx (from the URL above: "Each Sphinx HTML build creates a file named `objects.inv` that contains a mapping from object names to URIs relative to the HTML set’s root"). Once that exists someplace on the internet (i.e., once eeglabio has a website for its docs) you add the package name and URL to our conf.py and it should "just work".

Probably easiest is to try setting up intersphinx in the eeglabio docs `conf.py` file and try linking in those docs to some function within MNE-Python or NumPy.  That should be easier to test/iterate.  Once that works, it should work for us to link to eeglabio docs too.

jackz314: I published the docs at [eeglabio.readthedocs.io](https://eeglabio.readthedocs.io/). Also changed `doc/conf.py`.

jackz314:
FYI, the docs of eeglabio is published at [eeglabio.readthedocs.io](https://eeglabio.readthedocs.io/). `intersphinx` has been setup, although currently it only references the module itself and one util function.

jackz314:
I'm not sure what to do in terms of testing, since exporting is now done in `eeglabio` and it's optional, should it still be tested here? And should I add the requirement to `requirements_testing.txt`?

drammock:
CircleCI errors:
```
WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.BaseEpochs.export':
  GL03: Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
  PR09: Parameter "fmt" description should finish with "."

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.Epochs.export':
  GL03: Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
  PR09: Parameter "fmt" description should finish with "."

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.EpochsArray.export':
  GL03: Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
  PR09: Parameter "fmt" description should finish with "."

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.io.BaseRaw.export':
  PR09: Parameter "fmt" description should finish with "."

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.io.Raw.export':
  PR09: Parameter "fmt" description should finish with "."

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.io.RawArray.export':
  PR09: Parameter "fmt" description should finish with "."
```

drammock:
The intersphinx mapping is working, nice job! https://27261-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Epochs.html?highlight=eeglabio#mne.Epochs.export

jackz314:
I fixed the tests by adding a check for eeglabio. Everything should work now.

cbrnr (approved):
Nice work @jackz314! Make sure to add an entry to what's new!

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+            format. To save in native MNE format (fif) without info loss,
+            use :func:`save` instead.
+
+        Parameters
+        ----------
+        fname : str
+            Name of the export file.
+        fmt : str
+            Format of the export. Defaults to auto, which will infer
+            format from fname's extension.
```
agramfort, line 1831: please specify what values are valid for fmt param.

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+        -----
+        For EEGLAB format, channel locations are expanded to full EEGLAB
+        format. For more details see :func:`eeglabio.utils.cart_to_eeglab`.
+        """
+        if fmt == "auto":
+            fmt = op.splitext(fname)[1]
+            if fmt:
+                fmt = fmt[1:]
+            else:
+                raise ValueError("Couldn't infer format from filename "
```
agramfort, line 1843:
```suggestion
                raise ValueError(f"Couldn't infer format from filename {fname} "
```

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+                # (-y x z) to (x y z)
+                cart_coords[:, 0] = -cart_coords[:, 0]  # -y to y
+                # swap x (1) and y (0)
+                cart_coords[:, [0, 1]] = cart_coords[:, [1, 0]]
+            else:
+                cart_coords = None
+
+            export_set(fname, self.get_data(picks=pick_chs),
+                       self.info['sfreq'], self.events, self.tmin, self.tmax,
+                       pick_chs, self.event_id, cart_coords)
```
agramfort, line 1871: I would name parameters in the function call to explicit the function signature of export_set

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+        elif fmt == "edf":
+            raise NotImplementedError("Export to EDF format not implemented.")
+        else:
+            raise ValueError("Format not supported (%s)" % fmt)
```
agramfort, line 1875:
```suggestion
            raise ValueError(f"Format not supported {fmt}. Supported formats are 'set'.")
```

mne/io/base.py
```diff
@@ -1450,6 +1451,72 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
...
+                cart_coords = None
+
+            annotations = [self.annotations.description,
+                           self.annotations.onset,
+                           self.annotations.duration]
+            export_set(fname, self.get_data(picks=pick_chs),
+                       self.info['sfreq'],
+                       pick_chs, cart_coords, annotations)
+        else:
+            raise ValueError("Format not supported (%s)" % fmt)
```
agramfort, line 1518: too much copy paste @jackz314 I would have to type my comments here too.

can you avoid this with a util private function?

mne/io/tests/test_raw.py
```diff
@@ -690,3 +691,25 @@ def test_get_data_units():
     # not the good type
     with pytest.raises(TypeError, match='instance of None, str, or dict'):
         raw.get_data(units=['fT/cm', 'fT', 'uV'])
+
+
+def test_export_set():
+    """Test saving a Raw instance to EEGLAB's set format."""
+    if not _check_eeglabio_installed(strict=False):
+        return
```
agramfort, line 699: use pytest skip

mne/tests/test_epochs.py
```diff
@@ -3096,6 +3097,32 @@ def test_save_complex_data(tmpdir, preload, is_complex, fmt, rtol):
     assert_allclose(data_read, data, rtol=rtol)
 
 
+@pytest.mark.parametrize('preload', (True, False))
+def test_export_set(tmpdir, preload):
+    """Test saving an Epochs instance to EEGLAB's set format."""
+    if not _check_eeglabio_installed(strict=False):
+        return
```
agramfort, line 3104: use pytest skip

agramfort (approved):
looking good @jackz314 !

mne/epochs.py
hoechenberger, line 1831:
```suggestion
            Format of the export. Defaults to ``'auto'``, which will infer
            the format from the filename extension.
```

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+        ----------
+        fname : str
+            Name of the export file.
+        fmt : str
+            Format of the export. Defaults to auto, which will infer
+            format from fname's extension.
+
+        Notes
+        -----
+        For EEGLAB format, channel locations are expanded to full EEGLAB
```
hoechenberger, line 1835:
```suggestion
        For EEGLAB exports, channel locations are expanded to full EEGLAB
```

mne/epochs.py
```diff
@@ -1810,6 +1811,69 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
+        format. For more details see :func:`eeglabio.utils.cart_to_eeglab`.
+        """
+        if fmt == "auto":
+            fmt = op.splitext(fname)[1]
```
hoechenberger, line 1839: Your'e using double-quotes here and single-quotes in other places. Can you please make this consistent? (My personal preference being single-quotes :))

mne/io/base.py
```diff
@@ -1450,6 +1451,72 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
...
+        fname : str
+            Name of the export file.
+        fmt : str
+            Format of the export. Defaults to auto, which will infer
+            format from fname's extension.
+
+        Notes
+        -----
+        For EEGLAB format, channel locations are expanded to full EEGLAB
+        format. For more details see :func:`eeglabio.utils.cart_to_eeglab`.
```
hoechenberger, line 1476: We have the same docstring in the method. Can you please move it into the `docdict` (in `mne/utils/docs.py`) to remove the redundancy?

jackz314: Since raw and epochs can technically have different export targets in the future, should I just move the warning and eeglab's notes to `docdict`?

jackz314: Also, I looked around but I'm not quite sure how to use them, are there any examples elsewhere in the code that I can copy from?

hoechenberger: I would move the entire thing for now, except for the very first line, which is different.

You add the documentation string to an entry in the `docdict`, and then you can reference it in the function's/method's docstring via `%(the_docdict_key_you_picked)s`

If the function/method has a `@verbose` or `@fill_doc` decorator, the `%(...)s` will be replaced with the corresponding bits from the `docdict` on documentation build time. 

I'd suggest to add the `@verbose` decorator and a `verbose` kwarg to allow users to control the verbosity level; and this then enables the `docdict` support automatically :)

jackz314: What does the verbose kwarg do exactly? Isn't this supposed to be used for logging?

mne/io/base.py
jackz314: moved to io.uitls and utils.check

mne/io/base.py
jackz314: I decided to only move the parts that don't contain supported formats to `docdict` and used `@verbose` (still not entirely sure what it does). Everything else should work now.

drammock: we have 2 decorators, `@verbose` and `@fill_doc`. You need one of them in order for docdict replacements to work.  The `@verbose` decorator additionally adds a `verbose` kwarg, which (as you thought) allows users to control logging.  `verbose` is usually passed down to private/helper function calls, so even if there are no `warn()` or `logger` calls in the top-level function, `verbose` can still affect how much output a user sees.

jackz314:
All suggestions are implemented and it seems like all the errors are fixed. Let me know if it needs anything else.

hoechenberger:
API docs for Raw: https://27339-1301584-gh.circle-artifacts.com/0/dev/generated/mne.io.Raw.html?highlight=export#mne.io.Raw.export
API docs for Epochs: https://27339-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Epochs.html?highlight=export#mne.Epochs.export

mne/utils/docs.py
```diff
@@ -2262,6 +2262,27 @@
...
+.. warning::
+    Since we are exporting to external formats, there's no guarantee that all
+    the info will be preserved in the external format. To save in native MNE
+    format (fif) without info loss, use :func:`save` instead.
+"""
+docdict['export_params_base'] = """
+Parameters
+----------
+fname : str
+    Name of the export file.
```
hoechenberger, line 2276:
```suggestion
    Name of the output file.
```

mne/utils/docs.py
```diff
@@ -2262,6 +2262,27 @@
     fail irrespective of this parameter.
 """
 
+# Export
+docdict['export_warning'] = """
+.. warning::
+    Since we are exporting to external formats, there's no guarantee that all
+    the info will be preserved in the external format. To save in native MNE
+    format (fif) without info loss, use :func:`save` instead.
```
hoechenberger, line 2270:
```suggestion
    format (``.fif``) without information loss, use :func:`save` instead.
```

mne/io/base.py
```diff
@@ -1450,6 +1453,48 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
...
                    split_size, split_naming, 0, None, overwrite)
 
+    @verbose
+    def export(self, fname, fmt='auto', verbose=None):
+        """Export Raw to external formats.
+
+        Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
+        %(export_warning)s
+        %(export_params_base)s
+            Valid options are ``'auto'`` for auto inferred and
```
hoechenberger, line 1463:
```suggestion
            Valid options are ``'auto'`` for automatic inferred and
```

mne/io/base.py
```diff
@@ -1450,6 +1453,48 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
...
+            ``'set'`` for EEGLAB.
```
hoechenberger, line 1464: shouldn't we maybe call the format "eeglab" instead of "set"? WDYT?

mne/io/base.py
```diff
@@ -1450,6 +1453,48 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
                    start, stop, buffer_size, projector, drop_small_buffer,
                    split_size, split_naming, 0, None, overwrite)
 
+    @verbose
+    def export(self, fname, fmt='auto', verbose=None):
+        """Export Raw to external formats.
+
+        Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
+        %(export_warning)s
+        %(export_params_base)s
```
hoechenberger, line 1462: I think it would make sense to include the supported parameters right there. As in,

```
fmt : 'set' | 'abc' | 'def'

**Review by hoechenberger — COMMENTED** [review]
Nice job! Just a few nitpicks

File: mne/io/base.py (hunk 20 above)
Comment (jackz314) [reply mne/io/base.py:1464]: I think eeglab probably will make more sense once there are more formats. I used set just because it was easier to get the format automatically.

File: mne/io/base.py (hunk 21 above)
Comment (jackz314) [reply mne/io/base.py:1462]: I see, but we should also keep the "supported formats" section to include more details, right?

File: mne/io/base.py (hunk 20 above)
Comment (jackz314) [reply mne/io/base.py]: I just updated the supported format list to a dict (e.g. 'eeglab': 'set').

File: mne/utils/check.py [hunk 22]
```diff
@@ -771,3 +784,28 @@ def _ensure_events(events):
         raise ValueError(
             f'events must be of shape (N, 3), got {events.shape}')
     return events
+
+
+def _infer_check_export_fmt(fmt, fname, supported_formats):
```
hoechenberger, line 789: Can you (very briefly) document the parameters here so one knows which data types the function expects?

mne/utils/check.py
```diff
@@ -771,3 +784,28 @@ def _ensure_events(events):
...
+            # find fmt in supported formats dict's tuples
+            fmt = next((k for k, v in supported_formats.items() if fmt in v),
+                       fmt)  # default to original fmt for raising error later
+        else:
+            raise ValueError(f"Couldn't infer format from filename {fname}"
+                             " (no extension found)")
+
+    if fmt not in supported_formats:
+        raise ValueError(f"Format '{fmt}' is not supported. "
+                         f"Supported formats are {supported_formats}.")
```
hoechenberger, line 810: I think we only want to print the keys here, separated by commas, i.e. something like

`', '.join(supported_formats.keys()`

hoechenberger (commented):
Two more nitpicks and then I'm happy 😃

mne/utils/check.py
jackz314: Well someone could attempt to use file extensions to infer automatically, and if the extension is not supported I think printing out the dictionary might be more helpful compared to just the formats.

jackz314: I can also print the dictionary in a better way (e.g. `eeglab (.set), brainvision (.eeg, .vmrk, ...)`)

hoechenberger: Great idea!

mne/utils/check.py
```diff
@@ -771,3 +784,42 @@ def _ensure_events(events):
...
+            fmt = next((k for k, v in supported_formats.items() if fmt in v),
+                       fmt)  # default to original fmt for raising error later
+        else:
+            raise ValueError(f"Couldn't infer format from filename {fname}"
+                             " (no extension found)")
+
+    if fmt not in supported_formats:
+        supported_str = ', '\
+            .join(f'{k} ({", ".join(["*." + ext for ext in v])})'
+                  for k, v in supported_formats.items())
```
hoechenberger, line 822: More code, but I believe also more readable and easier to change in the future:
```suggestion
        supported = []
        for format, extensions in supported_formats.items():
            ext_str = ', '.join(f'*.{ext}' for ext in extensions)
            supported.append(f'{format} ({ext_str})')

        supported_str = ', '.join(supported)
```

jackz314: Agreed, the nested list comprehension does feel a bit hacky and is probably difficult to understand. Fixed.

hoechenberger (approved):
👌👌👌👌👌👌👌❤️

hoechenberger:
@cbrnr Feel free to merge if you're happy!

Fantastic work, and thank you so much for your endless patience, @jackz314!

mne/tests/test_epochs.py
```diff
@@ -3096,6 +3097,34 @@ def test_save_complex_data(tmpdir, preload, is_complex, fmt, rtol):
     assert_allclose(data_read, data, rtol=rtol)
 
 
+@pytest.mark.skipif(not _check_eeglabio_installed(strict=False),
+                    reason='eeglabio not installed')
+@pytest.mark.parametrize('preload', (True, False))
+def test_export_set(tmpdir, preload):
+    """Test saving an Epochs instance to EEGLAB's set format."""
+    if not _check_eeglabio_installed(strict=False):
+        return
```
agramfort, line 3106: this can be removed now that you have the pytest skip above

mne/utils/docs.py
```diff
@@ -2262,6 +2262,29 @@
...
+    the info will be preserved in the external format. To save in native MNE
+    format (``.fif``) without information loss, use :func:`save` instead.
+"""
+docdict['export_params_base'] = """
+Parameters
+----------
+fname : str
+    Name of the output file.
+"""
+docdict['export_params_fmt'] = """
```
agramfort, line 2278:
```suggestion
docdict['export_params_fmt'] = """
fmt : 'auto' | 'eeglab'
```

mne/io/base.py
```diff
@@ -1450,6 +1453,54 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
...
+        fmt : 'auto' | 'eeglab'
```
agramfort, line 1463:
```suggestion
```

mne/epochs.py
```diff
@@ -1810,6 +1813,50 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
...
             _save_split(this_epochs, fname, part_idx, n_parts, fmt)
 
+    @verbose
+    def export(self, fname, fmt='auto', verbose=None):
+        """Export Epochs to external formats.
+
+        Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
+        %(export_warning)s
+        %(export_params_base)s
+        fmt : 'auto' | 'eeglab'
```
agramfort, line 1823:
```suggestion
```

mne/utils/docs.py
jackz314: The idea is raw and epochs might have different supported formats in the future. So I intentionally left the supported formats out of the docdict. What do you think?

cbrnr:
@hoechenberger can you check what needs to be done for @agramfort's latest comments? I think we're not yet ready.

jackz314:
> @hoechenberger can you check what needs to be done for @agramfort's latest comments? I think we're not yet ready.

Yeah I removed the redundant code but I'm not sure about the docstring part.

mne/io/base.py
```diff
@@ -20,7 +20,9 @@
 import numpy as np
 
 from .constants import FIFF
-from .utils import _construct_bids_filename, _check_orig_units
+from .utils import _construct_bids_filename, _check_orig_units, \
+    _get_cart_ch_coords_from_inst
+from ..utils.check import _infer_check_export_fmt
```
larsoner, line 25: Should be down in the `from ..utils` line, you'll need to add `_infer_check_export_fmt` to `mne/utils/__init__.py`

mne/io/base.py
```diff
@@ -1450,6 +1453,54 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
                    start, stop, buffer_size, projector, drop_small_buffer,
                    split_size, split_naming, 0, None, overwrite)
 
+    @verbose
+    def export(self, fname, fmt='auto', verbose=None):
+        """Export Raw to external formats.
+
+        Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
+        %(export_warning)s
+        %(export_params_base)s
```
larsoner, line 1462: I think I'd prefer:
```
%(export_warning)s

Parameters
----------
%(export_params_fname)
```
I know it's less DRY, but it's also clearer that you've entered the Parameters section of the docstring at this point, which is a good tradeoff to me. The first time I read this I had to read extra and stop and think "okay where does Parameters actually start" and then dig into a variable in another file. I think the `docdict` invention is challenging enough for contributors even without the section headers being buried in them :)

mne/io/utils.py
```diff
@@ -312,3 +312,24 @@ def _construct_bids_filename(base, ext, part_idx):
     if dirname:
         use_fname = op.join(dirname, use_fname)
     return use_fname
+
+
+def _get_cart_ch_coords_from_inst(chs, drop_chs=None):
```
larsoner, line 317: I don't think this name reflects what it does -- our channels are already in Cartesian coordinates, we just use a RAS definition for our dimensions and it looks like EEGLAB uses ALS (based on the changes). So I would name this `_get_als_coords_from_ras` or `_get_eeglab_coords_from_inst` or something

mne/utils/check.py
```diff
@@ -771,3 +784,45 @@ def _ensure_events(events):
...
+    fmt : str
+        Format of the export, will only infer the format from filename if fmt
+        is auto.
+    fname : str
+        Name of the target export file, only used when fmt is auto.
+    supported_formats : dict of str : tuple/list
+        Dictionary containing supported formats (as keys) and each format's
+        corresponding file extensions in a tuple/list (e.g. 'eeglab': ('set',))
+    """
+    fmt = fmt.lower()
```
larsoner, line 807:
```suggestion
    _validate_type(fmt, str, 'fmt')
    fmt = fmt.lower()
```

larsoner (commented):
Code looks good, just minor suggestions really!

However, I don't think we should merge until this is solved, i.e., eeglabio makes a PyPi release:
```
$ pip install --user eeglabio
ERROR: Could not find a version that satisfies the requirement eeglabio
ERROR: No matching distribution found for eeglabio
```
and then we add `eeglabio` to `github_actions_dependencies.sh` in the same place as `nitime`. I suspect that none of these lines are really being tested by CIs, and codecov agrees:
```
codecov/patch — 22.32% of diff hit (target 95.00%)
```
I know that there is already a `pip install -i https://test.pypi.org/simple/ eeglabio` in the eeglabio instructions, but I think users will rightly assume they should be able to `pip install eeglabio`, and we should wait until that's available. Also you might consider setting up a conda-forge recipe for it, maybe @hoechenberger can help with that end?

larsoner:
@jackz314 now that #9337 is merged, if you rebase or merge with `upstream/main` you should be able to add `eeglabio` to `requirements_testing_extra.txt` once you make a PyPi release, and it should make CIs actually test the new functionality

jackz314:
> @jackz314 now that #9337 is merged, if you rebase or merge with `upstream/main` you should be able to add `eeglabio` to `requirements_testing_extra.txt` once you make a PyPi release, and it should make CIs actually test the new functionality

I moved eeglabio from Test PyPI to [PyPI](https://pypi.org/project/eeglabio/) and added it as an requirement. The tests should now run on CIs as well.

agramfort:
you have some remaining comments from @larsoner to address.

thx !

>

mne/io/utils.py
jackz314: I decided on `_get_als_coords_from_chs`.

mne/io/base.py
jackz314: Makes sense, I was confused by `docdict` at first as well. It'll be nice if it somehow supported formatting.

jackz314:
> you have some remaining comments from @larsoner to address. thx !
> […](#)

For some reason, I didn't see the comments before. I just addressed them.

larsoner:
Thanks for this @jackz314 !

PR #9197, opened 2021-03-25
BUG: Fix example dtype

larsoner:
Should fix recent CircleCI failures that started when we merged #9166, cc @NeuroLaunch @hoechenberger . The PR was not the problem, it just exposed an error in the example. I added some extra sanity checking of types so that hopefully the error is nicer when users make this mistake.


agramfort (approved):

NeuroLaunch:
Thanks for fixing that, @larsoner.

drammock:
thx @larsoner

PR #9201, opened 2021-03-26
[MRG] Add test for fake click, failing with mpl 3.4

alexrockhill:
#### Reference issue
Linked to issues in https://github.com/mne-tools/mne-bids/pull/756

There isn't a test for the click, which failed in a use-case for `mne-bids` so this PR adds one.

### DIFF
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -13,7 +13,7 @@ raw_fname = op.join(base_dir, 'test_raw.fif')
 
 
 def test_click_ch_name():
-    """Test fake clicks"""
+    """Test fake clicks."""
     import matplotlib
     import matplotlib.pyplot as plt
     matplotlib.use('Agg')
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -16,7 +16,7 @@ def test_click_ch_name():
     """Test fake clicks"""
     import matplotlib
     import matplotlib.pyplot as plt
-    matplotlib.use('Qt5Agg')
+    matplotlib.use('Agg')
     plt.close('all')
     raw = read_raw_fif(raw_fname)
     raw_fig = raw.plot(block=False, show=False, verbose='warning')
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -4,7 +4,25 @@ import numpy as np
 import pytest
 
 from mne.datasets import testing
+from mne.utils._testing import _click_ch_name
 from mne.utils import (_TempDir, _url_to_local_path, buggy_mkl_svd)
+from mne.io import read_raw_fif
+
+base_dir = op.join(op.dirname(__file__), '..', '..', 'io', 'tests', 'data')
+raw_fname = op.join(base_dir, 'test_raw.fif')
+
+
+def test_click_ch_name():
+    """Test fake clicks"""
+    import matplotlib
+    import matplotlib.pyplot as plt
+    matplotlib.use('Agg')
+    plt.close('all')
+    raw = read_raw_fif(raw_fname)
+    raw_fig = raw.plot(block=False, show=False, verbose='warning')
+    _click_ch_name(raw_fig, ch_index=0, button=1)
+    raw_fig.canvas.key_press_event(raw_fig.mne.close_key)
+    assert raw.info['bads'] == ['MEG 0113']
 
 
 def test_buggy_mkl():
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -16,7 +16,7 @@ def test_click_ch_name():
     """Test fake clicks"""
     import matplotlib
     import matplotlib.pyplot as plt
-    matplotlib.use('Agg')
+    matplotlib.use('Qt5Agg')
     plt.close('all')
     raw = read_raw_fif(raw_fname)
     raw_fig = raw.plot(block=False, show=False, verbose='warning')
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -23,7 +23,7 @@ def test_click_ch_name():
     _click_ch_name(raw_fig, ch_index=0, button=1)
     # close events don't fire with non-interactive backends,
     # we have to call the event directly
-    raw_fig.close(event=None)
+    raw_fig._close(event=None)
     assert raw.info['bads'] == ['MEG 0113']
 
 
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -13,7 +13,7 @@ raw_fname = op.join(base_dir, 'test_raw.fif')
 
 
 def test_click_ch_name():
-    """Test fake clicks."""
+    """Test fake click on channel."""
     import matplotlib
     import matplotlib.pyplot as plt
     matplotlib.use('Agg')
--- a/mne/utils/tests/test_testing.py
+++ b/mne/utils/tests/test_testing.py
@@ -21,7 +21,9 @@ def test_click_ch_name():
     raw = read_raw_fif(raw_fname)
     raw_fig = raw.plot(block=False, show=False, verbose='warning')
     _click_ch_name(raw_fig, ch_index=0, button=1)
-    raw_fig.canvas.key_press_event(raw_fig.mne.close_key)
+    # close events don't fire with non-interactive backends,
+    # we have to call the event directly
+    raw_fig.close(event=None)
     assert raw.info['bads'] == ['MEG 0113']
 
 
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -105,8 +105,10 @@ class MNEFigure(Figure):
     def _keypress(self, event):
         """Handle keypress events."""
         if event.key == self.mne.close_key:
-            from matplotlib.pyplot import close
+            from matplotlib.pyplot import close, rcParams
             close(self)
+            if rcParams['backend'].lower() == 'agg':
+                self._close(event)
         elif event.key == 'f11':  # full screen
             self.canvas.manager.full_screen_toggle()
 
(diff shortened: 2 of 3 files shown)


larsoner:
Yeah there are tests failing here, too. They should fail on this PR. Then we'll need to fix them. Do you know how @alexrockhill ? If not, maybe @drammock has ideas.

alexrockhill:
> Yeah there are tests failing here, too. They should fail on this PR. Then we'll need to fix them. Do you know how @alexrockhill ? If not, maybe @drammock has ideas.

It looks like it's an issue with the `_fake_click` (mne.viz.utils.py line 692) that doesn't work on the `Agg` backend but works on `Qt5Agg` (I think). Not sure if that's the best backend for testing but looking through the options (`['GTK3Agg', 'GTK3Cairo', 'MacOSX', 'nbAgg', 'Qt4Agg', 'Qt4Cairo', 'Qt5Agg', 'Qt5Cairo', 'TkAgg', 'TkCairo', 'WebAgg', 'WX', 'WXAgg', 'WXCairo', 'agg', 'cairo', 'pdf', 'pgf', 'ps', 'svg', 'template']`) I don't see a lot of other choices. Probably don't want to go back to Tk ;)

alexrockhill:
If the tests pass on the PR linked above, then the `Qt5Agg` backend works, they are running now

alexrockhill:
It would be nice if the click threw an error if it didn't work too, that's a hard one but the issue was that it failed silently which made it a lot harder to track down

alexrockhill:
Update, looks like `Qt5Agg` works just fine for the backend of the tests. Might slow things down as I assume that was the reason for using `Agg` was that it was quicker...

larsoner:
Also we don't want windows popping up during tests. It's better if we figure out how to make `_fake_click` work with Agg, for example by forcing an event to be processed even though `Agg` usually doesn't.

alexrockhill:
Hmmm well it seems like the changes for the new mpl version effected what is handled by `fig.canvas` and what is handled by `fig.canvas.manager` so maybe the clicks need to go through the manager now

larsoner:
Yeah, someone will need to investigate...

alexrockhill:
Since it works for other backends, it might just be a bug in `matplotlib` specific to `Agg`... I looked and I don't think the manager is handling button presses now.

alexrockhill:
Since it works for other backends, it might just be a bug in `matplotlib` specific to `Agg`... I looked and I don't think the manager is handling button presses now.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.15.4
 scipy>=1.1.0
-matplotlib
+matplotlib<3.4
```
larsoner, line 3: You'll have to change the `pip-pre` builds, too, if you're trying to just work around the error for now. Those reqs live in `tools/*dependencies*.sh` and are more of a pain to deal with. But maybe we can merge this PR as is so that it's at least just the `pip-pre` builds that are left failing until we find a fix

alexrockhill: Ok, I'll take it out and leave it failing

mne/utils/tests/test_testing.py
```diff
@@ -4,7 +4,25 @@
...
+
+base_dir = op.join(op.dirname(__file__), '..', '..', 'io', 'tests', 'data')
+raw_fname = op.join(base_dir, 'test_raw.fif')
+
+
+def test_click_ch_name():
+    """Test fake clicks."""
+    import matplotlib
+    import matplotlib.pyplot as plt
+    matplotlib.use('Agg')
```
drammock, line 19: calls to `matplotlib.use` need to happen before you import pyplot.

mne/utils/tests/test_testing.py
```diff
@@ -4,7 +4,25 @@
...
+    plt.close('all')
+    raw = read_raw_fif(raw_fname)
+    raw_fig = raw.plot(block=False, show=False, verbose='warning')
+    _click_ch_name(raw_fig, ch_index=0, button=1)
+    raw_fig.canvas.key_press_event(raw_fig.mne.close_key)
+    assert raw.info['bads'] == ['MEG 0113']
```
drammock, line 25: if you really just want to test that clicks are happening on Agg, do you even need to load and plot a raw here?  would it work to just create fig & axes, put a box in the middle of the axes, fake a click on it, and check that its `pick` event got fired?

mne/utils/tests/test_testing.py
alexrockhill: Then you get a style error that imports aren't at the top I thought. That's copied from `mne_bids.tests.test_inspect.py`. Maybe I'm wrong though.

mne/utils/tests/test_testing.py
alexrockhill: Hmmm, that would work too but it wouldn't test the functionality of `_click_ch_name`...

mne/utils/tests/test_testing.py
drammock: it used to be the case that importing pyplot sets the backend under the hood if it's not set already, and that it couldn't be changed by subsequent calls to `matplotlib.use`.  It's possible that that's changed by now.  If not, you'd need to use an appropriate `noqa` comment to work around the style error.

mne/utils/tests/test_testing.py
drammock: ok, maybe I misunderstood exactly what you're trying to test.

alexrockhill: What you mentioned already exists here https://github.com/mne-tools/mne-python/blob/main/mne/viz/tests/test_utils.py on line 57 but the test passed so there was something wrong in between I guess

mne/utils/tests/test_testing.py
alexrockhill: I believe that has been changed but again could be wrong, but I used to do the noqa work around and now I don't and things seem okay

alexrockhill:
Ok narrowing down the problem... I ran the test part way through and then used `fig.savefig` and the first channel was grayed out so I think the issue is somewhere during the processing of the click when it gets converted into a channel being marked as bad. I'm not too familiar with that part of the code so any help would be great.

alexrockhill:
So somewhat of a solution, it looks like `MNEBrowseFigure._close` is called with `Qt5Agg` but not with `Agg` after `matplotlib.pyplot.close(self)` is called in mne.viz._figure.py on line 109.

alexrockhill:
Minimally reproducible example.
```
import matplotlib
matplotlib.use('Agg')

def _close(event):
    print(event)


fig, ax = plt.subplots()
fig.canvas.mpl_connect('close_event', _close)
plt.close(fig)

matplotlib.use('Qt5Agg')
fig, ax = plt.subplots()
fig.canvas.mpl_connect('close_event', _close)
plt.close(fig)
```

alexrockhill:
Any idea why the tests are hanging up? I tested locally and the test that failed passed but now the tests won't run...

drammock:
> So somewhat of a solution, it looks like `MNEBrowseFigure._close` is called with `Qt5Agg` but not with `Agg` after `matplotlib.pyplot.close(self)` is called in mne.viz._figure.py on line 109.

this is a known bug that ought to be fixed in MPL main, not sure which release it will go into (I thought 3.4 but maybe they punted until 3.5) https://github.com/matplotlib/matplotlib/issues/18609

mne/viz/_figure.py
```diff
@@ -105,8 +105,10 @@ def _close(self, event):
     def _keypress(self, event):
         """Handle keypress events."""
         if event.key == self.mne.close_key:
-            from matplotlib.pyplot import close
+            from matplotlib.pyplot import close, rcParams
             close(self)
+            if rcParams['backend'].lower() == 'agg':
+                self._close(event)
```
drammock, line 111: I'm not crazy about having backend-triaging code paths in the main codebase (less objection if it's in the tests)

alexrockhill: Hmmm, okay I can just call _close from the test, we can try that

alexrockhill:
Yay, okay I think this is fixed now

alexrockhill:
Huh, not sure what's causing the version/OS-specific failures

alexrockhill:
Huh, related to https://github.com/mne-tools/mne-bids/pull/757, it appears that the child figures can't close themselves with the agg backend which complicates doing it all in the tests

alexrockhill:
Okay, this is what I would do. I checked with a search and if the tests pass, that would be good evidence that nothing else uses the `plt.close` with a non-interactive backend so I think the best solution is to cut out the middleman call to `plt.close` and call the `fig._close` function directly within the figure. Since all mne figures inherit from `MNEFigure` they must all have the `_close(self, event)` function signature so I think this works.

alexrockhill:
Well, actually like Dan said, you could just close all the child figures in the tests, sorry should have read that more carefully, I just put it back

drammock:
OK, I think I've finally caught up / figured out why you opened this in the first place...  I hadn't yet seen the failing tests elsewhere / on other PRs which @larsoner alluded to and which @alexrockhill is seeing on MNE-BIDS apparently.  But I don't quite understand how the test added here will help?  I think it duplicates tests we already have in `mne/viz/tests/test_raw.py`.

Another possibly relevant issue is https://github.com/matplotlib/matplotlib/issues/19576 which has again already been fixed in MPL main (https://github.com/matplotlib/matplotlib/pull/19611).  That issue was causing pick events to happen twice, effectively marking and then un-marking a channel in one go.  I forget exactly how we worked around that to get the tests to pass, it's possible we'll need to undo / revisit that fix.

I'll make a debugging push on monday to try to sort this out, if @alexrockhill doesn't beat me to it.

alexrockhill:
> Another possibly relevant issue is [matplotlib/matplotlib#19576](https://github.com/matplotlib/matplotlib/issues/19576) which has again already been fixed in MPL main ([matplotlib/matplotlib#19611](https://github.com/matplotlib/matplotlib/pull/19611)). That issue was causing pick events to happen twice, effectively marking and then un-marking a channel in one go. I forget exactly how we worked around that to get the tests to pass, it's possible we'll need to undo / revisit that fix.

Ah maybe that's the mystery that was causing the tests to fail in `mne-bids`. Great, we can close this then, it does look redundant.

hoechenberger:
There's half a dozen stalled tests associated with this PR that didn't get canceled when the PR was closed. Canceling them manually now.

PR #9202, opened 2021-03-27
DOC: more website fixes

### LINKED ISSUE #8876: MAINT: Website issues
- A dedicated sponsors page would be nice (see how pandas does it) - we could include a link to it in the left sidebar, and then we can discuss what to do with the right sidebar (remove, keep but link the title to the sponsors page, show the link in the left sidebar only when the right sidebar disappears).
- Some numbers/characters in the sponsors sidebar are still different from the main text (e.g. P41, MIND, IDEX).
- There are still duplicate links in the left sidebar and the top nav (Docs, Get help); I'd only keep them in the top navbar.
- My preference was having all sidebars on the right; people didn't like this, and now we have two sidebars, which I think looks OK, but the problem is that the right sidebar disappears below a certain width with no replacement. This is not ideal. If we remove two links from the Version box (see previous bullet point), the sponsors sidebar could go below the version sidebar on the left.
- Often, searching for a particular answer on Google takes me to the dev version of our website. If I switch to the stable version, it always takes me to the homepage instead of the corresponding page in stable. Is this something that could be solved?

_Originally posted by @cbrnr in https://github.com/mne-tools/mne-python/issues/8757#issuecomment-779644396_

drammock:
closes #8876 

The only (?) change raised in that issue that isn't addressed here is the "sticky sidebars when scrolling" issue raised by @marsipu.  I can't find a config option for it in the theme, so I suspect that it ought to work and I broke it when I made it so that the footer isn't *always* just off the bottom edge of the screen even on short pages.  If I'm right about the cause, then I think maybe the best solution is to ditch our custom footer (with all the institution logos) for something simpler (just the copyright) and put the list of institutional sponsors only on the new funding page.  WDYT @cbrnr @larsoner @agramfort?


cbrnr:
I think this is a very good idea!

drammock:
Renderings so far:

Homepage: https://26577-1301584-gh.circle-artifacts.com/0/dev/index.html

New funders page: https://26577-1301584-gh.circle-artifacts.com/0/dev/funding.html

agramfort (approved):
don't know why CIs complain

cbrnr:
I still see the footer with all contributed logos?

drammock:
> I still see the footer with all contributed logos?

I haven't made that change yet, was waiting for feedback (as it's the weekend I thought I should give people until Monday to weigh in)

larsoner:
To me two things would be nice to preserve on the landing page:

1. Contributing institutions are visible
2. Funders are visible

For (1) it would be nice to keep the institutions at the bottom, but if we can't do that, maybe we could just add them to the homepage? Those might look cool in the right bar where the funders used to be. I don't like having them just at the bottom of the main page (non-sticky) because I think most people will never see it, and I think they should be prominent.

For (2) I don't like that all of our funders lost prominence -- they're basically hidden and hard to find now. Maybe we should have a second box in the left column (below the version box with link) that just shows the funder logos, and the entire box links to the funders page?

drammock:
> Maybe we should have a second box in the left column (below the version box with link) that just shows the funder logos, and the entire box links to the funders page?

As mentioned in https://github.com/mne-tools/mne-python/issues/8876#issuecomment-798770097, the homepage left sidebar disappears below a certain width.  Theme default in that case is to put that *inline before the main content* which was nixed long ago by @agramfort because you have to scroll past tons of funders before even seeing our logo.  The remedy was to not show the left sidebar content at all once it collapses, so anything we put there becomes inaccessible to mobile users.  For the "quicklinks" box this is OK because it links to pages that are in the TOCtree, so they are accessible in other ways (though with more clicks) via the main nav.

So if it's important to have funders and institutions on the homepage, they can't go in either sidebar. Which leaves the footer, or the main content area below the gallery.  Since that's the only thing we haven't tried yet, I'll see if I can work something up that looks nice.  Having logos only and linking to a detailed funders page seems like a good approach so I'll try that first.

larsoner:
> As mentioned in #8876 (comment), the homepage left sidebar disappears below a certain width. 

Ahh sorry, forgot about that...

> So if it's important to have funders and institutions on the homepage, they can't go in either sidebar. Which leaves the footer, or the main content area below the gallery. Since that's the only thing we haven't tried yet, I'll see if I can work something up that looks nice. Having logos only and linking to a detailed funders page seems like a good approach so I'll try that first.

I think we have to consider two use cases: mobile and desktop.

On mobile, the sidebar disappears, and I'm okay with no funders or institutions being very visible there (or maybe they are at the bottom).

On desktop, the sidebar is there, so it would be nice to have a funders box, and maybe an institition box below that. I don't think putting them at the bottom is a good solution because they won't be seen very often.

So a workable solution might be to add to the sidebar, but then also have a `.. container` in the main text that more or less duplicates that content (in some mobile-friendly form), and only shows up when the left bar disappears. I think this maximizes desktop visibility while maintaining content on but not distrupting the mobile experience.

drammock:
> On mobile, the sidebar disappears, and I'm okay with no funders or institutions being very visible there (or maybe they are at the bottom).

this is the opposite of what @cbrnr expressed (https://github.com/mne-tools/mne-python/issues/8876#issuecomment-798880931) 😖 

I'll try your suggestion of big screen -> funders in sidebar, narrow screen -> funders below homepage gallery.  This is still not ideal from the perspective of someone using a screen reader, since site content will be in a different order for them depending on which device they're using.

larsoner:
One way to get a more uniform experience and to promote contributors and funders would be to demote the gallery. For example we could put everything in the main contents with the title box (col-xs-12), then funders (col-sm-12, col-md-6) and contributors (col-sm-12, col-md-6) then the gallery (col-xs-12) or so.

PR #9203, opened 2021-03-27
Add support for importing channel location from XYZ/CSV files.

jackz314:
#### Reference issue
"Spin-off" from #9192. 

Add support for importing channel location from XYZ/CSV files as an extra XYZ option (to be finalized) for [`read_custom_montage`](https://mne.tools/stable/generated/mne.channels.read_custom_montage.html) as per suggestion from [here](https://github.com/mne-tools/mne-python/pull/9192#issuecomment-806463148).

#### Additional information
The naming (XYZ?) & formatting of this feature isn't fully finalized yet, I'm still waiting for input from more qualified people on that.


agramfort (commented):
tell us when you added test and a what's new entry and it's good to go from your end.

jackz314:
I think feature-wise it's done, but I'm not sure what's good enough for the tests. Should I just import from a CSV and compare the resulting montage to the CSV data? That's how I "tested" it locally, but I'm not sure if it's good and comprehensive enough.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
         pos *= head_size / np.median(np.linalg.norm(pos, axis=1))
 
     return make_dig_montage(ch_pos=_check_dupes_odict(ch_names, pos))
+
+
+def _read_csv(fname, delimiter=','):
+    """Import eeg channel locations from CSV files into MNE instance.
```
agramfort, line 340:
```suggestion
    """Import EEG channel locations from CSV files into MNE instance.
```

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    Optionally the first column can contain the channel names.
+
+    Parameters
+    ----------
+    fname : str
+        Name of the csv file to read channel locations from
+    delimiter : str
+        Delimiter used by the CSV file
+    include_ch_names : bool
+        Whether the CSV file include channel names as the first column
```
agramfort, line 352: this parameter does not exist

agramfort (commented):
do you need help to add some unit tests?

jackz314:
> do you need help to add some unit tests?

Some help would be great since I'm still not really familiar with the MNE testing standards.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    """
+    include_ch_names = True
+    import csv
```
drammock, line 355: python standard library imports should go top-of-file, not nested inside function. We only nest optional dependencies (or internal imports to avoid circular import errors)

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    # if not include_ch_names:
+    #     for ch, coord in zip(chs, coords):
+    #         ch['loc'][:3] = coord
+    # else:
+    #     for ch in chs:
+    #         try:
+    #             coord = coords[ch_names.index(ch['ch_name'])]
+    #         except ValueError:  # ch_name not in channel list, default to 0
+    #             coord = (0, 0, 0)
+    #         ch['loc'][:3] = coord
```
drammock, line 379: cruft?

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    Parameters
+    ----------
+    fname : str
+        Name of the csv file to read channel locations from
+    delimiter : str
+        Delimiter used by the CSV file
+    include_ch_names : bool
+        Whether the CSV file include channel names as the first column
+    """
+    include_ch_names = True
```
drammock, line 354: this is now always true. remove it and eliminate the conditionals that depend on it.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    ch_names = []
+    pos = []
+    for row in csv.reader(f, delimiter=delimiter):
+        if include_ch_names:
+            ch_name, x, y, z, *_ = row
+            ch_names.append(ch_name)
+        else:
+            x, y, z, *_ = row
+        pos.append((float(x), float(y), float(z)))
+    f.close()
```
drammock, line 367: we typically use the `with open(fname, 'r') as fid:` context manager, so we don't have to expressly close the file afterward. For consistency we should do the same here.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
         pos *= head_size / np.median(np.linalg.norm(pos, axis=1))
 
     return make_dig_montage(ch_pos=_check_dupes_odict(ch_names, pos))
+
+
+def _read_csv(fname, delimiter=','):
+    """Import eeg channel locations from CSV files into MNE instance.
+
+    CSV files should have columns x, y, and z, each row represents one channel.
+    Optionally the first column can contain the channel names.
```
drammock, line 343: the way the function is now written, channel names as first column are not in fact optional. Revise this text to reflect that.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
...
+    f.readline()
+    ch_names = []
+    pos = []
+    for row in csv.reader(f, delimiter=delimiter):
+        if include_ch_names:
+            ch_name, x, y, z, *_ = row
+            ch_names.append(ch_name)
+        else:
+            x, y, z, *_ = row
+        pos.append((float(x), float(y), float(z)))
```
drammock, line 366: (nitpick) you could omit the three `float()` calls here, and just add `dtype=float` when you later convert `pos` to a numpy array.  Seems cleaner to me that way.

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +334,46 @@ def _read_brainvision(fname, head_size):
         pos *= head_size / np.median(np.linalg.norm(pos, axis=1))
 
     return make_dig_montage(ch_pos=_check_dupes_odict(ch_names, pos))
+
+
+def _read_csv(fname, delimiter=','):
```
drammock, line 339: not clear to me why `delimiter` is exposed here.  It's a private function (so users can't actually change it) and it's only called once in the codebase and there's no triaging based on file extension (csv vs tsv).  Do you know there to be XYZ channel location text files that use a different delimiter?  If so, we should expose delimiter in the *public* API somehow.  If not, we should assume comma-delimiter and remove this param.

drammock (changes requested):

agramfort:
pushed a commit to add a test and address some of the comments

>

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +336,31 @@ def _read_brainvision(fname, head_size):
...
 
     return make_dig_montage(ch_pos=_check_dupes_odict(ch_names, pos))
+
+
+def _read_csv(fname):
+    """Import EEG channel locations from CSV files.
+
+    CSV files should have columns 4 columns containing
+    ch_name, x, y, and z. Each row represents one channel.
+    The first column can contain the channel names.
```
drammock, line 346:
```suggestion
```

drammock: (cruft)

mne/channels/_standard_montage_utils.py
```diff
@@ -334,3 +336,31 @@ def _read_brainvision(fname, head_size):
...
+
+    Parameters
+    ----------
+    fname : str
+        Name of the csv file to read channel locations from
```
drammock, line 351:
```suggestion
        Name of the csv file to read channel locations from.
```

drammock (approved):
just 2 small docstring fixes.

agramfort:
@jackz314 do you want to take over?

mne/channels/_standard_montage_utils.py
jackz314: `delimiter` is a leftover parameter, the only different ones are probably tabs in TSV files. Should this be checked in `read_custom_montage` or in `_read_csv`? As for the channel names, I've seen files with channel names (mainly CSV files) and ones without (mainly XYZ files).

jackz314:
I'm finalizing the changes now. Should we change `_read_csv` to `_read_xyz`? Since it should handle all three formats (csv, tsv, and xyz)? Also, it seems like the norm for XYZ files is that channel names are in the last column, but I've also seen some, where it's the first column, what should we do about this?

agramfort:
Csv is not a standard. We cannot do magic read. With what you say I am now tempted to say user should just call make dig montage and deal with reading the csv on their side. We cannot claim to support more. Take home message : avoid black magic as it will backfire

jackz314:
I pushed a commit with some code to adapt for TSV, CSV, and XYZ files. Since CSV isn't a standard, I just specified and used the ch_name, x, y, z format, which follows the TSV standard used elsewhere. For XYZ files I used the EEGLAB standard (or just the standard EEGLAB uses), which is row count, x, y, z, ch_name.

jackz314:
Thanks for the tests, I ran them locally and it seems like they are passing. I think this should be ready unless there's anything else needed.

agramfort (commented):
@jackz314 you need now to add a what's new entry in the file latest.inc in the doc folder

also I feel that we should update the page https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html

@drammock you confirm it would be the right place?

drammock:
> also I feel that we should update the page https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html
> @drammock you confirm it would be the right place?

Not sure I agree.  `read_custom_montage` isn't really covered anywhere in the tutorials (it is used once in passing in https://mne.tools/dev/auto_tutorials/source-modeling/plot_eeg_mri_coords.html).  I think it might make more sense to create a new how-to example: "how to load a custom EEG sensor montage" in the input/output category.

jackz314:
Just added the what's new entry. 

I think it would make a lot of sense to add `read_custom_montage` to a tutorial somewhere since otherwise, it might be confusing for people just learning about MNE. I think it would make sense to add it to the existing page [https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html](https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html) since it's titled "Working with sensor locations", which seems like a natural place for adding/reading custom sensor locations.

drammock:
> Just added the what's new entry.
> 
> I think it would make a lot of sense to add `read_custom_montage` to a tutorial somewhere since otherwise, it might be confusing for people just learning about MNE. I think it would make sense to add it to the existing page https://mne.tools/dev/auto_tutorials/intro/plot_40_sensor_locations.html since it's titled "Working with sensor locations", which seems like a natural place for adding/reading custom sensor locations.

I guess it would be OK to add a section between "working with built-in montages" and "controlling channel projection" called "working with custom montages" that shows a quick example of `read_custom_montage`.  It should probably list the various formats that we support (not just XYZ).

doc/changes/latest.inc
```diff
@@ -55,6 +55,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add CSV, TSV, and XYZ support to :func:`mne.channels.custom_montage` (:gh:`9203` **by new contributor** |Jack Zhang|_)
```
agramfort:
```suggestion
- Add CSV, TSV, and XYZ support to :func:`mne.channels.read_custom_montage` (:gh:`9203` **by new contributor** |Jack Zhang|_)
```

doc/changes/latest.inc
```diff
@@ -17,7 +17,7 @@
 Current (0.23.dev0)
 -------------------
 
-.. |New Contributor| replace:: **New Contributor**
```
agramfort, line 20: please keep this line. Add a line for you but do not replace it

jackz314: Not sure why this is showing up as removed. It is kept in my latest commit.

doc/changes/latest.inc
```diff
@@ -19,6 +19,10 @@ Current (0.23.dev0)
 
 .. |New Contributor| replace:: **New Contributor**
 
+.. |Jack Zhang| replace:: **Jack Zhang**
+
+.. |New Contributor| replace:: **New Contributor**
```
agramfort, line 24: now 


.. |New Contributor| replace:: **New Contributor**

is written twice

agramfort (approved):
anyone merge if happy

thx @jackz314

doc/changes/latest.inc
jackz314: Fixed.

drammock:
@jackz314 are you planning to add to the tutorial in this PR, or in a separate one?

jackz314:
> @jackz314 are you planning to add to the tutorial in this PR, or in a separate one?

I wasn't planning on adding the tutorials myself, mostly because I'm not really familiar with how the tutorial part or the other kinds of montages work yet. But if no one else is working on it, I can look into it and open another PR later if I have time.

drammock:
Ok I'll merge this one now then.

Issue #9193, opened 2021-03-25
Crashes when using read_epochs_eeglab on files with only one trial.


jackz314:
#### Describe the bug
When reading in single trial EEGLAB files as epochs, MNE will crash when initializing `EpochsEEGLAB`. This is because the trial count is 1, so creating the `event_id` dict is skipped, and when [this line](https://github.com/mne-tools/mne-python/blob/13a0de51bd0f13113d866e73522fb60bdde37809/mne/io/eeglab/eeglab.py#L507) references `event_id`, an `AttributeError` will be thrown. Like this one:

```
test_epochs.py:3046: in test_save_set
    epochs_read = read_epochs_eeglab(temp_fname)
../io/eeglab/eeglab.py:275: in read_epochs_eeglab
    epochs = EpochsEEGLAB(input_fname=input_fname, events=events, eog=eog,
<decorator-gen-240>:24: in __init__
    ???
../io/eeglab/eeglab.py:507: in __init__
    for key, val in event_id.items():
E   AttributeError: 'NoneType' object has no attribute 'items'
```

#### Steps to reproduce
I discovered this when writing tests for my PR #9192, I was following another similar test [`test_save_complex_data` ](https://github.com/mne-tools/mne-python/blob/13a0de51bd0f13113d866e73522fb60bdde37809/mne/tests/test_epochs.py#L3017) that also saves single-trial epochs to a file and reads it back as a test. I simply wrote the same data into an EEGLAB set file and read it back. Alternatively, you can also try `read_epochs_eeglab` on any single-trial set file.

#### Expected results
Since single-trial epochs aren't really useful or meaningful (from my understanding so far), I understand if this isn't supported, but some better clarifications on this would probably make more sense to other people, for example, throwing another exception specifically for single-trial data when attempting `read_epochs`. 

#### Actual results
N/A

#### Additional information
Platform:      Linux-5.8.0-45-generic-x86_64-with-glibc2.32
Python:        3.8.6 (default, Jan 27 2021, 15:42:20)  [GCC 10.2.0]
Executable:    /usr/bin/python
CPU:           x86_64: 8 cores
Memory:        31.3 GB

mne:           0.23.dev0
numpy:         1.19.5 {blas=openblas, lapack=openblas}
scipy:         1.4.1
matplotlib:    3.3.4 {backend=Qt5Agg}

sklearn:       0.24.1
numba:         0.52.0
nibabel:       Not found
nilearn:       Not found
dipy:          Not found
[...]

PR #9204, opened 2021-03-27
Support single-epoch files for EEGLAB epoch.

jackz314:
#### Reference issue
#9193 

Add support for single-epoch (zero trials) set files from EEGLAB when importing as Epochs.

#### Additional information
MNE no longer crashes when attempting `read_epochs_eeglab()` on a single-epoch set file. A placeholder `events` array and `event_id` dict is added to ensure compatibility with the rest of the Epochs' operations.


jackz314:
Added a test case for single epoch EEGLAB files in `test_eeglab.py`.

mne/io/eeglab/eeglab.py
```diff
@@ -488,16 +488,24 @@ def __init__(self, input_fname, events=None, event_id=None, tmin=0,
...
+                        prev_stim = event_id[event_name[idx - 1]]
+                    events[idx, 0] = event_latencies[idx]
+                    events[idx, 1] = prev_stim
+                    events[idx, 2] = event_id[event_name[idx]]
+            else:
+                # create placeholder event_id and events here to support
+                # single-epoch files
+                events = np.zeros((1, 3), dtype=int)
+                events[0, 2] = 1
+                event_id = {'1': 1}
```
agramfort, line 508: I am confused by the fix you suggest. TO me you should have eeg.trials == 1 and going through the for loop should work

mne/io/eeglab/tests/test_eeglab.py
```diff
@@ -430,3 +430,16 @@ def test_io_set_raw_2021():
     assert "EEG" not in io.loadmat(raw_fname_2021)
     _test_raw_reader(reader=read_raw_eeglab, input_fname=raw_fname_2021,
                      test_preloading=False, preload=True)
+
+
+@testing.requires_testing_data
+def test_read_single_epoch():
+    """Test reading a single epoch/trial set file as Epochs"""
+    raw = read_raw_eeglab(raw_fname_mat, preload=True)
+    epochs = read_epochs_eeglab(raw_fname_mat)
```
agramfort, line 439: why are you using read_epochs_eeglab to read a raw continuous file?

mne/io/eeglab/eeglab.py
jackz314: Because if there's only one trial the event-related variables won't exist, so using `event_id` wouldn't work. In EEGLAB, I don't think there's a distinct difference between "epochs" or "raw" data like MNE has, so this is one of the reasons I suggested throwing an exception earlier.

mne/io/eeglab/tests/test_eeglab.py
jackz314: This is the simplest way I can find to pretend there's a single-epoch SET file, since the purpose is only to test if `read_epochs` works with single-epoch data.

jackz314:
> but then to me it's not one epoch obtained by chunking a raw based on an event onset. 

The original problem that made me create this PR was mistakenly reading some set file that wasn't epoched in EEGLAB, which means there wasn't any event-related data. I don't think you can have just one trial and still have event data in EEGLAB because that would make it two trials at least.

Since the distinction between raw and epochs in EEGLAB is solely based on events (source of epochs), I think it might be more suitable to just throw an exception because of the clearer distinction between single epochs based on events and raw data in MNE.

> there is a clear difference in MNE between a Raw and one Epoch

I am aware of that, I was referring to EEGLAB, which doesn't have that distinction.

agramfort:
ok then I suggest that the right fix is to raise a meaningful error message
when one
tries to use read_eeglab_epochs on a raw EEGLAB file.

can you update the PR accordingly?

>

jackz314:
I added the exception and changed the test as well. The error I used is `TypeError("Can't read epochs from a raw file (trials less than 2)")`

agramfort (approved):

mne/io/eeglab/tests/test_eeglab.py
```diff
@@ -430,3 +430,10 @@ def test_io_set_raw_2021():
     assert "EEG" not in io.loadmat(raw_fname_2021)
     _test_raw_reader(reader=read_raw_eeglab, input_fname=raw_fname_2021,
                      test_preloading=False, preload=True)
+
+
+@testing.requires_testing_data
+def test_read_single_epoch():
+    """Test reading a single epoch/trial (raw) set file as Epochs"""
```
larsoner, line 437:
```suggestion
    """Test reading raw set file as an Epochs instance."""
```

larsoner:
Thanks @jackz314 !

PR #9208, opened 2021-03-28
DOC: Move many Epochs docstrings to docdict, and use them in BaseEpochs

hoechenberger:


agramfort (approved):
Assuming it is plain copy paste and you checked rendering on ci artifact

hoechenberger:
> Assuming it is plain copy paste and you checked rendering on ci artifact

Artifacts look good; ~only other change I did was I combined the docstrings of `tmin, tmax` and `reject_tmin, reject_tmax`, respectively.

PR #9212, opened 2021-03-29
CI: Cancel GH Actions jobs after 90 min max

hoechenberger:
In hope to prevent future stale CI jobs from blocking the queue again
Doesn't affect all jobs, only those where we've seen stalling

### DIFF
--- a/.github/workflows/codespell_and_flake.yml
+++ b/.github/workflows/codespell_and_flake.yml
@@ -9,6 +9,7 @@ on:
 
 jobs:
   style:
+    timeout-minutes: 90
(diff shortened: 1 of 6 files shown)


hoechenberger:
I've merged sth similar for MNE-BIDS and it seems to be working as intended!

PR: https://github.com/mne-tools/mne-bids/pull/760

In action:
https://github.com/mne-tools/mne-bids/pull/759/checks?check_run_id=2220090841

(the run got stalled there)

larsoner:
Thanks @hoechenberger !

PR #9215, opened 2021-03-29
Add parameter real_ori to LCMV, DICS

britta-wstnr:
This is work in progress, mostly to test and see if taking the complex part to compute the orientation selection makes any sense. 

CC @larsoner

### DIFF
--- a/mne/beamformer/_compute_beamformer.py
+++ b/mne/beamformer/_compute_beamformer.py
@@ -142,7 +142,7 @@ def _sym_inv_sm(x, reduce_rank, inversion, sk):
 
 def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
                         reduce_rank, rank, inversion, nn, orient_std,
-                        whitener):
+                        whitener, real_ori):
     """Compute a spatial beamformer filter (LCMV or DICS).
 
     For more detailed information on the parameters, see the docstrings of
@@ -174,6 +174,7 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
         The std of the orientation prior used in weighting the lead fields.
     whitener : ndarray, shape (n_channels, n_channels)
         The whitener.
+    real_ori : make tests happy for now
 
     Returns
     -------
@@ -273,7 +274,11 @@ def _compute_beamformer(G, Cm, reg, n_orient, weight_norm, pick_ori,
         assert ori_pick.shape == (n_sources, n_orient, n_orient)
 
         # pick eigenvector that corresponds to maximum eigenvalue:
-        eig_vals, eig_vecs = np.linalg.eig(ori_pick.real)  # not Hermitian!
+        if real_ori is True:
+            eig_vals, eig_vecs = np.linalg.eig(ori_pick.real)  # not Hermitian!
+        else:
+            eig_vals, eig_vecs = np.linalg.eig(ori_pick)  # not Hermitian!
+
         # sort eigenvectors by eigenvalues for picking:
         order = np.argsort(np.abs(eig_vals), axis=-1)
         # eig_vals = np.take_along_axis(eig_vals, order, axis=-1)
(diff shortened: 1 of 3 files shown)


britta-wstnr:
@larsoner also exposed the parameter `real_ori` in LCMV for testing, although it does not make much sense there from a user perspective.
Also did not bother to make the documentation useful yet.

larsoner:
@britta-wstnr I cooked up a max-power ori test. For some reason with oct-4 (~500 sources) it's still really close to chance for the orientations (average dot product around 0.5, being 1 would be correct) but with oct-6 it at least goes up to 0.78, which still isn't great but is at least better:

https://gist.github.com/larsoner/0c3e43057f82106245d4eec9238fb9a6

britta-wstnr:
An update on this, following a discussion with @larsoner:
Based on this PR, @larsoner did some simulations, see this [gist](https://gist.github.com/larsoner/6dcf350225a682d0a1b7c8dab8987c4e).
The outcome of this is that `real_filter=True` is better than complex filters (the current default!) and that an eigenvalue decomposition behaves better than an SVD.
Thus, the eigenvalue decomposition we have currently in the code can stay - but the default should probably be updated. Another reason to do so is that a complex filter is actually hard to interpret biophysically.

Apart from this, one thing that could be nice to have is the possibility to compute coherence with respect to a reference sensor (or grid point), a thing that FieldTrip offers, see [for example here](https://www.fieldtriptoolbox.org/tutorial/coherence/).

larsoner:
@wmvanvliet okay to change by deprecation the default for DICS to `real_filter=True`?

@britta-wstnr want to push that change to this PR, or want me to do it? It would be good to get this change into 0.23 if we can

britta-wstnr:
If @wmvanvliet agrees I can give it a stab later today!

britta-wstnr:
@larsoner: we revert all changes regarding `real_ori` here, right? Or should we keep this around for a little longer for testing and make the `real_filter` changes in a separate PR?

larsoner:
I would just revert all of those, I don't think we need it. Also you could just make a clean PR and close this if you want

britta-wstnr:
Let's make a new PR then, then I can keep the branch around in case we need it

wmvanvliet:
I agree to put `real_filter=True` as a default if that is actually better. I'll see if I can add the param to my own simulations and do a run.

wmvanvliet:
> Apart from this, one thing that could be nice to have is the possibility to compute coherence with respect to a reference sensor (or grid point), a thing that FieldTrip offers, see for example here.

We're adding this to the [conpy](https://github.com/aaltoimaginglanguage/conpy) package, which is a package for using DICS to compute connectivity.

wmvanvliet:
the connectivity computation functions in that package should be updated to take in a DICS beamformer as constructed through MNE-Python.

wmvanvliet:
Triton is done with the DICS simulation runs. Comparing orientation error with `real_ori=True` vs `real_ori=False`
Figure_1
Doesn't seem to make any difference?

larsoner:
Running a variant of your DICS example:

https://gist.github.com/larsoner/6dcf350225a682d0a1b7c8dab8987c4e

It seemed like almost any variant of `vertidx` I chose had lower orientation error for `real_filter=True` (granted both were small), and also less spread (these errors are in degrees, so both quite small errors). For example for the vertex pair `[[1000, 2000]]`:
```
real_filter: [0.2 0. ]
complex_filter: [1.2 2.9]
```
Screenshot from 2021-04-22 16-44-50
Screenshot from 2021-04-22 16-44-55

And for ``[[0], [1000]]`:
```
real_filter: [0.2 0.6]
complex_filter: [1.2 3.8]
```
Screenshot from 2021-04-22 16-47-35
Screenshot from 2021-04-22 16-47-38

I'm not sure if the difference is that my sampling was not systematic (it certainly wasn't, but not sure if that's the cause of the differenc), or that our simulations differed, or the quantification metrics are different...

wmvanvliet:
Another big difference is that the number of sources are different. OK then, if it doesn't seem to matter in our extensive sins and does seem to help in your sim here, then I guess it's a good idea.

britta-wstnr:
Deprecation is implemented now in #9340

larsoner:
@wmvanvliet @britta-wstnr where did we land on this? Should we discuss at the next dev meeting if it's not clear?

wmvanvliet:
I thought we landed on deprecating and at some point removing the complex filters.

PR #9216, opened 2021-03-29
CI: debugging workflows via circle

drammock:
attempting to replicate the CI failures in GH Workflows within CircleCI, so we can access via SSH (neither @larsoner nor I can reproduce them locally).

### DIFF
--- a/tutorials/misc/plot_hack_circle.py
+++ b/tutorials/misc/plot_hack_circle.py
@@ -1,0 +1,101 @@
+"""
+=========
+Debugging
+=========
+
+Abuse CircleCI's 'rerun with SSH' capabilities for debugging a thorny MPL
+failure.
+"""
+
+import os.path as op
+
+import numpy as np
+import pytest
+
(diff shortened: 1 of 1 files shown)


tutorials/misc/plot_hack_circle.py
```diff
@@ -0,0 +1,101 @@
+"""
```
larsoner, line 1: In the future you might want to instead just add a CircleCI step that does the commands:
```
pip install pytest pytest-cov
pytest mne/viz/tests/test_raw.py
```
or so. You could do it before the docs even get built so it happens faster. But in theory it shouldn't matter as long as what you have here actually fails

larsoner: looks like it's not failing https://app.circleci.com/pipelines/github/mne-tools/mne-python/7552/workflows/eedd2a4b-6040-4c47-9ffe-c05330c26e38/jobs/26615

PR #9218, opened 2021-03-29
DOC: Last few footbibs

### LINKED ISSUE #8919: use bibtex in sleep_physionet/_utils.py
convert references in `mne/datasets/sleep_physionet/_utils.py` to use footcite / footbibliography

### LINKED ISSUE #8923: use bibtex in mxne_optim.py
convert references in `mne/inverse_sparse/mxne_optim.py` to use footcite / footbibliography

drammock:
closes #8923
closes #8919

### DIFF
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -197,6 +197,16 @@
 	journal = {Journal of Neural Engineering},
 }
 
+@techreport{BurdakovMerkulov2001,
+  title={On a new norm for data fitting and optimization problems},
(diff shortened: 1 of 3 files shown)


drammock:
nb: it turns out none of these are actually rendered anywhere (the physionet one is private, the mxne ones are technically public but aren't in `python_reference.rst`).  I'm inclined to merge it anyway for completeness, and in case one of these becomes public and/or its docstring gets copy/pasted into something that is public.

larsoner:
Agreed, thanks @drammock

PR #9220, opened 2021-03-30
ENH: add _repr_html_ method in the class mne.Evoked

vagechirkov:
Reference issue #9177

Add `_repr_html_` for mne.Evoked:

image
(updated image)


mne/data/html_templates.py
```diff
@@ -100,3 +100,32 @@
     </tr>
 </table>
 """)
+
+evoked_template = Template("""
+<table class="table table-hover">
+    <tr>
+        <th>Name</th>
+        <td>{{evoked.comment}}</td>
```
hoechenberger, line 108: Maybe a term like "Content" would fit better here? Check for example what `evoked.comment` looks like when you average epochs from, say, 4 event types and with different counts. I wouldn't want to call the monster you'll find in `evoked.comment` a "name" anymore 😅😱

doc/changes/latest.inc
```diff
@@ -57,6 +57,8 @@ Current (0.23.dev0)
 
 Enhancements
 ~~~~~~~~~~~~
+- Add HTML representation for `~mne.Evoked` in Jupyter Notebooks (:gh:`9220` by |Valerii Chirkov|_)
+
 - Add HTML representation for `~mne.Epochs` in Jupyter Notebooks (:gh:`9174` by |Valerii Chirkov|_)
```
hoechenberger, line 62: I'd suggest to merge those two entries

mne/data/html_templates.py
```diff
@@ -100,3 +100,38 @@
...
+    <tr>
+        <th>Content</th>
+        <td>{{evoked.comment}}</td>
+    </tr>
+    <tr>
+        <th>Data kind</th>
+        <td>{{evoked.kind}}</td>
+    </tr>
+    <tr>
+        <th>The data shape</th>
```
hoechenberger, line 115:
```suggestion
        <th>Data shape</th>
```

mne/data/html_templates.py
```diff
@@ -100,3 +100,38 @@
...
+        <td>{{f'{evoked.data.shape[0]} channels'}},
+        {{f'{evoked.data.shape[1]} times'}}</td>
```
hoechenberger, line 117: maybe better?
```suggestion
        {{f'{evoked.data.shape[1]} time points'}}</td>
```

mne/data/html_templates.py
```diff
@@ -100,3 +100,38 @@
...
+        <td>{{evoked.comment}}</td>
+    </tr>
+    <tr>
+        <th>Data kind</th>
+        <td>{{evoked.kind}}</td>
+    </tr>
+    <tr>
+        <th>Data shape</th>
+        <td>{{f'{evoked.data.shape[0]} channels'}},
+        {{f'{evoked.data.shape[1]} time points'}}</td>
```
drammock, line 117:
```suggestion
        {{f'{evoked.data.shape[1]} samples'}}</td>
```

drammock: I don't feel strongly about this suggestion, but to me it's good to be agnostic about whether the last dimension is actually "time" since users can put any data they want into our objects using e.g. `EvokedArray`

hoechenberger (approved):
LGTM!

mne/data/html_templates.py
```diff
@@ -100,3 +100,38 @@
     </tr>
 </table>
 """)
+
+evoked_template = Template("""
+<table class="table table-hover">
+    <tr>
+        <th>Content</th>
```
hoechenberger, line 107: I'm starting to think that maybe we should just call this `Conditions`? WDYT?

vagechirkov: I think the original `Comment` is also an option :)

hoechenberger:
@vagechirkov Are you willing to revive this? I think it would still be awesome to have this improvement in the next MNE-Python release! <3

PR #9227, opened 2021-03-30
FIX: don't use op.realpath when reading raw

### LINKED ISSUE #9221: Some trouble with version control tools for data
This is in part a request for better integration between two tools, and in part an FYI.

We are developing datalad (www.datalad.org), a tool to version control and share not only code, but also data and software environments. Internally, datalad uses [git-annex](https://git-annex.branchable.com/) to version control  data. The version control mechanism that git-annex uses is clever, but it prevents mne-python from automatically reading in split data files even if they are correctly named. As more and more MEG dataset are shared openly, and portals that host or analyze these datasets such as OpenNeuro or brainlife.io use DataLad as a backend or frontend, I wanted to raise awareness of the problem, and ask whether its possible to integrate better.
Scripts and datasets that rely on conforming to BIDS standards or mne filenaming standards will fail to read in split files correctly by default.

When a file is version controlled, it is transformed into a symlink with the original file name that points to a path in the "annex" (an internal, non-human readable representation from which for example file transport is managed). This path consists of a hash build from the file content.  
Here is a typical example from a public, BIDS conforming dataset - small files are "as is", larger files are symlinked:

```
 ls -l sub-01/ses-day1/meg 
total 39
-rw-rw---- 1 adina BnB  1211 Mar 30 17:51 sub-01_ses-day1_coordsystem.json
-rw-rw---- 1 adina BnB 25565 Mar 30 17:51 sub-01_ses-day1_task-AVLearn_run-01_channels.tsv
-rw-rw---- 1 adina BnB 34145 Mar 30 17:51 sub-01_ses-day1_task-AVLearn_run-01_events.tsv
-rw-rw---- 1 adina BnB   539 Mar 30 17:51 sub-01_ses-day1_task-AVLearn_run-01_meg.json
lrwxrwxrwx 1 adina BnB   143 Mar 30 17:51 sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif -> ../../../.git/annex/objects/Xz/5P/MD5E-s2146079382--dca9c1e4ee05b6b32bdcd4b5aef8c9bb.fif/MD5E-s2146079382--dca9c1e4ee05b6b32bdcd4b5aef8c9bb.fif
lrwxrwxrwx 1 adina BnB   143 Mar 30 17:51 sub-01_ses-day1_task-AVLearn_run-01_part-02_meg.fif -> ../../../.git/annex/objects/1F/4J/MD5E-s1231308362--bd7d70c27e61d20e79466544e8873954.fif/MD5E-s1231308362--bd7d70c27e61d20e79466544e8873954.fif
```
The symlink is committed into Git, allowing to version control data of arbitrary size. As  the symlink contains precise file identity information (the hash), a content change of the file results in a change of this symlink and in turn transparently into a "modification" that can be committed. 

Internally, mne-python resolves the symlink to read the file (cool!), but then uses only the realpath to construct consecutive splits. This fails, as it searches in the internal data annex (where files are named based on their content, not based on standards) instead of the human readable symlink names.

Here is a reproducer:

```
# get public data from openneuro
$ datalad install https://github.com/OpenNeuroDatasets/ds002598.git
$ cd ds002598/
# retrieve some relevant files (here: two fif files of one subject)
$ datalad get  sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-*.fif

# try to open the first split:
$ ipython
>>> import mne
>>> raw =  mne.io.read_raw_fif('sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif')
Opening raw data file sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif...
    Read a total of 13 projection items:
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
        generated with autossp-1.0.1 (1 x 306)  idle
    Range : 27000 ... 2060999 =     27.000 ...  2060.999 secs
Ready.
<ipython-input-2-cb1c10b217d7>:1: DeprecationWarning: The default for on_split_missing is "warn" in 0.22 but will change to "raise" in 0.23, set it explicitly to avoid this message
  raw = mne.io.read_raw_fif('sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif')
<ipython-input-2-cb1c10b217d7>:1: RuntimeWarning: Split raw file detected but next file /home/adina/scratch/ds002598/.git/annex/objects/Xz/5P/MD5E-s2146079382--dca9c1e4ee05b6b32bdcd4b5aef8c9bb.fif/sub-01_ses-day1_task-AVLearn_run-01_part-02_meg.fif does not exist. Ensure all files were transferred properly and that split and original files were not manually renamed on disk (split files should be renamed by loading and re-saving with MNE-Python to preserve proper filename linkage).
  raw = mne.io.read_raw_fif('sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif')
```

drammock:
closes #9221 

This one-line fix is almost certainly too simplistic and will probably break something, so I'm hoping for some intense scrutiny here.  Here's my understanding of the problem and solution(?):

1. `datalad` uses `git-annex` to manage large data files (`datalad` is the recommended access method for OpenNeuro, so it's really something we should support)
2. for split raw files, this poses a problem because when the first raw is loaded, we call `os.path.realpath()` on the filename, so something like `sub-01/ses-day1/meg/sub-01_ses-day1_task-AVLearn_run-01_part-01_meg.fif` becomes `/home/user/Desktop/test-datalad/ds002598/.git/annex/objects/Xz/5P/MD5E-s2146079382--dca9c1e4ee05b6b32bdcd4b5aef8c9bb.fif/MD5E-s2146079382--dca9c1e4ee05b6b32bdcd4b5aef8c9bb.fif`
3. that works fine for singleton raw files, but for split files, we determine whether it's split based on the original fname, but then go looking for the subsequent parts based on the `realpath`-transformed fname.

my solution is to just not call `realpath()`.  Doing so doesn't seem to hurt much as far as I can tell; I can still load sample data, and after creating a symlink on my desktop to a random `.fif` file on another drive, passing the symlink into `read_raw()` loads it just fine.

The only case I can think of where this will break something that previously worked, is if you had this:
```console
$ tree bar
bar
├── foo
│   ├── qux_part-01_meg.fif
│   └── qux_part-02_meg.fif
└── qux_part-01_meg.fif -> foo/qux_part-01_meg.fif
```
...and you tried to load the symlink in `bar`.  This would fail because now we would look for `...part-02...` in `bar/` instead of in `bar/foo/`.  But honestly that seems like an edge case that we shouldn't support anyway --- allowing users to access all parts of a split raw by providing a symlink to just the first part (when symlinks to the other parts aren't also present).

### DIFF
--- a/mne/io/fiff/raw.py
+++ b/mne/io/fiff/raw.py
@@ -144,7 +144,6 @@ class Raw(BaseRaw):
                 endings += tuple([f'{e}.gz' for e in endings])
                 check_fname(fname, 'raw', endings)
             # filename
-            fname = op.realpath(fname)
             ext = os.path.splitext(fname)[1].lower()
             whole_file = preload if '.gz' in ext else False
             del ext
--- a/mne/io/fiff/raw.py
+++ b/mne/io/fiff/raw.py
@@ -25,7 +25,7 @@ from ...annotations import Annotations, _read_annotations_fif
 
 from ...event import AcqParserFIF
 from ...utils import (check_fname, logger, verbose, warn, fill_doc, _file_like,
-                      _on_missing)
+                      _on_missing, _check_fname)
 
 
 @fill_doc
@@ -75,13 +75,13 @@ class Raw(BaseRaw):
     def __init__(self, fname, allow_maxshield=False, preload=False,
                  on_split_missing='raise', verbose=None):  # noqa: D102
         raws = []
-        do_check_fname = not _file_like(fname)
+        do_check_ext = not _file_like(fname)
         next_fname = fname
         while next_fname is not None:
             raw, next_fname, buffer_size_sec = \
                 self._read_raw_file(next_fname, allow_maxshield,
-                                    preload, do_check_fname)
-            do_check_fname = False
+                                    preload, do_check_ext)
+            do_check_ext = False
             raws.append(raw)
             if next_fname is not None:
                 if not op.exists(next_fname):
@@ -132,18 +132,19 @@ class Raw(BaseRaw):
 
     @verbose
     def _read_raw_file(self, fname, allow_maxshield, preload,
-                       do_check_fname=True, verbose=None):
+                       do_check_ext=True, verbose=None):
         """Read in header information from a raw file."""
         logger.info('Opening raw data file %s...' % fname)
 
         #   Read in the whole file if preload is on and .fif.gz (saves time)
         if not _file_like(fname):
-            if do_check_fname:
+            if do_check_ext:
                 endings = ('raw.fif', 'raw_sss.fif', 'raw_tsss.fif',
                            '_meg.fif', '_eeg.fif', '_ieeg.fif')
                 endings += tuple([f'{e}.gz' for e in endings])
                 check_fname(fname, 'raw', endings)
             # filename
+            fname = _check_fname(fname, 'read', True, 'fname')
             ext = os.path.splitext(fname)[1].lower()
             whole_file = preload if '.gz' in ext else False
             del ext
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -782,7 +782,7 @@ def read_annotations(fname, sfreq='auto', uint16_codec=None):
     _validate_type(fname, 'path-like', 'fname')
     fname = _check_fname(
         fname, overwrite='read', must_exist=True,
-        allow_dir=str(fname).endswith('.ds'),  # allow_dir for CTF
+        need_dir=str(fname).endswith('.ds'),  # for CTF
         name='fname')
     name = op.basename(fname)
     if name.endswith(('fif', 'fif.gz')):
--- a/mne/io/artemis123/artemis123.py
+++ b/mne/io/artemis123/artemis123.py
@@ -8,7 +8,7 @@ import datetime
 import calendar
 
 from .utils import _load_mne_locs, _read_pos
-from ...utils import logger, warn, verbose
+from ...utils import logger, warn, verbose, _check_fname
 from ..utils import _read_segments_file
 from ..base import BaseRaw
 from ..meas_info import _empty_info
@@ -308,6 +308,7 @@ class RawArtemis123(BaseRaw):
         from scipy.spatial.distance import cdist
         from ...chpi import (compute_chpi_amplitudes, compute_chpi_locs,
                              _fit_coil_order_dev_head_trans)
+        input_fname = _check_fname(input_fname, 'read', True, 'input_fname')
         fname, ext = op.splitext(input_fname)
         if ext == '.txt':
             input_fname = fname + '.bin'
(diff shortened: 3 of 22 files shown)


agramfort:
I think it’s the right fix

drammock:
@agramfort any idea why `realpath` was there in the first place?

cbrnr:
This looks correct. I agree that all parts of a file need to be in the same folder, symlink or not. I don't see the reason why we resolved symlinks previously, this shouldn't be necessary at all.

agramfort:
>
> This looks correct. I agree that all parts of a file need to be in the
> same folder, symlink or not. I don't see the reason why we resolved
> symlinks previously, this shouldn't be necessary at all.
>

agreed.

Sorry no I don't remember. Maybe @larsoner knows.

hoechenberger (approved):

agramfort (approved):
@larsoner I let you merge if happy

mne/io/fiff/raw.py
```diff
@@ -144,7 +144,6 @@ def _read_raw_file(self, fname, allow_maxshield, preload,
                 endings += tuple([f'{e}.gz' for e in endings])
                 check_fname(fname, 'raw', endings)
             # filename
-            fname = op.realpath(fname)
```
larsoner, line 147: I tracked it down to a comment years ago:

https://github.com/mne-tools/mne-python/pull/1223#discussion-diff-11646059

I think we should at least keep `abspath` in there in case someone is using a relative path and then `os.chdir`s.

I can add a quick regression test for this use case, make sure it fails, and push a commit.

drammock: sounds good, although I wonder how could someone `os.chdir` in the middle of an operation where a multi-part raw file was being loaded?

larsoner: For `preload=False` the filename gets stored and accessed/used on the fly later

larsoner (approved):
Added a regression test for the linked issue, then added a general `io` test for the `chdir` case I mentioned. It exposed problems in many readers, so I fixed those, too, by making more uniform use of `_check_fname`, which always returns an `abspath` now.

agramfort:
CIs complain

larsoner:
Failure is just a timeout, feel free to merge if you're happy @agramfort

PR #9229, opened 2021-03-31
MRG, MAINT: Improve orientation tests for LCMV and DICS

larsoner:
1. Improve the LCMV orientation test to test only the orientations of the subset of vertices that were correctly localized.
2. Add DICS equivalents of the LCMV tests.

@britta-wstnr I think this might help with #9215

### DIFF
--- a/mne/beamformer/tests/test_dics.py
+++ b/mne/beamformer/tests/test_dics.py
@@ -781,6 +781,7 @@ def test_orientation_max_power(bias_params_fixed, bias_params_free,
     perc = mask.mean() * 100
     assert lower <= perc <= upper
     # Compute the dot products of our forward normals and
+    # assert we get some hopefully reasonable agreement
     assert fwd['coord_frame'] == FIFF.FIFFV_COORD_HEAD
     nn = np.concatenate(
         [s['nn'][v] for s, v in zip(fwd['src'], filters['vertices'])])
--- a/mne/beamformer/tests/test_dics.py
+++ b/mne/beamformer/tests/test_dics.py
@@ -7,23 +7,25 @@ import copy as cp
 import os.path as op
 
 import pytest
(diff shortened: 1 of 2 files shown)


larsoner:
@britta-wstnr can you look and we can merge? I think this is a useful testing enhancement

mne/beamformer/tests/test_dics.py
```diff
@@ -738,15 +740,61 @@ def test_localization_bias_free(bias_params_free, reg, pick_ori, weight_norm,
...
+    loc = np.abs(apply_dics(evoked, filters).data)
+    ori = filters['max_power_ori'][0]
+    assert ori.shape == (246, 3)
+    loc = np.abs(loc)
+    # Compute the percentage of sources for which there is no loc bias:
+    max_idx = np.argmax(loc, axis=0)
+    mask = want == max_idx  # ones that localized properly
+    perc = mask.mean() * 100
+    assert lower <= perc <= upper
+    # Compute the dot products of our forward normals and
```
britta-wstnr, line 783: `and ....?` cliffhanger!

britta-wstnr (commented):
I am still puzzled that agreement is so low. Both `fwd` models are surface only, right?

larsoner:
@britta-wstnr indeed it's disappointing that it's so low, not sure why. But this at least makes the test clearer/cleaner so I think it's a step in the right direction. Okay to merge once CIs come back happy (and I update the comment)?

larsoner:
Okay with you @britta-wstnr ?

britta-wstnr:
Yes, I agree it makes things easier to read.
I've been trying to think why we get such bad results, but I did not come up with anything useful ..... So fine with me to merge here, although it would be great to figure out eventually why this is so low!

PR #9230, opened 2021-04-01
fix API parameter column width [skip github][skip azp]

### LINKED ISSUE #9222: docstring rendering problem
I'm seeing a rendering problem related to responsive layout:

https://user-images.githubusercontent.com/2046265/112759858-31c9cb80-8ff5-11eb-8615-c047bf8e8213.mov

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-python/issues/8876#issuecomment-808922304_

drammock:
closes #9222 

The CSS controlling this is inherited from sphinx.  It works the same on, e.g., [the pandas site](https://pandas.pydata.org/docs/reference/api/pandas.array.html) but is less of a problem there because their default font size is one px smaller than ours, and possibly also because their preferred font is Lato instead of Source Sans Pro (I think Lato is slightly lighter / has tighter spacing). This solution fixes the problem of the colon being wrapped even with full-width windows. It does not fix the problem of `Parameters:` becoming
```
Pa
ra
me
te
rs:
```
when the window is narrowed. For that, adjustments to the theme's media breakpoints would be required, and it should really be fixed in the theme rather than overridden here.

### DIFF
--- a/doc/_static/style.css
+++ b/doc/_static/style.css
@@ -8,6 +8,11 @@
 
 }
 
+/* ************************************************************ Sphinx fixes */
+dl.field-list {
+    grid-template-columns: auto 1fr;
+}
+
 /* ********************************************************** Sphinx-gallery */
 
 /* backreference links: restore hover decoration that SG removes */


agramfort (approved):

larsoner:
Thanks @drammock

PR #9235, opened 2021-04-01
Fix apply function

### LINKED ISSUE #9234: BUG: apply_function arg/kwarg handling
`raw/epochs/evoked.apply_function` all have an `*args, **kwargs` in their signatures. But they don't work the way I think they should / the docs are rather misleading. Preparation:

```py
import os
import mne
from scipy.ndimage import gaussian_filter

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False, preload=False)
raw.crop(tmax=5).load_data()
```

`gaussian_filter1d` requires a `sigma` parameter in addition to the data. This works as expected:
```py
raw.apply_function(gaussian_filter1d, sigma=4)
```
This also works:
```py
kwargs = dict(sigma=4)
raw.apply_function(gaussian_filter1d, **kwargs)
```

This next one is expected to fail **if you already know how argument unpacking works**, and does fail (`4.` is passed as `picks` param of `apply_function`, so you get `picks must be a list of int or list of str, got a data type of float64`):
```py
args = (4.,)
raw.apply_function(gaussian_filter1d, *args)
```
But note that if `args = (4,)` (int instead of float) you get a different error, that `gaussian_filter1d` is missing required argument `sigma`).  From the perspective of a new user, this is really bad: `*args` is in the docstring, so I pass in `*args` and it fails, in 2 different ways depending on what seems like a trivial difference in input.

This also fails (`gaussian_filter1d() got an unexpected keyword argument 'args'`) though might have been expected/tried by a new user:
```py
args = (4.,)
raw.apply_function(gaussian_filter1d, args=args)
```

This also fails due to python syntax restrictions, but might have been expected to work based on the docstring:
```py
raw.apply_function(gaussian_filter1d, verbose=None, *args)
```

So in sum we don't actually support passing `args` to the applied function.  Possible solutions:

1. change function signature from `*args, *kwargs` to `args, kwargs`, and accept only tuples for `args` and only dicts for `kwargs`
2. drop `*args` from the function signature, leave kwarg handling as it is

drammock:
closes #9234 

I've locally updated the test for `epochs.apply_function()` and I'm a bit surprised / wondering if there's another bug in there.  Marking as draft until I've figured it out one way or another.  It seems like when applied channel-wise it gets applied to an ndarray that is `(n_epochs, n_times)` and it's not clear to me that that's the right thing to do (maybe it should really be epoch-wise and channel-wise?)

cc @nordme and @vpeterson

### DIFF
--- a/mne/io/tests/test_apply_function.py
+++ b/mne/io/tests/test_apply_function.py
@@ -7,7 +7,7 @@ import pytest
 
 from mne import create_info
 from mne.io import RawArray
-from mne.utils import logger, catch_logging, run_tests_if_main
+from mne.utils import logger, catch_logging
 
 
 def bad_1(x):
@@ -63,6 +63,3 @@ def test_apply_function_verbose():
         assert out is raw
         raw.apply_function(printer, verbose=True)
         assert sio.getvalue().count('\n') == n_chan
-
-
-run_tests_if_main()
--- a/mne/tests/test_epochs.py
+++ b/mne/tests/test_epochs.py
@@ -3391,7 +3391,7 @@ def test_apply_function():
 
     # apply_function to all channels at once
     def fun(data):
-        """This should reverse channel order without changing values."""
+        """Reverse channel order without changing values."""
         return np.eye(data.shape[1])[::-1] @ data
 
     want = data_epochs[:, ::-1]
--- a/mne/tests/test_epochs.py
+++ b/mne/tests/test_epochs.py
@@ -3380,36 +3380,34 @@ def test_empty_constructor():
     BaseEpochs(info, None, None, event_id, tmin, tmax, baseline)
 
 
-def fun(data_epochs):
-    """Auxiliary function to test_apply_function."""
-    nc = np.size(data_epochs, 1)
-    matrix = np.random.rand(nc, nc)
-    return matrix @ data_epochs
-
-
-def fun_ch(data_epochs):
-    """Auxiliary function to test_apply_function channel wise."""
-    sign = -1
-    return data_epochs * sign
-
-
 def test_apply_function():
     """Test apply function to epoch objects."""
-    data = np.random.rand(2, 10, 1000)
+    n_channels = 10
+    data = np.arange(2 * n_channels * 1000).reshape(2, n_channels, 1000)
     events = np.array([[0, 0, 1], [INT32_MAX, 0, 2]])
-    info = mne.create_info(10, 1000., 'eeg')
+    info = mne.create_info(n_channels, 1000., 'eeg')
     epochs = mne.EpochsArray(data, info, events)
     data_epochs = epochs.get_data()
-    # check apply_function in all channels at the time
-    out = epochs.apply_function(fun, channel_wise=False)
-    assert np.shape(out) == np.shape(data_epochs)
-    # check apply_function channel-wise
-    # change sing 3 first channels
-    picks = [0, 1, 2]
-    no_picks = np.arange(3, 10, 1)
-    out_ch = epochs.apply_function(fun_ch, picks=picks, channel_wise=True)
-    out_ch_data = out_ch._data
-    # check whether those channels have been sign inverted
-    check_picks = (np.sign(out_ch_data[:, picks, :]) == -1).all()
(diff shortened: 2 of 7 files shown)


agramfort (approved):

mne/io/tests/test_apply_function.py
```diff
@@ -63,6 +63,3 @@ def test_apply_function_verbose():
         assert out is raw
         raw.apply_function(printer, verbose=True)
         assert sio.getvalue().count('\n') == n_chan
-
-
-run_tests_if_main()
```
larsoner, line 68:
```
Warning: ./mne/io/tests/test_apply_function.py:10:1: F401 'mne.utils.run_tests_if_main' imported but unused
```

mne/tests/test_epochs.py
```diff
@@ -3380,36 +3380,34 @@ def test_empty_constructor():
...
+
+    # apply_function to all channels at once
+    def fun(data):
+        """Reverse channel order without changing values."""
+        return np.eye(data.shape[1])[::-1] @ data
+
+    want = data_epochs[:, ::-1]
+    got = epochs.apply_function(fun, channel_wise=False).get_data()
+    assert np.array_equal(want, got)
+    assert got.shape == data_epochs.shape
```
drammock, line 3400:
```suggestion
```

drammock: this is redundant (we already check shape using `_check_fun` when it runs, and `np.array_equal` also checks shape)

mne/tests/test_epochs.py
```diff
@@ -3380,36 +3380,34 @@ def test_empty_constructor():
...
+
+    # apply_function channel-wise (to first 3 channels) by replacing with mean
+    picks = np.arange(3)
+    non_picks = np.arange(3, n_channels)
+
+    def fun(data):
+        return np.full_like(data, data.mean())
+
+    out = epochs.apply_function(fun, picks=picks, channel_wise=True)
+
```
drammock, line 3410:
```suggestion
```

mne/tests/test_epochs.py
```diff
@@ -3380,36 +3380,32 @@ def test_empty_constructor():
...
-    assert check_picks and check_nopicks
+
+    # apply_function to all channels at once
+    def fun(data):
+        """Reverse channel order without changing values."""
+        return np.eye(data.shape[1])[::-1] @ data
+
+    want = data_epochs[:, ::-1]
+    got = epochs.apply_function(fun, channel_wise=False).get_data()
+    assert np.array_equal(want, got)
```
larsoner, line 3399: assert_array_equal is our standard way to do this (the failure traceback is better, etc.)

mne/tests/test_epochs.py
```diff
@@ -3380,36 +3380,32 @@ def test_empty_constructor():
...
+    picks = np.arange(3)
+    non_picks = np.arange(3, n_channels)
+
+    def fun(data):
+        return np.full_like(data, data.mean())
+
+    out = epochs.apply_function(fun, picks=picks, channel_wise=True)
+    expected = epochs.get_data(picks).mean(axis=-1, keepdims=True)
+    assert np.all(out.get_data(picks) == expected)
+    assert np.array_equal(out.get_data(non_picks), epochs.get_data(non_picks))
```
larsoner, line 3411: Same for these

mne/tests/test_epochs.py
drammock:
```suggestion
    assert_array_equal(want, got)
```

mne/tests/test_epochs.py
drammock:
```suggestion
    assert_array_equal(out.get_data(non_picks), epochs.get_data(non_picks))
```

drammock:
FYI for @vpeterson and @nordme: after a dev discussion on Discord, it was decided to change the behavior here.  What you implemented for `epochs.apply_function(..., channel_wise=True)` would pass an `n_epochs × n_times` array to the applied function.  What I've now implemented will pass only a 1-D time series (one channel within 1 epoch) to the applied function.

larsoner:
Thanks @drammock !

PR #9237, opened 2021-04-02
DOC: Some improvements to linear_regression()

hoechenberger:


agramfort (approved):
Provided it looks good on artifact

hoechenberger:
Artifact looks good:

https://26690-1301584-gh.circle-artifacts.com/0/dev/generated/mne.stats.linear_regression.html?highlight=linear_regression#mne.stats.linear_regression

mne/stats/regression.py
```diff
@@ -32,28 +32,30 @@ def linear_regression(inst, design_matrix, names=None):
...
     names : array-like | None
-        Optional parameter to name the regressors. If provided, the length must
-        correspond to the number of columns present in regressors
-        (including the intercept, if present).
-        Otherwise the default names are x0, x1, x2...xn for n regressors.
+        Optional parameter to name the regressors (i.e., the columns in the
+        design matrix). If provided, the length must correspond to the number
+        of columns present in design matrix (including the intercept, if
+        present). Otherwise the default names are ``'x0'``, ``'x1'``,
+        ``'x2', …, 'xn'`` for ``n`` regressors.
```
drammock, line 39: chiming in too late here since you already merged, but this is not correct.  If there are `n` regressors and you start at `x0` then your last regressor will be `x(n-1)`.

hoechenberger: good point. I actually didn't check this for logical correctness, simply adapted to formatting of the existing text. Will fix.

hoechenberger: This has been fixed via #9239

PR #9241, opened 2021-04-03
ENH : add new interpolate_mark_bads function

### LINKED ISSUE #9231: mne.preprocessing.interpolate_bads taking list of instances
the issue is the following:

we need to be able to concatenate raws or epochs with different sets of bads.

currently mne.concatenate_raws and mne.concatenate_epochs enforce (for good reasons)
that the list of bads are the same for all instances.

The question is how to simplify the life of users to it's easy to have such "concatenable" instances.

Currently some users interpolate all bads for all instances using interpolate_bads methods.
It works but I think it is not ideal. For me you need the big picture of all instances to make
the best decision. For example I think it's NOT a good idea to systematically interpolate a channel that
is bad for all runs, however interpolation certainly makes sense if one channel is bad for let's say 50%
of the runs.

Suggestion:

`insts = mne.preprocessing.interpolate_bads([inst1, inst2, ...], good_fraction=0.5)`

would just interpolate if a channel is good for at least 50% of the runs and otherwise the
channel would be marked as bad for all instances.

then all interpolate_bads method would use this new public function with `[self]` as parameter.

we'll need to be careful about inplace computation options.

thoughts?

agramfort:
closes #9231


mne/channels/interpolation.py
```diff
@@ -132,11 +132,18 @@ def _interpolate_bads_eeg(inst, origin, verbose=None):
     ----------
     inst : mne.io.Raw, mne.Epochs or mne.Evoked
         The data to interpolate. Must be preloaded.
+    origin : array-like, shape (3,) | str
+        Origin of the sphere in the head coordinate frame and in meters.
+        Can be ``'auto'`` (default), which means a head-digitization-based
+        origin fit.
+    exclude : list | tuple
+        The channels to exclude from interpolation. If excluded a bad
+        channel will stay bad.
```
larsoner, line 141: I would rather kill this docstring entirely rather than have to continue to update it, I think it actually reduces maintenance burden

mne/channels/channels.py
```diff
@@ -1134,6 +1135,9 @@ def interpolate_bads(self, reset_bads=True, mode='accurate',
                 method=dict(meg="MNE", eeg="spline", fnirs="nearest")
 
             .. versionadded:: 0.21
+        exclude : list | tuple
+            The channels to exclude from interpolation. If excluded a bad
+            channel will stay bads.
```
larsoner, line 1140: "will stay *in* bads", or "will stay *bad*"

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,64 @@
+"""Tools for interpolation data."""
```
larsoner, line 1:
```suggestion
"""Tools for data interpolation."""
```

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,64 @@
...
+# Authors: Alexandre Gramfort <alexandre.gramfort@inria.fr>
+
+from itertools import chain
+import numpy as np
+
+from ..epochs import BaseEpochs
+
+
+def interpolate_mark_bads(insts, good_fraction=0., copy=True):
+    """Interpolate or mark bads consistently for a list of data.
```
larsoner, line 12:
```suggestion
    """Interpolate or mark bads consistently for a list of instances.
```

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,64 @@
...
+    Once called on a list of instances, the instances can be concatenated
+    as they will have the same list of bad channels.
+
+    Parameters
+    ----------
+    insts : list
+        The list of instances (Evoked, Epochs or Raw) to consider
+        for interpolation. Each instance should have marked channels.
+    good_fraction : float
+        A float between 0 and 1 that specifies the fraction of time
```
larsoner, line 23:
```suggestion
        A float between 0 (default) and 1 that specifies the fraction of time
```
and then you'll need to re-PEP8-ify it because the line is probably too long

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Authors: Alexandre Gramfort <alexandre.gramfort@inria.fr>
+
+from itertools import chain
+import numpy as np
+
+from ..epochs import BaseEpochs
+
+
+def interpolate_mark_bads(insts, good_fraction=0., copy=True):
```
larsoner, line 11: The default I think should be to always just mark all bad (this seemed to be the consensus when we discussed), so `good_fraction=1.`.

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,64 @@
...
+    Parameters
+    ----------
+    insts : list
+        The list of instances (Evoked, Epochs or Raw) to consider
+        for interpolation. Each instance should have marked channels.
+    good_fraction : float
+        A float between 0 and 1 that specifies the fraction of time
+        a channel should be good to be eventually interpolated for
+        certain instances. For example if 0.5, a channel which is
+        good more than half of the time will be interpolated in the
```
larsoner, line 26: What you wrote here implies `>` but the conditional below (which operates on the opposite) is `<`. So the wording here should be changed to reflect the effective `>=` operation.
```suggestion
        good at least half of the time will be interpolated in the
```

larsoner (commented):
Some preliminary comments

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,63 @@
...
+        as bad. If 1 then channels will never be interpolated and if 0 all bad
+        channels will be systematically interpolated.
+    copy : bool
+        If True then the returned instances will be copies.
+
+    Returns
+    -------
+    insts_bads : list
+        The instances will consistent list of bad channels after
+        potential interpolation.
```
hoechenberger, line 36:
```suggestion
        The instances with with the same list of bad channels after
        potential interpolation.
```

hoechenberger (commented):
I'm not so sure the name for the new function is the best. To me personally, it's a bit confusing or at least not very intuitive.

WDYT about something like `equalize_bads()`, for consistency with `equalize_epoch_counts()` and `equalize_event_counts()`?

agramfort:
why not for equalize_bads

thoughts anyone?

>

drammock:
> why not for equalize_bads 

Summer we're talking about API names, WDYT about changing "good_fraction" to something like "interp_thresh"? (Since it's the**thresh**old below which**interp**olation will happen instead of making bad)

hoechenberger:
@drammock +1!!

agramfort:
rename done

>

larsoner (approved):
LGTM +1 for merge, @drammock feel free to merge if you're happy

mne/preprocessing/interpolate.py
```diff
@@ -0,0 +1,67 @@
...
+        as bad. If 1 then channels will never be interpolated and if 0 all bad
+        channels will be systematically interpolated.
+    copy : bool
+        If True then the returned instances will be copies.
+
+    Returns
+    -------
+    insts_bads : list
+        The instances will have the same list of bad channels after
+        potential interpolation.
```
drammock, line 36:
```suggestion
        The list of instances, with the same channel(s) marked as bad in all of
        them, possibly with some formerly bad channels interpolated.
```

drammock (approved):

drammock:
thanks @agramfort!

PR #9246, opened 2021-04-05
FIX : add_channels now resets picks

### LINKED ISSUE #9214: epochs.add_channels does not update picks
running 

```
import os
import mne

subjects_dir = os.path.join(data_path, "subjects")
data_dir = os.path.join(data_path, 'MEG', 'sample')
raw_fname = os.path.join(data_dir, 'sample_audvis_raw.fif')

raw = mne.io.read_raw_fif(raw_fname, verbose=False)
raw.pick([2, 3, 310])  # take some MEG and EEG
raw.crop(0, 15)  # take litte data

events = mne.make_fixed_length_events(raw, id=3000, start=0)
epochs = mne.Epochs(raw, events, event_id=3000, tmin=0, tmax=3,
                    proj=True, baseline=None, reject=None, preload=True,
                    decim=1)

epochs_final = epochs.copy()
epochs_bis = epochs.copy().rename_channels(lambda ch : ch + '_bis')
epochs_final.add_channels([epochs_bis], force_update_info=True)
epochs_final.drop_channels(epochs.ch_names)
```

on current main branch crashes. It seems that add_channels forgets to update the picks
attribute.

@larsoner wdyt?

agramfort:
closes #9214


mne/channels/channels.py
```diff
@@ -1070,6 +1070,11 @@ def add_channels(self, add_list, force_update_info=False):
             self._read_picks = [
                 np.concatenate([r, extra_idx]) for r in self._read_picks]
             assert all(len(r) == self.info['nchan'] for r in self._read_picks)
+        elif isinstance(self, BaseEpochs):
+            self.picks = np.arange(self._data.shape[1])
+            if hasattr(self, '_projector'):
+                del self._projector
```
larsoner, line 1076: I suspect that this will be a problem with delayed projection. Probably better to do some variant of what's done in `__init__`:
```
        activate = False if self._do_delayed_proj else proj
        self._projector, self.info = setup_proj(self.info, False,
                                                activate=activate)
```
and also add a regression test that would have caught this error (e.g., use delayed projection, get data, add channel, get data again, ensure they match -- which they shouldn't until you make the above suggested change).

agramfort:
@larsoner merge if happy

larsoner:
Okay for me to push a commit to do my suggestion above?

> also add a regression test that would have caught this error (e.g., use delayed projection, get data, add channel, get data again, ensure they match -- which they shouldn't until you make the above suggested change).

agramfort:
go ahead !

>

agramfort:
thx @larsoner !

PR #9248, opened 2021-04-05
expose colorbar label parameter in plot_sensors_connectivity

### LINKED ISSUE #9228: Image title of spectal connectivity
There is a bug on the title of spectral connectivity the colorbar (always shows PLI method). Probably this is a relic of a time when phase lag index was the only measure you offered. See discussion here: https://mne.discourse.group/t/image-title-of-spectal-connectivity/2905

drammock:
closes #9228 

Note that I couldn't figure out how to test this when backend is Mayavi.

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2944,7 +2944,7 @@ def snapshot_brain_montage(fig, montage, hide_sensors=True):
 
 @fill_doc
 def plot_sensors_connectivity(info, con, picks=None,
-                              cbar_label='connectivity'):
+                              cbar_label='Connectivity'):
     """Visualize the sensor connectivity in 3D.
 
     Parameters
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2943,7 +2943,8 @@ def snapshot_brain_montage(fig, montage, hide_sensors=True):
 
 
 @fill_doc
-def plot_sensors_connectivity(info, con, picks=None):
+def plot_sensors_connectivity(info, con, picks=None,
+                              cbar_label='connectivity'):
     """Visualize the sensor connectivity in 3D.
 
     Parameters
@@ -2954,6 +2955,8 @@ def plot_sensors_connectivity(info, con, picks=None):
         The computed connectivity measure(s).
     %(picks_good_data)s
         Indices of selected channels.
+    cbar_label : str
+        Label for the colorbar.
 
     Returns
     -------
@@ -3007,7 +3010,7 @@ def plot_sensors_connectivity(info, con, picks=None):
                              vmin=vmin, vmax=vmax,
                              reverse_lut=True)
 
-    renderer.scalarbar(source=tube, title='Phase Lag Index (PLI)')
+    renderer.scalarbar(source=tube, title=cbar_label)
 
     # Add the sensor names for the connections shown
     nodes_shown = list(set([n[0] for n in con_nodes] +
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2972,7 +2972,7 @@ def plot_sensors_connectivity(info, con, picks=None,
     picks = _picks_to_idx(info, picks)
     if len(picks) != len(con):
         raise ValueError('The number of channels picked (%s) does not '
-                         'correspond the size of the connectivity data '
+                         'correspond to the size of the connectivity data '
                          '(%s)' % (len(picks), len(con)))
 
     # Plot the sensor locations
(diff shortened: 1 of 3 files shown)


larsoner:
> Note that I couldn't figure out how to test this when backend is Mayavi.

We can hopefully trust our abstraction of the backends to take care of this. But in theory you should be able to get to it by going through the Mayavi traits-based internals. Generally these reflect the VTK classes but are Python-ized (snake instead of camel, remove vtk prefix, and using settrs and getters instead of functions), but I think it's okay not to do this here.

mne/viz/_3d.py
```diff
@@ -2943,7 +2943,8 @@ def snapshot_brain_montage(fig, montage, hide_sensors=True):
 
 
 @fill_doc
-def plot_sensors_connectivity(info, con, picks=None):
+def plot_sensors_connectivity(info, con, picks=None,
+                              cbar_label='connectivity'):
```
larsoner, line 2947:
```suggestion
                              cbar_label='Connectivity'):
```
Since it's a title maybe capitolize?

larsoner (commented):
Otherwise LGTM, maybe a quick `latest.inc` update would be good?

drammock:
> in theory you should be able to get to it by going through the Mayavi traits-based internals.

In theory maybe, but not in practice. I tried for about 15 minutes to find the object-based accessor for the scalar bar.  Eventually resorting to the mayavi scene recording feature and click-dragging the bar yielded something like `mlab_interface.children[18].children[0].children[0].children[0].scalar_bar_interactor...` (reproducing from memory, I'm sure that's inaccurate) which I wasn't about to put into a test.

mne/viz/tests/test_3d.py
```diff
@@ -717,14 +717,14 @@ def test_plot_sensors_connectivity(renderer):
...
-
-    plot_sensors_connectivity(info=info, con=con, picks=picks)
+    with pytest.raises(TypeError, match='must be an instance of Info'):
+        plot_sensors_connectivity(info='foo', con=con, picks=picks)
+    with pytest.raises(ValueError, match='does not correspond to the size'):
+        plot_sensors_connectivity(info=info, con=con[::2, ::2], picks=picks)
+
+    fig = plot_sensors_connectivity(info=info, con=con, picks=picks)
+    if hasattr(fig, 'plotter'):
+        assert fig.plotter.scalar_bar.GetTitle() == 'connectivity'
```
larsoner, line 727: > Eventually resorting to the mayavi scene recording feature and click-dragging the bar yielded something like mlab_interface.children[18].children[0].children[0].children[0].scalar_bar_interactor... (reproducing from memory, I'm sure that's inaccurate) which I wasn't about to put into a test.

Indeed that's ugly but there is generally more than one path to a given vtk instance, and the recorder just chooses one based on a long line of children. But this can be simplified and relied upon since Mayavi should construct the pipeline the same way every time. For example in your test this works for Mayavi:
```suggestion
    if renderer._get_3d_backend() == 'pyvista':
        title = fig.plotter.scalar_bar.GetTitle()
    else:
        assert renderer._get_3d_backend() == 'mayavi'
        # the last thing we add is the Tube, so we need to go
        # vtkDataSource->Stripper->Tube->ModuleManager
        mod_man = fig.children[-1].children[0].children[0].children[0]
        title = mod_man.scalar_lut_manager.scalar_bar.title
    assert title == 'Connectivity'
```
I deduced sort of how to get to it from `mne/viz/backends/_pysurfer_mayavi.py` then looking at how `bgcolor` gets used there, and using `--pdb` then `interact` to mess around.

larsoner: (not 100% sure we should bother doing this, but it is one way to get it tested under Mayavi...)

mne/viz/tests/test_3d.py
```diff
@@ -717,14 +717,14 @@ def test_plot_sensors_connectivity(renderer):
...
-        plot_sensors_connectivity(info='foo', con=con,
-                                  picks=picks)
-    with pytest.raises(ValueError):
-        plot_sensors_connectivity(info=info, con=con[::2, ::2],
-                                  picks=picks)
-
-    plot_sensors_connectivity(info=info, con=con, picks=picks)
+    with pytest.raises(TypeError, match='must be an instance of Info'):
+        plot_sensors_connectivity(info='foo', con=con, picks=picks)
+    with pytest.raises(ValueError, match='does not correspond to the size'):
```
larsoner, line 722: This fails locally because there is a typo in the error message code (missing a "to ")

larsoner (approved):

drammock:
all green, in it goes!

PR #9253, opened 2021-04-05
make raw.plot() behave under MPL 3.4.x

drammock:
closes #8955 

I'd rather figure out *why* this problem arose, but the allure of a fairly inexpensive one-line fix was too seductive... we'll see if it slips past the code-quality censors :)


larsoner:
Weird, but if it works why not!

agramfort (approved):
🤦 but ok....

agramfort:
thx @drammock

PR #9259, opened 2021-04-07
epochs-tut

hoechenberger:
The tutorial demonstrating `make_fixed_length_epochs()` is titled "Creating epochs of equal length" in `main`. I always found this confusing, as Eppchs are *always* supposed to have the same duration.

I've change the title and reworked the first paragraph in an attempt to make things a little clearer


hoechenberger:
Doc artifact:
https://26883-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/epochs/plot_60_make_fixed_length_epochs.html

larsoner (approved):

tutorials/epochs/plot_60_make_fixed_length_epochs.py
```diff
@@ -2,11 +2,14 @@
 """
 .. _tut-fixed-length-epochs:
 
-Creating epochs of equal length
-===============================
-
-This tutorial shows how to create equal length epochs and briefly demonstrates
-an example of their use in connectivity analysis.
+Divide continuous data into consecutive epochs
```
drammock, line 5: IMO this should really say something like "Creating equally-spaced epochs". Crossref to #6390, where I proposed to change the name of the function (and was outvoted).  I think ultimately it's the function having a confusing name that is the source of the problem this PR seeks to address.

drammock: More constructively, I don't think "consecutive" is the right word choice here, since the function can create partially-overlapping epochs.  What is special about that function is the equal spacing.

hoechenberger: > "Creating equally-spaced epochs".

Perfect, yes, thanks for this proposal. This is the formulation I'd been looking for.

I'll make this adjustment.

> I think ultimately it's the function having a confusing name that is the source of the problem this PR seeks to address.

No it's actually both, the function name **and** the way parts of the tutorial are phrased.

I'd be +100 on renaming the function. Or at least adding a new one with a more intuitive name. We could deprecate the old one and keep it around for many release cycles if need be. But it should be possible to discover and use this functionality via an approachable, intuitive API – this is currently not the case.

I will open an issue about this.

agramfort (approved):

PR #9261, opened 2021-04-07
MRG, ENH: Epochs.equalize_event_counts() operates on all events by default

### LINKED ISSUE #9260: RFC: Epochs.equalize_event_counts() should operate on all events by default
Currently, `Epochs.equalize_event_counts()` has a positional argument `event_ids`, which expects a list of event names (for some reason called "event ids" in the docstring) whose counts to equalize.

Now a common scenario is that I have, say, 2 or 3 conditions (event types), and before calculating their contrast, I want to equalize their count. Currently, I'll have to do something like this:

```python
epochs.equalize_event_counts(epochs.event_id.keys())
```

I'd like to be able to just do this instead
```python
epochs.equalize_event_counts()
```

To achieve this, the signature of the method should be changed from currently:
```python
    def equalize_event_counts(self, event_ids, method='mintime'):
```

to
```python
    def equalize_event_counts(self, event_ids=None, method='mintime'):
```

and if `event_ids is None`, we simply set `event_ids = list(self.event_id.keys())`

WDYT?

hoechenberger:
Fixes #9260

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -169,7 +169,7 @@ Enhancements
 
 - Add :func:`mne.preprocessing.equalize_bads` to interpolate bad channels in a list of `~mne.Evoked`, `~mne.Epochs` or `~mne.io.Raw` having different sets of bad channels (:gh:`9241` by `Alex Gramfort`_)
 
-- :meth:`mne.Epochs.equalize_event_counts` can now be called without providing a list of event names, and will euqlize the counts of **all** event types present in the `~mne.Epochs` (:gh:`9261` by `Richard Höchenberger`_)
+- :meth:`mne.Epochs.equalize_event_counts` can now be called without providing a list of event names, and will equalize the counts of **all** event types present in the `~mne.Epochs` (:gh:`9261` by `Richard Höchenberger`_)
 
 Bugs
 ~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -169,6 +169,7 @@ Enhancements
 
 - Add :func:`mne.preprocessing.equalize_bads` to interpolate bad channels in a list of `~mne.Evoked`, `~mne.Epochs` or `~mne.io.Raw` having different sets of bad channels (:gh:`9241` by `Alex Gramfort`_)
 
+- :meth:`mne.Epochs.equalize_event_counts` can now be called without providing a list of event names, and will euqlize the counts of **all** event types present in the `~mne.Epochs` (:gh:`9261` by `Richard Höchenberger`_)
 
 Bugs
 ~~~~
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1810,7 +1810,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
             this_epochs.event_id = self.event_id
             _save_split(this_epochs, fname, part_idx, n_parts, fmt)
 
-    def equalize_event_counts(self, event_ids, method='mintime'):
+    def equalize_event_counts(self, event_ids=None, method='mintime'):
         """Equalize the number of trials in each condition.
 
         It tries to make the remaining epochs occurring as close as possible in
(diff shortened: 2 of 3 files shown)


mne/epochs.py
```diff
@@ -1819,52 +1819,76 @@ def equalize_event_counts(self, event_ids, method='mintime'):
...
+        the example above), it is not guaranteed that after equalization the
+        conditions will contribute equally. E.g., it is possible to end up
+        with 70 ``'Nonspatial'`` epochs, 69 ``'Left'`` and 1 ``'Right'``.
+
+        .. versionchanged:: 0.23
+            Default to equalizing all events in the passed instance if no
+            event names were specified explicitly.
         """
-        if len(event_ids) == 0:
+        _validate_type(event_ids, types=(list, dict, None),
```
hoechenberger, line 1885: `dict` is here to ensure backward-compatibility: even some of our tests pass a `dict` to this method.

hoechenberger:
Doc artifact looks good:

https://26892-1301584-gh.circle-artifacts.com/0/dev/generated/mne.BaseEpochs.html?highlight=equalize_event_counts#mne.BaseEpochs.equalize_event_counts

mne/epochs.py
larsoner: Almost everywhere we allow `list` we allow `tuple`, too. I don't see a reason not to here (and it's more backward compatible)

larsoner (commented):
Otherwise LGTM

mne/epochs.py
hoechenberger: Yes actually we should allow `Collection`, is there a way to do that without using the `typing` module?

larsoner: In theory we could use `Iterable` from `collections.abc` or something, but I would probably just go with something like:
```
if x is None:
    x = list(self.event_id)
x = np.asarray(x, dtype=str)
if not x.ndim == 1 and x.size > 0:
    raise ValueError(...)
```
then your `x` is an iterable of str

hoechenberger: I'm now using `Iterable`, LMKWYT!

doc/changes/latest.inc
```diff
@@ -169,6 +169,7 @@ Enhancements
 
 - Add :func:`mne.preprocessing.equalize_bads` to interpolate bad channels in a list of `~mne.Evoked`, `~mne.Epochs` or `~mne.io.Raw` having different sets of bad channels (:gh:`9241` by `Alex Gramfort`_)
 
+- :meth:`mne.Epochs.equalize_event_counts` can now be called without providing a list of event names, and will euqlize the counts of **all** event types present in the `~mne.Epochs` (:gh:`9261` by `Richard Höchenberger`_)
```
drammock, line 172:
```suggestion
- :meth:`mne.Epochs.equalize_event_counts` can now be called without providing a list of event names, and will equalize the counts of **all** event types present in the `~mne.Epochs` (:gh:`9261` by `Richard Höchenberger`_)
```

mne/epochs.py
```diff
@@ -1819,52 +1819,81 @@ def equalize_event_counts(self, event_ids, method='mintime'):
...
-        if len(event_ids) == 0:
+        from collections.abc import Iterable
+        _validate_type(event_ids, types=(Iterable, None),
+                       item_name='event_ids', type_name='list-like or None')
+        if isinstance(event_ids, str):
+            raise TypeError(f'event_ids must be list-like or None, but '
+                            f'received a string: {event_ids}')
+
+        if event_ids is None:
+            event_ids = list(self.event_id.keys())
```
drammock, line 1893:
```suggestion
            event_ids = list(self.event_id)
```

drammock: calling `list()` on a dict already returns a list of the keys

mne/tests/test_epochs.py
```diff
@@ -1914,6 +1914,21 @@ def test_epoch_eq():
     assert_equal(len(epochs['a/x']), 0)
     assert_equal(len(epochs['a/y']), 0)
 
+    # test default behavior (event_ids=None)
+    epochs = Epochs(raw, events, {'a': 1, 'b': 2, 'c': 3, 'd': 4},
+                    tmin, tmax, picks=picks, reject=reject)
+    epochs_1, _ = epochs.copy().equalize_event_counts()
+    epochs_2, _ = (epochs.copy()
+                   .equalize_event_counts(list(epochs.event_id.keys())))
```
drammock, line 1922:
```suggestion
                   .equalize_event_counts(list(epochs.event_id)))
```

drammock (approved):
other than a couple small comments LGTM

mne/tests/test_epochs.py
hoechenberger, line 1922:
```suggestion
    epochs_2, _ = epochs.copy().equalize_event_counts(list(epochs.event_id))
```

agramfort (approved):

drammock:
thanks @hoechenberger!

PR #9262, opened 2021-04-07
MRG, ENH: Add projoff option

larsoner:
I prefer to browse data with projections off when looking for bad channels, and adding this `--projoff` option allows me to use a one-liner command and start immediately arrow-keying around to look for flux jumps:
```
mne browse_raw --allowmaxshield --lowpass 40 --duration 30 --projoff my_raw.fif
```

I also snuck in a fix for the matplotlib deprecations:
```
/usr/lib/python3.9/_collections_abc.py:1023: MatplotlibDeprecationWarning: 
The modification of the Axes.lines property was deprecated in Matplotlib 3.5 and will be removed two minor releases later. Use Artist.remove() instead.
  del self[self.index(value)]
/usr/lib/python3.9/_collections_abc.py:1023: MatplotlibDeprecationWarning: 
The modification of the Axes.texts property was deprecated in Matplotlib 3.5 and will be removed two minor releases later. Use Artist.remove() instead.
  del self[self.index(value)]
```
The 3.0.3 docs have `Artist.remove` so in theory this should be backward compatible with our minimal MPL dep.

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -1385,8 +1385,8 @@ class MNEBrowseFigure(MNEFigure):
         for annot in self.mne.hscroll_annotations[::-1]:
             self.mne.ax_hscroll.collections.remove(annot)
             self.mne.hscroll_annotations.remove(annot)
-        for text in self.mne.annotation_texts[::-1]:
-            self.mne.ax_main.texts.remove(text)
+        for text in list(self.mne.annotation_texts):
+            text.remove()
             self.mne.annotation_texts.remove(text)
 
     def _draw_annotations(self):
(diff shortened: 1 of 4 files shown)


drammock (approved):

larsoner:
@drammock I also want to add an option to press "p" to do the same thing as clicking the "Toggle all" button, okay to add here?

drammock:
> @drammock I also want to add an option to press "p" to do the same thing as clicking the "Toggle all" button, okay to add here?

`p` is already used for toggling annotation snapping

agramfort (approved):

larsoner:
"j" pops up the projection figure, shift-j could be okay for toggling all projectors

larsoner:
Added shift-j (known as "J" in the code) support and pass-through of jJ from the proj fig to the main fig

mne/viz/_figure.py
```diff
@@ -630,7 +630,8 @@ def _keypress(self, event):
         else:
             last_time = self.mne.inst.times[-1]
         # scroll up/down
-        if key in ('down', 'up'):
+        if key in ('down', 'up', 'shift+down', 'shift+up'):
```
larsoner, line 633: Also to make it so that you can navigate full pages in up/down or left/right while holding shift (without needing to release it) I added an alias here for shift+down to move down

drammock:
all green, @larsoner merge if you're finished here.

PR #9268, opened 2021-04-08
ENH: Add support for missing dir

larsoner:
This makes it so we at least get a nice warning about a corrupted file if the tag directory is missing. In real cases there is usually more than this and eventually we hit an error about channel information missing, but the warning now is much more informative than what we had before (something along the lines of "NoneType has no attribute pos").

I also improved an error message for raw concat incompatibility while I was messing around with some incompatible files.

### DIFF
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -2353,13 +2353,13 @@ def _check_raw_compatibility(raw):
     """Ensure all instances of Raw have compatible parameters."""
     for ri in range(1, len(raw)):
         if not isinstance(raw[ri], type(raw[0])):
-            raise ValueError('raw[%d] type must match' % ri)
-        if not raw[ri].info['nchan'] == raw[0].info['nchan']:
-            raise ValueError('raw[%d][\'info\'][\'nchan\'] must match' % ri)
-        if not raw[ri].info['bads'] == raw[0].info['bads']:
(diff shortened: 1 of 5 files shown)


mne/io/base.py
```diff
@@ -2353,13 +2353,13 @@ def _check_raw_compatibility(raw):
...
     for ri in range(1, len(raw)):
         if not isinstance(raw[ri], type(raw[0])):
-            raise ValueError('raw[%d] type must match' % ri)
-        if not raw[ri].info['nchan'] == raw[0].info['nchan']:
-            raise ValueError('raw[%d][\'info\'][\'nchan\'] must match' % ri)
-        if not raw[ri].info['bads'] == raw[0].info['bads']:
-            raise ValueError('raw[%d][\'info\'][\'bads\'] must match' % ri)
-        if not raw[ri].info['sfreq'] == raw[0].info['sfreq']:
-            raise ValueError('raw[%d][\'info\'][\'sfreq\'] must match' % ri)
+            raise ValueError(f'raw[{ri}] type must match')
```
hoechenberger, line 2356: Shouldn't this better raise a `TypeError`?

hoechenberger (approved):
Small nitpick / question, otherwise LGTM

agramfort (approved):

agramfort:
thx @larsoner

PR #9274, opened 2021-04-09
FIX: Prepare for PyVista 0.30.0

GuillaumeFavelier:
On the latest version, the API for `_add_observer()` and `use_ipyvtk` have been modified. This PR tries to catch up to those changes. Changes are also necessary in PyVistaQt.

Follows https://github.com/mne-tools/mne-python/pull/9010#issuecomment-816740584

PR #9275, opened 2021-04-09
Example for using to_data_frame method on EpochsTFR

eioe:
Related to PR #9124

To demonstrate possible usage of the `to_data_frame` method, I adapted an example that uses an `EpochsTFR` instance. 
@dengemann WDYT? 
@cbrnr since I'm mutilating your (very nice) example - you should have a veto before this goes out. 🙂 
Thx!

Also, in the original PR I forgot to do the housekeeping in terms of updating `latest.inc` and `names.inc`. So I'm sneaking this in here.


cbrnr:
Will take a look next week!

eioe:
tbh, I do not understand why it crashes building the docs now and the CircleCI output is unclear to me. Can you maybe help @drammock ?

drammock:
> tbh, I do not understand why it crashes building the docs now and the CircleCI output is unclear to me. Can you maybe help @drammock ?

Looks like you're hit by the problem that @GuillaumeFavelier is trying to fix in #9274

eioe:
ah yes, I see, thanks @drammock

eioe:
@cbrnr fyi, building the docs now worked. If you want you can take a look at the rendered result here: 
https://27276-1301584-gh.circle-artifacts.com/0/dev/auto_examples/time_frequency/plot_time_frequency_erds.html
I'm all open for any kind of suggestions.

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
     fig.colorbar(axes[0].images[-1], cax=axes[-1])
     fig.suptitle("ERDS ({})".format(event))
     fig.show()
+
+##############################################################################
+# Similar as for :class:`~mne.Epochs` objects, we can also export data from
+# :class:`~mne.time_frequency.EpochsTFR` and
+# `~mne.time_frequency.AverageTFR` objects to a
+# :class:`Pandas DataFrame <pandas.DataFrame>`:
```
drammock, line 121:
```suggestion
# Similar to `~mne.Epochs` objects, we can also export data from
# `~mne.time_frequency.EpochsTFR` and `~mne.time_frequency.AverageTFR` objects
# to a :class:`Pandas DataFrame <pandas.DataFrame>`:
```

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
...
+
+df = tfr.to_data_frame()
+df.time = df.time / 1000  # time to s
```
drammock, line 124:
```suggestion
df = tfr.to_data_frame(time_format=None)
```
This should make the time column already in seconds.

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
...
+freq_bands = {'delta': (0.5, 4),
+              'theta': (5, 7),
+              'alpha': (8, 14),
+              'beta': (15, 35)}
+fig, axes = plt.subplots(4, 3, sharey=True, sharex=True)
+for f_idx, f_band in enumerate(['beta', 'alpha', 'theta', 'delta']):
+    data = df.loc[(df.freq >= freq_bands[f_band][0]) &
+                  (df.freq <= freq_bands[f_band][1])]
+    for ch_idx, ch in enumerate(["C3", "Cz", "C4"]):
+        g = sns.lineplot(data=data.reset_index(),
```
drammock, line 140: why do you need `.reset_index()` here?  `.to_data_frame()` should default to an integer index, no?

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -30,11 +30,13 @@
 .. footbibliography::
```
drammock, line 30: move references to the bottom of the file

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
...
+# `~mne.time_frequency.AverageTFR` objects to a
+# :class:`Pandas DataFrame <pandas.DataFrame>`:
+
+df = tfr.to_data_frame()
+df.time = df.time / 1000  # time to s
+df.head()
+
+##############################################################################
+# This allows us to use additional plotting functions like
+# :func:`seaborn.lineplot` to easily plot confidence bands:
```
drammock, line 129: the nested for-loops in this section can be eliminated by using a long-format dataframe, setting the channel and band columns as categorical, and making use of `sns.FacetGrid`. Here's a starting point:

```py
df = tfr.to_data_frame(time_format=None, long_format=True)
df['channel'].cat.reorder_categories(['C3', 'Cz', 'C4'], ordered=True, inplace=True)

freq_bands = {'delta': (0.5, 4),
              'theta': (5, 7),
              'alpha': (8, 14),
              'beta': (15, 35)}

def map_bands(freq):
    for band, (low_lim, high_lim) in freq_bands.items():
        if low_lim <= freq <= high_lim:
            return band

df['band'] = pd.Categorical(
    df['freq'].map(map_bands),
    categories=['beta', 'alpha', 'theta', 'delta'],
    ordered=True)

g = sns.FacetGrid(df, row='band', col='channel', margin_titles=True)
g.map(sns.lineplot, 'time', 'value', 'condition', n_boot=10)
```

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
...
+plt.figlegend(handles=reversed(handles), labels=reversed(labels),
+              loc='lower center', bbox_to_anchor=(0.1, 0.01, 0.9, 0.01),
+              ncol=2, title=None)
+plt.show()
+
+
+##############################################################################
+# Having the data in form of a DataFrame also facilitates subsetting,
+# grouping, and other transforms.
+# Here, we use seaborn to plot average ERDS in the motor-imaginery interval
```
drammock, line 166:
```suggestion
# Here, we use seaborn to plot average ERDS in the motor-imagery interval
```

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +113,85 @@
...
+dat = (df.query('time > 1')
+       .filter(regex=r'condition|epoch|C3|Cz|C4|freq_band')
+       .groupby(['condition', 'epoch', 'freq_band'])
+       .mean()
+       .reset_index()
+       .melt(id_vars=['condition', 'epoch', 'freq_band'],
+             var_name='channel',
+             value_name='mean power'))
+fig, axes = plt.subplots(1, 2, sharey=True)
+for cond, ax in zip(['hands', 'feet'], axes.flatten()):
```
drammock, line 181: here also I think you can avoid the for-loop by using `sns.FacetGrid` (no need to specify `row` this time though)

drammock (changes requested):

examples/time_frequency/plot_time_frequency_erds.py
eioe: yes, you're absolutely right. This is cruft.

cbrnr:
@eioe can you ping me once you've resolved @drammock's comments? Then I will take a look.

examples/time_frequency/plot_time_frequency_erds.py
eioe: Done.

examples/time_frequency/plot_time_frequency_erds.py
eioe: Done.

examples/time_frequency/plot_time_frequency_erds.py
eioe: Thanks! That looks way better and I implemented it that way.

examples/time_frequency/plot_time_frequency_erds.py
eioe: Done.

examples/time_frequency/plot_time_frequency_erds.py
eioe: Thanks! You're right. Also implemented.

eioe:
@drammock I implemented all requested changes. Unfortunately, there's now a merging conflict that I don't understand. Can you point me into a direction how to fix it?

drammock:
> @drammock I implemented all requested changes. Unfortunately, there's now a merging conflict that I don't understand. Can you point me into a direction how to fix it?

probably related to #9292.  All the examples/tutorials were renamed to no longer have filenames that start with `plot_` (they originally did so because of a default setting in sphinx-gallery).  You'll need to put your changes into the newly-renamed file and delete the old `plot_*` file.

examples/time_frequency/plot_time_frequency_erds.py
```diff
@@ -111,3 +109,81 @@
...
+    ordered=True)
+
+g = sns.FacetGrid(df, row='band', col='channel', margin_titles=True)
+g.map(sns.lineplot, 'time', 'value', 'condition', n_boot=10)
+for ax in g.axes.ravel():
+    ax.axhline(0, color='black', linestyle='dashed', linewidth=0.5,
+               alpha=0.5)
+    ax.axvline(0, color='black', linestyle='dashed', linewidth=0.5,
+               alpha=0.5)
+    ax.set(ylim=(None, 1.5))
```
drammock, line 153: these can also use `g.map()` to eliminate the for-loop, for example:

``py
g.map(plt.axhline, y=0, color='black', linestyle='dashed', linewidth=0.5,
      alpha=0.5)
```

Comment (drammock) [reply examples/time_frequency/plot_time_frequency_erds.py:153]: for ylim, you have `g.set(ylim=(None, 1.5))`

Comment (eioe) [reply examples/time_frequency/plot_time_frequency_erds.py]: thanks! Indeed way nicer. Both implemented.

File: examples/time_frequency/time_frequency_erds.py [hunk 9]
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
-               alpha=0.5)
-    ax.set(ylim=(None, 1.5))
+g.map(plt.axhline, y=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
+g.map(plt.axvline, x=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
+g.set(ylim=(None, 1.5))
 g.set_axis_labels("Time (s)", "ERDS (%)")
 g.set_titles(col_template="{col_name}", row_template="{row_name}")
 g.add_legend()
```
drammock, line 155: this one needs a `g.fig.subplots_adjust()`; in [the latest CI rendering](https://27718-1301584-gh.circle-artifacts.com/0/dev/auto_examples/time_frequency/time_frequency_erds.html) the `left` and `bottom` margins need to be bigger, the axis labels are getting cut off.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
+g = sns.FacetGrid(df_mean, col='condition', col_order=['hands', 'feet'],
                   margin_titles=True)
 g = (g.map(sns.violinplot, 'channel', 'value', 'band', n_boot=10,
            palette='deep', order=['C3', 'Cz', 'C4'],
            hue_order=['delta', 'theta', 'alpha', 'beta'])
-      .add_legend(ncol=4, loc=3, bbox_to_anchor=(0.1, 0, 0.9, 0)))
-for ax in g.axes.ravel():
-    ax.axhline(0, color='black', linestyle='dashed', linewidth=0.5,
-               alpha=0.5)
+      .add_legend(ncol=4, loc=3, bbox_to_anchor=(0.1, 0.1, 0.9, 0.1)))
```
drammock, line 175:
```suggestion
      .add_legend(ncol=4, loc='lower center'))
```

right now in [the latest rendering](https://27718-1301584-gh.circle-artifacts.com/0/dev/auto_examples/time_frequency/time_frequency_erds.html) the legend does not end up centered in the figure; I think this will fix that.

drammock (approved):
ok @eioe, I think this is my last round of comments!  This is looking really good.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
-for ax in g.axes.ravel():
-    ax.axhline(0, color='black', linestyle='dashed', linewidth=0.5,
-               alpha=0.5)
-    ax.axvline(0, color='black', linestyle='dashed', linewidth=0.5,
-               alpha=0.5)
-    ax.set(ylim=(None, 1.5))
+g.map(plt.axhline, y=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
+g.map(plt.axvline, x=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
```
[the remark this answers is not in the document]
eioe: way more elegant. implemented!

examples/time_frequency/time_frequency_erds.py
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
-               alpha=0.5)
+      .add_legend(ncol=4, loc=3, bbox_to_anchor=(0.1, 0.1, 0.9, 0.1)))
+
+g.map(plt.axhline, y=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
 g.set_axis_labels("", "ERDS (%)")
 g.set_titles(col_template="{col_name}", row_template="{row_name}")
-plt.show()
+g.fig.tight_layout()
+g.fig.subplots_adjust(left=0.1, right=0.9, top=0.9, bottom=0.3)
```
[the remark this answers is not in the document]
eioe: thx. Good to know! implemented.

examples/time_frequency/time_frequency_erds.py
eioe: well spotted! Thanks for the in-depth review. Fixed.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
            palette='deep', order=['C3', 'Cz', 'C4'],
            hue_order=['delta', 'theta', 'alpha', 'beta'])
-      .add_legend(ncol=4, loc=3, bbox_to_anchor=(0.1, 0, 0.9, 0)))
-for ax in g.axes.ravel():
-    ax.axhline(0, color='black', linestyle='dashed', linewidth=0.5,
-               alpha=0.5)
+      .add_legend(ncol=4, loc=3, bbox_to_anchor=(0.1, 0.1, 0.9, 0.1)))
+
+g.map(plt.axhline, y=0, color='black', linestyle='dashed', linewidth=0.5,
+      alpha=0.5)
```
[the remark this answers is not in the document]
eioe: Done.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -145,43 +145,43 @@ def map_bands(freq):
...
 g.set_axis_labels("", "ERDS (%)")
 g.set_titles(col_template="{col_name}", row_template="{row_name}")
-plt.show()
+g.fig.tight_layout()
+g.fig.subplots_adjust(left=0.1, right=0.9, top=0.9, bottom=0.3)
 
+plt.show()
```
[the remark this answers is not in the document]
eioe: No, you are of course right. Also works without it here. Deleted.

examples/time_frequency/time_frequency_erds.py
eioe: Thx, implemented.

eioe:
@drammock thanks again for the in-depth review and all the constructive suggestions. Learned a lot. 
I implemented all of them. 
Only, the original figures of the tutorial (the TFR plots) are now very small. I tried a lot to change their size (`figsize=(...)`, `fig.set_size_inches(...)`, `fig.tight_layout()`, fig.adjust_subplots(...)`, and various combinations) but without success. So I put it back in it's original state. Help is appreciated.   
Running time of the script on CircleCI seems to be ~24s - if I interpret that correctly. Is that too long for an example? (Original version, before my changes, was at 7s. 😐 )

drammock:
> the original figures of the tutorial (the TFR plots) are now very small.

This was not caused by you; they are small in the current website too: https://mne.tools/dev/auto_examples/time_frequency/time_frequency_erds.html  (they look bigger in [the stable docs](https://mne.tools/stable/auto_examples/time_frequency/plot_time_frequency_erds.html#sphx-glr-auto-examples-time-frequency-time-frequency-erds-py) because that site has a wider width for the main content of the page).

This happens because if a code cell generates more than 1 figure, sphinx-gallery will always put them into two columns. If you want to *force* plots to be full width instead of side-by-side, you can do something like this:

```
fig1 = raw.plot()

fig2 = raw.plot_psd()
```

That approach is harder with the code cell in this example because the 2 figures are generated by iterations of a for-loop, so you can't easily separate out the second plotting call from the first one.

@cbrnr I'm happy with this one, your turn to review/merge :)

cbrnr (changes requested):
Very nice addition and beautiful plots! I have a couple of (minor) comments that I think should be addressed before merging.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -111,3 +109,75 @@
...
+df = tfr.to_data_frame(time_format=None)
+df.head()
+
+##############################################################################
+# This allows us to use additional plotting functions like
+# :func:`seaborn.lineplot` to easily plot confidence bands:
+
+df = tfr.to_data_frame(time_format=None, long_format=True)
+df['channel'].cat.reorder_categories(['C3', 'Cz', 'C4'], ordered=True,
+                                     inplace=True)
```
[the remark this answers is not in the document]
drammock: It is actually already a categorical variable (this happens automatically in our dataframe export code), here we're just changing the order.  it determines what order the columns appear in the seaborn FacetGrid.  To me it made sense to have `Cz` in the middle.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -111,3 +109,75 @@
...
+def map_bands(freq):
+    for band, (low_lim, high_lim) in freq_bands.items():
+        if low_lim <= freq <= high_lim:
+            return band
+
+
+df['band'] = pd.Categorical(
+    df['freq'].map(map_bands),
+    categories=['beta', 'alpha', 'theta', 'delta'],
+    ordered=True)
```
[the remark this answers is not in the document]
drammock: yes!  I knew there was a function for that but I couldn't remember its name.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -57,14 +55,14 @@
 
 picks = mne.pick_channels(raw.info["ch_names"], ["C3", "Cz", "C4"])
 
-# epoch data ##################################################################
+# epoch data #################################################################
```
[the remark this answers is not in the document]
eioe: ha, nice. I now realized that the ruler in my VSC was set to 78 instead of 79. Thx!

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
eioe: That was also my thinking - also for consistency with the time-freq plots above. I suggest keeping it in this order but can of course also take it out. Your call @cbrnr 🙂

examples/time_frequency/time_frequency_erds.py
```diff
@@ -111,3 +109,75 @@
...
+# This allows us to use additional plotting functions like
+# :func:`seaborn.lineplot` to easily plot confidence bands:
+
+df = tfr.to_data_frame(time_format=None, long_format=True)
+df['channel'].cat.reorder_categories(['C3', 'Cz', 'C4'], ordered=True,
+                                     inplace=True)
+
+freq_bands = {'delta': (0.5, 4),
+              'theta': (5, 7),
+              'alpha': (8, 14),
```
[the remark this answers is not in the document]
eioe: yea, I took more or less arbitrary values. I'll change it to 4-7 and 8-13 (13Hz feels more like alpha than beta to me).

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
eioe: Ok, I'm doing this now: 
``` 
freq_bounds = {'_': 0,
               'delta': 3,
               'theta': 7,
               'alpha': 13,
               'beta': 35,
               'gamma': 140}

df['band'] = pd.cut(df['freq'], list(freq_bounds.values()),
                    labels=list(freq_bounds.keys())[1:])
```
wdyt?

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
eioe: Done.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -111,3 +109,75 @@
     fig.colorbar(axes[0].images[-1], cax=axes[-1])
     fig.suptitle("ERDS ({})".format(event))
     fig.show()
+
+##############################################################################
+# Similar to `~mne.Epochs` objects, we can also export data from
+# `~mne.time_frequency.EpochsTFR` and `~mne.time_frequency.AverageTFR` objects
+# to a :class:`Pandas DataFrame <pandas.DataFrame>`:
+
+df = tfr.to_data_frame(time_format=None)
```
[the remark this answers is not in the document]
eioe: Done.

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
eioe: Fixed.

examples/time_frequency/time_frequency_erds.py
```diff
@@ -111,3 +109,75 @@
...
+                  margin_titles=True)
+g = (g.map(sns.violinplot, 'channel', 'value', 'band', n_boot=10,
+           palette='deep', order=['C3', 'Cz', 'C4'],
+           hue_order=['delta', 'theta', 'alpha', 'beta'])
+      .add_legend(ncol=4, loc='lower center'))
+
+g.map(plt.axhline, **axline_kw)
+g.set_axis_labels("", "ERDS (%)")
+g.set_titles(col_template="{col_name}", row_template="{row_name}")
+g.fig.subplots_adjust(left=0.1, right=0.9, top=0.9, bottom=0.3)
```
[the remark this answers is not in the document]
eioe: I decreased the linewidth. Looks better to me than no outline at all. Agree?

eioe:
Thanks @cbrnr for the review. Fixes, comments and suggestions are commited.

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
drammock: `list(freq_bounds.keys())` is the same as `list(freq_bounds)`, FYI.  Otherwise looks fine

drammock:
CircleCI failure is real:

```
/home/circleci/project/doc/auto_examples/time_frequency/time_frequency_erds.rst:503: WARNING: py:obj reference target not found: time_format=None
```

probably you used single-backticks but needed 2 backticks on each side.

examples/time_frequency/time_frequency_erds.py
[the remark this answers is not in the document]
drammock: actually looking at [the rendering on CircleCI](https://27782-1301584-gh.circle-artifacts.com/0/dev/auto_examples/time_frequency/time_frequency_erds.html#sphx-glr-auto-examples-time-frequency-time-frequency-erds-py) I wonder if you got the call to `pd.cut` correct?  there's an empty row at the bottom of the plot and the top row looks really noisy.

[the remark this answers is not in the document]
eioe: argh, that's what you get when "quickly finishing something friday night" ... thanks for spotting. 

---
> there's an empty row at the bottom of the plot

Putting this fixed the empty row for gamma frequencies:
`df['band'] = df['band'].cat.remove_unused_categories()`

---
> the top row looks really noisy

I think, that's because now only two freqs (2 & 3 Hz) go into this plot and the # of reps for the CI bootstrap is very low. It's a tiny bit better with 3 freqs (aka, having the boundary between delta and theta at 4 Hz instead) - but not so much. 
I _think_ that the actual call to to `pd.cut` is ok. I also tested it against the previous mapping method. Results were identical.

larsoner:
All looks good, thanks @eioe !

PR #9276, opened 2021-04-09
fix vline when scrolling in epochs browser

drammock:
this is item 4 from #8761

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -713,8 +713,6 @@ class MNEBrowseFigure(MNEFigure):
                 if self.mne.t_start + self.mne.duration > last_time:
                     self.mne.t_start = last_time - self.mne.duration
                 self._update_hscroll()
-                if key == 'end' and self.mne.vline_visible:  # prevent flicker
-                    self._show_vline(None)
                 self._redraw(annotations=True)
         elif key == '?':  # help window
             self._toggle_help_fig(event)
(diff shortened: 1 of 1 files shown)


jasmainak:
Thanks @drammock for giving a stab at this. When I tried to make a green vertical line by left clicking, this happened:

image

here's how I got to epochs:

```py
import os
import mne

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False).crop(tmax=120)

              'visual/right': 4, 'face': 5, 'button': 32}
epochs = mne.Epochs(raw, events, tmin=-0.2, tmax=0.5, event_id=event_dict,
                    preload=True)
```

drammock:
are you using the macos backend for MPL?

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

jasmainak:
so I had started using 

```sh
$ ipython --matplotlib
```

what I thought was the recommendation but I just checked and it seems the recommendation is:


```sh
$ ipython --matplotlib=qt
```

and with this, it works!

jasmainak:
Actually no, it doesn't work on that backend either. I spoke too soon. It depends where you click. Should I try another backend?

agramfort:
I replicate the problem on my machine.

larsoner:
I cannot replicate on my MacOSX backend, but I do get this:
```
(base) $ MPLBACKEND=MacOSX python -ui rep.py 
  File "/Users/larsoner/opt/miniconda3/lib/python3.8/site-packages/matplotlib/cbook/__init__.py", line 224, in process
    func(*args, **kwargs)
  File "/Users/larsoner/python/mne-python/mne/viz/_figure.py", line 563, in _resize
    old_width, old_height = self.mne.fig_size_px
```
And interaction with this backend is pretty bad. I recommend people use Qt5Agg on macOS, the experience will probably be better.

We should still fix this, but I wish the default backend on macOS was Qt5Agg so we didn't have to keep working around the MacOSX backend flaws in our code :(

drammock:
> I wish the default backend on macOS was Qt5Agg so we didn't have to keep working around the MacOSX backend flaws in our code :(

Another alternative is that one of you folks who has a mac laptop can do all the 2D viz work (or ship a mac laptop to me) :)

PR #9027, opened 2021-03-11
MRG, ENH: Improve Epochs._name and Evoked.comment

hoechenberger:
Rework `Epochs._name` and `Evoked.comment`:

- `*` replaced with `×` in `Epochs._name`
    - old: `0.50 * auditory/left + 0.50 * visual/left`
    - new: `0.50 × auditory/left + 0.50 × visual/left`
- `Evoked.comment` in evoked differences created via `mne.combine_evoked()` now groups conditions via parens:
    - old: `1.500 × 0.50 * auditory/left + 0.50 * visual/left + -1.500 × × 0.51 * auditory/right + 0.49 * visual/right`
    - new: `1.500 × (0.50 × auditory/left + 0.50 × visual/left) - 1.500 × (0.51 × auditory/right + 0.49 × visual/right)`
-  special-casing for `1` and `-1`:
    - old: `-1.000 × 0.50 * auditory/left + 0.50 * visual/left + 1.000 × 0.51 * auditory/right + 0.49 * visual/right`
    - new: `- (0.50 × auditory/left + 0.50 × visual/left) + (0.51 × auditory/right + 0.49 × visual/right)`

Example using `mne.viz.plot_compare_evokeds()`:

old:
main

new:
new

What do you think?

(btw unfortunately `Evoked.comment` cannot contain a proper minus character, as saving will fail then)

PR #9277, opened 2021-04-10
MRG: Return empty list of SSP projectors if no ECG, EOG events were found

hoechenberger:


agramfort (commented):
LGTM

can you use add a line in what API section in what's new?

thx @hoechenberger

hoechenberger:
I'm not sure if we should have a deprecation cycle though, WDYT?

agramfort:
no deprecation is ok

>

doc/changes/latest.inc
```diff
@@ -298,3 +298,5 @@ API changes
 - ``mne.beamformer.tf_dics`` has been deprecated and will be removed in MNE-Python 0.24 (:gh:`9122` by `Britta Westner`_)
 
 - Fitting `~mne.preprocessing.ICA` on baseline-corrected `~mne.Epochs`, and / or applying it on baseline-corrected `~mne.Epochs` or `~mne.Evoked` data will now display a warning. Users are advised to only baseline correct their data after cleaning is completed (:gh:`9033` by `Richard Höchenberger`_)
+
+- `mne.preprocessing.compute_proj_eog` and `mne.preprocessing.compute_proj_ecg` now return empty lists if no EOG or ECG events, respectively, could be found. Previously, we'd return ``None`` in these situations, which does not match the documented behavior of returning a list of projectors (:gh:`9277` by `Richard Höchenberger`_)
```
agramfort, line 302: missing empty line

agramfort (approved):
just a nitpick

doc/changes/latest.inc
agramfort, line 302:
```suggestion
- :func:`mne.preprocessing.compute_proj_eog` and :func:`mne.preprocessing.compute_proj_ecg` now return empty lists if no EOG or ECG events, respectively, could be found. Previously, we'd return ``None`` in these situations, which does not match the documented behavior of returning a list of projectors (:gh:`9277` by `Richard Höchenberger`_)
```

hoechenberger: That's actually not required anymore since to some changes @drammock made to our doc building process a while ago :)

hoechenberger:
ECG tests are failing – of course they are, I only took care of EOG tests. Will fix this ASAP.

doc/changes/latest.inc
larsoner: I think this should go in BUG rather than API because it's a backward-incompatible change with no deprecation cycle

hoechenberger: > because it's a backward-incompatible change with no deprecation cycle

Yeah that's specifically why I put it into the API section …

larsoner: I think we went through this a bit already in https://github.com/mne-tools/mne-python/pull/9027#issuecomment-804348211 and https://github.com/mne-tools/mne-python/pull/8870#discussion_r577136501 but just to be more specific I think our current practice is:

1. "Enhancements": new, non-breaking code changes
2. "Bug": bugfixes, can immediately break existing code without any warning or deprecation cycle, but for a good reason (i.e., the existing code probably did not work correctly)
3. "API": changes to the API that do not break code right away but will probably do so in the next release, and almost always emit a deprecation warning to tell you

Should we add this to the developer guide so that it's explicitly written somewhere?

hoechenberger: Ok I will move this to "Bug", even though I believe since been present for so long it's expected behavior that we're changing, so as a user I would expect this in the "API" section – also considering that bug section is so long that it's impossible to read in its entirety …

larsoner: > for so long it's expected behavior now that we're changing

The point is not about the duration the bug has existed, but rather there is a release cycle to change your code (API) or if it breaks immediately (BUG)

> as a user I would expect this in the "API" section – also considering that the bug section is so long that it's impossible to read it in its entirety …

Again please open an issue about this

hoechenberger: > The point is not about the duration the bug has existed, but rather there is a release cycle to change your code (API) or if it breaks immediately (BUG)

Ok, got it!

hoechenberger:
@larsoner I think this should be good to merge now!

agramfort (approved):
@larsoner merge if happy

larsoner:
Thanks @hoechenberger !

PR #9289, opened 2021-04-13
Add XDF to "Reading EEG data" tutorial

### LINKED ISSUE #5180: Add support for XDF (labstreaminglayer format)
I think this should be added. I am happy to do this. There is already a python loader for the xdf format so it is just a matter of making output structure compliant with mne.

cbrnr:
Closes #5180.

Besides tidying up and improving the docstring a bit, I have added a section on how to import XDF data.

### DIFF
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1,4 +1,7 @@
 % Encoding: UTF-8
+%
+% If available, include a DOI (preferred) *or* a URL for a given reference, but
+# not both, as the DOI turns into a link which is redundant with the URL.
 
 % MNE-C reference
 @article{GramfortEtAl2014,
@@ -518,7 +521,6 @@
   title = {Graph-Based Variability Estimation in Single-Trial Event-Related Neural Responses},
   journal = {{IEEE} Transactions on Biomedical Engineering},
   doi = {10.1109/tbme.2009.2037139},
-  url = {https://doi.org/10.1109%2Ftbme.2009.2037139},
   year = {2010},
   publisher = {Institute of Electrical and Electronics Engineers ({IEEE})},
   volume = {57},
@@ -1306,7 +1308,6 @@
 @article{Pascual-Marqui2011,
   title = {Assessing interactions in the brain with exact low-resolution electromagnetic tomography},
   volume = {369},
-  url = {https://royalsocietypublishing.org/doi/full/10.1098/rsta.2011.0081},
   doi = {10.1098/rsta.2011.0081},
   number = {1952},
   journal = {Philosophical Transactions of the Royal Society A: Mathematical, Physical and Engineering Sciences},
@@ -1980,7 +1981,6 @@
   title = {Structural templates for imaging {EEG} cortical sources in infants},
   volume = {227},
   issn = {1053-8119},
-  url = {http://www.sciencedirect.com/science/article/pii/S1053811920311678},
   doi = {10.1016/j.neuroimage.2020.117682},
   urldate = {2021-01-12},
   journal = {NeuroImage},
@@ -1994,7 +1994,6 @@
   title = {A database of age-appropriate average {MRI} templates},
   volume = {124},
   issn = {1053-8119},
-  url = {http://www.sciencedirect.com/science/article/pii/S1053811915003559},
   doi = {10.1016/j.neuroimage.2015.04.055},
   journal = {NeuroImage},
   author = {Richards, John E. and Sanchez, Carmen and Phillips-Meek, Michelle and Xie, Wanze},
@@ -2062,7 +2061,6 @@
   year = {2002},
   issn = {1053-8119},
   doi = {https://doi.org/10.1006/nimg.2001.1037},
-  url = {https://www.sciencedirect.com/science/article/pii/S1053811901910377},
   author = {Christopher R. Genovese and Nicole A. Lazar and Thomas Nichols},
 }
 
@@ -2082,7 +2080,6 @@
   title = {The {Parallel} {Auditory} {Brainstem} {Response}},
   volume = {23},
   issn = {2331-2165},
-  url = {https://doi.org/10.1177/2331216519871395},
   doi = {10.1177/2331216519871395},
   language = {en},
   journal = {Trends in Hearing},
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -2084,8 +2084,6 @@
   language = {en},
   journal = {Trends in Hearing},
   author = {Polonenko, Melissa J. and Maddox, Ross K.},
-  month = jan,
   year = {2019},
-  note = {Publisher: SAGE Publications Inc},
   pages = {2331216519871395}
 }
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -2054,16 +2054,16 @@
 }
 
 @article{GenoveseEtAl2002,
-title = {Thresholding of Statistical Maps in Functional Neuroimaging Using the False Discovery Rate},
-journal = {NeuroImage},
-volume = {15},
-number = {4},
-pages = {870-878},
-year = {2002},
-issn = {1053-8119},
-doi = {https://doi.org/10.1006/nimg.2001.1037},
-url = {https://www.sciencedirect.com/science/article/pii/S1053811901910377},
-author = {Christopher R. Genovese and Nicole A. Lazar and Thomas Nichols},
+  title = {Thresholding of Statistical Maps in Functional Neuroimaging Using the False Discovery Rate},
+  journal = {NeuroImage},
+  volume = {15},
+  number = {4},
+  pages = {870-878},
+  year = {2002},
+  issn = {1053-8119},
+  doi = {https://doi.org/10.1006/nimg.2001.1037},
+  url = {https://www.sciencedirect.com/science/article/pii/S1053811901910377},
+  author = {Christopher R. Genovese and Nicole A. Lazar and Thomas Nichols},
 }
 
 @article{YaoEtAl2019,
@@ -2076,4 +2076,19 @@ author = {Christopher R. Genovese and Nicole A. Lazar and Thomas Nichols},
   year={2019},
   doi = {10.1007/s10548-019-00707-x},
   publisher={Springer}
+}
+
+@article{PolonenkoMaddox2019,
+  title = {The {Parallel} {Auditory} {Brainstem} {Response}},
+  volume = {23},
+  issn = {2331-2165},
+  url = {https://doi.org/10.1177/2331216519871395},
+  doi = {10.1177/2331216519871395},
+  language = {en},
+  journal = {Trends in Hearing},
+  author = {Polonenko, Melissa J. and Maddox, Ross K.},
+  month = jan,
+  year = {2019},
+  note = {Publisher: SAGE Publications Inc},
+  pages = {2331216519871395}
 }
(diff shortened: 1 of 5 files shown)


tutorials/io/plot_20_reading_eeg_data.py
```diff
@@ -128,31 +130,33 @@
...
 
 .. _import-set:
 
 EEGLAB set files (.set)
 =======================
 
-EEGLAB .set files can be read in using :func:`mne.io.read_raw_eeglab`
-and :func:`mne.read_epochs_eeglab`.
+EEGLAB .set files can be read using :func:`mne.io.read_raw_eeglab` and
+:func:`mne.read_epochs_eeglab`.
```
sappelhoff, line 152: would it be worthwhile to also add `.fdt` here? I know that it's "more or less" no longer being advertised, but EEGLAB still supports exporting set+fdt ... and MNE-Python can read it.

So if a user uses the "search" function of the docs with `fdt`, they should end up here.

tutorials/io/plot_20_reading_eeg_data.py
```diff
@@ -211,15 +243,16 @@
...
 
+
 Reading electrode locations and head shapes for EEG recordings
 ==============================================================
 
-Some EEG formats (EGI, EDF/EDF+, BDF) neither contain electrode location
-information nor head shape digitization information. Therefore, this
-information has to be provided separately. For that purpose all raw instances
-have a :meth:`mne.io.Raw.set_montage` method to set electrode locations.
+Some EEG formats (EGI, EDF/EDF+, BDF) contain neither electrode locations nor
```
sappelhoff, line 250: I think there are more, so let's make sure this is an example, not an exhausitve list.

For instance, I have never seen a BrainVision file with **digitized** coordinates embedded. It's possible (see https://github.com/bids-standard/pybv/issues/44), but I have never seen it - have you? Usually **if** there are coords at all, they are idealized template locations.

```suggestion
Some EEG formats (e.g., EGI, EDF/EDF+, BDF) contain neither electrode locations nor
```

tutorials/io/plot_20_reading_eeg_data.py
```diff
@@ -161,45 +165,73 @@
...
+encounter a problem.
+
+
+XDF data (.xdf, .xdfz)
+======================
+
+MNE-Python does not support loading
+`XDF <https://github.com/sccn/xdf/wiki/Specifications>`_ files out of the box,
+mainly because XDF is an extremely flexible file format which supports signals
+from various modalities recorded with different sampling rates. However, it is
```
sappelhoff, line 213: sounds good in general, but I suggest a rephrasing to make it sound less like: "we don't provide it because it's so great" :slightly_smiling_face: 

```suggestion
MNE-Python does not support loading
`XDF <https://github.com/sccn/xdf/wiki/Specifications>`_ files out of the box,
because the inherent flexibility of the XDF format makes it difficult to provide a 
one-size-fits-all function.
For example, XDF supports signals from various modalities recorded with different
sampling rates. However, it is
```

sappelhoff (approved):
LGTM, three nitpicks

tutorials/io/plot_20_reading_eeg_data.py
cbrnr: Yes, good point!

sappelhoff (approved):

cbrnr:
Any idea why CircleCI is not showing up?

larsoner:
No idea, it doesn't show up in https://app.circleci.com/pipelines/github/mne-tools/mne-python and even when I try to force it using https://app.circleci.com/pipelines/github/mne-tools/mne-python?branch=pull%2F9289 it doesn't show up. Sometimes in the past CircleCI has had problems when people forked and then enabled builds on their fork, did you mess with any of your own GitHub user's CircleCI settings recently?

cbrnr:
Hm, maybe I changed some CircleCI setting accidentally when I was cleaning out (supposedly) unused apps. Do you know where I need to look?

larsoner:
I have CircleCI listed in "Authorized OAuth Apps" in https://github.com/settings/applications. Also you can try navigating to something like https://app.circleci.com/pipelines/github/mne-tools/mne-python/7809/workflows/d85b7b1f-5a4f-4193-bd35-68d6b582e243/jobs/27188/artifacts, I think CircleCI only lets you do it if you have given it some permissions.

cbrnr:
I didn't have CircleCI in my list of authorized apps. Following the link did prompt me to log in, which in turn created the entry again. Thanks @larsoner! I'll create some dummy commit to trigger the build.

cbrnr:
Rendered doc looks good: https://27223-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_20_reading_eeg_data.html#sphx-glr-auto-tutorials-io-plot-20-reading-eeg-data-py

@larsoner @drammock ready for review.

cbrnr:
@drammock not directly related to this PR, but the TOC link "Setting EEG references" never gets highlighted (not even if I click on it). Maybe it's because the corresponding text is too little, but I think it would be nice if it highlighted at least when clicking it.

tutorials/io/plot_20_reading_eeg_data.py
```diff
@@ -161,45 +164,74 @@
...
 
 Persyst EEG data (.lay, .dat)
 =============================
 
-EEG data from the Persyst system can be read in using the
-:func:`mne.io.read_raw_persyst` function.
+EEG data from the Persyst system can be read with
+:func:`mne.io.read_raw_persyst`.
+
+Note that not all subject metadata may be properly imported because Persyst
```
drammock, line 179:
```suggestion
Note that subject metadata may not be properly imported because Persyst
```

tutorials/io/plot_20_reading_eeg_data.py
```diff
@@ -161,45 +164,74 @@
...
+.. code-block:: python
+
+   import pyxdf
+
+   streams, header = pyxdf.load_xdf(filename)  # replace with actual file name
+   data = streams[0]["time_series"].T
+   n_chans = int(streams[0]["info"]["channel_count"][0])
+   sfreq = float(streams[0]["info"]["nominal_srate"][0])
+   info = mne.create_info(n_chans, sfreq, "eeg")
+   raw = mne.io.RawArray(data, info)
```
drammock, line 227: This will not get executed (only formatted/highlighted) and so therefore it's continued correct functioning won't get tested upon doc builds.  (I realize it's not meant to be executed, since you don't provide a filename, but I strongly prefer to avoid code samples in our docs that aren't actually run/tested).  I prefer either:

1. update this code so it works and end the triple-quoted docstring just before this code block so that this gets executed, or
2. follow my previous suggestion (https://github.com/mne-tools/mne-python/issues/5180#issuecomment-810371278) and add this content as a very short I/O example and crossreference it from within this tutorial.

perhaps it goes without saying that I prefer option 2.

cbrnr: Where do you want me to add this as a very short I/O example? Can you point me to the LSL example? The [NEO example](https://mne.tools/dev/auto_examples/io/plot_read_neo_format.html) (I think that's what you meant) creates toy data, so it looks like we will have to add a short XDF sample file somewhere (where?).

drammock: > Where do you want me to add this as a very short I/O example?

in `doc/examples/io/plot_xx_some_new_filename.py`

> Can you point me to the LSL example?

That is what I'm asking you to create.  I was being a bit careless between "LSL" and "XDF" because XDF is the default format for the LSL recorder.

> it looks like we will have to add a short XDF sample file somewhere (where?)

MNE-testing-data?  Or put a small file on OSF?  Or use one that is [already on OSF](https://osf.io/search/?q=xdf&filter=file&page=1)?  @larsoner maybe has stronger feelings than I do about where that should live.

larsoner: If it's not used in testing, don't put it in `mne-testing-data`. If it's small (< 50 MB?) it can live in https://github.com/mne-tools/mne-misc-data . If it's big it should get its own datasets downloader probably.

cbrnr: OK. Does anyone have a small example file? I think it would be good to have something that contains a few seconds of EEG and maybe even some markers. There's a sample file [here](https://github.com/xdf-modules/example-files/blob/master/data_with_clock_resets.xdf), but the EEG data it contains is incorrectly scaled:
Screen Shot 2021-04-13 at 18 18 47

sappelhoff: Not sure if it helps, but I have a normal-sized (I think ~30min recording of a small EEG stream) example file from my "eeg_matchingpennies" dataset: https://osf.io/cj2dr/

see e.g., https://osf.io/agj2q/

that could be truncated, or load as a whole from OSF :man_shrugging: 

you could also get it with datalad, I have a git-annex mirror here: https://github.com/sappelhoff/eeg_matchingpennies

cbrnr: I think this file is too large (>500MB) so we'd need to truncate and then store it somewhere else. However, I don't know how to truncate an XDF file.

cbrnr: @tstenner @cboulay do you happen to have a short demo XDF file, ideally with a few seconds of EEG, which we could use here? Alternatively, do you know how we could create such a file (e.g. crop existing XDF, create XDF with toy data)?

larsoner: In the coming days I can record some using an actiChamp LSL connector / streamer paired with the standard LSL stream recorder (which outputs XDF)

larsoner (approved):
I added this tiny proof-of-concept example:

- https://27538-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/io/plot_20_reading_eeg_data.html#xdf-data-xdf-xdfz
- https://27538-1301584-gh.circle-artifacts.com/0/dev/auto_examples/io/plot_read_xdf.html

+1 for merge from me

examples/io/plot_read_xdf.py
```diff
@@ -0,0 +1,42 @@
...
+
+import pyxdf
+
+import mne
+from mne.datasets import misc
+
+fname = op.join(
+    misc.data_path(), 'xdf',
+    'sub-P001_ses-S004_task-Default_run-001_eeg_a2.xdf')
+streams, header = pyxdf.load_xdf(fname)  # replace with actual file name
```
drammock, line 28:
```suggestion
streams, header = pyxdf.load_xdf(fname)
```

doc/references.bib
```diff
@@ -2077,3 +2075,17 @@ @article{YaoEtAl2019
...
+  title = {The {Parallel} {Auditory} {Brainstem} {Response}},
+  volume = {23},
+  issn = {2331-2165},
+  doi = {10.1177/2331216519871395},
+  language = {en},
+  journal = {Trends in Hearing},
+  author = {Polonenko, Melissa J. and Maddox, Ross K.},
+  month = jan,
+  year = {2019},
+  note = {Publisher: SAGE Publications Inc},
```
drammock, line 2089:
```suggestion
  year = {2019},
```
`month` and `note` fields actually show up in the rendered bibentry.  Hopefully some day we'll be able to provide a custom CSL to sphinxcontrib-bibtex (that doesn't print such things) so we won't have to curate our bibtex fields so carefully.

drammock (approved):
other than 2 small details LGTM

larsoner:
Thanks @cbrnr !

PR #9291, opened 2021-04-13
FIX: Prepare for PyVista 0.30.0 (part 2)

GuillaumeFavelier:
This PR brings the second wave of changes to adapt to PyVista 0.30.0 and PyVistaQt 0.3.0+

ToDo:
- [x] Revert Azure, Circle, `requirements.txt` and `environment.yml` to `pyvista/master` once https://github.com/pyvista/pyvistaqt/pull/95 is merged

Follows https://github.com/mne-tools/mne-python/pull/9274
Related to https://github.com/pyvista/pyvista/pull/1257

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1692,8 +1692,10 @@ class Brain(object):
         # Remove the default key binding
         if getattr(self, "iren", None) is not None:
             try:
+                # pyvista<=0.30.0
                 self.plotter._key_press_event_callbacks.clear()
             except AttributeError:
+                # pyvista>=0.30.0
                 self.plotter.iren.clear_key_event_callbacks()
 
     def _clear_widgets(self):
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1692,7 +1692,7 @@ class Brain(object):
         # Remove the default key binding
         if getattr(self, "iren", None) is not None:
             try:
-                # pyvista<=0.30.0
+                # pyvista<0.30.0
                 self.plotter._key_press_event_callbacks.clear()
             except AttributeError:
                 # pyvista>=0.30.0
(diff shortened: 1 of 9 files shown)


mne/viz/backends/_pyvista.py
```diff
@@ -668,18 +668,13 @@ def _update_picking_callback(self,
...
-            on_button_press
-        )
-        self.plotter.iren.AddObserver(
-            vtk.vtkCommand.EndInteractionEvent,
-            on_button_release
-        )
+        try:
+            add_obs = self.plotter.iren.AddObserver
+        except AttributeError:
+            add_obs = self.plotter.iren.add_observer
```
hoechenberger, line 674: I think it would be good to add a comment why we do this / which line corresponds to which version

GuillaumeFavelier: Very good advice, thank you

mne/viz/_brain/_brain.py
```diff
@@ -1692,8 +1692,10 @@ def _clear_callbacks(self):
         # Remove the default key binding
         if getattr(self, "iren", None) is not None:
             try:
+                # pyvista<=0.30.0
                 self.plotter._key_press_event_callbacks.clear()
             except AttributeError:
+                # pyvista>=0.30.0
```
hoechenberger, line 1698: `== 0.30.0` will match both though?? 😅

GuillaumeFavelier: Whoops

GuillaumeFavelier: It should be fixed in 6f98ca81ce618495e6bf1a3673cf637f64d0d45b

GuillaumeFavelier:
I think Circle does not want to install PyVistaQt master because `dev0` is missing in the version number.

GuillaumeFavelier:
I opened https://github.com/pyvista/pyvistaqt/pull/97#issuecomment-821192485

GuillaumeFavelier:
Okay, with this we should (hopefully) be better prepared for `pyvista 0.30.0`.

mne/viz/_brain/_brain.py
larsoner: should probably be `0.3.0` not `0.30.0`

larsoner:
Pushed a tiny cosmit, thanks @GuillaumeFavelier !

PR #9301, opened 2021-04-15
WIP: Remove codecov token

larsoner:
Thanks to [codecov's oopsie](https://about.codecov.io/security-update/) we need to cycle any env var secrets. For us that's just CODECOV_TOKEN, but they claim you don't need it anymore for Azure. Let's see if we can just remove it. (I already regenerated a new one on the codecov site, so if this fails I'll just change it in the Azure interface.)


larsoner:
Okay looks like the new token is working

PR #9292, opened 2021-04-13
MAINT: Rename examples and tutorials

larsoner:
Adds automated redirect from `plot_*` to all of our examples, and renames them. @drammock for:

> I should note that some, but not all, of the tutorials have had a number added to their filename, to enforce a specific sort order within their category. With the links changing anyway, now would be a good time to add numbers to the rest of them.

I implemented (but did not test) some `custom_redirects` code that you can use. I think you'll want to adapt it to take care of the number renaming. I added the conditional `fname not in needed_plot_redirects.values()` to the auto-`plot_`-redirect code, but you'll need to do a bit of testing to make sure things work. Can you take over and make the necessary code changes and `git mv`s?

closes #9082 
closes #9290

PR #9311, opened 2021-04-18
Introduce lower bounds for ExG peak detection thresholds

### LINKED ISSUE #9273: find_eog_events() finds events when there are literally none
### Problem description & MWE

```python
import os
import mne

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, preload=True)

# Replace all EOG data with 1's, and drop all channels but EOG
raw._data[raw.ch_names.index('EOG 061'), :] = 1.
raw.pick_channels(['EOG 061'])

# EOG detection finds ocular activity
eog_events = mne.preprocessing.find_eog_events(raw, ch_name='EOG 061')
print(eog_events)
```
Output:
```python
Found 560 significant peaks
Number of EOG events detected : 560
[[ 25847      0    998]
 [ 26542      0    998]
 [ 27055      0    998]
 ...
 [191898      0    998]
 [192414      0    998]
 [192599      0    998]]
```

We can even create Epochs …
```python
eog_epochs = mne.Epochs(raw=raw, events=eog_events)
eog_epochs.plot(picks='eog')
```
Screen Shot 2021-04-09 at 10 54 00

I assumed this has something to do with the filtering that happens in `find_eog_events()`, so I changed the filter length:
```python
eog_events = mne.preprocessing.find_eog_events(raw, ch_name='EOG 061',
                                               filter_length='30s')
```
Output:
```python
Found 140 significant peaks
Number of EOG events detected : 140
```
Ok so I thought the problem could be related to small numerical fluctuations at the edges of the filter windows. So I looked at the thresholding we use for peak detection:

```
    thresh : float | None
        Threshold to trigger the detection of an EOG event. This controls the
        thresholding of the underlying peak-finding algorithm. Larger values
        mean that fewer peaks (i.e., fewer EOG events) will be detected.
        If ``None``, use the default of ``(max(eog) - min(eog)) / 4``,
        with ``eog`` being the filtered EOG signal.
```

And with the following, all went as expected:
```python
eog_events = mne.preprocessing.find_eog_events(raw, ch_name='EOG 061',
                                               thresh=1e-15)
```
Output:
```
No significant peaks found
Number of EOG events detected : 0
```

Bingo.

### Proposal
`thresh` should get a lower boundary, so it cannot become "0." and floating point imprecisions won't trigger peak detection.

hoechenberger:
Fixes #9273

We set `100 * eps` as the lower bound.

### DIFF
--- a/mne/preprocessing/_peak_finder.py
+++ b/mne/preprocessing/_peak_finder.py
@@ -50,11 +50,19 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
     s = x0.size
 
     if x0.ndim >= 2 or s == 0:
-        raise ValueError('The input data must be a non empty 1D vector')
+        raise ValueError('The input data must be a non-empty 1D vector')
 
+    eps = 1e2 * np.finfo(x0.dtype).eps
     if thresh is None:
         thresh = (np.max(x0) - np.min(x0)) / 4
+        if thresh < eps:
+            thresh = eps
         logger.debug('Peak finder automatic threshold: %0.2g' % (thresh,))
+    elif thresh < eps:
+        raise ValueError(
+            f'The selected threshold value is too small for peak detection '
+            f'due to inevitable floating-point imprecisions.\nPlease pick a '
+            f'value >= {eps:g} (You passed thresh={thresh})')
 
     assert extrema in [-1, 1]
 
@@ -160,10 +168,6 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
             peak_mags = []
             peak_inds = []
 
-    # Change sign of data if was finding minima
-    if extrema < 0:
-        peak_mags *= -1.0
-
     # ensure output type array
     if not isinstance(peak_inds, np.ndarray):
         peak_inds = np.atleast_1d(peak_inds).astype('int64')
@@ -171,8 +175,11 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
     if not isinstance(peak_mags, np.ndarray):
         peak_mags = np.atleast_1d(peak_mags).astype('float64')
 
-    # Plot if no output desired
-    if len(peak_inds) == 0:
+    # Change sign of data if was finding minima
+    if extrema < 0 and peak_mags.size > 0:
+        peak_mags *= -1.0
+
+    if peak_inds.size == 0:
         logger.info('No significant peaks found')
     else:
         logger.info('Found %d significant peak%s'
--- a/mne/preprocessing/tests/test_ecg.py
+++ b/mne/preprocessing/tests/test_ecg.py
@@ -86,12 +86,21 @@ def test_find_ecg():
     assert 'ECG-SYN' not in raw.ch_names
     assert 'ECG-SYN' not in ecg_epochs.ch_names
 
-    # Test behavior if no peaks can be found -> achieve this by providing
-    # all-zero'd data
-    raw._data[ecg_idx] = 0.
+    # test behavior if no peaks can be found -> achieve this by setting a high
+    # threshold
     ecg_events, _, average_pulse, ecg = find_ecg_events(
-        raw, ch_name=raw.ch_names[ecg_idx], return_ecg=True
+        raw, ch_name=raw.ch_names[ecg_idx], return_ecg=True, qrs_threshold=1e5
     )
+    assert ecg_events.size == 0
+    assert average_pulse == 0
+    assert np.allclose(ecg, np.zeros_like(ecg))
+
+    # test behavior for an entirely flat channel
+    raw._data[ecg_idx] = 0.
+    with pytest.warns(RuntimeWarning, match='first 3 seconds.*flat'):
+        ecg_events, _, average_pulse, ecg = find_ecg_events(
+            raw, ch_name=raw.ch_names[ecg_idx], return_ecg=True
+        )
     assert ecg_events.size == 0
     assert average_pulse == 0
     assert np.allclose(ecg, np.zeros_like(ecg))
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -297,6 +297,8 @@ Bugs
 
 - Fix bug where ``backend='notebook'`` could not be used in :meth:`mne.SourceEstimate.plot` (:gh:`9305` by `Jean-Remi King`_)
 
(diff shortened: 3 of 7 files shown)


mne/preprocessing/tests/test_ecg.py
```diff
@@ -86,12 +86,21 @@ def test_find_ecg():
...
-    raw._data[ecg_idx] = 0.
+    # test behavior if no peaks can be found -> achieve this by setting a high
+    # threshold
     ecg_events, _, average_pulse, ecg = find_ecg_events(
-        raw, ch_name=raw.ch_names[ecg_idx], return_ecg=True
+        raw, ch_name=raw.ch_names[ecg_idx], return_ecg=True, qrs_threshold=1e5
     )
     assert ecg_events.size == 0
     assert average_pulse == 0
     assert np.allclose(ecg, np.zeros_like(ecg))
```
hoechenberger, line 96: I'm surprised this one passes …

hoechenberger:
I forgot to update `test_peak_finder.py`, will do

hoechenberger:
@drammock @larsoner It seems there's a problem with Circle that's unrelated to my PR, could you please have a look?

drammock:
Looks like a deprecation warning from jinja, which will need to be fixed upstream in Sphinx.

hoechenberger:
So should I pin our jinja to an older version perhaps?

drammock:
> So should I pin our jinja to an older version perhaps?

we don't directly depend on jinja; it's a sphinx dependency.  I think the right thing to do is to open an issue upstream in the sphinx repo, and add this particular deprecation warning to the `reset_warnings()` function in `doc/conf.py` so that in the meantime it won't affect our doc builds (I think that's the right place, as opposed to `mne/conftest.py`... I think conftest is for warnings generated by our tests, and `reset_warnings()` in `conf.py` is for warnings generated during doc builds).  @larsoner  can correct me if I'm wrong.

larsoner:
That's correct!

mne/preprocessing/_peak_finder.py
```diff
@@ -50,11 +50,26 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
     s = x0.size
 
     if x0.ndim >= 2 or s == 0:
-        raise ValueError('The input data must be a non empty 1D vector')
+        raise ValueError('The input data must be a non-empty 1D vector')
+
+    try:
+        eps = 1e2 * np.finfo(x0.dtype).eps
+    except ValueError:  # int dtypes are exact
+        eps = 0.
```
agramfort, line 58: I don't like try except blocks. They are too implicit. Here i would say that either we support integer inputs and you should check that dtype is int-like or if not you should cast to a float with the same nbytes. See

https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/validation.py#L173

how it's done in sklearn.

hoechenberger: I'm now simply casting unless it's already a float, see 20361b4

You okay with that?

mne/preprocessing/_peak_finder.py
```diff
@@ -52,10 +52,11 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
     if x0.ndim >= 2 or s == 0:
         raise ValueError('The input data must be a non-empty 1D vector')
 
-    try:
-        eps = 1e2 * np.finfo(x0.dtype).eps
-    except ValueError:  # int dtypes are exact
-        eps = 0.
+    # cast to float, but only if we must
+    if not isinstance(x0, (np.single, np.double, np.longdouble)):
```
agramfort, line 56:
```suggestion
    if x0.dtype.kind in 'uib':   # integer type
```

doc/conf.py
```diff
@@ -791,6 +791,15 @@ def reset_warnings(gallery_conf, fname):
         'ignore', message="can't resolve package from", category=ImportWarning)
     warnings.filterwarnings(
         'ignore', message='.*mne-realtime.*', category=DeprecationWarning)
+    warnings.filterwarnings(
```
larsoner, line 794: This was errantly duplicated

Also perhaps you don't need these changes anymore? The latest CircleCI PR doc builds have been green at least

hoechenberger: I reverted the changes and now we get the deprecation warnings again and an exception :(

drammock: looks like you reverted too much!  the jinja glue line was supposed to stay...  it's just the accidentally dupllicated mne-realtime line that needed reverting

larsoner: I thought even the jinja-glue wouldn't be needed since builds are passing on other PRs. But maybe this PR changes something that makes it necessary? Or the caching is making different builds use different versions of Sphinx or something?

drammock: version of sphinx is the same between this failed circleCI build and the successful one on #9316 (both use `4.0.0b1`).  Jinja is different (`3.0.0rc1` here versus `3.0.0a1` there)

mne/preprocessing/_peak_finder.py
```diff
@@ -50,11 +50,26 @@ def peak_finder(x0, thresh=None, extrema=1, verbose=None):
...
     if thresh is None:
         thresh = (np.max(x0) - np.min(x0)) / 4
+        if thresh < eps:
+            thresh = eps
         logger.debug('Peak finder automatic threshold: %0.2g' % (thresh,))
+    elif thresh < eps:
+        raise ValueError(
+            f'The selected threshold value is too small for peak detection '
+            f'due to inevitable floating-point imprecisions.\nPlease pick a '
+            f'value >= {eps:g} (You passed thresh={thresh})')
```
larsoner, line 72: I am not convinced this is right. We have to keep in mind `eps` is relative to 1., it's not an absolute minimum delta between floats (that would be `finfo(x0.dtype).tiny`, and even that's before going denormalized). For example consider this on `main`, which seems correct:
```
>>> import mne
>>> import numpy as np 
>>> x = np.full(1000, 1e-20)
>>> x[500] = 2e-20
>>> mne.preprocessing.peak_finder(x)
(array([500]), array([2.e-20]))
```
But we have
```
>>> (np.max(x) - np.min(x)) / 4
2.5e-21
>>> 1e2 * np.finfo(x.dtype).eps
2.220446049250313e-14
```
So this PR will find nothing significant.

For EEG/EOG data in V, this might be okay. For data in T or T/M, these hang out around 1e-12 so this can become problematic.

doc/conf.py
drammock: the jinja error is happening on #9292 now also.

larsoner: Feel free to port the fix over. Personally I would do it just by adding another line to the `DeprecationWarnings` list a few lines above

larsoner: (around line 770 of conf.py)

larsoner:
Moving to 0.24 so that whatever solution we come up with we have time to test properly before release

PR #9323, opened 2021-04-20
MAINT: Check name URLs

larsoner:
Currently we don't check anyone's `names.inc` URLs but we should. This adds a little extension that outputs a `generated/_names.rst` when using the linkcheck builder.

I see some failures locally with the check so I'll push that first to make sure we see it on CIs, then I'll add another commit to fix them.


larsoner:
@agramfort feel free to merge if you're happy with the URLs. I basically tried to find a personal url; then a current research-group-page url; then a GitHub URL.

drammock (approved):
looks reasonable.  I didn't know mark had moved to JPL!

PR #9324, opened 2021-04-20
MAINT: Mark more tests as slow for macOS

larsoner:
On most PRs we end up waiting for the macOS CI to finish. This is probably really because it takes 20 minutes (!) to set up the conda env, but we might as well also cut our testing time down from the 30 minutes that it currently runs there. This PR does it without sacrificing much testing by skipping some tests that are mostly redundant once we've run them on one OS, so let's let Linux do the heavy lifting and make macOS and Windows (which both `pytest -m "not slowtest"`) a bit faster.


agramfort (approved):

agramfort:
thx @larsoner

PR #9327, opened 2021-04-21
WIP, RFC: Improve readability of sensor names in plot_sensors()

hoechenberger:
Instead of a long textual description, let me just share two cropped screenshots

`main`:
Screen Shot 2021-04-21 at 10 09 18

This branch:
Screen Shot 2021-04-21 at 10 08 59

It's super crude in its current state but it already improves usability for me.

Would like to hear your input & ideas on **if** and **how** to move forward with this?

Thanks!

-----

Code for MWE:
```python
from pathlib import Path
import matplotlib
import mne

matplotlib.use('Qt5Agg')

sample_dir = Path(mne.datasets.sample.data_path())
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname, preload=False)
raw.plot_sensors()
```

### DIFF
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -996,7 +996,8 @@ def _onpick_sensor(event, fig, ax, pos, ch_names, show_names):
     if len(this_pos) == 3:
         ax.text(this_pos[0], this_pos[1], this_pos[2], ch_name)
     else:
-        ax.text(this_pos[0], this_pos[1], ch_name)
+        ax.text(this_pos[0] + 0.01, this_pos[1], ch_name, va='center_baseline',
+                backgroundcolor='whitesmoke')
     fig.canvas.draw()
 
 


mmagnuski:
Nice! Although for `show_names=True` this could lead to most channel markers be obscured by the text boxes. Maybe some transparency for the text box would be ok?

hoechenberger:
> Nice! Although for `show_names=True` this could lead to most channel markers be obscured by the text boxes. Maybe some transparency for the text box would be ok?

Oh, I hadn't even thought about this case. I only played with "interactive" mode. 
Yes some sort of transparency could be good. I will look into this.

I was also thinking that it could be cool to display the names on hover. But that's another discussion & PR. :)

mmagnuski:
given that you modify only `_onpick_sensor` - it should not affect show_names=True, I think. Sorry, I was previously on my phone and didn't read the code changes.

hoechenberger:
But you're bringing up another important issue anyway. This is close to unusable:




But also something for another PR :)

drammock:
I think I would try placing the labels below their dot, centered (or maybe above & centered, if the mouse cursor gets in the way).  For the single-sensor case I think this will obviate the need for a (maybe semi-transparent) box.  For the `show_names=True` case I think it will improve things somewhat but not fix them entirely, as there will still be some text-on-text issues even if the text-on-dot issues are resolved.

A more involved solution would be to expose a color parameter.  The color param would make it possible to make the dots light grey and the text black for example.  Combined with changing the label position to above/below I think this would probably resolve the issue fully.

Exposing a font size parameter might also help, though I think better to add just 1 param if we can get away with it, and dot color I think is more useful by itself (in terms of legibility) than font size by itself.

Another possibility (even more involved) is to allow `color` to take a list/array of colors if you wanted to color-code by ROI or something.  But in this case users might also want the *text* to be different colors for different ROIs, and things start to get complicated.  For this more complicated case I'd say YAGNI.

PR #9332, opened 2021-04-21
maint: fix scraper test

drammock:
fixes a failing test of our sphinx-gallery scraper


mne/tests/test_report.py
```diff
@@ -605,7 +605,6 @@ def test_scraper(tmpdir):
     rst = scraper(block, block_vars, gallery_conf)
```
drammock, line 605: this line is the one that creates `op.join(app.builder.outdir, 'auto_examples')` (which then leads to a `FileExists` error on the deleted line)

larsoner:
Thanks @drammock !

PR #9337, opened 2021-04-22
MRG, MAINT: Better split of reqs

larsoner:
Inspired by #9192 I think we need a little bit more granularity in our test infrastructure, so:

1. Remove `ipytest`, it was errantly added
2. Make SG dep on the latest dev version (see https://github.com/sphinx-gallery/sphinx-gallery/issues/817#issuecomment-824925909, I think we should always use latest SG dev to catch bugs etc. -- we are good early users)
3. Split off some reqs into `requirements_testing_extra.txt`
4. Add comments to our `requirements_*` files to make it clearer why we have/need them
5. Update `azure-pipelines.yml` to include the `requirements_testing_extra.txt` installs

Overall I hope this makes things a bit more DRY and well defined. @drammock I know you've been disappointed with the structure of our requirements, can you see if you think this at least helps instead of making things worse?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ include LICENSE.txt
 include SECURITY.md
 include requirements.txt
 include requirements_testing.txt
+include requirements_testing_extra.txt
 include requirements_doc.txt
 include mne/__init__.py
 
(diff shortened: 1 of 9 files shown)


drammock (approved):
LGTM, seems like an improvement

PR #9340, opened 2021-04-23
DICS beamformer: deprecate the use of complex filter

britta-wstnr:
As discussed in #9215, this PR implements the deprecation of `real_filter=False` in `make_dics`. Instead of using a complex filter by default, the filter should be real-valued.


larsoner (approved):
LGTM +1 for merge

hoechenberger (commented):
Thanks @britta-wstnr! Could you please add an entry to `latest.inc`, `API` section?

hoechenberger:
Thanks @britta-wstnr!

PR #9345, opened 2021-04-23
MRG: Some tutorial tweaks

drammock:
Here are some tweaks to a couple tutorials, and an overhaul of the EEG ERP one.  Marking as draft because the EEG ERP one takes too long on my system, so I'm sure Circle will complain, any suggestions to speed it up are welcomed.  I also wonder if there's a better dataset to use than `sample`... thoughts (@hoechenberger maybe?)


larsoner:
24 sec on circle is longer than ideal but I can live with it if there isn't a good way to speed it up. There is this error though
```
/home/circleci/project/doc/auto_tutorials/evoked/30_eeg_erp.rst:724: WARNING: py:obj reference target not found: matplotlib
```

drammock:
> py:obj reference target not found: matplotlib

Dang, `mod` didn't work either. I'll just remove the crossref. 

There probably are ways to speed it up, I just didn't have time to try yet

larsoner:
Thanks @drammock !

Issue #9308, opened 2021-04-16
SNIRF loader not working without the optional sourcePos3D


HanBnrd:
#### Describe the bug
The function [*read_raw_snirf*](https://mne.tools/dev/generated/mne.io.read_raw_snirf.html#mne.io.read_raw_snirf) is not working if the [optional](https://github.com/fNIRS/snirf/issues/41#issuecomment-620159229) field [sourcePos3D](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsiprobesourcepos3d) does not exist in the SNIRF file.

#### Steps to reproduce
Given a *nback.snirf* SNIRF file with only sourcePos2D and no sourcePos3D:
```python
from mne.io import read_raw_snirf

snirf_intensity = read_raw_snirf('nback.snirf')
```

#### Expected results
```
Loading nback.snirf
```

#### Actual results
```
Loading nback.snirf
Traceback (most recent call last):
  File "...\snirf.py", line 3, in <module>
    snirf_intensity = read_raw_snirf('nback.snirf')
  File "...\mne-python\mne\io\snirf\_snirf.py", line 43, in read_raw_snirf
    return RawSNIRF(fname, preload, verbose)
  File "<decorator-gen-248>", line 24, in __init__
  File "...\mne-python\mne\io\snirf\_snirf.py", line 119, in __init__
    assert len(sources) == srcPos3D.shape[0]
IndexError: tuple index out of range
```

#### Additional information
This may be complex to solve as the [*beer_lambert_law* function](https://github.com/mne-tools/mne-python/blob/main/mne/preprocessing/nirs/_beer_lambert_law.py#L18-L61) requires the 3D positions of optodes.

HanBnrd:
Ping @rob-luke

rob-luke:
I think this is too complex to solve as we need 3d positions to calculate distances for beer lambert law.

HanBnrd:
> I think this is too complex to solve as we need 3d positions to calculate distances for beer lambert law.

I agree, I think it's not a priority now but I would keep this issue open in case we ever modify the Beer Lamber law implementation to use source-detector distances instead of the 3D positions.

This may become a bigger issue that we may want to address if every vendor decides to export data with only 2D positions and no sourcePos3D.

rob-luke:
> if every vendor decides to export data with only 2D positions

In that case I will move to M/EEG research 😆

kdarti:
[...]

PR #9347, opened 2021-04-24
MRG: Add SNIRF support for 2D locations and NIRx files

### LINKED ISSUE #9338: Add support for NIRx fNIRS data in SNIRF format
We currently have a SNIRF reader that works with files produced using the society for fNIRS approved generator/converter.

However, there is quite some flexibility in the specification. Further, we only support a subset of the specifiation (for example we do not support anything other than continuous wave spectroscopy, and we dont support 2D locations).

NIRx now export their data as SNIRF. This is great. However, they do not export some fields that are optional in the spec. And currently our reader expects these data fields. So the reader needs to be modified to be forgiving if these optional fields are missing, or it needs to throw a usefull error.

I have recorded some data with a NIRSport2 device in snirf format. I will add this to the test data repository and I will then update our reader to gracefully handle these missing fields.

My initial investigation indicates that the fields we need to not expect are
* /nirs(i)/probe/sourceLabels
* /nirs(i)/probe/detectorLabels

so this shouldnt be too much work for me.

I will also use this oppurtunity to all other fields we expect are "required" and try not to rely on optional fields.

#### Additional comments

I am writing this here so everyone knows and does not try to duplicate effort.

I will also address https://github.com/mne-tools/mne-nirs/issues/257

FYI @robquant @HanBnrd @larsoner

rob-luke:
#### Reference issue
Closes #9338 mne-tools/mne-nirs#257 #9308.

This PR enables reading of SNIRF files from the NIRx vendor and it also enables reading of SNIRF files with 2D locations from the SfNIRS team.

Support for 2D locations was enabled by setting the z coordinate to 0.

Support for NIRx was enabled by not relying on the `sourceLabels` field in SNIRF. Which was optional in the spec, so I shouldn't have expected it to be there.

#### Additional information
People may wish to support no sensor location info, but they can do that in a future PR.

It may also be possible to do some fancy 2D to 3D conversion for data, but that would be significant work.

I wanted to use an exact copy of some code, so I turned it in to a function called `_validate_nirs_info`

Specifically validating and testing Artinis files will have to wait till someone can provide test data. However, I think this will achieve the 2D support they require.

### DIFF
--- a/tools/get_testing_version.sh
+++ b/tools/get_testing_version.sh
@@ -1,8 +1,8 @@
 #!/bin/bash -ef
 
 TESTING_VERSION=`grep -o "testing='[0-9.]\+'" mne/datasets/utils.py | cut -d \' -f 2 | sed "s/\./-/g"`
-# This can be incremented to start fresh when the cache misbehaves
-TESTING_VERSION=${TESTING_VERSION}-1
+# This can be incremented to start fresh when the cache misbehaves, e.g.:
+# TESTING_VERSION=${TESTING_VERSION}-1
 if [ ! -z $GITHUB_ENV ]; then
 	echo "TESTING_VERSION="$TESTING_VERSION >> $GITHUB_ENV
 elif [ ! -z $AZURE_CI ]; then
--- a/tools/get_testing_version.sh
+++ b/tools/get_testing_version.sh
@@ -1,6 +1,8 @@
 #!/bin/bash -ef
 
 TESTING_VERSION=`grep -o "testing='[0-9.]\+'" mne/datasets/utils.py | cut -d \' -f 2 | sed "s/\./-/g"`
+# This can be incremented to start fresh when the cache misbehaves
+TESTING_VERSION=${TESTING_VERSION}-1
 if [ ! -z $GITHUB_ENV ]; then
 	echo "TESTING_VERSION="$TESTING_VERSION >> $GITHUB_ENV
 elif [ ! -z $AZURE_CI ]; then
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -187,6 +187,9 @@ Enhancements
 
 - Add support for setting montages on fNIRS data, with built in standard montages for Artinis OctaMon and Artinis Brite23 devices (:gh:`9141` by `Johann Benerradi`_, `Robert Luke`_ and `Eric Larson`_)
 
+- Enable support for reading SNIRF files with 2D optode positions (:gh:`9347` `Robert Luke`_)
+
+
 Bugs
 ~~~~
 - Fix bug with :func:`mne.viz.plot_evoked_topo` where set ylim parameters gets swapped across channel types. (:gh:`9207` by |Ram Pari|_)
--- a/mne/channels/montage.py
+++ b/mne/channels/montage.py
@@ -735,16 +735,9 @@ def _set_montage_fnirs(info, montage):
     stored. This function modifies info['chs'][#]['loc'] and info['dig'] in
     place.
     """
-    from ..preprocessing.nirs import (_channel_frequencies,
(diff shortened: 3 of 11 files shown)


rob-luke:
This will fail until https://github.com/mne-tools/mne-testing-data/pull/84 is merged.

rob-luke:
@larsoner @agramfort @drammock Could you please review and merge if happy. I expect CIs to go green

rob-luke:
>  I expect CIs to go green

Famous last words, I forgot to update the testing hash and version

rob-luke:
Ok there is some weird CI behaviour going on. The new files are being found and read in the Mac and windows CI tests (I opened the logs and checked all snirf tests were run and passed). But the linux CIs cant find the files.

Could it be a caching thing?

rob-luke:
Ok I'm truly stumped here. @larsoner could you give me a clue when you have a chance?

rob-luke:
I just redid the experiment for our tutorial at https://mne.tools/stable/auto_tutorials/preprocessing/plot_70_fnirs_processing.html

I measured 25 minutes worth of data on myself with the NIRSport2 device and it was saved in the SNIRF format. The tutorial above worked simple by changing `read_raw_nirx` to `read_raw_snirf`, so I am confident the reader is now working correctly. We just need to squish these pesky CI errors.

tools/get_testing_version.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash -ef
 
 TESTING_VERSION=`grep -o "testing='[0-9.]\+'" mne/datasets/utils.py | cut -d \' -f 2 | sed "s/\./-/g"`
+# This can be incremented to start fresh when the cache misbehaves
+TESTING_VERSION=${TESTING_VERSION}-1
```
rob-luke, line 5: bash magic, thanks @larsoner

larsoner: Really StackOverflom magic :)

https://stackoverflow.com/questions/63521430/clear-cache-in-github-actions#answer-64819132

larsoner:
I can replicate the failure locally:
```
mne/io/snirf/tests/test_snirf.py:44: in test_basic_reading_and_min_process
    raw = read_raw_snirf(fname, preload=True)
mne/io/snirf/_snirf.py:47: in read_raw_snirf
    return RawSNIRF(fname, preload, verbose)
<decorator-gen-250>:24: in __init__
    ???
mne/io/snirf/_snirf.py:78: in __init__
    fname = _check_fname(fname, 'read', True, 'fname')
mne/utils/check.py:178: in _check_fname
    raise FileNotFoundError(f'{name} does not exist: {fname}')
E   FileNotFoundError: fname does not exist: /home/larsoner/python/MNE-testing-data/snirf/SfNIRS/snirf_homer3/1.0.3/snirf_1_3_nirx_15_2_recording_w_short.snirf
```
Is it perhaps a real path problem that only Linux tests because on other CIs there is a `slowtest` marker that makes it get skipped for example?

larsoner:
@rob-luke I take it you're either on macOS or Windows, which both play more loosely with case than Linux...?

rob-luke:
@larsoner Mac at home, windows at work. And currently very embarrassed.

larsoner:
I think we should all blame the OS and not the user here :)

rob-luke:
Im actually quite surprised that they handle case differently, you learn something every day.

larsoner:
Thanks @rob-luke !

larsoner:
> Closes #9338 mne-tools/mne-nirs#257 #9308.

FYI @rob-luke another nugget for the day, these need to be on separate lines -- only the first closed automagically on merge

rob-luke:
Thanks @larsoner. And I am pleased as squeezed this in before the release. Cheers

PR #9348, opened 2021-04-24
WIP: Add NIRSport support

rob-luke:
I am trying to determine how to use the GitHub magic to do some work on #7936. I tried pushing to https://github.com/uge-lescot/mne-python/pull/1 but made a mess of it. Now I am trying this.


rob-luke:
@rderollepot thanks for your support with this PR (and I apologise for jumping between branches, hopefully someone better at git can merge this back in to your PR). Im going to try and explain my understanding below and see if we can debug my issue. I will explain things you are already aware of, so that others (@larsoner) can also follow along.

I have added some tests for the files which you contributed and diligently documented in https://github.com/mne-tools/mne-testing-data/pull/78

However, my tests are failing. When I read the data it appears to not contain any NaNs in the channel combinations specified in the montage. I have confirmed that your code does read the correct wave file based on the saturation argument 👍 . And I have confirmed that NaNs exist in the wl1 file 👍  and not in the nosatflag file 👍 . 

The data on disk (specifically the files `nirx_15_3_recording_w_full_saturation` and `nirx_15_3_recording_w_occasional_saturation`) contains the measurements between all source and detector pairs. However, only certain pairs are valid, and these are specified in your montage. When I load the raw wl1 data file with all combinations of channels it has NaNs 👍 but when the code selects/subsets just the channels (optode pairings of interest) in your montage, all those channels are NaN free. Based on your great notes in  https://github.com/mne-tools/mne-testing-data/pull/78 I expected that at least the data for detector 2 would contain NaNs. Is this your understanding too?

Is it possible that in the test data the channels in your montage never saturated? I dont know the software/hardwware you used, so Im not sure if a big red light flashes, or if you just have to guess. I'd really like to ensure there is a test that returns Nans so we can ensure the saturation flag works. To make things move quicker, would you like to email me another file (doesn't have to be a test file like before, just anything you have where you are certain it saturates) so I can confirm the behaviour in this PR. Im totally open to other suggestions too.

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,47 @@
...
+    nans = np.any(np.isnan(data), axis=0)
+    starts, stops = _mask_to_onsets_offsets(nans)
+
+    if len(starts) > 0:
+        starts, stops = np.array(starts), np.array(stops)
+        onsets = (starts + raw.first_samp) * sampling_duration
+        durations = (stops - starts) * sampling_duration
+    else:
+        warnings.warn("The dataset you provided does not contain 'NaN' "
+                      "values. No annotation were made.")
```
larsoner, line 39: Use our internal `from ..utils import warn ... warn(...)` because it does some nice `stacklevel` stuff automagically

Also "annotation" -> "annotations"

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,47 @@
...
+        return
+
+    if annot is None:
```
larsoner, line 42: `annot` is never `None` anymore, we changed this a version or two ago, e.g.:
```
>>> import mne
>>> raw = mne.io.read_raw_fif(mne.datasets.sample.data_path() + '/MEG/sample/sample_audvis_raw.fif')
>>> raw.annotations
<Annotations | 0 segments>
```
The `.copy()` above would fail if it were None since None does not have a `copy` method

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,47 @@
...
+        Updated annotations for raw data.
+    """
+    annot = raw.annotations.copy()
+    data, times = raw.get_data(return_times=True)
+    sampling_duration = 1 / raw.info['sfreq']
+
+    nans = np.any(np.isnan(data), axis=0)
+    starts, stops = _mask_to_onsets_offsets(nans)
+
+    if len(starts) > 0:
```
larsoner, line 33: I would switch this conditional to be `if len(starts) == 0: warn() return` that way the lines below this can be unidented.

mne/preprocessing/annotate_nan.py
```diff
@@ -0,0 +1,47 @@
...
+
+import numpy as np
+
+import warnings
+
+from ..annotations import Annotations
+from ..utils import _mask_to_onsets_offsets
+
+
+def annotate_nan(raw):
```
larsoner, line 13: And add the `%(verbose)s` in the parameters
```suggestion
@verbose
def annotate_nan(raw, *, verbose=None):
```
It's always good to give a `verbose` arg to things that can potentially warn so that you can easily disable the warnings

mne/io/nirx/nirx.py
```diff
@@ -21,15 +21,23 @@
...
 
-    This function has only been tested with NIRScout devices.
-
     Parameters
     ----------
     fname : str
         Path to the NIRX data folder or header file.
+    saturated : str
+        Replace saturated segments of data with NaNs.
+        If ignore (default) the measured data is returned, even if it
```
larsoner, line 33:
```suggestion
        If "ignore" (default) the measured data is returned, even if it
```

mne/io/nirx/nirx.py
```diff
@@ -21,15 +21,23 @@
...
+        contains measurements while the amplifier was saturated.
+        If nan the returned data will contain NaNs during time segments
```
larsoner, line 35:
```suggestion
        If "nan" the returned data will contain NaNs during time segments
```

mne/io/nirx/nirx.py
```diff
@@ -21,15 +21,23 @@
...
+        when the amplifier was saturated.
+        If annotate the returned data will contain annotations specifying
```
larsoner, line 37:
```suggestion
        If "annotate" the returned data will contain annotations specifying
```
And I would say this should be the default actually

larsoner:
@rob-luke it looks like we're close (today or tomorrow maybe) to pushing out 0.23. Do you think we should wait for this, or merge this early in the 0.24 cycle?

rob-luke:
> However, I just got off a phone call with NIRx and they said that the sensors very rarely saturate, they said they might go a year without seeing saturation. But is that your experience @rderollepot? Or do you see this more often? I will email NIRx

They just responded. Apparently saturation was more common in nirsport1. So we should be sure to ensure we handle it correctly.

rderollepot:
First of all, thanks a lot @rob-luke and @larsoner for your work on this PR !

> However, my tests are failing. When I read the data it appears to not contain any NaNs in the channel combinations specified in the montage. I have confirmed that your code does read the correct wave file based on the saturation argument 👍 . And I have confirmed that NaNs exist in the wl1 file 👍 and not in the nosatflag file 👍 .
> 
> The data on disk (specifically the files `nirx_15_3_recording_w_full_saturation` and `nirx_15_3_recording_w_occasional_saturation`) contains the measurements between all source and detector pairs. However, only certain pairs are valid, and these are specified in your montage. When I load the raw wl1 data file with all combinations of channels it has NaNs 👍 but when the code selects/subsets just the channels (optode pairings of interest) in your montage, all those channels are NaN free 👎 . Based on your great notes in [mne-tools/mne-testing-data#78](https://github.com/mne-tools/mne-testing-data/pull/78) I expected that at least the data for detector 2 would contain NaNs. Is this your understanding too?

Yes, that is what I would have expected too 👍 

> Is it possible that in the test data the channels in your montage never saturated?

I thought it did, but while reading your very clear analysis I realized there was a mistake in the process I followed to record those test datasets. To verify that I could trigger some saturations (which I was unsure), I just checked for NaNs presence in the .wl1/.wl2 files as you did, but I didn't verify that those channels were indeed in my montage as you pointed out, therefore not ensuring that those saturations are problematic. Sorry for that, I'll plan to record new test datasets if those cannot be used.

> To make things move quicker, would you like to email me another file (doesn't have to be a test file like before, just anything you have where you are certain it saturates) so I can confirm the behaviour in this PR.

I definitely have experimental recordings with saturated channels, I'll look for the cleaner one I can find and email it to you quickly.

> However, I just got off a phone call with NIRx and they said that the sensors very rarely saturate, they said they might go a year without seeing saturation. But is that your experience @rderollepot? Or do you see this more often? I will email NIRx to ask about this specifically.

We have two NIRx fNIRS in the lab: the first NIRSport, and the NIRScout. We rarely end up with saturations (if any ?) using the NIRScout, but we often have some using the NIRSport v1, which confirms the answer you just got from NIRx.

rob-luke:
> Yes, that is what I would have expected too

Great, I'm glad we are all of the same understanding.

> I'll plan to record new test datasets if those cannot be used

If possible this would be great. But I am aware that getting time in the lab can be tough. So let me know if not possible. Perhaps I’m being too pedantic with tests?

We would only need to replace one measurement. Then we would have 1) a recording with no saturation (already have) 2) a recording with saturation, but not on montage channels (already have) 3) recording that has saturation in the channel of interest (new measurement, could just overwrite one that you previously committed). 3 would allow us to test the new code features.

Thoughts?

larsoner:
Instead of recording new data, could you modify the existing montage definition (can be on the fly in the code in memory with `monkeypatch` for example) to have a saturated source-detector pair so that the `nan` get read from the wl1/wl2 files?

rob-luke:
Both @rderollepot and NIRx have sent me data to test with locally. If these larger files operate correctly locally then I'd be more confident with a testing hack as you describe. I'll report back here in a week on my progress.

rob-luke:
Testing will fail now as I am using a local test file.

rderollepot:
I will have access both to the NIRSport v1 and to the lab today (and possibly until the end of the week if further needs come up), so I will try to record the third dataset as described by @rob-luke here:


> We would only need to replace one measurement. Then we would have 1) a recording with no saturation (already have) 2) a recording with saturation, but not on montage channels (already have) 3) recording that has saturation in the channel of interest (new measurement, could just overwrite one that you previously committed). 3 would allow us to test the new code features.

I am unsure of the correct github workflow here though ?

Should I raise an issue related to mne-tools/mne-testing-data#78 pointing out the fact that saturations do not appear in montage channels as it would have been expected, and then open a new PR linked to that issue in which I overwrite the third dataset with a new one ? Or do you recommend some other way ?

rob-luke:
Good question. I think the easiest (least work on you) way is to open a pr that replaces the directory nirx_15_3_recording_w_full_saturation with a recording that contains some saturation. Seem ok to you? Other ideas?

drammock:
> I am unsure of the correct github workflow here though

I agree with @rob-luke; since we've talked through the right solution here already, no need to open a new issue in the testing repo first. Instead, just open a PR into the testing repo with the new file(s), and in the PR description add cross-references to the original PR (https://github.com/mne-tools/mne-testing-data/pull/78) and this one (https://github.com/mne-tools/mne-python/pull/9348).

rderollepot:
Thanks @rob-luke and @drammock for your inputs, I just finished submitting the PR mne-tools/mne-testing-data#85

Let me know @rob-luke if you see anything strange, or if an additional test dataset would be needed.

mne/io/nirx/nirx.py
```diff
@@ -133,15 +133,27 @@ def __init__(self, fname, saturated, preload=False, verbose=None):
...
+                    # flags in it.
+                    satidx = np.where(['nosatflags' not in op.basename(k)
+                                       for k in files[key]])[0][0]
+
                     if saturated == 'nan':
+                        # In this case data is returned with NaNs in it.
                         warn('The measurement contains saturated data. '
                              'Saturated values will be replaced by NaNs.')
-                        files[key] = files[key][1]
+                        files[key] = files[key][satidx]
```
rob-luke: In addition to the test file issue, there was a bug in the original authors code. And I guess that's why we need tests :)

rob-luke:
With the above bug fixed, once https://github.com/mne-tools/mne-testing-data/pull/85 is merged and a new release is tagged and we bump the test version here, then I hope everything goes green.

mne/io/nirx/nirx.py
```diff
@@ -21,15 +21,23 @@
...
     fname : str
         Path to the NIRX data folder or header file.
+    saturated : str
+        Replace saturated segments of data with NaNs.
+        If "ignore" (default) the measured data is returned, even if it
+        contains measurements while the amplifier was saturated.
+        If "nan" the returned data will contain NaNs during time segments
+        when the amplifier was saturated.
+        If "annotate" the returned data will contain annotations specifying
+        sections the saturate segments and the data will contain NaNs.
```
rob-luke, line 38: This seems unexpected to me. I would have expected that the purpose of this option would be to return real data (no NaNs) so you can filter etc without nan issues. And could then exclude the bad segments.

Or I guess you could then use the annotations to fill in the bad segments using interpolation between the end points of good data.

@rderollepot sorry if this was discussed already, but is this the intended behaviour?

rob-luke: Not returning NaNs here will be a big rewrite though :(

larsoner: I agree it might be nicer in `annotate` mode to give the actual values but annotate with `BAD_SATURATED`, and this makes sense as the default behavior. I think it's worth rewriting for example to pass a list of the start/stop intervals that would need to be added as annotations.

Then a test could be that `annotate` mode should give the same results as `nan` mode + `annotate_nan`, other than the `description` of the annotations being different (the former being `BAD_SATURATED` and the latter being `BAD_NAN` or so)

rob-luke: > I agree it might be nicer in annotate mode to give the actual values but annotate with BAD_SATURATED, and this makes sense as the default behavior.

Agreed, this seems a logical default behaviour. However, I don’t have time to do this now. Getting the PR fixed, adding tests, and getting CIs green already took more time than anticipated. This is now in a good state for anyone to take over and change default behaviour.

larsoner:
I took a stab at it @rob-luke, see if it makes sense to you

mne/io/nirx/nirx.py
```diff
@@ -299,43 +330,68 @@ def prepend(li, str):
...
+        # make onset/duration/description
+        onset, duration, description = list(), list(), list()
+        if annot_mask is not None:
+            on, dur = _mask_to_onsets_offsets(annot_mask.any(0))
+            on = on / info['sfreq']
+            dur = dur / info['sfreq']
+            dur -= on
+            onset.extend(on)
+            duration.extend(dur)
+            description.extend(['BAD_NAN'] * len(on))
```
rob-luke, line 362: Would this be more specific? Maybe helpful if there are other causes of NaN in other data types M/EEG etc.
```suggestion
            description.extend(['BAD_SATURATION'] * len(on))
```

rob-luke:
Code looks good. I am just testing the behaviour locally against some additional private files that rderollepot kindly shared.

rob-luke:
Behaviour looks good to me locally. Here is an example

image

Thanks @larsoner for doing the heavy lifting.

rob-luke:
I would approve, but you cant approve your own pull requests 😄

larsoner:
I also realized this is a perfect use case for channel specific annotations. I'll add that next I think

mne/io/nirx/nirx.py
rderollepot: > This seems unexpected to me. I would have expected that the purpose of this option would be to return real data (no NaNs) so you can filter etc without nan issues. And could then exclude the bad segments.
> 
> Or I guess you could then use the annotations to fill in the bad segments using interpolation between the end points of good data.
> 
> @rderollepot sorry if this was discussed already, but is this the intended behaviour?

No I think you are right, after some digging into the original PR, it seems that this topic had been quite heavily discussed, and the conclusion that cbrnr drawn at that time can be found [here](https://github.com/mne-tools/mne-python/pull/7936#discussion_r449058046).

I also agree that 'annotate' (without NaNs) would make a better default behaviour than 'ignore', as 'ignore' now seems to me a quite dangerous option that one should carefully choose.

larsoner:
Okay made them channel-specific. If we know that they are always saturated in source-detector pairs like this (or that we should always annotate both as saturated if one of them is):
```
(Pdb) p raw.annotations.ch_names
array([('S5_D2 760',), ('S5_D2 850',), ('S5_D2 760',), ('S5_D2 850',), (),
       ('S5_D2 760',), ('S5_D2 850',), ()], dtype=object)
```
Then I could simplify the code a bit to produce half the number of channel-specific annotations. Currently it makes one annotation for S5_D2 760, and a separate annotation for S5_D2 850, but if these are always paired, we could make a single annotation for the time window that lists S5_D2 760 and S5_D2 850 together.

rderollepot:
Great !

I do not know the 'hardware based' answer to that question, but I made a simple script to find out where NaNs were located in both wl1 and wl2 matrices, and it turns out that, on the 47 datasets I have that saturated, wl1 and wl2 always share the exact same 'NaN mask'. It is not enough to guarantee that they are always saturated in source-detector pairs though.

larsoner:
My vote is to make it so that if either is saturated they both get annotated as saturated. It seems to be the case in 47/47 of your cases so in practice might be true anyway, and even if it isn't the case for some files, we probably will end up ignoring that S-D pair downstream anyway. WDYT?

rob-luke:
> My vote is to make it so that if either is saturated they both get annotated as saturated. 

Agreed. 

> we probably will end up ignoring that S-D pair downstream anyway

Correct. The current behaviour is to mirror the behaviour within S-D pairs. So if one channel is marked as bad, then so is the other. See...

https://github.com/mne-tools/mne-python/blob/8ccc3999da6c15efa03840230c13aeb7bab5618d/mne/preprocessing/nirs/nirs.py#L198-L203

rob-luke:
Actually I need to add more internal calls to those functions. I'll do that in another PR.

rderollepot:
> My vote is to make it so that if either is saturated they both get annotated as saturated.

I also agree with that 👍

doc/changes/names.inc
```diff
@@ -389,3 +389,7 @@
 .. _Jack Zhang: https://github.com/jackz314
 
 .. _Felix Klotzsche: https://github.com/eioe
+
+.. _David Julien: https://github.com/Swy7ch
+
+.. _Romain Derollepot: https://github.com/rderollepot
```
larsoner, line 395: @Swy7ch @rderollepot let me know if you want different URLs here

rderollepot: That is great for me, thanks 👍 

I also want to take this opportunity to say that I am really grateful for your help @larsoner and @rob-luke in seing that the great contribution @Swy7ch started almost a year ago can come to an end, improving it with new features along the way. We are still very new to MNE in the lab (it started with David's internship), and I am not sure I would have been able to finish it myself (or it would have take me ages 😅).

rob-luke:
🥳

PR #9351, opened 2021-04-26
BUG: Fix concat on Windows

### LINKED ISSUE #9336: brain.screenshot(time_viewer=True) crash
I was running the mne-bids-pipeline (so with mne-23dev cloned from GitHub in backend), but a bug appeared just after the creation of the reports of the first subject. After investigation, it turns out that the bug comes from MNE, from the option time_viewer=True in the screenshot method. By setting it to false, the image plots nicely.

I tried to make a reproducible example with sample data, but this does not reproduce the issue (Else). So I've also included here the (IF True) my [stc file](https://drive.google.com/file/d/1rcaBOMTbLnmyWqvUn9dRy4G3V5PEJehY/view?usp=sharing) and my [subjects file](https://drive.google.com/drive/folders/13TeY0L88dsag6AD3VfYj7GTOFqS_He_k?usp=sharing) which isolate the code on my computer.
```python
from pathlib import Path
import matplotlib
import mne
matplotlib.use('Qt5Agg')

if True:
    stc_fname = Path("mne_data/ds003392/derivatives",
                     'mne-bids-pipeline/sub-01/meg/',
                     'sub-01_task-localizer_coherent+dSPM+hemi-lh.stc')

    subjects_dir = Path("mne_data/ds003392/derivatives/freesurfer",
                        'subjects')
    subject = 'sub-01'
    stc = mne.read_source_estimate(stc_fname, subject=subject)
else:
    fname = data_path + '/MEG/sample/sample_audvis-meg'

    stc = mne.read_source_estimate(fname)
    subject = "sample"
    subjects_dir = Path("C:/Users/charb/mne_data/MNE-sample-data/subjects")
brain = stc.plot(subject=subject, subjects_dir=subjects_dir,
                 initial_time=0.13, views=['lat'], hemi='split')
brain.toggle_interface()
brain._renderer.plotter.reset_camera()
brain._renderer.plotter.subplot(0, 0)
brain._renderer.plotter.reset_camera()
figs, ax = plt.subplots(figsize=(15, 10))

# BUG : Set time_viewer to False to remove the Bug
image = brain.screenshot(time_viewer=True)
ax.imshow(image)
ax.axis('off')
plt.show()

```

#### Expected results
An interactive plot.

#### Actual results
When I run the code, the figure appears one second and then crashes

```
(mne_dev) C:\Users\charb\Desktop\parietal>C:/Users/charb/anaconda3/envs/mne_dev/python.exe c:/Users/charb/Desktop/parietal/test.py
Using pyvista 3d backend.

Using control points [13.01218636 16.52222281 48.36637508]
  File "c:\Users\charb\Desktop\parietal\test.py", line 51, in <module>
    image = brain.screenshot(time_viewer=True)
  File "C:\Users\charb\anaconda3\envs\mne_dev\lib\site-packages\mne\viz\_brain\_brain.py", line 2631, in screenshot
    trace_img = np.reshape(
  File "<__array_function__ internals>", line 5, in reshape
  File "C:\Users\charb\anaconda3\envs\mne_dev\lib\site-packages\numpy\core\fromnumeric.py", line 299, in reshape
    return _wrapfunc(a, 'reshape', newshape, order=order)
  File "C:\Users\charb\anaconda3\envs\mne_dev\lib\site-packages\numpy\core\fromnumeric.py", line 58, in _wrapfunc
    return bound(*args, **kwds)
ValueError: cannot reshape array of size 1082952 into shape (1015,4)

(mne_dev) C:\Users\charb\Desktop\parietal>    

```
#### Additional information
I'm working on Windows.

```
(mne_dev) C:\Users\charb\Desktop\parietal\mne-bids-pipeline>python
Python 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 04:59:43) [MSC v.1916 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import mne; mne.sys_info()
Platform:      Windows-10-10.0.19041-SP0
Python:        3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 04:59:43) [MSC v.1916 64 bit (AMD64)]
Executable:    C:\Users\charb\anaconda3\envs\mne_dev\python.exe
CPU:           Intel64 Family 6 Model 158 Stepping 10, GenuineIntel: 8 cores
Memory:        7.9 GB

mne:           0.23.dev0
numpy:         1.20.2 {blas=NO_ATLAS_INFO, lapack=lapack}
scipy:         1.6.2
matplotlib:    3.4.1 {backend=Qt5Agg}

sklearn:       0.24.1
numba:         0.53.1
nibabel:       3.2.1
nilearn:       0.7.1
dipy:          1.4.0
cupy:          Not found
pandas:        1.2.4
mayavi:        4.7.2
pyvista:       0.29.1 {pyvistaqt=0.3.0, OpenGL 4.5.0 - Build 26.20.100.6911 via Intel(R) UHD Graphics 630}
vtk:           9.0.1
PyQt5:         5.12.3
>>>
```

larsoner:
Closes #9336

@crsegerie can you see if this fixes your problem? 

I was able to replicate on Windows. Basically I think the problem was that after our DPI calculations, the resulting size was something like `1406.99999999` instead of `1407`, so MPL made something with only 1406 pixels instead of 1407 and this broke things. We could try tweaking our DPI calculations, but it seems safer just to output a PNG then read it back, as this has the dimensions, then make use of our smart concat function to take care of any difference in dimension.

Okay with you @GuillaumeFavelier ?

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1229,7 +1229,7 @@ class Brain(object):
     def _configure_tool_bar(self):
         self._renderer._tool_bar_load_icons()
         self._renderer._tool_bar_set_theme(self.theme)
-        self._renderer._tool_bar_initialize()
+        self._renderer._tool_bar_initialize(name="Toolbar")
         self._renderer._tool_bar_add_file_button(
             name="screenshot",
             desc="Take a screenshot",
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -2611,9 +2611,11 @@ class Brain(object):
             Image pixel values.
         """
         img = self._renderer.screenshot(mode)
+        logger.debug(f'Got screenshot of size {img.shape}')
         if time_viewer and self.time_viewer and \
                 self.show_traces and \
                 not self.separate_canvas:
+            from matplotlib.image import imread
             canvas = self.mpl_canvas.fig.canvas
             canvas.draw_idle()
             fig = self.mpl_canvas.fig
@@ -2624,13 +2626,22 @@ class Brain(object):
                 # (e.g., macOS w/Qt 5.14+ and VTK9) then things won't work,
                 # so let's just calculate the DPI we need to get
                 # the correct size output based on the widths being equal
-                dpi = img.shape[1] / fig.get_size_inches()[0]
-                fig.savefig(output, dpi=dpi, format='raw',
+                size_in = fig.get_size_inches()
+                dpi = fig.get_dpi()
+                want_size = tuple(x * dpi for x in size_in)
+                n_pix = want_size[0] * want_size[1]
+                logger.debug(
+                    f'Saving figure of size {size_in} @ {dpi} DPI '
+                    f'({want_size} = {n_pix} pixels)')
+                # Sometimes there can be off-by-one errors here (e.g.,
+                # if in mpl int() rather than int(round()) is used to
+                # compute the number of pixels) so rather than use "raw"
+                # format and try to reshape ourselves, just write to PNG
+                # and read it, which has the dimensions encoded for us.
+                fig.savefig(output, dpi=dpi, format='png',
                             facecolor=self._bg_color, edgecolor='none')
                 output.seek(0)
-                trace_img = np.reshape(
-                    np.frombuffer(output.getvalue(), dtype=np.uint8),
-                    newshape=(-1, img.shape[1], 4))[:, :, :3]
+                trace_img = imread(output, format='png')[:, :, :3]
             img = concatenate_images(
                 [img, trace_img], bgcolor=self._brain_color[:3])
         return img
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -2642,8 +2642,10 @@ class Brain(object):
                             facecolor=self._bg_color, edgecolor='none')
                 output.seek(0)
                 trace_img = imread(output, format='png')[:, :, :3]
-            img = concatenate_images(
-                [img, trace_img], bgcolor=self._brain_color[:3])
+                trace_img = np.clip(
+                    np.round(trace_img * 255), 0, 255).astype(np.uint8)
+            bgcolor = np.array(self._brain_color[:3]) / 255
+            img = concatenate_images([img, trace_img], bgcolor=bgcolor)
         return img
 
     @contextlib.contextmanager
--- a/mne/viz/backends/_qt.py
+++ b/mne/viz/backends/_qt.py
@@ -391,7 +391,7 @@ class _QtWindow(_AbstractWindow):
 
     def _window_adjust_mplcanvas_layout(self):
         canvas = self._mplcanvas.canvas
-        dock, dock_layout = _create_dock_widget(
+        self._mpl_dock, dock_layout = _create_dock_widget(
             self._window, "Traces", Qt.BottomDockWidgetArea)
         dock_layout.addWidget(canvas)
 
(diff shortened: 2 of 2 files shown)


crsegerie:
Yeah, this solves my problem! Thank you!

agramfort:
this PR breaks the screenshots for me on my mac. I get screenshots like this now

toto2

GuillaumeFavelier:
Thanks for pointing it out @agramfort because `test_brain_screenshot` is skipped for the `macos/conda` job:

```
mne/viz/_brain/tests/test_brain.py::test_brain_screenshot[pyvista] SKIPPED [ 95%]
...
SKIPPED [2] mne/viz/_brain/tests/test_brain.py:351: Unreliable/segfault on macOS CI
```

https://github.com/mne-tools/mne-python/pull/9351/checks?check_run_id=2440366851#step:12:2839
https://github.com/mne-tools/mne-python/pull/9351/checks?check_run_id=2440366851#step:12:3040

GuillaumeFavelier:
@larsoner what do you think about trying the alternative calculation only when the `reshape` fails?

larsoner:
> Thanks for pointing it out @agramfort because test_brain_screenshot is skipped for the macos/conda job:

I think it's just a 0-1 vs 0-255 mapping problem, I'll test locally and push a fix

> @larsoner what do you think about trying the alternative calculation only when the reshape fails?

I'd rather not have two code paths. If we're forced to go back to `raw` mode, we need to figure out somehow what shape the output actually needs to be. But like I mention above I doubt it will be necessary

larsoner:
This code:

<details>
<summary>Code</summary>

```
import os.path as op
import mne
from mne.minimum_norm import read_inverse_operator, apply_inverse

# Set dir
subject = 'sample'
data_dir = op.join(data_path, 'MEG', subject)
fname_evoked = data_dir + '/sample_audvis-ave.fif'
condition = 'Left Auditory'
evoked = mne.read_evokeds(fname_evoked, condition=condition,
                          baseline=(None, 0))
inv = read_inverse_operator(
    data_path + '/MEG/sample/sample_audvis-meg-oct-6-meg-inv.fif')
stc_vec = apply_inverse(evoked, inv, pick_ori='vector')
    brain = stc_vec.plot(hemi='both', views=['lat', 'med'],
                         initial_time=0.1, subjects_dir=subjects_dir)
plt.figimage(brain.screenshot(time_viewer=True))
```

</details>

Gives on `main`:

Screen Shot 2021-04-27 at 7 11 06 AM

On what @agramfort tested you can see two bugs -- one I introduced and another involving `bgcolor` that is on main but not seen there in this example (white line running vertically along the lower left part of the image):

Screen Shot 2021-04-27 at 7 11 31 AM

Pushed a commit to fix both problems:

Screen Shot 2021-04-27 at 7 13 54 AM

Clearly the image of the brain is wrong, and the matplotlib plot is not the right size either. But these are separate issues on `main`, too, so let's fix them separately.

@GuillaumeFavelier do you want to try my code snippet to see if you see the same problems on your system?

larsoner:
Okay the problem was really with my `figimage` call, I needed `resize=True` (edited now above) and I get something correct:

Screenshot from 2021-04-27 07-42-38

But there *is* a problem with the update of the display in that the views don't respond to the fact that the `matplotlib` traces have forced them to resize:

Screenshot from 2021-04-27 07-42-43

GuillaumeFavelier:
I would have to switch to KDE for a better test but on `i3` (which maximizes the new windows by default) I think I can reproduce:

image

When I leave the maximized mode:

image

Could it be yet another process events?

larsoner:
For me, deleting the `self._interactor.resize(ren_sz.width(), ren_sz.height())` in `_window_ensure_minimum_sizes` makes it look correct. Going to see what it breaks on Linux and macOS locally

larsoner:
... actually it doesn't make it work, it makes it look not cut off, which is good, but it's the wrong size, which is bad. I'll see if I can figure out what to do to make it work

larsoner:
Okay should be fixed, can you review and test locally @GuillaumeFavelier ?

GuillaumeFavelier (approved):
For me, everything works locally and the changes look good.

hoechenberger:
Thanks @larsoner!!

PR #9353, opened 2021-04-26
make graphviz diagram a link

### LINKED ISSUE #9223: graphviz diagrams not clickable/zoomable
The flowchart in the ICA tutorial is too small because it's scaled to fit into the center column. Would be good if one could easily zoom in; easiest would be to link to the full-sized image; most convenient from a user's perspective would be using some kind of lightbox.

https://user-images.githubusercontent.com/2046265/112762090-4e1e3600-8ffe-11eb-8fc7-f0da263dce4c.mov

_Originally posted by @hoechenberger in https://github.com/mne-tools/mne-python/issues/8876#issuecomment-808931273_

drammock:
closes #9223 

This is hackish because the graphviz-generated image has a dynamically generated filename like `../../_images/graphviz-7483cb1cf41f06e2a4ef451b17f073dbe584ba30.png`. It appears to be deterministic though (same filename on the live site as is generated for me locally), so as long as the diagram's source code remains the same (and possibly, as long as the same version of graphviz is in use?), the link should work.

I think the right way to do it would be for someone to enhance `sphinx.ext.graphviz` to add a new option `:href:` or `:link:` or similar that would cause an `<a>` tag to get generated that wraps around the `` (instead of what's done here, which wraps around the `img` and 2 containing `div`s).  But I'm skeptical whether such an enhancement would be accepted upstream, since SVG output can have links on specific graph elements...  not sure how that would play out.

### DIFF
--- a/tutorials/preprocessing/40_artifact_correction_ica.py
+++ b/tutorials/preprocessing/40_artifact_correction_ica.py
@@ -136,9 +136,18 @@ raw.crop(tmax=60.)
 # parameters that control dimensionality at various stages are summarized in
 # the diagram below:
 #
+#
+# .. raw:: html
+#
+#    <a href="../../_images/graphviz-7483cb1cf41f06e2a4ef451b17f073dbe584ba30.png">
+#
 # .. graphviz:: ../../_static/diagrams/ica.dot
 #    :alt: Diagram of ICA procedure in MNE-Python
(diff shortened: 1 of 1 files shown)


agramfort (approved):

larsoner:
Thanks @drammock !

PR #9356, opened 2021-04-27
API: Complete deprecations

larsoner:
Post-release deprecation changes.


agramfort (approved):

drammock:
Here are a handful of `XXX` bits in our code that might(?) make sense to address now:

```
mne/channels/montage.py:683:    # XXX: to change to the new naming in v.0.20 (all this block should go)
mne/commands/mne_compute_proj_ecg.py:64:                      default=False)  # XXX: change to default=True in 0.17
mne/commands/mne_compute_proj_eog.py:68:                      default=False)  # XXX: change to default=True in 0.17
mne/fixes.py:32:    XXX: Can be removed when numpy introduces a fix.
mne/tests/test_annotations.py:65:    assert raw.annotations is not None  # XXX to be fixed in #5416
mne/tests/test_annotations.py:66:    assert len(raw.annotations.onset) == 0  # XXX to be fixed in #5416
mne/tests/test_annotations.py:250:    assert raw_read.annotations is not None  # XXX to be fixed in #5416
mne/tests/test_annotations.py:251:    assert len(raw_read.annotations.onset) == 0  # XXX to be fixed in #5416
mne/tests/test_cov.py:323:        # XXX don't use "shrunk" here, for some reason it makes Travis 2.7
mne/utils/docs.py:1892:"""  # XXX make sure we deal with this deprecation in 0.23
mne/viz/_brain/tests/test_brain.py:438:    # XXX disable for sprint because it's too unreliable
mne/viz/utils.py:993:    # XXX: Bug in matplotlib won't allow setting the position of existing
server_environment.yml:18:- nbformat <5.1  # XXX remove pinning once https://github.com/jupyter/nbformat/issues/206 has been fixed
```

larsoner:
Those comments probably just need to be updated, will look a bit more in-depth tomorrow

larsoner:
I'll take care of those in a separate PR so that it's easier to review

PR #9358, opened 2021-04-28
MRG, MAINT: Bump requirements, remove cruft

larsoner:
1. Bump requirements to those ~2 years old (this includes Python 3.7 as minimum, SciPy et al. have done similarly so I think it's okay)
2. Remove things from `fixes.py` that required old versions (np.einsum, np.take_along_axis, etc.)
3. Update `old` CI run
4. Move some stuff from `viz/utils.py` that really belongs in `fixes.py`, which is meant to keep version-specific backward compat shims
5. [Went through](https://github.com/mne-tools/mne-python/pull/9356#issuecomment-827940308) the `XXX` and `TODO` in our codebase and took care of some of them, including:
    1. Remove `run_tests_as_main`
    2. Remove uses of `_TempDir` in tests in most places
    3. Move `morph_maps` code to their own file rather than living in `surface.py`
    4. Remove `DigMontage.dev_head_t` as it's never used or set, and probably was never meant to be there (we have a warning that this class should not be instantiated by users anyway)
    5. Bump pydata-sphinx-theme to 0.6.3 so we can get rid of our custom template fix

I'm going to update the "make a release" instructions with steps 1-3 since I think it should come right after the deprecations PR (as it is here).

@drammock I left a fun one behind that you can tackle in a follow-up PR in `mne/viz/_figure.py`:
```
        # XXX when we support a newer matplotlib (something >3.0) the
        # instructions can have inline bold formatting:
        # instructions = '\n'.join(
        #     [r'$\mathbf{Left‐click~&~drag~on~plot:}$ create/modify annotation',  # noqa E501
        #      r'$\mathbf{Right‐click~on~plot~annotation:}$ delete annotation',
        #      r'$\mathbf{Type~in~annotation~window:}$ modify new label name',
        #      r'$\mathbf{Enter~(or~click~button):}$ add new label to list',
        #      r'$\mathbf{Esc:}$ exit annotation mode & close this window'])
```


requirements_doc.txt
```diff
@@ -1,7 +1,7 @@
 # requirements for building docs
 sphinx
 https://github.com/numpy/numpydoc/archive/main.zip
-pydata-sphinx-theme==0.6.1
+pydata-sphinx-theme>=0.6.3
```
drammock, line 4: I would keep this pinned.  The theme is still under pretty active development, and its test suite is still being refined.
```suggestion
pydata-sphinx-theme==0.6.3
```

larsoner: I assumed that was the original rationale, but I'm hoping that by doing this we can catch their bugs earlier, similar to how we build CircleCI with `sphinx-gallery:master`. I figure if it starts breaking things in a way we can't easily fix then we can easily add a fixed pin later

drammock: fair enough.  I guess staying on latest point release is not as dangerous as being on their latest main, anyway.

agramfort (approved):
what a cleanup !

drammock (approved):

PR #9365, opened 2021-04-30
better docs for permutation_cluster_test

### LINKED ISSUE #9364: make it easier to specify adjacency in permutation_cluster_test
I tried to do a time-frequency permutation_cluster_test over 126 channels for different observations and for two groups.
So I have 4D data (observations, times, frequencies, channels) for two groups of users but it didn't work. 

```py
adjacency = sparse.csr_matrix(np.zeros((126,126)))
sample = np.zeros((4,100,6,126))

print(sample.shape)
print(adjacency.shape)

F_obs, clusters, cluster_pv, H0 = mne.stats.permutation_cluster_test(X=[sample, sample], adjacency=adjacency)
```

the error raised is 

> ValueError: adjacency (len 126) must be of the correct size, i.e. be equal to or evenly divide the number of tests (75600). If adjacency was computed for a source space, try using the fwd["src"] or inv["src"] as some original source space vertices can be excluded during forward computation

at line 606 of "mne/stats/cluster_level.py" file.
(and 75600 is evenly dividable by 126, but "got_times" is different from "n_times" for the dimension more)

It seems like at most there can be three dimensions and not four.
I thought that with this version it was possible to do it.

Is it not?

Best,
Giovanni

_Originally posted by @GiovanniGr in https://github.com/mne-tools/mne-python/issues/7915#issuecomment-829705335_

drammock:
closes #9364 

@larsoner following up on our conversation today, this should clarify the docstring at least.  I also snuck in a code simplification, though I realized that the functionality for omitting adjacency along a particular dimension is already there, you just need to pass a matrix of zeros for that dimension.

### DIFF
--- a/mne/stats/_adjacency.py
+++ b/mne/stats/_adjacency.py
@@ -37,8 +37,8 @@ def combine_adjacency(*structure):
 
     >>> _time = n_times  # regular lattice adjacency
     >>> _freq = np.zeros((n_freqs, n_freqs))  # no adjacency between freq. bins
-    >>> _chan = scipy.sparse.diags([1., 1.], offsets=(-1, 1),
-                                   shape=(n_chans, n_chans))  # just for examp.
+    >>> _chan = scipy.sparse.diags(
+            [1., 1.], offsets=(-1, 1), shape=(n_chans, n_chans))  # for example
     >>> combine_adjacency(_time, _freq, _chan)
     """
     from scipy import sparse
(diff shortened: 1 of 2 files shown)


larsoner:
Doctest problem:

https://github.com/mne-tools/mne-python/pull/9365/checks?check_run_id=2479764391#step:13:3306

adding a `# doctest: +NORMALIZE_WHITESPACE` at the end of the offending Python call might fix it

larsoner:
Thanks @drammock

PR #9366, opened 2021-05-01
When plotting GFP comparison in Report, don't show sensor layout by default

hoechenberger:
Showing the head circle with sensors often simply leads to overplotting (GFP traces / sensors) and is unnecessary, as the GFP calculation employs all channels anyway.


agramfort (approved):

agramfort:
thx @hoechenberger

PR #9367, opened 2021-05-02
MRG: Add fNIRS to html output

rob-luke:
#### Reference issue
None. But I really love these html outputs in notebooks, so cool.

Add fNIRS to the html output for raw.

#### Additional information
I follow #9172, but am unsure how to test this change. Can we merge without a test?

#### Alternative solution

If you don't want to add fNIRS channels as they are less common we could have an optional row using:
```
    {{if n_fnirs > 0}}
        <th>fNIRS channels</th>
        <td>{{n_fnirs}}</td>
    {{endif}}
```
I tested this locally and it works too, I could do the same for EOG and ECG too if you want.

### DIFF
--- a/mne/data/html_templates.py
+++ b/mne/data/html_templates.py
@@ -33,7 +33,7 @@ info_template = Template("""
     <tr>
         <th>Good channels</th>
         <td>{{n_mag}} magnetometer, {{n_grad}} gradiometer,
-            and {{n_eeg}} EEG channels</td>
+            {{n_eeg}} EEG channels, and {{n_fnirs}} fNIRS channels.</td>
     </tr>
     <tr>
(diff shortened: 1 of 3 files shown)


rob-luke:
Render of change can be viewed here: https://28136-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/70_fnirs_processing.html#preprocessing-functional-near-infrared-spectroscopy-fnirs-data

And here is a screen grab...

image

rob-luke:
@larsoner @drammock @agramfort are you ok with this or would you prefer my alternative option above?

agramfort (approved):

agramfort:
thx @rob-luke

PR #9368, opened 2021-05-02
WIP: Use nirx name field to populate his_id

rob-luke:
#### Reference issue
None

Use the name field to populate the `subject_info['his_id']` field. This then prints nicely in the html _repr_.

#### Additional information
When starting a recording with a nirx devices the user is presented with a GUI screen and can enter age, gender, experiment type, and name. I always this name field to enter the participant ID, like P1, P2, etc. Can we use this field to populate the his_id variable in mne? The name is a free field text input, so I guess some labs could actually enter a participant name in here, but I think its more common to enter an ID. But this is probably lab variable, so close if it seems a bad idea. I just like the nice html printing of his_id.

### DIFF
--- a/mne/io/nirx/nirx.py
+++ b/mne/io/nirx/nirx.py
@@ -247,11 +247,13 @@ class RawNIRX(BaseRaw):
         #       "Additional Notes", "Contact Information" and this information
         #       is currently discarded
         # NIRStar does not record an id, or handedness by default
+        # The name field is used to populate the his_id variable.
         subject_info = {}
         if is_aurora:
             names = inf["subject"].split()
         else:
             names = inf['name'].split()
+        subject_info['his_id'] = "_".join(names)
         if len(names) > 0:
             subject_info['first_name'] = \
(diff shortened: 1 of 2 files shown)


larsoner:
Seems reasonable to me but CI failures look legitimate

rob-luke:
Thanks @larsoner for the feedback. I should have opened this as an issue not a PR.

But since you think its not a terrible idea I will fix the code and tests etc then ping for a review. Cheers.

larsoner:
Issue-as-PR is no problem, it's even sometimes easier to comment on. The risk is mostly taken by the person who opens it because they might have wasted some time on commits that might not be merged, rather than for a reviewer since it's mostly going to be the same comments either way.

rob-luke:
@larsoner, I finally got around to fixing those errors, could you please review.

agramfort (approved):

drammock (approved):

drammock:
thanks @rob-luke!

PR #9369, opened 2021-05-02
MRG, ENH: Make _get_hpi_info public

hoechenberger:
I need this functionality in MNE-BIDS to enrich our MEG sidecar files with information about cHPI coil frequencies. Therefore I decided to make this function public.

WIP: Need to add tests for the newly supported data types

cc @agramfort

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,7 +23,7 @@ Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
 
-- Nothing yet
+- New function :func:`mne.chpi.get_hpi_info` to retrieve basic information about the cHPI system used when recording MEG data (this functionality was hidden in a private function before) (:gh:`9369` by `Richard Höchenberger`_)
 
 
 Bugs
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,7 +23,7 @@ Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
 
-- New function :func:`mne.chpi.get_chpi_info` to retrieve basic information about the cHPI system used when recording MEG data (this functionality was hidden in a private function before) (:gh:`9369` by `Richard Höchenberger`_)
+- New function :func:`mne.chpi.get_chpi_info` to retrieve basic information about the cHPI system used when recording MEG data (:gh:`9369` by `Richard Höchenberger`_)
 
 
 Bugs
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -23,7 +23,7 @@ Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
 
-- New function :func:`mne.chpi.get_hpi_info` to retrieve basic information about the cHPI system used when recording MEG data (this functionality was hidden in a private function before) (:gh:`9369` by `Richard Höchenberger`_)
+- New function :func:`mne.chpi.get_chpi_info` to retrieve basic information about the cHPI system used when recording MEG data (this functionality was hidden in a private function before) (:gh:`9369` by `Richard Höchenberger`_)
 
 
 Bugs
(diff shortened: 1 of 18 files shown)


doc/preprocessing.rst
```diff
@@ -156,6 +156,7 @@ EEG referencing:
 .. autosummary::
    :toctree: generated/
 
+   get_hpi_info
```
hoechenberger, line 159: Is this supposed to be in alphabetic order here?

tutorials/preprocessing/59_head_positions.py
```diff
@@ -11,6 +11,11 @@
 of time. Here we show an example of how to do this, and how to visualize
 the result.
 
+Authors: Eric Larson <larson.eric.d@gmail.com>
+         Richard Höchenberger <richard.hoechenberger@gmail.com>
+
+License: BSD (3-clause)
+
```
agramfort, line 18: we add this after the __doc__ header usually.

tutorials/preprocessing/59_head_positions.py
```diff
@@ -20,9 +25,6 @@
 frequencies, plus other peaks related to power line interference (60 Hz and
 harmonics).
 """
-# Authors: Eric Larson <larson.eric.d@gmail.com>
-#
-# License: BSD (3-clause)
```
agramfort, line 25: it was at the right place

doc/changes/latest.inc
```diff
@@ -23,7 +23,7 @@ Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
 
-- Nothing yet
+- New function :func:`mne.chpi.get_chpi_info` to retrieve basic information about the cHPI system used when recording MEG data (this functionality was hidden in a private function before) (:gh:`9369` by `Richard Höchenberger`_)
```
larsoner, line 26: Users shouldn't ever need to care or think about or private functions (I would say it's better if they don't know of their existence, lest they get curious about what other functionality we're hiding :) )
```suggestion
- New function :func:`mne.chpi.get_chpi_info` to retrieve basic information about the cHPI system used when recording MEG data (:gh:`9369` by `Richard Höchenberger`_)
```

mne/chpi.py
```diff
@@ -295,14 +295,51 @@ def extract_chpi_locs_kit(raw, stim_channel='MISC 064', *, verbose=None):
...
+    from .io import BaseRaw
+    from . import BaseEpochs
+    from .evoked import Evoked
+
+    _validate_type(item=inst, item_name='inst',
+                   types=(Info, BaseRaw, BaseEpochs, Evoked))
+    _check_option(parameter='on_missing', value=on_missing,
+                  allowed_values=['ignore', 'raise'])
+
+    info = inst if isinstance(inst, Info) else inst.info
```
larsoner, line 336: I think I would rather just have people pass `inst.info` here

mne/chpi.py
```diff
@@ -295,14 +295,51 @@ def extract_chpi_locs_kit(raw, stim_channel='MISC 064', *, verbose=None):
...
+    """Retrieve cHPI information from the data.
+
+    Parameters
+    ----------
+    inst : instance of Info | Raw | Epochs | Evoked
+        The object from which to extract the HPI information.
+    on_missing : 'ignore' | 'raise'
+        What to do if no HPI information can be found. If ``'ignore'``
+        (default), all return values will be empty arrays or ``None``. If
+        ``'raise'``, an exception will be raised.
```
larsoner, line 310: Better / more DRY to use `on_missing_base` in `docs.py` (see how we do it for other `on_*` in other functions) and the helper `mne.utils. _on_missing`. I don't see a reason why `'warn'` shouldn't be supported here.

mne/chpi.py
hoechenberger: I thought supporting more object types could provide slightly more convenience for users … no? :)

agramfort: I would stick to passing info.

opening more options means more tests, more documentation etc.

hoechenberger:
I believe I have addressed all of the review comments

larsoner:
Failures look legitimate

larsoner:
Thanks @hoechenberger !

PR #9375, opened 2021-05-06
WIP: Add more checks to ensure fNIRS bads are set correctly for optode pairs

rob-luke:
#### Reference issue
Now that users can create and modify their own fnirs data structures (#9141) this adds more checks that the data structure is correct. Follows #9184 #9185

fNIRS channels come in pairs (either multiple wavelengths or chromaphore). If one of the pairs is set to bad then they both should be (see similar discussion about saturation here: https://github.com/mne-tools/mne-python/pull/9348#issuecomment-832123559). An internal function already existed to check that the bids are spread correctly across pairs. Here I just add another usage of this function.

I also modify these bad checking/spreading functions to use info rather than raw. As the checking other nirs checking functions now operate on info as suggested by @HanBnrd.

### DIFF
--- a/mne/preprocessing/nirs/nirs.py
+++ b/mne/preprocessing/nirs/nirs.py
@@ -196,17 +196,17 @@ def _fnirs_check_bads(info):
             raise RuntimeError('NIRS bad labelling is not consistent')
 
 
-def _fnirs_spread_bads(raw):
+def _fnirs_spread_bads(info):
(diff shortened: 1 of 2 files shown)


HanBnrd:
Awesome, thanks @rob-luke 🙂

rob-luke:
@agramfort  @larsoner could you please review

PR #9378, opened 2021-05-06
FIX: make epoch cropping idempotent

### LINKED ISSUE #9374: Repeated Epoch/Evoked cropping is not idempotent when include_tmax=False
#### Describe the bug

Repeated cropping of `mne.Epoch` when `include_tmax=False` cuts off the last time sample.

#### Steps to reproduce

```python
import mne
import numpy as np

n_epochs, n_channels, n_timepoints = 10, 4, 21
fake_data = np.random.random((n_epochs, n_channels, n_timepoints))

epo = mne.EpochsArray(fake_data, info)

for i in range(10):
    epo.crop(tmin=0, tmax=0.4, include_tmax=False)
    print(f"{epo.tmax=}")
```

#### Expected results

The `epo.tmax` should be set to 0.375 on the first and all subsequent iterations.

#### Actual results

Every loop iteration cuts off the last time sample of the epoch.

#### Additional information

```
Platform:      Linux-5.4.0-72-generic-x86_64-with-glibc2.29
Python:        3.8.5 (default, Jan 27 2021, 15:41:15)  [GCC 9.3.0]
Executable:    /home/jan/dev/python/spot/spot_venv/bin/python
CPU:           x86_64: 8 cores
mne:           0.24.dev0
numpy:         1.19.5 {blas=openblas, lapack=openblas}
scipy:         1.5.4
matplotlib:    3.3.4 {backend=module://backend_interagg}

sklearn:       0.23.2
numba:         Not found
nibabel:       3.2.1
nilearn:       Not found
dipy:          Not found
cupy:          Not found
pandas:        1.1.5
mayavi:        Not found
pyvista:       Not found
vtk:           Not found
```

The cause of the issue is that when the `tmax` argument for `crop` is out of range of the `Epoch`, it is set to `epo.tmax`. However, this behaviour is (at least for me) unintuitive when `include_tmax=False` as I do not want to cut off the last time sample, if my desired interval is larger than the epoch.

Possible solution would be to set `tmax` to `tmax + 1/sfreq` when `include_tmax=True` and the desired tmax-crop is outside the epoch interval.

jsosulski:
#### Reference issue
Fixes #9374 

Set `include_tmax=True` when `tmax` is larger than the epoch.

#### Additional Information
I think this cannot be sensibly done in the `_time_mask` function as we do not know whether `tmax` was set automatically or not in that function. However, I do now knot if `_time_mask` should do this kind of error handling or if it is up to the calling functions.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -28,7 +28,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix bug with :meth:`mne.Epochs.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** `Jan Sosulski`_)
+- Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** `Jan Sosulski`_)
 
 
 API changes
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -28,7 +28,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Nothing yet
+- Fix bug with :meth:`mne.Epochs.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** `Jan Sosulski`_)
 
 
 API changes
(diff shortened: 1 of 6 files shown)


agramfort:
The test failure seems unrelated.

@jsosulski can you add a line in the bug fix section of our what’s new page? See file latest.inc 

Thx !

agramfort (approved):
thx @jsosulski that's perfect !

agramfort:
@larsoner I let you merge when you have a minute !

jsosulski:
@agramfort : I just finished checking all the `_time_mask` calls in MNE and `Evoked` has the same issue. Should I append that fix to this PR or create a new Issue for `Evoked` first?

agramfort:
if you can fix everything in one PR it's great. All the fixes are related

>

agramfort (approved):
thx @jsosulski 

@larsoner ok with the fix?

larsoner:
Thanks @jsosulski !

PR #1234, opened 2014-04-22
FIX/ENH: loading EDF annotations

mbillingr:
I made sure that the stim channel does not get resampled, and that the int16 values (each containing two characters) are not clamped.

Parsing the stim channel to get events of the form [onset, duration, 'annotation'] remains to be done.

I'm not sure where to store these events, though. Currently I consider not to store the events at all. Instead, a member function would simply parse and return the event list. Any thoughts?

PR #9379, opened 2021-05-07
[ENH] add exclude kw to psd_plot

eort:
Deals with [this comment ](https://github.com/mne-tools/mne-python/issues/7751#issuecomment-832755795]) in #1234.

This PR adds the `exclude` keyword to `raw.plot_psd()` and `epochs.plot_psd()`. I essentially copied the exclude functionality from `evoked.plot()` to `_psd_figure.py` and fixed up the dependencies. Also two tests were added. 

#### Additional information
`pytest viz/tests/test_figure.py` failed on my system, but it also did so on the main branch, so I don't think it is related to the changes.


agramfort:
thx @eort 

can you see why CIs fail?

eort:
> thx @eort
> 
> can you see why CIs fail?

Yeah, I guess so (see last commit)

rob-luke:
I am getting that same fine cal error in #9375.

mne/viz/_figure.py
```diff
@@ -2351,6 +2351,21 @@ def _psd_figure(inst, proj, picks, axes, area_mode, tmin, tmax, fmin, fmax,
...
+        if isinstance(exclude, str) and exclude == 'bads':
+            exclude = bad_ch_idx
+        elif (isinstance(exclude, list) and
+              all(isinstance(ch, str) for ch in exclude)):
+            exclude = [inst.info['ch_names'].index(ch) for ch in exclude]
+        else:
+            raise ValueError(
+                'exclude has to be a list of channel names or "bads"')
+
+        picks = np.array([pick for pick in picks if pick not in exclude])
```
larsoner, line 2367: We can use some internal functions to simplify this. We should probably also check that `include` are valid channel names
```suggestion
    inst.info._check_consistency()
    _validate_type(exclude, (list, tuple, str), 'exclude')
    if isinstance(exclude, str):
        _check_option('exclude', exclude, 'bads', extra='when str')
        exclude = info['bads']
    for ei, ex in enumerate(exclude):
        if not isinstance(ex, str) or ex not in info['ch_names']:
            raise ValueError(f'exclude[{ei}] ({repr(ex)}) '
                             'not found in info["ch_names"]')
    if len(exclude) > 0:
        picks = np.array([pick for pick in picks if
                          info['ch_names'][pick] not in exclude], int)
```

eort: Looks good to me!

As I simply copied the code from the erp.plot(), should I port the suggested change here, to that function as well?

PR #9381, opened 2021-05-07
Reading events info from ceo file

dddd1007:
#### Reference issue
Fixes #9380

#### What does this implement/fix?
The read_raw_curry() cannot read the event file from .ceo file. So I try to make the function recognize .ceo files, a type of event file generated by Curry.

PR #9389, opened 2021-05-10
Update latest.inc

dddd1007:
FIX: Make `mne.io.curry` can read '.cdt.ceo' file.


larsoner:
437df52 should be part of https://github.com/mne-tools/mne-python/pull/9381, can you switch to that branch and add it there? Something like this should work:
```
git checkout patch-1  # assuming you already have this locally
git cherry-pick 437df52
git push origin patch-1
```

PR #9390, opened 2021-05-10
add codemeta generation comment

drammock:
follow up for https://github.com/mne-tools/mne-python/pull/9387#issuecomment-836732605 and https://github.com/mne-tools/mne-python/pull/9387#issuecomment-836738190


larsoner:
Thanks @drammock

PR #9394, opened 2021-05-13
MRG, ENH: Add FieldLine and Kernel OPMs

larsoner:
Needs https://github.com/mne-tools/fiff-constants/pull/35, then I need to change the REPO and COMMIT


mne/data/coil_def.dat
```diff
@@ -3,7 +3,7 @@
 #
 #       Copyright 2005 - 2019
 #
-#       Matti Hämäläinen
+#       Matti Hamalainen
```
wmvanvliet, line 6: why this?

larsoner: Matti generates this with his C code, I'm guessing it doesn't get along well with unicode

larsoner:

PR #9396, opened 2021-05-14
MRG, ENH: Extend usability of stc_near_sensors

larsoner:
I have some fNIRS collaborators who generate figures where the fNIRS electrodes are projected onto the nearest brain areas, like in ECoG:

brain

or:

brain

Yes, I know the units on the colorbar are painful. :) But at least it's clear it's just a proximity-based projection and not actual inverse imaging!

This PR makes this possible by:

1. Adding a `picks` argument to `stc_near_sensors` so that fNIRS electrodes (HbO usually) can be used.
2. Adding a `smoothing_steps=0` meaning "no smoothing steps at all". Even `smoothing_steps=1` makes some smooth transitions where discrete ones are really desired. This caused me to shift "nearest" to -1 in the `brain` interface, which I think is fine.
3. Adding a `mode='weighted'` argument to `stc_near_sensors` that ensures that each STC vertex value is either zero, or a weighted sum of values from the sensors. This is more like what our smoothing/interpolation functions usually do. For example, if all sensors have a value of 1, all STC values are either 1 or 0. The default mode `"sum"` makes it so that some of these values can be greater than 1, which is a bit wacky (it's probably not a great default in this sense).

cc @rob-luke since you might be interested in this. Like I said it's not the greatest, but people do it (like in ECoG / sEEG) so I think it's probably worth having at least some way to do this.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -39,6 +39,8 @@ Enhancements
 
 - New function :func:`mne.label.find_pos_in_annot` to get atlas label for MRI coordinates.  (:gh:`9376` by **by new contributor** |Marian Dovgialo|_)
 
+- Add support for ``picks`` in :func:`mne.stc_near_sensors` (:gh:`9396` by `Eric Larson`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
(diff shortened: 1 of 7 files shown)


rob-luke:
@larsoner this is amazing! I didn't know this would be possible!

This is a great intermediate step until https://github.com/mne-tools/mne-nirs/issues/62 is completed (if ever, that's lots of work I think).

Want me to add this to the fNIRS tutorial? Or did you already plan that? I'm keen to play around with this new feature.

rob-luke (approved):

agramfort (approved):

agramfort:
Thx @larsoner

rob-luke:
@larsoner I have to say this again, this is so cool, I love it. Thanks

larsoner:
> I can add this to the fNIRS tutorial in another PR. Or did you already plan that? I'm keen to play around with this new feature.

Feel free to add it to a tutorial!

PR #9397, opened 2021-05-17
add support to save annotations as tsv

### LINKED ISSUE #9395: Add option to save Annotions as `.tsv`
The BIDS specification requires `.tsv` rather than `.csv`  files for tabular data. Currently, annotations can only be written to `fif`, `txt`, and `csv`. So it would be nice to add `tsv` support to that list. 

Should be easy enough by adding a separate call to `read/write_annotations.py` for `tsv` file, and calling internally read/write_csv with `sep='\t'`.

eort:
Closes #9395

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -433,7 +433,7 @@ class Annotations(object):
 
     @verbose
     def save(self, fname, *, overwrite=False, verbose=None):
-        """Save annotations to FIF, CSV or TXT.
+        """Save annotations to FIF, CSV, TSV or TXT.
 
         Typically annotations get saved in the FIF file for raw data
         (e.g., as ``raw.annotations``), but this offers the possibility
@@ -451,12 +451,14 @@ class Annotations(object):
         """
         check_fname(fname, 'annotations', ('-annot.fif', '-annot.fif.gz',
                                            '_annot.fif', '_annot.fif.gz',
-                                           '.txt', '.csv'))
+                                           '.txt', '.csv', '.tsv'))
         fname = _check_fname(fname, overwrite=overwrite)
         if fname.endswith(".txt"):
             _write_annotations_txt(fname, self)
         elif fname.endswith(".csv"):
             _write_annotations_csv(fname, self)
+        elif fname.endswith(".tsv"):
+            _write_annotations_csv(fname, self, sep='\t')
         else:
             with start_file(fname) as fid:
                 _write_annotations(fid, self)
@@ -706,12 +708,12 @@ def _write_annotations(fid, annotations):
     end_block(fid, FIFF.FIFFB_MNE_ANNOTATIONS)
 
 
-def _write_annotations_csv(fname, annot):
+def _write_annotations_csv(fname, annot, sep=','):
     annot = annot.to_data_frame()
     if 'ch_names' in annot:
         annot['ch_names'] = [
             _prep_name_list(ch, 'write') for ch in annot['ch_names']]
-    annot.to_csv(fname)
+    annot.to_csv(fname, sep=sep)
 
 
 def _write_annotations_txt(fname, annot):
@@ -803,6 +805,9 @@ def read_annotations(fname, sfreq='auto', uint16_codec=None):
     elif name.endswith('csv'):
         annotations = _read_annotations_csv(fname)
 
+    elif name.endswith('tsv'):
+        annotations = _read_annotations_csv(fname, sep='\t')
+
     elif name.endswith('cnt'):
         annotations = _read_annotations_cnt(fname)
 
@@ -834,7 +839,7 @@ def read_annotations(fname, sfreq='auto', uint16_codec=None):
     return annotations
 
 
-def _read_annotations_csv(fname):
+def _read_annotations_csv(fname, sep=','):
     """Read annotations from csv.
 
     Parameters
@@ -848,7 +853,7 @@ def _read_annotations_csv(fname):
         The annotations.
     """
     pd = _check_pandas_installed(strict=True)
-    df = pd.read_csv(fname, keep_default_na=False)
+    df = pd.read_csv(fname, keep_default_na=False, sep=sep)
     orig_time = df['onset'].values[0]
     try:
         float(orig_time)
--- a/mne/tests/test_annotations.py
+++ b/mne/tests/test_annotations.py
@@ -808,6 +808,10 @@ def dummy_annotation_file(tmpdir_factory, ch_names, fmt):
         content = ("onset,duration,description\n"
                    "2002-12-03 19:01:11.720100,1.0,AA\n"
                    "2002-12-03 19:01:20.720100,2.425,BB")
+    elif fmt == 'tsv':
+        content = ("onset\tduration\tdescription\n"
+                   "2002-12-03 19:01:11.720100\t1.0\tAA\n"
+                   "2002-12-03 19:01:20.720100\t2.425\tBB")
     elif fmt == 'txt':
         content = ("# MNE-Annotations\n"
                    "# orig_time : 2002-12-03 19:01:11.720100\n"
@@ -823,6 +827,12 @@ def dummy_annotation_file(tmpdir_factory, ch_names, fmt):
         if isinstance(content, Annotations):
             # this is a bit of a hack but it works
             content.ch_names[:] = ((), ('MEG0111', 'MEG2563'))
+        elif fmt == 'tsv':
+            content = content.splitlines()
+            content[-3] += '\tch_names'
+            content[-2] += '\t'
+            content[-1] += '\tMEG0111:MEG2563'
+            content = '\n'.join(content)
         else:
             content = content.splitlines()
             content[-3] += ',ch_names'
@@ -841,6 +851,7 @@ def dummy_annotation_file(tmpdir_factory, ch_names, fmt):
 @pytest.mark.parametrize('ch_names', (False, True))
 @pytest.mark.parametrize('fmt', [
     pytest.param('csv', marks=needs_pandas),
+    pytest.param('tsv', marks=needs_pandas),
     'txt',
     'fif'
 ])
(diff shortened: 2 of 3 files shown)


cbrnr:
I think it would be nice if users could select the columns separator independently of the file extension. CSV is not really standardized, and I've seen a lot of `.csv` files that used tabs. Similarly, I think it would also be useful to choose between starting at sample 0 or the first timestamp independently of the extension.

mne/tests/test_annotations.py
```diff
@@ -808,6 +808,10 @@ def dummy_annotation_file(tmpdir_factory, ch_names, fmt):
         content = ("onset,duration,description\n"
                    "2002-12-03 19:01:11.720100,1.0,AA\n"
                    "2002-12-03 19:01:20.720100,2.425,BB")
+    elif fmt == 'tsv':
+        content = ("onset\tduration\tdescription\n"
+                   "2002-12-03 19:01:11.720100\t1.0\tAA\n"
+                   "2002-12-03 19:01:20.720100\t2.425\tBB")
```
agramfort, line 814: this is not how BIDS events tsv files should be formatted

see for example

https://openneuro.org/datasets/ds000248/file-display/sub-01:meg:sub-01_task-audiovisual_run-01_events.tsv

eort: You mean `'    '` rather than `'\t'`?

mne/annotations.py
```diff
@@ -706,12 +708,12 @@ def _write_annotations(fid, annotations):
...
 
 
-def _write_annotations_csv(fname, annot):
+def _write_annotations_csv(fname, annot, sep=','):
     annot = annot.to_data_frame()
     if 'ch_names' in annot:
         annot['ch_names'] = [
             _prep_name_list(ch, 'write') for ch in annot['ch_names']]
-    annot.to_csv(fname)
+    annot.to_csv(fname, sep=sep)
```
hoechenberger, line 716: Unrelated to this PR, but shouldn't we pass `index=False` here?

mne/tests/test_annotations.py
agramfort: onset should be in seconds not time stamps
description should probably be called trial_type

hoechenberger:
I see several problem with the existing and this new behavior.

- The docstring doesn't mention which extensions, exactly, are allowed. It's only an implicit remark.
- We check if extensions match expected values in a case-sensitive way. I don't think this is a good idea.
- Looking at the docstring, it's entirely unclear to me what a the method will do to a file with a `.txt` extension.
- I wouldn't add a separate `.tsv` option if we already have `.csv`. Instead, add a `delim` param, and allow users to pick whichever extension they want.

agramfort:
the background is the following. I wanted to write time stamps to have
annotations
that are less data analysis step dependent (eg cropping,
resampling). @cbrnr wanted
to write plain integers in samples. I suggested we have .csv with time
stamps and txt
with samples.

cbrnr:
I think a function with several parameters such as `delim=','` (pandas calls it `delimiter`, and I've also seen `sep` somewhere), `time_zero=0`, and `unit='samples'` would be nice to unify writing annotations to plain text files.

mne/annotations.py
cbrnr: Yes!

hoechenberger:
+1 for a consolidation & API extension as proposed by @cbrnr 
But I don't understand the `time_zero` param? Shouldn't that be `raw.first_time` / `raw.first_samp`?

cbrnr:
> But I don't understand the time_zero param? Shouldn't that always be raw.first_time / raw.first_samp?

Sometimes it's useful to just set the beginning of the data to zero, thereby ignoring any `raw.first_time`/`raw.first_samp`.

agramfort:
my concern is that if we allow that flexibility in the writer we need to
open the door to
options in the reader. It's not anymore certain how to read these files
properly.
It will make th reader doing black magic to guess as best as possible how to
interpret the content.

cbrnr:
@agramfort that's true, but our current approach to infer the format from the extension is also brittle. CSV supports comments, so we could put all parameters into 1-2 header comment lines.

mne/tests/test_annotations.py
sappelhoff: > onset should be in seconds not time stamps
description should probably be called trial_type

Yes re: the first comment.

re: the second point --> "trial_type" is an optional column so for the test it may make more sense here. "description" would fall under an "additional arbitrary, user-defined column", so it's absolutely fine, but would have to be described in events.json

hoechenberger:
One could argue that users requiring that much flexibility should just use `Annotations.to_data_frame.to_*()`…

agramfort:
>
> One could argue that users requiring that much flexibility should just use
> Annotations.to_data_frame.to_*()…
>

+1

cbrnr:
> One could argue that users requiring that much flexibility should just use Annotations.to_data_frame.to_*()…

True, but how do I influence the units and starting point?

agramfort:
with pandas operations?

>

cbrnr:
> with pandas operations?

Of course. I thought it would be convenient to have a parameter for that, but if you think the distinction between .txt, .csv, and .tsv is clear enough that's fine with me.

hoechenberger:
> True, but how do I influence the units and starting point?

We'd need to add this to `Annotations.to_data_frame()`. Should be easy to do?

eort:
So, the consensus seems to be following: 

- pass `index=False` to all `to_csv()` calls
- let users choose the delimiter and (arbitrary) file extension with a `sep` keyword (for both writer and reader)
- properly describe the options in the doc
- make the io_test write a properly formatted bids tsv (seconds instead of timestamp)

Still open: 

- should there be other keywords? (i.e. unit, time_zero). if not, there should be some more information on how to manipulate those settings somewhere.

Am I missing anything?

Personally, I don't really have a strong opinion on whether those keywords would be added or not. However, I find it quite confusing to implicitly map a format (timestamp vs. samples) to a file extension (csv vs. txt). So, it should be made explicit one way or another. Either documentation, or adding a dedicated keyword to reader and writer, I lean towards latter.

agramfort:
my concern is the following. If users use Annotations.save we want to
guarantee ideally that
in the future they can just do read_annotations on their file and it just
works.
if we allow flexibility in the writer you ask future user to pass the
correct parameters in the
read_annotations.

if you want flexibility you can do:

df = annot.to_data_frame()
# hack as you want the df DataFrame then save with options you want
df.to_csv(..., )

see:

https://mne.tools/stable/generated/mne.Annotations.html#mne.Annotations.to_data_frame

for .tsv file for BIDS I would suggest you see how it's done in mne_bids
https://github.com/mne-tools/mne-bids/blob/main/mne_bids/write.py#L144
and make sure on _events.tsv file from a BIDS dataset can be read and saved
back with the new code you propose.

my 2c

>

eort:
Yeah that makes sense, but then the behaviors that each extension implies should be described in the docs somewhere. 

So following things are on the to-do list:
- `index=False` to the `.save('*.csv')` method
- a `.save('*.tsv')` option that has its own private writer `_write_annotations_tsv()`, producing BIDS compatible tsv files 
- documentation on that file extension determines which format the stored annotations will have (possibly also in the annotations tutorial)
- add a note that more flexibility can be achieved by first exporting annotations to a dataframe.

hoechenberger:
I see this has gone stale, what shall we do here?

eort:
The summer has kept me busy. It's still on my radar though. I hope to get to it during the fall. If you want it done earlier, feel free to take it on!

eort:
This would be how it looks without any adjustment
```
In [1]: raw.annotations.save('saved-annotations.tsv', overwrite=True)
   ...: raw.annotations.to_data_frame()
Overwriting existing file.
Out[1]: 
                       onset  duration description
0 2002-12-03 19:01:56.676071      1.00         AAA
1 2002-12-03 19:01:57.426071      0.75         AAA
2 2002-12-03 19:01:58.676071      0.50         BBB
3 2002-12-03 19:02:00.676071      0.25         CCC

In [2]: annot_from_file = mne.read_annotations('saved-annotations.tsv')
   ...: annot_from_file.to_data_frame()
Out[2]: 
                       onset  duration description
0 1970-01-01 00:00:45.955971      1.00         AAA
1 1970-01-01 00:00:46.705971      0.75         AAA
2 1970-01-01 00:00:47.955971      0.50         BBB
3 1970-01-01 00:00:49.955971      0.25         CCC
```

agramfort:
@eort I am not sure we agree on the specifics on the .tsv format you want to write? for me the idea is that you then use this function to write .tsv files in mne_bids although mne_bids does not have a dependency on pandas.

Well we need to agree of the objective I think

cbrnr:
Alright, I'm OK with adding a third file format then, provided that there are no additional parameters (e.g. time column starting at 0 or at `meas_date` etc.) and that this .tsv file can be used with BIDS without further changes. In other words, I wouldn't change the existing .csv and .txt writers and add a BIDS-compatible .tsv option.

agramfort:
well a .tsv does not make it BIDS compliant. So either we really make it
bids compliant or I don't see the strong need.

should we add this possibility in mne then maybe we could simplify code in
mne-bids but we will still have a pb
which is that mne-bids should not require pandas.

So maybe we can agree that we have more important battles to fight at this
point?

>

agramfort:
>
> If .tsv makes it easier to get annotations (events) into BIDS I still
> think this would be a nice addition (but I'm also OK if we decide against
> it). @eort <https://github.com/eort> please feel free to improve the docs
> for the two different formats, I find it difficult to understand the
> differences as well (actually, the differences between .csv and .txt are
> not documented at all).
>

+1 to improve the doc.


> IMO the best solution would still be to have parameters for delimiter
> (comma or tab) and time zero (0 or orig_time), which both writer and
> reader functions should share.
>

the pb is that these formats are underspecified. If you have a .tsv in bids
you know what you get on read (what column names / types etc.)
Here mne aims to force a specification so when you read it does the right
thing automatically. Once I have a written a file with certain
options I anticipate user will struggle to read back unless with made the
reader do black magic to guess things

eort:
Closed in favor of #10188

PR #9406, opened 2021-05-18
MRG, ENH: Export evokeds MFF

ephathaway:
In this PR we add support for exporting evoked(s) objects to MFF, as described in #8038 

For the API we add function `mne.export_evokeds`, which supports exporting to MFF format, but leaves room for adding other formats in the future.


mne/evoked.py
```diff
@@ -1436,6 +1436,77 @@ def _write_evokeds(fname, evoked, check=True):
...
+    %(export_params_fname)s
+    evoked : Evoked instance, or list of Evoked instances
+        The evoked dataset, or list of evoked datasets, to export to one file.
+        Note that the measurement info from the first evoked instance is used,
+        so be sure that information matches.
+    fmt : 'auto' | 'mff'
+        Format of the export. Defaults to ``'auto'``, which will infer the
+        format from the filename extension. See supported formats above for
+        more information.
+    device : None (default) | str
```
larsoner, line 1462: We have to be careful about future compat here. Each export could have zero or more such extra arguments, and I foresee a nasty parameter list if, say, one of these took 10 such arguments.

I don't see a great way around it (`**kwargs` seems worse for readability, and `mff_kwargs=None, eeglab_kwargs=None, ...` just punts on the issue. So how about we prefix the relevant exporter name:
```suggestion
    mff_device : None (default) | str
```
That way it will clearly group visually in the parameter list:
```
mff_device : ...
    ...
mff_history : ...
    ...
eeglab_foo : ...
    ...
brainvision_bar : ...
    ...
```

mne/evoked.py
```diff
@@ -1436,6 +1436,77 @@ def _write_evokeds(fname, evoked, check=True):
         end_file(fid)
 
 
+@fill_doc
+def export_evokeds(fname, evoked, fmt='auto', device=None, history=None):
```
larsoner, line 1440: If we force the `mff_*` entries to be provided as kwargs, it makes for nice future compat:
```suggestion
def export_evokeds(fname, evoked, fmt='auto', *, mff_device=None, mff_history=None):
```
because later we can add to the signature *before* the MFF-specific entries if there happens to be something common to all formats for example.

larsoner (commented):
Looks good, mostly minor things and one future compat idea!

mne/io/egi/egimff.py
```diff
@@ -9,6 +9,7 @@
 from xml.dom.minidom import parse
 
 import numpy as np
+import pytz
```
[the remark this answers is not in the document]
ephathaway: Good catch. Yeah sounds like a plan.

mne/evoked.py
```diff
@@ -1436,6 +1436,77 @@ def _write_evokeds(fname, evoked, check=True):
...
+    """Export evoked dataset to external formats.
+
+    Supported formats: MFF (mff, uses module mffpy)
+
+    .. warning::
+        Since we are exporting to external formats, there's no guarantee that
+        all the info will be preserved in the external format. To save in
+        native MNE format (``.fif``) without information loss, use
+        :func:`mne.write_evokeds` instead.
```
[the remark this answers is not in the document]
ephathaway: Yeah, I originally wanted to reuse that export warning message from raw and epochs, but it says to use the `save` function, which isn't quite equivalent in this case. I think we want to point the user to `write_evokeds`.

Are you suggesting change the shared warning message to something like:

```
To save in native MNE format (``.fif``) without information loss,
use :func:`save` instead or :func:`mne.write_evokeds` if exporting evokeds.
```

mne/evoked.py
ephathaway: Good call.

mne/evoked.py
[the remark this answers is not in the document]
larsoner: I would do:
```
docs.py:

docdict['export_warning'] = """\
.. warning: ... without information loss, use"""

evoked.py:

...
    %(export_warning)s :func:`mne.write_evokeds` instead.

epochs.py:

...
    %(export_warning)s :meth:`save` instead.
```
Make sense?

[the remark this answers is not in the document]
ephathaway: Ah yes! Didn't think of that.

mne/io/egi/tests/test_egi.py
```diff
@@ -369,7 +369,6 @@ def test_io_egi_evokeds_mff(idx, cond, tmax, signals, bads):
     assert evoked_cond.info['nchan'] == 259
     assert evoked_cond.info['sfreq'] == 250.0
     assert not evoked_cond.info['custom_ref_applied']
-    assert evoked_cond.info['dig'] is None
```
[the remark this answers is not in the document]
ephathaway: Yeah, so before with no categories.xml in the example MFF, no digitization data could be determined, so it `info['dig']` defaulted to `None`. Now, with the categories.xml, the digitization data can be read in.

mne/evoked.py
```diff
@@ -1436,6 +1436,73 @@ def _write_evokeds(fname, evoked, check=True):
...
+        more information.
+    mff_device : None (default) | str
+        If exporting to MFF format, it is required to specify the device on
+        which EEG was recorded (e.g. 'HydroCel GSN 256 1.0'). This is necessary
+        for determining the sensor layout and coordinates specs.
+    mff_history : None (default) | list of dict
+        If exporting to MFF format, it is optional to provide a list of history
+        entries (dictionaries) to be written to history.xml. This must adhere
+        to the format described in mffpy.xml_files.History.content. If None, no
+        history.xml will be written.
```
agramfort, line 1466: I am a bit uneasy about this. Ideally the export_evokeds function should have a format agnostic signature. Otherwise we have one export function per format. Also I see the word "required" while it defaults to None. I assume it would be highly recommended to save this information. Can we infer the device from channel names and/or their locations?

mne/io/egi/egimff.py
```diff
@@ -907,6 +909,142 @@ def _read_evoked_mff(fname, condition, channel_naming='E%d', verbose=None):
                        nave=nave, verbose=verbose)
 
 
+@fill_doc
+def export_evokeds_to_mff(fname, evoked, device, history):
```
agramfort, line 913: this function should be private.

export_evokeds will be AFAIK the public entry point to the API.

mne/io/egi/egimff.py
```diff
@@ -907,6 +909,142 @@ def _read_evoked_mff(fname, condition, channel_naming='E%d', verbose=None):
...
+    sampling_rate = int(info['sfreq'])
+
+    # Initialize writer
+    mffpy = _import_mffpy('Export evokeds to MFF.')
+    import pytz
+    writer = mffpy.Writer(fname)
+    record_time = info['meas_date'] or \
+        pytz.utc.localize(datetime.datetime.utcnow())
+    writer.addxml('fileInfo', recordTime=record_time)
+    writer.add_coordinates_and_sensor_layout(device)
```
agramfort, line 953: this writer checks that device is a valid string? I foresee many typos in the device name otherwise

mne/evoked.py
larsoner: > Otherwise we have one export function per format.

@agramfort I could live with this, as to me it's mostly the same. However, it goes against how things work for Raw and Epochs currently, so we'd want to switch those to have separate functions, too, otherwise the inconsistent interface isn't worth it to me. And this would require a deprecation cycle because we already released with the EEGLAB export.

agramfort:
ok we do what is done in read_raw

https://github.com/mne-tools/mne-python/blob/main/mne/io/_read_raw.py#L60

we accept kwargs and we also make the export_epochs_mff a public function
but not in a top level module

>

mne/evoked.py
ephathaway: Maybe it makes more sense to just have `mne.io.egi.export_evokeds_to_mff` as the API for exporting to MFF.

The device argument determines which `sensorLayout.xml` and `coordinates.xml` to write to the MFF directory. These are generic files that are determined only by the net type used to record EEG and they give the 2D and 3D positions of the sensors (see https://github.com/BEL-Public/mffpy/tree/develop/mffpy/resources/coordinates). The tricky part is that sometimes we have two different net types (e.g. 'HydroCel GSN 32 1.0' and 'MicroCel GSN 100 32 1.0') with the exact same sensor positions.

I think it's possible to determine the device type (or at least a device with the same coordinates) from `info['dig']`, but it would definitely be a chore.

As far as history, this is not something that is required for writing a valid MFF, we would appreciate the option to include it so that we can log processing steps in the output file.

ephathaway: There is also this `info['device_info']` key that could be utilized to store the device type upon reading MFFs and then used to determine the device upon write. I think in any case, at least having the option to provide the device as an argument upon write is a good idea.

mne/io/egi/egimff.py
ephathaway: Yes, the device string is used to lookup the corresponding generic coordinates and sensorLayout XMLs. If these are not found, an error is thrown.

mne/evoked.py
agramfort: yes good idea. We could set properly the "device_info" on read from mff file and you remind that there is an info['proc_history'] that we don't use enough. It's just used for maxfilter information AFAIK.
the less stuff we do that is custom the MFF the better. It's more work but I think it's better in the long term.

my 2x

agramfort:
I am fine if you do it in the same PR.

ephathaway:
The `info['proc_history']` does seem like a nice analog to the way we store processing history in MFF. I could see reading the entries from history.xml in the MFF read functions, storing this in `info['proc_history']`, then having a `.add_proc_history_entry` method so that a user can log an MNE processing step they took. Exporting to MFF could then write the contents of `info['proc_history']` to a history.xml file. This probably goes beyond the scope of this PR though.

mne/evoked.py
```diff
@@ -1436,6 +1436,66 @@ def _write_evokeds(fname, evoked, check=True):
...
+@fill_doc
+def export_evokeds(fname, evoked, fmt='auto', **kwargs):
+    """Export evoked dataset to external formats.
+
+    This function is a wrapper for format-specific export functions. The export
+    function is selected based on the inferred file format. All arguments are
+    passed to the respective export function.
+
+    Supported formats
+        MFF (mff, uses `mne.io.egi.egimff.export_evokeds_to_mff`)
```
ephathaway, line 1448: I'm not sure how to correctly reference this function. Would I add it to the file i/o python reference?

agramfort: I would suggest to expose it as `mne.io.egi.export_evokeds_to_mff`

I would insert section in the API page just before `Creating data objects from arrays` which I would call `Exporting to other file formats`

ok for you? thx

agramfort:
and yes I would keep the handling of proc_history for a future PR. It made we realize that we heavily under use this field in info. We could use it for provenance tracking....

mne/evoked.py
larsoner: As we have decided in the past that `mne.io` is for *raw* I/O only, I'm inclined to make it `mne.evokeds.export_evokeds_mff`

agramfort:
that would move a lot of EGI/MFF code outside of mne/io/egi and I don't
like that much.
For two reasons:
- it makes it easier to know who is most knowledgeable for certain
submodules
- it certainly limits circular imports

if you are concerned about our discourse that mne.io if for raw we could
say it's for
raw and exporting for non-fif files.

>

larsoner:
The only code outside mine/io/egi/ would be a single import line in mne/evoked.py, and there should be no circular import because mne/evoked.py already imports from IO, so I don't think those concerns are valid

agramfort:
this is case I prefer to have a new submodule mne.export that contains
functions to export to other file formats

>

larsoner:
So for Raw we have a method `raw.export(...)` and for Epochs we have `epochs.export(...)` that were part of 0.23. Is the proposal to have `mne.export` where just evoked exporting goes...? To me it makes more sense to use the namespaces of the objects (Raw, Epochs) or submodules (mne.evoked.export..., mne.io.export_raw_whatever, mne.epochs.export...) than to use a separate `mne.export` submodule. But I can live with a new `mne.export` if others agree it's cleaner to split evoked into its own submodule rather than splitting by object type / method / group

drammock:
@agramfort expressed the concern of keeping all the EGI/MFF-related code in the `mne.io.egi` submodule. I think that's a good idea (even if there were past discussions about `io` being only for raw data).  But I don't see why that means that *the API entry point* has to be inside `mne.io.egi`.  As @larsoner points out, all that would be needed is a relative import from `mne.io.egi` into whatever namespace where we want the function to live.  Here are the candidates:

1. `mne.io.egi.export_evokeds_mff`
2. `mne.evokeds.export_evokeds_mff`
3. `mne.export_evokeds_mff`  (same namespace as `mne.read_evokeds`, `mne.read_evokeds_mff`, and `mne.write_evokeds`)

If I'm reading everyone correctly, @agramfort prefers option 1, @larsoner prefers option 2, and I prefer option 3.  Anybody want to change their vote?  Anyone else want to weigh in?

agramfort:
my concern is that the mne namespace is already super crowded. We have it
seems more than 200 functions / classes already in the mne namespace...
maybe it's too late to raise this concern but I felt it needed to be said :)

I would vote for 1 as I don't expect users to use this function what that
want to do. Evoked.export should already do the trick I think.

>

drammock:
> I would vote for 1 as I don't expect users to use this function what that want to do. Evoked.export should already do the trick I think.

I thought `Evoked.export` was not an option because we want to support writing *multiple* evokeds to one file?

agramfort:
sorry I meant export_evokeds (not Evoked.export)

PS @ephathaway can you have a look at https://mne.discourse.group/t/unable-to-load-mff-files-parseerror-not-well-formed-invalid-token/3121/7 ?

larsoner:
@agramfort what is the advantage of having the API entry point be `mne.io.egi.export_evokeds_mff` (what you prefer) instead of `mne.evoked.export_evokeds_mff`?

agramfort:
I want to hide it a bit as to me it's more of an expert feature.

>

larsoner:
Given that we have `raw.export(...)` and `epochs.export(...)`, I don't think we should try that hard to hide the equivalent Evoked functionality (i.e., three levels deep, in the module for raw file I/O). Having it in `mne.evoked` rather than the `mne` already seems hidden / well-scoped / expert-y enough.

larsoner:
I had a quick chat with @agramfort and this is what we converged on:

1. All export-related code lives in a new `mne.export` submodule, which is for exporting data to any format other than our two standard ones (FIF and H5)
2. This PR would add `mne.export.export_evokeds` (with no writer-specific kwargs) implemented in `mne/export/_export.py` and `mne.export.export_evokeds_mff`, with actual implementation living in `mne/export/mff/mff.py` or so
3. Existing raw and epochs exporting code can be moved to `mne.export.export_raw` and `mne.export.export_epochs`
4. A new Evoked method `Evoked.export(...)` that calls `mne.export.export_evokeds` under the hood

@ephathaway @drammock okay for you?

If so, I propose:

1. I make a PR to initialize the `mne.export` submodule and move Epochs and Raw code there, and we merge
2. I merge `upstream/main` into this PR
3. I push commits to this PR to reflect the structure described above (it will be faster than trying to explain all details I left out above I think)

drammock:
Ok for me.

larsoner:
@ephathaway let me know if you're okay with me pushing a couple of commits to do a merge with `upstream/master` and reorganize the code a bit according to the plan above, I should be able to do it Tuesday

ephathaway:
> @ephathaway let me know if you're okay with me pushing a couple of commits to do a merge with `upstream/master` and reorganize the code a bit according to the plan above, I should be able to do it Tuesday

Your plan sounds good to me. Go for it!

My only (minor) concern is that it may confuse users a bit to have an `io` module and an `export` module that both deal with foreign file formats.

larsoner:
> My only (minor) concern is that it may confuse users a bit to have an io module and an export module that both deal with foreign file formats.

The `io` name has always been a bit of a misnomer. I think here if we make it as clear as possible (and stick to the notion) that `mne.io.*` publict funtions are only for reading raw data I think we'll be okay.

larsoner:
@ephathaway I can't push, can you tick the box "Allow edits by maintainers"?

Screenshot from 2021-06-02 09-48-58

ephathaway:
> @ephathaway I can't push, can you tick the box "Allow edits by maintainers"?

Since the forked repo is under our organization instead of my own GH account, this option doesn't show up. I gave you write access to our fork, which should do the trick.

mne/io/egi/egimff.py
```diff
@@ -887,16 +891,21 @@ def _read_evoked_mff(fname, condition, channel_naming='E%d', verbose=None):
...
-        if entry['method'] == 'Montage Operations Tool':
-            if 'Average Reference' in entry['settings']:
-                # Average reference has been applied
-                projector, info = setup_proj(info)
-            else:
-                # Custom reference has been applied that is not an average
-                info['custom_ref_applied'] = True
+    try:
+        fp = mff.directory.filepointer('history')
+    except ValueError:  # should probably be FileNotFoundError upstream...
```
larsoner, line 896: FYI @ephathaway I had to add this, with the code before writing with `history=None` raised an error when the resulting file was read with `read_evokeds_mff`. This catches the error that the history file is not written. Upstream in mffpy this should probably be a FileNotFoundError rather than ValueError. I can open a PR for this in mffpy if you agree

ephathaway: That would be great. Thanks! And yes, I think we should treat history.xml as optional.

larsoner: https://github.com/BEL-Public/mffpy/pull/89

Pushed a commit to make this PR compatible with ValueError (backward compat) and FileNotFoundError (future compat)

larsoner (approved):
LGTM +1 for merge. @agramfort want to see if this reflects the design we had in mind?

agramfort (approved):

agramfort:
merge when green

larsoner:
Thanks @ephathaway !

larsoner:
I just realized we forgot `latest.inc`, will push a fix to `main`

Issue #8986, opened 2021-03-03
BUG: Possible inefficient updates to Brain


larsoner:
One reason `Brain` is (probably) slower on CIs and example builds is that it does at least some unnecessary updates. For example on #8985 + #8975 if I add a `mne.set_log_level('debug')` in some test script, each time I tweak a clim or update a time point I get:
```
>>> brain.set_time_point(0)
Color mapping 'curv' with Greys colormap and range [-1, 2]
Color mapping 'data' with <class 'numpy.ndarray'> colormap and range [4.73, 14.18]
Color mapping 'curv' with Greys colormap and range [-1, 2]
Color mapping 'data' with <class 'numpy.ndarray'> colormap and range [4.73, 14.18]
```
So it's being mapped twice. And if I modify the `_Brain` code to have this:
```
    def _update(self):
        logger.debug('Update')
```
Just on init I see:
```
Update
Update
Update
Update
Update
Update
```
So we get 6 renderer updates. This last part might not actually matter since just having `_update` return immediately it's still ~4 seconds to open the window and exit.

So this is mostly an issue to remind us to track down and profile to see where time is being taken.

GuillaumeFavelier:
We opted-out from `auto_update` so for this one, I would suggest to add an `update` parameter to the public functions that would default to `True`. With this solution, the internal calls can have `update=False`. What do you think?

larsoner:
I think right now we don't have to worry about any public updates, and can just assume that if someone does something, update should be called. The real problem as I see it so far is that update is called more than once for a single "do something" command from the user's end. For example, `set_time_point` above causes a double recomputation of the overlay scalars/colormapping, even though it should only happen once.

With the changes from #8985 for example we could update a smoke test of `set_time_point(0)` to check the update count for example with:
```
with catch_logging('debug') as log:
    brain.set_time_point(0)
log = log.getvalue()
assert log.count('colormap and range') == 2, log  # one curv + one hemi data
```
and this would fail on `main` because the count would be 4.

PR #9407, opened 2021-05-19
MNT: Reduce number of calls to _update

GuillaumeFavelier:
1) This PR reduces the amount of renderer updates. I modified the two main entry points: `__init__()` and `setup_time_viewer()`.
For now I use a context manager but if big chunks of code need to be nested maybe `enable/disable` functions are better, not sure.
I'm still trying to download 0.119 locally so testing is delayed but in the example that I use regularly, it went from 16 to 1.

2) I'll see how to reduce the amount of overlay updates next.
3) I'll also investigate why the window is so slow to setup (suggested in https://github.com/mne-tools/mne-python/pull/9407#issuecomment-844199125)

Closes https://github.com/mne-tools/mne-python/issues/8986

### DIFF
--- a/mne/morph.py
+++ b/mne/morph.py
@@ -1142,7 +1142,7 @@ def _hemi_morph(tris, vertices_to, vertices_from, smooth, maps, warn):
     e = mesh_edges(tris)
     e.data[e.data == 2] = 1
     n_vertices = e.shape[0]
-    e = e + sparse.eye(n_vertices)
+    e += sparse.eye(n_vertices, format='csr')
     if isinstance(smooth, str):
         _check_option('smooth', smooth, ('nearest',),
                       extra=' when used as a string.')
--- a/mne/source_space.py
+++ b/mne/source_space.py
@@ -1357,22 +1357,22 @@ def read_talxfm(subject, subjects_dir=None, verbose=None):
     return mri_mni_t
 
 
-def _read_mri_info(path, units='m', return_img=False):
+def _read_mri_info(path, units='m', return_img=False, use_nibabel=False):
     # This is equivalent but 100x slower, so only use nibabel if we need to
     # (later):
-    #
-    # import nibabel
-    # hdr = nibabel.load(path).header
-    # hdr = mgz.header
-    # n_orig = hdr.get_vox2ras()
-    # t_orig = hdr.get_vox2ras_tkr()
-    # dims = hdr.get_data_shape()
-    # zooms = hdr.get_zooms()[:3]
-    hdr = _get_mgz_header(path)
-    n_orig = hdr['vox2ras']
-    t_orig = hdr['vox2ras_tkr']
-    dims = hdr['dims']
-    zooms = hdr['zooms']
+    if use_nibabel:
+        import nibabel
+        hdr = nibabel.load(path).header
+        n_orig = hdr.get_vox2ras()
+        t_orig = hdr.get_vox2ras_tkr()
+        dims = hdr.get_data_shape()
+        zooms = hdr.get_zooms()[:3]
+    else:
+        hdr = _get_mgz_header(path)
+        n_orig = hdr['vox2ras']
+        t_orig = hdr['vox2ras_tkr']
+        dims = hdr['dims']
+        zooms = hdr['zooms']
 
     # extract the MRI_VOXEL to RAS (non-zero origin) transform
     vox_ras_t = Transform('mri_voxel', 'ras', n_orig)
--- a/mne/source_space.py
+++ b/mne/source_space.py
@@ -2251,6 +2251,8 @@ def _vol_vertex(width, height, jj, kk, pp):
 
 def _get_mgz_header(fname):
     """Adapted from nibabel to quickly extract header info."""
+    fname = _check_fname(fname, overwrite='read', must_exist=True,
+                         name='MRI image')
     if not fname.endswith('.mgz'):
         raise IOError('Filename must end with .mgz')
     header_dtd = [('version', '>i4'), ('dims', '>i4', (4,)),
--- a/mne/source_space.py
+++ b/mne/source_space.py
@@ -1358,21 +1358,21 @@ def read_talxfm(subject, subjects_dir=None, verbose=None):
 
 
 def _read_mri_info(path, units='m', return_img=False):
-    if has_nibabel():
-        import nibabel
-        mgz = nibabel.load(path)
-        hdr = mgz.header
-        n_orig = hdr.get_vox2ras()
-        t_orig = hdr.get_vox2ras_tkr()
-        dims = hdr.get_data_shape()
-        zooms = hdr.get_zooms()[:3]
-    else:
-        mgz = None
-        hdr = _get_mgz_header(path)
-        n_orig = hdr['vox2ras']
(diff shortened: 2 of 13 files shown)


larsoner:
Notebook test is failing, seems related. 

In any case, this is the slowest `_brain` test on macOS run on this PR:
```
9.10s call     mne/viz/_brain/tests/test_brain.py::test_brain_init[pyvista]
```
and on `main` it's:
```
9.36s call     mne/viz/_brain/tests/test_brain.py::test_brain_init[pyvista]
```
So not a huge difference. Do you notice any timing difference locally @GuillaumeFavelier ? It might be the case that calling `._update()` multiple times does not have much effect on performance if VTK is smart about not bothering to re-render until actually needed. For example I just replaced a single call to `_update` in `set_time_point` to be:
```
        print('up')
        for _ in range(1000):
            self._renderer._update()
```
and I see "up" printed a bunch and get the same FPS in both cases -- it's low on my laptop, like 3, so if it's actually re-rendering each time it seems like it should drop to updating less than once a minute. So I'm not sure this will gain us too much?

On the other hand, if setting the camera forces a re-render somehow at the PyVista end (by doing a .Render call directly rather than plotter.update() maybe?) -- which seems possible since sometimes the brains switch views? Might also be worth checking the logic in their update method to understand things if you haven't already (I had a quick look and don't fully understand the logic):

https://github.com/pyvista/pyvista/blob/master/pyvista/plotting/plotting.py#L986-L1012

GuillaumeFavelier:
Locally, the average of 10 calls for `test_brain_init`:

main | PR
--|--
7.79s | 7.71s

The impact is limited to say the least.

I won't lie, this `update` method is quite a mystery for me. I just assumed that in the case of `BackgroundPlotter` it was equivalent to calling `render()` since `force_redraw=True`. I will verify this.

larsoner:
VTK might also be smart enough internally to figure out if anything in the scene has changed, so even if you `force_render` it might not actually do it. Not sure...

In any case, worth looking into!

GuillaumeFavelier:
Because of the inheritance of `BackgroundPlotter`, this `udpate()` is the not the one we think...

```
In [9]: brain._renderer.plotter.update?
Docstring:
update(self)
update(self, QRect)
update(self, QRegion)
update(self, int, int, int, int)
Type:      builtin_function_or_method
```

It is probably a Qt or VTK function :sweat_smile:

GuillaumeFavelier:
Sice this is clarified, I'll revert and I'll switch to the second part of the PR namely minimizing the number overlay updates.

There is not much I can do for the first part except maybe minimize the number of calls to `processEvents` and I doubt the effect will be so important.

larsoner:
One thing I've noticed is slow is loading the window in the first place. I would start by making a script that just loads and closes maybe 10 times, `@profile` some functions, and `kernprof -lv python whatever.py` it and see what's actually eating all the time.

GuillaumeFavelier:
The latest commit 8067b94 drastically improves the caching system of overlays:
* new partial update code to avoid iterating over all the overlays (basically, the cost of `set_time_point()` has decreased)
* no unnecessary full updates during overlay removal (this happened before in picking routines)
* the cache is updated only during full update or when a new overlay is added
I also updated the tests.

I think it can be improved further but a complexity analysis would probably be necessary to evaluate.

GuillaumeFavelier:
So far, using `line_profiler` on `_plot_stc()`:
- `30%` by `brain = Brain()`
  - `60%` by `geo.load_geometry()`
  - `20%` by `self._setup_canonical_rotation()`
- `30%` by `brain.add_data()`
- `40%` by `brain.setup_time_viewer()`
  - `60%` by `self._configure_dock()`
    - `98%` by `self._configure_dock_trace_widget()`
      - `99%` by `_set_annot('None')`
  - `30%` by `self._renderer.show()`

GuillaumeFavelier:
Yes, you are right it takes all the time

GuillaumeFavelier:
So far, using `line_profiler` on `_plot_stc(show_traces=False)`:

- `42%` by `brain = Brain()`
- `40%` by `brain.add_data()`
- `12%` by `brain.setup_time_viewer()`

<details>
<summary>Script</summary>

```py
import os
import mne

initial_time = 0.096

for i in range(10):
    brain = stc.plot(
        size=600,
        hemi='rh',
        show_traces=False,
        views=['lat', 'med'],
    )
    brain.close()
```

</details>

GuillaumeFavelier:
So far, using `line_profiler` on `_plot_stc(time_viewer=False, show_traces=False)`:

- `45%` by `brain = Brain()`
- `45%` by `brain.add_data()`

<details>
<summary>Script</summary>

```py
import os
import mne

initial_time = 0.096

for i in range(10):
    brain = stc.plot(
        size=600,
        hemi='rh',
        time_viewer=False,
        show_traces=False,
        views=['lat', 'med'],
    )
    brain.close()
```

</details>

GuillaumeFavelier:
After 0a31936, the average of 3 x 10 runs:

master | PR
--|--
22.6s | 19.3s

GuillaumeFavelier:
Same observation as https://github.com/mne-tools/mne-python/pull/9407#issuecomment-845193301, interfacing with `matplotlib` seems costly. I tried to keep updates minimal at least at init but now the path forward becomes blurry.

What I can say is that any significant improvement to `load_geometry()`, `_hemi_morph()` or `vertex_to_mni()` should benefit `Brain`'s loading time.

GuillaumeFavelier:
This PR is ready for reviews on my end @larsoner

larsoner:
`main` I get 20.0, this branch I get 17.9, with my commit (speeding up `vertex_to_mni`) I get 16.9 sec. I say we merge this and tackle the morphing separately. I'll fix the MFF problem on `main` first, then restart CIs, then let's merge!

larsoner:
@GuillaumeFavelier I pushed a commit that uses `reset_camera=False` and `update=False` to avoid some costly `camera.reset()` and `process_events` calls, can you take a look and re-benchmark at your end? Once CIs come back green if you're happy with my changes and resulting benchmark options, feel free to merge

larsoner:
(my changes optimize `load_geometry()`, `_hemi_morph()`, and `vertex_to_mni()` to the extent I found possible BTW)

GuillaumeFavelier:
After 0a31936, the average of 3 x 10 runs:

master | PR
--|--
22.6s | 15s 

It feels way better in my opinion so I am happy with those changes :)

The failure of `compat / minimal / py3.7` seems legit though.

GuillaumeFavelier:
Probably `test_scale_mri` and `test_volume_source_space` could use `requires_nibabel`

larsoner:
Thanks @GuillaumeFavelier !

larsoner:
A couple of nice things testing-wise, on another PR that ran recently we had:
```
515.01s total  mne/viz/_brain/tests/
...
25.19s call     mne/viz/_brain/tests/test_brain.py::test_brain_traces[pyvista-mixed-split]
```
and on the last run for this PR it was:
```
426.48s total  mne/viz/_brain/tests/
...
17.86s call     mne/viz/_brain/tests/test_brain.py::test_brain_traces[pyvista-mixed-split]
```

PR #9412, opened 2021-05-20
FIX : not disappearing vline in Raw/Epochs-Plot

marsipu:
The vertical green line appearing after a left-click inside the Raw/Epochs-Plot should disappear after a right-click somewhere in the plot, which is fixed by this PR.

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -2218,7 +2218,7 @@ class MNEBrowseFigure(MNEFigure):
                 artist.set_visible(visible)
                 self.draw_artist(artist)
         self.mne.vline_visible = visible
-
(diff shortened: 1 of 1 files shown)


marsipu:
I first tried the existing `_redraw`-function, but this failed with `test_plot_epochs_keypresses` due to a conflict during `_draw_traces`. I originally thought it would be good to keep the `_recompute_epochs_vlines`.
Do you think this is still important when just the visible-status changes for the vlines?
I saw no flickering with just `draw_idle`, how was it originally provoked?

drammock (approved):

cbrnr:
Thanks @marsipu!

PR #9414, opened 2021-05-20
Fix error when reducing shown time-window while vline visible

### LINKED ISSUE #9413: BUG: Raw.plot() error when changing duration while vline visible
#### Describe the bug
When the vertical line is visible and the shown duration is reduced by pressing `End`, nothing happens and an error occurs due to l.713-714 in mne.viz._figure.py:
```python
if key == 'end' and self.mne.vline_visible:  # prevent flicker
    self._show_vline(None)
```
```python
  File "/home/martins/miniconda3/envs/mnedev/lib/python3.9/site-packages/matplotlib/cbook/__init__.py", line 270, in process
    func(*args, **kwargs)
  File "/home/martins/PycharmProjects/mne-python/mne/viz/_figure.py", line 714, in _keypress
    self._show_vline(None)
  File "/home/martins/PycharmProjects/mne-python/mne/viz/_figure.py", line 2210, in _show_vline
    self.mne.vline_text.set_text(f'{xdata:0.2f} s ')
TypeError: unsupported format string passed to NoneType.__format__
```

What is the preventing of flicker about?
Would you rather remove the line or handle None as argument?

#### Steps to reproduce

1. Plot Raw as in:
```python
sample_data_raw_file = os.path.join(sample_data_folder, 'MEG', 'sample', 'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file)
raw.plot(block=True)
```
2. Make vertical line visible by clicking inside plot
3. Press `End` on keyboard.

#### Expected results
Duration changes.

#### Actual results
Duration doesn't change, error is thrown instead.

marsipu:
#### Reference issue
Example: Fixes #9413.

The error happening when changing the shown duration by pressing `End` is fixed.

#### Additional information
I am not sure, what the problem with the flickering was which motivated this line. I assumed that redrawing the vline was the intention, so I replaced `_show_vline` directly with `_toggle_vline`.

### DIFF
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -710,8 +710,6 @@ class MNEBrowseFigure(MNEFigure):
                 if self.mne.t_start + self.mne.duration > last_time:
                     self.mne.t_start = last_time - self.mne.duration
                 self._update_hscroll()
-                if key == 'end' and self.mne.vline_visible:  # prevent flicker
-                    self._toggle_vline(True)
                 self._redraw(annotations=True)
         elif key == '?':  # help window
             self._toggle_help_fig(event)
--- a/mne/viz/_figure.py
+++ b/mne/viz/_figure.py
@@ -711,7 +711,7 @@ class MNEBrowseFigure(MNEFigure):
                     self.mne.t_start = last_time - self.mne.duration
                 self._update_hscroll()
                 if key == 'end' and self.mne.vline_visible:  # prevent flicker
-                    self._show_vline(None)
+                    self._toggle_vline(True)
                 self._redraw(annotations=True)
         elif key == '?':  # help window
             self._toggle_help_fig(event)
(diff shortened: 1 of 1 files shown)


marsipu:
The commit-name is false, it should be "expanding" instead of reducing.

drammock:
I've pushed a commit to fix a misleading comment (left over from an earlier refactoring).  The "flickering problem" relates to epoch figures, where if you increased the number of epochs displayed (via the `end` key) then the additional epoch would show up, and then a bit later the green line would appear in it.  But it turns out that this is no longer a problem now that blitting has been disabled, so we can just remove those two lines.

drammock (approved):

agramfort (approved):

agramfort:
thx @marsipu @drammock

Issue #9335, opened 2021-04-22
Show times relative to `meas_date` in `raw.plot`


cbrnr:
Based on my question in our forum (https://mne.discourse.group/t/show-times-relative-to-meas-date-in-raw-plot/3027/3), I think it would be useful to show time stamps based on `meas_date` on the x-axis of `raw.plot`. This should be optional, and by default `raw.plot` should stick to showing time starting at 0.

larsoner:
In terms of a mechanism for doing it, I would make a new tick formatter. Matplotlib decides where the ticks go, and the formatter takes the x values (which are just from `raw.times` starting from zero) and incorporates whatever information about `first_samp`, `raw.annotations.meas_date` etc. and returns the datetime string.

cbrnr:
I can give it a try, but not for 0.23 (obviously).

cbrnr:
Fixed in #9419.

PR #9419, opened 2021-05-21
ENH: Add time_format to Raw.plot()

marsipu:
#### Reference issue
Tackles #9335 (@cbrnr).

This adds a parameter (time_format) to Raw.plot() to display the real time derived from Info['meas_date'].


marsipu:
I tried many ways of escaping in RST, but none allowed me to include "%" from the datetime format code.
Do you know another way?

marsipu:
When I try to build the documentation with `make html_dev-noplot`, I get:
```
Handler <function make_redirects at 0x0000027ECEBCDEE0> for event 'build-finished' threw an exception (exception: 'auto_tutorials' is not in list)
```
I assume it is not related to the changes from this PR, right?. I already updated my sphinx-installation from `requirements_doc.txt`.

agramfort:
your doc CIs are green. Do you have this pb locally?

mne/viz/ica.py
```diff
@@ -59,6 +60,11 @@ def plot_ica_sources(ica, inst, picks=None, start=None,
         plotter. For evoked, this parameter has no effect. Defaults to False.
     show_first_samp : bool
         If True, show time axis relative to the ``raw.first_samp``.
+    show_real_time : bool | str
+        If True or a `datetime format code <https://docs.python.org/3/library
+        /datetime.html#strftime-and-strptime-format-codes>`_, the labels for
+        x-axis ticks will be the real time derived from the Info['meas_date'].
+        Default format code if True will be ``H:M:S``.
```
larsoner, line 67: To escape `%` when you plan on using `%` formatting later (this is what `docdict` / `docs.py` does under the hood) you double `%`:
```
>>> '%(test)s %%H:%%M%%S' % dict(test='whatever')
'whatever %H:%M%S'
```
so:
```suggestion
        Default format code if True will be ``'%%H:%%M:%%S'``.
```

marsipu: > 
> 
> your doc CIs are green. Do you have this pb locally?

Yes, its locally (on Windows). Maybe I redo the mnedev-environment and see if the pb persists.

marsipu: Thank you!

marsipu:
> 
> 
> your doc CIs are green. Do you have this pb locally?

Yes, its locally (on Windows). Maybe I redo the mnedev-environment and see if the pb persists.

marsipu:
And locally I can't replicate the error in test_epochs::test_metadata from the CIs:
```
mne\tests\test_epochs.py:2766: in test_metadata
    meta = np.array([[1.] * 5 + [3.] * 5,
E   FutureWarning: Promotion of numbers and bools to strings is deprecated. In the future, code such as `np.concatenate((['string'], [0]))` will raise an error, while `np.asarray(['string', 0])` will return an array with `dtype=object`.  To avoid the warning while retaining a string result use `dtype='U'` (or 'S').  To get an array of Python objects use `dtype=object`. (Warning added in NumPy 1.21)
```

larsoner:
I pushed a fix for that in #9422, feel free to ignore it. It's due to using latest NumPy where they've deprecated something

mne/io/base.py
```diff
@@ -1525,15 +1525,17 @@ def plot(self, events=None, duration=10.0, start=0.0, n_channels=20,
              event_color='cyan', scalings=None, remove_dc=True, order=None,
              show_options=False, title=None, show=True, block=False,
              highpass=None, lowpass=None, filtorder=4, clipping=_RAW_CLIP_DEF,
-             show_first_samp=False, proj=True, group_by='type',
-             butterfly=False, decim='auto', noise_cov=None, event_id=None,
-             show_scrollbars=True, show_scalebars=True, verbose=None):
+             show_first_samp=False, show_real_time=False, proj=True,
+             group_by='type', butterfly=False, decim='auto', noise_cov=None,
+             event_id=None, show_scrollbars=True, show_scalebars=True,
+             verbose=None):
```
cbrnr, line 1531: I know that this parameter logically belongs next to `show_first_samp`, but usually we tack new parameters to the end (just before `verbose`, which should always be last). I'm not sure it makes a lot of sense in this function though, because we already have so many parameters and nobody should rely on using positional arguments. But we haven't enforced keyword args, so we should probably stick to our conventions.

doc/changes/latest.inc
```diff
@@ -45,6 +45,8 @@ Enhancements
 
 - Add projections when printing a :class:`mne.Info` in the notebook (:gh:`9403` by `Alex Gramfort`_)
 
+- Add "show_real_time" to :meth:`mne.io.Raw.plot` and :meth:`mne.preprocessing.ICA.plot_sources` to display acquisition time on x-axis (:gh:`9419` by `Martin Schulz`_)
```
cbrnr, line 48: I have the feeling that there should be a better parameter name than `show_real_time`, but currently I can't think of a better name. Maybe someone else has a suggestion?

mne/viz/_figure.py
```diff
@@ -447,6 +447,15 @@ def __init__(self, inst, figsize, ica=None, xlabel='Time (s)', **kwargs):
             # hide some ticks
             ax_main.tick_params(axis='x', which='major', bottom=False)
             ax_hscroll.tick_params(axis='x', which='both', bottom=False)
+        else:
+            # Add Timestamp X-Ticks if activated
+            if self.mne.show_real_time:
```
cbrnr, line 452: Why not combine this into an `elif` branch?

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,22 @@ def _update_yaxis_labels(self):
                    else 'normal')
             text.set_style(sty)
 
+    def _xtick_timestamp_formatter(self, _, xval):
```
cbrnr, line 1866: What does the `_` parameter do?

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,22 @@ def _update_yaxis_labels(self):
...
+        import datetime
+        first_time = self.mne.inst.first_time
+        xdatetime = self.mne.inst.info['meas_date'] + \
+            datetime.timedelta(seconds=xval + first_time,
+                               milliseconds=int(first_time % 1 * 1000))
+        try:
+            xdtstr = xdatetime.strftime(self.mne.show_real_time)
+        except ValueError:
+            logger.warning(f'{self.mne.show_real_time} is not a valid '
+                           f'datetime-format-string!')
```
cbrnr, line 1877:
```suggestion
                           f'datetime format string!')
```

mne/viz/_figure.py
```diff
@@ -2204,7 +2229,11 @@ def _show_vline(self, xdata):
         else:
             self.mne.vline.set_xdata(xdata)
             self.mne.vline_hscroll.set_xdata(xdata)
-        self.mne.vline_text.set_text(f'{xdata:0.2f} s ')
+        if self.mne.show_real_time:
+            text = self._xtick_timestamp_formatter(xdata, None)
+        else:
+            text = f'{xdata:0.2f} s '
```
cbrnr, line 2235: I wonder if we should have a `start` parameter in `self._xtick_timestamp_formatter`, which would make this condition unnecessary? Also, what about the `None` argument (see my previous question)? Could you set the second argument to `None` by default?

mne/viz/ica.py
```diff
@@ -28,7 +28,8 @@
 @fill_doc
 def plot_ica_sources(ica, inst, picks=None, start=None,
                      stop=None, title=None, show=True, block=False,
-                     show_first_samp=False, show_scrollbars=True):
+                     show_first_samp=False, show_real_time=False,
+                     show_scrollbars=True):
```
cbrnr, line 32: Same comment regarding parameter order, the new parameter should probably be appended to the end.

mne/viz/ica.py
```diff
@@ -59,6 +60,11 @@ def plot_ica_sources(ica, inst, picks=None, start=None,
         plotter. For evoked, this parameter has no effect. Defaults to False.
     show_first_samp : bool
         If True, show time axis relative to the ``raw.first_samp``.
+    show_real_time : bool | str
+        If True or a `datetime format code <https://docs.python.org/3/library
+        /datetime.html#strftime-and-strptime-format-codes>`_, the labels for
+        x-axis ticks will be the real time derived from the Info['meas_date'].
+        Default format code if True will be ``'%%H:%%M:%%S'``.
```
cbrnr, line 67:
```suggestion
        If True or a `datetime format code <https://docs.python.org/3/library
        /datetime.html#strftime-and-strptime-format-codes>`_, the labels for
        x-axis ticks will be the time derived from ``Info['meas_date']``.
        Default format code if True is ``'%%H:%%M:%%S'``.
```

cbrnr (changes requested):

doc/changes/latest.inc
marsipu: Maybe `time_ticks` or `clock_ticks`?

mne/viz/_figure.py
marsipu: One of the two is the position on the axis while the other one is the displayed value. Thanks, I just realised that I mixed those two and time was not showing correctly.

marsipu: The first should be the value and the second the position

mne/viz/_figure.py
marsipu: I'll do it. Do you agree with the decision to exclude the epochs-plot? I thought, the application of this feature would be most helpful with continous data.

mne/viz/_figure.py
marsipu: > I wonder if we should have a start parameter in self._xtick_timestamp_formatter, which would make this condition unnecessary?

This one I don't understand, what would the start-parameter do?

mne/viz/_figure.py
cbrnr: Yes, I agree it's most useful for continuous plots.

mne/viz/_figure.py
cbrnr: I thought that we could always call `self._xtick_timestamp_formatter`, even for the default case where time starts at zero and is always shown in seconds. You could handle this case inside the method.

doc/changes/latest.inc
agramfort: some thoughts:

time_axis = 'relative' | 'absolute'

or

time_axis = 'datetime' | ???

mne/viz/_figure.py
```diff
@@ -447,6 +447,13 @@ def __init__(self, inst, figsize, ica=None, xlabel='Time (s)', **kwargs):
             # hide some ticks
             ax_main.tick_params(axis='x', which='major', bottom=False)
             ax_hscroll.tick_params(axis='x', which='both', bottom=False)
+        elif self.mne.show_real_time:
+            if not isinstance(self.mne.show_real_time, str):
+                self.mne.show_real_time = "%H:%M:%S"
+            for _ax in (ax_main, ax_hscroll):
+                _ax.xaxis.set_major_formatter(
+                    FuncFormatter(self._xtick_timestamp_formatter))
+                _ax.set_xlabel(f'Time ({self.mne.show_real_time})')
```
agramfort, line 456:
```suggestion
                _ax.set_xlabel(f'Time (hour/min/sec)')
```
less geeky

agramfort (commented):
I tested on some over night sleep EEG data and it worked great.

I am thinking we should be able to toggle the time format with eg the 't' shortcut. Doing so I am less worried about parameter API so I will expect users to toggle from the GUI the option.

my 2c

mne/viz/_figure.py
cbrnr: Or maybe `Time (hh:mm:ss)` (should also be less geeky but more exact)...

mne/viz/_figure.py
marsipu: I understand. This would probably also go nicely with the planned toggle-behaviour. I can look into it

mne/viz/_figure.py
marsipu: This was supposed to change, when the user provides its own datetime-format.

Originally I thought it would be nice, if you can decide the datetime-format yourself by providing the datetime format code.
But probably there is not that much variation from the default format H/M/S, which would be more useful for the user, right?
**Should we just provide the default?**
That would also make implementing the zoom-in/zoom-out format-changes more straight-forward?

drammock: Letting users pass a custom datetime format code seems like overkill to me. HH:MM:SS.SSS ought to be adequate for all users needs. If it's not, we can add that enhancement only when someone requests it.

marsipu:
@agramfort, @cbrnr Thank you for the comments&ideas, I will look into the toggle-behaviour and the zoom-dependent format-changes.

drammock:
For the API, what about `time_format = 'float' | 'datetime'` ?

agramfort:
>
> For the API, what about time_format = 'float' | 'datetime' ?
>
I like that !

mne/viz/_figure.py
marsipu: But in this specific case for the vline where an "s" is appended to the displayed time only for `time_format='float'`, I think it is still necessary to have this check inside `_show_vline`.

marsipu:
If you view the plot with a "usual" duration-window, milliseconds aren't probably that important. By adding up `first_time` and `meas_date` considering milliseconds (and even microseconds for larger Zoom-Levels), these will introduce an offset in milliseconds which will be visible at every tick even at smaller Zoom-Levels. A solution could be subtracting this tiny offset to make the ticks look cleaner at smaller Zoom-Levels. But this would also mean to loose +-0.5s of accuracy on the plot (relative to the "real" time).
WDYT?

marsipu:
Now, the X-Axis for a plot with `duration=10` of the sample-dataset looks like this:
grafik

With my workaround, this Zoom-Level would look like this:
grafik

And a higher Zoom-Level would look like this:
grafik

So we would round the time to seconds for visibility. I would implement it like this:
```python
            import datetime
            first_time = self.mne.inst.first_time
            meas_date = self.mne.inst.info['meas_date']
            seconds = int(xval + first_time)
            milliseconds = int(xval % 1 * 1e3)
            # Add rounded offset from first_time and meas_date to improve
            # readability on smaller zoom-levels.
            # Introduces inaccuracy of max. 0.5 s from real time.
            seconds += round(first_time % 1 + meas_date.timestamp() % 1)
            # Round to seconds if not milliseconds from xval
            # (higher zoom-level).
            meas_date = meas_date.replace(microsecond=0)
            # Determining datetime with accuracy of microseconds to add
            # up on first_time and meas_date correctly.
            xdatetime = meas_date + datetime.timedelta(seconds=seconds,
                                                       milliseconds=
                                                       milliseconds)
            if milliseconds != 0:
                xdtstr = xdatetime.strftime('%H:%M:%S.%f')[:-3]
            else:
                xdtstr = xdatetime.strftime('%H:%M:%S')
```

drammock:
@marsipu at the higher level zoom the times are out of order (55, 56.5, 56, 57.5, 57) in your screenshot

marsipu:
> 
> 
> @marsipu at the higher level zoom the times are out of order (55, 56.5, 56, 57.5, 57) in your screenshot

Ah thanks, I overlooked it. The problem was:
```python
seconds = int(xval + first_time)
```
which also added up milliseconds. I changed it to:
```python
seconds = int(xval) + int(first_time)
```

What do you think about introducing this small artificial offset in favor of visualization? I also added the offsetto the x-axis-label now so it would be transparent:
grafik

cbrnr:
I would only show fractions of a second when necessary (i.e. when zoomed in enough). So even if `meas_date` is something like 19:01:53.676 I wouldn't show it when viewing 10s segments of data, but only in cases where these fractional seconds are significant. Furthermore, I'd never show trailing zeros (e.g. 19:01:54.5 instead of 19:01:54.500) and I wouldn't show the offset. My initial concern was not that we don't represent time exactly (it's OK to round), but when zoomed in it is very confusing if all time stamps are identical because of rounding. In that case, I'd show more exact time stamps.

marsipu:
> 
> 
> I would only show fractions of a second when necessary (i.e. when zoomed in enough). So even if `meas_date` is something like 19:01:53.676 I wouldn't show it when viewing 10s segments of data, but only in cases where these fractional seconds are significant. Furthermore, I'd never show trailing zeros (e.g. 19:01:54.5 instead of 19:01:54.500) and I wouldn't show the offset. My initial concern was not that we don't represent time exactly (it's OK to round), but when zoomed in it is very confusing if all time stamps are identical because of rounding. In that case, I'd show more exact time stamps.

To summarize:

- [ ] Remove the display of the offset
- [ ] Avoid trailing spaces

We could also display the exact time (not rounded) only when zooming in. The only thing that could be confusing for the user would be then, that `19:01:53` would then be `19:01:53.676`.
Would you prefer the rounding solution or the exact solution?

larsoner:
Could we programmatically change the sub-second rounding level based on duration shown (zoom)? For example something like this, showing how `np.pi` seconds could be formatted, seems reasonable:
```
>>> for duration in (0.09, 0.2, 1.5, 8., 20., 101.):
...     fmt = '{t:0.%df}' % max(int(np.ceil(-np.log10(duration))) + 2, 0)
...     print(f'{duration}: ' + fmt.format(t=np.pi))
... 
0.09: 3.1416
0.2: 3.142
1.5: 3.14
8.0: 3.14
20.0: 3.1
101.0: 3
```

marsipu:
> 
> 
> It doesn't quite work as expected. First, I still see values such as 22:06:03.0 or 22:06:03.00 even though the trailing zeros are not necessary (because all displayed values show whole seconds). I'd really remove them to increase readability.
> 
> Second, when zooming in I still get identical labels, e.g. 22:06:06.00 and 22:06:06.00.
> 
> So I think there must be something wrong with the heuristic.

Thank you for the comment. I am not sure how you can still get ticks with .00 because with the current implementation it should only be possible, if the subsecond parts of `first_time` and `meas_date` are both `0` or add up to `1`. Or am I missing something? What dataset are you using?
When I test it with the sample-dataset, it looks like this:

**duration=10**
grafik
**duration=1**
grafik

I removed the part where I cropped trailing zeros again because I thought the above case would be very rare and it was also changing e.g. `.470` to `.47`. Would this be desirable too?

drammock:
seems like this might be a good opportunity to write a test.  Something like:

```py
@pytest.mark.parametrize('dur', [10, 1, 0.1, 0.01])
def my_test(dur):
    # load testing raw
    fig = raw.plot(duration=dur)
    ticklabels = fig.mne.ax_main.get_xticklabels()
    if dur == 10:
        assert ticklabels[0] == whatever
    # ...etc
```

caveat: code not tested, but hopefully it gives the idea.  Write the test so that it asserts what you want/expect the formatting to look like for each duration level.  Think about what the cutoff should be for duration above which you no longer display decimal seconds... you probably don't need to test 4 levels like I've shown here, just one above and one below (and maybe the boundary duration).   Also watch out for changes in ticks when the window size changes... not sure if that happens with our Raw browser but it might.

marsipu:
> 
> 
> It doesn't quite work as expected. First, I still see values such as 22:06:03.0 or 22:06:03.00 even though the trailing zeros are not necessary (because all displayed values show whole seconds). I'd really remove them to increase readability.
> 
> Second, when zooming in I still get identical labels, e.g. 22:06:06.00 and 22:06:06.00.
> 
> So I think there must be something wrong with the heuristic.

Thank you for the comment, I found the problem. It appeared if the subsecond values of `first_time` and `meas_date`  are both 0 or add up to 1 (which dataset did you use to test?)
I also added the part of removing trailing zeros again (I removed it before because I thought that the above case was highly unlikely and it also set values as e.g `12:34:59.470` from `.470` to `.47`. Should these zeros be also removed?)

marsipu:
@drammock Thank you for the suggestion, I will do a test.

>  Also watch out for changes in ticks when the window size changes...
Thanks for the hint, I observed that the tick-values don't update when I change the size of the window once opened.

marsipu:
This would be my test:
```python
@pytest.mark.parametrize('dur', [20, 4.5, 0.01])
def test_datetime_xticks(dur, raw):
    """Test if decimal seconds of xticks have appropriate length
    depending on duration."""
    fig = raw.plot(duration=dur, time_format='datetime')
    ticklabels = fig.mne.ax_main.get_xticklabels()
    tick_texts = ticklabels[0]._text.split('.')
    len_decsec = max(int(np.ceil(-np.log10(dur))) + 2, 0)
    if len(tick_texts) == 2:
        assert len(tick_texts[1]) == len_decsec
    plt.close('all')
```
I struggle with a problem:
Somehow the xticks always at first have the values s`[-0.06, -0.04, -0.02, 0.0, 0.02, 0.06, None, None]` regardless of the duration.
I already encountered this when debugging, those values seem to be updated to the real ones when the plot is shown.
But for the tests with 'agg' I can't get to the real values.
Now the test always passes but it defies its purpose.
@drammock Do you know where these first values come from?

mne/viz/_figure.py
```diff
@@ -1886,9 +1886,10 @@ def _xtick_formatter(self, xval, _=None):
             lim = max(int(np.ceil(-np.log10(self.mne.duration))) + 3, 0)
             if lim > 0:
                 ustr = ustr[:lim]
-                # Remove trailing zeros (if there are only zeros
-                # as decimal points)
-                if int(ustr) != 0:
+                # Remove trailing zeros at the end
+                while len(ustr) > 0 and ustr[-1] == '0':
+                    ustr = ustr[:-1]
```
cbrnr, line 1891:
```suggestion
                ustr = ustr.rstrip("0")
```

doc/changes/latest.inc
```diff
@@ -49,6 +49,8 @@ Enhancements
 
 - Add support for interpolating oxy and deoxyhaemoglobin data types (:gh:`9431` by `Robert Luke`_)
 
+- Add "time_format" to :meth:`mne.io.Raw.plot` and :meth:`mne.preprocessing.ICA.plot_sources` to display acquisition-time on x-axis (:gh:`9419` by `Martin Schulz`_)
```
cbrnr, line 52:
```suggestion
- Add "time_format" to :meth:`mne.io.Raw.plot` and :meth:`mne.preprocessing.ICA.plot_sources` to display acquisition time on x-axis (:gh:`9419` by `Martin Schulz`_)
```

mne/viz/_figure.py
```diff
@@ -1039,6 +1047,7 @@ def _get_help_text(self):
             ('p', 'Toggle draggable annotations' if is_raw else None),
             ('s', 'Toggle scalebars' if not is_ica else None),
             ('z', 'Toggle scrollbars'),
+            ('t', 'Toggle time-format'),
```
cbrnr, line 1050:
```suggestion
            ('t', 'Toggle time format'),
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
             text.set_style(sty)
 
+    def _xtick_formatter(self, xval, _=None):
+        """Change the x-axis labels."""
+        if self.mne.time_format == 'datetime':
+            import datetime
+            first_time = self.mne.inst.first_time
+            meas_date = self.mne.inst.info['meas_date']
+            seconds = int(xval + first_time)
+            ms = int((xval + first_time) % 1 * 1e3)
```
cbrnr, line 1873:
```suggestion
            ms = int((xval + first_time) % 1e3)
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+            if ms == 0:
+                us = int(((xval + first_time) % 1 * 1e6))
```
cbrnr, line 1875:
```suggestion
                us = int((xval + first_time) % 1e6)
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+            else:
+                us = int(((xval + first_time) % 1 * 1e6) % (ms * 1e3))
```
cbrnr, line 1877:
```suggestion
                us = int(((xval + first_time) % 1e6) % (ms * 1e3))
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+            # Adding time from first_time (only seconds) and xval to meas_date.
+            xdatetime = meas_date + datetime.timedelta(seconds=seconds,
+                                                       milliseconds=ms,
+                                                       microseconds=us)
+            xdtstr = xdatetime.strftime('%H:%M:%S')
+            # Crop time-string for microseconds
+            # depending on duration (zoom-level).
```
cbrnr, line 1884:
```suggestion
            # Crop time string for microseconds depending on duration
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+            import datetime
+            first_time = self.mne.inst.first_time
+            meas_date = self.mne.inst.info['meas_date']
+            seconds = int(xval + first_time)
+            ms = int((xval + first_time) % 1 * 1e3)
+            if ms == 0:
+                us = int(((xval + first_time) % 1 * 1e6))
+            else:
+                us = int(((xval + first_time) % 1 * 1e6) % (ms * 1e3))
+            # Adding time from first_time (only seconds) and xval to meas_date.
```
cbrnr, line 1878:
```suggestion
            # Add time from first_time (only seconds) and xval to meas_date
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+                                                       microseconds=us)
+            xdtstr = xdatetime.strftime('%H:%M:%S')
+            # Crop time-string for microseconds
+            # depending on duration (zoom-level).
+            ustr = xdatetime.strftime('%f')
+            lim = max(int(np.ceil(-np.log10(self.mne.duration))) + 3, 0)
+            if lim > 0:
+                ustr = ustr[:lim]
+                # Remove trailing zeros at the end
+                ustr = ustr.rstrip("0")
```
cbrnr, line 1890:
```suggestion
                ustr = ustr[:lim].rstrip("0")
```

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,56 @@ def _update_yaxis_labels(self):
...
+
+    def _toggle_time_format(self):
+        if self.mne.time_format == 'float':
+            self.mne.time_format = 'datetime'
+            x_axis_label = 'Time (HH:MM:SS)'
+        else:
+            self.mne.time_format = 'float'
+            x_axis_label = 'Time (s)'
+
+        # Change x-axis-label
```
cbrnr, line 1906:
```suggestion
        # Change x-axis label
```

cbrnr (changes requested):
Some style suggestions, I'll test the behavior next.

cbrnr:
The trailing zeros problem seems to be fixed now! However, there seem to be rounding issues that lead to some weird labels. Assuming I start with these labels at a particular zoom level:

Now zooming in once again shows `22:06:23.199` instead of `22:06:23.2`:

And zooming in one more level:

This is a huge step that is way too much - that's probably also why the data traces are not even drawn.

mne/viz/_figure.py
marsipu, line 1875:
```suggestion
                us = int((xval + first_time) % 1 * 1e6)
```

mne/viz/_figure.py
marsipu: This is not the same because associativity doesn't apply here:
```python
123.456 % 1 * 1e3 != 123.456 % 1e3
```

mne/viz/_figure.py
marsipu: Same as l.1873

marsipu:
> Now zooming in once again shows 22:06:23.199 instead of 22:06:23.2

I can't reproduce this behaviour:
image
Could it be that this comes from an inaccuracy introduced by the [binary approximation](https://docs.python.org/3/tutorial/floatingpoint.html) of the float-values as xval, first_time etc.?

mne/viz/_figure.py
cbrnr: Actually here's a mistake. `%` has precedence over `*`, so you're really calculating the remainder of the division by one, and then multiply by `1e6`. Either use parens or better yet just write `1e6`, i.e. `us = int((xval + first_time) % 1e6)`.

mne/viz/_figure.py
cbrnr: Ah I see what you're doing here. Maybe use `math.frac` to get the fractional part to be clear? Or at least add a comment saying what you're doing here.

mne/viz/_figure.py
cbrnr: Sorry, I didn't see your previous comment - see above.

marsipu:
> This is a huge step that is way too much - that's probably also why the data traces are not even drawn.

As I understand it, for zooming in it is basically `duration -= 1` until the minimum zoom-leveldefined by 3 times the spacing of timepoints is reached (_figure.py, l.711)
```python
min_dur = 3 * np.diff(self.mne.inst.times[:2])[0]
```
It probably wouldn't feel that much of a big step if zoom-steps were decreased logarithmically with the duration. But would one need that extra resolution?

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,53 @@ def _update_yaxis_labels(self):
...
             text.set_style(sty)
 
+    def _xtick_formatter(self, xval, _=None):
+        """Change the x-axis labels."""
+        if self.mne.time_format == 'datetime':
+            import datetime
+            first_time = self.mne.inst.first_time
+            meas_date = self.mne.inst.info['meas_date']
+            seconds = int(xval + first_time)
+            ms = int((xval + first_time) % 1 * 1e3)
```
marsipu, line 1873:
```suggestion
            # Get decimals with remainder of division by 1.
            ms = int((xval + first_time) % 1 * 1e3)
```

cbrnr:
> EDIT: I think I found a possible explanation:
When zooming in to the minimum and then zooming back out, min_dur is added to all subsequent durations, because of duration += 1 (for your example-data it would be about 5ms). So every zoom-level afterwards is slightly different than before which is probably irrelevant but could show up in some zoomed-in cases in subsecond-rounding. Should I fix this and where (here or new PR)?

I'd do it here if it's not too much of a change.

drammock:
agree that we should fix the extra-3-samples-after-zooming-in-then-zooming-back-out problem here.  I am -1 on making duration changes logarithmic however.

cbrnr:
> I am -1 on making duration changes logarithmic however.

Yes, same!

mne/io/base.py
```diff
@@ -1492,14 +1492,16 @@ def plot(self, events=None, duration=10.0, start=0.0, n_channels=20,
...
-             show_scrollbars=True, show_scalebars=True, verbose=None):
+             show_scrollbars=True, show_scalebars=True, time_format='float',
+             verbose=None):
         return plot_raw(self, events, duration, start, n_channels, bgcolor,
                         color, bad_color, event_color, scalings, remove_dc,
                         order, show_options, title, show, block, highpass,
-                        lowpass, filtorder, clipping, show_first_samp, proj,
-                        group_by, butterfly, decim, noise_cov=noise_cov,
+                        lowpass, filtorder, clipping, show_first_samp,
+                        proj, group_by, butterfly, decim, noise_cov=noise_cov,
```
drammock, line 1501: (nitpick) It is best to avoid cosmetic changes like this, they are distracting to PR reviewers (who must think "what changed here, is it related, is it correct?" for each change)

mne/viz/_figure.py
```diff
@@ -1775,7 +1784,7 @@ def _check_update_vscroll_clicked(self, event):
         return False
 
     # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
-    # SCALEBARS & Y-AXIS LABELS
+    # SCALEBARS & Y-AXIS LABELS & X-AXIS LABELS
```
drammock, line 1787:
```suggestion
    # SCALEBARS & AXIS LABELS
```

mne/viz/ica.py
```diff
@@ -60,6 +61,9 @@ def plot_ica_sources(ica, inst, picks=None, start=None,
     show_first_samp : bool
         If True, show time axis relative to the ``raw.first_samp``.
     %(show_scrollbars)s
+    time_format : 'float' | 'datetime
+        If 'datetime', the labels for x-axis ticks will be the time derived
+        from ``Info['meas_date']``. Default is 'float'.
```
drammock, line 66:
```suggestion
    time_format : 'float' | 'datetime'
        Style of time labels on the horizontal axis. If 'float', labels will be
        number of seconds from the start of the recording. If 'datetime',
        labels will show "clock time" (hours/minutes/seconds) inferred from
        ``raw.info['meas_date']``. Default is 'float'.
        
        .. versionadded:: 0.24
```

writing this makes me wonder if "datetime" is really the right parameter value (since we don't show dates).  Is "timestamp" better?  Or is "datetime" understood to mean "date and/or time in ISO 8601 format"?

drammock (changes requested):
some stylistic comments.  Please commit your WIP test into the repo too, so I can review the functionality and the test at the same time

mne/viz/ica.py
cbrnr: I agree that "datetime" is not optimal. What about "clocktime"? Or just "clock"?

mne/viz/_figure.py
```diff
@@ -1854,6 +1863,54 @@ def _update_yaxis_labels(self):
...
+            ustr = xdatetime.strftime('%f')
+            lim = max(int(np.ceil(-np.log10(self.mne.duration))) + 3, 0)
+            if lim > 0:
+                ustr = ustr[:lim].rstrip("0")
+                if len(ustr) > 0:
+                    xdtstr += '.' + ustr
+
+            return xdtstr
+        else:
+            return xval
```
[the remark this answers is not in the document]
drammock: two other things that would be great to see here.

1. the ticks on the scrollbar never change, so they don't really need to get decimals added when zooming in.
2. the green text added when the vertical bar shows is often overlapping the first tick.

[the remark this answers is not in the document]
cbrnr: How would you solve the overlapping green text issue?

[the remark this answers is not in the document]
drammock: Not sure. One option is to give the green text a (semi-)opaque background; another is to put the green text below the green line, though it would still need the background in case it overlaps one of the other tick labels... Could also try just pushing it further left from its current position but I'm not sure that will suffice for the datetime case

mne/io/base.py
marsipu: I understand, that one wasn't necessary. Some other times I had to to cosmetic shifts to comply with the line limit. Or is there a better way to handle these cosmetic things?

mne/viz/_figure.py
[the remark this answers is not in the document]
marsipu: Wow nice @drammock , I didn't know that you could pass float-seconds to `datetime.timedelta()`.

mne/viz/ica.py
marsipu: I like clock, because the ticks look like one and we never show a date.

mne/io/base.py
drammock: the other parameter-line-reflowing changes in this PR are acceptable.  However, it is possible to avoid changing too many lines simply by permitting yourself to leave a line of parameters *shorter* than it could be.  The logical conclusion of this is codified in, e.g.,  [black code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) where every parameter gets its own line, always, so that exactly 1 line changes when adding 1 parameter.  

We don't currently follow black code style, but note that even if your conscience (or your BDFL) insists that you reflow parameter lines, a good practice would be to separate (in different commits) the change where the new param was added from the change where the param list was reflowed.  That allows the PR reviewer to look at changes from all commits *except* ones with messages like "whitespace" or "reflow parameter list".

drammock:
one other point of GitHub etiquette, @marsipu:  don't put @drammock into commit messages (if you need to reference someone in a commit message, exclude the @ symbol).  What will happen now is that, when we squash the commits, "@drammock" will end up in the extended commit message, and whenever someone merges that commit from `main` into their fork/branch, I will get an email about it.

Hopefully when this PR is done I (or @cbrnr or @larsoner or @agramfort or whoever merges) will remember to edit the squashed commit message before confirming the merge and remove the @ symbols.  Alternatively you could do a `git rebase -i b0bbe6e` and mark the two commits with my handle in them as `reword` so you can edit the commit messages yourself, and then `git push --force` to rewrite the PR history.

agramfort:
I'll let you merge this one when happy

>

mne/viz/tests/test_raw.py
```diff
@@ -732,3 +732,17 @@ def test_scalings_int():
     """Test that auto scalings access samples using integers."""
     raw = RawArray(np.zeros((1, 500)), create_info(1, 1000., 'eeg'))
     raw.plot(scalings='auto')
+
+
+@pytest.mark.parametrize('dur', [20, 4.5, 0.01])
+def test_clock_xticks(dur, raw):
+    """Test if decimal seconds of xticks have appropriate length
+    depending on duration."""
+    fig = raw.plot(duration=dur, time_format='datetime')
```
drammock, line 741:
```suggestion
    fig = raw.plot(duration=dur, time_format='clock')
```

mne/io/base.py
marsipu: Understood, thank you.

mne/viz/_figure.py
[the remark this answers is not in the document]
marsipu: Concerning the green link:
I assume the vertical-line-text would probably interfere more often with the xticks when it is put directly under the line then if it stays at the side.
I tried pushing the text a bit up so it is aligned with the x-axis. How would you like that?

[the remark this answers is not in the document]
drammock: can you post a screenshot?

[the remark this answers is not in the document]
marsipu: grafik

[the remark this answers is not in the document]
marsipu: Just for the max. Zoom-Level this would exceed the available space. Maybe we could set a limit there ? (#97dfa0a)

drammock (approved):
a couple more minor tiny suggestions.  I'm satisfied with the green text solution, so I think now it's time to move on to the failing CI tests.

marsipu:
> I think now it's time to move on to the failing CI tests.

I think they are all related to my new test. I couldn't solve the problem with the initialization of the xticks yet. Do you think I could create some artificial test-values to test it?

drammock:
> Quick Git-Question: Is it a problem if two commits have the same commit-message?

no

drammock:
In figuring out why the tests fail, I had basically fixed them, so I just pushed a couple commits to do so.  The main problem was needing a call to `fig.canvas.draw()` before querying the ticklabels; I also reverted my earlier change relating to how many decimals to show (removed the factor-of-2 mutiplying the log-tickdiff).

cbrnr:
Looking good! I have only three more questions/comments:

1. Are the clock tick marks located exactly at the time they specify? Or is the tick really at a slightly different time and the mark is just rounded? I'd very much prefer the former (showing exactly the locations of the tick marks).
2. The last zoom step is still way too large. This is probably an unrelated issue and could be tackled in a separate PR.
3. `"float"` mode now sometimes shows unnecessary decimal places, presumably due to rounding issues, e.g. `0.0`, `0.2`, `0.4`, `0.6000000000000001`, `0.8`. Can we make sure that this doesn't happen?

marsipu:
> Are the clock tick marks located exactly at the time they specify? Or is the tick really at a slightly different time and the mark is just rounded? I'd very much prefer the former (showing exactly the locations of the tick marks).

They should be as exact as the shown decimals, additional rounding as I applied it before intermediately is gone.

> "float" mode now sometimes shows unnecessary decimal places, presumably due to rounding issues, e.g. 0.0, 0.2, 0.4, 0.6000000000000001, 0.8. Can we make sure that this doesn't happen?

I included the 'float'-case now with the decimal-solution too, so this should not happen anymore.

marsipu:
I also thought it may be handy if the values shown at the vline would have a higher number of decimals even when being zoomed-out. I set the minimum to 3 there, which may be not the optimal way. WDYT?

drammock:
>     1. Are the clock tick marks located exactly at the time they specify? 

Yes, to within rounding, as Martin said, but it yields some weird results (sometimes ticks only show h:m:s and the diff between successive ticks changes as you look across the axis). I think the "simple" way to fix this would be to change the x values of the data from seconds to timestamps, and let matplotlib decide where the ticks should go. That would probably slow down the redraw (possibly by a lot). Not sure if it's worth it.

marsipu:
> the diff between successive ticks changes as you look across the axis

I tared the dynamic decimals a bit which would remove the difference-problem in `'float'-mode` (assuming `tdelta` for duration-change stays 1). But now there would be 1 decimal in `'clock'-mode` for duration<100 (where meas_date and first_time seconds don't add up to an integer). Maybe this is not as disturbing as the differences in 'float'-mode?

drammock:
I was playing around with the possibility of using timestamps as the x-data and letting matplotlib handle tick placement.  This does not work because it breaks our clipping masks (and would probably break our annotations too, not sure because the clipping mask problem prevents the plot from loading at all).  FYI this might be fixable in MPL, by casting rectangle artist coords using `np.float64()` instead of builtin `float()`.

In testing that approach, I came across a couple simplifications / improvements that I've just pushed, including a proof-of-concept for a pseudo-logarithmic zooming (zoom in: 80%, zoom out: 125%, so it's invertable).  @cbrnr I think I'd be ok with this as a replacement for the additive zooming, but LMK what you think.

cbrnr:
@drammock I don't think there was anything wrong conceptually with the current zooming approach, only the last zoom level seems to be broken. So I wouldn't change it and instead fix the last level.

drammock:
> @drammock I don't think there was anything wrong conceptually with the current zooming approach, only the last zoom level seems to be broken. So I wouldn't change it and instead fix the last level.

The current approach is "always change by one second, unless that would be less than 3 samples, in which case show 3 samples (because anything less than 3 samples is unintelligible)".  I don't consider that "broken" (@marsipu has already made it so that going back from 3 samples to 1 second works properly --- i.e., it no longer goes from 3 samples to 1 second + 3 samples), but I can agree that it might sometimes be useful to have access to zoom levels less than 1 second but more than 3 samples.  That is what I thought you wanted, which is why I implemented the pseudo-logarithmic approach.  Did you actually try it out, and find it worse than the 1-second-at-a-time zooming?  If so, then what other approach do you propose?

cbrnr:
So in short currently it is not possible to zoom in less than one second because it immediately goes to three samples? I haven't tried your suggestion because I wanted to understand the current behavior, so I will now try it out! It would probably make sense to always zoom by a factor of two, let's see.

marsipu:
> So in short currently it is not possible to zoom in less than one second because it immediately goes to three samples?

This was right before @drammock's solution because dur_delta was always `±1` which resulted in the direct jump to the lower limit for `duration<1`. Currently dur_delta is 20% of the duration up to the lower limit which allows much more zoom-steps for `duration<1`. I like this approach, because it provides small zoom-steps while it also makes zooming-out faster which I would appreciate as a user.

cbrnr:
@drammock @marsipu I tested the new behavior and it is *much* better than the old one (I actually think the old one was broken because not being able to zoom in less than a second should have resulted in not zooming in any further instead of showing 3 samples). If we wanted to do some bikeshedding, we could try the factor 2 approach, but I'm also fine with what you already implemented.

@marsipu regarding the clock labels, I think this is ready now. Very nice work, this is a badly needed feature!

cbrnr (approved):

drammock:
OK, I've done one final pass over the code and played around some more with the feature interactively. I'm quite happy with how this turned out @marsipu!  I added a couple more code comments, and tweaked the behavior very slightly:

1. always shows millisecond-precision in the green vline text (prev. sometimes showed only 2 decimals if zoomed out very far)
2. still truncate to integers when possible (`if int(x) == x`) but do not force this to always happen on the hscroll axis (i.e., even when `int(x) != x`).  Most of the time matplotlib will choose clean integer-valued ticks anyway, and better to be accurate in case someone loads a very short Raw file where the hscroll axis only spans a couple seconds.

+1 for merge when CIs are finished.

cbrnr:
Works great! I have one final (minor) nitpick, something which has been bothering me for a long time, and maybe this can be fixed here: there is no tick mark to the very right of the visualized time period. For example, the default duration is 10s, so I get labels 0, 2, 4, 6, 8 - but not 10. I'd find it super useful if the plot also showed the last time stamp (for both float and clock formats). Is this possible?

marsipu:
> Is this possible?

The problem as I see it is, that `xlim` of `ax_main` is set to the last time-point of `times` (l. 2185 of _figure.py).
Currently for duration = 10s, this last time-point is 0.999 (for `sfreq=1000`, l.1954 of `_figure.py`). If that is not on purpose this Off-by-one-error would be one reason.
Another reason could be float-inaccuracies when `sfreq!=1000`. (with the off-by-one-fix I for example still see the problem for the sample-data with `sfreq=600.614...`). I tried just rounding the upper xlim with `np.ceil` but the problem is, that we then couldn't accept float-values for duration anymore, because we would than either have whitespace in the plot or clip a part of it.

cbrnr:
Nice! I think it's really the range - with this change the last tick is now included!

marsipu:
> 
> 
> Nice! I think it's really the range - with this change the last tick is now included!

But not for the sample-data, right?

cbrnr:
Which sample data do you mean? I tried with `n1.edf` from the URL I provided previously, and it works.

marsipu:
> 
> 
> Which sample data do you mean? I tried with `n1.edf` from the URL I provided previously, and it works.

I meant the dataset from `mne.datasets.sample.data_path()`

cbrnr:
Any particular file from this data set (can you provide the code)?

marsipu:
```python
import os
import mne
                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw(sample_data_raw_file)
raw.plot(block=True, duration=10)
```

cbrnr:
I see. That's a bit unfortunate, I'll think about it. But for some data sets at least it seems to improve things at least.

drammock:
I would try increasing the user-requested duration by one or two sampling periods, rather than using `np.ceil`

marsipu:
@drammock you mean just like that? Yeah, nice idea, this is so small that you wouldn't notice on the plot

cbrnr:
Good idea, now it also works for the sample data! 🥳

drammock:
in it goes!  Thanks @marsipu and @cbrnr!

PR #9424, opened 2021-05-24
MRG, ENH: Add label geometry methods

### LINKED ISSUE #9415: ENH: Quantify location in label
I want to get a measure of how confident I can be that a given vertex within a label actually belongs to that label versus a neighboring one. One option is, for each vertex in the label, compute the minimum distance (in m) required to travel along the cortical surface before you reach a bounding vertex. I think this can be computed pretty easily using `dijkstra` in SciPy -- we just compute the minimum distance from each vertex in the label to the set of all edge vertices in the label. So maybe:
```
class Label:

    def distances_to_edge(self, subject=None, subjects_dir=None, surface='white'):
        """Compute the distance from each vertex to the edge of the label.
        ...
```
Then it might also be useful to have also a:
```
    def compute_area(self, subject=None, subjects_dir=None, surface='white')
```
or similar that computes the area of a given label. Then the distance can be compared to the total area (probably via a an equivalent-circle-radius `r = np.sqrt(a/np.pi)`) to give some sense of how close to the middle it likely is given the label's size, if you're more interested in relative rather than absolute distances.

larsoner:
Closes #9415

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -51,6 +51,8 @@ Enhancements
 
 - Add support for ``picks`` in :func:`mne.stc_near_sensors` (:gh:`9396` by `Eric Larson`_)
 
+- Add label methods `mne.Label.compute_area` and `mne.Label.distances_to_outside` (:gh:`9424` by `Eric Larson`_)
+
 - Add projections when printing a :class:`mne.Info` in the notebook (:gh:`9403` by `Alex Gramfort`_)
 
 - Add new function :func:`mne.bem.make_scalp_surfaces` to compute scalp surfaces for coregistration (:gh:`9463` by `Alex Gramfort`_)
--- a/mne/label.py
+++ b/mne/label.py
@@ -19,13 +19,15 @@ from .parallel import parallel_func, check_n_jobs
 from .source_estimate import (SourceEstimate, VolSourceEstimate,
                               _center_of_mass, extract_label_time_course,
                               spatial_src_adjacency)
-from .source_space import add_source_space_distances, SourceSpaces
+from .source_space import (add_source_space_distances, SourceSpaces,
+                           _ensure_src)
 from .stats.cluster_level import _find_clusters, _get_components
-from .surface import read_surface, fast_cross_3d, mesh_edges, mesh_dist
+from .surface import (complete_surface_info, read_surface, fast_cross_3d,
+                      _mesh_borders, mesh_edges, mesh_dist)
 from .transforms import apply_trans
 from .utils import (get_subjects_dir, _check_subject, logger, verbose, warn,
                     check_random_state, _validate_type, fill_doc,
-                    _check_option, check_version)
+                    _check_option, check_version, _check_fname)
 
 
 def _blend_colors(color_1, color_2):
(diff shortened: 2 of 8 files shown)


doc/changes/latest.inc
```diff
@@ -45,6 +45,8 @@ Enhancements
 
 - Add projections when printing a :class:`mne.Info` in the notebook (:gh:`9403` by `Alex Gramfort`_)
 
+- Add label methods `mne.Label.compute_area` and `mne.Label.distance_to_edge` (:gh:`9424` by `Eric Larson`_)
```
drammock, line 48:
```suggestion
- Add label methods `mne.Label.compute_area` and `mne.Label.distances_to_edge` (:gh:`9424` by `Eric Larson`_)
```

doc/changes/latest.inc
```diff
@@ -45,6 +45,8 @@ Enhancements
 
 - Add projections when printing a :class:`mne.Info` in the notebook (:gh:`9403` by `Alex Gramfort`_)
 
+- Add label methods `mne.Label.compute_area` and `mne.Label.distances_to_edge` (:gh:`9424` by `Eric Larson`_)
```
agramfort, line 48:
```suggestion
- Add label methods :meth:`mne.Label.compute_area` and :meth:`mne.Label.distances_to_edge` (:gh:`9424` by `Eric Larson`_)
```
?

mne/label.py
```diff
@@ -792,9 +788,86 @@ def center_of_mass(self, subject=None, restrict_vertices=False,
                                  subject, subjects_dir, restrict_vertices)
         return vertex
 
+    @verbose
+    def distances_to_edge(self, subject=None, subjects_dir=None,
+                          surface='white', *, verbose=None):
+        """Compute the distance from each vertex to the edge of the label.
```
agramfort, line 794: edge with cortical mesh has an ambiguous meaning. Maybe use boundary or border instead of edge here?

agramfort (commented):
also if you find a place to document this it would be a bonus

thx @larsoner

tutorials/simulation/70_point_spread.py
```diff
@@ -116,6 +115,15 @@
...
     labels[i].values[cent_idx] = 1.
 
+    # Print some useful information about this vertex and label
+    if 'transversetemporal' in label.name:
+        dist = label.distances_to_edge(subjects_dir=subjects_dir)[cent_idx]
+        area = label.compute_area(subjects_dir=subjects_dir)
+        # convert to equivalent circular radius
+        r = np.sqrt(area / np.pi)
+        print(f'{label.name} COM vertex is {dist * 1e3:0.1f} mm from edge '
+              f'(label area equivalent to a circle with r={r * 1e3:0.1f} mm)')
```
larsoner, line 125: > also if you find a place to document this it would be a bonus

There is an example here already, rendered as:

https://28834-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/simulation/70_point_spread.html#find-the-center-vertices-in-source-space-of-each-label

larsoner: (At some point we should probably have a tutorial for Label specifically but we don't, so this was the best place I could find)

doc/changes/latest.inc
larsoner: No need, our default role takes care of it. Most PRs lately haven't included these. I think @drammock has moved away from including them

tutorials/simulation/70_point_spread.py
agramfort: I meant to show the distance from label on the cortex as image but I can live without

Issue #9425, opened 2021-05-25
save_movie does not seem to interpret arguments


timpx:
#### Describe the bug
save_movie function does not seem to interpret arguments. For instance, with the time_viewer argument set to True, the saved movie does not have a visible time viewer. The time_dilatation argument does not seem to have any impact either. Etc.

#### Steps to reproduce
Run the [sEEG tutorial](https://mne.tools/stable/auto_tutorials/clinical/20_seeg.html) and uncomment the following line:
```python
brain.save_movie(time_dilation=3, interpolation='linear', framerate=10,
                  time_viewer=True, filename='./mne-test-seeg.m4')
```

#### Expected results
I would expect the arguments to work. Especially time_dilatation, because right now for me it takes forever to save the movie.

#### Actual results
Resulting movie is 1 min 40 sec without time courses, unlike what is shown on the sEEG tutorial webpage.

#### Additional information
Platform:      Linux-5.4.0-72-generic-x86_64-with-glibc2.31
Python:        3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:46)  [GCC 9.3.0]
Executable:    ~/anaconda3/envs/neo_mne/bin/python
CPU:           x86_64: 24 cores
Memory:        250.2 GB

mne:           0.23.0
numpy:         1.19.2 {blas=openblas, lapack=openblas}
scipy:         1.5.3
matplotlib:    3.3.4 {backend=Qt5Agg}

sklearn:       0.23.2
numba:         0.53.1
nibabel:       3.2.1
nilearn:       0.7.1
dipy:          1.2.0
cupy:          Not found
pandas:        1.2.4
mayavi:        4.7.2
pyvista:       0.29.1 {pyvistaqt=0.3.0, OpenGL 4.5.0 NVIDIA 390.143 via Quadro P5000/PCIe/SSE2}
vtk:           9.0.1
PyQt5:         5.12.3

larsoner:
@timpx are you running this in a Jupyter notebook where the viewer is embedded in a webpage, or plain Python (or IPython) interpreter where the viewer pops up as a Qt5 application window?

timpx:
I tried both, for both the pyvista 3d backend is used, and the plot_3d pops up a window. Same issue for both though.

GuillaumeFavelier:
Do you mind updating `pyvista` and `pyvistaqt` to the latest version (`0.30.1` and `0.4.0` respectively) and try again?

[...]

PR #9426, opened 2021-05-25
FIX: brain save_movie

GuillaumeFavelier:
This PR fixes parameter forwarding in the `save_movie()` method and also the `Save movie...` action in the tool bar.

**ToDo:**

- [x] Test generated movie duration against estimation given by time range and time dilation (suggested in https://github.com/mne-tools/mne-python/pull/9426#pullrequestreview-667897094)

Closes https://github.com/mne-tools/mne-python/issues/9425

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -3138,11 +3138,6 @@ class Brain(object):
         %(brain_screenshot_time_viewer)s
         **kwargs : dict
             Specify additional options for :mod:`imageio`.
-
-        Returns
-        -------
-        dialog : object
-            The opened dialog is returned for testing purpose only.
         """
         if filename is None:
             filename = _generate_default_filename(".mp4")
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1261,7 +1261,8 @@ class Brain(object):
         self._renderer._tool_bar_add_file_button(
             name="movie",
             desc="Save movie...",
-            func=lambda: self.save_movie(
+            func=lambda filename: self.save_movie(
+                filename=filename,
                 time_dilation=(1. / self.playback_speed)),
             shortcut="ctrl+shift+s",
         )
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1261,7 +1261,8 @@ class Brain(object):
         self._renderer._tool_bar_add_file_button(
             name="movie",
             desc="Save movie...",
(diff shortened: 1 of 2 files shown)


mne/viz/_brain/_brain.py
```diff
@@ -1261,7 +1261,8 @@ def _configure_tool_bar(self):
         self._renderer._tool_bar_add_file_button(
             name="movie",
             desc="Save movie...",
-            func=self.save_movie,
+            func=lambda: self.save_movie(
+                time_dilation=(1. / self.playback_speed)),
```
larsoner, line 1265: I don't think this will do what we want. This will create a function with a *static* time dilation at the time of the toolbar button being added. So if you create a window, whatever the default playback speed is will be fixed here, and then no matter how you tweak the options in the interface it won't change.

What we probably need is a tiny `def _save_movie_clicked(...)` that calls `self.save_movie(..., time_dilation = 1. / self.playback_speed)`

larsoner: ... another way around this would be to add the option `'widget'`, which means "use whatever value the interactive widget is set to". Then this call can be `lambda: self.save_movie(time_dilation='widget', ...)`

GuillaumeFavelier: It's not partial, you are sure the lambda won't be evaluated and then the updated value of `playback_speed` used?

larsoner: You're right, I was thinking about `partial` and passing kwargs, never mind!

larsoner (commented):
One thing that did occur to me when reading the related issue is that in the `brain` test where we save a movie, we should load it using imageio-ffmpeg or something similar that can give the duration of the movie, and we can `assert_allclose` it to the expected duration given the time range and time dilation. This could be done for both the widget interface and the toolbar one. WDYT?

GuillaumeFavelier:
> This could be done for both the widget interface and the toolbar one. 

Sure, I'll add a test for it in this PR :+1:

GuillaumeFavelier:
The test of `save_movie()` passed on `linux / pip-pre`:

```
mne/viz/_brain/tests/test_brain.py::test_brain_save_movie[pyvista] PASSED [ 95%]
```

https://github.com/mne-tools/mne-python/pull/9426/checks?check_run_id=2676293949#step:13:3143

GuillaumeFavelier:
The PR is ready to merge on my end @larsoner

larsoner:
Thanks @GuillaumeFavelier

PR #9427, opened 2021-05-26
MRG, ENH: Add mne.export

larsoner:
Implements the first part of https://github.com/mne-tools/mne-python/pull/9406#issuecomment-848766291

Having implemented it, it seems like a really nice separation of functionality actually. `mne/epochs.py` and `mne/io/base.py` are cleaner, tests are in `mne/export/tests`, and `mne/export/_eeglab.py` can have un-nested imports for its implementation, which is nice.


mne/export/_export.py
```diff
@@ -0,0 +1,121 @@
...
+#
+# License: BSD (3-clause)
+
+import os.path as op
+
+from ..utils import verbose, _validate_type
+
+
+@verbose
+def export_raw(fname, raw, fmt='auto', verbose=None):
```
agramfort, line 12: should we have both BaseRaw.export and Epochs.export? it seems quite redundant here no?

so maybe?

```suggestion
def _export_raw(fname, raw, fmt='auto', verbose=None):
```

larsoner: My idea is to expose a uniform interface `mne.export.export_*` to people for all exporting functions (raw, epochs, and evoked), as it's nice to expose all exporting methods there. It is redundant with `mne.Epochs.export(...)`, but so are `mne.Epochs.plot` / `mne.viz.plot_epochs`, `mne.Covariance.save` / `mne.write_cov`, etc. -- I think of this as a similar method-to-function mapping.

larsoner: (I can live without it though if you're not convinced)

agramfort: @drammock you decide :)

drammock: I don't object to the function/method redundancy.

Question about Evokeds: the class method on `Evoked` would pass `[self]` to the `mne.export.export_evokeds` right?  So the function can handle multiple instances, but the class method only one instance?  That's how I would expect it to work anyway.

I don't see the class-method versions in this changeset?  Is that in a separate PR?

larsoner: > I don't see the class-method versions in this changeset? Is that in a separate PR?

You mean `Epochs.export(...)` and `Raw.export(...)`? Those are already in `main`. `Evoked.export` will be part of #9406

> the class method on Evoked would pass [self] to the mne.export.export_evokeds right...

Yep, that's the plan

drammock: > > I don't see the class-method versions in this changeset? Is that in a separate PR?
> 
> You mean `Epochs.export(...)` and `Raw.export(...)`? Those are already in `main`.

Never mind, I just wasn't reading the diff carefully enough.  I see now the nested `from .export import export_epochs`

agramfort:
let's go this way with public export_xxx functions then

>

PR #9434, opened 2021-05-30
[MRG] change utils.logger.warning -> utils.warn

### LINKED ISSUE #9432: Use of "warn", "warnings.warn", "logger.warning" in the codebase
I recently got a warning raised by [`plot_compare_evokeds`](https://mne.tools/stable/generated/mne.viz.plot_compare_evokeds.html) that I somehow couldn't switch off using `mne.utils.use_log_level(0)`:

> <ipython-input-8-bd89d5f2150c>:50: RuntimeWarning: Only 1 channel in "picks"; cannot combine by method "mean".
  cmap=cmap,
<ipython-input-8-bd89d5f2150c>:50: RuntimeWarning: Cannot find channel coordinates in the supplied Evokeds. Not showing channel locations.
  cmap=cmap,

That led me down a bit of a rabbit hole and I saw that within the code base, there are **several ways how warnings are raised**. Is each of those ways desirable for the particular situation? Or is that something to be unified, to consistently make use of a feature (such as setting a log level that all warning calls respect).

Examples:

- `mne.utils.warn` versus `mne.utils.logger.warning` in [brainvision.py](https://github.com/mne-tools/mne-python/blob/f42f61284c280145c4f396aa3e9fa9ed504d7114/mne/io/brainvision/brainvision.py#L22), [misc.py](https://github.com/mne-tools/mne-python/blob/f42f61284c280145c4f396aa3e9fa9ed504d7114/mne/utils/misc.py#L152),  and [kit.py](https://github.com/mne-tools/mne-python/blob/main/mne/io/kit/kit.py#L515)
- some other modules use `warnings.warn` ([fixes.py](https://github.com/mne-tools/mne-python/blob/f42f61284c280145c4f396aa3e9fa9ed504d7114/mne/fixes.py#L21), [_coreg_gui.py](https://github.com/mne-tools/mne-python/blob/f42f61284c280145c4f396aa3e9fa9ed504d7114/mne/gui/_coreg_gui.py#L898))

 --> should these all be just `mne.utils.warn` instead?

PS: I also didn't solve that issue of not being able to suppress the warning yet, so if somebody has an idea, please let me know. Here is a MWE:

```Python
import numpy as np
import mne

# create data
sampling_freq = 200
times = np.linspace(0, 1, sampling_freq, endpoint=False)
sine = np.sin(20 * np.pi * times)

info = mne.create_info(ch_names=['sine'],
                       ch_types=['misc'],
                       sfreq=sampling_freq)

data = np.array([[0.2 * sine],
                 [0.4 * sine],
                 [0.6 * sine],
                 [0.8 * sine],
                 [1.0 * sine]])
evoked_array = mne.EvokedArray(data.mean(axis=0), info, tmin=-0.5,
                               nave=data.shape[0], comment='simulated')

# This will still raise warnings ...
with mne.utils.use_log_level(0):
    mne.viz.plot_compare_evokeds([evoked_array] * 2, picks=0, combine="mean")
```

sappelhoff:
closes #9432

A simple PR to improve consistency in the code base: using `mne.utils.warn`, instead of different warning calls.

### DIFF
--- a/mne/utils/misc.py
+++ b/mne/utils/misc.py
@@ -149,6 +149,12 @@ def run_subprocess(command, return_code=False, verbose=None, *args, **kwargs):
                     break
                 else:
                     err = err.decode('utf-8')
+                    # Leave this as logger.warning rather than warn(...) to
+                    # mirror the logger.info above for stdout. This function
+                    # is basically just a version of subprocess.call, and
+                    # shouldn't emit Python warnings due to stderr outputs
+                    # (the calling function can check for stderr output and
+                    # emit a warning if it wants).
                     logger.warning(err)
                     all_err += err
             if do_break:
--- a/mne/utils/misc.py
+++ b/mne/utils/misc.py
@@ -123,7 +123,7 @@ def run_subprocess(command, return_code=False, verbose=None, *args, **kwargs):
     # non-blocking adapted from https://stackoverflow.com/questions/375427/non-blocking-read-on-a-subprocess-pipe-in-python#4896288  # noqa: E501
     out_q = Queue()
     err_q = Queue()
-    with running_subprocess(command, *args, **kwargs) as p:
+    with running_subprocess(command, *args, **kwargs) as p, p.stdout, p.stderr:
         out_t = Thread(target=_enqueue_output, args=(p.stdout, out_q))
         err_t = Thread(target=_enqueue_output, args=(p.stderr, err_q))
         out_t.daemon = True
@@ -149,12 +149,10 @@ def run_subprocess(command, return_code=False, verbose=None, *args, **kwargs):
                     break
                 else:
                     err = err.decode('utf-8')
-                    warn(err)
+                    logger.warning(err)
                     all_err += err
             if do_break:
                 break
-    p.stdout.close()
-    p.stderr.close()
     output = (all_out, all_err)
 
     if return_code:
--- a/mne/utils/misc.py
+++ b/mne/utils/misc.py
@@ -149,7 +149,7 @@ def run_subprocess(command, return_code=False, verbose=None, *args, **kwargs):
                     break
                 else:
                     err = err.decode('utf-8')
-                    logger.warning(err)
+                    warn(err)
                     all_err += err
             if do_break:
                 break
--- a/mne/fixes.py
+++ b/mne/fixes.py
@@ -22,8 +22,6 @@ import warnings
 
 import numpy as np
 
-from .utils import warn
-
 
 ###############################################################################
 # Misc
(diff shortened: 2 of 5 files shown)


sappelhoff:
CIs were failing when I pushed the changes for the `fixes.py` file due to a circular import: https://github.com/mne-tools/mne-python/pull/9434/commits/8ececbed4b6ee7e9e52475decc2196235333f38a

in particular, this function was the problem: https://github.com/mne-tools/mne-python/blob/f42f61284c280145c4f396aa3e9fa9ed504d7114/mne/fixes.py#L29-L41

and it looks like the scipy PR for this temporary fix was merged: https://github.com/scipy/scipy/pull/12676/. So perhaps it could be removed now (not sure, because the numpy issue related to that is still open: https://github.com/numpy/numpy/issues/12943)

agramfort (approved):

larsoner:
We try to keep code in fixes as close to copy-paste as possible to the upstream coffee, so I wouldn't change anything in there.

Even if an upstream fix has been merged we support several versions back so there's a ~2 year delay before we can remove things from fixes

mne/utils/misc.py
```diff
@@ -149,7 +149,7 @@ def run_subprocess(command, return_code=False, verbose=None, *args, **kwargs):
                     break
                 else:
                     err = err.decode('utf-8')
-                    logger.warning(err)
+                    warn(err)
```
sappelhoff, line 152: This seems to confuse the tests. I don't really see how, because the traceback is not very informative: https://github.com/mne-tools/mne-python/pull/9434/checks?check_run_id=2704866135

Except for three failures where coverage.py warns about something, it always looks like this:

```
 <decorator-gen-2>:24: in run_subprocess
    ???
mne/utils/misc.py:152: in run_subprocess
    warn(err)
mne/utils/_logging.py:358: in warn
    warnings.warn_explicit(
E   RuntimeWarning:
```

Do you have an idea what's going on @larsoner?

larsoner: Warnings are treated as errors in tests. Now you are emitting a warning, rather than logging with a WARNING level of verbosity, hence it raises an error. The easiest thing is probably to add an ignore line to `mne/conftest.py` for this "include is ignored because"

larsoner: ... actually in this case I think `logger.warn` is arguably the correct behavior since this is basically an advanced wrapper for `subprocess`. I think we should restore the `logger.warning` for this one to mirror the `logger.info` that is used for `stdout`-level messages.

sappelhoff: thanks, that makes sense. Also thanks for directly pushing those commits. It all passes now.

PR #9435, opened 2021-05-30
FIX : rank computation from info now uses SSS proc history if only gr…

### LINKED ISSUE #9428: Info rank computation fails if picking only 'grad' or 'mag'
try to run this:

```python
import os.path as op
import mne
from mne.datasets import somato

data_path = somato.data_path()
subject = '01'
task = 'somato'
raw_fname = op.join(data_path, 'sub-{}'.format(subject), 'meg',
                    'sub-{}_task-{}_meg.fif'.format(subject, task))

picks = mne.pick_types(raw.info, meg=True, eeg=False, eog=True, stim=False)
event_id, tmin, tmax = 1, -1., 3.
baseline = (None, 0)
epochs = mne.Epochs(raw, events, event_id, tmin, tmax, picks=picks,
                    baseline=baseline, reject=dict(grad=4000e-13, eog=350e-6),
                    preload=True)
evoked = epochs.average()

print('Rank with all MEG')
print(mne.compute_rank(evoked, 'info'))

print('Rank with only GRAD')
print(mne.compute_rank(evoked.copy().pick('grad'), 'info'))
```

you will get:

```
Rank with all MEG
{'meg': 64}
Rank with only GRAD
{'grad': 204}
```

it shows you that if you pick the grad it ignores the rank computation derived from maxfilter.

@larsoner thoughts?

thx @PABannier for spotting the bug...

agramfort:
…ad or mag are present

closes #9428

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -59,6 +59,8 @@ Bugs
 
 - Fix bug when printing a :class:`mne.io.RawArray` in the notebook (:gh:`9404` by `Alex Gramfort`_)
 
+- Fix bug when computing rank from info for SSS data with only gradiometers or magnetometers (:gh:`9435` by `Alex Gramfort`_)
+
 API changes
 ~~~~~~~~~~~
 - Nothing yet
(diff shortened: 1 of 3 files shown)


hoechenberger (approved):
LGTM

hoechenberger:
Thanks @agramfort!

PR #9438, opened 2021-06-01
MRG: concatenate_raws(), concatenate_epochs(), write_evokeds() gain on_mismatch param and raise by default

### LINKED ISSUE #9436: mne.concatenate_raws() doesn't check if dev_head_t is the same in all instances
MWE:
```python
from pathlib import Path
import mne


sample_dir = Path(mne.datasets.sample.data_path())
sample_fname = sample_dir / 'MEG' / 'sample' / 'sample_audvis_raw.fif'

raw = mne.io.read_raw_fif(sample_fname, preload=True)
raw.crop(tmax=60)

raws = [raw.copy(), raw.copy()]

# Mofidy one raw's dev_head_t
raws[1].info['dev_head_t']['trans'] += 0.1

# This works, but shouldn't
mne.concatenate_raws(raws)
```

With Epochs, it fails, as it should:
```python
epochs = []
for raw in raws:
    epochs.append(mne.make_fixed_length_epochs(raw=raw, duration=20))

mne.concatenate_epochs(epochs)
```

```
ValueError: epochs[1].info['dev_head_t'] must match. The instances probably come from different runs,
 and are therefore associated with different head positions. Manually change info['dev_head_t'] to avoid 
this message but beware that this means the MEG sensors will not be properly spatially aligned. 
See mne.preprocessing.maxwell_filter to realign the runs to a common head position.
```

cc @agramfort

hoechenberger:
`mne.concatenate_raws()` gains on_mismatch param and raises by default

Fixes #9436

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -59,6 +59,8 @@ Bugs
 
 - Fix bug when printing a :class:`mne.io.RawArray` in the notebook (:gh:`9404` by `Alex Gramfort`_)
 
+- Fix bug when computing rank from info for SSS data with only gradiometers or magnetometers (:gh:`9435` by `Alex Gramfort`_)
+
 API changes
 ~~~~~~~~~~~
 - Nothing yet
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -51,7 +51,6 @@ Enhancements
 
 - :func:`mne.concatenate_raws`, :func:`mne.concatenate_epochs`, and func:`mne.write_evokeds` gained a new parameter ``on_mismatch``, which controls behavior in case not all of the supplied instances share the same device-to-head transformation (:gh:`9438` by `Richard Höchenberger`_)
 
-
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
@@ -64,6 +63,8 @@ Bugs
 
 - Fix bug when computing rank from info for SSS data with only gradiometers or magnetometers (:gh:`9435` by `Alex Gramfort`_)
 
+- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_to_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
+
 API changes
 ~~~~~~~~~~~
-- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_to_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
+- Nothing yet
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,4 +63,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- Nothing yet
+- :func:`mne.concatenate_raws` now raises an exception if the device-to-head transformation is not identical across all files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -49,6 +49,9 @@ Enhancements
 
 - Add support for interpolating oxy and deoxyhaemoglobin data types (:gh:`9431` by `Robert Luke`_)
 
+- :func:`mne.concatenate_raws`, :func:`mne.concatenate_epochs`, and func:`mne.write_evokeds` gained a new parameter ``on_mismatch``, which controls behavior in case not all of the supplied instances share the same device-to-head transformation (:gh:`9438` by `Richard Höchenberger`_)
+
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,7 +63,7 @@ Bugs
 
 - Fix bug when computing rank from info for SSS data with only gradiometers or magnetometers (:gh:`9435` by `Alex Gramfort`_)
 
-- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_to_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
+- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
 
 API changes
 ~~~~~~~~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,4 +63,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- :func:`mne.concatenate_raws` now raises an exception if ``raw.info`` not identical across all files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
+- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_to_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,4 +63,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- :func:`mne.concatenate_raws` now raises an exception if the device-to-head transformation is not identical across all files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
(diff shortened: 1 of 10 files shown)


hoechenberger:
@larsoner I put the changelog entry into the API section, but it could just as well go to Bugs. WDYT?

mne/io/base.py
```diff
@@ -2422,6 +2428,28 @@ def concatenate_raws(raws, preload=None, events_list=None, verbose=None):
     events : ndarray of int, shape (n_events, 3)
         The events. Only returned if ``event_list`` is not None.
     """
+    _check_option(parameter='on_mismatch', value=on_mismatch,
+                  allowed_values=('raise', 'warn', 'ignore'))
+
+    first_dev_head_t = raws[0].info['dev_head_t']
+    for raw in raws[1:]:
```
agramfort, line 2435: Can you factorize the code with concatenate_epochs ?

hoechenberger: @agramfort I looked into this but we're going so much epochs-specific magic in `concatenate_epochs()` that I felt this would actually complicate things

hoechenberger: No, wait. I think I can (re)use `_compare_epochs_infos()`. Need to look into this.

larsoner: I expect there to be some pain involved (maybe a new param for `concatenate_epochs`, maybe another BUG entry, etc.), but agree with @agramfort that at least the info-checking functionality should be shared between the two. I don't in principle see why the info comparisons should differ. Let's see if you agree from looking at the code, I don't know how they are different...

hoechenberger: I've tried to address your comments in b087b5b

mne/evoked.py
```diff
@@ -1354,7 +1354,7 @@ def write_evokeds(fname, evoked):
 
 def _write_evokeds(fname, evoked, check=True):
     """Write evoked data."""
-    from .epochs import _compare_epochs_infos
+    from .io.meas_info import _ensure_infos_match
```
larsoner, line 1357: this does not need to be nested anymore and can be combined with the existing `from .io.meas_info` line at the top

mne/epochs.py
```diff
@@ -3366,7 +3334,7 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True):
     shift = int((10 + tmax) * out.info['sfreq'])
     events_offset = _update_offset(None, out.events, shift)
     for ii, epochs in enumerate(epochs_list[1:], 1):
-        _compare_epochs_infos(epochs.info, info, ii)
+        _ensure_infos_match(epochs.info, info, ii)
```
larsoner, line 3337:
```suggestion
        _ensure_infos_match(epochs.info, info, f'epochs[{ii}]')
```

mne/io/meas_info.py
```diff
@@ -2478,3 +2478,60 @@ def _write_ch_infos(fid, chs, reset_range, ch_names_mapping):
...
+        The name of the object appearing in the error message of the comparison
+        fails.
+    on_mismatch : 'raise' | 'warn' | 'ignore'
+        What to do in case of a mismatch between `info1` and `info2`.
+    """
+    _check_option(parameter='on_mismatch', value=on_mismatch,
+                  allowed_values=('raise', 'warn', 'ignore'))
+
+    if not isinstance(name, str):  # passed epochs index
+        name = f'epochs[{name:d}]'
```
larsoner, line 2500:
```suggestion
```

mne/io/meas_info.py
```diff
@@ -2478,3 +2478,60 @@ def _write_ch_infos(fid, chs, reset_range, ch_names_mapping):
...
+
+
+def _ensure_infos_match(info1, info2, name, *, on_mismatch='raise'):
+    """Check if infos match.
+
+    Parameters
+    ----------
+    info1, info2 : instance of Info
+        The infos to compare.
+    name : str or int
```
larsoner, line 2490:
```suggestion
    name : str
```

mne/io/meas_info.py
```diff
@@ -2478,3 +2478,60 @@ def _write_ch_infos(fid, chs, reset_range, ch_names_mapping):
...
+    if info1['nchan'] != info2['nchan']:
+        msg = f"{name}.info['nchan'] values don't match"
+    elif set(info1['bads']) != set(info2['bads']):
+        msg = f"{name}.info['bads'] don't match"
+    elif info1['sfreq'] != info2['sfreq']:
+        msg = f"{name}.info['sfreq'] don't match"
+    elif set(info1['ch_names']) != set(info2['ch_names']):
+        msg = f"{name}.info['ch_names'] don't match"
+    elif len(info2['projs']) != len(info1['projs']):
+        msg = f"SSP projectors in {name} don't match"
```
larsoner, line 2514: I don't think we want to do this `msg` for *all* options, just for `info['dev_head_t']`. The others should just directly raise an exception as they did before

mne/io/base.py
```diff
@@ -2409,10 +2410,15 @@ def concatenate_raws(raws, preload=None, events_list=None, verbose=None):
...
+        List of `~mne.io.Raw` instances to concatenate (in order).
     %(preload_concatenate)s
     events_list : None | list
-        The events to concatenate. Defaults to None.
+        The events to concatenate. Defaults to ``None``.
+    on_mismatch : 'raise' | 'warn' | 'ignore'
+        How to handle situations when we discover that specific aspects of the
+        `~mne.io.Raw` instances mismatch in a way that could cause problems
+        after concatenation. Currently, this check is limited to the
+        device-to-head transformation. Defaults to ``'raise'``.
```
larsoner, line 2421: What you say here about `device-to-head`-only is how it should work, but this is not how it's coded

mne/io/meas_info.py
```diff
@@ -2478,3 +2478,60 @@ def _write_ch_infos(fid, chs, reset_range, ch_names_mapping):
...
+               f"See mne.preprocessing.maxwell_filter to realign the "
+               f"runs to a common head position.")
+
+    if msg is not None:
+        if on_mismatch == 'raise':
+            raise ValueError(msg)
+        elif on_mismatch == 'warn':
+            warn(msg)
+        else:
+            logger.info(msg)
```
larsoner, line 2537: Should use `_on_missing`, `git grep on_missing` to see how it's used elsewhere

mne/io/meas_info.py
```diff
@@ -2478,3 +2478,60 @@ def _write_ch_infos(fid, chs, reset_range, ch_names_mapping):
...
+    info1, info2 : instance of Info
+        The infos to compare.
+    name : str or int
+        The name of the object appearing in the error message of the comparison
+        fails.
+    on_mismatch : 'raise' | 'warn' | 'ignore'
+        What to do in case of a mismatch between `info1` and `info2`.
+    """
+    _check_option(parameter='on_mismatch', value=on_mismatch,
+                  allowed_values=('raise', 'warn', 'ignore'))
```
larsoner, line 2497: Should use `_check_on_missing`

hoechenberger:
@larsoner I believe I've addressed all of your comments

mne/io/base.py
```diff
@@ -2409,10 +2410,13 @@ def concatenate_raws(raws, preload=None, events_list=None, verbose=None):
...
     raws : list
-        List of Raw instances to concatenate (in order).
+        List of `~mne.io.Raw` instances to concatenate (in order).
     %(preload_concatenate)s
     events_list : None | list
-        The events to concatenate. Defaults to None.
+        The events to concatenate. Defaults to ``None``.
+    on_mismatch : 'raise' | 'warn' | 'ignore'
+        How to handle situations where the device-to-head transformation
+        differs between raw instances. Defaults to ``'raise'``.
```
larsoner, line 2419: Should make use of `_on_missing_base` and docdict, see `mne/utils/docs.py`

Also needs `.. versionadded` directive

mne/epochs.py
```diff
@@ -3366,7 +3334,7 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True):
     shift = int((10 + tmax) * out.info['sfreq'])
     events_offset = _update_offset(None, out.events, shift)
     for ii, epochs in enumerate(epochs_list[1:], 1):
-        _compare_epochs_infos(epochs.info, info, ii)
+        _ensure_infos_match(epochs.info, info, f'epochs[{ii}]')
```
larsoner, line 3337: I think you should also add `on_mismatch` to `concatenate_epochs`, since it's very similar to raw. I guess it should also be done for evoked?

hoechenberger: Ok!

mne/io/base.py
hoechenberger: Argh... yes to all. Thanks for taking your time

hoechenberger:
@larsoner Ok next round :)

I've added the `on_mismatch` param to more public functions, as you proposed. I've added it as kw-only argument, which I **thought** could make sense here, but happy to revert this if you feel differently.

hoechenberger:
@larsoner Ok. I hope that's it. 😅👌

agramfort (approved):

hoechenberger:
The failing CI job actually ran all tests successfully. That is, all CI tests pass

agramfort:
Not sure if the CI failure is legit. Otherwise lgtm. @larsoner feel free to merge if happy

doc/changes/latest.inc
```diff
@@ -63,4 +66,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- Nothing yet
+- :func:`mne.concatenate_raws` now raises an exception if ``raw.info['dev_to_head_t']`` differs between files. This behavior can be controlled using the new ``on_mismatch`` parameter (:gh:`9438` by `Richard Höchenberger`_)
```
larsoner, line 69: This is an immediate change in behavior without a deprecation cycle, so it has to live in the BUG section

mne/utils/docs.py
```diff
@@ -58,6 +58,13 @@
 
     .. versionadded:: 0.22
 """ % (_on_missing_base,)
+docdict['on_info_mismatch'] = f"""
+on_mismatch : 'raise' | 'warn' | 'ignore'
+    {_on_missing_base} the device-to-head transformation differs between raw
```
larsoner, line 63: Not just raw anymore
```suggestion
    {_on_missing_base} the device-to-head transformation differs between
```

larsoner:
Other than two minor remaining gripes, LGTM +1 for merge

hoechenberger:
All done here. 😁😅

hoechenberger:
Circle was failing because of some HTTP timeouts. I restarted the Circle build.

agramfort:
Thx 🙏 @hoechenberger

PR #9440, opened 2021-06-02
MRG, MAINT: Better mem usage and errors

larsoner:
1. Fix problems with examples introduced by #9435
2. Reduce memory usage of envelope correlation examples that seem to keep killing CircleCI
3. Use more standard `_check_option` / `_validate_type` in `Brain` to give more informative messages for bad Brain args.


agramfort (approved):

agramfort:
thx @larsoner

PR #9442, opened 2021-06-02
[MRG] DOC: Add dates to versions in changelog, add entry for version 0.1

### LINKED ISSUE #9439: doc: In changelog, add release date after "version heading"
My analysis pipeline yields a slightly different result ... I updated MNE in the meantime, so now I investigate the changelog for potential culprits (functions that might have changed, or bugs that were squashed, etc.).

Having the release date next to each version in the changelog would help me with "debugging" this situation, as I know when I last ran the pipeline. I can dig out that information from pypi, or the github releases page, but it would be much more convenient to have it directly in the changelog.

My proposal is to change these lines: https://github.com/mne-tools/mne-python/blob/e6a029802bb2174eaee21f0ab544b4b64c71c1da/doc/changes/0.21.inc#L11-L13

to something like:

```
Version 0.21 (2021-04-02)
```

What do you think?

sappelhoff:
closes #9439

Next to adding the dates for released versions, I propose to also add a `0.1.inc` file that documents the first release.

I generated the author list for that version using `git shortlog d0b2b37c84572114727ae03804d29d8238c4037f..2e624ce85127f7d94991b6738294d5a8a8f2554b --summary`

which is the initial commit up to the 0.1 version bump commit

It's nice for consistency, and to have a nice overview of when the project was started.

---

PS: I updated the release protocol in the [wiki](https://github.com/mne-tools/mne-python/wiki/How-to-make-a-release):

The newly changed line now reads:

> Change "Current" to "Version 0.20 (YYYY-MM-DD)", replacing YYYY-MM-DD with the release date, and change .. _current:: to .. _changes_0_20_0:.


doc/changes/0.1.inc
```diff
@@ -0,0 +1,18 @@
+.. _changes_0_1:
+
+Version 0.1 (2011-09-23)
+------------------------
+
+- Initial release (initial commit: 2010-12-26).
```
agramfort, line 6: note the xmas period me arriving in Boston late 2010 :)

agramfort (approved):

doc/changes/0.1.inc
sappelhoff: Yes, I had a short laugh about these dates - there is even one on christmas eve :heart_eyes:

doc/changes/0.11.inc
```diff
@@ -1,7 +1,7 @@
 .. _changes_0_11:
 
-Version 0.11
-------------
+Version 0.11 (2015-12-24)
+-------------------------
```
sappelhoff, line 4: :christmas_tree:

agramfort:
maybe I felt I had to be as generous as Santa this year :)

>

drammock (approved):

drammock:
thanks @sappelhoff!

Issue #9417, opened 2021-05-21
Using multiple colors when multiple ICA components in the reconstructed latent source figure


crsegerie:
![image](https://user-images.githubusercontent.com/38733898/119115926-02996e80-ba28-11eb-9d22-92a37ea2230f.png)

In the above figure, the colors look the same for me, and I don't know which component is which one.

marsipu:
Does this come from something like:

```python
ecg_evoked = create_ecg_epochs(raw).average()
ica.plot_sources(ecg_evokeds)
```
And it probably should only apply for evokeds?
I could look into it in the next time if nobody else does it yet

hoechenberger:
> Does this come from something like:
> 
> ```python
> ecg_evoked = create_ecg_epochs(raw).average()
> ica.plot_sources(ecg_evokeds)
> ```

Yep!

> And it probably should only apply for evokeds?

I suppose so, yes!

marsipu:
Would you want different colors just for the excluded components or for every component?

hoechenberger:
Only for the excludes, they are already labeled and appear in the legend

drammock:
closed by #9444

PR #9444, opened 2021-06-03
ENH: multiple colors for ica.plot_sources(evokeds)

marsipu:
#### Reference issue
For #9417

This is supposed to make the differentiation of multiple ICA-components displayed for evokeds easier by displaying them in different colors.

#### Questions
**Are there more ICA-Categories than EOG/ECG which would be displayed simultaneously to consider?**
In this draft I accounted for different categories by assigning a different colormap for each one.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,6 +63,8 @@ Enhancements
 
 - Speed up :meth:`mne.decoding.TimeDelayingRidge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
 
+- Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` ny `Martin Schulz`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -63,7 +63,7 @@ Enhancements
 
 - Speed up :meth:`mne.decoding.TimeDelayingRidge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
 
-- Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` ny `Martin Schulz`_)
+- Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` by `Martin Schulz`_)
 
 Bugs
 ~~~~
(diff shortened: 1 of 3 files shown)


hoechenberger:
I would differentiate categories by line style and components by color

marsipu:
> Good start! I would differentiate categories by line style and components by color 💪
Thank you for the idea. How big is the possible number of different categories I should account for?

hoechenberger:
So far I've only seen two in real life (ECG and EOG), not sure if we even support anything else right now? 😅

hoechenberger:
Thanks @marsipu! Can you touch one of the ICA tutorials so we can actually see the effects in the generated report?

https://mne.tools/stable/auto_tutorials/preprocessing/40_artifact_correction_ica.html?highlight=ica#using-a-simulated-channel-to-select-ica-components might be a good place!

By "touching" I really mean, simply changing a tiny little thing, e.g. improving phrasing somewhere, to enforce re-rendering of the tutorial

marsipu:
Where do I find the generated report?

drammock:
Screenshot_2021-06-10 ENH multiple colors for ica plot_sources(evokeds) by marsipu · Pull Request #9444 · mne-tools mne-pyt

marsipu:
Thank you! I also created this [gist](https://gist.github.com/marsipu/212d764a18c205cfa69fba49b5e6a418) to test multiple component-categories.

hoechenberger:
It's working!

https://29347-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/40_artifact_correction_ica.html#using-a-simulated-channel-to-select-ica-components


I'm not a big fan of the color map though, can we try to see what it looks like e.g. with `tab10`?

agramfort (approved):

agramfort:
anyone feel free to MRG if happy. thx @marsipu

drammock:
> I'm not a big fan of the color map though, can we try to see what it looks like e.g. with `tab10`?

I tend to agree with @hoechenberger here, that we should stick with MPL default color choices (which in this case means `tab10`) unless we have a good reason not to.  I also found the gist example a little... overwhelming; it seemed to me difficult to keep track of which ones were ECG and which were EOG, etc.  I think that particular visualization would be better if *color* (rather than linestyle) were held fixed within a component type (so, ECG component 1 is blue solid, ECG 2 is blue dashed; EOG1 is orange solid, EOG2 is orange dashed; etc).  However, I acknowledge that this is a nitpick verging on bikeshedding, especially since I don't think it's all that common to view multiple component types at once (?).  So in that light maybe it is better to hold linestyle fixed (= all solid) and vary color, since the most common use case is to show 1-3 components of the same kind (EOG, ECG, etc) and nothing else.  But having talked myself out of that suggestion, I still think `tab10` is a better choice than `rainbow`.

marsipu:
@drammock thank you for the feedback. I agree that especially the example from the gist doesn't proof the point of differentiating the components more easily. I also think that showing multiple component-types together is probably more rare than multiple components from the same type. At the moment, I just use the 4 named matplotlib linestyles, so displaying more than 4 different components with the same color but different linestyles would then lead to the original problem that components could not be differentiated (these could of course expanded with custom linestyles, but that would maybe not help much in terms of differentiation). So as you suggest I would keep it the way it is currently implemented and change the colormap to `tab10` (for `excluded_components>10` this would result in equal colors again, but so many excluded components are probably unlikely, right?).

marsipu:
grafik

hoechenberger:
This looks fantastic!

drammock:
in it goes!  Thanks @marsipu

marsipu:
Thank you:). Somehow #9417 wasn't closed

hoechenberger:
Thanks @marsipu

PR #9445, opened 2021-06-04
MRG, ENH: Add annotate_break()

hoechenberger:
This PR adds `mne.preprocessing.annotate_breaks()`, a function designed to detect periods of "inactivity" during a recording (no events, no annotations occurring).

In some experimental protocols, one lets the recording continue even during break periods, yielding segments of overly noisy data as participants are free to move, swallow, blink, talk etc. during these times. This function helps to automatically mark those segments as bad.

WIP:

- [x] Changelog entry
- [x] Tutorial / example

Code is not very DRY but I tried focussing on making it easier to understand.


mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+    merged_intervals = np.array(merged_intervals)
+
+    # Now extract the actual break periods
+    break_onsets = []
+    break_durations = []
+
+    # Handle the time period up until the first annotation
+    if (raw.first_time < merged_intervals[0][0] and
+            merged_intervals[0][0] - raw.first_time >= min_duration):
+        onset = raw.first_time  # don't subtract start_after_offset here
```
hoechenberger, line 483:
```suggestion
        onset = raw.first_time  # don't add start_after_offset here
```

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
             durations.append(times[l_idx[-1]] - times[l_idx[0]])
         desc.append(art_name)
     return Annotations(onsets, durations, desc)
+
+
+@verbose
+def annotate_breaks(raw=None, events=None, *,
```
agramfort, line 340: Raw should be mandatory

hoechenberger:
I just realized the other `annotate_*` functions all have a singular naming:

```
   annotate_flat
   annotate_movement
   annotate_muscle_zscore
   annotate_nan
```

Should I name this function `annotate_break`, then? (Initially I thought the plural form is more intuitive)

cbrnr:
This is a great addition which is going to be really useful for me (and many others I'd assume)! Maybe for the tutorial you might want to mention how to annotate breaks based on *both* events and annotations (I'd do it with `annotations_from_events` or the other way round).

> Should I name this function too `annotate_break`, then? (Initially I thought the plural form is more intuitive)

I'd use the singular form to be constent.

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
             durations.append(times[l_idx[-1]] - times[l_idx[0]])
         desc.append(art_name)
     return Annotations(onsets, durations, desc)
+
+
+@verbose
+def annotate_breaks(raw, events=None, *,
```
larsoner, line 340: I don't see much value in the `*` here, I would just put it before `verbose` to make it safer for if/when we later add new kwargs

hoechenberger:
I've addressed all comments, the only thing missing is a tutorial

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,186 @@ def _annotations_from_mask(times, art_mask, art_name):
             durations.append(times[l_idx[-1]] - times[l_idx[0]])
         desc.append(art_name)
     return Annotations(onsets, durations, desc)
+
+
+@verbose
+def annotate_break(raw, events=None, min_duration=10, start_after_offset=3.5,
+                   stop_before_onset=3.5, ignore=('bad', 'edge'), *,
+                   verbose=None):
```
agramfort, line 342: I would default to 5s not 3.5s

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,186 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+        Specifies how far the "break" annotation extends towards the beginning
+        and end of the time period between the two annotations or events
+        spanning a break. This can be used to ensure e.g. that the break
+        annotation doesn't start and end immediately with a stimulation event.
+        If, for example, your data contains a break of 30 seconds between two
+        stimuli, and ``start_after_offset`` is set to ``5`` and
+        ``stop_before_onset`` is set to ``3``, the break annotation will start
+        5 seconds after the first stimulus, and end 3 seconds before the second
+        stimulus, yielding an annotated break of ``30 - 5 - 3 = 22`` seconds.
+        Both default to 3.5 seconds.
```
agramfort, line 369: 5?

agramfort (commented):
just a nitpick as I don't know where this 3.5comes from 

but the diff LGTM

thx @hoechenberger

hoechenberger:
I've changed the defaults for the "gap" between the to-be-created break annotation and the surrounding events from 3.5 to 5 sec, as proposed by @agramfort 

I've also renamed the two corresponding parameters:

`start_after_offset` has been renamed to `dist_to_previous`
`stop_before_onset` has been renamed to `dist_to_next`

I think this is more intuitive

larsoner:
I would reserve "dist" to refer to physical distance (I think this is how we use it everywhere else), not temporal. For time I would use t_after and t_before or something else with "t" in it.

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,184 @@ def _annotations_from_mask(times, art_mask, art_name):
             durations.append(times[l_idx[-1]] - times[l_idx[0]])
         desc.append(art_name)
     return Annotations(onsets, durations, desc)
+
+
+@verbose
+def annotate_break(raw, events=None, min_duration=15., dist_to_previous=5.,
+                   dist_to_next=5., ignore=('bad', 'edge'), *, verbose=None):
```
agramfort, line 341:
```suggestion
def annotate_break(raw, events=None, min_duration=15., t_after=5.,
                   t_before=5., ignore=('bad', 'edge'), *, verbose=None):
```

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,184 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+        in the raw data, and only operate based on these events.
+    min_duration : float
+        The minimum time span in seconds between the offset of one and the
+        onset of the subsequent annotation (if ``events`` is ``None``) or
+        between two consecutive events (if ``events`` is an array) to consider
+        this period a "break". Defaults to 15 seconds.
+    dist_to_previous, dist_to_next : float
+        Specifies how far the "break" annotation extends towards the beginning
+        and end of the time period between the two annotations or events
+        spanning a break. This can be used to ensure e.g. that the break
```
agramfort, line 360:
```suggestion
    t_after, t_before : float
        Specifies when to start the break after the last event (or annotation) and before
        the next one. This can be used to ensure e.g. that the break
```

agramfort (commented):
just a suggestion

sappelhoff (commented):
Just chiming in to say thanks: This will be helpful!

hoechenberger:
I renamed to parameters to:

```
t_start_after_previous
t_stop_before_next
```

Quite verbose but much more self-explanatory, WDYT?

hoechenberger:
Tutorial:
https://29132-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/20_rejecting_bad_data.html#detecting-and-annotating-breaks

API docs:
https://29132-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.annotate_break.html#mne.preprocessing.annotate_break

hoechenberger:
Updated API docs:
https://29139-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.annotate_break.html?highlight=annotate_break#mne.preprocessing.annotate_break

Tutorial:
https://29132-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/20_rejecting_bad_data.html#detecting-and-annotating-breaks

agramfort (approved):
just 2 nitpicks

@larsoner @drammock @sappelhoff feel free to MRG

thx @hoechenberger

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+                  until recording end) is at least ``min_duration`` seconds
+                  long.
+
+    ignore : iterable of str | None
+        Annotation descriptions starting with these strings will be ignored by
+        the break-finding algorithm. The string comparison is case-insensitive,
+        i.e., ``('bad',)`` and ``('BAD',)`` are equivalent. By default, all
+        annotation descriptions starting with "bad" and annotations
+        indicating "edges" (produced by data concatenation) will be
+        ignored. If ``None``, take all existing annotations into account.
```
larsoner, line 385: I would not allow None, they can just pass `()` or `[]`

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+        If ``events`` is passed, this parameter has no effect.
+    %(verbose)s
+
+    Returns
+    -------
+    break_annotations : mne.Annotations
```
larsoner, line 391:
```suggestion
    break_annotations : instance of Annotations
```

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+        The break annotations, each with the description ``'BAD_break'``. If
+        no breaks could be found given the provided function parameters, an
+        empty `~mne.Annotations` object will be returned.
+
+    Notes
+    -----
+    .. versionadded:: 0.24
+    """
+    from mne.io import BaseRaw
```
larsoner, line 400: all module imports should be relative, so this should be `from ..io.base import BaseRaw`. It also shouldn't create a circular import, so it should be moved to the top (not nested)

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+    _validate_type(item=raw, item_name='raw', types=BaseRaw, type_name='Raw')
+    _validate_type(item=events, item_name='events', types=(None, np.ndarray))
+
+    if min_duration - t_start_after_previous - t_stop_before_next <= 0:
+        raise ValueError(
+            f'The result of '
+            f'min_duration - t_start_after_previous - t_stop_before_next '
+            f'must be greater than 0, but it is: '
+            f'{min_duration - t_start_after_previous - t_stop_before_next}'
+        )
```
larsoner, line 410: Why can't I have a `min_duration` of 1 sec with `t_start_after_previous = t_start_before_next = 5`? Imagine I have events at 0, 10.5, and 30 sec, I would expect this function *will not* to annotate nothing between 0 and 10.5 (because the potential 0.5-sec annotation doesn't meet the `min_duration` threshold) but *will* add an annotation from 15.5 to 25 sec.

mne/preprocessing/artifact_detection.py
```diff
@@ -333,3 +334,189 @@ def _annotations_from_mask(times, art_mask, art_name):
...
+    # Finally, create the break annotations
+    break_annotations = Annotations(
+        onset=break_onsets,
+        duration=break_durations,
+        description=['BAD_break'],
+        orig_time=raw.info['meas_date']
+    )
+
+    logger.info(f'Detected {len(break_annotations)} break periods of >= '
+                f'{min_duration} sec duration.')
```
larsoner, line 520: I would add the total time annotated as well as total with something like
```suggestion
                f'{min_duration} sec '
                f *{sum(break_durations}:0.1f}/{raw.times[-1]:0.1f} sec total)')
```

cbrnr: Also, the SI symbol for second is "s" and not "sec".

hoechenberger: I have modified the logging, it now generates output like:

```
Ignoring annotations with descriptions starting with: bad, edge
Detected 3 break periods of >= 20 sec duration:
    43.0 – 71.0 sec [28.0 sec]
    138.0 – 221.0 sec [83.0 sec]
    288.0 – 320.7 sec [32.7 sec]
In total, 51.7% of the data (143.7 sec) have been marked as a break.
```

@cbrnr We're using `sec` for logging in many other places of MNE as well, so I decided to use it here too for the sake of consistency. If you think we should change this, I'd appreciate if we could perhaps do it consistently across the entire code base in a separate PR?

mne/preprocessing/artifact_detection.py
hoechenberger: It was in fact a conscious design choice I made, but I'm willing to discuss it.

So the parameter `min_duration` applies to the **break periods in the data**. This is essentially being used in a first pass: we're simply looking at the data (or actually, at its annotations or events, for that matter), checking if there are any periods of duration `min_duration` where no experimental events occurred. **These are break periods**.

Now, in a subsequent pass, we create the **break annotations**. Since one will typically not want them to extend all the way to the surrounding experimental events, we set `t_start_after_previous` and `t_stop_before_next` to control just how much of the break period will be "filled up" by the break annotation.

To me it felt more natural to do it this way:
- find breaks of sufficient duration
- fill up those breaks with annotations

It could be that `min_duration`, here, is a poor name choice or it needs better documentation.

I'm a little concerned now that it obviously got you confused, so I'd like to discuss the naming, docs, and functionality with again before we actually merge this …

@cbrnr @sappelhoff @agramfort WDYT?

mne/preprocessing/artifact_detection.py
hoechenberger: @larsoner I'm a bit at a loss whether I should take `raw.first_samp` into consideration when generating the logging output or not. I find it very confusing that I have to consider it when creating annotations, but `raw.plot()` doesn't set `show_first_samp=True` by default – now the logging output, based on `raw.annotations`, doesn't match what users see in the raw browser figure by default. elhp !

mne/preprocessing/artifact_detection.py
larsoner: > So the parameter min_duration applies to the break periods in the data. 

This is consistent with my interpretation

> we're simply ... Now, in a subsequent pass ... To me it felt more natural to do it this way:

It sounds like you're talking about *how you've implemented* the code, let's first talk about how it should work *conceptually*. We can implement it however we need to in order to make it work the way we think it should conceptually.

To clarify, do you still think it is correct behavior when you have events at 0, 10.5, and 30 seconds, with `min_duration=1, t_start_after_previous=5, t_start_before_next=5`? Because to me it makes sense that this should create a single BAD_BREAK annotation from 15.5 to 25 seconds rather than raise an error. I don't see any reason to restrict `min_duration` to be less than the total length of the "buffer" that we put between the annotations and actual data segments.

sappelhoff: I think this is a reasonable behavior and I found the docstring to be clear and understandable. One improvement may be to write two sentences between the one-line initial docstr and the "parameters" documentation --> these two sentences could clarify that this is a two step procedure, step 1 uses min_duration, step 2 uses t_start... and t_stop...

mne/preprocessing/artifact_detection.py
hoechenberger: I'm now subtracting `raw.first_time` from the annotation onsets I display in the log messages.

mne/preprocessing/artifact_detection.py
hoechenberger: Addressed in 70423eb

mne/preprocessing/artifact_detection.py
hoechenberger: Addressed in 2397b46

mne/preprocessing/artifact_detection.py
larsoner: ... in other words, to me it makes sense to think of the "breaks" as being any gap between events/annotation bounds that is at least `min_duration` long *and* `t_start_after_previous` after the previous *and* `t_stop_before_next` before the next event. This boolean logic is much more natural and simpler to parse as a user than needing to think about how we actually implemented a given algorithm, even if that algorithm is well described. And my example of 0/10.5/30 to me is a compelling use case for why the current implementation does not make sense...

larsoner: Ahh I think I see, I'm misunderstanding/wrong about what `min_duration` is -- it's not the minimum duration of the annotation that will be added but of the total break time. Let me look again at the docstring and see if what's here is okay and I just misread it!

drammock: I could be wrong, but I think there's still a misunderstanding between @hoechenberger and @larsoner.  I think @hoechenberger is saying "`min_duration` refers to length of a break period, i.e. the *total time* between events (before accounting for `t_start_before_prev` and `t_stop_before_next`), whereas I think @larsoner is thinking of `min_duration` as the minimum duration *of what ends up getting annotated*.

To me the naming is problematic.  Any parameter called `min_duration` should IMO definitely refer to the minimum duration *of the resulting annotations* (this is, I think, how @larsoner interpreted it).  If it turns out to make more sense to parameterize the duration between events (rather than the duration of resultant annotations) then we definitely need a new name for that parameter.

larsoner: Maybe we could call it `min_break_duration` so it's clear it's referring to the break and not the annotation itself

hoechenberger: > Maybe we could call it `min_break_duration` so it's clear it's referring to the break and not the annotation itself

Sounds good to me!

I just spent 5 min trying to wrap my head around the other possibility – letting the parameter define the minimum duration of the **resulting** annotation. And it just doesn't work out in my mind … might just be me, though.

This just to say: for now, I think I prefer the current behavior, and would opt to rename the parameter to the suggested name.

drammock: > I just spent 5 min trying to wrap my head around the other possibility – letting the parameter define the minimum duration of the **resulting** annotation. And it just doesn't work out in my mind … might just be me, though.

to me, it's more natural to think of this problem as:

1. how long of a buffer should I leave after an event before ignoring data?
2. how long before the next event do I need to start keeping data again?
3. if the remaining time in between is shorter than X, don't bother annotating it.

I think this indicates that we really need to get the param names and docstring super clear.  For the way I'm thinking about it I might say `left_buffer`, `right_buffer`, `min_duration` (where `buffer` could also be `margin` or `padding` or `gap`) and another possibility is `pre_event_buffer` and `post_event_buffer`).  Those names might also work for @hoechenberger's API (where you specify minimum *break* rather than minimum *annotation*) if everyone prefers that, assuming `min_duration` becomes `min_break_duration` or (better?) `min_inter_event_duration`.

hoechenberger: I have renamed `min_duration` to `min_break_duration` and expanded the docs a little in dd014b2

sappelhoff: > @sappelhoff OTOH seems to agree with the current approach – if I'm not mistaken?

Yes, I agree with the current approach. When I record and analyze my data, I know approximately how long the breaks in the paradigm were. For example every 20 minutes, there was a 1 minute break. And I know that apart from these breaks my paradigm is fast-paced. So I know I can set `min_break_duration = 30`, and then care about tstart... and tstop. Seems way simpler to me, but apparently that's controversial, so more opinions would be good.

drammock: To add a little context / argue for my version:  what I'm calling `left_buffer` and `right_buffer` are *always* going to be experiment-dependent, depending on how long your "trial" is / when during the trial you stamp events (i.e., do you include an end-of-trial event or not).  In contrast, the `min_annotation_duration` (as @larsoner and I conceived of it) is not experiment-dependent at all, and might have a sensible default like 1 second or 5 seconds.  People should still be able to control it, but it's possible to set a default that wouldn't break for certain experiment designs.  In @hoechenberger's proposed API, `min_break_duration` is confounded with experimental design, so there's no way to set a safe default (I don't think?)

hoechenberger: @drammock 
I suppose we're also thinking about different use cases here… I mostly agree with what @sappelhoff said: I would typically know that, for a given study I ran, breaks last, say, 1+ mins or so, and then I would just set `min_break_duration` to that value to capture those breaks, whereas …

> . In contrast, the `min_annotation_duration` (as @larsoner and I conceived of it) is not experiment-dependent at all, and might have a sensible default like 1 second or 5 seconds.

… this seems to indicate you're really thinking about capturing rather short event-free segments too? This is not what I initially had in mind when creating this function. So maybe this is also a reason why we have a bit of confusion here?

What I like about the `min_break_duration` approach is that I don't have to do any math to capture a (sufficiently long) break. It will cause headaches for short break periods, as you suggested, yes. But that's actually not what I typically consider a break in the paradigms I've worked with in the past

drammock: I'm thinking of past experiments I've done where I allowed participants to self-determine break length between blocks.  So some breaks might be 15 seconds, others might be 5 minutes.  But either one might involve big nasty EEG noise due to the participant stretching or scratching their nose.

agramfort (approved):
Thx @hoechenberger

hoechenberger:
@agramfort I don't think we were done discussing this (see https://github.com/mne-tools/mne-python/pull/9445#discussion_r646550173)

Can you revert?

hoechenberger:
(I'm off to bed otherwise I'd do it myself)

agramfort:
sorry I merged too quickly

let's iterate on a new PR

>

cbrnr:
Yes, @hoechenberger can you create a new PR and summarize the current state of the discussion? I haven't made up my mind what I'd prefer, but it definitely seems like this needs to be discussed in detail.

PR #9449, opened 2021-06-07
fix lingering hard-coded references

drammock:
I found a couple hard-coded refs today that slipped through the code sprint...  they weren't properly linked so my regex missed them.


agramfort (approved):
thx @drammock

PR #9450, opened 2021-06-08
Update URLs in contributing message

cbrnr:
The links to MNE-Python and our CoC didn't work. @rylaw pointed out that the message GitHub shows on the Issues page contains a dead URL as well, but I don't know where we can change it. Any ideas?

Here's a screenshot of the message:
image002


agramfort (approved):

drammock (approved):

larsoner:
> the message GitHub shows on the Issues page contains a dead URL as well, but I don't know where we can change it. Any ideas?

I poked around a bit and could not find it. @drammock any ideas?

drammock:
> the message GitHub shows on the Issues page contains a dead URL as well, but I don't know where we can change it.

I also can't figure out where that message comes from (and I don't see it when I view the issues page).  +1 for merge without dealing with it.

larsoner:
@cbrnr @rylaw what URL does it want / link to? We can just add a little stub file at that location. For example I wonder if it wants a `CONTRIBUTING.md` in the root rather than `CONTRIBUTING.rst`

larsoner:
I did a quick check on another repo that I've never opened an issue on and it pointed to CONTRIBUTING.md. +1 for reformatting and renaming our root CONTRIBUTING.rst to CONTRIBUTING.md

drammock:
> +1 for reformatting and renaming our root CONTRIBUTING.rst to CONTRIBUTING.md

OK with me (though I'd really rather that GitHub fixed the problem on their end...  but this is faster/easier than dealing with their support).  @cbrnr can you do the conversion as part of this PR?

NB: we maintain our contributing info in 2 places (the other one is here: https://github.com/mne-tools/mne-python/blob/main/doc/overview/development.rst) so if you make any tweaks *other than rst -> md formatting* you should mirror those content changes in both places.

cbrnr:
I did the conversion, reworded the text a little bit and synced that with the other location.

larsoner:
Thanks @cbrnr

cbrnr:
@larsoner why is it necessary to exclude `CONTRIBUTING.md` but not the previous `CONTRIBUTING.rst`?

larsoner:
Actually looks like there is `include *.rst` so I perhaps should have done `include CONTRIBUTING.md`. Not sure if it matters, though

PR #9453, opened 2021-06-08
MAINT: Better Azure reporting, fix codecov

### LINKED ISSUE #9188: MAINT: don't skip CIs when merging to main?
as I write this, the most recent merged PR was #9185 which had all-green CIs.  The commit message for the squashed commit (70709f5069ddb413c2719f4d2b41fd6bfea2349a) does not say "ci skip" or use any other of our skip codes, and yet, all the GitHub CIs are showing as "cancelled" (click the red X at the top of the repo homepage where it shows "most recent commit").  Questions:

1. why?
2. is this what is causing CodeCov badge to show 15% coverage?
3. is it possible to change the skip behavior so that skips can occur in open PRs, but never occur on commits to main?

larsoner:
Codecov reporting is [broken in that PRs show up as our current coverage](https://community.codecov.com/t/pr-commits-showing-up-in-default-main-branch-coverage/2945), which makes our homepage badge say wrong things like we have 32% coverage. This PR will fix this hopefully by switching to the Bash uploader, which hopefully is less buggy.

In the meantime, I'm also going to try to improve our Azure reporting to also have coverage. First commit will hopefully replicate the coverage-as-main problem, then I'll try to fix it with the bash uploader.

Closes #9188

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,8 @@ jobs:
             name: Check-skip
             command: |
               export COMMIT_MESSAGE=$(git log --format=oneline -n 1);
-              if [[ "$COMMIT_MESSAGE" == *"[skip circle]"* ]] || [[ "$COMMIT_MESSAGE" == *"[circle skip]"* ]]; then
-                echo "Skip detected, exiting job ${CIRCLE_JOB}."
+              if [[ -v CIRCLE_PULL_REQUEST ]] && ([[ "$COMMIT_MESSAGE" == *"[skip circle]"* ]] || [[ "$COMMIT_MESSAGE" == *"[circle skip]"* ]]); then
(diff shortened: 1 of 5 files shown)


larsoner:
FYI I regenerated the codecov token on the codecov UI, and removed it as a secret from Azure so that hopefully the bad reports at least stop. In theory this PR should make it "just work", too.

larsoner:
Looks like it's working, I'm going to push a commit to make `[skip azp]` and `[skip circle]` be checked only on PRs, not on our `main` builds. I don't see a way to do this with GitHub actions yet, but at least having Azure and Circle run regardless of message I think is a good start

drammock:
> I'm going to push a commit to make `[skip azp]` and `[skip circle]` be checked only on PRs, not on our `main` builds.

In that case this closes #9188

PR #9456, opened 2021-06-09
Remove "Question/Support/Other" issue template

hoechenberger:
These questions should be asked on the forum, as is also clearly indicated in the issue template description – and we have a link to the forum right at the bottom of the template selector. I don't think there's any value in having this option anymore: it only clutters the issue template selector.


agramfort (approved):

agramfort:
thx @hoechenberger

PR #9458, opened 2021-06-09
MRG, ENH: Speed up TimeDelayingRidge.predict()

### LINKED ISSUE #9455: ENH: Speed up TimeDelayingRidge prediction
Have been using `mne.decoding.time_delaying_ridge.TimeDelayingRidge` and noticed that its `fit` function is faster than its `predict` function, because only the former takes advantage of frequency domain computations. Fixing it might be as simple as changing the `np.convolve` to an FFT-based convolution function such as `scipy.signal.fftconvolve`. I have not done this because it is quite possible that there are other places in the `decoding` module that could benefit from the same (basically, any predictor that relies on a series of adjacent lags as features).

rkmaddox:
Addresses #9455 by switching `np.convolve` to `fftconvolve` from `scipy.signal`. Speeds up `predict` function.

Closes #9455

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -59,7 +59,7 @@ Enhancements
 
 - Add new function :func:`mne.preprocessing.annotate_break` to automatically detect and mark "break" periods without any marked experimental events in the continuous data (:gh:`9445` by `Richard Höchenberger`_)
 
-- Speed up :func:`mne.decoding.time_delaying_ridge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
+- Speed up :meth:`mne.decoding.TimeDelayingRidge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
 
 Bugs
 ~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -59,6 +59,8 @@ Enhancements
 
 - Add new function :func:`mne.preprocessing.annotate_break` to automatically detect and mark "break" periods without any marked experimental events in the continuous data (:gh:`9445` by `Richard Höchenberger`_)
 
+- Speed up :func:`mne.decoding.TimeDelayingRidge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
+
 Bugs
 ~~~~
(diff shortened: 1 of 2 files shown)


rkmaddox:
Only difference I can see in functionality is that `fftconvolve` always outputs floats, where `np.convolve` of two integer arrays returns integers. I cannot imagine a case where that would matter.

mne/decoding/time_delaying_ridge.py
```diff
@@ -6,6 +6,7 @@
 # License: BSD (3-clause)
 
 import numpy as np
+from scipy.signal import fftconvolve
```
larsoner, line 9: This needs to be nested inside `predict` (we lazy load scipy submodules because they take some time, and we want to keep `import mne` as lightweight as possible)

larsoner:
Can you add an entry to `doc/latest.inc` in the enhancements section mentioning this speedup? Then +1 for merge from me

Any quick timings you can report on real data before/after your changes?

rkmaddox:
Sorry, I don't see a `latest.inc` in the `doc` folder. Am I looking in the wrong place?

larsoner:
sorry doc/changes/latest.inc

rkmaddox:
Not really any real data on speedup to report; will be highly dependent on the nature of the data.

doc/changes/latest.inc
```diff
@@ -59,6 +59,8 @@ Enhancements
 
 - Add new function :func:`mne.preprocessing.annotate_break` to automatically detect and mark "break" periods without any marked experimental events in the continuous data (:gh:`9445` by `Richard Höchenberger`_)
 
+- Speed up :func:`mne.decoding.time_delaying_ridge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
```
larsoner, line 62:
```suggestion
- Speed up :meth:`mne.decoding.TimeDelayingRidge.predict` by switching to FFT-based convolution (:gh:`9458` by `Ross Maddox`_)
```

larsoner:
I think the changelog fix 0de2fb4 was not a fix :)

rkmaddox:
So close. My local branch is broken now because I am git-stupid so hopefully no more changes needed.

larsoner:
Thanks @rkmaddox !

PR #9460, opened 2021-06-10
DOC: add info about reading in digitization data from CTF dir

### LINKED ISSUE #9452: Add info about reading in Polhemus head scan file into raw instance
Currently, there is no documentation that says that the Polhemus head scan `.pos` file can be read into a `Raw` instance via `mne.io.read_raw_ctf` when the `.pos` file is placed in the CTF directory.

#### Proposal
Include this piece of information in the Notes of the [`mne.io.read_raw_ctf` doc](https://mne.tools/stable/generated/mne.io.read_raw_ctf.html?highlight=read_raw#mne.io.read_raw_ctf).

rylaw:
There is no documentation that says that the Polhemus digitization `.pos` file can be read into a `mne.ioRaw` instance via `mne.io.read_raw_ctf` when the .pos file is placed in the CTF directory. 

#### Reference issue
Fixes #9452. 

- Added this info in the Notes of `mne.io.read_raw_ctf`. 
- Added this as a note in the CTF tutorial in the following section.

https://github.com/mne-tools/mne-python/blob/c0d431acf0307a8bf92ab01995397194563f6920/tutorials/io/60_ctf_bst_auditory.py#L76-L89

### DIFF
--- a/mne/io/ctf/ctf.py
+++ b/mne/io/ctf/ctf.py
@@ -57,6 +57,10 @@ def read_raw_ctf(directory, system_clock='truncate', preload=False,
     Notes
     -----
     .. versionadded:: 0.11
+    .. note:: To read in the Polhemus digitization data (for example, from
+              a .pos file), include the file in the CTF directory. The
+              points will then automatically be read into the `mne.io.Raw`
+              instace via `mne.io.read_raw_ctf`.
     """
     return RawCTF(directory, system_clock, preload=preload,
                   clean_names=clean_names, verbose=verbose)
--- a/mne/io/ctf/ctf.py
+++ b/mne/io/ctf/ctf.py
@@ -60,7 +60,7 @@ def read_raw_ctf(directory, system_clock='truncate', preload=False,
     .. note:: To read in the Polhemus digitization data (for example, from
               a .pos file), include the file in the CTF directory. The
               points will then automatically be read into the `mne.io.Raw`
-              instace via `mne.io.read_raw_ctf`.
+              instance via `mne.io.read_raw_ctf`.
     """
     return RawCTF(directory, system_clock, preload=preload,
                   clean_names=clean_names, verbose=verbose)
--- a/mne/io/ctf/ctf.py
+++ b/mne/io/ctf/ctf.py
@@ -57,10 +57,11 @@ def read_raw_ctf(directory, system_clock='truncate', preload=False,
     Notes
     -----
     .. versionadded:: 0.11
-    .. note:: To read in the Polhemus digitization data (for example, from
-              a .pos file), include the file in the CTF directory. The
-              points will then automatically be read into the `mne.io.Raw`
-              instance via `mne.io.read_raw_ctf`.
+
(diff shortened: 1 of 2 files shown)


mne/io/ctf/ctf.py
```diff
@@ -57,6 +57,10 @@ def read_raw_ctf(directory, system_clock='truncate', preload=False,
     Notes
     -----
     .. versionadded:: 0.11
+    .. note:: To read in the Polhemus digitization data (for example, from
+              a .pos file), include the file in the CTF directory. The
+              points will then automatically be read into the `mne.io.Raw`
+              instace via `mne.io.read_raw_ctf`.
```
agramfort, line 63:
```suggestion
              instance via `mne.io.read_raw_ctf`.
```

but now I am also confused as

https://mne.tools/stable/generated/mne.channels.read_polhemus_fastscan.html
https://mne.tools/stable/generated/mne.channels.read_dig_polhemus_isotrak.html#mne.channels.read_dig_polhemus_isotrak

do not specify .pos files.

I think we replicated the behavior of brainstorm

tutorials/io/60_ctf_bst_auditory.py
```diff
@@ -90,6 +90,9 @@
 #   - 12 head tracking channels
 #   - 20 unused channels
 #
+# Notice also that the digitized electrode positions (stored in a .pos file)
+# were automatically loaded and added to the `mne.io.Raw` object.
```
drammock, line 94:
```suggestion
# were automatically loaded and added to the `~mne.io.Raw` object.
```

drammock: @rylaw FYI, the tilde will cause Sphinx to suppress the `mne.io.` part and will just display `Raw` in the rendered HTML (while still cross-linking to the appropriate API page).

mne/io/ctf/ctf.py
larsoner: Yes there is still some weirdness with how we handle these files. That's for another PR, though...

agramfort (approved):
@drammock or @larsoner merge if happy

mne/io/ctf/ctf.py
```diff
@@ -57,6 +57,10 @@ def read_raw_ctf(directory, system_clock='truncate', preload=False,
     Notes
     -----
     .. versionadded:: 0.11
+    .. note:: To read in the Polhemus digitization data (for example, from
+              a .pos file), include the file in the CTF directory. The
+              points will then automatically be read into the `mne.io.Raw`
+              instance via `mne.io.read_raw_ctf`.
```
larsoner, line 63: This is already the `Notes` section, putting the entire text in a `.. note::` seems... excessive :)
```suggestion

    To read in the Polhemus digitization data (for example, from
    a .pos file), include the file in the CTF directory. The
    points will then automatically be read into the `mne.io.Raw`
    instance via `mne.io.read_raw_ctf`.
```

drammock (approved):

agramfort:
thx @rylaw

PR #9401, opened 2021-05-18
MRG: Add support for NIRSport2 devices

rob-luke:
#### Reference issue
Closes #9371
Related to https://github.com/mne-tools/mne-testing-data/pull/86

#### What does this implement/fix?
1. Adds support for reading data from NIRSport2 devices. 
2. Modifications to snirf reader such that it works with nirx created snirf files.

#### Additional information
This device can export simultaneously as SNIRF and vendor format. So I made a simultaneous recording and then ensure that data from the two files match if read with MNE.
The data is identical (as demonstrated in tests). However, the metadata is different across files (e.g. subject ID). I am in contact with the vendor to discuss this. So I don't think it's a reader bug.

PR #9461, opened 2021-06-11
MRG: Honor optode_frame argument in read_raw_snirf 

rob-luke:
PR #9401 introduced the argument `optode_frame` to `read_raw_snirf`. This was being used to transform the coordinates of sensors, but was not being stored in `info['chs']`.  Also added a test.

### DIFF
--- a/mne/io/snirf/tests/test_snirf.py
+++ b/mne/io/snirf/tests/test_snirf.py
@@ -192,15 +192,15 @@ def test_snirf_nirsport2():
 def test_snirf_coordframe():
     """Test reading SNIRF files."""
     raw = read_raw_snirf(nirx_nirsport2_103, optode_frame="head").\
-          info['chs'][3]['coord_frame']
+        info['chs'][3]['coord_frame']
     assert raw == FIFF.FIFFV_COORD_HEAD
 
     raw = read_raw_snirf(nirx_nirsport2_103, optode_frame="mri").\
-          info['chs'][3]['coord_frame']
+        info['chs'][3]['coord_frame']
     assert raw == FIFF.FIFFV_COORD_HEAD
 
     raw = read_raw_snirf(nirx_nirsport2_103, optode_frame="unknown").\
-          info['chs'][3]['coord_frame']
+        info['chs'][3]['coord_frame']
     assert raw == FIFF.FIFFV_COORD_UNKNOWN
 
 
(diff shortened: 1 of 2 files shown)


rob-luke:
@larsoner @agramfort could you please review.

mne/io/snirf/tests/test_snirf.py
```diff
@@ -186,6 +186,20 @@ def test_snirf_nirsport2():
...
 
 
+@requires_testing_data
+@requires_h5py
+def test_snirf_coordframe():
+    """Test reading SNIRF files."""
+    assert read_raw_snirf(nirx_nirsport2_103, optode_frame="unknown").\
+        info['chs'][10]['coord_frame'] == 0
+    assert read_raw_snirf(nirx_nirsport2_103, optode_frame="head").\
+        info['chs'][10]['coord_frame'] == 4
```
agramfort, line 196: can you replace the 4 by FIFF.FIFFV_COORD_HEAD ?

mne/io/snirf/tests/test_snirf.py
```diff
@@ -186,6 +186,20 @@ def test_snirf_nirsport2():
...
+    assert read_raw_snirf(nirx_nirsport2_103, optode_frame="asdf").\
+        info['chs'][10]['coord_frame'] == 0
+    assert read_raw_snirf(nirx_nirsport2_103, optode_frame="mri").\
+        info['chs'][10]['coord_frame'] == 4
```
agramfort, line 200: why is it not

FIFF.FIFFV_COORD_MRI

? if there is a good reason can you put FIFFV_COORD_HEAD and explain why?

larsoner: > why is it not FIFF.FIFFV_COORD_MRI

Because we currently convert everything to head coordinates.

As of a year ago there wasn't even a way to store the coordinate frame of individual channels. Now there is (via the extended channel info blocks). In some/many places in MNE, we also still assume that non-MEG channels are in the HEAD coordinate frame. This is gradually being changed but we're not there yet.

larsoner (approved):

rob-luke:
Thanks for the review @agramfort. I have changed to testing against FIFF. Is @larsoner's description above sufficient? Or do you think we should change to returning something other than head? I am happy to implement whatever is appropriate.

agramfort (approved):
+1 for MRG provided CIs are green.

thx @rob-luke

rob-luke:
Unfortunately the CI is failing as its hitting a 60 min timeout, so I will not merge myself. @larsoner could you merge if this is acceptable.

larsoner:
Thanks @rob-luke !

PR #9463, opened 2021-06-11
ENH : add make_scalp_surfaces function to public API

agramfort:
cc @hoechenberger

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -28,12 +28,13 @@ from .io.tree import dir_tree_find
 from .io.open import fiff_open
 from .surface import (read_surface, write_surface, complete_surface_info,
                       _compute_nearest, _get_ico_surface, read_tri,
-                      _fast_cross_nd_sum, _get_solids, _complete_sphere_surf)
+                      _fast_cross_nd_sum, _get_solids, _complete_sphere_surf,
+                      decimate_surface)
 from .transforms import _ensure_trans, apply_trans, Transform
 from .utils import (verbose, logger, run_subprocess, get_subjects_dir, warn,
                     _pl, _validate_type, _TempDir, _check_freesurfer_home,
                     _check_fname, has_nibabel, _check_option, path_like,
-                    _on_missing)
+                    _on_missing, ETSContext)
 from .externals.h5io import write_hdf5, read_hdf5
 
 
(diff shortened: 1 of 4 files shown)


mne/bem.py
```diff
@@ -2061,3 +2062,99 @@ def _ensure_bem_surfaces(bem, extra_allow=(), name='bem'):
...
+
+    This is useful for coregistration and checking aligngment between anatomy
+    and channel locations.
+
+    Parameters
+    ----------
+    %(subject)s
+    %(subjects_dir)s
+    force : bool
+        Force transformation of surface into bem. Default to True.
```
hoechenberger, line 2087:
```suggestion
    force : bool
        Force creation of the surface even if it has some topological defects.
        Defaults to ``True``.
```

mne/bem.py
```diff
@@ -2061,3 +2062,99 @@ def _ensure_bem_surfaces(bem, extra_allow=(), name='bem'):
...
+        incomplete=incomplete, extra=msg)[0]
+    write_bem_surfaces(dense_fname, surf, overwrite=overwrite)
+    levels = 'medium', 'sparse'
+    tris = [] if no_decimate else [30000, 2500]
+    if os.getenv('_MNE_TESTING_SCALP', 'false') == 'true':
+        tris = [len(surf['tris'])]  # don't actually decimate
+    for ii, (n_tri, level) in enumerate(zip(tris, levels), 3):
+        logger.info('%i. Creating %s tessellation...' % (ii, level))
+        logger.info('%i.1 Decimating the dense tessellation...' % ii)
+        with ETSContext():
```
larsoner, line 2149: We shouldn't need `ETSContext` for this as it's pure VTK rather than mayavi nowadays

mne/bem.py
```diff
@@ -2061,3 +2062,99 @@ def _ensure_bem_surfaces(bem, extra_allow=(), name='bem'):
...
+    Parameters
+    ----------
+    %(subject)s
+    %(subjects_dir)s
+    force : bool
+        Force transformation of surface into bem. Default to True.
+    %(overwrite)s
+    no_decimate : bool
+        Disable medium and sparse decimations (dense only).
+        Default to False.
```
hoechenberger, line 2091:
```suggestion
    no_decimate : bool
        Disable the "medium" and "sparse" decimations. In this case, only
        a "dense" surface will be generated. Defaults to ``False``, i.e.,
        create surfaces for all three types of decimations.
```

hoechenberger (commented):
Left some comments

Needs a changelog entry

Thanks @agramfort

mne/bem.py
```diff
@@ -2061,3 +2062,99 @@ def _ensure_bem_surfaces(bem, extra_allow=(), name='bem'):
...
+    if os.getenv('_MNE_TESTING_SCALP', 'false') == 'true':
+        tris = [len(surf['tris'])]  # don't actually decimate
+    for ii, (n_tri, level) in enumerate(zip(tris, levels), 3):
+        logger.info('%i. Creating %s tessellation...' % (ii, level))
+        logger.info('%i.1 Decimating the dense tessellation...' % ii)
+        with ETSContext():
+            points, tris = decimate_surface(points=surf['rr'],
+                                            triangles=surf['tris'],
+                                            n_triangles=n_tri)
+        dec_fname = dense_fname.replace('dense', level)
```
larsoner, line 2153: It just occurred to me that this is actually dangerous, if `dense` ever appears in SUBJECT we're in trouble. It would be better to have `fname_template.format(level)` with `fname_template = op.join(bem_dir, '%s-head-{0}.fif' % subject)` or so

larsoner (commented):
Other than a few new bugs I've noticed, LGTM. I can push commits for these if it would help

hoechenberger (approved):

hoechenberger:
@agramfort Can you add a changelog entry? Otherwise LGTM

hoechenberger:
@larsoner merge if you're happy :)

PR #9464, opened 2021-06-11
MAINT: Remove unnecessary context

larsoner:
ETSContext is a relic from when we used Mayavi to process things. Now it's done with vtk directly:

https://github.com/mne-tools/mne-python/blob/c40056c11023ff2d57b4e7bd16904deda280880a/mne/surface.py#L1140-L1179


agramfort (approved):

agramfort:
thx @larsoner

PR #9465, opened 2021-06-11
ENH: Make config.get_subjects_dir() work with pathlib.Path

hoechenberger:
I tried to follow the approach we've taken elsewhere:
according to docs, only support `str`, but in fact allow `path-like`
and cast to `str` allow support for `pathlib.Path`.


mne/utils/config.py
```diff
@@ -358,9 +358,12 @@ def get_subjects_dir(subjects_dir=None, raise_error=False):
     value : str | None
         The SUBJECTS_DIR value.
     """
+    _validate_type(item=subjects_dir, types=('path-like', None),
+                   item_name='subjects_dir', type_name='str or path-like')
+
     if subjects_dir is None:
         subjects_dir = get_config('SUBJECTS_DIR', raise_error=raise_error)
-    return subjects_dir
+    return str(subjects_dir)
```
larsoner, line 366: This can end up being `'None'` whereas before it could be `None`:
```
$ _MNE_FAKE_HOME_DIR=$PWD python -c "import mne; print(mne.utils.get_subjects_dir(None, raise_error=False))"
None
```
I think you need to protect the cast with a `if subjects_dir is not None`
```suggestion
    if subjects_dir is not None:
        subjects_dir = str(subjects_dir)
    return subjects_dir
```

agramfort (approved):

mne/utils/config.py
larsoner: (please also add a regression test for this if `pytest mne/utils` passes on your system without this change)

hoechenberger: But we already check for `None` in line 364, am I missing something?

hoechenberger: Hm okay seems one test is failing, never mind. I will look into this.

larsoner: `get_config` can return None (see my example above where it does this)

hoechenberger: Got it, thanks

hoechenberger:
I will add a test, please refrain from merging for now

hoechenberger:
@larsoner I've added a test that fails on `main`

mne/utils/tests/test_config.py
```diff
@@ -87,3 +87,20 @@ def test_sys_info():
...
+    assert get_subjects_dir(subjects_dir) == subjects_dir
+
+    # Path
+    subjects_dir = Path('/foo')
+    assert get_subjects_dir(subjects_dir) == str(subjects_dir)
+
+    # `None` in config
+    subjects_dir = None
+    set_config(key='SUBJECTS_DIR', value=subjects_dir, home_dir=tmpdir,
+               set_env=True)
```
larsoner, line 105: Tests should leave global state unmodified. Using set_env=True here will modify is.environ, hence possible affecting subsequent tests (for example if someone calls SUBJECTS_DIR=bad pytest...). You should use monkeypatch.delenv with raising=False instead.

hoechenberger: Thanks @larsoner, I didn't know about `monkeypatch.*`, good to know!

agramfort (approved):

agramfort:
Thx @hoechenberger

mne/utils/tests/test_config.py
```diff
@@ -87,3 +87,18 @@ def test_sys_info():
...
+    subjects_dir = '/foo'
+    assert get_subjects_dir(subjects_dir) == subjects_dir
+
+    # Path
+    subjects_dir = Path('/foo')
+    assert get_subjects_dir(subjects_dir) == str(subjects_dir)
+
+    # `None`
+    monkeypatch.delenv('SUBJECTS_DIR', raising=False)
+    assert get_subjects_dir() is None
```
larsoner, line 104: Reverting the `tmpdir` was an error, anyone with a `.mne/mne-python.json` file with `SUBJECTS_DIR` will have this test fail, such as for me locally on my macOS machine where I define it to be the MNE-sample-data dir:
```
____________________________________________________________________ test_get_subjects_dir _____________________________________________________________________
mne/utils/tests/test_config.py:104: in test_get_subjects_dir
    assert get_subjects_dir() is None
E   AssertionError: assert '/Users/larsoner/mne_data/MNE-sample-data/subjects' is None
E    +  where '/Users/larsoner/mne_data/MNE-sample-data/subjects' = get_subjects_dir()
```
I would just `monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmpdir))` so that `get_config` reroutes to a fake home dir

hoechenberger: Aaarghhh! Will fix in a followup PR

PR #9468, opened 2021-06-12
Add color normalization parameter to plot_topomap

### LINKED ISSUE #9466: Add norm parameter to plot_topomap
I would like to plot ERD/ERS values on a topomap using `mne.viz.plot_topomap`. I can already select a suitable diverging colormap such as `RdBu_r` (red&ndash;white&ndash;blue), but my values are not symmetrical around 0 (they range from e.g. &ndash;1 to +2.5). Because I would like to map white to zero, I need to set a different [colormap normalization](https://matplotlib.org/stable/tutorials/colors/colormapnorms.html) such as the [TwoSlopeNorm](https://matplotlib.org/stable/tutorials/colors/colormapnorms.html#twoslopenorm-different-mapping-on-either-side-of-a-center). The `mne.viz.plot_topomap` function uses [imshow](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.imshow.html) internally, which already has a `norm` parameter. Therefore, we could pretty easily expose this new parameter in `plot_topomap`. I'd have to test if this really works for my use case, but in principle this seems like a promising approach. Does anyone have any objections or alternative suggestions? Otherwise I'll submit a PR.

cbrnr:
Fixes #9466. Things to do/discuss:

- I've named the parameter `cnorm` (for "color normalization"), because I think this better describes its purpose. It also avoids a name clash with an already existing local name `norm` (which has something to do with the L2 norm, not entirely sure). However, Matplotlib calls this parameter `norm`, so if we want to mimic their naming scheme we could also call our parameter `norm` instead.
- I need to add a test.
- When passing `cnorm`, simultaneously passing `vmin` and `vmax` is deprectated, because these are already defined in the normalizer. I've implemented a pretty blunt workaround, but we could instead define a default linear normalizer with `vmin` and `vmax` and then never pass these explicitly.

### DIFF
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -795,6 +795,11 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     for more details on colormap normalization.
     """
     sphere = _check_sphere(sphere)
+    if check_version("matplotlib", "3.2.0"):
+        from matplotlib.colors import TwoSlopeNorm
+    else:
+        from matplotlib.colors import DivergingNorm as TwoSlopeNorm
+    _validate_type(cnorm, (TwoSlopeNorm, None), 'cnorm')
     if cnorm is not None:
         if vmin is not None:
             warn(f"vmin={cnorm.vmin} is implicitly defined by cnorm, ignoring "
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -758,7 +758,7 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     %(topomap_border)s
     %(topomap_ch_type)s
 
-    ..versionadded:: 0.24.0
+        ..versionadded:: 0.24.0
     cnorm : matplotlib.colors.Normalize | None
         Colormap normalization, default None means linear normalization. If not
         None, ``vmin`` and ``vmax`` arguments are ignored. See Notes for more
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -788,9 +788,9 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
 
     Note that because we define ``vmin`` and ``vmax`` in the normalization,
     arguments ``vmin`` and ``vmax`` to ``plot_topomap`` will be ignored if a
-    normalization is provided.
-    See https://matplotlib.org/stable/tutorials/colors/colormapnorms.html for
-    more details on colormap normalization.
+    normalization is provided. See the
+    :doc:`matplotlib docs <matplotlib:tutorials/colors/colormapnorms>`
+    for more details on colormap normalization.
     """
     sphere = _check_sphere(sphere)
     if cnorm is not None:
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -958,7 +958,6 @@ def _plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
 
     # plot interpolated map
     if cnorm is None:
-        from matplotlib.colors import Normalize
         cnorm = Normalize(vmin=vmin, vmax=vmax)
     im = ax.imshow(Zi, cmap=cmap, origin='lower', aspect='equal',
                    extent=extent, interpolation=image_interp, norm=cnorm)
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -25,7 +25,8 @@ from ..io.pick import (pick_types, _picks_by_type, pick_info, pick_channels,
                        _pick_data_channels, _picks_to_idx, _get_channel_types,
                        _MEG_CH_TYPES_SPLIT)
 from ..utils import (_clean_names, _time_mask, verbose, logger, fill_doc,
-                     _validate_type, _check_sphere, _check_option, _is_numeric)
+                     _validate_type, _check_sphere, _check_option, _is_numeric,
+                     warn)
 from .utils import (tight_layout, _setup_vmin_vmax, _prepare_trellis,
                     _check_delayed_ssp, _draw_proj_checkbox, figure_nobar,
                     plt_show, _process_times, DraggableColorbar,
@@ -790,6 +791,13 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     more details on colormap normalization.
     """
     sphere = _check_sphere(sphere)
+    if cnorm is not None:
+        if vmin is not None:
+            warn(f"vmin={cnorm.vmin} is implicitly defined by cnorm, ignoring "
+                 f"vmin={vmin}.")
+        if vmax is not None:
+            warn(f"vmax={cnorm.vmax} is implicitly defined by cnorm, ignoring "
+                 f"vmax={vmax}.")
     return _plot_topomap(data, pos, vmin, vmax, cmap, sensors, res, axes,
                          names, show_names, mask, mask_params, outlines,
                          contours, image_interp, show,
@@ -941,12 +949,10 @@ def _plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
 
     # plot interpolated map
     if cnorm is None:
-        im = ax.imshow(Zi, cmap=cmap, vmin=vmin, vmax=vmax, origin='lower',
-                       aspect='equal', extent=extent,
-                       interpolation=image_interp, norm=cnorm)
-    else:
-        im = ax.imshow(Zi, cmap=cmap, origin='lower', aspect='equal',
-                       extent=extent, interpolation=image_interp, norm=cnorm)
+        from matplotlib.colors import Normalize
+        cnorm = Normalize(vmin=vmin, vmax=vmax)
+    im = ax.imshow(Zi, cmap=cmap, origin='lower', aspect='equal',
+                   extent=extent, interpolation=image_interp, norm=cnorm)
 
     # gh-1432 had a workaround for no contours here, but we'll remove it
     # because mpl has probably fixed it
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -690,7 +690,7 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
                  contours=6, image_interp='bilinear', show=True,
                  onselect=None, extrapolate=_EXTRAPOLATE_DEFAULT,
                  sphere=None, border=_BORDER_DEFAULT,
-                 ch_type='eeg'):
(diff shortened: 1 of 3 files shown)


mne/viz/topomap.py
```diff
@@ -756,6 +756,8 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     %(topomap_sphere)s
     %(topomap_border)s
     %(topomap_ch_type)s
+    cnorm : matplotlib.colors.Normalize | None
+        Colormap normalization, default None means linear normalization.
```
drammock, line 760: I would add a "see notes" here, and add to the notes section a link to MPL's docs or tutorial explaining what this is / how it works.  This is based on an impression / guess that most users won't know what "colormap normalization" means and thus won't be confident that the default is in fact what they want.

drammock:
> we could instead define a default linear normalizer with `vmin` and `vmax` and then never pass these explicitly

That sounds slightly nicer to me, but the results should be the same either way (right?).  So I'd say do whatever you think leads to easier to understand / cleaner code.

cbrnr:
I think there's an even simpler (but less flexible) option: in addition to `vmin` and `vmax` we could also have `v0`, which by default is right in the middle of the data. If set, it would imply a `TwoSlopeNorm` (so no other norms would be possible, but maybe that's not even necessary). WDYT?

drammock:
I prefer passing a norm rather than passing vmin/vmid/vmax. More flexible, not hard to code on our end, and allows, e.g., logarithmic norms if someone wanted that.

cbrnr:
> That sounds slightly nicer to me, but the results should be the same either way (right?). So I'd say do whatever you think leads to easier to understand / cleaner code.

Yes, it's exactly the same. I'll think about it.

Any preference for `norm` vs. `cnorm`?

drammock:
> Any preference for `norm` vs. `cnorm`?

slight preference for `cnorm` for the reasons you mention above.

cbrnr:
I've added a Notes section, and I'm now always using a normalizer internally. If both `cnorm` and `vmin`/`vmax` are passed, I'm issuing a warning &ndash; should I perhaps just log as info instead?

drammock:
> should I perhaps just log as info instead?

Warning is preferred any time we're ignoring user input, I think.

cbrnr:
Any idea what's up with CircleCI (`ModuleNotFoundError: No module named 'pyvista.core.filters'`)?

drammock:
> Any idea what's up with CircleCI (`ModuleNotFoundError: No module named 'pyvista.core.filters'`)?

no idea, I always ask @GuillaumeFavelier when pyvista fails on the CIs https://app.circleci.com/pipelines/github/mne-tools/mne-python/8624/workflows/daf68146-3366-46a7-b809-1fdf7b221e6f/jobs/29702/steps?invite=true#step-118-71

larsoner:
Circle failures will be fixed by https://github.com/pyvista/pyvista/pull/1398

cbrnr:
@GuillaumeFavelier I'm pretty sure the error in Main Notebook are not related to my changes - do you know what's causing this?

GuillaumeFavelier:
It seems to be related to a recent change in `pyvista 0.31.2`, I'll push a fix :+1:

cbrnr:
Other than the pip-pre jobs timing out this is ready for review.

mne/viz/topomap.py
```diff
@@ -757,19 +758,53 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
...
+    3, and 0 should be white. However, white corresponds to the midpoint in the
+    data by default, i.e. 1. Therefore, we use the following colormap
+    normalization ``cnorm`` and pass it as the the ``cnorm`` argument:
+
+        from matplotlib.colors import TwoSlopeNorm
+        cnorm = TwoSlopeNorm(vmin=-1, vcenter=0, vmax=3)
+
+    Note that because we already defined ``vmin`` and ``vmax`` in the
+    normalization, arguments ``vmin`` and ``vmax`` are ignored if these are
+    also passed.
```
drammock, line 791:
```suggestion
    Note that because we define ``vmin`` and ``vmax`` in the normalization,
    arguments ``vmin`` and ``vmax`` to ``plot_topomap`` will be ignored if a
    normalization is provided.
```

mne/viz/topomap.py
```diff
@@ -757,19 +758,53 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
...
+    See https://matplotlib.org/stable/tutorials/colors/colormapnorms.html for
```
drammock, line 792: there's probably a way to link to this using an intersphinx crossref, but my attempt to get it to work locally failed, so just using the URL is fine I guess.  (the advantage of a proper crossref is that our doc build would notice if it failed to resolve someday, whereas a URL-as-text is not tested)

drammock (approved):
just a nitpick, I think it reads a little clearer now but if you disagree @cbrnr feel free to disregard.

mne/viz/topomap.py
cbrnr: In that case I'd also prefer a proper crossref. Where would I need to add it? Locally at the end of the docstring?

larsoner:
```
$ python -m sphinx.ext.intersphinx https://matplotlib.org/objects.inv > matplotlib.txt
$ cat matplotlib.txt | grep colormapnorms
	tutorials/colors/colormapnorms           Colormap Normalization                  : tutorials/colors/colormapnorms.html
	sphx_glr_download_tutorials_colors_colormapnorms.py tutorials/colors/colormapnorms.html#sphx-glr-download-tutorials-colors-colormapnorms-py
	sphx_glr_tutorials_colors_colormapnorms.py Colormap Normalization                  : tutorials/colors/colormapnorms.html#sphx-glr-tutorials-colors-colormapnorms-py
```
Then if you look at the `.txt` above this line you see it's in `std:doc` which means use can use `:doc:` to get to it. Pushed a commit for this:

Screenshot from 2021-06-23 11-57-58

mne/viz/topomap.py
```diff
@@ -757,19 +758,53 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     %(topomap_border)s
     %(topomap_ch_type)s
 
+        ..versionadded:: 0.24.0
+    cnorm : matplotlib.colors.Normalize | None
+        Colormap normalization, default None means linear normalization. If not
+        None, ``vmin`` and ``vmax`` arguments are ignored. See Notes for more
+        details.
```
larsoner, line 765:
```suggestion
        details.

        .. versionadded:: 0.24
```

mne/viz/topomap.py
```diff
@@ -25,7 +25,8 @@
                        _pick_data_channels, _picks_to_idx, _get_channel_types,
                        _MEG_CH_TYPES_SPLIT)
 from ..utils import (_clean_names, _time_mask, verbose, logger, fill_doc,
-                     _validate_type, _check_sphere, _check_option, _is_numeric)
+                     _validate_type, _check_sphere, _check_option, _is_numeric,
+                     warn)
```
larsoner, line 29:
```suggestion
                     warn, check_version)
```

mne/viz/topomap.py
```diff
@@ -757,19 +758,53 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
...
+    Note that because we define ``vmin`` and ``vmax`` in the normalization,
+    arguments ``vmin`` and ``vmax`` to ``plot_topomap`` will be ignored if a
+    normalization is provided. See the
+    :doc:`matplotlib docs <matplotlib:tutorials/colors/colormapnorms>`
+    for more details on colormap normalization.
     """
     sphere = _check_sphere(sphere)
+    if cnorm is not None:
+        if vmin is not None:
+            warn(f"vmin={cnorm.vmin} is implicitly defined by cnorm, ignoring "
```
larsoner, line 798: `cnorm` needs to be type-checked with `_validate_type` before this, otherwise `cnorm.vmin` might not even exist and the user will get a strange error:
```suggestion
    if check_version("matplotlib", "3.2.0"):
        from matplotlib.colors import TwoSlopeNorm
    else:
        from matplotlib.colors import DivergingNorm as TwoSlopeNorm
    _validate_type(cnorm, (TwoSlopeNorm, None), 'cnorm')
    if cnorm is not None:
        if vmin is not None:
            warn(f"vmin={cnorm.vmin} is implicitly defined by cnorm, ignoring "
```

mne/viz/topomap.py
```diff
@@ -915,9 +950,11 @@ def _plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
     patch_ = _get_patch(outlines, extrapolate, interp, ax)
 
     # plot interpolated map
-    im = ax.imshow(Zi, cmap=cmap, vmin=vmin, vmax=vmax, origin='lower',
-                   aspect='equal', extent=extent,
-                   interpolation=image_interp)
+    if cnorm is None:
+        from matplotlib.colors import Normalize
```
larsoner, line 954: better to move this up with the other nested MPL imports to keep them together as much as possible
```suggestion
```

mne/viz/topomap.py
```diff
@@ -828,7 +863,7 @@ def _plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
                   mask_params=None, outlines='head',
                   contours=6, image_interp='bilinear', show=True,
                   onselect=None, extrapolate=_EXTRAPOLATE_DEFAULT, sphere=None,
-                  border=_BORDER_DEFAULT, ch_type='eeg'):
+                  border=_BORDER_DEFAULT, ch_type='eeg', cnorm=None):
```
larsoner, line 866:
```suggestion
                  border=_BORDER_DEFAULT, ch_type='eeg', cnorm=None):
    from matplotlib.colors import Normalize
```

larsoner (commented):
Just a few minor things from me, otherwise LGTM

cbrnr:
Any idea what's up with CircleCI?

```
E: Failed to fetch http://deb.debian.org/debian/pool/main/g/graphviz/libpathplan4_2.40.1-6_amd64.deb  404  Not Found [IP: 151.101.250.132 80]
```

cbrnr:
Seems like this file doesn't exist: http://deb.debian.org/debian/pool/main/g/graphviz/

There's a file called libpathplan4_2.40.1-6+deb10u1_amd64.deb - maybe the package cache doesn't get refreshed?

cbrnr:
Yes, you're right, it does download the file locally.

larsoner:
There is a fix in #9484 (`sudo apt-get update`) which I'm going to merge imminently.

cbrnr:
Finally all CIs are green! @larsoner feel free to merge.

larsoner:
Thanks @cbrnr !

PR #9476, opened 2021-06-15
MAINT: Make Azure happy

larsoner:
Should hopefully fix the timeouts we've had lately. I'm pretty sure it's a problem with Mayavi building against this:

https://pypi.org/project/vtk/9.0.20210612.dev0/


hoechenberger:
Thanks @larsoner!

PR #9478, opened 2021-06-15
[MRG, ENH] Change the sEEG tutorial to support new data

alexrockhill:
Address new data in https://github.com/mne-tools/mne-misc-data/pull/5.

Ok, there is a bit of an issue with the coordinate transforms that I am still working out, but I think everything otherwise is okay.

### DIFF
--- a/tutorials/clinical/20_seeg.py
+++ b/tutorials/clinical/20_seeg.py
@@ -29,6 +29,7 @@ how to visualize surface grid channels on the brain.
 
 # Authors: Eric Larson <larson.eric.d@gmail.com>
 #          Adam Li <adam2392@gmail.com>
+#          Alex Rockhill <aprockhill@mailbox.org>
 #
 # License: BSD (3-clause)
 
--- a/tutorials/clinical/20_seeg.py
+++ b/tutorials/clinical/20_seeg.py
@@ -122,7 +122,7 @@ raw.info['bads'].extend([ch for ch in raw.ch_names if ch not in ch_names])
 raw.load_data()
 raw.drop_channels(raw.info['bads'])
 events, event_id = mne.events_from_annotations(raw)
-epochs = mne.Epochs(raw, events, event_id, baseline=None)
+epochs = mne.Epochs(raw, events, event_id, detrend=1, baseline=None)
 epochs = epochs['Response'][0]  # just process one epoch of data for speed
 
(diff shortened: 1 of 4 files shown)


alexrockhill:
Ok this should be good to go once the CIs pass!

alexrockhill:
The artifact looks good
https://29553-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/clinical/20_seeg.html

drammock:
this part is not formatted right:
Screenshot_2021-06-15 Working with sEEG data — MNE 0 24 dev0 documentation

alexrockhill:
> this part is not formatted right:
> Screenshot_2021-06-15 Working with sEEG data — MNE 0 24 dev0 documentation

Good catch! I was too focused on the plots turning out alright.

alexrockhill:
Ok fixed the note and my other note that I modeled it off that was also a mistake... oops

https://29569-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/clinical/20_seeg.html?highlight=seeg

tutorials/clinical/20_seeg.py
```diff
@@ -42,6 +43,8 @@
 
 print(__doc__)
 
+np.set_printoptions(suppress=True)  # suppress scientific notation
```
agramfort, line 46: is this really necessary?

agramfort (approved):
besides LGTM

thx @alexrockhill

tutorials/clinical/20_seeg.py
larsoner: I'm okay with having this, but we need to update `reset_warnings` to `np.set_printoptions` to whatever they are when `conf.py` is initially loaded (which can be obtained before any examples are run directly in `conf.py` using `get_printoptions` presumably)

alexrockhill: I have a hard time understanding the affines in scientific notation. I'm not sure what you mean @larsoner, is that just like putting it back at the end?

Also that's how `nibabel` does it and it looks nice.

larsoner:
I'll push a quick commit

alexrockhill:
Well I get what you did I think but I'm definitely glad you did it :) Thanks

larsoner:
Go go go @alexrockhill !

PR #9481, opened 2021-06-16
ENH Faster solver BCD

PABannier:
The goal of this PR is to make BCD solver faster.


PABannier:
For the record, with Anderson acceleration used with BCD, fitting a mixed norm solver using SURE on 60ms of data from Left Auditory takes 86s on my laptop, against 185s without acceleration, all other things being equal.

mne/inverse_sparse/mxne_optim.py
```diff
@@ -192,6 +192,9 @@ def dgap_l21(M, G, X, active_set, alpha, n_orient):
         The regularization parameter.
     n_orient : int
         Number of dipoles per locations (typically 1 or 3).
+    primal_only : bool
+        Only computes and returns the primal value. Useful when only the
+        primal is needed.
 
     Returns
     -------
```
agramfort, line 200: I guess you need to update the Returns section. I am not a huge fan that a function called dgap_... does not return a dual gap...

mne/inverse_sparse/mxne_optim.py
```diff
@@ -372,6 +382,35 @@ def _mixed_norm_solver_bcd(M, G, alpha, lipschitz_constant, maxit=200,
                              % (gap, tol))
                 break
 
+        # using Anderson acceleration for faster convergence
```
agramfort, line 385: specify that it's Anderson acceleration of the primal variable.

mne/inverse_sparse/mxne_optim.py
PABannier: Same here. I didn't want to duplicate code, but I can create a function primal_l21 if it is more appropriate.

doc/changes/latest.inc
```diff
@@ -51,6 +51,8 @@ Enhancements
 
 - Add projections when printing a :class:`mne.Info` in the notebook (:gh:`9403` by `Alex Gramfort`_)
 
+- Add support for SURE parameter selection in :func:`mne.inverse_sparse.mixed_norm` (:gh:`9430` by **new contributor** `Pierre-Antoine Bannier`_ and `Alex Gramfort`_)
+
```
agramfort, line 55:
```suggestion
```

it's already above

doc/changes/latest.inc
```diff
@@ -71,6 +73,8 @@ Enhancements
 
 - Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` by `Martin Schulz`_)
 
+- Speed up :func:`mne.inverse_sparse._mixed_norm_solver_bcd` by adding Anderson acceleration to BCD (:gh:`9481` by **new contributor** `Pierre-Antoine Bannier`_ and `Alex Gramfort`_)
```
agramfort, line 76:
```suggestion
- Speed up BCD solver in :func:`mne.inverse_sparse.mixed_norm` by adding Anderson acceleration (:gh:`9481` by **new contributor** `Pierre-Antoine Bannier`_ and `Alex Gramfort`_)
```

doc/changes/latest.inc
```diff
@@ -95,4 +99,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- Nothing yet
+- Parameter ``alpha`` now defaults to ``'sure'`` in :func:`mne.inverse_sparse.mixed_norm` (:gh:`9430` by **new contributor** `Pierre-Antoine Bannier`_ and `Alex Gramfort`_)
```
agramfort, line 102:
```suggestion
```

doc/references.bib
```diff
@@ -2099,3 +2099,11 @@ @article{DeledalleEtAl2014
...
   URL = {https://doi.org/10.1137/140968045}
 }
+
+@article{BertrandEtAl2020,
+  title = {{Anderson acceleration of coordinate descent}},
+  author = {Bertrand, Quentin and Massias, Mathurin},
+  url = {https://hal.archives-ouvertes.fr/hal-03019684},
+  year = {2021},
+  booktitle = {AISAS}
+}
```
agramfort, line 2109:
```suggestion
@InProceedings{BertrandEtAl2020,
  title = { Anderson acceleration of coordinate descent },
  author = {Bertrand, Quentin and Massias, Mathurin},
  booktitle = {Proceedings of The 24th International Conference on Artificial Intelligence and Statistics},
  pages = {1288--1296},
  year = {2021},
  editor = {Banerjee, Arindam and Fukumizu, Kenji},
  volume = {130},
  series = {Proceedings of Machine Learning Research},
  month = {13--15 Apr},
  publisher = {PMLR},
  pdf = {http://proceedings.mlr.press/v130/bertrand21a/bertrand21a.pdf},
  url = {http://proceedings.mlr.press/v130/bertrand21a.html}
}
```

mne/inverse_sparse/mxne_optim.py
```diff
@@ -173,6 +173,50 @@ def prox_l1(Y, alpha, n_orient):
     return Y, active_set
 
 
+def primal_l21(M, G, X, active_set, alpha, n_orient, return_primal_only=True):
+    """Primal objective for the mixed norme inverse problem.
```
agramfort, line 177:
```suggestion
    """Primal objective for the mixed-norm inverse problem.
```

mne/inverse_sparse/mxne_optim.py
```diff
@@ -173,6 +173,50 @@ def prox_l1(Y, alpha, n_orient):
...
+
+    Parameters
+    ----------
+    M : array, shape (n_sensors, n_times)
+        The data.
+    G : array, shape (n_sensors, n_active)
+        The gain matrix a.k.a. lead field.
+    X : array, shape (n_active, n_times)
+        Sources.
+    active_set : array of bool, shape (n_sources, )
```
agramfort, line 189:
```suggestion
    active_set : array of bool, shape (n_sources,)
```

mne/inverse_sparse/mxne_optim.py
```diff
@@ -173,6 +173,50 @@ def prox_l1(Y, alpha, n_orient):
     return Y, active_set
 
 
+def primal_l21(M, G, X, active_set, alpha, n_orient, return_primal_only=True):
```
agramfort, line 176:
```suggestion
def _primal_l21(M, G, X, active_set, alpha, n_orient, return_primal_only=True):
```

I would make this aux function private

mne/inverse_sparse/mxne_optim.py
```diff
@@ -173,6 +173,50 @@ def prox_l1(Y, alpha, n_orient):
...
+    R = M - GX
+    penalty = norm_l21(X, n_orient, copy=True)
+    nR2 = sum_squared(R)
+    p_obj = 0.5 * nR2 + alpha * penalty
+    if return_primal_only:
+        return p_obj
+    return p_obj, R, nR2, GX
+
+
+
```
agramfort, line 219:
```suggestion


```

mne/inverse_sparse/mxne_optim.py
```diff
@@ -208,12 +252,7 @@ def dgap_l21(M, G, X, active_set, alpha, n_orient):
     ----------
     .. footbibilography::
     """
-    GX = np.dot(G[:, active_set], X)
-    R = M - GX
-    penalty = norm_l21(X, n_orient, copy=True)
-    nR2 = sum_squared(R)
-    p_obj = 0.5 * nR2 + alpha * penalty
-
+    p_obj, R, nR2, GX = primal_l21(M, G, X, active_set, alpha, n_orient, False)
```
agramfort, line 255:
```suggestion
    p_obj, R, nR2, GX = _primal_l21(M, G, X, active_set, alpha, n_orient, False)
```

agramfort (approved):

doc/changes/latest.inc
```diff
@@ -71,6 +71,8 @@ Enhancements
 
 - Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` by `Martin Schulz`_)
 
+- Speed up BCD solver in :func:`mne.inverse_sparse.mixed_norm` by adding Anderson acceleration (:gh:`9481` by **new contributor** `Pierre-Antoine Bannier`_ and `Alex Gramfort`_)
```
larsoner, line 74: new contributor enhancements should be grouped toward the top of the section, and need the `|` substitution syntax (see other examples above)

mne/inverse_sparse/mxne_optim.py
```diff
@@ -173,6 +173,49 @@ def prox_l1(Y, alpha, n_orient):
...
+        Forward prediction.
+    """
+    GX = np.dot(G[:, active_set], X)
+    R = M - GX
+    penalty = norm_l21(X, n_orient, copy=True)
+    nR2 = sum_squared(R)
+    p_obj = 0.5 * nR2 + alpha * penalty
+    if return_primal_only:
+        return p_obj
+    return p_obj, R, nR2, GX
```
larsoner, line 216: I don't understand why you need `return_primal_only` if it just takes the first element in the returned tuple (i.e., doesn't save any computation or anything

mne/inverse_sparse/mxne_optim.py
```diff
@@ -372,13 +414,46 @@ def _mixed_norm_solver_bcd(M, G, alpha, lipschitz_constant, maxit=200,
...
+        if use_accel:
+            last_K_X[i % (K + 1)] = X
+
+            if i % (K + 1) == K:
+                for k in range(K):
+                    U[k] = last_K_X[k + 1].ravel() - last_K_X[k].ravel()
+                C = U @ U.T
+                one_vec = np.ones(K)
+
+                try:
```
larsoner, line 428: the `try` part should include the minimal set of lines to catch a given exception. In this case it looks like only the `np.linalg.solve` line should be inside the try/except. The other lines can go in an `else` clause (which occurs if no exception is hit)

Moreover, when there is an error, it silently just continues as if there wasn't a problem. This is abnormal -- usually we emit warnings in these cases -- so if it's really okay to silently ignore this in this case, some code comment saying why would be good, e.g.:
```
# This code is just an attempt at a speedup, so if it fails we can
# safely continue using the non-sped-up code
```

mne/inverse_sparse/mxne_optim.py
PABannier: Indeed,` _primal_l21` is called at two different locations. In `dgap_l21` where the tuple is needed, and in the acceleration piece where I only need `p_obj`. I added `return_primal_only` to avoid heavy notation like `p_obj, _, _, _ = _primal_l21(...)` .

mne/inverse_sparse/mxne_optim.py
PABannier: I'll change that. Indeed, it is ok to have a warning in `np.linalg.solve` sometimes. Sometimes, the matrix U is singular by construction and can't be inverted, hence the `try` part.

mne/inverse_sparse/mxne_optim.py
larsoner: You don't need to make it so heavy, you can just do `_primal_l21(...)[0]` and it's much more compact than return_*_only=True`

PABannier: Ok got it! Thx!

agramfort:
Thx @PABannier 🔥

PR #9480, opened 2021-06-15
[DOC] Hyphenation consistency

alexrockhill:
Most tutorials are like so `_tut-working-with-ecog` but a few used underscores as in `_tut_working_with_ecog`. This just makes them all consistent the way the majority are.

PR #9484, opened 2021-06-16
MRG, ENH: Create a tutorial for how to find intracranial electrode contacts from a CT and an MR image

alexrockhill:
Addresses part of #9155.

This is a tutorial for how to:

1) Align a CT to an MR using Dipy
2) Find the electrode contacts using a GUI (another PR for the GUI but it's the `img_pipe` one)
3) Warp the electrode contacts to `fsaverage` using Dipy

This prepares you to start the working with SEEG and working with ECoG tutorials which require you to already have electrode positions in MNI space before starting but don't explain how you find them.


larsoner:
@alexrockhill feel free to ping me when I should look

alexrockhill:
Just a WIP so far but probably tomorrow

alexrockhill:
Ok @larsoner, if you have time to review once the artifact renders, this is a good first draft.

It does look like the SDR does a better job with the morph than the Talairach:

SDR
Screen Shot 2021-06-17 at 10 33 21 AM

Talairach
Screen Shot 2021-06-17 at 10 37 29 AM

alexrockhill:
The references won't work until #9480 merges though....

alexrockhill:
Uh oh I think the SDR takes longer than 10 minutes... not sure how to fix that

alexrockhill:
I'm not sure what the latest error was but maybe rebasing will take care of it. I still think it's probably related to execution time. I'll look in the SDR tests for source space to see if there's a way to do things faster...

alexrockhill:
Ok so still failing with the weird error. I'm going to try profiling the code but the SDR function and maybe the affine registrations too might need to be added to misc data so that they can just be loaded instead of computed.

alexrockhill:
> I'm not sure what the latest error was but maybe rebasing will take care of it. I still think it's probably related to execution time. I'll look in the SDR tests for source space to see if there's a way to do things faster...
> 
> ```
> generating gallery for auto_tutorials/clinical... [ 25%] 10_ieeg_localize.py
>   19.0 s : 456.8 MB   
> ERROR:mprof:Program resulted with a non-zero exit code: -9
> make: *** [Makefile:65: html_dev-pattern-memory] Error 247
> ```

I have no idea what this error means, if anyone had a chance to give it a look that would be appreciated.

alexrockhill:
[program.prof.zip](https://github.com/mne-tools/mne-python/files/6678550/program.prof.zip)

Ok, I know it's not rendering but it does run locally and it would be great to get a review of the content @larsoner if you have a minute. Here is the `cProfile` output as well that can be visualized, it looks like the affine registration just takes a long time but it does run in just over ten minutes.

alexrockhill:
Hmmm so there's an issue that the `affine_registration` from Dipy doesn't work to align the CT and MR... I have used the nipy histogram registration but it isn't a dependency... Looking into reimplementing

larsoner:
Is it expected to work? If so, might be worth opening an issue on dipy, they might know a workaround

We could add another dependency if histogram registration is expected to work on most cases

alexrockhill:
> Is it expected to work? If so, might be worth opening an issue on dipy, they might know a workaround
> 
> We could add another dependency if histogram registration is expected to work on most cases

I don't think so, I don't think Dipy has any reason to use or have tested anything using CT images.

The Nipy histogram registration is written in cython so that would be a bit harder to reimplement, but I could give it a try and see how slow it is in pure python...

alexrockhill:
So this last version has all the parts working but there is some confusion converting between nibabel and nipy image objects that is not ideal. This adds a nipy dependency but I'm not sure how else to register the CT to the MR.

I think I'm going to start on the GUI next week and we can continue to iterate on this.

alexrockhill:
Asking on Neurostars about `HistogramRegistration` https://neurostars.org/t/histogramregistration-migration/19568

alexrockhill:
It appears as those there are some issues with the SDR morph to `fsaverage`. I think we'll have to use `cvs_avg35_inMNI152`.

Screen Shot 2021-06-20 at 1 07 08 PM

alexrockhill:
So weirdly some depreciation warnings crash Sphinx since ``nipy`` is not rigorously maintained currently. I made a PR here so that it Sphinx can run https://github.com/nipy/nipy/pull/488 but probably best to remove the dependency at some point in the near future.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,373 @@
...
+# the ACPC plane and the image should be rotated until they are symmetrical,
+# and the crosshairs should transect the midline of the brain.
+# Be sure to use both the rotate and the translate menus and save the volume
+# after you're finished using ``Save Volume As`` in the transform popup
+# :footcite:`HamiltonEtAl2017`.
+
+T1 = nib.freesurfer.load(op.join(misc_path, 'seeg', 'sample_seeg_T1.mgz'))
+viewer = T1.orthoview()
+viewer.set_position(0, 9.9, 5.8)
+viewer._axes[0].annotate('PC', (107, 108), xytext=(10, 75),
```
larsoner, line 92: Do not use private attributes in examples, can this be accessed via `viewer.figure.axes[0]` for example? Or maybe `viewer.fig.axes[0]`?

alexrockhill:
Here's the result (missing a few contacts). It does look like it works for fsaverage and like it's better than the Talairach transform. A bit more work than I thought though...

Screen Shot 2021-06-21 at 7 28 11 PM

alexrockhill:
Ok this should finally run in a reasonable time once https://github.com/mne-tools/mne-misc-data/pull/6 merges. Should be good to go if you have some time to review @larsoner!

larsoner:
I get this when running the example:
```
/home/larsoner/python/mne-python/tutorials/clinical/10_ieeg_localize.py:468: RuntimeWarning: Fiducial point nasion not found, assuming identity unknown to head transformation
  raw.set_montage(montage)
```
Can you use your GUI to mark nasion/lpa/rpa? Then we can have the coordinate axes properly labeled, and the head<->MRI coordinate transformation working properly.

alexrockhill:
> I get this when running the example:
> 
> ```
> /home/larsoner/python/mne-python/tutorials/clinical/10_ieeg_localize.py:468: RuntimeWarning: Fiducial point nasion not found, assuming identity unknown to head transformation
>   raw.set_montage(montage)
> ```
> 
> Can you use your GUI to mark nasion/lpa/rpa? Then we can have the coordinate axes properly labeled, and the head<->MRI coordinate transformation working properly.

Ok just pushed.

alexrockhill:
Wait we don't need to warp those points and can't since they're not in the brain, we need the fiducial points of `fsaverage` not the subject

alexrockhill:
Ok that makes the warning go away and everything lines up so that's better.

larsoner:
Ahh right. It would be nice if we could warp them and see how well it matches, but since we warp just `brain` now indeed it probably does not make sense.

So it runs okay on my machine, is the next step to find some way to speed it up? If it's for example in the affine or SDR computations, you could add the `mne.morph.write_dipy(...)` and `mne.morph.read_dipy` functions I mentioned somewhere, run locally, save, add to MNE-misc-data, comment out code in the example, and just load it. If it's not clear what I mean we can have a quick Discord chat...

mne/viz/utils.py
```diff
@@ -2351,3 +2351,56 @@ def _generate_default_filename(ext=".png"):
...
+    mc.Update()
+    polydata = mc.GetOutput()
+
+    verts = numpy_support.vtk_to_numpy(polydata.GetPoints().GetData())
+    triangles = np.zeros((polydata.GetNumberOfCells(), image.ndim),
+                         dtype=np.int64)
+    for i in range(triangles.shape[0]):
+        tri = polydata.GetCell(i)
+        for j in range(triangles.shape[1]):
+            triangles[i, j] = tri.GetPointId(j)
```
larsoner, line 2403:
```suggestion
    triangles = numpy_support.vtk_to_numpy(
        polydata.GetPolys().GetConnectivityArray()).reshape(-1, 3)
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+#
+# Then translate the image until the crosshairs meet on the AC and
+# run through the PC as shown in the plot. The eyes should be in
+# the ACPC plane and the image should be rotated until they are symmetrical,
+# and the crosshairs should transect the midline of the brain.
+# Be sure to use both the rotate and the translate menus and save the volume
+# after you're finished using ``Save Volume As`` in the transform popup
+# :footcite:`HamiltonEtAl2017`.
+
+T1 = nib.freesurfer.load(op.join(misc_path, 'seeg', 'sample_seeg_T1.mgz'))
```
larsoner, line 95:
```suggestion
T1 = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_T1.mgz'))
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+
+###############################################################################
+# Aligning the CT to the MR
+# =========================
+#
+# Let's load our CT image and visualize it with the T1 image.
+# As you can see, the CT is already aligned to the T1 image in this example
+# dataset.
+
+CT = nib.freesurfer.load(op.join(misc_path, 'seeg', 'sample_seeg_CT.mgz'))
```
larsoner, line 144:
```suggestion
CT = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_CT.mgz'))
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+# from the subject's brain to the template brain. We will use the symmetric
+# diffeomorphic registration (SDR) implemented by ``Dipy`` to do this.
+#
+# Before we can make a function to account for individual differences in the
+# shape and size of brain areas, we need to fix the alignment of the brains.
+# The plot below shows that they are not yet aligned.
+
+# load the freesurfer average brain
+template_brain = nib.freesurfer.load(
+    op.join(subjects_dir, subject, 'mri', 'brain.mgz'))
```
larsoner, line 307:
```suggestion
    op.join(subjects_dir, 'fsaverage', 'mri', 'brain.mgz'))
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+# is more complex, and the less-complex, linear Talairach transformation.
+# By accounting for the shape of this particular subject's brain using the
+# SDR to warp the positions of the electrode contacts, the position in the
+# template brain is able to be more accurately estimated.
+
+# load electrophysiology data
+raw = mne.io.read_raw(op.join(misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
+
+lpa, nasion, rpa = mne.coreg.get_mni_fiducials(
+    subject, subjects_dir=subjects_dir)
```
larsoner, line 466:
```suggestion
    'fsaverage', subjects_dir=subjects_dir)
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+ch_pos = dict(zip(ch_names, ch_coords / 1000))  # mm -> m
+montage = mne.channels.make_dig_montage(
+    ch_pos, coord_frame='mri', nasion=nasion, lpa=lpa, rpa=rpa)
+raw.set_montage(montage)
+
+# get trans
+trans = mne.channels.compute_native_head_t(montage)
+
+# plot the resulting alignment
+fig = mne.viz.plot_alignment(raw.info, trans, subject,
```
larsoner, line 479:
```suggestion
fig = mne.viz.plot_alignment(raw.info, trans, 'fsaverage',
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+
+print(__doc__)
+
+np.set_printoptions(suppress=True)  # suppress scientific notation
+
+# paths to mne datasets - sample sEEG and FreeSurfer's fsaverage subject
+# which is in MNI space
+misc_path = mne.datasets.misc.data_path()
+sample_path = mne.datasets.sample.data_path()
+subject = 'fsaverage'
```
larsoner, line 52: We usually reserve "subject" in tutorials like this to be the individual subject, so I wouldn't use `fsaverage` here. It's only used in a couple of places anyway.
```suggestion
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,487 @@
...
+        ax.axis('off')
+
+
+# Make an affine to transform the image
+unalign_affine = np.array([
+    [-1.01, 0.02, -0.01, 128],
+    [0.01, -0.02, 1.02, -135],
+    [0.06, -0.99, -0.02, 140],
+    [0, 0, 0, 1]])
+CT_unaligned = resample_from_to(CT, (CT.shape, unalign_affine))
```
larsoner, line 206: Do you not have the original-unaligned CT? Or did it come aligned already? I'm inclined to have some form of "unaligned CT" (can be this one) as the version stored in the dataset if it reflects the more general case that most people will have. It's better than faking it in the tutorial itself.

mne/viz/utils.py
```diff
@@ -2351,3 +2351,56 @@ def _generate_default_filename(ext=".png"):
     now = datetime.now()
     dt_string = now.strftime("_%Y-%m-%d_%H-%M-%S")
     return "MNE" + dt_string + ext
+
+
+def marching_cubes(image, level):
```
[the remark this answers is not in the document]
alexrockhill: What do you think about making `peak_to_volume` and `mne` function? Maybe renamed to `marching_voxels` or something. Where would it go if so?

alexrockhill:
> Ahh right. It would be nice if we could warp them and see how well it matches, but since we warp just `brain` now indeed it probably does not make sense.
> 
> So it runs okay on my machine, is the next step to find some way to speed it up? If it's for example in the affine or SDR computations, you could add the `mne.morph.write_dipy(...)` and `mne.morph.read_dipy` functions I mentioned somewhere, run locally, save, add to MNE-misc-data, comment out code in the example, and just load it. If it's not clear what I mean we can have a quick Discord chat...

That makes sense but is it necessary? Does it take too long?

tutorials/clinical/10_ieeg_localize.py
alexrockhill: Ok this was just copying the seeg tutorial

tutorials/clinical/10_ieeg_localize.py
alexrockhill: I thought you needed the freesurfer to get the header. I must be mistaken.

larsoner:
> That makes sense but is it necessary? Does it take too long?

Not for time reasons, no. But everything in a tutorial should be there for a reason and serve a clear pedagogical purpose. People reading them should think "I probably need to know this to understand this topic." Knowing how to make your CT *match worse* is not something that people should really need to know. We should start with something that is as close (and as general) as possible to what people will be working with, which I'm assuming is a mis-aligned CT and MRI. If this is the case, we shouldn't impose the misalignment in the tutorial, it should already be there from the beginning (i.e., the mismatch imposition should be hidden entirely from the person reading the tutorial.)

Blind code re-use can happen with tutorials, and we definitely don't want people copy-pasting the intentional deformation and not realizing that it's in there just because we had data that was too ideal already...

tutorials/clinical/10_ieeg_localize.py
larsoner: In that example, there is only ever one freesurfer subject: fsaverage. Here there are two: the actual subject, and the template (fsaverage).

mne/viz/utils.py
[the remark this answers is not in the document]
larsoner: To me marching cubes (subject of this comment) should go in `mne/surface.py`. But if you're talking about some iEEG-specific (probably) function that maps X/Y/Z locations from one MRI volume into another would probably be useful. But let's get this tutorial working and merged, then hash out an API for this second function in another issue / PR / discussion. Let's make it work, then make it pretty.

alexrockhill:
> > That makes sense but is it necessary? Does it take too long?
> 
> Not for time reasons, no. But everything in a tutorial should be there for a reason and serve a clear pedagogical purpose. People reading them should think "I probably need to know this to understand this topic." Knowing how to make your CT _match worse_ is not something that people should really need to know. We should start with something that is as close (and as general) as possible to what people will be working with, which I'm assuming is a mis-aligned CT and MRI. If this is the case, we shouldn't impose the misalignment in the tutorial, it should already be there from the beginning (i.e., the mismatch imposition should be hidden entirely from the person reading the tutorial.)
> 
> Blind code re-use can happen with tutorials, and we definitely don't want people copy-pasting the intentional deformation and not realizing that it's in there just because we had data that was too ideal already...

I was talking about `morph.save_dipy` and the length of the example. I already pushed the other CT. We'll see if it works, they are very far unaligned, may have to add a center of mass translation.

I think that's a good point, I had only done it the way I did because the nipy version didn't converge when it was too far away so I just manually aligned it roughly first.

I would say if you are copy-pasting something labeled `unalign_affine` into your code that is clearly explained in text right before, maybe that is on you for not being careful enough although I think the point is very well taken in general.

larsoner:
> I was talking about morph.save_dipy and the length of the example

Oh sorry I didn't read the quoted part properly at all! Yes the read/write is just a way to save time, both in CI and for users. If I were actually analyzing such data I'd want these functions so I could compute one good alignment and save it rather than having to wait over and over again. Locally it takes almost 5 minutes to run the tutorial (on a powerful workstation) which feels too long when going through it, and on CircleCI this will add way too much time, especially given that the way to fix it (read/write dipy) is so simple.

If it's not clear how to implement these or you want to move on to working on something else, I can push a quick commit I think.

alexrockhill:
Ok sounds good, I'll give it a shot.

The CT unaligned has too large a file size by the way.

larsoner:
> The CT unaligned has too large a file size by the way.

This should be fixable somehow -- it suggests that the file format it's being saved in isn't being preserved. Was the original uint8 (0-255) and you're saving float32? Did the original have a smaller domain? Different zooms? It should be possible to deform without the file size growing very much

alexrockhill:
> > The CT unaligned has too large a file size by the way.
> 
> This should be fixable somehow -- it suggests that the file format it's being saved in isn't being preserved. Was the original uint8 (0-255) and you're saving float32? Did the original have a smaller domain? Different zooms? It should be possible to deform without the file size growing very much

I think it's just the performance of the compression. I just used freeview, translated and saved as the same file type...

alexrockhill:
Actually probably the domain is way out of whack because it's way off to the side compared to the T1.

larsoner:
> I just used freeview, translated and saved as the same file type...

Internally freeview could be saving *any* transformed image as float32 instead of uint8. You can load the orig and transformed to confirm. You can also convert with nibabel and save back out again in whatever format you want

alexrockhill:
> > I just used freeview, translated and saved as the same file type...
> 
> Internally freeview could be saving _any_ transformed image as float32 instead of uint8. You can load the orig and transformed to confirm. You can also convert with nibabel and save back out again in whatever format you want

I figured it out, it was converted to nii and back, which made it larger. Pushing now.

mne/viz/utils.py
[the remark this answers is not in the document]
alexrockhill: Oops I missed seeing this comment and made a `voxel_neighbors` function in `mne/surface.py` it would be fine to pull out but the tutorial is a lot cleaner this way. Some of the slicing is really better not shown in a tutorial, seems unnecessarily confusing although generalizable to N dimensions.

alexrockhill:
Ok, I think that addressed all your comments @larsoner, I think the changes from your commit caused a few tests to fail though...

alexrockhill:
Once https://github.com/mne-tools/mne-misc-data/pull/6 merges, we should be good to go.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,505 @@
...
+# use mne-python's fsaverage data
+fetch_fsaverage(subjects_dir=subjects_dir, verbose=True)  # downloads if needed
+
+###############################################################################
+# Aligning the T1 to ACPC
+# =======================
+#
+# Let's load our T1-weighted image and visualize it. As you can see, the
+# image is already aligned to the anterior commissue and posterior commissure
+# (ACPC). This is recommended to do before starting. This can be done using
```
larsoner, line 58: Why is this recommended? it doesn't seem like it should be necessary for the affine to be computed correctly. The center-of-mass alignment should render this step unnecessary

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,505 @@
...
+        ax.invert_yaxis()
+        ax.axis('off')
+    fig.tight_layout()
+    return fig
+
+
+CT = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_CT.mgz'))
+
+# resample to T1 shape
+CT_resampled = resample(moving=CT.get_fdata(),
```
larsoner, line 165: I would call this `CT_aligned` and the one above `CT_orig`

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,505 @@
...
+    moving=mri_from,
+    static=mri_to,
+    moving_affine=affine_from,
+    static_affine=affine_to,
+    nbins=32,
+    metric='MI',
+    pipeline=[center_of_mass, translation, rigid, affine],
+    level_iters=[100, 100, 10],
+    sigmas=[3.0, 1.0, 0.0],
+    factors=[4, 2, 1])[1]
```
larsoner, line 383: This code is *very* similar to what we have with the code in `mne.morph._compute_morph_sdr`, and it's also referenced above. I think it's worth exposing some public function that calls that function internally to simplify this (the public version would probably only be a few lines at most). Maybe `mne.morph.align_volumes(img_from, img_to, zooms=None, n_iter_sdr=(...), ...)` and the call here would be something like:
```
affreg, _ = mne.morph.align_volumes(
    subject_brain, template_brain, zooms=(5, 5, 5),  # zooms != 1 here just for speed
    n_sdr_iter=())  # skip SDR step just for speed
reg_affine = affreg.affine
```
The second output here would be the SDR morph, which is None because `n_sdr_iter=()` here. And below we just say the complete call that was used actually didn't set `n_sdr_iter`, and is skipped here for speed because it takes too long. This is kind of where I was headed with the refactoring of `mne.morph._compute_morph_sdr`, so it shouldn't require too much code to get there I don't think. (The logging provided by that, including the R^2 values, is really nice, too, I think.)

But does it really still take too long when we use `zooms=(5, 5, 5)` and tweak the iterations? We use this in another example and it's fine:

https://github.com/mne-tools/mne-python/blob/653ada8b9b0d4937d0944afd5ad29712600892a5/examples/inverse/morph_volume_stc.py#L83-L87

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -0,0 +1,505 @@
...
+#     from dipy.align.metrics import CCMetric
+#     from dipy.align.imwarp import SymmetricDiffeomorphicRegistration
+#     # Compute registration
+#     sdr = SymmetricDiffeomorphicRegistration(
+#         metric=CCMetric(3), level_iters=[10, 10, 5])
+#     mapping = sdr.optimize(static=template_brain.get_fdata(),
+#                            moving=subject_brain.get_fdata(),
+#                            static_grid2world=template_brain.affine,
+#                            moving_grid2world=subject_brain.affine,
+#                            prealign=reg_affine)
```
larsoner, line 413: Ahh so one difference is that this is computed at the native resolution rather than the reduced space. So maybe we need an API like `zooms=dict(affine=(5, 5, 5), sdr=(1, 1, 1))` in the `mne.morph.align_volumes` function.

larsoner (commented):
Great progress! I think working on this example is useful for seeing what *actually* needs to be done to make this work, thanks for sticking with it. The `align_volumes` seems like the simplest solution I can think of, but if you see problems with it maybe we should have a quick chat tomorrow.

Also if you are sick of iterating on this and want to move onto GUI I can try implementing my suggested changes locally. I can also fix the tests I broke if it would help.

alexrockhill:
> Also if you are sick of iterating on this and want to move onto GUI I can try implementing my suggested changes locally. I can also fix the tests I broke if it would help.

If you wouldn't mind fixing the tests you broke, I'm not exactly sure what you were doing there. You're welcome to make some changes if you'd like, I stopped working on it today.

larsoner:
I won't get to it until tomorrow morning EDT but I'll push something then after pulling your latest changes

tutorials/clinical/10_ieeg_localize.py
alexrockhill: I guess this is extraneous, it was in `img_pipe` so I thought there was a good reason for it but I guess `recon-all` works just as well without it.

larsoner: I would ask whoever wrote the recommendation what it's based on. If it's just "nice to have" then maybe we don't need it.

Or maybe better we can update the dataset (again!) to have the actual original, not-ACPC-aligned T1, plus a voxel number (found in FreeView) of the ACPC, and write a little function `img_realigned = mne.morph.realign_acpc(img, acpc_voxel)` that does this. That could be a follow-up PR, though :)

larsoner: (My naive sense is that this is probably just "good FreeSurfer practice", in which case it should live in some separate FreeSurfer tutorial and we should talk about this there. But if it's for example standard in iEEG to do this then it should live here.)

tutorials/clinical/10_ieeg_localize.py
alexrockhill: Yes, it still takes too long because those are downsampled by a factor of 5 and for accuracy, the SDR needs to be computed on the full-resolution brain although the affine registration can be on the downsampled brain.

Maybe this should be another PR to get this moving. Or if you want to refactor that would be good too but that's more things that might break and have to be fixed on the way and this is already getting quite extensive.

tutorials/clinical/10_ieeg_localize.py
alexrockhill: Yeah, that was the gist of my comment above, sorry should have read this one first.

alexrockhill:
> I won't get to it until tomorrow morning EDT but I'll push something then after pulling your latest changes

Thanks for all the help! Sounds good, I think what needs to happen is that the tests need to be fixed and if you want to refactor the SDR morph you could do that but I might save it for a followup PR as it's going to affect the other SDR tests.

tutorials/clinical/10_ieeg_localize.py
larsoner: Okay for now I'll just fix the tests I broke and then we can hopefully merge. Then I'll make a quick follow-up to finish the refactoring / new function idea. That's mostly separable anyway so will make reviewing easier anyway!

mne/surface.py
```diff
@@ -1646,3 +1646,112 @@ def dig_mri_distances(info, trans, subject, subjects_dir=None,
...
+
+    Returns
+    -------
+    voxels : set
+        The set of locations including the ``seed`` voxel and
+        surrounding it that are above a threshold.
+    """
+    seed = np.array(seed).round().astype(int)
+    check_grid = image[tuple([
+        slice(idx - max_peak_dist, idx + max_peak_dist + 1) for idx in seed])]
```
larsoner, line 1743: How much of this can be replaced by https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html ?

alexrockhill: I gave it a college try, but it really just doesn't work very well or at least I don't know how to make it work. See 0e449d5 if you want to give it a try but I think having a specific case is really helpful for being computationally efficient. I think messing with parameters it could maybe work but there is an issue that one whole electrode gets labeled together which makes it tougher to sort out on the back end.

alexrockhill:
Ok, I'm happy to keep iterating if you have a chance to look at it tomorrow @larsoner but this looks like a viable tutorial to me

larsoner:
Thanks @alexrockhill !

requirements_doc.txt
```diff
@@ -10,3 +10,4 @@ seaborn
 sphinx_copybutton
 https://github.com/mne-tools/mne-bids/archive/main.zip
 pyxdf
+nipy
```
alexrockhill, line 13: We probably don't need this

PR #9486, opened 2021-06-17
ENH: Expand ~ to home dir in fetch_fsaverage()

hoechenberger:
I passed `subjects_dir='~/foo'` and the data ended up in `~/mne_data/~/foo`, which was not nice.


agramfort:
thx @hoechenberger

larsoner:
It's probably better to make this fix in `_check_fname`, then make a call to `_check_fname(..., overwrite='read', must_exist=True, need_dir=True)`. It's more general, has some nice fixes (e.g., checks that `SUBJECTS_DIR` actually exists, for example), and will make `~` expansion much more general than one isolated function.

Issue #9421, opened 2021-05-21
GSoC 2021 - Refactoring mne.connectivity into mne-connectivity repository


adam2392:
This issue is meant to track the progress of refactoring existing functionality inside mne-python's "connectivity" sub-module into an entirely new repository.

- [ ] Create new repository, `mne-connectivity`, inside mne.tools (I think I have access)
- [ ] Get CI and basic GH action tools setup, based on `mne-bids` and `mne-python` CI
- [ ] Refactor existing code into the `mne-connectivity` repo and include a validation notebook that validates that code in this refactored setting matches what is here

To track detailed progress inside `mne-connectivity`, such as getting CI up and running, I'll point to issues in that specific repository.

## Higher level goals

I think @agramfort spoke about this a bit with me and I agree: there is really no "great way" of saying which algorithm or model is better in certain instances. There are many "proposed connectivity algorithms", whether they are time-domain, or free-domain based. So the metric moving forward in this new repo is to at least standardize what you should expect on a dataset.

Given 1-3 datasets, running a specific algorithm, one should expect certain performance or values. The current goal is to leverage openneuro and pull datasets for doing this in CI.

## Reference Issues

#9060 #9030 #8729

larsoner:
@adam2392 I read this only after making my comment in the other repo, which seems to echo almost 1:1 what you wrote here -- so sounds good to me :)

> The current goal is to leverage openneuro and pull datasets for doing this in CI.

I would start smaller and try using toy or simulated datasets -- preferably there are some of these in the reference papers. Once that works, examples that build on CircleCI to give outputs for real-data inputs are great, but I wouldn't start from these if possible because debugging them will be harder both in terms of eyeballing errors and in terms of CI time (they are going to require a big download and a long processing time compared to a toy example).

adam2392:
@drammock @larsoner and @britta-wstnr and I were discussing today and they suggested I ask you:

What do we do with all the mne.connectivity examples inside mne-python? 
- should we have a link directly to the mne-connectivity package?
- what’s the best way to advertise this?

Any preferences there?

drammock:
> What do we do with all the mne.connectivity examples inside mne-python?

I can think of 2 broad options:

1. move the examples to the doc site of `mne-connectivity`
2. leave them on the MNE-Python site, but import the needed functions from `mne-connectivity`

[...]

PR #9493, opened 2021-06-21
[MRG, API] Initiate deprecation of mne.connectivity submodule

adam2392:
#### Reference issue
Fixes: #9421 

Initiates the deprecation of mne.connectivity into `mne-connectivity` 

#### Additional information
TODO:

- [x] figure out how to do CI releases on men-connectivity because I can't do a release manually. M1 Macs and even my Linux computer stalling when building the docs locally with the graphs. Not sure why. -> Make v0.1 release which is basically the same as what men-python.connectivity has rn

Afterwards, all changes to connectivity will occur in this new repo with added features documented in that codebase.


adam2392:
Not sure what's the best practice for dealing with deprecation warnings in the unit tests? Should I "explicitly ignore" these?

agramfort:
you should filter warnings manually in the affected tests

however merging this should be done many in a few weeks when mne-connectivity has been consolidated and released

adam2392:
Added a `pytest.ini` file for ignoring the deprecation warnings inside connectivity sub-module.

adam2392:
Rn still finagling with the mne-connectivity docs. They all build fine on a PR branch, but need to get deployed correctly still. 

Once that is done, the `examples/` for connectivity can be erased with a link to the `examples` in mne-conn

larsoner:
We have a global list of warnings in `mne/conftest.py`, we should not start a `pytest.ini` just for this (it creates a second location for global ignores). I suggest having a `mne/connectivity/conftest.py` that ignores warnings just in that submodule, similar to what we have in `mne/conftest.py` (the root `mne/conftest.py` one and the `mne/connectivitiy/conftest.py` should work together automagically via pytest). That way when we delete that folder via deprecation, it goes away automatically

adam2392:
@britta-wstnr just FYI this is what we were talking bout, not sure how these tests are now changing.

adam2392:
I have no idea why adding these depreciation warnings are all of a sudden making a unit test fail...? I even tried another fresh PR branch.

mmagnuski:
@adam2392 for example `test_epochs_tmin_tmax` expects that at some point it would get one warning, but now it gets two, which results in assertion error (2 == 1).

adam2392:
I'm not sure why, but the `pytest_configure` hook is not getting loaded in properly with the `conftest.py` I added. 

However, I can just decorate the tests to ignore that Deprecation Warning. This still accomplishes what we want, since come v0.25, the entire directory can just be deleted safely.

Reference:
- https://docs.pytest.org/en/latest/reference/fixtures.html

britta-wstnr:
The problem is that all the examples and tutorials now throw the deprecation warning. Usually we would change the code such that this does not happen (e.g. by explicitly changing parameters in a function call away from the deprecated parameter), but that  is of course not possible here. 
Pinging @drammock : what is the best thing to do with the examples in this case?

agramfort:
I would make a release mne Connectivity first and then deprecate here and have each old example just point to the new mne connectivity documentation. Clear ?

adam2392:
> I would make a release mne Connectivity first and then deprecate here and have each old example just point to the new mne connectivity documentation. Clear ?

I did end up releasing mne-connectivity v0.1 already, which is an exact copy of the code here. 

When you say point to the new mne-connectivity documentation, you mean just forward it to the new link? If so, this is clear. Lmk if I misunderstood.

Is there a way to specify in sphinx on how to do so? @drammock

drammock:
I think we discussed this before, and the decision was to do 3 things:

1. This file (https://github.com/mne-tools/mne-python/blob/main/examples/connectivity/README.txt) should be updated to say something like "the connectivity functionality has moved into the mne-connectivity package" (and should include a link to examples gallery in your docs)
2. The corresponding `.py` files here (https://github.com/mne-tools/mne-python/tree/main/examples/connectivity) should be removed from MNE-Python (not sure if we discussed page-by-page redirects...  I'm neutral on whether to bother with those)
3.  the homepage gallery tile titled "connectivity" will need to link to the correct example in your docs, and the image used to represent it on our homepage will need to be updated to be an image from your doc build (or alternatively, will need to be hard-added as a `_static` asset to our site).

adam2392:
> 1. This file (https://github.com/mne-tools/mne-python/blob/main/examples/connectivity/README.txt) should be updated to say something like "the connectivity functionality has moved into the mne-connectivity package" (and should include a link to examples gallery in your docs)

Ah okay done now.

> 2. The corresponding `.py` files here (https://github.com/mne-tools/mne-python/tree/main/examples/connectivity) should be removed from MNE-Python (not sure if we discussed page-by-page redirects...  I'm neutral on whether to bother with those)

Okay this makes it way easier then and we can get rid of the sphinx warnings.

> 3. the homepage gallery tile titled "connectivity" will need to link to the correct example in your docs, and the image used to represent it on our homepage will need to be updated to be an image from your doc build (or alternatively, will need to be hard-added as a `_static` asset to our site).

Not sure I follow here.

https://mne.tools/dev/auto_examples/index.html#connectivity-analysis-examples here shows many images. Do you mean, linking then each example now to the image used in my doc build on `mne-connectivity`?

Thanks!!

drammock:
"homepage gallery tile" means one of the six tiles on https://mne.tools/dev/

One of them is titled "connectivity" and shows the circular graph.  That is a link, and the link will fail after this PR because you just deleted the example that it links to.  Also, the image shown in that tile is auto-generated by Sphinx by executing that (now-deleted) example, so we need a source for that image; options are (1) add the jpg or png as a static asset within our source tree (e.g., in `doc/_static`) or (2) hard-code a URL that pulls the image from your docs.  I lean toward option 1

larsoner:
I think the `mne/connectivity/conftest.py` fails because our root-level conftest.py adds an `error` line. A simple fix is to add this to `mne/conftest.py`, which seems targeted enough:
```diff
diff --git a/mne/conftest.py b/mne/conftest.py
index 68ac1a8d6..df2c6e7d2 100644
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -104,6 +104,7 @@ def pytest_configure(config):
     ignore:Call to deprecated method.*Deprecated since.*:DeprecationWarning
     always:.*get_data.* is deprecated in favor of.*:DeprecationWarning
     ignore:.*rcParams is deprecated.*global_theme.*:DeprecationWarning
+    ignore:.*mne.connectivity sub-modules will be housed.*:DeprecationWarning
     always::ResourceWarning
     """  # noqa: E501
     for warning_line in warning_lines.split('\n'):
```
Then the only failure is:
```
mne/connectivity/tests/test_spectral.py:272: in test_epochs_tmin_tmax
    assert len(w) == 1  # just one even though there were multiple epochs
E   assert 2 == 1
E    +  where 2 = len(WarningsChecker(record=True))
```
and I think it's fine to just make that 2 instead of 1.

This should be simpler than adding a bunch of function-level decorators.

adam2392:
Kay I think this is good to go now.

I added `mne-connectivity` as a testing dependency for the sake of `60_make_fixed_length_epochs.py` tutorial. 

However, I imagine, we might want to remove that part of the tutorial? If so, do we want to add that to `mne-connectivity`? 

I can add it instead as a tutorial on "using mne-python to make fixed length epochs to apply connectivity measures to" on the mne-connectivity side.

adam2392:
> I added `mne-connectivity` as a testing dependency for the sake of `60_make_fixed_length_epochs.py` tutorial.

Spoke with @drammock on Discord and this seems sufficient to go for now. The sequence would be:

- once we release mne-connectivity v0.2+, we update a PR here in this tutorial for the API.

This brings about i) increased visibility for mne-conn and ii) more linkage between the two docs/APIs.

doc/changes/latest.inc
```diff
@@ -188,3 +188,5 @@ API changes
 - Split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting and make the default for ``row`` and ``col`` apply to all rows and columns (:gh:`9596` by `Alex Rockhill`_)
 
 - Deprecate ``solver='prox'`` in :func:`mne.inverse_sparse.mixed_norm` in favor of ``solver='cd'`` and ``solver='bcd'`` as coordinate descent solvers consistently outperform proximal gradient descent (:gh:`9608` by `Pierre-Antoine Bannier`_)
+
+- All `mne.connectivity` functions are deprecated to be fully removed in v0.25 (:gh:`9493` by `Adam Li`_).
```
britta-wstnr, line 192: Might be good to mention that they got moved - now it sounds as they just got deleted completely.

adam2392:
I get the following warnings now.

```
WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.degree':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'connectivity', 'threshold_prop'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.seed_target_indices':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'seeds', 'targets'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.degree':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'connectivity', 'threshold_prop'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.seed_target_indices':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'seeds', 'targets'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.circular_layout':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'start_between', 'group_sep', 'start_pos', 'node_names', 'group_boundaries', 'node_order'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_connectivity_circle':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'linewidth', 'colorbar', 'colorbar_size', 'node_edgecolor', 'fontsize_title', 'colormap', 'fig', 'interactive', 'facecolor', 'node_angles', 'colorbar_pos', 'node_colors', 'con', 'padding', 'indices', 'show', 'vmax', 'node_linewidth', 'fontsize_names', 'subplot', 'fontsize_colorbar', 'title', 'node_names', 'node_width', 'n_lines', 'vmin', 'textcolor'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_sensors_connectivity':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'picks', 'con', 'cbar_label', 'info'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.circular_layout':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'start_between', 'group_sep', 'start_pos', 'node_names', 'group_boundaries', 'node_order'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_connectivity_circle':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'linewidth', 'colorbar', 'colorbar_size', 'node_edgecolor', 'fontsize_title', 'colormap', 'fig', 'interactive', 'facecolor', 'node_angles', 'colorbar_pos', 'node_colors', 'con', 'padding', 'indices', 'show', 'vmax', 'node_linewidth', 'fontsize_names', 'subplot', 'fontsize_colorbar', 'title', 'node_names', 'node_width', 'n_lines', 'vmin', 'textcolor'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_sensors_connectivity':
  PR01: Parameters {'*args', '**kwargs'} not documented
  PR02: Unknown parameters {'picks', 'con', 'cbar_label', 'info'}
```

I tried adding something to `conf.py` to ignore `connectivity` keyword in numpydoc?

doc/changes/latest.inc
adam2392: Added mention to mne-conn.

tutorials/epochs/60_make_fixed_length_epochs.py
```diff
@@ -14,6 +14,9 @@
 epochs by segmenting the data around the onset of each stimulus, we will
 create 30 second epochs that allow us to perform non-event-related analyses of
 the signal.
+
+Note: Starting version v0.25, mne.connectivity sub-modules will be housed in
+``mne-connectivity``. Download it by  running ``pip install mne-connectivity``
```
drammock, line 19:
```suggestion
.. note::
    Starting in version 0.25, all functions in the ``mne.connectivity``
    sub-module will be housed in a separate package called
    ``mne-connectivity``. Download it by  running
    ``pip install mne-connectivity``.
```

doc/conf.py
```diff
@@ -733,8 +735,8 @@ def append_attr_meth_examples(app, what, name, obj, options, lines):
              alt='Clusters'),
         dict(title='Connectivity',
              text='All-to-all spectral and effective connec\u00ADtivity measures.',  # noqa E501
-             url='auto_examples/connectivity/mne_inverse_label_connectivity.html',  # noqa E501
-             img='sphx_glr_mne_inverse_label_connectivity_001.png',
+             url='https://mne.tools/mne-connectivity/stable/auto_examples/mne_inverse_label_connectivity.html',  # noqa E501
+             img='_static/temp_sphx_glr_mne_inverse_label_connectivity_001.png',  # noqa E501
```
drammock, line 739: it looks like you haven't actually added this file to the repo?  That will need to happen in this PR.  Also please give it a name that makes it more obvious that it's not temporary and not generated by sphinx, maybe `homepage_connectivity_tile.png` or similar.

examples/connectivity/README.txt
```diff
@@ -4,3 +4,5 @@ Connectivity Analysis Examples
 
 Examples demonstrating connectivity analysis in sensor and source space.
 
+Note: the connectivity functionality has moved into the mne-connectivity package.
+See https://mne.tools/mne-connectivity/stable/auto_examples/index.html.
```
drammock, line 8: The `:mod:` (module) link in this suggestion will only work if you've added `mne-connectivity` to the intersphinx mapping in `doc/conf.py` (which you should do).
```suggestion
Examples demonstrating connectivity analysis in sensor and source space.

.. note::
    Connectivity functionality has moved into the :mod:`mne-connectivity`
    package. Examples can be found at
    https://mne.tools/mne-connectivity/stable/auto_examples/index.html.
```

britta-wstnr:
I assume this happens because the function still show up in the API reference list on the homepage?

drammock (commented):
I can't immediately figure out why you're getting those numpydoc errors.  One thing I can think to try is maybe don't include `extra=` before each deprecation message?  That's the only thing I can see that is different from other currently deprecated funcs in our codebase (like `tf_dics`) and maybe passing the message as a positional argument will somehow behave differently with `deprecation_wrapped`.

If that *does* work, then we should also make `extra` into a positional-only (non-keyword) argument by adding a `/` into the signature for `deprecated`'s  `__init__` function (see https://stackoverflow.com/a/24735582/1664024)

drammock:
> I assume this happens because the function still show up in the API reference list on the homepage?

I think though that they *should* still be part of the API (until after we cut the 0.24 release, which will be in October probably). At that point we're on to 0.25.dev0 and can *remove* (rather than deprecate) them, including their API reference entries in `doc/connectivity.rst`

britta-wstnr:
In the built docs, the functions show up as `mne.viz.plot_connectivity_circle(*args, **kwargs)`  (but still with the list of parameters below) instead of `mne.viz.plot_connectivity_circle(con, node_names, indices=None, n_lines=None, node_angles=None, node_width=None, node_colors=None, facecolor='black', textcolor='white', node_edgecolor='black', linewidth=1.5, colormap='hot', vmin=None, vmax=None, colorbar=True, title=None, colorbar_size=0.2, colorbar_pos=(- 0.3, 0.1), fontsize_title=12, fontsize_names=8, fontsize_colorbar=8, padding=6.0, fig=None, subplot=111, interactive=True, node_linewidth=2.0, show=True)`
which is the mismatch that is found, right?

Any idea why this would happen, @drammock ? Maybe indeed the `extra`?

drammock:
> Any idea why this would happen, @drammock ? Maybe indeed the `extra`?

that is my best guess but I don't have time to test it myself at the moment so @adam2392 may need to do a little testing/digging. `*args, **kwargs` show up in `mne/utils/docs.py` in the definition of `deprecation_wrapped` which is what made me suspect that as the culprit.

doc/conf.py
adam2392: Ah it was in the gitignore

adam2392:
Unfortunately still getting the warnings even with the new changes :(

```
WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.degree':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'connectivity', 'threshold_prop'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.seed_target_indices':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'targets', 'seeds'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.degree':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'connectivity', 'threshold_prop'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.connectivity.seed_target_indices':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'targets', 'seeds'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.circular_layout':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'group_boundaries', 'node_order', 'start_between', 'group_sep', 'start_pos', 'node_names'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_connectivity_circle':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'con', 'colorbar_size', 'n_lines', 'padding', 'colorbar_pos', 'colormap', 'interactive', 'linewidth', 'indices', 'colorbar', 'fig', 'node_linewidth', 'node_colors', 'show', 'title', 'node_edgecolor', 'subplot', 'node_width', 'node_angles', 'textcolor', 'fontsize_title', 'fontsize_colorbar', 'vmax', 'vmin', 'node_names', 'fontsize_names', 'facecolor'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_sensors_connectivity':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'cbar_label', 'con', 'info', 'picks'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.circular_layout':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'group_boundaries', 'node_order', 'start_between', 'group_sep', 'start_pos', 'node_names'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_connectivity_circle':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'con', 'colorbar_size', 'n_lines', 'padding', 'colorbar_pos', 'colormap', 'interactive', 'linewidth', 'indices', 'colorbar', 'fig', 'node_linewidth', 'node_colors', 'show', 'title', 'node_edgecolor', 'subplot', 'node_width', 'node_angles', 'textcolor', 'fontsize_title', 'fontsize_colorbar', 'vmax', 'vmin', 'node_names', 'fontsize_names', 'facecolor'}

WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.viz.plot_sensors_connectivity':
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'cbar_label', 'con', 'info', 'picks'}
```

britta-wstnr:
It could be connected to the double backticks in the deprecation message. At least that is the commit, where thing diverted from correct API references to the `args, kwargs` issue in the doc build. (Commit `ca61c7e`)
Can you check if that is the case?

adam2392:
> It could be connected to the double backticks in the deprecation message. At least that is the commit, where thing diverted from correct API references to the `args, kwargs` issue in the doc build. (Commit `ca61c7e`)
> Can you check if that is the case?

Trying it w/o the double backticks and using quotation marks instead.

britta-wstnr:
Mh, still failing. 

> that is my best guess but I don't have time to test it myself at the moment so @adam2392 may need to do a little testing/digging. `*args, **kwargs` show up in `mne/utils/docs.py` in the definition of `deprecation_wrapped` which is what made me suspect that as the culprit.

This theory by @drammock seems still likely to be part of the problem, because if you click `[source]` on the rendered API reference page for one of the failing references (I tried `plot_connectivity_circle`), it takes you actually to:
https://github.com/mne-tools/mne-python/blob/4fd5b1d8dc0e89de4b0953cb4ca1825774e27588/mne/utils/docs.py#L3033-L3035

drammock:
@adam2392 I pushed a commit that I thought had fixed the numpydoc problems (at least they didn't happen locally for me anymore)... but after a `cd doc; make clean; make` they're back again.  Strangely the errors went away on the *second* build of the docs without doing a `make clean` in between...  I was hoping that the trick might be the relative ordering of `@verbose` and `@deprecated` decorators.  But it seems something even more subtle is at work here.  I'm starting to suspect a NumPyDoc bug.

drammock:
So I've restored the original order (with `@deprecated` decorators being the *outermost* decorator) because switching the order had broken the docstring tests.  The numpydoc warnings are still happening for the *first* local build of the docs after a `make clean` but not happening on subsequent builds, and I'm out of ideas at the moment for how to fix it.  So I expect we will still see a circleCI failure but there may be nothing to do other than wait for one of us to have time to bang our heads against it some more.

I also found and fixed a missing `footcite / footbibliography` in the `phase_slope_index` docstring, and also in that docstring fixed the formatting of the code sample.  @adam2392 you'll probably want to copy the changes in 660071e into the corresponding docstring in `mne-connectivity`.

drammock:
I've stumbled on a clue, which is that the numpydoc errors don't happen and the `[source]` links appear to go to the right place now for any of the functions that have a `@verbose` decorator.  The only ones failing are the ones that have only `@fill_doc` or have no other decorators besides the deprecation one.  I suspect this means that there is something insufficient about how we implement the `@deprecated` decorator... maybe it needs to be doing something additional that `@verbose` does?

adam2392:
But for example, in `mxne_optim` everything is fine w/o verbose?

```
@deprecated(extra="Starting version v0.25, MxNE problems will be solved "
                  "using exclusively (block) coordinate descent solvers. "
                  "Prefer using solver=`bcd` or solver=`cd`.")
```

drammock:
Can you look at git blame for the mxne deprecation and see if that pr did anything tricky like ignoring warnings?

Sent from ProtonMail mobile

britta-wstnr:
@drammock I already looked through recent deprecations lately. The one you mentioned did not deprecate functions that would also show up in the API reference. The deprecation of the time-frequency DICS beamformer a while ago did, but there I did not do anything beyond what we are doing here ... At least within MNE, the deprecation functions did not change since then.

mne/viz/circle.py
```diff
@@ -122,9 +122,8 @@ def _plot_connectivity_circle_onpick(event, fig=None, axes=None, indices=None,
         fig.canvas.draw()
 
 
-@deprecated("Starting version v0.25, mne.connectivity sub-modules "
-            "will be housed in 'mne-connectivity'. Download it "
-            "by running 'pip install mne-connectivity'.")
+@deprecated(CONNECTIVITY_DEPRECATION_MSG)
+@verbose
```
britta-wstnr, line 126: e.g. this function does not take `@verbose` so now test fails.

adam2392:
Okay, I'ma just wait for @larsoner  help :p I'll revert these changes to where only numpydoc warnings were giving the error.

mne/viz/circle.py
agramfort: yes you cannot decorate with verbose a function that does not have a verbose parameter

doc/conf.py
```diff
@@ -733,8 +734,8 @@ def append_attr_meth_examples(app, what, name, obj, options, lines):
              alt='Clusters'),
         dict(title='Connectivity',
              text='All-to-all spectral and effective connec\u00ADtivity measures.',  # noqa E501
-             url='auto_examples/connectivity/mne_inverse_label_connectivity.html',  # noqa E501
-             img='sphx_glr_mne_inverse_label_connectivity_001.png',
+             url='https://mne.tools/mne-connectivity/stable/auto_examples/mne_inverse_label_connectivity.html',  # noqa E501
+             img='_static/temp_homepage_connectivity.png',  # noqa E501
```
larsoner, line 738: This should be changed to a `https://mne.tools/mne-connectivity/stable/_images/...` URL

doc/overview/datasets_index.rst
```diff
@@ -119,9 +119,6 @@ Resting state
 
 Details can be found at the Brainstorm `resting state dataset tutorial`_.
 
-.. topic:: Examples
-
-    * :ref:`ex-envelope-correlation`
```
larsoner, line 124: Ideally instead of removing this, we would use intersphinx and be able to do
```
* :ref:`mne-connectivity:ex-envelope-correlation`
```
or similar...

examples/connectivity/README.txt
```diff
@@ -4,3 +4,7 @@ Connectivity Analysis Examples
 
 Examples demonstrating connectivity analysis in sensor and source space.
 
+.. note::
+    Connectivity functionality has moved into the :mod:`mne-connectivity`
+    package. Examples can be found at
+    https://mne.tools/mne-connectivity/stable/auto_examples/index.html.
```
larsoner, line 10: ... and instead of this being a static URL, would be another intersphinx `:ref:`

larsoner:
Pushing a commit to hopefully deal with my comments and the CircleCI failure soon...

larsoner:
> The numpydoc warnings are still happening for the first local build of the docs after a make clean but not happening on subsequent builds, and I'm out of ideas at the moment for how to fix it.

This is just because subsequent builds do not rebuild the API pages of the docs because no changes are made to them between builds (i.e., Sphinx change detection is working properly). The errors are still there.

The easy fix is just to add the functions to the numpydoc validation ignore set. This is effectively what we do automagically in `test_docstring` or whatever. It's easy enough to add them so I went that route.

I also fixed some docstring errors, and otherwise this should be good to go I think

larsoner (changes requested):
Actually it looks like there are still no HTML redirects in `conf.py`. @adam2392 do you want to look into doing that, or would you prefer if @drammock or I gave it a shot?

adam2392:
Horray! numpydoc warnings are fixed. Thanks @larsoner 

> Actually it looks like there are still no HTML redirects in `conf.py`. @adam2392 do you want to look into doing that, or would you prefer if @drammock or I gave it a shot?

Sorry not sure I understand. What do you mean by HTML redirect? All the connectivity examples now point to `mne-connectivity`.

larsoner:
> All the connectivity examples now point to mne-connectivity

Currently we have the URL people might have bookmarked (and is indexed by Google):

https://mne.tools/stable/auto_examples/connectivity/mne_inverse_label_connectivity.html

In this PR, the equivalent URL is now just a 404:

https://32739-1301584-gh.circle-artifacts.com/0/dev/auto_examples/connectivity/mne_inverse_label_connectivity.html

This URL should exist (not be a 404) as a redirect to:

https://mne.tools/mne-connectivity/stable/auto_examples/mne_inverse_label_connectivity.html

Same for all other removed examples. It's cheap/easy for us to add these redirects and they will make people's lives easier.

adam2392:
> Same for all other removed examples. It's cheap/easy for us to add these redirects and they will make people's lives easier.

Ah I see! Okay I added all the html redirects now.

larsoner:
```
File "/home/circleci/project/doc/conf.py", line 1091, in make_redirects
    assert os.path.isfile(to_path), to
AssertionError: https://mne.tools/mne-connectivity/stable//auto_examples/mne_inverse_label_connectivity.html
```
I would just nest this in a clause like `if not to_path.startswith('http'):` as we only need to check relative target paths not absolute

adam2392:
> I would just nest this in a clause like `if not to_path.startswith('http'):` as we only need to check relative target paths not absolute

It seems like the error is still coming up. Any thoughts on how to address that?

larsoner:
Thanks @adam2392 !

adam2392:
> Thanks @adam2392 !

Thanks @larsoner for helping address all the CI issues!

drammock:
@adam2392 there was one example that got missed:

```
  File "/home/circleci/project/examples/inverse/psf_ctf_label_leakage.py", line 130, in <module>
    node_angles = circular_layout(label_names, node_order, start_pos=90,
  File "/home/circleci/project/mne/utils/docs.py", line 3034, in deprecation_wrapped
    warnings.warn(msg, category=DeprecationWarning)
DeprecationWarning: Function circular_layout is deprecated; Functions in the ``mne.connectivity`` sub-module have moved to a new package (mne-connectivity) and will be removed in MNE-Python version 0.25. Install the new connectivity package by running ``pip install mne-connectivity`` in a system terminal or anaconda prompt.
```

Can you fix in a follow-up PR?

larsoner:
I'd be +1 on keeping `circular_layout` in MNE proper, since there are compelling uses outside MNE-Connectivity (such as PSF/CTF) actually

PR #9499, opened 2021-06-22
Don't warn when anonymizing w/o daysback and when meas_date is None

hoechenberger:
Don't warn when anonymizing w/o daysback and meas_date is None

required for https://github.com/mne-tools/mne-bids/pull/815

### DIFF
--- a/mne/io/meas_info.py
+++ b/mne/io/meas_info.py
@@ -2228,17 +2228,16 @@ def anonymize_info(info, daysback=None, keep_his=False, verbose=None):
     none_meas_date = info['meas_date'] is None
 
     if none_meas_date:
-        warn('Input info has \'meas_date\' set to None.'
-             ' Removing all information from time/date structures.'
-             ' *NOT* performing any time shifts')
-        info['meas_date'] = None
+        if daysback is not None:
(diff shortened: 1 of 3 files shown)


mne/io/meas_info.py
```diff
@@ -2228,17 +2228,16 @@ def anonymize_info(info, daysback=None, keep_his=False, verbose=None):
     none_meas_date = info['meas_date'] is None
 
     if none_meas_date:
-        warn('Input info has \'meas_date\' set to None.'
-             ' Removing all information from time/date structures.'
-             ' *NOT* performing any time shifts')
-        info['meas_date'] = None
```
agramfort, line 2234: This line was not necessary?

hoechenberger: Yes, I believe it's a cruft leftover from a previous refactoring.

The code goes like this:

```python
    none_meas_date = info['meas_date'] is None

    if none_meas_date:
```
so when we're inside the `if` branch, `info['meas_date']` is already `None`

hoechenberger:
Failing CI seems unrelated, this is good to merge.

agramfort:
Thx @hoechenberger

PR #9503, opened 2021-06-24
MRG, ENH: Refactor MRI/CT code

larsoner:
@alexrockhill can you take over the first two steps below:

- [x] Fix example so that all electrodes get assigned locations (currently the `np.where` can be length zero, which is bad)
- [x] Make sure example is still structurally okay

Then I can do for the functions I added:

- [x] Reduce footprint from `71.64 sec 2537.8 MB`
- [x] Add tests
- [x] Speed up tests
- [x] Add imposed translation, ensure translation is found in each pipeline
- [x] Update `SourceMorph` object attributes `niter_*` -> `niter` dict, add `pipeline`


larsoner:
Oh wow `71.64 sec   2537.8 MB` that's too much memory. I'll have to figure out how to reduce that...

larsoner:
Just pushed a commit to make CircleCI complete without error (my hack workaround for having none at the right value was wrong), feel free to push right over it @alexrockhill if you have a correct fix.

alexrockhill:
Hmmm, yeah I just tried and it doesn't work either. I think that's because the rigid registration can't be zoomed... Do you want to split it up into two steps? `mne.morph.compute_volume_translation` and `mne.morph.compute_volume_alignment`... I don't think warp is that accurate either, you're not changing the scaling, just rotating and translating

alexrockhill:
The reason `np.where` is empty is because the `elec_image` that got morphed was off because the alignment was off.

larsoner:
> I don't think warp is that accurate either, you're not changing the scaling, just rotating and translating

You must mean for the CT-MR warp step in the subject's own space. I'll I add a `rigid=False` param that skips the affine step that only does trans+rot.

> I don't think warp is that accurate either, you're not changing the scaling, just rotating and translating

Is this still in reference to the CT-MR warp, or the MR-template warp?

alexrockhill:
> > I don't think warp is that accurate either, you're not changing the scaling, just rotating and translating
> 
> You must mean for the CT-MR warp step in the subject's own space. I'll I add a `rigid=False` param that skips the affine step that only does trans+rot.
> 
> > I don't think warp is that accurate either, you're not changing the scaling, just rotating and translating
> 
> Is this still in reference to the CT-MR warp, or the MR-template warp?

In reference to the MR-CT registration which I don't think qualifies as a warp and so I think is misleading.

alexrockhill:
I think it should be split into two separate registration and warp steps, I can take a pass if you'd like

larsoner:
> I think it should be split into two separate registration and warp steps, I can take a pass if you'd like

To me it's all registration/warping/coreg, just with different levels of granularity (translation is one, rotation is another, affine-but-still-linear / shear is another, SDR is another). This is why I'm inclined toward a single function to do it...

larsoner:
I made the first step rigid and it still has invalid electrodes

alexrockhill:
> I made the first step rigid and it still has invalid electrodes

If it has the same `zoom=5` then it won't align properly, the CT to MR rigid can't be zoomed.

alexrockhill:
> > I think it should be split into two separate registration and warp steps, I can take a pass if you'd like
> 
> To me it's all registration/warping/coreg, just with different levels of granularity (translation is one, rotation is another, affine-but-still-linear / shear / scaling is another, SDR is another). This is why I'm inclined toward a single function to do it...

I just think when you're passing a CT into a function to align with an MR, you don't want it to be "warped", that would be bad, just registered

alexrockhill:
Maybe a good solution is to call it `mne.transforms.register` with a `sdr_warp=True` keyword argument to do the SDR as well all in one call. I think even having under in the `morph` module is misleading since you don't want to "morph" the CT either.

larsoner:
Maybe under `mne.source_space` is best actually, that's where a bunch of MRI wrangling things live

larsoner:
... I think you're right that `mne.transforms` is probably best, I'll push

larsoner:
Okay all electrodes get assigned now, feel free to look at the outputs to see if they make sense, and also the naming/structure etc.

alexrockhill:
Looks great!

I don't have write access but I would suggest having `zooms` be consistent with `niter` and either have `zooms_trans`, `zooms_affine` and `zooms_sdr` or have one `niter` dictionary. I would break up `zooms_trans` and `zooms_affine` as that will speed things up for the CT registration; it's nice to zoom the translation and not the rigid affine for how I would actually do it for a real analysis.

larsoner:
> . I would break up zooms_trans and zooms_affine as that will speed things up for the CT registration;

This suggests we might actually want the entire set `zooms_translation, zooms_rigid, zooms_affine, zooms_sdr`. I'm inclined to make both `niter` and `zooms` dicts for this reason, we'll end up with a ton of parameters otherwise. And maybe the logic should be that if a given `zooms` level is not provided, the zooms of the previous level are used. If `zooms['whatever'] = None` it means "use native". So for example `zooms=dict(translation=5, rigid=None)` would mean "use 5 for translation, then native from rigid onward"

alexrockhill:
I think also a keyword `warp` for whether to compute the SDR would be really helpful, to me `niters_sdr=()` is confusing

larsoner:
> I think also a keyword warp for whether to compute the SDR would be really helpful, to me niters_sdr=() is confusing

How about `niter=dict(affine=None, sdr=())` would mean "only do rigid" (i.e., use None as an alias for `()` meaning, "do not do it")?

alexrockhill:
> > . I would break up zooms_trans and zooms_affine as that will speed things up for the CT registration;
> 
> This suggests we might actually want the entire set `zooms_translation, zooms_rigid, zooms_affine, zooms_sdr`. I'm inclined to make both `niter` and `zooms` dicts for this reason, we'll end up with a ton of parameters otherwise. And maybe the logic should be that if a given `zooms` level is not provided, the zooms of the previous level are used. If `zooms['whatever'] = None` it means "use native". So for example `zooms=dict(translation=5, rigid=None)` would mean "use 5 for translation, then native from rigid onward"

That sounds great except that I wouldn't want it to carry on at a lower resolution. If the previous level used some downsampling, that doesn't mean the next level has to, in fact in the case of the CT to MR, you want to zoom and then not. On `None` I would just use the full resolution.

alexrockhill:
> > I think also a keyword warp for whether to compute the SDR would be really helpful, to me niters_sdr=() is confusing
> 
> How about `niter=dict(affine=None, sdr=())` would mean "only do rigid" (i.e., use None as an alias for `()` meaning, "do not do it")?

I think that's getting really confusing to me. Maybe a `steps` list would be good, with `['trans', 'rigid']` for CT to MR and `['trans', 'rigid', 'affine', 'sdr']` for the T1 to fsaverage.

larsoner:
> I think that's getting really confusing to me. Maybe a steps list would be good, with ['trans', 'rigid'] for CT to MR and ['trans', 'rigid', 'affine', 'sdr'] for the T1 to fsaverage.

That seems like a nice way to do it!

> That sounds great except that I wouldn't want it to carry on at a lower resolution. If the previous level used some downsampling, that doesn't mean the next level has to, in fact in the case of the CT to MR, you want to zoom and then not. On None I would just use the full resolution.

So the logic would be "native unless specified"? I can live with this, especially if we alias `zooms=5` to mean "use 5 for all steps". Even in the example here we use 6 then 4 for subject-to-fsaverage, I'll probably switch it to `zooms=5` in that case (and the results are still fine it seems).

alexrockhill:
> > I think that's getting really confusing to me. Maybe a steps list would be good, with ['trans', 'rigid'] for CT to MR and ['trans', 'rigid', 'affine', 'sdr'] for the T1 to fsaverage.
> 
> That seems like a nice way to do it!
> 
> > That sounds great except that I wouldn't want it to carry on at a lower resolution. If the previous level used some downsampling, that doesn't mean the next level has to, in fact in the case of the CT to MR, you want to zoom and then not. On None I would just use the full resolution.
> 
> So the logic would be "native unless specified"? I can live with this, especially if we alias `zooms=5` to mean "use 5 for all steps". Even in the example here we use 6 then 4 for subject-to-fsaverage, I'll probably switch it to `zooms=5` in that case (and the results are still fine it seems).

Sounds perfect, totally on the same page

larsoner:
Okay I'll get to it tomorrow, or if you want to give it a shot you can

alexrockhill:
I didn't finish and I need to fix the commit history but I pushed these in case you want to finish it tomorrow morning before I'm up, but I can also fix and finish tomorrow just a bit later, I'm on PST.

doc/changes/latest.inc
```diff
@@ -71,7 +71,11 @@ Enhancements
 
 - Show multiple colors and linestyles for excluded components with :class:`mne.Evoked` in :meth:`mne.preprocessing.ICA.plot_sources` (:gh:`9444` by `Martin Schulz`_)
 
-- Added tutorial for how to processes image (CT and MR) files in order to localize electrode contacts for intracranial recordings :ref:`tut-ieeg-localize` (:gh`9484` by `Alex Rockhill`_)
+- Add functions for aligning MRI and CT data `mne.transforms.compute_volume_registration` and `mne.transforms.apply_volume_registration` (:gh:`9503` by `Eric Larson`_ and `Alex Rockhill`_)
```
agramfort, line 74:
```suggestion
- Add functions for aligning MRI and CT data :func:`mne.transforms.compute_volume_registration` and :func:`mne.transforms.apply_volume_registration` (:gh:`9503` by `Eric Larson`_ and `Alex Rockhill`_)
```

mne/transforms.py
```diff
@@ -1521,3 +1522,155 @@ def _euler_to_quat(euler):
...
+        if step not in ('trans', 'rotate', 'scale', 'sdr'):
+            raise TypeError(f"``step`` must be one of ``'trans', "
+                            "'rotate', 'scale', 'sdr'`` "
+                            f"got {step} instead")
+        if validate_type is not None:
+            _validate_type(param[step], validate_type, f"{name}['{step}'']")
+
+
+@verbose
+def compute_volume_registration(moving, static, steps='warp', zooms=None,
```
agramfort, line 1539: I would call moving_img and static_img to suggest it's a nibabel Image

mne/viz/topomap.py
```diff
@@ -757,19 +758,60 @@ def plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
...
+    which correspond to the first and last colors in the colormap. This might
+    be undesired when vmin and vmax are not symmetrical around zero (or a value
+    that can be interpreted as some midpoint). For example, assume we want to
+    use the RdBu colormap (red to white to blue) for values ranging from -1 to
+    3, and 0 should be white. However, white corresponds to the midpoint in the
+    data by default, i.e. 1. Therefore, we use the following colormap
+    normalization ``cnorm`` and pass it as the the ``cnorm`` argument:
+
+        from matplotlib.colors import TwoSlopeNorm
+        cnorm = TwoSlopeNorm(vmin=-1, vcenter=0, vmax=3)
```
agramfort, line 789: all this is unrelated.

doc/changes/latest.inc
larsoner: We don't need this anymore (our default role is `:py:obj`) and at least some people have expressed a preference for not doing it

mne/transforms.py
larsoner: I had it that way at first but thought `moving` and `static` alone were clearer because it matches what dipy calls it in `affine_registration`:

https://dipy.org/documentation/1.4.1./reference/dipy.align/#affine-registration

So here I think being consistent with their naming is worth it. (Plus the name is shorter.) But I don't feel strongly so if you aren't convinced by the consistency argument I'll change it

mne/transforms.py
```diff
@@ -1524,30 +1524,51 @@ def _euler_to_quat(euler):
...
-        Can be None (default) to set ``niter_sdr=()`` automatically when
-        ``rigid=True``.
-    rigid : bool
-        If True, do a rigid alignment instead of a full affine transformation.
+    niter : dict | None
+        For each phase of the registration, ``niter`` is the number of
+        iterations per successive stage of optimization. The length of
+        ``niters`` defines the number of successive stages of optimization.
+        Default is ``(100, 100, 10)`` for the ``"trans"``, ``"rotate"`` and
+        ``"scale"`` steps and ``(5, 5, 3)`` for the ``"sdr"`` step. If a
```
larsoner, line 1568: @alexrockhill -1 on changing this name to scale. IIUC it's not just a scaling, it's a full affine, which includes shear

larsoner:
Working on this now @alexrockhill !

larsoner:
@alexrockhill @agramfort I tried to make the naming and wording as consistent as possible with `dipy`, which also made me realize we should deprecate `niter_affine` and `niter_sdr` in favor of `niter` dict. Feel free to have a look and see if it makes sense.

@alexrockhill please also confirm that the updated sEEG example looks okay once CircleCI completes! If everything is okay, I'll finish up the last couple of points in the TODO list (add tests, make SourceMorph store the complete `pipeline` and `niter` dict rather than the `niter_affine` and `niter_sdr` parameters)

alexrockhill:
Ok sorry just giving it a shot with the renaming, not trying to make things more confusing. I'll look things over a bit later today!

mne/transforms.py
alexrockhill: Sounds good, I was just throwing out a way to try and make it more accessible but I think you're right it's too simplified.

mne/morph.py
```diff
@@ -1011,95 +1032,173 @@ def _compute_r2(a, b):
...
+        types = (dict,) + types
+    _validate_type(zooms, types, 'zooms')
+    zooms = _handle_default('transform_zooms', zooms)
+    for key, val in zooms.items():
+        _check_option('zooms key', key, ('affine', 'sdr'))
+        if val is not None:
+            val = tuple(
+                float(x) for x in np.array(val, dtype=float).ravel())
+            _check_option(f'len(zooms[{repr(key)})', len(val), (1, 3))
+            if len(val) == 1:
```
alexrockhill, line 1085: Shouldn't this be before the check on the previous line that it is shape `(1, 3)`?

alexrockhill:
Looks great overall, that's what I was trying to do when I started the last commit but better :)

It seems a bit odd that all the affine code is in `morph.py` and the call is in transforms but is just a shallow wrapper. It might make a bit more sense to me to put all the affine code in `transforms.py` and then call a function from `morph.py` for the sdr but that's a bit of a nitpick.

I'm not sure the zooms still execute the way I would do it, the fastest and just-as-good-as-no-zooms way from my testing was to zoom on the translation and then not on the rigid registration. You can't do that because the zooms key is grouped together for those two steps under affine. It's a small consideration but it does make a difference of several minutes per subject. The translation step takes > 5 minutes on a macbook without `zooms` and 10 seconds with `zooms=5` and the output is functionally equivalent.

Just a thought, exposing all the pipeline steps as well as the shortcuts seems like it might be a bit overwhelming to users. I think the main thing is to be able to control the `zooms` and `niters` for the `translation`, `rigid`, `affine` and `sdr` steps. So I think it could be simplified, I know it's not much, by combining the `center_of_mass` and `translation` because those seek to do the same thing and I can't think of a use case when you wouldn't want to do both (they really don't work for the CT-MR on their own weirdly). I think the `affines` default option can also be omitted because it isn't a default use case. I think relying heavily on the word affine for the optimization is confusing because all the optimization optimize the affine just with different constraints on which parameters are allowed to vary. I agree that the way I had it before was over-simplified and confusing that it was inconsistent with `Dipy` so I do think this is much better and these are small points that are also a nitpicks.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -111,90 +113,38 @@ def plot_overlay(image, compare, title, thresh=None):
...
+# execute because it's slow) the call:
 #
 # .. code-block:: python
 #
-#     # normalize intensities
-#     mri_to = T1.get_fdata().copy()
-#     mri_to /= mri_to.max()
-#     ct_from = CT_orig.get_fdata().copy()
-#     ct_from /= ct_from.max()
+#    affine, _ = compute_volume_registration(CT_orig, T1, pipeline='rigid')
```
alexrockhill, line 131: This should be `rigids` and it doesn't give a good registration. It's not terribly off but it's not good enough. I'm not sure exactly why but I would guess from my testing that the zooms for the translation actually helped. Maybe if it's translated with a lot of optimization it can get stuck in a local minimum, not sure just thinking out loud.

alexrockhill: Also this took more than 10 minutes on my macbook, I think because of the translation and that it didn't work properly so then the rigid optimization took longer.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -377,46 +293,51 @@ def plot_overlay(image, compare, title, thresh=None):
...
+# Apply the mapping
+elec_image = nib.spatialimages.SpatialImage(elec_image, subject_brain.affine)
+del subject_brain, CT_aligned, CT_data  # not used anymore
+
+###############################################################################
+# Warp and plot the result
+_, warped_elec_image = apply_volume_registration(
+    elec_image, template_brain, pre_affine, sdr_morph, interpolation='nearest')
+fig = nilearn.plotting.plot_glass_brain(warped_elec_image)
+fig = fig.axes['x'].ax.figure
```
alexrockhill, line 329: Do you need a different figure, can't you just use the one that was returned?

alexrockhill:
Also, I get a negative `R**2` for the center of mass. (And the `R**2` values are very low but I think that's related to the zooms issue).
```
>>> affine, _ = compute_volume_registration(CT_orig, T1, pipeline='rigids')
Computing rigid registration ...
Reslicing to zooms=None for center_of_mass ...
Computing center of mass:
    Translation:  110.6 mm
    Translation:  110.6 mm
    R²:           -23.3%
Optimizing translation:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:  184.4 mm
    R²:            17.2%
Optimizing rigid:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:  184.4 mm
    Rotation:       7.7°
    R²:            16.4%
```

alexrockhill:
Let me know if you'd appreciate me taking a swing at any of these things, looks like a great improvement especially to the API of the example.

mne/morph.py
larsoner: The idea is that it should be one or three values, and if it's one, we promote it to three. In principle it can be done in either order, but to me this logically makes sense and follows our usual pattern: first validate that the user gave a correct input, then for our own later convenience unify the input so we know what it is (always a tuple of three float, or None).

tutorials/clinical/10_ieeg_localize.py
larsoner: The variable naming here is not great, it doesn't actually return a figure but some custom object. I'll update the var names to make it clearer

tutorials/clinical/10_ieeg_localize.py
larsoner: > This should be rigids and it doesn't give a good registration. It's not terribly off but it's not good enough. I'm not sure exactly why but I would guess from my testing that the zooms for the translation actually helped. Maybe if it's translated with a lot of optimization it can get stuck in a local minimum, not sure just thinking out loud.

Can you push a line that does work after I make a commit where `zooms` can be specified at each level?

larsoner:
> So I think it could be simplified, I know it's not much, by combining the center_of_mass and translation because those seek to do the same thing and I can't think of a use case when you wouldn't want to do both

Agreed!

> I think the affines default option can also be omitted because it isn't a default use case. I think relying heavily on the word affine for the optimization is confusing because all the optimization optimize the affine just with different constraints on which parameters are allowed to vary.

Agreed it's not useful for our use cases here. However, I think of "computing a full affine" as the default mode in MRI -- I think it's what `flirt` and most other MRI tools will do by default -- so it seems like we should provide a shortcut for it rather than tell people to do `pipeline=('translation', 'rigid', 'affine')` (and the maintenance overhead is minimal!).

> The translation step takes > 5 minutes on a macbook without zooms and 10 seconds with zooms=5 and the output is functionally equivalent.

I'll make `zooms` more fine-grained, agreed this is not good

> It might make a bit more sense to me to put all the affine code in transforms.py and then call a function from morph.py for the sdr but that's a bit of a nitpick.

I agree (both that it makes more sense there and that it's a minor issue) but `blame` gets worse/harder the more we move code around so I'd rather just leave it

> Also, I get a negative R**2 for the center of mass.

The way we compute the variance explained / R**2 this can happen. I think the negative values are useful (they indicate a very bad match) but agree it's a bit of a misnomer. The computation is really `100 * (1 - residual_sum_sq / original_sum_sq)` so it's not exactly R**2, maybe we should figure out a better name someday... but probably in another PR since it's not specific to these changes.

> (And the R**2 values are very low but I think that's related to the zooms issue).

IIRC they only go up to ~20% with different zooms, but feel free to push a change and we'll see once I fix the code!

larsoner:
Okay feel free to look and push @alexrockhill

alexrockhill:
I just pushed a few small changes after running through the example on the notebook side but I think there's a few fixes needed on the backend.

1) The affine registration for the CT takes forever, doesn't work and doesn't print any output on the progress (prints all info when done)
2) The SDR with the zoom decimation isn't better than the talairach transform, I think we need to do full resolution or go back to loading the results (go back seems to make more sense) although the API is so much cleaner it's fantastic
3) It seems to make more sense to only return the SDR when that it one of the steps as in itc in tfr https://mne.tools/stable/generated/mne.time_frequency.tfr_morlet.html
4) The electrode maps look a little wonky and inconsistent as far as size, I'll see if I can get that fixed

I'll work on those changes now.

larsoner:
> The affine registration for the CT takes forever, doesn't work and doesn't print any output on the progress (prints all info when done)

Can you change the commented-out code to make the call work?

> The SDR with the zoom decimation isn't better than the talairach transform, I think we need to do full resolution

Can you see if `zooms=dict(affine=5, sdr=2)` for example works? If not then we could load the SDR (somehow). Or really in the example we can just say that we've lost some accuracy for speed.

> It seems to make more sense to only return the SDR when that it one of the steps as in itc in tfr 

I'd actually rather not change the set of returns based on the inputs. With the current code no changes are needed beyond `pipeline` to change between affine / sdr / affine+SDR. It's just that you get `None` for some of them. If you change the returns then it'll require lots of conditionals in your code.

alexrockhill:
Ok @larsoner, not really sure where to go from here, I reverted a lot of the changes you made because they didn't work unfortunately. I don't understand `Dipy` well enough to do what you were trying to do and call each function individually instead of using the `affine_registration` wrapper and I'm not sure I see the point of doing that. I tested it as it was and I ended up just reverting most of the changes in order to get things working. I think what happened is you had a lot of the same ideas I had when I was writing the tutorial because I tried to do it that way but (again not really fully understanding `Dipy`) I couldn't get it to work that way. This version works and it wraps the lengthy function calls that are shared with the source morphing but I also did a lot of the things that were different than the way you were thinking. We can either revert and use some of this last commit to get things working/refactor or we can go from this back towards what you were thinking. Feel free to just revert, I was just making changes until I could get it working and it took until I reverted all the way back pretty much.

larsoner:
Can you indicate what outputs look better with the current code rather than the code I had?

Also it looks like there was a lot of code movement and refactoring that makes the diff a bit harder to follow (more than reverting, like moving things from morph to transforms). Once you let me know what looks better now compared to my last commit, I'll probably try to force push a much smaller commit that makes things work.

alexrockhill:
> Can you indicate what outputs look better with the current code rather than the code I had?

The CT alignment just didn't work the way it was and the SDR was way off with `zooms=5` and `zooms=2` as it was before.

> Also it looks like there was a lot of code movement and refactoring that makes the diff a bit harder to follow (more than reverting, like moving things from morph to transforms). Once you let me know what looks better now compared to my last commit, I'll probably try to force push a much smaller commit that makes things work.

Yeah sorry I moved things around there too so that they were in the same place.

larsoner:
So zooms=None is important for SDR to fsaverage. What did you change that fixed the CT to MR? I had that commented out with translation zooms 5 and all others none I thought. Was the code buggy or did you change the parameters?

alexrockhill:
> So zooms=None is important for SDR to fsaverage. What did you change that fixed the CT to MR? I had that commented out with translation zooms 5 and all others none I thought. Was the code buggy or did you change the parameters?

I changed it so that the `moving` image was changed every step. Before, the `pre_affine` was optimized but not applied correctly. It looked like the same things should be happening but the optimization wasn't working so I changed it back.

alexrockhill:
Here's the output and the registration using the last commit you pushed. It looks like it should work but it doesn't and I wasn't able to figure out why and didn't dig into the `Dipy`.

Screen Shot 2021-06-28 at 4 51 13 PM

```
>>> affine_map, _ = compute_volume_registration(CT_orig, T1, pipeline='rigids', zooms=dict(translation=5))
Computing rigid registration ...
Reslicing to zooms=(5.0, 5.0, 5.0) for translation ...
Optimizing translation:
    Initializing using center of mass:
    Translation:  114.7 mm
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:  183.0 mm
    R²:            20.0%
Reslicing to zooms=None for rigid ...
Optimizing rigid:

    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:  183.0 mm
    Rotation:       6.2°
    R²:            17.2%
>>> 
>>> CT_aligned = mne.transforms.apply_volume_registration(CT_orig, T1, affine_map)
Applying affine registration ...
[done]
>>> plot_overlay(T1, CT_aligned, 'Aligned CT Overlaid on T1', thresh=0.95)
```

larsoner:
> I changed it so that the moving image was changed every step

It shouldn't really matter if you change it (other than you linearly interpolate multiple times which is not great) every step, except that by doing so you do throw away some information at the edge of the image each time (e.g., by cutting it out once it's resampled outside the domain during the translation).

I suspect that right thing to do / best practice here is properly normalize the CT better using some thresholds before computing the transformation. We normalize by the max internally, but I suspect we might want to also subtract some lower threshold then set everything less than zero to zero (and then let the internal max normalization do its thing). Then resample-each-step or resample-at-end should be equivalent, other than the at-end being slightly better by only interpolating once.

alexrockhill:
Hmmm okay, looking into it more, it looks like the output for the version that makes the `affine` "loaded in"/typed out outputs this

```
>>> affine = compute_volume_registration(CT_orig, T1, pipeline='rigids')
Reslicing to zooms=(5.0, 5.0, 5.0)...
Optimizing translation:
    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:  183.0 mm
    R²:            17.4%
Reslicing to zooms=None...
Optimizing rigid:


CT_aligned = mne.transforms.apply_volume_registration(CT_orig, T1, affine)
plot_overlay(T1, CT_aligned, 'Aligned CT Overlaid on T1', thresh=0.95)


    Optimizing level 2 [max iter: 100]
    Optimizing level 1 [max iter: 100]
    Optimizing level 0 [max iter: 10]
    Translation:    3.4 mm
    Rotation:       8.9°
    R²:            16.6%
```

I'm not sure but I think it maybe have something to do with extra steps in the `affine_registration` `Dipy` function.

alexrockhill:
The R**2 actually goes down when it's working too which seems a bit confusing.

alexrockhill:
> > I changed it so that the moving image was changed every step
> 
> It shouldn't really matter if you change it (other than you linearly interpolate multiple times which is not great) every step, except that by doing so you do throw away some information at the edge of the image each time (e.g., by cutting it out once it's resampled outside the domain during the translation).

Well so what happens in the version I pushed is that the original `moving` is warped using the updated affine to compute translation, rotation and R**2 but the underlying images are modified for the computations from the original in one step...

mne/transforms.py
```diff
@@ -1521,3 +1523,225 @@ def _euler_to_quat(euler):
...
+        if step in ('center_of_mass', 'translation', 'rigid'):
+            dist = np.linalg.norm(reg_affine[:3, 3])
+            angle = np.rad2deg(_angle_between_quats(
+                np.zeros(3), rot_to_quat(reg_affine[:3, :3])))
+            logger.info(f'    Translation: {dist:6.1f} mm')
+            if step == 'rigid':
+                logger.info(f'    Rotation:    {angle:6.1f}°')
+        r2 = _compute_r2(_get_img_fdata(static), _get_img_fdata(moving))
+        logger.info(f'    R²:          {r2:6.1f}%')
+    return reg_affine
```
larsoner, line 1642: This is going to be incorrect I think, it needs to combine all affines computed at each step. (This arises because you now resample at each step, rather than always working from the starting moving image.)

alexrockhill: The way it resamples combines the affines ASAIK

alexrockhill: To my knowledge this is because each step is inclusive of all the previous steps so when you do the rigid registration on the image that has already been translated, the translation affine is included

larsoner: If you take a good T1 image, translate it by 10 mm (you can just set its `affine` to be something different), then apply this function is the output affine a translation by 10 mm? That is what you should get be for both `pipeline=('translate',)` and `pipeline=('translate', 'rigid')`, but I suspect that with the current code if it won't be (because you apply the translation in the first step, so no adjustment is necessary in the next). I also don't 100% know it would have worked on my commit, but I suspect it would have. I can check tomorrow unless you still have time to check today. It would be good to manually check the R**2 values of the actually transformed data using these pipelines, too. And the `affine_registration` of `dipy` can be a benchmark, it should give 99.9% or something...

But I could have this totally backward, in which case the old code was probably wrong! But that is also how `morph.py` does things, so we must have some bugs on main, too...

alexrockhill: You were right about this, the affines needed to be combined, I'm not sure why the example worked previously though that's odd, looking into it.

larsoner:
Okay @alexrockhill feel free to have a look and see if you're happy with the code, here is the updated example:

https://30168-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/clinical/10_ieeg_localize.html

larsoner:
Windows is having a bad day because of the VTK 9.0.2 release, can be ignored here as I'm working on it in #9512

alexrockhill:
> Okay @alexrockhill feel free to have a look and see if you're happy with the code, here is the updated example:
> 
> https://30168-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/clinical/10_ieeg_localize.html

Still looks off... I would rather have a maximally accurate example that has a lot of code that's not abstracted than one that doesn't work quite right. I think the existing morph functions may not be compatible with the what is on main now for the tutorial. I think first priority if you want to go this route of refactoring is getting a version that works and then maybe in a followup we can get the SDR morph from the volume source estimate working using the same code. I really would rather not spend a ton of time debugging the volume source estimate code so I think it's best to part ways. I'll take a pass now.

larsoner:
> Still looks off...

Can you clarify how/where? I'm assuming you're talking about subject->fsaverage and not CT-MRI. For "to fsaverage", those temporal electrodes are now nicely on/around the actual lobe, which is what I was always worried about, so to me it looks okay. Is there an image above somewhere (or in some artifact) that looks better?

> I'll take a pass now.

Can you see if just `zooms=None` in the SDR pass makes things look okay to you? If so I can go back to saving the warped MRIs.

> I really would rather not spend a ton of time debugging the volume source estimate code so I think it's best to part ways.

So maybe as an intermediate, if zooms=None does not work, we can keep the MR-CT alignment that is here currently since it works, then put back your custom dipy code for the subject->fsaverage alignment and save those warped electrode and brain images?

alexrockhill:
> > Still looks off...
> 
> Can you clarify how/where? I'm assuming you're talking about subject->fsaverage and not CT-MRI. For "to fsaverage", those temporal electrodes are now nicely on/around the actual lobe, which is what I was always worried about, so to me it looks okay. Is there an image above somewhere (or in some artifact) that looks better?
> 
> > I'll take a pass now.
> 
> Can you see if just `zooms=None` in the SDR pass makes things look okay to you? If so I can go back to saving the warped MRIs.
> 
> > I really would rather not spend a ton of time debugging the volume source estimate code so I think it's best to part ways.
> 
> So maybe as an intermediate, if zooms=None does not work, we can keep the MR-CT alignment that is here currently since it works, then put back your custom dipy code for the subject->fsaverage alignment and save those warped electrode and brain images?

The CT-MR is way off still, that's the main thing. The SDR looks pretty good.

larsoner:
> The CT-MR is way off still, that's the main thing.

So this is what we have now:

sphx_glr_10_ieeg_localize_002

And this is what we have in `main`:

sphx_glr_10_ieeg_localize_002_main

I think it only differs by 1.2mm and 1.9 degrees from what you had before. I have a hard time seeing meaningful differences here, and if I didn't know which is which I'm not sure which I'd say was better, but I am no expert here!

As much as I'm inclined to keep it, indeed we shouldn't keep something if you think it is "way off" even if I can't see it. I thought you had said before that the MR-CT was now okay, but since it's not I agree perhaps the easiest thing to do is just to revert to the old affine and commented out code for the MR-CT step, and revisit sometime later how to use our standard pipeline to get an affine of equivalent quality.

alexrockhill:
> > The CT-MR is way off still, that's the main thing.
> 
> So this is what we have now:
> 
> sphx_glr_10_ieeg_localize_002
> 
> And this is what we have in `main`:
> 
> sphx_glr_10_ieeg_localize_002_main
> 
> I think it only differs by 1.2mm and 1.9 degrees from what you had before. I have a hard time seeing meaningful differences here, and if I didn't know which is which I'm not sure which I'd say was better, but I am no expert here!
> 
> As much as I'm inclined to keep it, indeed we shouldn't keep something if you think it is "way off" even if I can't see it. I thought you had said before that the MR-CT was now okay, but since it's not I agree perhaps the easiest thing to do is just to revert to the old affine and commented out code for the MR-CT step, and revisit sometime later how to use our standard pipeline to get an affine of equivalent quality.

Way off is a relative term and I think you have to spend some time aligning these by hand to appreciate when the fit is right and when it's not. 1.2 mm and 1.9 degrees isn't a whole lot but those are clearly how much it's wrong and to me, that's too much. You can align by eye closer than that, the whole point of this is that it should be as precise as possible with the alignment, better than what can be done by human eye. I could fix the top image to be much closer by hand.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -111,89 +110,43 @@ def plot_overlay(image, compare, title, thresh=None):
...
-
-
-###############################################################################
-# The previous section takes several minutes to execute so the results are
-# presented here pre-computed for convenience.
-
-alignment_affine = np.array([
-    [0.99235816, -0.03412124, 0.11857915, -133.22262329],
-    [0.04601133, 0.99402046, -0.09902669, -97.64542095],
-    [-0.11449119, 0.10372593, 0.98799428, -84.39915646],
```
larsoner, line 189: @alexrockhill okay if I just revert the text removal here, then we're good to merge?

alexrockhill: That's the correct alignment but the commented out code should return that matrix and it doesn't right?

larsoner: By revert the text I mean both this affine and the old commented code, which I assume is what you used to generate it

alexrockhill:
Ok I just pushed a different pull request to split this up. First, I think a version that works with the tutorial should be merged which is the PR I just pushed. Then we can try and refactor the volume source estimate code to use the new transforms. I think combining the two steps is causing too much of an issue.

larsoner:
Closing for #9515

PR #9505, opened 2021-06-24
Fix plotting of events when there are overlapping epochs

### LINKED ISSUE #9502: plotting overlapping epochs: Events not properly aligned in plot?
I have to create overlapping epochs from scratch using [EpochsArray](https://mne.tools/dev/generated/mne.EpochsArray.html#mne.EpochsArray). I designed my events, tmin, and other parameters so that the main event for each epoch (the t0 event) is exactly in the middle of a 2 second timespan (-1s to 1s). However when plotting via `epochs.plot`, the events are **not** in the middle.

This plotting appearance is fixed when I subset my epochs to only contain those epochs that are not overlapping.

The following MWE is modified from [the docs](https://mne.tools/dev/auto_tutorials/simulation/10_array_objs.html#creating-mne-python-data-structures-from-scratch).

The only change from the docs example is to go from non-overlapping epochs: `np.arange(0, 1000, sampling_freq)` 
to overlapping epochs: `np.linspace(0, 200, 5, dtype=int)`.

```python
import mne
import numpy as np

sampling_freq = 200  # in Hertz

info = mne.create_info(ch_names=['10 Hz sine', '5 Hz cosine'],
                       ch_types=['misc'] * 2,
                       sfreq=sampling_freq)

times = np.linspace(0, 1, sampling_freq, endpoint=False)
sine = np.sin(20 * np.pi * times)
cosine = np.cos(10 * np.pi * times)

data = np.array([[0.2 * sine, 1.0 * cosine],
                 [0.4 * sine, 0.8 * cosine],
                 [0.6 * sine, 0.6 * cosine],
                 [0.8 * sine, 0.4 * cosine],
                 [1.0 * sine, 0.2 * cosine]])


events = np.column_stack((np.linspace(0, 200, 5, dtype=int),  # NOTE: used to be np.arange(0, 1000, sampling_freq),
                          np.zeros(5, dtype=int),
                          np.array([1, 2, 1, 2, 1])))
event_dict = dict(condition_A=1, condition_B=2)
simulated_epochs = mne.EpochsArray(data, info, tmin=-0.5, events=events,
                                   event_id=event_dict)

simulated_epochs.plot(picks='misc', events=events, event_id=event_dict)

```

Take a look at the resulting plot. It's hard to see in the following screenshot, but the events are NOT centered in the middle of the epoch -> rather, they are clustered at the end of the plot:

image

When subsetting these epochs to only contain non-overlapping epochs, the issue is solved and the plot shows the expected alignment of events in the middle of epochs.

```python
simulated_epochs[np.array((0, 4))].plot(picks='misc', events=events[np.array((0, 4))], event_id=event_dict)

```

image

**Is this somehow expected? Or is it a bug?**

drammock:
closes #9502 

cc @sappelhoff

### DIFF
--- a/mne/viz/epochs.py
+++ b/mne/viz/epochs.py
@@ -795,16 +795,24 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20, n_channels=20,
     # events
     if events is not None:
         event_nums = events[:, 2]
-        boundary_samps = epochs.events[1:, 0] - epochs.time_as_index(0)
-        event_ixs = np.searchsorted(boundary_samps, events[:, 0], side='right')
-        event_samp_offsets = (events[:, 0] - epochs.events[:, 0][event_ixs]
-                              + epochs.time_as_index(0))
-        event_times = ((np.arange(len(epochs)) * len(epochs.times))[event_ixs]
-                       + event_samp_offsets) / sfreq
-        # don't show events that are beyond the range of their epoch
-        mask = event_samp_offsets <= len(epochs.times)
-        event_times = event_times[mask]
-        event_nums = event_nums[mask]
+        event_samps = events[:, 0]
+        epoch_n_samps = len(epochs.times)
+        # handle overlapping epochs (each event may show up in multiple places)
+        boundaries = (epochs.events[:, [0]] + np.array([-1, 1])
+                      * epochs.time_as_index(0))
+        in_bounds = np.logical_and(boundaries[:, [0]] <= event_samps,
+                                   event_samps < boundaries[:, [1]])
+        event_ixs = [np.nonzero(a)[0] for a in in_bounds.T]
+        event_times = list()
+        event_numbers = list()
+        for samp, num, _ixs in zip(event_samps, event_nums, event_ixs):
+            relevant_epoch_events = epochs.events[:, 0][_ixs]
+            offsets = samp - relevant_epoch_events + epochs.time_as_index(0)
+            this_event_times = (_ixs * epoch_n_samps + offsets) / sfreq
+            event_times.extend(this_event_times)
+            event_numbers.extend([num] * len(_ixs))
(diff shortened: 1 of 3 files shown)


agramfort (approved):

mne/viz/epochs.py
```diff
@@ -795,16 +795,28 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20, n_channels=20,
...
+                                   event_samps < boundaries[:, [1]])
+        event_ixs = [np.nonzero(a)[0] for a in in_bounds.T]
+        warned = False
+        event_times = list()
+        event_numbers = list()
+        for samp, num, _ixs in zip(event_samps, event_nums, event_ixs):
+            relevant_epoch_events = epochs.events[:, 0][_ixs]
+            if len(relevant_epoch_events) > 1 and not warned:
+                logger.info('You seem to have overlapping epochs. Some event '
+                            'lines may be duplicated in the plot.')
```
sappelhoff, line 813:
```suggestion
                logger.info('You seem to have overlapping epochs. Some event '
                            'lines may be duplicated in the plot.')
                warned = True                            
```

sappelhoff (commented):
I tested this on my real data and it works like a charm, everything is as expected now.

I think you forgot to set the `warned` variable though, because it's printing one info per plotted epoch, which is overkill. See my code suggestion.

Thanks!

mne/viz/epochs.py
```diff
@@ -795,16 +795,29 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20, n_channels=20,
...
+                warned = True
```
drammock, line 814:
```suggestion
                warned = True
```

drammock:
all green, ready for review/merge @agramfort @larsoner

agramfort:
Thx @drammock

PR #9513, opened 2021-06-30
FIX: Fix simulate_evoked and apply_forward

larsoner:


mne/simulation/evoked.py
```diff
@@ -33,7 +33,7 @@ def simulate_evoked(fwd, stc, info, cov, nave=30, iir_filter=None,
         The source time courses.
     info : dict
         Measurement info to generate the evoked.
-    cov : Covariance object
+    cov : Covariance object | None
         The noise covariance.
```
agramfort, line 37: If None what is it that we do?

agramfort (approved):
if you can just update the docstring it would be great !

thx @larsoner !

agramfort:
cc @LaetitiaG it should fix your pb.

LaetitiaG:
Indeed, it works fine now! thanks a lot!

agramfort (approved):
@larsoner I pushed here. Merge if happy

mne/simulation/evoked.py
```diff
@@ -70,11 +70,14 @@ def simulate_evoked(fwd, stc, info, cov, nave=30, iir_filter=None,
     .. versionadded:: 0.10.0
     """
     evoked = apply_forward(fwd, stc, info, use_cps=use_cps)
+    if cov is None:
+        return evoked
+
     if nave < np.inf:
         noise = _simulate_noise_evoked(evoked, cov, iir_filter, random_state)
         evoked.data += noise.data / math.sqrt(nave)
         evoked.nave = np.int64(nave)
```
larsoner, line 79: This is incorrect as you now no longer set `nave` in the `cov=None` case

larsoner: ... oh I guess this was always the case when `nave=np.inf`, but I'll fix it here

larsoner: argh I see now why you did it this way. I'm okay with it!

PR #9514, opened 2021-06-30
use Polynomial.fit for poly fitting

nordme:
@larsoner  np.polyfit was deprecated when I tried to use the realign function to sync our eeg / meg data. This is an update to use the Polynomial.fit method suggested by Numpy in the deprecation warning. Take a look and see what you think.

### DIFF
--- a/mne/preprocessing/realign.py
+++ b/mne/preprocessing/realign.py
@@ -5,6 +5,7 @@
 
 import numpy as np
 
+from numpy.polynomial.polynomial import Polynomial
 from ..io import BaseRaw
 from ..utils import _validate_type, warn, logger, verbose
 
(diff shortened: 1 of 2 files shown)


larsoner:
LGTM, can you add a `latest.inc` entry in the enhancements section mentioning this code update?

mne/preprocessing/realign.py
```diff
@@ -62,7 +63,11 @@ def realign_raw(raw, other, t_raw, t_other, verbose=None):
         warn('Fewer than 20 times passed, results may be unreliable')
 
     # 1. Compute correction factors
-    coef = np.polyfit(t_other, t_raw, deg=1)
+    poly = Polynomial.fit(x=t_other, y=t_raw, deg=1)
+    converted = poly.convert(domain=(-1, 1))
+    [zero_ord, first_ord] = converted.coef
```
drammock, line 68: just a sanity check here, because I couldn't easily tell from the numpy docs:  is this order (zero_ord, first_ord) correct?  I notice below you're replacing the old `coef[0]` with `first_ord` and the old `coef[1]` with `zero_ord`.

drammock: never mind, I found the answer in the transition guide.  before I was looking only at the docs of the new class, which don't tell you the coef order.

nordme:
Yup, I will add a latest.inc entry. Q: I see the mne-tools.mne-python check failed because of timing out, but from the error message I don't understand why it took so long. What's up with that?

larsoner:
That build times out but that should be fixed by  #9518. If you rebase or merge your branch with upstream/main it should be fixed

mne/preprocessing/realign.py
```diff
@@ -5,6 +5,7 @@
 
 import numpy as np
 
+from numpy.polynomial.polynomial import Polynomial
```
agramfort, line 8:
```suggestion
from numpy.polynomial.polynomial import Polynomial

```

agramfort (approved):

larsoner:
Thanks @nordme !

Issue #8833, opened 2021-02-03
ENH: New interface for Coregistration


GuillaumeFavelier:
This PR summarizes the ideas to implement for `coreg`:

<details open>

<summary><b>Milestone 1.0</b></summary>

- [ ] increase fiducial picking performance (https://github.com/mne-tools/mne-python/pull/9689#issuecomment-951910957)
- [ ] add support for the `guess_mri_subject` parameter 
- [ ] add support for the `project_eeg` parameter
- [ ] add support for the `scale_by_distance` parameter
- [ ] add support for the `mark_inside` parameter
- [ ] add support for the `scale` parameter
- [ ] add support for the `head_inside` parameter
- [ ] find a better name for the "Digitization Source" section (which could be confusing)
- [ ] make the sections collapsible *(ipywidgets)*
- [ ] add units to UI (Omit more explicit message)
- [ ] console idea to add to the UI whatever is printed in the terminal
- [ ] add tooltips as much as possible on buttons
- [ ] add legend to plot alignment / coreg to see what colored points are etc.
- [ ] display dist estimation in mm.
- [ ] 📁 icon for file button widgets (instead of "Load")

</details>

<details>
<summary><b>Done</b></summary>

- [x] Update `Coregistration()` to require parameters (https://mne.discourse.group/t/mne-coreg-based-on-ctf-data/3022/4)
- [x] Decouple GUI code from model (suggested in https://github.com/mne-tools/mne-python/pull/6693#issuecomment-542364434, done in https://github.com/mne-tools/mne-python/pull/9516)
- [ ] Build a coreg GUI using `pyvista` (work in progress in https://github.com/mne-tools/mne-python/pull/9689)

</details>

<details open>
<summary><b>Bugs</b></summary>

- [ ] inconsistent behaviour of checkbox layout for ipywidgets

</details>

hbk008:
I am not able to get past this: 

from mne.coreg import Coregistration

it throws error: cannot import name 'Coregistration' from 'mne.coreg' 

The other function from mne.coreg work well like getting fiducials from fiducials = mne.coreg.get_mni_fiducials(subject, subjects_dir=subjects_dir)
but this import Coregistration from mne.coreg fails for some reason. 
Can someone help me resolve this?

hoechenberger:
> from mne.coreg import Coregistration

To start up the coregistration GUI, you need to call

```python
mne.gui.coregistation()
```

hbk008:
[...]

PR #9516, opened 2021-07-01
MRG,ENH: Refactor coreg/traits classes

GuillaumeFavelier:
This PR refactors some traits-dependant classes out of GUI code. The goal is to be able to perform coregistration without the need for an interface.

Getting rid of `traits` means reproducing the dependency relations between properties/methods. It's still a draft. 
It's an item of https://github.com/mne-tools/mne-python/issues/8833


GuillaumeFavelier:
Here is the modified version of [jhouck's work](https://github.com/mne-tools/mne-python/pull/6693/files) (that I use as baseline) using the new `Coregistration` class:

For now, `data_path` is still used but the goal is to require `info` instead. I'll modify that next.

Among the more pressing issues, I have to continue refactor the class to remove the following from `gui` somehow:

```py
class Coregistration(object):
    def __init__(self, data_path, subject, subjects_dir):
        from .gui._file_traits import DigSource
        from .gui._fiducials_gui import MRIHeadWithFiducialsModel
...
```

<details>

```py
"""
=========================================
Use automated approach to co-registration
=========================================
This example shows how to use the get_mni_fiducials routine and the
coregistration GUI functions to perform an automated MEG-MRI co-registration.
.. warning:: The quality of the co-registration depends heavily upon the
             quality of the head shape collected during subject prepration and
             the quality of your T1-weighted MRI. Use with caution and check
             the co-registration error.
"""

# License: BSD (3-clause)

import os.path as op
import numpy as np
import mne
from mne.coreg import get_mni_fiducials, Coregistration
from mne.surface import dig_mri_distances
from mne.io import write_fiducials
from mne.io.constants import FIFF

subject = 'sample'
fname_raw = op.join(data_path, 'MEG', subject, subject + '_audvis_raw.fif')
fname_fids = op.join(subjects_dir, subject, 'bem', subject + '-fiducials.fif')
fname_trans = op.join(data_path, 'MEG', subject,
                      subject + 'audvis_raw_auto-trans.fif')
src = mne.read_source_spaces(op.join(subjects_dir, subject, 'bem',
                                     'sample-oct-6-src.fif'))

fids_mri = get_mni_fiducials(subject, subjects_dir)

# Save mri fiducials. This is mandatory. as fit_fiducials uses this file

write_fiducials(fname_fids, fids_mri, coord_frame=FIFF.FIFFV_COORD_MRI)
print('\nSaving estimated fiducials to %s \n' % fname_fids)

coreg = Coregistration(data_path, subject, subjects_dir)
coreg.fit_fiducials()
coreg.fit_icp(iterations=6, nasion_weight=2.)
coreg.omit_hsp_points(distance=5. / 1000)
coreg.fit_icp(iterations=20, nasion_weight=10.)
coreg.save_trans(fname=fname_trans)
errs_icp = coreg.point_distance()

raw = mne.io.Raw(fname_raw)
errs_nearest = dig_mri_distances(raw.info, fname_trans, subject, subjects_dir)

fig = mne.viz.plot_alignment(raw.info, trans=fname_trans, subject=subject,
                             subjects_dir=subjects_dir, surfaces='head-dense',
                             dig=True, eeg=[], meg='sensors',
                             coord_frame='meg')
mne.viz.set_3d_view(fig, 45, 90, distance=0.6, focalpoint=(0., 0., 0.))

print('Median distance from digitized points to head surface is %.3f mm'
      % np.median(errs_icp * 1000))
print('''Median distance from digitized points to head surface using nearest
neighbor is %.3f mm''' % np.median(errs_nearest * 1000))
```

</details>

GuillaumeFavelier:
The latest version of the testing example using `info`:

<details>

```py
"""
=========================================
Use automated approach to co-registration
=========================================
This example shows how to use the get_mni_fiducials routine and the
coregistration GUI functions to perform an automated MEG-MRI co-registration.
.. warning:: The quality of the co-registration depends heavily upon the
             quality of the head shape collected during subject prepration and
             the quality of your T1-weighted MRI. Use with caution and check
             the co-registration error.
"""

# License: BSD (3-clause)

import os.path as op
import numpy as np
import mne
from mne.coreg import get_mni_fiducials, Coregistration
from mne.surface import dig_mri_distances
from mne.io import write_fiducials, read_info
from mne.io.constants import FIFF

subject = 'sample'
fname_raw = op.join(data_path, 'MEG', subject, subject + '_audvis_raw.fif')
fname_fids = op.join(subjects_dir, subject, 'bem', subject + '-fiducials.fif')
fname_trans = op.join(data_path, 'MEG', subject,
                      subject + 'audvis_raw_auto-trans.fif')
src = mne.read_source_spaces(op.join(subjects_dir, subject, 'bem',
                                     'sample-oct-6-src.fif'))

fids_mri = get_mni_fiducials(subject, subjects_dir)

# Save mri fiducials. This is mandatory. as fit_fiducials uses this file

write_fiducials(fname_fids, fids_mri, coord_frame=FIFF.FIFFV_COORD_MRI)
print('\nSaving estimated fiducials to %s \n' % fname_fids)

info = read_info(fname_raw)
coreg = Coregistration(info, subject, subjects_dir)
coreg.fit_fiducials()
coreg.fit_icp(iterations=6, nasion_weight=2.)
coreg.omit_hsp_points(distance=5. / 1000)
coreg.fit_icp(iterations=20, nasion_weight=10.)
coreg.save_trans(fname=fname_trans)
errs_icp = coreg.point_distance()

raw = mne.io.Raw(fname_raw)
errs_nearest = dig_mri_distances(raw.info, fname_trans, subject, subjects_dir)

fig = mne.viz.plot_alignment(raw.info, trans=fname_trans, subject=subject,
                             subjects_dir=subjects_dir,
                             surfaces=dict(head=0.4),
                             dig=True, eeg=[], meg=False,
                             coord_frame='meg')
mne.viz.set_3d_view(fig, 45, 90, distance=0.6, focalpoint=(0., 0., 0.))

print('Median distance from digitized points to head surface is %.3f mm'
      % np.median(errs_icp * 1000))
print('''Median distance from digitized points to head surface using nearest
neighbor is %.3f mm''' % np.median(errs_nearest * 1000))
```

</details>

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
                 F_mri_ras, 'ras', 'ras'),
             F_ras_mni, 'ras', 'mni_tal')
     _write_fs_xfm(fname_to, T_ras_mni['trans'], kind)
+
+
+class Coregistration(object):
+    """
+    Performs automated coregistration.
```
agramfort, line 1232:
```suggestion
    """Class for MRI<->MEG coregistration.
```

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+    Parameters
+    ----------
+    info : instance of Info
+        The measurement info.
+    subjects_dir : directory
```
agramfort, line 1238:
```suggestion
    subjects_dir : path-like
        Path to MRI subjects directory.
```

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        SUBJECTS_DIR.
+    subject : str
+        Subject, corresponding to a folder in SUBJECTS_DIR.
+    """
+
+    def __init__(self, info, subjects_dir, subject):
```
agramfort, line 1244:
```suggestion
    def __init__(self, info, subject, subjects_dir=None):
```
subject is commonly passed first and subjects_dir can be None in which case we find it from the os.environ
See how it's done elsewhere

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    def has_mri_data(self):
+        return len(self.transformed_high_res_mri_points) > 0
+
+    @property
+    def has_hsp_data(self):
+        return (self.has_mri_data and
+                len(self.nearest_transformed_high_res_mri_idx_hsp) > 0)
+
+    @property
+    def orig_hsp_point_distance(self):
```
agramfort, line 1435: docstring please

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        head_pts = np.concatenate(head_pts)
+        mri_pts = np.concatenate(mri_pts)
+        weights = np.concatenate(weights)
+        if self.n_scale_param == 0:
+            mri_pts *= self.parameters[6:9]  # not done in fit_matched_points
+        return head_pts, mri_pts, weights
+
+    def fit_icp(self, iterations=20, lpa_weight=1., nasion_weight=10.,
+                rpa_weight=1.):
+        """Find MRI scaling, translation, and rotation to match HSP."""
```
agramfort, line 1501: docstring please

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+                self.parameters[:] = est
+            angle, move, scale = self.changes
+            self._parameters_items_changed()
+            if angle <= self.icp_angle and move <= self.icp_distance and \
+                    all(scale <= self.icp_scale):
+                break
+
+    def omit_hsp_points(self, distance):
+        """Exclude head shape points that are far away from the MRI head."""
+        import warnings
```
agramfort, line 1529: import at the top

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        mask = self.orig_hsp_point_distance <= distance
+        n_excluded = np.sum(~mask)
+        logger.info("Coregistration: Excluding %i head shape points with "
+                    "distance >= %.3f m.", n_excluded, distance)
+        # set the filter
+        with warnings.catch_warnings(record=True):  # comp to None in Traits
+            self.hsp.points_filter = mask
+
+    def point_distance(self):
+        """Compute euclidean distance between the head-mri points."""
```
agramfort, line 1544: docstring
```suggestion
        """Compute Euclidean distance between the head-mri points."""
```

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            head_pts, mri_pts, weights = self._setup_icp()
+            est = fit_matched_points(mri_pts, head_pts,
+                                     scale=self.n_scale_param,
+                                     x0=est, out='params', weights=weights)
+            if self.n_scale_param == 0:
+                self.parameters[:6] = est
+            elif self.n_scale_param == 1:
+                self.parameters[:] = list(est) + [est[-1]] * 2
+            else:
+                self.parameters[:] = est
```
agramfort, line 1520: is it documentated what self.parameters can contain depending on self.n_scale_param?

why not using out='trans' and just work with the trans?

@larsoner any opinion?

larsoner: > is it documentated what self.parameters can contain depending on self.n_scale_param?

No but it can be figured out from the code and documented in this PR. IIRC it's rotation + quats + scale factors (so 6, 7, or 9 params)

> why not using out='trans' and just work with the trans?

Loss of precision and internal computations / optimization is done on quaternions so better just to quat_to_rot on the fly when necessary (it's very cheap in the grand scheme of ICP or viz)

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    @property
+    def mri_trans_noscale(self):
+        if self.coord_frame == 'head':
+            t = self.mri_head_t
+        else:
+            t = np.eye(4)
+        return t
+
+    @property
+    def mri_trans(self):
```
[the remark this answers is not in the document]
larsoner: Most are shortcuts to make some other processing easier. You can try to remove some but my guess is you will realize why they were needed/helpful. @GuillaumeFavelier take a look at where it's used and if you can't figure out why it's there I can dig in and look

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    def has_nasion_data(self):
+        return (np.any(self.mri.nasion) and np.any(self.hsp.nasion))
+
+    @property
+    def has_rpa_data(self):
+        return (np.any(self.mri.rpa) and np.any(self.hsp.rpa))
+
+    @property
+    def head_mri_t(self):
+        trans = rotation(*self.parameters[:3]).T
```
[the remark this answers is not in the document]
larsoner: -1 on this (see below), better to store the cleanest representation internally (quat + trans + scale) and convert to 4x4 affine when necessary

mne/coreg.py
```diff
@@ -1223,3 +1225,528 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        self.icp_fid_match = self.icp_fid_matches[0]
+        self.grow_hair = 0.
+        self.lpa_weight = 1.
+        self.nasion_weight = 10.
+        self.rpa_weight = 1.
+        self.hsp_weight = 1.
+        self.eeg_weight = 1.
+        self.hpi_weight = 1.
+
+        self.hsp = _DigSource(info)
```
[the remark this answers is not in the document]
larsoner: It's shorthand for "headshape points" and a holdover of Polhemus systems. I'm fine with standardizing around something else like "extra points" (which is what we call it in info/dig) but make sure that HSP here does not include other stuff (eeg points, HPI points, etc.) and really just includes these "extra" dig points

mne/coreg.py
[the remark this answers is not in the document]
agramfort: @GuillaumeFavelier I think you need now to write down in a docstring what self.parameters actually contains depending on scenarios and you should have tiny tests on the self.parameters attributes depending on those scenarios.

also @GuillaumeFavelier to follow the MNE good practices you should expose in the public API only the functions you consider part of the public interface.

mne/coreg.py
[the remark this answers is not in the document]
GuillaumeFavelier: I don't know. It seems to be the case:

```
  27     @property
  26     def _transformed_hsp_hpi(self):
  25         return apply_trans(self._hsp_trans, self._hsp.hpi_points)
  24 
  23     @property
  22     def _transformed_hsp_eeg_points(self):
  21         return apply_trans(self._hsp_trans, self._hsp.eeg_points)
  20 
  19     @property
  18     def _transformed_hsp_points(self):
  17         return apply_trans(self._hsp_trans, self._hsp.points)
```

mne/coreg.py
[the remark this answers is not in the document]
GuillaumeFavelier: I ended turning all the properties and attributes private except the methods used in the example script.

mne/coreg.py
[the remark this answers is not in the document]
GuillaumeFavelier: I ported the minimum to reproduce https://github.com/mne-tools/mne-python/pull/9516#issuecomment-876477760

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
                 F_mri_ras, 'ras', 'ras'),
             F_ras_mni, 'ras', 'mni_tal')
     _write_fs_xfm(fname_to, T_ras_mni['trans'], kind)
+
+
+class Coregistration(object):
+    """
+    Class for MRI<->MEG coregistration.
```
larsoner, line 1234:
```suggestion
    Class for MRI<->head coregistration.
```

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        The measurement info.
+    subject : str
+        Name of the subject the data are defined for.
+    subjects_dir : path-like
+        Path to MRI subjects directory.
+
+    Notes
+    -----
+    Internal computation quantities ``parameters`` are in units of (for X/Y/Z):
+    - ``parameters[:3]`` are in radians
```
larsoner, line 1248: I was wrong, this is correct! These are Euler angles, not unit quaternions...

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    - ``parameters[3:6]`` are in m
+    - ``paramteres[6:9]`` are in scale proportion
```
larsoner, line 1250: I was also wrong about this, it's always 9 params (and are the x/y/z scale factors)

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        move = np.linalg.norm(old[3:6] - new[3:6]) * 1e3
+        angle = np.rad2deg(_angle_between_quats(
+            rot_to_quat(rotation(*new[:3])[:3, :3]),
+            rot_to_quat(rotation(*old[:3])[:3, :3])))
+        percs = 100 * (new[6:] - old[6:]) / old[6:]
+        return move, angle, percs
+
+    @property
+    def _transformed_hsp_hpi(self):
+        return apply_trans(self._hsp_trans, self._hsp.hpi_points)
```
larsoner, line 1300: We can make these properties, but the advantage of caching from `traitsui` goes away, which is sad. Some of these will end up being a bit slow to recompute, hence the recompute-only-when-necessary behavior in `main` is nice.

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    @property
+    def _mri_trans_noscale(self):
+        if self._coord_frame == 'head':
+            t = self._mri_head_t
+        else:
+            t = np.eye(4)
+        return t
+
+    @property
+    def _mri_trans(self):
```
larsoner, line 1430: This is the transformation to apply to the MRI points, depending on the chosen coordinate frame (head or MRI)

The reason we have head and MRI coord frames is to allow people to think about moving the dig points to the MRI (when viewed in the MRI coord frame) or about stretching/reconfiguring the MRI to conform to the head (when viewed in the head coordinate frame). If this seems too complex, we could get rid of it and always use the MRI coordinate frame.

Really though this difference in views can be implemented at the view level rather than the visualization level. It's reasonable in this class to do everything in MRI coordinates I think.

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        points = apply_trans(self._mri_trans,
+                             self._processed_high_res_mri_points)
+        return points
+
+    @property
+    def _has_mri_data(self):
+        return len(self._transformed_high_res_mri_points) > 0
+
+    @property
+    def _has_hsp_data(self):
```
larsoner, line 1446: we should change hsp->dig because really what we're talking about is called "dig" / Digitization everywhere else in MNE

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+    def save_trans(self, fname):
+        """Save the head-mri transform as a fif file.
+
+        Parameters
+        ----------
+        fname : str
+            Target file path.
+        """
+        write_trans(fname, Transform('head', 'mri', self._head_mri_t))
```
larsoner, line 1622: We don't really need this in the class. Instead we can have a `.trans` property that returns the `Transform(...)`, and in the actual GUI we can just call `write_trans(fname, coreg.trans)`

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+    @property
+    def points(self):
+        if self.points_filter is None:
+            return self._hsp_points
+        else:
+            return self._hsp_points[self.points_filter]
+
+    @property
+    def _hsp_points(self):
```
larsoner, line 1653: `_extra_points` would be a better name

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        old = np.array(self._last_parameters, float)
+        move = np.linalg.norm(old[3:6] - new[3:6]) * 1e3
+        angle = np.rad2deg(_angle_between_quats(
+            rot_to_quat(rotation(*new[:3])[:3, :3]),
+            rot_to_quat(rotation(*old[:3])[:3, :3])))
+        percs = 100 * (new[6:] - old[6:]) / old[6:]
+        return move, angle, percs
+
+    @property
+    def _transformed_hsp_hpi(self):
```
larsoner, line 1299: _transformed_dig_hpi

larsoner (commented):
Looks like a nice clean start so far!

The existing/old names are not great because we have stuff like `hsp.hsp_points`. These names are not ideal. I propose we call all points dig points (so `DigSource` would become the `._dig` attribute of the class), and it has points of type `eeg` (`._dig.eeg_points`), `hpi` (`._dig.hpi_points`), and most importantly `extra` -- this is currently `hsp` so you end up with things like `._hsp.points` which is vague, but combined with the above suggestion this would become `._dig.extra_points`

mne/coreg.py
larsoner: Having this class separated out from the GUI stuff, I think the best way forward is to have a single `def _update_params(params)` method that looks at the `_last_params` and the new ones and updates the exact set of `_`-variables that need to be updated. This removes all the `property` business, and also makes it so that things can be updated in the right order (which was always a problem with the exsting `coreg` code actually!).

I suggest that the first step be the renaming ._hsp->._dig and DigSource.points - > DigSource.extra_points. Then the next step/commit is removal of all of these `@property`s using a single monolithic `_update_params` that does the right thing. (If you set for example `self._last_params = np.array([np.nan] * 9)` then call `self._update_params([0] * 6 + [1] * 3)` at the end of the class constructor, it should automatically populate all `_`-variables nicely.

larsoner: ... and if this is not clear, I'm happy to have a video chat on Discord to discuss, or I can push some commits to move things in this direction!

larsoner:
BTW the suggested monolithic `_update_params` also means we don't need any `traitlets` (or `traits`), which is nice...

mne/coreg.py
agramfort, line 1250:
```suggestion
    - ``parameters[6:9]`` are in scale proportion
```

agramfort: can we rather have more meaning variable names than parameters? eg rotations, translations, scales ?

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    - ``paramteres[6:9]`` are in scale proportion
+    """
+
+    def __init__(self, info, subject, subjects_dir):
+        self._default_parameters = [0., 0., 0., 0., 0., 0., 1., 1., 1.]
+        self._parameters = list(self._default_parameters)
+        self._last_parameters = list(self._default_parameters)
+
+        self._n_scale_params = (0, 1, 3)
+        self._n_scale_param = self._n_scale_params[0]
```
agramfort, line 1259: how do we expose this option of number of scale params here?

mne/coreg.py
```diff
@@ -1223,3 +1227,586 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        self._coord_frames = ('mri', 'head')
+        self._coord_frame = self._coord_frames[0]
```
agramfort, line 1261: I don't remember why we support both coord_frames but maybe we can start by supporting only one eg mri if it's the default? Would make things simpler....

mne/coreg.py
agramfort: +100

mne/coreg.py
agramfort: I also leaning towards doing everything in MRI space...

mne/coreg.py
agramfort: +1

mne/coreg.py
larsoner: Agreed, this class can do everything in MRI space and the viewer/GUI can take care of which coordinate frame to display everything in. It's easier to do the transformation at that level I think as it's really a viz problem

mne/coreg.py
larsoner: I actually don't think we need this `n_scale_params` at all -- we can just have the monolithic `_update_params` take into account the `self.scale_factors` being all identity (no scale), all the same and non-identity (uniform), or different (3-axis scaling). The n_scale_params is really a GUI concept/switch and I don't think it's needed here.

In terms of programatically setting things, I think we probably want a property `self.scale_factors = (1., 1., 1.)` or so, and maybe a `set_scale_factors(...)`, `set_euler_angles(...)`, `set_translation(...)` that all internally call `self._update_params(...)`

GuillaumeFavelier:
From https://github.com/mne-tools/mne-python/issues/9542, the remaining changes are:

- [ ] implement `self._update_params` and remove `@property`
- [ ] rename `self.parameters` -> `self.rotations`, `self.translations`, `self.scales`
- [ ] implement `self.set_rotations`, `self.set_translations` and `self.set_scales`
- [ ] refactor `self._n_scale_params` out

Also maybe more `hsp`->`dig`

larsoner:
> implement self.set_rotations, self.set_translations and self.set_scales

Yes but I see this as a one- or two-line wrapper around `self._update_params`. In fact if you make the signature `def _update_params(rotation=None, translation=None, scale=None)` then you know immediately what (at most) was changed based on what is non-None, and each `set_*` method becomes a one-line wrapper.

> Also maybe more hsp->dig

This should definitely be part of it. If it's not clear what I mean, I can do the renaming in a few minutes probably. Just let me know if/when I should push a commit. Otherwise I'm also happy to look at what you do and complain if it's not what I had in mind :)

GuillaumeFavelier:
2d30c1aa0cf450183ac3d177f335d115ed6aab0d is my first try at `_update_params()` (removing the properties: `_head_mri_t`, `_mri_head_t`, `_mri_trans_noscale`, `_mri_trans`) if I understood roughly the idea. Also, feel free to complain @larsoner

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
                 F_mri_ras, 'ras', 'ras'),
             F_ras_mni, 'ras', 'mni_tal')
     _write_fs_xfm(fname_to, T_ras_mni['trans'], kind)
+
+
+class Coregistration(object):
+    """
+    Class for MRI<->head coregistration.
+
```
agramfort, line 1235:
```suggestion
    """Class for MRI<->head coregistration.

```

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            rot_to_quat(rotation(*old[:3])[:3, :3])))
+        percs = 100 * (new[6:] - old[6:]) / old[6:]
+        return move, angle, percs
+
+    @property
+    def _transformed_dig_hpi(self):
+        return apply_trans(self._hsp_trans, self._dig.hpi_points)
+
+    @property
+    def _transformed_dig_eeg_points(self):
```
agramfort, line 1324:
```suggestion
    def _transformed_dig_eeg(self):
```

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        return apply_trans(self._hsp_trans, self._dig.eeg_points)
+
+    @property
+    def _transformed_dig_points(self):
```
agramfort, line 1328:
```suggestion
    def _transformed_dig_extra(self):
```

would be more consistent.

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    @property
+    def _hsp_trans(self):
+        if self._coord_frame == 'head':
+            t = np.eye(4)
+        else:
+            t = self._head_mri_t
+        return t
+
+    @property
+    def _transformed_orig_hsp_points(self):
```
agramfort, line 1340:
```suggestion
    def _transformed_orig_dig_extra(self):
```

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        # find the new filter
+        mask = self._orig_hsp_point_distance <= distance
+        n_excluded = np.sum(~mask)
+        logger.info("Coregistration: Excluding %i head shape points with "
+                    "distance >= %.3f m.", n_excluded, distance)
+        # set the filter
+        with warnings.catch_warnings(record=True):  # comp to None in Traits
+            self._dig.extra_points_filter = mask
+
+    def point_distance(self):
```
agramfort, line 1584: can you we have a more explicit name? it's all dig points and scalp surface? also I would use plural to explicit it's all distances. eg


```suggestion
    def compute_dig_head_distances(self):
```

?

mne/coreg.py
```diff
@@ -1223,3 +1227,576 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    def reset(self):
+        """Reset all the parameters affecting the coregistration."""
+        self._n_scale_param = self._n_scale_params[0]
+        self._grow_hair = 0.
+        self._parameters[:] = list(self._default_parameters)
+        self._last_parameters[:] = list(self._default_parameters)
+        self.omit_hsp_points(np.inf)
+
+
+class _DigSource(object):
```
agramfort, line 1623: do we really need the next 3 classes?

mne/tests/test_coreg.py
```diff
@@ -279,3 +279,27 @@ def test_get_mni_fiducials():
     fids_est = np.array([f['r'] for f in fids_est])
     dists = np.linalg.norm(fids - fids_est, axis=-1) * 1000.  # -> mm
     assert (dists < 8).all(), dists
+
+
+@testing.requires_testing_data
+def test_coregistration():
+    """Test automated coregistration."""
```
agramfort, line 286: can you test with the different scaling options?

agramfort (commented):
you have a merge conflict @GuillaumeFavelier

thanks @GuillaumeFavelier we're making great progress 💪

mne/coreg.py
larsoner: @GuillaumeFavelier I think you can just always use the `mri` coordinate frame in this class

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        move = np.linalg.norm(self._last_translation - self._translation) * 1e3
+        angle = np.rad2deg(_angle_between_quats(
+            rot_to_quat(rotation(*self._rotation)[:3, :3]),
+            rot_to_quat(rotation(*self._last_rotation)[:3, :3])))
+        percs = 100 * (self._scale - self._last_scale) / self._last_scale
+        return move, angle, percs
+
+    @property
+    def _transformed_dig_hpi(self):
+        return apply_trans(self._hsp_trans, self._dig.hpi_points)
```
larsoner, line 1374: Let's not `@property` any of these, instead update all of them as private attributes in `_update_params` as needed

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    def _has_nasion_data(self):
+        return (np.any(self._mri.nasion) and np.any(self._dig.nasion))
+
+    @property
+    def _has_rpa_data(self):
+        return (np.any(self._mri.rpa) and np.any(self._dig.rpa))
+
+    def _nearest_calc_default(self):
+        return _DistanceQuery(
+            self._processed_high_res_mri_points * self._scale)
```
larsoner, line 1450: No need for this function, all of the `_default` methods were just to make Traits happy. Just set this in the constructor

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        except Exception:  # some problem, leave at origin
+            self.fid.mni_points = None
+        else:
+            self.fid.mni_points = np.array([f['r'] for f in fids], float)
+
+        # find fiducials file
+        fid_files = _find_fiducials_files(subject, subjects_dir)
+        if len(fid_files) == 0:
+            self.fid.reset_traits(['file'])
+            self.lock_fiducials = False
```
larsoner, line 1844: locking of fiducials is really a GUI thing, no need to have this in the class

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+                except Exception:
+                    raise ValueError(
+                        "Error loading surface from %s (see "
+                        "Terminal for details)." % filename)
+            self.surf = _Surf(rr=bem['rr'], tris=bem['tris'], nn=bem['nn'])
+        else:
+            self.surf = _Surf()
+
+
+class _FiducialsSource(object):
```
larsoner, line 1778: Don't need this

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+
+class _Surf(object):
+    def __init__(self, rr=None, nn=None, tris=None):
+        self.rr = np.empty((0, 3)) if rr is None else rr
+        self.nn = np.empty((0, 3)) if nn is None else nn
+        self.tris = np.empty((0, 3)) if tris is None else tris
+
+
+class _SurfaceSource(object):
```
larsoner, line 1759: Don't think we need this, just use `read_surface` etc. directly when needed

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    @property
+    def lpa(self):
+        return self._cardinal_point(FIFF.FIFFV_POINT_LPA)
+
+    @property
+    def rpa(self):
+        return self._cardinal_point(FIFF.FIFFV_POINT_RPA)
+
+
+class _Surf(object):
```
larsoner, line 1752: Don't need this

mne/coreg.py
larsoner: Agreed we should get rid of this class and just use things like `mne.io.digitization._get_data_as_dict_from_dig`, I think it'll get us most of the way there.

mne/coreg.py
```diff
@@ -1242,3 +1245,613 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        try:
+            return _fiducial_coords(*read_fiducials(self.filename))
+        except Exception as err:
+            raise ValueError(
+                "Error reading fiducials from %s: %s (See terminal "
+                "for more information)" % (self.filename, str(err)),
+                "Error Reading Fiducials")
+
+
+class _MRIHeadWithFiducialsModel(object):
```
larsoner, line 1796: Don't need this, just load head surfaces directly in the Coreg class. Maybe some of this should become a private function, but really we should refactor with the MRI-head-surface-finding code in `mne.viz.plot_alignment`

larsoner (commented):
I like that it's getting smaller and simpler on each iteration, keep up the good work @GuillaumeFavelier !

I think once this class has a clean interface we should add a little tutorial using an updated version of the code @jhouck wrote:

https://gist.github.com/jhouck/c3a2a3ceae1c5a27af6f568f6cd9641c

It will make sure our new Coregistration class does everything it needs to since @jhouck used most of the functionality we need

mne/coreg.py
```diff
@@ -1242,3 +1245,572 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+                apply_trans(self._head_mri_t, self._dig.hpi_points)
+            self._transformed_dig_eeg = \
+                apply_trans(self._head_mri_t, self._dig.eeg_points)
+            self._transformed_dig_extra = \
+                apply_trans(self._head_mri_t, self._dig.extra_points)
+            self._transformed_orig_dig_extra = \
+                apply_trans(self._head_mri_t, self._dig._extra_points)
+            self._nearest_transformed_high_res_mri_idx_orig_hsp = \
+                self._nearest_calc.query(
+                    apply_trans(self._head_mri_t, self._dig._extra_points))[1]
```
larsoner, line 1321: The order helps us here, as we can do things like this:
```suggestion
                self._nearest_calc.query(self._transformed_orig_dig_extra)[1]
```

mne/coreg.py
```diff
@@ -1242,3 +1245,572 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            self._nearest_transformed_high_res_mri_idx_hpi = \
+                self._nearest_calc.query(
+                    apply_trans(self._head_mri_t, self._dig.hpi_points))[1]
```
larsoner, line 1324: Same here and probably others

mne/coreg.py
```diff
@@ -1242,3 +1245,572 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            return
+
+        # find the new filter
+        mask = self._orig_hsp_point_distance <= distance
+        n_excluded = np.sum(~mask)
+        logger.info("Coregistration: Excluding %i head shape points with "
+                    "distance >= %.3f m.", n_excluded, distance)
+        # set the filter
+        with warnings.catch_warnings(record=True):  # comp to None in Traits
+            self._dig.extra_points_filter = mask
```
larsoner, line 1594: This `omit_hsp_points` will need to call some kind of update. Maybe we should add a `force_update_omitted=False` to `self._update_params` and then this can call:
```
self._update_params(force_update_omitted=True)
```
And in the `_update_params` the filtered points will be updated as above (when things change) *or* when `force_update_omitted is True`.

This also made me realize that this class is missing a `self.set_grow_hair` method that exists in `mne coreg`. Basically it "pushes out" the MRI points. This method can just set the value and then do `self._update_params(self._rotation, self._translation, self._scale)` or so, because when the hair grows I think everything needs to be updated IIRC.

larsoner (commented):
Looks like it's getting closer, looking forward to the next iteration @GuillaumeFavelier !

mne/coreg.py
GuillaumeFavelier: Why do we have to recompute here? I thought the goal with `omit_hsp_points()` was to use the mask and hide the points in the GUI.

larsoner: Not just the GUI, it's used during fitting. Let's say you have 100 "extra" points but 3 were accidentally digitized 5 cm from the head (it happens sometimes if you mis-click the digitizing pen for example in Polhemus). You omit by distance and use the remaining points for ICP fitting and computed distances, not just viz.

agramfort:
@GuillaumeFavelier please have a look at the coverage report of coreg.py

you can get it with:

`pytest mne/tests/test_coreg.py --cov=mne --cov-report html`

it seems the grow hair option is not tested. 
Alse the call to omit_hsp_points with distance <= 0

I would suggest you add to this PR the example script you started from by @jhouck 

thx @GuillaumeFavelier !

agramfort:
@jhouck can you tell us what you think of this PR? We started from your example

GuillaumeFavelier:
The rendered example is available:

https://31645-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/coregistration.html?highlight=co%20registration

mne/io/_digitization.py
```diff
@@ -259,7 +259,10 @@ def _get_data_as_dict_from_dig(dig):
         elif d['kind'] == FIFF.FIFFV_POINT_EXTRA:
             hsp.append(d['r'])
         elif d['kind'] == FIFF.FIFFV_POINT_EEG:
-            if d['ident'] != 0:  # ref channel
+            if exclude_ref_channel:
+                if d['ident'] != 0:  # ref channel
+                    dig_ch_pos_location.append(d['r'])
+            else:
```
larsoner, line 265: Simpler logic
```suggestion
            if d['ident'] != 0 or not exclude_ref_channel:
                dig_ch_pos_location.append(d['r'])
```

mne/tests/test_coreg.py
```diff
@@ -279,3 +279,32 @@ def test_get_mni_fiducials():
...
+    assert coreg._extra_points_filter is None
+    coreg.omit_hsp_points(distance=-1)
+    coreg.omit_hsp_points(distance=5. / 1000)
+    assert coreg._extra_points_filter is not None
+    coreg.fit_icp()
+    assert isinstance(coreg.trans, Transform)
+    errs_icp = coreg.compute_dig_head_distances()
+    assert np.median(errs_icp * 1000) < 4
+    coreg.reset()
+    assert np.allclose(coreg._parameters, default_params)
```
larsoner, line 310: It would be good to load the actual `-trans.fif` and assert that the translation and rotation (using `angle_between_quats`) is sufficiently small

mne/tests/test_coreg.py
larsoner: This is still pending.

For scalings you could for example test first that even with scalings enabled, you still end up with a similar head<->MRI trans and near-unity scale factors.

Then you can test that if you for example:
```
for d in info['dig']:
    d['r'] *= [0.8, 1.1, 1.1]
```
you get approximately these scale factors back from the automated routine. (Or maybe their inverse, not sure). I chose Y=Z here because IIRC the x axis is pretty well aligned between the MRI and head, but there is a rotation about the X axis, so the Y and Z scale factors will end up mixed, so setting them the same way just makes things easier.

examples/visualization/coregistration.py
```diff
@@ -0,0 +1,71 @@
+"""
+=========================================
+Use automated approach to co-registration
```
larsoner, line 3: @drammock it would be good if you could comment on where this should live, and maybe push some commits to restructure to have a more narrative form.

Also I think that the functions that the `fit_*` functions should `logger.info` the dig<->MRI distances before the fit and after the fit. Then they can just be called with `verbose=True` in the tutorial in separate code blocks (`# %%`) along with a `plot_alignment` call after each step), and then the outputs will be nice and readable in the SG output.

mne/tests/test_coreg.py
GuillaumeFavelier: so my idea is:

```
assert_allclose(_angle_between_quats(rot_to_quat(coreg._rotation), X)
```

where `X` is the quat obtained from the rot and trans matrices in the `-trans.fif` file but I don't know how to extract those.

GuillaumeFavelier: After an offline discussion with @agramfort, I tried b6abb03 to test the angle between trans.

GuillaumeFavelier:
In 078b2f5, I added testing for scale estimation but it will fail since scale fitting is not enabled in `fit_matched_point()`. Should this parameter be exposed in `fit_fiducials()` and `fit_icp()` or enabeld by default? Just using `scale=True` does not improve the result in my experiments.

examples/visualization/coregistration.py
drammock: I feel like `examples/visualization` is not the right place for this, as the point of coreg is mostly about finding the right transform, and the visualization is just the tool we use to do that. IMO there isn't really an appropriate section within `examples`...  it would fit pretty naturally as `tutorials/forward/25_automated_coreg.py` (coming right after the tutorial about general coreg / coord frame concepts).

To fit in as a tutorial it would need a more verbose / extended narrative, but wouldn't need to be *hugely* longer.  I'd recommend moving it to that location now, and expand the narrative either now (if @jhouck wants to expand it) or in a separate PR (I can do it if @jhouck doesn't want to).


> Also I think that the functions that the `fit_*` functions should `logger.info` the dig<->MRI distances before the fit and after the fit. Then they can just be called with `verbose=True` in the tutorial in separate code blocks (`# %%`) along with a `plot_alignment` call after each step), and then the outputs will be nice and readable in the SG output.

+1 for this suggestion; I often think that if we're `print`ing something in a tutorial then maybe it's something we ought to be logging anyway.

GuillaumeFavelier:
I updated the tutorial, I mainly removed the call to `write_fiducials()` (does not seem necessary?) and added more calls to `plot_alignment()`.

examples/visualization/coregistration.py
GuillaumeFavelier: I moved the tutorial and I added `_log_dig_mri_distance()`.

jhouck:
> The rendered example is available:
> 
> https://31645-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/coregistration.html?highlight=co%20registration

This looks good! Nice work upstream dealing with traits and the various private functions.

larsoner:
> I mainly removed the call to write_fiducials() (does not seem necessary?) and added more calls to plot_alignment().

It might be necessary if someone has a vanilla fsaverage. Nowadays we fix this just by calling `fetch_fsaverage` instead (it adds this file), so that's what the tutorial should probably do.

> Should this parameter be exposed in fit_fiducials() and fit_icp() or enabled by default? Just using scale=True does not improve the result in my experiments.

I think we probably want a `set_scale_mode(None | 'uniform' | '3-axis')` or so. Usually you want to be in one of these three modes and not change very often.

Indeed `scale=True` should not improve anything for the real data case. It's a good sanity check that the scale factors in the 1- and 3-factor case are all close to unity, and the head<->MRI transformations are all very similar.

Where it should matter is if you take the `info['dig']` and modify them [as suggested above](https://github.com/mne-tools/mne-python/pull/9516#discussion_r676841119). In this case, the scale factors should matter. More concretely, I would test five cases:

1. scale=None, real dig
2. scale='uniform', real dig: head<->MRI should match scale=None and one unity scale factor recovered
3. scale='3-axis', real dig: head<->MRI should match scale=None and three unity scale factors recovered
4. scale='uniform', dig scaled down by some scale factor: that scale factor recovered (not sure what will happen with head<->MRI actually)
5. scale='3-axis', dig scaled down by 3 scale factors: those three scale factors recovered (again not sure about head<->MRI)

We don't particularly care about the other cases (e.g., dig scaled, but scale=None), so I would just focus on making sure these 5 work.

larsoner:
@GuillaumeFavelier does that make sense? If not, I could push a commit with the test code that I have in mind that should pass. (True TDD!)

GuillaumeFavelier:
>  I could push a commit with the test code that I have in mind that should pass. 

Please do, it will help greatly

I just pushed d2db626f3d8b6e4b1dc8c38c3303a472f69288b8 to add `fetch_fsaverage()`.

GuillaumeFavelier:
I parametrized the test:

```
@pytest.mark.parametrize('ref_scale', [
    (None, [1., 1., 1.]),
    ('uniform', [1., 1., 1.]),
    ('3-axis', [1., 1., 1.]),
    ('uniform', [0.8, 1., 1.]),
    ('3-axis', [0.8, 0.8, 0.8])])
```

Hopefully, it is similar to what you have in mind @larsoner 

Locally, `ref_scale2` ('3-axis', real dig) fails with:

```
______________________________________________________________________ test_coregistration[ref_scale2] _______________________________________________________________________
mne/tests/test_coreg.py:321: in test_coregistration
    assert np.median(errs_icp * 1000) < 10
E   assert 209903.46365501382 < 10
E    +  where 209903.46365501382 = <function median at 0x7f0e2d0e8b80>((array([214.4902833 , 209.8582842 , 209.90683272, 209.88562759,\n       209.8588608 , 209.93050095, 209.91853943, 209.84...16795, 209.85467031,\n       209.82849843, 209.85099058, 209.83717712, 209.93365572,\n       209.92734148, 209.81050887]) * 1000))
E    +    where <function median at 0x7f0e2d0e8b80> = np.median
```

and `ref_scale3` ('uniform', dig scaled) fails with:

```
______________________________________________________________________ test_coregistration[ref_scale3] _______________________________________________________________________
mne/tests/test_coreg.py:328: in test_coregistration
    assert_allclose(coreg._scale, ref_scale[1], atol=0.1)
E   AssertionError: 
E   Not equal to tolerance rtol=1e-07, atol=0.1
E   
E   Mismatched elements: 1 / 3 (33.3%)
E   Max absolute difference: 0.13845061
E   Max relative difference: 0.17306327
E    x: array([0.938451, 0.938451, 0.938451])
E    y: array([0.8, 1. , 1. ])
```

GuillaumeFavelier:
I added `set_scale_mode()` and reintroduced the local variable `n_scale_params` to enable fitting of scale parameters. I used the GUI code as reference for that.

mne/tests/test_coreg.py
```diff
@@ -279,3 +280,51 @@ def test_get_mni_fiducials():
...
     assert (dists < 8).all(), dists
+
+
+@testing.requires_testing_data
+@pytest.mark.parametrize('ref_scale', [
+    (None, [1., 1., 1.]),
+    ('uniform', [1., 1., 1.]),
+    ('3-axis', [1., 1., 1.]),
+    ('uniform', [0.8, 1., 1.]),
+    ('3-axis', [0.8, 0.8, 0.8])])
```
larsoner, line 291: These are reversed (the uniform should be uniform), and let's try having the Y and Z be non-unity
```suggestion
    ('uniform', [0.8, 0.8, 0.8]),
    ('3-axis', [0.8, 1.2, 1.2])])
```

GuillaumeFavelier:
It appears that resolving the `traits` dependencies are way trickier than I thought. I discovered the main issue: I underestimated `_update_nearest_calc()`. In `_update_params` all the properties using `_nearest_calc` were outdated, that's where the error was coming from. I pushed those at the end to solve this. Now after ad9a31a8a94ca7ffc2715701b0be7e5018667501, the result of the PR is similar to the reference code in my test of '3-axis'. I'll double-check 'uniform' now.

GuillaumeFavelier:
Did I miss anything with `fetch_fsaverage()`? Circle did not seem to like it.

```
  File "/home/circleci/project/tutorials/forward/25_automated_coreg.py", line 42, in <module>
    coreg = Coregistration(info, subject, subjects_dir)
  File "/home/circleci/project/mne/coreg.py", line 1306, in __init__
    self._setup_mri(subjects_dir=self._subjects_dir,
  File "/home/circleci/project/mne/coreg.py", line 1351, in _setup_mri
    fid_filename = fid_files[0].format(subjects_dir=subjects_dir,
IndexError: list index out of range
```

My last commit 9d66b7c94b16ea3b142f054fa76fb67c77af89a5 rolled back to `write_fiducials()` and now it's green.

mne/coreg.py
```diff
@@ -1245,3 +1248,558 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    """Class for MRI<->head coregistration.
+
+    Parameters
+    ----------
+    info : instance of Info
+        The measurement info.
+    subject : str
+        Name of the subject the data are defined for.
+    subjects_dir : path-like
+        Path to MRI subjects directory.
```
agramfort, line 1263: Can you document the trans attribute ?

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,59 @@
+"""
+=========================================
+Use automated approach to co-registration
```
agramfort, line 3:
```suggestion
Use automated approach to coregistration
```

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,59 @@
...
+=========================================
+
+This example shows how to use the get_mni_fiducials routine and the
+coregistration functions to perform an automated MEG-MRI co-registration.
```
agramfort, line 7:
```suggestion
coregistration functions to perform an automated MEG-MRI coregistration.
```

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,59 @@
...
+
+.. warning:: The quality of the co-registration depends heavily upon the
```
agramfort, line 9:
```suggestion
.. warning:: The quality of the coregistration depends heavily upon the
```

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,59 @@
...
+             quality of the head shape collected during subject prepration and
+             the quality of your T1-weighted MRI. Use with caution and check
+             the co-registration error.
```
agramfort, line 12:
```suggestion
             the coregistration error.
```

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,59 @@
...
+from mne.io import write_fiducials, read_info
+from mne.io.constants import FIFF
+
+
+data_path = mne.datasets.sample.data_path()
+subjects_dir = op.join(data_path, 'subjects')
+subject = 'sample'
+fname_fids = op.join(subjects_dir, subject, 'bem', subject + '-fiducials.fif')
+fids_mri = get_mni_fiducials(subject, subjects_dir)
+write_fiducials(fname_fids, fids_mri, coord_frame=FIFF.FIFFV_COORD_MRI)
```
agramfort, line 31: Writing something to disk should not be necessary

mne/coreg.py
```diff
@@ -1245,3 +1248,558 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+        Parameters
+        ----------
+        lpa_weight : float
+            Relative weight for LPA. The default value is 1.
+        nasion_weight : float
+            Relative weight for nasion. The default value is 10.
+        rpa_weight : float
+            Relative weight for RPA. The default value is 1.
+        %(verbose)s
```
agramfort, line 1617: Methods that do not return anything should return self do we can chain operations. Don’t forget to update docstring when you update this.

agramfort (commented):
We’re getting there @GuillaumeFavelier 

did you also take pieces from my commit eg to prevent fit_fiducial with 3 scales and only 3 points ?

mne/coreg.py
GuillaumeFavelier: It's a property, not an attribute. Should I still document it?

tutorials/forward/25_automated_coreg.py
GuillaumeFavelier: I did not manage to make it work for Circle when I used `fetch_fsaverage()`.

mne/coreg.py
agramfort: I would say yes. Although I agree that it should already be visible in the doc

GuillaumeFavelier:
It's ready for review @larsoner, @agramfort

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        Path to MRI subjects directory.
+
+    Attributes
+    ----------
+    trans : instance of Transform
+        MRI<->Head coordinate transformation.
+
+    Notes
+    -----
+    Internal computation quantities parameters are in units of (for X/Y/Z):
```
agramfort, line 1272:
```suggestion
    Internal computation quantities parameters are in the following units:
```

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        # find high-res head model (if possible)
+        high_res_path = _find_head_bem(subject, subjects_dir, high_res=True)
+        low_res_path = _find_head_bem(subject, subjects_dir, high_res=False)
+        if high_res_path is None and low_res_path is None:
+            raise RuntimeError("No standard head model was "
+                               f"found for subject {subject}")
+        if high_res_path is not None:
+            self.bem_high_res = self._read_surface(high_res_path)
+        else:
+            self.bem_high_res = self._read_surface(low_res_path)
```
agramfort, line 1330: Can you log with info the head surface used?

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+                                             high_res_path,))
+            # Create one from the high res one, which we know we have
+            rr, tris = decimate_surface(self.bem_high_res.surf.rr,
+                                        self.bem_high_res.surf.tris,
+                                        n_triangles=5120)
+            # directly set the attributes of bem_low_res
+            self.bem_low_res = complete_surface_info(dict(rr=rr, tris=tris),
+                                                     copy=False, verbose=False)
+        else:
+            self.bem_low_res = self._read_surface(low_res_path)
```
agramfort, line 1344: This a public attribute that is not documented. Make it private?

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+        # Set MNI points
+        try:
+            fids = get_mni_fiducials(subject, subjects_dir)
+        except Exception:  # some problem, leave at origin
+            self.mni_points = None
```
agramfort, line 1350: Are you actually testing this? I fear it cannot work in this case. I would consider a pure crash and allow users to specify ‘fids’ while if set to ‘auto’ we get is from MNI fsaverage.

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        fid_files = _find_fiducials_files(subject, subjects_dir)
+        if len(fid_files) > 0:
+            fid_filename = fid_files[0].format(subjects_dir=subjects_dir,
+                                               subject=subject)
+        else:
+            fid_filename = None
+        if fid_filename is None or not op.exists(fid_filename):
+            self.fid_points = self.mni_points
+        else:
+            self.fid_points = _fiducial_coords(*read_fiducials(fid_filename))
```
agramfort, line 1364: I would not keep mni_points as attribute unless it’s necessary

Also I would put this logic of guessing the fiducials in a dedicated private function. Function that would only be called if ‘auto' mode is used

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            else:
+                try:
+                    bem = read_surface(filename, return_dict=True)[2]
+                    bem['rr'] *= 1e-3
+                    complete_surface_info(bem, copy=False)
+                except Exception:
+                    raise ValueError(
+                        "Error loading surface from %s (see "
+                        "Terminal for details)." % filename)
+        return bem
```
agramfort, line 1383: This function does not use self so it can be made a private function outside of the class

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        return self
+
+    def omit_hsp_points(self, distance):
+        """Exclude head shape points that are far away from the MRI head.
+
+        Parameters
+        ----------
+        distance : float
+            Exclude all points that are further away from the MRI head than
+            this distance. A value of distance <= 0 excludes nothing.
```
agramfort, line 1805: Please specific the unit

mne/coreg.py
```diff
@@ -1245,3 +1248,632 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        self._update_params(force_update_omitted=True)
+        return self
+
+    def compute_dig_head_distances(self):
+        """Compute Euclidean distance between the head-mri points.
+
+        Returns
+        -------
+        dist : float
+            The distance between the head-mri points.
```
agramfort, line 1832: It’s the distance of the head shape points to the MRI skin surface. Please rephrase

mne/coreg.py
GuillaumeFavelier: I tried 235f8b89fc1447a0806e7bf688922d5998bd4069 but I'm not 100% sure this is what you mean, feel free to take a look

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    ----------
+    info : instance of Info
+        The measurement info.
+    subject : str
+        Name of the subject the data are defined for.
+    subjects_dir : path-like
+        Path to MRI subjects directory.
+    fids : list | str
+        List of fiducials or if set to 'auto', the fiducials are initialized
+        automatically. Defaults to 'auto'.
```
agramfort, line 1283:
```suggestion
        automatically using fiducials defined in MNI template. Defaults to 'auto'.
```

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+    Parameters
+    ----------
+    info : instance of Info
+        The measurement info.
+    subject : str
+        Name of the subject the data are defined for.
+    subjects_dir : path-like
+        Path to MRI subjects directory.
+    fids : list | str
```
agramfort, line 1281: I would prefer a dict than a list. Keys could be lpa, rpa and nation. You should also specify in the doc string
In what coordinate systems it should be defined and in which unit.

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+
+        Parameters
+        ----------
+        scale_mode : None | str
+            The scale mode can be 'uniform', '3-axis' or disabled.
+            Defaults to None.
+
+            * 'uniform': 1 scale factor is recovered.
+            * '3-axis': 3 scale factors are recovered.
+            * None: no unity scale factor is recovered.
```
agramfort, line 1466:
```suggestion
            * None: do not scale the MRI.
```

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        self._update_params(rot=np.array(rot))
+        return self
+
+    def set_translation(self, tra):
+        """Set the translation parameter.
+
+        Parameters
+        ----------
+        tra : array, shape (3,)
+            The translation parameter.
```
agramfort, line 1515: Can you specify the unit?

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        self._update_params(self._rotation, self._translation, self._scale)
+        return self
+
+    def set_rotation(self, rot):
+        """Set the rotation parameter.
+
+        Parameters
+        ----------
+        rot : array, shape (3,)
+            The rotation parameter.
```
agramfort, line 1499: Can you specify the unit?

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+            angle, move, scale = self._changes
+            if angle <= self._icp_angle and move <= self._icp_distance and \
+                    all(scale <= self._icp_scale):
+                break
+        if verbose:
+            self._log_dig_mri_distance()
+        return self
+
+    def omit_hsp_points(self, distance):
+        """Exclude head shape points that are far away from the MRI head.
```
agramfort, line 1804: You need to be consistent in docstirngs. Above you only said HSP. I would suggest to avoid as much as possible acronyms.

mne/coreg.py
```diff
@@ -1245,3 +1248,637 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        n_excluded = np.sum(~mask)
+        logger.info("Coregistration: Excluding %i head shape points with "
+                    "distance >= %.3f m.", n_excluded, distance)
+        # set the filter
+        self._extra_points_filter = mask
+        self._update_params(force_update_omitted=True)
+        return self
+
+    def compute_dig_head_distances(self):
+        """Compute Euclidean distance between the head-mri points.
```
agramfort, line 1832: Please also rephrase here. Not just 3 lines below

tutorials/forward/25_automated_coreg.py
```diff
@@ -0,0 +1,62 @@
+"""
+=========================================
+Use automated approach to coregistration
+=========================================
+
+This example shows how to use the coregistration functions to perform an
+automated MEG-MRI coregistration.
+
+.. warning:: The quality of the coregistration depends heavily upon the
+             quality of the head shape collected during subject prepration and
```
agramfort, line 10:
```suggestion
             quality of the head shape points (HSP) collected during subject prepration and
```

agramfort (commented):
@jhouck can you also suggest some wording improvements in the example? Can you also test this branch now to confirm it works as it should? 

We’re getting there !

agramfort:
did you add a test for the fids parameter?

it's close to ready from my end. Maybe @larsoner wants to have a look before we merge.

I think you can safely start drafting the GUI on top of this class now @GuillaumeFavelier

jhouck:
> @jhouck can you also suggest some wording improvements in the example? Can you also test this branch now to confirm it works as it should?
> 
> We’re getting there !

@agramfort sure! I'm out of the office until Friday and can have a look then, if that's not too late.

mne/coreg.py
```diff
@@ -1245,3 +1248,642 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    info : instance of Info
+        The measurement info.
+    subject : str
+        Name of the subject the data are defined for.
+    subjects_dir : path-like
+        Path to MRI subjects directory.
+    fids : dict | str
+        List of fiducials or if set to 'auto', the fiducials are initialized
+        automatically using fiducials defined in MNI template. Defaults to
+        'auto'.
```
agramfort, line 1284: @GuillaumeFavelier can you be explicit in which unit and coordinate systems the fids (which I would actually rename fiducials) should be? Should it be provided in MRI Surface RAS ? in head ? in MRI voxel?

agramfort:
I think it is now good for review from my end.

@GuillaumeFavelier see my bug fixes in setup_fiducials. You were actually always reading
fiducials from disk and the dict path was broken. I also now allow to pass directly the output
of read_fiducials function.

@larsoner you have time for review?

mne/coreg.py
```diff
@@ -1245,3 +1248,656 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        if fids == 'estimated':
+            logger.info('Estimating fiducials from fsaverage.')
+            fids = get_mni_fiducials(self._subject, self._subjects_dir)
+
+        if isinstance(fids, list):
+            fid_coords = _fiducial_coords(fids)
+        else:
+            assert isinstance(fids, dict)
+            fid_coords = np.array([fids['lpa'], fids['nasion'], fids['rpa']],
+                                  dtype=float) / 1e3  # convert to meters
```
larsoner, line 1402: Conversion to meters should not be necessary, all units in MNE data structures should be SI (here, meters). This seems like a bug, especially since:
```
>>> import mne
>>> mne.io.read_fiducials('mne/data/fsaverage/fsaverage-fiducials.fif')[0][-1]['r'])
array([ 0.08436285, -0.02850276, -0.04127743], dtype=float32)
```
I'll change this to assume and be documented as `m`

mne/coreg.py
```diff
@@ -1245,3 +1248,656 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+    Internal computation quantities parameters are in the following units:
+    - rotation are in radians
+    - translation are in m
+    - scale are in scale proportion
+    """
+
+    def __init__(self, info, subject, subjects_dir, fiducials='auto'):
+        self._info = info
+        self._subject = subject
+        self._subjects_dir = subjects_dir
```
larsoner, line 1309: Not enough input validation / `None` default here, will push

mne/coreg.py
```diff
@@ -1245,3 +1248,656 @@ def _scale_xfm(subject_to, xfm_fname, mri_name, subject_from, scale,
...
+        return self
+
+    def compute_dig_head_distances(self):
+        """Compute distance between head shape points and MRI skin surface.
+
+        Returns
+        -------
+        dist : array, shape (n_points,)
+            The distance of the head shape points to the MRI skin surface.
+        """
```
larsoner, line 1857: Why not use `dig_mri_distances` here? Will refactor.

larsoner: Ohh I see why, because it's going to be slower to use the `surface` one. Hmm... I'll get rid of the uses of `dig_mri_distances` then

larsoner (approved):
LGTM after a refactoring, @GuillaumeFavelier just make sure you're happy with https://github.com/mne-tools/mne-python/pull/9516/commits/e13925e8e44f950d0c42d0abfb63f0f2c8eeb458

agramfort:
example looks good https://32845-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/forward/25_automated_coreg.html but the Coregistration class is not clickable. I'll look into this.

agramfort:
I am clueless why the Coregistration instance / import is not clickage in https://32845-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/forward/25_automated_coreg.html?highlight=coregistration

although I don't see a pb in the rendered doc in https://32845-1301584-gh.circle-artifacts.com/0/dev/generated/mne.coreg.Coregistration.html?highlight=coregistration#mne.coreg.Coregistration

@larsoner @drammock any hint?

larsoner:
It's a weird bug where having something documented at the level of `mne.coreg.Coregistration` but available at a higher level `mne.Coregistration` confused sphinx-gallery. I'll remove it from `mne/__init__.py` since it seems like it was meant to live just in the `mne.coreg` namespace (and I agree it's nicer than having it at the root level)

agramfort:
thx @GuillaumeFavelier !!!

PR #9519, opened 2021-07-01
Add exclude parameter to raw.plot_psd()

### LINKED ISSUE #9441: Plot all channels with raw.plot_psd
We have already discussed and fixed the problem that it was impossible to plot the PSD of bad channels (#6637, #6678). However, I find the solution still not ideal. I frequently need to plot PSDs of all channels, and the only way to do that is to pass all channel indices or names to `picks`. I'd prefer a string value, but unfortunately `"all"` and `"data"` are already taken and just include good channels. There's also the default third option (`None`), which seems to be equal to `"all"`.

Does anyone have a good suggestion how to solve this issue?

cbrnr:
Fixes #9441.

This adds an `exclude` parameter to `raw.plot_psd()`, which makes it possible to plot the PSD of *all* channels (including bads) using `raw.plot_psd(exclude=[])`.

Questions/To Do:
- [x] Add changelog in API or new section?
- [x] Add `.. versionadded :: 0.24.0` tag?
- [x] Add note somewhere (docstring?) to show how to plot the PSD for all channels (including bads)?

### DIFF
--- a/mne/viz/raw.py
+++ b/mne/viz/raw.py
@@ -421,7 +421,7 @@ def plot_raw_psd(raw, fmin=0, fmax=np.inf, tmin=None, tmax=None, proj=False,
         are excluded. Pass an empty list to plot all channels (including
         channels marked "bad", if any).
 
-        ..versionadded:: 0.24.0
+        .. versionadded:: 0.24.0
     %(verbose)s
 
     Returns
--- a/mne/viz/raw.py
+++ b/mne/viz/raw.py
@@ -373,7 +373,7 @@ def plot_raw_psd(raw, fmin=0, fmax=np.inf, tmin=None, tmax=None, proj=False,
                  area_mode='std', area_alpha=0.33, dB=True, estimate='auto',
                  show=True, n_jobs=1, average=False, line_alpha=None,
                  spatial_colors=True, sphere=None, window='hamming',
-                 verbose=None):
+                 exclude='bads', verbose=None):
     """%(plot_psd_doc)s.
 
     Parameters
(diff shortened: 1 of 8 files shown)


mne/viz/raw.py
```diff
@@ -416,6 +416,9 @@ def plot_raw_psd(raw, fmin=0, fmax=np.inf, tmin=None, tmax=None, proj=False,
     %(window-psd)s
 
         .. versionadded:: 0.22.0
+    exclude : list of str | 'bads'
+        Channels names to exclude from being shown. If 'bads', the
+        bad channels are excluded.
```
drammock, line 421:
```suggestion
        bad channels are excluded. Pass an empty list to plot all channels
        (including channels marked "bad", if any).
```

drammock (approved):
> Add changelog in API or new section?

I doubt this will break old code, so I'd put it as a new feature.

> Add .. versionadded :: 0.24.0 tag?

sure, why not?

> Add note somewhere (docstring?) to show how to plot the PSD for all channels (including bads)?

See my suggestion

mne/viz/raw.py
cbrnr: Thanks, included in latest commit.

agramfort (approved):
diff LGTM !

no objection to merge once CIs are green.

cbrnr:
@drammock any idea what is wrong with the docstring? It ends with a ".", so I don't understand the error message.

drammock:
> @drammock any idea what is wrong with the docstring? It ends with a ".", so I don't understand the error message.

missing space between `..` and `versionadded`

cbrnr:
Ultraslow is timing out, otherwise looking good.

agramfort:
thx @cbrnr !

cbrnr:
@larsoner there are two unused imports in `mne/preprocessing/nirs/_scalp_coupling_index.py` introduced in #9522 - do you want me to fix it as part of this PR or do you want a separate one?

larsoner:
Just pushed a hotfix

cbrnr:
All green, @drammock or @larsoner feel free to merge.

drammock (approved):

PR #9522, opened 2021-07-02
BUG: Fix bug with optical_density

larsoner:
cc @rob-luke I think the protection against negative + zero values was not quite done safely enough

### DIFF
--- a/mne/preprocessing/nirs/_optical_density.py
+++ b/mne/preprocessing/nirs/_optical_density.py
@@ -8,18 +8,20 @@ import numpy as np
 
 from ...io import BaseRaw
 from ...io.constants import FIFF
-from ...utils import _validate_type, warn
+from ...utils import _validate_type, warn, verbose
 from ...io.pick import _picks_to_idx
 from ..nirs import _channel_frequencies, _check_channels_ordered
 
 
-def optical_density(raw):
+@verbose
+def optical_density(raw, *, verbose=None):
     r"""Convert NIRS raw data to optical density.
 
     Parameters
     ----------
     raw : instance of Raw
         The raw data.
+    %(verbose)s
 
     Returns
     -------
@@ -32,7 +34,6 @@ def optical_density(raw):
         raw.info, np.unique(_channel_frequencies(raw.info, nominal=True)))
 
     picks = _picks_to_idx(raw.info, 'fnirs_cw_amplitude')
-    data_means = np.mean(raw.get_data(), axis=1)
 
     # The devices measure light intensity. Negative light intensities should
     # not occur. If they do it is likely due to hardware or movement issues.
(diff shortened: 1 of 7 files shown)


mne/preprocessing/nirs/_optical_density.py
```diff
@@ -32,20 +34,26 @@ def optical_density(raw):
...
-        np.log(raw._data[ii], out=raw._data[ii])
-        raw._data[ii] *= -1
-        raw.info['chs'][ii]['coil_type'] = FIFF.FIFFV_COIL_FNIRS_OD
+        min_ = np.inf
+        for pi in picks:
+            np.abs(raw._data[pi], out=raw._data[pi])
+            min_ = min(min_, raw._data[pi].min() or min_)
+        # avoid == 0
+        for pi in picks:
+            np.maximum(raw._data[pi], min_, out=raw._data[pi])
```
rob-luke, line 50: Why do you set it to _min and not something like eps?

mne/preprocessing/nirs/_scalp_coupling_index.py
```diff
@@ -59,7 +59,10 @@ def scalp_coupling_index(raw, l_freq=0.7, h_freq=1.5,
 
     sci = np.zeros(picks.shape)
     for ii in picks[::2]:
-        c = np.corrcoef(filtered_data[ii], filtered_data[ii + 1])[0][1]
+        with np.errstate(invalid='ignore'):
+            c = np.corrcoef(filtered_data[ii], filtered_data[ii + 1])[0][1]
+        if not np.isfinite(c):  # someone had std=0
```
rob-luke, line 64: Does this mean that the channel is flat? I.e. it's all the same value? If this is the case would it not be better to throw the user a warning and tell them to set it to bad?

mne/preprocessing/nirs/_tddr.py
```diff
@@ -132,6 +132,8 @@ def _TDDR(signal, sample_rate):
         sigma = 1.4826 * np.median(dev)
 
         # Step 3d. Scale deviations by standard deviation and tuning parameter
+        if sigma == 0:
```
rob-luke, line 135: Does this also indicate a flat channel? As above.

rob-luke: Do try and keep the code as close to the original as possible could we add the flat signal check as a block at the front?

rob-luke (commented):
I love that we are now getting down to the gritty details. That means things are broadly working.

Just a few questions. Mainly for my learning.

mne/preprocessing/nirs/_scalp_coupling_index.py
larsoner: Warning could be useful, yeah. This was mostly just laziness since I think/hope the outcome of the algorithm is still to clearly say the channel is bad anyway (e.g., with a SCI of 0). Maybe I should assert this in a test, then a warning isn't necessary? Or maybe we need both warning and sci=0?

rob-luke: You're right, as long as it comes out with a 0 value the result will be the same to the user.

We should probably have a flat signal check as a seperate function check. I think I have seen that around somewhere in MNE already, Ill look.

larsoner: Yes we have a flat marking / annotating function

https://mne.tools/stable/generated/mne.preprocessing.annotate_flat.html

mne/preprocessing/nirs/_tddr.py
larsoner: We could but I don't mind the small diff that's a clear fix here. It ends up being less clean to do it at the beginning I think because it creates redundant calculations

mne/preprocessing/nirs/_optical_density.py
larsoner: `eps` is relative to `1.` and is around 1e-12. Imagine you had data in the femto/1e-15 or smaller range, your `np.finfo(float).eps` would actually be a "large" value relative to this scale.

larsoner:
@rob-luke refactored a tiny bit and added tests, feel free to look again and merge if you're happy / convinced

mne/preprocessing/nirs/_optical_density.py
rob-luke: Thanks @larsoner, turns out I didnt understand eps properly. I wonder how many bugs I have introduced over the years!

rob-luke:
Thanks @larsoner, I'm convinced 😉

PR #9525, opened 2021-07-05
MRG: Add annotations.rename()

### LINKED ISSUE #9523: Add a method to rename annotation descriptions

I find myself repeatedly needing to explain how to rename annotations to end users. I rename annotations using something like...

```python
ann = raw_intensity.annotations
ann.description = [d.replace('1.0', 'Control') for d in ann.description]
ann.description = [d.replace('2.0', 'Tapping/Left') for d in ann.description]
ann.description = [d.replace('3.0', 'Tapping/Right') for d in ann.description]
raw_intensity.set_annotations(ann)
```

Which isn't much code. But the barrier to entry is quite high. It requires that a user understands list comprehensions, which I think is an advanced topic. New users testing out MNE usually first want to load their data, rename triggers, and plot the data. So I suggest we add a method to annotations for renaming descriptions.

Thoughts?

I would add an `annotations.rename(old_name, new_name)` method.

Leave as it is

#### Additional comments
I am happy to write this if others think its worthwhile

rob-luke:
#### Reference issue
Closes #9523

Adds a `annotations.rename()` method and `raw.rename_annotations()`.

#### Additional information
~Currently the code takes the old name and the new name as arguments. An alternative suggestion would be to passing in a mapping like `{"old_name": "new_name", "another_old_name": "fancy_new_name"}`. I’m on the fence about which way to implement it, happy to change if you want.~


cbrnr:
As @drammock suggested, I'd like to see the dict update argument here. I'd also keep the two parameter version for simplicity. If we really want to go overboard, we could also have it accept a function that performs the new mapping (like in `raw.rename_channels`.

However, all of this could also be added later in a separate PR.

drammock:
IIRC `raw.rename_channels()` accepts a mapping, so this should too. I'd also consider aliasing this as `raw.rename_annotations()` because I think it will aid discoverability.

rob-luke:
Excellent. I'll implement option two tomorrow.

rob-luke:
Ready for review. I think the CI failed due to time out and I dont have permission to restart a single job.

mne/annotations.py
```diff
@@ -567,6 +567,50 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
 
+    @verbose
+    def rename(self, mapping, verbose=None):
+        """Rename annotation description(s). Operates inplace.
+
+        Parameters
+        ----------
+        mapping : dict
+            A dictionary mapping the old description to a new description
+            name e.g. {‘1.0’ : ‘Control’, ‘2.0’ : ‘Stimulus’}.
```
agramfort, line 578:
```suggestion
            name e.g. {'1.0' : 'Control', '2.0' : 'Stimulus'}.
```

just to avoid copy-paste mistakes.

mne/io/base.py
```diff
@@ -1923,6 +1923,30 @@ def describe(self, data_frame=False):
...
 
+    @verbose
+    def rename_annotations(self, mapping, verbose=None):
+        """Rename annotation description(s). Operates inplace.
+
+        Parameters
+        ----------
+        mapping : dict
+            A dictionary mapping the old description to a new description
+            name e.g. {‘1.0’ : ‘Control’, ‘2.0’ : ‘Stimulus’}.
```
agramfort, line 1934: idem

agramfort (approved):
besides LGTM

mne/io/base.py
rob-luke:
```suggestion
            name e.g. {'1.0' : 'Control', '2.0' : 'Stimulus'}.
```

mne/annotations.py
```diff
@@ -567,6 +567,50 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+            raise ValueError(
+                "Annotation description(s) in mapping missing from data: "
+                "%s" % np.array(orig_annots)[np.array(missing)])
+    else:
+        raise ValueError("Renaming requires the mapping of descriptions "
+                         "to be provided as a dict. "
+                         f"Instead {type(mapping)} was provided.")
+
+    for map in mapping:
+        desc = [d.replace(map, mapping[map]) for d in desc]
```
drammock, line 611:
```suggestion
    for old, new in mapping.items():
        desc = [d.replace(old, new) for d in desc]
```
avoids shadowing the built-in `map()` function

mne/annotations.py
```diff
@@ -567,6 +567,50 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+        """
+        self.description = _rename_annotations(self.description,
+                                               mapping, verbose)
+        return self
+
+
+@verbose
+def _rename_annotations(desc, mapping, verbose=None):
+
+    if isinstance(mapping, dict):
```
drammock, line 598: this `if isinstance...  else ValueError` pattern could probably be replaced by our `_validate_types` helper.  If you're not up for that refactor, at the very least this should raise a `TypeError` instead.

mne/io/base.py
```diff
@@ -1923,6 +1923,30 @@ def describe(self, data_frame=False):
...
+        self : mne.Annotations
+            The modified Annotations object.
+
+        Notes
+        -----
+        .. versionadded:: 0.24.0
+        """
+        self.annotations.description = _rename_annotations(
+            self.annotations.description, mapping, verbose)
+        return self
```
drammock, line 1948: is there a particular reason not to use the public function here?  E.g.:
```suggestion
        self.annotations.rename(mapping, verbose=verbose)
        return self
```
?  Seems slightly simpler to me / makes it clearer that it's just an alias.

mne/io/base.py
```diff
@@ -1923,6 +1923,30 @@ def describe(self, data_frame=False):
...
+        ----------
+        mapping : dict
+            A dictionary mapping the old description to a new description
+            name e.g. {'1.0' : 'Control', '2.0' : 'Stimulus'}.
+        %(verbose_meth)s
+
+        Returns
+        -------
+        self : mne.Annotations
+            The modified Annotations object.
```
drammock, line 1940: Right now this is inaccurate; it will return the Raw instance, not the annotations instance.  And in fact the raw method *should* return the raw instance IMO, so the docstring should be updated.

rob-luke:
Thanks @cbrnr @agramfort @drammock for the reviews. I think I have addressed all your comments and am ready for another round.

cbrnr:
Have we resolved whether or not to copy this method to a `Raw` method (@agramfort didn't like it whereas @drammock made this suggestions)? Same thing for #9529 (I think these should be consistent whatever we decide). I'm slightly leaning towards *not* duplicating `Annotations` methods in `Raw`.

rob-luke:
I am not sure if they need to be the same. I think setting annotation names is much more common procedure than setting the durations.

drammock:
> I am not sure if they need to be the same. I think setting annotation names is much more common procedure than setting the durations.
> 
> If it has to be consistent then I vote for not duplicating in raw.

I agree 100%.

larsoner:
I would rather not duplicate at the `raw` level. The scope / nesting is better with the `raw.annotations.*` methods rather than `raw.*` operating on `raw.annotations`

rob-luke:
So I remove raw.rename_annotations()?

larsoner:
+1 for me on removing `raw.rename_annotations`

drammock:
> +1 for me on removing `raw.rename_annotations`

I can live with that.  My motivation was for tab-completion discoverability (I'd guess many folks do `raw.<TAB>` but few do `raw.annotations.<TAB>`) but it's not a big deal, as long as we document it well.

mne/tests/test_annotations.py
```diff
@@ -1361,5 +1361,5 @@ def test_annotation_rename():
         raw.rename_annotations({"aaa": "does not exist"})
     with pytest.raises(ValueError, match="[' a']"):
         raw.rename_annotations({" a": "does not exist"})
-    with pytest.raises(ValueError, match="<class 'str'> was provided"):
+    with pytest.raises(TypeError, match="dict, got <class 'str'> instead"):
         raw.rename_annotations("wrong")
```
drammock: It won't increase coverage to test these twice (once as `annot.rename()` and once as `raw.rename_annotations()`) since the latter is just an alias; it should suffice to call the alias *once*.  But I guess that no longer matters since there seems to be consensus to ditch the alias, so these will get removed anyway.

drammock (approved):

rob-luke:
Should be good to go now. I have removed `raw.rename_annotations`, refactored the code, slightly modified `annotations.set_durations` to use `_validate_type`. I also touched the function `utils.`_validate_type`, so please review code again before hitting merge.

doc/changes/latest.inc
```diff
@@ -81,10 +81,11 @@ Enhancements
 
 - Update :func:`mne.preprocessing.realign_raw` with Numpy-recommended polynomial fitting method (:gh:`9514` by `Erica Peterson`_)
 
-- Add :meth:`annotations.set_durations() <mne.Annotations.set_durations>` to set annotation durations (:gh:`9529` by `Robert Luke`_)
-
 - Add ``exclude`` parameter to :meth:`mne.io.Raw.plot_psd` and :meth:`mne.Epochs.plot_psd` (:gh:`9519` by `Clemens Brunner`_)
 
+- Add :meth:`annotations.rename() <mne.Annotations.rename>` to rename annotation descriptions (:gh:`9525` by `Robert Luke`_)
+
+- Add :meth:`annotations.set_durations() <mne.Annotations.set_durations>` to set annotation durations (:gh:`9529` by `Robert Luke`_)
```
cbrnr, line 88:
```suggestion
- Add `mne.Annotations.rename` to rename annotation descriptions (:gh:`9525` by `Robert Luke`_)

- Add `mne.Annotations.set_durations` to set annotation durations (:gh:`9529` by `Robert Luke`_)
```

mne/annotations.py
```diff
@@ -589,6 +589,8 @@ def set_durations(self, mapping, verbose=None):
         -----
         .. versionadded:: 0.24.0
         """
+        _validate_type(mapping, (float, dict))
```
cbrnr, line 592:
```suggestion
        _validate_type(mapping, (int, float, dict))
```

mne/annotations.py
```diff
@@ -610,6 +612,40 @@ def set_durations(self, mapping, verbose=None):
...
 
+    @verbose
+    def rename(self, mapping, verbose=None):
+        """Rename annotation description(s). Operates inplace.
+
+        Parameters
+        ----------
+        mapping : dict
+            A dictionary mapping the old description to a new description
+            name e.g. {'1.0' : 'Control', '2.0' : 'Stimulus'}.
```
cbrnr, line 623:
```suggestion
            A dictionary mapping the old description to a new description,
            e.g. {'1.0' : 'Control', '2.0' : 'Stimulus'}.
```

mne/annotations.py
```diff
@@ -610,6 +612,40 @@ def set_durations(self, mapping, verbose=None):
...
+        self : mne.Annotations
+            The modified Annotations object.
+
+        Notes
+        -----
+        .. versionadded:: 0.24.0
+        """
+        _validate_type(mapping, dict)
+
+        orig_annots = sorted(list(mapping.keys()))
```
cbrnr, line 637:
```suggestion
        orig_annots = sorted(mapping.keys())
```

mne/annotations.py
```diff
@@ -610,6 +612,40 @@ def set_durations(self, mapping, verbose=None):
...
+        missing = [annot not in self.description for annot in orig_annots]
+        if any(missing):
+            raise ValueError(
+                "Annotation description(s) in mapping missing from data: "
+                "%s" % np.array(orig_annots)[np.array(missing)])
```
cbrnr, line 642:
```suggestion
            raise ValueError(f"Annotation description(s) in mapping missing from data: "
                             f"{np.array(orig_annots)[np.array(missing)]}")
```

mne/tests/test_annotations.py
```diff
@@ -1286,6 +1286,57 @@ def test_annotation_ch_names():
...
+        a.rename({" a": "does not exist"})
+    with pytest.raises(TypeError, match="dict, got <class 'str'> instead"):
+        a.rename("wrong")
+    with pytest.raises(TypeError, match="dict, got <class 'list'> instead"):
+        a.rename(["wrong"])
+    with pytest.raises(TypeError, match="dict, got <class 'set'> instead"):
+        a.rename({"wrong"})
+
+
 def test_annotation_duration_setting():
```
cbrnr, line 1340: Can you add a test setting a new integer duration?

cbrnr (changes requested):

mne/annotations.py
cbrnr: It seems now this line is > 79 chars, can you break it up into multiple rows?

rob-luke:
Thanks @cbrnr I hit all your suggestions except one which caused a flake error.

cbrnr:
Thanks @rob-luke, yes, it's difficult to use GitHub's suggest feature because it doesn't show line lengths. But you did replace the percent operator with an f-string, and this was the main thing I wanted to change.

cbrnr (approved):

cbrnr:
Thanks @rob-luke!

Issue #8291, opened 2020-09-22
Create local topographic representations. Don't connect distant sensors in topomap.


rob-luke:
#### Describe the problem
When creating topoplots for montages that have clearly separated clusters of sensors the plotting connects these regions, often creating some visual distortion between regions of interest.

For example in the fNIRS example [here](https://mne.tools/stable/auto_tutorials/preprocessing/plot_70_fnirs_processing.html#compare-tapping-of-left-and-right-hands) I would like the left and right regions repeated. Here is what I would like removed.

![](https://files.gitter.im/536e5d0f048862e761fa0cfe/aQ9n/image.png)

And here is another example with a dataset with 4 seperate regions that I would prefer plotted independently

![image](https://user-images.githubusercontent.com/748691/93946018-8442da80-fd7b-11ea-8a44-8e65697b532a.png)

#### Describe your solution
I am not sure of the optimal solution, but my initial thoughts are I would like to be able to either specify how many independent regions I want plotted. Or perhaps specify a maximum distance between sensors, and anything above this distance is not connected in the plot. 

#### Additional context
I tried varying the existing parameters in the topo plot function, but nothing quite achieves my aim.

agramfort:
I would not use kmeans but
https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html
with single option to find the clusters here.

>

rob-luke:
Thanks for the pointers. As this seems like it will involve digging in to the lower level code I will put this on hold till I have some more time.

In the meantime, if anyone lands here with the same issue I have got around this by plotting each ROI independently and specifying the same axes. 

This is a different dataset and analysis than above, but it ends up looking like 
![image](https://user-images.githubusercontent.com/748691/94377710-0a3d9780-0167-11eb-95f5-3f990395b457.png)

Using something like...
```python
fig, axes = plt.subplots(nrows=1, ncols=1, figsize=(6, 6))
for r in rois:
    raw_subset = raw().pick(picks= rois[r])
    your_plotting_function(raw_subset, axes=axes)
```

larsoner:
You can probably pass `outlines=dict()` or something similar to get it to not re-plot the outlines in the subsequent calls, too. I say we make sure we can avoid re-plotting the head outline somehow, then close, as this is actually a nice solution! +1 for adding this to an fNIRS tutorial at some point, too.

rob-luke:
[...]

Issue #9511, opened 2021-06-28
evoked.plot_topomap extrapolation can fail when few electrodes present


drammock:
from the forum: https://mne.discourse.group/t/erp-plot-topomap-nan-to-integer-error/3317

extrapolate = 'box' or 'local' both work, but the default ('head') fails (at some point in the code we call `np.arcsin` and its argument is out of the domain for that function, thus introducing a NaN).

Link to example data: https://www.dropbox.com/l/AABzZrKRNcYbG7Z9m0OexH8WeBVSXRunqMc

```py
import numpy as np
import mne

exp3 = mne.io.read_raw_edf('exp3.edf', preload=True)
exp3.filter(1,30, method='iir')
events_exp3 = mne.make_fixed_length_events(exp3, duration=1.0)

ten_five = mne.channels.make_standard_montage('standard_1005')
exp3.set_montage(ten_five)

epochs_exp3 = mne.Epochs(
    exp3, events_exp3, event_id={'resting_state': 1}, tmin=-0.3, tmax=0.7,
    preload=True)

rest_exp3 = epochs_exp3['resting_state'].average()
fig = rest_exp3.plot_topomap()
```
Traceback:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-4-7729af823da6> in <module>
----> 1 fig = rest_exp3.plot_topomap()

/opt/mne/python/mne/evoked.py in plot_topomap(self, times, ch_type, vmin, vmax, cmap, sensors, colorbar, scalings, units, res, size, cbar_fmt, time_unit, time_format, proj, show, show_names, title, mask, mask_params, outlines, contours, image_interp, average, axes, extrapolate, sphere, border, nrows, ncols)
    443                      axes=None, extrapolate=_EXTRAPOLATE_DEFAULT, sphere=None,
    444                      border=_BORDER_DEFAULT, nrows=1, ncols='auto'):
--> 445         return plot_evoked_topomap(
    446             self, times=times, ch_type=ch_type, vmin=vmin,
    447             vmax=vmax, cmap=cmap, sensors=sensors, colorbar=colorbar,

/opt/mne/python/mne/viz/topomap.py in plot_evoked_topomap(***failed resolving arguments***)
   1714         adjust_for_cbar = colorbar and ncols is not None and idx >= ncols - 1
   1715         ax_idx = idx + 1 if adjust_for_cbar else idx
-> 1716         tp, cn, interp = _plot_topomap(
[...]

PR #9528, opened 2021-07-05
FIX: fix topo plot when large channel distance gives rise to invalid values for arcsin

### LINKED ISSUE #9526: Crash topomap with only 4 channels from 10-5 system
to replicate the problem:

```
import numpy as np
import mne

ch_names = ['TP9', 'AF7', 'AF8', 'TP10']

info = mne.create_info(ch_names, 100, ch_types='eeg')
evoked = mne.EvokedArray(np.random.randn(4, 10), info)
ten_five = mne.channels.make_standard_montage("standard_1005")
evoked.set_montage(ten_five)
evoked.plot_topomap()
```

now if I do:

`evoked.plot_topomap(sphere=0.14)`

it works.

mmagnuski:
Fixes #9526 and #9511. I'll add the example posted by @agramfort as a test later.


rob-luke:
Could this possibly fix/close #8291 too?

mmagnuski:
Unfortunatelly #8291 is much more complex, this PR is a simple bug-fix. For #8291 we would first need some logic for detecting sensor groups.

rob-luke:
Thanks for the clarification @mmagnuski, I was just hoping for a magic fix 😄

agramfort:
@mmagnuski can you just a what's new entry? thx for the quick reaction to fix this one !

mmagnuski:
No problem! what's new entry added, waiting for the CIs.

mmagnuski:
One CI task (Azure Main Ultraslow) timed out, rest is green.

mne/viz/tests/test_topomap.py
```diff
@@ -557,6 +557,24 @@ def test_plot_topomap_bads():
...
+    """
+    ch_names = ['TP9', 'AF7', 'AF8', 'TP10']
+
+    info = create_info(ch_names, 100, ch_types='eeg')
+    evoked = EvokedArray(np.random.randn(4, 10) * 1e-6, info)
+    ten_five = make_standard_montage("standard_1005")
+    evoked.set_montage(ten_five)
+
+    evoked.plot_topomap(sphere=0.05, res=8)
+    plt.close('all')
```
larsoner, line 575: FYI this is no longer necessary as we have a magic autouse mpl fixture that calls `plt.close`

https://github.com/mne-tools/mne-python/blob/2e200aa7778a1462b4da6ce7bf4253b99cde3d6a/mne/conftest.py#L131-L137

larsoner:
Thanks @mmagnuski !

mne/viz/tests/test_topomap.py
mmagnuski: ok, good, I just followed what other tests there used (and actually thought about a decorator like that), I can change this in the next PR.

mmagnuski:
Thanks @larsoner @agramfort! :rocket:

mne/viz/tests/test_topomap.py
larsoner: I am slowly eliminating the ones at the end of functions as I see them. (The ones in the middle of tests are still useful to prevent too many MPL windows from popping up, eating memory, and possibly emitting warnings about 20+ windows.)

PR #9529, opened 2021-07-06
MRG: Add set_durations() function for annotations

rob-luke:
This is an issue as PR. What do people think?

#### Reference issue
Following #9525

I find it a common operation to set the duration of stimuli annotations. Hence I suggest a convenient function to do this.

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -589,31 +589,26 @@ class Annotations(object):
         -----
         .. versionadded:: 0.24.0
         """
-        return _set_duration(self, mapping, verbose)
+        if isinstance(mapping, dict):
+            orig_annots = sorted(list(mapping.keys()))
+            missing = [key not in self.description for key in orig_annots]
+            if any(missing):
+                raise ValueError(
+                    "Annotation description(s) in mapping missing from data: "
+                    "%s" % np.array(orig_annots)[np.array(missing)])
+            for map in mapping:
+                map_idx = [desc == map for desc in self.description]
+                self.duration[map_idx] = mapping[map]
 
+        elif isinstance(mapping, (int, float)):
+            self.duration = np.ones(self.description.shape) * mapping
 
-@verbose
-def _set_duration(annot, mapping, verbose=None):
+        else:
+            raise ValueError("Renaming requires the mapping of descriptions "
+                             "to times to be provided as a dict. "
+                             f"Instead {type(mapping)} was provided.")
 
-    if isinstance(mapping, dict):
-        orig_annots = sorted(list(mapping.keys()))
-        missing = [key not in annot.description for key in orig_annots]
-        if any(missing):
-            raise ValueError(
-                "Annotation description(s) in mapping missing from data: "
-                "%s" % np.array(orig_annots)[np.array(missing)])
-        for map in mapping:
-            map_idx = [desc == map for desc in annot.description]
-            annot.duration[map_idx] = mapping[map]
-
-    elif isinstance(mapping, (int, float)):
-        annot.duration = np.ones(annot.description.shape) * mapping
-    else:
-        raise ValueError("Renaming requires the mapping of descriptions "
-                         "to times to be provided as a dict. "
-                         f"Instead {type(mapping)} was provided.")
-
-    return annot
+        return self
 
 
 def _combine_annotations(one, two, one_n_samples, one_first_samp,
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -568,7 +568,7 @@ class Annotations(object):
         return self
 
     @verbose
-    def set_duration(self, mapping, verbose=None):
+    def set_durations(self, mapping, verbose=None):
         """Set annotation durations(s). Operates inplace.
 
         Parameters
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -573,7 +573,7 @@ class Annotations(object):
 
         Parameters
         ----------
-        mapping : dict | number
+        mapping : dict | float | int
             A dictionary mapping the annotation description to a duration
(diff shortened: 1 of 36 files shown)


mne/io/base.py
```diff
@@ -1923,6 +1923,31 @@ def describe(self, data_frame=False):
             msg += f"{cols['max'][i]:>8.2f}"
             print(msg)
 
+    @verbose
+    def set_durations(self, mapping, verbose=None):
```
agramfort, line 1927: I am -1 on adding so many public functions to Raw due to annotations. People should use:

`raw.annotations.set_durations(...)`

mne/annotations.py
```diff
@@ -567,6 +567,54 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
 
+    @verbose
+    def set_durations(self, mapping, verbose=None):
+        """Set annotation durations(s). Operates inplace.
+
+        Parameters
+        ----------
+        mapping : dict | float | int
+            A dictionary mapping the annotation description to a duration
+            e.g. {‘ShortStimulus’ : 3, ‘LongStimulus’ : 12}.
```
agramfort, line 578:
```suggestion
            e.g. ``{'ShortStimulus' : 3, 'LongStimulus' : 12}``.
```
`` are optional but will make it look like code in the rendered doc.

mne/io/base.py
rob-luke: Agreed. I only added as it was suggested for the other PR. I will remove.

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+            missing = [key not in self.description for key in orig_annots]
+            if any(missing):
+                raise ValueError(
+                    "Annotation description(s) in mapping missing from data: "
+                    "%s" % np.array(orig_annots)[np.array(missing)])
+            for map in mapping:
+                map_idx = [desc == map for desc in self.description]
+                self.duration[map_idx] = mapping[map]
+
+        elif isinstance(mapping, (int, float)):
```
agramfort, line 603: this is dangerous as it will fail eg for an np.int32. You should use:

isinstance(mapping, numbers.Number)

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
 
         return self
 
+    @verbose
+    def set_durations(self, mapping, verbose=None):
+        """Set annotation durations(s). Operates inplace.
```
agramfort, line 572: maybe remind somewhere that durations are in seconds.

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+
+        Parameters
+        ----------
+        mapping : dict | float | int
```
agramfort, line 576:
```suggestion
        mapping : dict | float
```
I would say float only as it's seconds

mne/annotations.py
larsoner: For consistency you should use `_is_numeric(mapping)` from `.utils`. Maybe we should just use `numbers.Number` there, too, but we can do that later

mne/annotations.py
larsoner:
```suggestion
        """Set annotation duration(s). Operates inplace.
```

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+        -------
+        self : mne.Annotations
+            The modified Annotations object.
+
+        Notes
+        -----
+        .. versionadded:: 0.24.0
+        """
+        if isinstance(mapping, dict):
+            orig_annots = sorted(list(mapping.keys()))
```
drammock, line 593:
```suggestion
            orig_annots = sorted(list(mapping))
```

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+            missing = [key not in self.description for key in orig_annots]
+            if any(missing):
+                raise ValueError(
+                    "Annotation description(s) in mapping missing from data: "
+                    "%s" % np.array(orig_annots)[np.array(missing)])
+            for map in mapping:
+                map_idx = [desc == map for desc in self.description]
+                self.duration[map_idx] = mapping[map]
```
drammock, line 601: don't shadow the built-in `map()` function

mne/annotations.py
```diff
@@ -567,6 +567,49 @@ def crop(self, tmin=None, tmax=None, emit_warning=False, verbose=None):
...
+
+        elif isinstance(mapping, (int, float)):
+            self.duration = np.ones(self.description.shape) * mapping
+
+        else:
+            raise ValueError("Renaming requires the mapping of descriptions "
+                             "to times to be provided as a dict. "
+                             f"Instead {type(mapping)} was provided.")
```
drammock, line 609: message talks about "renaming" but we're not doing that in this method, we're changing durations.

drammock (changes requested):

rob-luke:
Thanks @agramfort @larsoner @drammock for the review. I think all points have been addressed.

drammock (approved):

drammock:
you have a merge conflict in `latest.inc` @rob-luke

rob-luke:
> you have a merge conflict in `latest.inc` @rob-luke 

Fixed. Thanks

cbrnr:
Usually we don't change the order of our changelog entries though. Not super important here, but just a general thing.

agramfort (approved):

agramfort:
thx @rob-luke

PR #9531, opened 2021-07-06
MRG, BUG: Deal with flats better in maxwell autobad

### LINKED ISSUE #9479: find_bad_channels_maxwell breaks if during some chunk all channels are flat
Hello everyone!

I faced with a problem using `mne.preprocessing.find_bad_channels_maxwell()`. At some point it returns an error:
```
        Interval 103:  510.000 -  514.999
            Flat ( 2): MEG1243 MEG1532
        Interval 104:  515.000 -  519.999
  File "<input>", line 172, in <module>
  File "<decorator-gen-415>", line 24, in find_bad_channels_maxwell
  File "E:\Egor_Levchenko\Software\Miniconda\envs\MCCA-MEG\lib\site-packages\mne\preprocessing\maxwell.py", line 2190, in find_bad_channels_maxwell
    delta = chunk_raw.get_data(these_picks)
  File "<decorator-gen-170>", line 22, in get_data
  File "E:\Egor_Levchenko\Software\Miniconda\envs\MCCA-MEG\lib\site-packages\mne\io\base.py", line 819, in get_data
    picks = _picks_to_idx(self.info, picks, 'all', exclude=())
  File "E:\Egor_Levchenko\Software\Miniconda\envs\MCCA-MEG\lib\site-packages\mne\io\pick.py", line 1058, in _picks_to_idx
    '(%r)' % (orig_picks,))
ValueError: No appropriate channels found for the given picks ([])
```

I checked the code and my data. I think the problem is that my data is corrupted a little bit and I have all flat channels for around  20 seconds. After detecting the flat channels the code takes only `good_meg_picks` which are not in `chunk_flats`. As a result, `these_picks` appears to be empty if all channels are in `chunk_flats`:

https://github.com/mne-tools/mne-python/blob/8b9616dfbc377ce468a6d52b43f7a973eac47cab/mne/preprocessing/maxwell.py#L2167-L2175
https://github.com/mne-tools/mne-python/blob/8b9616dfbc377ce468a6d52b43f7a973eac47cab/mne/preprocessing/maxwell.py#L2187

You can see a snapshot of raw data below:

image

I think this is more a problem of my data, but I guess it should be nice to ignore "all flat" chunks with a warning or something like that. Correct me please if I am wrong and thank you for any comments!

System properties:
```
OS: Windows Server 2012 R2
python=3.7.7
mne=0.22.0
```

larsoner:
Just add a test that things behave how we expect for #9479, and a warning to help people. We could maybe do something smarter than just quit, but I think it makes sense to just point people to `annotate_flat` so that they can fix things there instead. With this code a variant of the minimal example from #9479 works as expected:
```
import mne
raw = mne.io.read_raw_fif('sub-1011_ses-20190430_task-vid2_meg.fif').load_data()
raw.set_channel_types(dict(EEG061='eog', EEG062='eog'))
annot, _ = mne.preprocessing.annotate_flat(raw, bad_percent=100, verbose=True)
raw.set_annotations(annot)
mne.preprocessing.find_bad_channels_maxwell(
    raw, skip_by_annotation=('bad_flat',), verbose=True)
```
Closes #9479

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -93,6 +93,8 @@ Bugs
 
 - Fix bug when computing rank from info for SSS data with only gradiometers or magnetometers (:gh:`9435` by `Alex Gramfort`_)
 
+- Fix bug with `mne.preprocessing.find_bad_channels_maxwell` where all-flat segments could lead to an error (:gh:`9531` by `Eric Larson`_)
+
 - Fix bug with `mne.io.Raw.set_montage` and related functions where the channel coordinate frame was not properly set to head (:gh:`9447` by `Eric Larson`_)
 
 - Fix bug with `mne.io.read_raw_fieldtrip` and `mne.read_epochs_fieldtrip` where channel positions were not set properly (:gh:`9447` by `Eric Larson`_)
--- a/mne/preprocessing/maxwell.py
+++ b/mne/preprocessing/maxwell.py
@@ -2173,6 +2173,14 @@ def find_bad_channels_maxwell(
         chunk_flats = sorted(all_flats)
         these_picks = [pick for pick in good_meg_picks
                        if raw.ch_names[pick] not in chunk_flats]
+        if len(these_picks) == 0:
+            logger.info(f'            Flat ({len(chunk_flats):2d}): <all>')
+            warn('All-flat segment detected, all channels will be marked as '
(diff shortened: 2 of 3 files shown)


agramfort (approved):

agramfort:
Thx @larsoner

PR #9537, opened 2021-07-08
Improve scalings description in Raw.plot()

cbrnr:
I found it rather difficult to understand the `scalings` values used in `Raw.plot()`, because the description does not include any details. @drammock please feel free to revise/improve, I'm not really happy with it but cannot think of anything better right now. I think the key point is to explicitly mention that the visualized range is actually twice the scaling value.

### DIFF
--- a/mne/viz/raw.py
+++ b/mne/viz/raw.py
@@ -78,6 +78,12 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
                  emg=1e-3, ref_meg=1e-12, misc=1e-3, stim=1,
                  resp=1, chpi=1e-4, whitened=1e2)
 
+        A particular scaling value ``s``corresponds to half of the visualized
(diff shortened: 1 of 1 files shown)


mne/viz/raw.py
```diff
@@ -78,6 +78,12 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
                  emg=1e-3, ref_meg=1e-12, misc=1e-3, stim=1,
                  resp=1, chpi=1e-4, whitened=1e2)
 
+        A particular scaling value ``s``corresponds to half of the visualized
```
agramfort, line 81:
```suggestion
        A particular scaling value ``s`` corresponds to half of the visualized
```

agramfort (approved):

wmvanvliet (approved):

agramfort:
thx @cbrnr !

PR #9538, opened 2021-07-08
MRG: Put repeated code in a function

rob-luke:
#### Reference issue
Following #9525 #9523 

I noted some repeated code and put it in a function in utils

#### Additional information
There are lots of checking code already. Does something already exist to do this check?

### DIFF
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1208,7 +1208,7 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        _check_dict_keys(mapping, ch_names, key_description="Channel name(s)",
+        _check_dict_keys(mapping, ch_names, key_description="channel name(s)",
                          valid_key_source="info")
         new_names = [(ch_names.index(ch_name), new_name)
                      for ch_name, new_name in mapping.items()]
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1208,8 +1208,8 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        _check_dict_keys(mapping, ch_names, dict_name="Channel name(s)",
-                         valid_name="info")
+        _check_dict_keys(mapping, ch_names, user_opt_name="Channel name(s)",
+                         validator_name="info")
         new_names = [(ch_names.index(ch_name), new_name)
                      for ch_name, new_name in mapping.items()]
     elif callable(mapping):
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -22,7 +22,7 @@ from ..defaults import HEAD_SIZE_DEFAULT, _handle_default
 from ..transforms import _frame_to_str
 from ..utils import (verbose, logger, warn,
                      _check_preload, _validate_type, fill_doc, _check_option,
-                     _get_stim_channel, _check_fname)
+                     _get_stim_channel, _check_fname, _check_dict_keys)
 from ..io.compensator import get_current_comp
 from ..io.constants import FIFF
 from ..io.meas_info import (anonymize_info, Info, MontageMixin, create_info,
@@ -1208,11 +1208,7 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        orig_names = sorted(list(mapping.keys()))
-        missing = [orig_name not in ch_names for orig_name in orig_names]
-        if any(missing):
-            raise ValueError("Channel name(s) in mapping missing from info: "
-                             "%s" % np.array(orig_names)[np.array(missing)])
+        _check_dict_keys(mapping, ch_names, dict_name="Channel name(s)", valid_name="info")
         new_names = [(ch_names.index(ch_name), new_name)
                      for ch_name, new_name in mapping.items()]
     elif callable(mapping):
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1208,7 +1208,8 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        _check_dict_keys(mapping, ch_names, dict_name="Channel name(s)", valid_name="info")
+        _check_dict_keys(mapping, ch_names, dict_name="Channel name(s)",
+                         valid_name="info")
         new_names = [(ch_names.index(ch_name), new_name)
                      for ch_name, new_name in mapping.items()]
     elif callable(mapping):
--- a/mne/channels/channels.py
+++ b/mne/channels/channels.py
@@ -1208,8 +1208,8 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        _check_dict_keys(mapping, ch_names, user_opt_name="Channel name(s)",
-                         validator_name="info")
+        _check_dict_keys(mapping, ch_names, key_description="Channel name(s)",
+                         valid_key_source="info")
         new_names = [(ch_names.index(ch_name), new_name)
                      for ch_name, new_name in mapping.items()]
     elif callable(mapping):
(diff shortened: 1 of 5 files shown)


mne/utils/check.py
```diff
@@ -562,6 +562,40 @@ def _check_depth(depth, kind='depth_mne'):
...
+    ------
+    ValueError
+        When the key of the dict is not one of the valid options.
+
+    Returns
+    -------
+    user_dict
+        When the keys are deemed acceptable the dictionary is returned.
+    """
+    sorted_dict = sorted(list(user_dict))
```
larsoner, line 589: `sorted` iterates so no need to do both
```suggestion
    sorted_dict = sorted(user_dict)
```

mne/utils/check.py
```diff
@@ -562,6 +562,40 @@ def _check_depth(depth, kind='depth_mne'):
...
+    missing = [val not in valid_keys for val in sorted_dict]
```
larsoner, line 590:
```suggestion
    missing = [val for val in sorted_dict if val not in valid_keys]
```

mne/utils/check.py
```diff
@@ -562,6 +562,40 @@ def _check_depth(depth, kind='depth_mne'):
...
+    if any(missing):
```
larsoner, line 591:
```suggestion
    if len(missing):
```

mne/utils/check.py
```diff
@@ -562,6 +562,40 @@ def _check_depth(depth, kind='depth_mne'):
...
+        raise ValueError(
+            f"{dict_name} is missing from {valid_name}: "
+            f"{np.array(sorted_dict)[np.array(missing)]}")
```
larsoner, line 594:
```suggestion
            f"{missing}")
```

larsoner (commented):
Thanks, this is helpful / DRY and AFAIK we don't have anything like it so +1 from me

mne/utils/check.py
drammock: Not a great variable name choice here, as it's no longer a dict.

Also this might be cleaner / faster to read if this line and next were done using `set()`

drammock (commented):
I don't know of an  existing check function for valid keys, so no objection to adding this one

mne/utils/check.py
larsoner: I thought about `set` but it shouldn't be needed since keys of dicts should be unique anyway

drammock: True, but I thought `set(user_dict) - set(valid_keys)` would be cleaner than these 2 lines. Not a blocker though, just a nitpick. I feel that `set()` is underutilized 😆

larsoner: Ahh yes that would work and I agree it's cleaner

cbrnr: I would also use the set difference.

mne/utils/check.py
```diff
@@ -562,6 +562,40 @@ def _check_depth(depth, kind='depth_mne'):
...
 
 
+def _check_dict_keys(user_dict, valid_keys,
+                     dict_name="Channel name(s)", valid_name="info"):
+    """Check that the keys in dictionary are valid against a set list.
+
+    Return the input dictionary if it is valid,
+    otherwise raise a ValueError with a readable error message.
+
+    Parameters
```
cbrnr, line 572: `dict_name` and `valid_name` are missing here. Furthermore, I have trouble associating these names with their function - can we find better names? Or can we not just issue a generic error message, in which case we could drop these parameters?

rob-luke: I'm open to alternative suggestions

larsoner: I'd rather keep the more specific names, it makes for better readability at the user end and at least for `_check_option` and `_validate_type` does not seem to add too much burden during development

larsoner: (but I agree the variable names used here could be clearer, like maybe `user_options` and `allowed_options` or something would be clearer)

cbrnr: But what does this error mean?

```
Channel name(s) is missing from info: ...
```

Do you mean:

```
Channel name(s) missing in info: ...
```

?

mne/utils/check.py
rob-luke: Using set now, thanks for the tip, I didnt know that one.

mne/utils/check.py
rob-luke: I have changed the variable names and error text to suggested values

rob-luke:
This is ready for another review.

mne/utils/check.py
```diff
@@ -562,6 +562,43 @@ def _check_depth(depth, kind='depth_mne'):
...
+    ----------
+    user_options : dict
+        The name of the parameter to check. This is used in the error message.
+    allowed_options : list
+        All possible valid key names.
+    user_opt_name : str
+        Meaningful name of the items stored in the input dictionary.
+    validator_name : str
+        Meaningful name of the values against which the keys are
+        validated against.
```
drammock, line 582: this doesn't look right / I find this hard to understand.  Suggestion:
```suggestion
    mapping : dict
        The user-provided dict whose keys we want to check.
    valid_keys : iterable
        The valid keys.
    key_description : str
        Description of the keys in ``mapping``, e.g., "channel name(s)" or
        "annotation(s)".
    valid_key_source : str
        Description of the ``valid_keys`` source, e.g., "info dict" or
        "annotations in the data".
```

Note: `valid_keys` could even be a dict, and it would still work. No need to restrict to `list`.

mne/utils/check.py
```diff
@@ -562,6 +562,43 @@ def _check_depth(depth, kind='depth_mne'):
...
+
+    Raises
+    ------
+    ValueError
+        When the key of the dict is not one of the valid options.
```
drammock, line 587: we don't usually include a `Raises` section in our docstrings

mne/utils/check.py
```diff
@@ -562,6 +562,43 @@ def _check_depth(depth, kind='depth_mne'):
...
+
+    Returns
+    -------
+    user_dict
+        When the keys are deemed acceptable the dictionary is returned.
```
drammock, line 592: if you accept my proposed argument renaming, then this becomes
```suggestion
    mapping
        If all keys are valid the input dict is returned unmodified.
```

mne/utils/check.py
```diff
@@ -562,6 +562,43 @@ def _check_depth(depth, kind='depth_mne'):
...
+    """
+    missing = set(user_options) - set(allowed_options)
+    if len(missing):
+        raise ValueError(
+            f"{user_opt_name} missing in {validator_name}: {missing}")
```
drammock, line 597:
```suggestion
        _is = 'are' if len(missing) > 1 else 'is'
        msg = (f'Invalid {key_description} {missing} {_is} not present in '
               f'{valid_key_source}')
        raise ValueError(msg)
```

drammock (commented):
some suggestions to improve the param names and error message

mne/utils/check.py
drammock: (you will need to update the tests accordingly, if you accept this rewrite)

rob-luke:
Ops I forgot about this PR. I completely agree with drammock and will implement his changes in the coming days.

larsoner:
Re-titled so that nobody accidentally merges in the meantime

rob-luke:
Ok I think I addressed all comments. @larsoner, ready for review again. Thanks

mne/utils/check.py
```diff
@@ -562,6 +562,42 @@ def _check_depth(depth, kind='depth_mne'):
     return _handle_default(kind, depth)
 
 
+def _check_dict_keys(mapping, valid_keys,
+                     key_description="Channel name(s)",
+                     valid_key_source="info"):
```
drammock, line 567: it's not obvious to me that "Channel name(s)" and "info" are sensible defaults here...  is that far-and-away the most widespread usecase?  To me it seems this is a pretty general-purpose function and those should be required args with no default value.  @rob-luke feel free to convince me otherwise... at the very least, if we keep the default values, "Channel name(s)" should start with a lowercase "c" (as it gets inserted mid-sentence in the error message)
```suggestion
                     key_description, valid_key_source):
```

mne/channels/channels.py
```diff
@@ -1208,11 +1208,8 @@ def rename_channels(info, mapping, allow_duplicates=False, verbose=None):
 
     # first check and assemble clean mappings of index and name
     if isinstance(mapping, dict):
-        orig_names = sorted(list(mapping.keys()))
-        missing = [orig_name not in ch_names for orig_name in orig_names]
-        if any(missing):
-            raise ValueError("Channel name(s) in mapping missing from info: "
-                             "%s" % np.array(orig_names)[np.array(missing)])
+        _check_dict_keys(mapping, ch_names, key_description="Channel name(s)",
```
drammock, line 1211:
```suggestion
        _check_dict_keys(mapping, ch_names, key_description="channel name(s)",
```

drammock (approved):
this looks great, only one question/suggestion

larsoner:
Thanks @rob-luke !

PR #9539, opened 2021-07-08
MRG: Fix return type for raw.annotations.rename and add to tutorial

rob-luke:
#### Reference issue
Following #9525 #9523

Demonstrate new annotation name and duration setting in a tutorial.

### DIFF
--- a/mne/tests/test_annotations.py
+++ b/mne/tests/test_annotations.py
@@ -1289,6 +1289,7 @@ def test_annotation_ch_names():
 def test_annotation_rename():
     """Test annotation renaming works."""
     a = Annotations([1, 2, 3], [5, 5, 8], ["a", "b", "c"])
+    assert isinstance(a.description, np.ndarray)
     assert len(a) == 3
     assert "a" in a.description
     assert "b" in a.description
@@ -1297,6 +1298,7 @@ def test_annotation_rename():
 
     a = Annotations([1, 2, 3], [5, 5, 8], ["a", "b", "c"])
     a.rename({"a": "new_name"})
+    assert isinstance(a.description, np.ndarray)
     assert len(a) == 3
     assert "a" not in a.description
     assert "new_name" in a.description
@@ -1314,6 +1316,7 @@ def test_annotation_rename():
 
     a = Annotations([1, 2, 3], [5, 5, 8], ["a", "b", "c"])
     a.rename({"b": "new_name", "c": "new name c"})
+    assert isinstance(a.description, np.ndarray)
     assert len(a) == 3
     assert "b" not in a.description
     assert "new_name" in a.description
@@ -1340,6 +1343,7 @@ def test_annotation_rename():
 def test_annotation_duration_setting():
     """Test annotation duration setting works."""
     a = Annotations([1, 2, 3], [5, 5, 8], ["a", "b", "c"])
+    assert isinstance(a.duration, np.ndarray)
     assert len(a) == 3
     assert a.duration[0] == 5
     assert a.duration[2] == 8
@@ -1358,6 +1362,7 @@ def test_annotation_duration_setting():
     assert a.duration[0] == 5
     assert a.duration[2] == 8
     a.set_durations(7.2)
+    assert isinstance(a.duration, np.ndarray)
     assert a.duration[0] == 7.2
     assert a.duration[2] == 7.2
     a.set_durations(2)
--- a/tutorials/preprocessing/70_fnirs_processing.py
+++ b/tutorials/preprocessing/70_fnirs_processing.py
@@ -34,7 +34,7 @@ raw_intensity.load_data()
 # stored as annotations. Second, we include information about the duration of
 # each stimulus, which was 5 seconds for all conditions in this experiment.
 # Third, we remove the trigger code 15, which signaled the start and end
-# of the experiment and is not relevant to analysis.
+# of the experiment and is not relevant to our analysis.
 
 raw_intensity.annotations.set_durations(5)
 raw_intensity.annotations.rename({'1.0': 'Control',
--- a/tutorials/preprocessing/70_fnirs_processing.py
+++ b/tutorials/preprocessing/70_fnirs_processing.py
@@ -31,9 +31,9 @@ raw_intensity.load_data()
 # ------------------------------------------------
 #
 # First, we attribute more meaningful names to the trigger codes which are
-# stored as annotations. Further, we include information about the duration of
+# stored as annotations. Second, we include information about the duration of
 # each stimulus, which was 5 seconds for all conditions in this experiment.
-# Further, we remove the trigger code 15, which signaled the start and end
+# Third, we remove the trigger code 15, which signaled the start and end
 # of the experiment and is not relevant to analysis.
 
 raw_intensity.annotations.set_durations(5)
@@ -41,7 +41,7 @@ raw_intensity.annotations.rename({'1.0': 'Control',
                                   '2.0': 'Tapping/Left',
                                   '3.0': 'Tapping/Right'})
 raw_intensity.annotations.delete(
-    np.where([d == '15.0' for d in raw_intensity.annotations.description]))
+    raw_intensity.annotations.description == '15.0')
 
 
 ###############################################################################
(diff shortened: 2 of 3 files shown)


rob-luke:
See render here https://30570-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/70_fnirs_processing.html#providing-more-meaningful-annotation-information

tutorials/preprocessing/70_fnirs_processing.py
```diff
@@ -26,6 +26,24 @@
 raw_intensity.load_data()
 
 
+###############################################################################
+# Providing more meaningful annotation information
+# ------------------------------------------------
+#
+# First, we attribute more meaningful names to the trigger codes which are
+# stored as annotations. Further, we include information about the duration of
```
larsoner, line 34:
```suggestion
# stored as annotations. Second, we include information about the duration of
```

tutorials/preprocessing/70_fnirs_processing.py
```diff
@@ -26,6 +26,24 @@
...
+# each stimulus, which was 5 seconds for all conditions in this experiment.
+# Further, we remove the trigger code 15, which signaled the start and end
```
larsoner, line 36:
```suggestion
# Third, we remove the trigger code 15, which signaled the start and end
```

tutorials/preprocessing/70_fnirs_processing.py
```diff
@@ -26,6 +26,24 @@
...
+# of the experiment and is not relevant to analysis.
+
+raw_intensity.annotations.set_durations(5)
+raw_intensity.annotations.rename({'1.0': 'Control',
+                                  '2.0': 'Tapping/Left',
+                                  '3.0': 'Tapping/Right'})
+raw_intensity.annotations.delete(
+    np.where([d == '15.0' for d in raw_intensity.annotations.description]))
```
larsoner, line 44: In theory (to be tested) this should "do the right thing" because NumPy should be able to do the array-of-strings comparison to yield a bool array, and `np.delete` should be bool-mask-aware
```suggestion
raw_intensity.annotations.delete(
    raw_intensity.annotations.description == '15.0')
```

larsoner (commented):
Just some minor suggestions but LGTM!

rob-luke:
Thanks @larsoner

larsoner (approved):

tutorials/preprocessing/70_fnirs_processing.py
rob-luke: Looks like this didn't work @larsoner. I'll revert tomorrow, but now it's late.

larsoner: Sounds good, no rush. Would be good to see why it didn't work...

rob-luke:
@larsoner I have fixed the return type so your suggested improvement should work now.

mne/tests/test_annotations.py
```diff
@@ -1289,6 +1289,7 @@ def test_annotation_ch_names():
 def test_annotation_rename():
     """Test annotation renaming works."""
     a = Annotations([1, 2, 3], [5, 5, 8], ["a", "b", "c"])
+    assert isinstance(a.description, np.ndarray)
```
cbrnr, line 1292: I don't think these `isinstance` tests are necessary here. The `Annotations` class initializer takes care of converting descriptions to np.array, so it should be tested there. It looks like we don't have a test yet, so maybe add that to `test_basics()`?

tutorials/preprocessing/70_fnirs_processing.py
```diff
@@ -26,6 +26,24 @@
...
 
+###############################################################################
+# Providing more meaningful annotation information
+# ------------------------------------------------
+#
+# First, we attribute more meaningful names to the trigger codes which are
+# stored as annotations. Second, we include information about the duration of
+# each stimulus, which was 5 seconds for all conditions in this experiment.
+# Third, we remove the trigger code 15, which signaled the start and end
+# of the experiment and is not relevant to analysis.
```
cbrnr, line 37:
```suggestion
# of the experiment and is not relevant to our analysis.
```

tutorials/preprocessing/70_fnirs_processing.py
```diff
@@ -26,6 +26,24 @@
...
+
+raw_intensity.annotations.set_durations(5)
+raw_intensity.annotations.rename({'1.0': 'Control',
+                                  '2.0': 'Tapping/Left',
+                                  '3.0': 'Tapping/Right'})
+raw_intensity.annotations.delete(
+    raw_intensity.annotations.description == '15.0')
```
cbrnr, line 44:
```suggestion
raw_intensity.annotations.delete(
    raw_intensity.annotations.description == '15.0'
)
```
If you use this style, I think the closing parenthesis should go into a new line (at least that's what black does).

mne/tests/test_annotations.py
rob-luke: I added these tests as the function was not returning an ndarray, so now I test that it does after my new modification to the rename function. Is there any harm in adding these tests?

tutorials/preprocessing/70_fnirs_processing.py
rob-luke: Thanks

tutorials/preprocessing/70_fnirs_processing.py
hoechenberger: That's what `black` does and the way I prefer it too; but in MNE-Python, the dominant style is the one @rob-luke proposed. 

@cbrnr Would you like to make an "official" proposal to adopt the `black` style? You'd have my support.

mne/tests/test_annotations.py
cbrnr: No harm, you can leave them here if you want. I'm just always a bit wary if I see the `isinstance` function in Python, but it seems like we really need this to be an `np.ndarray` (and not just something array-like).

tutorials/preprocessing/70_fnirs_processing.py
cbrnr: I never saw that style; I only know these two variants:

```
some_function(
    x,
    y,
    z
)

some_function(x,
              y,
              z)
```

Do we really use a mixture of the two?

Re black style, not sure if I'd want to go full in. Sometimes black formatting looks weird.

larsoner: -1 on this suggestion. Most places we deviate from `black` and keep the closing paren on the end, so I'd rather keep doing it that way at least for now. Or at least not force @rob-luke to deviate from our current practices.

@hoechenberger feel free to open a separate issue to discuss `black` adoption and/or if we should change our style practices/standards, let's not do it here.

hoechenberger: > I never saw that style

really it's all over the place in MNE's sources … I always though it looks very ugly :)

hoechenberger: > @hoechenberger feel free to open a separate issue to discuss `black` adoption and/or if we should change our style practices/standards, let's not do it here.

I'll leave that to @cbrnr :)

cbrnr: > really it's all over the place in MNE's sources … I always though it looks very ugly :)

Can you show an example? I have never seen using a newline after `(`, indenting one level, and the putting `)` in the same line. We normally indent up to `(` and not just 4 spaces.

cbrnr: It seems like sometimes we do use it (e.g. https://github.com/mne-tools/mne-python/blob/f1e6097c36e7232514d67d4b04af568e36ac5758/mne/beamformer/_compute_beamformer.py#L226-L228), but it is much less common. I'm fine with having all these variants, I just thought it was a mistake.

rob-luke:
So we can merge?

agramfort (approved):

agramfort:
Thx @rob-luke

rob-luke:
Thank you @cbrnr @larsoner @hoechenberger for the comments

PR #9541, opened 2021-07-08
MRG, ENH: Add warning when distances are bad

larsoner:
I was doing something silly during data analysis and this warning would have helped me see and fix it earlier.

@rob-luke can you look?

### DIFF
--- a/mne/io/hitachi/tests/test_hitachi.py
+++ b/mne/io/hitachi/tests/test_hitachi.py
@@ -204,31 +204,40 @@ def test_hitachi_basic(preload, version, n_ch, n_times, lowpass, sex, date,
     assert raw.info['subject_info']['sex'] == sex
     assert np.isfinite(raw.get_data()).all()
     assert raw.info['meas_date'] == dt.datetime(*date, tzinfo=dt.timezone.utc)
-    # bad distances
+    # bad distances (zero)
     distances = source_detector_distances(raw.info)
     want = [0.] * (n_ch - 4)
     assert_allclose(distances, want, atol=0.)
     raw_od_bad = optical_density(raw)
     with pytest.warns(RuntimeWarning, match='will be zero'):
         beer_lambert_law(raw_od_bad)
+    # bad distances (too big)
+    if version == '1.18':
+        need = sum(([f'S{ii}', f'D{ii}'] for ii in range(1, 9)), [])[:-1]
+        have = 'P7 FC3 C3 CP3 P3 F5 FC5 C5 CP5 P5 F7 FT7 T7 TP7 F3'.split()
+        assert len(need) == len(have)
+        mon = make_standard_montage('standard_1020')
+        mon.rename_channels(dict(zip(have, need)))
+        raw.set_montage(mon)
+        raw_od_bad = optical_density(raw)
+        with pytest.warns(RuntimeWarning, match='greater than 10 cm'):
+            beer_lambert_law(raw_od_bad)
     # good distances
     mon = make_standard_montage('standard_1020')
     if version == '1.18':
         need = sum(([f'S{ii}', f'D{ii}'] for ii in range(1, 9)), [])[:-1]
         have = 'F3 FC3 C3 CP3 P3 F5 FC5 C5 CP5 P5 F7 FT7 T7 TP7 P7'.split()
-        dist_tol = 0.01
-        bl_min = 1e-6
     else:
         need = sum(([f'S{ii}', f'D{ii}'] for ii in range(1, 18)), [])[:-1]
-        have = mon.ch_names[:len(need)]  # wrong but good enough for tests
-        dist_tol = 0.15  # really big because the list above is wrong
-        bl_min = 1e-8  # really small for the same reason
+        have = ('FT9 FT7 FC5 FC3 FC1 FCz FC2 FC4 FC6 FT8 FT10 '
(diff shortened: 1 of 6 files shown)


mne/io/hitachi/tests/test_hitachi.py
```diff
@@ -195,21 +196,46 @@ def test_hitachi_basic(preload, version, n_ch, n_times, lowpass, sex, date,
...
+    # good distances
+    mon = make_standard_montage('standard_1020')
+    if version == '1.18':
+        need = sum(([f'S{ii}', f'D{ii}'] for ii in range(1, 9)), [])[:-1]
+        have = 'F3 FC3 C3 CP3 P3 F5 FC5 C5 CP5 P5 F7 FT7 T7 TP7 P7'.split()
+        dist_tol = 0.01
+        bl_min = 1e-6
+    else:
+        need = sum(([f'S{ii}', f'D{ii}'] for ii in range(1, 18)), [])[:-1]
+        have = mon.ch_names[:len(need)]  # wrong but good enough for tests
```
larsoner, line 223: @rob-luke in particular if you have the correct mapping for the data file you supplied that would be great, as we could fix the `dist_tol` and `bl_min` tolerances

rob-luke: Sorry I don't have any more information about the dataset as I didnt measure it, it was provided to me.

mne/io/hitachi/tests/test_hitachi.py
```diff
@@ -195,21 +196,46 @@ def test_hitachi_basic(preload, version, n_ch, n_times, lowpass, sex, date,
...
+    raw.set_montage(mon)
+    distances = source_detector_distances(raw.info)
+    want = [0.03] * (n_ch - 4)
+    assert_allclose(distances, want, atol=dist_tol)
     test_rank = 'less' if n_times < n_ch else True
     _test_raw_reader(read_raw_hitachi, fname=fname,
                      boundary_decimal=1, test_rank=test_rank)  # low fs
+
+    # TODO: eventually we should refactor these to be in
+    # mne/io/tests/test_raw.py and run them for all fNIRS readers
```
rob-luke, line 237: Can you not just add the Hitachi files to the existing parametize setting in `mne/preprocessing/nirs/tests`? e.g.

https://github.com/mne-tools/mne-python/blob/720a95c7599d933526b5db7fae823e4349176a01/mne/preprocessing/nirs/tests/test_beer_lambert_law.py#L30-L33

https://github.com/mne-tools/mne-python/blob/720a95c7599d933526b5db7fae823e4349176a01/mne/preprocessing/nirs/tests/test_scalp_coupling_index.py#L29-L32

mne/preprocessing/nirs/_beer_lambert_law.py
```diff
@@ -38,10 +38,13 @@ def beer_lambert_law(raw, ppf=0.1):
     picks = _check_channels_ordered(raw.info, freqs)
     abs_coef = _load_absorption(freqs)
     distances = source_detector_distances(raw.info)
+    if (distances == 0).any():
+        warn('Source-detector distances are zero, some resulting '
+             'concentrations will be zero. Consider setting a montage '
+             'with raw.set_montage.')
```
rob-luke, line 45: I think we should add a check for distances which are too large too. fNIRS channels should be 3ish cm distance so that enough light is received by the detector. Could we throw a warning if the distance is greater than an average head? So something like 8-10cm? This would catch the situations where people are storing the distances as cm rather than m. Just a thought

```suggestion
if (distances > 0.1).any():
    warn('Source-detector distances are greater than 10 cm. '
              'Large distances will result in invalid data, and are '
              'likely due to optode locations being stored in a '
              ' unit other than meters.')
```

mne/io/hitachi/tests/test_hitachi.py
larsoner: But this means you have no information about positions, so you aren't using BLL? If so that's fine and this test can just stay as is, but it's also a bit :(

I can probably figure out some 1020 positions from the standard Hitachi configuration so that the distances are at least sane, though. So if you can't make it easy on me by providing them, then I'll at least sort out a reasonable possible arrangement :)

rob-luke: I never processed the data. I could ask the owner for info if required. I know it was a standard montage array.

rob-luke: Give me a few days and ill try and track down more details

mne/io/hitachi/tests/test_hitachi.py
larsoner: Not really easily because of the existing parametrize I think. Let's leave this as as TODO for another PR...

mne/io/hitachi/tests/test_hitachi.py
larsoner: I documented the layout structure and am pushing a commit with a reasonable choice

larsoner:
Okay should be good now @rob-luke

larsoner:
Azure is just an unrelated timeout hopefully fixed by #9549

mne/io/hitachi/tests/test_hitachi.py
rob-luke: Yeah I meant in future PR, but just as a note so that the above tests are expanded rather than reimplemented in `mne/io/tests/test_raw.py`

rob-luke (approved):

Issue #9520, opened 2021-07-01
[ENH] Add "roi" feature to `mne.viz.Brain`


alexrockhill:
One thing that is nice for plotting in 3D with stereoEEG is to be able to plot ROIs. There is already a `surfaces` keyword argument in `mne.viz.plot_alignment` does something similar. We've just added the marching cubes implementation in #9484 so it would be as simple as passing an atlas in the subjects directory, locating the aseg file based on the atlas, running marching cubes which only takes a second at most and then plotting the surface.

A call could look like this:

```
mne.viz.plot_alignment(raw.info, atlas='desikan-killiany', rois=('hippocampus', 'putamen', 'caudate'))
```

larsoner:
I'm inclined instead to improve the `Brain` interface, where it's easy to plot labels, parcellations, etc. We're adding sensor support there, and you can already combine the figure there with `plot_alignment` if you want something static.

alexrockhill:
Sounds good, that's why I raised the issue was to figure out the best way to do that

larsoner:
Give `fig = plot_alignment(...)` plus `mne.viz.Brain(...figure=fig)` a try and see how far you can get!

alexrockhill:
After thinking about it a bit and looking at the code, what would you think about an `mne.viz.plot_rois` where the `mne.viz.plot_alignment` is refactored into helper functions which can be shared. I'm not sure I understand the reason for the complexity of `mne.viz.Brain`, as I remember it in `Pysurfer` it was mostly for surface plotting and colormaps which I'm not sure has the best overlap. Just bringing up a suggestion of an idea for how to do this.

larsoner:
> what would you think about an mne.viz.plot_rois where the mne.viz.plot_alignment is refactored into helper functions which can be shared

I would not add a third method that allows plotting brain surfaces. Arguably `plot_alignment` should already use `Brain`...

> Give fig = plot_alignment(...) plus mne.viz.Brain(...figure=fig) a try and see how far you can get!

Along these lines, what does this not do that you need it to do? (Just using fnirs data here because it's the quickest route to 
```
import mne
subjects_dir = mne.datasets.sample.data_path() + '/subjects'
raw = mne.io.read_raw_nirx(mne.datasets.fnirs_motor.data_path() + '/Participant-1')
fig = mne.viz.plot_alignment(
    raw.info, subject='fsaverage', trans='fsaverage', fnirs=True, dig=True,
[...]

PR #9543, opened 2021-07-08
[MRG, MAINT] Move freesurfer functions in `mne/source_space.py` to `mne/_freesurfer.py`

alexrockhill:
Helps with refactoring for https://github.com/mne-tools/mne-python/issues/9520.

The freesurfer lookup table access seems oddly placed in `source_space.py`. I think `utils.config.py` makes sense because that is where `get_subjects_dir` is.

This would allow the lookup table to be accessed for plotting ROIs in other modules.


alexrockhill:
I don't think this should get a `whatsnew` entry since it's just a `util` but happy to do so if requested.

larsoner:
`source_space.py` is where a lot (most?) of our FreeSurfer mangling code lives. I agree `source_space.py` is at least a misnomer but we can't rename it for backward compat reasons -- perhaps we should make a new `mne/_freesurfer.py` or so rather than pack more stuff into `utils`, especially `utils/config.py` which should really just be about dealing with user config vars

alexrockhill:
> `source_space.py` is where a lot (most?) of our FreeSurfer mangling code lives. I agree `source_space.py` is at least a misnomer but we can't rename it for backward compat reasons -- perhaps we should make a new `mne/_freesurfer.py` or so rather than pack more stuff into `utils`, especially `utils/config.py` which should really just be about dealing with user config vars

You know a lot better about where things go than I do, sounds good as far as naming.

alexrockhill:
Ok @larsoner, this turned out to take a lot more changes than I thought. Now I get a circular import that `get_subjects_dir` is everywhere the basic import structure is off. I put an hour of work into it so if you want to advice on how to clean it up/clean it up I think this implements what you're suggesting. Otherwise we can just scale back if this is too much to do right now.

larsoner:
Let's just scale back, there is also a cost to moving things in terms of tracking blame and changes, and the tangible benefit is also small

larsoner:
... but now having read the PR's diff it is nice to have all this stuff in one place.

I'll see if I can push a commit to fix the circular import problem

larsoner:
I think the core problem is `get_subjects_dir`. That one I think actually makes at least as much sense to have in `config.py` as `_freesurfer.py` since it mostly uses `get_config` then does some value checks and raising (both of which are common in `utils/` files) -- I think reverting those changes should fix things.

larsoner (approved):

agramfort:
there is no public API change here?

alexrockhill:
> there is no public API change here?

The only one would be that `mne.utils.get_subjects_dir` was moving to `mne.get_subjects_dir` but Eric put that back because of import issues so everything is exactly the same that is a public function in the API.

doc/mri.rst
```diff
@@ -14,7 +14,7 @@ Step by step instructions for using :func:`gui.coregistration`:
 .. autosummary::
    :toctree: generated/
 
-   get_mni_fiducials
+   coreg.get_mni_fiducials
```
agramfort: is this an API change @alexrockhill

agramfort (approved):
@larsoner merge if happy

thx @alexrockhill !

doc/mri.rst
alexrockhill: That looks like it is the API that it was before, I just corrected it because it would have been a change but instead we just imported and noqa'd to shallow reference it in coreg

larsoner:
Thanks @alexrockhill !

PR #9544, opened 2021-07-09
[ENH, MRG] Encapsulate warp elec image in function

alexrockhill:
I thought the checks that the image warp worked properly and handling of the coordinate frames would be better handled by a function than leaving to copy paste. After writing it, the function seems rather cumbersome so it would be nice to get a review and maybe this isn't the best way to go about it. I tried splitting it into a piece for making an image and a piece for warping an image but that was cumbersome as well.

Also I didn't write a test because it is just too cumbersome but the example did work...

### DIFF
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1756,17 +1756,14 @@ def _check_dig(dig):
 
 
 def _ch_coords_to_vox(ch_coords, fs_t1):
-    """Convert channel coordinates from head space to voxel space."""
-    # convert electrode positions to voxels
+    """Convert channel coordinates from surface RAS to voxel space."""
     ch_coords = apply_trans(
         np.linalg.inv(fs_t1.header.get_vox2ras_tkr()), ch_coords * 1000)
     return ch_coords
 
 
-def _ch_coords_to_head(ch_coords, fs_t1):
-    """Convert back to the head coordinate frame."""
-    # first, convert back to surface RAS but
-    # to the template surface RAS this time
+def _ch_coords_to_surface_RAS(ch_coords, fs_t1):
+    """Convert back to the surface RAS coordinate frame."""
     ch_coords = apply_trans(
         fs_t1.header.get_vox2ras_tkr(), ch_coords) / 1000
     return ch_coords
@@ -1892,7 +1889,7 @@ def warp_montage_volume(montage, image, reg_affine, sdr_morph,
     montage_warped = montage.copy()
 
     # assign to the montage
-    ch_coords = _ch_coords_to_head(ch_coords, fs_t1)
+    ch_coords = _ch_coords_to_surface_RAS(ch_coords, fs_t1)
     for idx, ch_coord in zip(use_idx, ch_coords):
         montage_warped.dig[idx]['r'] = ch_coord
     return montage_warped, dig_image, warped_dig_image
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1821,13 +1821,9 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
     _require_version('dipy', 'SDR morph', '0.10.1')
     from .channels import DigMontage
     import nibabel as nib
-    from dipy.align.imwarp import DiffeomorphicMap
 
     _validate_type(montage, DigMontage, 'montage')
     _validate_type(base_image, nib.spatialimages.SpatialImage, 'base_image')
-    _validate_type(reg_affine, np.ndarray, 'reg_affine')
-    _check_option('reg_affine.shape', reg_affine.shape, ((4, 4),))
-    _validate_type(sdr_morph, DiffeomorphicMap, 'sdr_morph')
     _validate_type(thresh, float, 'thresh')
     if thresh < 0 or thresh >= 1:
         raise ValueError(f'`thresh` must be between 0 and 1, got {thresh}')
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -20,12 +20,13 @@ import numpy as np
 
 from .channels.channels import _get_meg_system
 from .fixes import (_serialize_volume_info, _get_read_geometry, jit,
-                    prange, bincount)
+                    prange, bincount, _get_img_fdata)
 from .io.constants import FIFF
 from .io.pick import pick_types
 from .parallel import parallel_func
 from .transforms import (transform_surface_to, _pol_to_cart, _cart_to_sph,
-                         _get_trans, apply_trans, Transform)
+                         _get_trans, apply_trans, Transform, invert_transform,
+                         apply_volume_registration)
 from .utils import (logger, verbose, get_subjects_dir, warn, _check_fname,
                     _check_option, _ensure_int, _TempDir, run_subprocess,
                     _check_freesurfer_home, _hashable_ndarray)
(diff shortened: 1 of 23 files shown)


mne/surface.py
```diff
@@ -1697,6 +1699,194 @@ def marching_cubes(image, level):
...
+        set(np.unique(_get_img_fdata(dig_image))))
+    missing_ch = [montage.ch_names[idx - 1] for idx in missing]
+    if missing_ch:
+        warn('Channels ' + ', '.join(missing_ch) + ' were not assigned '
+             'voxels' + (' after applying SDR warp' if after_warp else ''))
+
+
+@fill_doc
+def warp_montage_volume(montage, image, reg_affine, sdr_morph,
+                        subject, template, subjects_dir=None,
```
larsoner, line 1777: let's default to `temalpte='fsaverage'`. It's a reasonable choice and I assume many people will want it

mne/surface.py
```diff
@@ -1697,6 +1699,194 @@ def marching_cubes(image, level):
...
+                        template_subjects_dir=None,
```
larsoner, line 1778: I don't think it's too onerous to require that the template is in the `subjects_dir` so we can just have one (i.e., get rid of `template_subjects_dir` and just use `subjects_dir`). `ln -s` exists for people who hate `cp -a`

alexrockhill: In our example they are in different directories so that would make things complicated. I think this is okay

larsoner: So far the other functions that operate on two structurals in MNE (`mne coreg` / `mne.gui.coregistration`, `mne.compute_source_morph`) all use a single `subjects_dir`. I'm +1 or refactoring our sEEG example to adhere to this standard rather than adopt a new standard here, but I can live with something like this if we decide we want to do it all three places (here, coreg, morph).

Also we should consider if it's worth having the argument names here match what we have either for `compute_volume_registration` (moving/static) or `compute_source_morph` (subject_from/subject_to), or at least get close to that. My vote is to standardize around `subject_from` and `subject_to`, and maybe even update `compute_volume_registration` to have `image_from` and `image_to` to match.

larsoner:
> Also I didn't write a test because it is just too cumbersome but the example did work...

Everything in MNE should be regression tested -- it's an important part of the work and worth the effort! It's also a useful skill to learn in terms of how to take a complicated problem and distill it to something that proves things work properly, without taking too long to do so.

One idea would be to fake `sample`'s T1 into being CT-like. For example, eyeball two or three 3x3 blocks of voxels (in some known small volume regions) in freeview to call "electrodes". In the test, `nib.load` the T1, "color in" those hard-coded 3x3 blocks (using numpy slicing to assign values), save the image back out. Use the function in this PR to warp this "sample CT" to fsaverage, assert that you get your 2 or 3 electrodes out, and that they live in the correct fsaverage volumetric ROIs. It sounds like a lot, but I think it will end up being < 30 lines, and should make sure that this function works.

This also lends itself nicely to testing other cases, for example if you color in a single voxel, you might not get this point back, and the function should warn -- and you can assert via `pytest.warns(...)` that it does, etc.

mne/surface.py
agramfort: +1 to only have one subjects_dir

that's why we fetch all the time fsaverage in our study. it's not big

agramfort: and yes +1 for image_from and image_to and subject_from/subject_to
I find it very readable.

alexrockhill:
> > Also I didn't write a test because it is just too cumbersome but the example did work...
> 
> Everything in MNE should be regression tested -- it's an important part of the work and worth the effort! It's also a useful skill to learn in terms of how to take a complicated problem and distill it to something that proves things work properly, without taking too long to do so.
> 
> One idea would be to fake `sample`'s T1 into being CT-like. For example, eyeball two or three 3x3 blocks of voxels (in some known small volume regions) in freeview to call "electrodes". In the test, `nib.load` the T1, "color in" those hard-coded 3x3 blocks (using numpy slicing to assign values), save the image back out. Use the function in this PR to warp this "sample CT" to fsaverage, assert that you get your 2 or 3 electrodes out, and that they live in the correct fsaverage volumetric ROIs. It sounds like a lot, but I think it will end up being < 30 lines, and should make sure that this function works.
> 
> This also lends itself nicely to testing other cases, for example if you color in a single voxel, you might not get this point back, and the function should warn -- and you can assert via `pytest.warns(...)` that it does, etc.

Sounds good, I wrote that after a long day of working on too many PRs at once but also after thinking about it I don't think it's that cumbersome, it's 4 required arguments which is a lot but I think that's the best way for making it modular and making sure you can check step by step. I'm almost done with the test, will push soon.

mne/surface.py
alexrockhill: Maybe this will help teach someone reading the tutorial about symbolic links :)
Sounds convincing to me

alexrockhill:
Ok, this should pass all the tests, but it does take 50 seconds locally to test. @larsoner if you have any ideas to make it faster that would be nice but I'm not sure I have any. Other than that, should be good to go.

larsoner (commented):
Just a couple of high-level comments, I can look a bit deeper once you address those. Also 50sec will be too long, I can see if there are ways to speed it up later this week! If you want to try in the meantime, I would do something like add `@profile` to the top of the test function:
```
...
@pytest.mark.slowtest
@profile
def test_warp_montage_volume():
    ...
```
and do:
```
kernprof -lbv ~/.local/bin/pytest mne/tests/test_surface.py -k warp_montage
```
it should tell you which lines of the test are slow

mne/surface.py
```diff
@@ -1697,6 +1700,219 @@ def marching_cubes(image, level):
...
+    fs_t1 = nib.load(op.join(subjects_dir, subject_from, 'mri', 'T1.mgz'))
+    if not np.allclose(base_image.affine, fs_t1.affine, atol=1e-6):
+        raise RuntimeError('The `base_image` is not aligned to Freesurfer '
+                           'surface RAS space. This space is required as '
+                           'it is the space where the anatomical '
+                           'segmentation and reconstructed surfaces are')
+    return base_image, fs_t1
+
+
+def _check_dig(dig):
```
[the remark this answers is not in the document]
alexrockhill: Good point, thanks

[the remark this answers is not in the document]
alexrockhill: Hmmm this is was a lot more complicated than I thought because the `dig` doesn't have the channel names...

I'll push what I have, I think it's better but it's still a bit convoluted I think.

alexrockhill:
Hmmm so

```
reg_affine, sdr_morph = compute_volume_registration(
        subject_brain, template_brain, zooms=5, niter=[10, 10, 5],
        pipeline=['affine', 'sdr'])
```

raises

```
  File "<decorator-gen-22>", line 24, in compute_volume_registration
  File "/Users/alexrockhill/projects/mne-python/mne/transforms.py", line 1635, in compute_volume_registration
    return _compute_volume_registration(
  File "/Users/alexrockhill/projects/mne-python/mne/transforms.py", line 1653, in _compute_volume_registration
    pipeline = _validate_pipeline(pipeline)
  File "/Users/alexrockhill/projects/mne-python/mne/transforms.py", line 1578, in _validate_pipeline
    raise ValueError(
ValueError: Steps in pipeline are out of order, expected ('affine', 'sdr') but got ['affine', 'sdr'] instead
```

but works with a tuple.

I might fix that on this PR if that's okay

alexrockhill:
Ok, now the test is down to 20 seconds :)

alexrockhill:
Ok ready for review @larsoner, whenever you get a chance

mne/tests/test_surface.py
```diff
@@ -236,8 +239,104 @@ def test_voxel_neighbors():
...
+                          [-21.1130506, 21.98310911, 13.25658887]])
+    ch_pos = dict(zip(['1', '2', '3'], ch_coords / 1000))  # mm -> m
+    montage = make_dig_montage(ch_pos, coord_frame='mri')
+    montage_warped, image_from, image_to = warp_montage_volume(
+        montage, CT, reg_affine, sdr_morph, 'sample',
+        subjects_dir=subjects_dir, thresh=0.99)
+    # checked with nilearn plot from `tut-ieeg-localize`
+    # check montage in surface RAS
+    np.testing.assert_almost_equal(
+        montage_warped.dig[0]['r'], np.array([-0.31, 0.29, -0.35]), decimal=1)
```
larsoner, line 288: `decimal=1` means `0.1` i.e., 10 cm (units here are in meters). Surely this can be better? Also you don't need to explicitly `np.array` this and we always `from numpy.testing import assert_almost_equal` so I would do:
```
    assert_almost_equal(montage_warped.dig[0]['r'], [-0.31, 0.29, -0.35], decimal=3)
```
but this fails
```
mne/tests/test_surface.py:287: in test_warp_montage_volume
    np.testing.assert_almost_equal(
E   AssertionError: 
E   Arrays are not almost equal to 3 decimals
E   
E   Mismatched elements: 3 / 3 (100%)
E   Max absolute difference: 0.062
E   Max relative difference: 0.2137931
E    x: array([-0.313,  0.352, -0.325])
E    y: array([-0.31,  0.29, -0.35])
```
This actually seems quite far off (5cm!)

mne/surface.py
```diff
@@ -1697,6 +1700,211 @@ def marching_cubes(image, level):
...
+    _require_version('dipy', 'SDR morph', '0.10.1')
+    from .channels import DigMontage
+    import nibabel as nib
+    from dipy.align.imwarp import DiffeomorphicMap
+
+    _validate_type(montage, DigMontage, 'montage')
+    _validate_type(base_image, nib.spatialimages.SpatialImage, 'base_image')
+    _validate_type(reg_affine, np.ndarray, 'reg_affine')
+    _check_option('reg_affine.shape', reg_affine.shape, ((4, 4),))
+    _validate_type(sdr_morph, DiffeomorphicMap, 'sdr_morph')
```
larsoner, line 1822: This should really be:
```
    _validate_type(reg_affine, np.ndarray, 'reg_affine')
    _check_option('reg_affine.shape', reg_affine.shape, ((4, 4),))
    _validate_type(sdr_morph, (DiffeomorphicMap, None), 'sdr_morph')
```
because users shouldn't have to use SDR if affine is good enough. But these arguments are just passed unmodified to `apply_volume_registration` so these checks should live there (if they don't already) and here omitted (DRY):
```suggestion
```

larsoner (changes requested):

mne/tests/test_surface.py
alexrockhill: So I can make the tests a lot more precise but it depends a lot on the SDR and how many iterations. I thought it was rather frustrating how precise the arbitrary numbers were in the SDR morph for ranges of numbers so I tried to make this test a bit looser so that it wouldn't fail in the future if things were changed slightly since it's really not that precise unless you use many more iterations. i.e. at this level of test with the roughness it is actually off by 5 cm from the high-precision truth so I think it makes sense to make tests that are appropriately loose. Happy to change to be more precise if that's not convincing to you.

mne/surface.py
alexrockhill: Good point!

mne/tests/test_surface.py
larsoner: 5 cm is really far. I would expect even an affine transformation alone (and probably even a rigid affine!) to be off by less for any given point. Hence this makes me worry a bit that something is systematically wrong with the code or the test...

alexrockhill:
So I changed the tests so that the `zooms=5` but full iterations version is the ground truth but the euclidean distance is still ~0.11 which I think is in meters so that's 110 cm which is really far off...

"Ground truth" (with `zooms=5`)

```
[[-0.27778788, 0.24251515, -0.35693939],
 [-0.30033333, 0.24785714, -0.35014286],
 [-0.32261947, 0.25295575, -0.34614159]]
```

Extra quick version (affine + SDR with low iterations)

```
[[-0.31253333,  0.352     , -0.3247    ]
 [-0.33082353,  0.35317647, -0.31488235]
 [-0.34751786,  0.34966071, -0.30701786]]
```

alexrockhill:
I'll try adding back translation and rigid and see how much of a time hit that is for the precision.

larsoner:
Yes it's possible that the `affine` step with low zooms breaks things, so doing translation+rigid+SDR in the test (skipping affine) or even just translation+rigid (no affine or SDR) with `zooms=5` if it gets us speed + good enough accuracy (< 1 cm would be nice)

alexrockhill:
> Yes it's possible that the `affine` step with low zooms breaks things, so doing translation+rigid+SDR in the test (skipping affine) or even just translation+rigid (no affine or SDR) with `zooms=5` if it gets us speed + good enough accuracy (< 1 cm would be nice)

Great, I'll try that!

alexrockhill:
Hmmm with `zooms=5` even without the affine, we're still at about 10 cm

alexrockhill:
I'll try messing with the parameters a bit more, it's possible that a few more translation iterations will fix things

larsoner:
I would use `nibabel.orthoview()` on the before and after images to see what's going wrong. I wouldn't even really expect an identity "registration" for sample-fsaverage to be off by 10cm for most points...

larsoner:
(but maybe I underestimate how far off they are)

alexrockhill:
Wait, I'm thinking in mm, they are off my ~1 cm 🤦 

```
0.010855916835345538
0.008643073755143938
0.007148324234890798
```

larsoner:
Oh good, I think `atol=0.01` / `decimal=2` i.e. 1 cm is good enough for the 5mm-zooms version!

alexrockhill:
Ok good to go but it was 28 seconds locally, I tried turning down the iterations and skipping steps but everything I did lost precision so I think this is about the best to be ~1 cm precise

alexrockhill:
Ok this is good to go by me if it's okay with you

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -288,48 +305,12 @@ def plot_overlay(image, compare, title, thresh=None):
...
-
-# %%
-# Warp and plot the result.
-
-warped_elec_image = mne.transforms.apply_volume_registration(
-    elec_image, template_brain, reg_affine, sdr_morph, interpolation='nearest')
-
-# ensure that all the electrode contacts were warped to the template
-assert set(np.arange(1, ch_coords.shape[0] + 1)).difference(
-    set(np.unique(warped_elec_image.get_fdata()))) == set()
```
larsoner, line 332: Hooray simplification!

larsoner (approved):

larsoner:
Will merge once CIs come back happy, thanks @alexrockhill!

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -199,24 +200,38 @@ def plot_overlay(image, compare, title, thresh=None):
...
+# Let's plot the electrode contact locations on the subject's brain.
+
 # load the subject's brain
-subject_brain = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_brain.mgz'))
+subject_brain = nib.load(
+    op.join(misc_path, 'seeg', 'sample_seeg', 'mri', 'brain.mgz'))
 
-# make brain surface from T1
+# Make brain surface from T1
 verts, triangles = mne.surface.marching_cubes(
```
larsoner, line 235: Actually now that we have proper `pial` surfaces in the dataset, can you just load and plot them instead? It should further simplify the example because you don't have to marching cubes + transform, just load and plot

larsoner: (and actually you can use `mne.viz.plot_alignment` to do this for you in one line!)

alexrockhill: Sure, sounds good

larsoner: Actually can you do this in one of the other PRs? It seems like getting this PR merged is a bottleneck for the others, so I'd rather wait for CIs now and merge this ASAP

alexrockhill: So we took out the `bem` surfaces in the paring down so we can't plot the head...

Maybe a PR for a later date

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -231,23 +234,21 @@ def plot_overlay(image, compare, title, thresh=None):
...
 fig_kwargs = dict(size=(800, 600), bgcolor='w', scene=False)
 renderer = mne.viz.backends.renderer.create_3d_figure(**fig_kwargs)
-renderer.mesh(*verts.T, triangles=triangles, color='gray',
-              opacity=0.05, representation='surface')
-for ch_coord in ch_coords:
-    renderer.sphere(center=tuple(ch_coord / 1000.), color='y', scale=0.005)
+fig = mne.viz.plot_alignment(raw.info, trans, 'sample_seeg',
+                             fig=renderer.figure,
+                             subjects_dir=op.join(misc_path, 'seeg'),
+                             show_axes=True, surfaces=dict(pial=0.2))
```
larsoner, line 249: Better to have `coord_frame=`mri` here for view comparisons?

tutorials/clinical/10_ieeg_localize.py
larsoner: I don't think it's critical to show the head but if we need to update the dataset at some point we could add it

tutorials/clinical/10_ieeg_localize.py
alexrockhill: Not sure I get your meaning. Isn't it in `mri`?

larsoner: default `coord_frame='head'`

https://mne.tools/stable/generated/mne.viz.plot_alignment.html#mne-viz-plot-alignment

larsoner: (you can see in [my comment](https://github.com/mne-tools/mne-python/issues/9520#issuecomment-880968547) that I pass `coord_frame='mri'`, if you don't do this then `plot_alignment` plots everything in head coordinates, while `Brain` always plots in MRI coords)

larsoner:
Pushed a commit to set `coord_frame='mri'` and will merge once CIs come back happy!

(After doing that I realized that it doesn't matter much for this PR which coordinate frame things are plotted in, but it'll be useful when we combine with `Brain` in #9545, so no harm in having it here.)

alexrockhill:
Ok, I think good to go!

larsoner:
Indeed, thanks @alexrockhill ! Time for some fun rebasing elsewhere :)

alexrockhill:
> Indeed, thanks @alexrockhill ! Time for some fun rebasing elsewhere :)

The plan is that the dominos fall perfectly!

PR #9540, opened 2021-07-08
[ENH, MRG] Add mne.viz.Brain.add_volume_labels

alexrockhill:
Closes #9520

Here is a very, very rough draft of what I was thinking, I didn't want to go too far in a bad direction so if you have a second to look at it @larsoner, that would be very much appreciated.

Basically, the ROIs are a very different kind of surfaces than the SUBJECTS_DIR/SUBJECT/surf surfaces and so I think need their own class but I think inheritance works nicely; they are named very differently and don't all have hemispheres.

PR #9545, opened 2021-07-09
[MRG, ENH] Find aseg labels near montage

alexrockhill:
The idea here is to find which voxels are nearby an electrode contact in order to know which surfaces to plot and for analysis.

This shares internal functions with https://github.com/mne-tools/mne-python/pull/9544/ and is a good reason to make `mne.surfaces._find_voxel_neighbors` private so that it can be used in both scenarios without confusion and user-facing function.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -109,6 +109,8 @@ Enhancements
 
 - The :meth:`mne.Evoked.get_data` method now has a ``units`` parameter (:gh:`9578` by `Stefan Appelhoff`_)
 
+- Add :func:`mne.get_montage_rois` to find the regions of interest in a Freesurfer atlas anatomical segmentation for an intracranial electrode montage and :func:`mne.viz.plot_channel_labels_circle` to plot them (:gh:`9545` by `Alex Rockhill`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -109,7 +109,7 @@ Enhancements
 
 - The :meth:`mne.Evoked.get_data` method now has a ``units`` parameter (:gh:`9578` by `Stefan Appelhoff`_)
 
-- Add :func:`mne.get_montage_rois` to find the regions of interest in a Freesurfer atlas anatomical segmentation for an intracranial electrode montage and :func:`mne.viz.plot_channel_labels_circle` to plot them (:gh:`9545` by `Alex Rockhill`_)
+- Add :func:`mne.get_montage_volume_labels` to find the regions of interest in a Freesurfer atlas anatomical segmentation for an intracranial electrode montage and :func:`mne.viz.plot_channel_labels_circle` to plot them (:gh:`9545` by `Alex Rockhill`_)
 
 Bugs
 ~~~~
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1889,9 +1889,9 @@ _VOXELS_MAX = 100  # define constant to avoid runtime issues
 
 
 @fill_doc
-def get_montage_rois(montage, subject, subjects_dir=None,
-                     aseg='aparc+aseg', dist=5):
-    """Get the regions of interest near the channels in a montage.
+def get_montage_volume_labels(montage, subject, subjects_dir=None,
+                              aseg='aparc+aseg', dist=5):
+    """Get regions of interest near channels from a Freesurfer parcellation.
 
     .. note:: This is applicable for channels inside the brain
               (intracranial electrodes).
@@ -1907,10 +1907,10 @@ def get_montage_rois(montage, subject, subjects_dir=None,
 
     Returns
     -------
-    rois : dict
-        The regions of interest within ``dist`` of each channel.
+    labels : dict
+        The regions of interest labels within ``dist`` of each channel.
     colors : dict
-        The Freesurfer lookup table colors for the rois.
+        The Freesurfer lookup table colors for the labels.
     """
     from .channels import DigMontage
     from ._freesurfer import read_freesurfer_lut, _get_aseg
@@ -1922,10 +1922,10 @@ def get_montage_rois(montage, subject, subjects_dir=None,
 
     # read freesurfer lookup table
     lut, fs_colors = read_freesurfer_lut()
-    labels = {v: k for k, v in lut.items()}
+    label_lut = {v: k for k, v in lut.items()}
 
     # assert that all the values in the aseg are in the labels
-    assert all([idx in labels for idx in np.unique(aseg_data)])
+    assert all([idx in label_lut for idx in np.unique(aseg_data)])
 
     # get transform to surface RAS for distance units instead of voxels
     vox2ras_tkr = aseg.header.get_vox2ras_tkr()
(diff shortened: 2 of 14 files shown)


alexrockhill:
@larsoner, I think it might be best to change the default back to `aseg.mgz` because `fsaverage` doesn't have any other segmentations in the sample data.

alexrockhill:
This is ready for review too, but this PR has to merge first https://github.com/mne-tools/mne-misc-data/pull/9

larsoner:
Even after fetch_fsaverage is called?

We can also update the sample dataset pretty easily, so I wouldn't let it be a limiting factor

alexrockhill:
> Even after fetch_fsaverage is called?
> 
> We can also update the sample dataset pretty easily, so I wouldn't let it be a limiting factor

Yes but good to know, maybe we should update that then

alexrockhill:
This also updates the seeg tutorial which was out-of-date because of the new data format

alexrockhill:
Here's what the plots I added look like by the way.

Screen Shot 2021-07-14 at 8 23 04 PM

larsoner:
>  I think it might be best to change the default back to aseg.mgz because fsaverage doesn't have any other segmentations in the sample data.

I just did `mne.datasets.sample.data_path(force_update=True, verbose=True)` (i.e., not even running `fetch_fsaverage`) and I see:
```
$ ls ~/mne_data/MNE-sample-data/subjects/fsaverage/mri
aparc.a2005s+aseg.mgz  aparc+aseg.mgz  brainmask.mgz  lh.ribbon.mgz   orig      p.aseg.mgz     ribbon.mgz        subcort.prob.mgz  transforms
aparc.a2009s+aseg.mgz  aseg.mgz        brain.mgz      mni305.cor.mgz  orig.mgz  rh.ribbon.mgz  subcort.prob.log  T1.mgz
```
so `aparc+aseg.mgz` is in there. Maybe your version of `sample` was outdated?

larsoner (commented):
Can anything be refactored with `mne.label.find_pos_in_annot`?

alexrockhill:
> find_pos_in_annot

I didn't see that one, yeah looks like it. I think it might have to be refactored to accept multiple positions though since you wouldn't want to be loading the aseg file over and over.

alexrockhill:
> > I think it might be best to change the default back to aseg.mgz because fsaverage doesn't have any other segmentations in the sample data.
> 
> I just did `mne.datasets.sample.data_path(force_update=True, verbose=True)` (i.e., not even running `fetch_fsaverage`) and I see:
> 
> ```
> $ ls ~/mne_data/MNE-sample-data/subjects/fsaverage/mri
> aparc.a2005s+aseg.mgz  aparc+aseg.mgz  brainmask.mgz  lh.ribbon.mgz   orig      p.aseg.mgz     ribbon.mgz        subcort.prob.mgz  transforms
> aparc.a2009s+aseg.mgz  aseg.mgz        brain.mgz      mni305.cor.mgz  orig.mgz  rh.ribbon.mgz  subcort.prob.log  T1.mgz
> ```
> 
> so `aparc+aseg.mgz` is in there. Maybe your version of `sample` was outdated?

Yep, mine must have just been way out-of-date.

alexrockhill:
Ok so I changed one thing that I was very unsure about: `label.get_pos_in_annot` went to `label.get_pos_in_aseg` which included backdating a function by someone else. Since it was before the next release, I think this might be okay since I think calling it `annot` is confusing because `annot` files are something slightly different and already in use to mean a different thing.

Other than that, I tried to refactor to use the function but realized after doing so that it actually would be really inefficient because each channel of the montage is looped over so the aseg would have to be loaded in every time to use that function. I did refactor a bit into a `_get_aseg` function in `_freesurfer.py` so that some of the codebase is shared.

If all that is okay, should be good to go.

larsoner:
> label.get_pos_in_aseg which included backdating a function by someone else. Since it was before the next release, I think this might be okay since I think calling it annot is confusing because annot files are something slightly different and already in use to mean a different thing.

Agreed this naming is better, I'll look at the refactoring now

larsoner:
Actually I'll wait until #9544 is merged and this is rebased, feel free to ping me once that's done @alexrockhill

alexrockhill:
I think that looks pretty cool and informative

Screen Shot 2021-07-16 at 1 18 59 PM

alexrockhill:
> > label.get_pos_in_aseg which included backdating a function by someone else. Since it was before the next release, I think this might be okay since I think calling it annot is confusing because annot files are something slightly different and already in use to mean a different thing.
> 
> Agreed this naming is better, I'll look at the refactoring now

I took that part out and put it in a separate PR, this should be good to go as well after I rebase

larsoner:
> I took that part out and put it in a separate PR, this should be good to go as well after I rebase

The refactoring PR should be merged first, then, right?

alexrockhill:
> > I took that part out and put it in a separate PR, this should be good to go as well after I rebase
> 
> The refactoring PR should be merged first, then, right?

They are actually independent/use the exact same `_freesurfer.py` `_get_aseg` code from a `git checkout aseg -- mne/_freesurfer.py` call.

alexrockhill:
Looks good to merge by me!

doc/changes/latest.inc
```diff
@@ -109,6 +109,8 @@ Enhancements
 
 - The :meth:`mne.Evoked.get_data` method now has a ``units`` parameter (:gh:`9578` by `Stefan Appelhoff`_)
 
+- Add :func:`mne.get_montage_rois` to find the regions of interest in a Freesurfer atlas anatomical segmentation for an intracranial electrode montage and :func:`mne.viz.plot_channel_labels_circle` to plot them (:gh:`9545` by `Alex Rockhill`_)
```
agramfort, line 112: Why tend to calls ROIs labels everywhere in mne / freesurfer world. So here it would be volume labels?

Get_montage_volume_labels or get_montage_vol_labels ?

mne/surface.py
```diff
@@ -1906,24 +1885,101 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
...
 
 
-def _get_neighbors(loc, image, thresh, voxels):
+_VOXELS_MAX = 100  # define constant to avoid runtime issues
+
+
+@fill_doc
+def get_montage_rois(montage, subject, subjects_dir=None,
+                     aseg='aparc+aseg', dist=5):
+    """Get the regions of interest near the channels in a montage.
```
agramfort, line 1894: Say where the ROIs / volume labels come from. Say it’s coming from a freesurfer volume parcellation

mne/surface.py
```diff
@@ -1906,24 +1885,101 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
...
+    %(aseg)s
+    dist : float
+        The distance in mm to use for identifying regions of interest.
+
+    Returns
+    -------
+    rois : dict
+        The regions of interest within ``dist`` of each channel.
+    colors : dict
+        The Freesurfer lookup table colors for the rois.
```
agramfort, line 1913: Please check of the acronym ROI is used anywhere in mne and use caps for an acronym when not a variable name.

mne/viz/circle.py
```diff
@@ -424,3 +425,77 @@ def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
 
     plt_show(show)
     return fig, axes
+
+
+def plot_channel_labels_circle(labels, colors=None, picks=None, **kwargs):
+    """Plot labels for each channel in a circle plot.
```
agramfort, line 431: Can you say here what is the rationale for this function. I would suggest to say it makes sense for sEEG channels a not any type of channels.

mne/viz/circle.py
```diff
@@ -424,3 +425,77 @@ def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
...
+
+    Parameters
+    ----------
+    labels : dict
+        Lists of labels (values) associated with each channel (keys).
+    colors : dict
+        The color (value) for each label (key).
+    picks : list | tuple
+        The channels to consider.
+    **kwargs : kwargs
```
agramfort, line 441: I prefer to avoid kwargs if possible.

doc/changes/latest.inc
larsoner: + 1 for `get_montage_volume_labels`

mne/surface.py
alexrockhill: It's used here https://github.com/mne-tools/mne-python/blob/main/examples/visualization/roi_erpimage_by_rt.py but it looks maybe an EEGLAB thing to mean general regions of sensor space (also several places here https://github.com/mne-tools/mne-python/blob/main/mne/channels/channels.py).

Also here in matlab.rst

```
 +--------------------------------+--------------------------------------------------------------+
 | mne_read_label_file            | Read a label file (ROI).                                     |
 +--------------------------------+--------------------------------------------------------------+
 | mne_write_label_file           | Write a label file (ROI).                                    |
 +--------------------------------+--------------------------------------------------------------+
```

And finally also the one that I assume you were thinking of is in several locations, which is using it synonymously with label.

```
# load the anatomical ROI for comparison
anat_label = mne.read_labels_from_annot(subject, parc='aparc',
                                       subjects_dir=subjects_dir,
```

I'll just go ahead and change to `label`.

mne/viz/circle.py
alexrockhill: I think you could use it for any channel type provided you have some labels/classes associated with it but I agree there are not a lot of other use cases.

mne/viz/circle.py
alexrockhill: There are 24 arguments that could possibly be used from `mne.viz.plot_connectivity_circle` to change the plotting style so I'm not sure this is possible. It especially provides the ability to change fontsizes and colors which is really nice for plots.

mne/viz/circle.py
agramfort: Well the pb I see is that field propagation is not isotropic  ie distance to source is not monotic with field strength. I fear such a function without a warning can enourage the type of false conclusions you can see in certain EEG papers.

mne/viz/circle.py
agramfort: Hum… let’s say I am getting old and more flexible… I can live with that

agramfort (approved):

larsoner:
Thanks @alexrockhill !

drammock:
@alexrockhill The doc build failed on this PR, and it looks related, see https://app.circleci.com/pipelines/github/mne-tools/mne-python/9131/workflows/e0f6c14c-27ff-43dd-90c2-939a03863e81/jobs/31255

larsoner:
@alexrockhill can you add a fix to #9540 ? I think that's very close to merge

alexrockhill:
> @alexrockhill can you add a fix to #9540 ? I think that's very close to merge

Sure right on it

PR #9552, opened 2021-07-12
DOC: use regexps for sphinx-copybutton

sappelhoff:
We use https://sphinx-copybutton.readthedocs.io/en/latest/ in our docs. This PR makes use of sphinx-copybutton configurations to improve copying.

Namely, fields like this one:

```
$ make html
```

or like this one:

```
>>> 1+1
2
```

Will copy without the respective prompts (--> `make html` instead of `$ make html`), and can thus be directly pasted into the terminal / editor.

Examples for what will be improved:

- Python example blocks in the API docs, like here: https://mne.tools/stable/generated/mne.io.Raw.html
- Contributor docs, like here: https://mne.tools/stable/install/contributing.html#building-the-documentation

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -126,6 +126,9 @@ pygments_style = 'default'
 # A list of ignored prefixes for module index sorting.
 modindex_common_prefix = ['mne.']
 
+# -- Sphiny-Copybutton configuration -----------------------------------------
+copybutton_prompt_text = r">>> |\.\.\. |\$ "
+copybutton_prompt_is_regexp = True
 
 # -- Intersphinx configuration -----------------------------------------------
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -126,7 +126,7 @@ pygments_style = 'default'
 # A list of ignored prefixes for module index sorting.
 modindex_common_prefix = ['mne.']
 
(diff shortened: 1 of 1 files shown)


agramfort (approved):

doc/conf.py
```diff
@@ -126,6 +126,9 @@
 # A list of ignored prefixes for module index sorting.
 modindex_common_prefix = ['mne.']
 
+# -- Sphiny-Copybutton configuration -----------------------------------------
```
agramfort, line 129:
```suggestion
# -- Sphinx-Copybutton configuration -----------------------------------------
```

sappelhoff:
rendered, check here (copy and paste to somewhere):

- https://30716-1301584-gh.circle-artifacts.com/0/dev/install/contributing.html#configuring-git
- https://30716-1301584-gh.circle-artifacts.com/0/dev/generated/mne.io.Raw.html (search for "examples", then copy paste first one you see)
- gallery code blocks work as before: https://30716-1301584-gh.circle-artifacts.com/0/dev/auto_examples/io/read_xdf.html#sphx-glr-auto-examples-io-read-xdf-py

larsoner:
Thanks @sappelhoff !

PR #9553, opened 2021-07-12
[MRG] Add `units` parameter to get_data for Epochs

### LINKED ISSUE #8473: FR: Add "unit" parameter to get_data method for Epochs

sappelhoff:
closes #8473 

Follow up to #9136

Adds the `units` parameter to the `get_data` method of `Epochs`.

~~**Work in Progress**~~

To do

- [x] refactor units in Raw
- [x] add units to Epochs
- [x] add tests for Epochs
- [x] pull units docs into docdict
- [x] document all params
- [x] add whatsnew entry


mne/epochs.py
```diff
@@ -3268,7 +3304,7 @@ def add_channels_epochs(epochs_list, verbose=None):
         raise ValueError('All epochs must be preloaded.')
 
     info = _merge_info([epochs.info for epochs in epochs_list])
-    data = [epochs.get_data() for epochs in epochs_list]
+    data = [epochs._data for epochs in epochs_list]
```
sappelhoff, line 3307: I made this change because this test would fail otherwise: https://github.com/mne-tools/mne-python/blob/0156284d20f0b1ae2ca8630acf3a507b4b50b21d/mne/tests/test_epochs.py#L2488-L2490

that test is popping something from `self.info['chs']`, **not** updating `self._data`, to later trigger a RuntimeError in the test. The info and _data are thus "out of sync" (something that's not going to happen in normal use) and trip up the `get_data`.

Is my fix of using `._data` here safe? Do you have better ideas?

larsoner: > Is my fix of using ._data here safe? Do you have better ideas?

Probably good enough. The test is messing with stuff in non-standard ways (and using private functions) so it's okay to have to resort to something like `._data` if you couldn't find a cleaner way to do it after thinking about it.

sappelhoff:
@HanBnrd perhaps you want to review this PR, as it builds on the stuff you coded during the last sprint :-)

mne/epochs.py
```diff
@@ -1325,7 +1325,8 @@ def _project_epoch(self, epoch):
         return epoch
 
     @verbose
-    def _get_data(self, out=True, picks=None, item=None, verbose=None):
+    def _get_data(self, out=True, picks=None, item=None, units=None,
```
larsoner, line 1328: Might help ensure internally we don't make a mistake
```suggestion
    def _get_data(self, out=True, picks=None, item=None, *, units=None,
```

mne/epochs.py
```diff
@@ -1355,16 +1359,30 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
                 data = data[select]
-                if picks is None:
+                if orig_picks is None:
+                    data *= ch_factors[np.newaxis, :, np.newaxis]
                     return data
                 else:
-                    picks = _picks_to_idx(self.info, picks)
-                    return data[:, picks]
+                    data = data[:, picks] * ch_factors[np.newaxis, :,
+                                                       np.newaxis]
```
larsoner, line 1384: Avoid copies in memory, let broadcasting simplify things, and make slightly more computationally efficient when we can (which will be most of the time)
```suggestion
                    data = data[:, picks]
                    if units is not None:
                        data *= ch_factors[np.newaxis]
```

mne/epochs.py
```diff
@@ -1443,11 +1461,12 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
 
         if out:
-            if picks is None:
+            if orig_picks is None:
+                data *= ch_factors[np.newaxis, :, np.newaxis]
                 return data
             else:
-                picks = _picks_to_idx(self.info, picks)
-                return data[:, picks]
+                data = data[:, picks] * ch_factors[np.newaxis, :, np.newaxis]
```
larsoner, line 1468:
```suggestion
                data = data[:, picks]
                if units is not None:
                    data *= ch_factors[np.newaxis]
```

mne/epochs.py
```diff
@@ -1474,13 +1493,14 @@ def get_data(self, picks=None, item=None):
             None (default) is an alias for ``slice(None)``.
 
             .. versionadded:: 0.20
+        %(units)s
```
larsoner, line 1496: versionadded

HanBnrd:
> @HanBnrd perhaps you want to review this PR, as it builds on the stuff you coded during the last sprint :-)

Thanks for implementing this @sappelhoff! Just went through the code, looks good to me 🙂

mne/epochs.py
HanBnrd: @larsoner @sappelhoff, should it be done also L1380 and L1465? And also in *base.py* at L850, L852, L862 and L904?

mne/tests/test_epochs.py
```diff
@@ -183,6 +183,39 @@ def _get_data(preload=False):
...
+    np.testing.assert_array_equal(d1, d2)
+
+    with pytest.raises(ValueError, match="is not a valid unit for eeg"):
+        epochs.get_data(picks="eeg", units="")
+
+    with pytest.raises(ValueError, match="cannot be str if there is more"):
+        epochs.get_data(picks=["eeg", "meg"], units="V")
+
+    # Check combination of units with item param, scale only one ch_type
+    d3 = epochs.get_data(item=[1, 2, 3], units={"grad": "fT/cm"})
```
HanBnrd, line 206: Would it make sense to also test that the number of items is okay if the *item* parameter is used together with *units*? (Just to ensure that *units* is not breaking *item* in the future)

mne/epochs.py
larsoner: Agreed above it could be done. I also updated my comment because I realized I had the broadcasting logic wrong (whoops!), needs to be `[:, np.newaxis]` not `[np.newaxis]` (to replace the PR's `[np.newaxis, :, np.newaxis]). Then this will also match what's in `base.py` (and those shouldn't be changed).

mne/epochs.py
```diff
@@ -1443,11 +1461,12 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
                           select_data=False)
 
         if out:
-            if picks is None:
+            if orig_picks is None:
+                data *= ch_factors[np.newaxis, :, np.newaxis]
```
sappelhoff, line 1465:
```suggestion
                if units is not None:
                    data *= ch_factors[:, np.newaxis]
```

mne/epochs.py
```diff
@@ -1355,16 +1359,30 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
+            ch_factors = _get_ch_factors(self, units, picks)
+
         if self._bad_dropped:
             if not out:
                 return
             if self.preload:
                 data = data[select]
-                if picks is None:
+                if orig_picks is None:
+                    data *= ch_factors[np.newaxis, :, np.newaxis]
```
sappelhoff, line 1380:
```suggestion
                    if units is not None:
                    data *= ch_factors[:, np.newaxis]
```

mne/epochs.py
sappelhoff:
```suggestion
        %(units)s
        
        .. versionadded:: 0.24.0
```

mne/tests/test_epochs.py
sappelhoff: I added a test, not sure if that is what you meant :)

sappelhoff (commented):
Thanks for the review @larsoner @HanBnrd I addressed your points and the PR is ready for more feedback or for merging.

doc/changes/latest.inc
```diff
@@ -93,6 +93,8 @@ Enhancements
 
 - Add :meth:`mne.Evoked.get_data` method to :class:`mne.Evoked` (:gh:`9555` by `Stefan Appelhoff`_)
 
+- The :class:`mne.Epochs` class :meth:`mne.Epochs.get_data` method now has a ``Units`` parameter (:gh:`9553` by `Stefan Appelhoff`_)
```
larsoner, line 96:
```suggestion
- The :meth:`mne.Epochs.get_data` method now has a ``units`` parameter (:gh:`9553` by `Stefan Appelhoff`_)
```

mne/epochs.py
```diff
@@ -1355,16 +1359,32 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
+                    if units is not None:
+                        data *= ch_factors[:, np.newaxis]
                     return data
                 else:
-                    picks = _picks_to_idx(self.info, picks)
-                    return data[:, picks]
+                    data = data[:, picks]
+                    if units is not None:
+                        data *= ch_factors[:, np.newaxis]
+                    return data
```
larsoner, line 1387: Simpler as:
```
if orig_picks is not None:
    data = data[:, picks]
if units is not None:
    data *= ch_factors[:, np.newaxis]
return data
```

mne/epochs.py
```diff
@@ -1443,11 +1463,15 @@ def _get_data(self, out=True, picks=None, item=None, verbose=None):
...
+                if units is not None:
+                    data *= ch_factors[:, np.newaxis]
                 return data
             else:
-                picks = _picks_to_idx(self.info, picks)
-                return data[:, picks]
+                data = data[:, picks]
+                if units is not None:
+                    data *= ch_factors[:, np.newaxis]
+                return data
```
larsoner, line 1474: Probably some similar refactoring as above could be done

larsoner (commented):
Other than a couple of minor simplification ideas, LGTM!

HanBnrd (approved):
Looks good to me 👍

larsoner (approved):

larsoner:
Thanks @sappelhoff !

PR #9554, opened 2021-07-12
MAINT: Try simpler color workaround

larsoner:
@drammock requested a version comment in #9541 but it exists in our minimum mpl 3.1.0 so we should just be able to use it:

https://matplotlib.org/3.1.0/api/collections_api.html?highlight=pathcollection#matplotlib.collections.PathCollection.get_edgecolors


agramfort (approved):
@drammock merge if happy

thx @larsoner

Issue #9548, opened 2021-07-09
ENH: [raw|epochs|evoked].get_data() specify tmin, tmax in seconds to get that segment


sappelhoff:
I often find myself wanting data from epochs in a certain time window, such as from -0.1 to 0.3 seconds. I don't want to look at `epochs.times` and compute the corresponding sample index every time, so I usually do something like `epochs.copy().crop(tmin, tmax).get_data()`. But copying the epochs is of course memory (and time) intensive.

I suggest we add two new parameters to the `get_data` method for raw, epochs, and evoked (evoked still needs the get_data method to be implemented, see #9244): `tstart` and `tstop` (or `tmin`, and `tmax`)

[The `get_data` method for raw](https://mne.tools/stable/generated/mne.io.Raw.html?highlight=get_data#mne.io.Raw.get_data) already has `start` and `stop` parameters, based on sample indices. The corresponding `tstart`/`tstop` params would operate the same way, but in seconds.

Adding something like `tstart` and `tstop` (in seconds) would also mean that we should add a `return_times=True|False` parameter to the [epochs.get_data](https://mne.tools/stable/generated/mne.Epochs.html?highlight=get_data#mne.Epochs.get_data) (and evoked.get_data, once we have it). That param is already there for raw.

Finally, maybe this would be a good opportunity to make the `get_data` interface equal between raw, epochs, and evoked.

epochs and evoked would then need to get the following that are already present in raw:

- units (#8473)
- start, stop (in samples)
- reject_by_annotation

What do you think?

larsoner:
Annotations are not preserved for epochs and evoked, so no reason to add them there.

I'd rather not add start/stop to epochs and evokeds. I kind of regret even having the sample-based one for raw -- I think it's mostly for backward compat, and it doesn't cost us much there so why not keep it. But I'd rather not add the sample-based stuff to more objects, I think it's not going to see much (if any) use. I'd rather showcase time-based limits in tutorials and examples, (but leave start/stop for raw for the reasons above).

Maybe we should use tmin/tmax rather than tstart/tstop? Not sure.

+1 for adding units to the epoch/evoked functions

drammock:
> I often find myself wanting data from epochs in a certain time window, such as from -0.1 to 0.3 seconds. I don't want to look at `epochs.times` and compute the corresponding sample index every time

Assuming you're aware of `epochs.time_as_index()`... is that the approach that you find too cumbersome?

[...]

PR #9555, opened 2021-07-12
Add get_data method to Evoked

### LINKED ISSUE #9244: ENH: Add evoked.get_data
In #9241 there are lines like:
```
evoked_eeg_interp.data[evoked_eeg_interp.ch_names.index('EEG 004'), :]
```
We should have `evoked.get_data` like we have for epochs and raw so this can just become:
```
evoked_eeg_interp.get_data('EEG 004')
```
and become more API-consistent.

sappelhoff:
fixes #9244 

working on #9553 I thought I'd give this a shot as well. Please let me know what I should add - I can add the `units` parameter here or in #9553, depending on which one gets merged earlier.

~EDIT: the tmin/tmax params from https://github.com/mne-tools/mne-python/issues/9548 could also be added here.~ better in #9556 

To Do
- [x] have somebody tell me what's missing, or if this is really all that was needed
- [x] add whatsnew

### DIFF
--- a/mne/evoked.py
+++ b/mne/evoked.py
@@ -158,6 +158,22 @@ class Evoked(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         """Set the data matrix."""
         self._data = data
 
+    def get_data(self, picks=None):
+        """Get evoked data as 2D array.
+
+        Parameters
+        ----------
+        %(picks_all)s
+
+        Returns
+        -------
+        data : ndarray, shape (n_channels, n_times)
+            A view on evoked data.
+        """
+        picks = _picks_to_idx(self.info, picks, "all", exclude=())
+        data = self.data[picks, :]
+        return data
+
     @verbose
     def apply_function(self, fun, picks=None, dtype=None, n_jobs=1,
                        verbose=None, **kwargs):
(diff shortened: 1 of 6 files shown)


mne/evoked.py
```diff
@@ -158,6 +158,23 @@ def data(self, data):
...
+        """Get evoked data as 2D array.
+
+        Parameters
+        ----------
+        %(picks_all)s
+
+        Returns
+        -------
+        data : ndarray, shape (n_channels, n_times)
+            A view on evoked data.
```
larsoner, line 172:
```suggestion
            A view on evoked data.

        Notes
        -----
        .. versionadded:: 0.24
```

larsoner (changes requested):
Needs `latest.inc` update and a `versionadded` would be good, otherwise LGTM!

mne/evoked.py
sappelhoff: I squeezed that into the whatsnew commit. thanks

larsoner (approved):

larsoner:
pip pre failure is https://github.com/numpy/numpy/issues/19469 and we can ignore it for now

agramfort (approved):

drammock:
thanks @sappelhoff

PR #9556, opened 2021-07-12
[MRG] ENH, FIX: Add tmin/tmax parameters to get_data methods, fix bug in find_bads_ecg

### LINKED ISSUE #9548: ENH: [raw|epochs|evoked].get_data() specify tmin, tmax in seconds to get that segment

sappelhoff:
closes #9548 

~~**Work in progress**~~

To do

- [x] add to Raw
    - [x] add tests
- [x] add to Epochs
    - [x] pending #9553 to avoid conflicts
    - [x] add tests
- [x] add to Evoked
    - [x] pending #9555 because until then Evoked has no get_data method
    - [x] add tests
- [x] fix `find_bads_ecg`: https://github.com/mne-tools/mne-python/pull/9556#issuecomment-879106098
    - [x] add tests
- [x] add whatsnew (1) tmin/tmax, (2) fix of find_bads_ecg


sappelhoff (commented):
CircleCI is failing, see screenshot below - if somebody immediately knows why, a hint would be appreciated :-)

image

mne/io/base.py
```diff
@@ -833,6 +833,12 @@ def get_data(self, picks=None, start=0, stop=None,
             ``dict(grad='fT/cm', mag='fT')`` will scale the corresponding types
             accordingly, but all other channel types will remain in their
             channel-type-specific default unit.
+        tmin : float | None
+            Start time of data to get in seconds. The `tmin` parameter is
+            ignored if the `start` parameter is defined.
```
drammock, line 838: use double-backticks

sappelhoff: thank you, that was it.

sappelhoff:
I think I found a bug along the way, see this docstr of `find_bads_ecg`:

https://github.com/mne-tools/mne-python/blob/308a0abf7c8b8b14c9afc6db2e3ed5b57787be31/mne/preprocessing/ica.py#L1251-L1256

`start` and `stop` are then passed to `_make_ecg` just here:

https://github.com/mne-tools/mne-python/blob/308a0abf7c8b8b14c9afc6db2e3ed5b57787be31/mne/preprocessing/ica.py#L1305-L1308

Then in `_make_ecg`, weird things happen: If the instance is `Epochs`, `.crop` is used, interpreting start and stop as seconds (even if they might have been int, so sample indices, not seconds) ... if it's `Raw`, these are passed to `start` and `stop` of `raw.get_data`, regardless of whether they are int (sample indices, that would be fine) or float (which would raise an error, because `raw.get_data` start and stop cannot deal with floats)

https://github.com/mne-tools/mne-python/blob/308a0abf7c8b8b14c9afc6db2e3ed5b57787be31/mne/preprocessing/ecg.py#L379-L385

In particular, the error raised would be:

> TypeError: slice indices must be integers or None or have an __index__ method

Just try it yourself by modifying our ICA tutorial, add a `tstart=2` to these lines:

https://github.com/mne-tools/mne-python/blob/308a0abf7c8b8b14c9afc6db2e3ed5b57787be31/tutorials/preprocessing/40_artifact_correction_ica.py#L402-L404

That should work. But now modify it to `2.3`, which raises an error although according to the docstr it shouldn't.

mne/io/base.py
```diff
@@ -800,7 +800,7 @@ def __setitem__(self, item, value):
     @verbose
     def get_data(self, picks=None, start=0, stop=None,
                  reject_by_annotation=None, return_times=False, units=None,
-                 verbose=None):
+                 tmin=None, tmax=None, verbose=None):
```
larsoner, line 803: Easy way to prevent people from doing something silly
```suggestion
                 *, tmin=None, tmax=None, verbose=None):
```

mne/io/base.py
```diff
@@ -847,6 +853,15 @@ def get_data(self, picks=None, start=0, stop=None,
         -----
         .. versionadded:: 0.14.0
         """
+        # validate types
+        _validate_type(start, types='int', item_name='start')
+        if stop is not None:
+            _validate_type(stop, types='int', item_name='stop')
```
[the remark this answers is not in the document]
sappelhoff: @larsoner `"int"` works because of these lines: https://github.com/mne-tools/mne-python/blob/3843c36cec7310eb7b7c793e825be608f81af852/mne/utils/check.py#L391-L393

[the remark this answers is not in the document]
sappelhoff: I amended the docstr of _validate_type to mention `'int-like'` as an option for the `types` parameter.

```
    types : type | str | tuple of types | tuple of str
         The types to be checked against.
         If str, must be one of {'int', 'int-like', 'str', 'numeric', 'info',
         'path-like', 'callable'}.
         If a tuple of str is passed, use 'int-like' and not 'int' for
         integers.
```

mne/preprocessing/ecg.py
```diff
@@ -366,7 +366,24 @@ def create_ecg_epochs(raw, ch_name=None, event_id=999, picks=None, tmin=-0.5,
...
+
+         Epochs: ints are treated as seconds, NOT sample indices, floats
+                 are processed correctly, except that `ecg` and `times`
+                 seem to be out of sync.
+
+         Evoked: start and stop are erroneously not processed at all.
+
+         See: https://github.com/mne-tools/mne-python/pull/9556#issuecomment-879106098
+
+    """  # noqa: E501
```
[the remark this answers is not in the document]
sappelhoff: I think this specific bug can be solved within `_make_ecg` by checking the type of start and stop:

- if int and Raw -> use start and stop of raw.get_data
- if float and Raw -> use the new tmin and tmax of raw.get_data (same for Epochs and Evoked, which will get their tmin/tmax)
- if int and Epochs -> some custom code to index into the _data array (same for Evoked)

`_make_ecg` is only used in `find_bads_ecg` (where the current bug is), and in `find_ecg_events`, which is affected a little differently: it always passes `start=0, stop=None` to `_make_ecg` (no customization by user via params possible)

- For Raw, this is fine --> corresponds to the full signal

For other (Epochs / Evoked) this would be buggy (see Details) ... BUT `find_ecg_events` only accepts Raw, so it's fine.

So yes, I think the proposed fix would solve all instances.


<details>

- For Epochs, this is buggy --> will get the **end** of the signal right, but the beginning is always at 0 SECONDS (because .crop is used)

- For Evoked, this is fine --> "lucky accident" corresponds to full signal because Evoked start/stop is erroneously not handled in `_make_ecg`

</details>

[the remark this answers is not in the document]
sappelhoff: proposed fix in 14a70b92cbd7041c81d73e1b4d9482e662113d1a

I just need the reverse function to `inst.time_as_idx` --> so something like `inst.idx_as_time` :wink:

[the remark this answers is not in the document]
sappelhoff: if there is something in the codebase to gracefully handle this already, I'd be thankful for pointers @larsoner ... it's a bit tricky otherwise, because times can be negative (baseline), so simple division/multiplication with sfreq won't do

mne/preprocessing/ecg.py
```diff
@@ -376,14 +393,35 @@ def _make_ecg(inst, start, stop, reject_by_annotation=False, verbose=None):
...
+
+    # Handle start/stop
+    begin_param_name = 'tmin'
+    if isinstance(start, int_like):
+        if isinstance(inst, BaseRaw):
+            # Raw has start param, can just use int
+            start = 0 if start is None else start
+            begin_param_name = 'start'
+        # for Epochs/Evoked, convert time idx to time in s
+        start = 99999.
```
larsoner, line 405: I would rather just not support sample numbers for epochs/evoked. My thinking is that we really only want to / need to support sample numbers for raw data. For epochs and evoked we should always work in time units. Would this work?

sappelhoff: yes definitely ... this is just a placeholder: 999999. is a nonsensical value: What I need at that place is a `start = inst.index_to_time(start)` method, see: https://github.com/mne-tools/mne-python/pull/9556#discussion_r670472858

larsoner: > What I need at that place is

I saw that, I'm saying let's not do it at all :) I don't think we really want to start supporting index-based time-axis access for Epochs and Evoked if we can avoid it

sappelhoff: ahh okay, I'll see. it'll probably be less pretty - but let's see.

larsoner: I think you can check for int-like-ness; if not instance of raw, raise, otherwise convert. Should be just a few lines...

sappelhoff: mmh why should we raise? 

- if None --> take start/end of signal
- if int --> integer index into signal
- if float --> time (s) index into signal

for Raw, all of this is easily suported via start/stop and tmin/tmax

For Epochs/Evoked everything works as well except INT (no start/stop params available there) ... but I could just directly index into the data, right? Why not?

What I meant with not pretty is that:
1. it's "custom indexing code" and
2. need to somehow handle the case where a combination of float and int is passed.

Or are you suggesting that for Epochs/Evoked we only accept floats and None, and raise for ints?

larsoner: > Or are you suggesting that for Epochs/Evoked we only accept floats and None, and raise for ints?

Yes exactly

> For Epochs/Evoked everything works as well except INT (no start/stop params available there) ... but I could just directly index into the data, right? Why not?

Every use case we support increases our maintenance burden, and also cognitive load for users. I doubt any user really wants to index these objects using ints and if they do, their use case is rare and they can round/etc. as they see fit. I kind of regret having sample-level options even for raw, but there are at least some use cases there and we have to for backward compat.

sappelhoff: re-reading,

> Or are you suggesting that for Epochs/Evoked we only accept floats and None, and raise for ints?

you probably are :laughing: 

I'll push a commit

mne/preprocessing/ecg.py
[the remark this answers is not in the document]
sappelhoff: continued in https://github.com/mne-tools/mne-python/pull/9556#discussion_r670481055

mne/preprocessing/ecg.py
sappelhoff: okay, done --> Currently with CI Skip, because I haven't implemented tmin/tmax for Epochs yet --> will do after the "Units" for Epochs are in

sappelhoff (commented):
Okay, I think I am done here (ready for review/merge). I have one failure that I can't get to the bottom to: `mne/tests/test_epochs.py::test_own_data FAILED                           [ 50%]`

```
mne/tests/test_epochs.py:517: in test_own_data
    assert epochs._data.flags['OWNDATA']
E   assert False
```


One more point: I am using the following short snippet two-and-a-half times:

```python
        _validate_type(tmin, types=('numeric', None), item_name='tmin',
                       type_name="float, None")
        _validate_type(tmax, types=('numeric', None), item_name='tmax',
                       type_name='float, None')

        # handle tmin/tmax as start and stop indices into data array
        n_times = self.times.size
        start = 0 if tmin is None else self.time_as_index(tmin)[0]
        stop = n_times if tmax is None else self.time_as_index(tmax)[0]

        # truncate start/stop to the open interval [0, n_times]
        start = min(max(0, start), n_times)
        stop = min(max(0, stop), n_times)
```

Should I turn that into a private function in `mne.utils` and import + use it? Or is it reasonably short / repeated few enough times to just ignore "DRY"? :-)

doc/changes/latest.inc
```diff
@@ -137,6 +139,8 @@ Bugs
 
 - Fix bug in :func:`mne.io.read_raw_egi` where reading data from a data segment that is part of an acquisition skip would lead to an error (:gh:`9565` by `Eric Larson`_)
 
+- Fix bug in ::meth:`mne.preprocessing.ICA.find_bads_ecg` where passing ``start`` and ``stop`` lead to erroneous data windows depending on the combintation of Raw, Epochs, Evoked, and the type (int, float, None) of ``start`` and ``stop`` (:gh:`9556` by `Stefan Appelhoff`_)
```
sappelhoff, line 142:
```suggestion
- Fix bug in ::meth:`mne.preprocessing.ICA.find_bads_ecg` where passing ``start`` and ``stop`` lead to erroneous data windows depending on the combination of Raw, Epochs, Evoked, and the type (int, float, None) of ``start`` and ``stop`` (:gh:`9556` by `Stefan Appelhoff`_)
```

larsoner:
I would make it a private method in the TimeMixin in `base.py`, DRY and scoped to the same level as `time_as_index` (which it uses) which is nice

mne/epochs.py
```diff
@@ -1380,7 +1398,7 @@ def _get_data(self, out=True, picks=None, item=None, *, units=None,
                     data = data[:, picks]
                 if units is not None:
                     data *= ch_factors[:, np.newaxis]
-                return data
+                return data[..., start:stop]
```
larsoner, line 1401: Slices are always views (even if they are effectively otherwise a no-op) so the OWNDATA will fail

sappelhoff: so should I do a `.copy` here? :thinking:

larsoner: You could special-case `tmin = tmax = None` and not slice in that case...

sappelhoff: Thanks, that did it:

```python
            if start != 0 or stop != self.times.size:
                data = data[..., start:stop]
            return data
```

sappelhoff:
> I would make it a private method in the TimeMixin in base.py, DRY and scoped to the same level as time_as_index (which it uses) which is nice

agreed, thanks for the suggestion, this is much better!

larsoner:
Thanks @sappelhoff !

hoechenberger:
Amazing, thanks, @sappelhoff!

PR #9558, opened 2021-07-12
Accept regex for exclude in `read_raw_{e,b,g}df`

cbrnr:
So far the `exclude` parameter in `read_raw_{e,b,g}df` only accepts a list of channel names to be excluded. I think it would be nice if it also accepted a regex.

### DIFF
--- a/mne/io/edf/tests/test_edf.py
+++ b/mne/io/edf/tests/test_edf.py
@@ -512,3 +512,16 @@ def test_degenerate():
                  partial(_read_header, exclude=())):
         with pytest.raises(NotImplementedError, match='Only.*txt.*'):
             func(edf_txt_stim_channel_path)
+
+
+def test_exclude():
+    """Test exclude parameter."""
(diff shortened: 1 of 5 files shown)


mne/io/edf/tests/test_edf.py
```diff
@@ -512,3 +512,16 @@ def test_degenerate():
...
+
+def test_exclude():
+    """Test exclude parameter."""
+    exclude = ["I1", "I2", "I3", "I4"]  # list of excluded channels
+
+    raw = read_raw_edf(edf_path, exclude=["I1", "I2", "I3", "I4"])
+    for ch in exclude:
+        assert ch not in raw.ch_names
+
+    raw = read_raw_edf(edf_path, exclude = "I[1-4]")
```
larsoner, line 525:
```suggestion
    raw = read_raw_edf(edf_path, exclude="I[1-4]")
```

larsoner (approved):

larsoner:
Thanks @cbrnr

PR #9561, opened 2021-07-12
MRG, BUG: Fix behavior with constrained_layout=True

larsoner:
Fixes evoked.plot_topomap when passing axes from a fig created with `constrained_layout=True`

### DIFF
--- a/mne/viz/utils.py
+++ b/mne/viz/utils.py
@@ -1240,15 +1240,15 @@ class DraggableColorbar(object):
 
     def connect(self):
         """Connect to all the events we need."""
-        self.cidpress = self.cbar.patch.figure.canvas.mpl_connect(
+        self.cidpress = self.cbar.ax.figure.canvas.mpl_connect(
             'button_press_event', self.on_press)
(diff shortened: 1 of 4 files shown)


mne/viz/utils.py
```diff
@@ -1240,15 +1240,15 @@ def __init__(self, cbar, mappable):
 
     def connect(self):
         """Connect to all the events we need."""
-        self.cidpress = self.cbar.patch.figure.canvas.mpl_connect(
+        self.cidpress = self.cbar.ax.figure.canvas.mpl_connect(
```
larsoner, line 1243: This is just because `cbar.patch` is deprecated now, and all we need is the figure so we might as well go through `ax` instead

drammock (approved):
so the problem is with calling `subplots_adjust`?  seems like we might need similar changes elsewhere?

agramfort (approved):
@drammock merge if happy

larsoner:
> so the problem is with calling subplots_adjust? seems like we might need similar changes elsewhere?

Yes and yes. I'm fine with doing the updates piecemeal as we encounter them. This at least takes care of `plot_topomap`

PR #9565, opened 2021-07-13
MRG, BUG: Fix MFF reading with skips

larsoner:
On `main` the added test yields a variant of:
<details>

```
mne/io/egi/tests/test_egi.py:118: in test_egi_mff_pause_chunks
    raw.save(fname_temp)
<decorator-gen-192>:24: in save
    ???
mne/io/base.py:1449: in save
    _write_raw(fname, self, info, picks, fmt, data_type, reset_range,
mne/io/base.py:2108: in _write_raw
    final_fname = _write_raw_fid(
mne/io/base.py:2197: in _write_raw_fid
    data, times = raw[picks, first:last]
mne/io/base.py:773: in __getitem__
    return self._getitem(item)
mne/io/base.py:780: in _getitem
    data = self._read_segment(start=start, stop=stop, sel=sel,
<decorator-gen-185>:24: in _read_segment
    ???
mne/io/base.py:419: in _read_segment
    _ReadSegmentFileProtector(self)._read_segment_file(
mne/io/base.py:2037: in _read_segment_file
    return self.__raw.__class__._read_segment_file(
mne/io/egi/egimff.py:658: in _read_segment_file
    one[eeg_one, disk_use_idx[s_start:s_end]] = block_data[eeg_in]
E   ValueError: shape mismatch: value array of shape (257,250) could not be broadcast to indexing result of shape (257,0)
```

</details>

This PR fixes that, and by changing a `<` to `<=` makes a tiny efficiency gain where we don't accidentally read a buffer and then use zero samples from it.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -127,7 +127,9 @@ Bugs
 
 - Fix bug in :func:`mne.concatenate_epochs` when concatenating :class:`mne.Epochs` objects with 0 events (:gh:`9535` by `Marijn van Vliet`_)
 
-- Fix bug in :func:`mne.viz.Brain.screenshot` where the RGBA mode was not supported (:gh:`9561` by `Guillaume Favelier`_).
+- Fix bug in :func:`mne.viz.Brain.screenshot` where the RGBA mode was not supported (:gh:`9564` by `Guillaume Favelier`_).
+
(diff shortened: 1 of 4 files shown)


larsoner:
pip pre failure is https://github.com/numpy/numpy/issues/19469 and we can ignore it for now

agramfort (approved):

PR #9568, opened 2021-07-14
matrix-based head decimation

kingjr:
Head decimation can be *very* slow when the original file is highly resolved (e.g. from a laser scan) because it uses a loop.

This fixes it.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -43,6 +43,8 @@ Enhancements
 
 - Speed up BCD solver in :func:`mne.inverse_sparse.mixed_norm` by adding Anderson acceleration (:gh:`9481` by **new contributor** |Pierre-Antoine Bannier|_ and `Alex Gramfort`_)
 
+- Speed up point decimation in :func:`mne.io.read_raw_kit` by vectorization and use of :class:`sklearn.neighbors.BallTree` (:gh:`9568` by `Jean-Remi King`_ and `Eric Larson`_)
+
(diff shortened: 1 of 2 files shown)


agramfort (approved):
@larsoner or @GuillaumeFavelier merge if happy

larsoner:
@kingjr in local testing using `cKDTree` sped things up by a factor of about two, can you try it on your real data and see if there are also gains there? You can easily switch between the two implementations by commenting/uncommenting the leading `"""`

larsoner:
So on `main` this code (with ~3MB hsp file):
```
import os.path as op
import mne
data_dir = op.dirname(mne.__file__) + '/io/kit/tests/data/'
mrk_path = op.join(data_dir, 'test_mrk.sqd')
elp_path = op.join(data_dir, 'test.elp')
hsp_fname = 'big_dig.txt'
raw_fname = data_dir + 'test.sqd'
mne.io.read_raw_kit(raw_fname, elp=elp_path, mrk=mrk_path, hsp=hsp_fname)
```
Takes 8.6 sec inside `_decimate_points`. On 76d4f15 (your commit) it's 9.5 sec, so it looks like the original commits did not help, at least on my machine. On f632d72 it was down to 2.4 sec. The rest of the time was spent in `use_pts = pts[mid_idx == mi]`, which we can easily avoid using `np.argsort` (it otherwise includes a ton of redundant comparisons, since they're mutually exclusive). I pushed 4d72188 to take `_decimate_points` down to 0.3 sec, which is hopefully fast enough for you :)

kingjr:
Great, thanks!

agramfort:
thx @kingjr and @larsoner

PR #9570, opened 2021-07-14
add HPI SNR calculation and plotting

drammock:
adds computation of SNR of the HPI coils.

### DIFF
--- a/mne/chpi.py
+++ b/mne/chpi.py
@@ -962,8 +962,7 @@ def compute_chpi_snr(raw, t_step_min=0.01, t_window='auto', ext_order=1,
     raw : instance of Raw
         Raw data with cHPI information.
     t_step_min : float
-        Minimum time step to use. If correlations are sufficiently high,
-        t_step_max will be used.
+        Minimum time step to use.
     %(chpi_t_window)s
     %(chpi_ext_order)s
     %(raw_tmin)s
@@ -1000,8 +999,7 @@ def compute_chpi_amplitudes(raw, t_step_min=0.01, t_window='auto',
     raw : instance of Raw
         Raw data with cHPI information.
     t_step_min : float
-        Minimum time step to use. If correlations are sufficiently high,
-        t_step_max will be used.
+        Minimum time step to use.
     %(chpi_t_window)s
     %(chpi_ext_order)s
     %(raw_tmin)s
--- a/mne/chpi.py
+++ b/mne/chpi.py
@@ -30,7 +30,7 @@ from .io.kit.constants import KIT
 from .io.kit.kit import RawKIT as _RawKIT
 from .io.meas_info import _simplify_info, Info
 from .io.pick import (pick_types, pick_channels, pick_channels_regexp,
-                      pick_info)
+                      pick_info, _picks_to_idx, _get_channel_types)
 from .io.proj import Projection, setup_proj
 from .io.constants import FIFF
 from .io.ctf.trans import _make_ctf_coord_trans_set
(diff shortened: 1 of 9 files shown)


mne/chpi.py
```diff
@@ -678,17 +685,19 @@ def _fit_chpi_amplitudes(raw, time_sl, hpi):
         if not (n_on >= 3).all():
             return None
     return _fast_fit(this_data, hpi['proj_op'], len(hpi['freqs']),
-                     hpi['model'], hpi['inv_model_reord'])
+                     hpi['model'], hpi['inv_model'], hpi['inv_model_reord'],
+                     hpi['mag_picks'], hpi['grad_picks'], return_snr)
 
 
-@jit()
```
larsoner, line 684: We definitely want this back in there eventually, optimizing these is important. To avoid making numba angry (which is probably due to differing `return` statements), you can make this param `compute_snr` and if it's False, just return all-zero arrays of the correct size.

drammock: that was an accidental omission, I meant to put it back.  (removed it while debugging)

drammock:
still a draft because I don't have tests yet, but I think it's worth another quick look @larsoner.  Locally it passes existing tests and runs under Numba / jit, so right now I'm just looking for general feedback on whether the API makes sense / if you spot any obvious non-DRY parts of the code.

mne/chpi.py
```diff
@@ -590,30 +590,39 @@ def _setup_hpi_amplitude_fitting(info, t_window, remove_aliased=False,
...
+    model = _setup_hpi_glm(hpi_freqs, line_freqs, info['sfreq'], window_nsamp)
     inv_model = np.linalg.pinv(model)
     inv_model_reord = _reorder_inv_model(inv_model, len(hpi_freqs))
     proj, proj_op, meg_picks = _setup_ext_proj(info, ext_order)
-
+    # include mag and grad picks separately, for SNR computations
+    all_ch_types = _get_channel_types(info, unique=True)
+    meg_types = all_ch_types.intersection({'mag', 'grad'})
+    mag_picks = (_picks_to_idx(info, 'mag') if 'mag' in meg_types else
+                 np.array([], dtype=int))
```
larsoner, line 602:
```suggestion
    mag_picks = _picks_to_idx(info, 'mag', allow_empty=True)
```

mne/chpi.py
```diff
@@ -590,30 +590,39 @@ def _setup_hpi_amplitude_fitting(info, t_window, remove_aliased=False,
...
+    grad_picks = (_picks_to_idx(info, 'grad') if 'grad' in meg_types else
+                  np.array([], dtype=int))
```
larsoner, line 604:
```suggestion
    grad_picks = _picks_to_idx(info, 'grad', allow_empty=True)
```

larsoner (commented):
Otherwise seems like a reasonable approach to me. CIs appear angry about docstrings

drammock:
@larsoner this is still WIP because I want to add the new viz function to an example/tutorial. But meanwhile, there seems to be some numerical instability when dealing with gradiometers.  Here are plots of 4 SNR computations of the same data, run one after another in the same ipython session:

Figure_4
Figure_3
Figure_2
Figure_1

Is this likely to be a problem with gradiometers in general? Or something weird about the sample data?  Here's the snippet:

```py
import os.path as op
import mne

data_path = mne.datasets.testing.data_path(download=False)
chpi_fif_fname = op.join(data_path, 'SSS', 'test_move_anon_raw.fif')
raw = mne.io.read_raw_fif(chpi_fif_fname, allow_maxshield='yes')
for ix in range(1, 5):
    result = mne.chpi.compute_chpi_snr(raw)
    fig = mne.viz.plot_chpi_snr(result)
    fig.savefig(f'figure_{ix}.png')
plt.close('all')
```

drammock:
cc @jjnurminen, as this implementation is based on your hpimon code.  Maybe you have some insight into the numerical instability seen above?

mne/chpi.py
```diff
@@ -976,14 +1067,26 @@ def compute_chpi_amplitudes(raw, t_step_min=0.01, t_window='auto',
...
+    if snr:
+        del sin_fits['proj']
+        sin_fits['freqs'] = hpi['freqs']
+        ch_types = raw.get_channel_types()
+        grad_offset = 3 if 'mag' in ch_types else 0
+        for ch_type in ('mag', 'grad'):
+            if ch_type in ch_types:
+                for key in ('snr', 'power', 'resid'):
+                    cols = 1 if key == 'resid' else n_freqs
+                    sin_fits[f'{ch_type}_{key}'] = np.empty((n_times, cols))
```
larsoner, line 1085: This is the non-deterministic part. `np.empty` will not initialize to anything, so you just get whatever junk was in that block of memory that was allocated. When I change this to `np.zeros`, I get flat lines for all five runs:

Screenshot from 2021-07-20 12-19-35


So it looks like this never gets values assigned to it for grads for some reason (?).

drammock:
I ask

> Is this likely to be a problem with gradiometers in general? Or something weird about the sample data?

and as usual, the answer is "neither, it is a bug in the code that I wrote" 😆

drammock:
copy-paste error, was overwriting the mag results with grad results, so grad results remained as whatever `np.empty` initialized to.

drammock:
OK I think this is ready for review @larsoner.  A couple questions to prime the pump:

1. for compatibility with `Report`, does the plotting function need to accept an `axes` parameter?
2. I'm worried that the tutorial is getting too computationally long:
```
computation time summary:
    - ../tutorials/preprocessing/59_head_positions.py:                        22.89 sec   237.9 MB
```
Do you think it makes sense to do the SNR demo in a short example and cross-ref it from the tutorial?

larsoner:
> for compatibility with Report, does the plotting function need to accept an axes parameter?

I'm not sure about Report, but most of our plotting functions support this param so I'd add it in this PR anyway.

> I'm worried that the tutorial is getting too computationally long:

20s isn't so long, but if you can make the same point by operating on a cropped segment of data (and I think you probably can?) then that's worth doing

mne/chpi.py
```diff
@@ -913,6 +952,44 @@ def _unit_quat_constraint(x):
...
+                     tmin=0, tmax=None, verbose=None):
+    """Compute time-varying estimates of cHPI SNR.
+
+    Parameters
+    ----------
+    raw : instance of Raw
+        Raw data with cHPI information.
+    t_step_min : float
+        Minimum time step to use. If correlations are sufficiently high,
+        t_step_max will be used.
```
larsoner, line 966: there is no t_step_max?

mne/viz/misc.py
```diff
@@ -1308,3 +1308,54 @@ def plot_csd(csd, info=None, mode='csd', colorbar=True, cmap=None,
...
+
+def plot_chpi_snr(snr_dict):
+    """Plot time-varying SNR estimates of the HPI coils.
+
+    Parameters
+    ----------
+    snr_dict : dict
+        The dictionary returned by `~mne.chpi.compute_chpi_snr`. Must have keys
+        ``times``, ``freqs``, ``TYPE_snr``, ``TYPE_power``, and ``TYPE_resid``
+        (where ``TYPE`` can be ``mag`` or ``grad`` or both).
```
larsoner, line 1321: How about an `axes` argument that must be the same length as `snr_dict` (3.7+ dicts are always ordered) and in the same order? Internally we can (and probably should) always construct in the order `'snr', 'power', 'residual'` and note this in the docstring here

larsoner: (the alternative is to do something like an `axes` dict but that's annoying when almost all the time people will want to plot in the order snr/power/resid, and this will correspond to the `axes` in order one gets from a `fig, axes = plt.subplots(3)` call for example)

larsoner (commented):
Looks good to me other than ideas for how to address your comments

mne/chpi.py
drammock: I meant to ask about that...  it's a copy-paste from the `compute_chpi_amplitudes` docstring, which also doesn't have a `t_step_max`...  I was hoping you would know why

mne/viz/misc.py
drammock: +1 for testing `len(axes) == len(snr_dict) - 2` (because of `freqs` and `times` entries).  Also I'm fine with your proposed ordering... right now it's just alphabetical.

drammock:
CIs all green! Here's the new tutorial section: https://31423-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/59_head_positions.html#computing-snr-of-the-hpi-signal

mne/chpi.py
```diff
@@ -923,8 +999,7 @@ def compute_chpi_amplitudes(raw, t_step_min=0.01, t_window='auto',
     raw : instance of Raw
         Raw data with cHPI information.
     t_step_min : float
-        Minimum time step to use. If correlations are sufficiently high,
-        t_step_max will be used.
+        Minimum time step to use.
```
larsoner, line 1002: Indeed the old description was a bit misleading, it's from when there was a single monolithic function. Now that head pos calculation is split into three separate steps, the stuff about `t_step_max` doesn't make much sense here

larsoner:
LGTM other than needing `latest.inc` with your name and @jjnurminen's

drammock:
found a bug in the plotting legend, which shows up in the tutorial.  Fixing now.

drammock:
CIS are all green again, and the legend is fixed:  https://31437-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/59_head_positions.html#computing-snr-of-the-hpi-signal

agramfort (approved):

larsoner:
Thanks @drammock !

PR #9575, opened 2021-07-16
[MRG, ENH] Get annotation description from snirf stim name

### LINKED ISSUE #9571: Get annotation descriptions from the stim name when reading snirf files

When loading a SNIRF file via `mne.io.read_raw_snirf`, the descriptions for `Annotations` are currently extracted from the [keys of the SNIRF stim indexed groups](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimj). According to the [SNIRF file format spec](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md), what is currently being extracted as the Annotation description is the index of the stim rather than a descriptive string.

> group: a structure containing sub-fields (defined in the H5G object class). Arrays of groups, also known as the indexed-groups, are denoted with numbers at the end (e.g. /nirs/data1, /nirs/data2) starting with index 1. Array indices should be contiguious with no skipped values (an empty group with no sub-member is permitted).

A better field to get a description from is the [stimulus group `name` field](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimjname) which is described as:

> This is a string describing the jth stimulus condition.

The proposed change is to get `Annotation` descriptions from the stimulus name rather than the index.

Update [`RawSNIRF.__init__` lines 362:369](https://github.com/mne-tools/mne-python/blob/3843c36cec7310eb7b7c793e825be608f81af852/mne/io/snirf/_snirf.py#L362-L369
) to take the description field from the `name` field of the stim group rather than the stim group key.

The descriptions can be left as-is. Although the match between an `Annotation` description and a SNIRF stimulus name isn't perfect, it seems like an improvement over using the stimulus index as-is.

If the description is not changed, then tests for the MNE-NIRS SNIRF writer will need to be changed (see [Additional comments](#additional-comments) below).

#### Additional comments

This is related to https://github.com/mne-tools/mne-nirs/pull/326, where the MNE-NIRS SNIRF writer is being updated to be more consistent with the SNIRF spec.

MNE-NIRS currently has tests which use the MNE-Python SNIRF reader to make a round trip write/read test. Resolving this issue will ensure that the annotation descriptions remain the same after a round trip.

dsleiter:
before submitting.

#### Reference issue
Fixes #9571

Uses the 'name' field from SNIRF file stim groups as the description for Annotations rather than the index of the stim group.

#### Additional information
This is in response to https://github.com/mne-tools/mne-nirs/pull/326

This could be a breaking change for anyone who depends on annotation descriptions when reading from SNIRF files that have not been created from the update to the MNE-NIRS SNIRF writer. I'd like some guidance on whether or not I should implement a deprecation or not. I listed this as an ENH because it isn't obvious that this is a bug fix, but it is in response to a fix of what is a bug in the SNIRF writer in order to maintain consistency.

### DIFF
--- a/mne/io/snirf/_snirf.py
+++ b/mne/io/snirf/_snirf.py
@@ -365,7 +365,8 @@ class RawSNIRF(BaseRaw):
                     data = np.atleast_2d(np.array(
                         dat.get('/nirs/' + key + '/data')))
                     if data.size > 0:
-                        annot.append(data[:, 0], 1.0, key[4:])
+                        desc = dat.get('/nirs/' + key + '/name')[0]
+                        annot.append(data[:, 0], 1.0, desc.decode('UTF-8'))
             self.set_annotations(annot)
 
             # Reorder channels to match expected ordering in MNE
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -31,9 +31,13 @@ Current (0.24.dev0)
 
 .. |Pierre-Antoine Bannier| replace:: **Pierre-Antoine Bannier**
 
+.. |Darin Erat Sleiter| replace:: **Darin Erat Sleiter**
+
 Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
+
+- Get annotation descriptions from the name field of SNIRF stimulus groups when reading SNIRF files via `mne.io.read_raw_snirf` (:gh:`9575` **by new contributor** |Darin Erat Sleiter|_)
 
 - Add support for NIRSport and NIRSport2 devices to `mne.io.read_raw_nirx` (:gh:`9348` and :gh:`9401` **by new contributor** |David Julien|_, **new contributor** |Romain Derollepot|_, `Robert Luke`_, and `Eric Larson`_)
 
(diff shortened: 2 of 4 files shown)


dsleiter:
I believe this is ready for review!

mne/io/snirf/_snirf.py
```diff
@@ -365,7 +365,8 @@ def natural_keys(text):
                     data = np.atleast_2d(np.array(
                         dat.get('/nirs/' + key + '/data')))
                     if data.size > 0:
-                        annot.append(data[:, 0], 1.0, key[4:])
+                        desc = dat.get('/nirs/' + key + '/name')[0]
```
rob-luke, line 368: For anyone else reviewing, this corresponds to https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimjname

rob-luke:
Thanks @dsleiter, this looks good to me. Lets see if we can get a review from @larsoner too.

As there is a matching PR at https://github.com/mne-tools/mne-nirs/pull/326 I suggest we wait till this PR has all green CI, then I will pull both PR branches and test locally (I will test a few real files have no loss of info when converted back and forth). Then once once all CIs are green, my local tests are passing, and we have a few review approvals we merge. Ok?

rob-luke (approved):
I tested this locally with the associated MNE-NIRS branch and all results seemed correct to me. I also tested locally with some real measurements.

larsoner:
Thanks @dsleiter !

dsleiter:
Thanks @rob-luke and @larsoner!

PR #9578, opened 2021-07-16
[MRG] Add `units` parameter to get_data for Evoked

sappelhoff:
follow up to #9553 and #9555 

This should conclude the recent series on PRs for `get_data` methods for Raw, Epochs, and Evoked, and adding tmin/tmax, and units params for them :slightly_smiling_face: 

Also fixes some `:gh:` directives (broken links) and removes a couple of unneeded `np.testing.` prefixes.


mne/evoked.py
```diff
@@ -159,12 +159,13 @@ def data(self, data):
         self._data = data
 
     @fill_doc
-    def get_data(self, picks=None, tmin=None, tmax=None):
+    def get_data(self, picks=None, units=None, tmin=None, tmax=None):
```
larsoner, line 162: This is a within-cycle change (you just added this in one of those last PRs), right? Just want to make sure we aren't breaking existing code that has been in a release like `.get_data(picks, tmin, tmax)`

larsoner (approved):

mne/evoked.py
sappelhoff: yup, get_data for Evoked is completely new --> and I did it this way, because it aligns with how units, tmin, tmax are ordered in Epochs.get_data

larsoner: Great, feel free to merge once CIs come back happy as this seems pretty straightforward!

larsoner:
Thanks @sappelhoff !

Issue #9577, opened 2021-07-16
[BUG] `mne.label.find_pos_in_annot` doesn't check `coord_frame`


alexrockhill:
It doesn't seem like a good idea to allow any position to be passed to `find_pos_in_annot` without checking that it is in the `mri` coordinate frame. If someone has the wrong coordinate frame, the error will be silent and they will get the wrong ROI which to me seems like a pretty good risk for incorrect results that could be avoided by enforcing the coordinate frame or maybe rather making this a class function of `Dipole`.

larsoner:
I don't think `Dipole` would be the right subclass, in general the point you're talking about could be anything (iEEG sensor, part of a mesh, an elementary current dipole, etc.). I think we should do two things:

1. Document clearly that if an `ndarray` or list-like is passed, these must be in the MRI coordinate frame
2. Add support for other objects as we need them, such as DigPoint instances (?), where we can check the coordinate frame

FWIW `Dipole` always lives in the *head* coordinate frame, so if we wanted to add support for that in (2) then we'd have to have a `trans` argument in that function, if we don't already.

But really I think we should do the doc update (1) first and then let actual use cases guide (2)

alexrockhill:
Maybe it should be in a mixin for things with positions?

larsoner:
What objects + use cases do you have in mind?

alexrockhill:
I think just `DigMontage` and `Dipole` unless you know more that have positions... now that I say that it might be better just to make an ``inst`` that takes either

larsoner:
`Dipole` is always in head coords and I don't think it's a common use case anyway. Let's just start from `montage` (and not worry about Dipole) since it's the use case we actually want to support for iEEG?

PR #9579, opened 2021-07-16
[ENH, MRG] Fix find aseg for Dipole

alexrockhill:
Fixes https://github.com/mne-tools/mne-python/issues/9577.

https://github.com/mne-tools/mne-python/pull/9545 does this for `DigMontage`s and so I think the only use case for this is `Dipole`s. There is already a `to_mni` function, I think there should just be an `to_aseg` function as well, that makes sense to me.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -37,7 +37,7 @@ Enhancements
 
 - Add support for NIRSport and NIRSport2 devices to `mne.io.read_raw_nirx` (:gh:`9348` and :gh:`9401` **by new contributor** |David Julien|_, **new contributor** |Romain Derollepot|_, `Robert Luke`_, and `Eric Larson`_)
 
-- New function :func:`mne.Dipole.to_aseg_labels` to get atlas label for MRI coordinates.  (:gh:`9376` by **by new contributor** |Marian Dovgialo|_)
+- New function :func:`mne.Dipole.to_volume_labels` to get atlas label for MRI coordinates.  (:gh:`9376` by **by new contributor** |Marian Dovgialo|_)
 
 - Add support for SURE parameter selection in :func:`mne.inverse_sparse.mixed_norm` and make ``alpha`` parameter now default to ``'sure'`` (:gh:`9430` by **new contributor** |Pierre-Antoine Bannier|_ and `Alex Gramfort`_)
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -37,7 +37,7 @@ Enhancements
 
 - Add support for NIRSport and NIRSport2 devices to `mne.io.read_raw_nirx` (:gh:`9348` and :gh:`9401` **by new contributor** |David Julien|_, **new contributor** |Romain Derollepot|_, `Robert Luke`_, and `Eric Larson`_)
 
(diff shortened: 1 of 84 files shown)


larsoner:
Do you need `Dipole.to_aseg`? I'd rather not add functionality until people have use cases that need to be met

larsoner:
(in other words, assume YAGNI until proven otherwise)

alexrockhill:
It's in the dipole tutorial that uses `label.find_pos_in_annot/aseg`. It's related to the mentioned issue.

This can be closed without merging but I made the changes and was going to keep it open for a minute until the issue is resolved or we just say it's not a big deal. I think it's actually a pretty big liability though because the coordinate frames are not that easy to understand.

alexrockhill:
Okay @larsoner, maybe take a quick look at this diff, it refactors `find_label_in_annot/aseg` to share `_freesurfer.py` code with #9545 and it makes the API much more consistent since there was already a `dip.to_mni` it makes sense to make a `dip.to_mri`.

mne/label.py
```diff
@@ -2,7 +2,7 @@
 #          Martin Luessi <mluessi@nmr.mgh.harvard.edu>
 #          Denis Engemann <denis.engemann@gmail.com>
 #
-# License: BSD-3-Clause
+# License: BSD (3-clause)
```
larsoner, line 5: Revert, we settled on BSD-3-Clause in #9569

mne/tests/test_label.py
```diff
@@ -1,6 +1,6 @@
 # Author: Eric Larson <larson.eric.d@gmail.com>
 #
-# License: BSD-3-Clause
+# License: BSD (3-clause)
```
larsoner, line 3: Revert

larsoner (commented):
@mdovgialo is this API change okay with you?

mne/label.py
alexrockhill: Sorry incidental

alexrockhill:
Weird, the test passes locally for me and it hasn't been effected by the changes...

alexrockhill:
And it passed before I fixed the licenses...

mdovgialo:
> @mdovgialo is this API change okay with you?
I'm fine with this

doc/changes/latest.inc
```diff
@@ -37,7 +37,7 @@ Enhancements
 
 - Add support for NIRSport and NIRSport2 devices to `mne.io.read_raw_nirx` (:gh:`9348` and :gh:`9401` **by new contributor** |David Julien|_, **new contributor** |Romain Derollepot|_, `Robert Luke`_, and `Eric Larson`_)
 
-- New function :func:`mne.label.find_pos_in_annot` to get atlas label for MRI coordinates.  (:gh:`9376` by **by new contributor** |Marian Dovgialo|_)
+- New function :func:`mne.Dipole.to_aseg_labels` to get atlas label for MRI coordinates.  (:gh:`9376` by **by new contributor** |Marian Dovgialo|_)
```
larsoner, line 40: Same as in #9545 with @agramfort , I'm +1 on making the `to_volume_labels` because this is the verbiage we use elsewhere in MNE

larsoner (commented):
Other than the naming issue we should be good here, the failure is unrelated

agramfort (approved):

alexrockhill:
Looks good to me :)

PR #9585, opened 2021-07-18
[ENH, MRG] Add `mne.viz.Brain.plot_sensors` and refactor `mne.viz.plot_alignment`

alexrockhill:
This adds simple sphere plotting to `mne.viz.Brain` for sensors. The color is controllable by channel name and type as is scale and alpha/opacity.

Hopefully this is helpful to @GuillaumeFavelier to build off for a more advanced version in https://github.com/mne-tools/mne-python/pull/8803, I didn't mean to supplant that PR only to provide a simple version to build from.

Here is a simple script to view the results on the sample data:

```py
import os.path as op
import numpy as np
import mne

evk = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
# should not be needed but see #9638
brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, units='mm')
brain.add_skull()
brain.add_head()
brain.add_sensors(evk.info,  trans=trans)

# to compare
mne.viz.plot_alignment(evk.info, surfaces=['brain', 'head'], subject='sample',
                                         subjects_dir=subjects_dir, trans=trans, meg=True, dig=True)
```

Closes #8803


mne/viz/_brain/_brain.py
```diff
@@ -2368,6 +2372,76 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        colors : str | dict
+            The color of the sensors. If dict, channel names or types can be
+            keys with values of matplotlib colors. Default is all sensors
+            yellow.
+        scales : float | dict
+            The size of the sensors. If dict, channel names or types can be
+            keys with values for the size in mm. Default is all sensors 5 mm.
+        alphas : float | dict
+            The opacity of the sensors or, if dict, each sensor by name or
+            type. Default is all sensors 0.5 opacity.
```
larsoner, line 2397: This API (and code) should be refactored with `mne.viz.plot_alignment` to the extent possible. We many options for plotting sensor types there:

- MEG with helmet, or just as sensors
- EEG as original, or projected to the scalp
- fNIRS in many configurations

These are all useful for different use cases. If we're going to have a generic `plot_sensors` function, we should port that functionality (for usability) and API (for consistency) over as much as possible.

mne/viz/_brain/_brain.py
```diff
@@ -2368,6 +2372,76 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+                                     'the "meg" or head" coordinate frames')
+                trans = _ensure_trans(trans, 'head', 'mri')
+                ch_coord = apply_trans(trans['trans'], ch_coord) * 1000
+                coord_frame = FIFF.FIFFV_COORD_MRI  # now in surface RAS
+            if coord_frame != FIFF.FIFFV_COORD_MRI:
+                raise RuntimeError(
+                    '`info` coordinate frame must be "meg", "head" or "mri", '
+                    f'got {coord_frame} for {info.ch_names[idx]}. You may '
+                    'want to use `ch_types=dict(meg=True, eeg=True)` '
+                    'for instance')
```
larsoner, line 2427: Same comment here, a lot of work has gone into `plot_alignment` to do coordinate frame gymnastics, so code should be deduplicated / refactored to be shared between the functions

larsoner (commented):
My major comment is that I think even without plotting any data values I think this is / should be a larger undertaking involving deduplication (and shared functionality) with plot_alignment. I'd rather not add new code that does 20-50% of what `plot_alignment` allows for plotting sensors using totally different code (and somewhat different API) to accomplish it...

larsoner:
... one long-term view / option would be to make `Brain` the basis for `plot_alignment`, so that someday `plot_alignment` is just a think wrapper around `Brain` constructor + methods. But I think this is a bit of work, so for now I'd rather stick with just calling `plot_alignment` and `Brain` separately for the sensor + brain parts, respectively. Maybe once the GUI is done you could work on such a refactoring @alexrockhill :)

alexrockhill:
> My major comment is that I think even without plotting any data values I think this is / should be a larger undertaking involving deduplication (and shared functionality) with plot_alignment. I'd rather not add new code that does 20-50% of what `plot_alignment` allows for plotting sensors using totally different code (and somewhat different API) to accomplish it...

So I agree and ideally this would be refactored but, after looking at the `plot_alignment` code, it's kind of this beast of a monolithic block and not modularized at all which made it hard for me to understand what's going on/refactor. The easiest way to do it seems to me from the ground up with replicating results. I might suggest that a smaller PR like this can add the functionality on the way to the larger undertaking like you said about major refactoring to share much more code. I'll work on refactoring a bit more to share, but since `plot_alignment` can be in "head" or "meg" as well, it actually uses a much more complicated alignment protocol that I don't think ports over well at all.

larsoner:
> I might suggest that a smaller PR like this can add the functionality on the way to the larger undertaking like you said about major refactoring to share much more code

In general I don't like this approach. To me it's much cleaner to take the existing code and refactor it first. Then everything is modularized and understandable, and no functionality has changed. Then the next PR can be to reuse code from the private function(s) now called by `plot_alignment` in a new `add_sensors` method.

alexrockhill:
> In general I don't like this approach. To me it's much cleaner to take the existing code and refactor it first. Then everything is modularized and understandable, and no functionality has changed. Then the next PR can be to reuse code from the private function(s) now called by `plot_alignment` in a new `add_sensors` method.

Ok, I'll take a shot at refactoring in this PR.

larsoner:
Sure but let's get the GUI and everything working first. It's a higher priority, and even though combining `plot_alignment` with `Brain` is clunky, it works. So this `add_sensors` to me is mostly just a nicer syntax, not new functionality that we need...

alexrockhill:
> Sure but let's get the GUI and everything working first. It's a higher priority, and even though combining `plot_alignment` with `Brain` is clunky, it works. So this `add_sensors` to me is mostly just a nicer syntax, not new functionality that we need...

For sure, I didn't mean right now :)

mne/viz/_brain/_brain.py
```diff
@@ -2368,6 +2372,76 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
                                   opacity=alpha, resolution=resolution)
             self._renderer.set_camera(**views_dicts[hemi][v])
 
+    @fill_doc
+    def add_sensors(self, info, trans=None, ch_types=None, colors=None,
```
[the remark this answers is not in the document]
alexrockhill: Why can't you just pass the `Evoked.info`?

mne/viz/_brain/_brain.py
alexrockhill: What would you think about making fNIRS it's own "add" function? It seems different enough and all the overloads on dictionaries in `plot_alignment` seem really confusing to me. I think both on the user and developer side it's easier to understand and maintain if the functions are a bit smaller in scope.

larsoner: To me the API `eeg=..., meg=..., fnirs=...` is simple enough. I'd rather have this than `add_sensors` and `add_sensors_fnirs`, to me it makes more sense to have the fnirs options in a keyword like this. The API for `fnirs` was hashed out with fNIRS people in the last year or two over a few PRs, so it should be "good enough" for them. But maybe  @rob-luke has thoughts / regrets ...

mne/viz/_brain/_brain.py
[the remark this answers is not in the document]
larsoner: My point is more that in #8803 we want to be able to plot *data on the sensors* (or maybe via some fancy mapping for MEG at least), so we need some way to have info *and* data to be #8803-compatible. One way to do this is to rename here `info` to `inst`, and allow it to be `Evoked | Info`, and if it's Evoked, we just take the `.info` for now, but it's forward-compatible with eventually doing some mapping of the data to the sensors.

mne/viz/_brain/_brain.py
alexrockhill: That's persuasive enough for me, I'll take a quick shot at adding that and then we can either merge or pause on this for a while as I was waiting for the BEM refactoring to be done so I think I'll go back to getting going on the GUI.

mne/viz/_brain/_brain.py
alexrockhill: So there's a lot more gymnastics in `plot_alignment` because the plot can be in device, head or mri coordinates. It makes sense to me that everything should just be in mri/surface RAS but that would be a depreciation. What do you think, make `mne.viz.Brain` support coordinate frame shifts or deprecate the coordinate frame support in `plot_alignment` or a third option?

larsoner: > It makes sense to me that everything should just be in mri/surface RAS but that would be a depreciation

Not always, this is why we provide the option. As a rule in MNE we try to only provide options that people need, and it's useful to be able to align to each coordinate frame (esp. for example in combined M-EEG recordings where you can look at alignment with no MRI whatsoever!).

> make mne.viz.Brain support coordinate frame

This can/should just stay MRI-only since it's brain/recon-centric

> a third option?

Yes:

1. Make a private helper function `_convert_sensors_to(info, ..., coord_frame)` that converts sensor coordinates from whatever coordinate frame they're natively in to whatever coordinate frame they need to be in for plotting. (This code exists already in `plot_alignment` and "just" needs to be pulled out.) 
2. Have plot_alignment call this, with `coord_frame=coord_frame`
3. Have `Brain` call this with `frame_to='mri'`

This kind of kills the dream of having `plot_alignment` be a thin wrapper to Brain, but I'm okay with that. It's 98% as readable and maintainable if both just call a set of shared private functions like this, and we keep an optimal user-facing API in terms of coordinate frames (i.e., allow option to change it in `plot_alignment` where it is useful, always use MRI for Brain since that's brain/recon-centric).

mne/viz/_brain/_brain.py
rob-luke: I would prefer not to have a plot_sensors_fNIRS function, and instead have a fNIRS argument like we currently do. I think the current system works fine, has there been some issues with the current API?

I'm happy to test any new code with fNIRS data for you. Just ping me.

rob-luke: > I think both on the user and developer side it's easier to understand and maintain if the functions are a bit smaller in scope.

I disagree. From a user standpoint, I think it would be confusing if some types work with plot_sensors, but others don't and require a special function. Users don't see the internal wrangling, and just expect a function to work with data regardless if it's eeg, meg, fnirs.

alexrockhill:
Ok, I think this does some really nice refactorization and makes `plot_sensors` complete as far as the functionality of `plot_alignment`. The tutorial will fail with a warning about the coordinate frame being unknown for the fNIRS and SEEG channel location examples and who-knows-how-many if/when all the `plot_alignment`s are switched over. I think best would be to fix the `set_montage` and I can prioritize that first as that will probably solve many of the examples.

I think the coordinate frame handling is a big improvement and clearer to me at least now.

rob-luke:
@alexrockhill MY APPOLOGIES!! I meant to open a PR to your PR, but instead I have just pushed directly to your branch.

rob-luke:
Mistake aside, I think the new API significantly simplified the example I just pushed. However, the colours of the optodes have changed. Is this intentional? If so, we should update the tutorial text. But I think the colours for the new version are too similar between sources and detectors (both look red to me).


### Old colors

image


### New colors

image


I also noticed that distance is now in mm? But was previously in m? Was this change also intentional?

mne/viz/_brain/_brain.py
alexrockhill: This was more in regards to adding the head separately and the sensors separately etc. I already pushed to put fNIRS in

rob-luke: Awesome, appreciate it!

alexrockhill:
> @alexrockhill MY APPOLOGIES!! I meant to open a PR to your PR, but instead I have just pushed directly to your branch.

That's okay, you're welcome to push commits.

alexrockhill:
> Mistake aside, I think the new API significantly simplified the example I just pushed. However, the colours of the optodes have changed (brain too, but I dont mind that). Is this intentional? If so, we should update the tutorial text. But I think the colours for the new version are too similar between sources and detectors (both look red to me).
> 
> ### Old colors
> image
> 
> ### New colors
> image
> 
> I also noticed that distance is now in mm? But was previously in m? Was this change also intentional?

Brain has a keyword argument units that can be set to either mm or m so that can go either way.

The colors were just my rough matplotlib "gray"/"red"/"brown" but should really be copied directly, I can update that.

drammock (commented):
@alexrockhill your comment on this PR says it "does some nice refactorization" but as far as I can tell this is still a completely separate implementation from what is done in `plot_alignment` (with no refactored-out shared helper functions).  Am I missing something?

alexrockhill:
Ok I refactored quite a bit to clean everything up and make nice `_freesurfer.py` functions.

> @alexrockhill your comment on this PR says it "does some nice refactorization" but as far as I can tell this is still a completely separate implementation from what is done in `plot_alignment` (with no refactored-out shared helper functions). Am I missing something?

Hmmm, yes I haven't gotten to the refactoring `plot_alignment`, that is quite an undertaking. I was just talking about refactoring from the previous commit so that the functions could be used more easily in `plot_alignment` in the future.

mne/_freesurfer.py
```diff
@@ -422,6 +422,61 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+        type_slice = slice(3, 6)
+    if detector:
+        type_slice = slice(6, 9)
+    ch_coord = info['chs'][idx]['loc'][type_slice]
+    # coordinate frame/location info
+    coord_frame = info['chs'][idx]['coord_frame']
+    if coord_frame == FIFF.FIFFV_COORD_UNKNOWN:
+        if verbose is None or verbose or verbose == 'warn':
+            warn('Got coordinate frame "unknown", assuming '
+                 '"head" coordinates')
```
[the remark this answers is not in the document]
alexrockhill: Yeah I didn't know how that worked, my bad, fixing it

drammock:
OK...  can you remove `MRG` from the PR title then?  Seems like I needn't have reviewed just yet.

mne/_freesurfer.py
```diff
@@ -422,6 +422,61 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+        ch_coord = apply_trans(info['dev_head_t']['trans'], ch_coord)
+        coord_frame = FIFF.FIFFV_COORD_HEAD  # now it's in head
+    if coord_frame == FIFF.FIFFV_COORD_HEAD:
+        trans = invert_transform(get_mri_head_trans(subject, subjects_dir))
+        ch_coord = apply_trans(trans['trans'], ch_coord) * 1000  # mm -> m
+        coord_frame = FIFF.FIFFV_COORD_MRI  # now in surface RAS
+    if coord_frame != FIFF.FIFFV_COORD_MRI:
+        raise RuntimeError(
+            '`inst` coordinate frame must be "meg", "head" or "mri", '
+            f'got {coord_frame} for {info.ch_names[idx]}')
```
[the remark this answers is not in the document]
alexrockhill: 1. Very unlikely, it should never happen unless the user modifies the coordinate frame manually with `info['chs'][idx]['coord_frame']`.
2. Sure changing to that.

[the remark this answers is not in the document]
alexrockhill: I guess maybe that's too strong of a thing to say, to my knowledge other coordinate frames are not regularly set in the info, it should really always be "head" but I don't know for sure.

alexrockhill:
> OK... can you remove `MRG` from the PR title then? Seems like I needn't have reviewed just yet.

I'm giving it a shot modifying `plot_alignment` now, maybe it won't be so bad. Also this will involve fixing the `set_montage` not setting the coordinate frame to "head" (https://github.com/mne-tools/mne-python/issues/9627) so that might cause widespread changes in examples + tutorials and maybe even sample data (mne sample is in the proper coordinate frame but the misc seeg data is "unknown").

alexrockhill:
Fixes #9635 now and will fix #9627 in a minute.

I gave it a real try to do larger refactoring but the lack of a requirement of the `subjects_dir` and `subject` and the support for "head" and "meg" coordinate frames makes there really not all that much that can be done with the same helper function, unfortunately. I think they are just going to have to rely on their own internal functions for coordinate frames for the most part. Happy to chat about this.

mne/_freesurfer.py
```diff
@@ -422,6 +425,122 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
     return fids
 
 
+@verbose
+def compute_mri_head_t(subject, subjects_dir, verbose=None):
+    """Get the head to surface RAS transform using the Freesurfer recon.
```
rob-luke, line 430:
```suggestion
    """Get the head to surface RAS transform using the Freesurfer reconstruction.
```

alexrockhill: It's too long unfortunately, that's the one-line summary

rob-luke: Then please define the abbreviation below. I was unsure if you meant reconstruction or recon, as in a first exploratory observation.

alexrockhill:
Ok @drammock, it works now and all the tests should pass. Everything is refactored and actually the differences were reconcilable with a bit more effort. You're welcome to review or I'm happy to wait until @larsoner gets back. I just wanted to finish this so that I knew what the `_freesurfer.py` functions would end up looking like for the GUI so that it wasn't a difficult merge.

tutorials/io/30_reading_fnirs_data.py
```diff
@@ -241,15 +242,12 @@
...
-    raw.info, trans=trans, subject='fsaverage', subjects_dir=subjects_dir,
-    surfaces=['brain', 'head'], coord_frame='mri', dig=True, show_axes=True,
-    fnirs=['channels', 'pairs', 'sources', 'detectors'], fig=fig)
-mne.viz.set_3d_view(figure=fig, azimuth=90, elevation=90, distance=0.5,
-                    focalpoint=(0., -0.01, 0.02))
+brain = mne.viz.Brain('fsaverage', subjects_dir=subjects_dir,
+                      alpha=0.5, cortex='low_contrast')
+brain.add_head()
+brain.add_sensors(raw)
+brain.enable_depth_peeling()
```
rob-luke, line 252: I understand this is from another PR, but I am just curious. Why is depth peeling required here, but not in the other example? I looked at the doc string and it doesn't explain what depth peeling is or in what situations you should use it. So I looked up wikipedia for an explanation. But I am still missing an understanding of in which situations you should enable this option.

alexrockhill: It has to do with the rendering and you should probably ask Eric to get a better explanation of technically what is going on, but my understanding and from working through examples what I've seen is that if there are many transparent surfaces overlaid, you really want depth peeling but, if not, it can cause some funky effects like the MEG sensors in this PR.

rob-luke: Thanks Alex, you've been a real gem teaching me along side these PRs.

mne/_freesurfer.py
drammock: FYI, it's a holdover from the freesurfer command called `recon-all` I believe.  But I agree it would be nice to spell out "reconstruction" later in the docstring if there's a place where it makes sense to do so.

alexrockhill:
I fixed the warnings that caused the tests to fail that I should have thought to check for, just waiting for https://github.com/mne-tools/mne-misc-data/pull/14 to merge so that the SEEG tutorial works. Other than that, I think a good review would be helpful but it looks okay to me. I put some of the changes in https://github.com/mne-tools/mne-python/pull/9586 that had more to do with the SEEG tutorial and less to do with adding sensors to a plot. I think the API looks nice and everything seems to work.

alexrockhill:
Alright! Struggled with the tests a lot on this one but I think ready for review. I thought it was a lot closer than it was, no hurry just working on the GUI now.

mne/_freesurfer.py
```diff
@@ -632,3 +748,43 @@ def _get_head_surface(surf, subject, subjects_dir, bem=None, verbose=None):
                 return _read_mri_surface(fname)
     raise IOError('No head surface found for subject '
                   f'{subject} after trying:\n' + '\n'.join(try_fnames))
+
+
+@verbose
+def _get_skull_surface(surf, subject, subjects_dir, bem=None, verbose=None):
```
agramfort, line 754: I suspect this is redundant with some BEM functions that will read all surfaces. You can just pick the skull from it.

mne/_freesurfer.py
```diff
@@ -422,6 +425,119 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+                ch_coord_frame = FIFF.FIFFV_COORD_HEAD
+            ch_coord = apply_trans(
+                to_cf_t[_frame_to_str[ch_coord_frame]], ch_coord)
+            if ch_type in _MEG_CH_TYPES_SPLIT + ('ref_meg',):
+                chs[type_name][info.ch_names[idx]] = (ch_coord, triangles)
+            else:
+                chs[type_name][info.ch_names[idx]] = ch_coord
+    if unknown_chs:
+        warn(f'Got coordinate frame "unknown" for {unknown_chs}, assuming '
+             '"head" coordinates')
```
agramfort, line 520:
```suggestion
             '"head" coordinates.')
```

mne/viz/_brain/_brain.py
```diff
@@ -2354,6 +2360,45 @@ def add_head(self, dense=True, color=None, alpha=0.5):
...
+    @fill_doc
+    def add_skull(self, outer=True, inner=False, color=None, alpha=0.5):
+        """Add a mesh to render the skull surface.
+
+        Parameters
+        ----------
+        outer : bool
+            Whether render the outer skull.
+        inner : bool
+            Whether to render the inner skull.
```
agramfort, line 2372: it feels only one parameter should be sufficient.

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2541,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+            if color is None:
+                color = DEFAULTS['coreg']['helmet_color']
+            alpha = _assign_by_name_or_ch_type(alphas, 'meg_helmet', 'meg')
+            alpha = 0.25 if alpha is None else alpha
+            self._renderer.mesh(*verts.T, triangles,
+                                color=colorConverter.to_rgba(color, alpha),
+                                opacity=alpha, reset_camera=False,
+                                render=False)
+
+        self._renderer._update()
```
agramfort, line 2693: this block of code above is not too redundant with the code in plot_alignment?

@GuillaumeFavelier what do you think?

agramfort (commented):
@GuillaumeFavelier please have a look at this refactoring. You are the one with the best vision of this part of our codebase. Thx !

mne/viz/_brain/_brain.py
alexrockhill: It's a bit redundant but the API is a bit different so I think that makes it difficult to entirely avoid the redundancy.

alexrockhill:
Is there a problem with Circle? I'll push an empty commit to see if it was just bad luck...

alexrockhill:
Looks like `apt-get` might have been/had some part depreciated today so I changed it to `apt` in circle dependencies, we'll see if that solves it.

mne/viz/_brain/_brain.py
```diff
@@ -2348,6 +2347,43 @@ def add_head(self, dense=True, color=None, alpha=0.5):
...
+    def add_skull(self, outer=True, color=None, alpha=0.5):
+        """Add a mesh to render the skull surface.
+
+        Parameters
+        ----------
+        outer : bool
+            Adds the outer skull if ``True``, otherwise adds the inner skull.
+        color : matplotlib-style color | None
+            A list of anything matplotlib accepts: string, RGB, hex, etc.
+            (default: "gray").
```
agramfort, line 2373: can you just put "gray" in the function signature then?

alexrockhill:
Okay? Want to refactor more? This has the circle CI fix in it as well so might be nice to merge or put that in a separate PR

drammock (commented):
this is a partial review I started last week and now I'm swamped, so take it with a grain of salt, it may be based on an outdated diff

mne/_freesurfer.py
```diff
@@ -422,6 +425,119 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+        for type_name, type_slice in type_slices.items():
+            if ch_type in _MEG_CH_TYPES_SPLIT + ('ref_meg',):
+                coil_trans = _loc_to_coil_trans(info['chs'][idx]['loc'])
+                coil = _create_meg_coils([info['chs'][idx]], acc='normal')[0]
+                # store verts as ch_coord
+                ch_coord, triangles = _sensor_shape(coil)
+                ch_coord = apply_trans(coil_trans, ch_coord)
+                if len(ch_coord) == 0 and warn_meg:
+                    warn(f'MEG sensor {info.ch_names[idx]} not found. '
+                         'Cannot plot MEG location.')
```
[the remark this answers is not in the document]
alexrockhill: It returns coordinates that will be plotted, there's really not a good way to put that back in `mne.viz._3d.py`... not sure how to improve that

drammock:
> Okay? Want to refactor more? This has the circle CI fix in it as well so might be nice to merge or put that in a separate PR

I already pulled the CircleCI fix out into another PR which just merged, so no rush on this one for that reason.

mne/_freesurfer.py
```diff
@@ -422,6 +425,120 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
     return fids
 
 
+@verbose
+def estimate_head_mri_t(subject, subjects_dir=None, verbose=None):
+    """Estimate the head->mri transform.
```
agramfort, line 430:
```suggestion
    """Estimate the head->mri transform from fsaverage fiducials.
```

mne/viz/_brain/_brain.py
agramfort: I don't see why. If you have a private function that takes a renderer and the parameters it needs it should be possible to add sensor data to any rendered in the right coordinate system. This is a big block of complex code.

mne/viz/tests/test_3d.py
```diff
@@ -576,7 +593,8 @@ def test_snapshot_brain_montage(renderer):
     """Test snapshot brain montage."""
     info = read_info(evoked_fname)
     fig = plot_alignment(
-        info, trans=None, subject='sample', subjects_dir=subjects_dir)
+        info, trans=Transform('head', 'mri'), subject='sample',
```
agramfort, line 596: we are deprecating the trans=None parameter? if so I don't think the what's new reflects this

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -320,17 +321,22 @@ def plot_overlay(image, compare, title, thresh=None):
...
+# (this step uses ``template_trans`` but it is recomputed behind the scenes)
 raw.set_montage(montage_warped)
 
 # plot the resulting alignment
-renderer = mne.viz.backends.renderer.create_3d_figure(**fig_kwargs)
-fig = mne.viz.plot_alignment(raw.info, fsaverage_trans, 'fsaverage',
-                             fig=renderer.figure, **al_kwargs)
-mne.viz.set_3d_view(fig, **view_kwargs)
+brain = mne.viz.Brain('fsaverage', **brain_kwargs)
+brain.add_sensors(raw, trans=template_trans)
```
agramfort, line 338:
```suggestion
brain.add_sensors(raw.info, trans=template_trans)
```
seems more natural

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2536,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+    @verbose
+    def add_sensors(self, inst, trans, picks=None, colors=None,
+                    scales=None, alphas=None, meg_helmet=True,
+                    project_eeg=False, fnirs_pairs=None, fnirs_sources=None,
+                    fnirs_detectors=None, verbose=None):
+        """Add mesh objects to represent sensor positions.
+
+        Parameters
+        ----------
+        inst : mne.Info | mne.io.Raw | mne.Epochs | mne.Evoked
```
agramfort, line 2548: I would suggest you only accept an Info.

mne/viz/_3d.py
```diff
@@ -2602,22 +2531,22 @@ def plot_sparse_source_estimates(src, stcs, colors=None, linewidth=2,
...
+        ('loosely dashed',        (0, (5, 10))),                # noqa: E241
+        ('dashed',                (0, (5, 5))),                 # noqa: E241
+        ('densely dashed',        (0, (5, 1))),                 # noqa: E241
+        ('loosely dashdotted',    (0, (3, 10, 1, 10))),         # noqa: E241
+        ('dashdotted',            (0, (3, 5, 1, 5))),           # noqa: E241
+        ('densely dashdotted',    (0, (3, 1, 1, 1))),           # noqa: E241
+        ('dashdotdotted',         (0, (3, 5, 1, 5, 1, 5))),     # noqa: E241
+        ('loosely dashdotdotted', (0, (3, 10, 1, 10, 1, 10))),  # noqa: E241
+        ('densely dashdotdotted', (0, (3, 1, 1, 1, 1, 1)))      # noqa: E241
     ]
```
agramfort, line 2550: I find the diff in this file super hard to read. Can you explain what you did? thx

doc/mri.rst
```diff
@@ -15,6 +15,7 @@ Step by step instructions for using :func:`gui.coregistration`:
    :toctree: generated/
 
    coreg.get_mni_fiducials
+   coreg.estimate_head_mri_t
```
agramfort, line 18: this new function is not documented in what's new

mne/_freesurfer.py
```diff
@@ -422,6 +425,120 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+        head=_ensure_trans([dev_head_t, head_mri_t, Transform('head', 'head')],
+                           fro='head', to=coord_frame),
+        mri=_ensure_trans([head_mri_t, mri_dev_t, Transform('mri', 'mri')],
+                          fro='mri', to=coord_frame))
+    return to_cf_t
+
+
+@verbose
+def _ch_pos_in_coord_frame(info, to_cf_t, warn_meg=True, verbose=None):
+    """Transform a channel location to "mri" (surface RAS)."""
```
agramfort, line 474: this function is made from code that was in plot_alignment and now plot_alignement uses this function? it's the refactoring?

mne/viz/tests/test_3d.py
alexrockhill: This was what @larsoner wanted, will update the `whatsnew`

alexrockhill: It just warns as well so I'm not sure if that counts as a depreciation, I think it should have warned before, an identity head->mri transform is rather odd to assume based on our conversation here https://github.com/mne-tools/mne-misc-data/pull/14

tutorials/clinical/10_ieeg_localize.py
alexrockhill: Sure, it will support data viewing in the future as I understand but good for now

mne/viz/_brain/_brain.py
alexrockhill: This is @larsoner's request for forward compatibility when data may be added as well as channel positions

mne/viz/_3d.py
alexrockhill: I think I modified the file in some way that I got a flake error when it was recomputed without ignoring this part so I just did this to satisfy the CIs

mne/_freesurfer.py
alexrockhill: Yes this is the main refactoring, and yes it's roughly based on the code in `plot_alignment` but is a bit more elegant

mne/surface.py
```diff
@@ -1736,27 +1736,11 @@ def _marching_cubes(image, level, smooth=0):
...
-                           'segmentation and reconstructed surfaces are')
-    return base_image, fs_t1
-
-
 def _warn_missing_chs(montage, dig_image, after_warp):
     """Warn that channels are missing."""
     # ensure that each electrode contact was marked in at least one voxel
     missing = set(np.arange(1, len(montage.ch_names) + 1)).difference(
-        set(np.unique(_get_img_fdata(dig_image))))
+        set(np.unique(np.array(dig_image.dataobj))))
```
larsoner, line 1743: `np.asanyarray` is preferred as it won't necessarily make a copy (unless you are going to modify values, in which case `np.array` is fine -- but here you're not), and it's also nibabel's recommendation:

https://nipy.org/nibabel/reference/nibabel.dataobj_images.html#nibabel.dataobj_images.DataobjImage.get_data

mne/surface.py
```diff
@@ -1840,28 +1824,40 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
...
     # convert to freesurfer voxel space
     ch_coords = apply_trans(
-        np.linalg.inv(fs_t1.header.get_vox2ras_tkr()), ch_coords * 1000)
+        np.linalg.inv(fs_from_img.header.get_vox2ras_tkr()), ch_coords * 1000)
 
     # take channel coordinates and use the image to transform them
     # into a volume where all the voxels over a threshold nearby
     # are labeled with an index
-    image_data = _get_img_fdata(image)
+    image_data = np.array(base_image.dataobj)
```
larsoner, line 1856: `np.array` is okay here as you want a copy anyway. But keep in mind that the values can now in general be non-float (and have over/underflow problems like `np.uint8`), unlike `_get_img_fdata`. So I actually think `_get_img_fdata` might be a safer choice here. The existing code will probably break quite badly if the dtype is `np.uint8` and there are more than 255 electrodes...

mne/surface.py
```diff
@@ -1893,25 +1889,20 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
     _warn_missing_chs(montage, image_to, after_warp=True)
 
     # recover the contact positions as the center of mass
-    warped_data = _get_img_fdata(image_to)
+    warped_data = np.array(image_to.dataobj)
```
larsoner, line 1892: No copy needed and native dtype is fine so `np.asanyarray` is probably better here

mne/tests/test_surface.py
```diff
@@ -314,50 +312,57 @@ def test_warp_montage_volume():
...
+        assert ch_coords_vox.round()[idx] + 5 not in voxels
     # check image_to, too many, just check center
     ground_truth_warped_voxels = np.array(
         [[135.5959596, 161.97979798, 123.83838384],
          [143.11111111, 159.71428571, 125.61904762],
          [150.53982301, 158.38053097, 127.31858407]])
     for i in range(len(montage.ch_names)):
         assert np.linalg.norm(
-            np.array(np.where(_get_img_fdata(image_to) == i + 1)
+            np.array(np.where(np.array(image_to.dataobj) == i + 1)
```
larsoner, line 365: `asanyarray` should be better here

mne/viz/_brain/_brain.py
```diff
@@ -2342,6 +2346,42 @@ def add_head(self, dense=True, color=None, alpha=0.5):
         surf = _get_head_surface('seghead' if dense else 'head',
                                  self._subject_id, self._subjects_dir)
         verts, triangles = surf['rr'], surf['tris']
+        verts *= 1e3 if self._units == 'mm' else 1
+        if color is None:
+            color = DEFAULTS['coreg']['head_color']
```
larsoner, line 2351: This was added in this cycle so I would not have this, I'd remove support for `None` entirely

mne/viz/_brain/_brain.py
```diff
@@ -2328,7 +2333,6 @@ def add_head(self, dense=True, color=None, alpha=0.5):
             (``head``).
         color : matplotlib-style color | None
```
larsoner, line 2334: We document everywhere else as just `color`, and no need for `None` here
```suggestion
        color : color
```

mne/viz/_brain/_brain.py
larsoner: Let's start with just `info` and add support for the others later, when we actually want to plot data values

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2536,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+            EEG sensors are present in ``info``).
+        fnirs_pairs : str | bool
+            The color to plot fnirs pairs. If False, fnirs pairs will not
+            be plotted. If None, the default color will be used.
+        fnirs_sources : str | bool
+            The color to plot fnirs sources. If False, fnirs sources will not
+            be plotted. If None, the default color will be used.
+        fnirs_detectors : str | bool
+            The color to plot the fnirs detectors. If False, fnirs detectors
+            will not be plotted. If None, the default color will be used.
```
larsoner, line 2579: As mentioned somewhere previously, -1 on this, I'd rather keep the API of `plot_alignment` and `add_sensors` the same regarding how to plot them. We had discussions in the past of the right/best API (e.g., during the addition of fNIRS stuff) and settled on that...

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2536,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        .. versionadded:: 0.24
+        """
+        from matplotlib.colors import colorConverter
+        if isinstance(inst, Info):
+            info = inst.copy()
+        else:
+            info = inst.info.copy()
+        picks = _picks_to_idx(info, picks, none='all', exclude=(),
+                              allow_empty=False)
+        info = info.pick_channels([inst.ch_names[idx] for idx in picks])
```
larsoner, line 2593: This is simpler, and your existing code has a problematic bug in that it will modify `info` in place
```suggestion
        _validate_type(info, Info, 'info')
        picks = _picks_to_idx(info, picks, none='all', exclude=(),
                              allow_empty=False)
        info = pick_info(info, picks)  # makes a copy
```

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2536,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        if trans is None:
+            warn('No `trans` provided, assuming identity head->mri transform')
+            trans = Transform('head', 'mri')
+        head_mri_t = _get_trans(trans, 'head', 'mri')[0]
```
larsoner, line 2597: Let's not support `None` here at all. In `plot_alignment` we can be more flexible, but it's almost certainly a bug if someone has a None trans for `Brain.add_sensors` because sensors are (pretty much always) in head and Brain is always in MRI. Let's make them construct an identity one if they really want it:
```suggestion
        head_mri_t = _get_trans(trans, 'head', 'mri', allow_none=False)[0]
        del trans
```
```

File: mne/viz/_brain/_brain.py [hunk 30]
```diff
@@ -2496,6 +2536,157 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        to_cf_t = _get_transforms_to_coord_frame(
+            info, head_mri_t, coord_frame='mri', verbose=verbose)
+        try:
+            ch_pos, sources, detectors = _ch_pos_in_coord_frame(
+                info, to_cf_t, verbose=verbose)
+        except RuntimeError as err:
+            raise RuntimeError(
+                f'{err} You may want to use `picks=("meg", "eeg")` '
+                'to avoid sensors that should not be plotted for instance')
+        eeg_indices = pick_types(info, eeg=True)
```
larsoner, line 2608: Don't exclude bads from the subset already picked
```suggestion
        eeg_indices = pick_types(info, eeg=True, exclude=())
```

mne/viz/_brain/_brain.py
larsoner: I agree with @agramfort, to me these should share the same code structure. The only difference really is that here the sensors are always plotted in the MRI coordinate frame, so this would just use a subset of some private function like:
```
def _plot_sensors(renderers, info, trans, meg, eeg, fnirs, coord_frame='mri', units='m'):
```
function or so. The `units` in `plot_alignment` are always `'m'`, and the `coord_frame` in `Brain.add_sensors` will always be `'mri'`, but otherwise both can just call this function with the correct arguments (one renderer for `plot_alignment` and potentially multiple for `Brain`) and get identical visualizations (and DRY code) that way...

mne/viz/_brain/tests/test_brain.py
```diff
@@ -315,20 +315,30 @@ def __init__(self):
...
+    # add head and skull
     brain.add_head(color='red', alpha=0.1)
+    brain.add_skull(outer=True, color='green', alpha=0.1)
 
     # add volume labels
     brain.add_volume_labels(
         aseg='aseg', labels=('Brain-Stem', 'Left-Hippocampus',
                              'Left-Amygdala'))
+    # add sensors
+    raw = read_raw_fif(fname_raw_testing)
```
larsoner, line 327:
```suggestion
    info = read_info(fname_raw_testing)
```

mne/viz/tests/test_3d.py
```diff
@@ -206,7 +209,7 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
                   subject='fsaverage', subjects_dir=subjects_dir,
                   src=sample_src)
     sample_src.plot(subjects_dir=subjects_dir, head=True, skull=True,
-                    brain='white')
+                    brain='white', trans=trans_fname)
```
larsoner, line 212: This should not be necessary and should not raise an error. You should only need `trans` if it's needed to get from the original coordinate frame (here, MRI, as that's where source spaces live) to the destination plotting frame (here, still MRI).

mne/viz/tests/test_3d.py
```diff
@@ -217,8 +220,17 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
     renderer.backend._close_all()
     # no-head version
     renderer.backend._close_all()
+    # trans required
+    with pytest.raises(ValueError, match='transformation matrix is required'):
+        plot_alignment(info, trans=None, src=src_fname)
```
larsoner, line 225: This should not raise an error, as the default for `plot_alignment` is `coord_frame='head'` and it should be possible to plot all channel positions in the head coordinate frame without requiring a head<->MRI transformation

mne/viz/tests/test_3d.py
```diff
@@ -217,8 +220,17 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
...
+    with pytest.raises(ValueError, match='transformation matrix is required'):
+        plot_alignment(info, trans=None, coord_frame='mri')
+    with pytest.raises(ValueError, match='transformation matrix is required'):
+        plot_alignment(info, trans=None, mri_fiducials=True)
+    with pytest.raises(ValueError, match='transformation matrix is required'):
+        plot_alignment(info, trans=None, surfaces=['brain'])
     # all coord frames
-    plot_alignment(info)  # works: surfaces='auto' default
+    plot_alignment(info, trans=trans_fname)  # works: surfaces='auto' default
```
larsoner, line 233: Should not be required, revert

mne/viz/tests/test_3d.py
larsoner: Probably better to ensure it actually warns by using `trans=None` and wrapping in a
```
with pytest.warns(..., match='...'):
```
But as above, it should (in all cases/tests) only emit a warning if a head<->MRI transform is actually needed. Here it is, so the `pytest.warns` is useful to make sure it does the right thing.

mne/viz/tests/test_3d.py
alexrockhill: This is to suppress the warning...

larsoner: See my other comments, plotting sensors in `head` coordinate frame (the default for `plot_alignment`) `trans` should not be required, so having it be `None` is not a problem/no need to warn. I'm asking you to be more selective about when the warning is emitted...

mne/viz/tests/test_3d.py
alexrockhill: The source space is in "mri"...

alexrockhill:
Actually, I realized I changed it to a `RuntimeError` not a warning if `trans` is required

alexrockhill:
Ok so I tried both a `_plot_sensors` and even a `_plot_sensor` approach to refactoring the shared code that you both @agramfort and @larsoner pointed out is redundant. The issue is that the refactored code is very inelegant and loses support for the enhanced coloring scheme that I had added to `add_sensors`. I agree that the function calls are almost identical but the logic is done a different way that makes it very inelegant when combined. In `mne.viz.Brain`, which is the way I much prefer, things are plotted with boolean flags in smaller steps that put all the pieces together elegantly. In `plot_alignment`, everything is passed in complicated lists of options. When trying to combine these two styles in one API, the complexity multiplies and it's just a mess.

larsoner:
>  In mne.viz.Brain, which is the way I much prefer, things are plotted with boolean flags in smaller steps that put all the pieces together elegantly. In plot_alignment, everything is passed in complicated lists of options. When trying to combine these two styles in one API, the complexity multiplies and it's just a mess

I am still -1 on a divergent API. I'd rather stick to and expand the plot_alignment API somehow rather than introduce a new interface, as above: https://github.com/mne-tools/mne-python/pull/9585#discussion_r690660363

I'd rather have slightly less clean code in a shared private function than these divergent APIs with non-DRY code

alexrockhill:
Okay, now the API is the same... there go my nice color plotting options though, guess it's we're just stuck with the defaults

larsoner:
>  there go my nice color plotting options though, guess it's we're just stuck with the defaults

For this PR, sure. I think if we want to add color options (which I agree is a good idea and worth doing!) we should figure out how to do it in `plot_alignment` and `Brain.plot_sensors` jointly in a follow-up PR

doc/changes/latest.inc
```diff
@@ -133,7 +133,9 @@ Enhancements
 
 - Add :func:`mne.channels.read_dig_localite` to read Localite electrode location files (:gh:`9658` by `Clemens Brunner`_)
 
-- Add :meth:`mne.viz.Brain.add_sensors` to plot sensor locations (:gh:`9580` by `Alex Rockhill`_)
+- Add :meth:`mne.viz.Brain.add_sensors` to plot sensor locations (:gh:`9585` by `Alex Rockhill`_)
+
+- Add :func:`mne.coreg.estimate_head_mri_t` estimate the head->mri transform from fsaverage fiducials (:gh:`9585` by `Alex Rockhill`_)
```
agramfort, line 138:
```suggestion
- Add :func:`mne.coreg.estimate_head_mri_t` to estimate the head->mri transform from fsaverage fiducials (:gh:`9585` by `Alex Rockhill`_)
```

mne/viz/_3d.py
```diff
@@ -1036,20 +880,124 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
...
+    for xi, x in enumerate(eeg):
+        _check_option(f'eeg[{xi}]', x, ('original', 'projected'))
+    for xi, x in enumerate(fnirs):
+        _check_option(f'fnirs[{xi}]', x, ('channels', 'pairs',
+                                          'sources', 'detectors'))
+    return meg, eeg, fnirs, warn_meg
+
+
+def _plot_sensors(info, to_cf_t, renderer, picks, meg, eeg, fnirs,
+                  warn_meg, head_surf, units, verbose):
```
agramfort, line 933: cc @GuillaumeFavelier this is doing what we discussed yesterday

mne/viz/_3d.py
agramfort: my comment here was concerning the diff of the entire file. What was the big picture logic? add such a function like _plot_sensors that takes a renderer param so Brain and plot_alignment can use the same code to display sensors?

mne/viz/_brain/_brain.py
agramfort: +1

alexrockhill:
> my comment here was concerning the diff of the entire file. What was the big picture logic? add such a function like _plot_sensors that takes a renderer param so Brain and plot_alignment can use the same code to display sensors?

I just wanted to add `add_sensors` to `Brain` but @larsoner said I had to refactor `plot_alignment` to do that ;)

mne/_freesurfer.py
```diff
@@ -632,3 +749,43 @@ def _get_head_surface(surf, subject, subjects_dir, bem=None, verbose=None):
...
+            return _bem_find_surface(bem, surf + '_skull')
+        except RuntimeError:
+            logger.info('Could not find the surface for '
+                        'skull in the provided BEM model, '
+                        'looking in the subject directory.')
+    fname = op.join(get_subjects_dir(subjects_dir, raise_error=True),
+                    subject, 'bem', surf + '_skull.surf')
+    if not op.isfile(fname):
+        raise ValueError('Skull surface cannot be found, should be '
+                         f'located at {fname}')
```
larsoner, line 790: Should use `fname = _check_fname(fname, overwrite='read', must_exist=True, name=f'{surf} skull surface')`

mne/viz/_3d.py
```diff
@@ -1036,20 +884,124 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
...
+    """Render sensors in a 3D scene."""
+    defaults = DEFAULTS['coreg']
+    ch_pos, sources, detectors = _ch_pos_in_coord_frame(
+        info.pick_channels([info.ch_names[idx] for idx in picks]),
```
larsoner, line 941: I feel like I made this comment already... but in any case, `info` needs to be copied here because in `add_sensors` it's not copied and this operates in place. Also `pick_info` is more succinct here (and does a copy=True by default):
```suggestion
        pick_info(info, picks),
```

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2532,81 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+            Can be "channels", "pairs", "detectors", and/or "sources" to show
+            the fNIRS channel locations, optode locations, or line between
+            source-detector pairs, or a combination like
+            ``('pairs', 'channels')``. True translates to ``('pairs',)``.
+        ecog : bool
+            If True (default), show ECoG sensors.
+        seeg : bool
+            If True (default), show sEEG electrodes.
+        dbs : bool
+            If True (default), show DBS (deep brain stimulation) electrodes.
```
larsoner, line 2571: These should be `(whatever)s`'ed in the `docdict` to avoid duplication

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2532,81 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        meg, eeg, fnirs, warn_meg = _handle_sensor_types(meg, eeg, fnirs)
+        picks = pick_types(info, meg=('sensors' in meg),
+                           ref_meg=('ref' in meg), eeg=(len(eeg) > 0),
+                           ecog=ecog, seeg=seeg, dbs=dbs,
+                           fnirs=(len(fnirs) > 0))
+        head_mri_t = _get_trans(trans, 'head', 'mri', allow_none=False)[0]
+        del trans
+        # get transforms to "mri"window
+        to_cf_t = _get_transforms_to_coord_frame(
+            info, head_mri_t, coord_frame='mri', verbose=verbose)
```
larsoner, line 2588: no need to pass `verbose` here

mne/_freesurfer.py
```diff
@@ -422,6 +425,120 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+    subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
+    lpa, nasion, rpa = get_mni_fiducials(subject, subjects_dir)
+    montage = make_dig_montage(lpa=lpa['r'], nasion=nasion['r'], rpa=rpa['r'],
+                               coord_frame='mri')
+    return invert_transform(compute_native_head_t(montage))
+
+
+@verbose
+def _get_transforms_to_coord_frame(info, trans, coord_frame='mri',
+                                   verbose=None):
```
larsoner, line 456: No need to accept `verbose` here

mne/_freesurfer.py
```diff
@@ -422,6 +425,120 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+    """
+    from .channels.montage import make_dig_montage, compute_native_head_t
+    subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
+    lpa, nasion, rpa = get_mni_fiducials(subject, subjects_dir)
+    montage = make_dig_montage(lpa=lpa['r'], nasion=nasion['r'], rpa=rpa['r'],
+                               coord_frame='mri')
+    return invert_transform(compute_native_head_t(montage))
+
+
+@verbose
```
larsoner, line 454: No need for decorator

mne/viz/_3d.py
```diff
@@ -1036,20 +884,124 @@ def plot_alignment(info=None, trans=None, subject=None, subjects_dir=None,
...
+        to_cf_t=to_cf_t, warn_meg=warn_meg,
+        verbose=verbose)
```
larsoner, line 943: No need to pass `verbose` here

mne/viz/_3d.py
larsoner: No need to take `verbose` here

mne/viz/_brain/_brain.py
```diff
@@ -2496,6 +2532,81 @@ def add_foci(self, coords, coords_as_verts=False, map_surface=None,
...
+        if pick_types(info, eeg=True, exclude=()).size > 0 and \
+                'projected' in eeg:
+            head_surf = _get_head_surface(
+                'seghead', self._subject_id, self._subjects_dir)
+        else:
+            head_surf = None
+        # Do the main plotting
+        if picks.size > 0:
+            _plot_sensors(info, to_cf_t, self._renderer, picks, meg, eeg,
+                          fnirs, head_surf, warn_meg, self._units, verbose)
```
larsoner, line 2598: No need to pass `verbose` here

mne/viz/tests/test_3d.py
larsoner: Ahh yes I missed that!

mne/viz/tests/test_3d.py
```diff
@@ -230,32 +240,40 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
...
                                         'sample-1280-1280-1280-bem-sol.fif'))
     bem_surfs = read_bem_surfaces(op.join(subjects_dir, 'sample', 'bem',
                                           'sample-1280-1280-1280-bem.fif'))
     sample_src[0]['coord_frame'] = 4  # hack for coverage
-    plot_alignment(info, subject='sample', eeg='projected',
-                   meg='helmet', bem=sphere, dig=True,
+    plot_alignment(info, trans_fname, subject='sample',
+                   eeg='projected', meg='helmet', bem=sphere, dig=True,
                    surfaces=['brain', 'inner_skull', 'outer_skull',
                              'outer_skin'])
```
larsoner, line 264: `trans` should not be required here, and shouldn't actually be used for that matter. A spherical BEM is defined in head coordinates actually... thus, as below, nothing here is in MRI coords.

mne/viz/tests/test_3d.py
```diff
@@ -230,32 +240,40 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
...
     plot_alignment(info, trans_fname, subject='sample', meg='helmet',
                    subjects_dir=subjects_dir, eeg='projected', bem=sphere,
                    surfaces=['head', 'brain'], src=sample_src)
+    # no trans okay, no mri surfaces
+    plot_alignment(info, bem=sphere, surfaces=['brain'])
```
larsoner, line 269: Good that this works

larsoner (commented):
Looks very close!

mne/viz/tests/test_3d.py
alexrockhill: In this case the `trans` is required because the Freesurfer surfaces are in "mri" and we're plotting in "head"

alexrockhill: i.e. "brain", "inner_skull" etc.

larsoner: Those should not come from freesurfer but rather from the "bem" i.e. the sphere conductor model

alexrockhill: Ah okay, fixed. Yeah the way it was done previously, it was a bit unclear what was happening.

mne/_freesurfer.py
```diff
@@ -422,6 +425,118 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+                          fro='meg', to=coord_frame),
+        head=_ensure_trans([dev_head_t, head_mri_t, Transform('head', 'head')],
+                           fro='head', to=coord_frame),
+        mri=_ensure_trans([head_mri_t, mri_dev_t, Transform('mri', 'mri')],
+                          fro='mri', to=coord_frame))
+    return to_cf_t
+
+
+@verbose
+def _ch_pos_in_coord_frame(info, to_cf_t, warn_meg=True, verbose=None):
```
larsoner, line 471: This is probably not the right place for this function. `mne/channels` or, since this really has to do with plotting more than anything else, `mne/viz/_3d.py` seems like a better place for it.

mne/_freesurfer.py
```diff
@@ -422,6 +425,118 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
...
+    """
+    from .channels.montage import make_dig_montage, compute_native_head_t
+    subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
+    lpa, nasion, rpa = get_mni_fiducials(subject, subjects_dir)
+    montage = make_dig_montage(lpa=lpa['r'], nasion=nasion['r'], rpa=rpa['r'],
+                               coord_frame='mri')
+    return invert_transform(compute_native_head_t(montage))
+
+
+def _get_transforms_to_coord_frame(info, trans, coord_frame='mri'):
```
larsoner, line 454: `mne/transforms.py` is probably a better place for this, or `mne/channels` somewhere

mne/viz/tests/test_3d.py
```diff
@@ -303,12 +323,13 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
     info['dig'] = None
     info_cube['chs'][0]['coil_type'] = 9999
     with pytest.raises(RuntimeError, match='coil definition not found'):
-        plot_alignment(info_cube, meg='sensors', surfaces=())
+        _create_meg_coils([info_cube['chs'][0]], acc='normal')
```
larsoner, line 326: Why this change from public to private call? Sticking to public interfaces as much as possible in tests is better

mne/viz/tests/test_3d.py
```diff
@@ -303,12 +323,13 @@ def test_plot_alignment(tmpdir, renderer, mixed_fwd_cov_evoked):
...
     coil_def_fname = op.join(tempdir, 'temp')
     with open(coil_def_fname, 'w') as fid:
         fid.write(coil_3d)
     with use_coil_def(coil_def_fname):
-        plot_alignment(info_cube, meg='sensors', surfaces=(), dig=True)
+        plot_alignment(info_cube, trans=trans_fname, meg='sensors',
```
larsoner, line 331: `trans` should not be required here, right? If so, revert

larsoner (commented):
Hopefully the last round...

larsoner:
Thanks @alexrockhill !

rob-luke:
Awesome

larsoner:
@alexrockhill this is wrong:

https://mne.tools/dev/auto_tutorials/forward/35_eeg_no_mri.html#sphx-glr-auto-tutorials-forward-35-eeg-no-mri-py

image

See how it looks on stable:

https://mne.tools/stable/auto_tutorials/forward/35_eeg_no_mri.html#sphx-glr-auto-tutorials-forward-35-eeg-no-mri-py

image

I did not check but I strongly suspect this PR broke the `projected` alignment / coordinate frame. Can you look into it?

PR #9586, opened 2021-07-19
[ENH, MRG] Add intracranial electrode localization GUI

### LINKED ISSUE #9559: [ENH] Add Intracranial Electrode Contact Picking GUI (PyQt)
The intracranial electrophysiology analysis pipeline could use a manual or semiautomated GUI to pick the locations of the electrode contacts based on a CT that has been aligned to the MR of that subject so that brain structures can be used to aid contact identification.

Basic Requirements:
- A CT and MR viewer that can scroll through slices like freeview
- A list selection to scroll through channel names from a ``raw`` object and associate them with a location on the CT in MR (surface RAS) space
- Ability to change the colormap to see color ranges within the CT more clearly as well as for the MR to make brain structures clearer
- An RAS locator to describe where you are in RAS space (helpful for use and this is also like freeview)

Advanced Requirements:
- A 3D viewer of the head with representations of "devices" or groups of contacts like an ECoG grid or sEEG electrode which more closely resembles surgical plans and so makes it easier to identify which device is which
- Speed of execution
- An automated algorithm to identify locations where electrode contacts are likely to be (likely using `voxel_neighbors`) to speed up location
- Contact coloring should be roughly orthogonal in ~24 dimensional space for sEEG patients with approximately that number of devices, also it looks nice if the device is colored the same

cc @larsoner

alexrockhill:
Add a GUI that allows users to find their intracranial electrode montage from a CT, an MRI and the Freesurfer recon-all from it, and an `info` object that supplies the names to be found and is return with a modified montage.

Closes #9559

Very rough version, no review yet.

### DIFF
--- a/mne/gui/__init__.py
+++ b/mne/gui/__init__.py
@@ -239,3 +239,44 @@ def kit2fiff():
     from ._kit2fiff_gui import Kit2FiffFrame
     frame = Kit2FiffFrame()
     return _initialize_gui(frame)
+
+
+@verbose
+def locate_ieeg(info, trans, aligned_ct, subject=None, subjects_dir=None,
+                groups=None, verbose=None):
+    """Locate intracranial electrode contacts.
+
+    Parameters
+    ----------
+    %(info_not_none)s
+    %(trans_not_none)s
+    aligned_ct : str | pathlib.Path | nibabel.spatialimages.SpatialImage
+        The CT image that has been aligned to the Freesurfer T1. Path-like
+        inputs and nibabel image objects are supported.
+    %(subject)s
+    %(subjects_dir)s
+    groups : dict | None
+        A dictionary with channels as keys and their group index as values.
+        If None, the groups will be inferred by the channel names. Channel
+        names must have a format like ``LAMY 7`` where a string prefix
+        like ``LAMY`` precedes a numeric index like ``7``. If the channels
+        are formatted improperly, group plotting will work incorrectly.
+        Group assignments can be adjusted in the GUI.
+    %(verbose)s
+
+    Returns
+    -------
+    gui : instance of IntracranialElectrodeLocator
+        The graphical user interface (GUI) window.
+    """
+    from ._ieeg_locate_gui import IntracranialElectrodeLocator
+    from PyQt5.QtWidgets import QApplication
+    # get application
+    app = QApplication.instance()
+    if app is None:
+        app = QApplication(["Intracranial Electrode Locator"])
+    gui = IntracranialElectrodeLocator(
+        info, trans, aligned_ct, subject=subject,
+        subjects_dir=subjects_dir, groups=groups, verbose=verbose)
+    gui.show()
+    return gui
--- a/mne/gui/_ieeg_locate_gui.py
+++ b/mne/gui/_ieeg_locate_gui.py
@@ -1,0 +1,864 @@
+# -*- coding: utf-8 -*-
+"""Intracranial elecrode localization GUI for finding contact locations."""
+
+# Authors: Alex Rockhill <aprockhill@mailbox.org>
+#
+# License: BSD (3-clause)
+
+import os.path as op
+import numpy as np
+from functools import partial
+
+from matplotlib.colors import LinearSegmentedColormap
+
+from PyQt5 import QtCore, QtGui, Qt
+from PyQt5.QtCore import pyqtSlot
+from PyQt5.QtWidgets import (QMainWindow, QGridLayout,
+                             QVBoxLayout, QHBoxLayout, QLabel,
+                             QMessageBox, QWidget,
+                             QListView, QSlider, QPushButton,
+                             QComboBox, QPlainTextEdit)
+from matplotlib.backends.backend_qt5agg import FigureCanvas
+from matplotlib.figure import Figure
+
+from .._freesurfer import _check_subject_dir, _import_nibabel
+from ..viz.backends.renderer import _get_renderer
+from ..surface import _read_mri_surface, _voxel_neighbors
+from ..transforms import (apply_trans, _frame_to_str, _get_trans,
+                          invert_transform)
+from ..utils import logger, _check_fname, _validate_type, verbose, warn
+
+_IMG_LABELS = [['I', 'P'], ['I', 'L'], ['P', 'L']]
+_CH_PLOT_SIZE = 1024
+_ZOOM_STEP_SIZE = 5
+
+# 20 colors generated to be evenly spaced in a cube, worked better than
+# matplotlib color cycle
+_UNIQUE_COLORS = [(0.1, 0.42, 0.43), (0.9, 0.34, 0.62), (0.47, 0.51, 0.3),
+                  (0.47, 0.55, 0.99), (0.79, 0.68, 0.06), (0.34, 0.74, 0.05),
+                  (0.58, 0.87, 0.13), (0.86, 0.98, 0.4), (0.92, 0.91, 0.66),
+                  (0.77, 0.38, 0.34), (0.9, 0.37, 0.1), (0.2, 0.62, 0.9),
+                  (0.22, 0.65, 0.64), (0.14, 0.94, 0.8), (0.34, 0.31, 0.68),
+                  (0.59, 0.28, 0.74), (0.46, 0.19, 0.94), (0.37, 0.93, 0.7),
+                  (0.56, 0.86, 0.55), (0.67, 0.69, 0.44)]
+_N_COLORS = len(_UNIQUE_COLORS)
+_CMAP = LinearSegmentedColormap.from_list(
+    'ch_colors', _UNIQUE_COLORS, N=_N_COLORS)
+
+
+def _load_image(img, name, verbose=True):
+    """Load data from a 3D image file (e.g. CT, MR)."""
+    nib = _import_nibabel('use iEEG GUI')
+    if not isinstance(img, nib.spatialimages.SpatialImage):
+        if verbose:
+            logger.info(f'Loading {img}')
+        _check_fname(img, overwrite='read', must_exist=True, name=name)
+        img = nib.load(img)
+    # get data
+    orig_data = np.array(img.dataobj).astype(np.float32)
+    # reorient data to RAS
+    ornt = nib.orientations.axcodes2ornt(
+        nib.orientations.aff2axcodes(img.affine)).astype(int)
+    ras_ornt = nib.orientations.axcodes2ornt('RAS')
+    ornt_trans = nib.orientations.ornt_transform(ornt, ras_ornt)
+    img_data = nib.orientations.apply_orientation(orig_data, ornt_trans)
+    orig_mgh = nib.MGHImage(orig_data, img.affine)
+    aff_trans = nib.orientations.inv_ornt_aff(ornt_trans, img.shape)
+    vox_ras_t = np.dot(orig_mgh.header.get_vox2ras_tkr(), aff_trans)
+    return img_data, vox_ras_t
+
+
+class ComboBox(QComboBox):
+    """Dropdown menu that emits a click when popped up."""
+
+    clicked = QtCore.pyqtSignal()
+
+    def showPopup(self):
+        """Override show popup method to emit click."""
+        self.clicked.emit()
+        super(ComboBox, self).showPopup()
+
+
+def _make_slice_plot(width=4, height=4, dpi=300):
+    fig = Figure(figsize=(width, height))
+    canvas = FigureCanvas(fig)
+    ax = fig.subplots()
+    fig.subplots_adjust(bottom=0, left=0, right=1,
+                        top=1, wspace=0, hspace=0)
+    ax.set_facecolor('k')
+    # clean up excess plot text, invert
+    ax.invert_yaxis()
+    ax.set_xticks([])
+    ax.set_yticks([])
+    return canvas, fig
+
+
+class IntracranialElectrodeLocator(QMainWindow):
+    """Locate electrode contacts using a coregistered MRI and CT."""
+
+    def __init__(self, info, trans, aligned_ct, subject=None,
+                 subjects_dir=None, groups=None, verbose=None):
+        """GUI for locating intracranial electrodes.
+
+        .. note:: Images will be displayed using orientation information
+                  obtained from the image header. Images will be resampled to
+                  dimensions [256, 256, 256] for display.
+        """
+        # initialize QMainWindow class
+        super(IntracranialElectrodeLocator, self).__init__()
+
+        if not info.ch_names:
+            raise ValueError('No channels found in `info` to locate')
+
+        # store info for modification
+        self._info = info
+        self._verbose = verbose
+
+        # load imaging data
+        self._subject_dir = _check_subject_dir(subject, subjects_dir)
+        self._load_image_data(aligned_ct)
+
+        self._ch_alpha = 0.5
+        self._radius = int(_CH_PLOT_SIZE // 100)  # starting 1/200 of image
+        # initialize channel data
+        self._ch_index = 0
+        # load data, apply trans
+        self._head_mri_t = _get_trans(trans, 'head', 'mri')[0]
+        self._mri_head_t = invert_transform(self._head_mri_t)
+        # load channels, convert from m to mm
+        self._chs = {name: apply_trans(self._head_mri_t, ch['loc'][:3]) * 1000
+                     for name, ch in zip(info.ch_names, info['chs'])}
+        self._ch_names = list(self._chs.keys())
+        # set current position
+        if np.isnan(self._chs[self._ch_names[self._ch_index]]).any():
+            self._ras = np.array([0., 0., 0.])
+        else:
+            self._ras = self._chs[self._ch_names[self._ch_index]].copy()
+        self._current_slice = apply_trans(
+            self._ras_vox_t, self._ras).round().astype(int)
+        self._group_channels(groups)
+
+        # GUI design
+
+        # Main plots: make one plot for each view; sagittal, coronal, axial
+        plt_grid = QGridLayout()
+        plts = [_make_slice_plot(), _make_slice_plot(), _make_slice_plot()]
+        self._figs = [plts[0][1], plts[1][1], plts[2][1]]
+        plt_grid.addWidget(plts[0][0], 0, 0)
+        plt_grid.addWidget(plts[1][0], 0, 1)
+        plt_grid.addWidget(plts[2][0], 1, 0)
+        self._renderer = _get_renderer(
+            name='IEEG Locator', size=(400, 400), bgcolor='w')
+        # TODO: should eventually make sure the renderer here is actually
+        # some PyVista(Qt) variant, not mayavi, otherwise the following
+        # call will fail (hopefully it's rare that people who want to use this
+        # have also set their MNE_3D_BACKEND=mayavi and/or don't have a working
+        # pyvistaqt setup; also hopefully the refactoring to use the
+        # Qt/notebook abstraction will make this easier, too):
+        plt_grid.addWidget(self._renderer.plotter)
+
+        # Channel selector
+        self._ch_list = QListView()
+        self._ch_list.setSelectionMode(Qt.QAbstractItemView.SingleSelection)
+        self._ch_list.setMinimumWidth(150)
+        self._set_ch_names()
+
+        # Plots
+        self._plot_images()
+
+        # Menus
+        button_hbox = self._get_button_bar()
+        slider_hbox = self._get_slider_bar()
+        bottom_hbox = self._get_bottom_bar()
+
+        # Put everything together
+        plot_ch_hbox = QHBoxLayout()
+        plot_ch_hbox.addLayout(plt_grid)
+        plot_ch_hbox.addWidget(self._ch_list)
+
+        main_vbox = QVBoxLayout()
+        main_vbox.addLayout(button_hbox)
+        main_vbox.addLayout(slider_hbox)
+        main_vbox.addLayout(plot_ch_hbox)
+        main_vbox.addLayout(bottom_hbox)
+
+        central_widget = QWidget()
+        central_widget.setLayout(main_vbox)
+        self.setCentralWidget(central_widget)
+
+        # ready for user
+        self._move_cursors_to_pos()
+        self._ch_list.setFocus()  # always focus on list
+
+    def _load_image_data(self, ct):
+        """Get MRI and CT data to display and transforms to/from vox/RAS."""
+        self._mri_data, self._vox_ras_t = _load_image(
+            op.join(self._subject_dir, 'mri', 'brain.mgz'),
+            'MRI Image', verbose=self._verbose)
+        self._ras_vox_t = np.linalg.inv(self._vox_ras_t)
+
+        self._voxel_sizes = np.array(self._mri_data.shape)
+        self._img_ranges = [[0, self._voxel_sizes[1], 0, self._voxel_sizes[2]],
+                            [0, self._voxel_sizes[0], 0, self._voxel_sizes[2]],
+                            [0, self._voxel_sizes[0], 0, self._voxel_sizes[1]]]
+
+        # ready ct
+        self._ct_data, vox_ras_t = _load_image(ct, 'CT', verbose=self._verbose)
+        if self._mri_data.shape != self._ct_data.shape or \
+                not np.allclose(self._vox_ras_t, vox_ras_t, rtol=1e-6):
+            raise ValueError('CT is not aligned to MRI, got '
+                             f'CT shape={self._ct_data.shape}, '
+                             f'MRI shape={self._mri_data.shape}, '
+                             f'CT affine={vox_ras_t} and '
+                             f'MRI affine={self._vox_ras_t}')
+
+        if op.exists(op.join(self._subject_dir, 'surf', 'lh.seghead')):
+            self._head = _read_mri_surface(
+                op.join(self._subject_dir, 'surf', 'lh.seghead'))
+            assert _frame_to_str[self._head['coord_frame']] == 'mri'
+        else:
+            warn('`seghead` not found, skipping head plot, see '
+                 ':ref:`mne.bem.make_scalp_surfaces` to add the head')
+            self._head = None
+        if op.exists(op.join(self._subject_dir, 'surf', 'lh.pial')):
+            self._lh = _read_mri_surface(
+                op.join(self._subject_dir, 'surf', 'lh.pial'))
+            assert _frame_to_str[self._lh['coord_frame']] == 'mri'
+            self._rh = _read_mri_surface(
+                op.join(self._subject_dir, 'surf', 'rh.pial'))
+            assert _frame_to_str[self._rh['coord_frame']] == 'mri'
+        else:
+            warn('`pial` surface not found, skipping adding to 3D '
+                 'plot. This indicates the Freesurfer recon-all '
+                 'has been modified and these files have been deleted.')
+            self._lh = self._rh = None
+
+    def _make_ch_image(self, axis):
+        """Make a plot to display the channel locations."""
+        # Make channel data higher resolution so it looks better.
+        ch_image = np.zeros((_CH_PLOT_SIZE, _CH_PLOT_SIZE)) * np.nan
+        vx, vy, vz = self._voxel_sizes
+
+        def color_ch_radius(ch_image, xf, yf, group, radius):
+            # Take the fraction across each dimension of the RAS
+            # coordinates converted to xyz and put a circle in that
+            # position in this larger resolution image
+            ex, ey = np.round(np.array([xf, yf]) * _CH_PLOT_SIZE).astype(int)
+            for i in range(-radius, radius + 1):
+                for j in range(-radius, radius + 1):
+                    if (i**2 + j**2)**0.5 < radius:
+                        # negative y because y axis is inverted
+                        ch_image[-(ey + i), ex + j] = group
+            return ch_image
+
+        for name, ras in self._chs.items():
+            # move from middle-centered (half coords positive, half negative)
+            # to bottom-left corner centered (all coords positive).
+            if np.isnan(ras).any():
+                continue
+            xyz = apply_trans(self._ras_vox_t, ras)
+            # check if closest to that voxel
+            dist = np.linalg.norm(xyz - self._current_slice)
+            if dist < self._radius:
+                x, y, z = xyz
+                group = self._groups[name]
+                r = self._radius - np.round(abs(dist)).astype(int)
+                if axis == 0:
+                    ch_image = color_ch_radius(
+                        ch_image, y / vy, z / vz, group, r)
+                elif axis == 1:
+                    ch_image = color_ch_radius(
+                        ch_image, x / vx, z / vx, group, r)
+                elif axis == 2:
+                    ch_image = color_ch_radius(
+                        ch_image, x / vx, y / vy, group, r)
+        return ch_image
+
+    @verbose
+    def _save_ch_coords(self, info=None, verbose=None):
+        """Save the location of the electrode contacts."""
+        logger.info('Saving channel positions to `info`')
+        if info is None:
+            info = self._info
+        for name, ch in zip(info.ch_names, info['chs']):
+            ch['loc'][:3] = apply_trans(
+                self._mri_head_t, self._chs[name] / 1000)  # mm->m
+
+    def _plot_images(self):
+        """Use the MRI and CT to make plots."""
+        # Plot sagittal (0), coronal (1) or axial (2) view
+        self._images = dict(ct=list(), chs=list(),
+                            cursor=list(), cursor2=list())
+        ct_min, ct_max = np.nanmin(self._ct_data), np.nanmax(self._ct_data)
+        text_kwargs = dict(fontsize=3, color='#66CCEE', family='monospace',
+                           weight='bold', ha='center', va='center')
+        xyz = apply_trans(self._ras_vox_t, self._ras)
+        for axis in range(3):
+            ct_data = np.take(self._ct_data, self._current_slice[axis],
+                              axis=axis).T
+            self._images['ct'].append(self._figs[axis].axes[0].imshow(
+                ct_data, cmap='gray', aspect='auto',
+                vmin=ct_min, vmax=ct_max))
+            self._images['chs'].append(
+                self._figs[axis].axes[0].imshow(
+                    self._make_ch_image(axis), aspect='auto',
+                    extent=self._img_ranges[axis],
+                    cmap=_CMAP, alpha=self._ch_alpha, vmin=0, vmax=_N_COLORS))
+            self._images['cursor'].append(
+                self._figs[axis].axes[0].plot(
+                    (xyz[axis], xyz[axis]), (0, self._voxel_sizes[axis]),
+                    color=[0, 1, 0], linewidth=1, alpha=0.5)[0])
+            self._images['cursor2'].append(
+                self._figs[axis].axes[0].plot(
+                    (0, self._voxel_sizes[axis]), (xyz[axis], xyz[axis]),
+                    color=[0, 1, 0], linewidth=1, alpha=0.5)[0])
+            # label axes
+            self._figs[axis].text(0.5, 0.05, _IMG_LABELS[axis][0],
+                                  **text_kwargs)
+            self._figs[axis].text(0.05, 0.5, _IMG_LABELS[axis][1],
+                                  **text_kwargs)
+            self._figs[axis].axes[0].axis(self._img_ranges[axis])
+            self._figs[axis].canvas.mpl_connect(
+                'scroll_event', self._on_scroll)
+            self._figs[axis].canvas.mpl_connect(
+                'button_release_event', partial(self._on_click, axis))
+        # add head and brain in mm (convert from m)
+        if self._head is not None:
+            self._renderer.mesh(
+                *self._head['rr'].T * 1000, triangles=self._head['tris'],
+                color='gray', opacity=0.2, reset_camera=False, render=False)
+        if self._lh is not None and self._rh is not None:
+            self._renderer.mesh(
+                *self._lh['rr'].T * 1000, triangles=self._lh['tris'],
+                color='white', opacity=0.2, reset_camera=False, render=False)
+            self._renderer.mesh(
+                *self._rh['rr'].T * 1000, triangles=self._rh['tris'],
+                color='white', opacity=0.2, reset_camera=False, render=False)
+        self._3d_chs = dict()
+        self._plot_3d_ch_pos()
+        self._renderer.set_camera(azimuth=90, elevation=90, distance=300,
+                                  focalpoint=tuple(self._ras))
+        # update plots
+        self._draw()
+        self._renderer._update()
+
+    def _scale_radius(self):
+        """Scale the radius to mm."""
+        shape = np.mean(self._ct_data.shape)  # this is Freesurfer shape (256)
+        scale = np.diag(self._ras_vox_t)[:3].mean()
+        return scale * self._radius * (shape / _CH_PLOT_SIZE)
+
+    def _update_camera(self, render=False):
+        """Update the camera position."""
+        self._renderer.set_camera(
+            # needs fix, distance moves when focal point updates
+            distance=self._renderer.plotter.camera.distance * 0.9,
+            focalpoint=tuple(self._ras),
+            reset_camera=False)
+
+    def _plot_3d_ch(self, name, render=False):
+        """Plot a single 3D channel."""
+        if name in self._3d_chs:
+            self._renderer.plotter.remove_actor(self._3d_chs.pop(name))
+        if not any(np.isnan(self._chs[name])):
+            radius = self._scale_radius()
+            self._3d_chs[name] = self._renderer.sphere(
+                tuple(self._chs[name]), scale=radius * 3,
+                color=_UNIQUE_COLORS[self._groups[name] % _N_COLORS],
+                opacity=self._ch_alpha)[0]
+        if render:
+            self._renderer._update()
+
+    def _plot_3d_ch_pos(self, render=False):
+        for name in self._chs:
+            self._plot_3d_ch(name)
+        if render:
+            self._renderer._update()
+
+    def _get_button_bar(self):
+        """Make a bar with buttons for user interactions."""
+        hbox = QHBoxLayout()
+
+        hbox.addStretch(10)
+
+        hbox.addWidget(QLabel('Snap to Center'))
+        self._snap_button = QPushButton('Off')
+        self._snap_button.setMaximumWidth(25)  # not too big
+        hbox.addWidget(self._snap_button)
+        self._snap_button.released.connect(self._toggle_snap)
+        self._toggle_snap()  # turn on to start
+
+        hbox.addStretch(1)
+
+        mark_button = QPushButton('Mark')
+        hbox.addWidget(mark_button)
+        mark_button.released.connect(self._mark_ch)
+
+        remove_button = QPushButton('Remove')
+        hbox.addWidget(remove_button)
+        remove_button.released.connect(self._remove_ch)
+
+        self._group_selector = ComboBox()
+        group_model = self._group_selector.model()
+
+        for i in range(_N_COLORS):
+            self._group_selector.addItem(' ')
+            color = QtGui.QColor()
+            color.setRgb(*(255 * np.array(_UNIQUE_COLORS[i % _N_COLORS])
+                           ).round().astype(int))
+            brush = QtGui.QBrush(color)
+            brush.setStyle(QtCore.Qt.SolidPattern)
+            group_model.setData(group_model.index(i, 0),
+                                brush, QtCore.Qt.BackgroundRole)
+        self._group_selector.clicked.connect(self._select_group)
+        self._group_selector.currentIndexChanged.connect(
+            self._update_group)
+        hbox.addWidget(self._group_selector)
+
+        # update background color for current selection
+        self._update_group()
+
+        return hbox
+
+    def _get_slider_bar(self):
+        """Make a bar with sliders on it."""
+
+        def make_label(name):
+            label = QLabel(name)
+            label.setAlignment(QtCore.Qt.AlignCenter)
+            return label
+
+        def make_slider(smin, smax, sval, sfun=None):
+            slider = QSlider(QtCore.Qt.Horizontal)
+            slider.setMinimum(int(round(smin)))
+            slider.setMaximum(int(round(smax)))
+            slider.setValue(int(round(sval)))
+            slider.setTracking(False)  # only update on release
+            if sfun is not None:
+                slider.valueChanged.connect(sfun)
+            slider.keyPressEvent = self._key_press_event
+            return slider
+
+        slider_hbox = QHBoxLayout()
+
+        ch_vbox = QVBoxLayout()
+        ch_vbox.addWidget(make_label('ch alpha'))
+        ch_vbox.addWidget(make_label('ch radius'))
+        slider_hbox.addLayout(ch_vbox)
+
+        ch_slider_vbox = QVBoxLayout()
+        self._alpha_slider = make_slider(0, 100, self._ch_alpha * 100,
+                                         self._update_ch_alpha)
+        ch_plot_max = _CH_PLOT_SIZE // 50  # max 1 / 50 of plot size
+        ch_slider_vbox.addWidget(self._alpha_slider)
+        self._radius_slider = make_slider(0, ch_plot_max, self._radius,
+                                          self._update_radius)
+        ch_slider_vbox.addWidget(self._radius_slider)
+        slider_hbox.addLayout(ch_slider_vbox)
+
+        ct_vbox = QVBoxLayout()
+        ct_vbox.addWidget(make_label('CT min'))
+        ct_vbox.addWidget(make_label('CT max'))
+        slider_hbox.addLayout(ct_vbox)
+
+        ct_slider_vbox = QVBoxLayout()
+        ct_min = int(round(np.nanmin(self._ct_data)))
+        ct_max = int(round(np.nanmax(self._ct_data)))
+        self._ct_min_slider = make_slider(
+            ct_min, ct_max, ct_min, self._update_ct_scale)
+        ct_slider_vbox.addWidget(self._ct_min_slider)
+        self._ct_max_slider = make_slider(
+            ct_min, ct_max, ct_max, self._update_ct_scale)
+        ct_slider_vbox.addWidget(self._ct_max_slider)
+        slider_hbox.addLayout(ct_slider_vbox)
+        return slider_hbox
+
+    def _get_bottom_bar(self):
+        """Make a bar at the bottom with information in it."""
+        hbox = QHBoxLayout()
+
+        hbox.addStretch(10)
+
+        self._intensity_label = QLabel('')  # update later
+        hbox.addWidget(self._intensity_label)
+
+        RAS_label = QLabel('RAS =')
+        self._RAS_textbox = QPlainTextEdit('')  # update later
+        self._RAS_textbox.setMaximumHeight(25)
+        self._RAS_textbox.setMaximumWidth(200)
+        self._RAS_textbox.focusOutEvent = self._update_RAS
+        self._RAS_textbox.textChanged.connect(self._check_update_RAS)
+        hbox.addWidget(RAS_label)
+        hbox.addWidget(self._RAS_textbox)
+        self._update_moved()  # update text now
+        return hbox
+
+    def _group_channels(self, groups):
+        """Automatically find a group based on the name of the channel."""
+        if groups is not None:
+            for name in self._ch_names:
+                if name not in groups:
+                    raise ValueError(f'{name} not found in ``groups``')
+                _validate_type(groups[name], (float, int), f'groups[{name}]')
+            self.groups = groups
+        else:
+            i = 0
+            self._groups = dict()
+            base_names = dict()
+            for name in self._ch_names:
+                # strip all numbers from the name
+                base_name = ''.join([letter for letter in name if
+                                     not letter.isdigit() and letter != ' '])
+                if base_name in base_names:
+                    # look up group number by base name
+                    self._groups[name] = base_names[base_name]
+                else:
+                    self._groups[name] = i
+                    base_names[base_name] = i
+                    i += 1
+
+    def _set_ch_names(self):
+        """Add the channel names to the selector."""
+        self._ch_list_model = QtGui.QStandardItemModel(self._ch_list)
+        for name in self._ch_names:
+            self._ch_list_model.appendRow(QtGui.QStandardItem(name))
+            self._color_list_item(name=name)
+        self._ch_list.setModel(self._ch_list_model)
+        self._ch_list.clicked.connect(self._go_to_ch)
+        self._ch_list.setCurrentIndex(
+            self._ch_list_model.index(self._ch_index, 0))
+        self._ch_list.keyPressEvent = self._key_press_event
+
+    def _select_group(self):
+        """Change the group label to the selection."""
+        group = self._group_selector.currentIndex()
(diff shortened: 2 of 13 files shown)


larsoner:
> a raw object that supplies the names to be found and is return with a modified montage.

It should be slightly more general to make this an instance of `info` that gets modified in-place.

> Very rough version, no review yet.

Let me know when to look and test! Ideally you would not need to tell me what to do and I can figure it out from the docs/tutorial etc.

alexrockhill:
> It should be slightly more general to make this an instance of `info` that gets modified in-place.

Yeah that's how it is in the code, the comments were a bit out-of-date

mne/gui/tests/test_ieeg_locate_gui.py
```diff
@@ -0,0 +1,22 @@
...
+data_path = testing.data_path(download=False)
+subject = 'sample'
+subjects_dir = op.join(data_path, 'subjects')
+sample_dir = op.join(data_path, 'MEG', subject)
+raw_path = op.join(sample_dir, 'sample_audvis_trunc_raw.fif')
+fname_trans = op.join(sample_dir, 'sample_audvis_trunc-trans.fif')
+
+
+def test_ieeg_elec_locate_gui_display():
+	mne.gui.locate_ieeg(subject, subjects_dir=subjects_dir)
```
agramfort, line 22: can you micmic clicks ? I suspect this test leads to limited coverage no?

alexrockhill: Yes, hopefully that will be done soon, I've been sorting out the API (e.g. other PR) but this is next

alexrockhill:
Still a work in progress, just for a demo, no need to review

alexrockhill:
Here's a script for use on the misc data

```
import os
import os.path as op

import numpy as np
import pandas as pd

import nibabel as nib
from dipy.align import resample

import mne

misc_path = mne.datasets.misc.data_path()
subject = 'sample_seeg'
subjects_dir = op.join(misc_path, 'seeg')

CT_orig = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_CT.mgz'))
raw = mne.io.read_raw(op.join(misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
raw.set_montage(None)
T1 = nib.load(op.join(misc_path, 'seeg', 'sample_seeg', 'mri', 'T1.mgz'))
reg_affine = np.array([
    [0.99270756, -0.03243313, 0.11610254, -133.094156],
    [0.04374389, 0.99439665, -0.09623816, -97.58320673],
    [-0.11233068, 0.10061512, 0.98856381, -84.45551601],
    [0., 0., 0., 1.]])
CT_aligned = mne.transforms.apply_volume_registration(CT_orig, T1, reg_affine)

mne.gui.locate_ieeg(raw.info, CT_aligned, subject='sample_seeg', subjects_dir=op.join(misc_path, 'seeg'))

```

alexrockhill:
I didn't find anything that helpful in `mne.viz.plot_bem` but I figured out the transforms by following this https://github.com/moloney/dcmstack/blob/master/src/dcmstack/dcmstack.py#L246:L252

larsoner:
> I didn't find anything that helpful in mne.viz.plot_bem

It produces outputs like this:

https://mne.tools/stable/auto_tutorials/forward/30_forward.html#sphx-glr-auto-tutorials-forward-30-forward-py

The slicing order and labeling of each axis has already been done in there somewhere. Following the trail:

1. https://github.com/mne-tools/mne-python/blob/474da964415c5233e955663b56e85531093be0a4/mne/viz/misc.py#L561
2. https://github.com/mne-tools/mne-python/blob/474da964415c5233e955663b56e85531093be0a4/mne/viz/misc.py#L322-L323
3. https://github.com/mne-tools/mne-python/blob/474da964415c5233e955663b56e85531093be0a4/mne/_freesurfer.py#L84-L97

And to see how data should be manipulated to plot properly, you want to look in `_plot_mri_contours`, though it gets a bit complicated:

https://github.com/mne-tools/mne-python/blob/474da964415c5233e955663b56e85531093be0a4/mne/viz/misc.py#L391

Ideally you could just use `_plot_mri_contours` directly, for example, and improve it to have any additional functionality you need, probably via refactoring.

larsoner:
... and note that doing it this way, the `affine` / `vox2ras` / `vox2ras_tkr` are those *of the original volume*, no messing around needed. You just plot such that you're in canonical orientation, and if someone clicks, you figure out which voxel that was in the volume (based on which slice you're plotting and whether or not it's flipped) and apply the affine/vox2ras/vox2ras_tkr from the original volume.

alexrockhill:
> https://github.com/mne-tools/mne-python/blob/474da964415c5233e955663b56e85531093be0a4/mne/viz/misc.py#L391
> 
> Ideally you could just use `_plot_mri_contours` directly, for example, and improve it to have any additional functionality you need, probably via refactoring.

So I followed to that point to but what I really need is the affine transformation as you were saying. Those are all really convoluted ways of getting what `nibabel.orientations.aff2axcodes` and `nibabel.orientations.axcodes2ornt` gives you in one or two lines that should be refactored. I'm trying to focus on the GUI right now and refactoring `plot_bem` when it doesn't contribute to the GUI doesn't seem like a good use of time until after the main things are finished at which point I would be happy to go back.

This part is the key part that is kind of in `plot_bem` but it's spread over many lines of code in a very confusing way

```
ras_ornt = nib.orientations.axcodes2ornt('RAS')
ornt_trans = nib.orientations.ornt_transform(ornt, ras_ornt)
img_data = nib.orientations.apply_orientation(orig_data, ornt_trans)
orig_mgh = nib.MGHImage(orig_data, img.affine)
aff_trans = nib.orientations.inv_ornt_aff(ornt_trans, img.shape)
vox_ras_t = np.dot(orig_mgh.header.get_vox2ras_tkr(), aff_trans)
```

Again, I'd be happy to go back and refactor it in `plot_bem` but it's not going to help the GUI.

Just to make the point clear, `flip_x` which is an output of the `_mri_orientation` in `_freesurfer.py` is used four separate times. The code I need is the above and refactoring that is going to take some time...

alexrockhill:
I'll try a separate PR to refactor the BEM code so that maybe it can share that code above maybe that will make things easier

larsoner:
> So I followed to that point to but what I really need is the affine transformation as you were saying... This part is the key part that is kind of in plot_bem but it's spread over many lines of code in a very confusing way

What I don't follow is *what affine do you need*? Why aren't you just using `img.affine`? What is your procedure for plotting data so that you need something else?

In the BEM code what we do is use the image's affine as is (if we need it), and then figure out how to slice into the volume to show standard orientations (sagittal, axial, coronal). To me this is a simple enough procedure and does what you should need for the GUI.

Perhaps if you describe conceptually what you do in the GUI and why you need some other affine it would help.

> refactoring plot_bem when it doesn't contribute to the GUI doesn't seem like a good use of time until after the main things are finished at which point I would be happy to go back

The BEM code 1) resolves orientations so that they can be plotted sagittal/axial/coronal, 2) shows the slice number, 3) displays L/R/A/P/I/S to the sides, 4) has contours-from-surfaces built in, 5) has been unit-tested and user-tested to make sure that non-standard orientations are shown properly. These all seem like useful features for the GUI, no?

larsoner:
... maybe you're talking about the mapping from GUI click->RAS? For this I would just figure out the click->volume index `i,j,k`, which is easy enough to get from knowing the sag/cor/ax slice that was clicked, where it was clicked, and the flip/slicing that was used in the current plot. Once you have this volume `i,j,k`, you just dot it with `img.affine` / `img.header.get_vox2ras_tkr`.

alexrockhill:
> What I don't follow is _what affine do you need_? Why aren't you just using `img.affine`? What is your procedure for plotting data so that you need something else?

I need an affine that goes from surface RAS to the RAS voxels of the reoriented image.

> In the BEM code what we do is use the image's affine as is (if we need it), and then figure out how to slice into the volume to show standard orientations (sagittal, axial, coronal). To me this is a simple enough procedure and does what you should need for the GUI.
> 
> Perhaps if you describe conceptually what you do in the GUI and why you need some other affine it would help.
> 
> The BEM code 1) resolves orientations so that they can be plotted sagittal/axial/coronal, 2) shows the slice number, 3) displays L/R/A/P/I/S to the sides, 4) has contours-from-surfaces built in, 5) has been unit-tested and user-tested to make sure that non-standard orientations are shown properly. These all seem like useful features for the GUI, no?

Yes, that is what is needed for the GUI but the flips and coordinate frames are a mess and all of that functionality is already in the GUI.

I'll work on a PR to reorient the MRI surface first so that it's not reoriented four separate times.

larsoner:
> I need an affine that goes from surface RAS to the RAS voxels of the reoriented image.

Ahh I think this is the sticking point. In the BEM code no explicit reorientation is done. The image is just sliced (and maybe flipped) as necessary to show properly oriented views. To me this is simple enough, and easy (enough) to know which voxel i/j/k is clicked, then use the standard `img.affine` to get RAS.

But if there is some way to reorient that doesn't actually involve resampling the image (the current code doesn't need to) and makes things simpler, that's great! (And the BEM code should benefit from this -- DRY, one right way to do a given thing, etc.!)

alexrockhill:
I think something is incompatible on Windows with the way I've implemented PyQt from the tests, it looks like there are segfaults in the tests. Any idea what's going on @larsoner or anyone else?

mne/gui/__init__.py
```diff
@@ -239,3 +239,39 @@ def kit2fiff():
...
+        Group assignments can be adjusted in the GUI.
+    %(verbose)s
+
+    """
+    from ._ieeg_locate_gui import IntracranialElectrodeLocator
+    from PyQt5.QtWidgets import QApplication
+    import matplotlib as mpl
+    # change to correct backend
+    mpl.use('Qt5Agg')
+    app = QApplication(['Intracranial Electrode Locator'])
```
larsoner, line 272: You should not instantiate a QApplication directly, first you check for `.instance` and if it's none you instantiate. Look at how our own `mne/viz/backends/_pyvista.py` and `mne/conftest.py` do it. In this case I think we should refactor the version in `_pyvista.py` to some new `mne.viz._qt` perhaps that just contains code to instantiate a QApplication and set it to use our icon (if we're the ones instantiating it). The `_pyvista.py` code already does this check-and-set, so this code just needs to be moved and used in multiple places

alexrockhill: Ok I can give that a shot

alexrockhill:
Ok, I pushed this because I needed it for some analysis now that the coordinate frames are all worked out but it is almost feature complete.

Things still left to do:
- Fix extra empty figure popping up from the renderer
- Speed up initialization
- Fix too many local maxima found (almost done)
- Add tube for each automatically detected device with a sphere for which contact is currently selected
- Add tests

I hope about a week away from being fully finished. We'll see though, some things can be quite finicky with the approach so we should test it on Liberty's data too and make sure everything works. In general it would be nice to avoid parameters about the contact size, spacing etc. as those will make it more confusing but they can limit the number of false positives so that takes some finagling to find a good balance.

mne/_freesurfer.py
```diff
@@ -422,6 +432,119 @@ def get_mni_fiducials(subject, subjects_dir=None, verbose=None):
     return fids
 
 
+@verbose
+def estimate_head_mri_t(subject, subjects_dir=None, verbose=None):
+    """Use the estimate the head->mri transform.
```
mmagnuski, line 437:
```suggestion
    """Estimate the head->mri transform.
```

agramfort:
@alexrockhill do you need feedback from me here?

alexrockhill:
> @alexrockhill do you need feedback from me here?

Now is fine or I'm adding tests so I will change a few more things and maybe I will find my own errors first over the next day or two but it is very close to being done.

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -124,8 +127,12 @@ def plot_overlay(image, compare, title, thresh=None):
...
 # here::
 #
-#    reg_affine, _ = mne.transforms.compute_volume_registration(
-#        CT_orig, T1, pipeline='rigids', verbose=True)
+#    from dipy.align import center_of_mass, translation, rigid
+#    reg_data, reg_affine = dipy.align.affine_registration(
+#         moving=np.array(CT_orig.dataobj), moving_affine=CT_orig.affine,
+#         static=np.array(T1.dataobj), static_affine=T1.affine,
+#         pipeline=[center_of_mass, translation, rigid])
+#    CT_aligned = nib.MGHImage(reg_data.astype(np.float32), T1.affine)
```
larsoner, line 135: -1 on this, it means we have something wrong with our `compute_volume_registration` code that should be fixed. Our pipeline should do something equivalent to theirs. Maybe the simplest thing is to actually use their pipeline under the hood in our code?

alexrockhill: Yeah, we could do that

alexrockhill: Ah so the main thing we added that wasn't supported by their pipeline is `zooms`

alexrockhill:
@adam2392, if you want to review the core automatic detection algorithm that would be a huge help. I haven't tried it on ECoG yet but I wrote it so that hopefully it will work there too

alexrockhill:
The tutorial looks much better when you use relative thresholds for the contacts (0.5 of max), I guess this will be a moot point when Dipy merges their improvement but that doesn't look like it will be ready imminently

adam2392:
> @adam2392, if you want to review the core automatic detection algorithm that would be a huge help. I haven't tried it on ECoG yet but I wrote it so that hopefully it will work there too

Where is this located?

alexrockhill:
> > @adam2392, if you want to review the core automatic detection algorithm that would be a huge help. I haven't tried it on ECoG yet but I wrote it so that hopefully it will work there too
> 
> Where is this located?

In the GUI, you press `Auto-Find`. The code is here https://github.com/mne-tools/mne-python/pull/9586/files#diff-ebee30ced30075bc78562aee0f3378e49e86d117f5edf6a9e1199296e287dfb7R955-R1016

alexrockhill:
> Here's a script for use on the misc data
> 
> ```
> import os.path as op
> import numpy as np
> import pandas as pd
> 
> import nibabel as nib
> from dipy.align import resample
> 
> import mne
> 
> misc_path = mne.datasets.misc.data_path()
> subject = 'sample_seeg'
> subjects_dir = op.join(misc_path, 'seeg')
> 
> CT_orig = nib.load(op.join(misc_path, 'seeg', 'sample_seeg_CT.mgz'))
> raw = mne.io.read_raw(op.join(misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
> 
> T1 = nib.load(op.join(misc_path, 'seeg', 'sample_seeg', 'mri', 'T1.mgz'))
> reg_affine = np.array([
>     [0.99270756, -0.03243313, 0.11610254, -133.094156],
>     [0.04374389, 0.99439665, -0.09623816, -97.58320673],
>     [-0.11233068, 0.10061512, 0.98856381, -84.45551601],
>     [0., 0., 0., 1.]])
> CT_aligned = mne.transforms.apply_volume_registration(CT_orig, T1, reg_affine)
> 
> trans = mne.coreg.estimate_head_mri_t(subject, subjects_dir)
> gui = mne.gui.locate_ieeg(raw.info, trans, CT_aligned, subject=subject, subjects_dir=subjects_dir)
> ```

@adam2392, see this script for running on the sEEG data. Maybe we should put the ECoG data in "head" too now that I think about it. I'll write a quick PR on misc data and mne-python.

alexrockhill:
Well I guess the ECoG in https://github.com/mne-tools/mne-misc-data is no longer used actually since it's pulled from BIDS. I don't see a CT there though so that wouldn't work to check for this.

You could test ECoG with Liberty's data from here: https://zenodo.org/record/996814

```
import os.path as op
import numpy as np
import pandas as pd

import nibabel as nib

import mne

subject = 'test_subj'
subjects_dir = '.'  # might have to change this
CT_aligned = nib.load(op.join(subjects_dir, subject, 'CT', 'rCT.nii'))
info = mne.create_info([f'chan{i}' for i in range(343)], 1000, ['ecog'] * 343)
raw = mne.io.RawArray(np.zeros((343, 1000)), info)

trans = mne.coreg.estimate_head_mri_t(subject, subjects_dir)
gui = mne.gui.locate_ieeg(raw.info, trans, CT_aligned, subject=subject, subjects_dir=subjects_dir)
```

tutorials/clinical/10_ieeg_localize.py
alexrockhill: After thinking about it, I think this might be okay, the CT-MR alignment is incredibly finicky and does not refactor well with the much more stable MR-MR alignments. If you think it's worth looking into more detail what is going wrong, I'm fine but seems like not the best use of time to me. We have a solution that works and the API is pretty good. To me, it's much better that it works optimally and looks a little bit worse than if everything is all nicely factored but it's slightly inaccurate and I've spent a lot of time trying to fix this part already.

larsoner: >  ... the CT-MR alignment is incredibly finicky and does not refactor well with the much more stable MR-MR alignments... To me, it's much better that it works optimally and looks a little bit worse than if everything is all nicely factored but it's slightly inaccurate and I've spent a lot of time trying to fix this part already.

To me it sounds like you are implicitly arguing for getting rid of our own `compute_volume_registration` / `apply_volume_registration` code in favor of directly calling `dipy` functions. This is because the real use case we currently have for our own versions of these functions is exactly for doing MR-CT alignment. Should we get rid of them?

alexrockhill: Sounds reasonable to me

alexrockhill:
@larsoner, would you happen to know why this is failing?

drammock:
Did you look at the CI log?
```
/home/circleci/.local/lib/python3.8/site-packages/mprof.py:76: SyntaxWarning: "is" with a literal. Did you mean "=="?
[...]
/home/circleci/project/doc/auto_tutorials/clinical/10_ieeg_localize.rst:348: WARNING: Enumerated list ends without a blank line; unexpected unindent.
[...]
/home/circleci/project/doc/auto_tutorials/intro/40_sensor_locations.rst:311: WARNING: undefined label: plot_source_alignment
/home/circleci/project/doc/auto_tutorials/io/30_reading_fnirs_data.rst:259: WARNING: undefined label: plot_source_alignment
/home/circleci/project/doc/auto_tutorials/io/30_reading_fnirs_data.rst:293: WARNING: undefined label: plot_source_alignment
/home/circleci/project/mne/channels/montage.py:docstring of mne.channels.montage.add_estimated_fiducials:38: WARNING: undefined label: plot_source_alignment
/home/circleci/project/mne/gui/__init__.py:docstring of mne.gui.locate_ieeg:54: WARNING: py:obj reference target not found: IntracranialElectrodeLocator
/home/circleci/project/doc/overview/cookbook.rst:316: WARNING: undefined label: plot_source_alignment
_includes/forward.rst:674: WARNING: undefined label: plot_source_alignment_without_mri
changes/0.21.inc:45: WARNING: undefined label: plot_source_alignment
```
The first one is a problem in nilearn that I've already patched but they haven't released a new point-version yet. The rest look like they're probably related to this PR?

adam2392:
@thebrainchain feel free to test this out which may significantly improve your existing workflow!

alexrockhill:
Ok so I optimized the algorithm I had and fixed a few bugs. I think this could be merged now, but I'm going to give an alternate Hough transformation approach a try and see how it does. It looks like it has quite a few parameters so we'll see, but I think this is about as good as this approach can get which I said was a "C-" and I don't think is good enough.

tutorials/clinical/10_ieeg_localize.py
larsoner: @alexrockhill can you try reverting the resample-each-step in favor of `starting_affine`? Having recently made a PR to dipy I see that's all their code does really. If it indeed works as well as `affine_registration` (when all `niter` are used the same way) then I wouldn't mind keeping `compute_volume_registration`, because 1) it supports `zooms` and 2) stays in sync with our `compute_source_morph` code.

We can set `niter_*` to whatever they need to be to be equivalent with `affine_registration` because `compute_volume_registration` is a new function.

alexrockhill:
Hmmm so it turned out the real reason it was missing so many electrodes was due to the preprocessing; they were zeroed out before the line algorithm started

alexrockhill:
Ok, removing the auto-find feature this PR after talking to @agramfort so, ready for review.

The plan is to add the `auto-find` feature as a separate PR.

tutorials/clinical/10_ieeg_localize.py
larsoner: This has not been addressed

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+#   - Click in each image to naviate to each electrode contact
+#   - Select the contact name in the right panel
+#   - Press "Mark" to associate that position with that contact
+#
+#   .. note: Using the scroll or +/- arrow keys you can zoom in and
+#            out, and the up/down, left/right and page up/page down keys
+#            allow you to move one slice in any direction. This information
+#            is available in the help menu, accessible by pressing "h".
+#
+# 2) Using "Auto-Find"
```
larsoner, line 272: This needs to be removed

larsoner:
> The plan is to add the auto-find feature as a separate PR.

Makes sense to me, that should be easier to review.

@adam2392 can you look and try again and see if you're happy? @jasmainak as well? You can try following the instructions in https://github.com/mne-tools/mne-python/pull/9586/files#diff-f802a9dfc7b418c33db03c47e5eb7353f3e124121ff07c4b867fa274d83d0d88R241 to mark electrodes and see how it works! If it's not clear from the instructions what to do, that's already useful feedback on how to improve the tutorial so you can test it more easily (and others can use it effectively).

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
 # point in CT space, it is the same point in MR space. So now everything is
 # ready to determine the location of each electrode contact in the
-# individual subject's anatomical space (T1-space). To do this, can make
-# list of ``TkReg RAS`` points from the lower panel in freeview or use the
-# mne graphical user interface (coming soon). The electrode locations will then
-# be in the ``surface RAS`` coordinate frame, which is helpful because that is
-# the coordinate frame that all the surface and image files that freesurfer
-# outputs are in, see :ref:`tut-freesurfer-mne`.
+# individual subject's anatomical space (T1-space). To do this, we can use the
+# MNE intracranial electrode location graphical user interface. The electrode
```
adam2392, line 249: Leftover sentence?

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
-#     $ freeview $MISC_PATH/seeg/sample_seeg_T1.mgz \
-#           $MISC_PATH/seeg/sample_seeg_CT.mgz
-
-# %%
-# Now, we'll make a montage with the channels that we've found in the
-# previous step.
+# 1) Manually
 #
-# .. note:: MNE represents data in the "head" space internally
+#   - Click in each image to naviate to each electrode contact
```
adam2392, line 263:
```suggestion
#   - Click in each image to navigate to each electrode contact
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+#   - Select the contact name in the right panel
+#   - Press "Mark" to associate that position with that contact
```
adam2392, line 265:
```suggestion
#   - Press "Mark", or "M" to associate that position with that contact
```

IIRC, that was the keyboard shortcut right?

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+#
+#   .. note: Using the scroll or +/- arrow keys you can zoom in and
+#            out, and the up/down, left/right and page up/page down keys
+#            allow you to move one slice in any direction. This information
+#            is available in the help menu, accessible by pressing "h".
```
adam2392, line 270: To be more explicit, I would then perhaps add something along the lines of:

- Repeat for each contact, until all contacts are marked. You can tell contacts are marked if they are colored(?) in. 
- When you are done, you can close the GUI and the contact coordinates will be saved in the Raw object

doc/preprocessing.rst
```diff
@@ -183,6 +183,7 @@ EEG referencing:
    :toctree: generated/
 
    Transform
+   hough_transform_3D
```
agramfort, line 186: remove?

mne/gui/tests/test_ieeg_locate_gui.py
```diff
@@ -0,0 +1,90 @@
...
+    center = np.array(brain.shape) / 2
+    # make image
+    ct_data = np.zeros(brain.shape, dtype=np.float32)
+    # make skull
+    for vert in verts:
+        x, y, z = np.round(vert).astype(int)
+        for i in range(-skull_size, skull_size + 1):
+            for j in range(-skull_size, skull_size + 1):
+                for k in range(-skull_size, skull_size + 1):
+                    ct_data[x + i, y + j, z + k] = 1000
```
agramfort, line 44: can you vectorize this?

mne/gui/tests/test_ieeg_locate_gui.py
```diff
@@ -0,0 +1,90 @@
...
+    # add electrode with contacts
+    for (x, y, z) in ch_coords:
+        # make sure not in skull
+        assert np.linalg.norm(center - np.array((x, y, z))) < 50
+        for i in range(-contact_size, contact_size + 1):
+            for j in range(-contact_size, contact_size + 1):
+                for k in range(-contact_size, contact_size + 1):
+                    ct_data[x + i, y + j, z + k] = 1000
```
agramfort, line 52: please vectorize this

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -46,6 +49,45 @@
 # use mne-python's fsaverage data
 fetch_fsaverage(subjects_dir=subjects_dir, verbose=True)  # downloads if needed
 
+###############################################################################
+# Aligning the T1 to ACPC
+# =======================
+#
+# For intracranial electrophysiology recordings, the Brain Imaging Data
+# Structure standard requires that coordinates be aligned to the anterior
+# commissue and posterior commissure (ACPC-aligned). This is recommended
```
agramfort, line 58:
```suggestion
# commissure and posterior commissure (ACPC-aligned). This is recommended
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+# see :ref:`tut-source-alignment`). For sEEG, it is convenient to get an
+# estimate of the location of the fiducial points for the subject
+# using the Talairach transform (see :func:`mne.coreg.get_mni_fiducials`)
+# to use to define the coordinate frame so that we don't have to manually
+# identify their location.
+
+# create symbolic link to share ``subjects_dir``
+if not op.exists(op.join(subjects_dir, 'sample_seeg')):
+    os.symlink(op.join(misc_path, 'seeg', 'sample_seeg'),
+               op.join(subjects_dir, 'sample_seeg'))
```
agramfort, line 236: I don't like this. You alter the mne-sample-data folder. Why not fetching fsaverage in the sample_seeg folder?

tutorials/clinical/10_ieeg_localize.py
agramfort, line 263:
```suggestion
#   - Click in each image to navigate to each electrode contact
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+#            allow you to move one slice in any direction. This information
+#            is available in the help menu, accessible by pressing "h".
+#
+# 2) Using "Auto-Find"
+#
+#   - Find a typical electrode contact err toward the largest type used
+#   - Press "Mark" to put a marker in that location
+#   - Adjust the radius slider so that it is just bigger than the contact
+#   - Adjust the CT threshold slider so that the contacts have black
+#     space between them
```
agramfort, line 278:
```suggestion
#   - Find a typical electrode contact err toward the largest type used.
#   - Press "Mark" to put a marker in that location.
#   - Adjust the radius slider so that it is just bigger than the contact.
#   - Adjust the CT threshold slider so that the contacts have black
#     space between them.
```

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
+#   - Press "Auto-Find" and wait several seconds for it to compute.
+#   - Navigate using the "Next/Previous Device" and "Next" and "Previous"
+#     to each automatically found contact and associate that position with
```
agramfort, line 281: automatically still relevant?

tutorials/clinical/10_ieeg_localize.py
```diff
@@ -173,42 +219,81 @@ def plot_overlay(image, compare, title, thresh=None):
...
 raw = mne.io.read_raw(op.join(misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
 
-# create symbolic link to share ``subjects_dir``
-if not op.exists(op.join(subjects_dir, 'sample_seeg')):
-    os.symlink(op.join(misc_path, 'seeg', 'sample_seeg'),
-               op.join(subjects_dir, 'sample_seeg'))
+gui = mne.gui.locate_ieeg(raw.info, subj_trans, CT_aligned,
+                          subject='sample_seeg', subjects_dir=subjects_dir)
+# The `raw` object is modified to contain the channel locations
+# after finishing the gui and can now be saved
```
agramfort, line 295:
```suggestion
# after closing the GUI and can now be saved.
```

agramfort (commented):
disclaimer I did not review the GUI code just the example and tests

tutorials/clinical/10_ieeg_localize.py
alexrockhill: This was @larsoner 's suggested way of doing it

agramfort: why @larsoner ? I see this a side effects from a script

larsoner: Seems like a fairly harmless one. `fetch_fsaverage` to `mne-misc-data` directory also modifies / creates a side effect, just to that dataset instead of `sample`. This could be done, but it will use some hundreds of MBs and some download time rather than being instantaneous. I'm fine either way

alexrockhill: Can we just remove the `symlink` when we're done?

alexrockhill:
> > @adam2392 can you look and try again and see if you're happy? @jasmainak as well? You can try following the instructions in https://github.com/mne-tools/mne-python/pull/9586/files#diff-f802a9dfc7b418c33db03c47e5eb7353f3e124121ff07c4b867fa274d83d0d88R241 to mark electrodes and see how it works! If it's not clear from the instructions what to do, that's already useful feedback on how to improve the tutorial so you can test it more easily (and others can use it effectively).
> 
> Mostly is fine to me! I added some minor comments. Looking forward to seeing the "auto-find"/"auto-initiate" feature work out.
> 
> Is it worth adding "tips" on sEEG and ECoG, such as:
> 
> 1. for Seeg, probably best to start at the deepest channel first and then locate along the electrode one by one?
> 2. For ECoG, it is best to determine using anatomical information, where the corners of the grid are first and annotate them? Then once can easily fill in the rest?
> 
> ^ I tell all my ugrads that so it's nice to have it just somewhere for consistency?

I think think the best option depends on the implantation and there isn't really a better order IMO. I think maybe wait for the `auto-find` feature and then that will dictate the order

alexrockhill:
Why do we `wrapped_stdout` for the `dipy.align.affine_registration` call? It prevents the text from being displayed after every level iteration and instead causes all the text to be displayed at the end

alexrockhill:
Ok this is okay to me... it's the version that worked with most of the subjects for the CT-MRI registration (although `affine_registration` and this `starting_affine` registration are close but not identical) but it's still a bit off for this example subject. This affine registration with the CT is a bit of a dark magic; it's really finicky. If you want to take a look at it @larsoner, you're welcome to. Otherwise, we can do it in a separate PR and this one can be good to go without it.

tutorials/clinical/10_ieeg_localize.py
agramfort: hum... keep it like this then but I would suggest to do this really at the top of the file with a big warning that it's just for the demo here. You should not have to do this on your own data.

alexrockhill: I originally had a `subjects_dir` parameter for the template as well, perhaps we should just put that back

larsoner: Works for me

mne/gui/_ieeg_locate_gui.py
```diff
@@ -0,0 +1,859 @@
...
+
+        # Main plots: make one plot for each view; sagittal, coronal, axial
+        plt_grid = QGridLayout()
+        self._plts = [SlicePlot(self), SlicePlot(self), SlicePlot(self)]
+        plt_grid.addWidget(self._plts[0], 0, 0)
+        plt_grid.addWidget(self._plts[1], 0, 1)
+        plt_grid.addWidget(self._plts[2], 1, 0)
+        self._renderer = create_3d_figure(
+            size=(400, 400), bgcolor='w', scene=False)
+        plt_grid.addWidget(self._renderer.figure.plotter)
```
larsoner, line 153: This has a failure in the `--pre` branch that uses latest GitHub `main`/`master` of PyVista and PyVistaQt, can you replicate and fix?

mne/defaults.py
```diff
@@ -105,10 +105,10 @@
...
         translation=None, rigid=None, affine=None, sdr=None),
     transform_niter=dict(
-        translation=(100, 100, 10),
-        rigid=(100, 100, 10),
-        affine=(100, 100, 10),
-        sdr=(5, 5, 3)),
+        translation=(10000, 1000, 100),
+        rigid=(10000, 1000, 100),
+        affine=(10000, 1000, 100),
+        sdr=(10, 10, 5)),
```
larsoner, line 111: Revert this too until we can make a separate PR to sort out `compute_volume_registration`? Also please open an issue for this and I'll milestone for 0.24 so we don't forget

mne/gui/_ieeg_locate_gui.py
```diff
@@ -0,0 +1,859 @@
...
+from ..transforms import (apply_trans, _frame_to_str, _get_trans,
+                          invert_transform)
+from ..utils import (logger, _check_fname, requires_nibabel,
+                     _validate_type, verbose, warn)
+
+_IMG_LABELS = [['I', 'P'], ['I', 'L'], ['P', 'L']]
+_CH_PLOT_SIZE = 1024
+_ZOOM_STEP_SIZE = 5
+
+_UNIQUE_COLORS = [(0.1, 0.42, 0.43), (0.9, 0.34, 0.62), (0.47, 0.51, 0.3),
```
larsoner, line 36: Some comment on how these were derived would be good. Even better would be to use the matplotlib color cycle, as it's designed for this purpose IIRC. If this is the matplotlib color cycle, it would be better to use a private function like `_get_unique_colors` that quickly computed this array and returned it. This would allow for example easy customization using `with plt.rc_context({'prop_cycle': ...}):` before calling the GUI

mne/gui/_ieeg_locate_gui.py
```diff
@@ -0,0 +1,859 @@
...
+
+# Authors: Alex Rockhill <aprockhill@mailbox.org>
+#
+# License: BSD (3-clause)
+
+import os.path as op
+import numpy as np
+from functools import partial
+
+import matplotlib.pyplot as plt
```
larsoner, line 12: You should not do this if possible because it will cause matplotlib to instantiate its GUI backend. If you refactor the creation below you end up not needing it anyway.

larsoner (commented):
Just some overarching code thoughts so far, I haven't tested the GUI myself yet

mne/gui/__init__.py
```diff
@@ -239,3 +239,44 @@ def kit2fiff():
...
+    -------
+    gui : instance of IntracranialElectrodeLocator
+        The graphical user interface (GUI) window.
+    """
+    from ._ieeg_locate_gui import IntracranialElectrodeLocator
+    from PyQt5.QtWidgets import QApplication
+    # get application
+    app = QApplication.instance()
+    if app is None:
+        app = QApplication(["Intracranial Electrode Locator"])
```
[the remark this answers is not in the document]
alexrockhill: I thought we were in agreement with @agramfort that we would do this in the future when the overhead cost was lower...

mne/gui/_ieeg_locate_gui.py
```diff
@@ -0,0 +1,859 @@
...
+from matplotlib.colors import LinearSegmentedColormap
+
+from PyQt5 import QtCore, QtGui, Qt
+from PyQt5.QtCore import pyqtSlot
+from PyQt5.QtWidgets import (QMainWindow, QGridLayout,
+                             QVBoxLayout, QHBoxLayout, QLabel,
+                             QMessageBox, QWidget,
+                             QListView, QSlider, QPushButton,
+                             QComboBox, QPlainTextEdit)
+from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
```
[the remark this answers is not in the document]
alexrockhill: Sure, doing that now

mne/gui/_ieeg_locate_gui.py
alexrockhill: I used these colors because the color cycle didn't work very well...

I'll add a comment.

mne/gui/__init__.py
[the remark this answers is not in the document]
larsoner: Okay I just added it to #9710

alexrockhill:
Hey not sure if it's the development version of pyvista or the changes to the matplotlib figure rendering that you suggested but the small lag on my mac is totally gone and it looks cleaner, thanks @larsoner

jasmainak:
I'll give it a shot later tonight!

mne/surface.py
```diff
@@ -1889,14 +1901,14 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
     _warn_missing_chs(montage, image_to, after_warp=True)
 
     # recover the contact positions as the center of mass
-    warped_data = np.asanyarray(image_to.dataobj)
+    warped_data = np.array(image_to.dataobj)
```
larsoner, line 1904: Why? You shouldn't need to make a copy like this

mne/surface.py
```diff
@@ -1889,14 +1901,14 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
...
     for val, ch_coord in enumerate(ch_coords, 1):
-        ch_coord[:] = np.asanyarray(
+        ch_coord[:] = np.array(
             np.where(warped_data == val), float).mean(axis=1)
```
larsoner, line 1907: Same. In fact you shouldn't even need the `np.array` at all:
```suggestion
        ch_coord[:] = (np.where(warped_data == val), float).mean(axis=1)
```

larsoner (commented):
Looking into the PyVista errors now, they are weird...

mne/gui/tests/test_ieeg_locate_gui.py
```diff
@@ -0,0 +1,136 @@
...
+    trans = mne.read_trans(fname_trans)
+    with pytest.warns(RuntimeWarning, match='`pial` surface not found'):
+        gui = mne.gui.locate_ieeg(raw.info, trans, aligned_ct,
+                                  subject=subject, subjects_dir=subjects_dir)
+
+    gui._ras[:] = coords[0]  # start in the right position
+    gui._move_cursors_to_pos()
+    for coord in coords:
+        coord_vox = mne.transforms.apply_trans(gui._ras_vox_t, coord)
+        _fake_click(gui._figs[2], gui.figs[2].axes[0],
```
larsoner, line 100: From the logs:
```
mne\gui\tests\test_ieeg_locate_gui.py:100: in test_ieeg_elec_locate_gui_display
    _fake_click(gui._figs[2], gui.figs[2].axes[0],
E   AttributeError: 'IntracranialElectrodeLocator' object has no attribute 'figs'
```
It's possible the PyVista errors all just come from this (?), so I'll fix it here and hopefully that fixes CIs.
```suggestion
        _fake_click(gui._figs[2], gui._figs[2].axes[0],
```

jasmainak:
I get:

```py
AttributeError                            Traceback (most recent call last)
~/Documents/github_repos/mne-python/tutorials/clinical/10_ieeg_localize.py in <module>
     90 
     91 T1 = nib.load(op.join(misc_path, 'seeg', 'sample_seeg', 'mri', 'T1.mgz'))
---> 92 viewer = T1.orthoview()
     93 viewer.set_position(0, 9.9, 5.8)
     94 viewer.figs[0].axes[0].annotate(

~/anaconda3/envs/mne/lib/python3.8/site-packages/nibabel/spatialimages.py in orthoview(self)
    604         the figure.
    605         """
--> 606         return OrthoSlicer3D(self.dataobj, self.affine,
    607                              title=self.get_filename())
    608 

~/anaconda3/envs/mne/lib/python3.8/site-packages/nibabel/viewers.py in __init__(***failed resolving arguments***)
    211         self._set_volume_index(0, update_slices=False)
    212         self._set_position(0., 0., 0.)
--> 213         self._draw()
    214 
    215     def __repr__(self):

~/anaconda3/envs/mne/lib/python3.8/site-packages/nibabel/viewers.py in _draw(self)
    500         for ii in range(3):
    501             ax = self._axes[ii]
--> 502             ax.draw_artist(self._ims[ii])
    503             if self._cross:
    504                 for line in self._crosshairs[ii].values():

~/anaconda3/envs/mne/lib/python3.8/site-packages/matplotlib/axes/_base.py in draw_artist(self, a)
   2932         """
   2933         if self.figure._cachedRenderer is None:
-> 2934             raise AttributeError("draw_artist can only be used after an "
   2935                                  "initial draw which caches the renderer")
   2936         a.draw(self.figure._cachedRenderer)

AttributeError: draw_artist can only be used after an initial draw which caches the renderer
```

am I missing some dependency? I have latest `matplotlib` and `nibabel`

larsoner:
Are you using the macOS backend or PyQt5? From what I've seen, the macOS backend has some problems, I wouldn't be surprised if this were another of them. `T1.orthoview()` has worked for me for ages, and presumably for `nibabel`'s CIs as well...

larsoner:
(For now you can just skip that bit of orthoview plotting in the tutorial, too -- I don't think it's critical for testing the GUI)

jasmainak:
Humm ... I had been using:

```sh
$ ipython --matplotlib
```

Now I commented out that part but get a new error:

```py
AttributeError                            Traceback (most recent call last)
~/Documents/github_repos/mne-python/tutorials/clinical/10_ieeg_localize.py in <module>
    285 raw = mne.io.read_raw(op.join(misc_path, 'seeg', 'sample_seeg_ieeg.fif'))
    286 
--> 287 gui = mne.gui.locate_ieeg(raw.info, subj_trans, CT_aligned,
    288                           subject='sample_seeg',
    289                           subjects_dir=op.join(misc_path, 'seeg'))

<decorator-gen-535> in locate_ieeg(info, trans, aligned_ct, subject, subjects_dir, groups, verbose)

~/Documents/github_repos/mne-python/mne/gui/__init__.py in locate_ieeg(info, trans, aligned_ct, subject, subjects_dir, groups, verbose)
    276     if app is None:
    277         app = QApplication(["Intracranial Electrode Locator"])
--> 278     gui = IntracranialElectrodeLocator(
    279         info, trans, aligned_ct, subject=subject,
    280         subjects_dir=subjects_dir, groups=groups, verbose=verbose)

~/Documents/github_repos/mne-python/mne/gui/_ieeg_locate_gui.py in __init__(self, info, trans, aligned_ct, subject, subjects_dir, groups, verbose)
    148         plt_grid.addWidget(plts[1][0], 0, 1)
    149         plt_grid.addWidget(plts[2][0], 1, 0)
--> 150         self._renderer = _get_renderer(
    151             name='IEEG Locator', size=(400, 400), bgcolor='w')
    152         # TODO: should eventually make sure the renderer here is actually

~/Documents/github_repos/mne-python/mne/viz/backends/renderer.py in _get_renderer(*args, **kwargs)
     39 def _get_renderer(*args, **kwargs):
     40     _get_3d_backend()
---> 41     return backend._Renderer(*args, **kwargs)
     42 
     43 

~/Documents/github_repos/mne-python/mne/viz/backends/_qt.py in __init__(self, *args, **kwargs)
    515                 _QtStatusBar, _QtWindow, _QtPlayback):
    516     def __init__(self, *args, **kwargs):
--> 517         super().__init__(*args, **kwargs)
    518         self._window_initialize()
    519 

~/Documents/github_repos/mne-python/mne/viz/backends/_pyvista.py in __init__(self, fig, size, bgcolor, name, show, shape, notebook, smooth_shading)
    183                 self.figure.smooth_shading = False
    184             with _disabled_depth_peeling():
--> 185                 self.plotter = self.figure.build()
    186             self._hide_axes()
    187             self._enable_aa()

~/Documents/github_repos/mne-python/mne/viz/backends/_pyvista.py in build(self)
    103                 plotter.set_icon(f":/mne-{kind}icon.png")
    104         if self.plotter.iren is not None:
--> 105             self.plotter.iren.initialize()
    106         _process_events(self.plotter)
    107         _process_events(self.plotter)

AttributeError: 'vtkmodules.vtkRenderingUI.vtkGenericRenderWindowIn' object has no attribute 'initialize'
```

upgrading `pyvistaqt` didn't seem to help ...

larsoner:
> upgrading pyvistaqt didn't seem to help ...

I would try in a fresh MNE environment created according to [our current installation instructions](https://mne.tools/dev/install/index.html). Also if you're on an ARM macOS machine then you probably need to use the i686 version of Anaconda rather than the native ARM one, I doubt all packages are available and working for ARM

jasmainak:
Okay finally got this to work. Thanks @larsoner !

image

Disclaimer, I have never done this myself. So, this is mostly a naive user perspective.

1) Why is the CT/MRI image inside two rectangles?
2) I can't understand the bottom right plot. When I click on a marker, is it supposed to center and mark the corresponding channel in bottom right? That doesn't seem to be happening
3) Is there a way to turn off the overlay of the CT image over the MRI? Or does it only show MRI?
4) I tried changing the color of one channel. Didn't seem to work ...
5) Can there be a button for help instead of just a keyboard shortcut? Ideally, folks should be able to run the GUI without reading too many instructions
6) Wasn't clear to me what "snap to center" does
7) If I try to "jump to a location" by editing the RAS coordinates, it does not work when I put a crazy location like 150. I guess it would be more intuitive if it had up/down arrows for each coordinate and these don't go past a certain number

alexrockhill:
> Okay finally got this to work. Thanks @larsoner !
> 
> image
> 
> Disclaimer, I have never done this myself. So, this is mostly a naive user perspective.
> 
> 1. Why is the CT/MRI image inside two rectangles?

That's because the background is non-zero and the padding in `matplotlib` is zeros (black). It doesn't seem to distracting to me (you're zoomed in very close most of the time). Is it an issue to you?

> 2. I can't understand the bottom right plot. When I click on a marker, is it supposed to center and mark the corresponding channel in bottom right? That doesn't seem to be happening

The bottom-right plot is view only (clicking in 3D doesn't make sense, you are clicking based on what is in the other three plots). Clicking and scrolling on the plot only moves the camera.

> 3. Is there a way to turn off the overlay of the CT image over the MRI? Or does it only show CT?

The screenshot you sent is just of the CT. You can press 'b' to toggle the view of the brain. I should add back the help button though come to think of it.

> 4. I tried changing the color of one channel. Didn't seem to work ...

To change the "device" of a contact, you have to change the color and then mark the contact. I guess you could change the group of unmarked contacts as well.

> 5. Can there be a button for help instead of just a keyboard shortcut? Ideally, folks should be able to run the GUI without reading too many instructions

Yeah, I should add that back.

> 6. Wasn't clear to me what "snap to center" does

It snaps to the center of an area of high-intensity. The technical aspects are are that it looks for a local maxima and then includes all monotonically decreasing voxels from that voxel above 50% of the maximal value but from a typical user perspective the first part is all you really need to know.

> 7. If I try to "jump to a location" by editing the RAS coordinates, it does not work when I put a crazy location like 150. I guess it would be more intuitive if it had up/down arrows for each coordinate and these don't go past a certain number

If you put in something outside the range, it just resets to what it was before. That follows what Freeview does and I think it's not too hard to figure out with a few entries. It should be user-proof but I don't necessarily think you need warnings and errors for this when you can follow the existing format with small modifications to figure it out pretty easily.

jasmainak:
> The bottom-right plot is view only (clicking in 3D doesn't make sense, you are clicking based on what is in the other three plots, the 3D one is mostly for identifying devices). Clicking and scrolling on the plot only moves the camera.

Sorry, I should have said when I click on the marker name on the right panel, not on the plot. I thought it would somehow highlight the marker in the bottom right corner but it does nothing

alexrockhill:
It is centered as the focal point in the view but for forward compatibility with an automated algorithm (which would also have a current selected contact) it doesn't have an indicator for the latest contact marked, you just have to go off what changed in the plot. This is because two markers would be confusing in the future when the current auto-detected contact needs to be indicated.

larsoner:
> That's because the background is non-zero and the padding in matplotlib is zeros (black). It doesn't seem to distracting to me (you're zoomed in very close most of the time). Is it an issue to you?

Can you systematically use/choose the border constant value as the median of the first/last i/j/k planes `np.median(np.concatenate(img[[0, -1]].ravel(), img[:, [0, -1]].ravel(), img[:, :, [0, -1].ravel())` rather than just using zero? Seems like that would be robust and look nicer.

> The bottom-right plot is view only (clicking in 3D doesn't make sense, you are clicking based on what is in the other three plots, the 3D one is mostly for identifying devices). Clicking and scrolling on the plot only moves the camera.

Maybe a title for this widget "3D View" would make it clearer that this is just a *view* and not something that can be interacted with via clicks.

> The screenshot you sent is just of the CT. You can press 'b' to toggle the view of the brain. I should add back the help button though come to think of it.

In general in UI design, anything that can be done with a keyboard shortcut should have a corresponding mouse-based control (e.g., checkbox, radio button, drop-down, etc.) when possible/reasonable. In this case I think a "Brain" checkbox or something might be good.

larsoner:
... also, aesthetically if the background of all the MRI images is black, the 3D image should also have a black background, unless it messes up the visualization of electrode colors.

alexrockhill:
> Can you systematically use/choose the border constant value as the median of the first/last i/j/k planes `np.median(np.concatenate(img[[0, -1]].ravel(), img[:, [0, -1]].ravel(), img[:, :, [0, -1].ravel())` rather than just using zero? Seems like that would be robust and look nicer.

-1, I think this is solving a problem that doesn't exist and it's nice to know for me the boundaries of the CT so that you can see which parts got cutoff as sometimes they do.

> Maybe a title for this widget "3D View" would make it clearer that this is just a _view_ and not something that can be interacted with via clicks.

-1 I don't think that title really makes that clearer and I think directions for that would be better placed in the tutorial. I do want to hold the users hand, but if you don't read in the instructions, I don't think it has to have more instructions in the GUI, ideally there would be as little text as possible.

> In general in UI design, anything that can be done with a keyboard shortcut should have a corresponding mouse-based control (e.g., checkbox, radio button, drop-down, etc.) when possible/reasonable. In this case I think a "Brain" checkbox or something might be good.

Sure that sounds helpful.

larsoner:
> -1, I think this is solving a problem that doesn't exist

Hmm... I don't like the way it looks now, and I think @jasmainak does not either, so I guess we both at least see it as an aesthetic issue.

> it's nice to know for me the boundaries of the CT so that you can see which parts got cutoff as sometimes they do.

The way it's coded now, sometimes you'll have that information and other times you won't, depending on the normalization of the CT image. To me it's better to systematically fix the `cval`/border bug, then actually add a bounding box rectangle using a color of your choosing (red?) than rely on the `cval` to do this for you

larsoner:
(But I think it's okay to fix this `cval` business in a follow-up PR, it's easy enough and should clearly show an isolated improvement / we can discuss once it's implemented if it's actually better)

alexrockhill:
> > -1, I think this is solving a problem that doesn't exist
> 
> Hmm... I don't like the way it looks now, and I think @jasmainak does not either, so I guess we both at least see it as an aesthetic issue.
> 
> > it's nice to know for me the boundaries of the CT so that you can see which parts got cutoff as sometimes they do.
> 
> The way it's coded now, sometimes you'll have that information and other times you won't, depending on the normalization of the CT image. To me it's better to systematically fix the `cval`/border bug, then actually add a bounding box rectangle using a color of your choosing (red?) than rely on the `cval` to do this for you

I'm fine with a bounding box, that seems like a reasonable solution, the point was just that where the image ends is important information. The CT isn't normalized though so it will usually be some non-zero low number.

alexrockhill:
Ok I added buttons for help and show/hide brain. You still can only use keys to scroll slices but I'm not sure there's an easy place to put those six buttons but you can always click instead of scrolling. Also zoom is only scroll button or a keypress.

I didn't address the bounding box but that would be nice to do in a followup.

jasmainak:
Sorry I got lost with the bounding box discussion but looking at the image it seemed to me that there was some affine registration issue but I might be wrong. I have seen those kinds of images when you do some registration and the corners become black because there are no pixels to copy from.

It might be worth taking inspiration from MMVT. Maybe you were also a contributor @alexrockhill ? ;-)

image

I like how in the sliced view the CT appears in a different color and makes it really easy to spot things. But I don't know if it's a standard way to do this ...

alexrockhill:
> Sorry I got lost with the bounding box discussion but looking at the image it seemed to me that there was some affine registration issue but I might be wrong. I have seen those kinds of images when you do some registration and the corners become black because there are no pixels to copy from.
> 
> It might be worth taking inspiration from MMVT. Maybe you were also a contributor @alexrockhill ? ;-)
> 
> image
> 
> I like how in the sliced view the CT appears in a different color and makes it really easy to spot things. But I don't know if it's a standard way to do this ...

This is how it was previously in `img_pipe` but there were issues with this method because of setting the threshold. It's better to see the un-thresholded image to see the extent of the contact as clearly as possible. This is because the intensities in the CT vary by location and how directly the pixel that is most directly on the contact aligns with the contact.

Looking more closely, the black rectangle immediately surrounding the skull and the gray rectangle surrounding that are actually part of the underlying image. I really would prefer not to mess with this and the background because you wouldn't want to obscure something that might help you understand what's going on in the CT and it's bounding. The outermost black background is what happens when you zoom out of a matplotlib image so that one could be fixed but again I would strongly prefer to leave it as is.

alexrockhill:
Looks like the failure is unrelated. Maybe we can get this moving and make more improvements later

larsoner:
Agreed, okay for merge and incremental PRs after this @jasmainak ?

alexrockhill:
The plan is to do a JOSS paper after this merges so there will be a review there as well

agramfort (approved):
+1 for MRG if CIs are all green.

alexrockhill:
I'm redoing Liberty's ECoG grids with it (I don't have any ECoG that I'm doing for my own research) and besides it not being automated, it works really well

jasmainak:
yep +1 for MRG and incremental PRs. Thanks @alexrockhill for the massive effort !

alexrockhill:
@larsoner did you mean to remove the directional labels (eg superior, inferior)? I didn't see them when I just used it.

larsoner:
> @larsoner did you mean to remove the directional labels (eg superior, inferior)? I didn't see them when I just used it.

No, I don't think I meant to do it -- sorry if I did so in some commit. Good for a next PR :)

larsoner:
Thanks @alexrockhill !!!

PR #9587, opened 2021-07-19
FIX variables not defined before assignment in BCD solver

PABannier:
When running BCD solver, if the solver does not converge (fast enough), gap > tol. 

Yet, gap is only defined in the for scope and not in the else scope. This causes an UnboundLocalError.
Similar issue with X when carrying out debiasing.

I did a quick fix by defining the two variables in the function body. Tell me if the fix could be enhanced.

### DIFF
--- a/mne/inverse_sparse/tests/test_mxne_optim.py
+++ b/mne/inverse_sparse/tests/test_mxne_optim.py
@@ -116,6 +116,33 @@ def test_l21_mxne():
     assert_allclose(X_hat_bcd, X_hat_prox, rtol=1e-2)
 
 
+@pytest.mark.slowtest
+def test_non_convergence():
+    """Test non-convergence of MxNE solver to catch bugs."""
+    n, p, t, alpha = 30, 40, 20, 1.
+    rng = np.random.RandomState(0)
+    G = rng.randn(n, p)
+    G /= np.std(G, axis=0)[None, :]
+    X = np.zeros((p, t))
+    X[0] = 3
+    X[4] = -2
+    M = np.dot(G, X)
+
+    # Impossible to converge with only 3 iterations and tol 1e-12
+    # In case of non-convegence, we test that no error is returned.
+    args = (M, G, alpha, 3, 1e-12)
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(*args, active_set_size=None, debias=True,
+                          solver='prox')
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(*args, active_set_size=None, debias=True,
+                          solver='cd', return_gap=True)
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(M, G, alpha, maxit=1000, tol=1e-8,
+                          active_set_size=None, debias=True, solver='bcd',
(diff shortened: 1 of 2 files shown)


PABannier:
@agramfort just written some tests. Actually I don't know exactly what to test since we only expect the code not to crash.

What I've done: 
- Simulate a problem and made it such that solver cannot converge (tol @ 1e-12 and n_iter @ 3).
- Check that the test does not throw any errors.
WDYT?

larsoner:
> Actually I don't know exactly what to test since we only expect the code not to crash... Check that the test does not throw any errors.

Agreed, just having a smoke test that errors out on `main` but passes on this PR is good enough I think.

mne/inverse_sparse/tests/test_mxne_optim.py
```diff
@@ -116,6 +116,33 @@ def test_l21_mxne():
...
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(*args, active_set_size=None, debias=True,
+                          solver='prox')
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(*args, active_set_size=None, debias=True,
+                          solver='cd', return_gap=True)
+    with pytest.warns(None):  # CD
+        mixed_norm_solver(M, G, alpha, maxit=1000, tol=1e-8,
+                          active_set_size=None, debias=True, solver='bcd',
+                          return_gap=True)
```
larsoner, line 143: Do we really need to test the non-crash under three circumstances? It seems like we would only need three of these if there were three branches in the code that need to be hit, and based on the changes it doesn't look like there are.

FYI another option for testing if you want is to do:
```
with catch_logging() as log:
    mixed_norm_solver(..., verbose=True)
log = log.getvalue()
assert 'Convergence reached' not in log
```

PABannier: Agreed, it is an overkill. Thanks for the tips! I guess it is more suited for what we want and for readable. I'll change that.

PABannier: @larsoner where can I find catch_logging()? Is it from Pytest, logging library in Python?

larsoner:
```
$ git grep catch_logging
mne/beamformer/tests/test_dics.py:from mne.utils import object_diff, requires_h5py, catch_logging
mne/beamformer/tests/test_dics.py:    with catch_logging() as log:
mne/beamformer/tests/test_dics.py:    with catch_logging() as log:
mne/beamformer/tests/test_lcmv.py:from mne.utils import object_diff, requires_h5py, catch_logging
mne/beamformer/tests/test_lcmv.py:        with catch_logging() as log:
mne/beamformer/tests/test_lcmv.py:        with catch_logging() as log:
...
```

larsoner: (`git grep` is your friend :) )

larsoner (approved):
LGTM feel free to merge if you're happy @agramfort

agramfort (approved):
Thx @PABannier

PR #9589, opened 2021-07-19
change documentation about PTP rejection for diff ch types

### LINKED ISSUE #9551: Document that we allow for rejection based on EMG PTP
Request from the forum:
https://mne.discourse.group/t/reject-epochs-based-on-peak-to-peak-signal-amplitude-ptp-of-emg

GanshengT:
#### Reference issue
Example: Fixes #9551.

Document specification about PTP rejection based on different types of channels

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -2295,7 +2295,7 @@ reject_common = """
     respective epoch will be dropped.
 
     The dictionary keys correspond to the different channel types; valid
-    keys include ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, and ``'ecg'``.
+    **keys** can be any channel type present in the object.
 
     Example::
 
@@ -2326,8 +2326,7 @@ reject : dict | str | None
 """
 flat_common = """
     Reject epochs based on **minimum** peak-to-peak signal amplitude (PTP).
-    Valid **keys** can be any channel type present in the object. Examples 
-    include ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, ``'ecg'``. The 
+    Valid **keys** can be any channel type present in the object. The
     **values** are floats that set the minimum acceptable PTP. If the PTP
     is smaller than this threshold, the epoch will be dropped. If ``None``
     then no rejection is performed based on flatness of the signal."""
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -2283,7 +2283,7 @@ reject_common = """
     respective epoch will be dropped.
 
     The dictionary keys correspond to the different channel types; valid
-    keys are: ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, and ``'ecg'``.
+    keys include ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, and ``'ecg'``.
 
     Example::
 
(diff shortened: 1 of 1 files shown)


mne/utils/docs.py
```diff
@@ -2314,8 +2314,9 @@
 """
 flat_common = """
     Reject epochs based on **minimum** peak-to-peak signal amplitude (PTP).
-    Valid **keys** are ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, ``'ecg'``.
-    The **values** are floats that set the minimum acceptable PTP. If the PTP
+    Valid **keys** can be any channel type that is in ``self.info``. Examples
```
drammock, line 2317: Can you think of another way to say this that doesn't make reference to `self`?  Many end-users will not be familiar with what `self` means.  Maybe 
```
...any channel type present in the object.
```
?

drammock (approved):
other than one suggestion below, this LGTM. thanks!

mscheltienne:
One comment, I think you only changed the documentation for the flatness parameter and not also for the ptp parameter under `reject_common`.

GanshengT:
> One comment, I think you only changed the documentation for the flatness parameter (PTP minimum) and not also for the more common PTP maximum parameter under `reject_common`.

In the 'reject_common', it is stated The dictionary keys correspond to the different channel types; valid keys **include** (previously being 'are') grad, mag, etc. I think this statement is rather clear and does not cause confusion. As regards drammoc's comment, do you have a better idea?

mscheltienne:
@GanshengT Indeed, my bad, didn't see that the first modification valid/include was on reject_common!

drammock:
> As regards drammoc's comment, do you have a better idea?

I provided an idea in the comment itself.  If you can't think of a more elegant way, then use what I suggested. But it shouldn't refer to `self` because the meaning of `self` will likely not be clear to non-programmers.

hoechenberger:
Hello, what's the status here? :)

hoechenberger:
Specifically, I'm confused as I don't think `emg` was added here – wasn't this the premise of this PR..?

mne/utils/docs.py
```diff
@@ -2295,7 +2295,7 @@
     respective epoch will be dropped.
 
     The dictionary keys correspond to the different channel types; valid
-    keys are: ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, and ``'ecg'``.
+    keys include ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, and ``'ecg'``.
```
drammock, line 2298: I think the phrasing in the other docstring is better, let's standardize them:
```suggestion
    **keys** can be any channel type present in the object.
```

mne/utils/docs.py
```diff
@@ -2326,8 +2326,9 @@
 """
 flat_common = """
     Reject epochs based on **minimum** peak-to-peak signal amplitude (PTP).
-    Valid **keys** are ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, ``'ecg'``.
-    The **values** are floats that set the minimum acceptable PTP. If the PTP
+    Valid **keys** can be any channel type present in the object. Examples 
+    include ``'grad'``, ``'mag'``, ``'eeg'``, ``'eog'``, ``'ecg'``. The
```
drammock, line 2330:
```suggestion
    Valid **keys** can be any channel type present in the object. The
```

drammock (commented):
I've made a suggestion to just remove example channel names altogether, and make the docstrings for `reject` and `flat` more similar (since they work the same way).  I think my proposed changes will eliminate the trailing whitespace errors too (@GanshengT our continuous integration servers test for code style, including trailing whitespace, so in future contributions it would be good to use an editor that highlights or auto-removes trailing whitespace, or better yet run our style tests locally and fix violations before pushing commits to a PR).

mne/utils/docs.py
hoechenberger: Shouldn't it be plural ("channel type**S**")? Asking for a friend whose first language isn't American 😅

drammock: The quantifier "any" takes a singular noun. (consider: "anyone", not "anyones" 🙂)

mscheltienne:
@drammock The premise of this PR was EMG channels from a post on the MNE forum for a user which tried to reject based on EMG channels while not picking them. Thus the broader originial issue to adress is that you need to have the channels used for rejection inside the picks, i.e. inside the Epoch object.
Changes looks good and clear, no need to list examples.

drammock (approved):

agramfort (approved):

hoechenberger:
Thank you, @GanshengT!

PR #9593, opened 2021-07-19
MRG, ENH: Use discrete method

larsoner:
@alexrockhill I also refactored the `set_printoptions` stuff to improve `Transform.__repr__` itself, but let me know if you had that context manager in there for another reason.

Todo:
- [x] Make `marching_cubes` private

Also the `20_seeg` had an incorrect coordinate frame slip in through `plot_alignment`:

image

Now it looks more reasonable:

image

### DIFF
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -31,7 +31,7 @@ from .transforms import (transform_surface_to, _pol_to_cart, _cart_to_sph,
 from .utils import (logger, verbose, get_subjects_dir, warn, _check_fname,
                     _check_option, _ensure_int, _TempDir, run_subprocess,
                     _check_freesurfer_home, _hashable_ndarray, fill_doc,
-                    _validate_type, _require_version)
+                    _validate_type, _require_version, _is_numeric)
 
 
 ###############################################################################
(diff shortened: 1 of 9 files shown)


alexrockhill:
Oh wow, yeah that looks wrong like it is now

alexrockhill:
Maybe `marching_cubes` should be private as it is getting a bit complicated and not used in any tutorial anymore

mne/surface.py
```diff
@@ -1689,38 +1697,80 @@ def marching_cubes(image, level, smooth=0):
...
+    out.Update()
 
     # get verts and triangles
-    verts = numpy_support.vtk_to_numpy(polydata.GetPoints().GetData())
-    triangles = numpy_support.vtk_to_numpy(
-        polydata.GetPolys().GetConnectivityArray()).reshape(-1, 3)
-    verts = np.flip(verts, axis=1)
-    triangles = np.flip(triangles, axis=1)
-    return verts, triangles
+    polydatas = list()
```
alexrockhill, line 1745: data is already plural, something else datas is in `mne.viz.plot_alignment` too

mne/surface.py
```diff
@@ -1667,19 +1667,27 @@ def marching_cubes(image, level, smooth=0):
...
+        algorithm suitable for labeled volumes (e.g., ``'aseg.mgz'``).
     %(smooth)s
 
     Returns
     -------
-    verts : ndarray
-        The spatial coordinates for unique mesh vertices.
-    triangles : ndarray
-        The locations of connections between ``verts`` to form faces.
+    out : list | tuple
```
alexrockhill, line 1678: This could be called polydata to be more informative

mne/surface.py
```diff
@@ -1667,19 +1667,27 @@ def marching_cubes(image, level, smooth=0):
     ----------
     image : ndarray
         The image to compute marching cubes with.
-    level : float
+    level : float | int | list
```
alexrockhill, line 1670: It might simplify things to require a list and then just return a list of tuples, especially if it's a private function that would make sense

alexrockhill:
There are a few nitpicky suggestions to simplify but other than that looks good to go to me

mne/surface.py
larsoner: Yes but the object is `PolyData`, we sometimes just keep adding `s`'s to indicate more of something, e.g. `poss` as "positions-es"

mne/surface.py
larsoner: It's not polydata, that's a VTK concept. The output is really either a 2-tuple of ndarray or a list of 2-tuples of ndarray

larsoner:
Okay eliminated `polydata` altogether, made `_marching_cubes` private, and now always return a list of `(rr, tris)` tuple pairs. Okay for you @alexrockhill ?

alexrockhill:
> Okay eliminated `polydata` altogether, made `_marching_cubes` private, and now always return a list of `(rr, tris)` tuple pairs. Okay for you @alexrockhill ?

Nice, looks a bit cleaner to me :) I think it's good to go.

Issue #4926, opened 2018-02-07
Move test data out of MNE package


cbrnr:
I think it would be nice if we moved all test data out of `mne.io` and integrated them into `mne.datasets.testing` instead. Currently, the following subfolders of `mne/io` contain test files:

- brainvision/tests/data (1.1 MB)
- bti/tests/data (3.1 MB)
- edf/tests/data (3.1 MB)
- egi/tests/data (0.3 MB)
- kit/tests/data (5.1 MB)
- nicolet/tests/data (0.03 MB)
- tests/data (64.2 MB)

This would reduce the package size by 76.9 MB.

larsoner:
These do not get included in the release, see 5MB binary:

https://pypi.python.org/pypi/mne/0.15.2

For the repo to save the size we'd need to purge / rewrite history, probably not worth it.

larsoner:
(at least to save the size when doing `git clone`)

cbrnr:
OK that's fine then. It's still a bit inconsistent that some test files are in the package (e.g. EDF/BDF) whereas others (e.g. GDF) are in the testing datasets module.

cbrnr:
And in the repo, why do we need to rewrite history? Why not just `git rm`?

larsoner:
Agreed it's inconsistent it is historical cruft.

If you `git rm` the files "go away" but the (restore-able) record of them stays in the `git` history unless you purge (or do a sufficiently shallow clone) so there is no gain.

cbrnr:
> Agreed it's inconsistent it is historical cruft.

So OK to leave this open in case someone wants to work on it?

You're right, the size of the repo doesn't change, but this is not such a big issue. I think consistency is more important.

larsoner:
Since we seem to be stuck with the existing size in the repo, the current situation does have one advantage: even without access to the testing data (which is now getting a bit large) more core functionality can be tested / played with than if you remove those files.

agramfort:
we should be migrating to :

https://web.gin.g-node.org/NeuralEnsemble/ephy_testing_data

it's a git annex repo so we can selectively pull certiain from the this
potentially big repo

cbrnr:
Yeah, trading consistency for flexibility is not a good option. The testing data is not only getting a bit large (approaching 1GB I think), but the download is also extremely slow. I just aborted because I didn't want to wait for two hours. If this repo is a better option than what we have now we should look into it.

cbrnr:
[...]

PR #9596, opened 2021-07-20
[ENH, VIZ]: Refactor _figure.py-classes to prepare alternative backend

marsipu:
## Abstract
This PR is supposed to refactor all methods and attributes from the Classes associated with the current implementation of 2D-Data-Visualization (mainly MNEBrowserFigure) into a new parent class called MNEDataBrowser. This should make the addition of a new implementation based on pyqtgraph (currently developed [here](https://github.com/marsipu/gsoc2021) more feasible. It will contain all visualization-independent steps of the preparation of data. Furthermore the aim is to let both backends use similar data-structures to store gui-elements like Annotations, Event-Lines, etc. ... . And some backend-independent GUI-Methods could be outlined as abstractmethods.

## ToDo
- [X] Create MNEDataBrowser
- [X] Move MNEBrowserFigure-Initialization to MNEDataBrowser
- [X] Extract GUI-independent methods from MNEBrowserFigure
- [X] Refactored DataIO
- [X] use/set/get for browser-backend
- [x] Refactor Tests


larsoner:
I would make this first PR just the first three points above, and stop before "Adapt PyQtGraph". I think refactoring the existing classes into a data class and matplotilb subclass is a good first PR.

There is a separate also difficult question about how we want to refactor tests. Ideally some or all of the existing raw and epochs data browsing tests would work on the new backend. So maybe an additional bullet point that could be part of this first PR would be to make it so that the matplotlib-specific stuff in the *tests* is also separated out in some way.

In 3D we do this by having a `_renderer` abstraction, and things like `_renderer._close_all()` exist that close all Mayavi or PyVista figures, for example. And then there is a pytest fixture that yields both a Mayavi and Pyvista backend that can be used in the tests. I'm not sure if we want to copy this verbatim but it is something to think about. See for example here:

- `renderer` fixture itself, which parametrizes over `backend`: https://github.com/mne-tools/mne-python/blob/main/mne/conftest.py#L290-L294
- `renderer` fixture use: https://github.com/mne-tools/mne-python/blob/main/mne/viz/tests/test_3d.py#L168-L201

I think it seems reasonable to add a `browse_backend` fixture or something that in this first PR just (trivially) parametrizes over `['matplotlib']`, and then in the next PR that actually adds pyqtgraph support it adds `'pyqtgraph'` to its list of params.

larsoner:
... don't worry about this too much for now (unless you want to!), as we can discuss this more in depth on Thursday. In the meantime you can keep working toward the first two points above in this PR for the next couple of days probably!

marsipu:
@larsoner Thank you for your suggestions. As you say, an abstraction would be helpful to avoid having to switch between cases for every part in the tests which is matplotlib-specific (accessing axes, simulate clicking, pyplot-functions). Maybe I can have methods for those cases in `BrowserBase` (or what we plan to call it).

I saw that for PyQt-testing there are fixtures from `pytest-qt` which help with creating the application-context and widget-interaction. Would you combine a possible `browse_backend`-fixture with pytest-qt?

As you suggest I will continue with the abstraction of `BrowserBase` first, but I hope I can get to the tests too soon.

P.S. Please don't worry about the preliminary integration of pyqtgraph in _get_browser. I added this just to try some things out already with the PyQtGraph-Prototype, because that helps me think about which parts of BaseBrowser can be used by both backends. I will remove this part again for the final PR.

cbrnr:
> I saw that for PyQt-testing there are fixtures from pytest-qt which help with creating the application-context and widget-interaction. Would you combine a possible browse_backend-fixture with pytest-qt?

Yes, we'll probably need this to simulate mouse clicks (which we currently already do for our MPL-based window with [`_fake_click()`](https://github.com/mne-tools/mne-python/blob/308a0abf7c8b8b14c9afc6db2e3ed5b57787be31/mne/viz/utils.py#L694)).

marsipu:
I have a question about the browser(backend)-fixture:
As I understand the `renderer`-fixture yields the entire renderer-object depending on the current backend-parameter.
In most of the tests in `test_raw.py` the figure-object is got from `raw.plot()`. In some cases there are parameters added to raw.plot() if required for the test. If I would want to have one `browser`-fixture, I would need a way to pass arguments to `raw.plot()` inside the fixture (this is probably not the way fixtures are supposed to be used) or I would need a fixture for every use-case (which is probably also bad). Or is there another way?
If not, maybe a `browser_backend`-fixture just parametrizes the backend-names and returns the string to be then passed into `raw.plot()`. For that I would have to change the API of `raw.plot()` already by adding the `browser_backend`-parameter.
WDYT?

larsoner:
> If not, maybe a browser_backend-fixture just parametrizes the backend-names and returns the string to be then passed into raw.plot(). For that I would have to change the API of raw.plot() already by adding the browser_backend-parameter.

This is close to what I had in mind. I think eventually we'll probably want a `mne.viz.set_browse_backend` / `mne.viz.use_browse_backend` / `mne.viz.get_browse_backend` like we have [`mne.viz.set_3d_backend`](https://mne.tools/stable/generated/mne.viz.set_3d_backend.html) et al. So I think the browser_backend fixture can be structured roughly like:
```
@pytest.fixture
@pytest.parametrize('backend', ('matplotlib',)
def browser_backend(backend):
    with mne.viz.use_browse_backend(backend):
        yield
```
But we also know that each test now needs an instance of some sort of click-keypress class, so once that refactoring is done this could become:
```
...
    if backend == 'matplotlib':
        tester = _MatplotlibFakeInteractor()
    # else:
    #    assert backend == 'pyqtgraph'
    #    tester = _PyQtGraphFakeInteractor()
    with mne.viz.use_browse_backend(backend):
        yield tester
```
and then inside your tests you can have
```
def plot_raw(browser_backend):
    ...
    fig = raw.plot(...)
    browser_backend._fake_click(fig, ...)
```
and this `_fake_click` is a method of `_MatplotlibFakeInteractor` and "does the right thing" for matplotlib, etc.

larsoner:
Alternatively, if you want to skip the `browse_backend` stuff for now, you could have the FakeInteractor have a `.backend` property that is the string, and you'd call in every test:
```
raw.plot(..., backend=backend)
```
but this set/use/get pattern is pretty nice.

marsipu:
@larsoner Thank you, I see. So you would rather have use/set/get than an additional parameter? Or both?

marsipu:
I also thought I maybe could put the fake-interactions as abstract-methods into BrowserBase and have the backend-specific code inside the backend-classes to have it all together. Then interactions inside the tests could be done with methods from the figure/pyqt-object (and it would be accessible by users)
WDYT?

mne/viz/_browser.py
```diff
@@ -0,0 +1,516 @@
...
+    logger.info(f'Using {backend_name} as 2d backend.\n')
+
+
+def _get_browser(inst, **kwargs):
+    """Instantiate a new MNE browse-style figure."""
+    from .utils import _get_figsize_from_config
+    figsize = kwargs.pop('figsize', _get_figsize_from_config())
+
+    _get_browser_backend()
+    browser = backend._init_browser(inst, figsize, **kwargs)
```
drammock, line 374: the way this is done now (passed to the same function as the rest of `kwargs`), it's no longer clear why figsize gets `pop`ped.  Perhaps 

```
default_kwargs = dict(figsize=_get_figsize_from_config())
default_kwargs.update(kwargs)
...
_init_browser(inst, **default_kwargs)
```
WDYT?

mne/viz/tests/test_raw.py
```diff
@@ -62,26 +36,27 @@ def _annotation_helper(raw, events=False):
...
+    fig.press_key('a')  # annotation mode
+    assert fig.get_n_windows() == 2
     # +3 from the scale bars
     n_scale = 3
     assert len(data_ax.texts) == n_anns + n_events + n_scale
     # modify description to create label "BAD test"
     ann_fig = fig.mne.fig_annotation
     for key in ['backspace'] + list(' test;'):  # semicolon is ignored
-        ann_fig.canvas.key_press_event(key)
+        fig.press_key(key, target=ann_fig)
```
drammock, line 47: I don't love this (would have expected `ann_fig.press_key(key)`).  I guess it's a consequence of the fact that pyqtgraph backend will handle annotations through its toolbar and hence won't have a separate annotation window?  Would it work to have `ann_fig` still exist in pyqtgraph case, but have it be a non-displaying object that passes interactions back to the main window?  Not sure which way is easier / better; WDYT?

mne/viz/_figure.py
```diff
@@ -2295,6 +2074,88 @@ def _toggle_vline(self, visible):
         self.mne.vline_visible = visible
         self.canvas.draw_idle()
 
+    def close_all(self):
+        plt.close('all')
```
drammock, line 2078: no.  this will close ALL FIGURES, not just the main data browser and its associated child figs.  See how the `_close` method currently works (all you need to do is close the main data browser window and all the child figs will automatically close with it).  Also not clear why this needs to be public.

mne/viz/_figure.py
```diff
@@ -2295,6 +2074,88 @@ def _toggle_vline(self, visible):
...
+    def close_event(self, target=None):
+        """Force calling of the MPL figure close event."""
+        target = target or self
+        try:
+            target.canvas.close_event()
+        except ValueError:  # old mpl with Qt
+            pass  # pragma: no cover
+
+    def get_n_windows(self):
+        return len(plt.get_fignums())
```
drammock, line 2091: no. This will include any other MPL figure windows that the user happens to have open.

mne/viz/_figure.py
```diff
@@ -2295,6 +2074,88 @@ def _toggle_vline(self, visible):
...
+
+    def press_key(self, key, target=None):
```
drammock, line 2093: (nitpick) for consistency I'd call this `fake_keypress` or `fake_key_press`

mne/viz/_figure.py
```diff
@@ -2295,6 +2074,88 @@ def _toggle_vline(self, visible):
...
+                           button=button)
+        elif kind == 'release':
+            func = partial(target.canvas.button_release_event, x=x, y=y,
+                           button=button)
+        elif kind == 'motion':
+            func = partial(target.canvas.motion_notify_event, x=x, y=y)
+        else:
+            raise RuntimeError(f'{kind} is no suitabl kind for _fake_click')
+
+        func(guiEvent=None)
```
drammock, line 2120: would it work to have the figure method be a thin wrapper to a utility function like the existing `_fake_click` function?  I think we use fake clicks in testing for other plots besides just the data browser... might be more DRY

mne/viz/_figure.py
```diff
@@ -2295,6 +2074,88 @@ def _toggle_vline(self, visible):
...
+
+    def click_ch_name(self, ch_index, button):
+        self.canvas.draw()
+        text = self.mne.ax_main.get_yticklabels()[ch_index]
+        bbox = text.get_window_extent()
+        x = bbox.intervalx.mean()
+        y = bbox.intervaly.mean()
+        self._fake_click((x, y), xform='pix', button=button)
+
+    def test_childs(self, key, attr):
```
drammock, line 2130: there should not be a testing function here.  Tests belong in `test_*.py` files inside `*/tests/` folders

drammock (commented):
Big-picture comments:

A large portion of what is in `mne/viz/_browser.py` looks to be cut-and-pasted from `mne/viz/_figure.py`.  Cut-and-paste changes are really hard to review because we can't easily tell if you simply moved code, or moved and then changed something.  would it be possible to do one of these things?

1. rebase this PR and squash commits, so that there is 1 commit that does all the cut-and-paste moves (and nothing else).  That way we can exclude that commit when viewing the diff, so we can focus on substantive changes
2. instead of keeping MPL-specific stuff in `_figure.py` and moving backend-agnostic code into `_backend`, you could (should?) put the MPL-specific code in `_mpl_figure.py` and use `_figure.py` for the backend-agnostic stuff (and later, use `_pyqtgraph_figure.py` for the new backend.  That way, what is currently cut-paste would be staying where it is (and changes to it would be easier to see)...  the MPL-specific stuff should not be changing much in this PR anyway so having that stuff be the cut-paste part is preferable.

(of those I think I prefer option 2 the best, but could live with option 1) 

Also: lots of style errors: https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=14620&view=logs&j=3e7b9abc-cf32-5e9b-ccb3-9355b11c347c&t=9989323c-d330-5eb0-bad7-a6f163ea363a&l=14

after fixing those and before pushing, from the repo root you can run `make docstring` to run the docstring tests locally to make sure you got 'em all.

marsipu:
@drammock Thank you for the detailed feedback, I see I have a lot to work on.

> instead of keeping MPL-specific stuff in _figure.py and moving backend-agnostic code into _backend, you could (should?) put the MPL-specific code in _mpl_figure.py and use _figure.py for the backend-agnostic stuff (and later, use _pyqtgraph_figure.py for the new backend. That way, what is currently cut-paste would be staying where it is (and changes to it would be easier to see)... the MPL-specific stuff should not be changing much in this PR anyway so having that stuff be the cut-paste part is preferable.

I'll see to that I apply your proposed option 2.

marsipu:
I see why keeping BrowserBase in `_figure.py` is a huge advantage for reviewing. Just for the more distant future I think it might be clearer with `_browser.py` to avoid direct association to `matplotlib.figure.Figure`. But naming these files is something we could change way afterwards, right? So I will do as you propose and but everything new into `_figure.py` as proposed.

mne/viz/_figure.py
marsipu: Right, this doesn't need to be public. This method was just something I added where in the current tests `plt.close('all')` would be called. I can specify this more with `_close` as you suggest.

mne/viz/_figure.py
marsipu: Right, this shouldn't be public.

mne/viz/epochs.py
```diff
@@ -894,19 +894,24 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20, n_channels=20,
...
                   xlabel='Epoch number')
-    fig = _browse_figure(**params)
-    fig._update_picks()
 
-    # make channel selection dialog, if requested (doesn't work well in init)
-    if group_by in ('selection', 'position'):
-        fig._create_selection_fig()
+    fig = _get_browser(**params)
+
+    if get_browser_backend() == 'matplotlib':
```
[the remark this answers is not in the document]
marsipu: This block should wrap every method-call which is currently not implemented in the pyqtgraph-browser (that is why the selection-code is in there too because this is not implemented in pyqtgraph yet).

As you say some methods may be difficult or unnecessary to be transfered to the pyqtgraph-backend. Maybe a common and abstract `update`-method could make the generalization easier here?

mne/viz/_figure.py
marsipu: Yes thats right, I changed the signature a bit by adding defaults to target(former fig) and ax which results in reduced calls from the tests. But that can be done better by wrapping `_fake_click `from `mne.viz.utils`, thanks for the hint.

mne/viz/_figure.py
marsipu: Yes, the naming is misleading. This is too something I just refactored in the first step from the tests. Maybe I can leave it as WIP? Because I think especially this test can be challenging to generalize between matplotlib and pyqtgraph.

drammock: if there are a few tests that will only work for one backend or another, that's fine, but they should still live in a `test_*.py` file inside a `*/tests/` directory (and just not be parametrized to run under both backends).  Depending on how many such tests there are, they can either live in the existing `test_raw.py` or there could be separate `test_mpl_backend.py` and `test_pyqtgraph_backend.py` files to contain those tests.

mne/viz/epochs.py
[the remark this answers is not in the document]
drammock: > This block should wrap every method-call which is currently not implemented in the pyqtgraph-browser

why?  This PR is supposed to be a pure conceptual refactoring of existing matplotolib plotting, and is not supposed to implement the pyqtgraph browser yet, so there should be no need.

drammock:
> I see why keeping BrowserBase in `_figure.py` is a huge advantage for reviewing. Just for the more distant future I think it might be clearer with `_browser.py` to avoid direct association to `matplotlib.figure.Figure`.

I don't share that worry...  ordinary users won't ever look at `_figure.py` (only contributors), and presumably anyone wanting to look at it will be able to see that it's not matplotlib-specific.  I'm happy to be outvoted here if others disagree, but let's delay such renaming until the PR is otherwise (nearly) ready to be merged.

mne/viz/_browser.py
```diff
@@ -0,0 +1,516 @@
...
+    @abstractmethod
+    def test_childs(self, key, attr):
+        pass
+
+
+def _reload_backend(backend_name):
+    global backend
+    backend = importlib.import_module(name=_backend_name_map[backend_name],
+                                      package='mne.viz')
+    logger.info(f'Using {backend_name} as 2d backend.\n')
```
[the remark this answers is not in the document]
marsipu: Ok, good to know. I was taking `_get_renderer` in `renderer.py` as a blueprint and thought it was necessary.

mne/viz/_browser.py
```diff
@@ -0,0 +1,516 @@
...
+
+
+def _get_browser(inst, **kwargs):
+    """Instantiate a new MNE browse-style figure."""
```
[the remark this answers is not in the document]
marsipu: I understand, but in this case it could be interpreted as kind of a "get" because it calls the `_init_browser`-funtion of the selected backend. A second `_init_browser` would be maybe confusing?

mne/viz/_browser.py
marsipu: Yes, you are right, I also would like more if just the kwargs would be passed on to _init_browser

mne/viz/tests/test_raw.py
marsipu: Frankly applying additional structures from matplotlib onto pyqtgraph (and then probably also vice versa) seems to me a bit confusing and constraining regarding the further development of each backend. 
But if we want minimal change and we don't want to have if-cases for each backend in some of the tests, there probably has to be a dummy-counterpart or something similar. This might be more conservative on the tests if these cases are handled inside the respective Backend-Class (with target/fig as parameter and not as potential non-existing object).

What may be a solution to this specific problem could be putting this method and also e.g _fake_click into MNEFigure. 
But this would then make inheritance and having those methods as abstract methods in BrowserBase more difficult (wasn't able to solve that yet, so I first await for your response about this).
WDYT?

marsipu:
Update from today @drammock:
I implemented your feedback and tried to refactor the remaining tests in test_raw.py. I realized that I probably need to start now with the second PR for the integration of the pyqtgraph-backend because in many test-cases I am not sure how this will turn out for the pyqtgraph-backend in the end without having implemented it (since I don't have that much experience with pytest-qt yet).
Or I try to finish this PR with the outlook of probably having to change many tests again when I add the pyqtgraph-backend.
WDYT?

P.S. I know there is still a test failing in `test_plot_raw_traces`, I had a hard time finding reason because I thought I refactored everything without a change.

mne/viz/tests/test_raw.py
```diff
@@ -440,33 +423,32 @@ def test_plot_raw_traces(raw):
...
     fig = plot_raw(raw, events=events, event_color={-1: 'r', 998: 'b'})
-    plt.close('all')
+    browse_backend._close_all()
     for group_by, order in zip(['position', 'selection'],
                                [np.arange(len(raw.ch_names))[::-3],
                                 [1, 2, 4, 6]]):
         fig = raw.plot(group_by=group_by, order=order)
         x = fig.get_axes()[0].lines[1].get_xdata()[10]
         y = fig.get_axes()[0].lines[1].get_ydata()[10]
-        _fake_click(fig, data_ax, [x, y], xform='data')  # mark bad
```
marsipu, line 450: I think this line is the reason, why the refactored test fails. Is it intentional that the _fake_click is transformed to data-coordinates of data_ax instead of fig.ax_main from the fig instantiated three lines above?

mne/viz/tests/test_raw.py
drammock: > But if we want minimal change and we don't want to have if-cases for each backend in some of the tests, there probably has to be a dummy-counterpart or something similar.

This is certainly my preference (compared to using if-else logic for the different backends).  Having an explicit dummy method provides a place in the code to briefly explain *why* that dummy method is a dummy (e.g., a docstring that says "dummy method for X, because PyQtGraph handles X in a toolbar, not in a separate figure"). This makes understanding the various backends and how they differ much easier for future maintainers.

> What may be a solution to this specific problem could be putting this method and also e.g _fake_click into MNEFigure.

For this particular case it may be that we need to revisit how `_fake_click` and `_fake_keypress` work to allow them to interact with toolbar elements more easily. Or perhaps we will ultimately need higher-level methods that are part of each backend, like `_route_keypress_to_annotation_textbox(key)`, so that the tests simply call that method and the helper methods contain the logic to get the correct window / textbox in focus to receive the key.  But for now, *remember that this PR is supposed to split out matplotlib-specific code from backend-agnostic code*.  You don't need to make the code fully compatible with PyQtGraph yet; if you foresee troublesome areas it is enough to leave code comments saying, e.g., "this method won't translate well to PyQtGraph backend because annotations are handled in a toolbar".  It will be easier to tackle those incommensurabilities one at a time, once this refactoring is done and once the (relatively) straightforward, similar parts of the new backend have been implemented.

drammock:
> I realized that I probably need to start now with the second PR for the integration of the pyqtgraph-backend

I strongly advise to finish this one first.  I know it can be difficult to ignore what you know about PyQtGraph and resist changing things now that you know will need to be changed eventually, but there is a reason we keep recommending you to do it this way.

> Or I try to finish this PR with the outlook of probably having to change many tests again when I add the pyqtgraph-backend.

Yes!  It is 100% expected, part of the process, and okay to implement a conceptual split between backend-agnostic code and MPL-specific code, *knowing that some of it will have to change very soon*.  It is in most cases still easier to do it that way, even though it seems like more work.  And it is *definitely* less error-prone because (1) there is a defined halfway point where we get to ensure all the tests still pass, and (2) the separate steps are much easier to review, meaning human reviewers are more likely to catch mistakes.

mne/viz/tests/test_raw.py
marsipu: > Or perhaps we will ultimately need higher-level methods that are part of each backend, like _route_keypress_to_annotation_textbox(key)

If I understand correctly, we would then have a helper-method for each child-element we want to test the interaction with.
Wouldn't the current solution with passing the child-element as an argument be a bit cleaner?

Or we could create a second Base-Class like `FigureBase`, where we would but all abstract methods which would be used by the main browser and its children together (something like `_click_ch_name` for example would only be used by `MNEBrowseFigure` and all the other methods in `BrowserBase` would also be of no use or even interfering with for e.g. `MNEAnnotationFigure`.

drammock: > Wouldn't the current solution with passing the child-element as an argument be a bit cleaner?

maybe.  I'm not advocating any specific solution to this specific problem right now. The point I'm trying to make is that deciding among the many possible ways to adapt the tests to handle multiple backends will be *easier* if you *wait* to adapt the tests until *after* the conceptual separation of backend-agnostic and MPL-backend-specific code has been done.

mne/viz/_figure.py
```diff
@@ -2031,483 +323,197 @@ def _update_data(self):
...
+
+    return old_backend_name
+
+
+def _init_browser_backend():
+    global MNE_BROWSE_BACKEND
+
+    # check if MNE_BROWSE_BACKEND is not None and valid or get it from config
+    loaded_backend = MNE_BROWSE_BACKEND \
+        or get_config(key='MNE_BROWSE_BACKEND', default=None)
```
drammock, line 459: we usually avoid backslash-continuation lines and use parentheses instead

mne/viz/tests/test_raw.py
drammock: this was probably a mistake / cruft.

mne/viz/tests/test_raw.py
```diff
@@ -440,33 +428,34 @@ def test_plot_raw_traces(raw):
...
-        fig.canvas.key_press_event('down')  # change selection
-        _fake_click(fig, fig.get_axes()[2], [0.5, 0.5])  # change channels
+        # ToDo: Is this intentional?
+        #  (click in data-coordinates of data_ax from start of test)
+        fig._fake_click((x, y), ax=data_ax, xform='data')  # mark bad
```
drammock, line 440: I think then based on your other comment that this line should change

mne/viz/tests/test_raw.py
```diff
@@ -440,33 +428,34 @@ def test_plot_raw_traces(raw):
...
+        fig._fake_keypress('down')  # change selection
+        fig._fake_click((0.5, 0.5), ax=fig.get_axes()[2])  # change channels
```
drammock, line 442: I think here you can do `ax=fig.mne.ax_vscroll`?

mne/viz/tests/test_raw.py
```diff
@@ -476,65 +465,66 @@ def test_plot_raw_traces(raw):
...
         raw.plot(group_by='position', order=np.arange(8))
     for fig_num in plt.get_fignums():
-        fig = plt.figure(fig_num)
-        if hasattr(fig, 'radio'):  # Get access to selection fig.
+        sel_fig = plt.figure(fig_num)
+        if hasattr(sel_fig, 'radio'):  # Get access to selection fig.
             break
     for key in ['down', 'up', 'escape']:
-        fig.canvas.key_press_event(key)
+        fig._fake_keypress(key, fig=sel_fig)
```
drammock, line 472: you changed what is being tested here. it's now sending those fake keypresses to `fig` instead of `sel_fig`.  You should revert renaming `fig` to `sel_fig`.  I think also maybe we can get access to `sel_fig` through `fig.mne.fig_selection` instead of looping and looking for `radio` attribute?

drammock (commented):
here is a first-pass review just looking at the code. Tomorrow I'll pull down your changes and test interactively to make sure everything still works.

mne/viz/_figure.py
```diff
@@ -1,344 +1,75 @@
...
+        ica = kwargs['ica'] if 'ica' in kwargs else None
 
         # what kind of data are we dealing with?
-        if isinstance(ica, ICA):
-            self.mne.instance_type = 'ica'
-        elif isinstance(inst, BaseRaw):
-            self.mne.instance_type = 'raw'
-        elif isinstance(inst, BaseEpochs):
-            self.mne.instance_type = 'epochs'
+        if inst is not None:
```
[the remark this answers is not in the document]
marsipu: No you're right, as we discussed yesterday it probably makes no sense for users to start an empty browser-container.

mne/viz/tests/test_raw.py
marsipu: It should still be the same logic:
`sel_fig` is passed to the `_fake_keypress`-method so the `key_press_event` is called on `sel_fig`.
But the problem why I had to rename `fig` to `sel_fig` is that `_fake_keypress`-method only exists in `BrowserBase` which is inherited from `fig`(`MNEBrowseFigure`) and not `sel_fig`(`MNESelectionFigure`).
This was the cost of generalizing all backend-agnostic methods from the browser and interactivity-tests into one base-class, which can't be inherited by child figures as `MNESelectionFigure`.

drammock:
@marsipu I pushed one commit to split the big `test_plot_raw_traces` into 4 smaller tests, which made it easier to figure out what was going wrong.  I'm not 100% sure why it was happening, but changing the fake-click to be on the second trace (a grad channel) instead of the first one (IAS) was enough to fix it.

marsipu:
Thank you @drammock

mne/viz/tests/test_raw.py
drammock: ah, you're right, I was misreading the code

mne/viz/tests/test_raw.py
```diff
@@ -284,75 +266,78 @@ def test_plot_raw_ssp_interaction(raw):
...
     ssp_fig = fig.mne.fig_proj
+    # ToDo: These gui-elements might differ in pyqtgraph.
     t = ssp_fig.mne.proj_checkboxes.labels
     ax = ssp_fig.mne.proj_checkboxes.ax
     assert _proj_status(ax) == [True, True, True]
     # this should have no effect (proj 0 is already applied)
     assert t[0].get_text().endswith('(already applied)')
     pos = np.array(t[0].get_position()) + 0.01
-    _fake_click(ssp_fig, ssp_fig.get_axes()[0], pos, xform='data')
+    fig._fake_click(pos, ssp_fig, ssp_fig.mne.ax_main, xform='data')
```
drammock, line 279: only the main plot window has an `ax_main`.  It looks like in the SSP dialog figure we don't actually store a reference to the main axes, but a few lines above we've already tracked them down as `ax = ssp_fig.mne.proj_checkboxes.ax` 
```suggestion
    fig._fake_click(pos, ssp_fig, ax, xform='data')
```

drammock:
@cbrnr and @hoechenberger I leave it to one of you two to push the green button

marsipu:
@cbrnr @hoechenberger Is there still something I can improve?

cbrnr:
I didn't have time to review, I'll probably do it tomorrow. Some of the discussions have not been resolved, are these still open issues?

marsipu:
> Some of the discussions have not been resolved, are these still open issues?
Thanks for the hint, most of them are already outdated and resolved.

cbrnr:
I didn't know that we have `mne/conftest.py` to define fixtures. I believe that it only makes sense to define fixtures here that are used in several places. It looks like the ones you added are very specific to the browser viz test – would it make sense to move these fixtures closer to where they are needed (i.e. somewhere in `mne/viz/tests`)?

marsipu:
> 
> 
> I didn't know that we have `mne/conftest.py` to define fixtures. I believe that it only makes sense to define fixtures here that are used in several places. It looks like the ones you added are very specific to the browser viz test – would it make sense to move these fixtures closer to where they are needed (i.e. somewhere in `mne/viz/tests`)?

The reason I moved the fixtures for the data-import from mne.io.tests.data into `conftest.py` was that those were partly redundant in `test_raw.py`, `test_epochs.py` and `test_ica.py`. 
When searching across mne-python I see, that they could be used in several other tests too, as in `test_csp.py`, `test_montage.py`, `test_transformer.py`, `test_csd.py` [...].
Maybe this would be something to be resolved from #4926 in another PR?

But I can of course move the fixtures in a separate file in mne.viz.tests if you like that better, should I do that now?

cbrnr:
> But I can of course move the fixtures in a separate file in mne.viz.tests if you like that better, should I do that now?

I wasn't aware of these fixtures until now, so I'm probably not the best person to decide where they should go. However, I feel like it is not even straightforward to decide if test setup parts should go into a fixture or if they should be defined within the actual test. And if they go into a fixture they can either be defined in a module or more globally in a `conftest.py` file.

In short, so far I've written my tests to be entirely self-contained, e.g. generating test data directly inside the test function.

What is our recommended best practice for writing tests? I haven't found anything in our docs, and it seems like this is an important topic. On the one hand, fixtures can make tests run much faster, but on the other hand tests are harder to understand with all that fixture magic happening in the background (at least IMO).

This is not really relevant to this PR, so maybe we should discuss it somewhere else. But a quick statement from @drammock, @larsoner or @hoechenberger on how to proceed with the fixtures you added here would be good.

mne/viz/raw.py
```diff
@@ -347,10 +347,12 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
...
                   window_title=title)
 
-    fig = _browse_figure(**params)
+    fig = _get_browser(**params)
+
     fig._update_picks()
 
-    # make channel selection dialog, if requested (doesn't work well in init)
+    # make channel selection dialog,
+    # if requested (doesn't work well in init)
```
cbrnr, line 355: Why the newline here?

mne/viz/epochs.py
```diff
@@ -894,10 +894,13 @@ def plot_epochs(epochs, picks=None, scalings=None, n_epochs=20, n_channels=20,
...
                   xlabel='Epoch number')
-    fig = _browse_figure(**params)
+
+    fig = _get_browser(**params)
+
     fig._update_picks()
 
-    # make channel selection dialog, if requested (doesn't work well in init)
+    # make channel selection dialog,
+    # if requested (doesn't work well in init)
```
cbrnr, line 903: Why the newline here?

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
-     │                         group_by == 'selection' or 'position'
-     │
-     └ MNELineFigure           Interactive figure for non-scrollable data.
-                               Generated by:
-                               - raw.plot_psd()
-                               - evoked.plot()        TODO Not yet implemented
-                               - evoked.plot_white()  TODO Not yet implemented
-                               - evoked.plot_joint()  TODO Not yet implemented
-"""
+"""Base classes and functions for 2D-Browser-Backends."""
```
cbrnr, line 2:
```suggestion
"""Base classes and functions for 2D browser backends."""
```

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
+from .backends._utils import VALID_BROWSE_BACKENDS
+from .utils import _get_color_list, _setup_plot_projector
 
-# CONSTANTS (inches)
-ANNOTATION_FIG_PAD = 0.1
-ANNOTATION_FIG_MIN_H = 2.9  # fixed part, not including radio buttons/labels
-ANNOTATION_FIG_W = 5.0
-ANNOTATION_FIG_CHECKBOX_COLUMN_W = 0.5
+MNE_BROWSE_BACKEND = None
+_backend_name_map = dict(
```
cbrnr, line 23: Rename to `_backends`?

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
+    matplotlib='._mpl_figure',
+    pyqtgraph='._pyqtgraph'
+)
+backend = None
 
 
-class MNEFigParams:
-    """Container object for MNE figure parameters."""
+class BrowserParams:
+    """Container object for 2D-Data-Browser parameters."""
```
cbrnr, line 31:
```suggestion
    """Container object for 2D data browser parameters."""
```

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
-            close(self)
-        elif event.key == 'f11':  # full screen
-            self.canvas.manager.full_screen_toggle()
-
-    def _buttonpress(self, event):
-        """Handle buttonpress events."""
-        pass
+class BrowserBase(ABC):
+    """
+    A base class containing for the 2D-Data-Browser.
```
cbrnr, line 41:
```suggestion
    A base class containing for the 2D data browser.
```

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
-            from matplotlib.pyplot import close
-            close(self)
-        elif event.key == 'f11':  # full screen
-            self.canvas.manager.full_screen_toggle()
-
-    def _buttonpress(self, event):
-        """Handle buttonpress events."""
-        pass
+class BrowserBase(ABC):
+    """
```
cbrnr, line 40: I think we usually start our docstrings directly after `"""` (no newline).

mne/viz/_figure.py
```diff
@@ -2031,483 +308,197 @@ def _update_data(self):
...
-            _ = self._recompute_epochs_vlines(None)
-        self._draw_traces()
-        if annotations and not self.mne.is_epochs:
-            self._draw_annotations()
-        self.canvas.draw_idle()
+    def _get_epoch_num_from_time(self, time):
+        epoch_nums = self.mne.inst.selection
+        return epoch_nums[np.searchsorted(self.mne.boundary_times[1:], time)]
+
+    def _redraw(self, **kwargs):
```
cbrnr, line 315: Missing `@abstractmethod`?

drammock:
We don't have a strict policy about fixtures. We did have fixtures before, including some in `conftest` and some defined in the test files they were used in. IMO the changes here are fine. I guess a future PR that describes how to write tests could also add comments to test files explaining where to find the fixtures.

mne/viz/_figure.py
```diff
@@ -2031,483 +308,197 @@ def _update_data(self):
...
-    psd_kwargs = dict()
-    for kw in welch_kwargs + multitaper_kwargs:
-        if kw in kwargs:
-            psd_kwargs[kw] = kwargs.pop(kw)
-    if isinstance(inst, BaseRaw):
-        psd_func = partial(psd_welch, window=window)
-    elif isinstance(inst, BaseEpochs):
-        psd_func = psd_multitaper
+        kwargs['figsize'] = (8, 8)
+    # Initialize Browser-Backend
```
cbrnr, line 358:
```suggestion
    # Initialize browser backend
```

mne/viz/_figure.py
```diff
@@ -2031,483 +308,197 @@ def _update_data(self):
...
+    return browser
+
+
+def _check_browser_backend_name(backend_name):
+    _validate_type(backend_name, str, 'backend_name')
+    _check_option('backend_name', backend_name, VALID_BROWSE_BACKENDS)
+    return backend_name
+
+
+# ToDo: This won't appear in documentation, has to be adjusted to the current
```
cbrnr, line 373: What do you mean by "this won't appear in the documentation"?

mne/viz/_figure.py
```diff
@@ -2031,483 +308,197 @@ def _update_data(self):
...
+       +--------------------------------------+------------+-----------+
+       | Annotations                          | ✓          | ✓         |
+       +--------------------------------------+------------+-----------+
+       | Toggle Projections                   | ✓          |           |
+       +--------------------------------------+------------+-----------+
+       | Butterfly-Mode                       | ✓          |           |
+       +--------------------------------------+------------+-----------+
+       | Smooth Scrolling                     |            | ✓         |
+       +--------------------------------------+------------+-----------+
+       | OpenGL-Acceleration                  |            | ✓         |
```
cbrnr, line 422:
```suggestion
       | OpenGL Acceleration                  |            | ✓         |
```

mne/viz/_figure.py
```diff
@@ -2031,483 +308,197 @@ def _update_data(self):
...
+        return MNE_BROWSE_BACKEND
+
+
+def get_browser_backend():
+    """Return the 2D backend currently used.
+
+    Returns
+    -------
+    backend_used : str | None
+        The 2D-Browser backend currently in use. If no backend is found,
```
cbrnr, line 473:
```suggestion
        The 2D browser backend currently in use. If no backend is found,
```

cbrnr:
Looks good, I haven't checked the tests though. General comment: please make sure to spell terms such as 2D browser and browser backend without dashes (please check all your changes, I haven't commented on all occurrences).

mne/viz/epochs.py
marsipu: Good question, probably came with trying to adhere to <79 chars and the line looked long (but wasn't too long in fact).

mne/viz/_figure.py
marsipu: At first I thought this is not necessary for pyqtgraph but now I got it too so it can be abstract.

mne/viz/_figure.py
```diff
@@ -1,333 +1,60 @@
...
+    def __init__(self, **kwargs):
         from .. import BaseEpochs
         from ..io import BaseRaw
         from ..preprocessing import ICA
 
-        super().__init__(figsize=figsize, inst=inst, ica=ica, **kwargs)
+        self._data = None
+        self._times = None
+
+        self.mne = BrowserParams(**kwargs)
```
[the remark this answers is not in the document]
marsipu: This is a convention introduced in the matplotlib-backend to access parameters and gui-elements across parent and childs or depending gui-objects. I adapted it (before I had something similar with passing the parent-class onto the childs). The matplotlib-backend relies havily on this convetion so changing something there would have big consequences.
I think the name is just short which is good since it is used very frequently.

mne/conftest.py
```diff
@@ -17,11 +17,17 @@
...
 import mne
+from mne import read_events, pick_types, Epochs
+from mne.channels import read_layout
 from mne.datasets import testing
 from mne.fixes import has_numba
+from mne.io import read_raw_fif, read_raw_ctf
 from mne.stats import cluster_level
 from mne.utils import _pl, _assert_no_instances, numerics
 
+# data from sample-dataset
```
cbrnr, line 28:
```suggestion
# data from sample dataset
```

mne/conftest.py
```diff
@@ -33,10 +39,20 @@
 fname_aseg = op.join(test_path, 'subjects', 'sample', 'mri', 'aseg.mgz')
 subjects_dir = op.join(test_path, 'subjects')
 fname_src = op.join(bem_path, 'sample-oct-4-src.fif')
-subjects_dir = op.join(test_path, 'subjects')
-fname_cov = op.join(s_path, 'sample_audvis_trunc-cov.fif')
 fname_trans = op.join(s_path, 'sample_audvis_trunc-trans.fif')
 
+# CTF-Dataset
```
cbrnr, line 44:
```suggestion
```

mne/viz/epochs.py
cbrnr: It's still two lines

mne/viz/tests/test_epochs.py
```diff
@@ -7,122 +7,83 @@
...
-import os.path as op
-
+import matplotlib.pyplot as plt
 import numpy as np
 import pytest
-import matplotlib.pyplot as plt
 
-from mne import (read_events, Epochs, pick_types, read_cov, create_info,
+from mne import (Epochs, create_info,
                  EpochsArray)
```
cbrnr, line 15:
```suggestion
from mne import Epochs, create_info, EpochsArray
```

cbrnr:
@hoechenberger if you have time please take a final look and merge if you're happy.

agramfort (approved):
LGTM

@drammock @cbrnr @hoechenberger please approve the PR so we can merge. 

Great work @marsipu !

cbrnr (approved):

drammock (approved):

drammock:
in it goes!  Thanks 1e6 @marsipu

PR #9598, opened 2021-07-20
[ENH, MRG] Update show_view API

### LINKED ISSUE #9595: [ENH] Add focalpoint to `mne.viz.Brain`
It would be nice to be able to set the focal point of the camera in the rendering without relying on any `pyvista` internal calls/calls to `brain._renderer`.

### LINKED ISSUE #9597: [MAINT] Refactor `mne.viz.Brain` to simplify
So right now `mne.viz.Brain` is more or less a wrapper around `pyvista.mesh` with some cool layers and curvatures as well that are really helpful for brains and electrophysiology 3D viewing. The part that I think is overly complicated in the internal representation is the views.

The question is could we just update `mne.viz.Brain` to be more like `matplotlib.pyplot.axes` with a `mne.viz.BrainFigure`? Then, the examples with multiple views can have the user handle two or more brain objects with the equivalent of `matplotlib.pyplot.Figure.axes` instead of handling them internally with all these complicated views and column and row subplots. That seems like it could be simpler and allow the user more control over which views they want.

alexrockhill:
Fixes #9597
Fixes #9595

So it looks like `focalpoint` was already an option as a keyword argument of `view` in `mne.viz.Brain.show_view`. I thought it would be helpful to pull all the arguments out of the keyword argument to be documented and explain clearly that if a string such as 'lateral' is passed, it will take precedence as that way seems clear to me. All the arguments could also all be keyword arguments of `view` as in `mne.viz.Brain.add_volume_labels(legend=...)`, that's an easy change, doesn't make a lot of sense to do it half and half. 

The other part that is changed is that if `hemi=None` then it is assigned to volume which is already in use as for volume data/as a catchall term. I think this makes it a bit cleaner and clearer what's going on with the views, to me at least.

Just an idea for how to make things a bit simpler hopefully.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -113,6 +113,8 @@ Enhancements
 
 - Add :func:`mne.viz.Brain.add_volume_labels` to plot subcortical surfaces and other regions of interest (:gh:`9540` by `Alex Rockhill`_ and `Eric Larson`_)
 
+- Add split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting (:gh:`9596` by `Alex Rockhill`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -113,8 +113,6 @@ Enhancements
 
 - Add :func:`mne.viz.Brain.add_volume_labels` to plot subcortical surfaces and other regions of interest (:gh:`9540` by `Alex Rockhill`_ and `Eric Larson`_)
 
-- Add split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting (:gh:`9596` by `Alex Rockhill`_)
-
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
@@ -160,3 +158,6 @@ Bugs
 API changes
 ~~~~~~~~~~~
 - In `mne.compute_source_morph`, the ``niter_affine`` and ``niter_sdr`` parameters have been replaced by ``niter`` and ``pipeline`` parameters for more consistent and finer-grained control of registration/warping steps and iteration (:gh:`9505` by `Alex Rockhill`_ and `Eric Larson`_)
+
+- Split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting and make the default for ``row`` and ``col`` apply to all rows and columns (:gh:`9596` by `Alex Rockhill`_)
+
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -2323,10 +2323,7 @@ class Brain(object):
 
         Parameters
         ----------
-        aseg : str
-            The anatomical segmentation file. Default ``aparc+aseg``. This may
-            be any anatomical segmentation file in the mri subdirectory of the
-            subject directory from the Freesurfer recon-all.
+        %(aseg)s
         labels : list
             Labeled regions of interest to plot. See
             :func:`mne.get_montage_volume_labels`
@@ -2463,7 +2460,7 @@ class Brain(object):
             self._renderer.set_camera(**views_dicts[hemi][v])
 
     def add_text(self, x, y, text, name=None, color=None, opacity=1.0,
-                 row=-1, col=-1, font_size=None, justification=None):
+                 row=None, col=None, font_size=None, justification=None):
         """Add a text to the visualization.
 
         Parameters
@@ -2482,10 +2479,10 @@ class Brain(object):
             background color).
         opacity : float
             Opacity of the text (default 1.0).
-        row : int
-            Row index of which brain to use.
-        col : int
-            Column index of which brain to use.
+        row : int | None
+            Row index of which brain to use. Default all rows.
+        col : int | None
+            Column index of which brain to use. Default all columns.
         font_size : float | None
             The font size to use.
         justification : str | None
@@ -2495,8 +2492,13 @@ class Brain(object):
         # are implemented
         # _check_option('name', name, [None])
 
-        self._renderer.text2d(x_window=x, y_window=y, text=text, color=color,
-                              size=font_size, justification=justification)
+        for h in self._hemis:
+            for ri, ci, v in self._iter_views(h):
+                if (row is None or row == ri) and (col is None or col == ci):
+                    self._renderer.subplot(ri, ci)
+                    self._renderer.text2d(x_window=x, y_window=y, text=text,
+                                          color=color, size=font_size,
+                                          justification=justification)
 
     def _configure_label_time_course(self):
         from ...label import read_labels_from_annot
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -502,36 +502,8 @@ class Brain(object):
         self.plotter = self._renderer.plotter
 
         self._setup_canonical_rotation()
-        self._add_hemis(surf, alpha, offset, geo_kwargs)
 
-        self.interaction = interaction
-        self._closed = False
-        if show:
-            self.show()
-        # update the views once the geometry is all set
-        for h in self._hemis:
-            for ri, ci, v in self._iter_views(h):
-                self.show_view(v, row=ri, col=ci, hemi=h)
-
-        if surf == 'flat':
-            self._renderer.set_interaction("rubber_band_2d")
-
-    def _setup_canonical_rotation(self):
-        from ...coreg import fit_matched_points, _trans_from_params
-        self._rigid = np.eye(4)
-        try:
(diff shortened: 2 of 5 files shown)


larsoner:
> and explain clearly that if a string such as 'lateral' is passed, it will take precedence as that way seems clear to me

I would expect the opposite actually, that if you supply kwargs other than just the standard view, whatever you pass would take precedence. In other words, if you did `show_view('lat', az=15)` it should be a lateral view with the `az` value overwritten.

alexrockhill:
> > and explain clearly that if a string such as 'lateral' is passed, it will take precedence as that way seems clear to me
> 
> I would expect the opposite actually, that if you supply kwargs other than just the standard view, whatever you pass would take precedence. In other words, if you did `show_view('lat', az=15)` it should be a lateral view with the `az` value overwritten.

Hmmm that seems reasonable too, I'll switch it.

mne/viz/_brain/_brain.py
```diff
@@ -504,10 +502,42 @@ def __init__(self, subject_id, hemi='both', surf='pial', title=None,
...
+            self.show()
+        # update the views once the geometry is all set
         for h in self._hemis:
+            for ri, ci, v in self._iter_views(h):
+                self.show_view(v, row=ri, col=ci, hemi=h)
+
+        if surf == 'flat':
+            self._renderer.set_interaction("rubber_band_2d")
+
+    def _setup_canonical_rotation(self):
```
larsoner, line 519: please move back, having it moved up here makes the diff harder to read and blame harder to follow with minimal gain

mne/viz/_brain/_brain.py
```diff
@@ -2654,24 +2653,32 @@ def show(self):
         """Display the window."""
         self._renderer.show()
 
-    def show_view(self, view=None, roll=None, distance=None, row=0, col=0,
+    def show_view(self, view=None, azimuth=None, elevation=None, roll=None,
```
larsoner, line 2656: do not change order of arguments, you will break existing working code

mne/viz/_brain/_brain.py
alexrockhill: Sure

mne/viz/_brain/tests/test_brain.py
```diff
@@ -340,12 +340,11 @@ def __init__(self):
     for a, b, p, color in zip(annots, borders, alphas, colors):
         brain.add_annotation(a, b, p, color=color)
 
-    view_args = dict(view=dict(focalpoint=(1e-5, 1e-5, 1e-5)),
-                     roll=1, distance=500)
+    view_args = dict(roll=1, distance=500, focalpoint=(1e-5, 1e-5, 1e-5))
```
larsoner, line 343: This is a backward incompatible API change that will break existing working code. We can make this change (I think it's worth it) but we'll need a proper deprecation cycle

mne/viz/_brain/tests/test_brain.py
```diff
@@ -358,12 +357,14 @@ def __init__(self):
...
-    assert np.allclose(brain._renderer.figure._azimuth, view_args["azimuth"])
-    assert np.allclose(
-        brain._renderer.figure._elevation, view_args["elevation"])
-    assert np.allclose(cam.GetFocalPoint(), fp)
+    azimuth, elevation = 180., 90.
+    for view_args in (dict(azimuth=azimuth, elevation=elevation,
+                           focalpoint='auto'),
+                      dict(view='lateral', hemi='lh')):
+        brain.show_view(**view_args)
+        assert np.allclose(brain._renderer.figure._azimuth, azimuth)
```
larsoner, line 365: use `assert_allclose` rather than `assert np.allclose`, the error message is better when it fails and it has better defaults for testing (`atol=0` which makes more sense than `atol=1e-7` in general)

mne/viz/_brain/_brain.py
alexrockhill: Better as keyword arguments in `view`? That might be more backward compatible, then just `roll` and `distance` are depreciated

alexrockhill:
Would it be confusing to reference the keyword arguments here https://mne.tools/stable/generated/mne.viz.set_3d_view.html? I don't see a stable place where descriptions of camera parameters live in `pyvista`...

mne/viz/_brain/_brain.py
larsoner: I think I actually prefer the way you have it now (other than order of arguments, new ones need to be at the end), otherwise people end up writing stuff like `show_view(dict(azimuth=..., elevation=..., focalpoint=...))` instead of just `show_view(azimuth=..., elevation=..., focalpoint=...)` so `dict` is just an extra unnecessary hoop to jump through.

Basically we need to support `view` as dict rather than `None | str` (what we're deprecating toward) until 0.25 comes out

larsoner:
Yes I think `set_3d_view` is the right place, but since I'm advocating for proper `kwargs` rather than being packed in a `view=dict(...)`, it would mean shared entries in `mne/utils/docs.py`

alexrockhill:
Ok I added the deprecation cycle and the doc entries. I thought it was clearer to put the focus on the camera in the entries but feel free to comment and change if you don't like how it's worded.

alexrockhill:
Ok this looks good, if it looks okay to you

alexrockhill:
Oops, actually missed one failure that used the deprecated method

doc/changes/latest.inc
```diff
@@ -113,6 +113,8 @@ Enhancements
 
 - Add :func:`mne.viz.Brain.add_volume_labels` to plot subcortical surfaces and other regions of interest (:gh:`9540` by `Alex Rockhill`_ and `Eric Larson`_)
 
+- Add split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting (:gh:`9596` by `Alex Rockhill`_)
```
larsoner, line 116: This should actually go under API because it introduces a deprecation

mne/viz/_brain/_brain.py
```diff
@@ -2482,10 +2479,10 @@ def add_text(self, x, y, text, name=None, color=None, opacity=1.0,
...
         opacity : float
             Opacity of the text (default 1.0).
-        row : int
-            Row index of which brain to use.
-        col : int
-            Column index of which brain to use.
+        row : int | None
+            Row index of which brain to use. Default all rows.
+        col : int | None
+            Column index of which brain to use. Default all columns.
```
larsoner, line 2485: It's not uncommon to want *text* in only one window (and bottom right is at least a somewhat reasonable default), and this is a backward incompatible change. I think I'd just leave the old default actually, we could deprecate but again I'm not 100% convinced text should be treated the same way as data.

mne/viz/_brain/_brain.py
```diff
@@ -2654,24 +2656,26 @@ def show(self):
         """Display the window."""
         self._renderer.show()
 
-    def show_view(self, view=None, roll=None, distance=None, row=0, col=0,
-                  hemi=None, align=True):
+    @fill_doc
+    def show_view(self, view=None, roll=None, distance=None,
+                  azimuth=None, elevation=None, focalpoint=None,
+                  row=None, col=None, hemi=None, align=True):
```
larsoner, line 2662: Order is still wrong, new args must go at the end or it will break people's previously working code with no warning, e.g. if someone had `brain.show_view(None, 15, 0.1, 0, 0)` it is now broken. If we really want to change the order, you at least need to do:
```
    def show_view(self, view=None, roll=None, distance=None,
                  *, azimuth=None, elevation=None, focalpoint=None,
                  row=None, col=None, hemi=None, align=True):
```
but even this will break people's code. Thus even though it's not as clean, I'd rather keep the old order:
```suggestion
    def show_view(self, view=None, roll=None, distance=None,
                  row=None, col=None, hemi=None, align=True,
                  azimuth=None, elevation=None, focalpoint=None):
```

mne/viz/_brain/_brain.py
```diff
@@ -2685,16 +2689,28 @@ def show_view(self, view=None, roll=None, distance=None, row=0, col=0,
...
+                           distance=distance, focalpoint=focalpoint)
+        if view is not None:  # view string takes precedence
+            view_params = {param: val for param, val in view_params.items()
+                           if val is not None}  # no overwriting with None
+            view_params = dict(views_dicts[hemi].get(view), **view_params)
         xfm = self._rigid if align else None
-        self._renderer.set_camera(**view, reset_camera=False, rigid=xfm)
+        for h in self._hemis:
+            for ri, ci, v in self._iter_views(h):
+                if (row is None or row == ri) and (col is None or col == ci):
```
larsoner, line 2710: To properly deprecate, you need to set the default of `row` to something like `'deprecated'`, and if it's not passed (and there is more than one view!) warn that the old behavior will be used in 0.24 (row=0) but new behavior (row=None) will occur in 0.25

doc/changes/latest.inc
larsoner: ... also there are more deprecations than this, i.e., `row` and `col` both need deprecation cycles

mne/viz/_brain/_brain.py
alexrockhill: Ok I'll just leave the default as `-1` and document the bottom left behavior

mne/viz/_brain/_brain.py
alexrockhill: Ok, I figured people didn't use that many ordered arguments but your right better to be safe

larsoner (approved):
LGTM, @GuillaumeFavelier can you review and merge if you're happy?

GuillaumeFavelier (approved):
LGTM

GuillaumeFavelier:
Thank you @alexrockhill !

larsoner:
@alexrockhill can you make a quick follow-up PR to fix these?

https://app.circleci.com/pipelines/github/mne-tools/mne-python/9192/workflows/843a1e8e-b046-48bc-9909-908f9b29b317/jobs/31440

alexrockhill:
> @alexrockhill can you make a quick follow-up PR to fix these?
> 
> https://app.circleci.com/pipelines/github/mne-tools/mne-python/9192/workflows/843a1e8e-b046-48bc-9909-908f9b29b317/jobs/31440

Sure doing it now

PR #9600, opened 2021-07-21
MAINT: Avoid bad VTK [skip actions] [skip circle]

larsoner:


server_environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
 - conda-forge/label/vtk_dev
 - conda-forge
 dependencies:
-- python>=3.7
+- python ==3.8
```
hoechenberger, line 6: why the pinning?

larsoner: segfaults on 3.9 for unknown reasons

hoechenberger: Ouch. Has this been reported upstream?

larsoner: No, I haven't tried to replicate locally or anything. Feel free if you have time

hoechenberger: Also, do we actually need the VTK dev channel any longer?

larsoner: Not sure, maybe that's the problem with the bump to 3.9

hoechenberger: Will have a look.

PR #9602, opened 2021-07-22
Update deps: more available from conda-forge; remove nbformat pinning

hoechenberger:


server_environment.yml
```diff
@@ -12,16 +12,16 @@ dependencies:
...
 - nilearn
 - nibabel
-- nbformat <5.1  # XXX remove pinning once https://github.com/jupyter/nbformat/issues/206 has been fixed
+- nbformat
 - nbclient
+- jupyter
+- ipympl
+- ipywidgets
+- ipyvtklink
+- jupyter_client!=6.1.5
```
hoechenberger, line 24:
```suggestion
- jupyter_client !=6.1.5
```

cbrnr:
Why the space before the version specifier?

hoechenberger:
> Why the space before the version specifier?

It's the conda-forge way to do things. Thought it could make sense for consistency with the conda ecosystem

cbrnr:
OK, this seems like the general way to define [conda package match specifications](https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#package-match-specifications), so you should probably update all occurrences of such tags?

larsoner:
Failures here should hopefully be fixed by #9600

larsoner:
```
mne/viz/_brain/tests/test_brain.py::test_brain_time_viewer[pyvista] qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem
```
Maybe this is because Qt shouldn't be installed on the server_environment, and the changes in this PR cause it to be installed?

hoechenberger:
> Maybe this is because Qt shouldn't be installed on the server_environment, and the changes in this PR cause it to be installed?

could be, i'll have a look

hoechenberger:
This is stale and not urgent, so I'll close for now.

PR #9603, opened 2021-07-22
MRG, BUG: Catch bad input to setup_volume_source_space

larsoner:
I did ~~a bad thing~~ two bad things in my code, and this would have made it clear sooner rather than later (when I got `np.nan` in my forward solution).


hoechenberger (approved):

hoechenberger:
Thanks @larsoner!

Issue #9606, opened 2021-07-23
Better error message for 3D plots when pyvistaqt is missing


cbrnr:
When creating a 3D plot with e.g. `mne.viz.plot_alignment` (not sure if there are any additional similar functions), I get an error if `pyvistaqt` is not installed:

```
RuntimeError: Could not load any valid 3D backend: pyvista, mayavi, notebook
```

Note that I actually have `pyvista` installed, but it seems like `pyvistaqt` is *also* required. A simple solution would be to update the error message like this (since `pyvistaqt` depends on `pyvista`):

```
RuntimeError: Could not load any valid 3D backend: pyvistaqt, mayavi, notebook
```

hoechenberger:
cc @GuillaumeFavelier

GuillaumeFavelier:
Hm... I agree it's better.

GuillaumeFavelier:
I'll open a PR to rename the backend

PR #9607, opened 2021-07-23
MNT: Rename pyvista 3d backend to pyvistaqt

GuillaumeFavelier:
This PR renames the default `pyvista` 3d backend to `pyvistaqt`. 

Closes https://github.com/mne-tools/mne-python/issues/9606


hoechenberger:
This is a breaking change, no? I think we should have a deprecation cycle because users may have Code where they set the 3D backend to "pyvista" -- I certainly do!

GuillaumeFavelier:
You're right. I can modify `set_3d_backend` to re-route to the new one and throw a `DeprecationWarning`?

hoechenberger:
Yes that would be great 👍

larsoner:
I'd rather not deprecate `'pyvista'`. It's almost trivial to continue supporting it and I imagine a lot of folks have it in their machine configs. So I would document that `'pyvistaqt'` is the option but in the code do something like:
```
_validate_type(backend, str, 'backend')
# backward compat wrapper
backend = 'pyvistaqt' if backend == 'pyvista' else backend
_check_option('backend', backend, ('pyvistaqt', 'notebook', 'mayavi'))
```

GuillaumeFavelier:
@larsoner , is 13279aa16b57c66a3699f32b1bb139f991d6bcf0 enough?
@hoechenberger, does it work for you too?

larsoner:
Also I think we'll need the latest.inc entry to go in BUG actually because now if someone had this working code:
```
if mne.get_3d_backend() == 'pyvista':
    ...
```
it will now fail without any deprecation cycle (and I can't think of a good way to deprecate it). I think this is maybe okay because this is a sort of bug fix (?)

GuillaumeFavelier:
> it will now fail without any deprecation cycle 

I am not sure to understand. I don't think it will fail but rather just avoid the `if` statement right? This kind of code looks more like a silent bug, I don't know.

larsoner:
Let's say I wrote some code to tweak my plots that I wanted to work on both Mayavi and PyVista. (IIRC I had scripts like this a version or two ago when we were in transition.) In <= 0.23, you could have code like this:
```
if mne.viz.get_3d_backend() == 'pyvista':
    brain.plotter...
else:  # this is now Mayavi if you're on desktop
    from mayavi import mlab
    mlab....
```
This PR will break such code because the conditional now needs to be (without any deprecation cycle) the following to continue to work:
```
if mne.viz.get_3d_backend() == 'pyvistaqt':  # now needs "qt" on the end
    ...
else:
    ...
```
Thus this is a backward incompatible change with no deprecation cycle, at least for the return value of `get_3d_backend`
```

**Comment by drammock** [comment]
to me this seems like an overly complicated solution, why not just change the error message like @cbrnr suggested in #9606?

**Comment by GuillaumeFavelier** [comment]
Thank you for clarifying.

> Thus this is a backward incompatible change with no deprecation cycle

Apart from having both entries: `pyvista` and `pyvistaqt`, I do not see any solution for this 

> to me this seems like an overly complicated solution

@drammock I wanted to address two issues at the same time since the confusion between `notebook` and `pyvista` has been reported a while back

**Comment by hoechenberger** [comment]
> Apart from having both entries: `pyvista` and `pyvistaqt`, I do not see any solution for this

Yes why not have those? And a deprecation warning that the first one will go away? Am I missing something here? It seems quite simple to me 🤔

**Comment by larsoner** [comment]
>> to me this seems like an overly complicated solution, why not just change the error message like @cbrnr suggested in #9606?
> Apart from having both entries: pyvista and pyvistaqt, I do not see any solution for this

I can live with a variant of the current solution, but I also agree with @drammock that it is not as simple as it seems

> Am I missing something here? It seems quite simple to me ￼

@hoechenberger look a few comments up for a breaking use case: https://github.com/mne-tools/mne-python/pull/9607#issuecomment-885671037

> And a deprecation warning that the first one will go away?

And a few comments farther up where I talk about this: https://github.com/mne-tools/mne-python/pull/9607#issuecomment-885649653

>

**Comment by drammock** [comment]
> I wanted to address two issues at the same time since the confusion between `notebook` and `pyvista` has been reported

in that case I think I'd vote for @larsoner's suggestion [here](https://github.com/mne-tools/mne-python/pull/9607#issuecomment-885649653) (accept `pyvista` as a synonym of `pyvistaqt`, but don't include it in docs / error messages)

**Comment by GuillaumeFavelier** [comment]
I tried to https://github.com/mne-tools/mne-python/pull/9607#issuecomment-885671037 in 3c958c432dcaff2c402dcb322ffda856bcac61b5 but I can't be less confident about it.

The PR is ready for review.

**Comment by larsoner** [comment]
@GuillaumeFavelier okay for me to push a quick commit to tweak latest.inc and check the `_check_option` output?

**Comment by GuillaumeFavelier** [comment]
Yes @larsoner, feel free to edit

**Comment by hoechenberger** [comment]
@larsoner Thanks, I somehow missed these comments!

**Comment by cbrnr** [comment]
I'm happy with whatever you decide to implement, but simply changing the error message would probably have been sufficient.

File: doc/conf.py [hunk 1]
```diff
@@ -346,7 +346,7 @@ def __call__(self, gallery_conf, fname):
         mlab.close()
         scrapers += ('mayavi',)
         push_exception_handler(reraise_exceptions=True)
-    elif backend in ('notebook', 'pyvista'):
+    elif backend in ('notebook', 'pyvistaqt'):
```
larsoner, line 349: We had our own use case that we broke :)

mne/utils/config.py
```diff
@@ -559,7 +559,7 @@ def sys_info(fid=None, show_paths=False):
                 try:
                     from pyvistaqt import __version__
                 except Exception:
-                    pass
+                    extras += 'pyvistaqt not found'
```
larsoner, line 562: More useful output in `mne sys_info`

mne/viz/backends/_qt.py
```diff
@@ -8,10 +8,7 @@
 from contextlib import contextmanager
 
 import pyvista
-try:
-    from pyvista.plotting.qt_plotting import FileDialog
-except ImportError:
-    from pyvistaqt.plotting import FileDialog
```
larsoner, line 14: Delete cruft

mne/viz/backends/renderer.py
```diff
@@ -37,10 +37,17 @@ def _reload_backend(backend_name):
 
 
 def _get_renderer(*args, **kwargs):
-    set_3d_backend(_get_3d_backend(), verbose=False)
+    _get_3d_backend()
```
larsoner, line 40: I don't see why we'd need to set the get when the get does a `_reload_backend` under the hood already?

mne/viz/backends/renderer.py
```diff
@@ -169,18 +193,18 @@ def use_3d_backend(backend_name):
 
     Parameters
     ----------
-    backend_name : {'mayavi', 'pyvista', 'notebook'}
+    backend_name : {'mayavi', 'pyvistaqt', 'notebook'}
         The 3d backend to use in the context.
     """
-    old_backend = _get_3d_backend()
-    set_3d_backend(backend_name)
```
larsoner, line 176: This behavior was wacky -- `_get_3d_backend` would set a backend (otherwise raise an error), only for it potentially to be un-set immediately by the following `set_3d_backend` call. The only reason I could see for this was that the old backend name was needed, so I just made `set_3d_backend` return the previous backend's name...

mne/viz/backends/renderer.py
```diff
@@ -169,18 +193,18 @@ def use_3d_backend(backend_name):
...
+    old_backend = set_3d_backend(backend_name)
     try:
         yield
     finally:
-        try:
-            set_3d_backend(old_backend)
-        except Exception:
-            pass
+        if old_backend is not None:
```
larsoner, line 203: only need to set back if one was loaded before

mne/viz/tests/test_3d.py
```diff
@@ -835,3 +837,22 @@ def test_renderer(renderer):
...
+        monkeypatch.delattr('pyvistaqt.plotting.FileDialog')
+    monkeypatch.setattr(
+        'mne.viz.backends.renderer.MNE_3D_BACKEND', None)
+    monkeypatch.setattr(
+        'mne.viz.backends.renderer.VALID_3D_BACKENDS', ('pyvistaqt',))
+    # avoid using the config
+    monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmpdir))
+    match = 'Could not load any valid 3D.*\npyvistaqt: cannot import.*'
+    with pytest.raises(RuntimeError, match=match):
+        _get_renderer()
```
larsoner, line 858: Now the error is explicit, e.g. in this test:
```
E   RuntimeError: Could not load any valid 3D backend:
E   pyvistaqt: cannot import name 'FileDialog' from 'pyvistaqt.plotting' (/home/larsoner/python/pyvistaqt/pyvistaqt/plotting.py)
```

mne/viz/backends/renderer.py
```diff
@@ -52,86 +59,102 @@ def set_3d_backend(backend_name, verbose=None):
...
+       | Inline plot in JupyterLab            | ✓      |           | ✓        |
+       +--------------------------------------+--------+-----------+----------+
+       | Inline plot in Google Colab          |        |           |          |
+       +--------------------------------------+--------+-----------+----------+
+       | Toolbar                              |        | ✓         | ✓        |
+       +--------------------------------------+--------+-----------+----------+
     """
     global MNE_3D_BACKEND
-    try:
-        MNE_3D_BACKEND
```
larsoner, line 109: I don't understand this `try` clause, `MNE_3D_BACKEND=None` is set at the top of the module so it should always exist unless someone does `del mne.viz.backends.renderer.MNE_3D_BACKEND` or so which I don't think we need to support?

larsoner (commented):
@GuillaumeFavelier in my quest to improve the error message I found a bunch of oddities in the 3D backend setting code. Can you see if my comments/changes make sense to you? Hopefully the CIs come back happy to indicate that I didn't break anything...

mne/viz/backends/tests/test_renderer.py
```diff
@@ -177,3 +174,34 @@ def test_get_3d_backend(renderer):
     orig_backend = renderer.MNE_3D_BACKEND
     assert renderer.get_3d_backend() == orig_backend
     assert renderer.get_3d_backend() == orig_backend
+
+
+def test_renderer(renderer, monkeypatch):
```
larsoner, line 179: moved here from `mne/viz/tests/test_3d.py`

mne/viz/backends/renderer.py
larsoner: it looks like it was just used in mocking in tests, I deleted it in favor of something simpler (monkeypatch.setattr)

mne/viz/backends/renderer.py
drammock: heh, this might explain why I sometimes see `setting pyvista 3d backend` twice in a row when running some of my scripts.

agramfort (approved):
@larsoner merge if happy

thanks a lot for completing this.

hoechenberger (approved):

hoechenberger:
Thanks @GuillaumeFavelier and @larsoner!

PR #9608, opened 2021-07-23
WIP, MNT: Deprecate PGD solver for MxNE

PABannier:
Based on benchmarks carried out with @agramfort , PGD is consistently slower than accelerated BCD. 

**The goal of this PR is to deprecate it**.

==============

Benchmark result:

Screen Shot 2021-07-23 at 13 34 50

In dark red, PGD (FISTA). In red, the BCD solver in MNE. 

PGD takes twice as much time. Similar results were found for different levels of regularization. Note that this benchmark is made on real data (sample, left auditory).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -167,4 +167,4 @@ API changes
 
 - Split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting and make the default for ``row`` and ``col`` apply to all rows and columns (:gh:`9596` by `Alex Rockhill`_)
 
-- Deprecate :func:`mne.inverse_sparse.mxne_optim._mixed_norm_solver_prox` due to performance reasons: coordinate descent solvers consistently outperforms proximal gradient descent (:gh:`9608` by `Pierre-Antoine Bannier`_)
+- Deprecate ``solver='prox'`` in :func:`mne.inverse_sparse.mxne_inverse` in favor of ``solver='cd'`` and ``solver='bcd'`` as coordinate descent solvers consistently outperform proximal gradient descent (:gh:`9608` by `Pierre-Antoine Bannier`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -163,3 +163,4 @@ API changes
 
 - Split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting and make the default for ``row`` and ``col`` apply to all rows and columns (:gh:`9596` by `Alex Rockhill`_)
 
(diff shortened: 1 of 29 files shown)


mne/inverse_sparse/mxne_optim.py
```diff
@@ -257,6 +261,9 @@ def dgap_l21(M, G, X, active_set, alpha, n_orient):
     return gap, p_obj, d_obj, R
 
 
+@deprecated(extra="Starting version v0.24, MxNE problems will be solved "
```
larsoner, line 264: That will start in 0.25 (0.24 is the upcoming release, where it will not yet be (B)CD-only)

larsoner:
Next step I think is to add to `latest.inc` and also wrap some test calls in `with pytest.deprecated_call(match='will be solved'):`

doc/changes/latest.inc
```diff
@@ -166,3 +166,5 @@ API changes
 - In `mne.compute_source_morph`, the ``niter_affine`` and ``niter_sdr`` parameters have been replaced by ``niter`` and ``pipeline`` parameters for more consistent and finer-grained control of registration/warping steps and iteration (:gh:`9505` by `Alex Rockhill`_ and `Eric Larson`_)
 
 - Split :func:`mne.viz.Brain.show_view` argument ``view`` into ``azimuth``, ``elevation`` and ``focalpoint`` for clearer view setting and make the default for ``row`` and ``col`` apply to all rows and columns (:gh:`9596` by `Alex Rockhill`_)
+
+- Deprecate :func:`mne.inverse_sparse.mxne_optim._mixed_norm_solver_prox` due to performance reasons: coordinate descent solvers consistently outperforms proximal gradient descent (:gh:`9608` by `Pierre-Antoine Bannier`_)
```
larsoner, line 170: Don't mention private functions/methods in latest.inc (this is a user-facing document and users should not have to care about private functions), and also give people a hint of what they need to do to change their code, e.g.:
```suggestion
- Deprecate ``solver='prox'`` in :func:`mne.inverse_sparse.mxne_inverse` in favor of ``solver='cd'`` and ``solver='bcd'`` as coordinate descent solvers consistently outperform proximal gradient descent (:gh:`9608` by `Pierre-Antoine Bannier`_)
```

PABannier: Ok! understood!

PABannier:
@larsoner the doc does not compile. I thought it was due to a misformatting error in latest.inc, but it seems that some HTML page is not created. Any idea why? Am I missing something?

larsoner:
Warnings are treated as errors so if you ctrl-F in the CircleCI output for `warn` you see:
```
changes/latest.inc:170: WARNING: py:func reference target not found: mne.inverse_sparse.mxne_inverse
```

agramfort (approved):

larsoner:
@PABannier ready to go from your end?

PABannier:
@larsoner ready to go for me!

PR #9613, opened 2021-07-26
WIP: Expand ~ in _check_fname() – Looking for help

hoechenberger:
Currently just a proof of concept for FIFF reading. Will expand to other places across the entire code base. Feel super free to help me and push to this branch.

### DIFF
--- a/mne/io/fiff/tests/test_raw_fiff.py
+++ b/mne/io/fiff/tests/test_raw_fiff.py
@@ -1786,3 +1786,17 @@ def test_corrupted(tmpdir):
     with pytest.warns(RuntimeWarning, match='.*tag directory.*corrupt.*'):
         raw_bad = read_raw_fif(bad_fname)
     assert_allclose(raw.get_data(), raw_bad.get_data())
+
+
+@testing.requires_testing_data
+def test_expand_user():
+    dir_sample = pathlib.Path('~/mne_data/MNE-testing-data/MEG/sample/')
+    fname_in = dir_sample / 'sample_audvis_trunc_raw.fif'
+    dir_out = dir_sample / 'tmp'
+    dir_out.expanduser().mkdir(exist_ok=True)
+    fname_out = dir_out / 'raw.fif'
+
+    raw = read_raw_fif(fname=fname_in, preload=True)
+    raw.save(fname=fname_out, overwrite=True)
+
+    shutil.rmtree(dir_out.expanduser())
--- a/mne/utils/check.py
+++ b/mne/utils/check.py
@@ -155,6 +155,8 @@ def _check_fname(fname, overwrite=False, must_exist=False, name='File',
                  need_dir=False):
     """Check for file existence, and return string of its absolute path."""
     _validate_type(fname, 'path-like', name)
+    fname = op.expanduser(fname)
+
     if op.exists(fname):
         if not overwrite:
             raise FileExistsError('Destination file exists. Please use option '
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -119,6 +119,8 @@ Enhancements
 
 - :meth:`mne.preprocessing.ICA.fit` now emits a warning if any of the ``start``, ``stop``, ``reject``, and ``flat`` parameters are passed when performing ICA on `~mne.Epochs`. These parameters only have an effect on `~mne.io.Raw` data and were previously silently ignored in the case of `~mne.Epochs` (:gh:`9605` by `Richard Höchenberger`_)
 
+- All functions for reading and writing files should now automatically handle ``~`` correctly on Linux and macOS machines and expand it to the user's home directory. Should you come across any function that doesn't do it, please do let us know! (:gh:`9613` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1416,7 +1416,6 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         or all forms of SSS). It is recommended not to concatenate and
         then save raw files for this reason.
         """
-        fname = op.abspath(fname)
         endings = ('raw.fif', 'raw_sss.fif', 'raw_tsss.fif',
                    '_meg.fif', '_eeg.fif', '_ieeg.fif')
         endings += tuple([f'{e}.gz' for e in endings])
@@ -1448,7 +1447,7 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                              '"double", not "short"')
 
         # check for file existence
-        _check_fname(fname, overwrite)
+        fname = _check_fname(fname, overwrite)
 
         if proj:
             info = deepcopy(self.info)


larsoner:
Sounds like a good idea to me! Then once we standardize _check_fname use in all functions that read and write (which we can do progressively / not in this PR) then `~` should work everywhere, which is nice!

cbrnr:
Can we use `pathlib.resolve()` instead?

hoechenberger:
> Then once we standardize _check_fname use in all functions that read and write (which we can do progressively / not in this PR)

Oh, my intention actually **was** to get this to work in this PR :) But maybe that's too ambitious. Should I limit things to just FIFF for now (as currently implemented) in this PR?

larsoner:
I think it's okay to go for as much as you're up for doing in this PR

> Can we use pathlib.resolve() instead?

No idea what the implications are of switching from our `os.path`-based functions to `pathlib.resolve`

hoechenberger:
> Can we use `pathlib.resolve()` instead?

That's resolving symlinks, not expanding `~`.

If the intention of the question is, "can we start moving our path handling over to pathlib", I'd be +9999, it's so much nicer and cleaner than dealing with strings and `os.path` in many situations.

hoechenberger:
> > Can we use `pathlib.resolve()` instead?
> 
> That's resolving symlinks, not expanding `~`.

Maybe you were thinking of `Path.expanduser()`?

cbrnr:
> Maybe you were thinking of `Path.expanduser()`?

Yes!

hoechenberger:
> > Maybe you were thinking of `Path.expanduser()`?
> 
> Yes!

I'm using that one in the test, it's **something** 😅🙈

hoechenberger:
Ok I think I'm now covering a large number (all? probably forgot some!) data types we deal with in MNE.

I'm too lazy to add tests for all of them now :( Somebody willing to give me a hand?

hoechenberger:
Also sometimes it's not absolutely clear at which **level** to expand the fname. I'm trying to to it early, but I'm not always super consistent here.

mne/minimum_norm/inverse.py
```diff
@@ -331,6 +331,8 @@ def write_inverse_operator(fname, inv, verbose=None):
     """
     check_fname(fname, 'inverse operator', ('-inv.fif', '-inv.fif.gz',
                                             '_inv.fif', '_inv.fif.gz'))
```
larsoner, line 333: Might be better / simpler just to make `check_fname` call `_check_fname` internally and return the `str`-ified absolute path, then our code ends up more DRY. It seems like every time we `check_fname` we probably want to call `_check_fname` anyway...

Or we could go the other way and make `_check_fname` accept a `ext` argument, and have it call `check_fname` internally. I think I'd rather go that way.

mne/source_estimate.py
```diff
@@ -250,7 +250,11 @@ def read_source_estimate(fname, subject=None):
     """  # noqa: E501
     fname_arg = fname
     _validate_type(fname, 'path-like', 'fname')
-    fname = str(fname)
+
+    # Expand `~` if present
+    # We're not using _check_fname() here due to the particularities involved
+    # in STC file handling (-lh, -rh suffix) executed below
+    fname = op.expanduser(fname)
```
larsoner, line 257: Can't you still use `fname = _check_fname(fname, overwrite='read', must_exist=False)` despite these particularities?

mne/utils/docs.py
```diff
@@ -267,6 +267,12 @@
     If proj is False no projections will be applied which is the
     recommended value if SSPs are not used for cleaning the data.
 """
+docdict['epochs_fname'] = """
+fname : str | file-like
+
```
larsoner, line 272:
```suggestion
```

mne/source_estimate.py
hoechenberger: I had a look yesterday and felt that it would increase code complexity and reduce readability. I can have another look tomorrow with hopefully fresher eyes.

mne/utils/docs.py
hoechenberger: should probably be `path-like | file-like` anyway

mne/minimum_norm/inverse.py
hoechenberger: Yes I was thinking the same, however I don't want to touch this in this PR to avoid unwelcome surprises and to keep the diff smaller.

mne/io/fiff/tests/test_raw_fiff.py
```diff
@@ -1786,3 +1786,18 @@ def test_corrupted(tmpdir):
...
+    dir_sample = pathlib.Path('~/mne_data/MNE-testing-data/MEG/sample/')
+    fname_in = dir_sample / 'sample_audvis_trunc_raw.fif'
+    dir_out = dir_sample / 'tmp'
+    dir_out.expanduser().mkdir(exist_ok=True)
+    fname_out = dir_out / 'raw.fif'
+
+    raw = read_raw_fif(fname=fname_in, preload=True)
+    raw.save(fname=fname_out, overwrite=True)
+
+    shutil.rmtree(dir_out.expanduser())
```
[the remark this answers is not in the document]
hoechenberger: Good idea, I'll give it a shot.

mne/io/eeglab/eeglab.py
```diff
@@ -446,6 +446,7 @@ def __init__(self, input_fname, events=None, event_id=None, tmin=0,
                  baseline=None, reject=None, flat=None, reject_tmin=None,
                  reject_tmax=None, eog=(), verbose=None,
                  uint16_codec=None):  # noqa: D102
+        input_fname = _check_fname(fname=input_fname, must_exist=True)
```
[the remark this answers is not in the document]
hoechenberger: Good catch, seems I forgot that one. Interesting the tests still seem to pass … I'll have a look.

mne/cov.py
```diff
@@ -153,7 +153,7 @@ def save(self, fname):
         """
         check_fname(fname, 'covariance', ('-cov.fif', '-cov.fif.gz',
                                           '_cov.fif', '_cov.fif.gz'))
-
+        fname = _check_fname(fname=fname, overwrite=True)
```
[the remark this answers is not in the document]
hoechenberger: I opt for doing it later :)

mne/evoked.py
```diff
@@ -277,7 +278,10 @@ def apply_baseline(self, baseline=(None, 0), *, verbose=None):
         return self
 
     def save(self, fname):
-        """Save dataset to file.
+        """Save evoked data to a file.
+
+        .. note:: To write multiple conditions into a single file, use
+                  `mne.write_evokeds` instead.
```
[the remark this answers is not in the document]
hoechenberger: Reverted.

mne/epochs.py
```diff
@@ -3119,8 +3114,12 @@ class EpochsFIF(BaseEpochs):
     def __init__(self, fname, proj=True, preload=True,
                  verbose=None):  # noqa: D102
         if isinstance(fname, str):
```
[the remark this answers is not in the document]
hoechenberger: Fixed by using `_check_path_like()`

mne/io/fiff/tests/test_raw_fiff.py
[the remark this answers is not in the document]
hoechenberger: I've implemented the approach you proposed.

mne/utils/docs.py
hoechenberger: Fixed.

mne/source_estimate.py
hoechenberger: Your proposal seems to do the trick. I've pushed this change.

hoechenberger:
I believe I have addressed all comments.

hoechenberger:
I still found some code snippets where I'm apparently still not expanding `~`, will need to investigate. Do not merge yet.

hoechenberger:
Ok I think this is good to merge once CI goes green.

hoechenberger:
Ok this is ready to go from my end!

mne/utils/check.py
```diff
@@ -156,6 +156,8 @@ def _check_fname(fname, overwrite=False, must_exist=False, name='File',
                  need_dir=False):
     """Check for file existence, and return string of its absolute path."""
     _validate_type(fname, 'path-like', name)
+    fname = op.expanduser(fname)
```
larsoner, line 159: now the modification of `fname` happens it two places -- here and below during the `return`. Can you move the `fname = str(op.abspath(fname))` after this line, then just `return fname` at the end? That way it's immediately clear what all operations we are actually doing to `fname` before returning it.

mne/utils/tests/test_config.py
```diff
@@ -89,17 +89,23 @@ def test_sys_info():
...
 
     # `None`
-    monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmpdir))
+    monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmp_path))
     monkeypatch.delenv('SUBJECTS_DIR', raising=False)
     assert get_subjects_dir() is None
+
+    # Expand `~`
+    monkeypatch.setenv('HOME', str(tmp_path))
+    monkeypatch.setenv('USERPROFILE', str(tmp_path))  # Windows
```
larsoner, line 110: In your `latest.inc` you just say macOS and Linux, but this makes it seem like Windows will work, too...?

larsoner (changes requested):
Other than some minor stuff, LGTM!

hoechenberger:
@larsoner This should be good to go.

PR #9614, opened 2021-07-27
MRG: Make viz.plot_alignment() continue on topological defects; make coreg display a suitable head surface even if it has topological defects

hoechenberger:
I wanted to run `viz.plot_alignment()` and it crashed with an exception because there were a few topological defects (as is often the case with our data).

So now, I'm passing `on_defects='warn'` to the `read_bem_surfaces()` call inside `plot_alignment()`.

At first I made this parameter user-configurable, but then I thought, `plot_alignment()` is used when the user wants to see the data they have, so I felt it's not only okay, but expected to always pass `on_defects='warn'`.

Not sure how to test this, though – maybe it's okay as-is? I can confirm that this works for me.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -121,6 +121,8 @@ Enhancements
 
 - :func:`mne.viz.plot_alignment` now emits a warning (instead of aborting with an exception) if the surfaces contain topological defects (:gh:`9614` by `Richard Höchenberger`_)
 
+- The coregistration GUI can now display the head surface even if there are topological defects. Previously, a low-resolution head devoid of most facial features and hence unsuitable for coregistration was displayed (:gh:`9614` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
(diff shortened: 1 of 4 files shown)


doc/changes/latest.inc
```diff
@@ -119,6 +119,8 @@ Enhancements
 
 - :meth:`mne.preprocessing.ICA.fit` now emits a warning if any of the ``start``, ``stop``, ``reject``, and ``flat`` parameters are passed when performing ICA on `~mne.Epochs`. These parameters only have an effect on `~mne.io.Raw` data and were previously silently ignored in the case of `~mne.Epochs` (:gh:`9605` by `Richard Höchenberger`_)
 
+- :meth:`mne.viz.plot_alignment` now emits a warning (instead of aborting with an exception) if the surfaces contain topological defects (:gh:`9614` by `Richard Höchenberger`_)
```
hoechenberger, line 122:
```suggestion
- :func:`mne.viz.plot_alignment` now emits a warning (instead of aborting with an exception) if the surfaces contain topological defects (:gh:`9614` by `Richard Höchenberger`_)
```

agramfort (approved):
thx @hoechenberger

hoechenberger:
I encountered a similar problem with the coreg UI, only that it wouldn't crash when the exception was raised, but the displayed head surface would be unusable for a proper coregistration. I'm addressing this in 43cd114

mne/utils/docs.py
```diff
@@ -2453,7 +2453,7 @@
 
 # BEM
 docdict['on_defects'] = """
-on_defects : str
+on_defects : 'raise' | 'warn'
```
larsoner, line 2456: This should use the `_on_missing` docstring base and functionality, can you open a PR to fix this? ('ignore'` is missing here, and the code is not DRY.)

hoechenberger: I didn't know `ignore` was supported here, ugh. Yes I can open a followup PR!

larsoner: It should be as long as `on_defects` uses the `_on_missing` machinery. If it doesn't do that, it should

PR #9617, opened 2021-07-27
[MAINT, MRG] Move `_check_mri` to _freesurfer.py

alexrockhill:
Puts `_check_mri` in a more appropriate module.

### DIFF
--- a/mne/viz/misc.py
+++ b/mne/viz/misc.py
@@ -24,6 +24,7 @@ import numpy as np
 
 from ..defaults import DEFAULTS
 from ..fixes import _get_img_fdata
+from .._freesurfer import _mri_orientation, _read_mri_info, _check_mri
 from ..rank import compute_rank
 from ..surface import read_surface
 from ..io.constants import FIFF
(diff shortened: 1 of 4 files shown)


mne/viz/misc.py
```diff
@@ -520,6 +519,7 @@ def plot_bem(subject=None, subjects_dir=None, orientation='coronal',
     on top of the midpoint MRI slice with the BEM boundary drawn for that
     slice.
     """
+    from .._freesurfer import _check_mri
```
larsoner, line 522: It doesn't seem like this should need to be nested, does it actually need to be?

larsoner (commented):
Otherwise LGTM

mne/viz/misc.py
alexrockhill: The other one is, I'll pull them both out

mne/viz/misc.py
```diff
@@ -309,7 +309,6 @@ def _plot_mri_contours(mri_fname, surfaces, src, orientation='coronal',
     """Plot BEM contours on anatomical slices."""
     import matplotlib.pyplot as plt
     from matplotlib import patheffects
-    from .._freesurfer import _mri_orientation, _read_mri_info
```
larsoner, line 312: This is better, internal MNE imports should be nested only when required by import cycle problems

larsoner (approved):

agramfort (approved):

larsoner:
Thanks @alexrockhill !

Issue #9594, opened 2021-07-20
[ENH] Add `mne.viz.Brain.remove` method


alexrockhill:
A coherent style would be to have actors added with `mne.viz.Brain.add_volume_labels`, `mne.viz.Brain.add_head`, etc. and then they could similarly be removed with a corresponding remove method that removes the meshes from the renderer and deletes them from an internally stored list. There could be some refactoring/general code to share for adding/removing.

Related to https://github.com/mne-tools/mne-python/issues/9584

alexrockhill:
@GuillaumeFavelier @larsoner

PR #9618, opened 2021-07-27
[ENH, MRG] Add `mne.viz.Brain.add_head` method to display head

### LINKED ISSUE #9584: [BUG] `mne.viz.Brain` raises cryptic error when plotting 'seghead', works for `mne.viz.plot_alignment`
Here's some minimally reproducible code:

```
import os.path as op
import mne

mne.viz.plot_alignment(evoked.info, trans, subject='sample', subjects_dir=subjects_dir, surfaces=['seghead'])  # works
mne.viz.Brain('sample', 'lh', 'seghead', subjects_dir=subjects_dir)  # raises error
```

Stack trace
```
  File "/Users/alexrockhill/projects/mne-python/mne/viz/_brain/_brain.py", line 523, in __init__
    mesh.add_overlay(
  File "/Users/alexrockhill/projects/mne-python/mne/viz/_brain/_brain.py", line 178, in add_overlay
    self._apply()
  File "/Users/alexrockhill/projects/mne-python/mne/viz/_brain/_brain.py", line 194, in _apply
    self._renderer._set_mesh_scalars(
  File "/Users/alexrockhill/projects/mne-python/mne/viz/backends/_pyvista.py", line 699, in _set_mesh_scalars
    mesh.point_arrays[name] = scalars
  File "/Users/alexrockhill/software/anaconda3/envs/mne/lib/python3.9/site-packages/pyvista/core/datasetattributes.py", line 57, in __setitem__
    self.append(narray=value, name=key)
  File "/Users/alexrockhill/software/anaconda3/envs/mne/lib/python3.9/site-packages/pyvista/core/datasetattributes.py", line 211, in append
    raise ValueError(f'narray length of ({narray.shape[0]}) != required length ({array_len})')
ValueError: narray length of (155407) != required length (267122)
```

alexrockhill:
Fixes #9584.

### DIFF
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -9,13 +9,14 @@ import os.path as op
 import numpy as np
 from gzip import GzipFile
 
+from .bem import _bem_find_surface, read_bem_surfaces
 from .io.constants import FIFF
 from .io.meas_info import read_fiducials
 from .transforms import (apply_trans, invert_transform, combine_transforms,
                          _ensure_trans, read_ras_mni_t, Transform)
-from .surface import read_surface
+from .surface import read_surface, _read_mri_surface
 from .utils import (verbose, _validate_type, _check_fname, _check_option,
-                    get_subjects_dir, _require_version)
+                    get_subjects_dir, _require_version, logger)
 
 
 def _check_subject_dir(subject, subjects_dir):
@@ -436,15 +437,16 @@ def read_talxfm(subject, subjects_dir=None, verbose=None):
 
 
 def _check_mri(mri, subject, subjects_dir):
+    """Check whether an mri exists in the Freesurfer subject directory."""
     _validate_type(mri, 'path-like', 'mri')
     if not op.isfile(mri):
         if subject is None:
             raise FileNotFoundError(
-                'MRI file %r not found and no subject provided' % (mri,))
+                f'MRI file {mri} not found and no subject provided')
         subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
         mri = op.join(subjects_dir, subject, 'mri', mri)
         if not op.isfile(mri):
-            raise FileNotFoundError('MRI file %r not found' % (mri,))
+            raise FileNotFoundError(f'MRI file {mri} not found')
     return mri
 
 
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
     lut = {d[0]: np.array(lut[d[0]], dtype=d[1]) for d in dtype}
     assert len(lut['name']) > 0
     return lut
+
+
+def get_head_surface(surf, subject, subjects_dir, bem=None):
+    """Get a head surface from the Freesurfer subject directory.
+
+    Parameters
+    ----------
+    surf : str
+        The name of the surface 'auto', 'head', 'outer_skin', 'head-dense'
+        or 'seghead'.
+    %(subject)s
+    %(subjects_dir)s
+
+    Returns
+    -------
+    verts : ndarray
+        The surface vertices or None if ``surf='auto'`` and none is found.
+    triangles : ndarray
+        The vertex triangles or None if ``surf='auto'`` and none is found.
+
+    Notes
+    -----
+    .. versionadded: 0.24
+    """
+    _check_option(
+        'surf', surf, ('auto', 'head', 'outer_skin', 'head-dense', 'seghead'))
+    if surf in ('auto', 'head', 'outer_skin'):
+        if bem is not None:
+            try:
+                return _bem_find_surface(bem, 'head')
+            except RuntimeError:
+                logger.info('Could not find the surface for '
+                            'head in the provided BEM model, '
+                            'looking in the subject directory.')
+    if subject is None:
+        if surf == 'auto':
+            return
+        raise ValueError('To plot the head surface, the BEM/sphere'
+                         ' model must contain a head surface '
+                         'or "subject" must be provided (got '
+                         'None)')
+    subject_dir = op.join(
+        get_subjects_dir(subjects_dir, raise_error=True), subject)
+    if surf in ('head-dense', 'seghead'):
+        try_fnames = [op.join(subject_dir, 'bem', f'{subject}-head-dense.fif'),
+                      op.join(subject_dir, 'surf', 'lh.seghead')]
+    else:
+        try_fnames = [op.join(subject_dir, 'bem', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', 'flash', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', f'{subject}-head.fif')]
+    for fname in try_fnames:
+        if op.exists(fname):
+            logger.info(f'Using {op.basename(fname)} for head surface.')
+            if op.splitext(fname)[-1] == '.fif':
+                return read_bem_surfaces(fname, on_defects='warn')[0]
+            else:
+                return _read_mri_surface(fname)
+    raise IOError('No head surface found for subject '
+                  f'{subject} after trying:\n' + '\n'.join(try_fnames))
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -442,11 +442,11 @@ def _check_mri(mri, subject, subjects_dir):
     if not op.isfile(mri):
         if subject is None:
             raise FileNotFoundError(
-                f'MRI file {mri} not found and no subject provided')
+                f'MRI file {mri!r} not found and no subject provided')
         subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
         mri = op.join(subjects_dir, subject, 'mri', mri)
         if not op.isfile(mri):
-            raise FileNotFoundError(f'MRI file {mri} not found')
+            raise FileNotFoundError(f'MRI file {mri!r} not found')
     return mri
 
 
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -541,7 +541,8 @@ def _get_lut(fname=None):
     return lut
 
 
-def get_head_surface(surf, subject, subjects_dir, bem=None):
+@verbose
+def _get_head_surface(surf, subject, subjects_dir, bem=None, verbose=None):
     """Get a head surface from the Freesurfer subject directory.
 
     Parameters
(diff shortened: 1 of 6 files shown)


rob-luke:
I love all the new brain functionality. But I have had some issues trying to use it (not just this PR, all of it), not due to code issues, but finding appropriate examples. @alexrockhill can you add this new functionality to an existing example?

And as a general issue for discussion, would it be appropriate to have a small tutorial that simply works through how to use brain e.g. creating a brain figure, adding some sensors, rotating it, taking a screenshot, marking a region, adding a etc, etc? Maybe this already exists?

alexrockhill:
> I love all the new brain functionality. But I have had some issues trying to use it (not just this PR, all of it), not due to code issues, but finding appropriate examples. @alexrockhill can you add this new functionality to an existing example?
> 
> And as a general issue for discussion, would it be appropriate to have a small tutorial that simply works through how to use brain e.g. creating a brain figure, adding some sensors, rotating it, taking a screenshot, marking a region, adding a stc, etc? Maybe this already exists?

I was going to work on #9594 some time in the near future, maybe that would be a good time to rework a tutorial? Or better yet #8803/#9585.

I can open an issue specifically for the tutorials and we can make sure there is ample time to comment on what should be in it.

rob-luke:
I see no reason not to start building the example in this PR. Start simple with a brain and then .add_head(), and build as you wish.

Thinking further, my understanding is that examples are the more appropriate location for small specialised instructions. So maybe it should go there?

alexrockhill:
> I see no reason not to start building the example in this PR. Start simple with a brain and then .add_head(), and build as you wish.
> 
> Thinking further, my understanding is that examples are the more appropriate location for small specialised instructions. So maybe it should go there?

This is a very small PR so I have no problem adding to it but I think one reason not to start now is that it might be helpful to discuss in the issue how best to implement the example as far as example vs tutorial and what to show so that we can make it easy to find and avoid redundancy.

mne/viz/_brain/_brain.py
```diff
@@ -2316,6 +2316,41 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
             self._labels[hemi].append(label)
         self._renderer._update()
 
+    @fill_doc
+    def add_head(self, color=None, alpha=0.5):
+        """Add labels to the rendering from an anatomical segmentation.
```
agramfort, line 2321: fix docstring

agramfort (commented):
besides LGTM

alexrockhill:
Good to go I think

alexrockhill:
@larsoner okay with you?

mne/viz/_brain/_brain.py
```diff
@@ -2316,6 +2316,41 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
...
+        -----
+        .. versionadded:: 0.24
+        """
+        from matplotlib.colors import colorConverter
+
+        # load seghead
+        seghead_fname = _check_fname(
+            op.join(self._subjects_dir, self._subject_id,
+                    'surf', 'lh.seghead'), overwrite='read', must_exist=True)
+        verts, triangles = read_surface(seghead_fname)
```
larsoner, line 2341: This should be refactored with the code and API in `plot_alignment` to support multiple sources for the head as well as dense/sparse representations, etc. Sometimes people will want the dense surface, sometimes the sparse, and `plot_alignment` has some control over this (including a nice `'auto'` option that should be the default here).

larsoner (changes requested):
You'll notice a theme in my reviews :)

mne/viz/_brain/_brain.py
alexrockhill: So is it worth making an intermediate step where `plot_alignment` and `mne.viz.Brain` share internal functions or should we just make one PR that makes `plot_alignment` a shallow wrapper for `mne.viz.Brain`?

larsoner: Let's just start with refactoring the head stuff. Refactoring it all is a much bigger undertaking. Piecemeal is much easier to review

alexrockhill: Ok one refactoring complete!

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+        if bem is not None:
+            try:
+                return _bem_find_surface(bem, 'head')
+            except RuntimeError:
+                logger.info('Could not find the surface for '
+                            'head in the provided BEM model, '
+                            'looking in the subject directory.')
+    if subject is None:
+        if surf == 'auto':
+            return
```
larsoner, line 578: This does not match the return type in the docstring (to do that, you'd want to return `(None, None)`. It seems like for public function this should raise an error.

This is in a private module, so maybe you want this to stay private? I think it should be private... In that case it's better to (still) have it prefixed by `_` in that case (even though the module also is) because if we ever import it in some public module (like `mne.bem` or something) it will become "accidentally public". So underscores on function names even in `_`-prefixed modules are still useful/helpful. It also makes it immediately clear in the places that it's used that we can change the API if we want.

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
     lut = {d[0]: np.array(lut[d[0]], dtype=d[1]) for d in dtype}
     assert len(lut['name']) > 0
     return lut
+
+
+def get_head_surface(surf, subject, subjects_dir, bem=None):
```
larsoner, line 544: You need a `@fill_doc` decorator here for the `%(subject)s` to fill in. But really you want `*, verbose=None` at the end and `@verbose` because some of the subfunctions call `logger.info`.

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+                      op.join(subject_dir, 'surf', 'lh.seghead')]
+    else:
+        try_fnames = [op.join(subject_dir, 'bem', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', 'flash', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', f'{subject}-head.fif')]
+    for fname in try_fnames:
+        if op.exists(fname):
+            logger.info(f'Using {op.basename(fname)} for head surface.')
+            if op.splitext(fname)[-1] == '.fif':
+                return read_bem_surfaces(fname, on_defects='warn')[0]
```
larsoner, line 596: This returns a `dict` not an `rr, tris` pair

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+    Notes
+    -----
+    .. versionadded: 0.24
+    """
+    _check_option(
+        'surf', surf, ('auto', 'head', 'outer_skin', 'head-dense', 'seghead'))
+    if surf in ('auto', 'head', 'outer_skin'):
+        if bem is not None:
+            try:
+                return _bem_find_surface(bem, 'head')
```
larsoner, line 571: This returns a dict

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+        try_fnames = [op.join(subject_dir, 'bem', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', 'flash', 'outer_skin.surf'),
+                      op.join(subject_dir, 'bem', f'{subject}-head.fif')]
+    for fname in try_fnames:
+        if op.exists(fname):
+            logger.info(f'Using {op.basename(fname)} for head surface.')
+            if op.splitext(fname)[-1] == '.fif':
+                return read_bem_surfaces(fname, on_defects='warn')[0]
+            else:
+                return _read_mri_surface(fname)
```
larsoner, line 598: This returns a dict

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+        or 'seghead'.
+    %(subject)s
+    %(subjects_dir)s
+
+    Returns
+    -------
+    verts : ndarray
+        The surface vertices or None if ``surf='auto'`` and none is found.
+    triangles : ndarray
+        The vertex triangles or None if ``surf='auto'`` and none is found.
```
larsoner, line 560: The return value always appears to be a dict or None, is this outdated?

mne/_freesurfer.py
```diff
@@ -537,3 +539,62 @@ def _get_lut(fname=None):
...
+def get_head_surface(surf, subject, subjects_dir, bem=None):
+    """Get a head surface from the Freesurfer subject directory.
+
+    Parameters
+    ----------
+    surf : str
+        The name of the surface 'auto', 'head', 'outer_skin', 'head-dense'
+        or 'seghead'.
+    %(subject)s
+    %(subjects_dir)s
```
larsoner, line 553: BEM is not documented

mne/_freesurfer.py
larsoner: After seeing all these doc errors, it might be cleaner just to remove the docstring altogether. Docstrings on private functions tend to get outdated rather quickly, so I tend not to use them (and often remove them on private functions) nowadays.

alexrockhill:
Sorry that was a bit sloppy, weird the CIs didn't complain about the docstring omissions.

Ok I cleaned it up but I left the docstring because I think the inputs and return type are not at all obvious and I'd like the future me to understand what's going on without having to execute code.

alexrockhill:
Ready to merge? https://github.com/mne-tools/mne-python/pull/9585 depends on this so it would be nice to get moving

mne/viz/_brain/_brain.py
```diff
@@ -2316,6 +2317,43 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
...
+        """Add a mesh to render the outer head surface.
+
+        Parameters
+        ----------
+        dense : bool
+            Whether to plot the dense head (``seghead``) or the less dense head
+            (``head``).
+        color : matplotlib-style color | None
+            A list of anything matplotlib accepts: string, RGB, hex, etc.
+            (default: "gray").
```
drammock, line 2331: just a note-to-self that we should consider unifying our "matplotlib-compatible color" docstrings, which are rather heterogeneous at the moment (`git grep matplotlib-` for a sampling).  But not necessary for this PR

mne/viz/_brain/_brain.py
```diff
@@ -2316,6 +2317,43 @@ def add_label(self, label, color=None, alpha=1, scalar_thresh=None,
...
+        alpha : float in [0, 1]
+            Alpha level to control opacity.
```
drammock, line 2333: I'm a little surprised this passed our docstring tests.  I think I would have done
```suggestion
        alpha : float
            Opacity of the head surface. Must be between 0 and 1 (inclusive).
            Default is 0.5.
```
because `float in [0, 1]` reads like Python code, confusingly suggesting this param can take on one of 2 values (`0` or `1`)

drammock (approved):
I've suggested a couple of improvements but overall LGTM.

mne/viz/_brain/_brain.py
alexrockhill: Eric wrote that :)

alexrockhill:
Ok good to merge, I think. @larsoner might want to review it if you haven't left already since it was your comments I was addressing

drammock:
> Ok good to merge, I think. @larsoner might want to review it if you haven't left already since it was your comments I was addressing

let's see if he gets around to it in the next ~24 hours.  If not I'll merge before the weekend; I saw what he had requested before and it looks like you addressed it

larsoner:
Thanks @alexrockhill !

PR #9621, opened 2021-07-28
MRG, DOC: Document 'ignore' on_defects value; make allowed on_missing values explicit

hoechenberger:
This is a follow-up to #9614, where @larsoner pointed out that the `on_defects` param should support an `'ignore'` value too. I've made the required adjustments, and also search-and-replaced `on_missing : str` -> `on_missing : 'raise' | 'warn' | 'ignore'` in the entire docdict.

### DIFF
--- a/mne/bem.py
+++ b/mne/bem.py
@@ -267,12 +267,7 @@ def _check_complete_surface(surf, copy=False, incomplete='raise', extra=''):
                'have fewer than three neighboring triangles [{}]{}'
                .format(_bem_surf_name[surf['id']], len(fewer), surf['ntri'],
                        ', '.join(str(f) for f in fewer), extra))
-        if incomplete == 'raise':
-            raise RuntimeError(msg)
-        elif incomplete == 'ignore':
-            pass
-        else:
-            warn(msg)
+        _on_missing(on_missing=incomplete, msg=msg, name='on_defects')
     return surf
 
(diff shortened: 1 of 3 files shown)


mne/bem.py
```diff
@@ -269,6 +269,8 @@ def _check_complete_surface(surf, copy=False, incomplete='raise', extra=''):
                        ', '.join(str(f) for f in fewer), extra))
         if incomplete == 'raise':
             raise RuntimeError(msg)
+        elif incomplete == 'ignore':
+            pass
         else:
             warn(msg)
```
larsoner, line 275: You should use `_on_missing` here

hoechenberger: Ok, done!

larsoner:
Failures are real, maybe just an error class change?

hoechenberger:
Arrrrrgh. Will have a look.

hoechenberger:
@larsoner I think we should be good now. Thanks for your feedback!

hoechenberger:
(Not sure why CI won't start up now?)

hoechenberger:
(oh, webhook incident. well … we'll just have to wait until it's resolved, I suppose)

hoechenberger:
Ok we're green!

agramfort:
Thx @hoechenberger

PR #9622, opened 2021-07-28
[MAINT, MRG] Better transform error that prints coordinate frames

alexrockhill:
I thought it would be helpful to have the allowed coordinate frames for a transform. I also updated to fstring syntax while I was there so that the PR would do a bit more maintenance but that can be reverted if it's not helpful.

### DIFF
--- a/mne/transforms.py
+++ b/mne/transforms.py
@@ -64,7 +64,9 @@ def _to_const(cf):
     """Convert string or int coord frame into int."""
     if isinstance(cf, str):
         if cf not in _str_to_frame:
-            raise ValueError('Unknown cf %s' % cf)
+            raise ValueError(
+                f'Unknown coordinate frame {cf}, '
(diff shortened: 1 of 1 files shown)


alexrockhill:
All the tests pass, good to go I think

mne/transforms.py
```diff
@@ -64,7 +64,9 @@ def _to_const(cf):
     """Convert string or int coord frame into int."""
     if isinstance(cf, str):
         if cf not in _str_to_frame:
-            raise ValueError('Unknown cf %s' % cf)
+            raise ValueError(
+                f'Unknown coordinate frame {cf}, '
+                'expected "' + '", "'.join(_str_to_frame.keys()) + '"')
```
larsoner, line 69: Someday we should just use `_check_option` for this

larsoner (approved):

larsoner:
Thanks @alexrockhill !

PR #9625, opened 2021-07-28
[MAINT, MRG] Refactor BEM code

alexrockhill:
This wasn't nearly as much of a pain as I thought. This simplifies things quite a bit and has the code from developing the GUI that can now be shared in `_freesurfer.py`.

Addresses comments in #9586.

Basically, instead of this very confusing which axis to flip and order of everything, the MRI is just reoriented to RAS and everything is follows from there.

Should be easy to use this code in the GUI, good idea @larsoner.

### DIFF
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -59,6 +59,7 @@ def _import_nibabel(why='use MRI files'):
 
 def _reorient_image(img, axcodes='RAS'):
     """Reorient an image to a given orientation.
+
     Parameters
     ----------
     img : instance of SpatialImage
@@ -66,12 +67,14 @@ def _reorient_image(img, axcodes='RAS'):
     axcodes : tuple | str
         The axis codes specifying the orientation, e.g. "RAS".
         See :func:`nibabel.orientations.aff2axcodes`.
+
     Returns
     -------
     img_data : ndarray
         The reoriented image data.
     vox_ras_t : ndarray
         The new transform from the new voxels to surface RAS.
+
     Notes
     -----
     .. versionadded:: 0.24
@@ -92,10 +95,12 @@ def _reorient_image(img, axcodes='RAS'):
 
 def _mri_orientation(orientation):
     """Get MRI orientation information from an image.
+
     Parameters
     ----------
     orientation : str
         Orientation that you want. Can be "axial", "saggital", or "coronal".
+
     Returns
     -------
     axis : int
@@ -104,6 +109,7 @@ def _mri_orientation(orientation):
         The dimension of the x axis.
     y : int
         The dimension of the y axis.
+
     Notes
     -----
     .. versionadded:: 0.21
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -57,44 +57,62 @@ def _import_nibabel(why='use MRI files'):
     return nib
 
 
-def _mri_orientation(img, orientation):
-    """Get MRI orientation information from an image.
-
+def _reorient_image(img, axcodes='RAS'):
+    """Reorient an image to a given orientation.
     Parameters
     ----------
     img : instance of SpatialImage
         The MRI image.
-    orientation : str
-        Orientation that you want. Can be "axial", "saggital", or "coronal".
-
+    axcodes : tuple | str
+        The axis codes specifying the orientation, e.g. "RAS".
+        See :func:`nibabel.orientations.aff2axcodes`.
     Returns
     -------
-    xyz : tuple, shape (3,)
-        The dimension indices for X, Y, and Z.
-    flips : tuple, shape (3,)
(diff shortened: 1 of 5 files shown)


mne/viz/misc.py
```diff
@@ -559,7 +546,7 @@ def plot_bem(subject=None, subjects_dir=None, orientation='coronal',
 
     # Plot the contours
     return _plot_mri_contours(mri_fname, surfaces, src, orientation, slices,
-                              show, show_indices, show_orientation)[0]
+                              show, show_indices, show_orientation)
```
larsoner, line 549: Did you `git grep _plot_mri_contours` to make sure no other returns need to be changed? IIRC this is shared across BEM plotting and dipole plotting

mne/_freesurfer.py
```diff
@@ -57,44 +57,68 @@ def _import_nibabel(why='use MRI files'):
...
+    .. versionadded:: 0.24
+    """
+    import nibabel as nib
+    orig_data = np.array(img.dataobj).astype(np.float32)
+    # reorient data to RAS
+    ornt = nib.orientations.axcodes2ornt(
+        nib.orientations.aff2axcodes(img.affine)).astype(int)
+    ras_ornt = nib.orientations.axcodes2ornt(axcodes)
+    ornt_trans = nib.orientations.ornt_transform(ornt, ras_ornt)
+    img_data = nib.orientations.apply_orientation(orig_data, ornt_trans)
```
larsoner, line 89: Can you verify in the nibabel code that this just ends up being a view on the data? slices are always views, and I think `.transpose`s are, too (?), and in theory this is all that should be needed and nibabel is generally very well coded so probably does this... but please do check so we can all sleep better at night.

If it doesn't create a view, we should maybe create an issue about enhancing their code to return a view if possible. (This is should be better memory--wise).

larsoner (commented):
Awesome, glad to see that it was so simple!

Since the BEM and GUI code also share a lot of common viz stuff (labeling, surface contours, slice numbers, etc.) hopefully we can continue to refactor that stuff as we go in subsequent PRs. To me the tasks of:

1. look at where this BEM boundary is on the MRI
2. look at where this dipole is in the MRI, and
3. display MRI slices so I can click on them

are all shared *EDIT: orthorview-like* tasks to it would be good to continue to refactor them. IIRC we unified the dipole plotting code with the BEM plotting code years ago and both benefitted from it, so I'm hoping similar advantages can come from the electrode GUI work!

Coming back to this PR, can you touch some examples that use `plot_dipole_locations`, `Report`, and `plot_bem`? That way we can visually verify that the output hasn't changed. IIRC there should be unit tests that yell at you if something is wrong with the plotting, but who knows. I'll see if I can dig up some old issues that led to refactoring before to see if we can visually verify they still work okay...

mne/viz/misc.py
alexrockhill: Yeah I had to change the one in `report.py` but that was it

larsoner:
... also I might be wrong about BEM and dipoles being shared. Looking a bit, that code uses `_plot_dipole_mri_orthoview`. I didn't follow the trail the whole way, though. Maybe there is some more refactoring we can do at some point. But for now I think continuing to refactor BEM + GUI to simplify / enhance both is good!

mne/_freesurfer.py
alexrockhill: https://github.com/nipy/nibabel/blob/62aea04248e70d7c4529954ca41685d7f75a0b1e/nibabel/orientations.py#L154:L172

Looks nothing here makes a copy.

alexrockhill: So I think we're good

alexrockhill:
> ... also I might be wrong about BEM and dipoles being shared. Looking a bit, that code uses `_plot_dipole_mri_orthoview`. I didn't follow the trail the whole way, though. Maybe there is some more refactoring we can do at some point. But for now I think continuing to refactor BEM + GUI to simplify / enhance both is good!

Hmmm yeah I was going to say, `_plot_mri_contours` was only used in report.

larsoner:
Okay just a commit that touches `Report` and `plot_bem` examples would be good, then. Alternatively you can run them locally and verify they haven't changed and report back

alexrockhill:
> Okay just a commit that touches `Report` and `plot_bem` examples would be good, then. Alternatively you can run them locally and verify they haven't changed and report back

Reporting back, they are unchanged.

larsoner (approved):

larsoner:
Ideally we would have some non-standard-orientation MRI to test with, but given that even standard FreeSurfer axes are not ordered RAS and this code generalizes everything, I think we'll be okay. I'll merge once CIs come back happy!

alexrockhill:
Ok good to go!

larsoner:
Thanks @alexrockhill !

PR #9626, opened 2021-07-28
[ENH, MRG] `mne.viz.Brain` visualization tutorial

### LINKED ISSUE #9619: [ENH] Add more complete `mne.viz.Brain` functionality tutorial
Per @rob-luke 's suggestion in #9618, it might be a good idea to explain all the methods of `mne.viz.Brain`.

We have from @rob-luke :
- creating a brain figure
- adding some sensors
- rotating it
- taking a screenshot
- marking a region
- adding a stc

I would add:
- Adding the head
- Removing mesh objects
- Explaining the different views

There are probably many more useful things not on here...

alexrockhill:
Fixes #9619.

Ok, a good start I think but there are a few PRs on deck that would make it more functionally complete. I think once those get done it would be a good time to revisit this and get it merged.

What do you think @rob-luke?

### DIFF
--- a/examples/visualization/brain.py
+++ b/examples/visualization/brain.py
@@ -1,0 +1,104 @@
+"""
+.. _ex-brain:
+
+===============================
+Plotting with ``mne.viz.Brain``
+===============================
+
+In this example, we'll show how to use :class:`mne.viz.Brain`.
+"""
+# Author: Alex Rockhill <aprockhill@mailbox.org>
+#
+# License: BSD-3-Clause
+
+# %%
+# Plot a brain
+# ------------
+#
+# In this example we use the ``sample`` data which is data from a subject
+# being presented auditory and visual stimuli to display the functionality
+# of :class:`mne.viz.Brain` for plotting data on a brain.
+
+import os.path as op
+import matplotlib.pyplot as plt
+
+import mne
+from mne.datasets import sample
+
+print(__doc__)
+
+data_path = sample.data_path()
+subjects_dir = op.join(data_path, 'subjects')
+sample_dir = op.join(data_path, 'MEG', 'sample')
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir,
+                      alpha=0.1, background='white', cortex='low_contrast')
+
+# %%
+# Add source information
+# ----------------------
+#
+# Plot source information.
+stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
+stc.crop(0.09, 0.1)
+
+fmin, fmax = stc.data.min(), stc.data.max()
+brain.add_data(stc.lh_data, hemi='lh', fmin=fmin, fmax=fmax,
+               vertices=stc.lh_vertno, alpha=0.25,
+               smoothing_steps='nearest', time=stc.times)
+brain.add_data(stc.rh_data, hemi='rh', fmin=fmin, fmax=fmax,
+               vertices=stc.rh_vertno, alpha=0.25,
+               smoothing_steps='nearest', time=stc.times)
+
+# %%
+# Modify the view of the brain
+# ----------------------------
+#
+# You can adjust the view of the brain using ``show_view`` method.
+brain.show_view(azimuth=190, elevation=70, distance=350, focalpoint=(0, 0, 20))
+
+# %%
+# Highlight a region on the brain
+# -------------------------------
+#
+# It can be useful to highlight a region of the brain for analyses.
+# To highlight a region on the brain you can use the ``add_label`` method.
+# Labels are stored in the Freesurfer label directory from the ``recon-all``
+# for that subject. Labels can also be made following the
+# `Freesurfer instructions
+# <https://surfer.nmr.mgh.harvard.edu/fswiki/mri_vol2label>`_
+# Here we will show Brodmann Area 44.
+#
+# .. note:: The MNE sample dataset contains only a subselection of the
+#           Freesurfer labels created during the ``recon-all``.
+brain.add_label('BA44', hemi='lh', color='green', borders=True)
+
+# %%
+# Include the head in the image
+# -----------------------------
+#
+# Add a head image using the ``add_head`` method.
+brain.add_head(alpha=0.5)
+
+# %%
+# Add sensors positions
+# ---------------------
+#
+# To put into context the data that generated the source time course,
+# the sensor positions can be displayed as well.
+evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
+trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
+brain.add_sensors(evoked.info, trans)
+brain.show_view(distance=500)  # move back to show sensors
+
+# %%
+# Create a screenshot for exporting the brain image
+# -------------------------------------------------
+#
+# For publication you may wish to take a static image of the brain,
+# for this use ``screenshot``.
+img = brain.screenshot()
+fig, ax = plt.subplots()
+ax.imshow(img)
+ax.axis('off')
+fig.suptitle('Brain')
--- a/examples/visualization/brain.py
+++ b/examples/visualization/brain.py
@@ -31,30 +31,30 @@ data_path = sample.data_path()
 subjects_dir = op.join(data_path, 'subjects')
 sample_dir = op.join(data_path, 'MEG', 'sample')
 
-brain = mne.viz.Brain('sample', subjects_dir=subjects_dir,
-                      alpha=0.1, background='white', cortex='low_contrast')
-
 # %%
 # Add source information
 # ----------------------
 #
 # Plot source information.
+
+brain_kwargs = dict(alpha=0.1, background='white', cortex='low_contrast')
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
+
 stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
 stc.crop(0.09, 0.1)
 
-fmin, fmax = stc.data.min(), stc.data.max()
-brain.add_data(stc.lh_data, hemi='lh', fmin=fmin, fmax=fmax,
-               vertices=stc.lh_vertno, alpha=0.25,
-               smoothing_steps='nearest', time=stc.times)
-brain.add_data(stc.rh_data, hemi='rh', fmin=fmin, fmax=fmax,
-               vertices=stc.rh_vertno, alpha=0.25,
-               smoothing_steps='nearest', time=stc.times)
+kwargs = dict(fmin=stc.data.min(), fmax=stc.data.max(), alpha=0.25,
+              smoothing_steps='nearest', time=stc.times)
+brain.add_data(stc.lh_data, hemi='lh', vertices=stc.lh_vertno, **kwargs)
+brain.add_data(stc.rh_data, hemi='rh', vertices=stc.rh_vertno, **kwargs)
 
 # %%
 # Modify the view of the brain
 # ----------------------------
 #
 # You can adjust the view of the brain using ``show_view`` method.
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
 brain.show_view(azimuth=190, elevation=70, distance=350, focalpoint=(0, 0, 20))
 
 # %%
@@ -71,13 +71,18 @@ brain.show_view(azimuth=190, elevation=70, distance=350, focalpoint=(0, 0, 20))
 #
 # .. note:: The MNE sample dataset contains only a subselection of the
 #           Freesurfer labels created during the ``recon-all``.
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
 brain.add_label('BA44', hemi='lh', color='green', borders=True)
+brain.show_view(azimuth=190, elevation=70, distance=350, focalpoint=(0, 0, 20))
 
 # %%
 # Include the head in the image
 # -----------------------------
 #
 # Add a head image using the ``add_head`` method.
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
 brain.add_head(alpha=0.5)
 
 # %%
@@ -86,6 +91,8 @@ brain.add_head(alpha=0.5)
 #
 # To put into context the data that generated the source time course,
 # the sensor positions can be displayed as well.
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
 evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
 trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
 brain.add_sensors(evoked.info, trans)
@@ -97,6 +104,8 @@ brain.show_view(distance=500)  # move back to show sensors
 #
 # For publication you may wish to take a static image of the brain,
 # for this use ``screenshot``.
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, **brain_kwargs)
 img = brain.screenshot()
 fig, ax = plt.subplots()
 ax.imshow(img)


rob-luke:
This is great! I am going to use this PR as a dumping ground for my confusions, so please don't take any comments and questions as being directed at you. I just want to list things that confuse me, so we can add links and pointers for others that might have the same questions (and maybe somethings will require separate PRs).

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+data_path = sample.data_path()
+sample_dir = op.join(data_path, 'MEG', 'sample')
+subjects_dir = op.join(data_path, 'subjects')
+evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
+trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
+stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
+
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir,
+                      alpha=0.5, cortex='low_contrast')
```
rob-luke, line 32:
```suggestion
# %%
# Plot a brain image
# ------------------
#
# First we plot a brain. This requires a directory with MRI? information.
# MNE provides functionality to automatically retrieve standard head images.
# In this example we use the ``sample`` data which is?
# Another common head used in the MNE documentation is fsaverage, which is?

data_path = sample.data_path()
subjects_dir = op.join(data_path, 'subjects')

brain = mne.viz.Brain('sample', subjects_dir=subjects_dir,
                      alpha=0.5, cortex='low_contrast')
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+# add sensors
+# brain.add_sensors(
+#     epochs.info, trans, colors=dict(eeg='green', grad='blue', mag='purple'),
+#     ch_types=dict(meg=True, eeg=True))
```
rob-luke, line 37:
```suggestion
# %%
# View sensors and brain
# ----------------------
#
# To view your sensors on the brain you can use the ``add_sensors`` method.
# This can be useful to validate your data has been
# stored and retrieved correctly.

# add sensors
# brain.add_sensors(
#     epochs.info, trans, colors=dict(eeg='green', grad='blue', mag='purple'),
#     ch_types=dict(meg=True, eeg=True))
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+# mark a region
+brain.add_label('BA3a', hemi='lh')
```
rob-luke, line 40:
```suggestion
# %%
# Highlight a region on the brain
# -------------------------------
#
# To highlight a region on the brain you can use the ``add_label`` method.
# Where do we get these names from?
# Is there an MNE tutorial listing these names in a table?
# How to I understand what BA3a is?
# Is there an external website that exaplains these acronyms?

# mark a region
brain.add_label('BA3a', hemi='lh')
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+# rotate brain by setting the camera position
+brain.show_view(azimuth=50, elevation=80, distance=250)
```
rob-luke, line 43:
```suggestion
# %%
# Modify the view of the brain
# ----------------------------
#
# You can adjust the view of the brain using ``show_view`` method.

brain.show_view(azimuth=50, elevation=80, distance=250)
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+# add a head
+# brain.add_head(color='red', alpha=0.1)
```
rob-luke, line 46:
```suggestion
# %%
# Include the head in the image
# -----------------------------
#
# Add a head image using the ``add_head`` method.

# add a head
# brain.add_head(color='red', alpha=0.1)
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,65 @@
...
+
+# take a screenshot
+img = brain.screenshot()
+fig, ax = plt.subplots()
+ax.imshow(img)
+ax.axis('off')
+fig.suptitle('Brain')
```
rob-luke, line 53:
```suggestion
# %%
# Create a screenshot for exporting the brain image
# -------------------------------------------------
#
# For publication you may wish to take a static image of the brain,
# for this use ``screenshot``.


# take a screenshot
img = brain.screenshot()
fig, ax = plt.subplots()
ax.imshow(img)
ax.axis('off')
fig.suptitle('Brain')

File: examples/visualization/brain.py [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+# add an stc
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir)
+
+fmin, fmax = stc.data.min(), stc.data.max()
+for i, hemi in enumerate(('lh', 'rh')):
+    hemi_data = stc.data[:len(stc.vertices[i])]
+    hemi_vertices = stc.vertices[i]
+    brain.add_data(hemi_data, fmin=fmin, hemi=hemi, fmax=fmax,
+                   vertices=hemi_vertices,
+                   smoothing_steps='nearest', time=stc.times)
```
rob-luke, line 65:
```suggestion
                   smoothing_steps='nearest', time=stc.times)

brain.show_view(azimuth=20, elevation=70, distance=500)```

File: examples/visualization/brain.py [hunk 8]
```diff
@@ -0,0 +1,65 @@
...
+# take a screenshot
+img = brain.screenshot()
+fig, ax = plt.subplots()
+ax.imshow(img)
+ax.axis('off')
+fig.suptitle('Brain')
+
+# %%
+# add an stc
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir)
```
rob-luke, line 57:
```suggestion
# %%
# Add source information
# ----------------------
#
# Plot source information.

sample_dir = op.join(data_path, 'MEG', 'sample')
evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
stc.crop(0.09, 0.1)

brain = mne.viz.Brain('sample', subjects_dir=subjects_dir, background='w')
```

examples/visualization/brain.py
rob-luke: I pushed the evoked and stc imports below as they may just confuse people who wish to just plot a brain. I think keep it simple, and then build up complexity.

examples/visualization/brain.py
rob-luke: The previous distance setting cropped out much of the brain.

examples/visualization/brain.py
rob-luke: Why cant we build on the brain from above? It would be nice to show an image at the end that has the brain, sensors, region of interest, and source analysis all in one figure. Thoughts?

examples/visualization/brain.py
rob-luke: And the suggested az and el make the final stc image look good

examples/visualization/brain.py
rob-luke, line 65:
```suggestion
                   smoothing_steps='nearest', time=stc.times)


# %%
# Highlight a region on the brain
# -------------------------------
#
# It can be useful to highlight a region of the brain to visually identify
# if the source activity is in an expected region.
# To highlight a region on the brain you can use the ``add_label`` method.
# Where do we get these names from?
# Is there an MNE tutorial listing these names in a table?
# How to I understand what BA3a is?
# Is there an external website that exaplains these acronyms?

# mark a region
brain.add_label('BA44', hemi='lh', color='green', borders=True)                   
```

examples/visualization/brain.py
rob-luke, line 40: I suggest moving this to the end. See below

examples/visualization/brain.py
rob-luke: I think this might be more appropriate to run at the end, so we can demonstrate how the region of interest is overlaid on the source results.

GuillaumeFavelier (approved):
Very nice initiative! I think it's a good opportunity to showcase what `Brain` is capable of :+1:

alexrockhill:
Thanks for the review @rob-luke!

examples/visualization/brain.py
alexrockhill: I don't think you can show the an ROI colored in with an stc, they are both colorings of parts of the brain...

alexrockhill: Oh nevermind, I forgot about borders!

drammock: yep.  Also I think the color used in `brain.add_label` can have an alpha, so depending on your STC's colormap it might sometimes even make sense to add a filled label on top of an STC.

alexrockhill:
Hmmm not sure what's changed and is causing the test failure... I'll have to look into it but it runs locally just fine

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+brain.show_view(azimuth=190, elevation=70, distance=500)
+
+
+# %%
+# Highlight a region on the brain
+# -------------------------------
+#
+# It can be useful to highlight a region of the brain for analyses.
+# To highlight a region on the brain you can use the ``add_label`` method.
+# Labels are stored in the Freesurfer label directory for that subject.
```
rob-luke, line 71: This is still unclear to me as a new user. Do I need to then install freesurfer to have that directory? Maybe adding the default location or link to another page discussing this?

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+# Here we will show Brodmann Area 44.
```
rob-luke, line 72: Are all Brodman areas available? Or just a subset?

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+#
+# Plot source information.
+sample_dir = op.join(data_path, 'MEG', 'sample')
+evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
+trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
+stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
+stc.crop(0.09, 0.1)
+
+fmin, fmax = stc.data.min(), stc.data.max()
+for i, hemi in enumerate(('lh', 'rh')):
```
rob-luke, line 49: General curiosity question here. This seems the trickiest part of working with brain. Are there plans to simplify this? I.e. add a brain.add_stc method or something?

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+import matplotlib.pyplot as plt
+
+import mne
+from mne.datasets import sample
+
+print(__doc__)
+
+data_path = sample.data_path()
+subjects_dir = op.join(data_path, 'subjects')
+brain = mne.viz.Brain('sample', subjects_dir=subjects_dir,
```
rob-luke, line 33: What is sample? Is there an existing MNE doc page we can link to that describes this?

The tutorials seem to switch between fsaverage and sample. But I can't figure out why some use one and some use another. I think this is confusing and make copying and pasting parts from multiple tutorials risky.

Are the two terms interchangeable, but you just need to be consistent? Can we add some text to that effect?

alexrockhill: `sample` is the subject from the sample dataset in MNE, `fsaverage` is a template space that can be linearly registered to using the Talairach transform or non-linearly registered to using Freesurfer's combined surface volumetric warp or symmetric diffeomorphic registration (that's what MNE uses) see https://mne.tools/dev/auto_tutorials/clinical/10_ieeg_localize.html for some more explanation

examples/visualization/brain.py
alexrockhill: I haven't touched it and it's not really what I'm working on right now but seems easy enough and could definitely be refactored. Just my opinion but `add_data` seems overly general/vague. Although I know stc data is in the brain space, it is not equivalent to the brain, just a more or less sparse sampling, that's just my two cents though

examples/visualization/brain.py
alexrockhill: In this example just a few because it's the sample data and no one wants all the bells and whistles and having to wait 15 minutes for it to download. In general for your own freesurfer recon, they should all be there I think

examples/visualization/brain.py
alexrockhill: They come with the recon, we could mention that. You can also make labels however you wish, which is described in the freesurfer documentation https://surfer.nmr.mgh.harvard.edu/fswiki/mri_vol2label

examples/visualization/brain.py
rob-luke: I think we need to explicitly say this then. As a user the first thing I did was run this example and change it to Brodman 41 and it errors.

I've been using MNE for a year now and have no experience with freesurfer. I don't think knowledge of freesurfer should be assumed. And if it's required for using Brain then we should add the required explanation or links here.

examples/visualization/brain.py
rob-luke: Thanks. This tutorial already greatly helped me. I will try this little snippet locally. Appreciate it!

examples/visualization/brain.py
rob-luke: I think as this is the Brain explain location we should add a brief explanation here. Or link to a specific subsection in other page.

rob-luke: > that's what MNE uses

If MNE uses fsaverage then why does this tutorial use something else? As this is a learning resource I think it should demonstrate the canonical usage. And I'd guess most users don't have their own MRI scans.

examples/visualization/brain.py
rob-luke: I.e the average user will just try and use their favourite Bradman area. When it doesn't work they will assume it's an MNE error. Not some complex freesurfer usage issue

If it's not expected to work then we need to spell that out in the tutorial.

examples/visualization/brain.py
alexrockhill: MNE uses symmetric diffeomorphic registration to register `sample` to `fsaverage` to take an individual subject's scan and morph it to a template (`fsaverage`).

So to answer your question MNE uses `sample` for individual-level results and `fsaverage` for group results in general (there may be exceptions to this).

rob-luke:
Thanks @alexrockhill and @GuillaumeFavelier. My questions come from genuine curiosity, not implying there is anything lacking from the code.

I've been analysing EEG and FNIRS for 10 years and not used freesurfer. I think there is a lot of assumed knowledge within this area of MNE. I hope by voicing what I don't understand that we can clarify it for other new users. Please feel free to point me (and in the example) to other relevant resources.

examples/visualization/brain.py
rob-luke: Wonderful. That's a great insight. I'm sure there will be exceptions to the rule of thumb. But great to know.

rob-luke: As a final suggestion then. Should we use fsaverage for the example then? As most users would not have individual subject MRIs. And then a final section to the tutorial that demonstrates individual subject scans are simply a case of changing to `sample`?

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+"""
+# Author: Alex Rockhill <aprockhill@mailbox.org>
+#
+# License: BSD-3-Clause
+
+# %%
+# Plot a brain image
+# ------------------
+#
+# First we plot a brain. This requires a directory with MRI? information.
```
agramfort, line 18:
```suggestion
# First we plot a brain. This requires a directory with MRI information.
```

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+# MNE provides functionality to automatically retrieve standard head images.
+# In this example we use the ``sample`` data which is data from an example
+# subject being shown auditory and visual stimuli.
```
agramfort, line 21: showing an auditory stimulus sounds weird to me

examples/visualization/brain.py
```diff
@@ -0,0 +1,106 @@
...
+# Plot source information.
+sample_dir = op.join(data_path, 'MEG', 'sample')
+evoked = mne.read_evokeds(op.join(sample_dir, 'sample_audvis-ave.fif'))[0]
+trans = mne.read_trans(op.join(sample_dir, 'sample_audvis_raw-trans.fif'))
+stc = mne.read_source_estimate(op.join(sample_dir, 'sample_audvis-meg'))
+stc.crop(0.09, 0.1)
+
+fmin, fmax = stc.data.min(), stc.data.max()
+for i, hemi in enumerate(('lh', 'rh')):
+    hemi_data = stc.data[:len(stc.vertices[i])]
```
agramfort, line 50: I doubt this is correct as this will always select the left hemisphere time courses

examples/visualization/brain.py
wmvanvliet: If you want to perform source estimation, you need an accurate head model, hence you need the subject's MRI. While it is technically possible to use `fsaverage` instead, this gives poor results. IMO, we should carry the message in the tutorials that when you are doing source estimation and visualizing it and stuff, **of course** you have the individual's MRI at hand. Not having it should be an absolute exception.

alexrockhill: Hmmm I've heard from Mark George's group at MUSC that this worked pretty well for TMS-EEG localization comparing known subject-specific MRI to template but I haven't tried first hand... We have fsaverage and sample's MRI maybe we should plot both and comment on how good or bad it is and maybe cite relevant sources

examples/visualization/brain.py
alexrockhill: What should we show?

examples/visualization/brain.py
alexrockhill: Good catch

examples/visualization/brain.py
rob-luke: "presented"

alexrockhill: Oh my bad, I thought he meant change the example data

alexrockhill:
Does anyone know with what's going on with Circle here? I'm pretty confused, it works locally

larsoner:
> Does anyone know with what's going on with Circle here? I'm pretty confused, it works locally

Locally doing:
```
$ PATTERN=brain.py make html_dev-pattern
```
I can replicate a problem:
```
WARNING: /home/larsoner/python/mne-python/examples/visualization/brain.py failed to execute correctly: Traceback (most recent call last):                                           
  File "/home/larsoner/python/mne-python/examples/visualization/brain.py", line 46, in <module>
    brain.add_data(stc.lh_data, hemi='lh', fmin=fmin, fmax=fmax,
  File "<decorator-gen-181>", line 24, in add_data
  File "/home/larsoner/python/mne-python/mne/viz/_brain/_brain.py", line 1937, in add_data
    self._data['vector_alpha'] = vector_alpha
TypeError: 'NoneType' object does not support item assignment
```
The way sphinx-gallery scrapers in general (including our [Brain scraper](https://github.com/mne-tools/mne-python/blob/main/mne/viz/_brain/_scraper.py)) work is that, after each code block, *the given visualization object is closed*. So when our scraper does its job, it closes the object.

An easy solution is to recreate a new `Brain` in each code block. This has the disadvantage of code duplication, but the advantage for example that running in notebook mode you'd end up with a new brain for each cell, which is probably what you want / more standard anyway.

Another easy solution is to use [`# sphinx_gallery_defer_figures = True`](https://sphinx-gallery.github.io/stable/configuration.html#defer-figures), but then you only get the brain shown at the end, and probably isn't that useful here.

A difficult solution is to modify the scraper to support not closing until the end. We can add `# sphinx_gallery_do_not_close` for example, then [look for this tag when scraping](). You can try this, but I think we will then have a problem where PyVista will do *its* scraping pass, and still find a plotter in its private store of all figures, so we'd have to sort out how to deal with that...

If you're more interested in the difficult solution, feel free to give it a shot, or I can see if I can figure out a way to do it without too much pain.

alexrockhill:
Ah that makes sense, thanks. I was just running it in a console locally. Perhaps the error should be more informative with a `_check_if_closed` function as well.

I'll look at refactoring.

agramfort (approved):
looks good https://34266-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/brain.html

thx @alexrockhill !

PR #9628, opened 2021-07-30
Fixes #9611 - Indexing error in Evoked.plot_topomap() when trying to mask channels.

### LINKED ISSUE #9611: Indexing error in Evoked.plot_topomap() when trying to mask channels
So either this is a documentation problem or a code bug: I supply `times` and `mask` with a shape of `(n_times, n_chans)` to `Evoked.plot_topomap()`, and it crashes with an `IndexError`.

```python
# %%
from pathlib import Path
import numpy as np

import mne

sample_data_dir = Path(mne.datasets.sample.data_path())
fname = sample_data_dir / 'MEG' / 'sample' / 'sample_audvis-ave.fif'

condition = 'Left Auditory'
evoked = mne.read_evokeds(fname=fname, condition=condition)
evoked.pick('mag')

times = (0.090, 0.110)
mask = np.empty(
    [len(evoked.ch_names), len(times)]
)
mask.fill(False)
mask[20, :] = True

fig = evoked.plot_topomap(times=times, time_unit='s', mask=mask)
```

```python
  File "/private/tmp/mwe_topo.py", line 22, in <module>
    fig = evoked.plot_topomap(times=times, time_unit='s', mask=mask)
  File "/Users/hoechenberger/Development/mne-python/mne/evoked.py", line 478, in plot_topomap
    return plot_evoked_topomap(
  File "/Users/hoechenberger/Development/mne-python/mne/viz/topomap.py", line 1697, in plot_evoked_topomap
    mask_ = mask[np.ix_(picks, time_idx)]
IndexError: index 174 is out of bounds for axis 1 with size 2
```

I expected a topoplot at the two time points, showing only the first 20 channels.

mscheltienne:
#### Reference issue
Fixes #9611 - Indexing error in Evoked.plot_topomap() when trying to mask channels.

Fixes the IndexError raised and the failing bit-wise inversion.

#### Additional information
The bitwise OR for gradiometer sensors is failing, but as I don't know what the purpose of the selection of indices even vs odd is, I did not fix it yet.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -33,6 +33,8 @@ Current (0.24.dev0)
 
 .. |Darin Erat Sleiter| replace:: **Darin Erat Sleiter**
 
+.. |Mathieu Scheltienne| replace:: **Mathieu Scheltienne**
+
 Enhancements
 ~~~~~~~~~~~~
 .. - Add something cool (:gh:`9192` **by new contributor** |New Contributor|_)
@@ -131,6 +133,8 @@ Bugs
 
 - Fix bug with `mne.io.read_raw_curry` to allow reading Curry 8 event files with '.cdt.ceo' extension (:gh:`9381` by **new contributor** |Xiaokai Xia|_ and `Daniel McCloy`_)
 
+- Fix bug in :func:`mne.viz.plot_topomap` and :func:`mne.viz.plot_evoked_topomap` where the use of the argument ``mask`` would lead to an error (:gh:`9628` **by new contributor** |Mathieu Scheltienne|_)
+
 - Fix bug with :func:`mne.io.read_raw_nihon` where latin-1 annotations could not be read (:gh:`9384` by `Alex Gramfort`_)
 
 - Fix bug when printing a :class:`mne.io.RawArray` in the notebook (:gh:`9404` by `Alex Gramfort`_)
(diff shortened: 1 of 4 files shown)


agramfort:
@mscheltienne can you add an entry in the file latest.inc in the doc folder? 🙏 
can you update possibly a test or touch an example that uses this code to eye ball the results? 🙏

mscheltienne:
@agramfort I added a bullet point in `latest.inc`, based on the others. Let me know if I did not respect the correct syntax.

For the test/example; I am not sure about what you expect. I am not (yet) familiar with the GitHub workflow, and I don't really know what are the steps to take to correctly contribute.

Anyway, this PR still need an additional fix for the gradiometers. Currently, the bitwise OR will raise an error; and as I do not know why the gradiometer sensors are treated differently; and with a difference between the odd and even pick indices; I don't know yet how to fix this. A similar comment is available on the original issue.

```
    if mask is not None:
        if ch_type == 'grad':
            mask_ = (mask[np.ix_(picks[::2], range(len(time_idx)))] |
                           mask[np.ix_(picks[1::2], range(len(time_idx)))])
        else:  # mag, eeg, planar1, planar2
            mask_ = mask[np.ix_(picks, range(len(time_idx)))]
```

mscheltienne:
Guessing the test failed due to the markdown syntax around the argument name `mask`? How can I know what failed the automatic tests?

agramfort:
should be ok now @mscheltienne 

have a look at the commit I just pushed.

mscheltienne:
@agramfort Ok, thank you, I understand now a bit better how the testing works. Should we add also a part for the gradiometers in the test as the mask is used differently, e.g. 

```
    evoked = read_evokeds(evoked_fname)[0]
    evoked.pick('grad')

    times = (0.090, 0.110)
    mask = np.zeros([len(evoked.ch_names), len(times)], dtype=bool)
    mask[:20, :] = True
    evoked.plot_topomap(times=times, time_unit='s', mask=mask)
```

For some reason, gradiometers are working fine now, Not really sure why the bitwise OR was failing earlier. Examples with `mask[20, :]` and `mask[:20, :]`.

image

mscheltienne:
Correction, the bitwise OR was crashing earlier because @hoechenberger used `mask = np.empty([len(evoked.ch_names), len(times)])` while the added test uses `mask = np.zeros([len(evoked.ch_names), len(times)], dtype=bool)`.

mmagnuski:
@agramfort take a look at the dicsussion in https://github.com/mne-tools/mne-python/issues/9611. Currently the code expects mask of `(n_channels, n_samples)`, which is ok (no need to change it). The main issue is that the docs are not clear enough about it.

mscheltienne:
The bitwise inversion I removed shouldn't have been removed. I revert the change and added `.astype(bool)` to make sure the inversion works, e.g. with float array of `0.` and `1.`.

mscheltienne:
Closing as this is not the proper fix for this issue.

PR #9630, opened 2021-07-30
[BUG, MRG] fix coordinate frame in fNIRS bug

### LINKED ISSUE #9627: [BUG] `mne.viz.plot_alignment` silently assumes unknown coordinate frame is in head
@larsoner suggested warning that the coordinate frame is unknown might be the best option but as in this example, an unknown coordinate frame is silently accepted as being in "head" coordinates:

```
import os.path as op
import numpy as np
import mne
data = np.random.normal(size=(16, 100))
ch_names = ['S1_D1 hbo', 'S1_D1 hbr', 'S2_D1 hbo', 'S2_D1 hbr',
            'S3_D1 hbo', 'S3_D1 hbr', 'S4_D1 hbo', 'S4_D1 hbr',
            'S5_D2 hbo', 'S5_D2 hbr', 'S6_D2 hbo', 'S6_D2 hbr',
            'S7_D2 hbo', 'S7_D2 hbr', 'S8_D2 hbo', 'S8_D2 hbr']
ch_types = ['hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr',
            'hbo', 'hbr', 'hbo', 'hbr']
sfreq = 10.  # in Hz
info = mne.create_info(ch_names=ch_names, ch_types=ch_types, sfreq=sfreq)
raw = mne.io.RawArray(data, info, verbose=True)

subjects_dir = mne.datasets.sample.data_path() + '/subjects'
mne.datasets.fetch_fsaverage(subjects_dir=subjects_dir)

montage = mne.channels.make_standard_montage('artinis-octamon')
raw.set_montage(montage)
trans = mne.channels.compute_native_head_t(montage)

fig = mne.viz.create_3d_figure(size=(800, 600), bgcolor='white')
fig = mne.viz.plot_alignment(
    raw.info, trans=trans, subject='fsaverage', subjects_dir=subjects_dir,
    surfaces=['brain', 'head'], coord_frame='mri', dig=True, show_axes=True,
    fnirs=['channels', 'pairs', 'sources', 'detectors'], fig=fig)
print({ch: raw.info['chs'][i]['coord_frame'] for i, ch in enumerate(raw.ch_names)})
```

Adapted from https://mne.tools/stable/auto_tutorials/io/30_reading_fnirs_data.html

alexrockhill:
Fixes #9627.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -175,6 +175,8 @@ Bugs
 
 - Fix bug in :func:`mne.viz.set_3d_backend` and :func:`mne.viz.get_3d_backend` where the PyVistaQt-based backend was ambiguously named ``'pyvista'`` instead of ``'pyvistaqt'``; use ``set_3d_backend('pyvistaqt')`` and expect ``'pyvistaqt'`` as the output of :func:`mne.viz.get_3d_backend` instead of ``'pyvista'``, and consider using ``get_3d_backend().startswith('pyvista')`` for example for backward-compatible conditionals (:gh:`9607` by `Guillaume Favelier`_)
 
+- Fix bug where setting of a montage with fNIRS data got set to "unknown" coordinate frame when it should have been in "head" (:gh:`9629` by `Alex Rockhill`_)
+
 API changes
 ~~~~~~~~~~~
 - In `mne.compute_source_morph`, the ``niter_affine`` and ``niter_sdr`` parameters have been replaced by ``niter`` and ``pipeline`` parameters for more consistent and finer-grained control of registration/warping steps and iteration (:gh:`9505` by `Alex Rockhill`_ and `Eric Larson`_)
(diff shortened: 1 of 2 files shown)


alexrockhill:
Look okay to you @rob-luke?

rob-luke:
Does this require that a user converts their montage to head frame in advance? Or is this done in the upstream function before this internal function is called?

I'm imagining the use case where your optode positions have been specified in MRI frame. How should a user load that montage?

Perhaps we should make it an option for users to specify the frame? Or if not, we should make it clear in the documentation that you must use head frame.

But I'm still learning about coordinatede frames. And I would defer to @larsoner and @agramfort opinions here.

alexrockhill:
> Does this require that a user converts their montage to head frame in advance? Or is this done in the upstream function before this internal function is called?
> 
> I'm imagining the use case where your optode positions have been specified in MRI frame. How should a user load that montage?
> 
> Perhaps we should make it an option for users to specify the frame? Or if not, we should make it clear in the documentation that you must use head frame.
> 
> But I'm still learning about coordinatede frames. And I would defer to @larsoner and @agramfort opinions here.

They are converted to head in `mne.channels.montage._set_montage`.

rob-luke:
Thanks for clarifying. The fNIRS IO page isn't rendering with the example usage of set montage, but i assume it looks identical to before?

alexrockhill:
Yes, should be identical

agramfort (approved):

larsoner:
Thanks @alexrockhill

PR #9634, opened 2021-08-01
update installation recommendations

### LINKED ISSUE #9623: RFC: installation process instructions
Based on some offline discussions with users, it may be time for some refinements to the install instructions. Specifically, we should make it clear that:

1. if they want to **update** mne-python, it's usually best to just `pip install -U mne`
2. if they want to update *everything in their environment* they should create a new environment instead of using the latest `environment.yml` to update an existing environment
3. (optional) given item 2, perhaps a "pro tip" would be to keep a list of other packages one installs into one's mne environment, to simplify the creation of new mne envs
4. consider revising our `environment.yml` to *not include* an env name, and update the recommended install command to `conda create --name mne --file environment.yml` instead.  The rationale here is that the current recommended command `conda env update --file environment.yml` implies that it can be used for updating

cc @NeuroLaunch @larsoner was there anything else from our discussion that I've forgotten to include here?

drammock:
closes #9623

- [x] add instructions about *updating* envs
- [x] add "pro tips" like using mamba, using different envs for each project, keeping a list of other packages you install (to make it easier to recreate envs), etc.

### DIFF
--- a/doc/install/advanced.rst
+++ b/doc/install/advanced.rst
@@ -100,7 +100,19 @@ starts off with a minimal set of around 30 packages instead of Anaconda's
 hundreds. See the `installation instructions for Miniconda`_ for more info.
 A similar alternative is `MiniForge`_, which uses the ``conda-forge`` channel
 as the default source for package installation (saving you the trouble of
-typing ``--channel conda-forge`` with each ``conda install`` command).
+typing ``--channel=conda-forge`` with each ``conda install`` command).
+
+.. warning::
+
+    If you have the ``PYTHONPATH`` or ``PYTHONHOME`` environment variables set,
+    you may run into difficulty using Anaconda. See the
+    `Anaconda troubleshooting guide`_ for more information. Note that it is
+    easy to switch between ``conda``-managed Python installations and the
+    system Python installation using the ``conda activate`` and ``conda
+    deactivate`` commands, so you may find that after adopting Anaconda it is
+    possible (indeed, preferable) to leave ``PYTHONPATH`` and ``PYTHONHOME``
+    permanently unset.
+
 
 It is also possible to use a system-level installation of Python (version
 |min_python_version| or higher) and use ``pip`` to install MNE-Python and its
--- a/doc/install/advanced.rst
+++ b/doc/install/advanced.rst
@@ -98,6 +98,9 @@ like Anaconda, Miniconda includes the ``conda`` command line tool for
 installing new packages and managing environments; unlike Anaconda, Miniconda
 starts off with a minimal set of around 30 packages instead of Anaconda's
 hundreds. See the `installation instructions for Miniconda`_ for more info.
+A similar alternative is `MiniForge`_, which uses the ``conda-forge`` channel
+as the default source for package installation (saving you the trouble of
+typing ``--channel conda-forge`` with each ``conda install`` command).
 
 It is also possible to use a system-level installation of Python (version
 |min_python_version| or higher) and use ``pip`` to install MNE-Python and its
--- a/doc/install/advanced.rst
+++ b/doc/install/advanced.rst
@@ -77,29 +77,14 @@ The notebook 3d backend requires PyVista to be installed along with other packag
 please follow :doc:`mne_python`.
 
 
-.. _installing_main:
-
 Using the development version
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-If you want access to the latest features and bugfixes, you can easily switch
-from the stable version of MNE-Python to the current development version.
+See :ref:`installing_main` for how to do a one-time update to the latest
+development version of MNE-Python. If you plan to contribute to MNE-Python, or
+just prefer to use git rather than pip to make frequent updates, there are
+instructions for installing from a ``git clone`` in the :ref:`contributing`.
 
-.. warning:: In between releases, function and class APIs can change without
-             warning.
-
-For a one-time update to latest main, make sure you're in the conda
-environment where MNE-Python is installed (if you followed the default install
-instructions, this will be ``mne``), and use ``pip`` to upgrade:
-
-.. code-block:: console
-
-   $ conda activate name_of_my_mne_environment
-   $ pip install --upgrade --no-deps https://github.com/mne-tools/mne-python/archive/main.zip
-
-If you plan to contribute to MNE-Python, or just prefer to use git rather than
-pip to make frequent updates, there are instructions for installing from a
-``git clone`` in the :ref:`contributing`.
 
 .. _other-py-distros:
 
--- a/doc/install/advanced.rst
+++ b/doc/install/advanced.rst
@@ -114,9 +114,9 @@ installing new packages and managing environments; unlike Anaconda, Miniconda
 starts off with a minimal set of around 30 packages instead of Anaconda's
 hundreds. See the `installation instructions for Miniconda`_ for more info.
 
-It is also possible to use a system-level installation of Python (version 3.6
-or higher) and use ``pip`` to install MNE-Python and its dependencies, using
-the provided `requirements file`_:
+It is also possible to use a system-level installation of Python (version
+|min_python_version| or higher) and use ``pip`` to install MNE-Python and its
+dependencies, using the provided `requirements file`_:
 
 .. code-block:: console
 
--- a/doc/install/index.rst
+++ b/doc/install/index.rst
@@ -1,7 +1,9 @@
+.. include:: links.inc
+
 Quick start
 ===========
 
-MNE-Python requires Python version 3.7 or higher.
+MNE-Python requires Python version |min_python_version| or higher.
 For users already familiar with Python:
 
 - If you only need 2D plotting capabilities with MNE-Python (i.e., most EEG/ERP
--- a/doc/install/index.rst
+++ b/doc/install/index.rst
@@ -1,4 +1,4 @@
-.. include:: links.inc
+.. include:: ../links.inc
 
 Quick start
 ===========
--- a/doc/install/index.rst
+++ b/doc/install/index.rst
@@ -3,7 +3,8 @@
 Quick start
 ===========
 
-MNE-Python requires Python version |min_python_version| or higher.
+MNE-Python requires Python version |min_python_version| or higher. If you've
+never worked with Python before, skip ahead to the last paragraph of this page.
 For users already familiar with Python:
 
 - If you only need 2D plotting capabilities with MNE-Python (i.e., most EEG/ERP
@@ -11,13 +12,15 @@ For users already familiar with Python:
 
   .. code-block:: console
 
-      $ pip install mne  # dependencies are numpy, scipy, matplotlib
+      $ pip install mne
+
+  The only hard dependencies are `NumPy`_ and `SciPy`_, though most users will
+  want to install `Matplotlib`_ too (for plotting).
 
 - If you need MNE-Python's 3D rendering capabilities (e.g., plotting estimated
   source activity on a cortical surface) it is a good idea to install
   MNE-Python into its own virtual environment. To do this with
-  `conda <anaconda>`_ (this will create a new ``conda`` environment called
-  ``mne``):
+  `conda <anaconda>`_:
 
   .. code-block:: console
 
@@ -25,6 +28,7 @@ For users already familiar with Python:
       $ #                   ↑↑↑                       ↑↑↑
       $ #             environment name            package name
 
+  This will create a new ``conda`` environment called ``mne``.
   If you need to convert structural MRI scans into models of the scalp,
   inner/outer skull, and cortical surfaces you also need
   :doc:`FreeSurfer <freesurfer>`.
@@ -34,8 +38,8 @@ for different operating systems (including instructions for installing Python
 if you don't already have it). The :ref:`advanced_setup` page has additional
 tips and tricks for special situations (servers, notebooks, CUDA, installing
 the development version, etc). The :ref:`contributing` has additional
-installation instructions for (future) contributors to MNE-Python (extra
-dependencies, etc).
+installation instructions for (future) contributors to MNE-Python (e.g, extra
+dependencies for running our tests and building our docs).
 
 .. toctree::
     :hidden:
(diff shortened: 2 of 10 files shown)


doc/install/index.rst
```diff
@@ -11,15 +13,17 @@ For users already familiar with Python:
 
       $ pip install mne  # dependencies are numpy, scipy, matplotlib
```
cbrnr, line 14: Technically, the only hard dependencies are `numpy` and `scipy`; `matplotlib` doesn't get installed automatically with `pip install mne`.

doc/install/mne_python.rst
```diff
@@ -129,10 +129,20 @@ GUI). This will install MNE-Python into the "base" conda environment, which
 should be active by default and should already have the necessary dependencies
 (``numpy``, ``scipy``, and ``matplotlib``).
 
+A second option is to install MNE-Python into its own conda environment
+(instead of installing into the "base" environment). This can be done via::
+
+    $ conda create --name new_environment_name --channel conda-forge mne
```
rob-luke, line 135: Will this environment be immediately available for use? Or do you need to activate it. If so, should we inform the user how to activate it?

I have multiple environments and need to activate the one I want. But maybe if you only have one it's activated automatically?

drammock: Newly-created environments are not automatically activated. The message shown in the terminal when creating a new environment (here, "foo") is:
```
#
# To activate this environment, use
#
#     $ conda activate foo
#
# To deactivate an active environment, use
#
#     $ conda deactivate
```
This message appears regardless of how many environments one already has.

The only case where this wouldn't be necessary would be a user who installs everything into "base" environment: with a typical conda installation, the base environment will be set to automatically activate when the terminal is opened (though even that is not a guarantee, it's configurable during anaconda install).

rob-luke: Yes! I forgot that it spits that message out. I think we should add a line stating users should run `conda activate new_environment_name` as stated in the message. I suggest this because users may either not read the message (unfortunate but common), or they will read the message and then not see the same instructions here and wonder which instructions to follow.

rob-luke: Just a suggestion obliviously, I find it hard not to sound bossy in these reviews.

rob-luke:
@drammock I have a few colleagues who are interested in getting started with MNE but havent installed it yet. They are scientists but not tech focused. If you wish, I could get them to follow the install instructions and give us user feedback, but it would happen sometime in the next week (not in the next day or two).

drammock:
> If you wish, I could get them to follow the install instructions and give us user feedback, but it would happen sometime in the next week (not in the next day or two).

Great! Let's get this merged in and backported after I address your other comment (tomorrow, it's late here). Then your colleagues' feedback can be addressed in a second PR

drammock:
@rob-luke @cbrnr @hoechenberger this one is ready for final review/merge.

doc/install/index.rst
```diff
@@ -1,42 +1,52 @@
...
 
+  This will create a new ``conda`` environment called ``mne``.
   If you need to convert structural MRI scans into models of the scalp,
   inner/outer skull, and cortical surfaces you also need
   :doc:`FreeSurfer <freesurfer>`.
 
-The :ref:`install_python_and_mne_python` page has more detailed instructions
-for different operating systems (including instructions for installing Python
-if you don't already have it). The :ref:`advanced_setup` page has additional
+The :ref:`standard_instructions` page has detailed instructions for different
```
rob-luke, line 36: For continuity with the first paragraph, and to make the next link seem more new user friendly

```suggestion
For users unfamiliar with python, the :ref:`standard_instructions` page has detailed instructions for different
```

doc/install/index.rst
```diff
@@ -1,42 +1,52 @@
...
+  MNE-Python into its own virtual environment. To do this with
+  `conda <anaconda>`_:
 
   .. code-block:: console
 
-      $ curl --remote-name https://raw.githubusercontent.com/mne-tools/mne-python/master/environment.yml
-      $ conda env update --file environment.yml
+      $ conda create --name mne --channel conda-forge mne
+      $ #                   ↑↑↑                       ↑↑↑
+      $ #             environment name            package name
```
rob-luke, line 29: 👍 awesome

doc/install/mne_python.rst
```diff
@@ -155,23 +57,19 @@ your operating system.
...
+       $ conda create --name mne --channel conda-forge mne
 
 .. collapse:: |windows| Windows
 
-   - Download the `environment file`_
    - Open an Anaconda command prompt
    - Run :samp:`conda install --name base nb_conda_kernels "spyder>=4.2.1"`
-   - :samp:`cd` to the directory where you downloaded the file
-   - Run :samp:`conda env update --file environment.yml`
+   - Run :samp:`conda create --name mne --channel conda-forge mne`
```
rob-luke, line 72: make this a code block like linux and Mac?

rob-luke:
PR is awesome

> Then your colleagues' feedback can be addressed in a second PR

Will do, ill try not to interfere while I watch them :)

rob-luke (approved):

doc/install/index.rst
cbrnr: Python with upper case P

cbrnr: But why this change? This is not about Python, it's about Freesurfer.

doc/install/install_python.rst
```diff
@@ -0,0 +1,103 @@
+.. include:: ../links.inc
+
+.. _install-python:
+
+Installing Python
+=================
+
+MNE-Python runs within Python, and depends on several other Python packages.
```
cbrnr, line 8: "runs within Python" sounds strange – what information does this convey that is not already contained in the name "MNE-Python"? If you want to explicitly mention that MNE-Python is a Python package, I'd write "MNE-Python is a Python package...".

doc/install/install_python.rst
```diff
@@ -0,0 +1,103 @@
...
+Version |version| requires Python version |min_python_version| or higher. We
+recommend the `Anaconda`_ distribution of Python, which comes with more than
```
cbrnr, line 10:
```suggestion
recommend the `Anaconda`_ Python distribution, which comes with more than
```

doc/install/install_python.rst
```diff
@@ -0,0 +1,103 @@
...
+250 scientific packages pre-bundled, and includes the ``conda`` command line
```
cbrnr, line 11:
```suggestion
250 scientific packages pre-bundled and includes the ``conda`` command line
```

doc/install/mne_python.rst
```diff
@@ -155,23 +57,19 @@ your operating system.
...
 .. collapse:: |apple| macOS
 
-   Update the ``base`` conda environment to include the ``nb_conda_kernels``
+   Update the "base" conda environment to include the ``nb_conda_kernels``
    package, so you can use MNE-Python in Jupyter Notebooks launched from the
-   Anaconda GUI. Then download the MNE-Python `environment file`_ (done here
-   with ``curl``) and use it to create a new environment (named ``mne`` by
-   default)::
+   Anaconda GUI. Then install MNE-Python into a new environment (here called
+   ``mne``, but you can name the environment whatever you want)::
```
cbrnr, line 63: This is potentially confusing (note that I'm not an Anaconda user though). Why do I need to install `nb_conda_kernel` in "base" when MNE-Python is in a separate environment? If I start a Jupyter notebook, how do I switch to the `mne` environment then? Maybe this should be briefly mentioned.

cbrnr (commented):
Very nice summary @drammock! Apart from my minor comments this is ready to merge IMO. On a somewhat unrelated note, it is sad how complicated package installation has become in Python (or rather, it has always been like that) 😢 . Don't ever try Julia because that's how package management should be in Python 😉 .

hoechenberger:
> Don't ever try Julia because that's how package management should be in Python 😉 .

or `npm`. Python packaging is really the worst, it was maybe okay-ish 10 years ago but these days, it's almost unacceptably bad.

doc/install/index.rst
rob-luke: The page I edited is entirely about is about python and nothing to do with FreeSurfer. And the link is to how to a detailed instructions on how install python. Am I confused?

cbrnr: Maybe I was confused. You're right, of course this is referring to Python installation instructions, so your suggestion is fine (but at least my comment on the capital P is still correct 😝 ).

rob-luke: That `P` gets me every time!

doc/install/advanced.rst
```diff
@@ -113,10 +98,13 @@ like Anaconda, Miniconda includes the ``conda`` command line tool for
...
+A similar alternative is `MiniForge`_, which uses the ``conda-forge`` channel
+as the default source for package installation (saving you the trouble of
+typing ``--channel conda-forge`` with each ``conda install`` command).
 
-It is also possible to use a system-level installation of Python (version 3.6
-or higher) and use ``pip`` to install MNE-Python and its dependencies, using
-the provided `requirements file`_:
+It is also possible to use a system-level installation of Python (version
+|min_python_version| or higher) and use ``pip`` to install MNE-Python and its
+dependencies, using the provided `requirements file`_:
```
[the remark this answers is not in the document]
cbrnr: Yes! IMO every proper Python package needs to be `pip` installable. I wouldn't completely switch to `conda`.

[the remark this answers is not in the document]
drammock: @hoechenberger this is in the advanced docs under "other python distros" so it's not an encouragement. I think it's entirely appropriate to say "you can do a system-install and it will probably work". By this point in the install docs it should be clear that it's not what we recommend.

doc/install/install_python.rst
drammock: this page is for people who have probably never used Python before.  If they're coming from MATLAB they may not have any exposure to the idea of "packages" (MATLAB has toolboxes, but they are often bundled / centrally managed / not installed by end-users).  So I don't think it's outlandish to explain "you need python, and several other python packages, in order for MNE-Python to work".  Saying "it's a python package" may not convey that to a naive user.  In light of that, do you have a different suggestion of how to re-word to sound less strange?

doc/install/install_python.rst
drammock: -1 on this suggestion. Saying "the Anaconda distribution of Python" is, I think, subtly clearer about the fact that what you're getting is Python, from a particular source.  Saying instead "the Anaconda Python distribution" could be read as saying what you're getting is "Anaconda Python", and subtly implies that "Anaconda Python" may be relevantly different from other Python flavors (by analogy to, say, food or clothing brands).

doc/install/install_python.rst
cbrnr: Maybe "MNE-Python requires Python and several other Python packages"?

doc/install/install_python.rst
cbrnr: OK, that's fine with me. I'd agree 100% if we were talking about using "Anaconda Python" though, but I was specifically suggestion "Anaconda Python distribution", which IMO conveys this subtle differences you're referring to.

doc/install/install_python.rst
cbrnr: Perhaps remove "other" to make it even clearer?

doc/install/install_python.rst
hoechenberger: Reconciliation proposal: put Anaconda in quotation marks?

```
... the "Anaconda" Python distribution
```

doc/install/mne_python.rst
drammock: if you use the anaconda navigator, you get a window of tiles for different apps (Jupyter, vscode, orange, rstudio, terminal...).  There is also a small dropdown that is easy to miss that says which environment you're in (and it defaults to "base" when you launch anaconda navigator).  Even if you know about how to switch envs, it's easy to forget because the UI doesn't offer good affordances for it, and doing it through the GUI is rather slow as well.

If you install `nb_conda_kernels` into base env, then you can launch Jupyter from base, but create and run notebooks who run their kernel through the MNE environment.  See the dropdown box for new notebook creation:

Screenshot_2021-08-04_09-27-28

Honestly I think this is a case where the explanation is too long and convoluted to be worth including.  I've reworded it to say 

> If you like using Jupyter notebooks, you should also update the "base" conda environment to include the ``nb_conda_kernels`` package; this will make it easier to use MNE-Python in Jupyter Notebooks launched from the Anaconda GUI.

...followed by the relevant code snippet. OK?

cbrnr: Absolutely! Perfect!

doc/install/mne_python.rst
```diff
@@ -129,23 +21,33 @@ GUI). This will install MNE-Python into the "base" conda environment, which
 should be active by default and should already have the necessary dependencies
 (``numpy``, ``scipy``, and ``matplotlib``).
 
+A second option is to install MNE-Python into its own virtual environment
+(instead of installing into conda's "base" environment). This can be done via::
+
+    $ conda create --name new_environment_name python=3
```
hoechenberger, line 27:
```suggestion
    $ conda create --name=new_environment_name python=3
```

doc/install/mne_python.rst
```diff
@@ -155,23 +57,35 @@ your operating system.
...
+    name the environment whatever you want)::
+
+        $ conda create --name=mne --channel=conda-forge mne
+
+    If you like using Jupyter notebooks, you should also update the "base"
+    conda environment to include the ``nb_conda_kernels`` package; this will
+    make it easier to use MNE-Python in Jupyter Notebooks launched from the
+    Anaconda GUI::
+
+        $ conda install --name base nb_conda_kernels
```
hoechenberger, line 70:
```suggestion
        $ conda install --name=base nb_conda_kernels
```

hoechenberger (commented):
Otherwise LGTM!

cbrnr (approved):

hoechenberger (approved):

cbrnr:
Thanks @drammock !

drammock:
backported

agramfort:
Thx @drammock

regarding comments on packaging. Npm is easier as it’s only JavaScript not python, c++, OpenGL / vtk etc. Regarding Julia yes maybe but I want to see something flying with the complexity of mne dependencies. You want to help packaging -> reduce dependencies and keep complexity low ! Packaging across platforms is just damned hard !

cbrnr:
> reduce dependencies and keep complexity low

Yes, that's true no matter what programming language.

PR #9636, opened 2021-08-02
[BUG, MRG] Fix coordinate frame loading from disc for non-scalp EEG recordings and fNIRS

### LINKED ISSUE #9635: [BUG] Coordinate Frame not Saved
I'm not sure if I'm missing something obvious but it looks like the coordinate frame doesn't survive the round trip to and from disk. I'm not sure why that is.

```
import numpy as np
import mne

for ch_type in ('eeg', 'seeg', 'ecog', 'dbs'):
    raw = mne.io.RawArray(
        np.random.random((10, 1000)),
        mne.create_info([f'ch{i}' for i in range(1, 11)],  1000, [ch_type] * 10))
    montage = mne.channels.make_dig_montage(
        {ch: np.random.random((3,)) for ch in raw.ch_names}, lpa=[1, 0, 0], nasion=[0, 0.1, 0], rpa=[-1, 0, 0],
        coord_frame='mri')
    raw.set_montage(montage)
    raw.save('tmp-raw.fif', overwrite=True, verbose=False)
    print(ch_type)
    print('Before: ', raw.info['chs'][0]['coord_frame'])
    info = mne.io.read_info('tmp-raw.fif')
    print('After : ', info['chs'][0]['coord_frame'])
```

Note, it appears this is specific to `seeg`, `ecog` and `dbs`, it works for `eeg`.

alexrockhill:
Fixes #9635.

Man this information was really hard to find it was so deeply buried in `mne/io/tag.py` it could probably also be refactored to be in `mne/io/constants.py` or something. Anyway this fixes the issue above and is another part in addressing #9627.


alexrockhill:
Also in https://github.com/mne-tools/mne-python/blob/main/mne/io/write.py#L377:L402 every other `info['chs']` attribute except for `coord_frame` is written to disk. I assume this is on purpose as it would involve changing the FIF standard but that would probably avoid these such issues.

alexrockhill:
@rob-luke, does this look okay to you? From what I was working on earlier, it looks like fNIRS is digitizations are stored the same way as EEG (transformed to the "head" coordinate frame but originally recorded in some arbitrary digitizer space). This PR just makes it so that when `info['chs'][idx]['kind']` is an fNIRS channel, it is assumed to be in head space. This is consistent with the last PR which made sure that the coordinate frame of the `raw` object was "head" when the montage was set (preparing for it to be saved to disk in "head")/

alexrockhill:
Ok from reasoning a bit, it seems an active choice was made to assume `seeg`, `ecog` and `dbs` are in the unknown coordinate frame when no information is given. I think that was a reasonable choice (when you do `mne.create_info(['ch1'], sfreq=1000, ['seeg'])` you probably don't expect the empty coordinates to be in the "head" coordinate frame) but it causes issues when saving and loading from disk because of the previously mentioned lack of a `coord_frame` field in the FIF format which makes it so that you *have* to assume the coordinate frame based on the type. For backward compatibility and for practical reasons, this seems like the best solution now because all `seeg`, `ecog` and `dbs` data is converted to head when the montage is set. It might be worth looking into specifying the coordinate frame in the future.

Also, this PR will have to be merged before #9585 because when `plot_sensors` is refactored, it will give warnings for unknown coordinate frames.

alexrockhill:
You know what, I think it might be better just to close this and put it in #9585

PR #9640, opened 2021-08-04
ENH: Add support to more than 3 stcs in mne.viz.plot_sparse_source_estimates

PABannier:
The goal of this PR is to add support to more than 3 secs in mne.viz.plot_sparse_source_estimates . 

To make a figure, I need to pass more than 3 source estimates to plot_sparse_source_estimates. I adapted the function to support more than 3 source estimates. The bottleneck is the limited number of line styles when plotting the activation intensity.

### DIFF
--- a/mne/viz/_3d.py
+++ b/mne/viz/_3d.py
@@ -2600,6 +2600,25 @@ def plot_sparse_source_estimates(src, stcs, colors=None, linewidth=2,
     # Update the backend
     from .backends.renderer import _get_renderer
 
+    linestyles = [
+        ('solid',                 'solid'),
+        ('dotted',                'dotted'),
+        ('dashed',                'dashed'),
+        ('dashdot',               'dashdot'),
+        ('loosely dotted',        (0, (1, 10))),
+        ('dotted',                (0, (1, 1))),
+        ('densely dotted',        (0, (1, 1))),
+        ('loosely dashed',        (0, (5, 10))),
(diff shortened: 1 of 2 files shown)


mne/viz/_3d.py
```diff
@@ -2613,7 +2613,8 @@ def plot_sparse_source_estimates(src, stcs, colors=None, linewidth=2,
     if colors is None:
         colors = _get_color_list()
 
-    linestyles = ['-', '--', ':']
+    linestyles = cycle(['-', '--', ':'])
+    linestyles = [next(linestyles) for _ in range(len(stcs))]
```
agramfort, line 2617: Can you exploit the bigger diversity of matplotlib line styles 

https://matplotlib.org/stable/gallery/lines_bars_and_markers/linestyles.html

mne/viz/_3d.py
```diff
@@ -2600,6 +2600,25 @@ def plot_sparse_source_estimates(src, stcs, colors=None, linewidth=2,
     # Update the backend
     from .backends.renderer import _get_renderer
 
+    linestyles = [
```
PABannier, line 2603: @agramfort wdyt?

agramfort: Put dashed before dotted to keep old behavior and we can merge

agramfort (approved):

agramfort:
Thx @PABannier

PR #9644, opened 2021-08-04
[BUG, MRG] Fix biosemi example coordinate frames

### LINKED ISSUE #9641: Regression in "How to plot topomaps the way EEGLAB does" example - montage.get_positions() returns different positions than the channel positons in evoked object.
Hi, I just noticed that the [How to plot topomaps the way EEGLAB does](https://mne.tools/dev/auto_examples/visualization/eeglab_head_sphere.html) example no longer shows correct EEGLAB-like channel layout:

the two layouts are the same on the current development and stable versions (0.23), which is incorrect. On version 0.22 these layouts are different (and correct):

The sphere parameters used in the example are not the same as before and this seems due to a change in how channel positions are obtained. Previously they were dug out from `.info['chs']` dictionary, now `montage.get_positions()` is used. However these two are not exchangable because the applied montage is in head coordinates, while the positions returned by `.get_positions()` show 'unknown' coord_frame.
In the same example doing:
```python
pos_mntg = biosemi_montage.get_positions()['ch_pos']['Oz']
ch_idx = fake_evoked.ch_names.index('Oz')
pos_evkd = fake_evoked.info['chs'][ch_idx]['loc'][:3]

print([f'{v:0.5f}' for v in pos_mntg])
print([f'{v:0.5f}' for v in pos_evkd])
```
gives different positions:
```
['0.00000', '-0.09494', '-0.00332']
['0.00000', '-0.09494', '0.03683']
```
I'm not sure what the best solution should be - is the intent of `montage.get_positions()` is to give channel positions in unknown coord frame? If so, the example should be reverted to use the old and ugly `.info['chs'][idx]['loc'][:3]` way of obtaining channel positions.

alexrockhill:
Fixes #9641.

### DIFF
--- a/examples/visualization/eeglab_head_sphere.py
+++ b/examples/visualization/eeglab_head_sphere.py
@@ -64,7 +64,7 @@ fake_evoked.set_montage(biosemi_montage)
 chs = ['Oz', 'Fpz', 'T7', 'T8']
 # when the montage is set, it is transformed to the "head" coordinate frame
 # that MNE uses internally, therefore we need to use
-# :meth:`mne.Evoked.get_montage` to get these properly transformed coordinates
+# ``fake_evoked.get_montage()`` to get these properly transformed coordinates
 montage_head = fake_evoked.get_montage()
 ch_pos = montage_head.get_positions()['ch_pos']
 pos = np.stack([ch_pos[ch] for ch in chs])
--- a/examples/visualization/eeglab_head_sphere.py
+++ b/examples/visualization/eeglab_head_sphere.py
@@ -62,7 +62,9 @@ fake_evoked.set_montage(biosemi_montage)
 
 # first we obtain the 3d positions of selected channels
 chs = ['Oz', 'Fpz', 'T7', 'T8']
-# we need the montage transformed to "head" coordinates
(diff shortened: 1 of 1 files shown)


mmagnuski (approved):
Great, thanks for a quick reaction. I'm a little behind with this `get_montage()` functionality :)

alexrockhill:
I just cleaned it up a bit

examples/visualization/eeglab_head_sphere.py
```diff
@@ -62,7 +62,10 @@
 
 # first we obtain the 3d positions of selected channels
 chs = ['Oz', 'Fpz', 'T7', 'T8']
-pos = np.stack([biosemi_montage.get_positions()['ch_pos'][ch] for ch in chs])
+# we need the montage transformed to "head" coordinates
```
mmagnuski, line 65: If we are adding more explanation, then maybe something like this:
```suggestion
# we want to have the channel positions in "head" coordinates - we can get
# these from the ``Evoked`` object via ``.get_montage()``
```

mmagnuski:
Great! I'll go ahead and merge.

PR #9647, opened 2021-08-05
Improve documentation for argument `mask` and fix documentation for argument `ch_type` for topographic plots.

mscheltienne:
#### Reference issue

Second attempt to properly solve issue #9611.

- [x] Support for non-bool mask array which would lead to error on the bitwise operations.
- [x] Update the docstring for the argument `mask`
- [ ] Add test for the mask argument
- [ ] Add example in the tutorial
- [ ] Update the changelog
- [ ] Fix the docstring for argument `ch_type`

#### Additional information

Per the discussion on issue #9611, the following tutorial could be updated with an example:
https://mne.tools/stable/auto_examples/visualization/evoked_topomap.html#sphx-glr-auto-examples-visualization-evoked-topomap-py

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -318,7 +318,7 @@ proj : bool | 'interactive' | 'reconstruct'
     .. versionchanged:: 0.21
        Support for 'reconstruct' was added.
 """
-docdict["topomap_ch_type"] = """
+docdict["evoked_topomap_ch_type"] = """
 ch_type : 'mag' | 'grad' | 'planar1' | 'planar2' | 'eeg' | None
     The channel type to plot. For 'grad', the gradiometers are collected in
     pairs and the RMS for each pair is plotted.
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -384,7 +384,7 @@ cbar_fmt : str
     String format for colorbar values.
 """
 docdict["topomap_mask"] = """
-mask : ndarray of bool, shape (n_channels, n_times) | None
+mask : ndarray of bool, shape (n_channels, n_samples) | None
     The channels to be marked as significant at a given time point.
     Indices set to ``True`` will be considered. Defaults to ``None``.
 """
--- a/mne/decoding/csp.py
+++ b/mne/decoding/csp.py
@@ -308,15 +308,8 @@ class CSP(TransformerMixin, BaseEstimator):
             only significant sensors will be shown.
         title : str | None
             Title. If None (default), no title is displayed.
-        mask : ndarray of bool, shape (n_channels, n_times) | None
-            The channels to be marked as significant at a given time point.
-            Indices set to `True` will be considered. Defaults to None.
-        mask_params : dict | None
-            Additional plotting parameters for plotting significant sensors.
-            Default (None) equals::
-
-                dict(marker='o', markerfacecolor='w', markeredgecolor='k',
-                     linewidth=0, markersize=4)
+        %(topomap_mask)s
+        %(topomap_mask_params)s
         %(topomap_outlines)s
         contours : int | array of float
             The number of contour lines to draw. If 0, no contours will be
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -1499,7 +1499,7 @@ def plot_evoked_topomap(evoked, times="auto", ch_type=None,
         automatically by checking for local maxima in global field power. If
         "interactive", the time can be set interactively at run-time by using a
         slider.
-    %(topomap_ch_type)s
+    %(evoked_topomap_ch_type)s
     %(topomap_vmin_vmax)s
     %(topomap_cmap)s
     %(topomap_sensors)s
--- a/mne/viz/topomap.py
+++ b/mne/viz/topomap.py
@@ -985,6 +985,7 @@ def _plot_topomap(data, pos, vmin=None, vmax=None, cmap=None, sensors=True,
                 col.set_clip_path(patch_)
 
     pos_x, pos_y = pos.T
+    mask = mask.astype(bool, copy=False) if mask is not None else None
     if sensors is not False and mask is None:
         _topomap_plot_sensors(pos_x, pos_y, sensors=sensors, ax=ax)
     elif sensors and mask is not None:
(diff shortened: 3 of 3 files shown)


mne/utils/docs.py
```diff
@@ -384,7 +384,7 @@
     String format for colorbar values.
 """
 docdict["topomap_mask"] = """
-mask : ndarray of bool, shape (n_channels, n_times) | None
+mask : ndarray of bool, shape (n_channels, n_samples) | None
     The channels to be marked as significant at a given time point.
     Indices set to ``True`` will be considered. Defaults to ``None``.
```
drammock, line 389: I think here we should be more explicit than to say "channels marked as significant".
```suggestion
mask : ndarray of bool, shape (n_channels, n_samples) | None
    Array indicating channel-time combinations to highlight with a distinct
    plotting style. Indices set to ``True`` will be plotted with the parameters
    given in ``mask_params``. Defaults to ``None``.
```

drammock (approved):
just one suggestion for clearer docstring, otherwise LGTM. Thanks @mscheltienne

mne/utils/docs.py
mscheltienne: Sure, but I'm not sure this docstring is good either. `mask` highlights sensors on a given topography at a given time. In your proposition, it is not clear what is highlighted. 

Also, instead of channels, I would use the word sensor to link to the `sensors` argument used for plotting. 

Maybe something along the lines of:

> Sensor-time combinations array indicating the sensors to highlight with a distinct plotting style at a given time point.

drammock: > Also, instead of channels, I would use the word sensor to link to the `sensors` argument used for plotting.

The "sensors" that are plotted are not necessarily all of the channels in the Evoked instance (since user can pass a `ch_type` argument --- but (I think) even if they do restrict the `ch_type` the mask still has to be the same dimension as `evoked.data` (in other words, the mask must account for all *channels* not just the subset of *sensors* that are drawn on the plot).  That's why the docstring of this parameter uses "n_channels" in the shape specification. 

> Maybe something along the lines of:
> 
> > Sensor-time combinations array indicating the sensors to highlight with a distinct plotting style at a given time point.

To me that is not more clear than my past proposal.  What about this:

```
mask : ndarray of bool, shape (n_channels, n_samples) | None
    Array indicating channel-time combinations to highlight with a distinct
    plotting style (useful for, e.g., marking which channels at which times a
    statistical test of the data reaches significance). Array elements set to
    ``True`` will be plotted with the parameters given in ``mask_params``.
    Defaults to ``None``, equivalent to an array of all ``False`` elements.
```

mscheltienne: Let's keep in mind that this is not only for evoked instances; but for all topographic map plots. This now leads me to question this `ch_type` argument. 

> [...] not necessarily all of the channels in the Evoked instance (since user can pass a `ch_type `argument [...]

So for an Evoked instance with multiple channels types, the user can pick channels to plot with `ch_type`?
However, [`mne.viz.plot_topomap()`](https://mne.tools/dev/generated/mne.viz.plot_topomap.html#mne.viz.plot_topomap), which is the function called in the end by `evoked.plot_topomap()`, looks different with the argument `pos` which must be an `Info` **with only one channel type**. Yet, it still has a `ch_type` argument with the same description. I'm getting a bit confused as to what `ch_type` does.

Back to the matter at hand, I will check tomorrow that the dimension of the mask must include all the channels present in the Evoked object and not only the channels plotted. 
The last proposition does look better to me.

drammock: > So for an Evoked instance with multiple channels types, the user can pick channels to plot with `ch_type`?

yes

> However, [`mne.viz.plot_topomap()`](https://mne.tools/dev/generated/mne.viz.plot_topomap.html#mne.viz.plot_topomap), which is the function called in the end by `evoked.plot_topomap()`, looks different with the argument `pos` which must be an `Info` **with only one channel type**. Yet, it still has a `ch_type` argument with the same description.

It should not have the same description. The docstring of `plot_evoked_topomap` should not say "The channel type being plotted. Determines the 'auto' extrapolation mode."  The problem appears to be that the `topomap_ch_type` key appears twice in the docdict.  Can you update the key [here](https://github.com/mne-tools/mne-python/blame/main/mne/utils/docs.py#L321) to be `evoked_topomap_ch_type` and also update the key used in the `plot_evoked_topomap` docstring?
 
> Back to the matter at hand, I will check tomorrow that the dimension of the mask must include all the channels present in the Evoked object and not only the channels plotted.

Yes please do double-check that, thanks

mscheltienne: Ok, that makes more sense. I will update the docdict key tomorrow.

mscheltienne: Also, now that I am looking at this `ch_type` argument.. 
In 0.23, for the function [`mne.viz.plot_topomap()`](https://mne.tools/stable/generated/mne.viz.plot_topomap.html?highlight=mne%20viz%20plot_topomap#mne.viz.plot_topomap), the `New in version` says `0.21`.

    The channel type being plotted. Determines the 'auto' extrapolation mode.
    New in version 0.21.

But now, in the code of 0.24, the doc is created with:

```
    %(topomap_ch_type)s

        .. versionadded:: 0.24.0
```

This leads to the very weird:

image

Any thoughts on this? Let's fix it at the same time.
`plot_topomap()` does appear to have a new argument, `cnorm` in 0.24, which is correctly annotated with `.. versionadded:: 0.24`. Typo which leads to a similar addition for an already existing argument?

mscheltienne: Yep, typo from the addition of `cnorm` I think: https://github.com/mne-tools/mne-python/commit/b6a36089a963a97380488845491fe38e8d3b610b

drammock: nice detective work @mscheltienne

drammock: I've updated the docstring suggestion to use my later wording.  with that and the removal of the spurious `versionadded` line I think this will be good to go.

hoechenberger (approved):

hoechenberger:
Thanks @mscheltienne!

mscheltienne:
@hoechenberger @drammock Am I missing something or was this merged before the docstring was updated with @drammock latest suggestion and before the removal of the suspicious `version added`?

hoechenberger:
Whoops. My bad, sorry – I overlooked this part. Will make a followup PR with a fix…

mscheltienne:
@hoechenberger If you are busy, I can do it quickly now.

hoechenberger:
> @hoechenberger If you need, I can do it quickly now.

Please do go ahead! Much appreciated

PR #9648, opened 2021-08-05
MAINT, CI: fix doc build

drammock:
#9634 broke documentation builds because it tried to parse `setup.py` to get the required python version (setup.py is not distributed with the package, so it can only be used that way in git-clone-based installations).

This uses either `importlib` (py 3.8+) or `pkg_resources` (py3.7) instead.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -879,12 +879,16 @@ for icon, cls in icons.items():
 
 # -- Dependency info ----------------------------------------------------------
 
-strip_chars = '=<>,"\''
-with open(os.path.join('..', 'setup.py'), 'r') as fid:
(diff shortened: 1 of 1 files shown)


drammock:
note: I'm expecting the CIs to be green here (we should wait and see), but the real test will be in the deploy step after this is merged.

drammock:
rendered doc looks okay:
- https://32254-1301584-gh.circle-artifacts.com/0/dev/install/index.html
- https://32254-1301584-gh.circle-artifacts.com/0/dev/install/install_python.html

(look at the first paragraph of each page, it says correctly "Python 3.7"; the 3.7 part is rendered from a variable)

PR #9650, opened 2021-08-06
FIX: Remove ambiguity in local filename check

### LINKED ISSUE #9639: BEM changes depending on calling path
#### Describe the bug
The BEM model is different depending on where the the code is called from.  From what I can tell it appears to be missing the CRAS offset.  This is  in the viz.plot_aligment, but I have also seen it in the source localization outputs.

#### Steps to reproduce
``` 
import mne
import os, os.path as op

subjects_dir=op.join(os.environ['HOME'], 
                 'mne_data/MNE-somato-data/derivatives/freesurfer/subjects')

#In mne_data directory
os.chdir(subjects_dir)
mne.viz.plot_bem('01', subjects_dir=subjects_dir)

#In home directory
os.chdir(os.environ['HOME'])
mne.viz.plot_bem('01', subjects_dir=subjects_dir)
```

#### Expected results
image

#### Actual results
image

#### Additional information
Platform:      Linux-5.4.0-80-generic-x86_64-with-glibc2.10
Python:        3.8.8 (default, Feb 24 2021, 21:46:12)  [GCC 7.3.0]
Executable:    /home/jstout/miniconda3/envs/enigma/bin/python
CPU:           x86_64: 20 cores
Memory:        125.5 GB

mne:           0.23.0
numpy:         1.20.1 {blas=openblas, lapack=openblas}
scipy:         1.6.1
matplotlib:    3.3.4 {backend=module://ipykernel.pylab.backend_inline}

sklearn:       0.24.1
numba:         0.53.0
nibabel:       3.2.1
nilearn:       0.7.1
dipy:          1.4.0
cupy:          Not found
pandas:        1.2.3
mayavi:        Not found
pyvista:       0.29.0 {pyvistaqt=0.3.0, OpenGL 4.5.0 NVIDIA 460.91.03 via Quadro RTX 8000/PCIe/SSE2}
vtk:           9.0.1
PyQt5:         5.15.4

jstout211:
#### Reference issue
Fixes #9639 

Prior to this fix, the _check_mri would proceed if it found the default mri (T1.mgz).  If the default filename happened to be in the calling directory, it would be returned from the function rather than performing the expected search of the freesurfer folder.

The fix does a check to confirm that if the file is found, it is not from the current path unless a relative or absolute path is provided.  The assumption being that having a relative or absolute path to the current directory was done with some intention rather than a coincidental default setting.

#### Additional information
It is possible that this could return a warning rather than an error.

### DIFF
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -474,8 +474,9 @@ def _check_mri(mri, subject, subjects_dir):
         if not op.isfile(mri):
             raise FileNotFoundError(f'MRI file {mri!r} not found')
     if op.basename(mri) == mri:
-        raise IOError(f'Ambiguous filename - found {mri!r} in current folder.\n'
-                      'If this is correct prefix with relative or absolute path')
+        err = (f'Ambiguous filename - found {mri!r} in current folder.\n'
+               'If this is correct prefix name with relative or absolute path')
+        raise IOError(err)
     return mri
 
 
--- a/mne/_freesurfer.py
+++ b/mne/_freesurfer.py
@@ -463,6 +463,8 @@ def read_talxfm(subject, subjects_dir=None, verbose=None):
 def _check_mri(mri, subject, subjects_dir):
     """Check whether an mri exists in the Freesurfer subject directory."""
     _validate_type(mri, 'path-like', 'mri')
+    if op.isfile(mri) and op.basename(mri) != mri:
+        return mri
     if not op.isfile(mri):
         if subject is None:
             raise FileNotFoundError(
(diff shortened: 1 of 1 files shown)


hoechenberger:
Thanks @jackz314
CI is failing because two lines are too long, could you have a look?

jackz314:
> Thanks @jackz314
> CI is failing because two lines are too long, could you have a look?

Wrong person I believe. Should be @jstout211 instead.

drammock (approved):

agramfort (approved):

hoechenberger:
Thanks @jstout211 (and sorry @jackz314 for the incorrect tagging)

PR #9651, opened 2021-08-06
Followup #9647 - Improve documentation of the mask argument for topographic maps.

mscheltienne:
#### Reference issue

Followup of PR #9647.

Example added using the argument `mask` to [_Plotting topographic maps of evoked data_](https://mne.tools/stable/auto_examples/visualization/evoked_topomap.html#sphx-glr-auto-examples-visualization-evoked-topomap-py). The example was created by @drammock on #9611.
Replace the docstring for `topomap_mask` with the proposition by @drammock on #9647.
Removed `.. versionadded:: 0.24.0` for argument `ch_type` in `mne.viz.plot_topomap()` introduced by b6a36089a963a97380488845491fe38e8d3b610b

### DIFF
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -123,7 +123,7 @@ plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
 # %%
 # We can also highlight specific channels by adding a mask, to e.g. mark
-# chanels exceeding a threshold at a given time:
+# channels exceeding a threshold at a given time:
 
 # Define a threshold and create the mask
 mask = evoked.data > 1e-13
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -122,7 +122,19 @@ evoked.plot_topomap(0.1, ch_type='mag', show_names=True, colorbar=False,
 plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
 # %%
-# We can also highlight specific channels by adding a mask:
+# We can also highlight specific channels by adding a mask, to e.g. mark
+# chanels exceeding a threshold at a given time:
+
+# Define a threshold and create the mask
+mask = evoked.data > 1e-13
+
+# Select times and plot
+times = (0.09, 0.1, 0.11)
+evoked.plot_topomap(times, ch_type='mag', time_unit='s', mask=mask,
+                    mask_params=dict(markersize=10, markerfacecolor='y'))
+
+#%%
+# Or by manually picking the channels to highlight at different times:
 
 times = (0.09, 0.1, 0.11)
 _times = np.searchsorted(evoked.times, times) - 1
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -137,7 +137,7 @@ evoked.plot_topomap(times, ch_type='mag', time_unit='s', mask=mask,
 # Or by manually picking the channels to highlight at different times:
 
 times = (0.09, 0.1, 0.11)
-_times = np.searchsorted(evoked.times, times) - 1
+_times = ((np.abs(evoked.times - t)).argmin() for t in times)
 significant_channels = [
     ('MEG 0231', 'MEG 1611', 'MEG 1621', 'MEG 1631', 'MEG 1811'),
     ('MEG 2411', 'MEG 2421'),
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -121,6 +121,23 @@ evoked.plot_topomap(0.1, ch_type='mag', show_names=True, colorbar=False,
                     time_unit='s')
 plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
+# We can also highlight specific channels by adding a mask:
+
+times = (0.09, 0.1, 0.11)
+_times = np.searchsorted(evoked.times, times) - 1
+significant_channels = [
+    ('MEG 0231', 'MEG 1611', 'MEG 1621', 'MEG 1631', 'MEG 1811'),
+    ('MEG 2411', 'MEG 2421'),
+    ('MEG 1621')]
+_channels = [np.in1d(evoked.ch_names, ch) for ch in significant_channels]
+
+mask = np.zeros(evoked.data.shape, dtype='bool')
+for _chs, _time in zip(_channels, _times):
+    mask[_chs, _time] = True
+
+evoked.plot_topomap(times, ch_type='mag', time_unit='s', mask=mask,
+                    mask_params=dict(markersize=10, markerfacecolor='y'))
+
 # %%
 # Animating the topomap
 # ---------------------
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -89,7 +89,7 @@ evoked.plot_topomap(times, ch_type='mag', cmap='Spectral_r', res=32,
 # the effect of extrapolation. There are three extrapolation modes:
 #
 # - ``extrapolate='local'`` extrapolates only to points close to the sensors.
-# - ``extrapolate='head'`` extrapolates out to the head head circle.
+# - ``extrapolate='head'`` extrapolates out to the head circle.
 # - ``extrapolate='box'`` extrapolates to a large box stretching beyond the
 #   head circle.
 #
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -121,6 +121,7 @@ evoked.plot_topomap(0.1, ch_type='mag', show_names=True, colorbar=False,
                     time_unit='s')
 plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
+# %%
 # We can also highlight specific channels by adding a mask:
 
 times = (0.09, 0.1, 0.11)
--- a/examples/visualization/evoked_topomap.py
+++ b/examples/visualization/evoked_topomap.py
@@ -133,7 +133,7 @@ times = (0.09, 0.1, 0.11)
 evoked.plot_topomap(times, ch_type='mag', time_unit='s', mask=mask,
                     mask_params=dict(markersize=10, markerfacecolor='y'))
 
-#%%
+# %%
 # Or by manually picking the channels to highlight at different times:
 
 times = (0.09, 0.1, 0.11)
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -384,9 +384,9 @@ cbar_fmt : str
     String format for colorbar values.
 """
 mask_base = """
-mask : ndarray of bool, shape %(shape)s | None
-    Array indicating channel%(shape-appendinx)s to highlight with a distrinct
-    plotting style%(example)s. Array elements set to ``True`` will be plotted
+mask : ndarray of bool, shape {shape} | None
+    Array indicating channel{shape-appendinx} to highlight with a distrinct
+    plotting style{example}. Array elements set to ``True`` will be plotted
     with the parameters given in ``mask_params``. Defaults to ``None``,
     equivalent to an array of all ``False`` elements.
 """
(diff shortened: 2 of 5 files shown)


mscheltienne:
@hoechenberger The complete PR ;) For documentation changes like this one, should a line be added to the changelog `latest.inc`?

mscheltienne:
Screenshot 2021-08-06 at 11 57 18

That is not the output I was looking for..

mmagnuski:
The same mask docstring is pasted to `mne.viz.plot_topomap` - but for `mne.viz.plot_topomap` it shouldn't say that mask is `(n_channels, n_samples)` (because `mne.viz.plot_topomap` is used to plot a single topography, so the mask shape is just `(n_sensors,)`.

mscheltienne:
@mmagnuski Good catch! 
I put the docstring key `%(topomap_mask)s` and `%(topomap_mask_params)s` everywhere the exact same docstring was appearing, with the advised shape `(n_channels, n_times)`. It appears in:

- `mne.cov.plot_topomap()`
- `mne.decoding.csp.plot_patterns()`
- `mne.viz.plot_topomap()`
- `mne.viz.plot_evoked_topomap()`
- `mne.viz.plot_arrowmap()`

Now, I am not familiar with all those functions, but it looks like the only function actually using a mask of shape `(n_channels, n_times)` is `mne.viz.plot_evoked_topomap()`; while the other lack the `times` argument and thus need a mask of shape `(n_channels, )`.

-> Need confirmation on that.

With that in mind, as for the argument `ch_type`, the docstring should be split in 2 docdict keys:

- `topomap_mask`
- `evoked_topomap_mask`

The docdict key `topomap_mask_params` should be valid for both.

If people with knowledge about the behavior of the different functions can confirm that the only one plotting multiple topographies at different times is `mne.viz.plot_evoked_topomap()`, I will add the second docdict key.

mmagnuski:
@mscheltienne Thanks for the extensive search for mask docstring, I will take a look at these functions tomorrow, but your intuition sounds correct. I also suspect that `plot_patterns` is a multi-axes plot, but I am not sure how it handles hask (and whether anyone would need to mark channels when plottin csp patterns).

drammock:
`mne.viz.plot_topomap()`, `mne.cov.plot_topomap()` and `mne.viz.plot_arrowmap()` have no second dimension to the data, so mask shape should be `n_channels`.

`mne.viz.plot_evoked_topomap()` has 2 dims, as we've extensively discussed.

`mne.decoding.csp.plot_patterns()` (and `mne.decoding.SPoC.plot_patterns()`) have channels x number of patterns and I don't know off the top of my head if it allows mask shapes of `(n_channels,)` or `(n_channels, n_patterns)` (or it might even already handle both if it's set up to do broadcasting).

> With that in mind, as for the argument `ch_type`, the docstring should be split in 2 docdict keys:
> 
>     * `%(topomap_mask)s`
> 
>     * `%(evoked_topomap_mask)s`

if `evoked_topomap_mask` is only used in one function, there's no point in putting it in the docdict, just copy the parameter description back into the actual function docstring. But it seems like a little digging is needed first to see what `plot_patterns()` needs.

mscheltienne:
I will have a look as well for the `plot_patterns()`. See what the correct docstring should be.
Obviously, if a docstring is used only in one location, I won't set it in the docdict.

mmagnuski:
Ok, it seems that `.plot_patterns` also expects a 2d mask but it should be `(n_channels, n_patterns)`. This is because `plot_patterns` turns the patterns into an Evoked object:

https://github.com/mne-tools/mne-python/blob/c9752455b79de10fac672d3ed25911f2618f71cc/mne/decoding/csp.py#L343

then it uses the `evoked.plot_topomap`, but the `.data` parameter is now `n_channels, n_patterns`.
(However, I still wonder why anyone would need to highlight specific channels when plotting patterns.)

> if evoked_topomap_mask is only used in one function, there's no point in putting it in the docdict, just copy the parameter description back into the actual function docstring. 

@drammock If these docstrings only differ in one, two places (the shape of the mask) maybe its worth constructing all the  docstrings from a common template using `.format()`? Also - keeping all related docstrings in one place might help with future changes to the mask parameter.

drammock:
> If these docstrings only differ in one, two places (the shape of the mask) maybe its worth constructing all the docstrings from a common template using `.format()`?

Agreed

mscheltienne:
I added a proposal based on your replies and on the previous docstring proposed by @drammock. I'm not sure what was the best formatting method to edit part in the middle of the base docstring, and I settled with a syntax similar to the one used by `fill_doc`.

mne/utils/docs.py
```diff
@@ -383,11 +383,29 @@
...
+    equivalent to an array of all ``False`` elements.
+"""
+topomap_mask_dstr = {
+    'shape': '(n_channels, )',
+    'shape-appendinx': '',
+    'example': ''}
+evoked_topomap_mask_dstr = {
+    'shape': '(n_channels, n_times)',
+    'shape-appendinx': '-time combinations',
+    'example': ' (useful for, e.g., marking which channels at which times ' +
```
mmagnuski, line 400: nitpick but I don't think you need commas before and after `e.g.` - only the one after is required AFAIK

mne/utils/docs.py
```diff
@@ -383,11 +383,29 @@
...
+mask : ndarray of bool, shape %(shape)s | None
+    Array indicating channel%(shape-appendinx)s to highlight with a distrinct
+    plotting style%(example)s. Array elements set to ``True`` will be plotted
+    with the parameters given in ``mask_params``. Defaults to ``None``,
+    equivalent to an array of all ``False`` elements.
+"""
+topomap_mask_dstr = {
+    'shape': '(n_channels, )',
+    'shape-appendinx': '',
+    'example': ''}
```
mmagnuski, line 396: I would prefer compressing these dictionary lines, but I leave it up to you and @drammock to decide.

examples/visualization/evoked_topomap.py
```diff
@@ -121,6 +121,24 @@
                     time_unit='s')
 plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
+# %%
+# We can also highlight specific channels by adding a mask:
+
+times = (0.09, 0.1, 0.11)
+_times = np.searchsorted(evoked.times, times) - 1
```
mmagnuski, line 128: is the `- 1` necessary? In my test you don't need the `- 1`:
```python
import numpy as np

times = np.arange(-0.25, 0.5, 0.01)
request = (0.1, 0.25, 0.3)
_times = np.searchsorted(times, request)

print(times[_times])
```
but in general I prefer `[(np.abs(evoked.times - t)).argmin() for t in times]` as its more accurate (although it is clearly less readable).

examples/visualization/evoked_topomap.py
```diff
@@ -121,6 +121,24 @@
                     time_unit='s')
 plt.subplots_adjust(left=0.01, right=0.99, bottom=0.01, top=0.88)
 
+# %%
+# We can also highlight specific channels by adding a mask:
```
mmagnuski, line 125: I'd add a simpler example before this one. For example creating the mask by finding channel-time points that exceed some value. This way we first showcase the easier approach (and arguably more common) to mask construction.

mmagnuski (changes requested):
just a few changes and we'll be done :)

drammock:
the CI failure is real / related:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/9508/workflows/602e6536-0740-40b4-a15a-f6f11d7a9225/jobs/32396?invite=true#step-139-1862

```
/home/circleci/project/mne/decoding/csp.py:docstring of mne.decoding.csp.CSP.plot_patterns:92: WARNING: py:obj reference target not found: n_patterns
``` 
Usually that error means double-backticks were needed where only single-backticks were used.

mscheltienne:
@mmagnuski Thanks for the feedback, I'll have a look. 
@drammock Thanks! I could not figure out what was failing! I should be able to found the cause with this error message.

mscheltienne:
@drammock Sorry, but I don't get why the HTML doc build raises this warning. Indeed, it looks like it's trying to find a reference to `n_patterns`.. but I did not use backticks around n_patterns.

The docstring of `CSP.plot_patterns()` is:

Screenshot 2021-08-10 at 13 24 21

The only place where `n_patterns` appears in the entire mne folder is in `docs.py`:

```
patterns_topomap_mask_dstr = {
    'shape': '(n_channels, n_patterns)',
    'shape-appendinx': '-pattern combinations',
    'example': ''}
```

mne/utils/docs.py
```diff
@@ -383,11 +383,29 @@
 cbar_fmt : str
     String format for colorbar values.
 """
-docdict["topomap_mask"] = """
-mask : ndarray of bool, shape (n_channels, n_samples) | None
-    The channels to be marked as significant at a given time point.
-    Indices set to ``True`` will be considered. Defaults to ``None``.
-"""
+mask_base = """
+mask : ndarray of bool, shape %(shape)s | None
```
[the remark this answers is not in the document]
mscheltienne: Done with `.format(**dict)`.

mne/utils/docs.py
mscheltienne: I made it a bit more compact, but it's a pain to keep it readable in the limit of 79 characters.

examples/visualization/evoked_topomap.py
mscheltienne: Yes, it is necessary to set the `True` mask value at the indices picked by plot_topomap. Without the `-1`, none of the 3 plots have a channel highlighted because the indices set to True on the times axis are off by one.

mmagnuski: Then it is specific to the time values in the example and may not work with other `.times` (as the ones in my example). It would be better then to find the closes timepoint to the one requrested using:
```python
_times = [(np.abs(evoked.times - t)).argmin() for t in times]
```
(bear in mind that people will likely copy that code and expect it to work on their data, where `.times` might be different)

drammock: I agree with @mmagnuski.  I included the `- 1` in my original example [here](https://github.com/mne-tools/mne-python/issues/9611#issuecomment-891966296) as a quick-and-dirty solution; using `argmin` is more robust.

mscheltienne: True, I will update later today.

drammock:
@mscheltienne notice how `n_patterns` is colored pink in the docstring for `plot_patterns()`, while everything else is either blue or black.  Words after the colon on the first line of a docstring entry are assumed to be linkable cross-references to data types (None, book, ndarray, etc) and are colored blue when successful.  Exceptions are explicitly listed in `doc/conf.py` and get printed in black.

I've added a commit that will fix this, as well as a typo fix in the docdict and a small refactor to make the new docdict entries more compact.

mscheltienne:
@drammock Thank you! Now I get it, and this is why `n_times` was doing fine, it was already in the exceptions. 
And thanks for all the small remaining typos correction, there were quite a few left!

mmagnuski:
The [example](https://32460-1301584-gh.circle-artifacts.com/0/dev/auto_examples/visualization/evoked_topomap.html#more-advanced-usage) looks good!

mscheltienne:
@mmagnuski Side note: is it useful to have the outputs `Removing projector [..]` in the example? It brings nothing, and I am not even sure why MNE outputs this in the first place when you are plotting a topographic map.

drammock:
> @mmagnuski Side note: is it useful to have the outputs `Removing projector [..]` in the example?

Generally speaking we want the output shown in our tutorials/examples to match what a user would see if running the code locally. So yes, it is useful.

> I am not even sure why MNE outputs this in the first place when you are plotting a topographic map.

Because of the `ch_type='mag'` argument. Internally the plotting function will call a channel picking function, and part of channel picking is keeping the projectors current with regard to the channels that are still present.

> Maybe adding a `mne.set_log_level('ERROR')` might improve the example?

If we do that, then we need to explain why we're doing it, and it would be code that is not strictly needed for the tutorial's pedagogical purpose, and that would have effects beyond just the tutorial's code when run locally (the logging level would be set that way for the remainder of the interactive session).

If you're really concerned about the messages, you could add some lines near the beginning of the tutorial that create copies of the evoked and pick the channel type in advance (concentrating those log lines all in one cell).  then instead of `evoked.plot_topomap()` you would call `evoked_mag.plot_topomap()` or `evoked_eeg.plot_topomap()` etc. and the log lines should no longer appear after each plot.  But personally I'm `-1` on that change because it doesn't reflect how users are likely to do this kind of thing during interactive data exploration --- I think they're more likely to just work off the `evoked` instance that has all channel types, and pick the types they need for a given plot.

drammock (approved):

drammock:
@mmagnuski are you satisfied? (if so, please re-review to remove the "changes requested" flag)

mscheltienne:
@drammock Thanks for the great explanation. It really helps to understand the vision and the design you are aiming for, and I hope it willé help for future PR ;)

mmagnuski (approved):

drammock:
in it goes!  thanks @mscheltienne!

mmagnuski:
Thanks @mscheltienne! 🚀

PR #9656, opened 2021-08-09
Remove executable bits from regular files

cbrnr:
There are several files which should not have their executable bit set (thanks `tree -C` for letting me know).


cbrnr:
All green except for LGTM Javascript. The only file I changed that could be related is `doc/_template/layout.html`, but IMO this file shouldn't be executable.

drammock:
Are you certain that the files in `commands/` can safely be `-x`?  I guess the tests ought to have caught it if it were a problem, but it's suspicious that so many (all?) were executable

cbrnr:
> Are you certain that the files in `commands/` can safely be `-x`? I guess the tests ought to have caught it if it were a problem, but it's suspicious that so many (all?) were executable

I think so, except if they are supposed to be executed directly (e.g. inside the `commands/` directory with `./mne_anonymize.py`). So yeah, this looks suspicious, but the following files were *not* executable so I thought they will not be executed directly:

- `mne_anonymize.py`
- `mne_flash_bem.py`
- `mne_prepare_bem_model.py`
- `mne_setup_forward_model.py`
- `mne_setup_source_space.py`
- `mne_sys_info.py`
- `mne_watershed_bem.py`
- `mne_what.py`
- `utils.py`

The Pythonic way to add executables is to list them in the `entry_points` array, that way users wouldn't even have to type the `.py` extension.

drammock (approved):

sappelhoff (commented):
lgtm, but how did you remove the "executable bits" from the files? I.e., which command(s) did you run to get the diff in this PR?

cbrnr:
What do you mean get the diff? I just did `ls -l` and `chmod a-x` where necessary.

sappelhoff (approved):
> I just did ls -l and chmod a-x where necessary.

okay, I thought you had some kind of a one liner. Thanks!

drammock:
thanks @cbrnr

cbrnr:
> okay, I thought you had some kind of a one liner. Thanks!

Nope, that wouldn't work because there's too much manual checking involved. But `tree -C` produces a nice overview (executable files are green with my settings).

PR #9663, opened 2021-08-13
Fix plot compare evokeds

drammock:
as reported on the forum, `plot_compare_evokeds` is failing to draw confidence bands when only one condition is passed.

https://mne.discourse.group/t/plot-compare-evokeds-from-mne-evokedarray-confidence-intervals-and-ylim/3522/3

This adds a test that fails on main and passes on this PR.

### DIFF
--- a/mne/viz/evoked.py
+++ b/mne/viz/evoked.py
@@ -2361,8 +2361,11 @@ def plot_compare_evokeds(evokeds, picks=None, colors=None,
         ci_dict = dict()
         for cond in conditions:
             this_evokeds = evokeds[cond]
-            # skip CIs when possible; assign ci_fun first to get arg checking
+            # assign ci_fun first to get arg checking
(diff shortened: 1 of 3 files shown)


mne/viz/evoked.py
```diff
@@ -2363,7 +2363,6 @@ def click_func(
             this_evokeds = evokeds[cond]
             # skip CIs when possible; assign ci_fun first to get arg checking
             ci_fun = _get_ci_function_pce(ci, do_topo=do_topo)
-            ci_fun = ci_fun if len(this_evokeds) > 1 else None
```
drammock, line 2366: I don't know why this condition was there in the first place. I think it was just a mistake.

drammock: turns out this *was* there for a good reason, but it just shouldn't apply if the user passes a custom CI function.

agramfort (approved):

agramfort:
thx @drammock !

PR #9664, opened 2021-08-16
MAINT: fix circle

drammock:
- workaround upstream bugs in nilearn & quantities
- when setting up circle, don't install mayavi & pysurfer only to uninstall them 2 lines later
- avoid building matplotlib from source when setting up circle
- use `apt` instead of `apt-get` (seems to fix problems due to recent debian-buster version bump, not sure why)

PR #9665, opened 2021-08-16
Fix circle

drammock:
revives #9664, which couldn't be re-opened because I force-pushed while it was closed.


drammock:
OK, circle is happy.  In it goes!

larsoner:
Thanks @drammock !

PR #9667, opened 2021-08-16
MAINT: more circle fixes

drammock:
something about seaborn 0.11.2 is making `tutorials/epochs/50_epochs_to_data_frame.py` time out.  I can reproduce locally and it happens on circle.  Nothing in the release notes points to an obvious culprit, so until someone has time to track it down, let's just avoid that version.


drammock:
circle green, in it goes!

PR #9672, opened 2021-08-18
Add setter for fiducial alignment strategy to Coregistration

drammock:
- Allows `mne.coreg.Coregistration.set_fid_match`
- Fixes bug when `Coregistration._icp_fid_match` == `'matched'` (now looks for attribute `self._lpa` instead of `self.lpa`, which doesn't exist)

### DIFF
--- a/mne/coreg.py
+++ b/mne/coreg.py
@@ -39,7 +39,7 @@ from .transforms import (rotation, rotation3d, scaling, translation, Transform,
                          rot_to_quat, _angle_between_quats)
 from .utils import (get_config, get_subjects_dir, logger, pformat, verbose,
                     warn, has_nibabel, fill_doc, _validate_type,
-                    _check_subject)
+                    _check_subject, _check_option)
 from .viz._3d import _fiducial_coords
 
 # some path templates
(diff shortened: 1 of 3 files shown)


larsoner:
Can you add one more entry `fid_match` to the parametrization list in `test_coreg.py` that would have caught this? https://github.com/mne-tools/mne-python/blob/a14e7e35cbbc5f55585492a8993196198dc392b7/mne/tests/test_coreg.py#L285-L292

larsoner:
(it can be False for every case except one for example and that's good enough)

mne/tests/test_coreg.py
```diff
@@ -282,15 +282,17 @@ def test_get_mni_fiducials():
...
-def test_coregistration(scale_mode, ref_scale, grow_hair, fiducials):
+@pytest.mark.parametrize(
+    'scale_mode,ref_scale,grow_hair,fiducials,fid_match', [
+        (None, [1., 1., 1.], 0., None, 'nearest'),
+        (None, [1., 1., 1.], 0., 'estimated', 'nearest'),
+        (None, [1., 1., 1.], 2., 'auto', 'nearest'),
+        ('uniform', [1., 1., 1.], 0., None, 'nearest'),
+        ('3-axis', [1., 1., 1.], 0., 'auto', 'nearest'),
+        ('uniform', [0.8, 0.8, 0.8], 0., 'auto', 'nearest'),
+        ('3-axis', [0.8, 1.2, 1.2], 0., 'auto', 'nearest')])
```
larsoner, line 293: These are all `'nearest'`, one needs to be `'matched'`, right?

drammock: 🤦🏻

mne/coreg.py
```diff
@@ -1756,6 +1756,22 @@ def _setup_icp(self, n_scale_params):
...
+        match : 'nearest' | 'matched'
+            Alignment strategy; ``'nearest'`` aligns anatomical landmarks to
+            any point on the head surface; ``'matched'`` aligns to the fiducial
+            points in the MRI.
+        """
+        if match not in self._icp_fid_matches:
+            msg = (f'"match" must be one of {self._icp_fid_matches}, got '
+                   f'{match}')
+            raise ValueError(msg)
+        self._icp_fid_match = match
```
agramfort, line 1773: can you make this function return self liek the other setters so we can chain them in one line?

drammock:
```suggestion
        self._icp_fid_match = match
        return self
```

mne/coreg.py
```diff
@@ -1756,6 +1756,23 @@ def _setup_icp(self, n_scale_params):
...
+        ----------
+        match : 'nearest' | 'matched'
+            Alignment strategy; ``'nearest'`` aligns anatomical landmarks to
+            any point on the head surface; ``'matched'`` aligns to the fiducial
+            points in the MRI.
+        """
+        if match not in self._icp_fid_matches:
+            msg = (f'"match" must be one of {self._icp_fid_matches}, got '
+                   f'{match}')
+            raise ValueError(msg)
```
larsoner, line 1772: Should be
```suggestion
        _check_option('match', match, self._icp_fid_matches)
```
But also needs an import at the top

larsoner:
Snuck in a CircleCI fix and a couple of code cleanups I noticed. Will merge once CIs come back happy, thanks @drammock

agramfort (approved):

PR #9674, opened 2021-08-18
MAINT: Fix Circle

larsoner:
Trying to fix/diagnose https://app.circleci.com/pipelines/github/mne-tools/mne-python/9675/workflows/78521055-7bf4-4549-a072-61fd6e93af74/jobs/32901


mne/viz/circle.py
```diff
@@ -119,7 +119,6 @@ def _plot_connectivity_circle_onpick(event, fig=None, axes=None, indices=None,
         fig.canvas.draw()
 
 
-@deprecated(CONNECTIVITY_DEPRECATION_MSG)
 def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
```
larsoner, line 122: @adam2392 @agramfort I think we need to un-deprecate this because `plot_channel_labels_circle` calls `plot_connectivity_circle` under the hood.

Another option would be to still deprecate `plot_connectivity_circle` but make it private in MNE-Python. That might be slightly cleaner, not sure. Thoughts on which is better?

adam2392: I'm okay with un-deprecate it and move it back into MNE-python if needed.

mne/viz/circle.py
```diff
@@ -222,6 +222,25 @@ def plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
...
     """
+    return _plot_connectivity_circle(
+        con, node_names, indices, n_lines, node_angles, node_width,
+        node_colors, facecolor, textcolor, node_edgecolor, linewidth, colormap,
+        vmin, vmax, colorbar, title, colorbar_size, colorbar_pos,
+        fontsize_title, fontsize_names, fontsize_colorbar, padding,
+        fig, subplot, interactive, node_linewidth, show)
+
+
+def _plot_connectivity_circle(con, node_names, indices=None, n_lines=None,
```
larsoner, line 233: @adam2392 @agramfort I decided to go the private route -- it seemed weird to keep `plot_connectivity_circle` in `mne` rather than `mne-connectivity`. Ideally at some point we should make it so that this code is not duplicated with `mne-connectivity`, maybe by having `mne-connectivity.viz.plot_connectivity_circle` call this private function for example

larsoner:
Ugh, now we have:
```
/home/circleci/project/mne/viz/circle.py:docstring of mne.viz.circle.plot_channel_labels_circle:20: WARNING: py:func reference target not found: mne.viz.plot_connectivity_circle
```
Should we move `plot_channel_labels_circle` to `mne_connectivity`? Another solution is to write:
```
    **kwargs : kwargs
        Keyword arguments for :func:`mne-connectivity:mne_connectivity.viz.plot_connectivity_circle`.
```
But it's weird to do this because people might think they need `mne_connectivity` to run this function, but they don't...

larsoner:
`plot_channel_labels_circle` was added within cycle by @alexrockhill . I'm thinking we should move that to `mne_connectivity`, then we can preserve our deprecation of `plot_connectivity_circle` (and not need to make it private or anything). Okay with you @alexrockhill @adam2392 ? If so I'll move the code.

In the meantime I'm going to push a commit just to get this PR to pass so that I can merge and get `main` to be green again.

drammock (approved):

alexrockhill:
Hmmm it's not really a connectivity thing, it just uses the same plot format...

agramfort (approved):

agramfort:
thx @larsoner

larsoner:
> Hmmm it's not really a connectivity thing, it just uses the same plot format...

One clean alternative is to refactor `plot_channel_labels_circle` to just use `circular_layout` internally rather than `plot_connectivity_circle`. @alexrockhill can you look into this solution if you agree it's cleaner?

alexrockhill:
Sure that would be nice, I used a lot of the helper functionality in `plot_connectivity_circle` though...

alexrockhill:
Hmmm yeah, it would be basically copy-pasting the whole function over... not sure that makes much sense

alexrockhill:
One of the things that might help is to put more of `plot_connectivity_circle` in `circular_layout`. If there is nothing else that uses it, that might just make it easier to use

Issue #9680, opened 2021-08-20
MAINT: plot_alignment newly failing examples


drammock:
@alexrockhill can you take a look at these CI failures from `plot_alignment`?  Looks related to your recent work.

`ValueError: A head->mri transformation matrix is required to plot pial surfaces, `trans=None` is not allowed`

https://app.circleci.com/pipelines/github/mne-tools/mne-python/9716/workflows/54dba24d-162b-43ba-87ba-3884689e8574/jobs/33025

alexrockhill:
Fixing it now

PR #9682, opened 2021-08-20
[MAINT + ENH, MRG] Fix identity transforms

alexrockhill:
Fixes https://github.com/mne-tools/mne-python/issues/9680

I guess there were more than a few examples where people assumed identity transforms...


examples/forward/left_cerebellum_volume_source.py
```diff
@@ -46,10 +49,11 @@
 # %%
 # Plot the positions of each source space
 
-fig = mne.viz.plot_alignment(subject=subject, subjects_dir=subjects_dir,
+fig = mne.viz.plot_alignment(trans=trans, subject=subject,
+                             subjects_dir=subjects_dir,
                              surfaces='white', coord_frame='head',
                              src=src)
```
larsoner, line 55: Rather than pass+use an incorrect trans here, I'd rather do `coord_frame='mri'`

larsoner (commented):
Other than an idea for one simplification LGTM

examples/forward/left_cerebellum_volume_source.py
alexrockhill: After I had adjusted the roll and everything ;)

alexrockhill:
Ok should be good to go here, I forgot to iterate over views which we actually used in the ECoG tutorial, my mistake!

larsoner:
Thanks @alexrockhill !

PR #9685, opened 2021-08-21
MRG: avoid warning in SURE when passing grid & allow to return empty stc

agramfort:
cc @PABannier 

let me know what you think

### DIFF
--- a/mne/inverse_sparse/mxne_inverse.py
+++ b/mne/inverse_sparse/mxne_inverse.py
@@ -12,7 +12,7 @@ from ..forward import is_fixed_orient
 from ..io.pick import pick_channels_evoked
 from ..io.proj import deactivate_proj
 from ..utils import (logger, verbose, _check_depth, _check_option, sum_squared,
-                     _validate_type, check_random_state)
+                     _validate_type, check_random_state, warn)
 from ..dipole import Dipole
 
 from .mxne_optim import (mixed_norm_solver, iterative_mixed_norm_solver, _Phi,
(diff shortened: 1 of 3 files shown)


mne/inverse_sparse/mxne_inverse.py
```diff
@@ -441,8 +442,9 @@ def mixed_norm(evoked, forward, noise_cov, alpha='sure', loose='auto',
 
     # Alpha selected automatically by SURE minimization
     if alpha == "sure":
-        alpha_grid = (np.geomspace(100, 10, num=15)
-                      if sure_alpha_grid == "auto" else sure_alpha_grid)
+        alpha_grid = sure_alpha_grid
+        if isinstance(sure_alpha_grid, str) and sure_alpha_grid == "auto":
+            alpha_grid = np.geomspace(100, 10, num=15)
```
PABannier, line 447: I don't know if it still makes sense to use np.geomspace, compared to np.arange. np.geomspace results in a sample with a lot of values near 10, and fewer around 100. I'm not sure it is what we want. WDYT?

PABannier:
For the empty STC + warning part, LGTM!

mne/inverse_sparse/mxne_inverse.py
agramfort: you mean np.linspace? can you test if np.linspace works better on some datasets @PABannier ?

PABannier: yes np.linspace. I have already tried it. I did not notice any change in performance. in some ways, np.linspace does encourage sparse solutions and might be faster.

Issue #9686, opened 2021-08-22
[WIP, ENH]: pyqtgraph as backend for 2D-Data-Browser


marsipu:
# pyqtgraph as backend for 2D-Data-Browser
In my 2021 Google Summer of Code project I worked on providing an additional backend for the 2D-Data-Browser currently implemented with matplotlib for `Raw.plot()`, `Epochs.plot()` and `ICA.plot_sources()`. You can find the final report [here]()
This is the [PR]() which contains the new backend.

In this issue I will host an updated list with the tasks left to be done and ideas for future features in the new pyqtgraph-backend.
I would be much interested in feedback about the desing & features of the new backend and/or new ideas for things you would like to see in the new backend. If you got some, leave them in the comments.

### Tests
- [ ] make all tests in `mne.viz.tests.test_raw.py` pass for pyqtgraph

### Features
- [x] Event Lines
- [x] Scale-Bars/Legends
- [x] Apply projections with GUI
- [ ] Channel-Clipping
- [ ] Settings-Dialog with pyqtgraph-specific settings

### Improvements (from feedback)
- [x] Crosshair without horizontal line but dot

## Other features from MPL
- [ ] adaption to `Epochs.plot()` (+ test_epochs.py)
- [ ] adaption to `ICA.plot_sources()` (+ test_ica.py)
- [ ] Per-channel-annotation (for Epochs)
- [ ] Selection-Figure (group_by-parameter)

## New feature ideas
- [ ] Right-Click Dragging for scaling
- [ ] Pinching for zooming with trackpad
- [ ] custom colors for annotations (like in event-colors, needs to be adapted in mne-python)
- [ ] Dark Mode (preferably with parameter accepting dict for colors to be adjusted by user)
- [ ] Filter view-range dynamically (set filter-parameters in gui)
- [ ] Initialize Browser without data (and load data from menu/with drag&drop)
- [ ] Start Data-Browser from command-line

marsipu:
Moved to [here](https://github.com/mne-tools/mne-qt-browser/projects/2?add_cards_query=is%3Aopen)

PR #9687, opened 2021-08-22
[ENH]: New pyqtgraph-backend for 2D-Data-Browser

marsipu:
# About this PR
This PR adds the new pyqtgraph-backend I developed during 2021's Google Summe of Code.
It covers basic functionality and is further developed in [mne-tools/mne-qt-browser](https://github.com/mne-tools/mne-qt-browser).
This PR is WIP and remaining ToDo's as well as ideas for improvements and new features can be submitted as issues in `mne-tools/mne-qt-browser`.
I would be glad about feedback for the new backend. I provided example code for testing below.

## Example code
```python
import mne
import os
import numpy as np

from mne.viz import use_browser_backend

# Load Raw
raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
                        'sample_audvis_raw.fif')
raw = mne.io.read_raw(raw_file)

# Load Events
events_path = os.path.join(sample_data_folder, 'MEG', 'sample',
                           'sample_audvis_raw-eve.fif')
events = mne.read_events(events_path)

# Add test-annotations
onsets = np.arange(2, 8, 2) + raw.first_time
durations = np.repeat(1, len(onsets))
descriptions = ['Test1', 'Test2', 'Test3']
for onset, duration, description in zip(onsets, durations, descriptions):
    raw.annotations.append(onset, duration, description)

with use_browser_backend('pyqtgraph'):
    raw.plot(events=events, block=True)
```

## Benchmark pyqtgraph-specific parameters
If you would like to compare performance of pyqtgraph-specific parameters (use of OpenGL, antialiasing, downsampling), you can fork my [gsoc2021-repository](https://github.com/marsipu/gsoc2021) and run `python -m prototypes` from the main directory.

## Showcase
https://user-images.githubusercontent.com/46245704/136716490-19768b32-e5e3-4da1-b811-9bab8b0ef589.mp4


agramfort:
to test should I just pip install git+https://github.com/pyqtgraph/pyqtgraph@master?

@cbrnr @marsipu what do you think if putting this feature in a different repo for now?
my recent experience with other mne side projects is that it really speeds up development
as you don't have to deal with mne-python full test suite etc.

cbrnr:
> @cbrnr @marsipu what do you think if putting this feature in a different repo for now?
my recent experience with other mne side projects is that it really speeds up development
as you don't have to deal with mne-python full test suite etc.

@larsoner @drammock you were more involved in the backend changes, what do you think?

marsipu:
> to test should I just pip install git+https://github.com/pyqtgraph/pyqtgraph@master?

Yes, and I would also recommend pyopengl because it improves performance a lot.

> @cbrnr @marsipu what do you think if putting this feature in a different repo for now?
> my recent experience with other mne side projects is that it really speeds up development
> as you don't have to deal with mne-python full test suite etc.

The goal for this PR would be to make the complete viz-test-suite pass for pyqtgtaph.
I couldn't finish that during GSoC, but after my 2 week holiday now I will try to find time to focus on that first.
If you still want this process outside I could continue in my gsoc2021-repository.
What do you prefer?

agramfort:
I think there is no rush here.

drammock:
> > @cbrnr @marsipu what do you think if putting this feature in a different repo for now?
> > my recent experience with other mne side projects is that it really speeds up development
> 
> @larsoner @drammock you were more involved in the backend changes, what do you think?

We discussed this at our last meeting, and @larsoner and I agreed that:

1. we want lots of folks to test out with their own data (and their own particular usage habits)
2. testing is easier if all people need to do is `pip install` 2 packages, check out this fork+branch, and copy-paste the example script above

I'm not too worried about speeding up development... @marsipu has a really nice prototyping framework at his other repo that runs automated performance benchmarks, and has been good about testing locally and only pushing large batches of commits so that our CIs don't get overwhelmed / backlogged.

marsipu:
Ok, when I get on with the test and the other ToDos for this PR I will commit them to the gsoc2021-repository.
Or would [ci skip] also suffice to save CI-resources until all tests are adapted? _(That may bring the advantage of tracking back the single commits in this PR)_

marsipu:
How would you suggest gathering feedback and potential bugs? Can you edit my list in #9686? Or can I make it editable?

larsoner:
> Or would [ci skip] also suffice to save CI-resources until all tests are adapted?

I would just push commits here with `[ci skip]` until you expect them to be green

> How would you suggest gathering feedback and potential bugs? Can you edit my list in #9686? Or can I make it editable?

Yes maintainers can edit issues. But in general it's better if you keep track most of the time @marsipu

larsoner:
@marsipu why is pyqtgraph `master` required? Is it just for turning off antialiasing or something? It would be nice if it's not required and we only use `master` functionality if it's present (e.g., by using `mne.fixes._get_args` or `mne.utils.check_version` to decide).

Do you know the minimal version of pyqtgraph that's needed? At the very least we'll need to add that to `README.md` in the optional dependencies section.

larsoner:
... If it's just for OpenGL to work, given that it doesn't work on all platforms I'd leave OpenGL disabled by default, and let people enable it who have usable OpenGL graphics on their machines and want to install PyQtGraph master. It would be best if the code snippet worked for people without OpenGL and without needing latest `master`.

drammock:
> ... If it's just for OpenGL to work, given that it doesn't work on all platforms I'd leave OpenGL disabled by default

I think the reason to use `master` is that they recently fixed it so that OpenGL now works on Windows, at least I think that's what @marsipu mentioned at our last meeting.  So it ought to work on all platforms now.

larsoner:
I would disable OpenGL by default and have people opt-in to trying it. Based on the number of failures we see for 3D viz I think it's the safest option.

marsipu:
@larsoner Yes, it is as @drammock says, there are two OpenGL-related changes in the current dev-version from pyqtgraph:
1. fix bug breaking OpenGL-Plots on Windows
2. make antialiasing disableable (my PR)

I hope there will be another release of pyqtgraph soon covering these changes.

On the machines I tested I had no other problems with OpenGL, but it seems reasonable to disable OpenGL by default since fixing problems with it seems to be complicated.
We might set it as a config-value then since potential problems are probably often hardware-dependent. WDYT?

cbrnr:
OpenGL is so much faster that I really like to have it turned on by default. We haven't seen any issues so far, so unless there are any known bugs there is no reason to disable it. I think it is also fine to require the `master` branch for now, because the new browser backend lives only within a PR and hasn't been integrated into MNE yet. I'm pretty confident that the PyQtGraph devs will release a new version soon anyway.

mne/viz/_pg_figure.py
```diff
@@ -0,0 +1,2296 @@
...
+            Defaults to "zscore".
+        """
+        self.pg_kwarg_defaults = dict(duration=20,
+                                      n_channels=30,
+                                      highpass=None,
+                                      lowpass=None,
+                                      ds='auto',
+                                      ds_method='peak',
+                                      antialiasing=False,
+                                      use_opengl=True,
```
marsipu, line 1233:
```suggestion
                                      use_opengl=False,
```

marsipu:
I agree to both points @larsoner and @cbrnr.
Maybe setting the default on the final integrated backend would depend on the experience you testers have with the version from this PR? Should I maybe activate openGL just for this testing-period by default to see if anyone got problems with it?

agramfort:
let me try to bring some arguments to start with a dedicated repo:
- you could have a test suite that runs in 5 mins
- you can more easily test different pyqtgraph version (stable, dev, with
opengl etc.)
- you can release a lot more often.

and I am +100 to make any relevant change to mne-python upstream to keep
the browser
code simpler like you did with the major refactoring you did on matplotlib
backend.

>

marsipu:
> Will you and @cbrnr be actively maintaining this code in mne-python?

I want to be available to maintain the functionality of the browser with pyqtgraph and to adapt to possible changes in the pyqtgraph-API in the future. I hope, that there won't be major API changes in pyqtgraph in future which would require rewriting a lot of the code.  But you're right, I don't know that yet. I am also looking forward to see how much work still has to be done after you developers tested the backend and found bugs and/or ideas for improvement.

agramfort:
I just manage to give it a try after updating the packaging.

it works really fast and it's already pretty usable congrats !

I see a few UI glitches when playing with annotations (selecting what is visible, removing one annotation etc.)
I am also not sure all shortcuts are working. Should it already?

great job @marsipu

agramfort:
also I see now a number of traceback errors in the terminal. Eg.

```

IndexError                                Traceback (most recent call last)
~/work/src/mne-python/mne/viz/_pg_figure.py in _remove_description(self)
    948                     self.mne.inst.annotations.description == rm_description)
    949                 for idx in rm_idxs:
--> 950                     self.mne.inst.annotations.delete(idx)
    951                 for rm_region in [r for r in self.mne.regions
    952                                   if r.description == rm_description]:

~/work/src/mne-python/mne/annotations.py in delete(self, idx)
    378         self.duration = np.delete(self.duration, idx)
    379         self.description = np.delete(self.description, idx)
--> 380         self.ch_names = np.delete(self.ch_names, idx)
    381
    382     def to_data_frame(self):

<__array_function__ internals> in delete(*args, **kwargs)

~/miniconda3/lib/python3.8/site-packages/numpy/lib/function_base.py in delete(arr, obj, axis)
   4404         else:
   4405             keep = ones(N, dtype=bool)
-> 4406             keep[obj,] = False
   4407
   4408         slobj[axis] = keep

IndexError: index 15 is out of bounds for axis 0 with size 15
```

marsipu:
Thank you @agramfort 

> I see a few UI glitches when playing with annotations (selecting what is visible, removing one annotation etc.)

I missed some bugs there which I fixed now. The cause for the traceback though came from the sample code I supplied (I forgot to set annotations.ch_names). I updated the sample code now above.

mne/viz/_pg_figure.py
```diff
@@ -0,0 +1,2409 @@
...
+from PyQt5.QtTest import QTest
+from PyQt5.QtWidgets import (QAction, QColorDialog, QComboBox, QDialog,
+                             QDockWidget, QDoubleSpinBox, QFormLayout,
+                             QGridLayout, QHBoxLayout, QInputDialog,
+                             QLabel, QMainWindow, QMessageBox,
+                             QPushButton, QScrollBar, QSizePolicy,
+                             QWidget, QStyleOptionSlider, QStyle,
+                             QApplication, QGraphicsView, QProgressBar,
+                             QVBoxLayout, QLineEdit, QCheckBox, QScrollArea)
+from pytestqt.exceptions import capture_exceptions
```
agramfort, line 27: pytestqt is not just for testing? as done here it must be installed for everyone.

mne/viz/_pg_figure.py
```diff
@@ -0,0 +1,2409 @@
...
+
+from ._figure import BrowserBase
+from ..annotations import _sync_onset
+from ..io.pick import _DATA_CH_TYPES_ORDER_DEFAULT
+from ..utils import logger
+from pyqtgraph import (AxisItem, GraphicsView, InfLineLabel, InfiniteLine,
+                       LinearRegionItem,
+                       PlotCurveItem, PlotItem, TextItem, ViewBox, functions,
+                       mkBrush, mkPen, setConfigOption, mkQApp, mkColor)
+from scipy.stats import zscore
```
agramfort, line 37: you need to update import order from more standard to local imports.

agramfort (commented):
a few comments but otherwise

it really really well for me !

many other devs from @mne-tools/mne-python-steering-committee can try too?

wmvanvliet:
Holy cow, this thing flies! A huge improvement over the matplotlib based renderer. :clap: 

I get an error message when selecting different annotation types and changing their color:

```
<mne.viz._pg_figure.PyQtGraphBrowser at 0x7f535f9c2940>Traceback (most recent call last):
  File "/m/home/home4/45/vanvlm1/data/projects/mne-python/mne/viz/_pg_figure.py", line 1067, in _set_color
    self.main._update_regions_colors()
AttributeError: 'PyQtGraphBrowser' object has no attribute '_update_regions_colors'
```

When I then select the color again, it works properly.

The speed of this browser begs fore some UI features that weren't feasible before, such as click+drag in the main window to scroll around, using scrollwheel+modifier key to zoom.

wmvanvliet:
oh, and can `=` also increase the scale? Then I don't have to press shift :)

marsipu:
@wmvanvliet Thank you for your feedback:)
I'll look into it during the coming days when I am back from holidays.

mne/viz/_pg_figure.py
marsipu: Oh right. But it is only used for testing. I will create a dummy to make it an optional dependency we could then put inside `requirements_testing.txt`

marsipu:
> The speed of this browser begs fore some UI features that weren't feasible before, such as click+drag in the main window to scroll around, using scrollwheel+modifier key to zoom.

Yes, thank you for the ideas. I thought about right-click click+drag to adjust shown channels(veritical) and duration(horizontal). WDYT?

wmvanvliet:
Sounds good!

marsipu:
As you suggested I moved the backend-code now into a [peripheral repository](https://github.com/marsipu/mne_pg_browser).
You will be asked for installation when "pyqtgraph" is selected as backend.

agramfort:
ok to move this repo to the mne-tools org @marsipu ?

if so give me admin access to this repo so I can transfer it.

marsipu:
> 
> 
> ok to move this repo to the mne-tools org @marsipu ?
> 
> if so give me admin access to this repo so I can transfer it.

Sure

mne/viz/_figure.py
```diff
@@ -443,9 +443,9 @@ def _load_backend(backend_name):
         backend = importlib.import_module(name=_backends[backend_name],
                                           package='mne.viz')
     else:
-        repo_link = "https://github.com/marsipu/mne_pg_browser/zipball/main"
+        repo_link = "https://github.com/mne-tools/mne_qt_browser/zipball/main"
```
agramfort, line 446:
```suggestion
        repo_link = "https://github.com/mne-tools/mne-qt-browser/zipball/main"
```

marsipu:
You see, I accidentally added commits from the main-branch.
I wanted to avoid to use `git push --force` again, because it resets the history in the PR.

But when I want to rebase my local branch onto `upstream/main`, pushing to origin is not allowed until I use `push --force` or rebase onto origin again which seems to result into adding somehow the commits from the main-branch to the history as above.

**What do you do to update your PR-branch with the changes from main? Is it possible without `push --force`?**
One solution is probably using the button from GitHub, but I am curious if it is also possible locally.

agramfort:
don't bother @marsipu we'll squash commits in the end. If the diff is clean we're good.

marsipu:
# Raw-Tests passing
Hello everyone,
I worked the past weeks on getting the pyqtgraph-backend to pass all the tests in `test_raw.py`. While doing that I added some missing features (Scale-Bars, Events, Projections-GUI, Selections-GUI) and reworked the Overview-Bar (just reimplementing paintEvent took too much performance, now it is also GraphicsView-based as the main-plot). Those changes can be seen in the [mne-qt-browser-repo](https://github.com/mne-tools/mne-qt-browser).
I finally got now to the point where all tests from `test_raw.py` pass  (at least locally on my Windows-machine) . But I am struggeling with the CIs:

## Failing CIs
### Failing tests on other OS
There still appear to be problems with `test_plot_raw_selection`,  `test_plot_raw_ssp_interaction` and/or `test_min_window_size` on other OS though. I'll try to get access to some machines with Linux/MacOS to figure those out too when I got time in the laboratory again (I will be off for a week now).

### Compatibility tests
I see that there are tests checking compatibility with older python/mne-versions. The problem is, that pyqtgraph needs at least `PyQt5 >= 5.12` and some of these tests still have `5.9`. How do we tackle this best?

### Docs failing
For some reason building the docs with `html_dev-noplot` is returning with error, on my machine as well in the CIs. In the CIs, I can't see the reason, maybe it is the same as on my machine:
grafik
I don't understand why this line in `mne-python\doc\conf.py` is failing and if/how that could be related to the changes from this PR. Could you help me there?

agramfort:
@larsoner you have anytime to look?

>

larsoner:
> The problem is, that pyqtgraph needs at least PyQt5 >= 5.12 and some of these tests still have 5.9. How do we tackle this best?

Any test that fails to import an optional module (in your case, `pyqtgraph`) should be skipped using an appropriate decorator/skip. With the backend framework in place, you should be able to try/except an import of pyqtgraph, and `pytest.skip` if it's not available. You can add a version check for Qt in there, too, if you need to. This should only really affect the `compat / minimal` CI in principle.

> For some reason building the docs with html_dev-noplot is returning with error, on my machine as well in the CIs.

The warning I see in CircleCI is:
```
/home/circleci/project/mne/io/eeglab/eeglab.py:docstring of mne.io.eeglab.eeglab.read_raw_eeglab:26: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```
So I don't see the failure there.

Locally it's probably because you're building on Windows. We should add a CI entry to at least `make html_dev-noplot` on Azure Windows. I can do that later this week and fix that bug at the same time.

>  I'll try to get access to some machines with Linux/MacOS to figure those out too when I got time in the laboratory again (I will be off for a week now).

MacOS will be a bit of a pain, but it should be fairly easy for you to set up an Ubuntu Linux guest VM (using your Windows machine as a host) and quickly get things running. I think there are pre-built VirtualBox Ubuntu images online; if you get one of these and run some/most of our Linux CI install steps (probably won't need the xvfb-ones?) then you should get to a very similar state as our remote CIs very quickly, no other physical machine required.

Also note that even Windows has some errors on CIs, e.g.:

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=15465&view=logs&jobId=fa98c7b3-575e-55b6-6946-22dea6f6daf8&j=fa98c7b3-575e-55b6-6946-22dea6f6daf8&t=c0866e65-eed5-5461-eadc-f0ff150eefd8

rob-luke:
Ok I'll switch over and use this as my daily driver for a few days.

marsipu:
This is a screencast with some of the enhanced features:

https://user-images.githubusercontent.com/46245704/136716490-19768b32-e5e3-4da1-b811-9bab8b0ef589.mp4

wmvanvliet:
I'm currently using this to manually annotate bad segments in a large-ish dataset. This new browser makes it so much easier. The overview bar helps a lot, because I can easily see the boundaries of the experiment (where there are lots of events) and when there are breaks.

One thing I do notice is that sometimes when zooming in/out, the channel labels disappear. Scrolling brings them back. Could there be some race condition that looses track of the channel--label positions when zooming?

wmvanvliet:
promo-vid needs some background music: https://www.youtube.com/channel/UCht8qITGkBvXKsR1Byln-wA

rob-luke:
> It's the bottom bar red/blue thing.

haha exactly, so far my understanding is that its some magical red blue bars. Is there a description somewhere? Is it appropriate for all data types?

drammock:
> 3. You could also toggle event-lines and annotations with pressing "e" or "Shift+a", WDYT?

I would still prefer to have the event lines at a lower stacking order than the data (in both the data window and in the overview bar).  But I withdraw the suggestion to show the event lines only in the data window.  The overview bar should show the same elements (events, annotations) shown in the data window.

> 6. since the main-plot and the Channel-Axis are two different scenes, it needed a more complicated workaround if you would like the label inside the Channel-Axis. Maybe we can find an even better solution (e.g. some kind of legend?)

If it's a big pain then maybe not worth it.  I'd rather leave the scalebar labels where they are than move them out into a legend.

> 7. Why is this unexpected? All elements of the Annotation-Dock change with the current selected Annotation.

Because a common usage scenario is to make a pass through the data looking specifically for one type of artifact (e.g., blinks) and if someone (accidentally or on purpose) modifies an existing annotation of a different type, then the next blink they try to annotate will get the wrong type.

> 9. What OS are you on? I can't reproduce on Windows... But that sounds annoying, maybe you could open an issue with a screencast?

Linux.
https://user-images.githubusercontent.com/1810515/136983483-5a6fde80-c8a0-4b45-b818-f2105d2450a2.mp4

> 12. This mode is reproducing the vertical scrollbar from the mpl-backend showing the boundaries between channel-types. What other information could we show there?

Not sure...  Right now the time dimension is unused so it would be good to think about how to make it more informative.  I guess I was expecting a really zoomed out version of the channel traces (sort of like the "minimap" that code editors often have on the right edge of the editor pane).  

> 17. I couldn't reproduce that. Now that Shift is gone, is this still happening?

Nope, this is no longer happening 👍🏻

drammock:
also, my ipython terminal shows "unrecognized OpenGL version" (twice) every time I run raw.plot() now.  FWIW my OpenGL version is:
`OpenGL version string: 4.6.0 NVIDIA 460.91.03`

marsipu:
> I just noticed that the browser fails catastrophically if `block=False`.  Is that expected?

Yes, in l.136 of utils is a comment stating, that block=False will only work if a Qt-event-loop is started elsewhere in the code (useful, if you'd like to integrate the browser into your own pyqt-application. But you're right, this should of course be explained to the user in the docs.

marsipu:
> also, my ipython terminal shows "unrecognized OpenGL version" (twice) every time I run raw.plot() now.  FWIW my OpenGL version is:
> `OpenGL version string: 4.6.0 NVIDIA 460.91.03`

Frankly I don't really know what pyopengl does and which version it uses. Maybe that would be a good issue to open to also see if others got the same message and on which platform. (I don't get it on Windows 10)

marsipu:
@mmagnuski Thank you for the feedback!
> when I enter zen mode the events disappear. I have to press e twice to bring them back.

I see, this deviates from the behaviour of the mpl-backend and since you also can toggle events and annotations (and the overviewbar) separately they don't need to be toggled with zen-mode.

**Color-Discussion**: I added items for this to the next [ToDo-project](https://github.com/mne-tools/mne-qt-browser/projects/2) if it's okay to do them in future PRs

> resizing the window by the height can lead to the event numbers to partially hide below the time axis:
I moved the label a bit up. Now this only happens with more extreme window-sizes.

> closing and reopening the annotation tools (pressing a) resets the active annotation. But I guess some users would expect this to happen :)

I see no reason why the selected region shouldn't be saved. I changed that, thanks!

> when the annotations are open, once an annotation is drawn, I can't change the scaling of the signal with - or +. Clicking somewhere outside the signal pannel (fro example on one of the channel names) "unblocks" - and + keys.

Fixed!

> @drammock I actually like the small step with ctrl + arrow to be 0.1 * window_duration :)

It should be for time (40% without, 10% with CTRL)

marsipu:
@larsoner I am on board for the 0.24-release. Tomorrow I got some time to look into the issues.

agramfort:
> > The new code from this PR is mainly just refactoring or adaptions for pyqtgraph in the tests so I wouldn't expect that it degrades coverage of mne-python.
> 
> It sounds like this is what @agramfort was concerned with, and at least codecov seems to agree with you "94.83% of diff hit (target 95.00%)". So I think we're good!

all good then !

mne/viz/tests/test_raw.py
```diff
@@ -733,29 +835,120 @@ def test_plot_sensors(raw):
...
+    from time import perf_counter
+
+    from PyQt5.QtCore import QTimer
+    from PyQt5.QtWidgets import QApplication
+
+    hscroll_diffs = list()
+    vscroll_diffs = list()
+
+    def _ininite_hscroll(pg_fig):
+        global bm_count
```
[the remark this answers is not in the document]
marsipu: Sure, i put it into mne_qt_browser

marsipu:
> Can we add your module to environment.yml (this means you need a conda-forge package, @hoechenberger and/or I could help with this) and requirements.txt so that users have it available?

For `requirements.txt` should I also upload mne_qt_browser on PyPi?
Yes, some help/tips for creating the conda-forge-PR would be great, thank you.

larsoner:
Yes, but more consistent to name the package mne-qt-browser for PyPi (like mne-connectivity, mne-kit-gui, etc.). Getting a package on PyPi first makes the conda forge process easy, so feel free to PyPi whenever you have time and I'll set up the conda forge stuff

mne/viz/_figure.py
```diff
@@ -341,11 +580,40 @@ def _click_ch_name(self, ch_index, button):
...
+                subprocess.run(
+                    [sys.executable, "-m", "pip", "install",
+                     repo_link],
+                    check=True)
+            except subprocess.CalledProcessError:
+                print('Installation of pyqtgraph-backend failed!\n'
+                      f'You can try to install the pyqtgraph-backend '
+                      f'manually with:\n'
+                      f'"pip install {repo_link}"')
+                return _load_backend('matplotlib')
```
marsipu, line 613: @larsoner This automatic installation of mne-qt-browser won't be necessary anymore when it is put into `requirements.txt`, right?

marsipu:
> Yes, but more consistent to name the package mne-qt-browser for PyPi (like mne-connectivity, mne-kit-gui, etc.). Getting a package on PyPi first makes the conda forge process easy, so feel free to PyPi whenever you have time and I'll set up the conda forge stuff

https://test.pypi.org/project/mne-qt-browser/0.1.0/
@larsoner Okay like this?
_Some parts of the README will be edited when this PR is merged._

mne/viz/_figure.py
agramfort: we need to discuss how to proceed here.

mne_qt_browser depends on mne and should take care of its release cycle and testing.

to me it's like mne-connectivity. We don't install or test mne-connectivity within mne-python repo.

anyway these lines of automatic install is really not recommended. Please remove them as otherwise we cannot merge and we would like to release in the coming days. This PR is to me the last big thing before we release.

thx @marsipu !

larsoner: Agreed no `pip install` should be done automatically

I think I agree with @agramfort that testing should be moved to the `mne-qt-browser` repo. @marsipu you can add a scheduled Azure run like we did for mne-kit-gui like [this](https://github.com/mne-tools/mne-kit-gui/blob/81fa55b64735ad0c6f35f9e4f4bfa5f78cf22589/azure-pipelines.yml#L13-L20) that runs every night, so you'll see quickly if things fail. I can also add an email notification via the Azure interface for you for the scheduled run -- I just did it for mne-kit-gui and it already pinged me over the weekend (somehow we managed to break things quickly!).

larsoner:
That looks correct except you need to upload to the real `pypi` and not `test.pypi.org`

marsipu:
> 
> 
> That looks correct except you need to upload to the real `pypi` and not `test.pypi.org`

I know, I wanted you to check first because I learned that it is hard to correct mistakes once the package is uploaded to PyPi.

mne/viz/_figure.py
marsipu: @agramfort @larsoner Do you mean with moving testing to the `mne-qt-browser`-repo, that the changes from this PR in `test_raw.py` should be reverted and all the adapted tests should be copied to the `mne-qt-browser`?

larsoner: > Do you mean with moving testing to the mne-qt-browser-repo, that the changes from this PR in test_raw.py should be reverted and all the adapted tests should be copied to the mne-qt-browser?

Hmm, that does seem a bit tough. I guess for maintenance purposes, if we change something about how a MNEFigure class works for example, or if we want to test some new interaction mode, we would want to change mne-qt-browser at the same time. Not sure the best path here.

Thinking about it more I think I'm okay with merging this with the testing infrastructure as is. We can always split it off later / during the 1.0 cycle if we think of a better solution.

Thoughts @drammock @agramfort ?

marsipu: > Hmm, that does seem a bit tough. I guess for maintenance purposes, if we change something about how a MNEFigure class works for example, or if we want to test some new interaction mode, we would want to change mne-qt-browser at the same time.

I am glad, that's what I am thinking too, at least for the moment.

> you can add a scheduled Azure run like we did for mne-kit-gui like this that runs every night, so you'll see quickly if things fail

Is this setting up an Azure-pipeline something which has to be done before the release?

larsoner:
> I know, I wanted you to check first because I learned that it is hard to correct mistakes once the package is uploaded to PyPi.

Nah, you just cut a 0.1.1 then 0.1.2 etc. ... :)

But more seriously:

1. I did:
    ```
    $ pip install -i https://test.pypi.org/simple/ mne-qt-browser==0.1.0
    $ MNE_BROWSE_BACKEND=pyqtgraph mne browse_raw ~/mne_data/MNE-sample- 
    data/MEG/sample/sample_audvis_raw.fif
    ```
    And there was a problem in that the browser shows up then immediately disappears. This should enter some sort of blocking mode, so I pushed a commit to this PR to do that in the `mne/commands/browse_raw.py` code, and tested that it fixes it for both backends.

2. Next there is this issue, doing `pip uninstall mne-qt-browser` I see:
    ```
    Uninstalling mne-qt-browser-0.1.0:
      Would remove:
        /home/larsoner/.local/lib/python3.9/site-packages/mne_qt_browser-0.1.0.dist-info/*
        /home/larsoner/.local/lib/python3.9/site-packages/mne_qt_browser/*
        /home/larsoner/.local/lib/python3.9/site-packages/tests/*
      Would not remove (might be manually added):
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bibtex_parser_test.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bst_parser_test/__init__.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bst_parser_test/apacite.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bst_parser_test/bst_parser_test.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bst_parser_test/jurabib.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/bst_parser_test/plain.py
        /home/larsoner/.local/lib/python3.9/site-packages/tests/data/IEEEtran.aux
        /home/larsoner/.local/lib/python3.9/site-packages/tests/data/IEEEtran.bib
        ...
    ```
    So it looks like your package adds a `tests` (rather than `mne_qt_browser/tests`), which is not good -- and you can see some other packages have also put junk in this location, but we shouldn't make the problem worse :)

3. Finally, I did `pip uninstall pyqtgraph pyopengl mne-qt-browser`, and then a subsequent call to `pip install mne-qt-browser` didn't install `pyqtgraph` or `pyopengl` as dependencies -- it seems like it should do these two things. If we can't do it because we need a new release of PyQtGraph, then we can add these as `install_requires` to `mne-qt-browser` later, but we should do it at some point.

mne/viz/_figure.py
larsoner: > Is this setting up an Azure-pipeline something which has to be done before the release?

No, you can set it up on the `mne-qt-browser` repo anytime. I just created one for `mne-kit-gui` that was pretty simple and runs for all OSes, you might want to base things off of that if you want to take a stab at it.

marsipu:
> Nah, you just cut a 0.1.1 then 0.1.2 etc. ... :)

Regarding the version, how do I set the `download_url` in `setup.py` with the GitHub-release-url if I have to set the tag there first? It seems like a paradox to me, because I can push the change with the `download_url ` only after I got the link from GitHub, but then this change in setup.py will be commited after the version-tag.

larsoner:
For mne-python we just set `DOWNLOAD_URL = 'http://github.com/mne-tools/mne-python'`

marsipu:
1. Thanks, I haven't thought of the command-line
2. Hm, maybe `packages=find_packages()` is wrong there, I just try it with just `packages=['mne_qt_browser']
3. That is peculiar, `pyqtgraph` and `pyopengl` should be already in `install_requires`.

larsoner:
... but if you want to be precise, I would just figure out what the URL will be ahead of time once I do the GitHub release (e.g., `https://github.com/mne-tools/mne-python/archive/refs/tags/<whatever_tag>.tar.gz`), set it, then do a GitHub release where it creates the tag `<whatever_tag>`. For MNE-Python that would end up as something like https://github.com/mne-tools/mne-python/archive/refs/tags/v0.23.4.tar.gz

larsoner:
Regarding the failure, on further inspection I think it's just because it's using test-pypi:
```
$ pip install -i https://test.pypi.org/simple/ mne-qt-browser==0.1.0
Defaulting to user installation because normal site-packages is not writeable
Looking in indexes: https://test.pypi.org/simple/
Collecting mne-qt-browser==0.1.0
  Downloading https://test-files.pythonhosted.org/packages/80/a9/0f9e1d5efe468b0a2135526dedfeb588b00cae0912ab769e6fe6269c3655/mne_qt_browser-0.1.0-py3-none-any.whl (37 kB)
Requirement already satisfied: scipy in /home/larsoner/python/scipy (from mne-qt-browser==0.1.0) (1.8.0.dev0+1889.135b734)
Requirement already satisfied: mne in /home/larsoner/python/mne-python (from mne-qt-browser==0.1.0) (0.24.dev0)
Requirement already satisfied: numpy in /home/larsoner/python/numpy (from mne-qt-browser==0.1.0) (1.22.0.dev0+1468.g7f61ddbf0)
Requirement already satisfied: qtpy in /home/larsoner/.local/lib/python3.9/site-packages (from mne-qt-browser==0.1.0) (1.9.0)
ERROR: Could not find a version that satisfies the requirement pyqtgraph (from mne-qt-browser) (from versions: none)
ERROR: No matching distribution found for pyqtgraph
```

mne/viz/_figure.py
drammock: I thought that having this in a separate repo was temporary until all the kinks are worked out; if that's the case then we'll need the tests here eventually anyway.  Either way I am OK leaving the tests where they are for now and merging, but let's discuss at next dev meeting the best way to handle testing here.

agramfort: yes keep the "common test" in mne-python and make the mne-qt-browser CI run the test functions in mne-python. I am thinking something like what we use to test a new Raw object. We have a function and if this function passes for a backend then we are good.

makes sense?

marsipu:
mne-qt-browser is now uploaded on [PyPi](https://pypi.org/project/mne-qt-browser/0.1.1/)

marsipu:
As @drammock already mentioned once:
Raw.plot() won't work out of the box for the new backend because `block` needs to be `True`. The rare case that you'd like to integrate the plot inside another Qt-Application where you need to set `block=False` to make it work probably applies to a small amount of users. That might startle many users if they want to try the new backend.
I am not sure what the best solution for this would be (discard the option of integration into another Qt-App by always setting `block=True` when pyqtgraph is selected? Or setting `block=True` by default?
WDYT?

larsoner:
Okay I have the conda-forge recipe ready to go:

https://github.com/conda-forge/staged-recipes/pull/16711

1. I had to remove `PyOpenGL` as it was creating problems in building. We can add it back in, but it will create complexity. I think we should just tell `conda` folks that they might get better performance using `conda install pyopengl` as well. This also reminds me we should probably keep it as an optional pip dependency.
2. We have to wait for 0.24 to release, because the recipe can't build until this PR is merged, 0.24 lands, and the conda-forge recipe for `mne-base` updates.

>  because block needs to be True

Can you remind me why this is the case? In general you can have Qt applications open in non-blocking mode in Python (matlptolib Qt5Agg, mayavi, pyvistaqt.BackgroundPlotter), and it would be nice if the new browser supported it, too.

marsipu:
> Can you remind me why this is the case? In general you can have Qt applications open in non-blocking mode in Python (matlptolib Qt5Agg, mayavi, pyvistaqt.BackgroundPlotter), and it would be nice if the new browser supported it, too.

I already tried to find a solution, but I couldn't find something that  it depends on how you start the plot:
## ipython (`raw.plot()` as single command)
If you use `%gui qt5` it works with `block=False` (otherwise the window is unresponsive)

## python (`raw.plot()` as single command)
It works with `block=False`.

## python (run a python script)
Only works with `block=True` (otherwise the window is unresponsive)

marsipu:
> Okay I have the conda-forge recipe ready to go
Thanks a lot @larsoner

> 1. I had to remove PyOpenGL as it was creating problems in building. We can add it back in, but it will create complexity. I think we should just tell conda folks that they might get better performance using conda install pyopengl as well. This also reminds me we should probably keep it as an optional pip dependency.

Do you mean putting pyopengl into the pip-section of environment.yml?

> 2. We have to wait for 0.24 to release, because the recipe can't build until this PR is merged, 0.24 lands, and the conda-forge recipe for mne-base updates.

Ok, so we add mne-qt-browser to environment.yml in a future PR.

agramfort:
@marsipu can you look at CI failures? otherwise it's good to go from your end API-wise?

marsipu:
> can you look at CI failures?
I think some of the failing CIs came from mne-qt-browser missing in some CI-dependencies. For the conda-CIs, I added mne-qt-browser now to the pip-section of environment.yml until the [PR at conda-forge](https://github.com/conda-forge/staged-recipes/pull/16711) is merged (which depends on this PR to be merged). Or would you suggest another solution?

But one failure from `linux/pip-pre/py3.9` wonders me and I can't explain it:
```
Run ./tools/github_actions_infos.sh
Usage: mne sys_info

mne sys_info: error: no such option: -p
Error: Process completed with exit code 2.
```
Do you have an idea what the problem is here (I haven't changed this script and it runs fine in my terminal)

> otherwise it's good to go from your end API-wise?
Yes, I specified the changes a bit more in `latest.inc`. The new thing for users will be that they now can specify the backend for `Raw.plot()`, e.g. with:
```
import mne
mne.viz.set_browser_backend('pyqtgraph')
```
In the Readme of mne-qt-browser is also a [small tutorial](https://github.com/mne-tools/mne-qt-browser).
The features of the pyqtgraph-backend are similar to matplotlib, deviations are explained in the docs.
Do you think this is sufficient for the users to get used to the new backend?

marsipu:
> But one failure from linux/pip-pre/py3.9 wonders me and I can't explain it:

Might it be a problem that `mne` is a dependency of `mne-qt-browser` and so the stable mne-version is installed before the PR-dev-version of `mne` is installed in the CIs? Or does that step overwrite the stable version with the PR-dev-version?

mne/defaults.py
```diff
@@ -7,9 +7,9 @@
 from copy import deepcopy
 
 DEFAULTS = dict(
-    color=dict(mag='darkblue', grad='b', eeg='k', eog='k', ecg='m', emg='k',
-               ref_meg='steelblue', misc='k', stim='k', resp='k', chpi='k',
-               exci='k', ias='k', syst='k', seeg='saddlebrown', dbs='seagreen',
+    color=dict(mag='#00008B', grad='b', eeg='k', eog='k', ecg='m', emg='k',
+               ref_meg='#4682B4', misc='k', stim='k', resp='k', chpi='k',
+               exci='k', ias='k', syst='k', seeg='#8B4513', dbs='#2E8B57',
```
larsoner, line 12: Let's revert these. You can make your code understand these by making use of 
```
mne/utils/check.py:def _to_rgb(*args, name='color', alpha=False):
```
which wraps matplotlib's color-name-to-rgb(a) functionality

mne/io/base.py
```diff
@@ -1514,22 +1514,23 @@ def _tmin_tmax_to_start_stop(self, tmin, tmax):
 
     @copy_function_doc_to_method_doc(plot_raw)
     def plot(self, events=None, duration=10.0, start=0.0, n_channels=20,
-             bgcolor='w', color=None, bad_color=(0.8, 0.8, 0.8),
+             bgcolor='w', color=None, bad_color='#CCCCCC',
```
larsoner, line 1517: same

mne/utils/docs.py
```diff
@@ -1515,6 +1515,32 @@
...
+    (Only pyqtgraph) Set to ``True``/``False`` to enable/disable loading
+    all data into RAM and performing most preprocessing for visualization
+    once to increase performance.
+    This happens in a separate thread and the effect applies
+    after precomputing is done.
+    With 'auto' the currently available RAM space will be compared to the
+    expected size of the precomputed data in RAM.
+    Default is ``'auto'``.
+
+    .. versionadded:: 1.0
```
larsoner, line 1530: 0.24

mne/utils/docs.py
```diff
@@ -1515,6 +1515,32 @@
...
+
+docdict['use_opengl'] = """
+use_opengl : bool
+    (Only pyqtgraph) Enables OpenGL with usually enhanced performance
+    (requires pyopengl to be installed).
+    It may depend on your system (CPU, Graphics-Card) if and how much
+    OpenGL enhances the performance.
+    Default is ``True``.
+
+    .. versionadded:: 1.0
```
larsoner, line 1541: 0.24

mne/viz/_figure.py
```diff
@@ -411,21 +658,21 @@ def set_browser_backend(backend_name, verbose=None):
...
        +--------------------------------------+------------+-----------+
-       | OpenGL Acceleration                  |            | ✓         |
+       | Selection Mode                       | ✓          | ✓         |
        +--------------------------------------+------------+-----------+
-       | Toolbar                              |            | ✓         |
+       | Smooth Scrolling                     |            | ✓         |
        +--------------------------------------+------------+-----------+
-       | Dark Mode                            |            |           |
+       | Overview-Bar (with Z-Score-Mode)     |            | ✓         |
        +--------------------------------------+------------+-----------+
```
larsoner, line 676:
```suggestion
       +--------------------------------------+------------+-----------+

       .. versionadded:: 0.24
```

mne/viz/ica.py
```diff
@@ -1042,7 +1042,7 @@ def _plot_sources(ica, inst, picks, exclude, start, stop, show, title, block,
         raise RuntimeError('Stop must be larger than start.')
 
     # misc
-    bad_color = (0.8, 0.8, 0.8)
+    bad_color = '#CCCCCC'
```
larsoner, line 1045: same

mne/viz/raw.py
```diff
@@ -20,21 +20,23 @@
...
+                    _get_channel_plotting_order, _make_event_color_dict,
+                    _show_browser)
 
 _RAW_CLIP_DEF = 1.5
 
 
 @verbose
 def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
-             bgcolor='w', color=None, bad_color=(0.8, 0.8, 0.8),
+             bgcolor='w', color=None, bad_color='#CCCCCC',
```
larsoner, line 31: same

mne/viz/raw.py
```diff
@@ -59,8 +61,8 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
     color : dict | color object | None
         Color for the data traces. If None, defaults to::
 
-            dict(mag='darkblue', grad='b', eeg='k', eog='k', ecg='m',
-                 emg='k', ref_meg='steelblue', misc='k', stim='k',
+            dict(mag='#00008B', grad='b', eeg='k', eog='k', ecg='m',
+                 emg='k', ref_meg='#4682B4', misc='k', stim='k',
```
larsoner, line 65: same

tools/circleci_dependencies.sh
```diff
@@ -23,6 +23,7 @@ if [[ "$CIRCLE_JOB" == "interactive_test" ]]; then
 	python -m pip install --progress-bar off vtk-9.0.20201117-cp39-cp39-linux_x86_64.whl
 	python -m pip install --progress-bar off https://github.com/pyvista/pyvista/zipball/main
 	python -m pip install --progress-bar off https://github.com/pyvista/pyvistaqt/zipball/main
+	python -m pip install --progress-bar off --upgrade --pre mne-qt-browser
```
larsoner, line 26: We don't need this for CircleCI. Eventually we should add it but let's do it later: #9942

larsoner (commented):
I'll push a commit to try to fix the pip-pre issue. Feel free to address my other tiny comments here, otherwise LGTM +1 for merge!

mne/defaults.py
larsoner: ... we can add a `output='float' | 'hex'` to `_to_rgba` if that would help. I can push but it should be pretty simple like:
```
try:
    out =  func(*args)
except ...:
    ...
if output == 'hex':
    out = np.clip([255 * o for o in out], 0, 255).round().astype(int)
    out = '#' + ''.join(f'{o:02X}' for o in out)
return out
```

marsipu: Hm. pyqtgraph seems to accept the matplotlib-color-names out of the box in the newest release. Only (0.8, 0.8, 0.8) is a problem because it expects RGB-values as integers from 0-255 and not float from 0-1.

_to_rgb would have to be added in +40 places in `mne-qt-browser`. Do you think that might justify changing (0.8, 0.8, 0.8) to (204, 204, 204)?

larsoner: > EDIT: I see, for matplotlib only the floats from 0-1 are accepted. What about changing to lightgray?

I think in MNE we should not have to change how we deal with / think about colors -- `matplotlib color` mindset has been around for a long time and is in a lot of our code / docstrings etc.

> _to_rgb would have to be added in +40 places in mne-qt-browser

You've only changed 5-10 (?) instances here, so I think you only really need to convert the problematic ones that come from MNE.

marsipu: > I think in MNE we should not have to change how we deal with / think about colors
I understand that. But since pyqtgraph seems to support matplotlib color-names in general, the only change now in MNE-Python would have to be changing the default bad-color from (0.8, 0.8, 0.8) to `lightgray`. Or is there a reason why the default-bad-color has to be a RGB-tuple?
(we could of course just change the handling of the bad-color in mne-qt-browser to convert the float-RGB to int-RGB, but that seems a bit inconsistent to me, WDYT?)

>You've only changed 5-10 (?) instances here, so I think you only really need to convert the problematic ones that come from MNE.

The colors from MNE go directly into methods to create the traces, ch-axis-names, overview-bar-representation, annotation-colors. Thats where all the changes would come from in `mne-qt-browser`.

larsoner: > (we could of course just change the handling of the bad-color in mne-qt-browser to convert the float-RGB to int-RGB, but that seems a bit inconsistent to me, WDYT?)

To me the inconsistency comes from pyqtgraph not handling tuple-of-float automatically, as this is an accepted matpltolib color. So they support some matplotlib color types but not all.

Does `'0.8'` work? It should since it's an understood matplotlib color. If so let's just go with that. If it doesn't, I can live with the `'lightgray'` change

marsipu: Yes, `0.8` does work too

larsoner:
@marsipu talking to @agramfort we want to optimally split CI burden between MNE-Python and MNE-Qt-Browser. One way to do this is to make it so that only one CI run (probably ultraslow?) runs the mne-qt-browser tests. This makes sure we don't break any interfaces at the MNE-Python end (e.g., MNEBrowseFigure) that break the qt backend. Then in the `mne-qt-browser` repo, you have CIs configured to test fully different operating systems, versions of dependencies, conda vs pip, etc. Does that sound reasonable to you?

If so, I'll try modifying the CIs here to go ahead with that plan. If you can push the color changes/minor suggestions based on my review, we might be able to merge today!

marsipu:
> @marsipu talking to @agramfort we want to optimally split CI burden between MNE-Python and MNE-Qt-Browser. One way to do this is to make it so that only one CI run (probably ultraslow?) runs the mne-qt-browser tests. This makes sure we don't break any interfaces at the MNE-Python end (e.g., MNEBrowseFigure) that break the qt backend. Then in the mne-qt-browser repo, you have CIs configured to test fully different operating systems, versions of dependencies, conda vs pip, etc. Does that sound reasonable to you?

Yes, that sounds reasonable to me.

> If you can push the color changes/minor suggestions based on my review, we might be able to merge today!

I pushed the changes requested by your review.

larsoner:
@marsipu I went to add `mne_qt_browser` to our `mne sys_info` but I get:
```
AttributeError: module 'mne_qt_browser' has no attribute '__version__'
```
Can you fix `mne-qt-browser` to have this attribute (it's pretty standard) and push to PyPi?

larsoner:
@marsipu feel free to merge https://github.com/mne-tools/mne-qt-browser/pull/11 if you're happy with it and push to PyPi, then I'll update `mne sys_info` here. Then I think we're good to go once @drammock has a last look

doc/cited.rst
```diff
@@ -3,7 +3,7 @@
 Papers citing MNE-Python
 ========================
 
-Estimates provided by Google Scholar as of 27 January 2021:
+Estimates provided by Google Scholar as of 02 November 2021:
 
-- `MNE (908) <https://scholar.google.com/scholar?cites=12188330066413208874&as_ylo=2014>`_
-- `MNE-Python (771) <https://scholar.google.com/scholar?cites=1521584321377182930&as_ylo=2013>`_
+- `MNE (1100) <https://scholar.google.com/scholar?cites=12188330066413208874&as_ylo=2014>`_
+- `MNE-Python (1060) <https://scholar.google.com/scholar?cites=1521584321377182930&as_ylo=2013>`_
```
larsoner, line 9: I snuck this in because we need it for release anyway

drammock (commented):
Left a few docstring improvement comments.  More importantly:  **keyboard shortcuts are not consistent between the two backends**.  pageup/pagedown do opposite things in pyqtgraph backend as in matplotlib backend.  Make them consistent.  I didn't thoroughly check all the keyboard shortcuts so I'm hoping this is an isolated oversight and there aren't other cases; @marsipu please check them all again.

larsoner:
@marsipu feel free to address these tomorrow. If we need to push the release until then or later in the week it's okay, let's get it right!

marsipu:
>Left a few docstring improvement comments. More importantly: keyboard shortcuts are not consistent between the two backends. pageup/pagedown do opposite things in pyqtgraph backend as in matplotlib backend. Make them consistent. I didn't thoroughly check all the keyboard shortcuts so I'm hoping this is an isolated oversight and there aren't other cases;

@drammock I checked all the keyboard-shortcuts and there shouldn't be any differences left now. PageUp/PageDown were confused.

drammock:
looks like the CI failures are related / relevant:
```
=================================== FAILURES ===================================
_____________________ test_min_window_size[pyqtgraph-None] _____________________
mne/viz/tests/test_raw.py:844: in test_min_window_size
    assert_allclose(fig._get_size(), (8, 8), atol=atol)
E   AssertionError: 
E   
E   Mismatched elements: 2 / 2 (100%)
E   Max absolute difference: 8.33333333
E   Max relative difference: 1.04166667
E    x: array([16.333333, 11.930556])
E    y: array([8, 8])
___________________ test_min_window_size[pyqtgraph-0.1,0.1] ____________________
mne/viz/tests/test_raw.py:844: in test_min_window_size
    assert_allclose(fig._get_size(), (8, 8), atol=atol)
E   AssertionError: 
E   
E   Mismatched elements: 2 / 2 (100%)
E   Max absolute difference: 8.33333333
E   Max relative difference: 1.04166667
E    x: array([16.333333, 11.930556])
E    y: array([8, 8])
=============================== warnings summary ===============================
```

larsoner:
Green! In it goes. Thanks for working through this @marsipu , this is amazing stuff!

rob-luke:
🎉 🎉 🎉 🎉 🎉

drammock:
👏👏👏👏👏👏👏👏👏

hoechenberger:
🥳 ❤️

sappelhoff:
awesome work! :rocket:

agramfort:
@marsipu what you have achieved is big ! Honestly I had almost no hope you could pull this off ! 👏 👏 👏

larsoner:
Hopefully just the first of many times you get to prove @agramfort wrong :)

marsipu:
Thank you all for your support and feedback on the way. I am really happy that the new backend can now be tested/used by the greater community.

> Honestly I had almost no hope you could pull this off !

Me too 😄 (I am currently preparing for my biggest exam in medicine next spring)

> Hopefully just the first of many times you get to prove @agramfort wrong :)

That's the plan (at least in a positive sense)

agramfort:
>
> Honestly I had almost no hope you could pull this off !
>
> Me too 😄 (I am currently preparing for my biggest exam in medicine next
> spring)
>

a medical doctor who can code for real ! you'll never stop bluffing me :)

> Hopefully just the first of many times you get to prove @agramfort
> <https://github.com/agramfort> wrong :)
>
> That's the plan (at least in a positive sense)
>
I am a very positive person don't hesitate ;)

PR #9688, opened 2021-08-23
[ENH, MRG] Add remove methods to `mne.viz.Brain`

### LINKED ISSUE #9594: [ENH] Add `mne.viz.Brain.remove` method

alexrockhill:
Fixes #9594.

### DIFF
--- a/mne/viz/_brain/_brain.py
+++ b/mne/viz/_brain/_brain.py
@@ -1802,13 +1802,14 @@ class Brain(object):
                             )
         return colormap_map[cortex]
 
-    def _remove(self, item):
+    def _remove(self, item, render=False):
         """Remove actors from the rendered scene."""
         if item in self._actors:
             for actor in self._actors[item]:
                 self._renderer.plotter.remove_actor(actor)
             self._actors.pop(item)  # remove actor list
-            self._renderer._update()
+            if render:
+                self._renderer._update()
 
     def _add_actor(self, item, actor):
         """Add an actor to the internal register."""
@@ -2067,7 +2068,7 @@ class Brain(object):
 
     def remove_data(self):
         """Remove rendered data from the mesh."""
-        self._remove('data')
+        self._remove('data', render=True)
 
     def _iter_views(self, hemi):
         """Iterate over rows and columns that need to be added to."""
@@ -2399,7 +2400,7 @@ class Brain(object):
 
     def remove_head(self):
         """Remove head objects from the rendered scene."""
-        self._remove('head')
+        self._remove('head', render=True)
 
     @fill_doc
     def add_skull(self, outer=True, color='gray', alpha=0.5):
@@ -2437,7 +2438,7 @@ class Brain(object):
 
     def remove_skull(self):
         """Remove skull objects from the rendered scene."""
-        self._remove('skull')
+        self._remove('skull', render=True)
 
     @fill_doc
     def add_volume_labels(self, aseg='aparc+aseg', labels=None, colors=None,
@@ -2526,7 +2527,7 @@ class Brain(object):
 
     def remove_volume_labels(self):
         """Remove the volume labels from the rendered scene."""
-        self._remove('volume_labels')
+        self._remove('volume_labels', render=True)
         self._renderer.plotter.remove_legend()
 
     def add_foci(self, coords, coords_as_verts=False, map_surface=None,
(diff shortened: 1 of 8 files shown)


mne/viz/_brain/_brain.py
```diff
@@ -360,12 +364,24 @@ class Brain(object):
...
        | remove_annotations        | -            | ✓             |
        +---------------------------+--------------+---------------+
+       | remove_sensors            |              | ✓             |
+       +---------------------------+--------------+---------------+
+       | remove_skull              |              | ✓             |
+       +---------------------------+--------------+---------------+
+       | remove_text               |              | ✓             |
+       +---------------------------+--------------+---------------+
+       | remove_volume_labels      |              | ✓             |
+       +---------------------------+--------------+---------------+
```
larsoner, line 384: Rather than a bunch of `remove_*` methods, an alternative would be to make all `add_*` functions return actors, and have a single `remove_actors` method that took an actor or list of actors. I think I like that design slightly better because it also allows for configuring actor VTK properties if people want to do so.

mne/viz/_brain/_brain.py
```diff
@@ -1785,6 +1802,21 @@ def _cortex_colormap(self, cortex):
                             )
         return colormap_map[cortex]
 
+    def _remove(self, item):
+        """Remove actors from the rendered scene."""
+        if item in self._actors:
+            for actor in self._actors[item]:
+                self._renderer.plotter.remove_actor(actor)
```
larsoner, line 1809: there can be multiple views/scenes, does this work when that's the case?

alexrockhill: Yes, it's a list of actors for each type

larsoner: I'm more concerned that the `.plotter` is never changed to any other subplot, I'm a bit surprised it works. But as long as you've at least tested by eye / manually that everything works in this case, I'm happy

mne/viz/_brain/_brain.py
alexrockhill: We could make `actors` a public property. One difficulty is that then for sensors you'd have to return a dictionary with all the types if you want to support removing by type (e.g. `fnirs`), that could work also, but it puts a bit more work on the user

larsoner: I think I prefer +0.5 a single `.remove` method if it would work. @GuillaumeFavelier @agramfort any opinion on having `remove_head(), remove_volume_labels(), ...` vs `remove_actors(actors)`?

larsoner: ... if we really want, we could add convenience aliases like `remove_actors('head')` or so, but I don't think we'd need it

alexrockhill: The only thing is that there are already `remove_annotation` and `remove_labels` so as it is would be consistent stylistically...

I'll push another commit the other way and then we can revert if the individual methods are better

larsoner: Ahh you're right. I think it makes sense then to leave `remove_*` rather than a generic `remove_actors`

mne/viz/_brain/_brain.py
alexrockhill: This works, can confirm

mne/viz/_brain/_brain.py
alexrockhill: Also in `add_text` someone spelled out the way to do it with adding a name etc. that would be an internal reference to the actor which would suggest doing it the way it is now, otherwise the name of the text is extraneous, that's something the user should keep track of

alexrockhill:
It also looks like their is an issue on my setup with `_pyvista.text_2d` and font family = `arial`, it doesn't render and it's just a silent error...

alexrockhill:
Looks like a new `DeprecationWarning` from the latest `Dipy` release but other than that everything is good. How do you fix that? I thought those were filtered by default?

mne/viz/_brain/_brain.py
GuillaumeFavelier: And in the case of `add_annotation()` and `add_label()` they do not rely on `add_actor()` but `add_overlay()` instead.

GuillaumeFavelier: They follow the `_LayeredMesh` logic instead of the usual `vtkActor` to avoid geometry duplication and opacity issues since VTK9.

larsoner:
> How do you fix that? I thought those were filtered by default?

We don't filter because we want to adapt our code to deprecations as soon as possible. Usually we triage to use the improved code path if possible. I can open a separate PR for this shortly.

larsoner:
In this case I think it's actually a bug at their end (see https://github.com/dipy/dipy/pull/2435) so I'll just pin the `--pre` CIs not to use the problematic version for now

larsoner:
Thanks @alexrockhill !

PR #9692, opened 2021-08-24
WIP: Add CODEOWNERS?

larsoner:
@mne-tools/mne-python-steering-committee members: have you disabled repo-wide notifications so you only get notified when pinged, or thought about doing so because you get too many MNE-Python notifications?

There is a potential solution: add yourself to `.github/CODEOWNERS`, giving you opt-in support for file- and directory-level control over which PRs you get pinged to review. For the last four months [in SciPy](https://github.com/scipy/scipy/blob/master/.github/CODEOWNERS) I have found this to be very effective to get emails about PRs I can help with and not get emails about PRs I can't (or don't want to).

I have opened this WIP PR just to show one hypothetical example -- let's say @rob-luke decides he's getting too many emails and only wants to hear about fNIRS-related stuff. He could add his name for fNIRS-related files as I've done here. If he then unsubscribes from MNE-Python as a whole (while keeping @-mentions enabled), GitHub will automatically ping him when people change fNIRS-specific code.

The only potential downside I see is that @rob-luke might miss an *issue* that gets opened about fNIRS, but the folks who triage most issues (like @hoechenberger, @drammock, @agramfort, @cbrnr, etc.) typically @-ping the appropriate people in these cases already anyway, so continuing this practice should mitigate this issue. And having names enumerated in `CODEOWNERS` makes it clear who wants to be pinged about what issues, too.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -14,16 +14,9 @@
 
 # Each line is a file pattern followed by one or more owners.
 
-# NIRS
-/mne/io/nirx/ @rob-luke
-/mne/io/nihon/ @rob-luke
-/mne/io/preprocessing/nirs/ @rob-luke
-
-
 # IO
 /mne/io/brainvision @sappelhoff
 /mne/export @sappelhoff
-
 
 # Beamforming
 /mne/beamformer/ @britta-wstnr


sappelhoff (commented):
I think this is a good GitHub feature and I would use it (i.e., enter my name for some files/folders) if we were to use it for MNE-Python.

larsoner:
... so I meant to say, if anyone does want to use this, feel free to push a commit with `[ci skip]` in the commit message to add your name to some files / directories (see [the instructions](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) for examples), or comment here what domain(s) you're only interested in and I'm happy to add them myself.

.github/CODEOWNERS
```diff
@@ -0,0 +1,22 @@
+# The purpose of this file is to trigger review requests when PRs touch
+# particular files. Those reviews are not mandatory, however it's often useful
+# to have an expert pinged who is interested in only one part of MNE and
+# doesn't follow general development.
```
sappelhoff, line 4: perhaps it'd be worth to mention that - despite the name of this file - having **your** name in this file does not really imply "ownership", but only that you want to be pinged?

agramfort:
I will continue to get all notifications?

larsoner:
Yes `CODEOWNERS` is opt-in and its only effect is to request a review from people who is marked as "owning" a modified file. It does not change anyone's repo "watch" status, so @agramfort your notify-for-everything status will not be affected :)

Someone like @sappelhoff who has added themselves to CODEOWNERS doesn't even technically have to disable "notify for all activity", but then they will get one or two notifications for relevant PRs, one that the PR was opened (because of watching the repo) then another requesting your review if they are also a "code owner" of some changed file. Hence as an email reduction method, adding to CODEOWNERS then setting the repo watch status to "mentions and participating" rather than "notify for all" would be my suggested strategy.

agramfort:
sounds good !

>

britta-wstnr:
Added myself to the beamformer module!

hoechenberger (approved):

PR #9643, opened 2021-08-04
[MRG] Enable export of EDF files

adam2392:
#### Reference issue
Closes: #9566 

#### What does this implement/fix?
Enables exporting of EDF+ ~~and BDF files,~~ which support annotations.

Note that the montage (i.e. info['dig']) is technically lost upon export.

#### Additional information
If you look at the Jupyter notebook, then I'm not getting lossless conversion. I think I am perhaps setting the physical max/min and digital max/min incorrectly.

PR #9694, opened 2021-08-25
[MRG, BUG] Fixed bug with EDF channel types and minor bug where other types of channels were not written and bug where annotations were not written to correct scale

adam2392:
#### Reference issue
Follow up fix from #9643 where not all the relevant "Voltage/uVoltage" channels were added.

In addition, it will allow channel types to be written to EDF files, in accordance with: https://github.com/mne-tools/mne-python/pull/9694#issuecomment-908135503

Closes: #9704 
Closes: #9730

- Adds corresponding unit test that makes sure all voltage channels are covered in exporting EDF
- Allows channel types to be read in from EDF files via their signal labels
- Allows exporting channel types to EDF files

### DIFF
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -58,17 +58,24 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
         stim_index = np.atleast_1d(stim_index.squeeze()).tolist()
         drop_chs.extend([raw.ch_names[idx] for idx in stim_index])
 
-    # add warning if any channel types are not voltage based
-    # note: we can write these other channels, such as 'misc'
-    # but they just won't work because EDF doesn't support
-    # the idea of a 'misc' channel
+    # Add warning if any channel types are not voltage based.
+    # Users are expected to only export data that is voltage based,
+    # such as EEG, ECoG, sEEG, etc.
+    # Non-voltage channels are dropped by the export function.
+    # Note: we can write these other channels, such as 'misc'
+    # but these are simply a "catch all" for unknown or undesired
+    # channels.
     voltage_types = list(units) + ['stim', 'misc']
-    if any([ch not in voltage_types for ch in orig_ch_types]):
+    non_voltage_ch = [ch not in voltage_types for ch in orig_ch_types]
+    if any(non_voltage_ch):
         warn('There are non-voltage channels that are not set as "misc". '
+             f'{non_voltage_ch}'
              'EDF only supports writing Voltage-based data. '
              'These channels will not be written to the EDF file. '
              'If you would still like these channels to be written, '
-             'set their channel type to "misc".')
+             'you can set their channel type to "misc", but they will '
+             'most likely not be written correctly. For example, EDF '
+             'does not support storing MEG channels.')
 
     ch_names = [ch for ch in raw.ch_names if ch not in drop_chs]
     ch_types = np.array(raw.get_channel_types(picks=ch_names))
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -69,6 +69,7 @@ def _export_raw(fname, raw, physical_range):
              'These channels will not be written correctly.')
 
     ch_names = [ch for ch in raw.ch_names if ch not in drop_chs]
+    ch_types = np.array(raw.get_channel_types(picks=ch_names))
     n_channels = len(ch_names)
     n_times = raw.n_times
 
@@ -101,7 +102,6 @@ def _export_raw(fname, raw, physical_range):
         ch_types_phys_max = dict()
         ch_types_phys_min = dict()
 
-        ch_types = np.array(raw.get_channel_types(picks=ch_names))
         for _type in np.unique(ch_types):
             _picks = np.nonzero(ch_types == _type)[0]
             _data = raw.get_data(units=units, picks=_picks)
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -107,8 +107,14 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
         for _type in np.unique(ch_types):
             _picks = np.nonzero(ch_types == _type)[0]
             _data = raw.get_data(units=units, picks=_picks)
+            if _type == 'ecog':
+                print(_data)
             ch_types_phys_max[_type] = _data.max()
             ch_types_phys_min[_type] = _data.min()
+        
+        from pprint import pprint
+        pprint(ch_types_phys_max)
+        pprint(ch_types_phys_min)
     else:
         # get the physical min and max of the data in uV
         # Physical ranges of the data in uV is usually set by the manufacturer
@@ -225,7 +231,10 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
             buf[:len(ch_data)] = ch_data
             err = hdl.writeSamples(buf)
             if err != 0:
-                raise RuntimeError(f"writeSamples() returned error: {err}")
+                print(ch_types[jdx])
+                print(jdx)
+                raise RuntimeError(f"writeSamples() for channel {ch_names[jdx]} "
+                                   f"returned error: {err}")
 
         # there was an incomplete datarecord
         if len(ch_data) != len(buf):
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -38,7 +38,7 @@ def _export_raw(fname, raw, physical_range):
     """
     # scale to save data in EDF
     phys_dims = 'uV'
-    
+
     # get EEG-related data in uV
     units = dict(eeg='uV', ecog='uV', seeg='uV', eog='uV', ecg='uV', emg='uV',
                  bio='uV', dbs='uV')
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -105,7 +105,7 @@ def _export_raw(fname, raw, physical_range):
             # XXX: not sure how to compare to get a robust check
             # - needs to account for maybe normalizing data onto
             # a consistent scale
-            # - then basically do hard-threshold check for 
+            # - then basically do hard-threshold check for
             # super outliers
             max_diffs = pdist(data_max)
             min_diffs = pdist(data_min)
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -68,14 +68,14 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     voltage_types = list(units) + ['stim', 'misc']
     non_voltage_ch = [ch not in voltage_types for ch in orig_ch_types]
     if any(non_voltage_ch):
-        warn('There are non-voltage channels that are not set as "misc". '
-             f'{non_voltage_ch}'
-             'EDF only supports writing Voltage-based data. '
-             'These channels will not be written to the EDF file. '
-             'If you would still like these channels to be written, '
-             'you can set their channel type to "misc", but they will '
-             'most likely not be written correctly. For example, EDF '
-             'does not support storing MEG channels.')
+        warn(f"Non-voltage channels detected: {non_voltage_ch}. MNE-Python's "
+             'EDF exporter only supports voltage-based channels, because the '
+             'EDF format cannot accommodate much of the accompanying data '
+             'necessary for channel types like MEG and fNIRS (channel '
+             'orientations, coordinate frame transforms, etc). You can '
+             'override this restriction by setting those channel types to '
+             '"misc" but no guarantees are made of the fidelity of that '
+             'approach.')
 
     ch_names = [ch for ch in raw.ch_names if ch not in drop_chs]
     ch_types = np.array(raw.get_channel_types(picks=ch_names))
@@ -214,8 +214,9 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     n_blocks = np.ceil(n_times / out_sfreq).astype(int)
 
     # increase the number of annotation signals if necessary
-    n_annotations = len(raw.annotations)
-    if n_annotations is not None:
+    annots = raw.annotations
+    if annots is not None:
+        n_annotations = len(raw.annotations)
         n_annot_chans = int(n_annotations / n_blocks)
         if np.mod(n_annotations, n_blocks):
             n_annot_chans += 1
@@ -252,7 +253,7 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
                  'were appended to all channels when writing the final block.')
 
     # write annotations
-    if raw.annotations:
+    if annots is not None:
         for desc, onset, duration in zip(raw.annotations.description,
                                          raw.annotations.onset,
                                          raw.annotations.duration):
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -30,7 +30,7 @@ def _try_to_set_value(header, key, value, channel_index=None):
                            f"{return_val}.")
 
 
-def _export_raw(fname, raw, physical_range):
+def _export_raw(fname, raw, physical_range, add_ch_type):
     """Export Raw objects to EDF files.
 
     TODO: if in future the Info object supports transducer or
@@ -132,7 +132,8 @@ def _export_raw(fname, raw, physical_range):
     # set channel data
     for idx, ch in enumerate(ch_names):
         ch_type = ch_types[idx]
-        signal_label = f'{ch_type.upper()} {ch}'
+        signal_label = f'{ch_type.upper()} ' if add_ch_type else ''
+        signal_label = signal_label + f'{ch}'
         if len(signal_label) > 16:
             raise RuntimeError(f'Signal label for {ch} ({ch_type}) is '
                                f'longer then 16 characters, which is not '
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -79,7 +79,8 @@ def _export_raw(fname, raw, physical_range):
     filter_str_info = f"HP:{highpass}Hz LP:{lowpass}Hz N:{linefreq}Hz"
 
     # get EEG-related data in uV
-    units = dict(eeg='uV', ecog='uV', seeg='uV', eog='uV', ecg='uV', emg='uV')
+    units = dict(eeg='uV', ecog='uV', seeg='uV', eog='uV', ecg='uV', emg='uV',
+                 bio='uV', dbs='uV')
 
     # get the entire dataset in uV
     data = raw.get_data(units=units, picks=ch_names)
@@ -89,7 +90,7 @@ def _export_raw(fname, raw, physical_range):
         ch_types_phys_max = dict()
         ch_types_phys_min = dict()
 
-        ch_types = np.array(raw.get_channel_types(picks=raw.ch_names))
+        ch_types = np.array(raw.get_channel_types(picks=ch_names))
         for _type in np.unique(ch_types):
             _picks = np.nonzero(ch_types == _type)[0]
             _data = raw.get_data(units=units, picks=_picks)
@@ -123,7 +124,6 @@ def _export_raw(fname, raw, physical_range):
             # take the channel type minimum and maximum
             ch_type = ch_types[idx]
             pmin, pmax = ch_types_phys_min[ch_type], ch_types_phys_max[ch_type]
-
         for key, val in [('PhysicalMaximum', pmax),
                          ('PhysicalMinimum', pmin),
                          ('DigitalMaximum', digital_max),
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -204,16 +204,19 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
         _try_to_set_value(hdl, 'DataRecordDuration', data_record_duration)
 
     # compute number of data records to loop over
-    n_blocks = n_times / out_sfreq
+    n_blocks = np.ceil(n_times / out_sfreq).astype(int)
 
     # increase the number of annotation signals if necessary
     n_annotations = len(raw.annotations)
-    if n_annotations > n_blocks:
-        n_annot_signals = np.ceil(n_annotations / n_blocks).astype(int)
-        hdl.setNumberOfAnnotationSignals(n_annot_signals)
+    if n_annotations is not None:
+        n_annot_chans = int(n_annotations / n_blocks)
+        if np.mod(n_annotations, n_blocks):
+            n_annot_chans += 1
+        if n_annot_chans > 1:
+            hdl.setNumberOfAnnotationSignals(n_annot_chans)
 
     # Write each data record sequentially
-    for idx in range(np.ceil(n_blocks).astype(int)):
+    for idx in range(n_blocks):
         end_samp = (idx + 1) * out_sfreq
         if end_samp > n_times:
             end_samp = n_times
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -4,8 +4,6 @@
 # License: BSD-3-Clause
 
 import numpy as np
-from numpy.testing import assert_allclose, assert_array_almost_equal
-from scipy.spatial.distance import pdist
 
 from ..utils import _check_edflib_installed, warn
 _check_edflib_installed()
@@ -107,8 +105,8 @@ def _export_raw(fname, raw, physical_range):
             # a consistent scale
             # - then basically do hard-threshold check for
             # super outliers
-            max_diffs = pdist(data_max)
-            min_diffs = pdist(data_min)
+            # max_diffs = pdist(data_max)
+            # min_diffs = pdist(data_min)
 
             ch_types_phys_max[_type].append(data_max)
             ch_types_phys_min[_type].append(data_min)
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -62,11 +62,13 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     # note: we can write these other channels, such as 'misc'
     # but they just won't work because EDF doesn't support
     # the idea of a 'misc' channel
-    voltage_types = list(units) + ['stim']
+    voltage_types = list(units) + ['stim', 'misc']
     if any([ch not in voltage_types for ch in orig_ch_types]):
-        warn('There are non-voltage channels. '
+        warn('There are non-voltage channels that are not set as "misc". '
              'EDF only supports writing Voltage-based data. '
-             'These channels will not be written correctly.')
+             'These channels will not be written to the EDF file. '
+             'If you would still like these channels to be written, '
+             'set their channel type to "misc".')
 
     ch_names = [ch for ch in raw.ch_names if ch not in drop_chs]
     ch_types = np.array(raw.get_channel_types(picks=ch_names))
@@ -183,7 +185,7 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     # set measurement date
     meas_date = raw.info['meas_date']
     if meas_date:
-        subsecond = meas_date.microsecond / 100.
+        subsecond = int(meas_date.microsecond / 100)
         if hdl.setStartDateTime(year=meas_date.year, month=meas_date.month,
                                 day=meas_date.day, hour=meas_date.hour,
                                 minute=meas_date.minute,
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -4,6 +4,8 @@
 # License: BSD-3-Clause
 
 import numpy as np
+from numpy.testing import assert_allclose, assert_array_almost_equal
+from scipy.spatial.distance import pdist
 
 from ..utils import _check_edflib_installed, warn
 _check_edflib_installed()
@@ -94,8 +96,22 @@ def _export_raw(fname, raw, physical_range):
         for _type in np.unique(ch_types):
             _picks = np.nonzero(ch_types == _type)[0]
             _data = raw.get_data(units=units, picks=_picks)
-            ch_types_phys_max[_type] = _data.max()
-            ch_types_phys_min[_type] = _data.min()
+            data_max = _data.max(axis=1)
+            data_min = _data.min(axis=1)
+
+            # perform a check to see that all data ranges are
+            # within two decimal places away, else it is possible
+            # that the user incorrectly set channel groupings
+            # XXX: not sure how to compare to get a robust check
+            # - needs to account for maybe normalizing data onto
+            # a consistent scale
+            # - then basically do hard-threshold check for 
+            # super outliers
+            max_diffs = pdist(data_max)
+            min_diffs = pdist(data_min)
+
+            ch_types_phys_max[_type].append(data_max)
+            ch_types_phys_min[_type].append(data_min)
     else:
         # get the physical min and max of the data in uV
         # Physical ranges of the data in uV is usually set by the manufacturer
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -62,7 +62,7 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     # note: we can write these other channels, such as 'misc'
     # but they just won't work because EDF doesn't support
     # the idea of a 'misc' channel
-    voltage_types = list(units.keys()) + ['stim']
+    voltage_types = list(units) + ['stim']
     if any([ch not in voltage_types for ch in orig_ch_types]):
         warn('There are non-voltage channels. '
              'EDF only supports writing Voltage-based data. '
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -206,6 +206,12 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     # compute number of data records to loop over
     n_blocks = n_times / out_sfreq
 
+    # increase the number of annotation signals if necessary
+    n_annotations = len(raw.annotations)
+    if n_annotations > n_blocks:
+        n_annot_signals = np.ceil(n_annotations / n_blocks).astype(int)
+        hdl.setNumberOfAnnotationSignals(n_annot_signals)
+
     # Write each data record sequentially
     for idx in range(np.ceil(n_blocks).astype(int)):
         end_samp = (idx + 1) * out_sfreq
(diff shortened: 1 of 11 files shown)


cbrnr:
The test doesn't cover the `bio` and `dbs` types you added here. Do we really want these types? The [docstring](https://github.com/mne-tools/mne-python/blob/main/mne/utils/docs.py#L2507) says EDF export only supports EEG, ECoG, and sEEG, so this should be updated in any case (whether or not we add `bio` and `dbs`, because we already support `eog`, `ecg`, and `emg`).

adam2392:
> The test doesn't cover the `bio` and `dbs` types you added here. Do we really want these types? The [docstring](https://github.com/mne-tools/mne-python/blob/main/mne/utils/docs.py#L2507) says EDF export only supports EEG, ECoG, and sEEG, so this should be updated in any case (whether or not we add `bio` and `dbs`, because we already support `eog`, `ecg`, and `emg`).

Oh apologies, forgot to push up the updated test.

Upon further inspection, I realized that EDF cannot write the type of channels in the header, as there is nowhere to store that. I also realized that `read_raw_edf` will end up losing the channel type information. With that being said, one should be able to write any type of "voltage" type signals into the EDF file, as I have seen EOG, ECG and even EMG before. DBS and Bio are just future-proofing since it doesn't hurt to add them.

What we might need to do though is "specify" that you lose channel type information in the doc string.

However, with packages like mne-bids, this issue is circumvented since one would just leverage the `channels.tsv` file in BIDS to set channel types.

Teuniz:
> Upon further inspection, I realized that EDF cannot write the type of channels in the header, as there is nowhere to store that.

There is:  
  
The header field 'label' offers 16 ASCII characters. The standard structure consists of three components, from left to right:  
  
    Type of signal (for example EEG).  
    A space.  
    Specification of the sensor (for example Fpz-Cz).  

https://edfplus.info/specs/edftexts.html#label

https://edfplus.info/specs/edftexts.html#signals

adam2392:
Sorry @agramfort there were some updates pertaining to how EDF should be read/written.

This is good to go again on my end, and I have added in all the thoughts from #9704

mne/export/tests/test_export.py
```diff
@@ -55,11 +55,14 @@ def test_export_raw_eeglab(tmpdir):
                     reason='edflib-python not installed')
 def test_integer_sfreq_edf(tmp_path):
     """Test saving a Raw array with integer sfreq to EDF."""
-    np.random.RandomState(12345)
+    prng = np.random.RandomState(12345)
```
agramfort, line 58:
```suggestion
    rng = np.random.RandomState(12345)
```
we use rng everywhere in mne

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
 
         nchan = int(_edf_str(fid.read(4)))
         channels = list(range(nchan))
-        ch_names = [fid.read(16).strip().decode('latin-1') for ch in channels]
+
+        # read in 16 byte labels and strip any extra spaces at the end
+        signal_labels = [fid.read(16).strip().decode('latin-1')
```
agramfort, line 646:
```suggestion
        ch_labels = [fid.read(16).strip().decode('latin-1')
```
is more similar to mne naming

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
...
+                         for ch in channels]
+
+        # ch_names = signal_labels.copy()
+        # get channel names and optionally channel type
+        # EDF specification contains 16 bytes that encode channel names,
+        # optionally prefixed by a string representing channel type separated
+        # by a space
+        ch_names = []
+        ch_types = []
+        for ch, new_label in enumerate(signal_labels):
```
agramfort, line 656:
```suggestion
        for ch_idx, this_label in enumerate(signal_labels):
```

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
...
+            # space is found, so the prefix is the channel type. 'Annotations'
+            # is also a keyword we search for
+            if new_label.count(' ') == 1 and ('Annotations' not in new_label):
+                ch_type, ch_name = new_label.split(' ')
+
+                # channel types should be upper case for easy comparison
+                ch_type = ch_type.upper()
+
+            ch_names.append(ch_name)
+            ch_types.append(ch_type)
```
agramfort, line 670: @Teuniz what do you think about this these changes to our EDF writer? thx

agramfort (commented):
what does eeglab do when reading files with ch types like this?

mne/io/edf/edf.py
Teuniz: Just one question, what is going to happen if, an inputfile in EDF format which contains (besides EEG signals) signals  coming from, for example, an 3-axis accelerometro, is going to be exported to EDF?  
In other words, what happens if a signals' type isn't covered by a standard list of types?  
  
Assigning the type EEG (because it's the fallback)  to an accelerometer signal doesn't seem to be a good idea.  
Because it's practically impossible to cover any kind of sensor and future developments, maybe it could be an idea  
 to have a type unknown or unspecified?  
Just brainstorming here...

cbrnr: We do have a `'MISC'` type that we could assign in this case.

cbrnr:
> what does eeglab do when reading files with ch types like this?

I assume that readers that do not parse the label field will just display the complete text as a channel label, e.g. "EEG Fz" instead of just "Fz". For this reason, I would like to have an option to turn channel type parsing off.

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
...
+        # get channel names and optionally channel type
+        # EDF specification contains 16 bytes that encode channel names,
+        # optionally prefixed by a string representing channel type separated
+        # by a space
+        ch_names = []
+        ch_types = []
+        for ch, new_label in enumerate(signal_labels):
+            # if no channel type, then we will default to eeg
+            ch_type = None
+            ch_name = new_label
```
adam2392, line 659:
```suggestion
            ch_name = this_label
```

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
...
+
+            # space is found, so the prefix is the channel type. 'Annotations'
+            # is also a keyword we search for
+            if new_label.count(' ') == 1 and ('Annotations' not in new_label):
```
adam2392, line 663:
```suggestion
            if this_label.count(' ') == 1 and ('Annotations' not in this_label):
```

mne/io/edf/edf.py
```diff
@@ -602,7 +641,33 @@ def _read_edf_header(fname, exclude):
...
+                ch_type, ch_name = new_label.split(' ')
```
adam2392, line 664:
```suggestion
                ch_type, ch_name = this_label.split(' ')
```

mne/io/edf/edf.py
adam2392: Upon reading, MNE has traditionally assumed EEG by default, so I think assuming it by default is fine, unless we want to deprecate that behavior?

Unless you mean the following:
1. if there **are** channel types encoded in the EDF signal labels, and then only some of them don't have it, then those should be easily assigned `'misc'`?
2. otherwise if there are no channel types at all, then the default backwards-compatible behavior is assuming `'eeg'`

Should I implement 1?
wdyt @agramfort ?

Upon writing... I agree writing the channel type is not always desirable if we want to control for erroneous user behavior. Thus I added the `add_ch_type` kwarg to export per @cbrnr 's suggestion.

mne/io/edf/edf.py
```diff
@@ -415,6 +447,13 @@ def _get_info(fname, stim_channel, eog, misc, exclude, preload):
             edf_info['units'][idx] = 1
         chs.append(chan_info)
 
+    # warn if channel type was not inferable
+    if len(bad_map):
+        bad_map = '\n'.join([f'{ch_name}: {ch_type}'
+                            for ch_name, ch_type in bad_map.items()])
```
drammock, line 453:
```suggestion
        bad_map = '\n'.join(f'{ch_name}: {ch_type}'
                            for ch_name, ch_type in bad_map.items())
```
don't need to construct list compr. here, `join` accepts generators

mne/io/edf/edf.py
```diff
@@ -602,7 +641,34 @@ def _read_edf_header(fname, exclude):
 
         nchan = int(_edf_str(fid.read(4)))
         channels = list(range(nchan))
-        ch_names = [fid.read(16).strip().decode('latin-1') for ch in channels]
+
+        # read in 16 byte labels and strip any extra spaces at the end
+        ch_labels = [fid.read(16).strip().decode('latin-1')
+                     for ch in channels]
+
+        # ch_names = signal_labels.copy()
```
drammock, line 649:
```suggestion
```

mne/io/edf/tests/test_edf.py
```diff
@@ -408,8 +409,8 @@ def test_bdf_multiple_annotation_channels():
 def test_edf_lowpass_zero():
     """Test if a lowpass filter of 0Hz is mapped to the Nyquist frequency."""
     raw = read_raw_edf(edf_stim_resamp_path)
-    assert raw.ch_names[100] == 'EEG LDAMT_01-REF'
-    assert len(raw.ch_names[100]) > 15
+    assert raw.ch_names[100] == 'LDAMT_01-REF'
+    assert len(raw.ch_names[100]) > 11
```
drammock, line 413:
```suggestion
    assert raw.ch_names[100] == 'LDAMT_01-REF'
```
we've already asserted exact match of the name, additionally asserting string length is redundant

mne/io/edf/tests/test_edf.py
```diff
@@ -525,3 +526,19 @@ def test_exclude():
...
+
+
+@testing.requires_testing_data
+def test_ch_types():
+    """Test reading of channel types from EDF channel label."""
+    raw = read_raw_edf(edf_chtypes_path)
+
+    # get the channel types for all channels
+    ch_types = raw.get_channel_types()
+    assert all([x in ch_types for x in ['eeg', 'ecg']])
```
drammock, line 538:
```suggestion
    assert all(x in ch_types for x in ['eeg', 'ecg'])
```
`all` can handle generators

mne/io/edf/tests/test_edf.py
```diff
@@ -525,3 +526,19 @@ def test_exclude():
...
+
+    # test certain channels were correctly detected as a channel type
+    test_ch_names = {"Fp1-Ref": "eeg", "P10-Ref": "eeg", "DC01": "eeg",
+                     "ECG1": "ecg", "ECG2": "ecg"}
+    assert all([raw.get_channel_types(picks=pick)[0] == ch_type
+                for pick, ch_type in test_ch_names.items()])
```
drammock, line 544:
```suggestion
    assert all(raw.get_channel_types(picks=pick)[0] == ch_type
               for pick, ch_type in test_ch_names.items())
```

mne/utils/docs.py
```diff
@@ -2499,13 +2499,31 @@
...
-addition, EDF does not support storing a montage.
+For EDF exports, only channels that are measured in terms of Voltage are
+supported. For example, 'eeg', 'ecog', 'seeg', 'bio', 'emg', 'eog',
+'ecg', 'dbs', 'bio' data are supported. However, EDF files do not
+require the storage of channel types, so when re-reading the file
+naively, you will only get 'eeg' types by default. For example, if you
+have 'misc' channels, they will be read back in as 'eeg' type. If you
+want to set the channel types, then set the ``add_ch_type`` keyword.
+This will then for example take EEG channel ``Fz1`` and export the
+signal label as ``EEG Fz1`` in the EDF file. In order to properly set
```
drammock, line 2521:
```suggestion
This will then for example take EEG channel ``Fz`` and export the
signal label as ``EEG Fz`` in the EDF file. In order to properly set
```
Let's use a realistic channel name in our example :)

mne/utils/docs.py
```diff
@@ -2499,13 +2499,31 @@
...
 For EEGLAB exports, channel locations are expanded to full EEGLAB format.
 For more details see :func:`eeglabio.utils.cart_to_eeglab`.
 """
 docdict['export_edf_note'] = """
-For EDF exports, only EEG, ECoG and sEEG data are supported. In
-addition, EDF does not support storing a montage.
+For EDF exports, only channels that are measured in terms of Voltage are
+supported. For example, 'eeg', 'ecog', 'seeg', 'bio', 'emg', 'eog',
+'ecg', 'dbs', 'bio' data are supported. However, EDF files do not
+require the storage of channel types, so when re-reading the file
```
drammock, line 2516: Maybe I missed something in the earlier discussion, but here you say "EDF does not require storing channel type".  That implies that it's possible, which means we can (and should) store channel type when we export.  Further down you give an example of "misc" being a problem, but "misc" channels are often in Volts (as are "stim", which are not included in your list of good types).  So reading this note, it's not clear to me exactly what situations will lead to channel type information being lost on round-trip.  Can you clarify what exactly is the issue, and make the note text here reflect that more closely?

Is the problem that *only a predefined lexicon of channel type names* are allowed?  Or is it that EDF does not even *allow* storing channel types (other than as part of the channel label)?  Or something else?

As appropriate, please distinguish between problems that are limitations of the EDF format, and problems due to other EDF writers that may not include all optional information.  (The latter case doesn't need to be addressed in this docstring, since this is the docstring for our *exporter*...  problems related to reading other writers' EDF files belong in the docstring for our EDF *reader*).

drammock (changes requested):

mne/export/_edf.py
```diff
@@ -53,7 +58,18 @@ def _export_raw(fname, raw, physical_range):
...
 
+    # add warning if any channel types are not voltage based
+    # note: we can write these other channels, such as 'misc'
+    # but they just won't work because EDF doesn't support
+    # the idea of a 'misc' channel
+    voltage_types = list(units.keys()) + ['stim']
+    if any([ch not in voltage_types for ch in orig_ch_types]):
+        warn('There are non-voltage channels. '
+             'EDF only supports writing Voltage-based data. '
+             'These channels will not be written correctly.')
```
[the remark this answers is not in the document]
adam2392: I updated the description to be more accurate. 

These channels are actually "dropped" and not exported. The reason being EDF does not support writing these non-voltage / misc channels. 

E.g. if there is a meg channel, that won't be written.

mne/utils/docs.py
adam2392: I updated the docs significantly to reflect what are the limitations of exporting and reading EDF files. 

The problem for writing/reading channel types is that there is no predefined lexicon of channel type names. Therefore, there are just commonly "suggested ones", such as EEG. 

lmk wyt?

mne/utils/docs.py
```diff
@@ -2499,13 +2499,36 @@
...
+that ``Fz`` is an EEG channel and ``MISC E`` would imply ``E`` is a MISC
+channel. However, EDF files do not require the storage of channel types,
+so many writers do not export the channel type. This is because
+there is no standard way for specifying all channel types.
+
+If you want to set the channel types, then set the ``add_ch_type`` keyword.
+This will then for example take EEG channel ``Fz`` and export the
+signal label as ``EEG Fz`` in the EDF file. In order to properly set
+the channel types before exporting, you will need to call
+:attr:`raw.set_channel_types <mne.io.Raw.set_channel_types>`.
```
drammock, line 2527: this paragraph is mostly redundant with the previous paragraph and the docstring of `add_ch_type`.  The only non-redundant part is the last sentence, which I'm not sure is even accurate?  It implies that in (many? most? all?) cases, the channel types will not be properly set, and users will need to do something extra to get them properly set.  Can you clarify in which cases you think users will want/need to use `raw.set_channel_types()`?

drammock (changes requested):
@adam2392 in addition to my in-line suggestions, take a look at [the rendered docstring](https://34002-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_raw.html), and you'll see a few other things that need fixing:

1. add EDF to the "supported formats" line (above the warning box)
2. fix the warning box (looks like a crossref is not rendering properly, as message ends mid-sentence)
3. the parameter `fmt` says only `eeglab` or `auto`, we should add `edf` there

mne/utils/docs.py
```diff
@@ -2499,13 +2499,36 @@
...
-addition, EDF does not support storing a montage.
+For EDF exports, only channels that are measured in terms of Voltage are
+supported. For example, 'eeg', 'ecog', 'seeg', 'bio', 'emg', 'eog',
+'ecg', 'dbs', and 'bio' data are supported. 'misc' channels are also
+allowed for a catch-all of "unknown" channel types. Channel types are able
+to be stored in the prefix of the signal label. For example, ``EEG Fz`` implies
+that ``Fz`` is an EEG channel and ``MISC E`` would imply ``E`` is a MISC
+channel. However, EDF files do not require the storage of channel types,
+so many writers do not export the channel type. This is because
+there is no standard way for specifying all channel types.
```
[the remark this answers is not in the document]
adam2392: Stim is dropped, so I can update this doc string here.

mne/utils/docs.py
adam2392: I updated the doc string here now. Lmk wyt.

adam2392:
> @adam2392 in addition to my in-line suggestions, take a look at [the rendered docstring](https://34002-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_raw.html), and you'll see a few other things that need fixing:
> 
> 1. add EDF to the "supported formats" line (above the warning box)

Done.

> 2. fix the warning box (looks like a crossref is not rendering properly, as message ends mid-sentence)

Looks like this was an issue in the new dev version. I fixed it by adding to `export_raw`.

https://mne.tools/dev/generated/mne.export.export_raw.html#mne.export.export_raw

> 3. the parameter `fmt` says only `eeglab` or `auto`, we should add `edf` there

Done.

adam2392:
The failure looks like something stalled on CI installing reqs? perhaps just needs a re-run?

https://github.com/mne-tools/mne-python/pull/9694/checks?check_run_id=3589203613

larsoner:
Will restart once #9731 is in

mne/utils/docs.py
```diff
@@ -2501,33 +2501,33 @@
...
+the EDF standard.
+
+If ``add_ch_type`` is True, then channel types are written based on what
+they are currently set in MNE-Python. One should double check that all
+their channels are set correctly. You can call
+:attr:`raw.set_channel_types <mne.io.Raw.set_channel_types>` to set
+channel types.
+
+Exporting EDF currently only supports the storage of one annotation
+per data record.
```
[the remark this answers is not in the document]
adam2392: https://github.com/mne-tools/mne-python/pull/9643#issuecomment-893272302

This thread talked about it.

The way that it is currently implemented only allows one annotation per data record. The underlying library allows us to modify the number of annotation channels there are. 

I can add this in if you think it is necessary? The algorithm would probably proceed as:

```
# compute number of data records there are

# compute number of annotations there are

# adjust the number of annotation signals there are using https://gitlab.com/Teuniz/EDFlib-Python/-/blob/master/src/EDFlib/edfwriter.py#L770


# write annotations 
```
```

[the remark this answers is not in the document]
Comment (adam2392) [reply mne/utils/docs.py:2527]: Actually it is trivial to add in, so I just added it in. :p sorry bout that.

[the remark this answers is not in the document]
Comment (Teuniz) [reply mne/utils/docs.py]: EDFlib writes max one annotation per **annotationchannel** per datarecord and it's up to the application to  increase either the number of annotation channels or decrease the datarecord duration in order to create  
more storage space for annotations.

File: mne/export/_edf.py [hunk 19]
```diff
@@ -206,6 +206,12 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
     # compute number of data records to loop over
     n_blocks = n_times / out_sfreq
 
+    # increase the number of annotation signals if necessary
+    n_annotations = len(raw.annotations)
```
drammock, line 210: this will fail if `raw.annotations` is `None`.

mne/export/_edf.py
```diff
@@ -206,6 +206,12 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
...
+    if n_annotations > n_blocks:
+        n_annot_signals = np.ceil(n_annotations / n_blocks).astype(int)
+        hdl.setNumberOfAnnotationSignals(n_annot_signals)
```
drammock, line 213: this will request 200 annotation channels if there are 200 annotations.  That seems... suboptimal.  If the Raw file is 600 seconds (10 minutes) long and has a 0.5-second annotation every 3 seconds, you would have 200 annotations and yet could write all of them on a single annotation channel (because they're guaranteed to be in different data blocks).

I don't know the ins and outs of how EDFLib handles the annotations (e.g., does a single annotation persist across data blocks until its offset, or are only the onsets and durations written in a given block?) but it seems like comparing the annotation onset times (and maybe durations) to the data blocks' start and stop times should allow you to efficiently figure out how many channels you will need.  `np.searchsorted` springs to mind, though there may be more elegant ways.

mne/export/_edf.py
```diff
@@ -206,6 +206,12 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
...
+
     # Write each data record sequentially
     for idx in range(np.ceil(n_blocks).astype(int)):
```
drammock, line 216: see above re: moving the `np.ceil(...).astype(int)`

drammock (commented):
in addition to the below comments, CI fails now with

`/home/circleci/project/mne/export/_export.py:docstring of mne.export._export.export_raw:4: WARNING: py:mod reference target not found: EDFlib-Python`

you would need to add EDFLib-Python to `conf.py` intersphinx list (if it supports it) for the `:mod:` reference to work.

adam2392:
> in addition to the below comments, CI fails now with
> 
> 
> 
> `/home/circleci/project/mne/export/_export.py:docstring of mne.export._export.export_raw:4: WARNING: py:mod reference target not found: EDFlib-Python`
> 
> 
> 
> you would need to add EDFLib-Python to `conf.py` intersphinx list (if it supports it) for the `:mod:` reference to work. 

Ah I don't think that works. Perhaps I'll just wrap in double ' to bold face it?

mne/export/_edf.py
Teuniz: Neither the onset time or duration of an annotation affects the number of annotations that can be written.  
EDFlib works basically like this: you can write as many annotations as you like using the api. They will  
be temporary stored in memory. Only when you close the file, EDFlib will set the filepointer to the first  
datarecord and write an annotation, then hops to the next datarecord and write the next annotation and  
so on untill there are no annotations or datarecords anymore, whichever one comes first.  
  
If the number of annotations and the number of datarecords are known before writing the datasamples,  
the following formula can be used to calculate the number of annotation channels needed:  
  
`annot_channs = (annots / datrecs) + (annots % datrecs);`  
`if(annot_channs > 1)`  
`{`  
`  setNumberOfAnnotationSignals(annot_channs);`  
`}`

adam2392: Thanks for the explanation @Teuniz! 

@drammock I've adjusted the code to address your comments.

drammock: Ah ok, I was wrong, I see it works very differently than I assumed. Thanks @Teuniz for the explanation. @adam2392 I'll review again first thing tomorrow and hopefully get this merged!

adam2392:
`mne/tests/test_cov.py::test_compute_whitener` seems to fail, but I don't think I made any changes affecting that.

mne/export/_export.py
```diff
@@ -15,8 +15,10 @@ def export_raw(fname, raw, fmt='auto', physical_range='auto',
     """Export Raw to external formats.
 
     Supported formats:
-    - EEGLAB (set, uses :mod:`eeglabio`)
-    - EDF (edf, uses ``EDFlib-Python``)
+
+        - EEGLAB (set, uses :mod:`eeglabio`)
```
cbrnr, line 19:
```suggestion
        - EEGLAB (.set, uses :mod:`eeglabio`)
```

mne/export/_export.py
```diff
@@ -15,8 +15,10 @@ def export_raw(fname, raw, fmt='auto', physical_range='auto',
...
+        - EDF (edf, uses ``EDFlib-Python``)
```
cbrnr, line 20:
```suggestion
        - EDF (.edf, uses ``EDFlib-Python``)
```

No `:mod:` prefix here?

adam2392: EDFlib-Python doesn't have a sphinx generated documentation, so I don't think it would work?

mne/export/_edf.py
```diff
@@ -53,7 +58,20 @@ def _export_raw(fname, raw, physical_range):
         stim_index = np.atleast_1d(stim_index.squeeze()).tolist()
         drop_chs.extend([raw.ch_names[idx] for idx in stim_index])
 
+    # add warning if any channel types are not voltage based
+    # note: we can write these other channels, such as 'misc'
+    # but they just won't work because EDF doesn't support
+    # the idea of a 'misc' channel
```
cbrnr, line 64: This comment should be improved. It doesn't really explain this issue (`but they just won't work`) and it somewhat contradicts the warning message (`...set their channel type to "misc"`). Can you clearly explain what users are expected to do?

mne/export/_edf.py
```diff
@@ -53,7 +58,20 @@ def _export_raw(fname, raw, physical_range):
...
+    voltage_types = list(units) + ['stim', 'misc']
+    if any([ch not in voltage_types for ch in orig_ch_types]):
+        warn('There are non-voltage channels that are not set as "misc". '
+             'EDF only supports writing Voltage-based data. '
+             'These channels will not be written to the EDF file. '
+             'If you would still like these channels to be written, '
+             'set their channel type to "misc".')
```
cbrnr, line 71: In addition to clarifying the warning, maybe we should list the channel names that cause this warning?

mne/export/_export.py
```diff
@@ -10,11 +10,16 @@
...
+def export_raw(fname, raw, fmt='auto', physical_range='auto',
+               add_ch_type=False, verbose=None):
     """Export Raw to external formats.
 
-    Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
-    %(export_warning)s
+    Supported formats:
+
+        - EEGLAB (.set, uses :mod:`eeglabio`)
+        - EDF (.edf, uses ``EDFlib-Python``)
```
cbrnr, line 20: What about the `:mod:` prefix? Why is it used for `eeglabio` but not for `EDFlib-Python`?

adam2392: Because eeglabio has sphinx generated docs, but EDFlib-Python does not

mne/export/_edf.py
adam2392: Done

mne/export/_edf.py
adam2392: Added some additional clarification.

cbrnr:
I'm sorry if you have already mentioned it in this long thread, but how does EDF only support voltage data? The specs just mention that the physical dimension for each channel are 8 ASCII characters, so any unit should be supported.

adam2392:
> I'm sorry if you have already mentioned it in this long thread, but how does EDF only support voltage data? The specs just mention that the physical dimension for each channel are 8 ASCII characters, so any unit should be supported.

Idk actually. I think possibly we could export any channel type, but I would prefer not implementing the ability to export MEG -> EDF. My impression was that EDF should only be used for EEG-like data. I don't see why anyone would export MEG/non-voltage data to EDF and not FIF unless it was by accident, so we have some catches in export to handle that.

Ref: https://www.edfplus.info/specs/edftexts.html

cbrnr:
I disagree. The table on the specs page even lists MEG as a signal type. I think we should not artificially restrict our export functionality by allowing only voltage data, which is not backed up by the EDF specs.

cbrnr:
Re exporting to EDF vs. FIFF, I think EDF is much more widely adopted/used than FIFF, but even if FIFF is the de facto standard for MEG signals the EDF format allows signals of any unit to be stored.

mne/export/_edf.py
```diff
@@ -30,14 +30,19 @@ def _try_to_set_value(header, key, value, channel_index=None):
...
     """Export Raw objects to EDF files.
 
     TODO: if in future the Info object supports transducer or
     technician information, allow writing those here.
     """
+    # scale to save data in EDF
     phys_dims = 'uV'
 
+    # get EEG-related data in uV
+    units = dict(eeg='uV', ecog='uV', seeg='uV', eog='uV', ecg='uV', emg='uV',
```
adam2392, line 43: @cbrnr can you lmk which data channels you would like included here? I'm not a MEG user, so I've only ever used MNE for EEG/iEEG.

Would it be:

```
mag='fT', grad='fT/cm'
```

Anything else?

adam2392:
> I disagree. The table on the specs page even lists MEG as a signal type. I think we should not artificially restrict our export functionality by allowing only voltage data, which is not backed up by the EDF specs.

Okay sure I can add in then MEG support. Can you lmk which channels and their corresponding default units I should add in to this comment? I'm not too familiar with non-EEG data types in MNE.

https://github.com/mne-tools/mne-python/pull/9694#discussion_r708294036

mne/export/_edf.py
adam2392: Actually the challenge is more-so on reading in MEG/fnirs data correctly using `read_raw_edf`. I'm not sure how to read in this data correctly. Sure I can easily append "MEG M1" to indicate M1 is a MEG channel, or "FNIRS M2" to indicate M2 is a FNIRS channel, but reading the data back in to do a unit test on the round trip is non-trivial for me.

Is it possible to leave this PR as is and then table it for a future improvement from someone who wants to export MEG/Fnirs to EDF and is familiar with the reader of MEG/Fnirs enough to add that functionality in?

larsoner: +1 for doing it later. I assume YAGNI, and for MEG important information will be lost anyway (at the very least, the head<->device transform, sensor orientations, sensor coil types)

adam2392:
https://34137-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_raw.html#mne.export.export_raw

This lgtm to go if we are okay w/ the future PR.

mne/export/_edf.py
drammock: please don't mark conversations as "resolved" without at least commenting on why you think it's not a problem @adam2392 (or changing something if the comment is valid).  Is this indeed not a problem?

mne/export/_edf.py
```diff
@@ -58,17 +58,24 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
...
+    if any(non_voltage_ch):
         warn('There are non-voltage channels that are not set as "misc". '
+             f'{non_voltage_ch}'
              'EDF only supports writing Voltage-based data. '
              'These channels will not be written to the EDF file. '
              'If you would still like these channels to be written, '
-             'set their channel type to "misc".')
+             'you can set their channel type to "misc", but they will '
+             'most likely not be written correctly. For example, EDF '
+             'does not support storing MEG channels.')
```
drammock: Here's I think a more accurate message (GitHub won't let me suggest this inline because it starts on a non-changed line)

```py
        warn(f"Non-voltage channels detected: {non_voltage_ch}. MNE-Python's "
             'EDF exporter only supports voltage-based channels, because the '
             'EDF format cannot accommodate much of the accompanying data '
             'necessary for channel types like MEG and fNIRS (channel '
             'orientations, coordinate frame transforms, etc). You can '
             'override this restriction by setting those channel types to '
             '"misc" but no guarantees are made of the fidelity of that '
             'approach.')
```

mne/export/_edf.py
adam2392: Ah sorry, I thought I resolved it by checking if `n_anotations is None`. I realize that's not solving the issue. Let me push a fix.

adam2392: Pushed a fix.

mne/export/_edf.py
adam2392: Changed the warning message.

drammock (approved):

adam2392:
Don't want to lose track of this PR since it had a lot of discussion.

Is this good to merge now?

mne/export/_edf.py
cbrnr: The same could be said for EEG as well because EDF doesn't store custom sensor locations. But let's stick to exporting EEG for now, the warning message and code comment are now clear that this is not a restriction imposed by EDF.

mne/io/edf/edf.py
```diff
@@ -384,6 +388,27 @@ def _get_info(fname, stim_channel, eog, misc, exclude, preload):
...
+        'BIO': FIFF.FIFFV_BIO_CH,
+        'RESP': FIFF.FIFFV_RESP_CH,
+        'MISC': FIFF.FIFFV_MISC_CH,
+        'SAO2': FIFF.FIFFV_BIO_CH,
+    }
+    bad_map = dict()
+
+    # montage is not able to be stored in EDF, so
+    # default to unknown locations for the channels
+    nan_arr = np.zeros(12)
```
cbrnr, line 409: Why `12`?

Also, if you want an array of all nans consider using `np.full`.

agramfort (approved):

agramfort:
thx @adam2392

mne/io/edf/edf.py
agramfort: mne uses 12 values for channel locations as it is what you need for MEG. You need location and rotation matrix.

cbrnr: Yes, I realized that, I thought a comment would have been nice. But it's all good, this is really minor.

mne/io/edf/edf.py
```diff
@@ -396,7 +421,21 @@ def _get_info(fname, stim_channel, eog, misc, exclude, preload):
         chan_info['coord_frame'] = FIFF.FIFFV_COORD_HEAD
         chan_info['coil_type'] = FIFF.FIFFV_COIL_EEG
         chan_info['kind'] = FIFF.FIFFV_EEG_CH
-        chan_info['loc'] = np.zeros(12)
+        chan_info['loc'] = nan_arr
```
larsoner, line 424: This needed to be `nan_arr.copy()` (as found with @drammock working together), after this PR all `info['chs'][ii]['loc']` point to the same array in memory, so setting values in one of them sets all of them. We need a unit test that would have caught this...

larsoner: (I think @drammock is working on a fix for this)

drammock: see #9743

PR #9689, opened 2021-08-23
WIP,ENH: Coreg GUI

GuillaumeFavelier:
This PR adds a `CoregistrationUI` class which acts as the the user interface for coregistration that relies on `pyvistaqt`.

This is still a work in progress, the UI components are not connected to `Coregistration` yet.

**Todo**
* the GUI API is not complete (radio buttons not ready, dock widgets to fetch files)
* multi-dock system needs rework (only temporary solution for now) *[ipywidget]*
* reuse of `plot_alignment()` temporarily but this is not viable long-term -> needs refactoring

**Known bugs**
* inconsistent behaviour of checkbox layout *[ipywidget]*
* because of `plot_alignment()`, there is a double plot after the one from coreg *[ipywidget]*

It's an item from https://github.com/mne-tools/mne-python/issues/8833

**Current status**
image

PR #9697, opened 2021-08-25
ENH: Add vertex number to vol viewer

larsoner:
I found an old branch where I did this. Unclear why I never opened the PR...

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -85,6 +85,8 @@ Enhancements
 
 - Add tutorial for how to processes image (CT and MR) files in order to localize electrode contacts for intracranial recordings :ref:`tut-ieeg-localize` (:gh:`9484` by `Alex Rockhill`_)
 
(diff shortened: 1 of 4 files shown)


larsoner:
Argh PyVista:
```
pyvista.utilities.misc.PyvistaDeprecationWarning: Use of `point_arrays` is deprecated. Use `point_data` instead.
```
Looks like we have a few uses. @GuillaumeFavelier I'll look into it unless you're already on it...

agramfort:
yes pyvista just merged a big API change....

>

GuillaumeFavelier:
> I'll look into it unless you're already on it

For now, I'm on https://github.com/mne-tools/mne-python/pull/9689. If you're busy, I can take care of it later.

> pyvista just merged a big API change

In https://github.com/pyvista/pyvista/pull/1593 but not released yet so only "pre"-jobs and... Circle should be affected.

PR #9702, opened 2021-08-26
[ENG, WIP] Hough Transform for 3D points

alexrockhill:
Adds a Hough transform for 3D points for MNE for use in https://github.com/mne-tools/mne-python/pull/9586.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -149,7 +149,7 @@ Enhancements
 
 - Add remove methods for mesh objects in :class:`mne.viz.Brain` (:gh:`9688` by `Alex Rockhill`_)
 
-- Add :func:`mne.transforms.hough_transform_3D` to find the lines that fit a cloud of points best (:gh:`XXX` by `Alex Rockhill`_)
+- Add :func:`mne.transforms.hough_transform_3D` to find the lines that fit a cloud of points best (:gh:`9702` by `Alex Rockhill`_)
 
 Bugs
 ~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -149,6 +149,8 @@ Enhancements
 
 - Add remove methods for mesh objects in :class:`mne.viz.Brain` (:gh:`9688` by `Alex Rockhill`_)
 
+- Add :func:`mne.transforms.hough_transform_3D` to find the lines that fit a cloud of points best (:gh:`XXX` by `Alex Rockhill`_)
+
 Bugs
 ~~~~
 - Fix bug with :meth:`mne.Epochs.crop` and :meth:`mne.Evoked.crop` when ``include_tmax=False``, where the last sample was always cut off, even when ``tmax > epo.times[-1]`` (:gh:`9378` **by new contributor** |Jan Sosulski|_)
--- a/doc/preprocessing.rst
+++ b/doc/preprocessing.rst
@@ -183,6 +183,7 @@ EEG referencing:
    :toctree: generated/
 
    Transform
+   hough_transform_3D
    quat_to_rot
    rot_to_quat
    read_ras_mni_t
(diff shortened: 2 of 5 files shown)


agramfort:
>
> I think that's a good idea except that it is a rather small function for a
> dedicated package. Maybe it would be better as a contribution to scipy?
>

start with a small package. Much simpler

alexrockhill:
Ok @agramfort, if you have a second to look sometime soonish, this is a working version that is pretty much an exact replication of the IPOL paper with the only exception to that being a bit of loosening of how close the point has to be to a line since these lines are very far apart in the iEEG use case and the point spread can sometimes be a bit high.

I tried to get the other version (exploring the full parameter space with k-means) to work, and in theory it sounded great, but in practice there were some magic number parameter specifying about distances from other Hough points that made it not good. Long story short it's really just too big a space to be doing any kind of KDE.

alexrockhill:
And as far as a new package for it, it's only ~100 lines of code which seems a bit short and it depends on mne internal functions which would be suboptimal for a standalone package (to have mne as a dependency for this kind of transform), IMO I think this is a good place for it and it might have other uses in MNE

larsoner:
So there is one parameter now, namely the distance tolerance? No other parameter like `ico` subdivision level is necessary?

alexrockhill:
The distance tolerance is relative to the ico subdivision. The grade, number of lines and xy resolution are parameters.

larsoner:
Any reason not to use something like `ico-6` which is very dense? What is the tradeoff? (Does it just take too long? Is it less accurate / finds too many lines?)

agramfort:
@alexrockhill if we merge this in main branch we maintain for life (almost)...

let's talk about this on discord

alexrockhill:
> Any reason not to use something like `ico-6` which is very dense? What is the tradeoff? (Does it just take too long? Is it less accurate / finds too many lines?)

As implemented in the paper, the lines are indexed so there is no relationship between the lines in finding the local maxima of best lines. This makes it so that the lines won't be that accurate anyway. In short, the icosahedron density is not the limiting factor for accuracy so it doesn't make more sense to spend more time computing with it.

alexrockhill:
> @alexrockhill if we merge this in main branch we maintain for life (almost)...
> 
> let's talk about this on discord

Hmmm, I see but it does have tests and it's really not that much more complicated than some of the other functions for instance `_fit_matched_points`. Happy to continue on Discord

alexrockhill:
Hey so it turned out to be pretty simple to replace the voting space icosahedron angle index with elevation and azimuth angles. The benefit of this is that nearby lines can be rounded to be together based on the `image_res`. Looked really good and worked fast. On to testing it on CT locations.

alexrockhill:
Well I guess this is all a moot point because it doesn't work on the CT data, it might find the electrodes but it also finds a bunch more lines in the skull and outside head in the noise of the CT... maybe a bit of a waste of time but very interesting, I didn't know about this before

alexrockhill:
image

This is what it looks like with the best 20 lines, it's just too much of a mess

alexrockhill:
So just an FYI, I ended up implementing something else that does this here https://github.com/mne-tools/mne-gui-addons/pull/23. I think finding the contacts is important but then at the end of the day, you still have to disambiguate them so as a use case, you'd practically want to do that with a target location e.g. from a surgical plan. That makes the problem much easier as then you can just follow a line. This would be a good solution too, first finding all the lines and then asking the user which is which but I think it's mostly academic as they should have some idea where the leads are supposed to be placed (i.e. this would be nice and even a bit better but the gain above the aforementioned PR vs difficulty implementing and pain of having false positives anyway make it not very worthwhile in my opinion). Just wanted to leave a note to follow up on this.

larsoner:
Okay, thanks for the update!

PR #9719, opened 2021-09-06
Fix ICA raising ValueError when bad channels are explicitly passed to .fit method.

### LINKED ISSUE #9716: ICA 'ValueError: array must not contain infs or NaNs' when using all channels (included marked bads)
#### Describe the bug

I have a dataset where I mark bad channels and then interpolate them. To keep track of which channel has been interpolated, I do not reset `raw.info['bads']`, and instead I prefer to pass `exclude=[]` to other MNE functions.
However, this leads to something that looks like a bug to me.

#### Steps to reproduce

Load sample dataset:
```
import mne
from pathlib import Path

data_folder = Path(mne.datasets.sample.data_path())
raw_file = data_folder / 'MEG' / 'sample' / 'sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(raw_file, preload=True)
raw.pick_types(eeg=True)
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015']
raw.interpolate_bads(reset_bads=False, mode='accurate')
```

Pick all EEG channels including bads and fit ICA:

```
picks = mne.pick_types(raw.info, eeg=True, exclude=[])
ica = mne.preprocessing.ICA(n_components=0.99, max_iter='auto')
ica.fit(raw, picks=picks, reject_by_annotation=True)
```

This code above fails with the traceback below.

However, resetting the list of bads and fitting the ICA does work, which in practice should be the same as in both case I'm picking all channels:

```
raw.info['bads'] = list()
ica = mne.preprocessing.ICA(n_components=0.99, max_iter='auto')
ica.fit(raw, picks='eeg', reject_by_annotation=True)
```

Traceback:

> /Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/preprocessing/ica.py:694: RuntimeWarning: overflow encountered in true_divide
>   data /= self.pre_whitener_
> Traceback (most recent call last):
> 
>   File "/var/folders/b1/kx2x00x95ls80zzkct4647sh0000gn/T/ipykernel_9109/2823392955.py", line 15, in <module>
>     ica.fit(raw, picks=picks, reject_by_annotation=True)
> 
>   File "<decorator-gen-420>", line 24, in fit
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/preprocessing/ica.py", line 570, in fit
>     self._fit_raw(inst, picks, start, stop, decim, reject, flat,
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/preprocessing/ica.py", line 614, in _fit_raw
>     self._fit(data, 'raw')
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/preprocessing/ica.py", line 707, in _fit
>     data = pca.fit_transform(data.T)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/utils/numerics.py", line 817, in fit_transform
>     U, S, _ = self._fit(X)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/utils/numerics.py", line 855, in _fit
>     U, S, V = _safe_svd(X, full_matrices=False)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/mne/fixes.py", line 69, in _safe_svd
>     return linalg.svd(A, **kwargs)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/scipy/linalg/decomp_svd.py", line 108, in svd
>     a1 = _asarray_validated(a, check_finite=check_finite)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/scipy/_lib/_util.py", line 293, in _asarray_validated
>     a = toarray(a)
> 
>   File "/Users/scheltie/Documents/pyvenv/neurotin/lib/python3.8/site-packages/numpy/lib/function_base.py", line 488, in asarray_chkfinite
>     raise ValueError(
> 
> ValueError: array must not contain infs or NaNs

#### Additional information

Platform:      macOS-10.16-x86_64-i386-64bit
Python:        3.8.10 (v3.8.10:3d8993a744, May  3 2021, 08:55:58)  [Clang 6.0 (clang-600.0.57)]
Executable:    /Users/scheltie/Documents/pyvenv/neurotin/bin/python3
CPU:           i386: 12 cores
Memory:        32.0 GB

mne:           0.23.0
numpy:         1.21.2 {blas=openblas, lapack=openblas}
scipy:         1.7.1
matplotlib:    3.4.3 {backend=Qt5Agg}

sklearn:       0.24.2
numba:         0.53.1
nibabel:       3.2.1
nilearn:       0.8.0
dipy:          1.4.1
cupy:          Not found
pandas:        1.3.1
mayavi:        4.7.3
pyvista:       0.31.3 {pyvistaqt=0.5.0, OpenGL 4.1 ATI-4.6.20 via AMD Radeon Pro 555X OpenGL Engine}
vtk:           9.0.1
PyQt5:         5.15.4

mscheltienne:
#### Reference issue
Fix #9716.

The marked bad channels were excluded by default in method `._compute_pre_whitener()`. However, the channel picking, excluding the bad channels IF they were not explicitly provided, is already done in `.fit` when the attribute `self.info` is defined with an info instance based on the selection of picked channels. 

I added `exclude=[]` to every pick in the if/else statement in `._compute_pre_whitener()`. 
I removed `meg=False` (default argument) and renamed `this_picks` to `picks_` to shorten the lines.

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -669,7 +669,8 @@ class ICA(ContainsMixin):
                 picks_ = pick_types(info, eog=True, exclude=[])
                 pre_whitener[picks_] = np.std(data[picks_])
         else:
-            pre_whitener, _ = compute_whitener(self.noise_cov, self.info)
+            pre_whitener, _ = compute_whitener(
+                self.noise_cov, self.info, picks=self.info.ch_names)
             assert data.shape[0] == pre_whitener.shape[1]
         self.pre_whitener_ = pre_whitener
 
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -827,16 +827,7 @@ class ICA(ContainsMixin):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
         start, stop = _check_start_stop(raw, start, stop)
-
-        picks = _picks_to_idx(
-            raw.info, self.ch_names, exclude=[], allow_empty=True)
-        if len(picks) != len(self.ch_names):
-            raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
-                               'fitted but %i channels supplied. \nPlease '
-                               'provide Raw compatible with '
-                               'ica.ch_names' % (len(self.ch_names),
-                                                 len(picks)))
-
+        picks = self._get_picks(raw)
         reject = 'omit' if reject_by_annotation else None
         data = raw.get_data(picks, start, stop, reject)
         return self._transform(data)
@@ -845,43 +836,41 @@ class ICA(ContainsMixin):
         """Aux method."""
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
-
-        picks = _picks_to_idx(
-            epochs.info, self.ch_names, exclude=[], allow_empty=True)
-        # special case where epochs come picked but fit was 'unpicked'.
-        if len(picks) != len(self.ch_names):
-            raise RuntimeError('Epochs don\'t match fitted data: %i channels '
-                               'fitted but %i channels supplied. \nPlease '
-                               'provide Epochs compatible with '
-                               'ica.ch_names' % (len(self.ch_names),
-                                                 len(picks)))
-
+        picks = self._get_picks(epochs)
         data = np.hstack(epochs.get_data()[:, picks])
         sources = self._transform(data)
-
         if not concatenate:
             # Put the data back in 3D
             sources = np.array(np.split(sources, len(epochs.events), 1))
-
         return sources
 
     def _transform_evoked(self, evoked):
         """Aux method."""
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
+        picks = self._get_picks(evoked)
+        return self._transform(evoked.data[picks])
 
+    def _get_picks(self, inst):
+        """Pick logic for _transform method."""
         picks = _picks_to_idx(
-            evoked.info, self.ch_names, exclude=[], allow_empty=True)
+            inst.info, self.ch_names, exclude=[], allow_empty=True)
         if len(picks) != len(self.ch_names):
-            raise RuntimeError('Evoked doesn\'t match fitted data: %i channels'
-                               ' fitted but %i channels supplied. \nPlease '
-                               'provide Evoked compatible with '
-                               'ica.ch_names' % (len(self.ch_names),
-                                                 len(picks)))
-
-        sources = self._transform(evoked.data[picks])
-
-        return sources
+            if isinstance(inst, BaseRaw):
+                kind, do = 'Raw', "doesn't"
+            elif isinstance(inst, BaseEpochs):
+                kind, do = 'Epochs', "don't"
+            elif isinstance(inst, Evoked):
+                kind, do = 'Evoked', "doesn't"
+            else:
+                raise ValueError('Data input must be of Raw, Epochs or Evoked '
+                                 'type')
+            raise RuntimeError("%s %s match fitted data: %i channels "
+                               "fitted but %i channels supplied. \nPlease "
+                               "provide %s compatible with ica.ch_names"
+                               % (kind, do, len(self.ch_names), len(picks),
+                                  kind))
+        return picks
 
     def get_components(self):
         """Get ICA topomap for components as numpy arrays.
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -830,7 +830,7 @@ class ICA(ContainsMixin):
         try:
             picks = _picks_to_idx(raw.info, self.ch_names, exclude=[])
             raise_ = False if len(picks) == len(self.ch_names) else True
-        except ValueError:
+        except RuntimeWarning:
             raise_ = True
             picks = list()
             for ch in self.ch_names:
@@ -858,7 +858,7 @@ class ICA(ContainsMixin):
             picks = _picks_to_idx(epochs.info, self.ch_names, exclude=[])
             # special case where epochs come picked but fit was 'unpicked'.
             raise_ = False if len(picks) == len(self.ch_names) else True
-        except ValueError:
+        except RuntimeWarning:
             raise_ = True
             picks = list()
             for ch in self.ch_names:
@@ -890,7 +890,7 @@ class ICA(ContainsMixin):
         try:
             picks = _picks_to_idx(evoked.info, self.ch_names, exclude=[])
             raise_ = False if len(picks) == len(self.ch_names) else True
-        except ValueError:
+        except RuntimeWarning:
             raise_ = True
             picks = list()
             for ch in self.ch_names:
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -59,7 +59,7 @@ from ..fixes import _get_args, _safe_svd
 from ..filter import filter_data
 from .bads import _find_outliers
 from .ctps_ import ctps
-from ..io.pick import pick_channels_regexp
+from ..io.pick import pick_channels_regexp, _picks_by_type
 
 __all__ = ('ICA', 'ica_find_ecg_events', 'ica_find_eog_events',
            'get_score_funcs', 'read_ica', 'read_ica_eeglab')
@@ -660,29 +660,14 @@ class ICA(ContainsMixin):
             # Scale (z-score) the data by channel type
             info = self.info
             pre_whitener = np.empty([len(data), 1])
-            for ch_type in _DATA_CH_TYPES_SPLIT + ('eog', "ref_meg"):
-                if _contains_ch_type(info, ch_type):
-                    if ch_type == 'seeg':
-                        picks_ = pick_types(info, seeg=True, exclude=[])
-                    elif ch_type == 'dbs':
-                        picks_ = pick_types(info, dbs=True, exclude=[])
-                    elif ch_type == 'ecog':
-                        picks_ = pick_types(info, ecog=True, exclude=[])
-                    elif ch_type == 'eeg':
-                        picks_ = pick_types(info, eeg=True, exclude=[])
-                    elif ch_type in ('mag', 'grad'):
-                        picks_ = pick_types(info, meg=ch_type, exclude=[])
-                    elif ch_type == 'eog':
-                        picks_ = pick_types(info, eog=True, exclude=[])
-                    elif ch_type in ('hbo', 'hbr'):
-                        picks_ = pick_types(info, fnirs=ch_type, exclude=[])
-                    elif ch_type == 'ref_meg':
-                        picks_ = pick_types(info, ref_meg=True, exclude=[])
-                    else:
-                        raise RuntimeError('Should not be reached.'
-                                           'Unsupported channel {}'
-                                           .format(ch_type))
-                    pre_whitener[picks_] = np.std(data[picks_])
+            for _, picks_ in _picks_by_type(info, ref_meg=False, exclude=[]):
+                pre_whitener[picks_] = np.std(data[picks_])
+            if _contains_ch_type(info, "ref_meg"):
+                picks_ = pick_types(info, ref_meg=True, exclude=[])
+                pre_whitener[picks_] = np.std(data[picks_])
+            if _contains_ch_type(info, "eog"):
+                picks_ = pick_types(info, eog=True, exclude=[])
+                pre_whitener[picks_] = np.std(data[picks_])
         else:
             pre_whitener, _ = compute_whitener(self.noise_cov, self.info)
             assert data.shape[0] == pre_whitener.shape[1]
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -663,26 +663,26 @@ class ICA(ContainsMixin):
             for ch_type in _DATA_CH_TYPES_SPLIT + ('eog', "ref_meg"):
                 if _contains_ch_type(info, ch_type):
                     if ch_type == 'seeg':
-                        this_picks = pick_types(info, meg=False, seeg=True)
+                        picks_ = pick_types(info, seeg=True, exclude=[])
                     elif ch_type == 'dbs':
-                        this_picks = pick_types(info, meg=False, dbs=True)
+                        picks_ = pick_types(info, dbs=True, exclude=[])
                     elif ch_type == 'ecog':
-                        this_picks = pick_types(info, meg=False, ecog=True)
+                        picks_ = pick_types(info, ecog=True, exclude=[])
                     elif ch_type == 'eeg':
-                        this_picks = pick_types(info, meg=False, eeg=True)
+                        picks_ = pick_types(info, eeg=True, exclude=[])
                     elif ch_type in ('mag', 'grad'):
-                        this_picks = pick_types(info, meg=ch_type)
+                        picks_ = pick_types(info, meg=ch_type, exclude=[])
                     elif ch_type == 'eog':
-                        this_picks = pick_types(info, meg=False, eog=True)
+                        picks_ = pick_types(info, eog=True, exclude=[])
                     elif ch_type in ('hbo', 'hbr'):
-                        this_picks = pick_types(info, meg=False, fnirs=ch_type)
+                        picks_ = pick_types(info, fnirs=ch_type, exclude=[])
                     elif ch_type == 'ref_meg':
-                        this_picks = pick_types(info, meg=False, ref_meg=True)
+                        picks_ = pick_types(info, ref_meg=True, exclude=[])
                     else:
                         raise RuntimeError('Should not be reached.'
                                            'Unsupported channel {}'
                                            .format(ch_type))
-                    pre_whitener[this_picks] = np.std(data[this_picks])
+                    pre_whitener[picks_] = np.std(data[picks_])
         else:
             pre_whitener, _ = compute_whitener(self.noise_cov, self.info)
             assert data.shape[0] == pre_whitener.shape[1]
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -830,8 +830,7 @@ class ICA(ContainsMixin):
 
         picks = _picks_to_idx(
             raw.info, self.ch_names, exclude=[], allow_empty=True)
-        raise_ = False if len(picks) == len(self.ch_names) else True
-        if raise_:
+        if len(picks) != len(self.ch_names):
             raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
                                'provide Raw compatible with '
@@ -850,8 +849,7 @@ class ICA(ContainsMixin):
         picks = _picks_to_idx(
             epochs.info, self.ch_names, exclude=[], allow_empty=True)
         # special case where epochs come picked but fit was 'unpicked'.
-        raise_ = False if len(picks) == len(self.ch_names) else True
-        if raise_:
+        if len(picks) != len(self.ch_names):
             raise RuntimeError('Epochs don\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
                                'provide Epochs compatible with '
@@ -874,8 +872,7 @@ class ICA(ContainsMixin):
 
         picks = _picks_to_idx(
             evoked.info, self.ch_names, exclude=[], allow_empty=True)
-        raise_ = False if len(picks) == len(self.ch_names) else True
-        if raise_:
+        if len(picks) != len(self.ch_names):
             raise RuntimeError('Evoked doesn\'t match fitted data: %i channels'
                                ' fitted but %i channels supplied. \nPlease '
                                'provide Evoked compatible with '
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -842,7 +842,7 @@ class ICA(ContainsMixin):
             raise RuntimeError('No fit available. Please fit ICA.')
         start, stop = _check_start_stop(raw, start, stop)
 
-        picks = pick_types(raw.info, include=self.ch_names, exclude='bads',
+        picks = pick_types(raw.info, include=self.ch_names, exclude=[],
                            meg=False, ref_meg=False)
         if len(picks) != len(self.ch_names):
             raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
@@ -860,7 +860,7 @@ class ICA(ContainsMixin):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        picks = pick_types(epochs.info, include=self.ch_names, exclude='bads',
+        picks = pick_types(epochs.info, include=self.ch_names, exclude=[],
                            meg=False, ref_meg=False)
         # special case where epochs come picked but fit was 'unpicked'.
         if len(picks) != len(self.ch_names):
@@ -884,7 +884,7 @@ class ICA(ContainsMixin):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        picks = pick_types(evoked.info, include=self.ch_names, exclude='bads',
+        picks = pick_types(evoked.info, include=self.ch_names, exclude=[],
                            meg=False, ref_meg=False)
 
         if len(picks) != len(self.ch_names):
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -827,17 +827,9 @@ class ICA(ContainsMixin):
             raise RuntimeError('No fit available. Please fit ICA.')
         start, stop = _check_start_stop(raw, start, stop)
 
-        try:
-            picks = _picks_to_idx(raw.info, self.ch_names, exclude=[])
-            raise_ = False if len(picks) == len(self.ch_names) else True
-        except RuntimeWarning:
-            raise_ = True
-            picks = list()
-            for ch in self.ch_names:
-                try:
-                    picks.extend(_picks_to_idx(raw.info, ch, exclude=[]))
-                except ValueError:
-                    pass
+        picks = _picks_to_idx(
+            raw.info, self.ch_names, exclude=[], allow_empty=True)
+        raise_ = False if len(picks) == len(self.ch_names) else True
         if raise_:
             raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
@@ -854,18 +846,10 @@ class ICA(ContainsMixin):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        try:
-            picks = _picks_to_idx(epochs.info, self.ch_names, exclude=[])
-            # special case where epochs come picked but fit was 'unpicked'.
-            raise_ = False if len(picks) == len(self.ch_names) else True
-        except RuntimeWarning:
-            raise_ = True
-            picks = list()
-            for ch in self.ch_names:
-                try:
-                    picks.extend(_picks_to_idx(epochs.info, ch, exclude=[]))
-                except ValueError:
-                    pass
+        picks = _picks_to_idx(
+            epochs.info, self.ch_names, exclude=[], allow_empty=True)
+        # special case where epochs come picked but fit was 'unpicked'.
+        raise_ = False if len(picks) == len(self.ch_names) else True
         if raise_:
             raise RuntimeError('Epochs don\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
@@ -887,17 +871,8 @@ class ICA(ContainsMixin):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        try:
-            picks = _picks_to_idx(evoked.info, self.ch_names, exclude=[])
-            raise_ = False if len(picks) == len(self.ch_names) else True
-        except RuntimeWarning:
-            raise_ = True
-            picks = list()
-            for ch in self.ch_names:
-                try:
-                    picks.extend(_picks_to_idx(evoked.info, ch, exclude=[]))
-                except ValueError:
-                    pass
+        picks = _picks_to_idx(evoked.info, self.ch_names, exclude=[])
+        raise_ = False if len(picks) == len(self.ch_names) else True
         if raise_:
             raise RuntimeError('Evoked doesn\'t match fitted data: %i channels'
                                ' fitted but %i channels supplied. \nPlease '
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -842,9 +842,18 @@ class ICA(ContainsMixin):
             raise RuntimeError('No fit available. Please fit ICA.')
         start, stop = _check_start_stop(raw, start, stop)
 
-        picks = pick_types(raw.info, include=self.ch_names, exclude=[],
-                           meg=False, ref_meg=False)
-        if len(picks) != len(self.ch_names):
+        try:
+            picks = _picks_to_idx(raw.info, self.ch_names, exclude=[])
+            raise_ = False if len(picks) == len(self.ch_names) else True
+        except ValueError:
+            raise_ = True
+            picks = list()
+            for ch in self.ch_names:
+                try:
+                    picks.extend(_picks_to_idx(raw.info, ch, exclude=[]))
+                except ValueError:
+                    pass
+        if raise_:
             raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
                                'provide Raw compatible with '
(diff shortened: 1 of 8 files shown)


mscheltienne:
New traceback following the fix of `_compute_pre_whitener()`:

```
Fitting ICA to data using 59 channels (please be patient, this may take a while)
Selecting by explained variance: 23 components

  File "/Users/scheltie/Downloads/test.py", line 12, in <module>
    ica.fit(raw, picks=picks, reject_by_annotation=True)

  File "<decorator-gen-443>", line 24, in fit

  File "/Users/scheltie/Documents/git/mne-python/mne/preprocessing/ica.py", line 591, in fit
    var = _ica_explained_variance(self, inst)

  File "/Users/scheltie/Documents/git/mne-python/mne/preprocessing/ica.py", line 2182, in _ica_explained_variance
    source_data = _get_inst_data(ica.get_sources(inst))

  File "/Users/scheltie/Documents/git/mne-python/mne/preprocessing/ica.py", line 944, in get_sources
    sources = self._sources_as_raw(inst, add_channels, start, stop)

  File "/Users/scheltie/Documents/git/mne-python/mne/preprocessing/ica.py", line 962, in _sources_as_raw
    data_ = self._transform_raw(raw, start=start, stop=stop)

  File "/Users/scheltie/Documents/git/mne-python/mne/preprocessing/ica.py", line 848, in _transform_raw
    raise RuntimeError('Raw doesn\'t match fitted data: %i channels '

RuntimeError: Raw doesn't match fitted data: 59 channels fitted but 56 channels supplied. 
Please provide Raw compatible with ica.ch_names
```

mscheltienne:
The last commit should be good to go. Waiting for review.

Additional TODO changes:
- Add a similar test for epochs and evoked instances
- Edit changelog?

mscheltienne:
Failing tests in vizualization:

```
    raw.info['bads'] = ['MEG 0113']
    with pytest.raises(RuntimeError, match="Raw doesn't match fitted data"):
        ica.plot_sources(inst=raw)
    epochs.info['bads'] = ['MEG 0113']
    with pytest.raises(RuntimeError, match="Epochs don't match fitted data"):
        ica.plot_sources(inst=epochs)
```

Is it actually not a mistake that those tests were successful previously? In `mne.viz.ica.py` in `plot_ica_sources`, if the argument `picks` is set to None, the picks are handled with:

`picks = _picks_to_idx(ica.n_components_, picks, 'all')`

Thus picking `'all'` channels, bads included.

mscheltienne:
Closed by mistake while changing computer.. deleted branch on remote by mistake.

mne/preprocessing/ica.py
```diff
@@ -884,7 +884,7 @@ def _transform_evoked(self, evoked):
         if not hasattr(self, 'mixing_matrix_'):
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        picks = pick_types(evoked.info, include=self.ch_names, exclude='bads',
+        picks = pick_types(evoked.info, include=self.ch_names, exclude=[],
```
agramfort, line 887: but now if I have marked bads and I don't pass picks to ICA.fit will it still ignore the bad channels? It should.

mscheltienne:
I did some more testing to make sure I didn't miss anything. You can find the complete result/script below. 
To sum up, I tested that the picks were correct by stopping line 557 and checking the return of `_picks_to_idx` and line 661 checking the content of `info`. 

Then I started adding `ica.get_sources`, as if I recall correctly that is where the test on `viz` is failing.
I feel like there is a mistake there; as when modifying the list of bad channels after fitting the ICA and then retrieving the sources with this modified instance, there is no error raised.

```
#%% Imports
import mne
import numpy as np

from mne.preprocessing import ICA


#%% Create RAW
FS = 160
rng = np.random.default_rng(12345)
data = rng.integers(low=1, high=100, size=(8, FS * 10))
ch_names = [f'EEG{i+1}' for i in range(4)] + \
           ['STI1'] + ['GRAD1', 'GRAD2', 'MAG1']
ch_types = ['eeg'] * 4 + ['stim'] + ['grad'] * 2 + ['mag']
data[4, :] = np.zeros(shape=(data[4, :].shape))
data[4, rng.integers(low=0, high=1600)] = 1
data = data.astype(np.float32)

info = mne.create_info(ch_names, ch_types=ch_types, sfreq=FS)
raw.filter(l_freq=1., h_freq=60., picks=['eeg', 'meg'])


#%% ICA default

raw_ = raw.copy()
ica = ICA(n_components=0.99, method='fastica', max_iter='auto')
ica.fit(raw_, picks=None)

# picks = array([0, 1, 2, 3, 5, 6, 7])
# info
# <Info | 7 non-empty values
#  bads: []
#  ch_names: EEG1, EEG2, EEG3, EEG4, GRAD1, GRAD2, MAG1
#  chs: 4 EEG, 2 GRAD, 1 MAG
#  custom_ref_applied: False
#  highpass: 1.0 Hz
#  lowpass: 60.0 Hz
#  meas_date: unspecified
#  nchan: 7
#  projs: []
#  sfreq: 160.0 Hz
# >

ica.get_sources(raw_)
# <RawArray | 7 x 1600 (10.0 s), ~99 kB, data loaded>


#%% ICA default with a marked bad channel

raw_ = raw.copy()
raw_.info['bads'] = ['EEG2']
ica = ICA(n_components=0.99, method='fastica', max_iter='auto')
ica.fit(raw_, picks=None)

# picks = array([0, 2, 3, 5, 6, 7])
# info
# <Info | 7 non-empty values
#  bads: []
#  ch_names: EEG1, EEG3, EEG4, GRAD1, GRAD2, MAG1
#  chs: 3 EEG, 2 GRAD, 1 MAG
#  custom_ref_applied: False
#  highpass: 1.0 Hz
#  lowpass: 60.0 Hz
#  meas_date: unspecified
#  nchan: 6
#  projs: []
#  sfreq: 160.0 Hz
# >

ica.get_sources(raw_)
# <RawArray | 6 x 1600 (10.0 s), ~86 kB, data loaded>
raw_.info['bads'] = ['EEG1', 'EEG2']
ica.get_sources(raw_)
# <RawArray | 6 x 1600 (10.0 s), ~86 kB, data loaded>


#%% ICA including the marked bad channel

raw_ = raw.copy()
raw_.info['bads'] = ['EEG2']
picks = mne.pick_types(raw_.info, meg=True, eeg=True, exclude=[])
ica = ICA(n_components=0.99, method='fastica', max_iter='auto')
ica.fit(raw_, picks=picks)

# picks = array([0, 1, 2, 3, 5, 6, 7])
# info
# <Info | 8 non-empty values
#  bads: 1 items (EEG2)
#  ch_names: EEG1, EEG2, EEG3, EEG4, GRAD1, GRAD2, MAG1
#  chs: 4 EEG, 2 GRAD, 1 MAG
#  custom_ref_applied: False
#  highpass: 1.0 Hz
#  lowpass: 60.0 Hz
#  meas_date: unspecified
#  nchan: 7
#  projs: []
#  sfreq: 160.0 Hz
# >

ica.get_sources(raw_)
# <RawArray | 7 x 1600 (10.0 s), ~99 kB, data loaded>
raw_.info['bads'] = ['EEG1', 'EEG2']
ica.get_sources(raw_)
# <RawArray | 7 x 1600 (10.0 s), ~99 kB, data loaded>
```

mne/preprocessing/ica.py
mscheltienne: It does. See below for the additional testing.

mscheltienne:
And to build upon my last comment, let's looks at the method `_transform_raw`:

```
        picks = pick_types(raw.info, include=self.ch_names, exclude='bads',
                           meg=False, ref_meg=False)
        if len(picks) != len(self.ch_names):
            raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
                               'fitted but %i channels supplied. \nPlease '
                               'provide Raw compatible with '
                               'ica.ch_names' % (len(self.ch_names),
                                                 len(picks)))
```

This is the error that should be raised and is not anymore if `exclude=[]` is set. This is the failed test in `viz`.
Now, if the ICA is fit with `picks` specified and including bad channels; then this will always raise as `ica.ch_names` will include the picked bad channels and `picks` in this method will not; leading to a mismatch. 
This is why I changed the `exclude` argument from `'bads'` to `[]`.

However, now if the instance is modified and e.g. an additional bad channel is added; this code snippet will not raise anymore as `picks` will include the bad channels and there will not be any mismatch.
This is what is illustrated by the calls to `ica.get_sources` in the comment above.

**Proposition:** This test between this hardcoded `picks` and the provided `raw` doesn't make much sense. Instead, how about testing that the provided `raw` includes the picked channel used to fit the ICA?

mscheltienne:
The last commit covers the proposal mentioned above. 

To avoid changing the error message, I had to create the variable `picks` even when a `ValueError` is raised by `_picks_to_idx` because some channels in the selection are missing from the instance. Maybe someone thinks of a cleaner solution/utils function I am not aware of.

I changed the (failing) test in `mne.viz.test.test_ica.py` to drop the channel instead of marking it as bad. 
The rationale is:
- If you drop the channel, the instance is modified and the channels do not match anymore the channels on which the ICA was fit.
- If you mark a channel as bad, the instance is not modified and the available channels still match the channels used to fit the ICA.

mne/preprocessing/ica.py
```diff
@@ -663,26 +663,26 @@ def _compute_pre_whitener(self, data):
             for ch_type in _DATA_CH_TYPES_SPLIT + ('eog', "ref_meg"):
                 if _contains_ch_type(info, ch_type):
```
agramfort, line 664: maybe there is a simpler way. We have access to the picked channels (channels on which ICA will be applied). So here for whitening we need to get the picked indices by channel types and select the picks_ from the "picked channels". Does it make sense?

mscheltienne: Let's see if I got it right. The channels on which the ICA will be applied are accessible in `._compute_pre_whitener` in `info = self.info`, info instance created by selecting the channels passed in arg `picks` in the method `.fit`. So now, for each channel type contained in this info instance, we need to compute the associated`pre_whitener`; essentially replacing this long list of if/else statements with a compact cleaner implementation. If that is what you want, i.e. to simplify the code and get rid of the long list of if/else statements, then sure I can look into it. I didn't do it to minimize the amount of code changed. If that's not what you meant, then I'm clueless.

agramfort:
you got me right. This would kill the long list of if and you would just
need to call the right function to get the channel types of all the ICA
channels.

>

mscheltienne:
Not sure why the test failed; I will have a look when I have time. Locally, the ICA tests all passed..

Screenshot 2021-09-08 at 16 45 27

mscheltienne:
@larsoner Thanks, I had no idea the workflow was not using the test from the PR branch, but merging with the latest updates.

mne/preprocessing/ica.py
```diff
@@ -830,7 +830,7 @@ def _transform_raw(self, raw, start, stop, reject_by_annotation=False):
         try:
             picks = _picks_to_idx(raw.info, self.ch_names, exclude=[])
             raise_ = False if len(picks) == len(self.ch_names) else True
-        except ValueError:
+        except RuntimeWarning:
```
larsoner: I don't think this is the right fix. The only reason this works (I think) is because we have warnings-as-errors turn on in our tests. For example:
```
>>> import warnings
>>> try:
...     warnings.warn('foo')
... except UserWarning:
...     print('caught')
... 
<stdin>:2: UserWarning: foo
```
i.e., in regular use this pattern will not work. I think the correct/better fix is to leave this, but in the places in the *tests* that were failing, you add an additional context manager like:
```
with pytest.warns(RuntimeWarning, match='not found'):
```

mscheltienne: Oh right... my bad.. I thought it was an error that was getting raised.. Then yes, in the method `_transform_xxxxx` I will use `_picks_to_idx(raw.info, self.ch_names, exclude=[], allow_empty=True)` which might log the warning if you look for a channel not present in the info, but will return every time an array to compare with `self.ch_names`, the picks used to fit the ICA.

mscheltienne: Is there a better syntax that I don't know than the nesting of content managers?

mne/preprocessing/ica.py
```diff
@@ -660,29 +660,14 @@ def _compute_pre_whitener(self, data):
...
-                                           .format(ch_type))
-                    pre_whitener[this_picks] = np.std(data[this_picks])
+            for _, picks_ in _picks_by_type(info, ref_meg=False, exclude=[]):
+                pre_whitener[picks_] = np.std(data[picks_])
+            if _contains_ch_type(info, "ref_meg"):
+                picks_ = pick_types(info, ref_meg=True, exclude=[])
+                pre_whitener[picks_] = np.std(data[picks_])
+            if _contains_ch_type(info, "eog"):
+                picks_ = pick_types(info, eog=True, exclude=[])
+                pre_whitener[picks_] = np.std(data[picks_])
```
agramfort, line 670: why do you need to special case this? This suggests there is something inconsistent somewhere.

mscheltienne: First of all, they were already both special cases before. 
Second, the problem is, as I started pointing out in a comment above, in `_picks_by_type` which picks only data channels (this is not a bug, this is what this function is intended for).

Example code:
```
import mne

ch_names = [f'EEG{i+1}' for i in range(4)] + \
           ['EOG1'] + ['GRAD1', 'GRAD2', 'MAG1'] + ['REF_MEG']
ch_types = ['eeg'] * 4 + ['eog'] + ['grad'] * 2 + ['mag'] + ['ref_meg']
info = mne.create_info(ch_names, ch_types=ch_types, sfreq=1000)
```

Missing the EOG channel:
```
_picks_by_type(info)
Out: [('mag', array([7])), ('grad', array([5, 6])), ('eeg', array([0, 1, 2, 3]))]
```

Adding the MEG REF as a MAG: **Bug?**
```
_picks_by_type(info, ref_meg=True)
Out: [('mag', array([7, 8])), ('grad', array([5, 6])), ('eeg', array([0, 1, 2, 3]))]
```

mscheltienne:
For the failed test, bug on the CI?

`ModuleNotFoundError: No module named 'transforms3d'`

larsoner:
Ignore the `conda` and `Notebook` CI runs for now, they'll be fixed once https://github.com/conda-forge/pyvista-feedstock/pull/51 is merged

mne/preprocessing/tests/test_ica.py
```diff
@@ -705,7 +705,8 @@ def test_ica_additional(method, tmpdir, short_raw_epochs):
     assert_equal(ica_sorted.labels_, dict(blink=[3], think=[2]))
 
     # epochs extraction from raw fit
-    pytest.raises(RuntimeError, ica.get_sources, epochs)
+    with pytest.warns(RuntimeWarning, match='could not be picked'):
+        pytest.raises(RuntimeError, ica.get_sources, epochs)
```
larsoner, line 709: `pytest.raises` is a context manager, so better to do:
```
with pytest.warns(...):
    with pytest.raises(RuntimeError, match='some matching regex'):
         ...
```
if you want to avoid so much nesting you could do:
```
warn_ctx = pytest.warns(RuntimeWarning, match='...')
err_ctx = pytest.raises(RuntimeError, match='...'):
with warn_ctx, err_ctx:
    ...
```

mne/preprocessing/ica.py
larsoner: >> This suggests there is something inconsistent somewhere.

> Second, the problem is, as I started pointing out in a comment above, in _picks_by_type which picks only data channels (this is not a bug, this is what this function is intended for).

IIRC the whitening function does not usually operate on EOG or ref_mag, but for ICA we want to include them. So indeed I agree what you have here is one option.

However, I'm not sure `compute_whitener` can *only* operate on these. Can you pass all `picks` to it as `compute_whitener(..., picks=all_data_plus_eog_and_ref_mag_picks)` and have it do the right thing? It might work...

mscheltienne: I don't follow. The changes here I limited to the case where the noise covariance matrix is not provided, and thus the case where `compute_whitener` is not called.
Note that I didn't test the case where the noise covariance matrix is provided. By the code in `compute_whitener`, I guess it will fail if you include bad channels as picks are handled with: `_picks_to_idx(info, picks, with_ref_meg=False)`.

mscheltienne:
Example with a noise covariance matrix provided:

```
import mne
from pathlib import Path

data_folder = Path(mne.datasets.sample.data_path())
raw_file = data_folder / 'MEG' / 'sample' / 'sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(raw_file, preload=True)
raw.pick_types(eeg=True)

raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015']
picks = mne.pick_types(raw.info, eeg=True, exclude=[])

cov = mne.compute_raw_covariance(raw, picks=picks)

ica = mne.preprocessing.ICA(n_components=0.99, max_iter='auto', noise_cov=cov)
ica.fit(raw, picks=picks, reject_by_annotation=True)
```

```
  File "C:\Users\Mathieu\AppData\Local\Temp/ipykernel_7708/3375348688.py", line 2, in <module>
    ica.fit(raw, picks=picks, reject_by_annotation=True)

  File "<decorator-gen-443>", line 24, in fit

  File "c:\users\mathieu\documents\git\mne-python\mne\preprocessing\ica.py", line 584, in fit
    self._fit_raw(inst, picks, start, stop, decim, reject, flat,

  File "c:\users\mathieu\documents\git\mne-python\mne\preprocessing\ica.py", line 628, in _fit_raw
    self._fit(data, 'raw')

  File "c:\users\mathieu\documents\git\mne-python\mne\preprocessing\ica.py", line 702, in _fit
    self._compute_pre_whitener(data)

  File "c:\users\mathieu\documents\git\mne-python\mne\preprocessing\ica.py", line 673, in _compute_pre_whitener
    assert data.shape[0] == pre_whitener.shape[1]

AssertionError
```

I'll add it as test case and try to fix it.

mscheltienne:
Fix for the case where a noise covariance matrix is provided is:
- In `ICA._compute_pre_whitener` change line 672 to `pre_whitener, _ = compute_whitener(self.noise_cov, self.info, picks=self.info.ch_names)`. As for before, the `info` instance is created by selecting the info/channels provided by picks. Thus, providing the list of channel names explicitly to `compute_whitener` includes all the picked channels; even if they were marked as bads. 

- #9727 must be fixed. I'll fix it in this PR since the bug is similar and was discovered thanks to this PR.

mne/preprocessing/ica.py
larsoner: > The changes here I limited to the case where the noise covariance matrix is not provided

Oh right, never mind then!

mne/rank.py
```diff
@@ -344,7 +344,7 @@ def compute_rank(inst, rank=None, scalings=None, info=None, tol='auto',
 
     simple_info = _simplify_info(info)
     picks_list = _picks_by_type(info, meg_combined=True, ref_meg=False,
-                                exclude='bads')
+                                exclude=[])
```
larsoner, line 347: In most cases I think we don't want to include bad channels actually, so this seems dangerous to me, assuming it does what I'm assuming it does (and I might be wrong!).

In a practical case, I think for 306-channel neuromag data with no projs for example you should have both on `main` and after this PR:
```
raw = mne.io.read_raw_fif(...)
raw.info['bads'] = []
raw.pick('meg')
assert len(raw.ch_names) == 306
assert compute_rank(raw) == dict(mag=102, grad=204)
cov = compute_raw_covariance(raw)
assert compute_rank(cov) == dict(mag=102, grad=204)
whitener = compute_whitener(cov, pca=True)[0]
assert whitener.shape[0] == 306
```
And then if you mark one channel bad, rank decreases by one:
```
# mark one mag bad, rank decreases by one
raw.info['bads'] = ['MEG 0111']
assert compute_rank(raw) == dict(mag=101, grad=204)
cov = compute_raw_covariance(raw)
assert compute_rank(cov) == dict(mag=101, grad=204)
whitener = compute_whitener(cov, pca=True)[0]
assert whitener.shape[0] == 305
```
Is it still the case after this PR? Either way, please add this sort of thing as an explicit test in `test_rank.py` or (maybe better) `test_cov.py`. 

If the above test is (still) the case and these ranks don't decrease when you pass an explicit `picks=np.arange(306)` for example, then that seems reasonable to me. In other words, I guess this could be an addition to the newly proposed test:
```
# explicit picks restore the old rank
whitener = compute_whitener(cov, pca=True, picks=np.arange(306))[0]
assert whitener.shape[0] == 306
```
Maybe the correct fix is to provide a `picks` argument to `compute_rank` that behaves the same way that it's supposed to in `compute_whitener`, i.e., `'bads'` are ignored by default (`picks_good_data_noref` just like `compute_whitener`) but if you pass picks explicitly then they can be included? That way the above test could also then have after marking one channel bad:
```
assert compute_rank(raw, picks=np.arange(306)) == dict(mag=102, grad=204)
```

mne/preprocessing/ica.py
mscheltienne: @larsoner In the end.. following your comment.. I did add and fixed as well the part with the covariance matrix provided.

mne/rank.py
mscheltienne: I agree that this PR should not change the default behavior, i.e. excluding bads. The goal of the PR is to impact only the behavior when explicit picks (bads or not) are provided. From just looking at the code, and the testing I've done until now, the modification should work as intended and should not have impacted the default behavior. However, I didn't have time today to check if the tests were already covering this correctly. I will check tomorrow, and add the additional tests if needed.

mscheltienne:
Additional testing. @larsoner As you spotted; the default b behavior with `picks=None` was broken for instances of `Raw` or `Epochs`. This time it should be good. I will check later which test is missing from the test files and add them accordingly. 

I avoided passing an additional `picks` argument to `compute_rank` as it might impact other functions/code. 

Testing for `compute_rank`:

```
#%% Load raw
import mne
from pathlib import Path

data_folder = Path(mne.datasets.sample.data_path())
raw_file = data_folder / 'MEG' / 'sample' / 'sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(raw_file, preload=True)
raw.pick_types(meg=True, eeg=True)

#%% Covariance rank default without bads
raw.info['bads'] = list()
noise_cov = mne.compute_raw_covariance(raw, picks=None)
rank = mne.rank.compute_rank(noise_cov, info=raw.info)
assert rank['meg'] == 302
assert rank['eeg'] == 58
assert raw.info['bads'] == []

#%% Covariance rank default with bads
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
noise_cov = mne.compute_raw_covariance(raw, picks=None)
rank = mne.rank.compute_rank(noise_cov, info=raw.info)
assert rank['meg'] == 301
assert rank['eeg'] == 55
assert raw.info['bads'] == ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']

#%% Covariance rank pick with bads
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
picks = mne.pick_types(raw.info, meg=True, eeg=True, exclude=[])
noise_cov = mne.compute_raw_covariance(raw, picks=picks)
rank = mne.rank.compute_rank(noise_cov, info=raw.info)
assert rank['meg'] == 302
assert rank['eeg'] == 58
assert raw.info['bads'] == ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']

#%% Raw rank without bads
raw.info['bads'] = list()
rank = mne.rank.compute_rank(raw)
assert rank['meg'] == 302
assert rank['eeg'] == 58

#%% Raw rank with bads
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
rank = mne.rank.compute_rank(raw)
assert rank['meg'] == 301
assert rank['eeg'] == 55
```

Testing for `Covariance` and `compute_whitener`:

```
#%% Load raw
import mne
from pathlib import Path

data_folder = Path(mne.datasets.sample.data_path())
raw_file = data_folder / 'MEG' / 'sample' / 'sample_audvis_filt-0-40_raw.fif'
raw = mne.io.read_raw_fif(raw_file, preload=True)
raw.pick_types(meg=True, eeg=True)

#%% Covariance default without bads
raw.info['bads'] = list()
noise_cov = mne.compute_raw_covariance(raw, picks=None)
assert len(noise_cov['names']) == len(raw.ch_names)
assert noise_cov['data'].shape == (364, 364)

whitener = mne.cov.compute_whitener(noise_cov, raw.info, picks=None)
assert whitener[0].shape == (364, 364)
assert whitener[1] == raw.ch_names

#%% Covariance default with bads
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
noise_cov_with_bads = mne.compute_raw_covariance(raw, picks=None)
assert len(noise_cov_with_bads['names']) == len(raw.ch_names) - 4
assert noise_cov_with_bads['data'].shape == (360, 360)

assert (noise_cov['data'] == noise_cov_with_bads_inc['data']).all()
idx = [k for k, ch in enumerate(raw.ch_names) if ch not in raw.info['bads']]
noise_cov_data = noise_cov['data'][idx, :][:, idx]
assert (noise_cov_data == noise_cov_with_bads['data']).all()

whitener = mne.cov.compute_whitener(noise_cov_with_bads, raw.info, picks=None)
assert whitener[0].shape == (360, 360)
assert whitener[1] == [ch for ch in raw.ch_names if ch not in raw.info['bads']]

#%% Covariance pick with bads
raw.info['bads'] = ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
picks = mne.pick_types(raw.info, meg=True, eeg=True, exclude=[])
noise_cov_with_bads_inc = mne.compute_raw_covariance(raw, picks=picks)
assert len(noise_cov_with_bads_inc['names']) == len(raw.ch_names)
assert noise_cov_with_bads_inc['data'].shape == (364, 364)

whitener = mne.cov.compute_whitener(
    noise_cov_with_bads_inc, raw.info, picks=picks)
assert whitener[0].shape == (364, 364)
assert whitener[1] == raw.ch_names
```

agramfort (approved):
@mscheltienne can you add an entry in the latest.inc file to document the bug fix?

🙏

mscheltienne:
I added the missing tests and the entry to latest.inc. Should be ready to merge.

mscheltienne:
Failed tests:

```
FAILED mne/tests/test_cov.py::test_compute_whitener[True-True] - assert False
FAILED mne/tests/test_cov.py::test_compute_whitener[True-False] - assert False
FAILED mne/tests/test_cov.py::test_compute_whitener[white-True] - assert False
FAILED mne/tests/test_cov.py::test_compute_whitener[white-False] - assert False
FAILED mne/tests/test_cov.py::test_compute_whitener[False-True] - assert False
FAILED mne/tests/test_cov.py::test_compute_whitener[False-False] - assert False
```
I'm not getting any fail/traceback with `test_cov.py` locally. Any idea?

mne/tests/test_cov.py
```diff
@@ -75,6 +76,24 @@ def test_compute_whitener(proj, pca):
         assert pca is False
         assert_allclose(round_trip, np.eye(n_channels), atol=0.05)
 
+    raw.info['bads'] = [raw.ch_names[0]]
+    picks = pick_types(raw.info, meg=True, eeg=True, exclude=[])
+    with pytest.warns(RuntimeWarning, match='Too few samples'):
+        cov2 = compute_raw_covariance(raw, picks=picks)
+        cov3 = compute_raw_covariance(raw, picks=None)
+    assert (cov2['data'][1:, 1:] == cov3['data']).all()
```
larsoner, line 84: It is better to use `assert_array_equal` here, it will give a more informative error
```suggestion
    assert_array_equal(cov2['data'][1:, 1:], cov3['data'])
```

mne/tests/test_cov.py
```diff
@@ -75,6 +76,24 @@ def test_compute_whitener(proj, pca):
...
+    W2, _, C2 = compute_whitener(cov2, raw.info, pca=pca, return_colorer=True,
+                                 picks=picks, verbose='error')
+    W3, _, C3 = compute_whitener(cov3, raw.info, pca=pca, return_colorer=True,
+                                 picks=None, verbose='error')
+    assert (W == W2).all()
+    assert (C == C2).all()
```
larsoner, line 90: Same with these
```suggestion
    assert_array_equal(W, W2)
    assert_array_equal(C, C2)
```

mne/tests/test_rank.py
```diff
@@ -261,3 +261,42 @@ def test_maxfilter_get_rank(n_proj, fname, rank_orig, meg, tol_kind, tol):
...
+    assert raw.info['bads'] == ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
+
+    # Explicit picks
+    picks = pick_types(raw.info, meg=True, eeg=True, exclude=[])
+    noise_cov_2 = compute_raw_covariance(raw, picks=picks)
+    rank = compute_rank(noise_cov_2, info=raw.info)
+    assert rank == dict(meg=303, eeg=60)
+    assert raw.info['bads'] == ['EEG 002', 'EEG 012', 'EEG 015', 'MEG 0122']
+
+    assert (noise_cov_1['data'] == noise_cov_2['data']).all()
```
larsoner, line 292:
```suggestion
    assert_array_equal(noise_cov_1['data'], noise_cov_2['data'])
```

mne/tests/test_cov.py
```diff
@@ -75,6 +76,24 @@ def test_compute_whitener(proj, pca):
         assert pca is False
         assert_allclose(round_trip, np.eye(n_channels), atol=0.05)
 
+    raw.info['bads'] = [raw.ch_names[0]]
+    picks = pick_types(raw.info, meg=True, eeg=True, exclude=[])
+    with pytest.warns(RuntimeWarning, match='Too few samples'):
+        cov2 = compute_raw_covariance(raw, picks=picks)
+        cov3 = compute_raw_covariance(raw, picks=None)
+    assert_array_equal(cov2['data'][1:, 1:], cov3['data'])
```
larsoner, line 84: With this `assert_array_equal` we should get some handle on how many entries are actually wrong. It could be a numerical imprecision issue, in which case `assert_allclose` should fix it. For example, accumulators using SIMD instructions could differ slightly and result in differences around epsilon, and `assert_allclose`'s `rtol` should take care of this (as long as the values are not zero-like, in which case we'd have to set `atol` properly -- but hopefully this is not the issue.)

Let's first see what the failing Azure CI says here, then hopefully committing this will fix the issue:

```suggestion
    assert_allclose(cov2['data'][1:, 1:], cov3['data'])
```

mne/preprocessing/ica.py
```diff
@@ -842,9 +828,10 @@ def _transform_raw(self, raw, start, stop, reject_by_annotation=False):
             raise RuntimeError('No fit available. Please fit ICA.')
         start, stop = _check_start_stop(raw, start, stop)
 
-        picks = pick_types(raw.info, include=self.ch_names, exclude='bads',
-                           meg=False, ref_meg=False)
-        if len(picks) != len(self.ch_names):
+        picks = _picks_to_idx(
+            raw.info, self.ch_names, exclude=[], allow_empty=True)
+        raise_ = False if len(picks) == len(self.ch_names) else True
+        if raise_:
```
larsoner, line 834: I think it was simpler before... I'm assuming you have this `raise_` just for debugging, or it's cruft from when this was a more complicated conditional or something?
```suggestion
        if len(picks) != len(self.ch_names):
```

mne/preprocessing/ica.py
```diff
@@ -884,10 +872,10 @@ def _transform_evoked(self, evoked):
...
             raise RuntimeError('No fit available. Please fit ICA.')
 
-        picks = pick_types(evoked.info, include=self.ch_names, exclude='bads',
-                           meg=False, ref_meg=False)
-
-        if len(picks) != len(self.ch_names):
+        picks = _picks_to_idx(
+            evoked.info, self.ch_names, exclude=[], allow_empty=True)
+        raise_ = False if len(picks) == len(self.ch_names) else True
+        if raise_:
```
larsoner, line 878:
```suggestion
        if len(picks) != len(self.ch_names):
```

mne/preprocessing/ica.py
```diff
@@ -842,9 +828,10 @@ def _transform_raw(self, raw, start, stop, reject_by_annotation=False):
...
-        picks = pick_types(raw.info, include=self.ch_names, exclude='bads',
-                           meg=False, ref_meg=False)
-        if len(picks) != len(self.ch_names):
+        picks = _picks_to_idx(
+            raw.info, self.ch_names, exclude=[], allow_empty=True)
+        raise_ = False if len(picks) == len(self.ch_names) else True
+        if raise_:
             raise RuntimeError('Raw doesn\'t match fitted data: %i channels '
                                'fitted but %i channels supplied. \nPlease '
                                'provide Raw compatible with '
```
larsoner, line 837: Looking at this code duplication (x3: raw, epochs, evoked), it's probably worth adding a `def _get_picks(self, inst, kind):` that has this picks, check, and raise logic and returns the picks to use from `inst` of kind `kind`.

larsoner (commented):
Other than a couple of minor remaining suggestions and the `assert_allclose` we might need to use, LGTM and a very nice simplification @mscheltienne !

mne/tests/test_cov.py
mscheltienne: I was guessing the problem was in those array equality assertions, but I didn't know about both of those convenient functions. Let's see the result, thanks!

larsoner: Quoth GitHub Actions:
```
Max relative difference: 1.93161015e-12
```
So this should indeed be fixed by assert_allclose, which defaults to `rtol=1e-7`. Let's find out!

larsoner:
@mscheltienne feel free to pull the commits I made using the GitHub UI, then push a commit to take care of my last few nitpicks if you want / agree with them

mne/preprocessing/ica.py
mscheltienne: Done. I noticed it too, but didn't want to bother at the time with the different error messages depending on the instance.

mne/preprocessing/ica.py
mscheltienne: Good catch, it was a simplification at the time; I think before I merged the changes to `_picks_to_idx` from main. Completely useless now.

larsoner:
Hopefully the `pip-pre` error will be fixed by https://github.com/mne-tools/mne-python/pull/9731, let's ignore it here

larsoner:
Argh I missed that one :(

I'll take a look... I'm guessing/hoping it's the last eigenvector corresponding to a zero eigenvalue in a `pca=False` case or something.

larsoner:
I put a fix in #9731

PR #9721, opened 2021-09-07
MRG, BUG: Fix hemi

### LINKED ISSUE #9720: Is the stc.plot hemi parameter broken?
#### Describe the bug
For my own code plotting source estimates only the right hemisphere is no showing, even if I set `hemi="both`

#### Steps to reproduce
I think this may also be evident in the examples like

https://mne.tools/dev/auto_tutorials/inverse/40_mne_fixed_free.html#fixed-orientation

which had code
```python
snr = 3.0
lambda2 = 1.0 / snr ** 2
kwargs = dict(initial_time=0.08, hemi='both', subjects_dir=subjects_dir,
              size=(600, 600))

stc = abs(apply_inverse(evoked, inv, lambda2, 'MNE', verbose=True))
brain = stc.plot(figure=1, **kwargs)
brain.add_text(0.1, 0.9, 'MNE', 'title', font_size=14)
```

larsoner:
Closes #9720

Fixes regression introduced by #9688 (specifically on [this line](https://github.com/mne-tools/mne-python/pull/9688/files#diff-4e3d2d919a60057d535ba733330d5b8338aa491d6392b3d1c4a1a3de7122e17eR1930)), adds a test that would have caught it, touches an example that is broken on `main`, and fixes a persistent `linkcheck` false alarm.


agramfort (approved):
Merge when green

drammock:
tx @larsoner!

PR #9723, opened 2021-09-09
update funders

drammock:
adds the new CZI award and updates our GSoC counter.  (question: does the GSoC counter count *years* or *students*?  I assumed it was years, if it's students than the change should be 6 -> 9 instead of 6 -> 7)


larsoner:
It's years, and FWIW we usually have 2-3 students each year. I could go back and figure out how many total students if we want to add that information

drammock:
> I could go back and figure out how many total students if we want to add that information

No strong feeling for or against.  I guess n_students is a little more accurate

drammock:
all green. rendered doc page:
https://33871-1301584-gh.circle-artifacts.com/0/dev/funding.html

hoechenberger (approved):

larsoner:
Thanks @drammock

PR #9724, opened 2021-09-09
MRG, DOC: Link to funding

larsoner:
Also fixes a warning-as-error bug on `main`, a build bug due to `mne-connectivity` release, plus adds URLs for funding sources that I could find.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -850,6 +850,9 @@ def reset_warnings(gallery_conf, fname):
         'ignore', message="can't resolve package from", category=ImportWarning)
     warnings.filterwarnings(
         'ignore', message='.*mne-realtime.*', category=DeprecationWarning)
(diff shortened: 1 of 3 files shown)


larsoner:
conda-forge failures are noted / fixed in https://github.com/conda-forge/pyvista-feedstock/pull/51, hopefully it lands very quickly so we don't have to work around it...

larsoner:
Much more blue here: https://33885-1301584-gh.circle-artifacts.com/0/dev/funding.html

Should be good to merge if you're happy @drammock

drammock:
you are OK with the ghActions failures (`module not found: transforms3d`)?

drammock (approved):

drammock:
(never mind, I just missed the earlier comment about conda-forge / pyvista)

PR #9729, opened 2021-09-13
MAINT: Fix link

larsoner:
Fixes this week's linkcheck legitimate error (there was another but it was a false alarm). A better solution might come from https://github.com/imageio/imageio/issues/663 but we can use this in the meantime.


agramfort:
thx @larsoner

PR #9731, opened 2021-09-13
FIX: Restore dipy

larsoner:
Let's see if dipy pre works now that https://github.com/dipy/dipy/pull/2435 has been merged


larsoner:
`numba` requires `numpy<1.21` which means we can't test `--pre` numpy (currently on 1.22.dev), so I'm removing `numba` from the `--pre` runs. In general I think `numba` prereleases lag both CPython and NumPy releases by many (~6+) months, so I think we have to choose between being able to test CPython/NumPy prereleases as often as possible and being able to test numba prereleases (which are rare anyway).

PR #9732, opened 2021-09-14
[BUG, MRG] Keep function globals

alexrockhill:
It looks like when the verbose decorator is added, the `__globals__` are lost because they aren't added as attributes to the newly made function. This is to fix that.

### DIFF
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -131,7 +131,8 @@ def %(name)s(%(signature)s):\n
     evaldict = dict(
         _use_log_level_=use_log_level, _function_=function)
     fm = FunctionMaker(function, None, None, None, None, function.__module__)
-    attrs = dict(__wrapped__=function, __qualname__=function.__qualname__)
+    attrs = dict(__wrapped__=function, __qualname__=function.__qualname__,
+                 __globals__=function.__globals__)
     return fm.make(body, evaldict, addsource=True, **attrs)
 
 


cbrnr:
Would [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) also work? This seems to be the go-to solution whenever writing a decorator, but I don't know if it also handles what we need here.

agramfort:
no opinion here.

larsoner:
This seems harmless enough, thanks @alexrockhill !

> Would functools.wraps also work?

I think a long time ago we went with the simplest solution until it wasn't sufficient, then used the easy parts of the `decorator` package until it wasn't sufficient, and now use lower-level parts of that to make `verbose` do the right thing. It's possible that something simpler might work, but I doubt it. Feel free to try separately if you are motivated to do so, but I don't see too much value in digging into it in this PR.

cbrnr:
> I think a long time ago we went with the simplest solution until it wasn't sufficient, then used the easy parts of the decorator package until it wasn't sufficient, and now use lower-level parts of that to make verbose do the right thing. It's possible that something simpler might work, but I doubt it. Feel free to try separately if you are motivated to do so, but I don't see too much value in digging into it in this PR.

Yeah, this was the reason I was mentioning this. The code seems to have evolved into something that's not easy to understand, so trying if something simpler also works might be worth it. I don't have the time to do it anytime soon though (and yes, this should of course be tackled in a new PR).

PR #9735, opened 2021-09-14
MRG, MAINT: Add ignore_namespaces to mne.utils.warn

larsoner:
Allow downstream modules (e.g., mne-bids, mne-connectivity, etc.) to do this:
```
<in mne_bids.utils>

def warn(msg, category=RuntimeWarning):
    from mne.utils import warn as _warn
    return _warn(msg, category, module='mne_bids',
                 ignore_namespaces=('mne', 'mne_bids')
```
Then when they use `mne_bids.utils.warn(...)` with functions that have a `verbose` decorator, the magical `stacklevel` traversal and `verbose`-conditional warning emission will work properly. Overhead at the MNE-Python end is minimal because our default `ignore_namespaces=('mne',)` will evaluate in the list comprehensions in negligible time.

@alexrockhill it would be great if you could test your MNE-BIDS PR in conjunction with this. In the meantime you'll probably need to vendor this version of the function in the case where `'ignore_namespaces' not in mne.fixes._get_args(mne.utils.warn)` (which will be true until this PR is merged, etc.).

No `latest.inc` update needed I think since this is an internal maintainability change.

### DIFF
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -5,6 +5,7 @@
 # License: BSD-3-Clause
 
 import contextlib
+import importlib
 import inspect
 from io import StringIO
 import re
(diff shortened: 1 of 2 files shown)


agramfort:
thx @larsoner

PR #9742, opened 2021-09-15
[MAINT, MRG] Enable pooch to perform fetching of datasets

### LINKED ISSUE #9736: Adding an optional token to the dataset fetcher code to allow optional fetching from private repositories
The dataset fetching code inside `mne/datasets/utils.py`, `mne/utils/fetching.py` are actually very general. I was hoping to leverage them without copy/pasting the code, so I can make use of upstream possible bug fixes / performance improvements (if they ever occur). 

However, in some cases, I would like to unit test against private data I have stored on Github, and they require an API token with the HTTP request. Eventually, then some of that data would be made public after say a publication, but it's then nice to build into a CI for myself for a private research project in the meantime. 

Is it possible to add an optional "token" into the dataset fetcher? This would also enable MNE to leverage private repos. In addition, it would lessen the code dependency for anyone trying to implement a data fetcher without copying every single function from MNE.

Add optional `token=None` kwarg to the following functions:

1. `_download`
2. `_fetch_file`
3. `_get_http`

Then one can easily add optional tokens in `_data_path`, depending on which dataset is being fetched. This would also enable any "mne" package, like mne-bids/connectivity/etc. to leverage private Github repo data that might get passed in via GH actions.

If we further refactor things, so that `key`, `urls`, `archive_names`, `folder_origs`, `folder_names`, `md5_hashes` are passed into `_data_path`, rather then set inside `_data_path`, then to create a MNE-fetcher, one simply needs to define a `data_path` that then passes these to `_data_path`, and they have a fully functional: `mne_downstream_package.testing.data_path()` that fetches their own datasets for testing without having to rely on MNE-Python for data fetching.

#### Additional Information
I think this also might be helpful in further cementing MNE-Python as a platform for developing neuroscience/clinical-neuroscience applications that sometimes might need data fetchers in their CI / testing pipeline for "private data".

Ref: https://chanzuckerberg.com/eoss/proposals/improving-usability-of-core-neuroscience-analysis-tools-with-mne-python/

### LINKED ISSUE #9756: CI, MAINT: Unclosed file ResourceWarning in EDF export
`ResourceWarning: unclosed file` related to EDF export

https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=15442&view=logs&j=fa98c7b3-575e-55b6-6946-22dea6f6daf8&t=c0866e65-eed5-5461-eadc-f0ff150eefd8&l=3120

_Originally posted by @drammock in https://github.com/mne-tools/mne-python/issues/9742#issuecomment-922065692_

cc @adam2392

adam2392:
#### Reference issue

First part of #8679 #9736 

closes #9756 

Builds on top of @drammock work in #8679 to:

- move all definition of a dataset to a `config.py` + `dataset_checksums.txt` file. These define urls, archive names, configuration keys, folder names, and md5 hashes
- use pooch to do the downloading, unzip/untar and QA checks
- replace all of `utils/fetching.py` with pooch usage in all datasets that used `_fetch_file()`.

~~I'm not sure how to handle bst, eegbci, hf_sef, limo and other complicated files yet. I think maybe I can copy @drammock's work? I'll have to investigate.~~

- [x] handle bst files
- [x] handle eegbci files
- [x] handle hf_sef files
- [x] handle limo files
- [x] handle `fsaverage` files
- [x] delete `_fetch_file` and related files

Note that the checksums file for eegbci is 3058 lines long, so in reality this code diff is:
- ~800 green lines
- 784 red lines
as of 09/15/21

#### Additional information
The next PR should handle:

1. passing in an optional authorization token to `_data_path` which would then address #9736 
2. generalize `_data_path` to a public function, perhaps... `mne_dataset_path()`, which then can be used by 3rd parties who don't want to store data in the MNE datasets section.


environment.yml
```diff
@@ -39,3 +39,4 @@ dependencies:
 - ipywidgets
 - pip:
   - ipyvtklink
+  - pooch
```
drammock, line 42: pooch is available on conda-forge, we should move this out of the `pip` section

mne/datasets/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+6. Failure to abide by these guidelines will result in termination of my
+   privileges to access WU-Minn HCP data.
+"""
+
+# To update the `testing` or `misc` datasets, push or merge commits to their
+# respective repos, and make a new release of the dataset on GitHub. Then
+# update the checksum in `mne/data/dataset_checksums.txt` and change version
+# here:                  ↓↓↓↓↓         ↓↓↓
+RELEASES = dict(testing='0.112', misc='0.8')
```
drammock, line 90: these numbers are out of date now (testing is on 0.123, not sure about misc)

mne/datasets/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+# To update any other dataset besides `testing` or `misc`, upload the new
+# version of the data archive itself (e.g., to https://osf.io or wherever) and
+# then update the corresponding checksum in `mne/data/dataset_checksums.txt`.
+TESTING_VERSIONED = f'mne-testing-data-{RELEASES["testing"]}'
+MISC_VERSIONED = f'mne-misc-data-{RELEASES["misc"]}'
+
+# remote locations of the various datasets
+URLS = dict(
```
drammock, line 99: might make sense to alphabetize these lines by key, just to make future devs a little happier.

drammock (commented):
a few prelim comments

adam2392:
So I think majority of the work was done by @drammock originally in his PR. Really great push! I've added some additional changes to basically remove as much redundant code as possible.

Here, I have basically removed almost the entire `utils/fetching.py` file because we frankly don't need it if we allow `pooch` to be an optional requirement for fetching datasets.

This PR does not change the user interface, and so far works pretty great for me to pull datasets. If this PR is approved, then I can easily add in a public interface (turning `_data_path` to a public function) that assumes some structure on top of pooch to basically enable any downstream MNE package to define their own data_paths, leveraging the `_data_path()` function.

adam2392:
Fixing the docs error.

The only other error is:

```
________________________________ test_downloads ________________________________
3404
mne/datasets/tests/test_datasets.py:79: in test_downloads
3405
    new_path = datasets._fake.data_path(update_path=True, **kwargs)
3406
```

However, I'm not entirely sure why that line is producing an error... Any ideas?

adam2392:
The error has to do w/ capsys, or calling data_path twice? I think?

This test succeeds:

```
def test_download(tmp_path, monkeypatch, capsys):
    kwargs = dict(path=str(tmp_path), verbose=True)
    # XXX we shouldn't need to disable capsys here, but there's a pytest bug
    # that we're hitting (https://github.com/pytest-dev/pytest/issues/5997)
    # now that we use pooch
    # with capsys.disabled(): 
    # path = datasets._fake.data_path(update_path=False, **kwargs)

    monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmp_path))
    with pytest.warns(RuntimeWarning, match='non-standard config'):
        new_path = datasets._fake.data_path(update_path=True, **kwargs)

    # assert path == new_path
    out, _ = capsys.readouterr()
    assert 'Downloading' not in out

```

But this one fails

```
def test_download(tmp_path, monkeypatch, capsys):
    kwargs = dict(path=str(tmp_path), verbose=True)
    # XXX we shouldn't need to disable capsys here, but there's a pytest bug
    # that we're hitting (https://github.com/pytest-dev/pytest/issues/5997)
    # now that we use pooch
    with capsys.disabled(): 
        path = datasets._fake.data_path(update_path=False, **kwargs)

    monkeypatch.setenv('_MNE_FAKE_HOME_DIR', str(tmp_path))
    with pytest.warns(RuntimeWarning, match='non-standard config'):
        new_path = datasets._fake.data_path(update_path=True, **kwargs)

    assert path == new_path
    out, _ = capsys.readouterr()
    assert 'Downloading' not in out
```

mne/datasets/utils.py
```diff
@@ -226,196 +157,93 @@ def _do_path_update(path, update_path, key, name):
...
+        logger.info(f'Dataset {name} version {data_version} out of date, '
+                    f'latest version is {want_version}')
+
+    # reasons to bail early (hf_sef has separate code for this):
+    if (not force_update) and (not outdated_dataset) and \
+            (not name.startswith('hf_sef_')):
+        # if target folder exists (otherwise pooch downloads every time,
+        # because we don't save the archive files after unpacking)
+        if op.isdir(final_path):
+            return final_path
```
drammock, line 241: @adam2392 this is why the RuntimeWarning is not getting raised the second time we call `datasets._fake.data_path()` in the test. the `_do_path_update` is never getting reached because we return here.

adam2392: Thanks for pointing this out. I think I fixed it now.

setup.cfg
```diff
@@ -26,7 +26,7 @@ addopts =
     --doctest-ignore-import-errors --junit-xml=junit-results.xml
     --ignore=doc --ignore=logo --ignore=examples --ignore=tutorials
     --ignore=mne/gui/_*.py --ignore=mne/externals --ignore=mne/icons
-    --capture=sys
+    # --capture=sys
```
drammock, line 29: don't change this.  (it probably won't help anyway, the default is `--capture=fd` which is even more aggresive than `--capture=sys`)

adam2392: Oh yeah sorry, this was scruff to help me test things. The CIs look good!

drammock: the perils of `git commit -a` come to life!  🙂

adam2392:
Remaining errors aren't due to this PR I think(?)

```
_____________________ test_compute_whitener[white-False] ______________________
mne\tests\test_cov.py:91: in test_compute_whitener
    assert_allclose(W, W2, rtol=rtol)
E   AssertionError: 
E   Not equal to tolerance rtol=1e-11, atol=0
E   
E   Mismatched elements: 20230 / 133956 (15.1%)
E   Max absolute difference: 2418.95239258
E   Max relative difference: 1.63754816e-06
E    x: array([[-1.282830e+09,  8.186675e+09, -7.612984e+11, ...,  0.000000e+00,
E            0.000000e+00,  0.000000e+00],
E          [-1.717234e+10, -3.549120e+10, -1.414323e+13, ...,  0.000000e+00,...
E    y: array([[-1.282830e+09,  8.186675e+09, -7.612984e+11, ...,  0.000000e+00,
E            0.000000e+00,  0.000000e+00],
E          [-1.717234e+10, -3.549120e+10, -1.414323e+13, ...,  0.000000e+00,...
______________________ test_compute_whitener[False-True] ______________________
mne\tests\test_cov.py:91: in test_compute_whitener
```

mne/datasets/utils.py
```diff
@@ -426,32 +260,97 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
...
+    fetcher = pooch.create(
+        path=path,
+        base_url='',    # Full URLs are given in the `urls` dict.
+        version=None,   # Data versioning is decoupled from MNE-Python version.
+        registry=None,  # Registry is loaded from file, below.
+        urls=pooch_urls,
+        retry_if_failed=2  # 2 retries = 3 total attempts
+    )
+    # load the checksum registry
+    registry = files('mne.data').joinpath('dataset_checksums.txt')
```
drammock, line 321: @larsoner are you OK with this use of `importlib_resources.files` here?  I had to add `importlib_resources` to `requirements_testing` for this to work.  (the alternative is `pkg_resources`, which we're supposed to be moving away from)

(FYI, the corresponding `importlib.resources.files()` was added in py3.9 so we can't yet use `importlib` directly for this)

larsoner: > I had to add importlib_resources to requirements_testing for this to work

This is not used only in `tests/*.py` but also in the main `mne` module, so this needs to be in `environment.yml` and `requirements.txt` instead. What's wrong with `op.join(op.dirname(__file__), '..', 'data', 'dataset_checksums.txt')`? Seems clear enough and simple enough to avoid a new dependency...

drammock: > This is not used only in `tests/*.py` but also in the main `mne` module, so this needs to be in `environment.yml` and `requirements.txt` instead.

that's why I asked (but admittedly I didn't make it clear; what I meant was "for the moment I hacked this in via `requirements_testing` but this will become a new dependency").

> What's wrong with `op.join(op.dirname(__file__), '..', 'data', 'dataset_checksums.txt')`?

1. "The use of `__file__` doesn’t work if your package lives inside a zip file, since in that case this code does not live on the file system" (https://importlib-resources.readthedocs.io/en/latest/using.html#example)  I interpret that to mean that there are potential problems with egg files, although admittedly I haven't dug deep on this.
2. using `__file__` this way is, IMO, hard to read and hard to use (I often struggle with this when I'm trying to interactively run a test line-by-line; I can't just `ctrl+enter` to run a line that defines a path if it has a `dirname(__file__)` in it).
3. `importlib` is part of the standard library and designed for just this kind of use case, and `importlib_resources` is a transitional package with the same API as `importlib.resources` backported to work with older versions of python. in that sense it is future-compatible in that the only thing that will need to change later is to change `_` to `.` in `from importlib_resources import foo`

drammock: if you want, I could skip the transitional package, try plain `importlib` (will work on py3.9 only) and fallback to `pkg_resources`, similar to what is done here:

https://github.com/mne-tools/mne-python/blob/3aeee826450c823464282657a38c6751d9cdc5a2/doc/conf.py#L898-L906

larsoner: Sure, maybe a new `mne.utils._mne_path()` that returns the path? Then eventually all of our `__file__`s can just use that thing, and that thing can implement the best try/except-based path logic based on our minimum python version

drammock (changes requested):
@adam2392 here's an unsolicited suggestion to avoid all these "cruft" reviews: when you comment out code temporarily, don't put in a space after the `#`.  That way your editor (properly configured) will highlight it as a style error so you are less likely to forget about it.

I'll work on @larsoner's suggestion for the importlib try/except utility function next.

adam2392:
@drammock apologies! Got excited that it worked. Also I use (shortcut keyboard) auto-commenting so I'll look into it if I can adjust the settings for that in vscdoe.

mne/datasets/eegbci/eegbci.py
```diff
@@ -71,7 +79,13 @@ def data_path(url, path=None, force_update=False, update_path=None,
...
         if not op.isdir(op.dirname(destination)):
             os.makedirs(op.dirname(destination))
-        _fetch_file(url, destination, print_destination=False)
+        pooch.retrieve(
+            # URL to one of Pooch's test files
+            url=url,
+            path=destination,
+            fname=fname
+        )
+        # _fetch_file(url, destination, print_destination=False)
```
drammock, line 88:
```suggestion
```

requirements_testing.txt
```diff
@@ -11,3 +11,4 @@ pydocstyle
 check-manifest
 twine
 wheel
+pooch
```
drammock, line 14: don't need this here if it's in `requirements.txt` and `environment.yml` already
```suggestion
```

setup.py
```diff
@@ -92,6 +92,7 @@ def package_tree(pkgroot):
           install_requires=install_requires,
           packages=package_tree('mne'),
           package_data={'mne': [
+              op.join('data', 'dataset_checksums.txt'),
```
drammock, line 95:
```suggestion
              op.join('data', 'dataset_checksums.txt'),
              op.join('data', 'eegbci_checksums.txt'),
```

mne/datasets/sleep_physionet/_utils.py
```diff
@@ -80,11 +91,18 @@ def _update_sleep_temazepam_records(fname=TEMAZEPAM_SLEEP_RECORDS):
...
+    subjects_fname = op.join(tmp, fname)
+    # _fetch_file(url=TEMAZEPAM_RECORDS_URL,
+    #             file_name=subjects_fname,
+    #             hash_=TEMAZEPAM_RECORDS_URL_SHA1,
+    #             hash_type='sha1')
+    pooch.retrieve(
+        url=TEMAZEPAM_RECORDS_URL,
+        known_hash=f"sha1:{TEMAZEPAM_RECORDS_URL_SHA1}",
+        path=tmp,
+        # fname=fname
```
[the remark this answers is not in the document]
drammock: @adam2392 I applied all the other code suggestions from my review, but can you double-check this one? I didn't look too carefully at it.

requirements_testing.txt
```diff
@@ -11,3 +11,5 @@ pydocstyle
...
+importlib_resources
```
drammock, line 15:
```suggestion
```

drammock (approved):
just one last bit of cruft in `requirements_testing.txt`. I think this is good to go after that.

adam2392:
Now there is an error with GH actions though:

```
    pooch = _soft_import('pooch', 'dataset downloading', strict=True)
61
  File "/usr/share/miniconda/envs/test/lib/python3.7/site-packages/mne-0.24.dev0-py3.7.egg/mne/utils/check.py", line 273, in _soft_import
62
    raise RuntimeError(f'For {purpose} to work, the {name} module is '
63
RuntimeError: For dataset downloading to work, the pooch module is needed, but it could not be imported.
```

Do we need pooch somewhere besides where we already added?

drammock:
> Now there is an error with GH actions though:

dang, OK, I'll fix it

drammock:
ci failure is unrelated (Windows 3.7 pip timeout)

mne/data/dataset_checksums.txt
```diff
@@ -0,0 +1,46 @@
+# include here the name of the zipped file and the md5 hash
```
adam2392, line 1: later PR: Put into python code and create temp file for internal

drammock:
TODO list (from live chat with @agramfort and @adam2392):

1. make sure `pip install mne` and then `import mne` will still work without pooch (i.e., make sure pooch imports are properly nested)
2. add pooch to `test_import_nesting`
3. restructure dataset info as nested dict of dicts; incorporate MD5s into python and change pooch to temporarily write the registry file
4. make it easier to add new datasets (separate PR)
5. add generic dataset downloader function, that allows API keys (separate PRs)

adam2392:
Unrelated CI failure.

drammock:
> Unrelated CI failure.

indeed they look unrelated to this PR, though did you see the `ResourceWarning: unclosed file` related to EDF export?
https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=15442&view=logs&j=fa98c7b3-575e-55b6-6946-22dea6f6daf8&t=c0866e65-eed5-5461-eadc-f0ff150eefd8&l=3120

adam2392:
Ah fixed the resource warning. It was an error with not closing the files when a runtime error is raised.

@drammock

drammock:
all green!  pooch PR is ready for final review

agramfort (approved):
@larsoner merge if happy

mne/export/_edf.py
```diff
@@ -261,6 +266,7 @@ def _export_raw(fname, raw, physical_range, add_ch_type):
             onset = onset * 10000
             duration = duration * 10000
             if hdl.writeAnnotation(onset, duration, desc) != 0:
+                hdl.close()
```
larsoner, line 269: This solution is less clean / DRY than a solution that uses a context manager like `with hdl:` or, if that's not an option:
```
@contextmanager
def _close_file(fid):
    try:
        yield
    finally:
        fid.close()
```
But we can clean this up in a follow-up PR

tools/azure_dependencies.sh
```diff
@@ -17,6 +17,7 @@ elif [ "${TEST_MODE}" == "pip-pre" ]; then
 	python -m pip install --progress-bar off --upgrade --only-binary ":all" "vtk<=9.0.1"
 	python -m pip install --progress-bar off https://github.com/pyvista/pyvista/zipball/main
 	python -m pip install --progress-bar off https://github.com/pyvista/pyvistaqt/zipball/main
+	python -m pip install --progress-bar off pooch
```
larsoner, line 20: This is not needed because the last line includes `pip install -r requirements_testing.txt` and it's in there

tools/github_actions_dependencies.sh
```diff
@@ -31,6 +31,8 @@ else
 	pip install --progress-bar off https://github.com/pyvista/pyvistaqt/zipball/main
 	echo "imageio-ffmpeg, xlrd, mffpy"
 	pip install --progress-bar off --pre mayavi imageio-ffmpeg xlrd mffpy
+	echo "pooch"
+	pip install --progress-bar off pooch
```
larsoner, line 35: Same

larsoner:
Thanks @adam2392 @drammock ! I'll open a quick PR for my tiny comments

mne/datasets/utils.py
```diff
@@ -426,32 +259,97 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
...
+        testing=untar,
+        visual_92_categories=untar,
+        phantom_4dbti=unzip,
+        refmeg_noise=unzip,
+        hf_sef_raw=pooch.Untar(
+            extract_dir=path, members=[f'hf_sef/{subdir}' for subdir in
+                                       ('MEG', 'SSS', 'subjects')]),
+        hf_sef_evoked=pooch.Untar(
+            extract_dir=path, members=[f'hf_sef/{subdir}' for subdir in
+                                       ('MEG', 'SSS', 'subjects')]),
```
adam2392, line 302: @drammock do you know why we needed hf_Sef to be like this?  These LOC I copied seems to break  circle CI in : https://github.com/mne-tools/mne-python/issues/9774

It seems to work perfectly fine w/ `processor = 'untar'` untaring to the path. I confirm that the error was not raised when I changed the processor to just regular `pooch.Untar(extract_dir=path`

adam2392: To clarify, I copied these from your original PR aimed to tackle #8674

PR #9748, opened 2021-09-16
Report

agramfort:
WIP / Test


larsoner:
See https://github.com/mne-tools/mne-python/pull/9749 which I think will be an easier route

PR #9749, opened 2021-09-16
MAINT: Move report code

larsoner:
Let's try this from another angle. This PR just moves `mne/report.py`->`mne/report/report.py` along with its accompanying paths and infrastructure. Then once that's done @hoechenberger you can easily overwrite and the diff should be more manageable.

@hoechenberger an easy way once this is merged (if we go this route) would be to:
```
git fetch upstream
git checkout main
git reset --hard upstream/main
git checkout -b report2
git checkout report mne/report/report.py mne/data/html_templates.py mne/report/__init__.py ... # all files you changed
rm mne/html/jquery.js ... # all files you deleted
git reset  # if you want to be able to see "git diff"
git commit -am 'ENH: Rework report'
git push origin -u report2
```
It'll only take a minute to enumerate the files changed and files deleted, so not too painful I think.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -31,6 +31,8 @@ recursive-include mne/gui/help *.json
 recursive-include mne/html *.js
 recursive-include mne/html *.css
 
+recursive-include mne/report *
+
 recursive-include mne/io/artemis123/resources *
 
 recursive-include mne mne/datasets *.csv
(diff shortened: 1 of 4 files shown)


agramfort:
thx @larsoner 

CI failures are unrelated

hoechenberger:
Thanks a bunch, @larsoner!!

PR #9754, opened 2021-09-16
WIP: Rework Reports (new history)

hoechenberger:
Rework of #9722 after #9749 had been merged


mne/report/__init__.py
```diff
@@ -1 +1 @@
-from .report import Report, open_report, _ReportScraper
+from .report import Report, open_report, _ReportScraper, _BEM_VIEWS
```
larsoner, line 1: You don't need this

mne/report/report.py
```diff
@@ -101,55 +245,47 @@ def _fig_to_img(fig, image_format='png', scale=None, auto_close=True,
...
-        warnings.simplefilter('ignore')  # incompatible axes
-        fig.savefig(output, format=image_format, dpi=fig.get_dpi(),
-                    bbox_to_inches='tight')
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            action='ignore',
+            message='.*not compatible with tight_layout.*',
+            category=UserWarning
+        )
+        fig.tight_layout()
```
larsoner, line 256: We have our own `mne.viz.utils` version of `tight_layout` for this reason

mne/report/report.py
```diff
@@ -101,55 +245,47 @@ def _fig_to_img(fig, image_format='png', scale=None, auto_close=True,
         fig = _ndarray_to_fig(img)
 
     output = BytesIO()
-    if scale is not None:
-        _scale_mpl_figure(fig, scale)
```
larsoner, line 105: There probably needs to be a deprecation cycle for this

mne/report/report.py
```diff
@@ -950,7 +509,7 @@ class Report(object):
 
     def __init__(self, info_fname=None, subjects_dir=None,
                  subject=None, title=None, cov_fname=None, baseline=None,
-                 image_format='png', raw_psd=False, projs=False, verbose=None):
+                 image_format='png', raw_psd=True, projs=False, verbose=None):
```
larsoner, line 512: Backward incompat change

mne/report/tests/test_report.py
```diff
@@ -264,14 +266,17 @@ def test_report_raw_psd_and_date(tmpdir):
...
+    raw.info['subject_info'] = dict(id=123, his_id='sample')
+
     raw_fname_new = op.join(tempdir, 'temp_raw.fif')
     raw.save(raw_fname_new)
     report = Report(raw_psd=True)
     report.parse_folder(data_path=tempdir, render_bem=False,
                         on_error='raise')
     assert isinstance(report.html, list)
     assert 'PSD' in ''.join(report.html)
-    assert 'GMT' in ''.join(report.html)
```
larsoner, line 274: Seems like a backward incompat change to go from what used to be GMT to something else. -1 on going to the current computer's time (this makes my result and your result differ when analyzing the same data), but +1 on going to the acquisition machine's time zone if available (can't remember if this is stored or not).

If we want, we can use JavaScript to convert to the viewer's time zone on the fly like we do with out dev doc footer:

https://github.com/mne-tools/mne-python/blob/main/doc/conf.py#L53-L55

This makes the contents of the file the same, but when viewed in a browser, it shows local.

mne/report/tests/test_report.py
```diff
@@ -649,3 +624,98 @@ def test_survive_pickle(tmpdir):
...
+                    tags=('ssp', 'ecg'))
+    r.add_ssp_projs(info=raw_fname, title='my proj', tags=('ssp'))
+    r.add_covariance(cov=cov_fname, info=raw_fname, title='my cov')
+    r.add_forward(forward=fwd_fname, title='my forward', subject='sample',
+                  subjects_dir=subjects_dir)
+
+    if get_3d_backend() is not None:
+        r.add_trans(trans=trans_fname, info=raw_fname, title='my coreg',
+                    subject='sample', subjects_dir=subjects_dir)
+        r.add_bem(subject='sample', subjects_dir=subjects_dir, title='my bem')
```
larsoner, line 654: This will probably take too long. We should make sure to heavily decimate the slicing here and elsewhere in new tests to minimize testing time.

mne/utils/docs.py
```diff
@@ -2486,7 +2503,7 @@
     Name of the output file.
 """
 docdict['export_params_fmt'] = """
-fmt : 'auto' | 'eeglab' | 'edf'
+fmt : 'auto' | 'eeglab'
```
larsoner, line 2506: Looks like a bad merge (probably from the `git checkout ...` line), you'll need to manually put the lines back on this one :(

larsoner (commented):
Great that there is a real diff now! I did not look at everything, but noticed a few minor things

hoechenberger:
Thanks for the initial suggestions, @larsoner! Some good catches. I'll try to address all of them and ping you.

mne/utils/docs.py
hoechenberger: I cherry-picked the respective commit from `main` (see 0658f29) and I think this got all those changes now.

mne/report/__init__.py
hoechenberger: Removed

mne/report/report.py
hoechenberger: I've reverted this change

mne/report/tests/test_report.py
hoechenberger: Actually this line was removed by accident, we don't have a regression here. I added the test back, no further changes required.

mne/report/report.py
hoechenberger: I've added the scaling back where it makes sense: `add_figs_to_section` and `add_slider_to_section`. I'm silently ignoring it for `add_images_to_section`, where it would previously not do any "actual" resizing, but simply scale the image via CSS – which would not make sense with our new, adaptive layout.

mne/report/report.py
```diff
@@ -757,6 +757,10 @@ def add_evokeds(self, evokeds, *, titles=None, noise_cov=None, projs=True,
         projs : bool
             Whether to add SSP projector plots if projectors are present in
             the data.
+        n_time_points : int | None
+            The number of equidistant time points to render. If ``None``,
+            will render each `~mne.Evoked` at 21 time points, unless the data
+            contains fewer time points, in which call all will be rendered.
```
agramfort, line 763:
```suggestion
            contains fewer time points, in which case all will be rendered.
```

mne/report/report.py
```diff
@@ -868,6 +874,7 @@ def add_raw(self, raw, title, *, psd=None, projs=True, tags=('raw',),
             raw=raw,
             add_psd=add_psd,
             add_projs=add_projs,
+            add_butterfly=butterfly,
```
agramfort, line 877:
```suggestion
            butterfly=butterfly,
```
Would be more natural

mne/report/report.py
```diff
@@ -908,6 +915,10 @@ def add_stc(self, stc, title, *, subject=None, subjects_dir=None,
             creation.
         subjects_dir : path-like | None
             The FreeSurfer ``SUBJECTS_DIR``.
+        n_time_points : int | None
+            The number of equidistant time points to render. If ``None``,
+            will render ``stc`` at 51 time points, unless the data
+            contains fewer time points, in which call all will be rendered.
```
agramfort, line 921:
```suggestion
            contains fewer time points, in which case all will be rendered.
```

mne/report/report.py
```diff
@@ -2083,6 +2110,18 @@ def parse_folder(self, data_path, pattern=None, n_jobs=1, mri_decim=2,
...
 
             .. versionadded:: 0.16
+        n_time_points_evokeds, n_time_points_stcs : int | None
+            The number of equidistant time points to render for `~mne.Evoked`
+            and `~mne.SourceEstimate` data, respectively. If ``None``,
+            will render each `~mne.Evoked` at 21 and each `~mne.SourceEstimate`
+            at 51 time points, unless the respective data contains fewer time
+            points, in which call all will be rendered.
+
+            .. versionadded:: 0.24.0
```
agramfort, line 2120: Stc param missing

hoechenberger: It's here, see line 2113:
```python
n_time_points_evokeds, n_time_points_stcs : int | None
```

mne/report/tests/test_report.py
hoechenberger: I'm decimating now much more (factor 100), so this shouldn't be much of a problem anymore, I hope

mne/report/report.py
agramfort: what's the reason why sometimes you have add_ and sometimes you don't?

hoechenberger:
Rendered docs:

[Tutorial](https://34972-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/70_report.html)
[API docs](https://34972-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Report.html)

hoechenberger:
@wmvanvliet @rob-luke If you're brave and curious, I'd like to hear how this works for you 🤞

mne/report/report.py
hoechenberger: Example with `projs`:

The public methods have `projs` (to be consistent with the rest of the MNE public API)
The private methods have `add_projs` params as this is actually a boolean value and I wanted to emphasize this fact

`add_projs` is produced by looking at `projs` and `self.projs` because in some cases we'll fall back to the value specified on `Report` instantiation:

```python
        add_projs = self.projs if projs is None else projs
```

For `butterfly` in the above example, we don't need to do this, as we don't store any "butterfly defaults" on `Report` instantiation. So the transformation from `butterfly` to `add_butterfly` is trivial, because they're identical.

… does that make sense? 😅

rob-luke:
Sorry @hoechenberger I won't have time to try this locally in the next week or so. But I checked out the tutorial and it looks great. The pages render very nicely and you still have the custom plot section, so that works for me. Cool stuff. Ill check back when I have a chance to code again.

hoechenberger:
Latest doc builds:

[Tutorial](https://35001-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/70_report.html#sphx-glr-auto-tutorials-intro-70-report-py)
[API docs](https://35001-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Report.html#mne.Report)

mne/report/report.py
```diff
@@ -1069,92 +688,812 @@ def add_custom_js(self, js):
...
+
+        self._add_or_replace(
+            dom_id=dom_id,
+            name=title,
+            tags=tags,
+            html=html,
+            replace=replace
+        )
+
+    @fill_doc
```
agramfort, line 1496: This should be deprecated no? Please make sure that doc strings of all mention the deprecation so it appears in the API page what method is deprecated.

hoechenberger: In fact, I [removed](fcec676) the `@deprecated` decorator and started "manually" emitting `DeprecationWarning`s because the decorator was causing issues with numpydoc. I posted this on Discord:

The following signature, for example:

```
    @deprecated(extra='Use `Report.add_image` instead')
    def add_images_to_section(self, fnames, captions, scale=None,
                              section='custom', comments=None, replace=False):
```

Would cause the following warning during the doc build:

```
WARNING: [numpydoc] Validation warnings while processing docstring for 'mne.Report.add_images_to_section':
  SS03: Summary does not end with a period
  PR01: Parameters {'**kwargs', '*args'} not documented
  PR02: Unknown parameters {'replace', 'section', 'captions', 'fnames', 'scale', 'comments'}
  RT01: No Returns section found
```

Any suggestions on how to fix this would be appreciated!

hoechenberger: @agramfort I've now hacked my way around this in 6d8f9ac

Not pretty but it works.

larsoner:
I'll push a commit to fix/modernize `@deprecated`, okay for me to `git merge upstream/main` while I'm at it to fix the conflict?

hoechenberger:
@larsoner yes absolutely, feel free to do anything except for force-pushes :)

mne/report/report.py
```diff
@@ -1191,27 +1529,115 @@ class construction.
...
+                f'Number of figs ({len(figs)}) must equal number of comments '
+                f'({len(comments)})'
+            )
+
+        tags = _clean_tags(section)
+
+        for fig, title, caption in zip(figs, captions, comments):
+            if scale is not None:
+                _scale_mpl_figure(fig, scale)
+            tight_layout(fig=fig)
```
larsoner, line 1587: So far I've observed loss of functionality in `add_figs_to_section`, it used to support `figs` as a `np.ndarray` directly (e.g., a `(h, w, 3)` array) directly and now I get:
```
  File "/home/larsoner/mne_data/mnefun/mnefun/_report.py", line 594, in gen_html_report
    report.add_figs_to_section(view, captions, section)
  File "/home/larsoner/python/mne-python/mne/report/report.py", line 1586, in add_figs_to_section
    image_format=image_format, tags=tags, replace=replace
  File "/home/larsoner/python/mne-python/mne/viz/utils.py", line 141, in tight_layout
    fig.canvas.draw()
AttributeError: 'numpy.ndarray' object has no attribute 'canvas'
```
(line numbers changed above because I'm making other changes)

I could push a fix that checks if it's an instance of `matplotlib.figure.Figure`, but I also don't like that this `tight_layout` is being forced. I think users should do this at their own end if they want it.

Moreover, I get warnings like this because it's being forced on me:
```
  report.add_slider_to_section(
/home/larsoner/mne_data/mnefun/mnefun/_report.py:943: UserWarning: tight_layout not applied: number of columns in subplot specifications must be multiples of one another.
```
So I'm going to remove this...

mne/report/report.py
```diff
@@ -1397,120 +1966,157 @@ class construction.
...
+                        for i in range(len(figs))]
 
-        slider_klass = sectionvar
+        if image_format is None:
+            image_format = self.image_format
 
+        for fig in figs:
+            if scale is not None:
+                _scale_mpl_figure(fig, scale)
+            tight_layout(fig=fig)
```
larsoner, line 1987: Same story here

mne/report/report.py
hoechenberger: Passing an array SHOULD work, if it doesn't then it's a bug
I can have a look at this

larsoner:
I found some bugs looking at reports I generated using old/deprecated methods. I suggest you fix these first, then once they all work, I can try moving over to new functions, and we can fix any bugs there. Finally I'll try using all old functions one more time, and if both deprecated + new functions render everything properly then hopefully we're good to go!

So to start, using old/deprecated functions -- I hope you can replicate locally using something simple (or immediately see what the problem can be from the images), but if not, I can whittle things down to some minimal example code:

1. Header occlusion bug -- first render of a custom report:
   Screenshot from 2021-09-30 08-22-06
    Resizing the window fixes it:
    Screenshot from 2021-09-30 08-22-11
2. BEMs not resized properly to fit in frame:
    Screenshot from 2021-09-30 08-22-34
3. Headers overlapping TOC column:
   Screenshot from 2021-09-30 08-22-48
4. Sliders don't actually change the shown images in a custom slider I made:
    Screenshot from 2021-09-30 08-23-03
.

mne/report/report.py
larsoner: Don't worry, I'm pushing a commit to remove `tight_layout` entirely here

hoechenberger:
Ouch.

None of these should happen. 

Which browser are you using?

larsoner:
Updated deprecation links nicely in our docs I think:
Screenshot from 2021-09-30 08-32-02

larsoner:
Chrome on Linux. Looking at the console, this seems problematic:
```
DevTools failed to load source map: Could not load content for file:///home/larsoner/mne_data/mnefun/examples/funloc/subj_01/bootstrap.bundle.min.js.map: System error: net::ERR_FILE_NOT_FOUND
DevTools failed to load source map: Could not load content for file:///home/larsoner/mne_data/mnefun/examples/funloc/subj_01/bootstrap.min.css.map: System error: net::ERR_FILE_NOT_FOUND
```
Perhaps this is causing the sliders not to work at least? Not sure why the path would be messed up here...

hoechenberger:
Thank you! Which version of chrome is this?

larsoner (commented):
Comments for everything but `report.py` because it's still a monster :) I'll look at that once all rendering bugs are fixed

hoechenberger:
Thanks @larsoner, will ping you when it's ready

larsoner:
Okay I can replicate all the problems above with this:

<details>

```
import os.path as op
import numpy as np
import matplotlib

import mne

data_path = mne.datasets.testing.data_path() + '/SSS/'
fname = data_path + 'test_move_anon_raw.fif'
struc = 'sample'
subjects_dir = data_path + '../subjects'
trans = data_path + '../MEG/sample/sample_audvis_trunc-trans.fif'
fname_inv = data_path + '../MEG/sample/sample_audvis_trunc-meg-eeg-oct-4-meg-inv.fif'

raw = mne.io.read_raw_fif(fname, allow_maxshield='yes').load_data()
raw.rename_channels(lambda x: f'{x[:3]} {x[3:]}' if 'MEG' in x else x)
fname_pos = data_path + 'test_move_anon_raw.pos'
style = {'axes.spines.right': 'off', 'axes.spines.top': 'off',
         'axes.grid': True}
matplotlib.use('Agg', force=True)
plt.rcParams.update(style)
report = mne.Report(verbose=False)


# %%
report.add_htmls_to_section(
    '<p align="center">Blah blah blah</p>',
    captions='Behavioral', section='Behavioral')

# %%
events = mne.make_fixed_length_events(raw)
epochs = mne.Epochs(raw, events, preload=True)
section = 'Drop log'
figs = [epochs.plot_drop_log()]
captions = [repr(epochs)]  # <--- Does not render properly!
report.add_figs_to_section(figs, captions, section, image_format='svg')  # <--- image_format='svg' BREAKS THINGS!
evoked = epochs.average()

# %%
section = 'BEM'
caption = '%s: %s' % (section, struc)
report.add_bem_to_section(
    struc, caption, section, decim=10, n_jobs=1,
    subjects_dir=subjects_dir)

report.save('report.html', overwrite=True)
```

</details>

Looks like two bugs:

- SVG support broken
- Sanitizing bugs for HTML (`repr(epochs)` as a caption is problematic)

Screenshot from 2021-09-30 10-01-33

hoechenberger:
@larsoner Oh, super good catch! I think I only really tested with PNGs!

I'm currently a little bit lost at to which problems still persist, I guess I'll be in touch on Discord! Talk to you soon

larsoner:
All of the problems [enumerated 2h ago](https://github.com/mne-tools/mne-python/pull/9754#issuecomment-931276147) persist for me, probably due to the [causes I just mentioned](https://github.com/mne-tools/mne-python/pull/9754#issuecomment-931351977), and you should be able to replicate locally by running the example script I posted above

larsoner:
(I pushed commits to fix some issues orthogonal to these, namely the `deprecation` inadequacy and a BEM axis labeling bug)

hoechenberger:
@larsoner Thanks, I'll get to work on this now!

mne/viz/misc.py
```diff
@@ -445,14 +464,14 @@ def plot_bem(subject=None, subjects_dir=None, orientation='coronal',
 
     Parameters
     ----------
-    subject : str
+    subject : str | None
```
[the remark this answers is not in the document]
hoechenberger: This very function here, `plot_bem()`, is not a method of a class. I've tried to make use of a the docdict anyway, LMK what you think. Changes in 33f8ab8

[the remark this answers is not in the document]
larsoner: Isn't `subject=None` not supported, then?

mne/viz/misc.py
```diff
@@ -426,16 +445,16 @@ def _plot_mri_contours(mri_fname, surfaces, src, orientation='coronal',
             if ai < n_col or n_col == 1:  # top
                 ax.text(dat.shape[1] / 2., dat.shape[0] - 1, ylabels[1],
                         ha='center', va='top', **kwargs)
-        if img_output:
-            output = BytesIO()
-            fig.savefig(output, bbox_inches='tight',
-                        pad_inches=0, format='png', dpi=dpi)
-            out.append(base64.b64encode(output.getvalue()).decode('ascii'))
```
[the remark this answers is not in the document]
hoechenberger: I like the idea of returning a list of ndarrays. As you mention correctly, the downstream plotting functions should handle this just fine, unless there are bugs somewhere (which obviously would need to be fixed anyway)

I will look into this after ensuring that ndarray plotting works in general – you reported some issue there with a deprecated function, which I'll look into just now.

mne/utils/docs.py
```diff
@@ -1389,7 +1390,19 @@
 """
 docdict['subject'] = """
 subject : str
-    The FreeSurfer subject name.
+    The FreeSurfer subject name (will be used to set the FreeSurfer
+    environment variable ``SUBJECT``).
```
larsoner, line 1394: Usually setting the env var is not actually done -- we only do this / set the env var when making actual calls to FreeSurfer functions. So I think this is not quite right...

It acts like the SUBJECT env var does for FreeSurfer, but I don't think it's worth mentioning here

mne/utils/docs.py
```diff
@@ -1389,7 +1390,19 @@
...
+"""
+docdict['subject_none'] = """
+subject : str | None
+    The FreeSurfer subject name. If ``None``, the value of the environment
+    variable ``SUBJECT`` will be used.
```
larsoner, line 1399: Most places we allow `subject=None`, it's taken from the class, like in `stc.plot(subject=None)` will use `stc.subject`. It doesn't use the SUBJECT env var I don't think. (I don't think many if any functions respect the SUBJECT env var and I'd rather not start doing so, I think it's going to be error prone.)

larsoner: ... so I think what you have below for `stc_subject` is really the only `subject_none` sort of thing we need in practice

mne/viz/misc.py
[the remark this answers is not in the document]
larsoner: You shouldn't need to do that, I pushed removal of the previously added (in this PR) `tight_layout` call. So ndarrays should work now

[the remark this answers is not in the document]
larsoner: (I added an edit to say by "that" I meant the ndarray fixes, I pushed a commit for that already -- but the changes to make this actually return ndarrays instead of a list of figures would still be good)

mne/utils/docs.py
hoechenberger: I just mentioned it because other docstrings mention(ed), "corresponding to the FreeSurfer SUBJECT variable".

But I get your point and will change this

mne/viz/misc.py
```diff
@@ -445,14 +464,14 @@ def plot_bem(subject=None, subjects_dir=None, orientation='coronal',
...
-    subject : str
+    subject : str | None
         Subject name.
     subjects_dir : str | None
         Path to the SUBJECTS_DIR. If None, the path is obtained by using
         the environment variable SUBJECTS_DIR.
     orientation : str
         'coronal' or 'axial' or 'sagittal'.
-    slices : list of int
+    slices : list of int | None
```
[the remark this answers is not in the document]
hoechenberger: Addressed in fa9993e

Took me a bit too long to figure out what was happening, so I reformatted the code and added a comment

mne/utils/docs.py
hoechenberger: btw the old docstrings for `subject` for the `plot_stc...`  functions were:

```
        The subject name corresponding to FreeSurfer environment
         variable SUBJECT. If None stc.subject will be used. If that
         is None, the environment will be used.
```

hoechenberger: … but I just checked `_check_subject()` and indeed it doesn't use the environment variable, so the docstrings were incorrect.

mne/viz/misc.py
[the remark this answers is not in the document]
hoechenberger: Good catch, I checked the function(s) it calls and they expect a string, not `None`. I had simply adjusted the docstring here because the default param value in the function signature was `subject=None`. So this is actually a bug. I fixed this in bba68f6

mne/viz/misc.py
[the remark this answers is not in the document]
hoechenberger: got you, thanks!

wmvanvliet:
@hoechenberger Just played around with the new report a little. Opening an existing report and adding/replacing figures in it works great!

hoechenberger:
Remaining CI complaints will go away once #9805 is merged

hoechenberger:
CI error seems to due to time limits on Azure, but unrelated to the changes in this PR

@larsoner lmk if you're happy or if you experience any remaining issues!

hoechenberger:
Rendered tutorial: https://35276-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/70_report.html#sphx-glr-auto-tutorials-intro-70-report-py

API docs: https://35276-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Report.html?highlight=report#mne.Report

agramfort (approved):
looks great !

I even just tested the `mne report` command it produced a beautiful report on my laptop !

🎉 👏

larsoner:
[Test time](https://github.com/mne-tools/mne-python/pull/9754/checks?check_run_id=3801018199#step:13:3517) is still a problem (maybe MRI rendering, maybe something else):
```
74.13s call     mne/report/tests/test_report.py::test_full_report
```
The only report-related ones that show up in [a different PR's recent build](https://github.com/mne-tools/mne-python/runs/3796599049#step:13:3518) are:
```
23.52s call     mne/report/tests/test_report.py::test_render_report[mayavi]
23.08s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
```

hoechenberger:
@larsoner Ouch, good catch. I managed to cut down the test time drastically on my local machine, and I'm surprised tests still take so long on Circle. Back to the profiler, I suppose!

larsoner:
3D plotting is much slower there, my guess is it's the new alignment plotting function. One option is to make it so the view/size there is very small so there is less to render

larsoner:
On Linux you can test this using `xvfb-run` because it will use software rendering like the CIs -- want me to try locally?

larsoner:
Locally there is at least some slowdown from software rendering, so I think it probably is the 3D rendered `plot_alignment` slowing things down:
```
$ pytest mne/report
25.12s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
15.49s call     mne/report/tests/test_report.py::test_full_report
...
$ xvfb-run pytest mne/report
23.49s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
21.96s call     mne/report/tests/test_report.py::test_full_report
```
And with `if get_3d_backend() is not None and False:` in `test_report.py:656` so that the 3D stuff is skipped:
```
23.22s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
9.85s call     mne/report/tests/test_report.py::test_full_report
```
So locally I would target the four calls in those to speed them up somehow. Making the render window smaller is one way.

larsoner:
... actually hold on, I think we have a regression with MNE_3D_BACKEND_TESTING where we automagically made it so that these things were faster. Let me see if I can fix it quickly.

hoechenberger:
Oh yes, I think you might be in the right track here!!

larsoner:
Nope, no regression -- really the slowdown here is caused because the `renderer` fixture is not used. But We can easily get around it by using the same actor-invisible trick. I've pushed a commit that locally fixes things, timings for `xvfb-run`:
```
22.97s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
12.97s call     mne/report/tests/test_report.py::test_full_report
```
I'll push and we'll see how the CIs do. But really these are both still too slow, this is what I see on `main` using the same `xvfb-run` call:
```
9.07s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
2.49s call     mne/report/tests/test_report.py::test_render_mri[pyvistaqt]
```

larsoner (changes requested):
Tests too slow currently

hoechenberger:
> But really these are both still too slow, this is what I see on `main` using the same `xvfb-run` call:

But keep in mind that we're **actually** plotting **much more** now than we used to, so a bit of slowdown is to be expected. But I agree, the current slowdown is too much.

hoechenberger:
@larsoner Test runtime summary for `Linux / pip-pre / py3.9`:

```
46.28s call     mne/report/tests/test_report.py::test_render_report[pyvistaqt]
3492
41.12s call     mne/tests/test_bem.py::test_io_head_bem
3493
35.44s call     mne/viz/tests/test_3d_mpl.py::test_plot_volume_source_estimates[stat_map-s-1-1-init_p3-want_p3-brain.mgz]
3494
33.58s call     mne/export/tests/test_export.py::test_export_raw_edf[misc-edf]
3495
31.82s call     mne/viz/tests/test_3d.py::test_plot_alignment_basic[testing_data-testing_data-pyvistaqt]
3496
29.56s call     mne/viz/tests/test_3d_mpl.py::test_plot_volume_source_estimates[stat_map-vec-1-1-None-want_p1-None]
3497
27.18s call     mne/tests/test_morph.py::test_volume_source_morph_round_trip[sample-sample-0.0-0.1-float-True]
3498
25.61s call     mne/tests/test_morph.py::test_volume_source_morph_round_trip[fsaverage-fsaverage-0.0-0.1-float-False]
3499
24.36s call     mne/tests/test_morph.py::test_volume_source_morph_basic
3500
22.71s call     mne/tests/test_docstring_parameters.py::test_docstring_parameters
3501
22.42s call     mne/tests/test_morph.py::test_volume_source_morph_round_trip[sample-fsaverage-7.0-7.4-float-True]
3502
22.39s call     mne/tests/test_source_space.py::test_setup_source_space_spacing[2]
3503
22.30s call     mne/tests/test_morph.py::test_volume_source_morph_round_trip[sample-sample-0.0-0.1-complex-False]
3504
20.70s call     mne/tests/test_morph.py::test_volume_source_morph_round_trip[sample-fsaverage-10.0-10.4-float-False]
3505
20.08s call     mne/viz/tests/test_3d_mpl.py::test_plot_volume_source_estimates[glass_brain-s-None-2-None-want_p0-None]
3506
19.97s call     mne/viz/tests/test_3d_mpl.py::test_plot_volume_source_estimates[glass_brain-vec-None-1-init_p2-want_p2-None]
3507
18.87s call     mne/inverse_sparse/tests/test_mxne_inverse.py::test_mxne_inverse_standard[_pytest_param]
3508
17.93s call     mne/report/tests/test_report.py::test_full_report
3509
```
You think this is good enough?

agramfort:
Thx @hoechenberger !

Issue #9752, opened 2021-09-16
Error checking in Info - setting unsupported keys in  `mne.Info` or `mne.Info` attributes partially works and doesn't show any warnings


mscheltienne:
In my preprocessing pipeline, I'm setting some information about the subjects. One of the interesting features is the age, or the birth year.  But MNE only supports the birthday as a tuple (year, month, day), as per the documentation of[ `mne.Info`](https://mne.tools/stable/generated/mne.Info.html).

Sadly, I do not have the exact birthday, and instead of setting it as (year, month, 1), I figured I could set a different `birthyear` or `age` argument directly. 

```
raw = mne.io.read_raw_fif(...)
raw.info['subject_info']['birthyear'] = 1995
{'id': 65, 'his_id': '065', 'birthyear': 1995, 'sex': 1}
raw.save(...)
```

And it worked really nicely. Until I saved and reloaded the file.

```
raw = mne.io.read_raw_fif(...)
raw.info['subject_info']
{'id': 65, 'his_id': '065', 'sex': 1}
```

The issue is that this emits no warning, no error, and yet the added data is not saved to disk.

If the dictionary stored in `'subject_info'` is not used by anything, and is just here for user-defined variables (*is this the case?*) then how about dropping the limitation on what the keys can be?

larsoner:
We have some history of trying to make things read-only, see for example https://github.com/mne-tools/mne-python/issues/7816#issuecomment-633226726. I don't remember the full discussion but that comment has a bunch of links

mscheltienne:
@larsoner Thanks for the bedside reading.
In my opinion, MNE and Python have both come a long way since the `Info` structure has been designed, and maybe it could (and need) to be reworked and improved.

As an end-user, I am always surprised by the structure of `Info`.
- Why is it a class where you access the attributes as you would with a dictionary?
- Why not a class with the attributes accessible as `Info.attribute`?
- Why only `ch_names` is accessible as `Info.ch_names`?

Obviously, you do get used to this structure; but it doesn't make it convenient. Most (if not all) IDE completion providers will not always work with the dictionary syntax `Info['` while they will with a plain attribute one `Info.`.

e.g. in IPython console:

![Screenshot 2021-09-17 at 11 11 35](https://user-images.githubusercontent.com/73893616/133757376-59415c1a-fd37-413e-b1ea-babae3fa88db.png)

e.g in Spyder IDE: (Kite is not that smart, but still..)

[...]

PR #9757, opened 2021-09-17
Make Info inherit from Bunch.

mscheltienne:
As discussed in #9752; a quick test of what would break if we change Info from a dict to a Bunch.
I gave it a quick test and the `copy()` method fails. `copy.deepcopy()` doesn't bring the set attributes and raises `AttributeError: 'Info' object has no attribute 'ch_names'`. Let's see if anything else breaks. 

For the `copy()` method, one quick solution would be to leave the only property `ch_names` in the code as it is now, but I think it might bite us later on.

If not too much code breaks, let's put some more thoughts into  `Bunch` and how to copy it to make sure it will work as intended.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -163,6 +163,7 @@ Enhancements
 
 - Reading EDF files via :func:`mne.io.read_raw_edf` now can infer channel type from the signal label in the EDF header (:gh:`9694` by `Adam Li`_)
 
+- Add :func:`mne.gui.locate_ieeg` to locate intracranial electrode contacts from a CT, an MRI (with Freesurfer ``recon-all``) and the channel names from an :class:`mne.Info` object (:gh:`9586` by `Alex Rockhill`_)
 
 Bugs
 ~~~~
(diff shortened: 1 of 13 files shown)


mscheltienne:
Notes following those 2 very quick tests: it doesn't look to me like it's going to be a pain to add. 

### Bunch from MNE

- Breaks `Info.copy()` method; deepcopy doesn't copy attributes and since I removed the `.ch_names` property, copied Info are missing `Info.ch_names`.

### Bunch from scikit-learn:

- Breaks `AssertionError: assert 'FIFFV_SSS_JOB_NOTHING' in ['__class__', '__contains__', '__delattr__', '__delitem__', '__dict__', '__dir__', ...]` in `mne.io.tests.test_constants`
- Breaks `Info(read_hdf5(fname_h5))` with `TypeError: __init__() takes 1 positional argument but 2 were given`. This is because I removed `*args` from `Info` initialization as I didn't see how it was fitting in.

I will pick this up next week!

mscheltienne:
Messed up CLI GitHub while reverting the 2 test changes.. Opening a new PR with dictionary keys exposed as attributes working for mne.Info as soon as possible.

PR #9758, opened 2021-09-17
Fix equalize event counts

drammock:
I hit a corner case today where `epochs.equalize_event_counts()` was failing because the smallest event_id had only one epoch in it, which caused `interp1d` to error out (we already handle the edge case where it has zero).  This fixes that.


agramfort (approved):
+1 for MRG after you add a tiny what's new entry

🙏

agramfort (approved):

agramfort:
thx @drammock

PR #9760, opened 2021-09-19
MRG, ENH: Use context manager, cleaner deps

larsoner:
Use a context manager for closing EDFlib's class -- at some point we should make an upstream PR to add `__enter__` and `__exit__` to their class. I started out by making a private function to minimize the diff, but it ended up needing so many args/kwargs, it seemed better to just indent the whole thing. So this PR does that, and other than removing all `.close()`s (which are no longer needed because the context manager should take care of it), it removes `pooch` from explicit pip installation because it's in `requirements_testing.txt` already.

### DIFF
--- a/mne/export/_edf.py
+++ b/mne/export/_edf.py
@@ -3,6 +3,7 @@
 #
 # License: BSD-3-Clause
 
+from contextlib import contextmanager
 import numpy as np
 
 from ..utils import _check_edflib_installed, warn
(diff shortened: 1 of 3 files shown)


adam2392:
@teuniz is it possible for me to make a PR to the EDFLIB-python here to have a enter and exit function to improve automatically closing of the file if an error occurs?

larsoner:
Either way we'll want to wait for a new release, and this code prepares us for it. Once the upstream library has context manager support, we can just remove the `_auto_close(...)` part and it will automatically work. So I say we merge this as is and eventually transition

drammock (approved):

PR #9761, opened 2021-09-20
[ENH] Allow custom r-peaks indices to be passed to `create_ecg_epochs`

DominiqueMakowski:

Currently, the `create_ecg_epochs()` convenience function runs the default built-in r-peak detection algorithm. However, many methods for detecting [R-peaks exist](https://neurokit2.readthedocs.io/en/latest/functions.html#neurokit2.ecg.ecg_peaks), that could be more appropriate in some cases. Users might want (need) to detect R-peaks externally, and then use this information for `create_ecg_epochs`. Alternatively, some devices only record r-peaks (instead of the ECG/PPG raw signal), so users might have this information without having an ECG channel. 

This small PR allows a list of r-peaks to be passed to `create_ecg_epochs` to be used instead of running mne's r-peak detector.

### DIFF
--- a/mne/preprocessing/ecg.py
+++ b/mne/preprocessing/ecg.py
@@ -277,7 +277,7 @@ def _get_ecg_channel_index(ch_name, inst):
 def create_ecg_epochs(raw, ch_name=None, event_id=999, picks=None, tmin=-0.5,
                       tmax=0.5, l_freq=8, h_freq=16, reject=None, flat=None,
                       baseline=None, preload=True, keep_ecg=False,
-                      reject_by_annotation=True, decim=1, verbose=None):
+                      reject_by_annotation=True, decim=1, rpeaks=None, verbose=None):
     """Conveniently generate epochs around ECG artifact events.
 
     %(create_ecg_epochs)s
@@ -309,6 +309,9 @@ def create_ecg_epochs(raw, ch_name=None, event_id=999, picks=None, tmin=-0.5,
         When ECG is synthetically created (after picking), should it be added
         to the epochs? Must be False when synthetic channel is not used.
         Defaults to False.
+    rpeaks : list
+        A list of R-peak location indices. If `None`, will attempt to find them
+        using MNE's built-in detector.
     %(reject_by_annotation_epochs)s
 
         .. versionadded:: 0.14.0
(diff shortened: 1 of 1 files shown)


larsoner:
I am -0.5 on this and +0.5 on telling people to use `raw_filt = raw.copy().filter(...)` + `mne.Epochs(raw, custom_peak_events, tmin=..., tmax=...)` or so directly if they want this sort of thing.

DominiqueMakowski:
> on telling people to use...

Yeah that's what I am currently doing, but I thought it's an easy and convenient addition... feel free to close then :)

larsoner:
Let's see what others think... it might be a bit annoying to force people to pass the filtering and tmin/tmax params, etc., thus justifying the simplification of being able to pass peaks directly.

cbrnr:
I agree with @larsoner, but I think we should explicitly mention this in the `create_ecg_epochs()` docstring. Also, @DominiqueMakowski we have a new R peak detector package with very good performance in case you didn't already know: https://sleepecg.readthedocs.io/en/stable/.

drammock:
to me this seems like a clean, useful addition with fairly minimal maintenance overhead.  I'd say it's worth it to save users the pain of manually filtering and epoching, especially if there are some systems that just return the R-peak times and not the whole ECG waveform. Also, since we already have sensible defaults chosen for the filtering and epoching steps, with this PR they get to use our defaults / are less likely to filter/epoch incorrectly.

cbrnr:
Filtering is specific to the detector, I don't quite understand why this should be an essential part when using a custom detector function. So this reduces the code to one line. But if you still think it makes things easier for users I'm OK with adding a new parameter.

drammock:
> Filtering is specific to the detector, I don't quite understand why this should be an essential part when using a custom detector function. So this reduces the code to one line. But if you still think it makes things easier for users I'm OK with adding a new parameter.

Filtering is also done to the raw instance before creating an SSP projector for the heartbeat.  But (I realize now) that is done in `compute_proj_ecg` not in `create_ecg_epochs`... so yeah, this PR doesn't actually help with that.  So I'm OK with just adding to the docstring a short example of what to do if you have pre-computed R-peak times.

PR #9762, opened 2021-09-20
[MRG] Move files in config to a python file

adam2392:
#### Reference issue
First part of #9736 

Moves dataset_checksums file into a dictionary of dictionaries that define each dataset uniquely.

#### Additional information
The next phase would be to:

- create a public facing function: `mne.datasets.fetch_dataset()`, which takes in a dictionary of parameters that must have `archive_name`, `folder_name`, `url`, `hash` defined and optionally can pass in `config_key` and `token`
- refactor `_data_path` to accept these parameters instead
- refactor functions that call `_data_path` internally to leverage `MNE_DATASETS` to pass in the necessary parameters to `_data_path`.


mne/datasets/config.py
```diff
@@ -95,123 +95,273 @@
...
-    ssvep='MNE_DATASETS_SSVEP_PATH',
-    testing='MNE_DATASETS_TESTING_PATH',
-    visual_92_categories='MNE_DATASETS_VISUAL_92_CATEGORIES_PATH',
-)
-assert set(ARCHIVE_NAMES) == set(URLS)
+# Each dataset should contain minimally:
+# - archive_name : the name of the compressed file that is downloaded
+# - hash : the md5 hash of the file
+# - url : the web url at which the file is downloaded
+# - folder_name : the folder in which the files are uncompressed to
```
drammock, line 102:
```suggestion
# - folder_name : the subfolder within the MNE data folder in which to save and
#                 uncompress (if needed) the file(s)
```

mne/datasets/config.py
```diff
@@ -95,123 +95,273 @@
...
+# - config_key : the environment variable for that dataset
```
drammock, line 103:
```suggestion
# - config_key : key to use with `mne.set_config` to store the on-disk location
#                of the downloaded dataset (ex: "MNE_DATASETS_EEGBCI_PATH").
```

mne/datasets/config.py
```diff
@@ -95,123 +95,273 @@
...
+# - (optional) token : An authorization token for private datasets.
+# include here the name of the zipped file and the ' : 'md5 hash
+# Note: other hashing algorithms can be enforced
```
drammock, line 106:
```suggestion
```

mne/datasets/config.py
```diff
@@ -95,123 +95,273 @@
...
-    visual_92_categories='MNE_DATASETS_VISUAL_92_CATEGORIES_PATH',
-)
-assert set(ARCHIVE_NAMES) == set(URLS)
+# Each dataset should contain minimally:
+# - archive_name : the name of the compressed file that is downloaded
+# - hash : the md5 hash of the file
+# - url : the web url at which the file is downloaded
+# - folder_name : the folder in which the files are uncompressed to
+# - config_key : the environment variable for that dataset
+# - (optional) token : An authorization token for private datasets.
```
drammock, line 104:
```suggestion
# - (optional) token : An authorization token, if one is needed to access the
#                      download URL.
```

mne/datasets/config.py
```diff
@@ -95,123 +95,273 @@
...
+rh_aparc_sub_annot = dict(
+    archive_name='rh.aparc_sub.annot',
+    hash='md5:dd6464db8e7762d969fc1d8087cd211b'
+)
+
+# below, do **not** include the release
+# number in the names of the testing or
+# misc datasets (it gets done automatically
+# in mne/datasets/utils.py). Only update
+# the ' : 'md5 hashes as needed.
```
drammock, line 318:
```suggestion
```
seems like this warning is no longer necessary; it should be obvious not to change the variable name, and it is clear from the URL that the version will be `.format()`ted in later.

mne/datasets/utils.py
```diff
@@ -207,11 +206,16 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
     # import pooch library for handling the dataset downloading
     pooch = _soft_import('pooch', 'dataset downloading', strict=True)
 
+    # extract configuration parameters
+    this_dataset = MNE_DATASETS[name]
+    config_key = this_dataset['config_key']
+    folder_name = this_dataset['folder_name']
```
drammock, line 212: you redefine `this_dataset` below, better to use different var names

```suggestion
    dataset_dict = MNE_DATASETS[name]
    config_key = dataset_dict['config_key']
    folder_name = dataset_dict['folder_name']
```

mne/datasets/utils.py
```diff
@@ -314,32 +340,33 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
...
+        registry = op.join(tmpdir, 'dataset_checksums.txt')
+
+        # write the md5 hashes
+        with open(registry, 'w') as fout:
+            for archive_name, dataset_hash in pooch_hash_mapping.items():
+                # write to the checksums file registry
+                fout.write(f'{archive_name} {dataset_hash}\n')
+
+        # load the checksum registry
+        fetcher.load_registry(registry)
```
drammock, line 357: this is actually not necessary.  The registry can be passed as a `dict` when the fetcher is instantiated: `fetcher = pooch.create(..., registry=pooch_hash_mapping)`

mne/datasets/utils.py
```diff
@@ -314,32 +340,33 @@ def _data_path(path=None, force_update=False, update_path=True, download=True,
         version=None,   # Data versioning is decoupled from MNE-Python version.
         registry=None,  # Registry is loaded from file, below.
         urls=pooch_urls,
-        retry_if_failed=2  # 2 retries = 3 total attempts
+        retry_if_failed=2  # 2 retries = 3 total attempts,
```
drammock, line 343:
```suggestion
        retry_if_failed=2  # 2 retries = 3 total attempts
```

drammock (changes requested):
looking good.  Mostly I'm suggesting clarifying the code comments, but one important improvement to make is to avoid writing the temporary file altogether.

mne/datasets/utils.py
adam2392: Ah nice, that's great.

adam2392:
Kay I think this is good to go now @drammock 

FYI: #9763 will have the downstream PR that starts to look at generalizing the dataset fetcher. Trying things out, I realized we also need to pull the `processor` for each dataset out. E.g. the Unzip/UnTar/etc.

drammock:
> the downstream PR that starts to look at generalizing the dataset fetcher. Trying things out, I realized we also need to pull the `processor` for each dataset out. E.g. the Unzip/UnTar/etc.

Most of the datasets work with the default values of the pooch processors, so I suggest:
```
processor : None | "zip" | "tar" | instance of pooch.Unzip | instance of pooch.Untar
```
where `None` is for files that aren't compressed at all, and the string arguments will use our internally defined `pooch.Unzip` or `Untar` similar to how we do it in `datasets/utils.py` line ~280

adam2392:
> Most of the datasets work with the default values of the pooch processors, so _for the generic function_ I suggest:
> 
> ```
> processor : None | "zip" | "tar" | instance of pooch.Unzip | instance of pooch.Untar
> ```
> 
> where `None` is for files that aren't compressed at all, and the string arguments will use our internally defined `pooch.Unzip` or `Untar` similar to how we do it in `datasets/utils.py` line ~280

Sounds good. That looks good to me.

> For this PR, it would probably be good to pull out the processors into the individual dicts that define each dataset. I can push a commit to do that if you're not sure how best to do it.

How would we do this actually since the `extract_dir=path` relies on the path variable defined internally?

drammock:
> How would we do this actually since the `extract_dir=path` relies on the path variable defined internally?

Oops, you're right.  I was thinking that when the `path` variable is passed in as a param to `_data_path()` if it's already a string (not None) it is left untouched, see the logic in the `_get_path()` helper function.  But we're not doing that --- I forgot we're building the path from the `config_key` for the internal datasets.  So never mind that suggestion.

adam2392:
I think we need to then separate out `_get_path` to within `fetch_dataset` before calling `_data_path`. That way, the path must be explicitly defined before calling `_data_path`. Then the processor can be defined outside `_data_path` and already have the path it needs.

Probably best to save that for the next PR since that depends on creation of the generic public API function?

If so, this is good to go by me.

drammock (approved):

adam2392:
Looks like the azure pipeline failures are timing out. I don't think due to this PR at least?

drammock:
> Looks like the azure pipeline failures are timing out. I don't think due to this PR at least?

that's been happening for several days, I haven't had a chance to look into it yet

mne/datasets/config.py
```diff
@@ -95,123 +95,270 @@
...
+# - hash : the cryptographic hash type of the file followed by a colon and then
+#          the hash value (examples: "sha256:19uheid...", "md5:upodh2io...")
+# - url : URL from which the file can be downloaded
+# - folder_name : the subfolder within the MNE data folder in which to save and
+#                 uncompress (if needed) the file(s)
+# - config_key : key to use with `mne.set_config` to store the on-disk location
+#                of the downloaded dataset (ex: "MNE_DATASETS_EEGBCI_PATH").
+# - (optional) token : An authorization token, if one is needed to access the
+#                      download URL.
+fnirs_motor = dict(
```
agramfort, line 109: @adam2392 sorry I was slow to review here. I would suggest you write it as:

```
MNE_DATASETS = dict()

MNE_DATASETS['fnirs_motor'] = dict(
```

this way you don't need to replicate below the list of datasets so you won't forget one.

can you open a new PR?

PR #9768, opened 2021-09-22
Fix drop log

### LINKED ISSUE #9726: Drop log handling in sliced epochs not intuitive
#### Describe the bug
Using `drop_log_stats()` or `drop_log` is unintuitive when working with sliced epochs.

#### Steps to reproduce

MWE
```python
import mne
import numpy as np

np.random.seed(10)
n_epochs, n_channels, n_timepoints = 10, 4, 21
fake_data = 1e-6*np.random.random((n_epochs, n_channels, n_timepoints))

# Have the first three epochs exceed minmax in one channel
fake_data[0:3, 1, 0] = 200e-6

epo = mne.EpochsArray(fake_data, info)
# This should drop the first three epochs
epo.drop_bad(reject=dict(misc=100e-6))

# Prints 30% as expected
print(f"I expect 30% and am: {epo.drop_log_stats()}%")

# select only the last epoch
subepos = epo[-1:]
# Prints 75%
print(f"I expect 0% or maybe 30% but am: {subepos.drop_log_stats()}")

# Drop log contains IGNORED only for events that were not dropped initially
print(subepos.drop_log)
# Output is:
# (('ch_1',), ('ch_1',), ('ch_1',), ('IGNORED',), ('IGNORED',), ('IGNORED',), ('IGNORED',), ('IGNORED',), ('IGNORED',), ())
```

#### Expected results
I would either expect the drop log stats to know the initial ratio of drops, i.e. 30%, or after slicing report only the dropped epochs in the slice, i.e. 0%. However, calculating the latter is hard for non-trivial drop/slice combinations.

#### Actual results
MNE remembers how many epochs were dropped initially, then ignores the epochs marked as IGNORED and calculates the drop log stats to be 75%, as 3 were dropped and the slice contains 1 non-dropped epoch.

I suspect that when doing slices of epochs, the `__getitem__` implementation only inserts `('IGNORED')` in the drop log when there is no entry yet.

#### Additional information
```
Platform:      Linux-5.11.0-34-generic-x86_64-with-glibc2.29
Python:        3.8.10 (default, Jun  2 2021, 10:49:15)  [GCC 9.4.0]
Executable:    /home/jan/dev/python/spot/spot_venv/bin/python
CPU:           x86_64: 8 cores
mne:           0.24.dev0
numpy:         1.21.2 {blas=openblas, lapack=openblas}
scipy:         1.7.1
matplotlib:    3.4.3 {backend=module://backend_interagg}
sklearn:       0.24.2
numba:         Not found
nibabel:       3.2.1
nilearn:       Not found
dipy:          Not found
cupy:          Not found
pandas:        1.3.2
mayavi:        Not found
pyvista:       Not found
vtk:           Not found
```

jsosulski:
#### Reference issue
Fixes #9726 

Instead of adding `'IGNORED'` only to empty tuples in the drop log, this will append it to all drop log entries when slicing an epoch object to a subset.

#### Additional information
I fixed one epoch unit test, which was straight forward, however the second one that fails is `test_drop_epochs_mult` where there is a section I don't really get:

https://github.com/mne-tools/mne-python/blob/6c0006dfc73588927936b95a1ae364556a55fbc0/mne/tests/test_epochs.py#L2367-L2373

There are several comparisons to empty lists `[]` which, as far as I understand the drop log correctly, can never be true/false as the entries can only be empty tuples, not empty lists. I could take a stab at fixing this unit test as well, once I understand it correctly.

Also currently I assume that a drop reason should be present only once in a drop log entry, however, one could argue that multiples are okay, to e.g. reconstruct the order of subset selections if for some reason someone would be interested in this.

OT: Probably too big of a change, but using `'UNSELECTED'` or `'DESELECTED'` as a drop reason when slicing epochs would make the drop log explain more precisely why the epochs were dropped. One could then add `'IGNORED'` as a special case that incorporates `'UNSELECTED'` and other common reasons, or just use an enum.

### DIFF
--- a/mne/tests/test_epochs.py
+++ b/mne/tests/test_epochs.py
@@ -2349,6 +2349,9 @@ def test_drop_epochs():
     assert_array_equal(events[epochs[3:].selection], events1[[5, 6]])
     assert_array_equal(events[epochs['1'].selection], events1[[0, 1, 3, 5, 6]])
 
+    # After subset selection, drop log should be 0
+    assert_equal(epochs[0:1].drop_log_stats(), 0)
+
 
 def test_drop_epochs_mult():
     """Test that subselecting epochs or making less epochs is equivalent."""
--- a/mne/tests/test_epochs.py
+++ b/mne/tests/test_epochs.py
@@ -1957,7 +1957,7 @@ def test_epoch_eq():
     old_shapes = [epochs[key].events.shape[0] for key in ['a', 'b', 'c', 'd']]
     epochs.equalize_event_counts(['a', 'b'])
     # undo the eq logging
-    drop_log2 = tuple(() if log == ('EQUALIZED_COUNT',) else log
+    drop_log2 = tuple(log[:-1] if 'EQUALIZED_COUNT' in log else log
                       for log in epochs.drop_log)
     assert_equal(drop_log1, drop_log2)
 
--- a/mne/utils/mixin.py
+++ b/mne/utils/mixin.py
@@ -198,8 +198,10 @@ class GetEpochsMixin(object):
             key_selection = inst.selection[select]
             drop_log = list(inst.drop_log)
             if reason is not None:
-                for k in np.setdiff1d(inst.selection, key_selection):
-                    drop_log[k] = (reason,)
+                select_all = list(range(len(inst.drop_log)))
+                for k in np.setdiff1d(select_all, key_selection):
+                    if reason not in drop_log[k]:
+                        drop_log[k] += (reason,)
             inst.drop_log = tuple(drop_log)
             inst.selection = key_selection
             del drop_log


larsoner:
> There are several comparisons to empty lists [] which, as far as I understand the drop log correctly, can never be true/false as the entries can only be empty tuples, not empty lists.

Yes this looks like a bug to me

> Also currently I assume that a drop reason should be present only once in a drop log entry

Agreed, I wouldn't bother appending more and more `'IGNORED'`s with each new `__getitem__` call, which it sounds like would be the resulting effect...

> OT: Probably too big of a change

Agreed we shouldn't go this route unless we really need to / have strong motivating use cases at some point.

jsosulski:
After some detachment of this issue, I think I was trying to misuse the droplog for my purposes. `epo.selection` should have all the info I need and I could subselect the droplog that is passed to `_drop_log_stats`.

Do you prefer I force push my changes (needed to rebase locally) to this PR, or close it and create a fresh one?

larsoner:
Force push is better I think, keeps all discussion in one place

jsosulski:
Nevermind, I cant figure out a way to make this work without re-structuring the drop_log / selection / events setup, because lets say you have first five epochs as target events, and the next five epoch as nontargets, and the last 2 of each class are bad:

```python
epo  # 5 target epochs, 5 nontarget epochs
epo.drop_bad()  # 3 target epochs, 3 nontarget epochs
epo.selection  #  is now [0, 1 ,2, 5, 6, 7]
len(epo.events)  # is now 6
epo.drop_log  # is now ((), (), (), ('d'), ('d'), (), (), (), ('d'), ('d'))
epo_t = epo['target']  # 3 target epochs, 0 nontarget epochs
epo_t.selection  # is now [0, 1, 2]
# Current MNE Version:
epo_t.drop_log  # is now ((), (), (), ('d'), ('d'), ('IGNORED'), ('IGNORED'), ('IGNORED'), ('d'), ('d'))
# This PR as of now:
epo_t.drop_log  # is now ((), (), (), ('IGNORED','d'), ('IGNORED','d'), ('IGNORED'), ('IGNORED'), ('IGNORED'), ('IGNORED','d'), ('IGNORED','d'))
```

A workaround would be to call `drop_bad()` only after all subset selections have been done.

What I would need is to have the event-wise information for all ('d') drop log entries, but the events have been dropped as far as I can tell, and I can no longer conclude which of the drops belonged to which event. Or am I missing something?

The only solution I can think of is to have the original events array somewhere in the epochs object available. Without this information I think only having `drop_log_stats()` reset to 0 after each Epochs subselection is feasible.

larsoner:
> The only solution I can think of is to have the original events array somewhere in the epochs object available

For this specific use case could you make use of `write_events` to write a little companion `-eve.fif` or `-eve.txt` file alongside your epochs? This could be a suitable workaround for now, and if lots of people request this feature as well we could start writing the original events, too. (We try not to add too many things to objects, so it would be nice to see if doing so would benefit other use cases / users rather than adding it here/now.)

> Without this information I think only having drop_log_stats() reset to 0 after each Epochs subselection is feasible.

I could see a `epochs.reset_selection()` or `reset_drop_log` being generally useful and not too hard to support. Would this be another different way to get the behavior/information you need? Basically it would make any given Epochs have `epochs.selection = np.arange(len(epochs.events))` and `epochs.drop_log = ((),) * len(epochs.events)`.

jsosulski:
> For this specific use case could you make use of `write_events` to write a little companion `-eve.fif` or `-eve.txt` file alongside your epochs?

Sure I can do that. I think I could also call `_drop_log_stats` directly, instead of the class method, where I pass my own drop log.

A `reset_selection` method would enable at least the behaviour where a subset selection resets the drop log stats to 0 and does not output spurious values. However, then #9726 would only be fixed when the user manually calls the reset method. Alternatively, I could make drop_log_stats emit a warning when drop_log entries outside the selection do not have 'IGNORED', e.g., `any(ignore not in epochs.drop_log[~epochs.selection])`.

larsoner:
> I think I could also call _drop_log_stats directly, instead of the class method, where I pass my own drop log.

It's not a good idea to use private functions, if this is the preferred solution we'd want to make it public for you.

I'm not sure what the best solution is here, what do you think allows us to make the smallest API changes / expansion while still enabling you to do what you need (and also preventing other users from being confused or making mistakes) @jsosulski ?

jsosulski:
It is no urgent need for my use case, as I can just call `drop` after doing epoch subselection. Then the drop log stats work as intended.
To fix the very wrong drop log stats with few API changes, I would propose to use a fix like the one in this PR, as this changes the drop log stats to slightly wrong 0.

After thinking about it a bit more, I think the core issues is that `epoch.selection` and `epoch.drop_log` have overlapping responsibilities. Drop log is set during selection operations and selection is set during dropping operations.
If `epoch.selection` is only set by indexing operations, i.e., `__getitem__`, and `epoch.drop_log` by dropping operations, it would be straight forward to calculate the drop log stats correctly. And the current behaviour that `epoch.selection` is the 'ground truth' of selected epochs is simply obtained by having something (in pseudocode) like `epoch.selection & ~epoch.drop_log`

Then drop log stats could then simply use `epoch.drop_log[epoch.selection]`

Anyway this is a imo larger change, and to preserve backwards compatibility I would instead add, e.g., `epoch.selected_epochs` and remove the `epoch.selection` attribute and replace it with a `@property` that automatically returns `epoch.selected_epochs & ~epoch.drop_log` which would result in the current output of `epoch.selection`

larsoner:
> Anyway this is a imo larger change, and to preserve backwards compatibility I would instead add, e.g., epoch.selected_epochs and remove the epoch.selection attribute and replace it with a @property that automatically returns epoch.selected_epochs & ~epoch.drop_log which would result in the current output of epoch.selection

I think there are many (years worth) `epochs` in the wild that have `epochs.selection` already, which we would have to figure out how to deal with. And we'd need to add a new, different constart to the FIF spec for this I think. So I agree it's a pretty big change. Not sure it's worth the effort if there are suitable workarounds.

> To fix the very wrong drop log stats with few API changes, I would propose to use a fix like the one in this PR, as this changes the drop log stats to slightly wrong 0.

Okay, want to update the PR this way? I don't fully follow the case(s) that have changed, I'm hoping that the updated tests will make it clearer. 

Another complementary option would be to add a short section in the tutorial (or example) docs about how to get the functionality and information you need using the current way of doing things (or way of doing things with the fix in this PR).

PR #9770, opened 2021-09-22
[ENH, MRG] Make directional labels more visible

alexrockhill:
The labels were too small and were hard to see on a light background so this makes them larger with a black stroke.


larsoner (approved):
Should we add something similar to the related code here?

https://github.com/mne-tools/mne-python/blob/main/mne/viz/misc.py#L410-L428

larsoner:
I opened  #9773, so actually no need to fix it here. We should get it for free whenever we refactor. Thanks @alexrockhill !

PR #9771, opened 2021-09-22
MRG, BUG: Fix bug with dipole raw data

larsoner:
I ran into a raw dataset in the wild that had `DIPOLE` coil type for raw data because the system had phantom dipole activations as a channel type (cool!). Our raw data plotting couldn't handle it, so I fixed it and added some consistency checks that hopefully make it so all channel types are covered now.

I also made it so that in `_html_repr_` of `Raw`, the `Filenames` are joined by `<br>` rather than `, `, as it makes it much more readable for long filenames (these are not so long, but many are):

Screen Shot 2021-09-22 at 8 49 56 AM


larsoner:
It looks like some of these tests are designed / expect that some channels have no units. @drammock do you think I should revert adding all channel types to `'units'` (e.g., `STIM`, the current cause of test failures)?

agramfort (approved):

drammock:
hmm...  I wasn't involved in implementing/approving this one (#9136).  Looking at the docstring for `raw.get_data(..., units=???)` I see this:

```
    If a string, must be a
    sub-multiple of SI units that will be used to scale the data from
    all channels of the type associated with that unit. This only works
    if the data contains one channel type that has a unit (unitless
    channel types are left unchanged) [...] if a
    dictionary is provided, keys must be channel types, and values must
    be units to scale the data of that channel type to.
```

So, options are:

1. change the code so that "AU" is recognized as a dimensionless unit (and possibly also "Z")
2. change the test to `raw.get_data(..., units=dict(eeg='V'))` or so

@HanBnrd FYI, the issue is that this test line: `raw.get_data(picks=['eeg', 'stim'], units='V')` no longer works, because previously dimensionless channels like "stim" now have a dimension ("AU" for arbitrary units). any opinion here?

larsoner:
> @HanBnrd FYI, the issue is that this test line: raw.get_data(picks=['eeg', 'stim'], units='V') no longer works, because previously dimensionless channels like "stim" now have a dimension ("AU" for arbitrary units). any opinion here?

Two other options, since this tradeoff is not so nice:

1. Keep a list of dimensionless units and special-case/ignore these like we did before in functions that need it, like `get_data(..., units=...)`
2. Remove my consistency checks for `units`, `scalings`, etc., and treat missing-from-units as dimensionless automatically.

I'm not sure which is better. In (1) we have to modify some code to look for dimensionless-ness. I guess this must have previously been done implicitly with some check like `ch_type in units`. So maybe that's the easiest way to go, since this would just (conceptually at least) need to change to `if ch_type not in dimensionless_units` or something.

drammock:
I think your option 1 and my option 1 are basically the same idea, and that's the approach I prefer.

HanBnrd:
Hi both,

If you would like to implement option 1 of @drammock, there is a bit more than handling *AU* since the stim appears to be in *V*.

Are we sure we don't want to convert stim? If we are sure, I think changing `ch_types = self.get_channel_types(picks=picks, unique=True)` into `ch_types = self.get_channel_types(picks=picks, unique=True, only_data_chs=True)` in [here](https://github.com/mne-tools/mne-python/pull/9136/files#diff-2441de960ad1e76f767c97ef4c93782cbc244cb073340510404f971e6a49359bR860) could solve the issue, since channels for which @larsoner added units are not in the *data channel* list I believe. So I'd recommend doing that instead.
(and also changing the comment to be clearer `# Check that there is only one channel type` int `# Check that there is only one data channel type`)

Regardless of the fix, is that something we would need to do for Epochs as well? (#9553)

CC @sappelhoff who worked on this as well

larsoner:
Okay I went with what I think is the least invasive change here. @drammock I ended up going with something like (2) because it required no adjustments to other code / was the path of least resistance. I think we can revisit something more complete if as we hit use cases that need rethinking about units, etc.

agramfort (approved):

drammock (approved):

Issue #9776, opened 2021-09-23
has_dataset should rely on `dataset_params` in mne.datasets.utils


adam2392:
#### Describe the new feature or enhancement
has_dataset is a public function, and therefore should rely on the same API as fetch_dataset. 

That is, feeding in a list of `dataset_params` rather then the `name`.

This would be an add-on to #9763 

#### Describe your proposed implementation
Refactor all unit tests and code that call `has_dataset`

PR #9781, opened 2021-09-26
[MRG, MAINT] Refactor "has_dataset" to be the same semantics as "fetch_dataset"

adam2392:
#### Reference issue
Fixes: #9776 

Refactors the has_dataset function which is "public", but not included in the docs API.

#### Additional information
Idk if we want to add `has_dataset` to the API docs?


adam2392:
Oh okay this PR reveals that it is probably better to add `dataset_name` to all the MNE_DATASETS dictionary such as:

```
MNE_DATASETS['sample'] = {
   'dataset_name': 'sample',
...
}
```

mne/datasets/utils.py
```diff
@@ -197,27 +197,26 @@ def _get_version(name):
                          return_version=True)[1]
 
 
-def has_dataset(name):
+def has_dataset(dataset_params):
```
agramfort, line 200: @adam2392 this is an API change of a public function without deprecation.

why not allowing name to be a str or a dict and if str is passed it's understood as the key of MNE_DATASETS

so API would say simple for simplest needs

adam2392: Ah okay, wasn't sure if it required deprecation cuz it wasn't in the API docs. I'll adjust this behavior.

agramfort (approved):

mne/datasets/utils.py
```diff
@@ -202,10 +202,14 @@ def has_dataset(name):
...
     ----------
-    name : str
-        The dataset name.
-        For brainstorm datasets, should be formatted like
-        "brainstorm.bst_raw".
+    name : str | dict
+        If it is a string, then it refers to an internal MNE dataset
+        name (e.g. 'sample' dataset). For brainstorm datasets, should
+        be formatted like "brainstorm.bst_raw".
+        If it is a a dictionary, then it is a dictionary of dataset
```
drammock, line 209:
```suggestion
        If it is a dictionary, then it is a dictionary of dataset
```

mne/datasets/utils.py
```diff
@@ -202,10 +202,14 @@ def has_dataset(name):
...
     Parameters
     ----------
-    name : str
-        The dataset name.
-        For brainstorm datasets, should be formatted like
-        "brainstorm.bst_raw".
+    name : str | dict
+        If it is a string, then it refers to an internal MNE dataset
+        name (e.g. 'sample' dataset). For brainstorm datasets, should
+        be formatted like "brainstorm.bst_raw".
```
drammock, line 208: this docstring can't be correct: 

```py
In [1]: import mne
In [2]: mne.datasets.utils.has_dataset('brainstorm.bst_raw')
KeyError                                  Traceback (most recent call last)
<ipython-input-2-ad99ed2b884a> in <module>
----> 1 mne.datasets.utils.has_dataset('brainstorm.bst_raw')

/opt/mne/python/mne/datasets/utils.py in has_dataset(name)
    224     else:
    225         dataset_name = 'spm' if name == 'spm_face' else name
--> 226         dataset_params = MNE_DATASETS[dataset_name]
    227         dataset_params['dataset_name'] = dataset_name
    228 

KeyError: 'brainstorm.bst_raw'
```

mne/datasets/utils.py
```diff
@@ -214,20 +218,25 @@ def has_dataset(name):
...
-    dataset_params = MNE_DATASETS[name]
-    dataset_params['dataset_name'] = name
-    config_key = MNE_DATASETS[name]['config_key']
+    if isinstance(name, dict):
+        dataset_name = name['dataset_name']
+        dataset_params = name
+    else:
+        dataset_name = 'spm' if name == 'spm_face' else name
+        dataset_params = MNE_DATASETS[dataset_name]
+        dataset_params['dataset_name'] = dataset_name
```
drammock, line 227: I find myself wondering why we don't just hard-code the dataset name into those dicts in the first place. IIRC there was some reason to want them there anyway (for the downloading?)

drammock (changes requested):

drammock:
One other comment:  I think for these dataset-fetching PRs you need to do some more aggressive local testing @adam2392, and not just rely on the CIs.  As we've already seen, CircleCI's caching has allowed some bugs to slip through on recent PRs.  Consider deleting your local `MNE_DATA` folder and doing an overnight `_download_all_example_data` as a sanity check

mne/datasets/utils.py
adam2392: Yea for the downloading, we would want them there. It's up to you all. I had removed it because then it's just cleaner config file, but I can add it back in to get rid of lines like these.

adam2392:
> One other comment: I think for these dataset-fetching PRs you need to do some more aggressive local testing @adam2392, and not just rely on the CIs. As we've already seen, CircleCI's caching has allowed some bugs to slip through on recent PRs. Consider deleting your local `MNE_DATA` folder and doing an overnight `_download_all_example_data` as a sanity check

Sure sounds good. I'll try it rn and make sure everything downloads.

drammock:
looking through all the usages of `has_dataset` now, I find 3 situations:

1. cases where `has_dataset` is used in a crufty way (a partial func like `has_erp_core_data` is defined but never used anywhere)
    - fieldtrip_cmc, erp_core, mtrf, bst_auditory, bst_phantom_ctf, bst_phantom_elekta, bst_resting 

2. cases where `has_dataset` is used only to create a public API entry (e.g., `mne.datsets.fnirs_motor.has_fnirs_motor_data()`)
    - epilepsy_ecog, fnirs_motor, multimodal, opm, phantom_4dbti, refmeg_noise, sample, somato, ssvep

3. cases where `has_dataset` is used more extensively, e.g., for datasets used in our tests.
    - bst_raw, testing, spm

I've pushed a commit that removes code described in (1).  My personal feeling is that all the cases in (2) are obsolete and should be deprecated now that we have a generic `has_dataset()` function: instead of calling `mne.datasets.fnirs_motor.has_fnirs_motor_data()` users can instead call `mne.datasets.has_dataset('fnirs_motor').  @larsoner @agramfort WDYT about deprecating in this way?

agramfort:
+100 !

>

larsoner:
@drammock feel free to merge if you're happy

drammock:
I snuck in some docstring fixes for `fetch_dataset` too.  Good to merge once Circle is green.

Issue #9747, opened 2021-09-16
make_fixed_length_epochs() should allow for custom specification of the event ID


hoechenberger:
Currently, `make_fixed_length_epochs()` assigns all events the ID `1`.

`make_fixed_length_events()`, on the other hand, already supports user-defined IDs.

It would be good to add a parameter to `make_fixed_length_epochs()`, allowing users to pass a custom event ID.

This was brought up on the forum, https://mne.discourse.group/t/fixed-epochs-and-fixed-events/3702

mscheltienne:
I wanted to quickly fix this as I might be using it soon as well but it's not as straightforward as I thought.

If we look at the discourse post, the user has a similar RAW to mine: multiple events of different lengths. In my case, I have an alternation of regulation phase (16s) and of non-regulation phase (8s). The beginning of each phase is marked with an event, 5 for regulation and 6 for non-regulation. The display on MNE browser with each event marked as an annotation is:

![image](https://user-images.githubusercontent.com/73893616/134648957-572b4e35-4430-4b73-a974-7983bcd4aaaa.png)

<details open>
<summary>MNE event array</summary>
<be>

```
mne.find_events(raw, 'TRIGGER')
array([[  5528,      0,      6],     # first non-regulation phase is longer, don't mind it for this example.
          [ 13309,      0,      5],
          [ 21527,      0,      6],
          [ 25699,      0,      5],
          [ 33965,      0,      6],
          [ 38088,      0,      5],
          [ 46357,      0,      6],
          [ 50478,      0,      5],
          [ 58785,      0,      6],
          [ 62929,      0,      5],
          [ 71182,      0,      6],
          [ 75315,      0,      5],
          [ 83511,      0,      6],
          [ 87646,      0,      5],
          [ 95866,      0,      6],
          [100035,      0,      5],
          [108247,      0,      6],
          [112425,      0,      5],
          [120693,      0,      6],
          [124814,      0,      5]], dtype=int64)
```

</details>

Now, from the way I understand the issue on the discourse forum; the user has something similar and wants to split each of the events in fix length epochs. If we consider my dataset as an example:
-> First regulation phase: 16 epochs with ID 5
-> First non-regulation phase: 8 epochs with ID 6
-> Second regulation phase: 16 more epochs with ID 5
-> Second non-regulation phase: 8 more epochs with ID 6
...

In my example, there is 10 alternation of regulation/non-regulation; thus leading to:
- 160 epochs with ID 5
[...]

PR #9782, opened 2021-09-27
Fix 9747 - Add ID arg to make_fixed_length_epochs to change the default value 1.

mscheltienne:
Tiny improvement fixing #9747.
I added a similar `id` argument for `make_fixed_length_epochs` as the one of `make_fixed_length_events`.

**Question:** to (more or less) match the argument order of `make_fixed_length_events`, I've placed `id` in second. 
Do you mind this type of change where user code with positional arguments might break, or do you always prefer to add new arguments to the end?

I'll change the order if needed.

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3679,8 +3679,6 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
     ----------
     raw : instance of Raw
         Raw data to divide into segments.
-    id : int
-        The id to use (default 1).
     duration : float
         Duration of each epoch in seconds. Defaults to 1.
     %(preload)s
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3670,9 +3670,9 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
 
 
 @verbose
-def make_fixed_length_epochs(raw, id=1, duration=1., preload=False,
+def make_fixed_length_epochs(raw, duration=1., preload=False,
                              reject_by_annotation=True, proj=True, overlap=0.,
-                             verbose=None):
+                             id=1, verbose=None):
     """Divide continuous raw data into equal-sized consecutive epochs.
 
     Parameters
@@ -3683,8 +3683,6 @@ def make_fixed_length_epochs(raw, id=1, duration=1., preload=False,
         The id to use (default 1).
     duration : float
         Duration of each epoch in seconds. Defaults to 1.
-
-        .. versionadded:: 0.24.0
     %(preload)s
     %(reject_by_annotation_epochs)s
 
@@ -3697,6 +3695,10 @@ def make_fixed_length_epochs(raw, id=1, duration=1., preload=False,
         ``0 <= overlap < duration``. Default is 0, i.e., no overlap.
 
         .. versionadded:: 0.23.0
+    id : int
+        The id to use (default 1).
+
+        .. versionadded:: 0.24.0
     %(verbose)s
 
     Returns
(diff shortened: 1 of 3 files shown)


mne/epochs.py
```diff
@@ -3670,7 +3670,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
 
 
 @verbose
-def make_fixed_length_epochs(raw, duration=1., preload=False,
+def make_fixed_length_epochs(raw, id=1, duration=1., preload=False,
```
agramfort, line 3673: this can break code. Please put the id param just before verbose

mne/epochs.py
```diff
@@ -3679,8 +3679,12 @@ def make_fixed_length_epochs(raw, duration=1., preload=False,
     ----------
     raw : instance of Raw
         Raw data to divide into segments.
+    id : int
+        The id to use (default 1).
```
agramfort, line 3683: add version added

mne/epochs.py
mscheltienne: Ok, so that was my question on the argument order guideline. I've changed it and place the new argument right before verbose.

mne/epochs.py
mscheltienne: Fixed. I made a mistake and placed it under the wrong argument.

agramfort (approved):
@mscheltienne can you just add a what's new entry?

mscheltienne:
@agramfort Done. I wasn't sure if small changes like this one should be added to the changelog.

agramfort (approved):

mscheltienne:
CI fails don't seem related to the changes.

hoechenberger:
I restarted CI, let's hope it all goes green now

mne/epochs.py
```diff
@@ -3672,7 +3672,7 @@ def average_movements(epochs, head_pos=None, orig_sfreq=None, picks=None,
 @verbose
 def make_fixed_length_epochs(raw, duration=1., preload=False,
                              reject_by_annotation=True, proj=True, overlap=0.,
-                             verbose=None):
+                             id=1, verbose=None):
```
hoechenberger, line 3675: I'd rather we used a more explicit name. Thinking of `event_id` – even though this is already used elsewhere in MNE with a different meaning.

hoechenberger (commented):
Otherwise looks good to me…

mscheltienne:
Remaining CI failure:

```
_________ test_make_inverse_operator_diag[testing_data-testing_data] __________
mne\minimum_norm\tests\test_inverse.py:775: in test_make_inverse_operator_diag
    _compare_inverses_approx(inverse_operator_diag, inv_op, evoked,
mne\minimum_norm\tests\test_inverse.py:182: in _compare_inverses_approx
    assert corr > ctol
E   assert 0.7766994175372354 > 0.99
```

Test passed before I modified the changelog in the documentation. Something unrelated changed on main?

larsoner:
> Something unrelated changed on main?

Something unrelated changed with the `--pre` builds of NumPy, almost certainly related to some new OpenBLAS bug. The failures we've seen with `cov.py` tests are probably related, too. Need to isolate and open an upstream bug at some point...

mne/epochs.py
mscheltienne: I'm in favor of `id` to match the argument of `mne.make_fixed_length_events`.

hoechenberger: > to match the argument of `make_fixed_length_events`

ok, so we have precedence. In this case, let's just stick with `id`.

hoechenberger (approved):
@larsoner I'm not sure whether you're still looking into the CI issue or whether we can just go ahead and merge? At any rate, LGTM!

mne/epochs.py
agramfort: yes agreed

hoechenberger:
Thanks @mscheltienne!

PR #9785, opened 2021-09-27
[PAPER, WIP, no MRG] Joss paper

alexrockhill:
This is the JOSS paper for intracranial EEG electrode location. It's still a work in progress at the moment but I am opening a PR so that @agramfort @larsoner and others can check that names and affiliations etc. are okay. I should have time to finish this pretty soon.


drammock:
@alexrockhill use draft mode on PRs that shouldn't get merged. It prevents accidents.

alexrockhill:
@drammock, I think `paper.md` has to be where it is in this PR for JOSS but would it be appropriate to have it live somewhere in `doc/` if it gets accepted just so it lives somewhere as markdown?

larsoner:
> would it be appropriate to have it live somewhere in doc/ if it gets accepted just so it lives somewhere as markdown?

Eventually will this be the entire paper text? I don't think this should live in the MNE-Python repository itself. Ideally it would live in a separate repo dedicated to the paper. Do they really require the paper text to live in the repo that houses the software? That seems like an onerous requirement actually. Even using `git submodule` would be annoying / we don't need the actual paper in our docs. At the end of the day once the paper is accepted, as far as the MNE-Python repo is concerned there should be a `:footcite:` and new entry in `doc/references.bib` and nothing more...

alexrockhill:
> > would it be appropriate to have it live somewhere in doc/ if it gets accepted just so it lives somewhere as markdown?
> 
> Eventually will this be the entire paper text? I don't think this should live in the MNE-Python repository itself. Ideally it would live in a separate repo dedicated to the paper. Do they really require the paper text to live in the repo that houses the software? That seems like an onerous requirement actually. Even using `git submodule` would be annoying / we don't need the actual paper in our docs. At the end of the day once the paper is accepted, as far as the MNE-Python repo is concerned there should be a `:footcite:` and new entry in `doc/references.bib` and nothing more...

No, it doesn't have to be in the repo, the branch is not required to be merged.

This was just to share the source code which is done in `mne-bids` but I'm fine not doing it, I thought it was a bit preferential since there are other mne-python papers and was unsure about doing it which is why I was checking.

larsoner:
> there are other mne-python papers and was unsure about doing it which is why I was checking.

Typically these papers have their own repos / source control of some sort, completely separate from MNE-Python itself. No need to mix changesets for a paper with changesets for code. I think the Frontiers paper was maybe an Overleaf document for example

agramfort:
JOSS works with .md files so no overleaf but I agree I would not put this
in mne-python repo.
Let's create a dedicated repo which can be on mne-tools org

>

alexrockhill:
If you have a chance to review a first draft @larsoner and @agramfort, that would be great

alexrockhill:
> repeating the request that this move to its own repo.  I see you using `skip` tags on some commits, but if you forget to do that it will waste CI cycles (which wastes electricity, which in most cases is generated with fossil fuel).

Good point, I think it has to be part of MNE-Python for the JOSS review though... that's how the paper preview and review works. The branch doesn't have to be merged, but after it's closed maybe that would be a good time to open the repo.

drammock:
I just looked at the JOSS submission guidelines, which say quite prominently

> Your paper (paper.md and BibTeX files, plus any figures) must be hosted in a Git-based repository together with your software (although they may be in a short-lived branch which is never merged with the default).

So it definitely needs to be part of either the MNE-Python repo, or possibly it's good enough if it just lives in your fork (?)

It would have been nice if you had simply pasted in that bit of the guidelines when this question originally arose...  your prior responses all said "I think..." and it would have saved us all some time if that guidance were clear from the get-go.

alexrockhill:
Yes in trying to sound like a nice guy I've lost a bit of the clarity, those are definitely the rules they lay out in the guidelines. I'm happy to open another repository but it will over complicate the review so we should do it after.

larsoner:
Sounds good to me

drammock:
based on my reading of the JOSS guidelines, the paper and the software must be in the same repo.  Therefore as I said above:

> I think the right approach is to work on the paper in a branch of your fork of MNE-Python

so basically do what you've been doing, but don't open a PR until you're done iterating w/ your co-authors and you're ready to submit to JOSS.  In the meantime, you should contact JOSS and ask whether it's OK if the paper is submitted directly from your fork, or if it must be on a (temporary) branch of the upstream repo.

drammock:
> But if there is no PR open then we can't use any GitHub web review features.

ah, good point.  In that case I'd do a PR from `alexrockhill/joss` to `alexrockhill/main`, yes.  And I'd do a draft PR so that you don't accidentally merge it and mess up your `main`.  When you and your co-authors are all satisfied, you can close that PR and open a new one into `mne-tools/joss` (we can create that branch if needed).

PR #9789, opened 2021-09-28
MRG, CI: Fix issue with Sphinx-Windows build

larsoner:
I think it's building docs for the stable version and not the dev version. This should tell us if it's the case.


larsoner:
Okay we're finally good to go here @drammock @agramfort . 

1. we had a legacy version of `mne` installed over the `python setup.py develop` / `pip install -e` version
2. I fixed a bunch of minor `linkcheck` bugs I came across during CI runs.
3. I added more verbose options to `mne sys_info` because they were useful in debugging CIs, and should be useful for the purposes of users' and developers' own debugging I think

agramfort (approved):

hoechenberger (approved):

hoechenberger:
Thanks, @larsoner!

Issue #9793, opened 2021-09-29
Averaging across frequencies - frequency band representation


adam2392:
#### Describe the new feature or enhancement
In `EpochsTFR` and `AverageTFR`, it would be nice to be able to average across frequencies and store the frequency band set name.

Taking the average over the frequency axis within a frequency band (e.g. delta would average power in frequencies 1-4 Hz) is a common workflow taken on by researchers.

#### Describe your proposed implementation
In the [current implementation](https://mne.tools/dev/generated/mne.time_frequency.AverageTFR.html), we would have to add:

- `average_freq(band_name)` function, which then assigns a `band_name` property
- `n_ave_freqs` which denotes the number of frequencies averaged. Analagous to `nave` for number of epochs averaging.
- `band_name` property storing the frequency band name

#### Describe possible alternatives
I currently do all this with custom numpy code, but it's very messy after awhile. + You can't really leverage the `AverageTFR.save()` functionalities, or have to hack around it.

#### Additional Information
https://mne.tools/dev/overview/roadmap.html seems to also point in the same direction where TFR analysis and data structures are still lacking in certain flexibility.

A workflow that I'm using rn is:
- use `make_fixed_length_epochs` to form an Epochs data structure
- compute `EpochsTFR` (e.g. using `tfr_multitaper`)
- save it for downstream processing
- read it back in with `read_tfrs`, and then format it back into a `(channels, freqs, times)` array, where the `times` axis is governed by the start time of each Epoch (stored in the `events` property). I then average across frequencies for e.g. the delta band.
- Now I can plot a spectrogram. 

By allowing averaging across frequencies, I can get rid of some intermediate storage.

larsoner:
Introducing a new `band_name` just in the case of averaging I'm not sure about -- classes having properties (or not) based on what methods have been makes everything more difficult. Instead, at least to start, can we just reassign `frequencies` to be the center of the band, and replace the entries of the frequencies averaged over with a single one for the average? So for example if you have 1, 2, 3, 4, ... Hz TFR data and you average over the frequencies from 1-4 you now get a single row in the data now at frequency 2.5 that replaces the rows and `frequencies` entries from 1-4?

PR #9794, opened 2021-09-29
[WIP] Enable TFR objects to average across frequencies

adam2392:
#### Reference issue
Fixes: #9793 

This demonstrates how the averaging over frequencies would occur.

#### Additional information
Lmk if this is along the lines of what you would expect?

### DIFF
--- a/mne/time_frequency/tfr.py
+++ b/mne/time_frequency/tfr.py
@@ -1091,6 +1091,34 @@ class _BaseTFR(ContainsMixin, UpdateChannelsMixin, SizeMixin):
                                default_index=default_index)
         return df
 
+    def average_freqs(self, band_name, method='mean'):
+        """Average frequencies over a band.
+
+        Averages the TFR object over the entire frequency axis.
+
+        Parameters
+        ----------
+        band_name : str
+            The name of the frequency band to assign to.
+
+        Returns
+        -------
+        inst : instance of AverageTFR | EpochsTFR
+            The modified instance.
+        """
+        if self.data.ndim == 4:
+            axis = 2
+        elif self.data.ndim == 3:
+            axis = 1
+
+        # return a lambda function for computing a combination metric
+        # over epochs
+        func = _check_combine(mode=method, axis=axis)
+        data = func(self.data)
+        self._data = data
+        self.freqs = np.mean(self.freqs)
+        self.band_name = band_name
+
 
 @fill_doc
 class AverageTFR(_BaseTFR):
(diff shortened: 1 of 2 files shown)


adam2392:
Any qualms with this approach in the context of #9793 ? Don't want to move forward if there's some issues w/ the general approach :p.

@agramfort @drammock @larsoner

adam2392:
@larsoner I changed this to take the average over frequencies as well now and remove `band_name`.

adam2392:
@agramfort @larsoner I was thinking about this and would it be better to add an "axis" parameter in the `average()` function?

adam2392:
> Here's an example that fails:
> 
> ```python
> import os
> import numpy as np
> import matplotlib.pyplot as plt
> import mne
> sample_data_folder = mne.datasets.sample.data_path()
> sample_data_raw_file = os.path.join(sample_data_folder, 'MEG', 'sample',
>                                     'sample_audvis_raw.fif')
> raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False, preload=False)
> events = mne.find_events(raw, stim_channel='STI 014')
> event_dict = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3}
> epochs = mne.Epochs(raw, events, event_id=event_dict, preload=False,
>                     proj=False)
> freqs = np.linspace(1,101,11)[5:]
> tfr = mne.time_frequency.tfr_morlet(epochs, freqs, 5, average=False, return_itc=False)
> tfr.average(dim='freqs')
> tfr.average()  # now average over epochs -> ValueError
> # ValueError: Number of frequencies and data size don't match (6 != 1).
> ```

Okay this should work now. The method used to change things in place, whereas the doc string does not dictate that.

drammock:
> Okay this should work now. The method used to change things in place, whereas the doc string does not dictate that.

Indeed it does seem to work now, but the problem was not (just) that things were being modified in-place, it's that not enough things were being changed (the length of `freqs` passed to the constructor didn't match the relevant axis size of `data.shape`).  

In fact, typically we adhere to the convention that methods *do* modify objects in-place, whereas functions return a copy.  However in this case, the `.average()` method was previously returning a different class (EpochsTFR → AverageTFR) so there was no possibility of modifying in-place.  But now, `.average(dim="freqs")` can (and probably should) modify in-place, and we should add a `copy=True|False` param to the method, and document the fact that with `.average(dim='epochs')` it will always return a copy.  @agramfort and @larsoner, I'd like to hear your opinions on this too.  Should averaging over frequencies (optionally) modify an `EpochsTFR` in-place?  If so, should the default be `copy=False`?

PR #9795, opened 2021-09-30
MRG: Fix read_raw_brainvision() handling of comma-separated data files

hoechenberger:
This fixes an issue reported on the forum:
https://mne.discourse.group/t/brainvision-raw-get-data-could-not-convert-string-to-float/3730

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -175,6 +175,8 @@ Enhancements
 
 - Add full ECoG dataset to MNE-misc-data and demonstrate its use in :ref:`ex-electrode-pos-2d` and :ref:`tut-ieeg-localize` (:gh:`9784` by `Alex Rockhill`_)
 
+- `mne.io.read_raw_brainvision` now handles ASCII data with comma-separated values, as may be exported from BrainVision Analyzer (:gh:`9795` by `Richard Höchenberger`_)
+
 Bugs
 ~~~~
 - Fix bug in :meth:`mne.io.Raw.pick` and related functions when parameter list contains channels which are not in info instance (:gh:`9708` **by new contributor** |Evgeny Goldstein|_)
@@ -250,8 +252,6 @@ Bugs
 - Fix bug where :func:`mne.io.read_raw_snirf` was including the landmark index as a spatial coordinate (:gh:`9777` by `Robert luke`_)
 
 - Fix bug where `mne.Annotations` were not appending channel names when being added together (:gh:`9780` by `Adam Li`_)
-
-- `mne.io.read_raw_brainvision` now handles ASCII data with comma-separated values correctly (:gh:`9795` by `Richard Höchenberger`_)
 
 API changes
 ~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


cbrnr:
What do the official file format specs say? Don't they define the separator?

sappelhoff:
re-iterating here again that this issue was about a "BrainVision Analyzer" exported file, not a standard BrainVision raw data file that is produced during recording: https://mne.discourse.group/t/brainvision-raw-get-data-could-not-convert-string-to-float/3730/7?u=sappelhoff

(just to have it here in case it comes up again)

hoechenberger:
> What do the official file format specs say? Don't they define the separator?

I couldn't find a spec of the ASCII data format. The header file only specifies the decimal separator.

In `main`, we test with spaces as column separators. In the Discourse posting linked above, OP has data with commas as column separators. No idea.

hoechenberger:
> that this issue was about a "BrainVision Analyzer" exported file

So are you saying that BrainProducts isn't following their own standard here? 🤯

sappelhoff:
> So are you saying that BrainProducts isn't following their own standard here? exploding_head

The situation is weird (as we've often realized while working on pybv) --> BrainVision files are not always BrainVision files: They can differ (a) depending on which recording software exports them, and (b) they are much more different when exported from BrainVision Analyzer (which IMHO exports derivative data, not raw data)

This is due to the fact that their data formats evolved over time and have only very recently been publicly released (well, a small part of it; and after we asked for it to have it available for BIDS) in the form of the spec that we also use at pybv.

hoechenberger:
@sappelhoff This seems to be working, not sure if it covers all use cases, obviously, but it doesn't seem to break existing behavior and allows me to load the BVA-exported data provided by OP. I hence think this is a good addition.

sappelhoff (approved):
thanks @hoechenberger. I feel there are some aspects we don't understand (yet?), but I think that's impossible to clear up without input by Brain Products.

And I agree that if this small tweak enables reading without impacting the other code, we might as well merge and move on.

hoechenberger:
@cbrnr Feel free to merge unless you have any remaining questions or concerns!

cbrnr:
Thanks @hoechenberger!

hoechenberger:
@sappelhoff It could be a good idea to run the MNE-BIDS CI just to be even more sure this change here doesn't break anything? WDYT?

sappelhoff:
won't it run tonight as scheduled? :) 

https://github.com/mne-tools/mne-bids/blob/91f9fb3868558887acc161650c8b2a1d8d9b69fb/.github/workflows/unit_tests.yml#L11-L12

Or do you need me to do something?
Also, we just released a new pybv, so we might want to update that in mne-bids as well (contains your new feature, and a fix for big-endian archs)

drammock:
Just a comment since this was already merged: if there is a published spec and we decide to support files that deviate from the spec, we might consider a `strict=True` kwarg to allow it (rather than silently accepting out-of-spec formats)

PR #9796, opened 2021-09-30
MRG, ENH: Add Otaniemi phantom based on CT

### LINKED ISSUE #9792: ENH: Add Neuromag phantom subject
I created an incomplete FreeSurfer subject directory for the Neuromag phantom that contains the head surface and CT-as-MRI T1.mgz. With this, you can do a plot for [this](https://mne.tools/dev/auto_tutorials/inverse/80_brainstorm_phantom_elekta.html?highlight=phantom) example that would look like:
```
import mne
import numpy as np

data_path = mne.datasets.brainstorm.bst_phantom_elekta.data_path()
info = mne.io.read_info(data_path + '/kojak_all_200nAm_pp_no_chpi_no_ms_raw.fif')
sphere = mne.make_sphere_model((0., 0., 0.), 0.08)
trans = mne.transforms.Transform('head', 'mri', np.eye(4))
fig = mne.viz.plot_alignment(
    info, trans, 'phantom', bem=sphere, surfaces=('inner_skull', 'head-dense'),
    coord_frame='meg', show_axes=True, mri_fiducials=True, subjects_dir='.',
    dig=True)
fig.plotter.screenshot('alignment.png')
```

alignment

FYI for folks who haven't seen it, this is what [the phantom](https://neuroimage.usc.edu/brainstorm/Tutorials/PhantomElekta#The_phantom) actually looks like:

<details>

</details>

Options:

1. Add to `bst_phantom_elekta`. I don't like this because the phantom is more generally usable than this, and it's not part of the data Brainstorm actually distributes, either.
2. Add to `mne-misc-data`. It's small so this is doable, but not the best organization-wise.
3. Add a `mne.datasets.fetch_phantom`. It's more stuff in that namespace (not great), but probably the cleanest because you can tell it explicitly which `subjects_dir` to put the subject into. It's also easier to update/add files in the future and for users to get them, similar to how `fetch_fsaverage` works.
4. Don't distribute this, nobody actually needs it (though I do, other MEG folks might, and it would make the bst example nicer!)

I prefer option (3), any thoughts?

larsoner:
Turns out the CT I already had was for the Otaniemi phantom. I'm waiting to hear back if we can get a CT of the newer phantom for the VectorView / TRIUX systems.

- [x] Add Otaniemi phantom
- [x] Add tests
- [ ] ~~Add VectorView phantom~~
- [x] Update VectorView example
- [x] Update latest.inc

Closes #9792

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -21,6 +21,7 @@ recursive-include mne/data/fsaverage *
 include mne/datasets/_fsaverage/root.txt
 include mne/datasets/_fsaverage/bem.txt
 include mne/datasets/_infant/*.txt
+include mne/datasets/_phantom/*.txt
 include mne/data/dataset_checksums.txt
 include mne/data/eegbci_checksums.txt
 
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -171,6 +171,8 @@ Enhancements
 
 - All data fetching code now relies on ``pooch``, which is an added optional requirement for dataset fetchers (:gh:`9742` by `Adam Li`_ and `Daniel McCloy`_)
 
+- Add phantom FreeSurfer subject fetcher :func:`mne.datasets.fetch_phantom` to allow fetching the Otaniemi phantom (:gh:`9796` by `Eric Larson`_)
+
 - Add :func:`mne.datasets.fetch_dataset` to allow packages to fetch datasets in a similar fashion to MNE (:gh:`9763` by `Adam Li`_)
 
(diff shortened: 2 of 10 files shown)


larsoner:
Okay so apparently the Otaniemi (old) Elekta phantom is the only one that I have / can currently get a CT scan of. Fortunately the newer VectorView/TRIUX phantom has the same (or at least similar enough) "scalp" geometry that I think we can use it as a suitable surrogate even in the VectorView phantom example, on `main` we have:

image

image

And on this PR:

img

img

Ready for review/merge from my end.

drammock (approved):
looks reasonable to me.

agramfort (approved):
thx @larsoner

PR #9799, opened 2021-10-01
MRG, BUG: Fix plot_alignment for OPMs

larsoner:
The "normal" coil_def for OPMs is a line (1D), which makes Qhull angry, and thus breaks plot_alignment. The "accurate" geometry is a cube, so it works better.

An alternative is to detect dimension deficiency and work around it, but even that's a pain as we'd have to deal with 2D geometry (probably `Delaunay`) and 1D geometry (probably convert to a skinny tube). However, it's easy enough just to try to use accurate if available first -- it will be in all built-in cases we have in our own `coil_def.dat`, but we keep the fallback for users supplying their own `coil_def.dat`. If we hit the really rare corner case of a custom `coil_def.dat` plus dimension-deficient geometry we can fix it later, but hopefully YAGNI

I also added a `fetch_phantom` to `_download_all_example_data` that I forgot to have in a recent PR

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -253,6 +253,8 @@ Bugs
 
 - Fix bug where :func:`mne.io.read_raw_persyst` was lower-casing events it found in the ``.lay`` file (:gh:`9746` by `Adam Li`_)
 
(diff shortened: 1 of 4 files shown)


agramfort (approved):

agramfort:
Thx @larsoner

PR #9802, opened 2021-10-01
Fix pooch

drammock:
Fixes a variety of problems with the implementation of `fetch_dataset()`.  Also updates the version number and hash of the MNE-misc dataset, which just got a new release today.

### DIFF
--- a/mne/datasets/utils.py
+++ b/mne/datasets/utils.py
@@ -93,9 +93,9 @@ def _get_path(path, key, name):
         if not isinstance(path, str):
             raise ValueError('path must be a string or None')
         return path
-    # 2. get_config(key)
+    # 2. get_config(key) — unless key is None or "" (special get_config values)
     # 3. get_config('MNE_DATA')
-    path = get_config(key or 'MNE_DATA')
+    path = get_config(key or 'MNE_DATA', get_config('MNE_DATA'))
     if path is not None:
         if not op.exists(path):
             msg = (f"Download location {path} as specified by MNE_DATA does "
--- a/mne/datasets/utils.py
+++ b/mne/datasets/utils.py
@@ -81,8 +81,8 @@ def _dataset_version(path, name):
     else:
         # Sample dataset versioning was introduced after 0.3
         # SPM dataset was introduced with 0.7
-        version = '0.3' if name == 'sample' else '0.7'
-
+        versions = dict(sample='0.7', spm='0.3')
+        version = versions.get(name, '0.0')
     return version
 
 
--- a/mne/datasets/utils.py
+++ b/mne/datasets/utils.py
@@ -82,7 +82,7 @@ def _dataset_version(path, name):
         # Sample dataset versioning was introduced after 0.3
         # SPM dataset was introduced with 0.7
         versions = dict(sample='0.7', spm='0.3')
-        version = versions.get(name, 'unknown')
+        version = versions.get(name, '0.0')
     return version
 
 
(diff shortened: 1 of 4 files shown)


larsoner:
> Also updates the version number and hash of the MNE-misc dataset, which just got a new release today.

This will conflict with #9800, but @alexrockhill is an expert merger at this point so I think it's okay :)

mne/datasets/utils.py
```diff
@@ -95,7 +95,7 @@ def _get_path(path, key, name):
         return path
     # 2. get_config(key)
     # 3. get_config('MNE_DATA')
-    path = get_config(key, get_config('MNE_DATA'))
+    path = get_config(key or 'MNE_DATA')
```
larsoner, line 98: I don't quite get this one. It means that before if the config for the given key did not exist, we would use `MNE_DATA`, but now we only do so if `key is None`? This seems like an incorrect change in behavior?

drammock: `None` is a sentinel value for `get_config()` that returns all config keys and their values that are defined in the config file.  But here, `key` being `None` actually means something different (viz, "this dataset is MNE-external and is not kept track of in the config")

drammock: so this change makes it so that, when `key is None`, we get the config value for `"MNE_DATA"` instead of a big dict of all defined config values.

drammock: put another way, it updates the `_get_path()` function to gracefully handle `key=None` input, which previously it was not designed to handle.

drammock:
@adam2392 can you review?

mne/datasets/utils.py
larsoner: Ahh that makes sense, thanks!

larsoner:
> @adam2392 can you review?

@drammock can you force CIs to do some review work for us, too, by creating some new test that fails on `main` but passes on this PR? Presumably downloading some `.mat` file like in https://github.com/alphacsc/alphacsc/pull/44#issuecomment-932315770 (I would look at `pymatreader` for some stable `.mat` testing file you could use) ? It looks like the only test adjustment is for a dataset version which looks reasonable, but wouldn't have caught the `alphacsc` error.

Feel free to ignore if this does not make sense or you were going to do this anyway...

mne/datasets/utils.py
drammock: Actually you're right, maybe this should still include the fallback of `get_config('MNE_DATA')` for cases of non-empty strings that aren't valid config keys?

larsoner: Or valid ones that aren't set yet in the config, which is pretty common -- people just set MNE_DATA once then all downloads just automatically go to that location

drammock:
> can you force CIs to do some review work for us, too, by creating some new test that fails on `main` but passes on this PR?

done.

adam2392 (approved):
This LGTM. 

Note: https://github.com/fatiando/pooch/pull/266 will fix our issues with uncompressing members that are sub-directories.

drammock:
CI fails are unrelated:

- `test_notebook_alignment` and `test_notebook_button_counts`: DeprecationWarning: Passing a schema to Validator.iter_errors is deprecated and will be removed in a future release. Call validator.evolve(schema=new_schema).iter_errors(...) instead.
- `test_3d_functions[mayavi]`:
    - ResourceWarning: unclosed context
    - ResourceWarning: unclosed event loop

larsoner:
Thanks @drammock !

PR #9804, opened 2021-10-04
MRG, BUG: Better coil_def and plot_alignment handling

larsoner:
Messing around with coil defs I ended up hitting a 2D `coil_def` case that's easy enough to fix. I also made custom `coil_def` errant usage easier to debug, which I do not think warrants a second `latest.inc` entry.


agramfort (approved):

larsoner:
Failures are just the recent notebook bugs

PR #9803, opened 2021-10-04
CI: Try VTK pre

larsoner:
1. Fix recent DeprecationWarnings
2. Try `--pre` build of VTK 9.1.0

PR #9805, opened 2021-10-05
MAINT: Pin jsonschema to < 4 to avoid DeprecationWarning

hoechenberger:
See https://github.com/jupyter/nbformat/issues/232

### DIFF
--- a/requirements_testing_extra.txt
+++ b/requirements_testing_extra.txt
@@ -1,6 +1,7 @@
 # requirements for full testing (on top of environment.yml/requirements.txt)
 nitime
 nbclient
+jsonschema < 4  # https://github.com/jupyter/nbformat/issues/232
 sphinx-gallery
 eeglabio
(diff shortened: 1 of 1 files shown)


agramfort (approved):
I think CI failure is unrelated

hoechenberger:
CI error seems unrelated?

@agramfort @larsoner Could you have a look and merge if you're happy?

larsoner:
FYI the preferred solution to me, already implemented in #9803 and #9800, is just to ignore the warning in conftest. It's not our problem and we know they'll probably fix it, so a simple ignore is less maintenance work for us here than a version pin we have to keep track of.

hoechenberger:
Oh, ok, sounds like a reasonable approach too 😅

So … should this PR be reverted…?

larsoner:
I'll fix it in #9803

PR #9808, opened 2021-10-06
WIP: Add support for data from NIRx Aurora 2021.9 software

rob-luke:
#### Reference issue
https://github.com/mne-tools/mne-testing-data/pull/89

The NIRx vendor has released their new acquisition software _Aurora 2021.9_. This software exports in two formats 1) their native format and 2) the community developed common SNIRF format.

A user already reported that SNIRF files from this software did not load with MNE (https://github.com/mne-tools/mne-nirs/issues/393) and I fixed the issue in https://github.com/mne-tools/mne-python/pull/9777. But we did not have any test data to confirm the issue was sufficiently resolved.

Since then, the vendor has kindly acquired some test data for us and emailed it to me (I don't have their newer devices that use the new software, my device is old and stuck with old software). They also sent the same data with the native file format.

The native file format has also changed.
1. The string used to encode dates has changed. 
2. And now the age field is optional. 

#### Additional information
Bit of a moving target 😢 these file formats.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -185,6 +185,8 @@ Enhancements
 
 - Add :func:`mne.preprocessing.ieeg.project_sensors_onto_brain` to project ECoG sensors onto the pial surface to compensate for brain shift (:gh:`9800` by `Alex Rockhill`_)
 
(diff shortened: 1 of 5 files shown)


rob-luke:
One test failed due to 60 min timeout. I just added @larsoner as a co-author

I think this is good to merge. @larsoner and @agramfort can you please review and merge if happy.

PR #9807, opened 2021-10-05
fix fetch_dataset() when force=False and no config key

drammock:
fixes #9806 (typo in commit message)

PR #9813, opened 2021-10-07
WIP, MAINT: Fix CircleCI

larsoner:
CircleCI [is failing on `main`](https://app.circleci.com/pipelines/github/mne-tools/mne-python?branch=main). I can replicate locally easily with:
```
$ make -C doc clean && PATTERN=intro make -C doc html_dev-pattern
...
generating gallery for auto_tutorials/intro... [ 14%] 10_overview.py
generating gallery for auto_tutorials/intro... [ 28%] 15_inplace.py                                                                                                                 
...
AssertionError: 1 Brain:
dict: len=35, 9 referrers: dict/frame/list/module
method: <bound method Brain.save_image of <mne.viz._brain._brain.Brain object at 0x7f732744f1c0>>
cell: <cell at 0x7f7325b8c8b0: Brain object at 0x7f732744f1c0>
cell: <cell at 0x7f7325ae1c10: Brain object at 0x7f732744f1c0>
dict: len=2, 4 referrers: UpdateLUT/dict/list
```
In other words, something in memory can no longer be cleaned up.

According to `git bisect` this actually starts for me on #9742, so I reverted #9807, then #9802, then #9742. Let's see if `[circle full]` actually works here. If it does, we can drill into why it's actually breaking things. I looked through `pooch`'s code a bit and I don't see any fancy mechanics that could be causing this problem (?).

I also have no idea why this is just happening now (after #9754 was merged), since the `pooch` PRs have has been merged for a while and builds came back green.

### DIFF
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -349,11 +349,22 @@ def _get_mri_contour_figs(*, sl, n_jobs, mri_fname, surfaces,
 
 def _iterate_trans_views(function, **kwargs):
     """Auxiliary function to iterate over views in trans fig."""
-    import matplotlib.pyplot as plt
-    from ..viz.backends.renderer import MNE_3D_BACKEND_TESTING
-    from ..viz._brain.view import views_dicts
+    from ..viz import create_3d_figure
+    fig = create_3d_figure((800, 800), bgcolor=(0.5, 0.5, 0.5))
+    from ..viz.backends.renderer import backend
+    try:
+        try:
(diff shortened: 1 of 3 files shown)


larsoner:
No, I just want to see first if it does indeed fix the problem in order to isolate it. If the problem is somewhere in those PRs I reverted, then we'll have to figure out where/why and make a more targeted fix rather than reverting everything.

larsoner:
Looks like #9814 is already on the way to working, so we'll go with that...

PR #9818, opened 2021-10-08
MRG: Allow _plot_mri_contours() to return arrays

### LINKED ISSUE #9812: BUG: _plot_mri_contours makes too many figures
@hoechenberger it looks like https://github.com/mne-tools/mne-python/pull/9754/files#r719376494 was never actually addressed in #9754 unless I'm missing it somehow in the diff?

hoechenberger:
Fixes #9812

I don't think we need a changelog entry for this one.

cc @larsoner


agramfort:
does it help already for something? it helps memory somewhere?

larsoner:
(In current `main` if you don't decimate your MRI you end up with 256 `matplotlib` figures...)

PR #9821, opened 2021-10-08
Easier pip install

drammock:
following dev discussion, this should allow things like `pip install mne[data]` to get just the basics plus what is needed for dataset downloading (numpy, scipy, pooch, tqdm).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,10 +34,8 @@ if version is None:
     raise RuntimeError('Could not determine version')
 
 
-descr = """MNE python project for MEG and EEG data analysis."""
-
 DISTNAME = 'mne'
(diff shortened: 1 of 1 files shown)


larsoner:
We could start small and just support [data] to make it easier to use dataset fetchers

drammock:
good idea, I'll do that in a new PR.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

cbrnr:
> We could start small and just support [data] to make it easier to use dataset fetchers

Is this a change from previous versions? Does `mne.datasets` require `pooch` now?

PR #9829, opened 2021-10-11
MAINT: Update broken link, fix rendering

larsoner:
@natalieklein FYI your site http://natklein.weebly.com appears to be down so I'm switching to https://github.com/natalieklein so that our `linkcheck` CIs don't complain, okay with you?

@hoechenberger I took care of the rendering problem here so no need for a separate PR


hoechenberger:
Alright, thanks @larsoner

PR #9831, opened 2021-10-11
MRG, BUG: Allow reading just stim

### LINKED ISSUE #9811: read_raw_edf fails if excluding all the signal channels
#### Describe the bug
When calling read_raw_edf(path, exclude=exclude, preload=False), if exclude contains all channels but stim ones, 
the call fails 

```
  File "/data/loris-mri/bin/mri/python_virtualenvs/loris-mri-python/lib/python3.7/site-packages/mne/io/edf/edf.py", line 433, in _get_info
    edf_info['record_length'][1] / edf_info['record_length'][0]
  File "/data/loris-mri/bin/mri/python_virtualenvs/loris-mri-python/lib/python3.7/site-packages/numpy/core/_methods.py", line 40, in _amax
    return umr_maximum(a, axis, None, out, keepdims, initial, where)
ValueError: zero-size array to reduction operation maximum which has no identity

```
In mne/io/edf/edf.py::_get_info
```
sfreq = np.take(n_samps, not_stim_ch).max() * \
        edf_info['record_length'][1] / edf_info['record_length'][0]

```
np.take(n_samps, not_stim_ch) returns an empty list and .max() is failing.

#### Steps to reproduce
Calling read_raw_edf(path, exclude=exclude, preload=False), with exclude set to all channels but stim ones 

#### Expected results
Avoid the ValueError and report the absence of data/signal channels to the user

#### Actual results
ValueError: zero-size array to reduction operation maximum which has no identity

larsoner:
Closes #9811

@laemtl @agramfort feel free to try this branch with your problematic file

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -191,6 +191,8 @@ Enhancements
 
 - Add :func:`mne.preprocessing.ieeg.project_sensors_onto_brain` to project ECoG sensors onto the pial surface to compensate for brain shift (:gh:`9800` by `Alex Rockhill`_)
 
+- Add support for data acquired with NIRx devices using Aurora software version 2021.9 (:gh:`9800` by `Robert Luke`_, and `Eric Larson`_)
+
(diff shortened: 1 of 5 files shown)


cbrnr:
> Just analog stim channels I think.

Ah OK. For some reason I thought the OP had an annotations-only file.

PR #9840, opened 2021-10-13
MRG: Fix pytest test discovery for MNE-Python devel versions

hoechenberger:
This should fix things for the test adapter in VS Code.

cc @cbrnr


hoechenberger:
@larsoner Im currently afk, feel free to commit your proposed change and merge if you want

PR #9842, opened 2021-10-13
WIP: Try to parallelize &  reduce memory usage of Report.add_evokeds()

hoechenberger:
cc @agramfort


hoechenberger:
Not seeing a huge effect locally :( in fact, peak memory consumption appears to be slightly worse now, even though we release memory in-between:

`main`:
Screen Shot 2021-10-14 at 11 50 31

This PR:
Screen Shot 2021-10-14 at 11 51 25

PR #9843, opened 2021-10-14
MRG: Change Beer Lambert law default ppf value to 6

rob-luke:
#### Reference issue
The fNIRS processing uses the modified Beer Lambert law (BLL) to convert optical density data to changes in haemoglobin concentration. This process uses a scaling factor called the partial path length factor.

The largest fNIRS group that wrote the [Homer software uses a ppf=6 ](https://github.com/BUNPC/Homer3/blob/c83f79ed5374ede69180e426041aca9cb9dd080e/FuncRegistry/UserFunctions/hmrR_OD2Conc.m#L13-L21)value. Whereas we use a default value of 0.1. @larsoner already ran in to this issue before, and for his code needed to change the MNE function to use ppf=6 (https://github.com/mne-tools/mne-python/pull/8711#issuecomment-758643565)

When I wrote the BLL code I had been using the default ppf value from the [Huppert AnalyzIR toolbox](https://github.com/huppertt/nirs-toolbox/blob/45769c83ec1226768693a76ed972efa0656a4781/%2Bnirs/%2Bmodules/BeerLambertLaw.m#L11) ([Santosa... Huppert 2018](https://www.mdpi.com/1999-4893/11/5/73)) of 0.1. Their approach was to use a partial volume correction (PVC=60) when determining the ppf. So their ppf = [dpf=6]/[pvc=60] = 0.1

I had assumed when writing our code that the 2018 paper was a better benchmark than the Homer 2009 paper (also by [Huppert](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2761652/)). However, I just had a phone call with Yucel (author of the [fNIRS best practices paper](https://doi.org/10.1117/1.NPh.8.1.012101)), and they still recommend ppf=6. For many reasons, but one that stood out for me is that the pvc value is not well known and varies across the head and task.

To be in line with best practices and with the most prominent fNIRS software and match community expectations from existing literature I change our default ppf=0.1 to ppf=6

#### Additional information
Neither approach is right or wrong. But using ppf=6 is more inline with the community expectations and makes our results easier compared to the majority of the existing literature.

I've been going back and forth in my head about this for a while, but I think we should make the jump before the 0.24 release.

The tests already had a hard coded ppf value (didnt use the default), so they didnt require changing.

### DIFF
--- a/mne/viz/conftest.py
+++ b/mne/viz/conftest.py
@@ -37,7 +37,7 @@ def fnirs_epochs():
                     'NIRx', 'nirscout', 'nirx_15_2_recording_w_overlap')
     raw_intensity = read_raw_nirx(fname, preload=False)
     raw_od = optical_density(raw_intensity)
-    raw_haemo = beer_lambert_law(raw_od)
+    raw_haemo = beer_lambert_law(raw_od, ppf=6.)
     evts, _ = events_from_annotations(raw_haemo, event_id={'1.0': 1})
     evts_dct = {'A': 1}
     tn, tx = -1, 2


rob-luke:
FYI: @HanBnrd (also looking forward to your talk next week!) @rderollepot @kylemath

@larsoner @agramfort ready for review, unless someone above disagrees

rob-luke:
> but unfortunately we need to do it via deprecation

Ah yes, of course. No worries, I will implement as you describe.

Thanks for explaining how to do it in detail, that saved me a good 20 mins!

PR #9845, opened 2021-10-14
MRG: Expand ~ in Report file operations

hoechenberger:
Followup of #9613


hoechenberger:
I don't think we need tests specifically for this, as we're already testing `_check_fname()` rather extensively.

larsoner:
Thanks @hoechenberger

PR #9851, opened 2021-10-14
[MRG] Add cnorm parameter to AverageTFR.plot()

### LINKED ISSUE #9809: Add color normalization parameter to AverageTFR.plot()
I've already added color normalization to `plot_topomap()` in #9468, which allows users to map a custom data value to a color value (e.g. white). For example, using `vmin=-1` and `vmax=1.5` would result in the center value being mapped to 1 / (1 + 1.5) = 0.4 instead of 0. Using a new parameter `cnorm=TwoSlopeNorm(vmin=-1, vcenter=0, vmax=1.5)` would solve this issue and map the data value 0 (instead of 0.4) to the center of the colormap (white for e.g. `cmap='RdBu_r'`).

Any objections to adding a `cnorm` parameter to `AverageTFR.plot()` just like we already have in `plot_topomap()`? If not I'll go ahead and submit a PR.

cbrnr:
Fixes #9809. This parameter can be used to center a colormap to an arbitrary value, effectively making `mne.viz.utils.center_cmap` unnecessary. Instead of having to change the colormap with `cmap = center_cmap(plt.get_cmap("RdBu"), vmin, vmax)`, it is now possible to define `cnorm` as e.g. `cnorm = matplotlib.colors.TwoSlopeNorm(vmin=-1, vcenter=0, vmax=1.5)` and pass that to `tfr.plot` (`vmin` and `vmax` are not required anymore in this case).

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -321,4 +321,4 @@ API changes
 
 - The output folder name for HF_SEF datasets is now ``hf_sef`` instead of ``HF_SEF`` (:gh:`9763` by `Adam Li`_)
 
-- Deprecate `mne.viz.utils.center_cmap` (:gh:`9851` by `Clemens Brunner`_)
+- Deprecate ``mne.viz.utils.center_cmap`` (:gh:`9851` by `Clemens Brunner`_)


cbrnr:
I've edited the ERDS maps example to use this parameter.

cbrnr:
Updated example: https://36038-1301584-gh.circle-artifacts.com/0/dev/auto_examples/time_frequency/time_frequency_erds.html

Issue #9857, opened 2021-10-15
Argument `threshold` and `measure` of ICA.find_bads_eXg methods are not clear


mscheltienne:
I wanted to compare different argument pairs for `threshold` and `measure` in both `ica.find_bads_eog` and `ica.find_bads_ecg` on EEG data (about 600 raw instances of 4/5 minutes each @ 512 Hz). 

The problem is, when reading the documentation for those 2 arguments, it is really not clear how I am supposed to set the `threshold` or even how it is set by default.

### Argument `measure`: 

> Which method to use for finding outliers. 'zscore' (default) is the iterated Z-scoring method, and 'correlation' is an absolute raw correlation threshold with a range of 0 to 1.

Is the z-score normalized between 0 and 1 as well? If not what is the range?
_Additional:_ `iterated` -> `iterative`?

### Argument `threshold`:

> The value above which a feature is classified as outlier. If ‘auto’ and method is ‘ctps’, automatically compute the threshold. If ‘auto’ and method is ‘correlation’, defaults to 3.0. The default translates to 0.25 for ‘ctps’ and 3.0 for ‘correlation’ in version 0.21 but will change to ‘auto’ in version 0.22.

Auto + CTPS -> Automatically compute the threshold.
Auto + Correlation -> 3.0
Wait, correlation is normalized between 0 and 1? So what does a threshold of 3.0 mean? 

> The default translates to 0.25 for ‘ctps’ and 3.0 for ‘correlation’ in version 0.21 but will change to ‘auto’ in version 0.22.

What does this mean? From the first sentences, I thought Auto + CTPS was a (somehow) automatically computed threshold. But it defaults to a fix value, 0.25?

_Additional:_ backquotes are missing around `'ctps'`, `'correlation'` and `'auto'`. 
Should the line about behavior in old versions 0.21 and 0.22 be removed?

### _Additional_

_Additional:_ for argument `method`, backquotes are missing around `'ctps'` and `'correlation'`.

## `ica.find_bads_eog`

Similar questions, but the documentation is even more sparse.
For argument `threshold`:

> The value above which a feature is classified as outlier.

-----------------

I would expect all metrics to be normalized between 0 and 1, and thus the threshold would be a value between 0 and 1 as well. If this is not the case, some more detailed explanation might be useful.

PR #9858, opened 2021-10-15
Improve documentation of `ica.find_bads_eXg` methods and fix interaction between `measure` and `threshold`.

mscheltienne:
Problems described in #9857.

I added the missing double backquotes around argument strings in the documentation.
I replaced iterated with iterative and removed the capital Z when writing `z-scoring` to match the syntax most used.
I move the common docstring for `measure` to the docdict, under ICA.

These 3 commits only address the minor problems. For the actual meaning of the docstring, I'll need some insights on what it could be.

### DIFF
--- a/mne/preprocessing/ica.py
+++ b/mne/preprocessing/ica.py
@@ -1301,7 +1301,8 @@ class ICA(ContainsMixin):
 
         Parameters
         ----------
-        inst : instance of Raw, Epochs or Evoked to compute sources from.
+        inst : instance of Raw, Epochs or Evoked
+            Object to compute sources from.
         ch_name : str
             The name of the channel to use for ECG peak detection.
             The argument is mandatory if the dataset contains no ECG


mscheltienne:
### Proposition:

If `measure` is set to `zscore`, then `threshold` is set to 3.0 (3 std seems reasonable).
If `measure` is set to `correlation`, then `threshold` is set to a value between [0, 1]. Does it need to be different depending on the type of artifact we are looking for? I'm guessing 0.9 would be a good default value for correlation with EOG related components, but it might be too high for ECG-related components. 

### Documentation would turn into:

#### `measure`

```
measure : 'zscore' | 'correlation'
    Which method to use for finding outliers among the components: 
        - ``'zscore'`` (default) is the iterative z-scoring method. This method computes 
          the z-score of the component's scores and masked the components with a z-score 
          above threshold. This process is repeated until no supra-threshold component remains. 
        - ``'correlation'`` is an absolute raw correlation threshold ranging from 0 to 1.
```

#### `threshold`

```
threshold : float | str
    Value above which a feature is classified as outlier. If ``measure`` is ``zscore``, defines the
    threshold on the z-score used in the iterative z-scoring method. If ``measure`` is ``correlation``,
    defines the absolute threshold on the correlation between 0 and 1. 
    If ``'auto'`, defaults to 3.0 if ``measure`` is ``zscore`` and 0.9 if ``measure`` is ``correlation``.
```

I am not a big fan of this fix 0.9 value. Do you have any better ideas?

#### Special case 1: `find_bads_ref`

If `method`is set to `together`, then `measure` is ignored and `zscore` is used. 
If this is the desired behavior, it should be clearly stated in the argument docstring.

The `Notes` section about the 2 `methods` has not been updated:

> Thresholding in both cases is based on adaptive z-scoring:
> The above-threshold components will be masked and the z-score will be
> recomputed until no supra-threshold component remains.

This is not the case, in the case `method` is `separate`, the method `_find_bads_ch` is called with the passed `measure` argument, which can be set to something else than `zscore`.

#### Special case 2: `find_bads_ecg`

If `method` is `ctps`,  then `measure` is ignored. There was an attempt to document this in both argument `threshold` and `method`, but IMO it's not clear at all.

mscheltienne:
@agramfort Currently it is hard-coded to 3.0 for values between 0 and 1, which makes absolutely no sense.
As much as I dislike hard-coded threshold, setting at least the default value as a very safe threshold within the range would be better, and would at least push people in the correct direction.

Current: Threshold of 3.0 for values between 0 and 1 - > What the heck does it mean?
Instead, how about setting it to 0.99 with a sentence in the docstring inviting the user to change it to his need when using correlations instead of zscore?

drammock:
thanks @mscheltienne !

PR #9862, opened 2021-10-15
MRG: In Report.add_ica, replace "n_components" with more versatile "picks" param

hoechenberger:
This allows us to e.g. only plot excluded components.


agramfort:
thx @hoechenberger

PR #9866, opened 2021-10-17
Fix: allow negative integers for `n_jobs` in 2 functions.

### LINKED ISSUE #9849: `mne.time_frequency.psd_welch` does not support negative numbers for `n_jobs` parameter.
Since it should be a rather unambiguous phenomenon, so please allow me to describe the bug without your recommended format.

Negative numbers like -1 cannot be used for `n_jobs` in the function `psd_welch` under the module `mne.time_frequency`, which fails to enable users to adaptively activate all cores when calling this function, unlike those functions (such as `plot_psd`) that allows -1 for `n_jobs` in MNE.

Riessarius:
#### Reference issue
Fixes #9849 .

Allow negative integers for n_jobs in psd_array_welch() and psd_array_multitaper().

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -43,6 +43,8 @@ Current (0.24.dev0)
 
 .. |Joshua Teves| replace:: **Joshua Teves**
 
+.. |Riessarius Stargardsky| replace:: **Riessarius Stargardsky**
+
 Enhancements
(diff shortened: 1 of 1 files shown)


Riessarius:
@agramfort It really is hard for newcomers like me to deal with this problem, but anyway I tried to manage it. I hope I did it in the right way.

PR #9867, opened 2021-10-18
MRG: Add a locking/unlocking context manager to Info for error checking.

mscheltienne:
Following the discussion on #9752 and the proposition in https://github.com/mne-tools/mne-python/pull/9767#issuecomment-934592039, here is a possible implementation of @larsoner proposition.
I removed the try/else/finally block, as at least for now, it might be masking potential bugs.

I added the unlocking context manager under the `_unlock` name; with the current state in a variable named `_unlocked`.
I added a list of all the valid attributes in `Info` in the form of the dictionary `_attributes`. For each valid key, the value is either:
- a string to raise an error message to guide users changing `Info` attributes without unlocking first towards the correct setter function.
- a function to check if the type/value looks correct before setting for attributes that can directly be set by the user.

I added the corresponding `__setitem__` method. I kept the possibility to set non-valid keys with a warning message.

This is obviously not going to pass the CI tests, as the MNE function setting attributes now have to be changed to use the context manager. Is this proposition with a dictionary containing either the error message to raise or the checker function OK for everyone?

-------------

What are the attributes that can be set directly by the users and that need a quick checker function?

- bads
- description
- experimenter
- subject_info
- device_info
- helium_info

Correct? Any others?

-------------

For the checker function, is a check of the type as proposed good enough for now, with potential improvements later on?
e.g. for `bads` we could check that the elements of the list are within the info `ch_names`.

### DIFF
--- a/mne/io/tests/test_meas_info.py
+++ b/mne/io/tests/test_meas_info.py
@@ -842,9 +842,9 @@ def test_repr():
     info = create_info(1, 1000, 'eeg')
     assert '7 non-empty values' in repr(info)
 
-    t = Transform(1, 2, np.ones((4, 4)))
+    t = Transform('meg', 'head', np.ones((4, 4)))
     info['dev_head_t'] = t
-    assert 'dev_head_t: MEG device -> isotrak transform' in repr(info)
+    assert 'dev_head_t: MEG device -> head transform' in repr(info)
 
 
 def test_repr_html():


mscheltienne:
I added a couple of error messages as examples. Now is a good time to settle on a common syntax.
In the examples, I used:

> This attribute can not be directly set. Please use instance methods ... instead.

Maybe replacing instance methods with a more explicit `<Raw | Epochs | Evoked> methods` would be better.

Note that I do not know what should be the error message for most of them. I do not know which function users are supposed to use, especially for all the MEG-related attributes. Some help on this part would be appreciated.

----------

I will start adding the unlocking context manager across MNE functions to get more and more CI tests working again.

mscheltienne:
Currently for `test_meas_info`: 11 failed, 26 passed in 1.89s.
10/11 with a similar error: 

```
git\mne-python\mne\io\fiff\raw.py:102: in __init__
    copy.deepcopy(raws[0].info), False,
C:\Program Files\Python38\lib\copy.py:153: in deepcopy
    y = copier(memo)
git\mne-python\mne\io\meas_info.py:846: in __deepcopy__
    result[k] = deepcopy(v, memodict)
git\mne-python\mne\io\meas_info.py:675: in __setitem__
    if isinstance(self._attributes[key], str) and self._unlocked:
E   AttributeError: 'Info' object has no attribute '_unlocked'
```

Something weird happens with I/O operations and with `deepcopy`. I'll look into it later.

hoechenberger:
> I am doing `subject_info`, too, in some code that predates all of MNE-Python's anonymization functionality.

Are you saying there won't be a way anymore to change e.g. the subject ID via the public API?

larsoner:
@mscheltienne ready for review/merge from your end, then? By convention we typically prepend `MRG` to the PR title and ping people who should review (you could "request review" if you like advanced GitHub features, too)

larsoner:
... there are already some errors popping up in https://app.circleci.com/pipelines/github/mne-tools/mne-python/10853/workflows/9b94c357-1ad4-41fe-9a82-562d459bc609/jobs/36499 like:
```
WARNING: /home/circleci/project/examples/time_frequency/temporal_whitening.py failed to execute correctly: Traceback (most recent call last):
  File "/home/circleci/project/examples/time_frequency/temporal_whitening.py", line 33, in <module>
    raw.info['projs'] += proj
  File "/home/circleci/project/mne/io/meas_info.py", line 686, in __setitem__
    raise AttributeError(self._attributes[key])
AttributeError: projs can not be set directly.
```
Feel free to start looking + fixing

mscheltienne:
Last bug with `Report` is a bit of a pain to debug. Traceback is not very helpful..

```
"""
  File "C:\Users\Mathieu\Documents\pyvenv\mne\lib\site-packages\joblib\externals\loky\process_executor.py", line 404, in _process_worker
    call_item = call_queue.get(block=True, timeout=timeout)
  File "C:\Program Files\Python38\lib\multiprocessing\queues.py", line 116, in get
    return _ForkingPickler.loads(res)
  File "c:\users\mathieu\documents\git\mne-python\mne\io\meas_info.py", line 691, in __setitem__
    if isinstance(self._attributes[key], str) and self._unlocked:
AttributeError: 'Info' object has no attribute '_unlocked'
"""
```

Missing a couple of tracebacks in between 😞 I'll have a look later. 
But it does show one vulnerability of the current code:

```
class Info(dict):
    def __init__(self, *args, **kwargs):
        with self._unlock(check_after=False):
            # do more stuff in unlockstate
            
    def __setitem__(self, key, val):
        if self._unlocked:
            super().__setitem__(key, val)
    
    @contextlib.contextmanager
    def _unlock(self, check_after=True):
        state = self._unlocked if hasattr(self, '_unlocked') else False
        self._unlocked = True
        try:
            yield
        except Exception:
            raise
        else:
            if check_after:
                self._check_consistency()
        finally:
            self._unlocked = state
```

`self._unlocked` is not defined by default, and I guess there is a call to `__new__` somewhere that is not followed by either `__init__` or and unlocking, but directly by attribute setting.

larsoner:
Okay, down to +844 −422 which has a nice symmetry. I'll see if I need to make any other changes to make tests pass, then push an empty commit with `[circle full]`

PR #9869, opened 2021-10-19
MRG: adding split bids support to epochs

dengemann:
#### Reference issue

This PR addresses https://github.com/mne-tools/mne-bids-pipeline/issues/461

The same way it exists already for raw files, here we implement support for writing splits into epochs filenames in a bids-compatible way.

#### Additional information

Now working on tests. 

cc @agramfort @hoechenberger

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -2599,7 +2599,7 @@ docdict['split_naming'] = """
 split_naming : 'neuromag' | 'bids'
     When splitting files, append a filename partition with the appropriate
     naming schema: for ``'neuromag'``, a split file ``fname.fif`` will be named
-    ``fname.fif``, ``fname-1.fif``, ``fname-2.fif`` etc; while for ``'bids'``,
+    ``fname.fif``, ``fname-1.fif``, ``fname-2.fif`` etc.; while for ``'bids'``,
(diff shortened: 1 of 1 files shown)


hoechenberger:
@dengemann feel free to poke me if you're getting frustrated / running out of time and would like me to take over. We need to get this blocker out of the way...

larsoner:
Probably some annoying thing with `warnings` not being thread-safe:

https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings

> If two or more threads use the catch_warnings context manager at the same time, the behavior is undefined.

For now the easy fix is to add a `@pytest.mark.filterwarnings('ignore:.*No contour levels.*:UserWarning')` to the test with a comment reminding us to find an actual fix (users should not see these, we know sometimes there are no contours within a given level)

hoechenberger:
```
mne\tests\test_epochs.py:1210: in test_split_saving
    os.remove(fname)
E   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\VssAdministrator\\AppData\\Local\\Temp\\pytest-of-VssAdministrator\\pytest-0\\test_split_saving_False_False_1\\test-epo.fif'
```
Ugh…

hoechenberger:
Thanks @dengemann!

PR #9871, opened 2021-10-19
MRG: Add ERP/ERF images via Report.add_epochs()

hoechenberger:
I think this is a aseful addition!
Only downside is that we have to preload epochs now but I guess that's okay?


hoechenberger:
@agramfort See here:
https://36214-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/70_report.html#adding-epochs

hoechenberger:
> Maybe it can help spotting data quality issues.

Yes that was my idea :)

PR #9878, opened 2021-10-20
WIP,ENH: 3d plot primitives

GuillaumeFavelier:
Those are the changes that I made to the private 3d functions in #9689.


GuillaumeFavelier:
Honestly, this is the part of the PR I am less confident with so I would really appreciate in-depth reviews. In particular, I want to avoid regressions 🙏 

I could add a test for `_plot_glyphs` and touch some plot_alignment examples?

PR #9881, opened 2021-10-20
MRG: Fix report rendering for data without channel locations

hoechenberger:
WIP:

- [x] Tests
- [x] Changelog
- [x] There might be more affected methods (ICA topos? SSP topos?)

cc @agramfort


hoechenberger:
Should be fixed now.

PR #9882, opened 2021-10-20
[PAPER] Intracranial location and analysis paper JOSS

alexrockhill:
See https://github.com/alexrockhill/mne-python/pull/1 for editing process


alexrockhill:
> Ok when I'm back at my desk I'll create a branch called joss or something, and you can pr into that

Great, thanks!

PR #9887, opened 2021-10-21
fix import location of _check_ch_locs

agramfort:
misc


agramfort:
CI failure seems unrelated

PR #9907, opened 2021-10-27
[BUG, MRG] Fix infinite recursion bug in `get_montage_volume_labels`

alexrockhill:
`get_montage_volume_labels` can enter an infinite recursion because the voxel neighbors on every step are not removed, causing it to go from one voxel to the neighbor and back. This fixes that.


larsoner:
We can ignore the `test_doc_filling` failures, those are now fixed in `main`. I'll merge assuming Azure comes back happy other than those and what #9917 fixed

larsoner:
Thanks @alexrockhill

PR #9913, opened 2021-10-28
FIX: fiducials locking in coreg gui

GuillaumeFavelier:
This PR fixes the issue with `CoregistrationUI` reported in https://github.com/mne-tools/mne-python/issues/8833 about the fiducial markers being unlocked when dig info is provided.

It's an item of https://github.com/mne-tools/mne-python/issues/8833

cc @larsoner


larsoner:
Thanks @GuillaumeFavelier !

PR #9916, opened 2021-10-28
Test new theme capabilities

drammock:
This PR tests whether the proposed version switcher being built into our theme will work cleanly as a replacement for our custom version switcher. #9915 already added the JSON file; this PR uses the WIP theme version in https://github.com/pydata/pydata-sphinx-theme/pull/436.  Along the way it removes some custom CSS that is no longer needed (because the theme has improved to handle those elements better).


drammock:
TestSphinxWindows failure has been reported upstream: https://github.com/pydata/pydata-sphinx-theme/issues/523

@larsoner WDYT, should we merge this anyway?  The docs are built on a linux instance and the switcher looks good in the CI artifact: https://39352-1301584-gh.circle-artifacts.com/0/dev/install/contributing.html

NB: in the CI artifact it's using the JSON file currently on main, which is why it shows 0.24 as "devel".  This PR also updates that JSON so after merge-build-deploy it should correctly show `1.0 (devel)` and `0.24 (stable)`

larsoner:
We don't want CIs red to mean "might still be okay to merge" -- it leads to too much cognitive burden and risk about "what's an okay failure", and "did I make sure it's only the right build that failed". If it's an upstream Windows bug that we're willing to temporarily ignore, let's either disable the windows sphinx build or (probably better, I think it's possible?) mark it as an allowed failure in the CI matrix at the Azure config end so that the merge button stays green but the build shows up gray on the GitHub interface.

drammock:
Ok. This PR isn't urgent, let's give upstream a few days before we go disabling tests.

drammock:
upstream fix is in, restarting the failed CI.

drammock:
different failure now. I'll look next week.

drammock:
second upstream fix is in (https://github.com/pradyunsg/sphinx-theme-builder/commit/85be6e927e1841af196c708e1bd7fd04158e7d02) so hopefully the azure build-docs-on-windows job will be green now

drammock:
ok the SphinxWindows job succeded. Ready for merge when remaining CIs come back green.

larsoner:
Thanks @drammock !

PR #9925, opened 2021-10-29
Replace tmpdir with tmp_path

### LINKED ISSUE #9662: Use `tmp_path` instead of `tmpdir` pytest fixture
Just a heads up that the `tmpdir` fixture has been replaced by `tmp_path` (see [here](https://docs.pytest.org/en/6.2.x/fixture.html#:~:text=pathlib.Path%20objects.-,tmpdir,-Provide%20a%20py)). New tests should use `tmp_path`, and if someone has the time to search and replace, we could also completely switch to the recommended fixture.

cbrnr:
Fixes #9662.

### DIFF
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1421,10 +1421,11 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                    '_meg.fif', '_eeg.fif', '_ieeg.fif')
         endings += tuple([f'{e}.gz' for e in endings])
         endings_err = ('.fif', '.fif.gz')
+        fname = _check_fname()
         check_fname(fname, 'raw', endings, endings_err=endings_err)
 
         split_size = _get_split_size(split_size)
-        if not self.preload and str(fname) in self._filenames:
+        if not self.preload and fname in self._filenames:
             raise ValueError('You cannot save data to the same file.'
                              ' Please use a different filename.')
 
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1421,7 +1421,7 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                    '_meg.fif', '_eeg.fif', '_ieeg.fif')
         endings += tuple([f'{e}.gz' for e in endings])
         endings_err = ('.fif', '.fif.gz')
-        fname = _check_fname()
+        fname = _check_fname(fname)
         check_fname(fname, 'raw', endings, endings_err=endings_err)
 
         split_size = _get_split_size(split_size)
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1421,7 +1421,9 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
                    '_meg.fif', '_eeg.fif', '_ieeg.fif')
         endings += tuple([f'{e}.gz' for e in endings])
         endings_err = ('.fif', '.fif.gz')
-        fname = _check_fname(fname)
+
+        # convert to str, check for overwrite a few lines later
+        fname = _check_fname(fname, overwrite=True)
         check_fname(fname, 'raw', endings, endings_err=endings_err)
 
         split_size = _get_split_size(split_size)
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1424,7 +1424,7 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         check_fname(fname, 'raw', endings, endings_err=endings_err)
 
         split_size = _get_split_size(split_size)
-        if not self.preload and fname in self._filenames:
+        if not self.preload and str(fname) in self._filenames:
             raise ValueError('You cannot save data to the same file.'
                              ' Please use a different filename.')
 
--- a/mne/tests/test_source_estimate.py
+++ b/mne/tests/test_source_estimate.py
@@ -437,7 +437,9 @@ def test_io_stc_h5(tmp_path, is_complex, vector):
         stc = _fake_vec_stc(is_complex=is_complex)
     else:
         stc = _fake_stc(is_complex=is_complex)
-    pytest.raises(ValueError, stc.save, tmp_path / 'tmp', ftype='foo')
+    match = 'can only be written' if vector else "Invalid value for the 'ftype"
+    with pytest.raises(ValueError, match=match):
+        stc.save(tmp_path / 'tmp.h5', ftype='foo')
     out_name = str(tmp_path / 'tmp')
     stc.save(out_name, ftype='h5')
     stc.save(out_name, ftype='h5')  # test overwrite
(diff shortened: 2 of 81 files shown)


cbrnr:
This has not been super fun (simply replacing `tmpdir` with `tmp_path` didn't cut it), but now it is almost working except for the following two issues (@larsoner maybe you can help):

Windows is crashing [here](https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=16350&view=logs&j=fa98c7b3-575e-55b6-6946-22dea6f6daf8&t=9997f553-f8cd-520b-0850-4d0fe0dea652&l=225) ("Windows fatal exception: access violation"), any ideas?

There's one failing test that I haven't fixed yet, because when saving an already loaded `Raw` object, it now raises a `FileExistsError` in https://github.com/mne-tools/mne-python/blob/main/mne/utils/check.py#L167, but `test_save` expects a different error (https://github.com/cbrnr/mne-python/blob/243da7562012baf4622ae9f8752557eb37709e72/mne/io/fiff/tests/test_raw_fiff.py#L1392) that used to be raised [here](https://github.com/cbrnr/mne-python/blob/fde89c53dedd89338346bebd1ac67335fb80375f/mne/io/base.py#L1428). How would you like me to proceed?

larsoner:
> There's one failing test that I haven't fixed yet, because when saving an already loaded Raw object, it now raises a FileExistsError

This to me suggests that there is some problem with the first check that's supposed to see "if data are not preloaded and we're trying to save to self._filenames[0], raise a ValueError", and instead you hit the second / subsequent check that is "if the filename exists and overwrite is False (default), then raise a FileExistsError". I would look into why using this new fixture has changed the behavior of the first check

mne/io/base.py
```diff
@@ -1424,7 +1424,7 @@ def save(self, fname, picks=None, tmin=0, tmax=None, buffer_size_sec=None,
         check_fname(fname, 'raw', endings, endings_err=endings_err)
 
         split_size = _get_split_size(split_size)
-        if not self.preload and fname in self._filenames:
+        if not self.preload and str(fname) in self._filenames:
```
larsoner, line 1427: Nowadays we don't cast using `str` but rather use `fname = _check_fname(fname)` because it does some nice standardization for us

cbrnr: Done.

cbrnr:
@larsoner the issue is that `fname = _check_fname(fname)` raises a `FileExistsError: Destination file exists. Please use option "overwrite=True" to force overwriting.` even before the other exception (`if not self.preload and fname in self._filenames:`) can be raised. I could of course set `overwrite=True`, but I guess this defeats the purpose of the other check. Any suggestions on how to proceed?

larsoner:
> I could of course set overwrite=True, but I guess this defeats the purpose of the other check.

I would use `overwrite=True` because you don't want/need *that* function to do an overwrite check, because you're doing a more targeted one later

cbrnr:
This is now ready for review. Maybe worth a changelog entry too, but we can do this as the final step.

agramfort:
@cbrnr can we now remove tmpdir from our fixtures? Did you already take care of this? 🙏 for the cleanup

cbrnr:
What do you mean by remove? I replaced `tmpdir` with `tmp_path` everywhere, is there some config file I need to update?

sappelhoff:
And if you're in the flow of it @cbrnr perhaps you wanna do a quick pass on mne-bids as well, replacing the tmpdir -> tmp_path there? :smirk: 

But I see your comment:

> This has not been super fun

so if you've had enough for now, we'll just do it some other time when one of us has time and energy

agramfort:
indeed tmpdir is a pytest builtin fixture.

git grep still says:

mne/gui/tests/test_coreg_gui.py:def
test_coreg_gui_pyvista_file_support(inst_path, tmpdir,
mne/gui/tests/test_coreg_gui.py:    tempdir = str(tmpdir)
mne/gui/tests/test_coreg_gui.py:def test_coreg_gui_pyvista(tmpdir,
renderer_interactive_pyvistaqt):
mne/gui/tests/test_coreg_gui.py:    tempdir = str(tmpdir)

is it expected to keep it there?

>

cbrnr:
@agramfort thanks for double-checking, I missed those. Now there are no `tmpdir` fixtures anymore.

@sappelhoff I can take a look, mne-bids is much smaller and therefore this should be much faster.

cbrnr:
@sappelhoff https://github.com/mne-tools/mne-bids/pull/896

cbrnr:
Any idea what's going on with CircleCI now? https://app.circleci.com/pipelines/github/mne-tools/mne-python/11173/workflows/849f4079-3350-4539-be79-8cdff6404ee1/jobs/37508/parallel-runs/0/steps/0-103

cbrnr:
@larsoner now in addition to CircleCI most (all?) Azure jobs are failing because of an issue with OpenGL install?

larsoner:
#9943

cbrnr:
Please ping me once that is merged.

cbrnr:
(This might be something where [merge queues](https://github.blog/changelog/2021-10-27-pull-request-merge-queue-limited-beta/) could be useful – we should definitely try them when they are available.)

larsoner:
> Please ping me once that is merged.

If you go to the PR and click the "subscribe" you'll get notified automatically

cbrnr:
@GuillaumeFavelier do you know why `coreg._lock_fids` is not `False`? https://github.com/mne-tools/mne-python/runs/4081446793?check_suite_focus=true#step:13:3462

mne/tests/test_source_estimate.py
```diff
@@ -425,21 +425,20 @@ def test_io_stc(tmpdir):
...
+def test_io_stc_h5(tmp_path, is_complex, vector):
     """Test IO for STC files using HDF5."""
     if vector:
         stc = _fake_vec_stc(is_complex=is_complex)
     else:
         stc = _fake_stc(is_complex=is_complex)
-    pytest.raises(ValueError, stc.save, tmpdir.join('tmp'),
-                  ftype='foo')
-    out_name = tmpdir.join('tmp')
+    pytest.raises(ValueError, stc.save, tmp_path / 'tmp', ftype='foo')
```
drammock, line 440:
```suggestion
    # different errors for (non)vector STCs
    match = (r"(Invalid value for the 'ftype' par)|"
             r"(VectorSourceEstimate objects can only be written ad HDF5)*")
    with pytest.raises(ValueError, match=match):
        stc.save(tmp_path / 'tmp', ftype='foo')
```

drammock (approved):

mne/tests/test_source_estimate.py
cbrnr: Should we go with this more specific error message catching or wait for the fix in #9944?

larsoner: I don't think my fix in #9944 will impact this comment -- it looks like this comment just improves/modernizes the `pytest.raises` usage. So yes feel free to commit this suggestion separately

larsoner (approved):
LGTM, will merge once CIs come back happy. Thanks @cbrnr

agramfort:
thx @cbrnr !

PR #9928, opened 2021-10-30
Coreg fixes

bloyl:
Handle two issues I raised in #8833,
1) support for `info['dig']` missing some types of dig points
2) support loading from instances only containing `dig` objects.

~~still needs `latest.inc`~~ hasn't been released so no need.

### DIFF
--- a/mne/gui/_coreg.py
+++ b/mne/gui/_coreg.py
@@ -12,6 +12,7 @@ from ..io import read_info, read_fiducials, read_raw
 from ..io.pick import pick_types
 from ..io.open import fiff_open, dir_tree_find
 from ..io.meas_info import _empty_info
+from ..io._read_raw import supported as raw_supported_types
 from ..coreg import Coregistration, _is_mri_subject
 from ..viz._3d import (_plot_head_surface, _plot_head_fiducials,
                        _plot_head_shape_points, _plot_mri_fiducials,
@@ -219,10 +220,23 @@ class CoregistrationUI(HasTraits):
     def _set_info_file(self, fname):
         if fname is None:
             return
-        if not self._check_fif('info', fname):
+
+        # info file can be anything supported by read_raw
+        try:
+            check_fname(fname, 'info', tuple(raw_supported_types.keys()),
+                        endings_err=tuple(raw_supported_types.keys()))
+        except IOError as e:
+            warn(e)
+            self._widgets["info_file"].set_value(0, '')
             return
-        self._info_file = _check_fname(
-            fname, overwrite=True, must_exist=True, need_dir=False)
+
+        # ctf ds `files` are actually directories
+        if fname.endswith(('.ds',)):
+            self._info_file = _check_fname(
+                fname, overwrite=True, must_exist=True, need_dir=True)
+        else:
+            self._info_file = _check_fname(
+                fname, overwrite=True, must_exist=True, need_dir=False)
 
     def _set_omit_hsp_distance(self, distance):
         self._omit_hsp_distance = distance
--- a/mne/gui/_coreg.py
+++ b/mne/gui/_coreg.py
@@ -6,9 +6,12 @@ import os.path as op
 import numpy as np
 from traitlets import observe, HasTraits, Unicode, Bool, Float
 
+from ..io.constants import FIFF
 from ..defaults import DEFAULTS
-from ..io import read_info, read_fiducials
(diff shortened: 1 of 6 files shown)


mne/tests/test_coreg.py
```diff
@@ -443,3 +444,27 @@ def test_coreg_class_gui_match():
...
     assert_allclose(coreg.scale, want_scale, atol=5e-4)
     assert_allclose(coreg.trans['trans'], want_trans, atol=1e-6)
+
+
+@testing.requires_testing_data
+@pytest.mark.parametrize(
+    'drop_point_kind', (FIFF.FIFFV_POINT_CARDINAL, FIFF.FIFFV_POINT_HPI,
+                        FIFF.FIFFV_POINT_EXTRA, FIFF.FIFFV_POINT_EEG))
+def test_coreg_class_init(drop_point_kind):
+    """Test that Coregistration can be instantiation with various digs."""
```
agramfort, line 454:
```suggestion
    """Test that Coregistration can be instantiated with various digs."""
```

agramfort (commented):
thx @bloyl 

does it full test what you adjusted? should we add a test of the Coregistration class using some CTF data?

mne/tests/test_coreg.py
bloyl: I trust the regression tests of the backend (`mne.coreg`) so I don't think we need full coregistration regression tests starting with CTF data are needed.

I did extended the gui test to test the non-fif file loading. It works with ctf ds files and should be support anything supported by `read_raw`.

mne/gui/_coreg.py
```diff
@@ -215,10 +220,23 @@ def _set_current_fiducial(self, fid):
...
+            warn(e)
+            self._widgets["info_file"].set_value(0, '')
             return
-        self._info_file = _check_fname(
-            fname, overwrite=True, must_exist=True, need_dir=False)
+
+        # ctf ds `files` are actually directories
+        if fname.endswith(('.ds',)):
+            self._info_file = _check_fname(
+                fname, overwrite=True, must_exist=True, need_dir=True)
```
bloyl, line 236: Are there any other datafile types that are directories and not files?

larsoner: Nirx I think (?), maybe others. It would probably be better to make _check_fname more flexible for example by using op.exists rather than isfile or isdir.

rob-luke: Yes, nirx is a directory.

agramfort (approved):
I don't think CI failures are related

Good enough to go from my end. we should not release without this fix.

thx heaps @bloyl

mne/gui/_coreg.py
bloyl: `nirx` is supported via `read_raw` with the `.hdr` extension, so as a file. `coreg` users will need to pick the `hdr` file not just the directory for `nirx` datasets.

I did add a smirf (fixed a bug too) and nirx test though

@rob-luke, Its worth noting that I tried testing with
`snirf_nirsport2_20219_raw_path = op.join(data_path, 'SNIRF', 'NIRx', 'NIRSport2', '2021.9', '2021-10-01_002.snirf')` which threw an exception because the `Nasion` isn't defined in that file?

I agree about improving `_check_fname` but that would require an api change and be wide spread, so best left to another PR.

rob-luke: Thanks for doing the additional testing @bloyl. It's good to know that the file doesn't contain nasion, that might be something we need to deal with in the future, or petition the vendor to export it if it's something we need.

mne/gui/_coreg.py
```diff
@@ -215,10 +220,23 @@ def _set_current_fiducial(self, fid):
     def _set_info_file(self, fname):
         if fname is None:
             return
-        if not self._check_fif('info', fname):
+
+        # info file can be anything supported by read_raw
```
larsoner, line 224: On `main` it can actually be more than this -- it could be an evoked or epochs etc. -- really anything that stored an `info` in a `.fif`. This change makes it so that, for FIF, only the raw variants are supported. I'll push a change to fix this just based on a `splitext` check

larsoner: Oh never mind, I misunderstood this code -- looks like you do handle `fif` using a `read-info`-like operation below!

larsoner (approved):
Thanks @bloyl !

PR #9934, opened 2021-11-01
MRG: Improve random_state docstring & some other small doc fixes

hoechenberger:


mne/preprocessing/ica.py
```diff
@@ -186,9 +186,7 @@ class ICA(ContainsMixin):
         are scaled to unit variance ("z-standardized") as a group by channel
         type prior to the whitening by PCA.
     %(random_state)s
-        As estimation can be non-deterministic it can be useful to fix the
-        random state to have reproducible results.
-    method : {'fastica', 'infomax', 'picard'}
+    method : 'fastica' | 'infomax' |'picard'
```
larsoner, line 189: I think what we had before was fine, see https://numpydoc.readthedocs.io/en/latest/format.html#parameters

drammock: I think what we had before is *consistent with the numpydoc standard* but I thought we purposefully deviated (using pipes instead of "or" or set notation)?  We use pipes throughout most of our docs I think...  If we keep @hoechenberger's suggestion, there should be a space between the last pipe and `'picard'` though

hoechenberger: Yes we switched to using pipe chars ~everywhere. Coming across `{}` seemed inconsistent to me while readint the documentation, esp. since other params of the same class **do** use pipes.

hoechenberger: I've added the missing space, thanks for the pointer, @drammock

hoechenberger:
Rendered docs for ICA:
https://37344-1301584-gh.circle-artifacts.com/0/dev/generated/mne.preprocessing.ICA.html?highlight=ica#mne.preprocessing.ICA

drammock (approved):

larsoner:
Thanks @hoechenberger

PR #9935, opened 2021-11-01
Attempt to make "verbose" docstring more use-friendly

hoechenberger:
The docstring for the `verbose` kwargs was quite technical and, therefore, cryptic. I tried to make it more approachable for users.

Also replaced the upper-case log levels with lower-case names in the examples.

The docstring is longer now, but I think it will be much more helpful.

### DIFF
--- a/mne/utils/_logging.py
+++ b/mne/utils/_logging.py
@@ -78,9 +78,10 @@ def verbose(function: _FuncT) -> _FuncT:
     -----
     This decorator is used to set the verbose level during a function or method
     call, such as :func:`mne.compute_covariance`. The `verbose` keyword
-    argument can be 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL', True (an
-    alias for 'INFO'), or False (an alias for 'WARNING'). To set the global
-    verbosity level for all functions, use :func:`mne.set_log_level`.
+    argument can be ``'debug'``, ``'info'``, ``'warning'``, ``'error'``,
+    ``'critical``', ``True`` (an alias for ``'info'``), or ``False`` (an alias
+    for ``'warning'``). To set the global verbosity level for all functions,
+    use :func:`mne.set_log_level`.
 
     This function also serves as a docstring filler.
 
@@ -89,8 +90,8 @@ def verbose(function: _FuncT) -> _FuncT:
     You can use the ``verbose`` argument to set the verbose level on the fly::
 
         >>> import mne
-        >>> cov = mne.compute_raw_covariance(raw, verbose='WARNING')  # doctest: +SKIP
-        >>> cov = mne.compute_raw_covariance(raw, verbose='INFO')  # doctest: +SKIP
+        >>> cov = mne.compute_raw_covariance(raw, verbose='warning')  # doctest: +SKIP
+        >>> cov = mne.compute_raw_covariance(raw, verbose='info')  # doctest: +SKIP
         Using up to 49 segments
         Number of samples used : 5880
         [done]
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -24,10 +24,20 @@ docdict = dict()
 
 # Verbose
 docdict['verbose'] = """
-verbose : bool | str | int | None
-    If not None, override default verbose level (see :func:`mne.verbose`
-    and :ref:`Logging documentation <python:tut-logging>` for more).
-    If used, it should be passed as a keyword-argument only."""
+verbose : bool | 'debug' | 'info' | 'warning' | 'error' | 'critical' | int | None
+    Controls the verbosity of the generated logging output. Logging messages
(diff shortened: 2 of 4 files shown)


mne/utils/docs.py
```diff
@@ -24,10 +24,20 @@
...
+    ``verbose='warning'``, all ``debug`` and ``info`` messages will be hidden.
+
+    You may also pass a boolean, where ``True`` is an alias for ``'info'`` and
+    ``False`` is an alias for ``'warning'``. See also :func:`mne.verbose`.
+
+    If ``None``, use the default logging level.
+
+    Integers are accepted for consistency with the underlying Python logging
+    system.
+
```
larsoner, line 39: 99% of the time I think users will want to use `True` or `False`. I'd rather just document these as the basic options, and for other options refer them to the logging documentation as we do now. Otherwise all of our docstrings get a lot longer, which means much longer API pages, etc. with very little gain for the end users.

hoechenberger: I also feel that nobody will want to pass integers here these days 😅

hoechenberger:
@larsoner I've tried to address your comments, docstring now looks like this:

Screen Shot 2021-11-01 at 17 32 15

WDYT?

mne/utils/_logging.py
```diff
@@ -89,8 +90,8 @@ def verbose(function: _FuncT) -> _FuncT:
     You can use the ``verbose`` argument to set the verbose level on the fly::
 
         >>> import mne
-        >>> cov = mne.compute_raw_covariance(raw, verbose='WARNING')  # doctest: +SKIP
-        >>> cov = mne.compute_raw_covariance(raw, verbose='INFO')  # doctest: +SKIP
+        >>> cov = mne.compute_raw_covariance(raw, verbose='warning')  # doctest: +SKIP
+        >>> cov = mne.compute_raw_covariance(raw, verbose='info')  # doctest: +SKIP
```
hoechenberger, line 94:
```suggestion
        >>> cov = mne.compute_raw_covariance(raw, verbose=False)  # doctest: +SKIP
        >>> cov = mne.compute_raw_covariance(raw, verbose=True)  # doctest: +SKIP
```

hoechenberger: Then again … nah.

larsoner:
> WDYT?

Good except I wouldn't say "accepts all Python logging levels" because 1) I have never seen people actually use these, and 2) it's less complete than linking to the logging documentation. For example, the fact that you can do `verbose='error'` is not covered by the new docstring. Hence I think "... use the default logging level. See the <....> for more information" as we had before that links to *our* logging docs and gives all options (even the rarely used ones) is better

mne/utils/docs.py
```diff
@@ -25,9 +25,13 @@
...
-    If not None, override default verbose level (see :func:`mne.verbose`
-    and :ref:`Logging documentation <python:tut-logging>` for more).
-    If used, it should be passed as a keyword-argument only."""
+    Controls the verbosity of the generated logging output. If ``True``, sets
+    the log level to ``INFO``, and if ``False``, sets it to ``WARNING``. If
+    ``None``, use the default logging level. Accepts all
+    `Python logging levels <https://docs.python.org/3/library/logging.html#logging-levels>`_
+    (case ignored).
+
+    Should only be passed as a keyword argument."""  # noqa: E501
```
drammock, line 34:
```suggestion
    Controls the verbosity of logging output. Accepts all
    `Python logging levels <https://docs.python.org/3/library/logging.html#logging-levels>`_ (case ignored). For most use cases, setting either
    ``True`` (equivalent to "info") or ``False`` (equivalent to "warning") will
    suffice. ``None`` selects the default logging level. Can only be passed as
    a keyword (not positional) argument."""  # noqa: E501
```

suggested edit to be a little more succinct (?)

hoechenberger:
> as we had before that links to _our_ logging docs

except that it didn't, so apparently that was a bug and I thought it was intentional … it brought me to the python.org logging docs, which I thought were quite unhelpful 😅

Will think a bit about your other points!

hoechenberger:
> I have never seen people actually use these

Does this imply that we should consider deprecating accepting integers as input? Or at least kill that bit from the docs? If noone uses it, why keep it?

mne/utils/docs.py
drammock: oh, never mind this suggestion, I just saw @larsoner's point about referring to our logging docs instead of the python docs.

hoechenberger:
> except that it didn't, so apparently that was a bug and I thought it was intentional … it brought me to the python.org logging docs, which I thought were quite unhelpful 😅

Yeah this seems to be the root of the problem. There's an issue in the `main` docs, linking to https://docs.python.org/3/tutorial/stdlib2.html#tut-logging instead of our own logging docs. With that fixed, I think most of my issues will be sorted out…

hoechenberger:
Will open another PR to fix this.

Issue #9938, opened 2021-11-01
runtime warning in tutorial "iEEG localize" breaking docs build


drammock:
@alexrockhill one for you?

```
Traceback (most recent call last):
  File "/home/circleci/project/tutorials/clinical/10_ieeg_localize.py", line 430, in <module>
    montage_warped, elec_image, warped_elec_image = mne.warp_montage_volume(
  File "<decorator-gen-56>", line 24, in warp_montage_volume
  File "/home/circleci/project/mne/surface.py", line 1902, in warp_montage_volume
    _warn_missing_chs(montage, image_to, after_warp=True)
  File "/home/circleci/project/mne/surface.py", line 1747, in _warn_missing_chs
    warn('Channels ' + ', '.join(missing_ch) + ' were not assigned '
  File "/home/circleci/project/mne/utils/_logging.py", line 365, in warn
    warnings.warn_explicit(
RuntimeWarning: Channels LAMY 1, LPM 8 were not assigned voxels after applying SDR warp
```

https://app.circleci.com/pipelines/github/mne-tools/mne-python/11144/workflows/f60acec5-586e-4508-a3a3-72ebbced8ce6/jobs/37417

alexrockhill:
Sure, I'll have a look as soon as I can

PR #9940, opened 2021-11-02
[BUG, MRG] Actually fix circular bug this time

alexrockhill:
Fixes https://github.com/mne-tools/mne-python/issues/9938.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -315,7 +315,7 @@ Bugs
 
 - Fix channel locations with ``NaN`` values causing all channel locations not to be plotted in :func:`mne.viz.Brain.add_sensors` (:gh:`9911` by `Alex Rockhill`_)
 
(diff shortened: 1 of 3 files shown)


alexrockhill:
Ok very sorry, I should have checked more carefully and was in too much of a hurry, the previous fix avoided the circular issue but introduced a short-circuit. This one shouldn't give the warning and should work as intended and I wrote comments to hopefully make it easier to understand in the future.

agramfort (approved):

agramfort:
it's however a bit a bad pattern to merge code without tests...

@larsoner I let you merge if you thing it's ok.

@alexrockhill if we could have a test that would be awesome

larsoner:
> @alexrockhill if we could have a test that would be awesome

Agreed, any way a unit test could/should have caught this? Maybe we need another test case, or our existing test case wasn't checked for correctness properly?

larsoner:
I'm going to merge this so that we can get 0.24 out the door sooner, but @alexrockhill a test would be appreciated if you can make one in a separate PR!

alexrockhill:
Will do

PR #9944, opened 2021-11-02
BUG: Force mayavi in test that uses it

larsoner:
I need this locally for the test to pass, otherwise it uses the PyVista backend. Bug introduced in #9913, cc @GuillaumeFavelier

### DIFF
--- a/mne/conftest.py
+++ b/mne/conftest.py
@@ -420,7 +420,7 @@ def renderer_pyvistaqt(request, garbage_collect):
 
 @pytest.fixture(params=["mayavi"])
 def renderer_mayavi(request, garbage_collect):
-    """Yield the PyVista backend."""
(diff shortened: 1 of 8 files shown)


mne/conftest.py
```diff
@@ -418,6 +418,13 @@ def renderer_pyvistaqt(request, garbage_collect):
         yield renderer
 
 
+@pytest.fixture(params=["mayavi"])
+def renderer_mayavi(request, garbage_collect):
+    """Yield the PyVista backend."""
```
drammock, line 423:
```suggestion
    """Yield the renderer backend."""
```

agramfort (approved):

larsoner:
This reproduces the coreg failure @cbrnr is seeing, I'll try to fix it here

PR #9945, opened 2021-11-03
MRG, MAINT: Remove Mayavi

larsoner:
Starting post-release deprecations with the biggest one, separate from the others: mayavi. I got rid of all instances of `pysurfer` and `mayavi` from our code except the ones that are comments about copying/adapting code.

cc @GuillaumeFavelier we need a public equivalent to `mayavi.core.api.Scene` for whatever we currently return for `create_3d_figure` that users can pass around to 3D functions as `fig=fig` etc. Can you work on this separately? Then we can `git grep PyVista renderer` after this PR and know all the places we need to put in a reference that links properly.


doc/overview/roadmap.rst
```diff
@@ -191,20 +191,6 @@ limitations and challenges with it. We should work to use some other backend
...
-
-1. Refactor code to use traitlets
-2. GUI elements to use PyQt5 (rather than TraitsUI/pyface)
-3. 3D plotting to use our abstracted 3D viz functions rather than Mayavi
-4. Refactor distance/fitting classes to public ones to enable the example
-   from :gh:`6693`.
-
-Once this is done, we can effectively switch to a PyVista backend.
-
 .. _documentation-updates:
```
larsoner, line 194: Congrats @GuillaumeFavelier for taking care of a roadmap item !!!

agramfort:
🍾 🎉

PR #9948, opened 2021-11-03
FIX: Azure

larsoner:
I errantly disabled azure in #9687 (!), this should reenable it.


larsoner:
I put this in #9945 #9947 #9949 directly so I changed the base here to `maint/0.24`

PR #9950, opened 2021-11-03
[MAINT, MRG] Better test for `mne.surface._voxel_neighbors`

alexrockhill:
Fixes https://github.com/mne-tools/mne-python/pull/9940 leftovers.

This is much more thorough and has a more realistic use-case for the algorithm (i.e. previously cube contours of integer type were used but now the intensities are more reasonably normally distributed (the problem with the cube contours is that the corners were the same intensity as the sides when they should be farther from the origin of the intensity) and are continuous floating point numbers).

### DIFF
--- a/mne/tests/test_surface.py
+++ b/mne/tests/test_surface.py
@@ -7,7 +7,6 @@ import os.path as op
 import pytest
 import numpy as np
 from numpy.testing import assert_array_equal, assert_allclose, assert_equal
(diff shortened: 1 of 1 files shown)


alexrockhill:
It doesn't look like the test failures are relevant, this is okay by me unless there are suggestions for more tests

agramfort (approved):

agramfort:
thx @alexrockhill

PR #9951, opened 2021-11-03
MRG, FIX: Fix PG tests

larsoner:
I think we should make OpenGL optional, and only try to use it based on a config var. But in the meantime, this will fix our PG tests after https://github.com/mne-tools/mne-qt-browser/pull/12 / 0.1.5


larsoner:
The one changed CI is green, merging

PR #9952, opened 2021-11-04
[ENH, MRG] Add local maxima button to ieeg locate GUI

alexrockhill:
This adds the capability to show in yellow which voxel is a local maxima (see below). This functionality is very useful for bridged electrodes where otherwise, the intensity values would have to be looked at carefully and/or the color scale changed (which has the added detriment of making it hard to see the context outside of the electrode) in order to determine the center of intensity of the electrode contact with assurance. This coordinates nicely with our method of warping which uses voxels monotonically decreasing from the local maximum (at least until Dipy finishes their PR https://github.com/dipy/dipy/pull/2369 to make it so that points can be warped). Lastly, the maximum computation is done efficiently only whenever the radius changes and the local maxima showing is turned on (it uses the radius to determine how far to count as local).

Screen Shot 2021-11-03 at 6 20 24 PM

### DIFF
--- a/mne/gui/tests/test_ieeg_locate_gui.py
+++ b/mne/gui/tests/test_ieeg_locate_gui.py
@@ -186,5 +186,6 @@ def test_ieeg_elec_locate_gui_display(_locate_ieeg, _fake_CT_coords):
     # test buttons
     gui._toggle_show_brain()
     assert 'mri' in gui._images
+    assert 'local_max' not in gui._images
     gui._toggle_show_max()
     assert 'local_max' in gui._images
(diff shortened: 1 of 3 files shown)


mne/gui/tests/test_ieeg_locate_gui.py
```diff
@@ -182,3 +182,9 @@ def test_ieeg_elec_locate_gui_display(_locate_ieeg, _fake_CT_coords):
     ct_sum_before = np.nansum(gui._images['ct'][0].get_array().data)
     gui._ct_min_slider.setValue(500)
     assert np.nansum(gui._images['ct'][0].get_array().data) < ct_sum_before
+
+    # test buttons
+    gui._toggle_show_brain()
+    assert 'mri' in gui._images
```
larsoner, line 188:
```suggestion
    assert 'mri' in gui._images
    assert 'local_max' not in gui._images
```

agramfort (approved):

larsoner:
Thanks @alexrockhill !

PR #9960, opened 2021-11-04
ENH: Make Qt default browser, add scraper

larsoner:
Let's see how it works. I think it won't work well because of https://github.com/mne-tools/mne-qt-browser/issues/19, but otherwise should be functional.


larsoner:
On [main](https://mne.tools/dev/auto_tutorials/intro/10_overview.html):



On [this PR](https://37792-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/10_overview.html):



so it works, except for (I think) https://github.com/mne-tools/mne-qt-browser/issues/19. Leaving this as draft and I'll try running it again once mne-qt-browser is fixed. Or we can close if we don't want it to be our default scraped view.

mne/viz/_scraper.py
```diff
@@ -0,0 +1,39 @@
...
+                img_fname = next(block_vars['image_path_iterator'])
+                inst = QApplication.instance()
+                assert inst is not None
+                for _ in range(30):  # max 30 sec
+                    inst.processEvents()
+                    inst.processEvents()
+                    if gui.mne.data_precomputed:
+                        inst.processEvents()
+                        break
+                    time.sleep(1.)
```
marsipu, line 31: With the threading reworked I think we can simplify that now to:
```suggestion
            if hasattr(self, 'load_thread') and self.load_thread is not None:
                if self.load_thread.isRunning():
                    gui.load_thread.wait(30000)
```

larsoner:
@marsipu feel free to push directly here, you can even install your branch of mne-qt-browser for testing purposes to CircleCI in this PR if you want. Please push with `[skip azp] [skip actions]` in your commit message, though, so that only CircleCI bothers to run for now

tools/circleci_dependencies.sh
```diff
@@ -31,7 +31,7 @@ elif [[ "$CIRCLE_JOB" == "linkcheck"* ]]; then
 else  # standard doc build
 	echo "Installing doc build dependencies"
 	python -m pip uninstall -y pydata-sphinx-theme
-	python -m pip install --upgrade --progress-bar off --only-binary matplotlib -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
+	python -m pip install --upgrade --progress-bar off --only-binary matplotlib mne-qt-browser -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
```
larsoner, line 34: for example you can replace `mne-qt-browser` here with `https://github.com/marsipu/mne-qt-browser/zipball/whateverbranchname` and it should use your threading branch (if you set the name properly)

larsoner:
```suggestion
	python -m pip install --upgrade --progress-bar off --only-binary matplotlib https://github.com/marsipu/mne-qt-browser/zipball/fix_threading -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
```

mne/viz/_scraper.py
```diff
@@ -0,0 +1,36 @@
...
+                assert inst is not None
+            if (hasattr(self, 'load_thread')
+                    and self.load_thread is not None):
+                if self.load_thread.isRunning():
+                    gui.load_thread.wait(30000)
+                pixmap = gui.grab()
+                pixmap.save(img_fname)
+                gui.close()
+                inst.processEvents()
+                inst.processEvents()
```
marsipu, line 33: Is this still necessary?
```suggestion
```

larsoner: No idea, when I was testing locally before I think it was. Locally you can test in the `doc/` directory with `PATTERN=10_overview make html_dev-pattern` and it will rebuild just the one example for you.

Hopefully what I just pushed will at least give the right result, then we can try removing these and let CircleCI tell us if they were necessary

marsipu: I would like to test but for some reason I got now this error and I don't understand what's the problem.
```python
Configuration error:
There is a programmable error in your configuration file:

  File "C:\Users\marti\anaconda3\envs\mnedev\lib\site-packages\sphinx\config.p
y", line 328, in eval_config_file
    exec(code, namespace)
  File "C:\Users\marti\PycharmProjects\mne-python\doc\conf.py", line 26, in <m
odule>
    from mne.tests.test_docstring_parameters import error_ignores
ModuleNotFoundError: No module named 'mne.tests'
```
Maybe you can help me?

larsoner: Do you have the dev version installed with setup.py develop or pip install -e ?

marsipu: Oh yes, I installed from the wrong folder. Thank you!

larsoner:
I found a bug, about to push...

larsoner:
Okay things build well locally for me now @marsipu, I think I'm done pushing for a bit. Feel free to push if it doesn't work for you!

I'll leave this as WIP until we finalize the threading PR and hopefully cut a new release of mne-qt-browser so that we can just pip install the stable version here

marsipu:
I had some little fixes but when I try to push to your fork, I get `Permission to larsoner/mne-python.git denied to marsipu`.
What am I doing wrong?

marsipu:
I also noticed that the scraped images were still showing the progressbar, so I added `QTest.qWait(100)` to give the gui some time to update

larsoner:
You should have permission now if you accept my collaboration invite

larsoner:
> I also noticed that the scraped images were still showing the progressbar, so I added QTest.qWait(100) to give the gui some time to update

Looks like adding one `inst.processEvents()` ~~is enough to fix it locally at least~~ [made the build hang](https://app.circleci.com/pipelines/github/mne-tools/mne-python/11653/workflows/4610251e-c078-491f-aac4-476cc38f6d20/jobs/38986), so I pushed one with more. I like the direct `processEvents` approach rather than the `QTest.qWait` one, since I guess during the wait time it might or might not process events? In any case, it looks like CircleCI [is happy with the latest version](https://38992-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/10_overview.html#sphx-glr-auto-tutorials-intro-10-overview-py), and locally the windows appear, get scraped, and close, which is nice and mimics the behavior of matplotlib and PyVista. The CircleCI output looks okay, too:

tools/circleci_dependencies.sh
```diff
@@ -33,7 +33,7 @@ elif [[ "$CIRCLE_JOB" == "linkcheck"* ]]; then
 else  # standard doc build
 	echo "Installing doc build dependencies"
 	python -m pip uninstall -y pydata-sphinx-theme
-	python -m pip install --upgrade --progress-bar off --only-binary matplotlib -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
+	python -m pip install --upgrade --progress-bar off --only-binary matplotlib https://github.com/marsipu/mne-qt-browser/zipball/fix_threading -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
```
marsipu, line 36: Reminder: Needs to be reverted before merging. This PR should be merged after mne-qt-browser was bumped to the next version (0.1.7)
```suggestion
	python -m pip install --upgrade --progress-bar off --only-binary matplotlib mne-qt-browser -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt
```

marsipu:
Should the scraper now also work for the other tutorials? Because in CircleCi it's not (e.g. for "Built-in plotting methods for Raw objects/40_visualize_raw"). Locally it is working:

## CircleCI
grafik
## Local build
grafik

drammock:
> Should the scraper now also work for the other tutorials?

Circle CI only builds tutorials that have changed. The are no changes in that file

marsipu:
Thanks @drammock , good to know

larsoner:
We forgot to decide at the dev meeting if we're going to make `mne-qt-browser` the default for MNE 1.0 :(

If we do that, we might want to resurrect this PR, too. So I'm leaving this marked as 1.0 just so we remember to discuss.

@agramfort any opinion?

.circleci/config.yml
```diff
@@ -106,6 +106,7 @@ jobs:
               echo "export MNE_3D_OPTION_ANTIALIAS=false" >> $BASH_ENV
               source tools/get_minimal_commands.sh
               echo "export MNE_3D_BACKEND=pyvistaqt" >> $BASH_ENV
+              echo "export MNE_BROWSER_BACKEND=pyqtgraph" >> $BASH_ENV
```
marsipu, line 109:
```suggestion
              echo "export MNE_BROWSER_BACKEND=qt" >> $BASH_ENV
```

doc/conf.py
```diff
@@ -354,6 +354,8 @@ def __call__(self, gallery_conf, fname):
     report_scraper = mne.report._ReportScraper()
     scrapers += (report_scraper,)
     del backend
+if mne.viz.get_browser_backend() == 'pyqtgraph':
```
marsipu, line 357:
```suggestion
if mne.viz.get_browser_backend() == 'qt':
```

marsipu (changes requested):
adjust for backend-renaming

larsoner:
@marsipu do you want to try changing the default from matplotlib to `qt` here? I think we want some behavior where by default it tries to use `'qt'` (unless the user has MNE_BROWSE_BACKEND=matplotlib set) and falls back to matplotlib if not available

marsipu:
Yes 😃

larsoner:
@marsipu CircleCI insists on failing:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/12948/workflows/0d40575e-ea9d-4c3f-8249-91c14d9db56b/jobs/41827

I just reverted a bunch of commits, but then I just realized the problem might actually be memory consumption. If this build with the commits reverted *also* fails, I'll try force-pushing back to what I think should have worked before the reverts (80ebe18) and then push something to prevent precomputation.

larsoner:
... actually I just noticed that the problem is a timeout, so it "just" gets stuck for 10 minutes, then dies. I think this is probably a variant of the old precomputation threading bug :(

I added config var support for precompute and turned it to False on CircleCI, let's see if it works...

marsipu:
Thank you @larsoner for looking into this problems, I hope disabling precompute solves the issue here.

mne/viz/raw.py
```diff
@@ -301,6 +301,13 @@ def plot_raw(raw, events=None, duration=10.0, start=0.0, n_channels=20,
 
     # gather parameters and initialize figure
     _validate_type(use_opengl, (bool, None), 'use_opengl')
+    _validate_type(precompute, (bool, str, None), 'precompute')
+    if precompute is None:
+        precompute = get_config('MNE_BROWSER_PRECOMPUTE', 'auto').lower()
+        _check_option('MNE_BROWSER_PRECOMPUTE',
+                      precompute, ('true', 'false', 'auto'),
+                      extra='when precompute=None is used')
+        precompute = dict(true=True, false=False, auto='auto')[precompute]
```
marsipu, line 310: Shouldn't this be also added to `mne.viz.epochs.py` and `mne.viz.ica.py`?

larsoner: Yep, I can do that

larsoner:
Looks like `MNE_BROWSER_PRECOMPUTE=false` on CircleCI worked:

https://41837-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/intro/10_overview.html#sphx-glr-auto-tutorials-intro-10-overview-py

@marsipu okay for you?

If so I'll look into any last failing tests, make the change you suggested, and mark as ready for merge.

larsoner:
> when you are handling files with different sizes, you probably want to set precompute in your code rather than changing the config-value everytime, right?

Yes. In most cases where there is a config var for something, we allow passing it as well. It allows for various levels of control It has been pretty useful over the years for example with SUBJECTS_DIR/subjects_dir

larsoner:
@marsipu it looks like the failures are real and related at least in part to the default-backend switch. For example now we get:
```
E   UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
```
which we shouldn't get -- any time we `show` a matplotlib figure, we should use our internal `plt_show` that catches this warning. It means that we are somehow hitting the wrong branch here, so I pushed a commit to expose the problem (these asserts should pass, but don't): 8870687ae46d855cbfb0b6cae2c3cf44b2056ef3

Can you look to see if this makes sense to you?

marsipu:
I am not sure about the `matplotlib`-problem yet (as you say, somehow `if backend == 'matplotlib':` in l.139 of `mne.viz.utils` does not work).
But I also saw that there are apparently tests with plot outside of `mne.viz.tests` like `mne.preprocessing.tests.test_ica.py::test_ica_core.py` which are not adapted yet.

I don't know if I can manage to look into this today because I am not finished with exam-preparation for today yet.

larsoner:
> But I also saw that there are apparently tests with plot outside of mne.viz.tests like mne.preprocessing.tests.test_ica.py::test_ica_core.py which are not adapted yet.

Makes sense, any test that uses `ica.plot_sources`, `epochs.plot`, and `raw.plot` (and related functions) probably needs to be parameterized over backend types.

Okay. The easy way out of this I think is to actually set `MNE_BROWSE_BACKEND=matplotlib` in `conftest.py` until we adapt all tests. We can probably even do this after release if we must, assuming that ICA and epochs are already somewhat well tested by the previous PRs that added support for them. I'll try pushing that here and see if things work.

marsipu:
>Makes sense, any test that uses ica.plot_sources, epochs.plot, and raw.plot (and related functions) probably needs to be parameterized over backend type

Yes, and I think some tests may use `matplotlib`-specific methods and attributes (e.g. fig.mne.ax_main) which need to be adapted to fit both backends (as I did for the viz-tests)

mne/utils/docs.py
```diff
@@ -1581,11 +1581,14 @@ def _reflow_param_docstring(docstring, has_first_line=True, width=75):
     Whether to load all data (not just the visible portion) into RAM and
     apply preprocessing (e.g., projectors) to the full data array in a separate
     processor thread, instead of window-by-window during scrolling. The default
-    ``'auto'`` compares available RAM space to the expected size of the
-    precomputed data, and precomputes only if enough RAM is available. ``True``
-    and ``'auto'`` only work if using the Qt backend.
+    None uses the ``MNE_BROWSER_PRECOMPUTE`` variable, which defaults to
+    `'auto'``. ``'auto'`` compares available RAM space to the expected size of
```
marsipu, line 1585:
```suggestion
    ``'auto'``. ``'auto'`` compares available RAM space to the expected size of
```

marsipu:
I am currently working on the tests

larsoner:
Looks like maybe just one failure left:
```
_______________________________ test_pg_scraper _______________________________
mne\viz\tests\test_scraper.py:20: in test_pg_scraper
    mne.viz._scraper._PyQtGraphScraper()(None, block_vars, gallery_conf)
mne\viz\_scraper.py:22: in __call__
    if getattr(gui, '_scraped', False):
E   RuntimeError: super-class __init__() of type PyQtGraphBrowser was never called
```

marsipu:
Hm, this could be a timing issue because browser-instances are added to _browser_instances shortly **before** the parent-classes are called. I fill fix and merge this in mne-tools/mne-qt-browser#56

larsoner:
> Bu maybe this could be a timing issue because browser-instances are added to _browser_instances shortly before the parent-classes are called.

I'm not sure how it's possible unless `__init__` somehow gets called from another thread...? It seems ilke the `_get_browser(...)` call cannot finish until `__init__` has been called, right?

marsipu:
> > Bu maybe this could be a timing issue because browser-instances are added to _browser_instances shortly before the parent-classes are called.
> 
> I'm not sure how it's possible unless `__init__` somehow gets called from another thread...? It seems ilke the `_get_browser(...)` call cannot finish until `__init__` has been called, right?

Oh you're right. Maybe there are somehow remnant-objects in `_browser_instances` from the other tests which are already deleted in C++ and doesn't count as initialized anymore? How would we test for this in the scraper?

larsoner:
> Maybe there are somehow remnant-objects in _browser_instances from the other tests which are already deleted in C++ and doesn't count as initialized anymore? How would we test for this in the scraper?

Indeed I just had the same thought -- some stale ref and the error message is misleading.

> How would we test for this in the scraper if getattr() isn't working?

b8c324853

larsoner:
... is it possible that the instance adds itself to the list in `__init__` before `super().__init__(...)` is called, and in between those two, an error of some sort could occur? In that case I think the fix is to move the "add-to-instances-list" to *after* the super init

larsoner:
... BTW I'm going to see if people are okay with moving 1.0 out to next week. It would be good to have at least a couple days of all these changes + installer testing done before we release.

marsipu:
> ... is it possible that the instance adds itself to the list in `__init__` before `super().__init__(...)` is called, and in between those two, an error of some sort could occur? In that case I think the fix is to move the "add-to-instances-list" to *after* the super init

Yes, if something happens during parent-init this might be. Probably we should se this in the tests too but I added a change for that in mne-tools/mne-qt-browser#56

larsoner:
Looks like this Linux CI probably also needs a PRECOMPUTE=false, this looks like a plotting hang:

https://github.com/mne-tools/mne-python/runs/5224295329?check_suite_focus=true#step:13:415

I pushed a commit to see if it fixes it, plus a change that hopefully will fix the stale window bug. Though thinking about it now, it's likely that there is just some test / there are some tests that still need to be decorated with the `pg_backend` or parametrized over backends. That would explain there being more than one instance in the scraper test. Assuming `[circle full]` comes back happy, I'll push a commit to make sure that does a `backend._close_all()` *before* the `yield` in addition to the one after the `yield` -- that should help the scraper test not have anything stale afterward.

larsoner:
@marsipu another one to look at sometime:
```
____________________________ test_array_epochs[qt] _____________________________
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/matplotlib/figure.py:2364: in show
    self.canvas.manager.show()
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/matplotlib/backend_bases.py:2815: in show
    raise NonGuiException(
E   matplotlib.backend_bases.NonGuiException: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.

During handling of the above exception, another exception occurred:
mne/tests/test_epochs.py:2769: in test_array_epochs
    epochs[0].plot()
mne/epochs.py:1167: in plot
    return plot_epochs(self, picks=picks, scalings=scalings,
mne/viz/epochs.py:910: in plot_epochs
    _show_browser(show, block=block, fig=fig)
mne/viz/utils.py:144: in _show_browser
    fig.show()
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/matplotlib/figure.py:2367: in show
    _api.warn_external(str(exc))
/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/matplotlib/_api/__init__.py:299: in warn_external
    warnings.warn(message, category, stacklevel)
E   UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
```
In the meantime I'm going to see if a naive `conftest` ignore will be enough to make things work. If it's not, we can set the default backend to matplotlib in the `Test Ultraslow_PG` test where this occurred.

larsoner:
Okay we already have one legitimate failure that we have to fix:
```
WARNING: /home/circleci/project/tutorials/intro/70_report.py failed to execute correctly: Traceback (most recent call last):
  File "/home/circleci/project/tutorials/intro/70_report.py", line 93, in <module>
    report.add_raw(raw=raw, title='Raw', psd=False)  # omit PSD plot
  File "/home/circleci/project/mne/report/report.py", line 1086, in add_raw
    htmls = self._render_raw(
  File "/home/circleci/project/mne/report/report.py", line 2729, in _render_raw
    butterfly_imgs_html = self._render_raw_butterfly_segments(
  File "/home/circleci/project/mne/report/report.py", line 2676, in _render_raw_butterfly_segments
    _constrain_fig_resolution(
  File "/home/circleci/project/mne/report/report.py", line 327, in _constrain_fig_resolution
    dpi = min(max_res, max_width / fig.get_size_inches()[0])
AttributeError: 'PyQtGraphBrowser' object has no attribute 'get_size_inches'
```
The easy fix here is to `use_browse_backend` in `report.py`. I'll push that

larsoner:
Looks like we have a timeout in 60_ctf_bst_auditory.py:

https://app.circleci.com/pipelines/github/mne-tools/mne-python/12966/workflows/92895aa4-1fe2-4e26-a259-0d256a995ac8/jobs/41862

@marsipu one option is to for now set `MNE_BROWSER_BACKEND=matplotlib` in CircleCI in this PR. It somewhat defeats the purpose, but it relieves the timeline stress here. That way we can push out 1.0, and weeks (or months!) later update our doc builds to show the Qt browser in a `1.0.1` release. WDYT?

Then all we have to fix here are the unit tests, and the point of the PR becomes mostly about setting the default to `qt`.

marsipu:
> @marsipu one option is to for now set MNE_BROWSER_BACKEND=matplotlib in CircleCI in this PR. It somewhat defeats the purpose, but it relieves the timeline stress here. That way we can push out 1.0, and weeks (or months!) later update our doc builds to show the Qt browser in a 1.0.1 release. WDYT?

Yes I think too that we could use more time to inspect all tutorials/examples or find a general solution for the issues arising with using `qt`-backend for rendering in all docs.

marsipu:
I just uploaded a fix which I think was causing `test_epochs_array` to fail

marsipu:
> Looks like we have a timeout in 60_ctf_bst_auditory.py:

In this tutorial I see in the middle:
```python
raw.plot(block=True
```

Theoretically this must have blocked the CI for matplotlib too (maybe this is then one of those constellations where block doesn't work for matplotlib). 
How would you solve this? (setting all plots non-blocking by default for CIs maybe?)

marsipu:
Just as a reminder: We might want to run the CIs once again here when we released `0.2.0` of mne-qt-browser before merging to run everything which is now skipped because of low version.

larsoner:
Ahh, `block=True` might also be why the Linux job dies after the `channels` tests, because the next one is probably `commands` and tests `mne browse_raw`, which runs with `block=True`. 

> Theoretically this must have blocked the CI for matplotlib too (maybe this is then one of those constellations where block doesn't work for matplotlib).

For tests and doc builds we force matplotlib to use `agg`, so `block` gets ignored.

> How would you solve this?

For now I'm going to try putting in a shim workaround env var `_MNE_QT_NO_BLOCK=true` that we can set in `conftest.py` and `conf.py` and check for wherever we would normally pass `block=True` to the Qt backend to make it so that it's always False when that env var is set.

marsipu:
What does "[circle full]" mean?

tutorials/preprocessing/20_rejecting_bad_data.py
```diff
@@ -49,7 +49,7 @@
 # existing annotation labels can be selected for use.
 
 fig = raw.plot()
-fig.canvas.key_press_event('a')
+fig._fake_keypress('a')
```
larsoner, line 52: Hmm... we shouldn't use private methods in tutorials. We can make this public I think

larsoner:
I should update the docs to say that by default for PRs Circle will only render things that change, but for your question:

https://github.com/mne-tools/mne-python/blob/ffb4ef3d623b9e7202e5078cd3301c8cfc40d5a5/.circleci/config.yml#L1-L6

larsoner:
This PR is exposing all sorts of fun stuff! https://github.com/sphinx-gallery/sphinx-gallery/pull/916

marsipu:
When do you think is the best time to do the `0.2.0`-release for `mne-qt-browser`?

larsoner:
You can go ahead and do it now. If we need to push out 0.2.1 almost immediately it's easy enough. Version numbers are cheap :)

marsipu:
🎉

larsoner:
This was not a circle full run, still working on tutorials locally. Hopefully just another hour or two, then I can push `[circle full]` and we can hope it works!

larsoner:
So close, it only had a few examples left!
```
generating gallery for auto_examples/datasets... [ 80%] opm_data.py
4672.6 s : 3.02 GB    
ERROR:mprof:Program resulted with a non-zero exit code: -9
Makefile:55: recipe for target 'html_dev-memory' failed
```
It suggests that the pyqtgraph windows are for some reason not being garbage collected properly, probably due to some circular refs that `gc` is not able to resolve :(

For now I'm going to set CircleCI to use matplotlib, but I'll see if I can figure out the memory leak somehow

PR #9963, opened 2021-11-04
fix missing connectivity image on homepage tile

drammock:
not sure when this crept in but due to the recent release now it's in both stable and dev sites, so I'm marking as backport candidate.


drammock:
rendered docs are good, merging

PR #9966, opened 2021-11-05
MRG, BUG: Fix regression with report due to bbox_inches

larsoner:
We shouldn't force `bbox_inches='tight'`. The effect is generally small, and if it's big, then people should fix their figures not to waste space. Here's an example of a bad effect (jumping around) on `main`:

Peek 2021-11-05 11-00

And it's fixed on this PR (and was fine before the big refactor):

Peek 2021-11-05 11-06


drammock (approved):

drammock:
```
changes/latest.inc:33: WARNING: py:class reference target not found: mne.report.Report
```

agramfort (approved):
@hoechenberger I let you merge this

hoechenberger (approved):

hoechenberger:
Thank you, @larsoner!

Issue #8376, opened 2020-10-15
ENH: Preserve annotations in epochs


adam2392:
#### Describe the new feature or enhancement
Is there a way to map annotations from the raw dataset onto the AverageTFR object? Since the tfr is occurring over windows, then the time scale of the annotations from the original time scale are inherently different.

E.g. if I have an event at sample 10000, with sfreq=1000, then it occurs at 10 seconds into the raw dataset. If i have an averageTFR that was computed with a moving window of say 1 second step size, then the new sample point of the event would be 10.

#### Describe your proposed implementation
Ideally this would consist of two things that I could see being useful:

1. A public function `map_annotations_to_sfreq` that allows mapping `Annotations` to different time-scales based on a passed say... `sfreq` value. So based on the above example, if I wanted to map `Annotations` that are based on `sfreq=1000`Hz, to some tfr object with `sfreq=1` Hz, then I would somehow pass that information to `map_annotations_to_sfreq` that returns me a new `Annotations` object that has `onset` and `duration` in the new time-scale.

2. Use the public function above to make `AverageTFR`, or `EpochsTFR`, etc. to keep their `Annotations` when the time-scale changes. 

I strongly would prefer having 1 because besides mne-python objects, I create other derivatives of the raw data that inherently have a "sliding" window component to them, and hence change the time-scale in the same way as a sliding FFT would.

#### Additional comments
@larsoner's comment on gitter for reference:

```
@adam2392 in principle we should be able to do that at least for EpochsTFR, can you open an issue feature request? For AverageTFR I'm not sure because we don't preserve annotations in Evoked even (just Raw and Epochs) so we'd probably want to do that first before doing it for AverageTFR
```

larsoner:
> A public function map_annotations_to_sfreq that allows mapping Annotations to different time-scales

I actually don't think we need this. Annotations [store data in seconds](https://mne.tools/stable/generated/mne.Annotations.html) not samples -- so they have no notion of (or need for) a `sfreq` (even if `sfreq` and sample numbers are used in the `Raw` instance to create the times in seconds). So I think all we really need to do is ...

> Use the public function above to make AverageTFR, or EpochsTFR, etc. to keep their Annotations when the time-scale changes.

... make EpochsTFR (and the functions that create them) respect the annotations properly.

mmagnuski:
That would be useful! I would first start with `.annotations` for Epochs as currently this attribute is only available for raw.
[...]

PR #9969, opened 2021-11-05
[ENH] Enable Epochs to store Annotations from Raw

adam2392:
#### Reference issue
First step for #8376 

- Implements `_set_annotations` inside Epochs data structures that get made from `Raw`, which essentially sets a copy of the `Raw.annotations` inside Epochs.
- Epochs now stores `_raw_sfreq` and `_first_samp` from Raw, which are used to reconstruct where in time 
- Epochs also has a `get_epoch_annotations` function that gets the Annotations as a nested list of 3-tuples, corresponding to onset, duration, description. The onset is relative to the Epoch the Annotation it is in now.
- Epochs also has a `map_annots_to_metadata` function that creates a new `Annotations` column inside `Epochs.metadata`, which has a list of 3-tuples per row corresponding to the onset, duration and description of each Annotation occurring within the Epoch. 

##### Changes
Setting annotations in Epochs relies on the assumption that Annotations MUST be set on the Raw object. Then when constructing an Epochs data structure from Raw, Annotations are passed through via `_set_annotations` private function.

Then we can enable two new Epochs functions:
- `get_epoch_annotations`, which obtains a list of `n_epochs` long containing tuples representing (onset relative to epoch times[0], duration, description)
- `map_annots_to_metadata`, which augments the `metadata` data frame with a new `Annotations` column containing the list of tuples (onset, duration, description) for each row of epochs.

##### Unit Tests
To walk you through the unit tests, there is the following cases:

- without meas_date, no first_samp, no orig_date
- without meas_date, first_samp, no orig_date
- with meas_date, no first_samp, no orig_date
- with meas_date, no first_samp, orig_date
- with meas_date, first_samp, no orig_date
- with meas_date, first_samp, orig_date

Note without `meas_date` and with `orig_date` results in error as it should when you do `raw.set_annotations`.

#### Additional information
The next step would be to incorporate IO for the new type of Epochs, by making a FIFF constant for `_raw_sfreq` and `_first_samp`.

Then the next step would be to test this functionality with something like `mne-connectivity`. 

Then the next step would be to incorporate concatenation ability.


larsoner:
> The main thing I need to handle is getting the analogous self._first_time in Raw from Epochs. I think... this is possible by using the self.events array and then the self.info['sfreq']?

No, we'll need to add a private `epochs._first_time` attribute (float) that comes from `raw._first_time` when Epochs are constructed, and make sure it survives round-trip IO (probably by using the FIRST_SAMP) tag.

adam2392:
> ... and we should probably also need a `self._raw_sfreq` or so that keeps track of the original sfreq from `raw` used so that `events[:, 0]` can still be used to get back to raw object times even after Epochs are resampled or decimated (both of which change `epochs.info['sfreq']`, so it's not safe just to use that)

Added this in.

> No, we'll need to add a private `epochs._first_time` attribute (float) that comes from `raw._first_time` when Epochs are constructed, and make sure it survives round-trip IO (probably by using the FIRST_SAMP) tag.

Added `epochs._first_time`. Re round-trip IO, are you referring to these lines for the saving of the file?

https://github.com/mne-tools/mne-python/blob/bdedc788b79f7b620167ba6358965ce312ec51be/mne/epochs.py#L156-L160

Then, where would it be necessary for the reading?

adam2392:
Another edge-case, I suspect I will have to implement logic in the concatenation of epochs too for Annotations?

For Annotations, should I implement them into EpochsArray, or just the Epochs from Raw object?

larsoner:
Ahh we already use FIRST_SAMPLE for `epochs.tmin`, I was hoping we could use it for `epochs._first_time` / `epochs._first_samp` but we can't. I think we probably want to add new constants:
```
FIFF.FIFF_MNE_EPOCHS_RAW_SFREQ = 3803
FIFF.FIFF_MNE_EPOCHS_RAW_FIRST_SAMP = 3804
```
@agramfort WDYT?

If people agree then this will require a PR to mne-tools/fiff-constants first

larsoner:
> Another edge-case, I suspect I will have to implement logic in the concatenation of epochs too for Annotations?

Yes and It's going to be a bit tricky. For now I would just `warn` that it's not supported (yet) and clear the annotations from all objects if they exist. Then we can implement it later.

> For Annotations, should I implement them into EpochsArray, or just the Epochs from Raw object?

Just from raw for now I think

agramfort:
> Ahh we already use FIRST_SAMPLE for `epochs.tmin`, I was hoping we could use it for `epochs._first_time` / `epochs._first_samp` but we can't. I think we probably want to add new constants:
> 
> ```
> FIFF.FIFF_MNE_EPOCHS_RAW_SFREQ = 3803
> FIFF.FIFF_MNE_EPOCHS_RAW_FIRST_SAMP = 3804
> ```
> 
> @agramfort WDYT?

Why is this needed?

larsoner:
> Why is this needed?

To go from `events[:, 0]` to actual times you need `raw.first_samp` plus `raw.info['sfreq']`. (And every attribute needs a FIFF constant to survive IO round-trip.)

In theory we could maybe get away with not storing the original data sample rate `raw.info['sfreq']` but we'd need to do things like `events[:, 0] = np.round(events[:, 0] / decim).astype(int)` when doing `decimate` and `resample`. It seems easier (for now at least) just to say "event samples correspond to those of the raw data" and then not modify them.

drammock:
Random idea, not thoroughly treated: annotations object gets a private attribute `_samps_changed` that is set to true if the epochs are resampled or decimated, and causes a warning if those annotations are added back to a raw. WDYT?

larsoner:
I'm not sure why we need it / what it fixes. The new attributes are needed to go to/from epochs to actual times. Either way, Annotations are always in units of actual times (not samples or anything) so I don't think it helps or adds much information.

agramfort:
>
> To go from events[:, 0] to actual times you need raw.first_samp plus
> raw.info['sfreq']. (And every attribute needs a FIFF constant to survive
> IO round-trip.)
>

really? the events are informed about the first_samp (it's as an offset of
events[:, 2]) and the meas_date as the same time origin AFAIK

>

larsoner:
> really? the events are informed about the first_samp (it's as an offset of events[:, 2]) and the meas_date as the same time origin AFAIK

Ahh okay. Then I think we just need `FIFF.FIFF_MNE_EPOCHS_RAW_SFREQ = 3803`, or we need to change `epochs.events[:, 0]` any time we resample or decimate Epochs.

agramfort:
but now we need to check what happens if meas_date is None and we have a first_samp != 0. MAybe it's the source of our Annotations pbs...

larsoner:
> but now we need to check what happens if meas_date is None and we have a first_samp != 0. MAybe it's the source of our Annotations pbs...

Could be. If that's the case maybe we do need the `FIFF.FIFF_MNE_EPOCHS_RAW_FIRST_SAMP = 3804`. My point is really that we're going to need at least one, maybe two constants for round-trip I/O...

Unfortunately given how tricky the meas_date/first_samp stuff is, we probably need to work all that stuff out *in this PR* (e.g., with various types of `raw` with and without `meas_date` and with zero and non-zero `first_samp`) in order to know which constants we need.

adam2392:
Okay, so from what I understand: I need to:
1. handle `meas_date`, `first_samp` properly when they are either set to a non-trivial number, or not set at all. 
2. handle FIFF constants to make `read_epochs` and `epochs.save()` work.
3. Add `map_annots_to_metadata()` convenience function to allow annotations to be mapped into the `epochs.metadata` property.

For 1. I have that handled right now by mainly copying the `raw.set_annotations` function. 

For 2, I think we need to keep both `raw._first_time` and the `raw.info['sfreq']` because yes i) events[:,0] are offsets from `raw.first_time` and then the raw sfreq is used to determine the actual time points of the Epochs on continuous time.

For 3, I am pushing up a commit soon.

adam2392:
Also a quick question regarding how to push Annotations into the metadata df:

Since the metadata df requires one row per Epoch, do I just create as many columns as there are Annotations? So each Annotation results in 2 new columns named: `Annotation_<description>_onset`, `Annotation_<description>_duration`, where the `description` is encoded in the column name?

larsoner:
> Since the metadata df requires one row per Epoch, do I just create as many columns as there are Annotations? So each Annotation results in 2 new columns named: Annotation_<description>_onset, Annotation_<description>_duration, where the description is encoded in the column name?

We also have to consider for example if there are two `BAD_BLINK` annotations within a given epoch, what do we do? I'm tempted to have a single column `Annotation_BAD_BLINK` where each entry is a `ndarray` of shape `(n_annot, 2)` of `(onset_within_epoch, duration_within_epoch)` entries.

> Is it possible to make metadata store say a list or tuple in the columns sometimes? If so, then we don't need to add an arbitrary # of columns.

I think Pandas allows you to do this sort of thing. To be tested...

adam2392:
So currently in Mne it says that metadata is just a constrained data frame that doesn't seem to allow lists in it. However, not sure if there's an internal check for it actually. 

I think if we could allow that data type for the new columns "Annotations_...", then we can achieve what you want.

adam2392:
Kay nvm, this works @larsoner 

The next step is to figure out how to save raw_sfreq and first_time in Epochs roundtrip IO. Those are the two necessary constants to go from Annotations -> metadata Epochs

adam2392:
@larsoner are these the unit tests you had in mind?

The key to making this one work was to just rely on `Raw.set_annotations`, and then making sure Epochs doesn't mess with the times / offsetting.

mne/epochs.py
```diff
@@ -3406,6 +3578,12 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
         if not isinstance(epochs, BaseEpochs):
             raise TypeError('epochs_list[%d] must be an instance of Epochs, '
                             'got %s' % (ei, type(epochs)))
+
+        if hasattr(epochs, 'annotations'):
+            warn('Concatenation of Annotations within Epochs '
+                 'is not supported yet. Annotations within these Epochs will '
+                 'be dropped.')
+            epochs.set_annotations(None)
```
adam2392, line 3586:
```suggestion
            epochs._set_annotations(None)
```

adam2392:
@agramfort @larsoner and @drammock  this is now ready for review. 

## Changes
Setting annotations in Epochs relies on the assumption that Annotations MUST be set on the Raw object. Then when constructing an Epochs data structure from Raw, Annotations are passed through via `_set_annotations` private function.

Then we can enable two new Epochs functions:
- `get_epoch_annotations`, which obtains a list of `n_epochs` long containing tuples representing (onset relative to epoch times[0], duration, description)
- `map_annots_to_metadata`, which augments the `metadata` data frame with a new `Annotations` column containing the list of tuples (onset, duration, description) for each row of epochs.

## Unit Tests
To walk you through the unit tests, there is the following cases:

- without meas_date, no first_samp, no orig_date
- without meas_date, first_samp, no orig_date
- with meas_date, no first_samp, no orig_date
- with meas_date, no first_samp, orig_date
- with meas_date, first_samp, no orig_date
- with meas_date, first_samp, orig_date

Note without `meas_date` and with `orig_date` results in error as it should when you do `raw.set_annotations`.

## TODO

1. Make these functions available for both `Epochs` and `EpochsFIF` class to enable IO.
2. Enable ch_names to be mapped to metadata for `map_annots_to_metadata()`

adam2392:
Notes from dev meeting:

- `map_annots_to_metadata` could be possibly named better. Maybe like... `add_annots_to_metadata`
- augment docstring for Epochs/EpochsFIF to specify that Annotations are only preserved from Raw, but can't be added to the Epochs data structure after the fact. Also augment EpochsArray to say it doesn't support Annotations.

adam2392:
> Notes from dev meeting:
> 
> * `map_annots_to_metadata` could be possibly named better. Maybe like... `add_annots_to_metadata`
> * augment docstring for Epochs/EpochsFIF to specify that Annotations are only preserved from Raw, but can't be added to the Epochs data structure after the fact. Also augment EpochsArray to say it doesn't support Annotations.

@drammock I've made the changes we spoke about. This is good to review now on my end.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+        # check if annotations exist
+        if self.raw_annotations is None:
+            return self
+
+        events = self.events
+
+        # create a list of annotations for each epoch
+        epoch_annot_list = [[] for _ in range(len(events))]
+
+        for annot in self.raw_annotations:
```
drammock, line 2766: I wonder if it's possible to avoid a for-loop here?  `self.raw_annotations.onset` is a numpy array of all onsets, and similarly for `.duration`.  Seems like from that and `self.events[:, 0]`, `tmin` and `tmax`, and some combination of `np.searchsorted()` / `<` / `>` / bitwise `&` and `|` you could get there?

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+            The modified instance (instance is also modified inplace).
+        """
+        pd = _check_pandas_installed()
+
+        # check if annotations exist
+        if self.annotations is None:
+            return self
+
+        # create a list of annotations for each epoch
+        epoch_annot_list = [[] for _ in range(len(self))]
```
drammock, line 2805: cruft?  it's redefined a few lines down and not used in between...

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+
+        # loop through each annotations and add to the metadata
```
drammock, line 2807: this comment doesn't describe what the next few lines actually do

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+            metadata = self._metadata
+        else:
+            data = np.empty((len(self), 0))
+            metadata = pd.DataFrame(data=data)
+
+        # get the Epoch annotations
+        epoch_annot_list = self.get_epoch_annotations()
+
+        # create a new Annotations column
+        metadata['Annotations'] = [[] for _ in range(metadata.shape[0])]
```
drammock, line 2818: why not directly assign `metadata['Annotations'] = epoch_annot_list` here?  or convert `epoch_annot_list` to a pandas Series and then add the whole thing as a column afterwards?

mne/epochs.py
```diff
@@ -3388,6 +3576,12 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
         if not isinstance(epochs, BaseEpochs):
             raise TypeError('epochs_list[%d] must be an instance of Epochs, '
                             'got %s' % (ei, type(epochs)))
+
+        if hasattr(epochs, 'annotations'):
+            warn('Concatenation of Annotations within Epochs '
+                 'is not supported yet. Annotations within these Epochs will '
+                 'be dropped.')
```
drammock, line 3583:
```suggestion
            warn('Concatenation of Annotations within Epochs is not supported '
                 'yet. Annotations within these Epochs will be dropped.')
```

mne/epochs.py
```diff
@@ -2479,6 +2480,50 @@ def _ensure_list(x):
...
+    -------
+    epoch_indices : np.ndarray
+        The epoch indices that the annotation overlaps in.
+    """
+    events = epochs.events
+    times = epochs.times
+    raw_sfreq = epochs._raw_sfreq
+
+    # convert to a 2D array of onset offset of each epoch
+    epoch_wins = np.zeros((len(events), 2))
```
drammock, line 2505:
```suggestion
    epoch_wins = np.zeros((len(events), 2), dtype=int)
```

mne/epochs.py
```diff
@@ -2479,6 +2480,50 @@ def _ensure_list(x):
...
+    for idx in range(len(events)):
+        # get the beginning and end sample of the epoch window
+        onset_samp = events[idx, 0] - times[0] * raw_sfreq
+        offset_samp = events[idx, 0] + times[-1] * raw_sfreq
```
drammock, line 2509: `np.rint(...).astype(int)` here too

drammock (changes requested):
OK I've done a first pass here @adam2392.  I didn't look closely at the tests at all, just the module changes.  Feel free to ping me on discord if I've made any comments here you don't understand or don't agree with, if you think a quick chat or pair-programming session will help sort it out faster.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+        %(on_missing_ch_names)s
+
+        Returns
+        -------
+        self : instance of Raw
+            The raw object with annotations.
+        """
+        meas_date = _handle_meas_date(self.info['meas_date'])
+        if annotations is None:
+            self._annotations = Annotations([], [], [], meas_date)
```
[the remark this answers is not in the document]
adam2392: Sure, we can make it be None.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+            if meas_date is None and annotations.orig_time is not None:
+                raise RuntimeError('Ambiguous operation. Setting an Annotation'
+                                   ' object with known ``orig_time`` to a raw'
+                                   ' object which has ``meas_date`` set to'
+                                   ' None is ambiguous. Please, either set a'
+                                   ' meaningful ``meas_date`` to the raw'
+                                   ' object; or set ``orig_time`` to None in'
+                                   ' which case the annotation onsets would be'
+                                   ' taken in reference to the first sample of'
+                                   ' the raw object.')
```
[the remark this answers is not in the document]
adam2392: Sure, we should remove this.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+            new_annotations = annotations.copy()
+            new_annotations._prune_ch_names(self.info, on_missing)
```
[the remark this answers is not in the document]
adam2392: Yes, we need these in case of the picks.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+        -------
+        epoch_annots : list
+            A list (with length equal to number of epochs) of annotations per
+            epoch. Each annotation is stored as (onset, duration, description),
+            where the onset is now relative to the epoch, rather then in
+            continuous time.
+        """
+        # check if annotations exist
+        if self.raw_annotations is None:
+            return self
```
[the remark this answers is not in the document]
adam2392: Done.

mne/epochs.py
adam2392: I don't think it works syntatcitcally, but I'll try a few different iterations.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
                                             self.reject_by_annotation)
         return data
 
+    @property
+    def raw_annotations(self):  # noqa: D102
```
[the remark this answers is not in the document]
adam2392: It was named this because `concatenate_epochs` checks for an attribute called `annotations`, so then a bunch of warnings were raised.

Instead, I added a line to `conftest.py` to ignore this warning for now.

mne/epochs.py
```diff
@@ -2640,6 +2701,121 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+        epoch_annots : list
+            A list (with length equal to number of epochs) of annotations per
+            epoch. Each annotation is stored as (onset, duration, description),
+            where the onset is now relative to the epoch, rather then in
+            continuous time.
+        """
+        events = self.events
+
+        # create a list of annotations for each epoch
+        epoch_annot_list = [[]] * len(events)
```
adam2392, line 2754: This actually doesn't work because it creates a pointer in RAM to the same list, rather then a separate list per epoch.

`epoch_annot_list = [[] for _ in range(len(events))]` is required.

drammock: Oh interesting! Could do `[list()] * len(events)` then?

mne/epochs.py
adam2392: Done.

mne/epochs.py
adam2392: Nope unfortunately that actually also doesn't work. Some weird Python magic. I think, if you do anything like:

```
[...] * len(N)
```

it allocates only 1 list and then points to that. I think maybe if you set it to be `None` inside, it bypasses this issue, but then Imo we might as well be explicit w/ the soln I found on SO.

larsoner: Indeed `[[]] * len(events)` will make a list of length N with identical (in memory) lists:
```
>>> x = [[]] * 3
>>> x[0].append(1)
>>> x
[[1], [1], [1]]
```

mne/epochs.py
adam2392: Okay pd.Series works.

mne/epochs.py
```diff
@@ -2640,6 +2696,134 @@ def _get_epoch_from_raw(self, idx, verbose=None):
...
+        epoch_annot_list = [[] for _ in range(len(events))]
+
+        for annot in self.raw_annotations:
+            onset_ = annot['onset']
+            duration_ = annot['duration']
+            description_ = annot['description']
+
+            # convert onset to samples and account for first time
+            onset_samp = int(onset_ * self._raw_sfreq + self._first_time)
+            duration_samp = int(duration_ * self._raw_sfreq)
```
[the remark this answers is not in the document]
adam2392: Done. and used first_samp instead.

mne/epochs.py
adam2392: Hmm not sure. I suppose this might be a performance issue if annotations is huge?

Not sure what you specifically have in mind here though. Would this remove just one for loop, or both for loops?

adam2392:
> OK I've done a first pass here @adam2392. I didn't look closely at the tests at all, just the module changes. Feel free to ping me on discord if I've made any comments here you don't understand or don't agree with, if you think a quick chat or pair-programming session will help sort it out faster.

I've addressed all your concerns raised here. The only one I am not sure on is your suggestion to avoid a for loop over annotations. 

https://github.com/mne-tools/mne-python/pull/9969#discussion_r763514409

adam2392:
Testing this downstream with `mne-connectivity` and found an issue in https://github.com/mne-tools/mne-connectivity/pull/58#issuecomment-989284960

where the `Annotations` column is too complex of a data structure. If you convert `metadata.to_numpy()` you end up getting an object, rather then basic numpy/python data types. I propose thus, `Annotations` column in metadata should probably be split into 3 columns of lists: `Annnotations_onset`, `Annotations_duration` and `Annotations_description`.

adam2392:
Is this mergeable now, or would it require some more discussion and work?

agramfort:
give me time to have a look. Is it good to go from your end and others?

>

mne/epochs.py
```diff
@@ -2479,8 +2480,183 @@ def _ensure_list(x):
...
+        events = self.events
+
+        # create a list of annotations for each epoch
+        epoch_annot_list = [[] for _ in range(len(events))]
+
+        # check if annotations exist
+        if self.annotations is None:
+            return epoch_annot_list
+
+        for annot in self.annotations:
```
drammock, line 2586: I still think you can eliminate the for-loop.

```py
epoch_tzeros = self.events[:, 0] / self._raw_sfreq  # no need to account for first_samp here
epoch_starts, epoch_stops = np.atleast_2d(epoch_tzeros) + np.atleast_2d(self.times[[0, -1]]).T
annot_starts = self._annotations.onset  # ... because first_samp isn't accounted for here either
annot_stops = annot_starts + self._annotations.duration

annot_spans_epoch_start = np.logical_and(
    np.atleast_2d(epoch_starts) > np.atleast_2d(annot_starts).T,
    np.atleast_2d(epoch_starts) < np.atleast_2d(annot_stops).T)

annot_spans_epoch_end = np.logical_and(
    np.atleast_2d(epoch_stops) > np.atleast_2d(annot_starts).T,
    np.atleast_2d(epoch_stops) < np.atleast_2d(annot_stops).T)

annot_fully_within_epoch = np.logical_and(
    np.atleast_2d(epoch_starts) < np.atleast_2d(annot_starts).T,
    np.atleast_2d(epoch_stops) > np.atleast_2d(annot_stops).T)

all_cases = (annot_spans_epoch_start + annot_spans_epoch_end +
             annot_fully_within_epoch)

for annot_ix, epo_ix in zip(np.where(all_cases)):
    this_annot = self._annotations[annot_ix]
    this_tzero = epoch_tzeros[epo_ix]
    annot = (this_annot['onset'] - this_tzero,
             this_annot['duration'],
             this_annot['description'])
    epoch_annot_list[epo_ix].append(annot)
```

To me this seems more straightforward (e.g. no need to deal with first_samp) and the variable names make clear what the numpy functions are doing.  But maybe I'm oversimplifying  / missing something?

drammock (changes requested):
I've paused my review midway because I'm pretty sure it's actually not working correctly.  If I do this:

```py
import os
import mne

                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False, preload=False)
annotations = mne.Annotations(onset=(2, 3, 5, 11), duration=(0.5, 2.5, 1, 8),
                              description=('foo', 'bar', 'bad_baz', 'qux'))
raw.set_annotations(annotations)

event_dict = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3}
epochs = mne.Epochs(raw, events, event_id=event_dict, preload=False,
                    proj=False)
```

if I look at `raw.plot(events=epochs.events)` it is clear that the first and second epochs both fall inside the "bar" annotation.  However, `epochs.get_epoch_annotations()` returns a list where *each annotation shows up only once* and where the first and second epochs show no annotations at all.  I conclude that something about your logic is off, and I encourage you to look again at my suggestion to eliminate the for-loop (it's more-fleshed-out this time).

mne/epochs.py
```diff
@@ -2479,8 +2480,183 @@ def _ensure_list(x):
...
+            self._annotations = None
+        else:
+            _validate_type(annotations, Annotations, 'annotations')
+            new_annotations = annotations.copy()
+            new_annotations._prune_ch_names(self.info, on_missing)
+            self._annotations = new_annotations
+
+        return self
+
+    def get_epoch_annotations(self):
```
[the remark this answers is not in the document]
adam2392: Perhaps `get_annotlist_per_epoch`, or `get_annotations_per_epoch`?

mne/epochs.py
adam2392: Ah I see yeah. I tested this out and this is way more concise. I added this in with a few minor syntax fixes and this actually fixed the issue you raised in your comment.

adam2392:
@drammock the more complex case with some non-fixed length Epochs that you raised using sample_data works now. Do you prefer that I come up some more complex unit tests? The ones right now are just sort of hard-coded examples that I worked out on paper.

Here is a notebook showing the Epoch Annotations correctly being obtained.

[plotting_epochs_and_testing.pdf](https://github.com/mne-tools/mne-python/files/7707628/plotting_epochs_and_testing.pdf)

drammock:
> Do you prefer that I come up some more complex unit tests?

IMO a complete test would cover the following:

- annotation start is {before, at, after} epoch start, crossed with: annotation end is {before, at, after} epoch end. (9 cases)
- 1 annotation that is fully outside all epochs (make sure it is dropped)
- 1 annotation that spans multiple epochs (make sure it shows up in both)

Personally I would create a fake raw with 1 channel, maybe 15 seconds long or so, and create equally spaced epochs (with some of separation between them) so that it's possible to cover all 11 cases without annotations crashing into adjacent epochs unwantedly.

mne/epochs.py
```diff
@@ -2596,29 +2551,43 @@ def get_epoch_annotations(self):
...
+        # when each epoch and annotation starts/stops
+        # no need to account for first_samp here...
+        epoch_tzeros = events[:, 0] / self._raw_sfreq
+        epoch_starts, epoch_stops = np.atleast_2d(
+            epoch_tzeros) + np.atleast_2d(self.times[[0, -1]]).T
+        # ... because first_samp isn't accounted for here either
+        annot_starts = self._annotations.onset
+        annot_stops = annot_starts + self._annotations.duration
+
+        # get all annotations that have a start point within epoch
```
drammock, line 2563: this comment is inaccurate.

mne/epochs.py
```diff
@@ -2596,29 +2551,43 @@ def get_epoch_annotations(self):
...
+        annot_straddles_epoch_start = np.logical_and(
+            np.atleast_2d(epoch_starts) >= np.atleast_2d(annot_starts).T,
+            np.atleast_2d(epoch_starts) < np.atleast_2d(annot_stops).T)
+
+        # get all annotations that have an endpoint within epoch
```
drammock, line 2568: this comment is inaccurate

mne/epochs.py
```diff
@@ -2596,29 +2551,43 @@ def get_epoch_annotations(self):
...
+        all_cases = (annot_straddles_epoch_start +
+                     annot_straddles_epoch_end +
+                     annot_fully_within_epoch)
+
+        for annot_ix, epo_ix in zip(*np.nonzero(all_cases)):
+            this_annot = self._annotations[annot_ix]
+            this_tzero = epoch_tzeros[epo_ix]
+
+            # get all annotations for this Epoch and offset the onset
+            # relative to the start of the Epoch
```
drammock, line 2586: this comment is inaccurate (specifically the "get all annotations" part; it's within a loop so it's only "getting" one annotation at a time.

mne/epochs.py
adam2392: Fixed.

adam2392: Lmk if this looks okay?

adam2392:
> IMO a complete test would cover the following:
> 
> * annotation start is {before, at, after} epoch start, crossed with: annotation end is {before, at, after} epoch end. (9 cases)
> * 1 annotation that is fully outside all epochs (make sure it is dropped)
> * 1 annotation that spans multiple epochs (make sure it shows up in both)
> 
> Personally I would create a fake raw with 1 channel, maybe 15 seconds long or so, and create equally spaced epochs (with some of separation between them) so that it's possible to cover all 11 cases without annotations crashing into adjacent epochs unwantedly.

Okay I updated the unit tests to be a bit more comprehensive. The first one does a hard coded test for fixed length epochs, where we test the behavior wrt `first_samp`, `meas_date` and `orig_time`.

The second test does what you suggest, where we test on non-contiguous epochs and make sure Annotations of different lengths and start points all show up in the right place.

mne/epochs.py
```diff
@@ -2492,8 +2493,166 @@ def _ensure_list(x):
...
+        else:
+            _validate_type(annotations, Annotations, 'annotations')
+            new_annotations = annotations.copy()
+            new_annotations._prune_ch_names(self.info, on_missing)
+            self._annotations = new_annotations
+
+        return self
+
+    def get_annotations_per_epoch(self):
+        """Get a list of which annotations occur during each epoch.
```
adam2392, line 2534:
```suggestion
        """Get a list of annotations that occur during each epoch.
```

Perhaps?

mne/epochs.py
```diff
@@ -2492,8 +2493,166 @@ def _ensure_list(x):
...
+        -------
+        epoch_annots : list
+            A list of lists (with length equal to number of epochs) where each
+            inner list contains the annotations that overlap the corresponding
+            epoch. Annotations are stored as a :class:`tuple` (onset, duration,
+            description), where the onset is now relative to time=0 of the
+            epoch, rather than time=0 of the original continuous (raw) data.
+        """
+        events = self.events
+
```
adam2392, line 2546:
```suggestion
```

mne/epochs.py
```diff
@@ -2492,8 +2493,166 @@ def _ensure_list(x):
...
+        # create a list of annotations for each epoch
+        epoch_annot_list = [[] for _ in range(len(events))]
+
+        # check if annotations exist
+        if self.annotations is None:
+            return epoch_annot_list
+
+        # when each epoch and annotation starts/stops
+        # no need to account for first_samp here...
+        epoch_tzeros = events[:, 0] / self._raw_sfreq
```
adam2392, line 2556:
```suggestion
        epoch_tzeros = self.events[:, 0] / self._raw_sfreq
```

mne/epochs.py
```diff
@@ -2492,8 +2493,166 @@ def _ensure_list(x):
...
+        onset, duration, description = [], [], []
+        for epoch_annot in epoch_annot_list:
+            if epoch_annot != []:
+                onset.append([x[0] for x in epoch_annot])
+                duration.append([x[1] for x in epoch_annot])
+                description.append([x[2] for x in epoch_annot])
+            else:
+                onset.append([])
+                duration.append([])
+                description.append([])
```
drammock, line 2641:
```suggestion
        onset, duration, description = [], [], []
        for epoch in epoch_annot_list:
            for ix, annot_prop in enumerate((onset, duration, description)):
                entry = [annot[ix] for annot in epoch] if len(epoch) else []
                annot_prop.append(entry)
```

mne/epochs.py
```diff
@@ -2600,6 +2768,12 @@ def __init__(self, raw, events, event_id=None, tmin=-0.2, tmax=0.5,
         proj = proj or raw.proj
 
         self.reject_by_annotation = reject_by_annotation
+
+        # store certain raw properties for dealing with annotations
+        self._first_samp = raw.first_samp.copy()
+        self._raw_sfreq = np.rint(raw.info['sfreq'])
+        self._first_time = raw._first_time.copy()
+
```
drammock, line 2776: two of these are cruft now; also the `np.rint` was on the wrong one (should go only on samples, non-integer sfreqs are valid)

```suggestion
        # keep track of original sfreq (needed for annotations)
        self._raw_sfreq = raw.info['sfreq']

```

drammock (changes requested):
OK, next task is to either make it work with non-preloaded Epochs, or raise a better / more informative error if the epochs are not preloaded.  If we decide to make it work with non-preloaded epochs, we should probably add a `.. warning::` to the docstring of `get_annots_per_epoch` (still not in love with the name...) that people should be careful / pay attention to whether epochs are preloaded / bads have been dropped yet, otherwise the indexing will get out of sync.

I'll go ahead and add all the simple changes (mostly code comments) to a batch and commit them now, and you can work from that.  Here's a MWE for the non-preloaded case:

```py
import os
import mne
                                    'sample_audvis_raw.fif')
raw = mne.io.read_raw_fif(sample_data_raw_file, verbose=False, preload=False)
raw.crop(tmax=30)
annotations = mne.Annotations(onset=(2, 3, 5, 11), duration=(0.5, 2.5, 1, 8),
                              description=('foo', 'bar', 'bad_baz', 'qux'))
raw.set_annotations(annotations)
event_dict = {'auditory/left': 1, 'auditory/right': 2, 'visual/left': 3}
epochs = mne.Epochs(raw, events, event_id=event_dict, preload=False,
                    proj=False)
epochs.add_annotations_to_metadata()

RuntimeError                              Traceback (most recent call last)
<ipython-input-7-262e45b97f16> in <module>
----> 1 epochs.add_annotations_to_metadata()

/opt/mne/python/mne/epochs.py in add_annotations_to_metadata(self)
   2623             metadata = self._metadata
   2624         else:
-> 2625             data = np.empty((len(self), 0))
   2626             metadata = pd.DataFrame(data=data)
   2627 

/opt/mne/python/mne/utils/mixin.py in __len__(self)
    289         from ..epochs import BaseEpochs
    290         if isinstance(self, BaseEpochs) and not self._bad_dropped:
--> 291             raise RuntimeError('Since bad epochs have not been dropped, the '
    292                                'length of the Epochs is not known. Load the '
    293                                'Epochs with preload=True, or call '

RuntimeError: Since bad epochs have not been dropped, the length of the Epochs is not known. Load the Epochs with preload=True, or call Epochs.drop_bad(). To find the number of events in the Epochs, use len(Epochs.events).
```

mne/epochs.py
```diff
@@ -2492,8 +2493,166 @@ def _ensure_list(x):
...
+        Returns
+        -------
+        self : instance of Epochs
+            The modified instance (instance is also modified inplace).
+        """
+        pd = _check_pandas_installed()
+
+        # check if annotations exist
+        if self.annotations is None:
+            return self
```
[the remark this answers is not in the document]
adam2392: Agreed. Added a warning message

mne/epochs.py
adam2392: We need to keep at least _first_samp in case the Raw was shifted. This is only for the case that `orig_time` and `_first_samp` is set in Annotations and Raw respectively.

adam2392:
> OK, next task is to either make it work with non-preloaded Epochs, or raise a better / more informative error if the epochs are not preloaded. If we decide to make it work with non-preloaded epochs, we should probably add a `.. warning::` to the docstring of `get_annots_per_epoch` (still not in love with the name...) that people should be careful / pay attention to whether epochs are preloaded / bads have been dropped yet, otherwise the indexing will get out of sync.

Epochs.add_annotations_to_metadata now works with non-preloaded Epochs. 

For `get_annotations_per_epoch`, one can call after preloading and it'll just return you a new list because now some of the Epochs are dropped. The underlying function uses `epochs.events` to query over the starting sample of each Epoch to then search for Annotations that lie within the Epoch.

 For `add_annotations_to_metadata`, when Epochs are loaded, and bad epochs are dropped, the metadata would be updated by dropping those corresponding rows.

Thus both functions match what you would expect before/after loading the actual data and dropping bads.

drammock (approved):
OK, this is looking pretty good!  I didn't have time to thoroughly review the tests, but they look to be fairly comprehensive...  my hunch is they could be simplified somewhat and/or some different design choices could make them easier to follow (e.g., why in the first test are the equally spaced epochs overlapping?  seems like epochs of 1s with zero overlap would make the various expectations / assertions easier to grasp).  Maybe @agramfort or @larsoner can suggest some ways to streamline the tests; if not I could try to look again on Friday or Monday.

adam2392:
> OK, this is looking pretty good! I didn't have time to thoroughly review the tests, but they look to be fairly comprehensive... my hunch is they could be simplified somewhat and/or some different design choices could make them easier to follow (e.g., why in the first test are the equally spaced epochs overlapping? seems like epochs of 1s with zero overlap would make the various expectations / assertions easier to grasp). 

I kept them as overlapping to add some complexity, but if you think that's unnecessary lmk. The test is somewhat hard to follow, probably because I mapped it out on paper :p. I can change if you all think necessary.

Otherwise I'll wait for Eric and Alexandre to review.

larsoner:
I should be able to look today or tomorrow @adam2392 !

mne/epochs.py
```diff
@@ -2492,8 +2493,158 @@ def _ensure_list(x):
...
+
+        Returns
+        -------
+        self : instance of Raw
+            The raw object with annotations.
+        """
+        if annotations is None:
+            self._annotations = None
+        else:
+            _validate_type(annotations, Annotations, 'annotations')
```
larsoner, line 2526: This will give a misleading message saying it has to be Annotations when it can also be None. Better to move it above the conditional with `(Annotations, None)` as the types

mne/epochs.py
```diff
@@ -2582,6 +2733,15 @@ class Epochs(BaseEpochs):
...
     and events to be: ``{'aud/vis': 3}`` and ``[[0, 0, 3]]`` respectively.
+
+    There is limited support for :class:`~mne.Annotations` in the
+    :class:`~mne.Epochs` class. Currently annotations that are present in the
+    :class:`~mne.io.Raw` object will be preserved in the resulting
+    :class:`~mne.Epochs` object, but it is not yet possible to add annotations
+    to the Epochs object programmatically (via code) or interactively (through
+    the plot window). Additionally, concatenating :class:`~mne.Epochs` objects
+    that contain annotations is not supported, and any annotations will be
+    dropped when concatenating.
```
larsoner, line 2744: An enumerated list should help readability, and I think the third point (which to me is the most important next step) was missing
```suggestion
    There is limited support for :class:`~mne.Annotations` in the
    :class:`~mne.Epochs` class. Currently annotations that are present in the
    :class:`~mne.io.Raw` object will be preserved in the resulting
    :class:`~mne.Epochs` object, but:
    
    1. It is not yet possible to add annotations
       to the Epochs object programmatically (via code) or interactively
       (through the plot window)
    2. Concatenating :class:`~mne.Epochs` objects
       that contain annotations is not supported, and any annotations will
       be dropped when concatenating.
    3. Annotations will be lost on save.
```

mne/epochs.py
```diff
@@ -3401,6 +3571,11 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
         if not isinstance(epochs, BaseEpochs):
             raise TypeError('epochs_list[%d] must be an instance of Epochs, '
                             'got %s' % (ei, type(epochs)))
+
+        if hasattr(epochs, 'annotations'):
+            warn('Concatenation of Annotations within Epochs is not supported '
+                 'yet. Annotations within these Epochs will be dropped.')
```
larsoner, line 3577:
```suggestion
                 'yet. All annotations will be dropped.')
```

mne/epochs.py
```diff
@@ -3401,6 +3571,11 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
         if not isinstance(epochs, BaseEpochs):
             raise TypeError('epochs_list[%d] must be an instance of Epochs, '
                             'got %s' % (ei, type(epochs)))
+
+        if hasattr(epochs, 'annotations'):
```
larsoner, line 3575: It only matters if they are non-None, I'm surprised this doesn't break some existing test where we create two `Epochs` instances from raw and concatenate them. Maybe we only currently test concat with EpochsArray or something? 
```suggestion
        if getattr(epochs, 'annotations', None) is not None and not warned:
            warned = True
```
And you'll need to add a `warned = False` before the loop. Basically when concatenating 10 Epochs instances with metadata you should get 1 warning, not 9.

mne/tests/test_epochs.py
```diff
@@ -3680,3 +3682,192 @@ def test_add_channels_picks():
...
+@pytest.mark.parametrize(
+    'first_samp, meas_date, orig_date', [
+        [0, None, None],
+        [10, None, None],
+        [0, np.pi, None],
+        [10, np.pi, None],
+        [0, np.pi, timedelta(seconds=1)],
+        [10, np.pi, timedelta(seconds=1)]
+    ]
+)
```
larsoner, line 3697:
```suggestion
@pytest.mark.parametrize('first_samp', [0, 10])
@pytest.mark.parametrize(
    'meas_date, orig_date', [
        [None, None],
        [np.pi, None],
        [np.pi, timedelta(seconds=1)]
    ]
)
```

mne/tests/test_epochs.py
```diff
@@ -3680,3 +3682,192 @@ def test_add_channels_picks():
...
+    Here, we test the following cases crossed:
+    - annotation start is before/after epoch start
+    - annotation end is before/after epoch end
+    - 1 annotation that is fully outside all epochs (make sure it is dropped)
+    - 1 annotation that spans multiple epochs (make sure it shows up in both)
+
+    In addition, tests functionality when Epochs are loaded vs not.
+    """
+    # do a more complicated case
+    data = np.random.randn(1, 600) * 10e-12
```
larsoner, line 3799: It's better practice to use a seeded RNG rather than whatever the current state is when pytest happens to call this (it makes the test more repeatable, even if the actual data values themselves do not matter).
```suggestion
    data = np.random.RandomState(0).randn(1, 600) * 10e-12
```

larsoner (commented):
Looks like a good start! I think we can further simplify the tests in follow-up PRs, too.

mne/epochs.py
adam2392: This is the way it was done in Raw too. Should I change it there as well?

mne/epochs.py
adam2392: Fair point :p. Will eliminate the 3rd point on the short follow-up PR.

mne/epochs.py
adam2392: Well in all the concatenation tests, they might not have set annotations inside the Raw when making them into `Epochs`, so probably wasn't caught.

drammock: Could it be that you're ignoring those warnings? I thought you added an ignore line to conftest? (Sorry, on mobile, hard to check for myself at the moment)

adam2392: Ah yeah true. So there's 23 places that `concatenate_epochs` is called unfortunately in `test_epochs.py`. I could add a catch warning statement in those files?

I suppose it depends on our timeframe or feasibility of adding concatenating Epochs with Annotations. I don't think it is possible because different Epochs might have been taken from completely different files right?

If so, I can add these catch warnings in and remove the ignore statement in `conftest.py`

larsoner: If you make my suggested change, I doubt many tests (if any) will break. They will only break off we have annotations on raw and make epochs and concat (right?), which I don't think we do

mne/conftest.py
```diff
@@ -127,6 +127,7 @@ def pytest_configure(config):
     ignore:.*np\.MachAr.*:DeprecationWarning
     ignore:.*Passing unrecognized arguments to super.*:DeprecationWarning
     always::ResourceWarning
+    ignore:.*Concatenation of Annotations within Epochs.*:RuntimeWarning
```
larsoner, line 130:
```suggestion
```
+1 for removing this, assuming it doesn't break a ton of tests

mne/epochs.py
adam2392: So the tests all do break actually because there are potentially Annotations within the Raw objects. Moreover, when you do `raw.set_annotations(None)`, the following lines of code ensure that Annotations are never None again:

```
if annotations is None:
            self._annotations = Annotations([], [], [], meas_date)
```

Therefore, we were on the right track with the suggested change, but we have to further check if the `epochs.annotations` is just an empty data structure or not.

Reference:
- https://github.com/mne-tools/mne-python/blob/3d4a4f66232448de542ac076df7b57115644978a/mne/io/base.py#L675-L676

mne/conftest.py
adam2392: Removed.

doc/changes/latest.inc
```diff
@@ -62,6 +62,8 @@ Enhancements
 
 - In :class:`mne.Report`, limit the width of automatically generated figures to a maximum of 850 pixels (450 pixels for :class:`mne.SourceEstimate` plots), and the resolution to 100 DPI to reduce file size, memory consumption, and – in some cases like :meth:`mne.Report.add_stc` – processing time (:gh:`10126`, :gh:`10129`, :gh:`10135` by `Richard Höchenberger`_)
 
+  - Annotations from a :class:`~mne.io.Raw` object are now preserved by the :class:`~mne.Epochs` constructor (:gh:`9969` by `Adam Li`_)
```
larsoner, line 65:
```suggestion
 - Annotations from a :class:`~mne.io.Raw` object are now preserved by the :class:`~mne.Epochs` constructor (:gh:`9969` by `Adam Li`_)
```

Errant space made this indented

https://39914-1301584-gh.circle-artifacts.com/0/dev/whats_new.html

mne/epochs.py
```diff
@@ -3398,10 +3572,21 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
...
+    # to make warning messages only occur once during concatenation
+    warned = False
+
     for ei, epochs in enumerate(epochs_list):
         if not isinstance(epochs, BaseEpochs):
             raise TypeError('epochs_list[%d] must be an instance of Epochs, '
                             'got %s' % (ei, type(epochs)))
+
+        if (getattr(epochs, 'annotations', None) is not None and
+                len(getattr(epochs, 'annotations', None)) != 0) and not warned:
```
larsoner, line 3585: `and` is a short-circuiting operation, so by this point in the conditional we know the first part has ben satisfied
```suggestion
        if (getattr(epochs, 'annotations', None) is not None and
                len(epochs.annotations) > 0 and
                not warned):
```

larsoner (approved):
Will push these two tiny changes then merge when green, thanks @adam2392 !

larsoner:
Thanks @adam2392 ! Do you plan to tackle one of the 1-3 issues next? It seems like I/O would be the next best step

PR #9970, opened 2021-11-05
MRG: Add extensive checking for Report tags

### LINKED ISSUE #9965: BUG: Tags with spaces are broken
This:
```
import mne
report = mne.Report()
figs, captions = list(), list()
for n in range(2):
    fig, ax = plt.subplots()
    ax.text(0, 0, str(n))
    figs.append(fig)
    captions.append(str(n))
report.add_figure(figs, 'numbers', caption=captions, tags=['My numbers'])
report.save('rep.html', overwrite=True)
```
is broken -- the top bar covers content, and the sliders do not work:

Screenshot from 2021-11-05 10-43-33

If I change it to `tags['My_numbers']` it works fine:

Screenshot from 2021-11-05 10-43-49

@hoechenberger can you look?

hoechenberger:
Fixes #9965

cc @larsoner

### DIFF
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -280,7 +280,7 @@ class _ContentElement:
 
 
 def _check_tags(tags) -> Tuple[str]:
-    # Must be iterable, not not a string
+    # Must be iterable, but not a string
     if (isinstance(tags, str) or not isinstance(tags, (Sequence, np.ndarray))):
         raise TypeError(
             f'tags must be a collection of str, but got {type(tags)} '
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -308,7 +308,7 @@ def _check_tags(tags) -> Tuple[str]:
     if bad_tags:
         raise ValueError(
             f'The following tags contained invalid characters: '
-            f'{", ".join(bad_tags)}'
+            f'{", ".join(repr(tag) for tag in bad_tags)}'
         )
 
     return tags
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -10,6 +10,7 @@ import io
 import dataclasses
 from dataclasses import dataclass
 from typing import Tuple
+from collections.abc import Sequence
 import base64
 from io import BytesIO, StringIO
 import contextlib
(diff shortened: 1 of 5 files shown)


mne/report/report.py
```diff
@@ -278,6 +279,41 @@ class _ContentElement:
     html: str
 
 
+def _check_tags(tags) -> Tuple[str]:
+    # Must be iterable, not not a string
```
hoechenberger, line 283:
```suggestion
    # Must be iterable, but not a string
```

agramfort (approved):
LGTM ! 🙏

mne/report/report.py
```diff
@@ -278,6 +279,41 @@ class _ContentElement:
     html: str
 
 
+def _check_tags(tags) -> Tuple[str]:
+    # Must be iterable, but not a string
+    if (isinstance(tags, str) or not isinstance(tags, (Sequence, np.ndarray))):
+        raise TypeError(
+            f'tags must be a collection of str, but got {type(tags)} '
+            f'instead: {tags}'
+        )
```
larsoner, line 288: This should be `_check_option('tags', tags, (str, 'array-like', Sequence))` or so (and I'm not sure you need the `Sequence` because maybe it's already handled by 'array-like'?

mne/report/report.py
```diff
@@ -278,6 +279,41 @@ class _ContentElement:
...
+    tags = tuple(tags)
+
+    # Check for invalid dtypes
+    bad_tags = [tag for tag in tags
+                if not isinstance(tag, str)]
+    if bad_tags:
+        raise TypeError(
+            f'tags must be strings, but got the following instead: '
+            f'{", ".join([str(tag) for tag in bad_tags])}'
+        )
```
larsoner, line 298:
```suggestion
    for ti, tag in tags:
        _check_option(f'tags[{ti}]', tag, str)
```

mne/report/report.py
```diff
@@ -278,6 +279,41 @@ class _ContentElement:
...
+    bad_tags = []
+    for tag in tags:
+        for invalid_char in invalid_chars:
+            if invalid_char in tag:
+                bad_tags.append(tag)
+                break
+    if bad_tags:
+        raise ValueError(
+            f'The following tags contained invalid characters: '
+            f'{", ".join(bad_tags)}'
```
larsoner, line 311: This will add quotes (and probably escape `"`) which will make it clearer I think
```suggestion
            f'{", ".join(repr(tag) for tag in bad_tags)}'
```

larsoner (commented):
Just some minor cleanup / simplification / more targeted error ideas

mne/report/report.py
hoechenberger: It's more complicated than that, I need to check raise

- if it **is** a string (here `_check_option()` could be used), or
- if it **is not** `array-like` (which I cannot test with `_check_option()` as it will raise if the condition is not met)

Looking through `check.py` now to see if there are some other functions I could use …

mne/report/report.py
hoechenberger: Here again I have the issue that `_check_option()` immediately raises if the condition is not met. But I want to collect all invalid tags and tell the user, so they don't have to start fixing them one after the other after each new run which shows a problem with yet another tag

hoechenberger:
@larsoner Please merge if you're happy :)

agramfort (approved):

larsoner:
Thanks @hoechenberger !

PR #9972, opened 2021-11-06
MRG: Add "interaction" param to CoregistrationUI

hoechenberger:
This adds an `interaction` kwarg to `CoregistrationUI`. `mne.gui.coregistration()` in `main` simply ignores its `interaction` parameter and doesn't pass it to `CoregistrationUI`. This PR fixes this behavior.

Following a recent conversation with @larsoner, this also changes the default interaction style to `'terrain'`.

In the `'terrain'` interaction mode, zooming by scrolling up/down with the mouse doesn't work for me on macOS, so this needs to be fixed.

```python
# %%
import mne

mne.gui.coregistration(interaction='trackball')  # behaves like `main`
```
https://user-images.githubusercontent.com/2046265/140620892-741b4d4a-72f8-4102-a7e2-9e8f7fcc5bc2.mov

```python
# %%
import mne

mne.gui.coregistration(interaction='terrain')  # new default
```
https://user-images.githubusercontent.com/2046265/140620903-d60ec482-7c43-4ef5-bf31-7f423817b4c8.mov

cc @agramfort

PR #9973, opened 2021-11-08
import fix for scipy 1.8 pre

drammock:
should hopefully fix the py3.9 CI error that popped up in #9972

### DIFF
--- a/mne/externals/pymatreader/utils.py
+++ b/mne/externals/pymatreader/utils.py
@@ -51,6 +51,15 @@ def _import_h5py():
     return h5py
 
 
+def get_matfile_version():
+    """Triage import path based on SciPy version."""
+    try:
+        from scipy.io.matlab.miobase import get_matfile_version
+    except ModuleNotFoundError:  # scipy 1.8
+        from scipy.io._matlab.miobase import get_matfile_version
+    return get_matfile_version
+
+
 def _hdf5todict(hdf5_object, variable_names=None, ignore_fields=None):
     """
     Recursively converts a hdf5 object to a python dictionary,
--- a/mne/externals/pymatreader/utils.py
+++ b/mne/externals/pymatreader/utils.py
@@ -33,6 +33,15 @@ from warnings import warn
 import numpy
 import scipy.io
 
+try:
+    from scipy.io.matlab.miobase import get_matfile_version
+    from scipy.io.matlab.mio5_params import MatlabOpaque
+    from scipy.io.matlab.mio5 import MatlabFunction
+except ModuleNotFoundError:  # scipy 1.8
+    from scipy.io._matlab.miobase import get_matfile_version
+    from scipy.io._matlab.mio5_params import MatlabOpaque
+    from scipy.io._matlab.mio5 import MatlabFunction
+
 if sys.version_info <= (2, 7):
     chr = unichr  # noqa This is needed for python 2 and 3 compatibility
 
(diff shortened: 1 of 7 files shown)


larsoner:
Can you port upstream as well

https://gitlab.com/obob/pymatreader

Happy to merge this if CIs come back happy and we can have a separate PR to incorporate any changes @thht asks for upstream

mne/externals/pymatreader/utils.py
```diff
@@ -51,6 +51,15 @@ def _import_h5py():
     return h5py
 
 
+def get_matfile_version():
+    """Triage import path based on SciPy version."""
+    try:
+        from scipy.io.matlab.miobase import get_matfile_version
+    except ModuleNotFoundError:  # scipy 1.8
+        from scipy.io._matlab.miobase import get_matfile_version
+    return get_matfile_version
```
larsoner, line 60: This might actually be a good function to have in the `scipy.io.matlab` namespace, any interest in making a PR to SciPy to add it @drammock ?

thht:
thanks for the notification.
i am happy to just add this in upstream pymatreader. i am working on it....

thht:
please note that this version does not work. the `get_matfile_version` function introduced here returns the actual function, not its result.

i am going to create a PR for a corrected version either today or tomorrow morning.

drammock:
> please note that this version does not work

yes, I just noticed the oversight and pushed a fix

thht:
i would rather fix this by using an `_import_get_matfile_version()` function.

https://gitlab.com/obob/pymatreader/-/merge_requests/29

in order to keep this version in sync with upstream, can we wait until tomorrow, when i can release and submit the PR here?

larsoner:
> in order to keep this version in sync with upstream, can we wait until tomorrow, when i can release and submit the PR here?

I would rather first merge this one ASAP to get things green (we don't like leaving CIs red), and whenever you get your version fixed tomorrow you can still make a PR here that is basically a `cp` then `git commit`. Then there is no time pressure (from our end) on you, and we still end up with the correct fix eventually

mne/externals/pymatreader/utils.py
drammock: I already made what I think are the needed code changes, but I don't have the time right now to get a build/test env working for scipy.  So: nope.

larsoner: https://github.com/scipy/scipy/pull/15004

larsoner:
@thht in particular you might want to wait to see if https://github.com/scipy/scipy/pull/15004 gets merged first. If it does, then you can use it instead. The bug will only hit people who have updated to scipy --pre in the last 6 days or so (our previously failing `--pre` CI run uses such a version)

larsoner:
We're going to need a `nilearn` fix, too:
```
/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/site-packages/nilearn/datasets/func.py:20: in <module>
    from scipy.io.matlab.miobase import MatReadError
E   ModuleNotFoundError: No module named 'scipy.io.matlab.miobase'; 'scipy.io.matlab' is not a package
```
@drammock I'll push a fix here that should get us around it in the meantime

drammock:
argh, CIs still going to fail even after pymatreader fixes, because `nilearn` has the same problem: https://github.com/mne-tools/mne-python/runs/4141568754?check_suite_focus=true#step:13:6934

larsoner:
I opened https://github.com/nilearn/nilearn/pull/3053 for `nilearn` to make sure we fix things there, too

PR #9975, opened 2021-11-08
MRG: Add argument overwrite to export for raws and epochs to match save.

mscheltienne:
I have to do a couple of I/O roundtrips to EEGLAB, and I could use an `overwrite` argument, set to False by default, for the export methods as in `inst.save` to prevent unintentional exports.

For `inst.save`, when overwrite was added, it was first initialized to True for one version and then changed to False.

> This defaults to True in 0.18 but will change to False in 0.19.
> New in version 0.18.

If you agree with the addition of this argument, do you want to follow the same path?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -36,7 +36,6 @@ Bugs
 
 - Add argument ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
 
-
 API changes
 ~~~~~~~~~~~
 - Nothing yet
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -26,6 +26,8 @@ Enhancements
 
 - Add show local maxima toggling button to :func:`mne.gui.locate_ieeg` (:gh:`9952` by `Alex Rockhill`_)
 
+- Add argment ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
+
 Bugs
 ~~~~
 - Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSE_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -26,8 +26,6 @@ Enhancements
 
 - Add show local maxima toggling button to :func:`mne.gui.locate_ieeg` (:gh:`9952` by `Alex Rockhill`_)
 
-- Add argment ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
-
 Bugs
 ~~~~
 - Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSE_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
@@ -35,6 +33,9 @@ Bugs
 - Fix bug with :class:`mne.Report` where figures were saved with ``bbox_inches='tight'``, which led to inconsistent sizes in sliders (:gh:`9966` by `Eric Larson`_)
 
 - Fix bug in :func:`mne.make_forward_solution` where sensor-sphere geometry check was incorrect (:gh:`9968` by `Eric Larson`_)
+
+- Add argument ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
+
 
 API changes
 ~~~~~~~~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -38,7 +38,7 @@ Bugs
 
 - Fix bug in :func:`mne.make_forward_solution` where sensor-sphere geometry check was incorrect (:gh:`9968` by `Eric Larson`_)
 
-- Add argument ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
+- Add argument ``overwrite`` to :func:`mne.export.export_raw`, :func:`mne.export.export_epochs`, :func:`mne.export.export_evokeds` and :func:`mne.export.export_evokeds_mff` (:gh:`9975` by `Mathieu Scheltienne`_)
 
 - :func:`mne.gui.coregistration` and the ``mne coreg`` command didn't respect the ``inspect`` parameter (:gh:`9972` by `Richard Höchenberger`_)
 
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1926,7 +1926,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         %(export_params_fmt)s
         %(overwrite)s
 
-        .. versionadded:: 1.0
+            .. versionadded:: 1.0
         %(verbose)s
 
         Notes
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1934,7 +1934,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         %(export_eeglab_note)s
         """
         from .export import export_epochs
-        export_epochs(fname, self, fmt, overwrite, verbose)
+        export_epochs(fname, self, fmt, overwrite=overwrite, verbose=verbose)
 
     def equalize_event_counts(self, event_ids=None, method='mintime'):
         """Equalize the number of trials in each condition.
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1914,7 +1914,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
                         split_naming, overwrite)
 
     @verbose
-    def export(self, fname, fmt='auto', verbose=None):
+    def export(self, fname, fmt='auto', overwrite=False, verbose=None):
         """Export Epochs to external formats.
 
         Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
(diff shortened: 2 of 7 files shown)


larsoner:
Agreed we should have it. However, the `export` stuff was added in 0.24 so I think we can just backport the addition of `overwrite` as a bugfix here to make our lives easier

larsoner:
(i.e., just have `overwrite=False` and don't bother with any deprecation cycle)

agramfort (approved):
good idea !

mne/epochs.py
```diff
@@ -1924,14 +1924,17 @@ def export(self, fname, fmt='auto', verbose=None):
         ----------
         %(export_params_fname)s
         %(export_params_fmt)s
+        %(overwrite)s
+
+        .. versionadded:: 1.0
```
larsoner, line 1929:
```suggestion
            .. versionadded:: 1.0
```

doc/changes/latest.inc
```diff
@@ -26,6 +26,8 @@ Enhancements
 
 - Add show local maxima toggling button to :func:`mne.gui.locate_ieeg` (:gh:`9952` by `Alex Rockhill`_)
 
+- Add argment ``overwrite`` to :func:`mne.export.export_raw` and :func:`mne.export.export_epochs` (:gh:`9975` by `Mathieu Scheltienne`_)
```
larsoner, line 29: Spelling error argment -> argument, and also move to BUG section

mne/io/base.py
```diff
@@ -1490,6 +1490,9 @@ def export(self, fname, fmt='auto', physical_range='auto',
         %(export_params_fmt)s
         %(export_params_physical_range)s
         %(export_params_add_ch_type)s
+        %(overwrite)s
+
+        .. versionadded:: 1.0
```
larsoner, line 1495:
```suggestion
            .. versionadded:: 1.0
```

mscheltienne:
Quick question while the CI is running, if you intend to backport this to 0.24 and to avoid the deprecation cycle, do you want to have `.. versionadded:: 1.0` or `.. versionadded:: 0.24`?

mscheltienne:
Ok, one more question: when I look at the documentation for the added export module, I see:

image

Why is `mne.export.export_evokeds_mff` the only public function out of the export functions? Shouldn't it be private as the export function for Raw/Epochs towards EEGLAB or EDF? It's covered by `mne.export.export_evokeds` and it's complicating the API without any gain IMO.

Also, I will add `overwrite` to `mne.export.export_evokeds` as well.

larsoner:
> Why is mne.export.export_evokeds_mff the only public function out of the export functions? 

It allows for extra arguments not usually needed or used by the other exporters. See https://github.com/mne-tools/mne-python/pull/9406#discussion_r635798097 and other points of discussion in that PR if interested

larsoner:
You can ignore pip-pre for now as it should be fixed by https://github.com/mne-tools/mne-python/pull/9973

mscheltienne:
Ok, it makes sense. Personally, I would have been in favor of passing `**kwargs` to the exporter + good documentation/check of the possible argument in `kwargs` per exporter rather than multiplying the public functions, but I see the point with the nasty parameter list.  

I will add `overwrite` to both evoked exporters then.

mscheltienne:
@larsoner Well adding `overwrite` for evoked is not as simple.. I think it would be best for MNE API to be consistent and to have this parameter everywhere, but `mffpy` doesn't support it,

```
git\mne-python\mne\export\_egimff.py:56: in export_evokeds_mff
    writer = mffpy.Writer(fname)
pyvenv\mne\lib\site-packages\mffpy\writer.py:34: in __init__
    self.filename = filename
pyvenv\mne\lib\site-packages\mffpy\writer.py:127: in filename
    assert not exists(fn), f"File '{fn}' exists already"
E   AssertionError: File 'C:\Users\Mathieu\AppData\Local\Temp\pytest-of-Mathieu\pytest-134\test_export_evokeds_to_mff_Fal1\evoked.mff' exists already
```

How do you feel about the idea of deleting the file prior to calling `mffpy.Writer`? I could also open a PR for a feature request on `mffpy`.

larsoner:
> How do you feel about the idea of deleting the file prior to calling mffpy.Writer? I could also open a PR for a feature request on mffpy.

I would open an issue upstream and for now delete at our end until they release a version that supports `overwriite` (if ever) and at that point triage what to do in MNE based on the version of `mmfpy` installed

mscheltienne:
@larsoner Alright, so let's leave this PR as it is. It doesn't yet include any change to the evoked export. 

I will open one upstream with the feature request + a second one on MNE with the change to include to export of evoked objects when the upstream change is merged and released.

larsoner:
> @larsoner Alright, so let's leave this PR as it is. It doesn't yet include any change to the evoked export.

Sorry, I meant for now delete *the file* at our end (not the code for `overwrite`). In other words, do add `overwrite` to `export_evokeds` and `export_evokeds_mff` in this PR

mscheltienne:
Oh alright, I will add that 'hack' a bit later this evening as I open the PR upstream.

mscheltienne:
I don't know if the proposal for Evoked is the best one... It's probably going to fail the tests as it does locally: 

```
git\mne-python\mne\export\_egimff.py:63: in export_evokeds_mff
    os.remove(fname)
E   PermissionError: [WinError 5] Access is denied: 'C:\\Users\\Mathieu\\AppData\\Local\\Temp\\pytest-of-Mathieu\\pytest-137\\test_export_evokeds_to_mff_Fal1\\evoked.mff'
```

larsoner:
This suggests that some file pointer has not been closed, maybe it's another MFF bug. If this is the case then maybe we skip MFF for now until they can handle an `overwrite` parameter *and* fix the file pointer bug (assuming that's what the issue really is)

mscheltienne:
Solved, I am not familiar with `mffpy`, and `fname` is in fact.. a directory.

mne/epochs.py
```diff
@@ -1914,7 +1914,7 @@ def save(self, fname, split_size='2GB', fmt='single', overwrite=False,
                         split_naming, overwrite)
 
     @verbose
-    def export(self, fname, fmt='auto', verbose=None):
+    def export(self, fname, fmt='auto', overwrite=False, verbose=None):
```
larsoner, line 1917: Easy enough to prevent people from accidentally doing `export(fname, fmt, True)` thinking they're doing `verbose=True)` (plus we actually want all `verbose` to come after a `*` anyway moving forward)
```suggestion
    def export(self, fname, fmt='auto', *, overwrite=False, verbose=None):
```

mne/export/_egimff.py
```diff
@@ -48,6 +53,15 @@ def export_evokeds_mff(fname, evoked, history=None, *, verbose=None):
...
     # Initialize writer
+    # Future changes: conditions based on version or mffpy requirement if
+    # https://github.com/BEL-Public/mffpy/pull/92 is merged and released.
+    try:
+        fname = _check_fname(fname, overwrite=False)
+    except FileExistsError:
+        if overwrite:
+            shutil.rmtree(fname)
+        else:
+            raise
```
larsoner, line 64: It looks like this doesn't actually *use* the `overwrite` parameter. How about:
```suggestion
    fname = _check_fname(fname, overwrite=overwrite)
    if op.exists(fname):
        os.remove(fname) if op.isfile(fname) else shutil.rmtree(fname)
```

mne/export/_egimff.py
```diff
@@ -3,17 +3,19 @@
...
 
 from ..io.egi.egimff import _import_mffpy
 from ..io.pick import pick_types, pick_channels
-from ..utils import verbose
+from ..utils import verbose, _check_fname
 
 
 @verbose
-def export_evokeds_mff(fname, evoked, history=None, *, verbose=None):
+def export_evokeds_mff(fname, evoked, history=None, overwrite=False, *,
```
larsoner, line 17:
```suggestion
def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
```

mne/export/_export.py
```diff
@@ -6,12 +6,12 @@
...
 
 from ._egimff import export_evokeds_mff
-from ..utils import verbose, logger, _validate_type
+from ..utils import verbose, logger, _validate_type, _check_fname
 
 
 @verbose
 def export_raw(fname, raw, fmt='auto', physical_range='auto',
-               add_ch_type=False, verbose=None):
+               add_ch_type=False, overwrite=False, verbose=None):
```
larsoner, line 14:
```suggestion
               add_ch_type=False, *, overwrite=False, verbose=None):
```

mne/export/_export.py
```diff
@@ -52,7 +56,7 @@ def export_raw(fname, raw, fmt='auto', physical_range='auto',
 
 
 @verbose
-def export_epochs(fname, epochs, fmt='auto', verbose=None):
+def export_epochs(fname, epochs, fmt='auto', overwrite=False, verbose=None):
```
larsoner, line 59:
```suggestion
def export_epochs(fname, epochs, fmt='auto', *, overwrite=False, verbose=None):
```

mne/export/_export.py
```diff
@@ -87,7 +95,7 @@ def export_epochs(fname, epochs, fmt='auto', verbose=None):
 
 
 @verbose
-def export_evokeds(fname, evoked, fmt='auto', verbose=None):
+def export_evokeds(fname, evoked, fmt='auto', overwrite=False, verbose=None):
```
larsoner, line 98:
```suggestion
def export_evokeds(fname, evoked, fmt='auto', *, overwrite=False, verbose=None):
```

larsoner:
FYI if you go to the "Files changed" pane, you can "add suggestion to batch" to commit all suggestions at once rather than going one-by-one (you can't do it from the main PR page)

mscheltienne:
@larsoner Yes I just noticed by hovering my mouse on the `add to batch` button midway through ahah Next time ;)

mne/export/_egimff.py
mscheltienne: It used the argument in the except, but that works too and is cleaner!

mscheltienne:
I'm adding one more quick fix to when pathlib.Path instances are passed as `fname`. `eeglabio` only supports `str`. For now, when a `Path` is provided, a TypeError is raised by `scipy`.

larsoner:
> I'm adding one more quick fix to when pathlib.Path instances are passed as fname. eeglabio only supports str. For now, when a Path is provided, a TypeError is raised by scipy.

To me this suggests `_check_fname` is not used properly somewhere, its output type is always `str`

https://github.com/mne-tools/mne-python/blob/a8522d213cf4b54ef682f2a874065bf62af65fdb/mne/utils/check.py#L149-L157

mscheltienne:
My bad.. wrong venv.. `_check_fname` is properly used everywhere now I think, so this problem when a `Path` is provided should not occur anymore. I'm leaving the 2 additional test cases in any way.

mscheltienne:
@larsoner This one should be good to go.

larsoner:
Thanks @mscheltienne !

PR #9977, opened 2021-11-08
Backport #9972

hoechenberger:

PR #9979, opened 2021-11-08
fix linkcheck

drammock:
this contributor's github page is 404, and their pages on both english and french versions of LinkedIn show up as "this profile is not available"


larsoner:
Thanks @drammock . @hoechenberger can you backport as part of #9977? Might as well bundle the backports in one PR

PR #9980, opened 2021-11-08
[BUG, MRG] Fix projection error

### LINKED ISSUE #9971: mistake in ECoG tutorial
Hello, 

In this tutorial:
https://mne.tools/stable/auto_tutorials/clinical/30_ecog.html

it is said: "We can project gamma power from the sensor data to the nearest locations on the pial surface and visualize that:"

However, to my understanding:
```
src = mne.read_source_spaces(
    op.join(subjects_dir, 'fsaverage', 'bem', 'fsaverage-ico-5-src.fif'))
```
reads out sources located on the white matter surface, not the pial surface.

If we want sources on the pial surface, I believe one should first do:
```
src = mne.setup_source_space('fsaverage', subjects_dir=subjects_dir,
                              spacing='ico5', surface='pial')
```

Indeed,
```
src = mne.read_source_spaces(
    op.join(subjects_dir, 'fsaverage', 'bem', 'fsaverage-ico-5-src.fif'))
```
and
```
src = mne.setup_source_space('fsaverage', subjects_dir=subjects_dir,
                              spacing='ico5', surface='white')
```
give the same result.

alexrockhill:
Fixes #9971.

I went ahead and did this one, looks better now I think.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -40,6 +40,8 @@ Bugs
 
 - :func:`mne.gui.coregistration` and the ``mne coreg`` command didn't respect the ``inspect`` parameter (:gh:`9972` by `Richard Höchenberger`_)
 
+- Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9978` by `Alex Rockhill`_)
+
 API changes
 ~~~~~~~~~~~
 - Nothing yet
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -40,7 +40,7 @@ Bugs
 
 - :func:`mne.gui.coregistration` and the ``mne coreg`` command didn't respect the ``inspect`` parameter (:gh:`9972` by `Richard Höchenberger`_)
 
-- Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9978` by `Alex Rockhill`_)
+- Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9980` by `Alex Rockhill`_)
 
 API changes
 ~~~~~~~~~~~
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3351,13 +3351,14 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     if src is None:  # fake a full surface one
         _validate_type(surface, str, 'surface', extra='when src is None')
-        rrs = [read_surface(op.join(subjects_dir, subject,
-                                    'surf', f'{hemi}.{surface}'))[0]
-               for hemi in ('lh', 'rh')]
+        rrs = [read_surface(op.join(
+            subjects_dir, subject, 'surf', f'{hemi}.{surface}'))[0]
+            for hemi in ('lh', 'rh')]
         src = SourceSpaces([
             dict(rr=rr / 1000., vertno=np.arange(len(rr)), type='surf',
                  coord_frame=FIFF.FIFFV_COORD_MRI)
             for rr in rrs])
+        rrs = np.concatenate(rrs, axis=0)
         keep_all = False
     else:
         if surface is None:
@@ -3365,9 +3366,9 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
             if src[0]['coord_frame'] == FIFF.FIFFV_COORD_HEAD:
                 rrs = apply_trans(trans, rrs)
         else:
-            rrs = [read_surface(op.join(subjects_dir, subject,
-                                        'surf', f'{hemi}.{surface}'))[0]
-                   for hemi in ('lh', 'rh')]
+            rrs = np.concatenate([read_surface(op.join(
+                subjects_dir, subject, 'surf', f'{hemi}.{surface}'))[0]
+                for hemi in ('lh', 'rh')], axis=0)
         keep_all = True
     # ensure it's a usable one
     klass = dict(
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3353,23 +3353,25 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     if surface is not None:
         surf_rr = [read_surface(op.join(subjects_dir, subject, 'surf',
-                                        f'{hemi}.{surface}'))[0]
+                                        f'{hemi}.{surface}'))[0] / 1000.
                    for hemi in ('lh', 'rh')]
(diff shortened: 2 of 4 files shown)


doc/changes/latest.inc
```diff
@@ -36,6 +36,8 @@ Bugs
 
 - Fix bug in :func:`mne.make_forward_solution` where sensor-sphere geometry check was incorrect (:gh:`9968` by `Eric Larson`_)
 
+- Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9978` by `Alex Rockhill`_)
```
larsoner, line 39:
```suggestion
- Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9980` by `Alex Rockhill`_)
```

mne/source_estimate.py
```diff
@@ -3261,10 +3261,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         The source space.
 
         .. warning:: If a surface source space is used, make sure that
-                     ``surf='pial'`` was used during construction.
+                     ``surface='pial'`` was used during construction.
```
larsoner, line 3264:
```suggestion
                     ``surface='pial'`` was used during construction,
                     or that you set ``surface='pial'`` here.
```

mne/source_estimate.py
```diff
@@ -3261,10 +3261,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
     %(picks_base)s good sEEG, ECoG, and DBS channels.
 
         .. versionadded:: 0.24
+    surface : str
```
larsoner, line 3268:
```suggestion
    surface : str | None
```

mne/source_estimate.py
```diff
@@ -3261,10 +3261,12 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
+        The surface to use if ``src=None``. Default is the pial surface.
```
larsoner, line 3269:
```suggestion
        The surface to use if ``src=None``. Default is the pial surface.
        If None, the source space surface will be used.
```

mne/source_estimate.py
```diff
@@ -3347,7 +3349,7 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     if src is None:  # fake a full surface one
         rrs = [read_surface(op.join(subjects_dir, subject,
-                                    'surf', f'{hemi}.pial'))[0]
+                                    'surf', f'{hemi}.{surface}'))[0]
```
larsoner, line 3352: This isn't quite the proposed solution [here](https://github.com/mne-tools/mne-python/issues/9971#issuecomment-962644529). It would be more like:
```
if surf is not None:
    surf_rrs = [...]
    rrs = surf_rrs
if src is None:
    _validate_type(surf, str, 'surf', extra='when src is None')
    rrs = surf_rrs
    ...
else:
    if surf is None:
        rrs = np.concatenate([s['rr'][s['vertno']] for s in src])
```
if surf is not None:
    rrs = ...

tutorials/clinical/30_ecog.py
```diff
@@ -181,12 +181,12 @@
 
 xyz_pts = np.array([dig['r'] for dig in evoked.info['dig']])
 
-src = mne.read_source_spaces(
-    op.join(subjects_dir, 'fsaverage', 'bem', 'fsaverage-ico-5-src.fif'))
+# use `src=None` for default source space, to specify spacing use
+# `src = mne.setup_source_space('fsaverage', subjects_dir=subjects_dir,
+#                               spacing='ico5', surface='pial')`
```
larsoner, line 186: ... then you don't actually need to make these changes because `pial` rrs will be used

mne/source_estimate.py
larsoner: ... and I guess we should use `surface` both places for consistency

alexrockhill:
Ok, I know this isn't the same logic as what you wrote @larsoner and so it has three lines copy-pasted but I think it's simpler/easier for me to understand what's going on. I'm happy to switch it back if you like what you wrote in the comment better.

mne/source_estimate.py
```diff
@@ -3261,10 +3261,14 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
...
         .. warning:: If a surface source space is used, make sure that
-                     ``surf='pial'`` was used during construction.
+                     ``surface='pial'`` was used during construction,
+                     or that you set ``surface='pial'`` here.
     %(picks_base)s good sEEG, ECoG, and DBS channels.
 
         .. versionadded:: 0.24
+    surface : str | None
+        The surface to use if ``src=None``. Default is the pial surface.
+        If None, the source space surface will be used.
```
larsoner, line 3271:
```suggestion
        If None, the source space surface will be used.
        
        .. versionadded:: 0.24.1
```

mne/source_estimate.py
```diff
@@ -3346,16 +3350,25 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
     subject = _check_subject(None, subject, raise_error=False)
     subjects_dir = get_subjects_dir(subjects_dir, raise_error=True)
     if src is None:  # fake a full surface one
-        rrs = [read_surface(op.join(subjects_dir, subject,
-                                    'surf', f'{hemi}.pial'))[0]
-               for hemi in ('lh', 'rh')]
+        _validate_type(surface, str, 'surface', extra='when src is None')
+        rrs = [read_surface(op.join(
+            subjects_dir, subject, 'surf', f'{hemi}.{surface}'))[0]
+            for hemi in ('lh', 'rh')]
```
larsoner, line 3356: This code is less DRY than a variant of what I proposed because it repeats the concatenation step. Code written this way makes maintenance harder. The flow here is essentially the same, but more DRY:
```
if surface is not None:
    surf_rr = np.concatenate([
        read_surface(op.join(subjects_dir, subject, 'surf',
                     f'{hemi}.{surface}'))[0])
        for hemi in ('lh', 'rh')]
if src is None:
    _validate_type(surface, str, 'surface', extra='when src is None')
    rrs = surf_rr
    src = SourceSpaces([
        dict(rr=rr / 1000., vertno=np.arange(len(rr)), type='surf',
             coord_frame=FIFF.FIFFV_COORD_MRI)
        for rr in rrs])
    keep_all = False
else:
    if surface is not None:
        rrs = surf_rr
    else:
        rrs = np.concatenate([s['rr'][s['vertno']] for s in src])
        if src[0]['coord_frame'] == FIFF.FIFFV_COORD_HEAD:
            rrs = apply_trans(trans, rrs)
    keep_all = True
```

larsoner (changes requested):

mne/source_estimate.py
alexrockhill: Sounds good to me, I was just confused by the first way you wrote it but this looks much more clear

larsoner:
Thanks @alexrockhill !

PR #9981, opened 2021-11-08
[MAINT, MRG] Fix Dipy link

alexrockhill:
Related to https://github.com/dipy/dipy/issues/2483. This will keep the documentation link to `Dipy` from going out of date every release.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -155,8 +155,8 @@ intersphinx_mapping = {
     'picard': ('https://pierreablin.github.io/picard/', None),
     'qdarkstyle': ('https://qdarkstylesheet.readthedocs.io/en/latest', None),
     'eeglabio': ('https://eeglabio.readthedocs.io/en/latest', None),
-    'dipy': ('https://dipy.org/documentation/1.4.0./',
-             'https://dipy.org/documentation/1.4.0./objects.inv/'),
+    'dipy': ('https://dipy.org/documentation/latest/documentation',
+             'https://dipy.org/documentation/latest/objects.inv/'),
     'pooch': ('https://www.fatiando.org/pooch/latest/', None),
 }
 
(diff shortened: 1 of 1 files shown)


larsoner:
Next time can you push doc-build-only changes like this with `[skip azp] [skip actions]`? Then only CircleCI will run and save some cycles

larsoner (changes requested):
https://38024-1301584-gh.circle-artifacts.com/0/dev/generated/mne.transforms.compute_volume_registration.html#mne.transforms.compute_volume_registration points to https://dipy.org/documentation/1.4.1./documentation/reference/dipy.align/#dipy.align.imwarp.DiffeomorphicMap which is a 404.

On main https://mne.tools/dev/generated/mne.transforms.compute_volume_registration.html#mne.transforms.compute_volume_registration points to https://dipy.org/documentation/1.4.0./reference/dipy.align/#dipy.align.imwarp.DiffeomorphicMap which is not 404. So we should not merge this until they fix their intersphinx (at least I think this is the problem?)

alexrockhill:
I think it's just that https://dipy.org/documentation/1.4.1./documentation/ is the documentation landing page but https://dipy.org/documentation/1.4.1. is a 404. I changed it to the one that was an actual landing page but I think even though the link doesn't work by itself, it works with intersphinx.

larsoner:
Can you push a commit (can be empty with `--allow-empty`) with `[skip azp] [skip actions]`? Locally with `make html_dev-noplot` are things fixed?

alexrockhill:
I tested this locally and it works, apparently the `/` at the end of the web address is vital. Anyway should be good to go now.

alexrockhill:
Just checked and the link works here as well https://38076-1301584-gh.circle-artifacts.com/0/dev/generated/mne.transforms.compute_volume_registration.html?highlight=compute_vol#mne.transforms.compute_volume_registration

larsoner:
Thanks @alexrockhill !

PR #9982, opened 2021-11-08
MNT: mne-qt-browser config and doc

marsipu:
This renames mne-qt-browser related config-variables for consistency:

MNE_BROWSE_BACKEND -> MNE_BROWSER_BACKEND
MNE_BROWSE_USE_OPENGL -> MNE_BROWSER_USE_OPENGL
~~MNE_BROWSE_RAW_SIZE -> MNE_BROWSER_SIZE~~

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -32,7 +32,7 @@ Enhancements
 
 Bugs
 ~~~~
-- Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSE_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
+- Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSER_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
 
 - Fix bug with :class:`mne.Report` where figures were saved with ``bbox_inches='tight'``, which led to inconsistent sizes in sliders (:gh:`9966` by `Eric Larson`_)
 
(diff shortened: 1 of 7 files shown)


larsoner:
`MNE_BROWSE_RAW_SIZE -> MNE_BROWSER_SIZE` this is a backward incompatible change, that variable has been around a very long time. I think we should leave that one alone probably

marsipu:
I put some other little fixes for the Docs in here, I hope that's ok

larsoner:
Still draft or ready for merge?

marsipu:
@larsoner I tried to add `PyQt5.QtWidgets.QMainWindow` to the docstring of Raw.plot() but sphinx.autodoc doesn't seem to find the target-reference. Do I need to add PyQt5 somewhere?

drammock:
Search `docs/conf.py` for "intersphinx" and you'll see how it's done.

marsipu:
> 
> 
> Search `docs/conf.py` for "intersphinx" and you'll see how it's done.

Thanks Dan. But I can't get it to work with PyQt5, even with the [suggested solution](https://stackoverflow.com/questions/47102004/how-to-properly-link-to-pyqt5-documentation-using-intersphinx).
And that seems to be a bit much for just this one class.
What would you suggest? Just having a local link maybe?

larsoner:
Intersphinx will only work if PyQt5 docs are built with Sphinx. Either don't bother linking and switch to double-backticks (probably what I would do) or manually link with something like
```
... with `QApplication <https://...>`__.
```

drammock:
hmm, both of the suggestions on that page are kind of onerous anyway.  I'd say just put it in double-backticks so it renders as code but doesn't link anywhere

larsoner:
oh they are built with sphinx. To debug this what you can do is:
```
python -m sphinx.ext.intersphinx https://www.riverbankcomputing.com/static/Docs/PyQt5/ > qt.txt
```
then inspect `qt.txt` for the thing you want to link to

larsoner:
... looks like
```
:sip:class:`PyQt5.QtWidgets.QApplication`
```
should in theory do it. But now I see people talk about this in the solution. And as @drammock says the workarounds are a pain. Just double-backtick it...

marsipu:
Ok, thanks for your help @drammock @larsoner

agramfort (approved):

Issue #9894, opened 2021-10-23
Use fixes._compare_version for version checks everywhere


hoechenberger:
#9754 introduced `fixes._compare_version()` that uses `setuptools`, as `distutils` – which we used so far – has been deprecated. We should ensure all our code that currently relies on `distutils` (`LooseVersion`) gets switched to `_compare_version`

PR #9988, opened 2021-11-10
MRG: Make check_version() use _compare_version() internally

hoechenberger:
Get rid of `LooseVersion` usage here.

Tiny step toward resolving #9894


larsoner:
Thanks @hoechenberger !

Issue #9989, opened 2021-11-10
Improve I/O with EEGLAB


mscheltienne:
I have to work a bit with EEGLAB, and I'm not very familiar with it (yet) but I already noticed a couple of improvements that could be made to the I/O in my opinion. 

### Reader

Channel type is hard-coded to `'eeg'` [here](https://github.com/mne-tools/mne-python/blob/main/mne/io/eeglab/eeglab.py#L152) while EEGLAB seems to support a larger variety of channel types. 
From the [documentation](https://eeglab.org/tutorials/04_Import/Channel_Locations.html):

> ‘EEG’, ‘MEG’, ‘EMG’, ‘ECG’, ‘Events’, etc.

- Try to retrieve the channel types saved by EEGLAB and convert it to one of MNE channel types.
  (e.g. `'EEG'` -> `'eeg'`, `'Events'` -> `'stim'`)

### Writer / Export

Looking at `export_set` in `eeglabio`, it has an argument `ref_channels` that is not used by the export code. If I'm not mistaken, the referencing information is present in `Info`, so it could be passed to the `.set` file via this argument.

```
eeglabio.raw.export_set(
        fname, data=raw.get_data(picks=ch_names), sfreq=raw.info['sfreq'],
        ch_names=ch_names, ch_locs=cart_coords, annotations=annotations)
```

Support for channel type is not included in `eeglabio`, I would like to add it https://github.com/jackz314/eeglabio/pull/4 
Once added to `eeglabio`, the exporter could make use of this additional argument.

cbrnr:
Sounds good! However, MNE does not keep track of the reference channel so unfortunately this won't work automatically.

mscheltienne:
@cbrnr Arf! Too bad for that, but to be honest my main priority is the channel types :)

Now that I think about it, what about projectors that are not yet applied? I think, for now, the export is not applying them.. maybe we could add a `proj` argument also to add the possibility to apply them prior to export. 
If so, we could also check for the presence of a common average reference projector, in which case we could set `ref_channels='average'` during the export. If MNE is not keeping track of the reference channel(s), then I think this is the only case in which `ref_channels` can be define.

cbrnr:
True, but I'd rather leave it to the user to define what they want to export. That way, we don't have to add another parameter (which as you correctly say only works for that single case).

mscheltienne:
[...]

Issue #9560, opened 2021-07-12
Make `mne.io.Raw` available as `mne.Raw`


cbrnr:
Two of the most important data classes, `mne.Epochs` and `mne.Evoked`, are available directly in `mne`. However, the class for continuous data is available as `mne.io.Raw` and not `mne.Raw` (because all `read_raw_*` readers also live in `mne.io`). For consistency, I propose to make `mne.io.Raw` available as `mne.Raw` (but keep `mne.io.Raw` for backward compatibility). WDYT?

larsoner:
-1 from me, I think if we think of `mne.io` as being for "raw IO" and all of our `mne.io.read_raw*` functions live there, the class should, too

agramfort:
-1 too

for the same reason. I also think our root namespace is already too big....

>

cbrnr:
+1 from me because IMO the three most important classes that represent data (`Raw`, `Epochs`, `Evoked`) should live together. @larsoner there is at least one `read_raw_*` function which produces an `Epochs` object (`read_raw_eeglab()`). @agramfort I'd rather move less important things into submodules and have these three classes in the root.

sappelhoff:
+1 because it seems natural to have the three main classes available from `mne`. @cbrnr's arguments make sense to me. But I won't personally spend time to get this done, so I guess I am a +0.5

drammock:
@cbrnr's argument makes sense to me too, but I agree with @agramfort that the root namespace is over-crowded.  I would not do this as a one-off change, but I would consider doing this as part of a larger namespace overhaul, done after careful consideration/planning, and with a longer-than-normal deprecation cycle.

larsoner:
Closing since I think we've converged on not changing this

PR #9990, opened 2021-11-10
MRG: Add support to retrieve channel type from ['chanlocs']['type'] with EEGLAB reader

mscheltienne:
Following #9989, a quick proposition to read channel types with EEGLAB reader.

In the reader, the data is read in a variable named `eeg` since EEGLAB is mostly focused on EEG datasets. However, in practice it does support more data types; and I could rename left and right this variable to e.g. `data` to make the code a bit more general if you want.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -44,7 +44,7 @@ Bugs
 
 - Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9980` by `Alex Rockhill`_)
 
-- Add channel type support when reading from EEGLAB ``.set`` format with :func:`mne.io.read_raw_eeglab` and :func:`mne.io.read_epochs_eeglab` (:gh:`9990` by `Mathieu Scheltienne`_)
+- Fix channel type support when reading from EEGLAB ``.set`` format with :func:`mne.io.read_raw_eeglab` and :func:`mne.read_epochs_eeglab` (:gh:`9990` by `Mathieu Scheltienne`_)
 
 API changes
 ~~~~~~~~~~~
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -44,6 +44,8 @@ Bugs
 
 - Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9980` by `Alex Rockhill`_)
 
+- Add channel type support when reading from EEGLAB ``.set`` format with :func:`mne.io.read_raw_eeglab` and :func:`mne.io.read_epochs_eeglab` (:gh:`9990` by `Mathieu Scheltienne`_)
+
 API changes
 ~~~~~~~~~~~
 - Nothing yet
--- a/mne/io/eeglab/eeglab.py
+++ b/mne/io/eeglab/eeglab.py
@@ -108,7 +108,7 @@ def _eeg_has_montage_information(eeg):
 def _get_montage_information(eeg, get_pos):
     """Get channel name, type and montage information from ['chanlocs']."""
     ch_names, ch_types, pos_ch_names, pos = list(), list(), list(), list()
-    unknown_types = set()
+    unknown_types = dict()
     for chanloc in eeg.chanlocs:
         # channel name
         ch_names.append(chanloc['labels'])
@@ -121,7 +121,10 @@ def _get_montage_information(eeg, get_pos):
             if try_type in _PICK_TYPES_KEYS:
                 ch_type = try_type
             else:
-                unknown_types.add(try_type)
+                if try_type in unknown_types:
+                    unknown_types[try_type].append(chanloc['labels'])
+                else:
+                    unknown_types[try_type] = [chanloc['labels']]
         ch_types.append(ch_type)
 
         # channel loc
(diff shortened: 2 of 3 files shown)


cbrnr:
I think a test would be really useful here. Can you use an existing `.set` file and patch the stuff that needs to be adapted on the fly?

mscheltienne:
@cbrnr Oh right.. I'm dumb, I can just edit the `['chanlocs']['type']` from the loaded file on the fly..

cbrnr:
@mscheltienne in case you're wondering why I edited the title, you don't have to include a period (super nitpicky I know).

mscheltienne:
Ahah alright, I was looking at it and had no idea what changed!

mscheltienne:
So.. it's actually a pain to test for now. The test I added is restricted to the modification to the private function `_get_montage_information` in the end.
IMO, a complete I/O roundtrip test FIF -> SET -> FIF test requires either a different `.set` file in the testing dataset OR the addition of https://github.com/jackz314/eeglabio/pull/4

By the way, I did test with the `eeglabio` PR branch the complete roundtrip FIF -> SET -> FIF, and both EEGLAB and MNE displayed the correct channel types.

larsoner:
Ready for review/merge @mscheltienne ?

mne/io/eeglab/eeglab.py
```diff
@@ -104,11 +105,26 @@ def _eeg_has_montage_information(eeg):
...
+        if 'type' in chanloc:
+            ch_type = chanloc['type']
+            if isinstance(ch_type, str) and 0 < len(ch_type):
+                ch_type = ch_type.strip().lower()
+                ch_types.append(ch_type if ch_type in _PICK_TYPES_KEYS
+                                else 'eeg')
+            else:
+                ch_types.append('eeg')  # guarantee same size as ch_names
+        else:
+            ch_types.append('eeg')  # guarantee same size as ch_names
```
larsoner, line 125: Instead of repeating a bunch of `ch_types.append` statements and using a lot of conditionals, it seems to be cleaner to say "assume EEG unless shown otherwise" like:
```suggestion
        ch_type = 'eeg'
        try_type = chanloc.get('type', None)
        if isinstance(try_type, str):
            try_type = try_type.strip().lower()
            if try_type in _PICK_TYPES_KEYS:
                ch_type = try_type
        ch_types.append(ch_type)
```

larsoner (commented):
Other than a simplification idea, LGTM!

larsoner (changes requested):
... oh actually you need a `latest.inc` update. I think this should qualify as a bugfix, as 1) it fixes type handling and 2) people's code could actually break from this PR if they've been picking types as `'eeg'` that will now be something else

mne/io/eeglab/eeglab.py
```diff
@@ -104,11 +105,23 @@ def _eeg_has_montage_information(eeg):
...
+        # channel name
         ch_names.append(chanloc['labels'])
+
+        # channel type
+        ch_type = 'eeg'
+        try_type = chanloc.get('type', None)
+        if isinstance(try_type, str):
+            try_type = try_type.strip().lower()
+            if try_type in _PICK_TYPES_KEYS:
+                ch_type = try_type
```
larsoner, line 121: Actually thinking about this functionality now, maybe it's a good idea to warn people if we encounter unknown types? For example something like this could help
```
        unknown_types = set()
        ...
            if try_type in _PICK_TYPES_KEYS:
                ch_type = try_type
            else:
                unknown_types.add(ch_type)
        if len(unknown_types):
            warn(f'Unknown types found, setting as type EEG: {sorted(unknown_types)}')
```

larsoner: ... or maybe better, keep a list of each channel name that was a given unknown type, then warn with
```
Unknown types found setting as type EEG:
'something': ['somechan1', 'somechan2']
'someelse': ['someelsechan1']
```

mscheltienne: True, I wanted to leave untouched the default behavior to set type as `'eeg'`, but a warning would be nice! I'll add it.

mscheltienne: Barely saw your second reply, but done.

mscheltienne:
I don't know why the doc build seems to fail.. Any idea?

larsoner:
It's https://github.com/pydata/pydata-sphinx-theme/issues/508, I'll open a PR to fix it

larsoner (approved):

mscheltienne:
@larsoner Any particular reason why this one lives in the `mne.` namespace?

https://github.com/mne-tools/mne-python/blob/main/mne/__init__.py#L34

mscheltienne:
Same remark for `read_epochs_kit` one line above.

mscheltienne:
And for `from .io import read_epochs_fieldtrip, read_evoked_fieldtrip, read_evokeds_mff` line 100.

cbrnr:
This discussion sounds familiar 😄

larsoner:
Discussion in https://github.com/mne-tools/mne-python/issues/9560

doc/changes/latest.inc
```diff
@@ -44,6 +44,8 @@ Bugs
 
 - Fix incorrect projection of source space onto white matter surface instead of pial in :ref:`tut-working-with-ecog` (:gh:`9980` by `Alex Rockhill`_)
 
+- Add channel type support when reading from EEGLAB ``.set`` format with :func:`mne.io.read_raw_eeglab` and :func:`mne.io.read_epochs_eeglab` (:gh:`9990` by `Mathieu Scheltienne`_)
```
larsoner, line 47:
```suggestion
- Fix channel type support when reading from EEGLAB ``.set`` format with :func:`mne.io.read_raw_eeglab` and :func:`mne.read_epochs_eeglab` (:gh:`9990` by `Mathieu Scheltienne`_)
```

mscheltienne:
Alright, I didn't read it entirely last time; and I didn't remember it also mentioned the `read_epochs` methods. So `io` is only for raw instance I/O because `Raw` lives in `mne.io`.. at least I'll know it now.

larsoner:
Thanks @mscheltienne !

PR #9991, opened 2021-11-10
[BUG, MRG] Revisit affine registration

alexrockhill:
The affine registration of the CT with the MR wasn't working optimally, alignments were slightly off, thanks for reporting @leosunpsy.

So I went back and looked and it looks like the `Dipy` method was to use the pre-alignment affines as we had at one point. I'm not sure why that wasn't working but it seems to work just fine now.

### DIFF
--- a/mne/transforms.py
+++ b/mne/transforms.py
@@ -1663,8 +1663,9 @@ def _compute_volume_registration(moving, static, pipeline, zooms, niter):
     _require_version('dipy', 'SDR morph', '0.10.1')
     import nibabel as nib
     with np.testing.suppress_warnings():
+        from dipy.align.imaffine import AffineMap
         from dipy.align import (affine_registration, center_of_mass,
-                                translation, rigid, affine, resample,
+                                translation, rigid, affine,
                                 imwarp, metrics)
 
     # input validation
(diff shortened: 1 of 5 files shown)


larsoner:
Did you run tests locally? I'd expect that perhaps some tolerances in `tests/test_transforms.py` might need to be adjusted, and/or some in `tests/test_morph.py` and `tests/test_source_estimate.py`

alexrockhill:
> Did you run tests locally? I'd expect that perhaps some tolerances in `tests/test_transforms.py` might need to be adjusted, and/or some in `tests/test_morph.py` and `tests/test_source_estimate.py`

I forgot which tests would fail so I was running all the tests here before going back and fixing those but I can check those two now.

alexrockhill:
Ok all the tests should pass now. I changed most of the `test_morph` back to how it was in `0.24` stable but the SDR morph caused one failed test with `test_morph::test_mixed_source_morph`. If you want to look and see if that looks okay @larsoner, that would be really helpful.

I will actually check that the ieeg tutorial that uses it runs okay now, I haven't done that yet.

alexrockhill:
The SDR morph on the tutorial looked terrible so I think we should not use pre-align for the SDR morph, which might also fix the test. I'll try that now.

alexrockhill:
Ok this should be the best solution. It involved applying the registration before the SDR morph but not on each of the steps which is a bit curious but seems to tick all the boxes for passing the tests. I just touched the ieeg tutorial as well so I'll need to double check that that turns out okay.

larsoner:
Looks like that's the strategy we used [originally and in 0.23](https://github.com/mne-tools/mne-python/blob/maint/0.23/mne/morph.py#L1075-L1086) -- we should probably backport this as a fix

larsoner:
```
mne/tests/test_surface.py:351: in test_warp_montage_volume
    assert np.linalg.norm(  # off by less than 1.5 cm
E   AssertionError: assert 0.029795600495524505 < 0.015
```

alexrockhill:
> ```
> mne/tests/test_surface.py:351: in test_warp_montage_volume
>     assert np.linalg.norm(  # off by less than 1.5 cm
> E   AssertionError: assert 0.029795600495524505 < 0.015
> ```

Yeah, it looks off in the tutorial as well, I checked and the affine registration works great but then the SDR misaligns them, looking into it, just a minute

alexrockhill:
Okay, not only did all the tests pass locally but I was able to increase the threshold of the montage warp to the default as long as `zooms=4`! All the tests pass locally, it was just a small affine typo, I just had to walk away and come back with fresh eyes.

mne/transforms.py
```diff
@@ -1694,31 +1693,27 @@ def _compute_volume_registration(moving, static, pipeline, zooms, niter):
...
+                                   static_zoomed.shape, static_affine,
+                                   moving_zoomed.shape, moving_affine)
+            moving_zoomed = affine_map.transform(moving_zoomed)
             sdr = imwarp.SymmetricDiffeomorphicRegistration(
                 metrics.CCMetric(3), niter[step])
             with wrapped_stdout(indent='    ', cull_newlines=True):
                 sdr_morph = sdr.optimize(static_zoomed, moving_zoomed,
-                                         static_affine, moving_affine)
-            moving_zoomed = sdr_morph.transform(moving_zoomed)
+                                         static_affine, static_affine)
```
alexrockhill, line 1708: This was the trouble, because it was aligned, the affines should both be the static affine

larsoner: Indeed that makes sense!

larsoner:
Thanks @alexrockhill !

PR #9994, opened 2021-11-11
MRG: Issue warning when exporting with unapplied projectors and improve docstring

mscheltienne:
I added a warning when exporting with unapplied projectors to all 4 export functions:
- mne.export.export_raw
- mne.export.export_epochs
- mne.export.export_evokeds
- mne.export.export_evokeds_mff

I added this warning in their docstrings:

```
.. warning::
        Export does not apply projector. Unapplied projectors will be lost.
        Consider using :meth:`mne.io.Raw.apply_proj` before exporting.
```

Test for warning when exporting evoked with unapplied proj is missing because I didn't have any good idea on how to add a projector to the EGI evoked instance loaded.

### DIFF
--- a/mne/export/_egimff.py
+++ b/mne/export/_egimff.py
@@ -42,6 +42,7 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
     Notes
     -----
     .. versionadded:: 0.24
+
     %(export_warning_note_evoked)s
 
     Only EEG channels are written to the output file.
--- a/mne/export/_egimff.py
+++ b/mne/export/_egimff.py
@@ -21,7 +21,6 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
     """Export evoked dataset to MFF.
 
     %(export_warning)s :func:`mne.write_evokeds` instead.
-
     .. warning::
         Export does not apply projector. Unapplied projectors will be lost.
         Consider using :meth:`mne.Evoked.apply_proj` before exporting.
--- a/mne/export/_egimff.py
+++ b/mne/export/_egimff.py
@@ -55,7 +55,7 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
     sampling_rate = int(info['sfreq'])
 
     # check for unapplied projectors
-    if any(not proj['active'] for proj in evoked.info['projs']):
+    if any(not proj['active'] for proj in evoked[0].info['projs']):
         warn('Evoked instance has unapplied projectors. Consider applying '
              'them before exporting with evoked.apply_proj().')
 
--- a/mne/export/_egimff.py
+++ b/mne/export/_egimff.py
@@ -20,8 +20,7 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
                        verbose=None):
     """Export evoked dataset to MFF.
 
-    %(export_warning)s :func:`mne.write_evokeds` instead.
-    %(export_warning_proj)s :meth:`mne.Evoked.apply_proj`.
+    %(export_warning)s
 
     Parameters
     ----------
@@ -43,6 +42,7 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
     Notes
     -----
     .. versionadded:: 0.24
+    %(export_warning_note_evoked)s
 
     Only EEG channels are written to the output file.
     ``info['device_info']['type']`` must be a valid MFF recording device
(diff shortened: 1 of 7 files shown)


mscheltienne:
I don't know how to fix this style issue.. any idea is welcome.

mne/io/base.py
```diff
@@ -1483,6 +1483,7 @@ def export(self, fname, fmt='auto', physical_range='auto',
 
         Supported formats: EEGLAB (set, uses :mod:`eeglabio`)
         %(export_warning)s :meth:`save` instead.
+        %(export_warning_proj)s :meth:`apply_proj`.
```
larsoner, line 1486: This level of detail seems better suited to the Notes section. Putting too many warnings or notes in the beginning makes it hard for people to get to the params section. I would modify `export_warning` to say `See Notes for details` then in Notes you can have the text about projections being one thing that will not be saved. This part I think you can do without having it in a `.. warning::` block anymore, too.

mne/export/_egimff.py
```diff
@@ -40,6 +42,7 @@ def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
     Notes
     -----
     .. versionadded:: 0.24
+    %(export_warning_note_evoked)s
```
larsoner, line 45: Not sure if it's here or above, but:
```
E   mne.export._egimff.export_evokeds_mff : GL03 : Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
```

mscheltienne:
Doc is looking good IMO:
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.io.Raw.html?highlight=export#mne.io.Raw.export
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.Epochs.html?highlight=export#mne.Epochs.export
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_raw.html?highlight=export
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_epochs.html?highlight=export
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_evokeds.html?highlight=export
- https://38350-1301584-gh.circle-artifacts.com/0/dev/generated/mne.export.export_evokeds_mff.html?highlight=export

mne/export/_egimff.py
mscheltienne: Yes, same as before this additional change... I'm sorry, but I don't get why this one is being raised. Is it because the docstring starts with a 'summary' line and does not have a 'description' before the warning?

drammock: the multi-line string `export_warning_note` starts and ends with a newline.  Probably you need to remove one or both of those.  I'd guess it's the trailing newline.

drammock (approved):

mscheltienne:
@cbrnr Ahah, I didn't notice I put one in the title again!

mscheltienne:
@drammock Let's see if this fixes it. I thought the `fill_doc` decorator was stripping before/after. By the way, is the `verbose` decorator also filling the docstring?

```
@verbose
def export_evokeds_mff(fname, evoked, history=None, *, overwrite=False,
                       verbose=None):
```

mne/export/_egimff.py
mscheltienne: Looks like that's not it.. or I missed something.

cbrnr:
> Ahah, I didn't notice I put one in the title again!

And in commit messages... 😄

drammock:
@mscheltienne here's how I found/fixed the problems:

in ipython REPL, our rendered docstrings are visible in plaintext via `mne.Epochs.export?` or `mne.export.export_evokeds_mff?`.  That made it pretty clear to see, e.g., where the "two blank lines" was happening.  The `Epochs.export` problem was more subtle, but the fix was based on the intuition that rST `.. whatever::` blocks usually need a blank line above and/or below (but numpydoc requires it be *at most* one line)

drammock:
...of course even that was apparently not enough, because I forgot (for the hundredth time) that subsequent `make html_dev-noplot` calls will ignore unchanged files, so fixing the errors one at a time actually *masked* some errors from me on subsequent runs.  Will push more fixes momentarily.

larsoner:
FWIW I usually find it easiest not to have a newline at the beginning or end in the replacement string. Then the newlines you use in the __doc__ will be the only newlines (at the beginning and end) that you get, so it generally simplifies things

mscheltienne:
Thanks for the tip with IPython and `?`, it makes it way easier to see the double line-skip indeed.

larsoner:
Thanks @mscheltienne @drammock !

PR #9998, opened 2021-11-12
MRG: Don't overwrite Report.fname upon saving

wmvanvliet:
When calling `mne.Report.save()`, we don't want to overwrite the existing `report.fname` attribute, as it breaks this pattern (which I happen to use a lot):

```python
with open_report('report.h5') as report:
    report.add_figure(fig)
    report.save('report.html')  # save HTML version
# End of context block. Report is saved to report.h5
```

I like to keep both an `.h5` and `.html` version of the report. Before this PR, saving to HTML will overwrite `report.fname` and hence at the end of the context it gets saved as HTML again, instead of the HDF5 file that was used when opening the report.

### DIFF
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -2437,7 +2437,8 @@ class Report(object):
         if open_browser and not is_hdf5 and not building_doc:
             webbrowser.open_new_tab('file://' + fname)
 
-        self.fname = fname
+        if self.fname is None:
+            self.fname = fname
         return fname
 
     def __enter__(self):
--- a/mne/report/report.py
+++ b/mne/report/report.py
@@ -705,7 +705,8 @@ class Report(object):
         return (
             'baseline', 'cov_fname', 'include', '_content', 'image_format',
             'info_fname', '_dom_id', 'raw_psd', 'projs',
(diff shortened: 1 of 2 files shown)


mne/report/report.py
```diff
@@ -2437,7 +2437,8 @@ def save(self, fname=None, open_browser=True, overwrite=False,
         if open_browser and not is_hdf5 and not building_doc:
             webbrowser.open_new_tab('file://' + fname)
 
-        self.fname = fname
+        if self.fname is None:
```
larsoner, line 2440: To fix the failing test, I think this either this needs to be `getattr(self, 'fname', None)` or you need to add `fname` to `_get_state_params`

larsoner (commented):
Don't forget `latest.inc` update in the bug section, we should also backport this. @hoechenberger  can you also look?

hoechenberger:
I will have a look, please don't merge until I Report back (pun intended)

hoechenberger (approved):

hoechenberger:
Ready to go once CI comes back green.

agramfort:
Thx @wmvanvliet @hoechenberger

PR #10002, opened 2021-11-13
Robust fif channel renaming of long channels names

bloyl:
I have some workflows the store source space label time courses in `Raw` instances.

The current renaming scheme (https://github.com/mne-tools/mne-python/blob/main/mne/io/meas_info.py#L110) works for most atlases but fails for `aparc_sub`, depending on the ordering of duplicated channels.

<details>
  <summary>Minimal working example</summary>

  ```
  from mne import create_info
  from mne.io import write_info
  
  ch_names = ['supramarginal_1-lh',
              'supramarginal_1-rh',
              'supramarginal_10-lh',
              'supramarginal_2-lh',
              'supramarginal_2-rh',
              'supramarginal_3-lh',
              'supramarginal_3-rh',
              'supramarginal_4-lh',
              'supramarginal_4-rh',
              'supramarginal_5-lh',
              'supramarginal_5-rh',
              'supramarginal_6-lh',
              'supramarginal_6-rh',
              'supramarginal_7-lh',
              'supramarginal_7-rh',
              'supramarginal_8-lh',
              'supramarginal_8-rh',
              'supramarginal_9-lh',
              'supramarginal_9-rh']
  
  info = create_info(ch_names, 1000., verbose='error')
  write_info('tmp-info.fif', info)
  ```
</details>

This PR extends the renaming strategy to first append a digit (current approach) and if that fails to generate a unique `ch_name` tries to append a lower case letter.

### DIFF
--- a/mne/io/meas_info.py
+++ b/mne/io/meas_info.py
@@ -109,6 +109,7 @@ def _get_valid_units():
 @verbose
 def _unique_channel_names(ch_names, max_length=None, verbose=None):
     """Ensure unique channel names."""
+    suffixes = tuple('abcdefghijklmnopqrstuvwxyz')
     if max_length is not None:
         ch_names[:] = [name[:max_length] for name in ch_names]
     unique_ids = np.unique(ch_names, return_index=True)[1]
@@ -129,14 +130,17 @@ def _unique_channel_names(ch_names, max_length=None, verbose=None):
             n_keep = min(len(ch_stem), n_keep)
             ch_stem = ch_stem[:n_keep]
             for idx, ch_idx in enumerate(overlaps):
-                ch_name = ch_stem + '-%s' % idx
+                # try idx first, then loop through lower case chars
+                for suffix in (idx,) + suffixes:
+                    ch_name = ch_stem + '-%s' % suffix
+                    if ch_name not in ch_names:
+                        break
                 if ch_name not in ch_names:
                     ch_names[ch_idx] = ch_name
                 else:
-                    raise ValueError('Adding a running number for a '
+                    raise ValueError('Adding a single alphanumeric for a '
                                      'duplicate resulted in another '
                                      'duplicate name %s' % ch_name)
-
     return ch_names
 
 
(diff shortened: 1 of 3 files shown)


mne/io/meas_info.py
```diff
@@ -109,6 +109,7 @@ def _get_valid_units():
 @verbose
 def _unique_channel_names(ch_names, max_length=None, verbose=None):
     """Ensure unique channel names."""
+    suffixes = tuple('abcdefghijklmnopqrstuvwxyz')
```
agramfort, line 112: doing something like:

```
import string
suffixes = tuple(string.ascii_lowercase)
```

would be nicer

agramfort (commented):
besides LGTM

@larsoner wdyt?

agramfort (approved):
I'll let a second reviewer merge.

maybe add a what's new entry while you're at it?

thx @bloyl

mne/io/tests/test_meas_info.py
```diff
@@ -141,8 +141,8 @@ def test_duplicate_name_correction():
     assert info['ch_names'] == ['A-0', 'A-1', 'A-2']
 
     # When running number is not possible
-    with pytest.raises(ValueError, match='Adding a running number'):
-        create_info(['A', 'A', 'A-0'], 1000., verbose='error')
+    info = create_info(['A', 'A', 'A-0'], 1000., verbose='error')
+    assert info['ch_names'] == ['A-a', 'A-1', 'A-0']
```
larsoner, line 145: Might be better to adjust the original failure mode test rather than replace it:
```suggestion
    with pytest.raises(ValueError, match='Adding a running alphanumeric'):
        create_info(['A'] * 27 + ['A-0'], 1000., verbose='error')
    info = create_info(['A', 'A', 'A-0'], 1000., verbose='error')
    assert info['ch_names'] == ['A-a', 'A-1', 'A-0']
```

larsoner (changes requested):
Also can you add an entry to the BUG section? I think this probably qualifies as a bug...

agramfort:
let's merge when green

PR #10004, opened 2021-11-15
Add argument overwrite to ICA.save()

mscheltienne:
Adding a check + overwrite argument to ``ICA.save()`` to avoid overwriting existing ICAs.

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -32,6 +32,8 @@ Enhancements
 
 - Improve docstring of export functions :func:`mne.export.export_raw`, :func:`mne.export.export_epochs`, :func:`mne.export.export_evokeds`, :func:`mne.export.export_evokeds_mff` and issue a warning when there are unapplied projectors (:gh:`9994` by `Mathieu Scheltienne`_)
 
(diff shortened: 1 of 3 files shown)


doc/changes/latest.inc
```diff
@@ -32,6 +32,8 @@ Enhancements
 
 - Improve docstring of export functions :func:`mne.export.export_raw`, :func:`mne.export.export_epochs`, :func:`mne.export.export_evokeds`, :func:`mne.export.export_evokeds_mff` and issue a warning when there are unapplied projectors (:gh:`9994` by `Mathieu Scheltienne`_)
 
+- Add argument ``overwrite`` to `mne.preprocessing.ICA.save` to check for existing files (:gh:`10004` by `Mathieu Scheltienne`_)
```
larsoner, line 35: If we are doing this without any deprecation cycle then this has to go in the BUG section

mscheltienne: I moved it to bug; if you prefer to have a deprecation cycle I can revert the change and add a warning instead.

larsoner: I think we've done it both ways before, so treating it as a fix is fine here

larsoner:
Thanks @mscheltienne !

PR #10007, opened 2021-11-16
DOC: stc_near_sensors

### LINKED ISSUE #9993: mne.stc_near_sensors options desciption
Hello,

If I understood different options of mne.stc_near_sensors correctly, I was wondering if in the documentation:

'sum'
    Activation is the sum across each sensor weighted by the fractional distance from each sensor. A sensor with zero distance gets weight 1 and a sensor at distance meters away (or larger) gets weight 0. If distance is less than the distance between any two electrodes, this will be the same as 'nearest'.

should be:

'sum'
    Activation is the sum across each sensor weighted by the fractional distance from each sensor. A sensor with zero distance gets weight 1 and a sensor at distance meters away (or larger) gets weight 0. If distance is less than **half** the distance between any two electrodes, this will be the same as **'single'**.

chapochn:
#### Reference issue
Fixes #9993

Rectified mistake in the description of the 'sum' option in stc_near_sensors
changed "electrode" to "sensor"

### DIFF
--- a/mne/source_estimate.py
+++ b/mne/source_estimate.py
@@ -3295,8 +3295,8 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         Activation is the sum across each sensor weighted by the fractional
         ``distance`` from each sensor. A sensor with zero distance gets weight
         1 and a sensor at ``distance`` meters away (or larger) gets weight 0.
-        If ``distance`` is less than the distance between any two electrodes,
-        this will be the same as ``'nearest'``.
+        If ``distance`` is less than half the distance between any two 
+        electrodes, this will be the same as ``'single'``.
     - ``'single'``
         Same as ``'sum'`` except that only the nearest electrode is used,
         rather than summing across electrodes within the ``distance`` radius.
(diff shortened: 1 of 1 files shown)


mne/source_estimate.py
```diff
@@ -3295,8 +3295,8 @@ def stc_near_sensors(evoked, trans, subject, distance=0.01, mode='sum',
         Activation is the sum across each sensor weighted by the fractional
         ``distance`` from each sensor. A sensor with zero distance gets weight
         1 and a sensor at ``distance`` meters away (or larger) gets weight 0.
-        If ``distance`` is less than the distance between any two electrodes,
-        this will be the same as ``'nearest'``.
+        If ``distance`` is less than half the distance between any two
+        electrodes, this will be the same as ``'single'``.
```
rob-luke, line 3299: This function also works with optodes. So change the word electrodes to sensors?

chapochn: done!

drammock (approved):

larsoner:
Thanks @chapochn !

PR #10009, opened 2021-11-16
MRG, MAINT: Update testing for datasets

larsoner:
A few fixes:

1. We errantly started to override the `fname` argument in [#9742](https://github.com/mne-tools/mne-python/pull/9742/files#diff-f5e016349dc0d44312eaaff21e34539015b5355b621f200cb99910a56dcb2820R95), which manifested as `SC-subjects.xls` and `ST-subjects.xls` appearing in whatever directory the tests were being run from, which was annoying because `git status` would show them as untracked files.
2. Fixes some path problems on Sphinx 5 using `:doc:`, matplotlib ones fixed by prepending `matplotlib:` and the MNE ones (which should probably work but don't, unclear what the problem is) just worked around using `:ref:` which points to the same place effectively.
3. Fixes our usage of autouse figures. I don't think the `plt.close('all')` one was actually being used until now, so this could perhaps help with memory usage and/or speed (or it might not matter, but it's still cleaner).

### DIFF
--- a/doc/install/contributing.rst
+++ b/doc/install/contributing.rst
@@ -441,7 +441,7 @@ not require a deprecation cycle.
 
 Note that any new API elements should be added to the main reference;
 classes, functions, methods, and attributes cannot be cross-referenced unless
-they are included in the :doc:`python_reference`
+they are included in the :ref:`api_reference`
 (:file:`doc/python_reference.rst`).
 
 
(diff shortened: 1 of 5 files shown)


drammock (approved):

agramfort (approved):

larsoner:
Failure appears to be due to https://www.statsmodels.org/ being down, so I'll go ahead and merge

drammock:
@larsoner I think we'll need to copy this `plt.close('all')` fix in mne_connectivity at least...  do you know off the top of your head where else it gets used?  in the pqtg browser perhaps?  in realtime?

larsoner:
Those would be the two repos where I'd expect it to be useful

PR #10013, opened 2021-11-17
update pymatreader to version 0.0.26

thht:
this is the version of the pymatreader changes in #10005 as implemented in upstream pymatreader. the changes are only minor so i do not expect any problems.

thanks again @larsoner for proposing the fixes!


larsoner:
Thanks @thht !

PR #10015, opened 2021-11-17
WIP,ENH: new coreg features

GuillaumeFavelier:
This PR adds support for the parameters `scale_by_distance`, `project_eeg`, `mark_inside` of `CoregistrationUI`.

It's part of https://github.com/mne-tools/mne-python/issues/8833

### DIFF
--- a/mne/gui/_coreg.py
+++ b/mne/gui/_coreg.py
@@ -85,6 +85,8 @@ class CoregistrationUI(HasTraits):
     _current_fiducial = Unicode()
     _info_file = Unicode()
     _orient_glyphs = Bool()
+    _scale_by_distance = Bool()
+    _project_eeg = Bool()
     _hpi_coils = Bool()
     _head_shape_points = Bool()
     _eeg_channels = Bool()
@@ -98,9 +100,9 @@ class CoregistrationUI(HasTraits):
                  fiducials='auto', head_resolution=None,
                  head_transparency=None, hpi_coils=None,
                  head_shape_points=None, eeg_channels=None, orient_glyphs=None,
-                 sensor_opacity=None, trans=None, size=None, bgcolor=None,
-                 show=True, standalone=False, interaction='terrain',
-                 verbose=None):
+                 scale_by_distance=None, project_eeg=None, sensor_opacity=None,
+                 trans=None, size=None, bgcolor=None, show=True,
+                 standalone=False, interaction='terrain', verbose=None):
         from ..viz.backends.renderer import _get_renderer
 
         def _get_default(var, val):
@@ -123,6 +125,8 @@ class CoregistrationUI(HasTraits):
             size=_get_default(size, (800, 600)),
             bgcolor=_get_default(bgcolor, "grey"),
             orient_glyphs=_get_default(orient_glyphs, True),
+            scale_by_distance=_get_default(scale_by_distance, True),
+            project_eeg=_get_default(project_eeg, True),
             hpi_coils=_get_default(hpi_coils, True),
             head_shape_points=_get_default(head_shape_points, True),
             eeg_channels=_get_default(eeg_channels, True),
@@ -177,6 +181,8 @@ class CoregistrationUI(HasTraits):
         self._set_subject(subject)
         self._set_info_file(info_file)
         self._set_orient_glyphs(self._defaults["orient_glyphs"])
+        self._set_scale_by_distance(self._defaults["scale_by_distance"])
+        self._set_project_eeg(self._defaults["project_eeg"])
         self._set_hpi_coils(self._defaults["hpi_coils"])
         self._set_head_shape_points(self._defaults["head_shape_points"])
         self._set_eeg_channels(self._defaults["eeg_channels"])
@@ -260,6 +266,12 @@ class CoregistrationUI(HasTraits):
     def _set_orient_glyphs(self, state):
         self._orient_glyphs = bool(state)
 
+    def _set_scale_by_distance(self, state):
+        self._scale_by_distance = bool(state)
+
+    def _set_project_eeg(self, state):
+        self._project_eeg = bool(state)
+
     def _set_hpi_coils(self, state):
         self._hpi_coils = bool(state)
 
@@ -340,8 +352,8 @@ class CoregistrationUI(HasTraits):
 
     @observe("_lock_fids")
     def _lock_fids_changed(self, change=None):
-        view_widgets = ["orient_glyphs", "show_hpi", "show_hsp",
-                        "show_eeg", "high_res_head"]
+        view_widgets = ["orient_glyphs", "scale_by_distance", "project_eeg",
+                        "show_hpi", "show_hsp", "show_eeg", "high_res_head"]
         fid_widgets = ["fid_X", "fid_Y", "fid_Z", "fids_file", "fids"]
         if self._lock_fids:
             self._forward_widget_command(view_widgets, "set_enabled", True)
@@ -389,6 +401,14 @@ class CoregistrationUI(HasTraits):
     @observe("_orient_glyphs")
     def _orient_glyphs_changed(self, change=None):
         self._update_plot(["hpi", "hsp", "eeg"])
+
+    @observe("_scale_by_distance")
+    def _scale_by_distance_changed(self, change=None):
+        self._update_plot(["hpi", "hsp", "eeg"])
+
+    @observe("_project_eeg")
+    def _project_eeg_changed(self, change=None):
+        self._update_plot("eeg")
 
     @observe("_hpi_coils")
     def _hpi_coils_changed(self, change=None):
@@ -629,7 +649,9 @@ class CoregistrationUI(HasTraits):
             hpi_actors = _plot_hpi_coils(
                 self._renderer, self._info, self._to_cf_t,
                 opacity=self._defaults["sensor_opacity"],
-                orient_glyphs=self._orient_glyphs, surf=self._head_geo)
+                orient_glyphs=self._orient_glyphs,
+                scale_by_distance=self._scale_by_distance,
+                surf=self._head_geo)
         else:
             hpi_actors = None
         self._update_actor("hpi_coils", hpi_actors)
@@ -639,7 +661,9 @@ class CoregistrationUI(HasTraits):
             hsp_actors = _plot_head_shape_points(
                 self._renderer, self._info, self._to_cf_t,
                 opacity=self._defaults["sensor_opacity"],
-                orient_glyphs=self._orient_glyphs, surf=self._head_geo,
+                orient_glyphs=self._orient_glyphs,
+                scale_by_distance=self._scale_by_distance,
+                surf=self._head_geo,
                 mask=self._coreg._extra_points_filter)
         else:
             hsp_actors = None
@@ -655,7 +679,10 @@ class CoregistrationUI(HasTraits):
                     meg=False, eeg=eeg, fnirs=False, warn_meg=False,
                     head_surf=self._head_geo, units='m',
                     sensor_opacity=self._defaults["sensor_opacity"],
-                    orient_glyphs=self._orient_glyphs, surf=self._head_geo)
+                    orient_glyphs=self._orient_glyphs,
+                    scale_by_distance=self._scale_by_distance,
+                    project_points=self._project_eeg,
+                    surf=self._head_geo)
                 eeg_actors = eeg_actors["eeg"]
             else:
                 eeg_actors = None
@@ -841,6 +868,19 @@ class CoregistrationUI(HasTraits):
             name="Orient glyphs",
             value=self._orient_glyphs,
             callback=self._set_orient_glyphs,
+            layout=layout
+        )
+        self._widgets["scale_by_distance"] = \
+            self._renderer._dock_add_check_box(
+            name="Scale glyphs",
+            value=self._scale_by_distance,
+            callback=self._set_scale_by_distance,
+            layout=layout
+        )
+        self._widgets["project_eeg"] = self._renderer._dock_add_check_box(
+            name="Project EEG",
+            value=self._project_eeg,
+            callback=self._set_project_eeg,
             layout=layout
         )
         self._widgets["show_hpi"] = self._renderer._dock_add_check_box(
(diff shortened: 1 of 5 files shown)


agramfort:
tell us when it's ready for review @GuillaumeFavelier

GuillaumeFavelier:
I double-checked some examples and fixed a visual bug introduced in here. This PR is now ready for reviews @agramfort, @larsoner

larsoner:
I ran:
```
mne coreg -s sample -d subjects --fif MEG/sample/sample_audvis_raw.fif
```
clicked "lock" (to get out of fiducials mode), then hit "fit fiducials", then turned on transparent mode, then recorded this, which is just a few clusters of rapid Z translation clicks:

https://user-images.githubusercontent.com/2365790/142451289-5c962555-2b83-4f1d-913b-e960457d19eb.mp4

1. After I stop clicking the Z-translate button, the points jump downward (probably correct?) but then "slide upward" incrementally (probably incorrect?). Note that this does *not* happen to HPI coils but it *does* to EEG and dig points, if that helps track down the bug.
2. The HPI-coil-size is seems wrong for the coil on the far side of the head. By the end of the video, it's basically on the head surface, but it has a huge radius

mne/gui/_coreg.py
```diff
@@ -56,6 +56,15 @@ class CoregistrationUI(HasTraits):
         If True, display the EEG channels. Defaults to True.
     orient_glyphs : bool
         If True, orient the sensors towards the head surface. Default to False.
+    scale_by_distance : bool
+        If True, scale the sensors based on their distance to the head surface.
+        Defaults to True.
+    project_eeg : bool
+        If True, project the EEG channels onto the head surface. Defaults to
+        True.
```
larsoner, line 64: It should probably default to False. They are actually useful for coreg. Projection to surface is mostly for understanding how they will appear during forward modeling, so it's more like extra information if people are curious rather than being super useful during the coreg process itself.

GuillaumeFavelier:
> just a few clusters of rapid Z translation clicks

I added a lock on `_set_parameter()` which should prevent those "out-of-sync" scenarii caused by spamming.

GuillaumeFavelier (commented):
https://github.com/mne-tools/mne-python/pull/10015/files#diff-b739315925255b273252922a00cd620817da5cf9fa20bd7322fa798928e78a19L201

Is this really necessary? Because it is not compatible with the parameter `head_resolution` (for example when the user sets `--low-res-head`).

larsoner:
> merging orient_glyphs and scale_by_distance into one checkbox

From talking to @agramfort a bit, I'm tempted to just *always* scale by distance and orient the glyphs, and not make these optional.

> keeping mark_inside=True and removing its checkbox

Agreed

> setting head_transparency depending on "lock fiducials" and removing its checkbox

Sometimes it's nice to be able to make it transparent, and even to varying degrees. But I guess we could add support for this later, and stick with what you propose for now

> Also is it necessary to have all "Show HPI", "Show EEG" and "Show HSP", maybe only one "Show sensors" could work?

Being able to toggle these separately can be nice. What if we show/hide based on the weight they are assigned for ICP? Zero weight = hide, non-zero weight = show?

larsoner:
> Is this really necessary? Because it is not compatible with the parameter head_resolution (for example when the user sets --low-res-head).

You need to use the high resolution head when picking fiducials, otherwise your clicks get locked to vertex locations that are some mm away from where you actually clicked. Also it's much harder to make out the correct fid locations when using the low resolution head.

In other words, I think if you are in "set fiducials" mode, you should always be shown an opaque, high-res head with no sensors. Once you leave that mode, you should switch to a transparent, high- or low- (depending on the option / config value they have set) res head with whatever sensors they've chosen to show

GuillaumeFavelier:
We discovered a bug with glyphs and `mark_inside` so to have a consistent behaviour, this PR requires `pyvista>=0.32`

The head shape points inside are not marked and are not scaled by distance:

bug | expected
--|--
image | image

GuillaumeFavelier:

agramfort (approved):
works great for me !

thx @GuillaumeFavelier 

@larsoner @drammock I let you test and merge.

larsoner (approved):
Works great!

As one last nitpick, can we make the HPI size the same as EEG and dig points? I don't really see a reason they should be different. After that, all point sizes mean the same distance away from the head, which is nice!

GuillaumeFavelier:
I'm on it

larsoner:
Actually it looks like the de-spamming isn't working -- the last/latest position isn't always shown properly. If I spam the Z translate to make the dig points move down, then stop, then press it once in the other direction to make the dig points move up, what you see is a downward jump first, suggesting that the last / newest position was ignored (not rendered), then on the next press you see the dig points move upward as they should:

Untitled

GuillaumeFavelier:
With the current implementation (simple locking), it's totally possible that the last request is locked/denied. I suppose I can work on a more complicated system.

What I have in mind at the moment is a queue to store all the `_set_parameter()` requests then "observe" the `_sync_lock` to wait for the good moment to `_update_plot()` :thinking: 

Feel free to share if you have any suggestions @larsoner

larsoner:
> What I have in mind at the moment is a queue to store all the _set_parameter() requests then "observe" the _sync_lock to wait for the good moment to _update_plot() 🤔

Maybe something with timers? A single interactive-click-redraw timer where any interactive click resets this timer sets it to update the display 1/60th of a second later might work. Not sure...

Another maybe simpler option would be just to set an `_redraw_pending = True`, and have a timer that runs every 1/60th of a second check `if self._redraw_pending:` and sets this to `False` then draws if necessary. That might be even cleaner. That way no matter what, the "final state" will always be shown, but we also don't try to draw everything (e.g., if it takes 500 ms to actually do a draw, I might have clicked three times in that interval, but the ending state / after the third click will eventually be drawn).

GuillaumeFavelier:
d723cf4ab664d30fbc47ab6602c865bf0511e216 is an attempt at a real lock: `threading.Lock`. That way, no request is denied but the time to draw everything increases with the number of requests. It can be mitigated if `_update_plot` is optimized (not at the moment).

GuillaumeFavelier:
> Maybe something with timers?

Good idea, I think we have something similar with `Brain`. I will it next and check that it is compatible with `ipywidgets` too.

GuillaumeFavelier:
0bac9b47f8a49d1ed096f484a09ad9cb2947f523 decreases the number of calls to `_update_plot()` and it does not hang for so long.

After the latest commit (add6a6e7924a58f842e4e99532dfc882d93251c4), it uses the lock for `_set_parameter` and the timer for `_update_plot`, what do you think @larsoner ?

larsoner:
> what do you think @larsoner ?

Did you test locally? At least for me on macOS the behavior is not fixed -- the last position isn't always shown (sometimes it is, sometimes it isn't):

https://user-images.githubusercontent.com/2365790/143779327-655b6458-f36c-4de0-96c3-4d6e4fc15144.mp4

GuillaumeFavelier:
To summarize my ideas:
- [x] using a lock and queue to serialize the requests
- [x] delaying the plot updates after the requests using `traits`
- [ ] reducing the number of requests: maybe with a Qt trick?
- [ ] switching the combo box widgets to label widgets: this prevent spamming and could gain me the time to optimize the viz

Feel free to push if you want to try something @larsoner

larsoner:
Sure, I can look today or early next week

larsoner:
1. Fixed the drawing (I think)
2. Last update always gets drawn by the timer. You can verify interactively, but also by watching e.g.:
    ```
    MNE_LOGGING_LEVEL=debug mne coreg -s sample -d subjects/ --fif MEG/sample/sample_audvis_raw.fif
    ```
3. Fixed "jerkiness" by removing a `_process_events` call in `_add_mesh`. We really shouldn't need it there. If we do, we should fix it elsewhere. I think this should speed things up globally, because now we can truly manually control when updates are made. @GuillaumeFavelier if you remember where this came from, we should double check that removing it doesn't break anything!t
4. Fixed CTRL-C behavior (now properly closes rather than aborting) both for `mne coreg` and for PyQtGraph-based browsing
5. Fixed iterative ICP updates so they show up on the fly now. I can't upload a video at the moment, but feel free to try locally! As usual, this is my command (but this time with debug prefix so I can see when it does draws):

    ```
    $ MNE_LOGGING_LEVEL=debug mne coreg -s sample -d subjects/ --fif MEG/sample/sample_audvis_raw.fif
    ```

mne/viz/backends/_utils.py
```diff
@@ -149,3 +150,18 @@ def _init_mne_qtapp(enable_icon=True, pg_app=False):
...
+    old_signal = signal.getsignal(signal.SIGINT)
+    is_python_signal_handler = old_signal is not None
+    if is_python_signal_handler:
+        signal.signal(signal.SIGINT, signal.SIG_DFL)
+    try:
+        app.exec_()
+    finally:
+        # reset the SIGINT exception handler
+        if is_python_signal_handler:
+            signal.signal(signal.SIGINT, old_signal)
```
larsoner, line 167: cc @marsipu you'll probably be interested in this -- it makes `ctrl-C` actually interrupt blocking mode. For me it actually kills the Python session, but matplotlib does the same. So I think it's okay (?)

mne/viz/utils.py
```diff
@@ -129,12 +129,13 @@ def _show_browser(show=True, block=True, fig=None, **kwargs):
...
     else:
         from PyQt5.QtWidgets import QApplication
+        from .backends._utils import _qt_app_exec
         if show:
             fig.show()
         # If block=False, a Qt-Event-Loop has to be started
         # somewhere else in the calling code.
         if block:
-            QApplication.instance().exec()
+            _qt_app_exec(QApplication.instance())
```
larsoner, line 138: ... and here is where I used that private function it so that it would affect `pyqtgraph`-based browsing

larsoner:
Pushed another couple of commits to:

1. Temporarily use my PyVista version branch until they change their policy (https://github.com/pyvista/pyvista/pull/1911)
2. Make `hpi` the same size as dig and EEG
3. Default to "matched" fid mode

@GuillaumeFavelier can you look and see if my changes and the code behavior make sense? Basically every `_update_plots` call just updates the set of things to draw, and every tick of the time (1/60th of a sec max) `_redraw` is called, which pops things from the list of things to draw. This ensures that whatever our last state is, eventually all items from it get drawn. In practice it seems that they're always drawn at once, which is nice. (Before they were drawn at the same time, but the _process_events in _add_mesh caused the draw to update, so if you updated say 5 items it would evoke 5 redraws rather than a single one at the end.)

GuillaumeFavelier:
Thank you for the great help @larsoner!

> can you look and see if my changes and the code behavior make sense?

I took the time to review your commits carefully and I tested locally with `_qt` and `_notebook`. I'm glad that 2 mutexes and a set did the trick :ok_hand: 

> Fixed CTRL-C behavior 

This is awesome!

> Fixed iterative ICP updates so they show up on the fly now

I was not aware of this issue, for me, the plot was updated after each iteration :open_mouth: 

> Fixed "jerkiness" by removing a _process_events call in _add_mesh. 

Same for the jerkiness but it was maybe more difficult for me to notice.

> Before they were drawn at the same time, but the _process_events in _add_mesh caused the draw to update, so if you updated say 5 items it would evoke 5 redraws rather than a single one at the end.

Anyway, thank you for the explanation, I'll keep this in mind.

> We really shouldn't need it there.

I have been digging to find where the call comes from and I tracked it down to https://github.com/mne-tools/mne-python/pull/7791#issuecomment-648127345. Apparently, it was necessary during the migration to `pyvistaqt` on MacOS with `vtk < 8.2`.

To be honest, we probably do not need it anymore.

agramfort:
trying to unselect "show high resolution head" on my mac produces a segfault. Can someelse replicate? Just untick the box just after starting the GUI.

❯ mne coreg -s sample -d ~/mne_data/MNE-sample-data/subjects --interaction=trackball -f ~/mne_data/MNE-sample-data/MEG/sample/sample_audvis_raw.fif
python3.8(72101,0x10aa1cdc0) malloc: Incorrect checksum for freed object 0x7fef1ba26a00: probably modified after being freed.
Corrupt value: 0x101010101010101
python3.8(72101,0x10aa1cdc0) malloc: *** set a breakpoint in malloc_error_break to debug
Fatal Python error: Aborted

Current thread 0x000000010aa1cdc0 (most recent call first):
  File "/Users/alex/miniconda3/lib/python3.8/site-packages/vtkmodules/qt/QVTKRenderWindowInteractor.py", line 381 in paintEvent
  File "/Users/alex/work/src/mne-python/mne/viz/backends/_utils.py", line 163 in _qt_app_exec
  File "/Users/alex/work/src/mne-python/mne/gui/_coreg.py", line 249 in __init__
  File "<decorator-gen-547>", line 24 in __init__
  File "/Users/alex/work/src/mne-python/mne/gui/__init__.py", line 192 in coregistration
  File "<decorator-gen-535>", line 24 in coregistration
  File "/Users/alex/work/src/mne-python/mne/commands/mne_coreg.py", line 89 in run
  File "/Users/alex/work/src/mne-python/mne/commands/utils.py", line 107 in main
  File "/Users/alex/miniconda3/bin/mne", line 33 in <module>
[1]    72101 abort      mne coreg -s sample -d ~/mne_data/MNE-sample-data/subjects  -f

GuillaumeFavelier:
> trying to unselect "show high resolution head" on my mac produces a segfault.

I have the same behaviour

agramfort:
you're on it?

>

GuillaumeFavelier:
yes, trying to see what is going on suddenly with this head :)

GuillaumeFavelier:
The crash seems related to the `_grow_hair_changed()` call in `_hair_resolution_changed()` so I pushed a few commits to rework `grow_hair` and obtain something compatible with the new redraw system. What do you think @larsoner, @agramfort  ?

larsoner:
> I pushed a few commits to rework grow_hair and obtain something compatible with the new redraw system. What do you think @larsoner, @agramfort ?

When I saw your comment about it being related to the hair growth stuff, my thought was "we should separate the update of the head geometry variable from the update of the PyVista plot", and it looks like indeed that's what you did, so I'll go ahead and fix the conflict and merge hopefully. I also noticed in @hoechenberger's PR some pyvista version comparisons we no longer need I think.

larsoner:
Thanks @GuillaumeFavelier !

PR #10018, opened 2021-11-18
MRG: Fix warning in KIT reader

mscheltienne:
The issue was raised on the discourse forum [here](https://mne.discourse.group/t/unable-to-load-data-with-mne-io-read-raw-kit/3983/2). This line is not covered by test cases.
As far as I can tell with this simple regex match `warn\(['"].*['"],.*\)`, it's the only place where this syntax for `logger.warning` was used.

### DIFF
--- a/mne/io/kit/kit.py
+++ b/mne/io/kit/kit.py
@@ -513,7 +513,7 @@ def get_kit_info(rawfile, allow_unknown_format, standardize_names=None,
             version_string = "V%iR%03i" % (version, revision)
             if allow_unknown_format:
                 unsupported_format = True
-                warn("Force loading KIT format %s", version_string)
+                warn("Force loading KIT format %s" % version_string)
             else:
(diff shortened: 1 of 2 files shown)


larsoner:
> This line is not covered by test cases.

FTFY by commit :)

larsoner (approved):
Will merge once CIs come back happy, thanks @mscheltienne

mscheltienne:
I gotta say, it feels a bit overkill for such a simple syntax typo.. but thanks!

larsoner:
> I gotta say, it feels a bit overkill for such a simple syntax typo.. but thanks!

Tell that to the user who wouldn't have encountered the error and spent time on it if we had this check in place before :)

mscheltienne:
Fair enough ;)

agramfort:
shall we backport?

>

larsoner:
Done

PR #10019, opened 2021-11-19
[MRG, ENH] Enable IO for saving/reading Epochs with Annotations

adam2392:
#### Reference issue
Closes: #8376 

Follow up to #9969 but requires that one to be merged first.

Enables IO for Epochs with Annotations:
- adds 1 FIFF constant for Raw_Sfreq. The First_time and Annotations constants already exist
- Adds saving logic to `def _save_part` in `epochs.py`
- [ ] Add logic for `def read_epochs`

#### Additional information
In order to add the IO portion, we need to add the Annotation functions and logic to `EpochsFIFF` object. Would the best way to do this is change all the Epoch annotation handling functions to be part of a `EpochAnnotationMixin` class that EpochsRaw and EpochsFIFF both use?


adam2392:
A few questions that will help me design a solution:

1. When concatenating Epochs, is `info['sfreq']` supposed to be the same?
2. If we want a variable for `_raw_sfreq` in each `Epochs*` class, should I just add it to `BaseEpochs` to make this simpler? Then each class passes in the correct one for functionality (i.e. `EpochsArray` passes in `info['sfreq']`, while everyone else passes in `raw.info['sfreq']`)

@larsoner

larsoner:
> When concatenating Epochs, is info['sfreq'] supposed to be the same?

Yes I'm pretty sure it's required

> If we want a variable for _raw_sfreq in each Epochs* class, should I just add it to BaseEpochs to make this simpler?

Sure. Agreed about how to handle subclasses. And to concat, these need to match

adam2392:
> Sure. Agreed about how to handle subclasses. And to concat, these need to match

Okay, so I've handled all the issues we ran into above by extending the base class `BaseEpochs`. However, there's an edge case file. Specifically `EpochsKit` stores an Epoch Array. I'm going to try handling this with an if/else statement in the class to check for the fields we want (i.e. Annotations and Raw_Sfreq).

Reference error message:
```
FAILED mne/io/kit/tests/test_kit.py::test_epochs - TypeError: events should b
```

mne/epochs.py
```diff
@@ -3146,6 +3173,19 @@ def _read_one_epoch_file(f, tree, preload):
...
+
+        raw_sfreq_node = dir_tree_find(tree, FIFF.FIFF_MNE_EPOCHS_RAW_SFREQ)
+        if len(raw_sfreq_node) > 0:
+            # before version 0.11 we errantly saved with this tag instead of
+            # an MNE tag
+            epochs_node = dir_tree_find(tree, FIFF.FIFFB_MNE_EPOCHS)
+            if len(epochs_node) == 0:
+                epochs_node = dir_tree_find(tree, 122)  # 122 used before v0.11
+                if len(epochs_node) == 0:
+                    raise ValueError('Could not find epochs data')
```
adam2392, line 3187: This is scruff from copy/paste, but figuring out how to get this working...

mne/epochs.py
```diff
@@ -3682,16 +3729,16 @@ def _concatenate_epochs(epochs_list, with_data=True, add_offset=True, *,
                     (offsets[-1], len(out.ch_names), len(out.times)),
                     dtype=this_data.dtype)
             data[start:stop] = this_data
-    return (info, data, events, event_id, tmin, tmax, metadata, baseline,
-            selection, drop_log, verbose)
+    return (info, data, raw_sfreq, events, event_id, tmin, tmax, metadata,
+            baseline, selection, drop_log, verbose)
```
drammock, line 3733: when you concatenate epochs that have annotations, not only does the resulting epochs object lose those annotations, *it doesn't even have an `.annotations` property anymore*.  I think we should set it to `None` rather than having it missing entirely.  Not sure if that should happen here or in `_finish_concat`, do what you think makes sense.

mne/tests/test_epochs.py
```diff
@@ -3878,5 +3878,63 @@ def test_epoch_annotations_cases():
     assert 'multiple' in first_epoch_ant
 
     # test that concatenation does not preserve epochs
```
drammock, line 3880:
```suggestion
    # test that concatenation does not preserve annotations
```

mne/tests/test_epochs.py
```diff
@@ -3878,5 +3878,63 @@ def test_epoch_annotations_cases():
...
+    old_epochs = epochs.copy()
     with pytest.warns(RuntimeWarning, match='Annotations'):
         concatenate_epochs([epochs, epochs_one])
+
+    # concatenation should not change the Epochs annotations
```
drammock, line 3885:
```suggestion
    # concatenation should not change the *input* Epochs' annotations
```

mne/tests/test_epochs.py
```diff
@@ -3878,5 +3878,63 @@ def test_epoch_annotations_cases():
...
+        onset=[0.4, 0.5, 1.4, 1.6, 3.0, 5.4],
+        duration=[ant_dur, ant_dur, ant_dur, 0, 2.0, 0],
+        description=['before', 'start', 'stop',
+                     'outside', 'multiple', 'bad_noisy'],
+    )
+    raw.set_annotations(ants)
+
+    events = np.zeros((3, 3), dtype=int)
+    events[:, 0] = [100, 300, 500]
+    epochs = Epochs(raw, events=events, tmin=-0.5, tmax=0.5)
```
drammock, line 3907: don't copy-paste this from the other test.  Either set up a (private) helper function that both tests call, or create a pytest fixture and pass it in as a test argument.

mne/tests/test_epochs.py
```diff
@@ -3878,5 +3878,63 @@ def test_epoch_annotations_cases():
...
+    assert_array_equal(
+        epochs.annotations.description,
+        loaded_epochs.annotations.description)
+
+    # if we set up EpochsArray and save it, it should have raw_sfreq
+    # and annotations even without explicit support
+    epoch_size = (3,) + data.shape
+    data = rng.random(epoch_size)
+    epochs = EpochsArray(data, info)
+    assert epochs._raw_sfreq is not None
```
drammock, line 3934:
```suggestion
    assert epochs._raw_sfreq == info['sfreq']
```

drammock (changes requested):

mne/epochs.py
adam2392: Okay, I've added a `BaseEpochs.set_annotations(None)` at the very end of BaseEpochs instantiation. That way, all `Epochs*` classes default to not having Annotations, unless they are instantiated outside of the `BaseEpochs.__init__()` call.

mne/epochs.py
```diff
@@ -142,6 +143,18 @@ def _save_part(fid, epochs, fmt, n_parts, next_fname, next_idx):
         elif fmt == 'double':
             write_function = write_double_matrix
 
+    #
+    # Epoch annotations
+    # Note that since EpochsArray does not support Annotations
+    # then we need to check if Annotations is a part of the API first.
+    #
```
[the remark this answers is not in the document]
adam2392: Done.

mne/tests/test_epochs.py
adam2392: Done.

drammock:
I just ran some analysis code while accidentally being on this branch instead of main.  I hit this:

```
UnboundLocalError                         Traceback (most recent call last)
~/Documents/academics/research/ilabs/ftof/connectivity/analysis/compute-connectivity.py in <module>
     68             slug = f'{subj}-{n_sec}sec'
     69             epo_fname = f'{subj}-{freq_band}-band-filtered-{n_sec}sec-epo.fif'
---> 70             epochs = mne.read_epochs(os.path.join(epo_dir, epo_fname))
     71             # do hilbert in sensor space (faster)
     72             epochs.apply_hilbert()

<decorator-gen-260> in read_epochs(fname, proj, preload, verbose)

/opt/mne/python/mne/epochs.py in read_epochs(fname, proj, preload, verbose)
   3180         The epochs.
   3181     """
-> 3182     return EpochsFIF(fname, proj, preload, verbose)
   3183 
   3184 

<decorator-gen-261> in __init__(self, fname, proj, preload, verbose)

/opt/mne/python/mne/epochs.py in __init__(self, fname, proj, preload, verbose)
   3244              baseline, selection, drop_log, epoch_shape, cals,
   3245              reject_params, fmt, annotations, raw_sfreq) = \
-> 3246                 _read_one_epoch_file(fid, tree, preload)
   3247 
   3248             if (events[:, 0] < 0).any():

/opt/mne/python/mne/epochs.py in _read_one_epoch_file(f, tree, preload)
   3159     return (info, data, data_tag, events, event_id, metadata, tmin, tmax,
   3160             baseline, selection, drop_log, epoch_shape, cals, reject_params,
-> 3161             fmt, annotations, raw_sfreq)
   3162 
   3163 

UnboundLocalError: local variable 'raw_sfreq' referenced before assignment
```

adam2392:
> I just ran some analysis code while accidentally being on this branch instead of main. I hit this:
> 
> ```
> ---------------------------------------------------------------------------
> UnboundLocalError                         Traceback (most recent call last)
> ~/Documents/academics/research/ilabs/ftof/connectivity/analysis/compute-connectivity.py in <module>
>      68             slug = f'{subj}-{n_sec}sec'
>      69             epo_fname = f'{subj}-{freq_band}-band-filtered-{n_sec}sec-epo.fif'
> ---> 70             epochs = mne.read_epochs(os.path.join(epo_dir, epo_fname))
>      71             # do hilbert in sensor space (faster)
>      72             epochs.apply_hilbert()
> 
> <decorator-gen-260> in read_epochs(fname, proj, preload, verbose)
> 
> /opt/mne/python/mne/epochs.py in read_epochs(fname, proj, preload, verbose)
>    3180         The epochs.
>    3181     """
> -> 3182     return EpochsFIF(fname, proj, preload, verbose)
>    3183 
>    3184 
> 
> <decorator-gen-261> in __init__(self, fname, proj, preload, verbose)
> 
> /opt/mne/python/mne/epochs.py in __init__(self, fname, proj, preload, verbose)
>    3244              baseline, selection, drop_log, epoch_shape, cals,
>    3245              reject_params, fmt, annotations, raw_sfreq) = \
> -> 3246                 _read_one_epoch_file(fid, tree, preload)
>    3247 
>    3248             if (events[:, 0] < 0).any():
> 
> /opt/mne/python/mne/epochs.py in _read_one_epoch_file(f, tree, preload)
>    3159     return (info, data, data_tag, events, event_id, metadata, tmin, tmax,
>    3160             baseline, selection, drop_log, epoch_shape, cals, reject_params,
> -> 3161             fmt, annotations, raw_sfreq)
>    3162 
>    3163 
> 
> UnboundLocalError: local variable 'raw_sfreq' referenced before assignment
> ```

Thanks good catch! I added a test for backwards compatability and also a fix to this issue.

mne/epochs.py
```diff
@@ -342,6 +352,8 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
     data : ndarray | None
         If ``None``, data will be read from the Raw object. If ndarray, must be
         of shape (n_epochs, n_channels, n_times).
+    raw_sfreq : float
+        The original Raw object sampling rate.
```
larsoner, line 356: Please add this last after a `*`, like `*, raw_sfreq=None`. If None, it should just use `info['sfreq']`. This will have better backward compatibility, e.g., prevent breaking `RtEpochs` in MNE-Realtime. This will also make it so you can revert a number of the test changes here, and maybe even some of the other BaseEpochs subclass changes

mne/annotations.py
```diff
@@ -647,6 +647,182 @@ def rename(self, mapping, verbose=None):
...
+            Annotations to set.
+
+        Returns
+        -------
+        self : instance of Raw
+            The raw object with annotations.
+        """
+        if annotations is not None:
+            raise RuntimeError('Setting Annotations within Epochs directly '
+                               'is not supported yet.')
```
larsoner, line 674: Why require this? I don't see why we need to restrict this. For example, if I have a raw, create epochs, go back to that raw, add some annotations, I should be able to do `epochs.set_annotations(raw.annotations)` and it should just work even with the code we have already, right?

adam2392: This was a quick fix for the tutorial bug found last week. 

For this function yes we could enable it to just receive Annotations. However, based on raw_sfreq being set correctly, this could be misconstrued? E.g. if someone sets up a epochsKit file that was resampled before Annotation support in Epochs. 

Should I just document a warning in the doc string or should I explicitly have a warning if they are one of the classes that I hard code raw_sfreq?

adam2392: I removed this error and added some notes to the doc string.

mne/epochs.py
adam2392: Done.

mne/io/kit/kit.py
```diff
@@ -417,6 +417,7 @@ def __init__(self, input_fname, events, event_id=None, tmin=0,
                               self.info['nchan'],
                               self._raw_extras[0]['frame_length'])
         tmax = ((data.shape[2] - 1) / self.info['sfreq']) + tmin
+
```
adam2392, line 420:
```suggestion
```

mne/tests/test_epochs.py
```diff
@@ -62,6 +62,34 @@
...
+
+    # Create Epochs that are spaced apart with a start point every
+    # two seconds in the Raw data starting at 1 seconds. The Epochs
+    # will contain the Raw data in these second windows:
+    # - (0.5, 1.5)
+    # - (2.5, 3.5)
+    # - (4.5, 5.5)
+    events = np.zeros((3, 3), dtype=int)
+    events[:, 0] = [100, 300, 500]
+    epochs = Epochs(raw, events=events, tmin=-0.5, tmax=0.5)
```
drammock, line 89:
```suggestion
    # epoch onsets will be at 0.5, 2.5, 4.5s and will be one second long
    events = np.zeros((3, 3), dtype=int)
    events[:, 0] = (np.array([0.5, 2.5, 4.5]) * sfreq).astype(int)
    # make annotations to test various kinds of overlap
    #         onset  dur  descr
    annots = [(0.3, 0.0, 'no_overlap'),
              (0.4, 0.1, 'coincident_onset'),   # only edge coincides
              (0.4, 0.2, 'straddles_onset'),
              (1.4, 0.2, 'straddles_offset'),
              (1.5, 0.0, 'coincident_offset'),  # only edge coincides, zero-dur
              (2.6, 0.0, 'within_epoch'),
              (4.4, 1.2, 'surround_epoch'),
              (3.4, 1.2, 'multiple')]
    annots = Annotations(*zip(*annots))
    raw.set_annotations(annots)
    epochs = Epochs(raw, events=events, tmin=0, tmax=1, baseline=None)
```

This is a proposal for what I think is much easier-to-follow code. **IT WILL BREAK THE TESTS** because it adds 2 cases that weren't there before, and it changes the annotation descriptions (which means the corresponding assertions in the tests will fail).  I can't easily update the tests through the PR review UI, since those lines are outside the diff; if you like the direction this change goes in but don't have time to update the tests yourself LMK and I can make changes locally and push them.

mne/tests/test_epochs.py
```diff
@@ -3877,6 +3884,59 @@ def test_epoch_annotations_cases():
...
+    old_epochs = epochs.copy()
     with pytest.warns(RuntimeWarning, match='Annotations'):
         concatenate_epochs([epochs, epochs_one])
+
+    # concatenation should not change the *input* Epochs' annotations
+    assert epochs.annotations == old_epochs.annotations
+
+
+def test_epochs_annotations_backwards_compat():
+    """Test backwards compatibility with Epochs saved."""
```
drammock, line 3897:
```suggestion
    """Test backwards compatibility with Epochs saved without annotations."""
```

drammock (approved):
I've made various suggestions for improving code comments, docstrings etc, and one suggestion of a moderate overhaul for the tests (to make them a bit easier to follow), but in general this is looking pretty good to me.  IMO this should get another review before merging though.

mne/annotations.py
```diff
@@ -647,6 +647,186 @@ def rename(self, mapping, verbose=None):
...
+        -----
+        Epoch Annotations were added in v1.0 in order to preserve
+        events that occurred in raw data when they were Epoched. For example,
+        this might occur in time-frequency analysis, or sliding-window
+        connectivity analysis. Setting Annotations directly on Epochs
+        requires an accurate ``_raw_sfreq``, which stores the raw sampling
+        rate to interpret the Annotation.
+        """
+        self._set_annotations(annotations)
+        return self
```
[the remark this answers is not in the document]
adam2392: I believe the main reason I had a private method was to prevent users from calling `set_annotations`, but now that we have a public method and we sufficiently warn the user of adding annotations after resampling, then I think we can get rid of the private method.

mne/annotations.py
```diff
@@ -647,6 +647,186 @@ def rename(self, mapping, verbose=None):
...
+        """Add raw annotations into the Epochs metadata data frame.
+
+        Adds three columns to the ``metadata`` consisting of a list
+        in each row:
+        - ``Annotations_onset``: the onset of each Annotation within
+        the Epoch relative to the start time of the Epoch (in seconds).
+        - ``Annotations_duration``: the duration of each Annotation
+        within the Epoch in seconds.
+        - ``Annotations_description``: the description of each
+        Annotation.
```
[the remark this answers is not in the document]
adam2392: Sure that works for me. I kept it very verbose because I wonder if there would ever be a case that metadata already has things like `annot_onset`, etc.?

mne/annotations.py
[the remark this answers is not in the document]
adam2392: Removed private method.

mne/io/constants.py
```diff
@@ -1033,6 +1033,7 @@
 FIFF.FIFF_MNE_EPOCHS_SELECTION     = 3800  # the epochs selection
 FIFF.FIFF_MNE_EPOCHS_DROP_LOG      = 3801  # the drop log
 FIFF.FIFF_MNE_EPOCHS_REJECT_FLAT   = 3802  # rejection and flat params
+FIFF.FIFF_MNE_EPOCHS_RAW_SFREQ     = 3803  # original raw sfreq
```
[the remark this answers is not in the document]
adam2392: No it has not been added. @larsoner suggested it would take some time. Any idea how to proceed? Should I make a PR to the FIFF standard package?

adam2392:
> I've made various suggestions for improving code comments, docstrings etc, and one suggestion of a moderate overhaul for the tests (to make them a bit easier to follow), but in general this is looking pretty good to me. IMO this should get another review before merging though.

Kay this is good to go for me once CI is green. I addressed all your comments in the new updated commit.

mne/annotations.py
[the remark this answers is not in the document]
drammock: if that's a concern, then add something that checks whether those column names already exist, and either (1) error out, or (2) warn that different column names had to be used (like `annot_onset_1` or similar).  To me it seems quite unlikely so I'd be inclined to just error out on the assumption that users will (almost) never hit that corner case.

agramfort:
allow me to have a look after the first approval before merging

thx

larsoner:
Test failures look real and there is a conflict

Thanks for working through this @adam2392 !

mne/io/constants.py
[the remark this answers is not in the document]
larsoner: It won't actually take too long, I just wanted to make sure that it's really all we need to add. So let's get this PR to mergeable state, then make the fiff-constants PR. It should only take a day or two to get that in I think since this constant is in the range of constants that MNE "owns"

larsoner (changes requested):
Just marking a change request so we don't forget to update fiff-constants before (adjusting this PR and then) merging this

mne/io/constants.py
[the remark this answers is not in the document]
adam2392: > Just marking a change request so we don't forget to update fiff-constants before (adjusting this PR and then) merging this

Kay I have added a PR to upstream FIFF-Constants

mne/annotations.py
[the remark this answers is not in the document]
adam2392: I added this error check and also a `overwrite` parameter to handle the edge case that someone might want to overwrite these columns.

mne/io/tests/test_constants.py
```diff
@@ -36,6 +36,7 @@
                      'viewitems', 'viewkeys', 'viewvalues',  # Py2
                      )
 _tag_ignore_names = (
+    'FIFF_MNE_EPOCHS_RAW_SFREQ',
```
larsoner, line 39: You can now remove this and change the commit above to `aa49e20cff5791fbaf01d77ad4ec2e0ecb69840d`

mne/epochs.py
```diff
@@ -2965,6 +2812,9 @@ def __init__(self, data, info, events=None, tmin=0, event_id=None,
             self._detrend_offset_decim(e, detrend_picks)
         self.drop_bad()
 
+        # also create an empty Annotations object
+        self._annotations = None
```
larsoner, line 2816: This should be done by the `BaseEpochs` class I think

mne/epochs.py
```diff
@@ -3305,9 +3161,16 @@ def _read_one_epoch_file(f, tree, preload):
         if drop_log is None:
             drop_log = ((),) * len(events)
 
+    # as of v1.0, Epochs support Annotations which require raw_sfreq
+    # to be stored. However, for backwards compatibility if raw_sfreq
+    # is None (not found) in the file, then we will default raw_sfreq
+    # # to info['sfreq']
+    if raw_sfreq is None:
+        raw_sfreq = info['sfreq']
```
larsoner, line 3169: Do you actually need to do this here? Even if it's `None`, the `BaseEpochs` should be able to use `raw_sfreq=None` as this alias already.

mne/epochs.py
```diff
@@ -3446,16 +3309,21 @@ def __init__(self, fname, proj=True, preload=True,
...
+            metadata=metadata, verbose=verbose, raw_sfreq=raw_sfreq,
+            **reject_params)
         self.baseline = baseline
         self._do_baseline = False
         # use the private property instead of drop_bad so that epochs
         # are not all read from disk for preload=False
         self._bad_dropped = True
 
+        # set annotations
+        self.set_annotations(annotations)
```
larsoner, line 3325: Maybe we should have `BaseEpochs.__init__(..., annotations=None)`. That would work well with my comment above that setting `self._annotations = None` should be handled by the base class.

mne/epochs.py
adam2392: Good point. Moved all these into BaseEpochs.

mne/epochs.py
adam2392: Removed.

mne/epochs.py
adam2392: Agreed. Moved setters into BaseEpochs.

mne/io/tests/test_constants.py
larsoner: It looks like you did not update the commit hash above in this file, I expect `pytest mne/io/tests/test_constants.py` to fail

adam2392: Oh sorry I misunderstood what you were saying. Updating now.

larsoner: Sounds good -- please do test that line (plus other epochs-annotatinos-relevant tests) locally before pushing... it's a lot faster and lower CI load than waiting for CIs to complain.

mne/tests/test_epochs.py
```diff
@@ -3813,84 +3841,122 @@ def test_epoch_annotations_cases():
...
+        concatenate_epochs([epochs, old_epochs])
+
+    # concatenation should not change the *input* Epochs' annotations
+    assert epochs.annotations == old_epochs.annotations
+
+
+def test_epochs_annotations_backwards_compat():
+    """Test backwards compatibility with Epochs saved without annotations."""
+    # loading an earlier saved file should work
+    fname = op.join(misc.data_path(), 'audio', 'audio-epo.fif')
```
larsoner, line 3910: This is bad because it will silently download the `misc` dataset for people. We could fix this and have a decorator, but really we don't really want to use the `misc` ataset in testing unless we absolutely must.

Instead of doing this, I would use `monkeypatch` like:
```

def no_sfreq_write_float(a, b, c):
    if b == FIFF.FIFFV_MNE_EPOCHS_RAW_SFREQ:
        return
    return write_tag(a, b, c)

monkeypatch.setattr(mne.epochs, 'write_float', no_sfreq_write_float)
```
Then create and save Epochs. @adam2392 feel free to give it a try, or if it's not clear what I mean I can quickly push a commit to do it

adam2392: @larsoner is this fix what you meant?

larsoner (approved):
@agramfort feel free to review and merge if happy!

mne/annotations.py
```diff
@@ -647,6 +647,189 @@ def rename(self, mapping, verbose=None):
...
+        self : instance of Epochs
+            The epochs object with annotations.
+
+        Notes
+        -----
+        Annotation onsets and offsets are stored as time in seconds (not as
+        sample numbers) so care must be taken when doing things like decimating
+        or resampling. If the user has an Epochs object that did not support
+        Annotations before, and the Epochs were resampled, then the Annotations
+        will not show up in the correct locations.
```
agramfort, line 683: I don't get this. As things are stored in seconds and not samples it becomes robust to resampling. I think this deserves some clarifications. I get that if raw sfreq is lost then it creates a problem but as is I find it confusing.

agramfort (approved):
just a nitpick.

otherwise LGTM

disclaimer I did not test locally and I did not time the running time of the new tests

thx @

mne/annotations.py
adam2392: How's this updated doc string instead?

drammock: so by "an Epochs object that did not support annotations before" what we really mean is "an Epochs object saved before Annotation support was introduced" right?  (because it won't have an entry for `_raw_sfreq`)?  And by "will not show up in the correct locations" we mean "the Annotations onsets and durations will be shifted and stretched by the resampling/decimation ratio"?  Are those accurate rephrasings?

adam2392: Yep!

drammock: ok, but... a quick test shows no problem, as @agramfort seemed to suspect:

```py
import mne

path = mne.datasets.kiloword.data_path() + '/kword_metadata-epo.fif'
epo = mne.read_epochs(path)

assert epo._decim == 1
epo.resample(epo.info['sfreq'] * 2)  # upsample, works with 0.5 too

ann = mne.Annotations(onset=[3,5], duration=[0.5, 0.6], description=['one', 'two'])
epo.set_annotations(ann)
for _ann in epo.annotations:
    print(_ann)
```

Can you provide an example @adam2392 where there *is* a problem with epochs saved prior to this PR?

Also along the way I noticed that `epo.set_annotations()` is part of the public API now, but its docstring says it "just copies annotations from Raw", so that docstring needs to be updated.

adam2392: Hmm.. okay will take a look. Perhaps along the way, we solved all the issues we initially foresaw?

drammock: converting to Draft mode until we get this sorted out (with green CIs and 3 approvals I was worried someone might instinctively merge without looking too closely)

adam2392:
```
path = mne.datasets.kiloword.data_path() + '/kword_metadata-epo.fif'
    epo = mne.read_epochs(path)
    assert epo._decim == 1
    ann = mne.Annotations(onset=[3,5], duration=[0.5, 0.6], description=['one', 'two'])
    epo.set_annotations(ann)
    epo.add_annotations_to_metadata()
    metadata = epo.metadata.copy()

    epo.resample(epo.info['sfreq'] * 1.5)  # upsample, works with 0.5 too
    epo.add_annotations_to_metadata(overwrite=True)
    new_metadata = epo.metadata  

    from pandas.testing import assert_frame_equal
    assert_frame_equal(metadata, new_metadata)
```

Okay this doesn't work because `events` are not altered when we do "resample". So therefore, `raw_sfreq` needs to be correctly set in order for this function to work.

adam2392: So this is one of the reasons why I had `set_annotations` as private now IIRC. It silently can introduce an error for downstream functions that leverage the Annotations. Currently, the only function we have is `add_annotations_to_metadata`. However, in the future we might have more.  

Now that I think about it though, I don't think it's necessarily an issue with `set_annotations` because that just puts Annotations into the `Epochs` class. It has to do with the fact that you can't really do anything useful w/ the Annotations if the Epochs object was resampled and you want to apply the original Annotations back to it. If you add to metadata, the Annotations placement are completely incorrect now.

To me this seems like an edge case that just needs to be documented.

larsoner: > Can you provide an example @adam2392 where there is a problem with epochs saved prior to this PR?

I think an Epochs constructed before this PR with `decim > 1`, saved, and loaded on the branch from this PR would show the problem, as `epochs.info['sfreq']` will not correspond to the sample rate of its `epochs.events`. I don't think anyone has tried this case in the above at least. I can try tomorrow unless someone beats me to it.

adam2392: > > Can you provide an example @adam2392 where there is a problem with epochs saved prior to this PR?
> 
> I think an Epochs constructed before this PR with `decim > 1`, saved, and loaded on the branch from this PR would show the problem, as `epochs.info['sfreq']` will not correspond to the sample rate of its `epochs.events`. I don't think anyone has tried this case in the above at least. I can try tomorrow unless someone beats me to it.

I added some additional clarification in the doc strings related to this issue. I can't find any test dataset that has the `decim > 1` property, so not sure how we could test this.

I also added a test case for `metadata` for Epochs resampled now with `_raw_sfreq` support that should match before/after resampling.

drammock: > I can't find any test dataset that has the `decim > 1` property, so not sure how we could test this.

You could create one while on the `main` branch, save it, and add it to the testing data?

larsoner: > You could create one while on the main branch, save it, and add it to the testing data?

Indeed that's what I was suggesting for self-testing purposes, but we already have a no-data-added way to do it with `monkeypatch` in the tests. I'll push something shortly that I think should help, namely, 1) expose the problem, 2) add a warning when it might be a problem. (2) is enabled because it should only possibly be an issue when working with `EpochsFIF`, so when people `set_annotations` on those we can warn them that they should regenerate them if they originally used decimation or resampling. This alongside adding a `verbose` kwarg (so that people can `verbose='error'` if they know it's not an issue) I think makes us safe enough.

larsoner: Okay @adam2392 @drammock @agramfort feel free to look at 05cbbbd -- the test should make clear both the corner case and the new behavior I added to warn people about it.

adam2392: Ah thanks @larsoner

mne/annotations.py
```diff
@@ -647,6 +647,216 @@ def rename(self, mapping, verbose=None):
...
+            The epochs object with annotations.
+
+        Notes
+        -----
+        Annotation onsets and offsets are stored as time in seconds (not as
+        sample numbers). If the user has an Epochs object that did not support
+        Annotations before, and the Epochs were resampled, then the Annotations
+        may not show up in the correct locations. So care must be taken when
+        doing things like decimating or resampling with Epochs objects saved
+        before Annotations were supported in Epochs.
```
larsoner, line 685: > I think we can still keep the doc string warnings and such tho right?

Yes we should keep them but adjust them.
```suggestion
        Annotation onsets and offsets are stored as time in seconds (not as
        sample numbers).
        
        If you have an ``-epo.fif`` file saved to disk created before 1.0,
        annotations can be added correctly only if no decimation or
        resampling was performed. We thus suggest to regenerate your
        :class:`mne.Epochs` from raw and re-save to disk with 1.0+ if you
        want to safely work with Annotations in Epochs.
```

mne/annotations.py
```diff
@@ -647,6 +647,216 @@ def rename(self, mapping, verbose=None):
...
+
+        Since this function does not handle offsetting the times based
+        on first_sampe or measurement dates, the recommended way to add
```
larsoner, line 688:
```suggestion
        on first_samp or measurement dates, the recommended way to add
```

mne/annotations.py
```diff
@@ -677,18 +678,34 @@ def set_annotations(self, annotations, on_missing='raise'):
...
-        before Annotations were supported in Epochs.
+        sample numbers).
+
+        If you have an ``-epo.fif`` file saved to disk created before 1.0,
+        annotations can be added correctly only if no decimation or
+        resampling was performed. We thus suggest to regenerate your
+        :class:`mne.Epochs` from raw and re-save to disk with 1.0+ if you
+        want to safely work with :class:`~mne.Annotations` in epochs.
+
+        Since this function does not handle offsetting the times based
```
drammock, line 689:
```suggestion
        Since this method does not handle offsetting the times based
```

mne/annotations.py
```diff
@@ -655,19 +655,20 @@ def annotations(self):  # noqa: D102
...
-    def set_annotations(self, annotations, on_missing='raise'):
+    def set_annotations(self, annotations, on_missing='raise', *,
+                        verbose=None):
         """Setter for Epoch annotations from Raw.
 
-        This function simply copies over Raw annotations, and
-        does not handle offsetting the times based on first_samp
-        or measurement dates, since that is expected to occur in
-        Raw.set_annotations().
+        This function does not handle offsetting the times based
```
drammock, line 662:
```suggestion
        This method does not handle offsetting the times based
```

mne/annotations.py
```diff
@@ -783,6 +800,17 @@ def add_annotations_to_metadata(self, overwrite=False):
...
             The modified instance (instance is also modified inplace).
+
+        Notes
+        -----
+        Annotation onsets and offsets are stored as time in seconds (not as
+        sample numbers). If the user has an Epochs object that did not support
+        Annotations before, and the Epochs were resampled, then the Annotations
+        will not show up in the correct locations. So care must be taken when
+        using this function with an Epochs object saved before Annotation
+        support was introduced.
```
drammock, line 811: do we still need this note?  Seems like we warn pretty loudly now (in console and in docstrings) at `epochs.set_annotations()` stage,  so maybe here at the `epochs.add_annotations_to_metadata()` stage we can assume they've been done correctly?

drammock (approved):
couple docstring suggestions, new tests LGTM

larsoner:
Thanks @adam2392 !

PR #10021, opened 2021-11-20
[MRG, MAINT] Fix constant

alexrockhill:
In using this code on my own data, I realized that 100 voxels is a very small amount in a 256 x 256 x 256 image and you are likely to want at least a few more for context when using `mne.get_montage_volume_labels`. I guess this could be changed to a parameter but I think it's better not to have to deal with. If it were done away with altogether though it could cause runtime issues.

### DIFF
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1925,7 +1925,7 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
     return montage_warped, image_from, image_to
 
 
-_VOXELS_MAX = 1024  # define constant to avoid runtime issues
+_VOXELS_MAX = 3 ** 8  # define constant to avoid runtime issues
 
 
 @fill_doc
--- a/mne/surface.py
+++ b/mne/surface.py
@@ -1958,6 +1958,9 @@ def get_montage_volume_labels(montage, subject, subjects_dir=None,
     _validate_type(montage, DigMontage, 'montage')
     _validate_type(dist, (int, float), 'dist')
 
+    if dist < 0 or dist > 10:
+        raise ValueError('`dist` must be between 0 and 10')
+
     aseg, aseg_data = _get_aseg(aseg, subject, subjects_dir)
 
     # read freesurfer lookup table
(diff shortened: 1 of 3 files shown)


mne/surface.py
```diff
@@ -1925,7 +1925,7 @@ def warp_montage_volume(montage, base_image, reg_affine, sdr_morph,
     return montage_warped, image_from, image_to
 
 
-_VOXELS_MAX = 100  # define constant to avoid runtime issues
+_VOXELS_MAX = 3 ** 8  # define constant to avoid runtime issues
```
larsoner, line 1928: I don't understand the logic of 3**8. I would expect maybe something cubed that could be understood as N cubic mm. Would it make sense to use something cubed instead? If not, could you add a comment explaining this choice for the next person who needs to work with this code?

alexrockhill: You're right that doesn't make sense, I was reversing the exponent. Maybe let's just go back to 1024 or 10 ** 3 = 1000

agramfort (commented):
Api change ? Should we backport?

alexrockhill:
> Api change ? Should we backport?

No API change and I think backporting would be a good idea.

I just added a value error so that you don't get an unexpected result with a very large distance (if the distance is too large, there will be recursion/runtime issues).

agramfort:
@alexrockhill can you see why CIs fail?

alexrockhill:
Looks like caching issues unrelated https://dev.azure.com/mne-tools/mne-python/_build/results?buildId=16770&view=logs&j=d2de176e-a8da-5d2d-82f1-e561f5072d99&t=949a5866-58c4-51d6-4de3-a03ab2c3cb49

agramfort:
can you push to see if it fixes the issue?

A

>

alexrockhill:
> can you push to see if it fixes the issue? A
> […](#)

I just pushed yesterday and it was the same issue, but I'll push an empty commit.

larsoner:
> No API change and I think backporting would be a good idea.

Can you update `latest.inc` then we can merge and backport? Thanks @alexrockhill

larsoner (approved):

agramfort (approved):

PR #10024, opened 2021-11-20
[MRG, ENH] Add `fill_holes` argument to `mne.viz.Brain.add_volume_labels`

### LINKED ISSUE #10023: [BUG] Swiss cheese structures in `mne.viz.Brain.add_volume_labels`
I'm getting structures that have a few holes in them when using the volume labels in `Brain`...
example

I wonder if it wouldn't be too hard to implement a binary closing or something of the type but it's in VTK so I'm not really sure what's available there to close structures but I'd assume they would have something like that.

alexrockhill:
Fixes #10023.

I'm a bit stuck, it doesn't look like anything is happening here. Anyone know how to fix this?

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -34,6 +34,8 @@ Enhancements
 
 - Add support for reading haemoglobin fNIRS data to :func:`mne.io.read_raw_snirf` (:gh:`9929` by `Robert Luke`_)
 
+- Add `fill_hole_size` keyword argument to :func:`mne.viz.Brain.add_volume_labels` to close holes in the mesh (:gh:`10024` by `Alex Rockhill`_)
+
 Bugs
 ~~~~
 - Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSER_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -34,7 +34,7 @@ Enhancements
 
 - Add support for reading haemoglobin fNIRS data to :func:`mne.io.read_raw_snirf` (:gh:`9929` by `Robert Luke`_)
 
-- Add `fill_hole_size` keyword argument to :func:`mne.viz.Brain.add_volume_labels` to close holes in the mesh (:gh:`10024` by `Alex Rockhill`_)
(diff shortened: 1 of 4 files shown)


larsoner:
> Anyone know how to fix this?

No, I would look up VTK tutorials and play around with commands directly until I got something that worked. I would also start simple, for example by making a 5x5x5 (or probably bigger, maybe 11x11x11 just have a nice big border?) cube of zeros, the center 3x3x3 of which is all ones except for the centermost voxel (1x1x1) which is zero, and 1) see that the first / naive version *does* have a hole, then 2) figure out what needs to be added to get the whole 3x3x3 filled in *without any hole in the middle*. It's much easier to work from data like this to understand what it's doing.

I would also look at `binary_dilation` and other morphology tools and examples here

https://scikit-image.org/docs/dev/api/skimage.morphology.html

And also related stackoverflow posts that talk about fixing broken meshes and/or volume-to-mesh conversions

alexrockhill:
Fair enough, I'll give it a look in a bit, thanks for checking in @larsoner, I thought you might have seen something obvious that could fix it since you wrote the other code that uses VTK but I'll just look what others have done and see if I can find a way that works.

larsoner:
I looked but didn't see anything obvious. You might also want to try to start from some VTK example, sometimes there is some little option you think should be the default but isn't that you need to enable, and you only realize you need it once you see them using it...

alexrockhill:
Sounds good, will do

alexrockhill:
Ok here's what I have:

Original

Screen Shot 2021-11-23 at 11 12 26 AM

One binary dilation:

Screen Shot 2021-11-23 at 11 13 08 AM

Note that if you add the erosion, you get back where you started

Screen Shot 2021-11-23 at 11 14 44 AM

I haven't quite been able to figure that one out, why the erosion doesn't work, but since the results are so good with only one iteration of binary dilation, I think it's acceptable. It uses a mask as well so it doesn't go into any other ROIs.

larsoner:
This is 100% for viz currently, so even if this isn't perfect but makes a useful improvement I think we can merge and come back to it later if it's not 100% correct.

I think the only remaining question I have is: is there a big performance penalty here? If so, we should look hard for a VTK alternative to see if it's faster, as their code is often highly optimized (I think!), and could be faster than `ndimage`

alexrockhill:
Seemed pretty quick, but I can try the vtk implementation later. The only thing is that it has to be converted into a binary image for each level which sounds like a pain in the butt in vtk.

larsoner:
Thanks @alexrockhill !

PR #10030, opened 2021-11-23
minor: specify nitime is a package

guiomar:
Hi! This is a very minor PR to specify nitime is a package.

Before it was: 

> (as in nitime)

Now:

> (as in nitime **package**)

I saw you don't often include hyperlinks in the documentation, so I didn't add http://nipy.org/nitime/ webpage.
Please ignore this PR if you think this is something already known that doesn't require specification. Sorry I'm new to python, and I'm leaning about all these different packages.

PR #10031, opened 2021-11-23
add nitime to intersphinx; deduplicate some docstrings

drammock:
follow-up to #10030, I realized too late that all those docstrings were identical.

@guiomar FYI, this PR does what I would have suggested you do in #10030 if I had thought of it before merging:

1. make `nitime` into a link that goes to their documentation; this is possible because both packages use `sphinx` to build their docs so we can use intersphinx to cross-link to their module
2. since all 5 instances of the `normalization` docstring entry were identical, I've abstracted them out into our `docdict` so that in future we only need to change it in one place.

### DIFF
--- a/mne/utils/docs.py
+++ b/mne/utils/docs.py
@@ -1247,7 +1247,7 @@ weight_norm : str | None
         The Neural Activity Index :footcite:`VanVeenEtAl1997` will be computed,
         which simply scales all values from ``'unit-noise-gain'`` by a fixed
         value.
-    - ``'unit-noise-gain-invariante'``
+    - ``'unit-noise-gain-invariant'``
         Compute a rotation-invariant normalization using the matrix square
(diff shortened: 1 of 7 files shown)


mne/utils/docs.py
```diff
@@ -1247,7 +1247,7 @@
         The Neural Activity Index :footcite:`VanVeenEtAl1997` will be computed,
         which simply scales all values from ``'unit-noise-gain'`` by a fixed
         value.
-    - ``'unit-noise-gain-invariante'``
+    - ``'unit-noise-gain-invariant'``
```
drammock, line 1250: this is just a random typo that I noticed while making the other fixes.

guiomar:
Oh! That's very smart, I didn't know that this thing existed (utils/docs.py), it seems to help a lot in homogenizing things!

larsoner:
Thanks @drammock

PR #10036, opened 2021-11-24
Revert change to fNIRS tutorial

rob-luke:
In #9843 I initially changed the default value of the beer lambert law, but then we had to leave it the same for a depreciation cycle. Somehow this change snuck through, this causes all the plots to have no data in the tutorial as reported at https://mne.discourse.group/t/fnirs-tutorial-does-not-display-visualization

in the coming weeks I will complete the transition to the new default. Can we back port this PR?

### DIFF
--- a/tutorials/preprocessing/70_fnirs_processing.py
+++ b/tutorials/preprocessing/70_fnirs_processing.py
@@ -135,7 +135,7 @@ raw_od.info['bads'] = list(compress(raw_od.ch_names, sci < 0.5))
 # Next we convert the optical density data to haemoglobin concentration using
 # the modified Beer-Lambert law.
 
-raw_haemo = mne.preprocessing.nirs.beer_lambert_law(raw_od, ppf=6)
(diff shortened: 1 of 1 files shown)


drammock:
tutorial looks reasonable except for the first figure, which is just a big black square
https://38736-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/preprocessing/70_fnirs_processing.html?highlight=functional%20infrared

rob-luke:
Thanks for posting the link and checking the output @drammock  The figures now look as they should. I think we can merge if youre happy @drammock @larsoner @agramfort 

The black figure occurs before the code change in this PR, so i think is unrelated and we can merge.
The black figure should be a 3d image, so I guess something has changed with the 3d plotting dependencies or MNE code? I have seen a similar thing when trying to use the github actions CI, but not with circle.

agramfort:
backported to 0.24 maint branch

rob-luke:
Thank you

PR #10038, opened 2021-11-25
add refmeg_noise and fnirs_motor to __all__

mscheltienne:
Is there any reason why only those 2 datasets `refmeg_noise` and `fnirs_motor` were left out of `__all__`?


agramfort (approved):
good catch !

larsoner:
Thanks @mscheltienne !

PR #10039, opened 2021-11-25
MRG: Deprecrate pick channels in info

mscheltienne:
Follow up on the discussion in #10035

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -68,4 +68,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- Nothing yet
+- :meth:`mne.Info.pick_channels` has been deprecated. Use ``inst.pick_channels`` to pick channels from Raw|Epochs|Evoked. Use :func:`mne.pick_info` to pick channels from :class:`mne.Info` (:gh:`10039` by `Mathieu Scheltienne`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -68,4 +68,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- :meth:`mne.Info.pick_channels` has been deprecated. Use ``inst.pick_channels`` to pick channels from Raw|Epochs|Evoked. Use :func:`mne.pick_info` to pick channels from :class:`mne.Info` (:gh:`10039` by `Mathieu Scheltienne`_)
+- ``mne.Info.pick_channels`` has been deprecated. Use ``inst.pick_channels`` to pick channels from Raw|Epochs|Evoked. Use :func:`mne.pick_info` to pick channels from :class:`mne.Info` (:gh:`10039` by `Mathieu Scheltienne`_)
--- a/mne/io/meas_info.py
+++ b/mne/io/meas_info.py
@@ -36,7 +36,7 @@ from .proc_history import _read_proc_history, _write_proc_history
 from ..transforms import (invert_transform, Transform, _coord_frame_name,
                           _ensure_trans)
 from ..utils import (logger, verbose, warn, object_diff, _validate_type,
-                     _stamp_to_dt, _dt_to_stamp, _pl, _is_numeric,
+                     _stamp_to_dt, _dt_to_stamp, _pl, _is_numeric, deprecated,
                      _check_option, _on_missing, _check_on_missing, fill_doc)
 from ._digitization import (_format_dig_points, _dig_kind_proper, DigPoint,
                             _dig_kind_rev, _dig_kind_ints, _read_dig_fif)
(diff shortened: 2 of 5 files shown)


mscheltienne:
To give some context, I wanted to see if `equalize_channels` had to operate on `Info` as well. It breaks a lot of stuff if it doesn't; so I ended up adding the selection of channels for `Info` directly in `equalize_channels` instead of using `Info.pick_channels`.

mscheltienne:
Doc build failure doesn't seem to be related to this PR.

doc/changes/latest.inc
```diff
@@ -68,4 +68,4 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-- Nothing yet
+- :meth:`mne.Info.pick_channels` has been deprecated. Use ``inst.pick_channels`` to pick channels from Raw|Epochs|Evoked. Use :func:`mne.pick_info` to pick channels from :class:`mne.Info` (:gh:`10039` by `Mathieu Scheltienne`_)
```
larsoner, line 71: This saves us some work later for when we actually remove it (the link will be dead at that point). Plus for most deprecated functions we remove them from the API page, so for consistency it's not bad to avoid linking to the docs.
```suggestion
- ``mne.Info.pick_channels`` has been deprecated. Use ``inst.pick_channels`` to pick channels from Raw|Epochs|Evoked. Use :func:`mne.pick_info` to pick channels from :class:`mne.Info` (:gh:`10039` by `Mathieu Scheltienne`_)
```

mne/io/meas_info.py
```diff
@@ -1029,6 +1029,11 @@ def pick_channels(self, ch_names, ordered=False):
 
         .. versionadded:: 0.20.0
         """
+        warn('This method is deprecated and will be removed after 1.1. To '
+             'pick channels from an instance, use '
+             'inst.pick_channels(ch_names). To pick channels from an Info, '
+             'use mne.pick_info(info, sel).', DeprecationWarning)
```
larsoner, line 1035: Instead of doing this, use the `@deprecated` decorator like:
```
@deprecated('info.pick_channels is deprecated, use inst.pick_channels instead.')
def pick_channels(...):
```
I wouldn't mention pick_info in this runtime-emitted DeprecationWarning because it's rare/expert mode, and the risk of a more basic user seeing `pick_info` as an option and just dropping that in is too high, I think. We can keep the `pick_info` note in the `latest.inc`, though, as a guide for the experts.

mne/io/tests/test_meas_info.py
```diff
@@ -1059,3 +1058,12 @@ def test_info_bad():
...
     with pytest.raises(ValueError, match='between meg<->head'):
         info['dev_head_t'] = Transform('mri', 'head', np.eye(4))
+
+
+def test_info_pick_channels():
+    """Test that info.pick_channels emits a deprecation warning."""
+    info = create_info(2, 1000., 'eeg')
+    with pytest.warns(DeprecationWarning,
+                      match='This method is deprecated and will be removed '
+                            'after 1.1. To pick channels from'):
```
larsoner, line 1068: Shorter as:

with pytest.deprecated_call(match=...):

larsoner (commented):
Just some minor stuff, thanks for taking care of this @mscheltienne !

mscheltienne:
Thanks for the tips, let's see if CIs are green; I wasn't quite sure on how the deprecated decorator works (especially as it doesn't throw any warning at me in the IPython console).

mne/io/meas_info.py
```diff
@@ -1007,6 +1007,7 @@ def _update_redundant(self):
             self['ch_names'] = [ch['ch_name'] for ch in self['chs']]
             self['nchan'] = len(self['chs'])
 
+    @deprecated('use inst.pick_channels instead.')
```
larsoner, line 1010: This should just automagically work. If it doesn't, I'll push a commit to fix the decorator (it's possible it has some bug)

mscheltienne: CIs are happy, so all good I guess, but I don't get any warning issued when using a function decorated with `@deprecated` in python console or in Spyder's IPython console. The docstring is edited however with `.. warning:: DEPRECATED:`. Is this the correct behavior for this decorator?

larsoner: I cannot replicate; on your branch I get:
```
>>> import mne
>>> import warnings
>>> warnings.simplefilter('always', DeprecationWarning)
>>> info = mne.create_info(2, 1000., 'eeg')
>>> info.pick_channels(info['ch_names'][:1])
<decorator-gen-25>:4: DeprecationWarning: Function pick_channels is deprecated; use inst.pick_channels instead.
```

larsoner:
Thanks @mscheltienne !

mne/io/meas_info.py
mscheltienne: Getting the warning as well on this mac laptop.. I guess the problem is with my windows computer and the venv.

PR #10040, opened 2021-11-25
Fix annotation cropping and I/O roundtrip

### LINKED ISSUE #9527: Annotation and crop - different onset points after reloading the saved croppend data
#### Describe the bug
I suspect that this issue is similar to the #9383.
The issue only occurs when a raw instance created with Annotation. Originally labeled data as brain vision files are work fine.
Cropping the instance, saving and reloading it shows the annotations in the original place before the cropping.

#### Steps to reproduce
```
import numpy as np
import mne
CH_NUM = 32
FS = 160

data = np.zeros((CH_NUM, FS * 20))
ch_names = [f'EEG{i}' for i in range(CH_NUM)]
ch_types = ['eeg'] * len(ch_names)
onset = [8, 12, 18]
duration = [2] * len(onset)
description = list(range(len(onset)))

info = mne.create_info(ch_names, ch_types=ch_types, sfreq=FS)
annotation = mne.Annotations(onset, duration, description)
raw = raw.set_annotations(annotation)

sess = raw.copy()
sess.crop(5, 18)
file = 'test_file_raw.fif'
sess.save(file, overwrite=True)
sess.plot(block=False)
r = mne.io.read_raw(str(file))
r.plot(block=False)
raw.plot(block=True)
```

#### Expected results
I expect that the annotated data should look like the same before and after saving it.

#### Actual results
The annotations are shifted...

#### Additional information
Platform:      Windows-10-10.0.19041-SP0
Python:        3.8.5 (default, Sep  3 2020, 21:29:08) [MSC v.1916 64 bit (AMD64)]
Executable:    C:\Programs\Miniconda3\envs\bci2\python.exe
CPU:           Intel64 Family 6 Model 158 Stepping 10, GenuineIntel: 12 cores
mne:           0.23.0
numpy:         1.19.3 {blas=D:\\a\\1\\s\\numpy\\build\\openblas_info, lapack=D:\\a\\1\\s\\numpy\\build\\openblas_lapack_info}
scipy:         1.6.0
matplotlib:    3.3.3 {backend=TkAgg}
sklearn:       0.24.0
numba:         Not found
nibabel:       Not found
nilearn:       Not found
dipy:          Not found
cupy:          Not found
pandas:        1.2.0
mayavi:        Not found
pyvista:       Not found
vtk:           Not found

mscheltienne:
PR to fix #9527 

WIP:
- [x] Test to expose failure
- [x] Fix


agramfort:
@mscheltienne I think I fixed the issue

mscheltienne:
Looks like it, awesome, thanks!

mscheltienne:
Any reason why you simplified the test by removing the 4th annotation?
I can't check on a computer/test until tomorrow evening.

agramfort:
it's because it was going beyond the duration of the signal and the
duration of the
last annotation was reduced by cropping hence was breaking the test.

>

agramfort:
this one is good to go

@larsoner @mscheltienne @drammock

mscheltienne:
OK, so when raw was cropped between t1, 18s; the 4th annotation starting at 18s was cropped and now lasting 1 sample (from memory, might be entirely wrong)? Isn't that a bug in itself? 

Also, I wonder if it's not better to correct the annotation onset at saving instead of loading, I don't know if there are third parties program that load - raw.fif files that might be affected.

agramfort:
>
> OK, so when raw was cropped between t1, 18s; the 4th annotation starting
> at 18s was cropped and now lasting 1 sample (from memory, might be entirely
> wrong)?
>
yes

> Isn't that a bug in itself?

when the idea is that annotations should be limited to the data you have
access to.

> Also, I wonder if it's not better to correct the annotation onset at
> saving instead of loading, I don't know if there are third parties program
> that load - raw.fif files that might be affected.
>
this is to me more dangerous and not backward compatible. Now we fix
reading back files that have been written correctly up to now

larsoner:
Will merge this after #10118 is in (and a possible merge with upstream here), thanks in advance @mscheltienne @agramfort 
!

PR #10041, opened 2021-11-25
decim is actually used for epochs

agramfort:
see https://github.com/mne-tools/mne-python/blob/main/mne/preprocessing/ica.py#L720


hoechenberger:
Thanks @agramfort!

This should be backported too. I'm off to bed, would be happy if sb else could take over :)

hoechenberger:
Argh we need to adjust this here too:
https://github.com/mne-tools/mne-python/blob/34ca144db4adab33d576bb60db1d7e8661b1082f/mne/preprocessing/ica.py#L617-L628

I will take care of this an the backport tomorrow.

agramfort:
indeed

can you send a PR?

PR #10044, opened 2021-11-26
Limit output of __repr__ of mne.BaseEpochs and mne.Evoked to improve performance

### LINKED ISSUE #10042: Improve MNE performance in IDEs / Limit object string representations

I have the unfortunate case of having MNE epochs that have a vast number of unique events (> 4000). While this may be a fault of my data not using the metadata property, this still leads to bad performance of MNE in, e.g., PyCharm.
PyCharm has a "Variable Explorer" window which (as I suspect) makes use of the string representation of current objects in memory and updates after each update, e.g., when executing a line in the interactive python console. Unfortunately, in the case of BaseEpoch the string representation sorts and enumerates all events, leading to quite the large string representation when an epoch has many events.

https://github.com/mne-tools/mne-python/blob/0789b7dade297a7e5cae10a343a870ed260ab9a0/mne/epochs.py#L1664-L1667

Even without sophisticated IDEs, in my work flow I often execute code in a console interactively. E.g. to take a quick look at the epoch object i simply type `epochs` and press enter, as the default for python consoles is to print the string representation of the objects. Currently this takes around 20 seconds to print on my machine.

Replace the above lines with
```python
        events_limit = 10
        counts = ['%r: %i' % (k, sum(self.events[:, 2] == v))
                  for k, v in sorted(self.event_id.items())[:events_limit]]
        if len(self.event_id) > 0:
            s += ',' + '\n '.join([''] + counts)
        if len(self.event_id) > events_limit:
            not_shown_events = len(self.event_id) - events_limit + 1
            s += f"\n 'and {not_shown_events} more events ...'"

```

For my epoch object, the above change reduces `len(epochs.__repr__())` from 257655 to 717 chars. Example output:

```
>>> epochs
<Epochs |  4283 events (all good), -0.2 - 0.75 sec, baseline off, ~20.3 MB, data loaded,
 'Letter_1/Sequence_1/NonTarget/A/B/G/J/L/Q/R/T/W/Y/ /?': 1
 'Letter_1/Sequence_1/NonTarget/A/C/D/J/K/L/O/S/X/./!/<': 1
 'Letter_1/Sequence_1/NonTarget/A/C/E/G/J/K/L/P/W/Z/,/<': 1
 'Letter_1/Sequence_1/NonTarget/A/D/G/H/K/M/N/U/V/W/Y/?': 1
 'Letter_1/Sequence_1/NonTarget/A/D/H/M/O/Q/T/W/Y/Z/,/<': 1
 'Letter_1/Sequence_1/NonTarget/A/D/H/M/P/Q/T/W/Y/!/?/<': 1
 'Letter_1/Sequence_1/NonTarget/A/E/G/I/L/O/Q/S/U/Y/./?': 1
 'Letter_1/Sequence_1/NonTarget/A/E/G/J/L/N/P/Q/S/U/ /?': 1
 'Letter_1/Sequence_1/NonTarget/A/E/H/I/K/O/R/T/U/./,/?': 1
 'Letter_1/Sequence_1/NonTarget/B/C/D/G/L/N/P/W/X/Z/,/!': 1
 'and 4273 more events ...'>
```

One could also add an optional keyword to `__repr__`, then one could replicate the old behaviour by calling `epochs.__repr__(full=True)`. Or one could instruct people to not use so damn many events when they could use metadata :)

jsosulski:
#### Reference issue
Closes #10042 .

Reduce the size and load of `__repr__` output/calculations. This function is called frequently in debuggers, IDEs and interactive python consoles to describe the currently available variables / objects. When handling epochs with many events, the current implementation leads to sluggish IDE performance / printing an object to console takes very long. You can still obtain the old behaviour by explicitly calling `__repr__(limit_events=None)`.

I also changed evoked to not show comments longer than 1000 chars. Evoked does not do excessive computations in `__repr__`, as such, the performance is not impacted. But the resulting output is still extremely long. You can still obtain the old behaviour by explicitly calling `__repr__(limit_comment=None)`. Additionally the whole comment is always available in `evoked.comment`.

Note that old behaviour is only available when explicitly calling `__repr__`. Implicit calls by IDEs debuggers etc. will always use the new limited output.

#### Additional Information

I am not sure why the dict_keys are first `sorted` for `BaseEpochs` object representation but not when converting to a comment for use in `Evoked`:

https://github.com/mne-tools/mne-python/blob/34ca144db4adab33d576bb60db1d7e8661b1082f/mne/epochs.py#L1664-L1665

https://github.com/mne-tools/mne-python/blob/34ca144db4adab33d576bb60db1d7e8661b1082f/mne/epochs.py#L1115-L1117

### DIFF
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -36,6 +36,8 @@ Enhancements
 
 - Add ``fill_hole_size`` keyword argument to :func:`mne.viz.Brain.add_volume_labels` to close holes in the mesh (:gh:`10024` by `Alex Rockhill`_)
 
+- Changed `mne.Epochs` and `mne.Evoked` to have a more concise `__repr__` to improve interactive MNE usage in Python Interactive Console, IDEs, and debuggers when many events are handled. (:gh:`10042` by `Jan Sosulski`_)
+
 Bugs
 ~~~~
 - Fix default of :func:`mne.io.Raw.plot` to be ``use_opengl=None``, which will act like False unless ``MNE_BROWSER_USE_OPENGL=true`` is set in the user configuration (:gh:`9957` by `Eric Larson`_)
--- a/doc/changes/latest.inc
+++ b/doc/changes/latest.inc
@@ -36,7 +36,7 @@ Enhancements
 
 - Add ``fill_hole_size`` keyword argument to :func:`mne.viz.Brain.add_volume_labels` to close holes in the mesh (:gh:`10024` by `Alex Rockhill`_)
 
-- Changed `mne.Epochs` and `mne.Evoked` to have a more concise `__repr__` to improve interactive MNE usage in Python Interactive Console, IDEs, and debuggers when many events are handled. (:gh:`10042` by `Jan Sosulski`_)
+- Changed :class:`mne.Epochs` and :class:`mne.Evoked` to have a more concise ``__repr__`` to improve interactive MNE usage in Python Interactive Console, IDEs, and debuggers when many events are handled. (:gh:`10042` by `Jan Sosulski`_)
 
 Bugs
 ~~~~
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1643,7 +1643,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         """Last time point."""
         return self.times[-1]
 
-    def __repr__(self):
+    def __repr__(self, events_limit=10):
         """Build string representation."""
         s = ' %s events ' % len(self.events)
         s += '(all good)' if self._bad_dropped else '(good & bad)'
@@ -1661,10 +1661,14 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         s += ', ~%s' % (sizeof_fmt(self._size),)
         s += ', data%s loaded' % ('' if self.preload else ' not')
         s += ', with metadata' if self.metadata is not None else ''
+
         counts = ['%r: %i' % (k, sum(self.events[:, 2] == v))
-                  for k, v in sorted(self.event_id.items())]
+                  for k, v in sorted(self.event_id.items())[:events_limit]]
         if len(self.event_id) > 0:
             s += ',' + '\n '.join([''] + counts)
+        if len(self.event_id) > events_limit:
+            not_shown_events = len(self.event_id) - events_limit + 1
+            s += f"\n 'and {not_shown_events} more events ...'"
         class_name = self.__class__.__name__
         class_name = 'Epochs' if class_name == 'BaseEpochs' else class_name
         return '<%s | %s>' % (class_name, s)
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -1643,7 +1643,7 @@ class BaseEpochs(ProjMixin, ContainsMixin, UpdateChannelsMixin, ShiftTimeMixin,
         """Last time point."""
         return self.times[-1]
 
-    def __repr__(self, limit_events=10):
+    def __repr__(self):
         """Build string representation."""
         s = ' %s events ' % len(self.events)
         s += '(all good)' if self._bad_dropped else '(good & bad)'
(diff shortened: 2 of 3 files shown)


mne/epochs.py
```diff
@@ -1643,7 +1643,7 @@ def tmax(self):
         """Last time point."""
         return self.times[-1]
 
-    def __repr__(self):
+    def __repr__(self, limit_events=10):
```
agramfort, line 1646: you cannot have a param in a __repr__ AFAIK

mne/epochs.py
```diff
@@ -1643,7 +1643,7 @@ def tmax(self):
...
         """Build string representation."""
```
agramfort, line 1647: I would also call this max_events

mne/epochs.py
jsosulski: Keywords seem to work, but I dont know if thats intended or not.
Alternatives that come to mind:
1) simply set `max_events` inside the function, disabling customization of the output
or
2) Somewhere in BaseEpochs add an attribute `repr_verboseness` or something and infer `max_events` from there on.

agramfort:
>
>
>    1. simply set max_events inside the function, disabling customization
>    of the output
>
>
+1 for this

mne/evoked.py
```diff
@@ -301,7 +301,13 @@ def save(self, fname):
         write_evokeds(fname, self)
 
     def __repr__(self):  # noqa: D105
-        s = "'%s' (%s, N=%s)" % (self.comment, self.kind, self.nave)
+        max_comment_length = 1000
+        if len(self.comment) > max_comment_length:
+            comment = self.comment[:max_comment_length]
+            comment += "...rest hidden. See .comment for the full description"
```
agramfort, line 307: I would just do
```suggestion
            comment += "..."
```

agramfort:
can you show how it displays now? thx

jsosulski:
MWE

```python
import mne
import numpy as np

np.random.seed(10)
n_epochs, n_channels, n_timepoints = 100, 2, 21
fake_data = 1e-6*np.random.random((n_epochs, n_channels, n_timepoints))

e = np.arange(1, n_epochs+1)
events = np.vstack([e, np.zeros_like(e), e]).T

epo = mne.EpochsArray(fake_data, info, events=events, event_id={f"event_id_{i}":i for i in e})
print(epo)
print(epo.average(picks="all"))
```

Output of `print(epo)`:
```
<EpochsArray |  100 events (all good), 0 - 0.5 sec, baseline off, ~40 kB, data loaded,
 'event_id_1': 1
 'event_id_10': 1
 'event_id_100': 1
 'event_id_11': 1
 'event_id_12': 1
 'event_id_13': 1
 'event_id_14': 1
 'event_id_15': 1
 'event_id_16': 1
 'event_id_17': 1
 'and 90 more events ...'>
```

Output of `print(epo.average(picks="all"))`:
```
<Evoked | '0.01 × event_id_1 + 0.01 × event_id_2 + 0.01 × event_id_3 + 0.01 × event_id_4 + 0.01 × event_id_5 + 0.01 × event_id_6 + 0.01 × event_id_7 + 0.01 × event_id_8 + 0.01 × event_id_9 + 0.01 × event_id_10 + 0.01 × event_id_11 + 0.01 × event_id_12 + 0.01 × event_id_13 + 0.01 × event_id_14 + 0.01 × event_id_15 + 0.01 × event_id_16 + 0.01 × event_id_17 + 0.01 × event_id_18 + 0.01 × event_id_19 + 0.01 × event_id_20 + 0.01 × event_id_21 + 0.01 × event_id_22 + 0.01 × event_id_23 + 0.01 × event_id_24 + 0.01 × event_id_25 + 0.01 × event_id_26 + 0.01 × event_id_27 + 0.01 × event_id_28 + 0.01 × event_id_29 + 0.01 × event_id_30 + 0.01 × event_id_31 + 0.01 × event_id_32 + 0.01 × event_id_33 + 0.01 × event_id_34 + 0.01 × event_id_35 + 0.01 × event_id_36 + 0.01 × event_id_37 + 0.01 × event_id_38 + 0.01 × event_id_39 + 0.01 × event_id_40 + 0.01 × event_id_41 + 0.01 × event_id_42 + 0.01 × event_id_43 + 0.01 × event_id_44 + 0.01 × event_id_45 + 0.01 × event_id_46 + 0.01 × event_id_47 + 0.01 × event_id_48 + 0...' (average, N=100), 0 – 0.5 sec, baseline off, 2 ch, ~8 kB>
```

Here we can see the difference in the repr handling between epo and evo by using sorted vs not using sorted that I mentioned in my first post. Should I unify this? If yes, make both sorted or both unsorted? Personally I prefer unsorted, as this then is just the order of my event_id dict, but I have no strong opinion on this.

agramfort (approved):
thx @jsosulski

hoechenberger:
> Personally I prefer unsorted, as this then is just the order of my event_id dict,

Yes, this makes sense to me too.

doc/changes/latest.inc
```diff
@@ -36,6 +36,8 @@ Enhancements
 
 - Add ``fill_hole_size`` keyword argument to :func:`mne.viz.Brain.add_volume_labels` to close holes in the mesh (:gh:`10024` by `Alex Rockhill`_)
 
+- Changed `mne.Epochs` and `mne.Evoked` to have a more concise `__repr__` to improve interactive MNE usage in Python Interactive Console, IDEs, and debuggers when many events are handled. (:gh:`10042` by `Jan Sosulski`_)
```
larsoner, line 39:
```suggestion
- Changed :class:`mne.Epochs` and :class:`mne.Evoked` to have a more concise ``__repr__`` to improve interactive MNE usage in Python Interactive Console, IDEs, and debuggers when many events are handled. (:gh:`10042` by `Jan Sosulski`_)
```

larsoner (approved):
LGTM +1 for merge

I pushed a commit to make CircleCI happy, will merge once CIs come back happy. Thanks in advance @jsosulski !

hoechenberger (changes requested):
> ```
>  'event_id_17': 1
>  'and 90 more events ...'>
> ```

I think the last line shouldn't contain quotation marks

jsosulski:
> I think the last line shouldn't contain quotation marks

Good catch! I also removed `sorted`. I suspect this is a relic from before mne required python 3.7, where dictionary entries were not guaranteed to be ordered in the same way you created `event_id`. Now the event order is no longer different between `mne.Epochs` and `mne.Evoked`

hoechenberger (approved):

PR #10045, opened 2021-11-26
MRG: Improve concatenate_epochs() docstring

### LINKED ISSUE #10043: Inconsistency in behavior between concatenate_raws() and concatenate_epochs()
I just ran into an issue because I blindly assumed `concatenate_epochs()` would modify the first element in-place, just like `concatenate_raws()` does. Turns out, this is not the case.

The docstring of the `Returns` portion says:

> The result of the concatenation (first Epochs instance passed in).

The part in the parentheses is misleading.

## Proposed fix

Since we'll probably not find a consensus to change the API, I'd propose to change the docstring of `concatenate_epochs()`:

- The `Return` bit should drop the `(first Epochs instance passed in)` part.
- A `Note` should be added, clearly informing users of the behavior difference between `concatenate_epochs()` and `concatenate_raws()`

Thoughts?

hoechenberger:
Fixes #10043

### DIFF
--- a/mne/epochs.py
+++ b/mne/epochs.py
@@ -3514,12 +3514,15 @@ def _finish_concat(info, data, events, event_id, tmin, tmax, metadata,
 @verbose
 def concatenate_epochs(epochs_list, add_offset=True, *, on_mismatch='raise',
                        verbose=None):
-    """Concatenate a list of epochs into one epochs object.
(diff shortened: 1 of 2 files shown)


agramfort (approved):

agramfort:
thx @hoechenberger

PR #10047, opened 2021-11-28
Add back pooch to sys_info; pin pooch min version

hoechenberger:
It appears that #9886 accidentally removed `pooch` from the `sys_info` output.

I also specify a lower version requirement for `pooch` (1.5+) after a user [reported](https://mne.discourse.group/t/issue-in-loading-data-in-overview-of-meg-eeg-with-mne-python/4042/) an issue with an outdated version on the forum.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,5 +33,5 @@ tqdm
 mffpy>=0.5.7
 ipywidgets
 ipyvtklink
-pooch
+pooch >= 1.5
 mne-qt-browser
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -33,5 +33,5 @@ tqdm
 mffpy>=0.5.7
 ipywidgets
 ipyvtklink
-pooch
+pooch >= 1.5
```
hoechenberger, line 36:
```suggestion
pooch>=1.5
```

agramfort (approved):

larsoner:
Thanks @hoechenberger !

Issue #7977, opened 2020-07-08
[BUG] 3d points with opacity < 1 are invisible


bloyl:
On (some?) linux virtual machines that use mesa_gl rendering, you can't visualize 3d points with any level of opacity. This affects `mne coreg`, `plot_alignment` etc.

You can recreate the issue by running https://mne.tools/dev/auto_examples/visualization/plot_eeg_on_scalp.html#sphx-glr-auto-examples-visualization-plot-eeg-on-scalp-py.

the figure below shows 2 plots. On the right is the default plot. and on the left is after I manually adjust the opacity using mayavi settings.

<img width="1328" alt="image" src="https://user-images.githubusercontent.com/5585821/86913833-3825b980-c0ed-11ea-9180-a8f592a3d41d.png">

this is with `export MESA_GL_VERSION_OVERRIDE=3.3` and isn't affected by `MNE_3D_OPTION_ANTIALIAS`.

This is most likely a vtk bug with old versions of LLVM but it would be good to figure out a minimum version of LLVM.

From looking at #7976 it seems one of our test systems is close to my system. I wonder if we could evaluate this there?  

> pyvista:       0.25.3 {pyvistaqt=0.1.1, OpenGL 3.3 (Core Profile) Mesa 18.3.6 via llvmpipe (LLVM 7.0, 256 bits)}

bloyl:
I should note this is a problem for both mayavi and pyvista backends.

here is the output from the new sys_info:

```
Platform:      Linux-3.10.0-1127.el7.x86_64-x86_64-with-glibc2.10
Python:        3.8.3 (default, Jul  2 2020, 16:21:59)  [GCC 7.3.0]
Executable:    /mnt/isilon/meg_lab/Linux_software/anaconda3_lab/envs/mne-vtk-test/bin/python
CPU:           x86_64: 4 cores
Memory:        23.4 GB

mne:           0.21.dev0
numpy:         1.18.5 {blas=mkl_rt, lapack=mkl_rt}
scipy:         1.5.0
matplotlib:    3.2.2 {backend=Qt5Agg}

sklearn:       0.23.1
numba:         0.50.1
nibabel:       3.1.1
cupy:          Not found
pandas:        1.0.5
dipy:          1.1.1
mayavi:        4.7.2.dev0
[...]

PR #10052, opened 2021-11-29
MRG, MAINT: Fix output

larsoner:
Run our circle-front to check to see if we have the same problem as https://github.com/QB3/sparse-ho/pull/13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,6 @@ jobs:
               echo "export MNE_FULL_DATE=true" >> $BASH_ENV
               echo "export MNE_3D_OPTION_ANTIALIAS=false" >> $BASH_ENV
               source tools/get_minimal_commands.sh
-              echo "export MESA_GL_VERSION_OVERRIDE=3.3" >> $BASH_ENV
               echo "export MNE_3D_BACKEND=pyvistaqt" >> $BASH_ENV
               echo "export PATH=~/.local/bin/:$PATH" >> $BASH_ENV
               echo "source ~/python_env/bin/activate" >> $BASH_ENV
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,8 +15,8 @@ _xvfb: &xvfb
 
 jobs:
     build_docs:
-      machine:
-        image: ubuntu-2004:202111-01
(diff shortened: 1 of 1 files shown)


larsoner:
Indeed we have a problem with it being mostly black (the colorbar shows up but not the mesh):

https://38888-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/inverse/30_mne_dspm_loreta.html

larsoner:
... I'll probably look into this early next week

larsoner:
Interesting, we do *not* have the problem with the docker image:

https://38909-1301584-gh.circle-artifacts.com/0/dev/auto_tutorials/inverse/30_mne_dspm_loreta.html

But we *do* have a problem with the machine config, as above.

It would be nice to be able to use the docker image... in any case, it would be worth trying our `_BrainScraper` for QB3/sparse-ho#13. I can push a commit for that tomorrow US time unless someone beats me to it

larsoner:
It looks like the machine config, which doesn't work, has a newer Mesa:
```
0.32.dev0 {OpenGL 4.5 (Core Profile) Mesa 21.0.3 via llvmpipe (LLVM 12.0.0, 256 bits)}
```
the docker config, which does work, has an older one:
```
0.32.dev0 {OpenGL 3.3 (Core Profile) Mesa 18.3.6 via llvmpipe (LLVM 7.0, 256 bits)}
```
I have no idea why this would be the case, but it's probably the culprit. I'll try some MESA tweaks to see if I can fix it. It would be nice if we could use the machine config, as it's not only faster + more mem but the docker ones are deprecated and will be removed according to an email I got today from CircleCI.

larsoner:
Found the fix, and I even used our [own docs](https://mne.tools/0.22/generated/mne.viz.set_3d_options.html) (and our [old issues](https://github.com/mne-tools/mne-python/issues/7977#issue-653234935)!

Will merge once I make sure `MNE_3D_OPTIONANTIALIAS=false` was the correct fix and that `MESA_GL_VERSION_OVERRIDE=3.3` was not necessary

agramfort (approved):

drammock (approved):
nice detective work @larsoner

PR #10054, opened 2021-11-29
DOC: add bad_acq_skip annotation behaviour

### LINKED ISSUE #10037: BUG: Annotated BAD_ACQ_SKIP times are skipped/lost when loaded from fif
Hi guys, I think there is a bug with the behaviour of ``BAD_ACQ_SKIP``.

#### Describe the bug
``times`` and ``n_times`` parameters of ``mne.io.Raw`` seem to be inconsistent when loading the object that has ``BAD_ACQ_SKIP`` annotation(s) : the part that is annotated ``BAD_ACQ_SKIP`` is not included in ``times``.

#### Steps to reproduce
``` python
import mne
import numpy as np

# create a fake raw instance based on 5-sec sin and cos signals
info = mne.create_info(ch_names=["cos", "sin"], sfreq=10)
x = np.linspace(0, 5, 50)
raw = mne.io.RawArray(
    np.array([np.cos(x), np.sin(x)]),
    info
)

# add BAD_ACQ_SKIP annotation on the first 2 seconds
raw.annotations.append(onset=0, duration=2, description="BAD_ACQ_SKIP")
# save to fif
raw.save("raw_ieeg.fif")

# load the fif
raw_loaded = mne.io.read_raw_fif("raw_ieeg.fif")

# compare
print(raw.times)
print(raw_loaded.times)
```
compare output:
``` python
[0.  0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.  1.1 1.2 1.3 1.4 1.5 1.6 1.7
 1.8 1.9 2.  2.1 2.2 2.3 2.4 2.5 2.6 2.7 2.8 2.9 3.  3.1 3.2 3.3 3.4 3.5
 3.6 3.7 3.8 3.9 4.  4.1 4.2 4.3 4.4 4.5 4.6 4.7 4.8 4.9]
[0.  0.1 0.2 0.3 0.4 0.5 0.6 0.7 0.8 0.9 1.  1.1 1.2 1.3 1.4 1.5 1.6 1.7
 1.8 1.9 2.  2.1 2.2 2.3 2.4 2.5 2.6 2.7 2.8 2.9]
```

#### Expected results
I would except ``times`` to take also the ``BAD_ACQ_SKIP`` into account, so in my example ``raw_loaded.times`` should be the same as ``raw.times``.

What do you think ?

etiennedemontalivet:
#### Reference issue
Closes #10037.

The changes add documentation about the specific behaviour of ``BAD_ACQ_SIP`` annotation in reading/writing file:
- when saving to FIF file, samples annotated ``BAD_ACQ_SKIP`` are not stored and will be loaded as 0s when reading back
- the first N samples annotated ``BAD_ACQ_SKIP`` are skipped when reading FIF file

### DIFF
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -236,7 +236,6 @@ class Annotations(object):
     ``BAD_ACQ_SKIP`` annotation leads to specific reading/writing file
     behaviours. See :meth:`mne.io.read_raw_fif` and
     :meth:`Raw.save() <mne.io.Raw.save>` notes for details.
-
     """  # noqa: E501
 
     def __init__(self, onset, duration, description,
--- a/mne/annotations.py
+++ b/mne/annotations.py
@@ -230,6 +230,13 @@ class Annotations(object):
              n                        |      |
              e                        +------+
          orig_time                 onset[0]'
+
+    **Specific annotation**
+
+    ``BAD_ACQ_SKIP`` annotation leads to specific reading/writing file
+    behaviours. See :meth:`mne.io.read_raw_fif` and
+    :meth:`Raw.save() <mne.io.Raw.save>` notes for details.
+
     """  # noqa: E501
 
     def __init__(self, onset, duration, description,
--- a/mne/io/base.py
+++ b/mne/io/base.py
@@ -1416,6 +1416,9 @@ class BaseRaw(ProjMixin, ContainsMixin, UpdateChannelsMixin, SetChannelsMixin,
         work properly on a saved concatenated file (e.g., probably some
         or all forms of SSS). It is recommended not to concatenate and
         then save raw files for this reason.
+
+        Samples annotated ``BAD_ACQ_SKIP`` are not stored in order to optimize
+        memory. Whatever values, they will be loaded as 0s when reading file.
         """
         endings = ('raw.fif', 'raw_sss.fif', 'raw_tsss.fif',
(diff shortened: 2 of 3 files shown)


mne/annotations.py
```diff
@@ -230,6 +230,13 @@ class Annotations(object):
              n                        |      |
              e                        +------+
          orig_time                 onset[0]'
+
+    **Specific annotation**
```
agramfort, line 234:
```suggestion
    .. note::
```

agramfort: then you will need to check that it renders well. Otherwise a section `Notes`

etiennedemontalivet: It renders ok. I did that in the first place because the ``Notes`` section of ``mne.Annotations`` is quit large and has already 2 paragraphs (+ the intro) named **ch_names** and **orig_time**. So adding **Specific annotation** would keep the format, but if you think adding a note is better, I'm ok with it. Your call :)

agramfort: it appears in bold in the html sphinx rendered doc?

etiennedemontalivet: Yes, as sub-section titles of the section ``Notes``. Here: https://mne.tools/dev/generated/mne.Annotations.html

agramfort (approved):
LGTM !

thx @etiennedemontalivet

larsoner:
Thanks @etiennedemontalivet !

